PR #28, opened 2014-12-17
tests: Begin supporting running a subset of tests via "make check"

cgwalters:
While I think InstalledTests is cool and powerful, it requires a lot
of infrastructure, and is unfamiliar to many people.  Let's also
support "make check".

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -20,6 +20,7 @@
 
 if BUILDOPT_INSTALL_TESTS
 insttest_PROGRAMS = 
+TESTS =
 
 insttestdir=$(pkglibexecdir)/installed-tests
 testfiles = test-basic \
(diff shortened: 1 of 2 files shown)


mbarnes:
Spotted a couple issues:

The entirety of Makefile-tests.am is within a "if BUILDOPT_INSTALL_TESTS ... endif" conditional, which means the "make check" tests don't run without the --enable-installed-tests switch.  That seems wrong; they should run regardless of the switch, shouldn't they?

Also you added "TESTS += test-rollsum" twice when it looks like maybe you meant to add "TESTS += test-varint"?

Maybe do "insttest_PROGRAMS += $TESTS" somewhere toward the bottom instead of adding each one individually (assuming installed tests are a superset)?  Less prone to oversight as test programs accumulate.

cgwalters:
I looked at this briefly.  It's going to be a bit of a pain to conditionalize the installed tests but keep the make check.   Should still be possible, and I'll get back to this patch at some point.

cgwalters:
@giuseppe ended up doing this in another commit.

PR #34, opened 2015-01-05
Port to libgsystem errno API, hard depend on 2014.3

cgwalters:
This is long overdue to make it to libgsystem.  Update our dependency.


mbarnes:
Pretty straight-forward.  LGTM.

PR #37, opened 2015-01-06
refs: Use G_IO_ERROR_NOT_FOUND when a ref lookup fails

cgwalters:
rpm-ostree had code to check for this, which didn't actually work.

I don't see a no backwards compatibility concern in changing this, as
it's unlikely a caller would try to sensibly disambiguate FAILED.


james-antill:
LGTM :+1:

cgwalters:
This was merged.

PR #38, opened 2015-01-07
Use *at() functions for native filesystem commits

cgwalters:
This is just an efficiency optimization.  We're getting fairly close
to all of the hot code paths using `*at()`.

Note that we end up maintaining a half-duplicate code path set here,
because we still need to support commits from an arbitrary GFile *,
which in a possible common case is an OSTree commit.

I think it's worth it though.


james-antill:
ACK :+1:

PR #40, opened 2015-01-12
Pull metadata objects before content objects

mbarnes:
This addresses items 1 and 2 of https://bugzilla.gnome.org/show_bug.cgi?id=740276

Not sure if we already have item 3 or what would be entailed there.  Testing this was more challenging than the patch itself, and I'm still trying to figure out a good way to induce a large pull on demand.  But from what I could tell the total objects to be downloaded didn't seem to increase once all the metadata was known.

I imagine there's some followup work to be done, but see what you think of this so far.

### DIFF
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -42,6 +42,7 @@ typedef struct {
   guint refcount;
   OstreeFetcher *self;
   SoupURI *uri;
+  int priority;
 
   OstreeFetcherState state;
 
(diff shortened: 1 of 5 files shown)


james-antill:
It looks like it should work, and I have no real objections, but some thoughts:

 * The first two patches are a lot of code to just change q_queue_push_head vs. g_queue_push_tail, are we going to use the extra functionality?
 * It also changes the order of requests (things of the same priority now go to the head of the queue instead of the tail).
 * (minor) Saying "metadata" in the progress patch seems wrong, as we are also downloading data if we are doing N > 1 requests.

mbarnes:
We use the extra functionality in the 2nd commit where I introduce a higher priority for metadata objects than content objects in enqueue_one_object_request().  It's true that equal-priority requests are now processed as as a LIFO instead of FIFO due to they way g_queue_insert_sorted() works, but I didn't think it mattered.

The "metadata" in the message comes from Colin's suggestion in the Bugzilla bug, but I'm fine with tweaking it.  The idea (AIUI) is to switch to the normal downloading message once we have all the metadata objects and can then accurately show progress -- the Y in "Receiving objects: ... (X/Y)" stays fixed.

cgwalters:
The only thing I can think of offhand where ordering matters is: We do support resuming partial downloads.  Objects are stored in tmp/$(sha256 of url).  But actually it's probably better if they go to the front as it makes it more likely we'll resume interrupted downloads first.

True that we'll download content here as well, and presently it's very likely that we download the 20MB `/boot/initramfs.img` object.  I don't have any good ideas for rewording the fetch line though, what you have now seems OK.

cgwalters:
So :+1:  from me, thanks!

PR #41, opened 2015-01-12
Fix bare user

cgwalters:


alexlarsson:
Looks good to me. I tested it and it fixed my problem.

PR #42, opened 2015-01-14
static-delta: do not accept both --from=REV and --empty for generate

giuseppe:
a small thing I've noticed while browsing the new code


cgwalters:
Bugfix looks good, but the test case doesn't really test the bug - the test would *also* pass with the old code.  Look at what e.g. `test-pull-metalink.sh` does:

    if ${CMD_PREFIX} ostree --repo=repo pull origin:main 2>err.txt; then
	assert_not_reached "pull unexpectedly succeeded"
    fi
    assert_file_has_content err.txt "${msg}"

cgwalters:
Merged, thanks!  Don't forget to ping in github after force-pushing, since AFAIK there's no notification.

PR #43, opened 2015-01-15
fetcher and deltas cleanup

cgwalters:
A small prelude, then a major rework of OstreeFetcher to use dirfd-relative API, leading to a few static deltas fixes.

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -574,6 +574,60 @@ ostree_content_stream_parse (gboolean                compressed,
 }
 
 /**
+ * ostree_content_file_parse_at:
+ * @compressed: Whether or not the stream is zlib-compressed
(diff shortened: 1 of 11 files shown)


james-antill:
First two and last two look fine, to me.

The middle one worries me a little bit, but might well be unfounded: The change from explicit ref counting for the GFile vs. just copying the tmpdir_fd around feels like weird bugs waiting to happen.

cgwalters:
Yeah, no way to refcount a fd without creating a wrapper (or duping it), but basically it's owned by the OstreeRepo, which has a lifetime >= everything else in `ostree_repo_pull()` etc.

james-antill:
Yeh, figured but thought I'd double check as it looks weird replace the ref/unref and with the strdup() calls. Anyway :+1:

PR #45, opened 2015-01-16
admin: Add set-origin command

cgwalters:
See https://github.com/projectatomic/rpm-ostree/issues/42 for
rationale.  There are two high level use cases:

 - If OS comes unconfigured, this is a way
   to point it at a repo of your choice.
 - To switch between repositories while keeping the same branch
   easily.

### DIFF
--- a/Makefile-ostree.am
+++ b/Makefile-ostree.am
@@ -60,6 +60,7 @@ ostree_SOURCES += \
 	src/ostree/ot-admin-builtin-instutil.c \
 	src/ostree/ot-admin-builtin-cleanup.c \
 	src/ostree/ot-admin-builtin-os-init.c \
+	src/ostree/ot-admin-builtin-set-origin.c \
 	src/ostree/ot-admin-builtin-status.c \
 	src/ostree/ot-admin-builtin-switch.c \
 	src/ostree/ot-admin-builtin-upgrade.c \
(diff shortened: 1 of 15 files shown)


cgwalters:
(Rebased with docs)

mbarnes:
Looks okay to me.  I think my only suggestion would be to break this into smaller commits, as it looks like there's a bit of cleanup/refactoring mixed in.

 * Add ot_parse_keyvalue()
 * Add ot_admin_get_indexed_deployment()
 * Add ostree_sysroot_write_origin_file()
 * Add set-origin command

I find it's worth it when I have to bisect a regression.

cgwalters:
Makes sense, wasn't too hard to do.  Rebased patch set uploaded.

mbarnes:
LGTM   :+1: 

Just a question about brace style -- for standalone code blocks do you typically prefer this:

    { statement;
      statement;
    }

over this:

    {
      statement;
      statement;
    }

The former looks a bit odd to me, but I can get used to it.

cgwalters:
I agree it looks odd, but the thing I kept tripping over with the latter is that my brain kept looking for the `if ()` or `while ()` block, and at least with the `{ variable` style it's more visibily different.

I'm not strongly attached to the style.  James used `if (TRUE)` for the same thing.  Maybe do a quick vote or something?

mbarnes:
No strong opinion on brace style; was just asking for future reference.

Not really a fan of `if (TRUE)`.

james-antill:
FWIW. I also thought the raw block looked weird, and would often look for an if/while/etc wondering if an extra line break had got in there somehow. I picked up the if (TRUE) thing from somewhere years ago, and it seemed like a good solution to the problem. The other style seems fine to me, at first glance.

PR #47, opened 2015-01-16
static-delta: limit the number of writes in process to 10

giuseppe:
This will avoid too many open files at the same time that could cause
an EMFILE error.

### DIFF
--- a/src/libostree/ostree-repo-static-delta-processing.c
+++ b/src/libostree/ostree-repo-static-delta-processing.c
@@ -265,6 +265,11 @@ _ostree_static_delta_part_execute_raw (OstreeRepo      *repo,
       guint8 opcode;
       OstreeStaticDeltaOperation *op;
 
+      /* Limit the number of outstanding writes to 1 to prevent too many open files
+         at the same time.  */
+      while (state->outstanding_content_writes > 1)
+        g_main_context_iteration (state->content_writing_context, TRUE);
+
       if (state->object_start)
         {
           if (!open_output_target (state, cancellable, error))


james-antill:
10 is a small number for local IO, default user FDs are 1k.

cgwalters:
This patch will cause thread safety issues as it will iterate the main context from multiple threads.

There's a more fundamental problem too in that we are downloading all delta parts as fast as possible, and then executing as many as possible in parallel.

This means that if say your network is fast and storage is slow (not uncommon, particularly with fsync), we end up using total storage proportional to the delta sizes plus unpacked objects (plus the original tree).  

I think the simplest thing is to process a queue in `static_deltapart_fetch_on_complete` and `on_static_delta_written`.

cgwalters:
Basically we only have one `_ostree_static_delta_part_execute_async()` in flight at a time.

giuseppe:
thanks for the very detailed clarification.

I've used the same sync mechanism already used in ```_ostree_static_delta_part_execute_raw()```, in fact I've copied the same lines of code that are used few lines below.

If the deltapart file contains too many small files, wouldn't this still possibly trigger the same EMFILE error even if we limit it to one ```_ostree_static_delta_part_execute_async()``` at a time?

What about limiting both the number of delta part files processed and the total number of files as my patch is doing?

cgwalters:
BTW, see: https://bugzilla.gnome.org/show_bug.cgi?id=687223 for more on parallelism vs EMFILE.

There's lots of hard problems there; for one, ironically the more processors your machine has the more likely it is to hit EMFILE and crash...

Ok, I didn't see you'd added the main context iteration in there...hmm.  So basically we're still spawning the worker threads, but they block.  I think the queue approach in ostree-repo-pull.c would be easier to understand and more efficient.

As far as total number of files, if we only execute one part at a time, that will ensure we're only executing one object at a time too, right?  I don't see any parallelism in part execution at the moment (and doing so would be hard).

giuseppe:
if I am understanding it correctly, _ostree_static_delta_part_execute_raw() spawns a new thread every time a new output target is found in the delta part stream (_ostree_static_delta_part_execute_raw() invokes open_output_target() that in turn calls ostree_repo_write_content_async).  If that is correct, forcing only one part to be executed at a time, won't be sufficient to ensure that many files are opened at the same time by the same part being processed.  This patch doesn't ensure that as well, to be completely safe we should also limit the number of part being processed as you suggested.  But for now, it seems much more difficult to hit EMFILE if we limit each partfile being processed to spawn at most 10 (or whatever magic number we decide about) other write threads.  Even if there is only one object being processed by _ostree_static_delta_part_execute_raw at a time, with the proper dispatch_close() being called, writer threads can live after this time as dispatch_close() closes only the write end of the pipe.

cgwalters:
Ah, I see what you're saying.  Yes, I was wrong, there is also unbounded async per delta part.  Hmm...we could also have a queue there, to ensure we're only writing one object at a time?  Or maybe the async here is overkill.

Honestly the thing we end up actually blocking on for the most part of course is `fdatasync()` - and doing that more intelligently would be the most dramatic speed up for most users.  (The trick of course is to do that while still retaining safety).

What do you think about just removing the async writes here for now?

giuseppe:
maybe we could maintain the possibility to have some sort of threading here in case the decompression of files will be too cpu intensive?  In this case we should move the decompression part in the spawned thread.  But probably it doesn't make sense now, I will take a look and try to remove the async writes as you suggested.

giuseppe:
it seems that removing the _async call is not so straightforward, as ostree_repo_write_content blocks reading on the input stream and the thread invoking this function is also feeding the input stream.

Could we modify my original patch to allow only 1 thread instead of 10?

cgwalters:
I'm OK with that for now yep.

giuseppe:
I pushed a new version that replaces 10 with 1

cgwalters:
:+1: pushed, thanks!

PR #48, opened 2015-01-16
Fail early when user lacks write permission

mbarnes:
When running certain yum commands as an unprivileged user it fails quickly with a helpful "You must be root to perform this command."

I tried to do something similar for OSTree -- just the repo-related commands so far.  If you like this I can do something similar for the admin commands.

When reviewing, look out for any commands I may have missed.

Related to: https://bugzilla.gnome.org/show_bug.cgi?id=743040

### DIFF
--- a/doc/ostree-sections.txt
+++ b/doc/ostree-sections.txt
@@ -201,6 +201,7 @@ ostree_repo_new_default
 ostree_repo_open
 ostree_repo_set_disable_fsync
 ostree_repo_is_system
+ostree_repo_is_writable
 ostree_repo_create
 ostree_repo_get_path
 ostree_repo_get_mode
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -65,6 +65,7 @@ struct OstreeRepo {
 
   gboolean inited;
   gboolean writable;
+  GError *writable_error;
   gboolean in_transaction;
   gboolean disable_fsync;
   GHashTable *loose_object_devino_hash;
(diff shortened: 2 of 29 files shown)


mbarnes:
Reworked this according to your previous suggestions.

mbarnes:
(Note to self to update the commit message for ostree_ensure_repo_writable() before merging -- it still names the "you must be root" error message.)

cgwalters:
Looks great to me.  I think we could get a basic test case for this if we did say `mkdir repo; ostree init --mode=archive-z2 repo; chmod 0 repo`.

mbarnes:
Okay, I'll amend a test case.  Do you think it's worth doing something similar for the admin commands?

mbarnes:
Getting back to this finally.  I'll go ahead and add something similar for admin commands.

cgwalters:
This needs a rebase.  Is it still waiting on the admin command update?

mbarnes:
Yeah, on my list for this week.  Unless you'd prefer a separate pull request?

mbarnes:
Rebased.

mbarnes:
Haven't tested these new commits yet because I'm having compose issues, but see if you're okay with what I'm doing here.  Permission tests for admin commands are a little less straight-forward.

mbarnes:
Finally got past my compose issues (local env issue).  Seems to work.

mbarnes:
Looks to me like you must be root for all the admin commands except "status" -- is that correct?

Checking UID would actually catch more cases than checking directory permissions, and it's simpler.

cgwalters:
Maybe `ostree_sysroot_new_writable()` ?

mbarnes:
Can you elaborate?  I'm wary of introducing new public API for this.

I was thinking more along the lines of a "must-be-root" flag for ostree_admin_option_context_parse().

cgwalters:
Ah yeah, I like the flag on the commandline approach better.  We could even use it to automatically flag in the help which commands require root.

mbarnes:
Rebased again but only changed the last two commits.  A --dry-run or some similar option that avoids needing root could derail this approach, but for now we're okay.

cgwalters:
Sorry about the delay - this had dropped off my radar.  :+1:  from me, merged!

PR #49, opened 2015-01-21
pull: Implement --enable-syncfs

giuseppe:
Differently than fsync that is called after each file is modified,
syncfs is used only once at the end of the pull to sync the file
system which contains the repository.

Results of pull on an empty repository:

$ LANG=C sudo time ./ostree --repo=repo --disable-fsync pull master master
0 metadata, 0 content objects fetched; 83524 KiB; 6 delta parts fetched, transferred in 5 seconds
17.09user 2.41system 0:05.77elapsed 337%CPU (0avgtext+0avgdata 296132maxresident)k
0inputs+868584outputs (0major+102828minor)pagefaults 0swaps

$ LANG=C sudo time ./ostree --repo=repo pull master master
0 metadata, 0 content objects fetched; 83524 KiB; 6 delta parts fetched, transferred in 401 seconds
17.31user 5.73system 6:41.33elapsed 5%CPU (0avgtext+0avgdata 296864maxresident)k
0inputs+868584outputs (0major+132664minor)pagefaults 0swaps

$ LANG=C sudo time ./ostree --repo=repo --enable-syncfs pull master master
0 metadata, 0 content objects fetched; 83524 KiB; 6 delta parts fetched, transferred in 6 seconds
17.14user 2.57system 0:10.60elapsed 185%CPU (0avgtext+0avgdata 296212maxresident)k
0inputs+868584outputs (0major+86783minor)pagefaults 0swaps

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -35,6 +35,7 @@
 #include "ostree-mutable-tree.h"
 #include "ostree-varint.h"
 #include <attr/xattr.h>
+#include <glib/gprintf.h>
 
 gboolean
 _ostree_repo_ensure_loose_objdir_at (int             dfd,
@@ -57,6 +58,17 @@ _ostree_repo_ensure_loose_objdir_at (int             dfd,
         }
     }
   return TRUE;
+}
+
+void
+_ostree_repo_get_tmpobject_path (char *output,
+                                 const char *checksum,
+                                 OstreeObjectType objtype)
+{
+  g_sprintf (output,
+             "tmpobject-%s.%s",
+             checksum,
+             ostree_object_type_to_string (objtype));
 }
 
 static GVariant *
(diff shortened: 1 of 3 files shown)


cgwalters:
See https://bugzilla.gnome.org/show_bug.cgi?id=728065

The problem with this is that if the system crashes in the middle, there's no guarantee that the objects written before the syncfs() are valid.  I actually hit that, see https://git.gnome.org/browse/ostree/commit/?id=2396608754060e9e7a68843b65e4d3b8f4a4b5f2

giuseppe:
I've pushed another version with the logic described here: https://bugzilla.gnome.org/show_bug.cgi?id=728065

All tests pass

cgwalters:
Hmm, didn't I have more comments on an older version of this commit?  I can't find it offhand.  But for example I thought I wrote up a concern about power cuts; this patch isn't addressing that yet is it?
I had the idea to use the boot uuid or something like that.

I wish github made it easier to find old comments...

giuseppe:
yes, I didn't implement that.  I believed it was a comment for a later improvement.  I am going to send another version that uses the boot uuid.

cgwalters:
Squashed these fixups on top and pushed.  :+1:

PR #50, opened 2015-01-21
Pull delta status

cgwalters:


mbarnes:
Changes look okay to me.  :+1: 

Couple minor asides:
 * For all the cleanup functions provided by libgsystem, GString seems overlooked.
 * The number of private string keys used with OstreeAsyncProgress for pulls seems to be growing.  Maybe worth some #define's with comments in ostree-repo-private.h for posterity / typo avoidance?

cgwalters:
I started to prototype out a while ago something like a GVariant object generator.  What we have with async progress is basically a big bag of readonly properties, then you can get a new snapshot of them.  So we could model it as an `a{sv}`, then have accessor functions generated like `ostree_async_progress_get_n_delta_parts(v)` that would really just be inline invocations of `g_assert (g_variant_lookup (v, "n-delta-parts", &nparts)); return nparts;` or so.

cgwalters:
Merged this but will keep the followup comments in mind.

PR #51, opened 2015-01-22
commit: Fix segfault on async writes if object exists and checksum reque...

cgwalters:
...sted

If an object already existed and we somehow tried to pull it, the
caller would still expect a returned checksum.

This appears to happen with static deltas for some reason; we might be
including duplicate metadata objects.  Regardless, this is a bug that
should be fixed.


james-antill:
LGTM :+1:

PR #52, opened 2015-01-22
build: Add --disable-static-deltas

cgwalters:
Since they're unstable, we want to allow organizations shipping ostree
now to be future proof against any changes.


mbarnes:
Looks okay except for the double negative in `AS_HELP_STRING`.

IMO, would be clearer as:
```
AS_HELP_STRING([--enable-static-deltas], [Enable static delta code (default: yes)])
```

cgwalters:
Thanks, pushed with that change.

PR #53, opened 2015-01-23
Explicitly label .origin files as configuration

cgwalters:
subscription-manager has a daemon that runs in a confined domain,
and it doesn't have permission to write usr_t, which is the default
label of /ostree/deploy/$osname/deploy.

A better long term fix is probably to move the origin file into the
deployment root as /etc/ostree/origin.conf or so.

In the meantime, let's ensure the .origin files are labeled as
configuration.

### DIFF
--- a/src/libostree/ostree-sepolicy.c
+++ b/src/libostree/ostree-sepolicy.c
@@ -45,6 +45,8 @@ struct OstreeSePolicy {
 
   GFile *path;
 
+  gboolean runtime_enabled;
+
 #ifdef HAVE_SELINUX
   GFile *selinux_policy_root;
   struct selabel_handle *selinux_hnd;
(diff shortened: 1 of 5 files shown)


cgwalters:
Downstream bug: https://bugzilla.redhat.com/show_bug.cgi?id=1182647

james-antill:
Wow, pretty hacky. Maybe don't use "/etc/ostree/remotes.d/dummy.conf" just in case remote configurations get put under a different configuration? Should work for now though, and moving it to origin.conf should be done before the label changes.

cgwalters:
No disagreement it's hacky.  I don't think we're at much risk of the labels for the remote config changing, because labels are a kind of ABI - on upgrades, unless the admin takes explicit action to relabel, the old labels will stay around, and new policy needs to handle old label configurations for user data.  With rpm-ostree we do consistently get updated labels for everything in `/usr`.

For moving `origin.conf`, you're saying do that patch before this one, then use the label for `origin.conf` for the legacy location?  I think that makes sense...that's going to need design, so if you don't object I'd like to keep this patch cherry-pick-able.

james-antill:
> on upgrades, unless the admin takes explicit action to relabel, the old labels will stay around, and new policy needs to handle old label configurations for user data

Yeh, that's true. I just thought we'd done it in the past, maybe not though.

> For moving origin.conf, you're saying do that patch before this one, then use the label for origin.conf for the legacy location?

No, I was just saying we should be able to do the patch to move origin before any labeling changes happen on the remote config. files. So :+1:

cgwalters:
Ok, thanks, merged.

PR #58, opened 2015-02-12
Add an API to set/unset a deployment tree's mutability

cgwalters:
This will be used by rpm-ostree to unset the immutable bit temporarily
in order to do package layering.  We could add an API to deploy a tree
without the immutable bit, but this is simpler.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1917,8 +1917,8 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
                                       cancellable, error))
     goto out;
 
-  if (!_ostree_linuxfs_alter_immutable_flag (new_deployment_path, TRUE,
-                                             cancellable, error))
+  if (!ostree_sysroot_deployment_set_mutable (self, new_deployment, FALSE,
+                                              cancellable, error))
     goto out;
 
   { ostree_cleanup_sepolicy_fscreatecon gpointer dummy = NULL;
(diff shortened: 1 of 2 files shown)


mbarnes:
`_ostree_linuxfs_alter_immutable_flag()` doesn't *really* block and could drop its `GCancellable` parameter.  Then you wouldn't need it for `ostree_sysroot_deployment_set_mutable()` either.
(I just find unnecessary `GCancellable` params to be misleading.)

Otherwise looks okay to me.

cgwalters:
I guess I've found it easier so far to just carry them along everywhere, but I've no problem with dropping them either.

mbarnes:
Just to note where I'm coming from on this -- on my old desktop project, blocking the user interface was a major problem.  In reworking the APIs I used GCancellable as one way to clearly indicate which functions can block and may require an asynchronous version.  I understand it's not as applicable here, but the tendency carries over.

cgwalters:
To me this gets to a big conceptual question - e.g. opening a local file a blocking operation?  What about directory enumeration?  Gio at least includes `GCancellable` pretty much everywhere, and that's why I was too.

Perhaps a middle ground is that "single operations" like `open()` or `unlink` aren't cancellable, but enumerating a directory (or operating recursively on dirs, like `glnx_shutil_rm_rf_at` are cancellable even though they operate locally.

mbarnes:
Right, multi-step operations have natural check points for cancellation, especially when the steps are repetitive.  I guess my objection is more with taking a `GCancellable` argument but then not using it.

For something like `_ostree_linuxfs_fd_alter_immutable_flag()` it only really makes sense to check up front if the `GCancellable` is already cancelled before proceeding.  But then I tend to think that should be left up to the caller.

cgwalters:
I'm still waffling on this.  If we added the stanza for checking the cancellable, i.e. match what e.g. `g_file_delete` does, would that address your objection? 

The dual factors of
 - Matching what Gio is doing
 - Also given the fact that OSTree has so many APIs that all take cancellables, we'd create divergence by not doing so

have me leaning towards keeping them.

mbarnes:
Didn't mean to make this into a big issue so I'll relax a bit.

I think as long as a function is just doing *something* with its `GCancellable` -- either checking it directly or passing it on -- I'm cool with keeping it.  So yeah, the added stanza would address my objection.

cgwalters:
I updated this to:
 - Check the cancellable
 - Use `*at()`

PR #59, opened 2015-02-13
deploy: Also look for /usr/lib/os-release

cgwalters:
With newer versions of systemd, the file has moved there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # Standard C/Automake goo
 .deps
 .libs
+.dirstamp
 *.typelib
 *.la
 *.lo
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "libglnx"]
+	path = libglnx
+	url = https://git.gnome.org/browse/libglnx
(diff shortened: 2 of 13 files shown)


mbarnes:
Seems a little weird using `g_file_load_contents()` for one directory and `g_file_get_contents()` for the other, but otherwise :+1:.

cgwalters:
Yeah, it was ugly...I was trying to do a partial transition away from GFile*.  But let's shave this yak a bit.  I just pushed https://git.gnome.org/browse/libglnx/commit/?id=1ebfefa565692ca033ba2d156c771d9e2c4017fc (post-review on that appreciated btw!)

Then the updated commit here uses it, and IMO it's much prettier, more efficient, etc.

mbarnes:
libglnx post-review: Looks nice.  Using a `GByteArray` in `glnx_fd_readall_malloc()` might shave off a few lines, but... eh.  Not too important.

The updated commit is much cleaner indeed.  :+1: :+1:

cgwalters:
I tried to use a `GString` but it's not really cleaner without a `g_string_ensure_capacity()` or something.  Actually I just filed https://bugzilla.gnome.org/show_bug.cgi?id=744992 to track this.

PR #60, opened 2015-02-17
Use libglnx

cgwalters:
Starting down the path of not using libgsystem, but this
breaks for me with:

    $ make
    Makefile:2445: libglnx/.deps/libglnx_la-glnx-console.Plo: No such file or directory

It's not immediately obvious to me what's broken here.


cgwalters:
Figured out a workaround.  Rebased and works now.

cgwalters:
Rebased again with some more fixes.

mbarnes:
LGTM  :+1: 

I'll try and chip away at some of the libgsystem->libglnx porting myself.

PR #61, opened 2015-02-19
Check for OSTREE_REPO in ostree_repo_new_default()

mbarnes:
Convenience feature to avoid having to pass --repo options repeatedly.

Before falling back to the default system repository path, check for a repository path defined by the OSTREE_REPO environment variable.


mbarnes:
I guess the correct DocBook tag would be `<envar>`, not `<literal>`.  I'll fix before pushing.

mbarnes:
https://git.gnome.org/browse/ostree/commit/?id=5aa0d51d7a55c48aa986ddbacbeb74e224c2be72

PR #62, opened 2015-02-20
repo: Port APIs used by prune to fd-relative *at calls

cgwalters:
More of the general trend away from GFile * to the faster and more
secure world of *at().

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2316,19 +2316,39 @@ ostree_repo_delete_object (OstreeRepo           *self,
                            GError              **error)
 {
   gboolean ret = FALSE;
-  gs_unref_object GFile *objpath = NULL;
+  int res;
+  char loose_path[_OSTREE_LOOSE_PATH_MAX];
+
+  _ostree_loose_path (loose_path, sha256, objtype, self->mode);
 
   if (objtype == OSTREE_OBJECT_TYPE_COMMIT)
     {
-      gs_unref_object GFile *detached_metadata =
-        _ostree_repo_get_commit_metadata_loose_path (self, sha256);
-      if (!ot_gfile_ensure_unlinked (detached_metadata, cancellable, error))
-        goto out;
+      char meta_loose[_OSTREE_LOOSE_PATH_MAX];
+
+      _ostree_loose_path_with_suffix (meta_loose, sha256,
+                                      OSTREE_OBJECT_TYPE_COMMIT, self->mode, "meta");
+
+      do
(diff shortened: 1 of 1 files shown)


mbarnes:
LGTM  :+1:

The pattern you're using for `fstatat()` and `unlinkat()` would make for nice little wrapper functions in `glnx-fdio.c`.

mbarnes:
As an aside, I get why the *at() functions are faster but how are they more secure?

cgwalters:
If a system administrator (running as root) types: `rm -rf /home/someuser/public_html/etc`, that gives `someuser` an opportunity to make a symlink `/home/someuser/public_html -> /` and destroy `/etc`. In contrast if the admin does `cd /home/someuser/public_html`, then they can use `ls` to verify they're in the right place, and just `rm -rf etc` securely.  The bash process holds an implicit reference to the original traversed dir (passed to the child `ls` and `rm`), and fd-relative APIs are a generalization of this process-owned pre-traversed path concept.

For OSTree, there are directories owned by non-root users that it traverses when making the new deployment root.  However, I think https://git.gnome.org/browse/ostree/commit/?id=b6704f8346c8f7392c69f30f0053800abc36fbc3 closes these holes for us, but if coreutils `rm` didn't use `unlinkat`, an admin would be in danger trying to remove a directory hierarchy owned by a user.

cgwalters:
As for the wrapper functions...WDYT about https://github.com/GNOME/libglnx/pull/1 ?

So far, I've avoided having purely trivial wrapper in libglnx.  The thinnest is perhaps `glnx_opendirat_with_errno` but that's because I found typing all of those flags a bother =)

If we were to create thin wrappers there'd be a number of them...`renameat`, `fstatat`, `unlinkat`, `linkat`, `symlinkat`, `mkdirat` are the main ones.  On the other hand, at least with the submodule approach we could just rev libglnx any time we wanted a new syscall with EINTR-handling/cancellable/gerror etc.  (As opposed to a new public API in libgsystem with `configure.ac` bumps etc.)

cgwalters:
Pushed, but will leave open for followups.

mbarnes:
Thanks for the security explanation; makes a lot of sense.

As for GNOME/libglnx#1, I like the macro but I think there should be an explicit error argument.  You might want to use a local GError* instead of a GError** argument (which I typically name 'local_error' or 'tmp_error') and also assuming a variable name feels kinda evil however strong the naming convention.

cgwalters:
I said I'd keep this open for followups, but eh...let's just try to do code cleanups as a normal gradual/continual process.

PR #63, opened 2015-02-20
checkout: Drop internal use of GFile *

cgwalters:
Checkout was one of the first complex code paths I tried to convert to
*at().  I ended up keeping both, because I hit the "xattrs for a
symlink" problem.  Later, Florian gave me a workaround, and we started
using it here, but the GFile * parameters weren't deleted.  They're
not used, so do so now.


mbarnes:
:+1:

PR #64, opened 2015-02-23
giuseppe/wip/deltas

giuseppe:
bsdiff support for static deltas + other minor fixes


cgwalters:
It looks like we're basically doing bsdiff for all "apparently modified" files that aren't rollsummed, and aren't too large?  That may be OK, though I think in the future we'll want to perhaps back off on using rollsum if it isn't a big win (i.e. if the delta is not significantly smaller than xz compressing the new object).

giuseppe:
the bsdiff code assumes that the delta file is going to be compressed afterwards (I learned this after I got shocked that the diff size was bigger than the new file), but since we are compressing the entire deltapart file, so I am relying on the xz compression we are performing there.  I am not sure how easily we can check the compressed size here, as probably the compression of just the diff between two files will be smaller than the compression done on it as part of the deltapart stream.
Probably this doesn't make much difference as we will anyway have some sort of heuristic to decide if it is worth or not to use bsdiff.

cgwalters:
On my test data, the previous deltas are 30MB total, this takes it to 10MB, which is pretty good!  And indeed I can see the deltapart which has lots of bsdiff'd objects:

    part 4 n:147 compressed:942021 uncompressed:31089940

In other words, xz takes that to 3% size.

However when I try to actually do a pull, I get:

    ERROR:src/libostree/ostree-repo-static-delta-processing.c:507:bspatch_read: assertion failed: (opaque->offset + length < opaque->state->payload_size)

I pushed two commits to https://github.com/cgwalters/ostree/tree/wip/deltas - want to take a look and add them to your branch?  Then do you think you'd have a chance to look at that assertion?

giuseppe:
Thanks for the testing.

It took a while to reproduce the issue here, but obviously the condition in the assert should be ```<=``` and not ```==```.

I had a look at your commits, I like the idea of disabling bsdiff when not needed.

About the second patch that adds a debug message: unfortunately the raw ```bsdiff``` output is not much different in size than its inputs (but it is mostly '\0' filled), and it needs an additional compression to turn to be useful, so the stats output per file won't be very helpful as they will show no compression at all.

giuseppe:
I have pushed a new version that fixes the assert, include your patch, and add ```--generate-static-delta``` to the commit command (plus some other minor cleanups).

cgwalters:
One time, I got:

    Receiving delta parts: 4/8 29.7 MB/s 29.7 MB/64.4 MB
    (ostree pull:18335): GLib-GIO-CRITICAL **: g_output_stream_write_all: assertion 'G_IS_OUTPUT_STREAM (stream)' failed

    (ostree pull:18335): GLib-GIO-CRITICAL **: g_simple_async_result_take_error: assertion 'error != NULL' failed

But I can't reproduce under `env G_DEBUG=fatal-warnings gdb` :confused: 

Will debug a bit more.

cgwalters:
Added on top https://github.com/GNOME/ostree/commit/67cec3a4defe601f8d05412008a909d93513b181 - make sense to you?

cgwalters:
Merged to master with the above, thanks!  Let's do more work as new PRs.

PR #65, opened 2015-02-23
tests: Move test gpg keyring into writable tmpdir

cgwalters:
At least the version of gpgme in RHEL7 wants to create a lock file,
work around this by copying it into the writable test temporary
directory.


mbarnes:
LGTM  :+1:

PR #66, opened 2015-02-24
repo: detached sigs: Use error prefixing instead of overwriting

cgwalters:
Noted when "rpm-ostree compose sign" failed to write to a repo due to
permissions.


cgwalters:
Actually, this is suboptimal, better commit inbound.

cgwalters:
Rewritten to put the error prefixing into the underlying functions instead of the wrapper.

mbarnes:
:+1:

cgwalters:
Merged.

PR #68, opened 2015-02-26
deploy: Use glnx file copy code

cgwalters:
This is centralizing more generic utility code in libglnx.


cgwalters:
Depends on https://github.com/GNOME/libglnx/pull/3

cgwalters:
Added some more cleanup commits here.

src/libostree/ostree-repo-commit.c
```diff
@@ -848,65 +848,57 @@ devino_equal (gconstpointer   a,
...
-                                              error);
-      if (!enumerator)
+      if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
         goto out;
+          
+      if (dent == NULL)
+        break;
 
-      dirname = gs_file_get_basename_cached (objdir);
+      if (strlen (dent->d_name) != 2)
```
mbarnes: This `strlen` check could use a comment.  I think it's to recurse only in the two-digit hex directories, right?

src/libostree/ostree-repo-commit.c
```diff
@@ -921,17 +913,26 @@ scan_loose_devino (OstreeRepo                     *self,
 
           dot = strrchr (name, '.');
           g_assert (dot);
-
+          
           if ((dot - name) != 62)
```
mbarnes: Suggest a comment here as well.  This is the rest of the 64-digit hash, I assume.

mbarnes:
Couple spots to clarify (if only for my sake), but overall looks like a nice cleanup.  :+1:

cgwalters:
Thanks, pushed with added comments there.

PR #69, opened 2015-03-02
Multiple "ostree.gpgsigs" keys in metadata

mbarnes:
When signing an already signed commit you get multiple `ostree.gpgsigs` keys in the detached metadata.  The 1st instance has the 1st signature, the 2nd instance has the 1st and 2nd signatures, etc.  Signature verification extracts the 1st instance from the metadata, and consequently the additional signatures are never verified.  This is bad.

I tracked this down to a bug in `_ostree_detached_metadata_append_gpg_sig()`.  The best solution was to use `GVariantDict` instead of `GVariantBuilder`, but this required a GLib dependency bump.

I tried to add a test case for this, but the CLI currently lacks a way to just dump all the metadata so I can test for duplicate `ostree.gpgsigs` keys.  Maybe `ostree show` needs a `--print-metadata` and `--print-detached-metadata` (separate from printing specific keys)?


cgwalters:
This all looks good to me.  I'd be more confident it fixed the bug it claimed with a test, but since I didn't write any when initially implementing the multiple signatures functionality, no hard requirement from me to add one.  So :+1:

mbarnes:
Cool, thanks.

If you like my suggestion for `ostree show`, I can follow up on that separately.  That would allow me to write a test case for this.  Better late than never...

cgwalters:
Ah sorry, I missed that show suggestion.  Might be kind of annoying parsing the output of shell script for this...I started using gjs for some tests, we just want to do `g_variant_n_children` after loading the detached metadata right?  But if you can think of a shell-script way for this I'm fine with that too.  Were you thinking say RFC882 format so it's somewhat grep-able?

(We could also use pygobject for tests if you prefer that)

mbarnes:
+1 to Python tests in general (or JavaScript... I suppose I need to learn it at some point).

But in this particular case it may be more annoying to have to reproduce the test environment in `test-gpg-signed-commit.sh` in another language.  For the output I was thinking of just one metadata key/value per line, so I can do something like verify `grep --only-matching ostree.gpgsigs $METADATA | wc --lines` equals 1.

PR #70, opened 2015-03-03
Rewrite OstreeGpgVerifier to use GPGME

mbarnes:
This sets the stage for more advanced signature management.

(Also, talking to GPG over pipes sucks.)

Previously we were spawning `gpgv2` with a bunch of `--keyring` options
for `/usr/share/ostree/trusted.gpg.d/` and whatever other keyring files
were explicitly added.  GPGME has no public API for multiple keyrings,
so we work around the issue by setting up a temp directory to serve as
a fake "home" directory for the crypto engine and then concatenate all
the keyring files into a single public keyring (`pubring.gpg`).

Unfortunately at present we do this on every signature verification.
There's a desire to cache this concatenation, but the problem is the
user may be unprivileged.  So it seems the cache would have to be per
user under `$XDG_CACHE_HOME`, which OSTree doesn't otherwise use.  I'm
open to suggestions.

We do at least clean up the temp directory when finished, and I have
further API changes planned to `OstreeGpgVerifier` to help mitigate the
performance impact.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -89,8 +89,6 @@ libostree_1_la_SOURCES = \
 	src/libostree/ostree-bootloader-syslinux.c \
 	src/libostree/ostree-bootloader-uboot.h \
 	src/libostree/ostree-bootloader-uboot.c \
-	src/libostree/ostree-gpg-verifier.c \
-	src/libostree/ostree-gpg-verifier.h \
 	src/libostree/ostree-repo-static-delta-core.c \
 	src/libostree/ostree-repo-static-delta-processing.c \
 	src/libostree/ostree-repo-static-delta-compilation.c \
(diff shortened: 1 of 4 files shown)


mbarnes:
Tacked on a small API change / optimization.

mbarnes:
I think I've address all your review comments except the backward-compat homedir cruft.

Still an open question on whether to include the user's personal trusted keyring in the signature verification as `gpgv2` was (perhaps unintentionally) doing before.

cgwalters:
Hmm...were we using the root keyring before for `ostree admin upgrade`?  It's maybe a compatibility issue if we suddenly stop doing so.  But on the other hand, we really shouldn't have system processes looking in home directories.  And adding things to root's keyring isn't a good model for trust too.

I'd lean towards stopping from reading it if it's not too hard to do.

mbarnes:
Yeah, we were before because we were spawning `gpgv2` without `--no-default-keyring`, so it was implicitly including root's `trustedkeys.gpg`.  I was leaning toward stopping as well, and fortunately it's easy to do and makes the code cleaner.

cgwalters:
:+1: from me, thanks!  Also I really like this `fixup!` flow.  It gives the "interdiffs" that I always liked from Gerrit.

mbarnes:
Merged, apparently only 99% cleanly to break GitHub's auto-close.

PR #71, opened 2015-03-04
tests: add tests for mutable tree.

giuseppe:


cgwalters:
Looks fine.  At some point I'd like to visit having these `make check` tests *also* be installed if we have a lot more of them, but it can come later.

PR #72, opened 2015-03-05
giuseppe/keyfile-utils-tests

giuseppe:


cgwalters:
:+1:  from me

giuseppe:
I pushed some more commits

cgwalters:
Other than that one small comment, :+1: from me, thanks!

giuseppe:
Thanks, pushed now with the amended comment.

PR #73, opened 2015-03-07
Sysroot fd relative

cgwalters:
Just some idle work on fd-relative bits while I watch installers build and such.


giuseppe:
it looks ok to me! +1

PR #74, opened 2015-03-10
Sysroot fd relative

cgwalters:
A few more idle commits.


giuseppe:
+1 from me with the small comment fix.

cgwalters:
Fixed, thanks!

PR #75, opened 2015-03-11
Make gpgme a hard dependency

mbarnes:
In anticipation of API enhancements for GPG signature verification, which would otherwise require a non-functional stub version were GPGME excluded.

GPGME is a pretty lightweight dependency, and the motivation to exclude it is not clear.

Note, this delists `+gpgme` as a feature in the `--version` output since the signing capability is now always present.


cgwalters:
With that change :+1: from me!

mbarnes:
I was unsure about removing the `+gpgme` bit.  No problem putting it back.

Thanks!

PR #76, opened 2015-03-16
OstreeGpgVerifier: Take the signed data as a GBytes

mbarnes:
Similar to c2b01ad.  For some reason I was thinking the commit data still needed to be written to disk prior to verifying, but it's just another artifact of spawning `gpgv2` (predates using GPGME).

Makes for a nice cleanup in `fetch_metadata_to_verify_delta_superblock()` as well.


mbarnes:
Merged with Colin's suggested changes (which are now lost in the ether).

PR #77, opened 2015-03-17
Enhanced GPG signature management

mbarnes:
This introduces `OstreeGpgVerifyResult` as a new public `libostree` class to wrapper GPG signature verification results, so they can be examined independently of running a verify operation.

A new `OstreeRepo` function `ostree_repo_verify_commit_ext()` returns such an instance.

This makes possible a few CLI enhancements:

 - The `ostree show` command now prints a signature count for signed commits, along with a human-readable blurb about each signature (validity, who signed it and when, etc.).

 - `ostree_repo_verify_commit_ext()` can be more generally used as a GPG signature parser.  Signing a commit now runs a verify operation beforehand and uses the `OstreeGpgVerifyResult` to check if the commit has already been signed with the given key ID.

 - The `ostree gpg-sign` command now has a `--delete` option to delete signatures by key ID.  This also runs a verify operation beforehand and uses the `OstreeGpgVerifyResult` object to find which signatures to delete.


cgwalters:
This looks pretty good overall.  For next steps, what do you think about showing the signature status in `ostree admin status` (and `rpm-ostree status`) as well as on updates?  It'd give me a warm fuzzy feeling if I typed `atomic host upgrade` and saw ... `Good signature from $DISTRO`

mbarnes:
That's doable.  I guess we would have to internalize most of 7d5854aec943b4d143e520f9ad69603d3636915e (hacked together as it is) into `libostree` somehow so we're not copying and pasting all that formatting gunk.

mbarnes:
Pushed a batch of "fixup" commits for the simple stuff.

I guess the main outstanding issue is whether you're okay with accepting key IDs as command-line arguments and in `ostree_gpg_verify_result_lookup()`.

My personal feeling is -- while I do acknowledge the risk of key collisions in general GPG usage, especially when talking to public key servers -- in this restricted context of local trusted keyrings I think they're fine.

cgwalters:
Re keyids - fair enough.  This can be solved indeed at a higher level.  The question really is around how those keyrings are created.  Basically I feel the way RPM (or maybe dnf/yum) do it for new repositories is fairly broken.  I think repository files should have a `gpg_fingerprint` value, and the key must match that.  Then if you fetch the repository file over TLS (perhaps even CA pinned TLS), you have a trusted path.

Anyways it's a complex topic, no need to block this commit on it.  Nice work!

mbarnes:
Agreed re: repo files.  That has struck me as fishy in the past but wasn't able to express it as well.

Good to merge then?

cgwalters:
:+1: from me!

mbarnes:
I hate losing review comments so I didn't force-push the rebase to github, but it's merged.

https://git.gnome.org/browse/ostree/commit/?id=4a2733f
..
https://git.gnome.org/browse/ostree/commit/?id=9b8bad3

Will follow up on some of the GPGME weirdness I noted.

mbarnes:
Started the follow up work in #80

PR #78, opened 2015-04-01
use glibc's xattr support instead of requiring libattr.

dsvensson:
Fixes build on Debian,

Here's some details of the same problem in systemd:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=756097
http://cgit.freedesktop.org/systemd/systemd/commit/?id=d2edfae0f9b


cgwalters:
This needed a corresponding update to libglnx, which I pushed.  I also fleshed out the commit message a bit with some of the info you added to the PR.

Thanks!

PR #79, opened 2015-04-01
test_rollsum needs -lz to link.

dsvensson:
Test fails to link on Debian as it uses crc32, not sure what the best strategy is here.

    /usr/bin/ld: src/libostree/tests_test_rollsum-ostree-rollsum.o: undefined reference to symbol 'crc32'
    //lib/x86_64-linux-gnu/libz.so.1: error adding symbols: DSO missing from command line

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -119,7 +119,7 @@ TESTS_LDADD = $(ostree_bin_shared_ldadd) $(OT_INTERNAL_GIO_UNIX_LIBS)
 
 tests_test_rollsum_SOURCES = src/libostree/ostree-rollsum.c tests/test-rollsum.c
 tests_test_rollsum_CFLAGS = $(TESTS_CFLAGS)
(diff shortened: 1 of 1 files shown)


cgwalters:
This is fixed by https://git.gnome.org/browse/ostree/commit/?id=12c8ef14254c634bffe59fa97c1e28c3cc7e930d right?

(Have you tried building git master?)

dsvensson:
Just did a silly verification by simply cloning the repo from scratch to avoid having any left overs, and no, it doesn't seem like the -lz is inherited properly. I'm building on Debian Sid.

mbarnes:
Confirmed.  Debian catches this stuff, Fedora does not.  I've yet to find the magic switch.

Pushed my own fix but credited Daniel:
https://git.gnome.org/browse/ostree/commit/?id=f01fceb5d7f685760aef1e3b6ad58a066e0b39a0

cgwalters:
This is fixed now, right?  Closing, reopen if not.

PR #80, opened 2015-04-01
Show signature status in "ostree admin status"

mbarnes:
The signature blurbs from `ostree show` are encapsulated in a new `libostree` function `ostree_gpg_verify_result_describe()` for easy reuse.

Here's a sample of the new status output:

```
* fedora-atomic 696f870fe5d8989032113ae10e0684ae44a4b02ead1c17688acf2064dab82f7b.0
    Version: 23
    origin refspec: local:fedora-atomic/f22/x86_64/local-testing
    Signature made Wed 01 Apr 2015 07:52:57 PM UTC using RSA key ID 7572B288944E4891
    Good signature from "Matthew Barnes <mbarnes@redhat.com>"
  fedora-atomic c98936ea67dae6628e2d35007185bda390b208e365994b5f103fb800ee4ea552.0
    Version: 23
    origin refspec: local:fedora-atomic/f22/x86_64/docker-host
```

Debated whether to add more spacing for readability.  What do you think?

Also, while we're on the topic, what should `rpm-ostree status` with signatures look like?  Don't know how to avoid disrupting the nice tabular formatting.


cgwalters:
:+1: from me, all comments above are optional/can-be-followed-up-later.

mbarnes:
I added the `GPG:` prefix before merging.  Not rebasing here to preserve review comments.

https://git.gnome.org/browse/ostree/commit/?id=18fdc7cb971d0d6395dd142c9960ef01ab7e6f8b

PR #81, opened 2015-04-02
build: Drop libattr from the spec file

giuseppe:
commit 534c4c20c3fa5ad9500ea96093a3ece7821a6056 already drops its
usage in the code.


cgwalters:
:+1:  However note that the spec files in downstreams are not automatically synchronized with this.  What I do currently is treat the Fedora spec as canonical and try to synchronize it with distros downstream of Fedora.

PR #83, opened 2015-04-02
giuseppe/auto summary

giuseppe:
it adds a configuration flag in the config file to automatically update the summary file on a commit.

It won't interfere if we decide to add signatures to the summary file, as we can use the same signature used to sign the commit (and perhaps have a way to add signatures to an existing summary file)


cgwalters:
Cool thanks, :+1:  from me.

PR #84, opened 2015-04-04
pull: Handle remote web server not honoring range requests

cgwalters:
It's valid for the remote server to say 200 OK and give us the entire
file instead of a 206 Partial Content, and in that case we should blow
away the previous cached data, rather than blindly appending to it and
thus creating multiple copies of the data inside the file.

This problem primarily occurs when we do have the complete file, and
we're interrupted, then try again, where the new process didn't record
the download was already complete.  We do a range request for bytes
past the end, and some web servers (e.g. Akamai) will return 200 OK
with the whole content again, rather than a 416 Requested Range Not
Satisfiable.

Thus we could also fix this by saner caching strategy - since we know
the file is complete, rename it again to $checksum.done or something
before it's processed.  (Or really, rework how we do caching more
intelligently in general).

This fixes the issue that interrupted pulls failed with such
webservers, although repeated attempts would eventually succeed
because we'd unlink files that failed to pull.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1207292

### DIFF
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -492,7 +492,18 @@ on_request_sent (GObject        *object,
 
   if (!pending->is_stream)
     {
-      int fd = openat (pending->self->tmpdir_dfd, pending->out_tmpfile, O_CREAT | O_WRONLY | O_APPEND | O_CLOEXEC, 0600);
+      int oflags = O_CREAT | O_WRONLY | O_CLOEXEC;
+      int fd;
+
+      /* If we got partial content, we can append; if the server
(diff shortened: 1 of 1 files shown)


mbarnes:
Looks correct to me.  Little easier to read now too, imo.  :+1:

mbarnes:
I suggest pasting the description above into the commit message for posterity.  Good write up, would like to preserve it in the commit log.

cgwalters:
The pull request description *is* the commit message, no?  Or did you mean the bugzilla?

mbarnes:
Bah, you're right, nevermind.  Brain not completely engaged since the weekend.

PR #85, opened 2015-04-06
 Teach fsck about partial commits

cgwalters:
An OSTree user noticed that `ostree fsck` would produce `missing
object` errors in the case of interrupted pulls.

It's possible to do e.g. `ostree pull --subpath=/usr/share/rpm ...`,
which gets you just that portion of the commit.  The use case for this
was being able to see what changes would appear in an update before
actually downloading all of it.

(I think this would be better covered by static deltas, but those
 aren't final yet, and `--subpath` predates it)

Further, `.commitpartial` is used as a successor to the `transaction`
symlink for more precise knowledge in the case where a pull was
interrupted that we needed to resume scanning.

So it makes sense for `ostree fsck` to be aware of it.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -48,6 +48,7 @@ testfiles = test-basic \
 	test-admin-locking \
 	test-repo-checkout-subpath	\
 	test-reset-nonlinear \
+	test-oldstyle-partial \
 	test-setuid \
 	test-delta \
 	test-xattrs \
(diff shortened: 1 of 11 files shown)


cgwalters:
Updated commit with `git add` for the test case.

cgwalters:
Rebased on master, fixed one compiler warning.

mbarnes:
Sorry, didn't even notice this PR til now.

Everything makes sense to me.  :+1:

cgwalters:
Merged.

PR #86, opened 2015-04-06
ostree_repo_checkout_tree_at: New API for checkouts

cgwalters:
rpm-ostree currently uses ostree_repo_checkout_tree(), which as a side
effect will use the uncompressed objects cache by default.  This is
rather annoying if you're using rpm-ostree on a server-side
repository, because if you then rsync the repo, you'll be syncing out
the uncompressed objects unless you exclude them.

We added the ability to disable the uncompressed cache in the
repository config to fix this, but it's better to allow application
control over this.  The uncompressed cache will in some future version
become opt in as well.

This new API further:
 - Drops the `GFile` usage in favor of `openat` APIs
 - Improves ergonomics by avoiding callers having to query the source
   `GFileInfo` (and carry around a copy of `OSTREE_GIO_FAST_QUERYINFO`)
 - Has a more extensible options structure

Per the comment, I rather crudely have the `ostree checkout` builtin
call both APIs to ensure some testing coverage.

However, I'd like to in the future have easier-to-set-up testing code
that calls `libtest.sh` to set up dummy data.

### DIFF
--- a/doc/ostree-sections.txt
+++ b/doc/ostree-sections.txt
@@ -277,6 +277,7 @@ ostree_repo_write_commit_detached_metadata
 OstreeRepoCheckoutMode
 OstreeRepoCheckoutOverwriteMode
 ostree_repo_checkout_tree
+ostree_repo_checkout_tree_at
 ostree_repo_checkout_gc
 ostree_repo_read_commit
 OstreeRepoListObjectsFlags
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -173,7 +173,7 @@ write_regular_file_content (OstreeRepo            *self,
 
 static gboolean
 checkout_file_from_input_at (OstreeRepo     *self,
-                             OstreeRepoCheckoutMode mode,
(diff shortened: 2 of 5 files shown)


mbarnes:
I definitely like the `OstreeRepoCheckoutOptions` struct, although I'm still trying to convince myself that exposing it on the public API is a good idea.  I guess the main benefit is getting to allocate it on the stack, I just worry that it may limit future extensibility.

Apart from that mild concern, it looks good to me.

cgwalters:
Were you thinking having a structure with public getters/setters?  That pattern feels heavyweight to me for this use case, but I could be convinced.  We do have padding members for future extensibility.

I think there's obviously not "one right answer" for extensible APIs - for simple functions it's usually best to just accept the cost of making a new one if the need appears, and then make the *new* one more extensible since obviously the area wasn't as simple as you thought.

mbarnes:
Right, I'm probably over-thinking it.  (Comes from too many years paying off technical debt.)

Progressive complexity as needed makes sense for this case.  So :+1:

cgwalters:
Cool, merged, thanks!

PR #87, opened 2015-04-08
reset: Don't enforce parent commits

cgwalters:
First, git doesn't do this, and whatever Linus thinks is right or
something.

Second specifically to OSTree, it's quite common to not have
intermediate commits.  If one wants to reset a ref in order to prune
data after a deployment, the parentage check will fail.


cgwalters:
See bug https://bugzilla.gnome.org/show_bug.cgi?id=747507

giuseppe:
:+1:

PR #88, opened 2015-04-09
refs: Use *at for writes, honor repo fsync flag

cgwalters:
I was looking at https://bugzilla.gnome.org/show_bug.cgi?id=738954
which wants us to ensure we chown() the refs.  As part of that,
I did a generic conversion to use `*at()` (which naturally gives
us more low level control so we can call `fchown` etc.

This patch also sneaks in a change to respect the repo's
`disable_fsync` flag - if fsync is not set, then we never
`fdatasync()` (unlike the `g_file_replace_contents()` default.  Also
unlike it, if fsync is enabled, we *always* sync even if the file
didn't exist.

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -60,19 +60,15 @@ add_ref_to_set (const char       *remote,
 }
 
 static gboolean
-write_checksum_file (GFile *parentdir,
-                     const char *name,
-                     const char *sha256,
-                     GCancellable *cancellable,
-                     GError **error)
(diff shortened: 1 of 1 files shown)


mbarnes:
Looks good.  The "always fsync" problem with `g_file_replace_contents()` is familiar to me.

I think this puts us pretty close to removing the `local_heads_dir` and `remote_heads_dir` `GFile` members of `OstreeRepo` after just a bit more porting work.

cgwalters:
Thanks, merged.

PR #89, opened 2015-04-14
Add an OstreeRepo signal for printing GPG status during pulls

mbarnes:
As discussed in https://github.com/projectatomic/rpm-ostree/pull/126.

The first commit is just some related cleanup work I had laying around.

Also, since I'm rewriting the logic again, shouldn't GPG verification on pulls fail if *any* signatures are bad?  I've preserved the old logic which is satisfied merely by one *good* signature, but it seems wrong to me.

### DIFF
--- a/doc/ostree-sections.txt
+++ b/doc/ostree-sections.txt
@@ -231,6 +231,7 @@ OstreeRepoRemoteChange
 ostree_repo_remote_change
 ostree_repo_remote_list
 ostree_repo_remote_get_url
+ostree_repo_remote_get_gpg_verify
 ostree_repo_get_parent
 ostree_repo_write_config
 OstreeRepoTransactionStats
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -962,13 +962,29 @@ scan_commit_object (OtPullData         *pull_data,
 
   if (pull_data->gpg_verify)
     {
-      if (!ostree_repo_verify_commit (pull_data->repo,
-                                      checksum,
-                                      NULL,
-                                      NULL,
-                                      cancellable,
-                                      error))
+      gs_unref_object OstreeGpgVerifyResult *result = NULL;
+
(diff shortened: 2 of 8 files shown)


cgwalters:
(If we're doing signals we could consider deprecating `OstreeAsyncProgress` in favor of a signal with a struct with all that data...I admit I sort of made it up and there are several things suboptimal about it)

mbarnes:
Addressed review feedback in the above "fixup" commits.

Gonna tack a bit more on to this pull request to help expedite https://github.com/projectatomic/rpm-ostree/pull/126.

cgwalters:
:+1: from me on the fixups

mbarnes:
Thinking let `ostree show` always print signature status, so even if GPG verification is disabled there's still a way to see signatures on commits.

cgwalters:
I have a mild concern about that in that for deployments who are currently protected by TLS (and who consider that sufficient), I'd like to not make the warning too strong.   For example, Red Hat Enterprise Linux configures both yum and rpm-ostree to use "pinned tls" with just one trusted CA.  I consider that quite strong security - it's not the same as GPG but it's good.

Something like:

`GPG: No signatures found, and remote is not configured with \`gpg-verify=true\`` ?

Ah, looks like currently we just don't print anything there?  That's OK by me, though I think in the future we should probably warn if somehow `gpg-verify=true` but there are no signatures (this shouldn't happen but just in case).

mbarnes:
Okay, I'll take that as another follow up.

cgwalters:
:+1: from me on both follow patches.  BTW don't forget to ping in the PR when there are new commits, otherwise there's no notification.

mbarnes:
Ah, I thought that was just for rebases.  I'll remember.

mbarnes:
Merged.
https://git.gnome.org/browse/ostree/commit/?id=6a7b9defb8d9d3ccc762fc11731092b4634d611c

PR #90, opened 2015-04-14
deploy: Use syncfs() on /, /boot instead of sync()

cgwalters:
This way we will actually catch errors, and we also avoid flushing
unrelated mount points.

I believe this change to be safe; OSTree only touches the `/` mount
point which covers the repository, the deployment roots (including
their copy of `/etc`), as well as `/boot`.  On initial system installs
we do relabel `/var` but that shouldn't happen at ostree time - any
new directories are taken care of via `systemd-tmpfiles` on boot.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -467,45 +467,39 @@ merge_etc_changes (GFile          *orig_etc,
 static gboolean
 checkout_deployment_tree (OstreeSysroot     *sysroot,
                           OstreeRepo        *repo,
-                          OstreeDeployment      *deployment,
-                          GFile            **out_deployment_path,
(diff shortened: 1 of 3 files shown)


cgwalters:
Rebased, back to a work in progress.  I'm trimmming out `fsync` calls along the deployment path.

mbarnes:
Looks good to me.  I appreciate the details given in the first couple commit messages, and the fd-relative porting looks pretty straight-forward as far as I can tell.

:+1:

cgwalters:
Thanks.  The last commit was very much a WIP but I think it's worth getting everything before it in now (and ideally letting it get some testing) and continue on other `fdatasync()` draining in further PRs.

cgwalters:
Merged this except the WIP which now lives in https://github.com/GNOME/ostree/pull/93

PR #92, opened 2015-04-17
sysroot: Cache an OstreeRepo instance

mbarnes:
Rather than returning a new `OstreeRepo` instance in each call to `ostree_sysroot_get_repo()`, cache one internally so the same instance is returned each time.

Need this to make https://github.com/projectatomic/rpm-ostree/pull/126 work.

Semantic changes like this tend to invite unforeseen side-effects, but as best as I can tell this seems okay.  Test cases pass and I can now see GPG status during `rpm-ostree upgrade`.


mbarnes:
Merged.
https://git.gnome.org/browse/ostree/commit/?id=a6bbcf2ba7b8c6e8c478bd776e1042bc7990b942

PR #94, opened 2015-04-21
Always request detached metadata for commits

mbarnes:
See https://bugzilla.gnome.org/748220

Always request detached metadata for commit objects, even if we already have the commit object.  This ensures we fetch any post facto detached metadata updates such as new GPG signatures.

Not the most elegant solution ever -- I'd say `enqueue_one_object_request()` is on the cusp of needing its own options enum -- but it seems to work.

The only weirdness I noticed is when the repo is otherwise up-to-date and you pull, you do get a metadata transfer count followed by `No upgrade available`, which to some might seem contradictory.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -101,6 +101,11 @@ typedef struct {
   OtPullData  *pull_data;
   GVariant    *object;
   gboolean     is_detached_meta;
+
+  /* Only relevant when is_detached_meta is TRUE.  Controls
+   * whether to fetch the primary object after fetching its
+   * detached metadata (no need if it's already stored). */
+  gboolean     object_is_stored;
 } FetchObjectData;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


mbarnes:
Thanks.  In light of the glitch I suffered in the demo this morning I think I'm going to give this at least a little more manual testing.  I still don't know what caused it, and I was running an `ostree` build with this patch in the demo.

mbarnes:
I still can't reproduce the glitch from the demo, but I've retested the patch and convinced myself that it wasn't the cause.  I also wrote a test case.  So I'll go ahead and merge.

It occurs to me the patch doesn't handle the case of *deleted* detached metadata, such as deleting all GPG signatures from a commit which may also delete the `.commitmeta` file.  The client will retain its `.commitmeta` file during a pull, even though it's been deleted on the server.  But this seems like a pretty remote corner case, so maybe not worth worrying about right now.

PR #95, opened 2015-04-23
Update the help for 'ostree reset'

miabbott:
As of `ostree` 2015.6, the help for the `reset` command does not state that it requires a REF and a COMMIT as arguments (even though the error message will tell you that).

This just makes the help a bit more useful in that regard.


cgwalters:
:+1:

cgwalters:
Merged.

PR #96, opened 2015-04-24
g_output_stream_splice: check correctly the error code

giuseppe:
Check that the error code < 0 as 0 is a success status.

The occurrences in ostree-repo-static-delta-*.c should never happen,
but change for completeness.


mbarnes:
Just to be pedantic, you could also assign the splice result to an `n_bytes_written` variable and then compare that in a separate statement.  That to me is the most readable form.

cgwalters:
I'm fine with that too.

giuseppe:
I pushed a new version that uses a temporary ```n_bytes_written```

Thanks for the review!

PR #97, opened 2015-04-27
giuseppe/gpg fixes

giuseppe:


giuseppe:
@mbarnes, just some minor things while I was looking on how to proceed with GPG support for static deltas

mbarnes:
Looks okay to me.  :+1: 

Note, I'm working on allowing GPG keys to be imported into repos so that certain keys are only valid for certain repos, and the repo knows where to find them.  That should alleviate if not eliminate the need for the `--gpg-homedir` option in most commands that currently have it.  But for consistency *now* I'm fine with adding the option to the `show` command.

giuseppe:
Thanks, I have pushed the patches now

PR #98, opened 2015-04-29
giuseppe/wip-deltas

giuseppe:
tests are still missing, I will add those once we agree on what is done so far.

I dropped any detached metadata from the static delta files, so no GPG signatures are there.  Instead I added the .sig file for the summary file and list the available delta files there.


giuseppe:
is there any pending request before this could get closed?  Beside rebase it onto master...

cgwalters:
:+1: from me, WDYT @mbarnes  ?

mbarnes:
Looks good to me.  Couple little niggles but nothing to get in a twist over.

:+1:

giuseppe:
rebased, used the new g_autoptr and pushed!

Thanks for the reviews

PR #100, opened 2015-05-04
repo: Stop creating "transaction" symlink

cgwalters:
This originally was a way that we detected the case where a pull was
interrupted.  Later, we added `.commitpartial` files which also cover
this case.

See also https://github.com/GNOME/ostree/pull/85

We still want to honor their existence (and unlink them) in case an
old version of ostree was in use, but I believe it's safe to stop
creating them now.

The only case where this would break is if you have a version of
ostree that predates commitpartial in your rollback history, but such
old versions are no longer in use by operating systems I support at
least.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1048,7 +1048,7 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
 
   g_return_val_if_fail (self->in_transaction == FALSE, FALSE);
 
-  /* We used to create a `transaction` symbolic link, but it's not
+  /* We used to create a `transaction` symbolic link, but it's now
    * obsoleted by the per-commit .commitpartial files.  We no longer
    * create it, but let's still read it if it exists, as well as
    * unlink it when we're done.
@@ -1056,7 +1056,7 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
   if (fstatat (self->repo_dir_fd, "transaction", &stbuf, AT_SYMLINK_NOFOLLOW) != 0)
     {
       if (errno == ENOENT)
-        ret_transaction_resume = TRUE;
+        ret_transaction_resume = FALSE;
       else
         {
           glnx_set_error_from_errno (error);
@@ -1064,7 +1064,7 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
         }
     }
   else 
-    ret_transaction_resume = FALSE;
+    ret_transaction_resume = TRUE;
 
   memset (&self->txn_stats, 0, sizeof (OstreeRepoTransactionStats));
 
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -1044,26 +1044,31 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
...
 
   g_return_val_if_fail (self->in_transaction == FALSE, FALSE);
 
-  if (self->transaction_lock_path == NULL)
-    self->transaction_lock_path = g_file_resolve_relative_path (self->repodir, "transaction");
-
-  if (g_file_query_file_type (self->transaction_lock_path, G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS, NULL) == G_FILE_TYPE_SYMBOLIC_LINK)
-    ret_transaction_resume = TRUE;
-  else
+  /* We used to create a `transaction` symbolic link, but it's not
```
mbarnes: s/not/now/ ?

src/libostree/ostree-repo-commit.c
```diff
@@ -1374,8 +1374,16 @@ ostree_repo_commit_transaction (OstreeRepo                  *self,
...
   self->in_transaction = FALSE;
 
-  if (!ot_gfile_ensure_unlinked (self->transaction_lock_path, cancellable, error))
-    goto out;
+  if (unlinkat (self->repo_dir_fd, "transaction", 0) != 0)
+    {
+      if (errno == ENOENT)
+        ;
+      else
+        {
```
mbarnes: Maybe just `if (errno != ENOENT)` ?

src/libostree/ostree-repo-commit.c
```diff
@@ -1044,26 +1044,31 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
...
-  else
+  /* We used to create a `transaction` symbolic link, but it's not
+   * obsoleted by the per-commit .commitpartial files.  We no longer
+   * create it, but let's still read it if it exists, as well as
+   * unlink it when we're done.
+   */
+  if (fstatat (self->repo_dir_fd, "transaction", &stbuf, AT_SYMLINK_NOFOLLOW) != 0)
+    {
+      if (errno == ENOENT)
+        ret_transaction_resume = TRUE;
```
mbarnes: Is this right?  Aren't we resuming a transaction if the symlink *does* exist?

cgwalters: Ooooh.  And that, friends, is why we have code review.

cgwalters:
Rebased, added fixup.

mbarnes:
Looks good.  :+1:

cgwalters:
Merged, thanks!

PR #101, opened 2015-05-04
sysroot: Add an API to lock

cgwalters:
If a system administrator happens to type `ostree admin upgrade`
multiple times, currently that will lead to a potentially corrupted
system.

I originally attempted to do locking *internally* in `libostree`, but
that didn't work out because currently a number of the commands
perform multi-step operations that all need to be serialized.  All of
the current code in `ostree admin deploy` is an example.

Therefore, allow callers to perform locking, as most of the higher
level logic is presently implemented there.

At some point, we can revisit having internal locking, but it will be
difficult.  A more likely approach would be similar to Java's approach
with concurrency on iterators - a "fail fast" method.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -45,6 +45,7 @@ testfiles = test-basic \
 	test-admin-deploy-uboot \
 	test-admin-instutil-set-kargs \
 	test-admin-upgrade-not-backwards \
+	test-admin-locking \
 	test-repo-checkout-subpath	\
 	test-reset-nonlinear \
 	test-setuid \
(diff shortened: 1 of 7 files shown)


cgwalters:
Depends on https://github.com/GNOME/libglnx/pull/7

mbarnes:
I should have read this one before https://github.com/GNOME/libglnx/pull/7.

Changes look good.   :+1:

Reminder to update `ostree-sections.txt` and *maybe* mention locking in the section header of the sysroot docs (but I won't hold you to that).

mbarnes:
:+1:

cgwalters:
Merged, thanks!

PR #102, opened 2015-05-05
Autocleanup cleanup

mbarnes:
Standardize on GLib's autocleanup macros throughout OSTree.

This also pretty significantly weakens our `libgsystem` dependency.

Sweeping cleanups like this can easily disrupt work-in-progress branches, so this can float around until we collectively decide it's a good time to merge it.  I expect to have to rebase this a few times.

It's mostly just a bunch of `sed` commands.  I can squash before merging if desired.


cgwalters:
Won't this break builds on GLib < 2.44?  The `g_auto()` syntax relies on the entire object chain using new style object definitions.  See https://bugzilla.gnome.org/show_bug.cgi?id=743640#c48

mbarnes:
That's what I thought too until I found this
https://git.gnome.org/browse/glib/tree/gio/gio-autocleanups.h

cgwalters:
I see.  Ok, but I get:

```
src/libostree/ostree-core.c: In function 'ostree_raw_file_to_content_stream':
src/libostree/ostree-core.c:428:3: error: unknown type name 'GInputStream_autoptr'
   g_autoptr(GInputStream) ret_input = NULL;
```

So that set may not be complete.

mbarnes:
Ah right, I forgot to pick up this:
https://git.gnome.org/browse/libglnx/commit/?id=013aa4feadb1fb7f2dd361b73f7d0fd623ae2daa

But if you're not digging GLib's way I don't mind backing that stuff out of `libglnx`.

cgwalters:
I'm cool with it, just wanted to make sure it worked on older glibs too.  I just pulled this and it builds for me on RHEL 7.1.

Well this will require some rebasing in existing PRs, we might as well pull the trigger now, there's not going to be a better time.

:+1:

mbarnes:
I was thinking of waiting until the D-Bus daemon gets merged.  We don't need to create any more friction for the Cockpit guys.  I'll just keep this updated until then.

cgwalters:
There's not much OSTree work for the rpm-ostree dbus daemon beyond the lock file bits, right?  Or am I missing something?

mbarnes:
Oh, pffft, you're right.  I was thinking the CLI changes but that's in rpm-ostree.

mbarnes:
Squashed some of the commits and rebased.

PR #104, opened 2015-05-07
sysroot: Add a try_lock() API

cgwalters:
The blocking locking API wasn't sufficient for use in the rpm-ostree
daemon; it really wants to know if the lock is held, then continue to
do other things (like service DBus requests), and get notification
when the lock is available.

Implement a higher level "loop until lock is available" method in the
`ostree admin` commandline.

Now you can see how returning a dirfd which we then
`g_file_monitor_directory(/proc/self/fd/%d)` on is pretty ugly...  I
debated a few other options like an async notification, but there'd
need to be a sane way to distinguish between "I got the lock now"
versus "I waited for the lock".

### DIFF
--- a/doc/ostree-sections.txt
+++ b/doc/ostree-sections.txt
@@ -369,6 +369,8 @@ ostree_sysroot_get_path
 ostree_sysroot_load
 ostree_sysroot_lock
 ostree_sysroot_try_lock
+ostree_sysroot_lock_async
+ostree_sysroot_lock_finish
 ostree_sysroot_unlock
 ostree_sysroot_get_fd
 ostree_sysroot_ensure_initialized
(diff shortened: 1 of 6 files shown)


mbarnes:
I don't know, the `try_lock()` API seems a little overly complex to me from a user's pov.

I kinda wonder if this could be split into a simple `try_lock()` function that works more like a mutex "try lock", and maybe a `lock_async()` / `lock_finish()` that does the file monitoring for you.

cgwalters:
Ok, :arrow_up: is simpler.  I was overthinking :thought_balloon: :boom: it before.

cgwalters:
Good point!  Fixup in :arrow_up:

mbarnes:
Looks good to me.  :+1:

cgwalters:
This was pushed.

PR #105, opened 2015-05-07
giuseppe/httpd port

giuseppe:


mbarnes:
Looks okay to me.  Is it worth having some test coverage for this?  You could probably just tweak one of the existing shell scripts.

mbarnes:
:+1: on the indentation cleanups at least.

giuseppe:
pushed with a test case and the amended commit message

PR #106, opened 2015-05-08
Moar GPG utils

mbarnes:
Snagged some handy GPG utilities from Seahorse.
https://git.gnome.org/browse/seahorse/tree/pgp/seahorse-gpgme-data.c

Found a use for `ot_gpgme_data_output()` in our `sign_data()` function, but mainly I need `ot_gpgme_data_input()` for the other GPG branch I'm working on.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3108,7 +3108,6 @@ sign_data (OstreeRepo     *self,
   gpgme_key_t key = NULL;
   gpgme_data_t commit_buffer = NULL;
   gpgme_data_t signature_buffer = NULL;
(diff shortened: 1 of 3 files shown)


cgwalters:
LGTM aside from the above comment :+1:

mbarnes:
Stef and I talked a little about Seahorse sharing back when I was on Evolution.  Those particular functions expose GPGME structs (that's their whole point) so GLib isn't the right place for them, nor even `libgcr` since that library is introspected.  But maybe a git submodule type deal like `libglnx` could work.  I'll talk to Stef.

cgwalters:
Ok, no need to block on it though.

PR #107, opened 2015-05-11
Add ostree remote gpg-import

mbarnes:
Implements what we agreed on in https://bugzilla.gnome.org/729592

Adds a new `remote` subcommand to the CLI:
````
ostree remote gpg-import {--keyring=FILE} [--stdin] NAME [KEY-ID...]
````

Also adds a convenience option `--gpg-import=FILE` to `ostree remote add`, which is equivalent to
````
ostree remote gpg-import --keyring=FILE
````

Remote-specific keyring files are named after their remote:
````
$REPODIR/NAME.trustedkeys.gpg
````

No new public APIs in `libostree`.

**NOTE:** For backward-compatibility, the old GPG verification behavior is retained.  So OSTree will still honor any existing keys in `$DATADIR/ostree/trusted.gpg.d/` or `$OSTREE_GPG_HOME` across all repos and all remotes.  If that's not what we want, I can rip that out easily enough.

As an aside, I think `OstreeChainInputStream` should be promoted to GIO.  That came in quite handy for the CLI part.


mbarnes:
Strange, the commits here are somehow listed out of order.

Should be all the `repo:` commits first, followed by the `ostree:` commits, followed by the `tests:` commit.

giuseppe:
LGTM!

cgwalters:
All minor bits aside, :+1: from me.

mbarnes:
As the review comments were pretty minor, I pushed a rebase.

PR #108, opened 2015-05-12
giuseppe/syntax check fixes

giuseppe:
a few fixes for syntax-check


mbarnes:
The syntactic rules look okay.  I think you could also prohibit all the `gs_free_*` macros.

The GNU specific bits in `maint.mk` I assume are just not applicable to OSTree?  If that's all it is, :+1:

giuseppe:
pushed.  Thanks for the review

PR #109, opened 2015-05-12
giuseppe/disable deltas

giuseppe:
give users the possibility to disable static deltas


mbarnes:
The changes all look good to me.  Long as @cgwalters has no objection to the option itself, :+1: from me.

cgwalters:
LGTM, but would slightly prefer the last patch to be squashed into the first.

giuseppe:
pushed with the last patch squashed into the first.  Thanks for the review!

PR #110, opened 2015-05-13
admin: Use locking for most sysroot commands

cgwalters:
The previous commit introduced locking for `ostree admin deploy`, but
we do expect people to possibly accidentally do e.g.
`ostree admin upgrade` concurrently.

Using consistent locking in the admin commands will help rpm-ostree.

### DIFF
--- a/src/ostree/ot-main.c
+++ b/src/ostree/ot-main.c
@@ -28,6 +28,7 @@
 #include <string.h>
 
 #include "ostree.h"
+#include "ot-admin-functions.h"
 #include "ot-main.h"
 #include "otutil.h"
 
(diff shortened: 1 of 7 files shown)


src/ostree/ot-main.c
```diff
@@ -362,6 +363,12 @@ ostree_admin_option_context_parse (GOptionContext *context,
       exit (EXIT_SUCCESS);
     }
 
+  if ((flags & OSTREE_ADMIN_BUILTIN_FLAG_UNLOCKED) == 0)
+    {
+      if (!ot_admin_sysroot_lock (sysroot, error))
```
mbarnes: As I had to look this up, I suggest a comment here to clarify the lock is implicitly released when the `OstreeSysroot` instance is finalized.

mbarnes:
Aside from my comment suggestion, :+1:

cgwalters:
Merged, thanks!

PR #111, opened 2015-05-14
repo: Bump mtime any time we write a ref

cgwalters:
External daemons like rpm-ostree want push notification any time a
change is made by an external entity.  inotify provides notification,
but a problem is there's no easy way to monitor all of the refs.

In the past, there has been discussion of opt-in recursive timestamps:
https://lkml.org/lkml/2013/4/5/307

But in today's world, let's just bump the mtime on the repo itself, as
a central inotify point.


petervo:
@cgwalters I was about 5mins away from opening a pr with a version of this. LGTM.

PR #113, opened 2015-05-15
admin: Ensure instutil commands and usage help don't grab lock

cgwalters:
When I was introducing the `_UNLOCKED` flag, I only audited
subcommands of `ostree admin`, but I missed that `ostree admin
instutil` also used the option parsing.  Those are only used by
Anaconda today so we can ignore them for locking purposes.

Also, the usage help generation was grabbing the lock unnecessarily.


mbarnes:
I missed those as well in my review.  :confused: 

:+1:

PR #116, opened 2015-05-26
sysroot: Sort returned boot loader configs

cgwalters:
This patch is not ready for merging, I'd like to have more tests.  Will mark WIP.

I haven't done a full dig through the history, but it seems quite
possible right now we've been relying on inode enumeration
order for generating bootloader configuration.

Most of the time, newer inodes (i.e. later written files) will win.
But that's obviously not reliable.

Fix this by sorting the returned configuration internally.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -75,6 +75,10 @@ insttest_DATA = tests/archive-test.sh \
 	tests/corrupt-repo-ref.js \
 	$(NULL)
 
+insttest_SCRIPTS += \
+	tests/syslinux-entries-crosscheck.py \
+	$(NULL)
+
 gpginsttestdir = $(pkglibexecdir)/installed-tests/gpghome
 gpginsttest_DATA = tests/gpghome/secring.gpg \
 	tests/gpghome/trustdb.gpg \
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -361,6 +361,35 @@ _ostree_sysroot_read_current_subbootversion (OstreeSysroot *self,
   return ret;
 }
 
+static gint
+compare_boot_loader_configs (OstreeBootconfigParser     *a,
+                             OstreeBootconfigParser     *b)
+{
+  const char *a_version = ostree_bootconfig_parser_get (a, "version");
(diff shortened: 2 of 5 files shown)


mbarnes:
I'm guessing you know about this already, but `scandirat()` looks to me like it could be more convenient than `opendirat()` for this particular case.

cgwalters:
I didn't know about that API, could be handy in the future.  But in this case, we don't actually want the filename order, we do want to respect the `version` key, as that's what links the total ordering on the deployments.  See this bit in `ostree-sysroot-deploy.c`:

```
  version_key = g_strdup_printf ("%d", n_deployments - ostree_deployment_get_index (deployment));
  ostree_bootconfig_parser_set (bootconfig, "version", version_key);
```

cgwalters:
See also http://www.freedesktop.org/wiki/Specifications/BootLoaderSpec/ from which this logic derives...although the current status quo is that grub2 doesn't quite implement it, and syslinux/uboot don't.  https://www.redhat.com/archives/anaconda-devel-list/2014-July/msg00002.html

cgwalters:
Ok, taking the WIP label off this, I think it's ready for review.

giuseppe:
:ok: from me

mbarnes:
Just a passing thought: `compare_deployments_by_boot_loader_version_reversed()` strikes me as a potentially handy public function `ostree_deployment_compare()`.  Is there any other way we'd want to sort deployments?

That aside, :+1: from me as well.

cgwalters:
I'm not sure it makes sense as public API because the version attribute is assigned internally; what the caller passes is a list whose order they've chosen.  But will keep in mind.

Pushed to master.  This issue is going to need more followup and publicity unfortunately though =(

cgwalters:
I was wondering - this bug appeared to have existed approximately forever, but I couldn't believe we hadn't seen it over the last few years.  It turns out it was actually introduced by https://git.gnome.org/browse/ostree/commit/?id=68ce55420211698d14065c20891a517102db3016

Because `GFileEnumerator` sorts by inode by default: https://git.gnome.org/browse/glib/tree/gio/glocalfileenumerator.c?id=2.44.1#n344

But when we switched to the raw `readdir()` wrapper we lost this sorting.

PR #117, opened 2015-05-27
giuseppe/fix summary crypt

giuseppe:


cgwalters:
(And what I think would be nice is writing to the local commit in some out of band metadata (xattrs?) a cached version of the signature validation, so you can still know that the commit was transitively signed.  This is nontrivial though)

cgwalters:
Can you take a look at https://github.com/GNOME/ostree/pull/118 and let's land it before we do this?

giuseppe:
I've pushed some new commits.  I've added a gpg-verify-summary option which can be used when gpg-verify is used too.  I had to fix some rebase conflicts, so I pushed again the whole series without fixup commits.

giuseppe:
any comments on the fixups?

cgwalters:
I know you said you were working on it, but didn't realize you'd uploaded new patches - remember there's no notification when you push new changes to a PR, so please add a comment like "Added fixups, please re-review" or so.  Will take a more detailed look soon.

One thing that jumps out at me - the default for gpg-verify-summary has to be `FALSE` if it isn't now, for backwards compatibility.

giuseppe:
I have changed the patches to have ```gpg-verify-summary=FALSE``` by default plus some other cleanups.

cgwalters:
I think this series looks OK now, but I'd like to do another pass with it after the merge is done.

PR #119, opened 2015-06-02
tests: Run all tests through a randomized readdir()

cgwalters:
Having undefined (but in practice rarely changing) ordering for
`readdir()` ended up screwing us over for bootloader config
generation; see https://bugzilla.redhat.com/show_bug.cgi?id=1226520

Let's make things significantly more likely to fail more quickly in
the future if similar bugs are introduced.  We accomplish this by
introducing a little `LD_PRELOAD` library that randomizes the results
of `readdir()`.

### DIFF
--- a/tests/readdir-rand.c
+++ b/tests/readdir-rand.c
@@ -1,0 +1,145 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright (C) 2015 Colin Walters <walters@verbum.org>.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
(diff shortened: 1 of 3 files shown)


cgwalters:
I did verify that reverting https://github.com/GNOME/ostree/commit/3300ee6259d191199f96b5f4e04ccfded3d11b86 on top of this, combined with the syslinux validation very quickly caused the test to fail.

james-antill:
The intent seems great, and given it works and is just used for testing +1.

But I'd worry about readdir_r/rewinddir/etc.

The other solution I've seen is to "always" sort readdir() returns (the only times you really don't want to are directories that can be big, and you know contain data that doesn't rely on ordering).

tests/readdir-rand.c
```diff
@@ -0,0 +1,145 @@
...
+      g_once_init_leave (&initialized, 1);
+    }
+}
+
+struct dirent *
+readdir (DIR *dirp)
+{
+  struct dirent *(*real_readdir)(DIR *dirp) = dlsym (RTLD_NEXT, "readdir");
+  struct dirent *ret;
+  gboolean doloop = TRUE;
```
mbarnes: Had to read this a few times to grok it.  I get it now.  Clever!  FWIW, the logic made more sense to me once I mentally swapped the name `doloop` with `cache_another`.

cgwalters:
Pushed a fixup :arrow_up: which should address both review comments.

Always sorting would be pretty invasive in the codebase, and would probably have a noticeable performance hit when dealing with the loose object directories.  I think this should give us enough coverage to be confident things work, while not affecting the real runtime.

(It would obviously be better if we had stronger test coverage of running as uid 0, as we're not testing the actual grub2 commands, but for now this is fairly good)

mbarnes:
LGTM  :+1:

cgwalters:
Merged.

PR #120, opened 2015-06-05
misc: New script to actively fix GRUB2 bootloader ordering

cgwalters:
This is a follow up to https://mail.gnome.org/archives/ostree-list/2015-June/msg00000.html

My plan here is to stick this script into a Docker container, and
allow people to pull and run it that way.


cgwalters:
No one ever reviewed this, it's pretty obsolete now anyways.

PR #121, opened 2015-06-07
ostree remote refs

mbarnes:
Implements the `ostree remote refs` command I proposed in https://bugzilla.gnome.org/749456 but uses the remote repo's `summary` file.  Any repo being served over HTTP really *ought* to have one.

The first 4 commits are just some refactoring that I think are worth adding regardless of the command.

### DIFF
--- a/Makefile-ostree.am
+++ b/Makefile-ostree.am
@@ -80,6 +80,7 @@ ostree_SOURCES += \
 	src/ostree/ot-remote-builtin-gpg-import.c \
 	src/ostree/ot-remote-builtin-list.c \
 	src/ostree/ot-remote-builtin-show-url.c \
+        src/ostree/ot-remote-builtin-refs.c \
 	$(NULL)
 
 ostree_bin_shared_cflags = $(AM_CFLAGS) -I$(srcdir)/src/libotutil -I$(srcdir)/src/libostree -I$(srcdir)/src/ostree \
--- a/doc/ostree-remote.xml
+++ b/doc/ostree-remote.xml
@@ -63,6 +63,9 @@ Boston, MA 02111-1307, USA.
             <cmdsynopsis>
                 <command>ostree remote gpg-import</command> <arg choice="opt" rep="repeat">OPTIONS</arg> <arg choice="req">NAME</arg> <arg choice="opt" rep="repeat">KEY-ID</arg>
             </cmdsynopsis>
+            <cmdsynopsis>
+                <command>ostree remote refs</command> <arg choice="req">NAME</arg>
+            </cmdsynopsis>
     </refsynopsisdiv>
 
     <refsect1>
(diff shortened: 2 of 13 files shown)


mbarnes:
**Rebased:** Did a little `OstreeMetalink` refactoring to make the API more convenient for fetching the `summary` file  (its only use case at present).  Added metalink support to `ostree_repo_remote_fetch_summary()` and wrote a small `remote refs` test case that uses a metalink.

mbarnes:
Rebased again to implement summary verification if GPG signatures are present.

But this uncovered some issues in `OstreeMetalink`, so the patch here needs to go in first:
https://bugzilla.gnome.org/750950

I made the signature verification automatic in `ostree_repo_remote_fetch_summary()`.  I'm only 80% sure that's a good idea, but I can't call `_ostree_repo_gpg_verify_with_metadata()` from the CLI layer.

cgwalters:
Should this come before or after https://github.com/GNOME/ostree/pull/117 ?

cgwalters:
A long time ago, OSTree used to have a per-remote cache directory.  See https://git.gnome.org/browse/ostree/commit/?id=2a0601efc790a0c8f783043f2db682eec9ceffaa which dropped it.

But I do wonder with this whether it would make sense to reintroduce it.  Then the architecture is clearer - `ostree_repo_pull()` could have an mode like `_FETCH_SUMMARY_ONLY`, and it gets GPG verified and stored in `repo/remote-cache/$remotename/summary` or so.

mbarnes:
@giuseppe mentioned something about rebasing #117 on this in a review comment I can't find anymore.  Doesn't matter much to me.

giuseppe:
yes, I think it is cleaner to rebase https://github.com/GNOME/ostree/pull/117 on top of this.

I've not tried to rebase it yet, but hopefully there won't be many conflicts.  If it is fine, I can do the refactoring after merging https://github.com/GNOME/ostree/pull/117 and use the new functions introduced by this series.

To make things more interesting, "remote refs" should honor ```gpg-verify-summary``` that is introduced by https://github.com/GNOME/ostree/pull/117, but obviously this  can be done only after https://github.com/GNOME/ostree/pull/117 is merged too.

mbarnes:
@giuseppe If you think you'll rebase #117 on this, maybe you just could add the `gpg-verify-summary` check for "remote refs" as part of your branch?

giuseppe:
@mbarnes yes, sure

giuseppe:
I have rebased my work on top of this branch:

https://github.com/giuseppe/ostree/tree/giuseppe/fix-summary-crypt-rebased

mbarnes:
Reviewed your branch.  LGTM overall.  A few minor comments, but no blockers.

giuseppe:
I have addressed some of the comments in the following fixup's patches.  I have not worked on refactoring the GPG signatures policy.  I agree with you that it will require some more work to be configured and allow users to choose how many signatures must be valid.

mbarnes:
Fixups look good.  :+1:

The signature policy thing doesn't need to block the merge.  It's just something I keep noticing and forgetting to mention.

giuseppe:
merged!  Could you close the merge request, it was pushed as part of giuseppe/fix-summary-crypt-rebased?

mbarnes:
Cool!

PR #122, opened 2015-06-27
pull: Error if gpg=true and summary is 404, add more tests

cgwalters:
I did a quick audit pass through the pull code.  What I focused on the
most is the case where `gpg-verify-summary=true`, and in particular
where `gpg-verify=false` too.  This should be a valid and secure
configuration.

The primary change here is to error out very quickly if either
`summary` or `summary.sig` are 404.  Previously, we'd only error out
if we were processing deltas.

Expand the existing test case to cover this, plus invalid summary and
invalid sig.  (The test case was failing with current git master too).


giuseppe:
:+1:

cgwalters:
Merged, thanks.

PR #123, opened 2015-06-29
build: autogen.sh no longer silently fails if gtk-doc is missing.

jeremycline:
If gtk-doc wasn't installed, the autogen.sh script would silently stop when it executed `which gtkdocize` because of `set -e`. This fixes that and makes the gtk-doc package optional. I'm brand new to autotools so I apologize in advance if I've done this all wrong.

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -16,7 +16,7 @@ set -e
 
 mkdir -p m4
 
-GTKDOCIZE=$(which gtkdocize 2>/dev/null)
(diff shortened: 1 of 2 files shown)


cgwalters:
I tweaked the commit message to match style, and a few minor tweaks to the code, and pushed. Thanks for the patch!

https://github.com/GNOME/ostree/commit/f0a02fbf20ebf4b1f2e248fee24a0babb2b22c03

PR #127, opened 2015-07-22
libostree: new API ostree_repo_remote_list_refs

giuseppe:
The new API permits to query a remote repository summary file and
retrieve the list of available refs.


src/libostree/ostree-repo.h
```diff
@@ -291,6 +291,12 @@ gboolean      ostree_repo_list_refs (OstreeRepo       *self,
                                      GCancellable     *cancellable,
                                      GError          **error);
 
+gboolean ostree_repo_remote_list_refs (OstreeRepo       *self,
+                                       const char       *remote_name,
+                                       GHashTable      **out_all_refs,
+                                       GCancellable     *cancellable,
+                                       GError          **error);
+
```
mbarnes: You could also just return the `GHashTable` directly here (or `NULL` on error).

Either way works though, just wanted to mention it.

src/ostree/ot-remote-builtin-refs.c
```diff
@@ -51,39 +51,17 @@ ot_remote_builtin_refs (int argc, char **argv, GCancellable *cancellable, GError
...
-
-      ref_map = g_variant_get_child_value (summary, 0);
-
-      /* Ref map should already be sorted by ref name. */
-      g_variant_iter_init (&iter, ref_map);
-      while ((child = g_variant_iter_next_value (&iter)) != NULL)
+      GHashTableIter hash_iter;
+      gpointer key, value;
+      g_hash_table_iter_init (&hash_iter, refs);
+      while (g_hash_table_iter_next (&hash_iter, &key, &value))
```
mbarnes: I think you're losing alphabetical ordering here.  You'll need to extract the keys as a list and sort them.

Or there's also `GTree` which never gets any love.

mbarnes:
:+1: to the idea.  Review comments above.

giuseppe:
thanks for the review.  I noticed also a regression in the tests suite, I am going to address that as well.

giuseppe:
the last commit is still WIP, I've pushed it just to get some early feedback.

cgwalters:
Hm, I'd been thinking of this as `OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY` or so.  (Note we need to fix that to actually be flags and not an enum), something like:

```
diff --git a/src/libostree/ostree-repo.h b/src/libostree/ostree-repo.h
index 84b0deb..dc74582 100644
--- a/src/libostree/ostree-repo.h
+++ b/src/libostree/ostree-repo.h
@@ -690,7 +690,8 @@ gboolean ostree_repo_prune (OstreeRepo        *self,
  */
 typedef enum {
   OSTREE_REPO_PULL_FLAGS_NONE,
-  OSTREE_REPO_PULL_FLAGS_MIRROR
+  OSTREE_REPO_PULL_FLAGS_MIRROR = (1 << 0),
+  OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY = (1 << 1)
 } OstreeRepoPullFlags;
 
 gboolean ostree_repo_pull (OstreeRepo             *self,
```

Then clients can `ostree_repo_load_commit()` as normal.  (And in this situation, `OSTREE_REPO_COMMIT_STATE_PARTIAL` would be true).  Also `ostree_repo_read_commit_detached_metadata()`.

mbarnes:
What would happen if both pull flags were passed?  They seem to conflict.

cgwalters:
I don't believe `MIRROR` and `COMMIT_ONLY` conflict in a technical sense, although I can't think of an immediate use case.

giuseppe:
@mbarnes, @cgwalters: Thanks for the review!  Good idea indeed, I am going to throw away ```ostree_repo_prefetch_commit``` and implement the new ```OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY``` pull flag

cgwalters:
One small comment, otherwise looks simple and clean to me.

cgwalters:
In order to make `ostree fsck` happy we needed:

```
index 6c98a19..aa8e635 100644
@@ -2167,7 +2167,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
     }
 
   /* iterate over commits fetched and delete any commitpartial files */
-  if (!dir_to_pull)
+  if (!dir_to_pull && !pull_data->is_commit_only)
     {
       g_hash_table_iter_init (&hash_iter, requested_refs_to_fetch);
       while (g_hash_table_iter_next (&hash_iter, &key, &value))
```

I've squashed this in.

cgwalters:
Ok, I did some commit squashing and pushed.  Thanks!

PR #128, opened 2015-07-28
static delta fixes

giuseppe:


cgwalters:
Looks fine to me.  Would have liked to see test coverage, but we can add it later.

PR #130, opened 2015-07-29
ostree: add colored output to tty

giuseppe:
DEPENDS FROM: https://github.com/GNOME/libglnx/pull/10


giuseppe:
I associated that with the git diff output where added lines are green and removed ones are red.  If it is confusing, I'll drop colors for ostree-diff.c.  Do you see any other place where we could use colors beside "commit %s" in git log and the error prompt?

PR #133, opened 2015-08-14
Giuseppe/rollsum test

giuseppe:


giuseppe:
@cgwalters, @mbarnes, do these changes make sense to you?

mbarnes:
Still coming up to speed on rolling checksums in general, but the changes here look okay to me.

cgwalters:
Hmm...there's nothing stopping two blocks of different size from having the same CRC!

So I think what we actually need to do here is:

```
if (to_offset != from_offset)
  continue;
```

right?

giuseppe:
True!  I was able to reproduce it with a crafted test case.  Pushed a new commit with the fix and a new test

cgwalters:
Nice test case!  I reworked the branch a little bit so that the bugfix is separate from the tests, to preserve bisectability.  Merged :+1:

PR #134, opened 2015-09-01
pull: Go back to using one main context

cgwalters:
xdg-app was hanging for me with v2015.8, but worked with v2015.7.
I narrowed things down to the GMainLoop/context commit, in which
we started pushing a temporary main context for synchronous
requests internally.

That's never really going to work with libsoup - there needs
to be a single main context which works on the socket.  Furthermore,
clients couldn't get progress messages that way.

For *other* internal uses where we added APIs that talk to the remote
repo, we cleanly push a temporary main context.

(Note that I kind of snuck in a change here around the GError handling
 in pulls that isn't strictly related but came up in testing)

### DIFF
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -731,6 +731,10 @@ fetch_uri_sync_on_complete (GObject        *object,
   data->done = TRUE;
 }
 
+/* Synchronously request a URI - will iterate the thread-default main
+ * context for historical reasons.  If you don't want that, push a
+ * temporary one.
+ */
 gboolean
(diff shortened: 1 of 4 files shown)


mbarnes:
Would it help to give the `SoupSession` in `OstreeFetcher` its own dedicated thread and main loop?  Then its main context wouldn't interfere with whatever you need to do with the public API.  I've done that in the past to get around similar issues with competing main contexts.

cgwalters:
We're unfortunately a bit tied in with backwards compatibility with respect to the `OstreeAsyncProgress` API.

cgwalters:
Merged as is.

PR #136, opened 2015-09-10
WIP: archive-{export,import}: New builtins

cgwalters:
I got an IRC request for support for "single file" *transport* of OSTree commits.

### DIFF
--- a/src/ostree/ot-builtin-archive-import.c
+++ b/src/ostree/ot-builtin-archive-import.c
@@ -1,0 +1,259 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright (C) 2015 Colin Walters <walters@verbum.org>
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ */
+
+#include "config.h"
+
+#include "ot-main.h"
+#include "ot-builtins.h"
+#include "ostree.h"
+#include "ostree-cmdprivate.h"
+#include "otutil.h"
+
+#ifdef HAVE_LIBARCHIVE
+#include <archive.h>
+#include <archive_entry.h>
+#include "ostree-libarchive-input-stream.h"
+#endif
+
+static GOptionEntry options[] = {
+  { NULL }
+};
+
+typedef struct {
+  OstreeRepo *repo;
+  char *commit_checksum;
+} ArchiveState;
+
+#ifdef HAVE_LIBARCHIVE
+
+static void
+propagate_libarchive_error (GError      **error,
+                            struct archive *a)
+{
+  g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+               "%s", archive_error_string (a));
+}
+
+static gboolean
+parse_object_name (const char *s,
+                   char **out_checksum,
+                   OstreeObjectType *out_objtype,
+                   GError **error)
+{
+  gboolean ret = FALSE;
+  const char *dot;
+  
+  dot = strrchr (s, '.');
+  if (!dot)
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_INVALID_FILENAME,
+                           "Missing '.' in object filename");
+      goto out;
+    }
+
+  if (strcmp (dot, ".file") == 0)
+    *out_objtype = OSTREE_OBJECT_TYPE_FILE;
+  else if (strcmp (dot, ".dirtree") == 0)
+    *out_objtype = OSTREE_OBJECT_TYPE_DIR_TREE;
+  else if (strcmp (dot, ".dirmeta") == 0)
+    *out_objtype = OSTREE_OBJECT_TYPE_DIR_META;
+  else if (strcmp (dot, ".commit") == 0)
+    *out_objtype = OSTREE_OBJECT_TYPE_COMMIT;
+  else
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_FILENAME,
+                   "Invalid object suffix: %s", dot);
+      goto out;
+    }
+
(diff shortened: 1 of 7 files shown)


giuseppe:
should the file be a .zip?  Could we reuse static-delta generate --empty here?

cgwalters:
That's a really good point, it should likely be just an arhcive of `static-delta --empty`.

src/ostree/ot-builtin-archive-import.c
```diff
@@ -0,0 +1,259 @@
...
+            break;
+
+          g_byte_array_append (buf, tmp_buf, read_size);
+        }
+      
+      {
+        g_autoptr(GVariant) metadata
+          = g_variant_new_from_data (ostree_metadata_variant_type (objtype),
+                                     buf->data, buf->len, TRUE, NULL, NULL);
+        if (!ostree_repo_write_metadata_trusted (self->repo, objtype, checksum, metadata,
```
alexlarsson: Are we sure its trusted here?

src/ostree/ot-builtin-archive-import.c
```diff
@@ -0,0 +1,259 @@
...
+          self->commit_checksum = g_strdup (checksum);
+        }
+
+      g_byte_array_unref (buf);
+    }
+  else
+    {
+      g_autoptr(GInputStream) archive_stream = _ostree_libarchive_input_stream_new (zipfile);
+
+      if (!ostree_repo_write_content_trusted (self->repo,
```
alexlarsson: We shouldn't just trust the imput here, should we. It may lie about its checksum...

alexlarsson:
So, here is my goal. Given a repository (either the local one for installed xdg-app, a remote one, or a local mode=z2 repo that you just commited your newly built app into) I want to be able to create a single file, representing this commit. The advantage of the single file is that it is easy to transport this in various ways, such as a uploading it to a target device using adb, storing on a FAT usb stick, or sticking it for easy download on a ftp server.

Once a user somehow has gotten hold of this image he should be able to install it to the local xdg-app ostree repo so he can run it. This involves importing all the objects from the commit, but also pointing the right branch to it. This implies that the file should also contain the branch name that was used. We also need to do proper sha1 checksumming of all objects, and at least optionally verify that the commit is signed with a trusted key. As a bonus, it would be nice if the file could contain a uri to a repository where updates can be found so that we can set up the origin to get automatic updates if this is available.

The patch seems to only add a new commandline option, but ideally I would like this available in libostree so that xdg-app can call it automatically.

alexlarsson:
I thought initially about using empty deltas, but as it looks now thats not quite possible, because deltas produce multiple files. First the delta metadata, and then the N delta parts (due to the max chunk size). There is also min-fallback-size that makes it skip large files. With the right options we can make it contain everything in only one chunk, but there will always be at least two files.

That said, I'm sure it wouldn't be hard to put the two in the same file, as well as some extra optional metadata (such as the origin repo and origin refname)

giuseppe:
the origin repo and the origin refname could go into the additional metadata of static deltas. I am not sure if ostree itself should set up the configuration once the exported commit has been received or let the user do it (xdg-app in this case).  Then it is a matter to have min-fallback-size=inf and pack all in a single file.

cgwalters:
Packing things into a single file is really easy with libarchive.

alexlarsson:
I don't see why libarchive should be needed at all. Why can't we just (optionally) add the parts to the superblock gvariant.

cgwalters:
Right now deltas are split into metadata and content block chunks as Giuseppe mentioned.  We could define a new format, but it seems simplest to just pack what we have now into a zip/tar, right?

Adding it all to the static delta superblock seems pretty sane, yeah.  The other important detail though is around signatures.  Right now static delta signatures are covered by the summary file.  We could generate one and pack it in, then have the signature stuff support verifying from local files?

Things might be a lot simpler if `ostree-repo-pull.c` knew how to operate on this single-file snapshot.  Which seems like it would be a easiest if it as a zip file or something that was a subset of the regular archive format, plus the other metadata you want like the target origin?

alexlarsson:
I'm looking at just putting the parts into the superblock atm. Looks simple.

alexlarsson:
The changes in https://github.com/GNOME/ostree/pull/139 do support thins kind of single-file static delta in ostree-repo-pull.c. However, i'm not sure how interesting it would be to do that if we're using e.g. a custom format with a zipfile containing a summary file and the delta metadata+parts. In that case it would probably be implemented more like ostree_repo_static_delta_execute_offline().

cgwalters:
If you want GPG verification (and you said you did) that's going to impact this a lot.  Right now OSTree's GPG verification is all based around a keyring per remote.  The nice thing about this is that it's easier to do nontrivial (but secure) configuration in a remote config file.  Things like TLS pinning.

Maybe one option to help bootstrap is to support embedding a public key with the file itself (kind of like what yum repos do with `gpgkey=http://example.com/repo/key.asc`.  Then if the user establishes a trust relationship with that key like:

`xdg-app install http://example.com/apps/foo '1CEC 7A9D F7DA 85AB EF84  3DC0 A866 D7CC AE08 7291'`

Where that fingerprint was copied/pasted from a known source (https://keybase.io/walters or something else).  Then xdg-app could import it into the remote's keyring too.

(These single file apps might also support embedding a public CA cert for certificate pinning too, etc.)?

alexlarsson:
Yeah, having a gpg signature + key in the file is a good start. There are then multiple ways to install the file:
 * Explicitly say --no-gpg-verify
 * Supply the key fingerprint to verify the key
 * Find the public key elsewhere and put it in /usr/share/ostree/trusted.gpg.d or the users keyring
 * Give a path to the key file

If we also include a repo uri in the file we can automatically set up a origin repo with a trusted gpg key for the app so that later updates work.

It seems to me like we could put all that in the extended metadata in the delta superblock, so the format would work as is. We just need offline-apply to find the gpg key and use it to verify the commit in the deltafile.

cgwalters:
Having the signature in the superblock is tricky because the signature has to *cover* the superblock.
Detached signatures are so much easier....

alexlarsson:
But is it *really* necessary to sign the superblock? I mean, in my case the superblock itself will be pretty simple (delta from empty, everything included), so having a signature on the commit object, plus a key in the superblock that you can optionally verify with a fingerprint would probably be enough, no?

cgwalters:
Theoretically, relying on commit signatures would work, except right now the delta apply code does not verify the checksum of each object, rather we assume that anyone who wants GPG signed deltas will be using the summary signatures.

We could investigate per-delta signatures, or alternatively add a mode where the object checksums are verified for deltas.

cgwalters:
Note too the superblock does presently contain the commit object.

giuseppe:
my suggestion is to include the ```summary``` file as it is, if a ```summary.sig``` which contains the static delta that is going to be used, or alternatively filter out anything that is not relevant from the ```summary``` file before sign and include it.  We will reuse what we already have and won't duplicate the same functionality in different places (that was one of the main reasons to have a ```summary``` file which lists the static deltas as well)

cgwalters:
See also https://github.com/appc/spec/blob/master/spec/discovery.md for some good ideas of bootstrapping

cgwalters:
Obsoleted by https://github.com/GNOME/ostree/pull/139

PR #137, opened 2015-09-10
pull: honor gpg-verify-summary=false when a summary signature is present

giuseppe:


cgwalters:
Merged, thanks!

PR #138, opened 2015-09-10
Mode-dependent default for "commit-update-summary" option

mbarnes:
Default to automatically (re)generating a summary file when committing to an `archive-z2` repo, since these are often served over HTTP and some commands like `ostree remote refs` depend on fetching a remote repo's summary file over HTTP.  Basically all `archive-z2` repos should keep an up-to-date summary file unless there's a specific reason not to.

**Ed:** This seems like a more reasonable default behavior to me, and it might help bypass some of our rel-eng woes.

### DIFF
--- a/src/ostree/ot-builtin-commit.c
+++ b/src/ostree/ot-builtin-commit.c
@@ -487,7 +487,9 @@ ostree_builtin_commit (int argc, char **argv, GCancellable *cancellable, GError 
 
   if (!skip_commit)
     {
+      OstreeRepoMode mode = ostree_repo_get_mode (repo);
       gboolean update_summary;
+
       if (!ostree_repo_write_commit (repo, parent, opt_subject, opt_body, metadata,
                                      OSTREE_REPO_FILE (root),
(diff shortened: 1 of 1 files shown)


cgwalters:
The reason I didn't do this is because ostree supports multiple processes writing to a single repo (but different refs) concurrently, and gnome-continuous (for example) actually does this.  But the summary file needs to be serialized.

mbarnes:
Seems like we could solve that with a lock file for the summary?

cgwalters:
Yeah...except several users of OSTree have it writing to NFS, and file locking on NFS is a sad story: http://0pointer.de/blog/projects/locking.html

mbarnes:
I pushed a comment commit to `master` that captures the rationale for the default, in case anyone else gets the bright idea.  See https://github.com/GNOME/ostree/commit/2233d443e76ba0557b6c57a76e0668b991003fb9.

cgwalters:
Thanks!  A lot of this is really my fault for not documenting things well from the start =/

PR #139, opened 2015-09-10
Single file deltas

alexlarsson:
Here is a new approach for https://github.com/GNOME/ostree/pull/136
We modify the delta part header to add a maybe variant which if set contains the actual parts instead of them being separate files on disk. This works both with offline apply and with pull, as well as not needing any new requirements. Things like origin url and original refname can be stored in the existing metadata field.

The bad part about this is that the format of the delta file is changed. We hardcode the full typename when we open the superblock, so there is no way to easily extend the format. The delta meta entry has a version number, but we can't actually use it because we need to know at open time the full type, which includes the full meta entry format. So, one can bump the version field, but not change the type of the struct.

The file format really needs some tweaks to be future safe. For instance, the delta-meta-entry should probably be "av" to be able to use the versioning as per above. And we should probably shove a "av" at the end of the main superblock for future extensibility.


alexlarsson:
btw, the first commit in this series is a pure bugfix and should go in no matter what we decide on the rest.

cgwalters:
Yep, pushed that patch along with a subsequent test case commit.

cgwalters:
Can you take a look at my comment in https://github.com/GNOME/ostree/pull/136#issuecomment-139945635 ? Note currently `execute-offline` does not perform any GPG check.

alexlarsson:
I will:
 * update the code to store the delta parts inline in the metadata a{sv} field, to avoid breaking the format
 * Make sure we verify the checksum of files in the delta part in case the summary is not signed or we manually applied the delta.
 * optionally make the delta entry format an array of variants so it can be changed later.

alexlarsson:
Update to store parts in metadata. Will look at verifying checksums now.

alexlarsson:
I also added a commit that lets you optionally add detached metadata (i.e. commit signatures) to the superblock. While this is not very important for pulling its quite important for the "standalone commit file" case.

With this i'm able to use single-file bundles in xdg-app, here is the branch for that: https://github.com/alexlarsson/xdg-app/tree/single-file-apps

giuseppe:
I have reviewed the patches and played around a bit.  LGTM

cgwalters:
I'd like to get a chance to look at this too because it's security sensitive.  Hopefully soon.

cgwalters:
BTW have you seen https://robots.thoughtbot.com/autosquashing-git-commits ?  Most of those last 4 could be fixups that'd be autosquashed.

We aren't using the github merges, so doing the manual autosquash beforehand 

Anyways...I think I'm ok with this series.  I have some lingering uncertainties around how the whole thing gets put together by xdg-app, particularly around GPG key management etc.  That shouldn't block this though, particularly because static deltas are not yet stable.

So OK to merge as is, or with autosquash.

cgwalters:
i.e. :+1:

alexlarsson:
Pushed to master

cgwalters:
A sanity check; we didn't end up breaking the delta format, because we just stuck this in the metadata, right?

alexlarsson:
Yeah, no break.

PR #140, opened 2015-09-14
Don't prune repository before a deploy

mbarnes:
Continuing the discussion in https://bugzilla.gnome.org/732526...

With these changes rolled into a test tree, I was able to jump backward a few versions on my atomic host instance with:
```
ostree pull <remote> <checksum>
ostree admin deploy <checksum>
** reboot **
```


mbarnes:
Thanks for the review, @giuseppe.

@cgwalters, is the 2nd commit kinda what you had in mind?  It doesn't actually get called anywhere (except for the trivial case of `ostree_sysroot_cleanup()`) so I'm a little hesitant to merge it without more use cases.

cgwalters:
My thought was we'd replace the `ostree_sysroot_cleanup()` call in the first commit with one which did everything except prune - i.e. squash the two together.

The alternative is to audit the code to ensure we're cleaning up any partial state anywhere we write data - I knew off the top of my head that the deployment root was one of these, but bootloader files, bootlinks, origin files etc. would also need verification.  Which is why it felt simpler to me to do cleanup-but-not-prune, as in the end we need to do that anyways.

mbarnes:
Ah, got'cha.  So this needs to be new public API then.

cgwalters:
Hmm, right...I forgot we were doing cleanup in each program before =/  Maybe we just expose `ostree_sysroot_prepare_cleanup()` as the new `ostree_sysroot_cleanup()` that programs should call?  I'm fine with keeping the flags version internally, but having each program hardcode `OSTREE_SYSROOT_CLEANUP_ALL & ~OSTREE_SYSROOT_CLEANUP_PRUNE_REPO` would feel a bit ugly.

It's currently asymmetrical because `ostree_sysroot_write_deployments()` does the post-cleanup internally itself, it's not something programs worry about.  But I'm not sure we can change that easily now.

mbarnes:
Reworked everything to match (I think) what you suggested.

Not sure if we still need the extra `glnx_shutil_rm_rf_at()` call in `ostree-sysroot-deploy.c` but I left it in for review.

cgwalters:
LGTM, thanks!

mbarnes:
Merged this, not sure why GitHub isn't detecting it...

PR #141, opened 2015-09-17
reset: Simplify argument checking logic

mbarnes:
Just a minor cleanup I noticed while reading the `ostree reset` code.


cgwalters:
:+1:

PR #145, opened 2015-09-24
tests: Add a test for pull+deploy of specific "bare" commit

cgwalters:
This is the "pull an older version" that's now supported since
36d65b3cfcc9557552314d112493516437d6fcd4

However...this test does *not* fail when I revert the commits that were supposed to make it work.  I'm still debugging that.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -49,6 +49,7 @@ testfiles = test-basic \
 	test-admin-deploy-uboot \
 	test-admin-instutil-set-kargs \
 	test-admin-upgrade-not-backwards \
+	test-admin-pull-deploy-commit \
 	test-admin-locking \
 	test-repo-checkout-subpath	\
 	test-reset-nonlinear \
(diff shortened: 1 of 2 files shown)


cgwalters:
And now that I had a chance to test on top of https://github.com/GNOME/ostree/pull/146 - this test correctly fails with the f4d832475edcb2bf6fdae2eaa8f8cb1b6a02abf0 commit reverted locally.

cgwalters:
To clarify, what was confusing is that the "no prune" behavior also occurred when transitioning from `0 > 1` deployments, as well as `1 -> 0`.  In the normal case of a running system you have at least 1, but my test only had 0.

cgwalters:
@mbarnes ^

mbarnes:
Looks good, and I gotta remember to use `rev-parse` in future tests.  That's handy.

I think it wouldn't hurt to link to this PR in the test script for posterity.  Sometimes it's not entirely clear what a test is testing for.

Otherwise, :+1:

cgwalters:
Not having a link to the PR conversation is a major gap in this workflow now.  =(

Maybe we can just do it manually - submit your PR, then use `git rebase` to add the PR link to the bottom of each commit?  Could be automated.

cgwalters:
Thanks though, pushed as https://github.com/GNOME/ostree/commit/b6e2eaf212bac6231dcaf03491b251b0a0850002

PR #146, opened 2015-09-24
sysroot: Cleanup refs and prune even on last undeployment

cgwalters:
I was working on a different test, and ended up being very confused at
the behavior where removing the last deployment didn't remove the last
`ostree/X/X/X` ref pointing to its commit.

There's no reason to special case the last undeployment AFAIK, and the
existing code handles this.


mbarnes:
LGTM, I think I had wondered about that myself while studying that code recently.  :+1:

PR #147, opened 2015-10-02
upgrader: Allow overriding the checksum to fetch

mbarnes:
This enhancement to `OstreeSysrootUpgrader` allowed me to use it throughout my demonstration script here: https://github.com/mbarnes/ostree-scripts/blob/master/clunky-rebase

Setting an override checksum causes `OstreeSysrootUpgrader` to pull a specific commit and reset the origin refspec to it in `ostree_sysroot_upgrader_pull_one_dir()`.


cgwalters:
One thing we could do is extend the origin file format.  Something like:

```
[origin]
refspec=someos:someos/42/x86_64/buildmaster
temporary-commit=cf92d91ef907afb6f9fda879421be93ccb5a33e9a69f2bebc6f3d3817b5228a0
```

Where `temporary-commit` tells the upgrader to pull that commit if the *current* one is different, but if the current deployment is already on it, then upgrade to latest?

cgwalters:
Though the way I had originally thought of this is that pinning to a specific commit is a "permanent" operation, i.e. `ostree admin upgrade` would just say "You're already on that commit".  And if you wanted to unpin, you'd `ostree rebase someos:someos/42/x86_64/buildmaster`.

If we wanted to go that approach, then I think it would probably work to just do:

```
[origin]
refspec=someos:cf92d91ef907afb6f9fda879421be93ccb5a33e9a69f2bebc6f3d3817b5228a0
```

?

Either way...it seems like it would be nice to reflect this somehow in the origin, even if it is just a comment.

mbarnes:
Do you have a preference for one method over the other?  Certainly we could support both at the `ostree` level.

For the `rpm-ostree rebase` enhancement I'm working toward, since my proposed syntax takes a refspec argument (e.g. `rpm-ostree rebase --version=1.2.3 [REFSPEC]`), I though it would be less surprising to stay on that refspec.  At least by default.  We could add an option to make it permanent.

This is just my initial thinking, though.  I can be talked out of all of it.

cgwalters:
Reading my comments, I realized it's a bit unclear.  What I was thinking is about a scenario where you have multiple sysadmins, and one of them downgrades a set of machines in the cluster due to a bug; maybe that subset's has a different ethernet card and the new kernel is buggy or whatever.

It'd be nice to have something permanent in the origin (or UI) that reflects this is an explicit decision.  Otherwise another admin might wonder why that subset is on a downgraded version.

If we're going for the "* upgrade takes you to latest" approach, then I like the
`override-version=1.2.3` or `override-commit=98ea6e4f216f2fb4b69fff9b3a44842c38686ca685f3f55dc48c5d3fb1107be4` in the origin file (and teaching the upgrader to honor it) over having an API that tools use, as it's at least recorded.

mbarnes:
I do like the idea of avoiding new API here.

> Where temporary-commit tells the upgrader to pull that commit if the current one is different, but if the current deployment is already on it, then upgrade to latest?

If the daemon is upgrading to latest by way of the `Upgrade()` method, couldn't it just strip the override line from the origin file prior to starting the upgrader?  That seems a little more intuitive to me than giving the upgrader a double interpretation of the override line.

Also, one thing that's still fuzzy to me is how much `ostree` should know about versions.  I know `ostree admin status` looks for them, but I still had the impression they were meant to be an `rpm-ostree` concept.  Perhaps we should just go with `override-commit` in the origin file and have the daemon resolve the version number to a checksum beforehand?

cgwalters:
Yeah...we could teach the sysroot upgrader to strip out `override-*` when doing a regular upgrade.  So it would only be honored when the caller explicitly uses `_set_origin()` ?

As far as ostree knowing about versions, I'm OK if we make more use of them in ostree/libostree.  Version numbers for software are a pretty universal concept =)

mbarnes:
Almost.  I had imagined the higher-level upgrade logic stripping out the override, and then handing what's left to `ostree_sysroot_upgrader_set_origin()`.  So the upgrader itself stays fairly dumb and just acts on what's present in the origin info; no surprises or implicit behaviors.

> As far as ostree knowing about versions, I'm OK if we make more use of them in ostree/libostree. Version numbers for software are a pretty universal concept =)

Okay, makes sense.

I do still think it wise to keep version-to-checksum resolution out `libostree` for now -- I expect less stupid solutions than mine in the future -- and just go with `override-commit` in the origin file.  The daemon could still record the version number in the origin file as a comment for human sysadmins.

Seems like we're on the same page, so I'll get started on a redo.

mbarnes:
**Sidenote:** If ostree is going to start becoming aware of versions, there's an obvious parallel here to git tags.  That would certainly make version resolution easier.  I imagine you've had some thoughts along these lines?

cgwalters:
If we put the version-to-checksum support into the libostree pull code it'd make it easier and more efficient, right?  We could pretty easily add `version=<string>` into `ostree_repo_pull_with_options()` or so?

As far as tags...yeah.  It would make a lot of sense.  Just hasn't been the time to "just do it" yet.

mbarnes:
> If we put the version-to-checksum support into the libostree pull code it'd make it easier and more efficient, right? We could pretty easily add version=<string> into ostree_repo_pull_with_options() or so?

Easier maybe, we'll see about more efficient.

Main thing I'm paranoid about is not making any `libostree` commitments for this yet.  We will definitely want the freedom to revamp it later, especially if git-style tags might be on the horizon.  But yeah, I guess we could sneak it in through options for now.

mbarnes:
Rebased on a rewrite.

I'll tackle the version-to-checksum stuff separately but it helped to talk it through here.

Don't have a proper test case for this, but I did verify it with my little Python script which is a bit shorter now: https://github.com/mbarnes/ostree-scripts/blob/master/clunky-rebase

mbarnes:
I'm having some difficulties coming up with a shell script test for this.

Nothing in ostree currently inserts the `override-commit` origin key, so I would have to do so manually.  But the only commands that even use `OstreeSysrootUpgrader` is `admin upgrade` and `admin switch`.

`admin upgrade` explicitly *removes* `override-commit` each time, so that's out.

`admin switch` always sets the upgrader's origin with `ostree_sysroot_origin_new_from_refspec()`, so I don't even have a chance to insert the option.

I suppose I could write a `libostree` test using the bindings -- and will do so if you think it necessary -- but I think it'd be easier to test this from rpm-ostree since that will actually have a CLI for exercising this code.

We could just hold off on merging this PR until I have I have more pieces put together.

cgwalters:
Could we add `ostree admin upgrade --override-commit` ?

mbarnes:
Yes, that would allow me to write a test case.

I'm not yet sure what I think about it from a design perspective, just because in my mind the main use case here is deploying *older* commits and it seems weird to use an `upgrade` command to downgrade.

But if you're okay with it I'll stop straining my brain so hard.  :confounded:

cgwalters:
It is a bit weird I agree, and it is unfortunate to have code duplication between ostree/rpm-ostree here.  But I don't have a fix for that.  I'm not opposed to relying on rpm-ostree tests for this, but having tests in ostree is nice if it's not too hard.

mbarnes:
Rebased on `master` and added the `--override-commit` option and a test case.

cgwalters:
LGTM!

PR #148, opened 2015-10-06
make ostree more u-boot friendly

gatispaeglis:


gatispaeglis:
What I am trying to pull request here is https://github.com/gatispaeglis/ostree/commit/6eac56d706c01b0f82767178e1e9cc7d396c45a1

gatispaeglis:
There are 2 ways to provide a custom env to u-boot from user space;
1) compiled binary that is sourced from u-boot. 
2) text file (usually /uEnv.txt) that is imported into env from u-boot.

I talked with the author of the current OSTree u-boot integration code from where I found out that it was designed with the 1st case in mind. I am using OSTree on several devices,  some use approach nr 1 and some use approach nr 2.

PR #149, opened 2015-10-06
Don't require /boot/uEnv.txt for u-boot support

gatispaeglis:
The current code checks if /boot/uEnv.txt is a symlink to
decice if sysroot requires u-boot support. Why this is bad:

There are 2 ways to provide a custom env to u-boot from user space:

1) A compiled binary that is sourced from u-boot.
2) A text file (usually /uEnv.txt) that is imported into env from u-boot.

The current OSTree u-boot integration code was designed with the 1st
case in mind.

Many bootscripts provided by an embedded device vendors expect
to find uEnv.txt in the top level directory, it is often hardcoded
when building u-boot and is difficult to change later on. Or in other
cases it is stored in read-only memory so changing it would require
re-flushing boot loader with a new env. So the issue here is that
OSTree's and vendor uEnv.txt want to exist and the same path and OSTree
would throw away any changes added to /uEnv.txt by user on the next
upgrade/deploy.

This patch "hides" away the OSTree's env file loader/uEnv.txt from users
who are used to edditing uEnv.txt at the top level directory. Now to add
OSTree support on such boards you can simply add a custom logic in uEnv.txt
that loads ostree env from /loader/uEnv.txt

This change is backward compatible with the previous ostree releases and
solves the issue described in:

https://bugzilla.gnome.org/show_bug.cgi?id=755787

### DIFF
--- a/src/libostree/ostree-bootloader-uboot.c
+++ b/src/libostree/ostree-bootloader-uboot.c
@@ -52,7 +52,7 @@ _ostree_bootloader_uboot_query (OstreeBootloader *bootloader,
 {
   OstreeBootloaderUboot *self = OSTREE_BOOTLOADER_UBOOT (bootloader);
 
-  *out_is_active = g_file_query_file_type (self->config_path, G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS, NULL) == G_FILE_TYPE_SYMBOLIC_LINK;
+  *out_is_active = g_file_query_file_type (self->config_path, G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS, NULL) == G_FILE_TYPE_REGULAR;
   return TRUE;
 }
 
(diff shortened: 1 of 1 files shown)


gatispaeglis:
"exist and the same" -> "exist *on* the same"

cgwalters:
It'd be nice to have review from the person who originally added the uboot backend, which is: Javier Martinez Canillas <javier.martinez@collabora.co.uk>.  I'll send him a mail.

cgwalters:
I also pinged the Fedora maintainer of u-boot for review of this.

gatispaeglis:
I did share the link to the patch with Javier earlier this week on IRC u-boot channel and he seemed to agree that the patch is OK and is backwards compatible. But we can wait for him to reply to your mail.

Other bonus what this patch adds is that now you are not required to store your /boot (in case you keep the /boot on a separate partition) on a file system which has support for symbolic links. For example, if somebody really wants to use FAT FS for some reason, now they can do that.

martinezjavier:
@cgwalters the patch looks correct to me. As @gatispaeglis said the patch is safe and backward compatible (since current deployments will have a boot/uEnv.txt symlink so won't be affected) and in insight, I should had checked for boot/loader/uEnv.txt instead of boot/uEnv.txt since boot/loader/ is a symlink to boot/loader{0,1} which is what OSTree really updates.

BTW, sorry for the delay. I did not get your email and @gatispaeglis pointed me to this pull request. Probably because you sent me to my old email address since I moved to another company.

Best regards,
Javier

cgwalters:
Makes sense, thanks!  Merged:

https://github.com/GNOME/ostree/commit/b69a88b111d1993247c2530de576fdde3bc7dcc7

PR #150, opened 2015-10-09
Better descriptive variable name on ostree commit

leandrosansilva:
- On the commit command, uses commit_subject as a better variable name
  for the directory, archive or ref to be commited, instead of just 'arg'.

- This is just an aesthetic change in the source code, having no impact
  anywhere else.


cgwalters:
I tweaked the commit message to match the existing style (e.g. using `commit: ` subject prefix) and pushed, thanks!

https://git.gnome.org/browse/ostree/commit/?id=5e98820b579eb06a3478f6496662fc071e7f7d2d

PR #151, opened 2015-10-12
pull: allow pulling delta update from different refs

leandrosansilva:
Added the option --force-from= to ostree pull where a delta does not
need to be applied between commits from different refs.

The implementation is probably not good, I made it for personal need,
but I'm sure the idea can be better evaluated.

Some background for this request at:
https://mail.gnome.org/archives/ostree-list/2015-June/msg00012.html

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1694,6 +1694,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   char **refs_to_fetch = NULL;
   GSource *update_timeout = NULL;
   gboolean disable_static_deltas = FALSE;
+  const char *force_from = NULL;
 
   if (options)
     {
@@ -1706,6 +1707,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       (void) g_variant_lookup (options, "override-remote-name", "s", &pull_data->remote_name);
       (void) g_variant_lookup (options, "depth", "i", &pull_data->maxdepth);
       (void) g_variant_lookup (options, "disable-static-deltas", "b", &disable_static_deltas);
+      (void) g_variant_lookup (options, "force-from", "&s", &force_from);
     }
 
   g_return_val_if_fail (pull_data->maxdepth >= -1, FALSE);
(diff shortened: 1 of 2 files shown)


giuseppe:
another alternative would be to check in the summary file what static delta files are available remotely and then fetch any of them where TO is the destination commit and FROM is a commit that is already fetched locally?

Let's say that locally you have commits A, B, C and you want to upgrade to Z.  The remote index files list delta files for B->Z, D->Z, E->Z, then the static delta B->Z can be fetched, since B is already fetched locally and use it to upgrade to Z.

What do you think of this idea?

cgwalters:
Are you writing a custom upgrade program on the client side to make use of this?

leandrosansilva:
@cgwalters yes, I am. Currently I'm using the gobject python interface.

I agree the implementation is bad, my idea was showing the idea might work, as a proof of concept, and in this current state it's useful for me.

@giuseppe I totally agree a implementation which automatically finds FROM is better than passing it explicitly. But I confess my knowledge on ost internals so far is not good enough for implementing it. But I'll take a look on how to implement your idea.

Thank you all for the feedback :-)

cgwalters:
Ok, it just helps to know the higher level use case.  I would like to support this "move between branches" model - but, we're doing some work in rpm-ostree which seems a bit similar.  See this:

https://github.com/projectatomic/rpm-ostree/pull/178

Basically we're orienting more strongly towards the `ostree.version` metadata on a single branch.  Would that style of approach also work for you?

If you're trying to present a user interface around clearly versioned builds - that is the intent of the version number approach.

That said again I'm open to improvements in the branch-for-versioning model too.

leandrosansilva:
@cgwalters my use case is that in the system I'm working on, a machine can change its current system arbitrary (to any other branch), as I keep different versions in the  same repository. It's like parallel development branches with the possibility of more than one "stable" branch. That's why I could not fit the ostree.version system on it.

Do you see any technical drawbacks on allowing those cross-branch delta?

cgwalters:
Ok, makes sense.  Maybe instead of `force-from`, call it `delta-previous-commit`?  

Actually, we should just do this by default via `OstreeSysrootUpgrader` at least, as it's more precise.

jlebon:
bot, test pull request

cgwalters:
Would still consider this with some followup to review comments and a test case.

cgwalters:
I want this now to support static deltas across rebases (in this case between C7AH -> CAHC).  Though I'm increasingly thinking what we want here isn't quite "force from" but `--add-from`.  And by default, `OstreeSysrootUpgrader` should provide the current deployment's commit (and possibly the rollback commit too).

cgwalters:
Going to close this in favor of https://github.com/ostreedev/ostree/pull/710 which I think will do this more "automagically".  Thank you for the patch though!  Let me know if my proposed version works for you.

PR #152, opened 2015-10-13
sysroot: Write symlinks before calling fsync(), then rename after

cgwalters:
There might be a race here in that we create new symlink files *after*
calling `syncfs`, and they are not guaranteed to end up on disk.

Rework the code so that we create symlinks before, and then only
rename them after (and `fsync()` the directory for good measure).

(Needs more testing)

https://bugzilla.gnome.org/show_bug.cgi?id=755595


cgwalters:
Thanks!  Merged with that fix.

PR #154, opened 2015-10-30
giuseppe/tombstone

giuseppe:


giuseppe:
@mbarnes thanks for all the comments.  I have left the ```tombstone-commits=true``` and didn't change the logic in ostree-repo-traverse.c for now.
If @cgwalters is fine, I would leave tombstone-commits=true instead of a features list, as it is easier to handle and even to read in the config file.

cgwalters:
`tombstone-commits=true` is fine by me yep.

cgwalters:
I like it, thanks!

giuseppe:
I have reworked all the comments, if it fine for you now, I will autosquash the patches and push the series

mbarnes:
LGTM!  :+1:

giuseppe:
pushed thanks!

PR #155, opened 2015-11-05
giuseppe/prune delete commit

giuseppe:
I added the possibility to delete a commit from CLI.  Also, I changed the login of fsck --add-tombstones to check for missing parent commits, instead of looking at the partial state.


mbarnes:
Code changes look correct to me.

I'm a little bit wondering if this should be a top-level command like `ostree delete`, as it seems a bit hidden to me as just an option of `ostree prune`.  Also, I don't have a specific use case in mind but it feels like deleting a single commit and pruning an entire repo should be separate operations in case a user wants to do something in between.

I could be wrong though.  What do you think?

giuseppe:
I decided to add it under ```prune``` as deleting a commit could leave objects without references in the repository, so embedding it into prune solves this problem; but I agree the connection between these two operations is a bit forced.  If we decide to move it to a top level command, then I would slightly change how it works and allows to remove a commit pointed to by a reference, and make the reference point to its parent instead.

cgwalters:
No objections to the command line, but in the original bug https://bugzilla.gnome.org/show_bug.cgi?id=743882

I think most people will want something like a script to auto-prune based on criteria like:

 - time: "Prune all commits from last week and past"
 - length: "Keep the last 100 commits"
 - Size: "Keep enough commits to fill up 100G" (kind of nontrivial)

We already have the API which is probably enough for this, but just noting that.

mbarnes:
`ostree prune` vs a separate command was just a thought, not a blocker for me.

So :+1: from me if you'd rather keep it as is.

mbarnes:
> I think most people will want something like a script to auto-prune based on criteria like:
>
>   * time: "Prune all commits from last week and past"
>   * length: "Keep the last 100 commits"
>   * Size: "Keep enough commits to fill up 100G" (kind of nontrivial)
>
> We already have the API which is probably enough for this, but just noting that.

Would it make sense to be able to express this in the repo config so ostree can trim the repo automatically after pulling down more commits?  Syntax would require a some design work but I'm sure there's precedent elsewhere.

giuseppe:
I added another commit to support ```prune --before=DATE```.  It uses the parser used by coreutils date for ```date --date="DATE"``` so it supports things like "-1 year", "1 week ago".

I added some trivial commits that are not really dependent from this series, which are not worth a separate PR.

giuseppe:
oops small issue I didn't consider, the parse-datetime module is distributed under GPLv3. I will ask the authors to relax the license to GPLv2

cgwalters:
Hmm...I'd been trying to keep ostree consistently under *L*GPLv2.

I'm not opposed to this but I think in the big picture we'd probably be better off just having core APIs in OSTree, then having something like an official/more productized https://github.com/cgwalters/ostree-scripts ?  Open to discussion though.

I definitely see us as different from git in that we expect most users to be *automating* commits and management.

giuseppe:
I've asked the author if the license can be relaxed so we can use it.  The only disadvantage I see is that if not enough users will use it, then it will be secondary and won't get enough maintainance. Having feature in the main repository we will keep testing them and keep in sync with the rest of the software.
On the other hand, if we decide to have more features in OSTree, then we will get stuck with them and have to support them forever.

giuseppe:
I added some more commits.  git offers the possibility to override the commit date, now sure how in general that is useful for users but for completeness I added an API and exposed it.  At least it facilitates the testing of the ```prune --before=DATE``` feature, in case we decide to merge it.

I am waiting for an answer from RMS if he agrees to change the license for parse-datetime to GPLv2 so that can be used by this feature.

cgwalters:
Note OSTree is *Lesser* GPLv2+, or LGPLv2+.  It'd be nice to keep it that way just from a simplicity perspective.

mbarnes:
> I am waiting for an answer from RMS if he agrees to change the license for parse-datetime to GPLv2+ so that can be used by this feature.

You're asking RMS to (in his mind) *downgrade* the license?  Oye.  I'm curious to know what he says.

giuseppe:
In the meanwhile, I am already looking for alternatives :)

```getdate_r``` seems the closest one, though there is no support for relative arguments.

giuseppe:
ok, in the meanwhile I have added two fixup to revert the usage of parse-datetime and parse the same format as used in the ```ostree log``` output

cgwalters:
OK, thanks.  Looks pretty good to me now!  :+1:

cgwalters:
BTW, related: https://docs.openshift.com/enterprise/3.0/admin_guide/pruning_resources.html
(Mostly `--keep-younger-than`, but worth linking to anyways)

giuseppe:
Yes good idea, OK to rename ```--before``` to ```--keep-younger-than``` before pushing?

cgwalters:
Sounds good to me, yeah.

giuseppe:
renamed and pushed. Thanks for the review.

PR #156, opened 2015-11-05
Allow updating files on the boot partition

gatispaeglis:


gatispaeglis:
I can also add some auto tests after after we will agree on this solution.

cgwalters:
Can we try to drive some standardization at the bootloader/distribution level around this area?

I'm a bit uncertain about just copying everything in `/boot`.  Currently in Fedora+ostree, we do this once at initial installation time in Anaconda, but not thereafter.  This does mean we don't get updates for things like GRUB splash screens.

I am a bit worried at a minimum this will break an existing ostree user who has content in `/boot` they don't want copied.

One further thing - we actually moved to supporting both `/usr/lib/ostree-boot` and `/boot` for content.  See... https://bugzilla.gnome.org/show_bug.cgi?id=740947 and https://github.com/projectatomic/rpm-ostree/commit/15ecaacd36bfcb91181400a7399461deef57e062

The basic rationale there is that this ensures that `/boot` is just an empty mount point in the deployment root, so we can mount the *real* `/sysroot/boot` (or a physical device) there.

gatispaeglis:
Hi, sorry for a late reply. 

Copying in my cases is needed because I want to keep "/boot" on a separate partition where it's file system type is independent from "rootfs" partition's file system type. The main reason for this is that u-boot might not have support for reading from the file system of rootfs - one example would be BTRFS.

The "boot_location" property from rpm-ostree does not seem to solve this issue, but I see that it nicely solves the boot shadowing issue.

Ideally there should be a solution that back ports "boot_location" to ostree.git repo + optionally copy boot contents (for backward compatibility). Is there a plan to move "boot_location" to ostree.git repo?

cgwalters:
I haven't forgotten about this, but I don't have quite enough expertise to weigh in confidently yet.

What specific files are we copying here?  Is this about devicetrees?

Also not required, but I've been porting more of ostree away from the `GFile*` APIs to use libglnx; see https://git.gnome.org/browse/ostree/commit/?id=ca57ec4aa56e0e3ecfb4c5f4f34b84633a22fb66 and https://git.gnome.org/browse/ostree/commit/?id=9020fe25473ee9f71aad416f01de040574e2052f etc.

gatispaeglis:
> What specific files are we copying here? Is this about devicetrees?
Yeah, so far I needed this only for device tree files and u-boot script file as both of these can change when kernel version changes.

I have been digging into GRUB/UEFI world as well as you might notice soon on bugzilla :) We can revisit this patch later as I might learn how and if this patch could benefit GRUB systems too.. and if patch needs any adjustments. But in general it would be cool to get my local patches upstreamed at some point ( I have few more ;) ) . Currently I am working with ostree version v2015.10, and resolving merge conflicts for every ostree version in future does not sound too exciting.

cgwalters:
bot, test pull request

gatispaeglis:
> Can we try to drive some standardization at the bootloader/distribution level around this area?
I am a bit unsure how to proceed with this.

According to rpm-ostree docs and the current code in ostree, the usr/lib/ostree-boot directory takes precedence over the /boot dir. Also /boot is marked as a legacy.

So one solution to enforce people to move from legacy /boot to use usr/lib/ostree-boot is to enabled the "update files in boot/ostree/os-bootcum" feature only for files that are stored in usr/lib/ostree-boot. And to keep backward compatibility, this feature must be enabled via configure argument. 

--update-ostree-boot for example.

How does this sound?

cgwalters:
It might be that some content is "copy once", others are "copy always"?

Thinking about this a bit more...I can't think of anything really problematic with processing everything in `usr/lib/ostree-boot` by default.   For Fedora/CentOS Atomic Host it means we'd start grabbing other bits from the kernel package like `symvers-3.10.0-327.13.1.el7.x86_64.gz`, but that's OK.

Other things like `usr/lib/ostree-boot/grub/splash.xpm.gz` have me less certain.  It seems like the default should probably be "things from the kernel component/package get updated, things from the bootloader don't"?

cgwalters:
Maybe for backwards compatibility we could introduce a set of directories like:
`usr/lib/ostree-boot/kernel` `usr/lib/ostree-boot/boot` ?  Then e.g. rpm-ostree would drop everything from the `kernel` package in the former, and everything else (mostly from the `grub` package) in the second?  And the semantics would be the first is all part of the bootcsum, the second is processed only by installers?

gatispaeglis:
Hm, I definitely have seen use cases where boot loader files need to be updated. For example this patch https://github.com/gatispaeglis/ostree/commit/013279d6de9b49bd9d5ffb0be3593b1971c00ef0 depends on being able to update u-boot's uEnv.txt.
It is a bit risky to have a dedicated sub-paths [kernel, boot] as we can't predict all possible use cases, unless we add [kernel, boot, misc] or simply one extra subdir (for the backward compatibility reasons) - usr/lib/ostree-boot/extra-updates

cgwalters:
Can you reopen this PR so it stays on the radar?

cgwalters:
Maybe it's simplest to special-case the `uEnv.txt` case for now?

gatispaeglis:
Special-casing won't scale here. Few more examples for the devices I have been working with:

- U-Boot on Boundary Devices by default searches for u-boot bootscript (this is different from uEnv.txt) in /boot. This script changes between kernel releases and thus needs to be shipped with the OSTree update.

- Beagle Bone Black keeps the u-boot binary on the sdcard in /boot, because it does not have enough internal storage space. This binary might require an update. 

- Theoretical: I have read about "conventional BIOS" [1]. In this setup GRUB binary is kept in /boot. It could be desirable to update it. 

The way I have documented this in my internal repo is:
Updates in boot partition are supported only for major releases - when kernel/initramfs versions change.
The kernel/initramfs version is considered to change when 'bootcsum' changes in the following expression:                  
`bootcsum=$(cat vmlinuz initramfs | sha256sum | cut -f 1 -d ' ')`

It makes sense to me, as updating on boot/ should not happen frequently, but only when kernel/initramfs changes.

[1] http://www.linuxquestions.org/questions/linux-hardware-18/efi-on-ext4-4175510688/

> Can you reopen this PR so it stays on the radar?

I will try that tomorrow.

gatispaeglis:
Not sure why it shows all those other commits here. This is the initial version of PR for v2015.10. Doesn't even compile with current master. Still thinking about best possible solution..

PR #157, opened 2015-11-14
Rework OstreeFetcher to allow for temporary main contexts

mbarnes:
This is an attempt to solve the issue with `OstreeFetcher` iterating the thread-default main context.

I basically moved the `SoupSession` to a separate thread with its own isolated main context and main loop, so it won't be blocked by pushing a temporary main context in the caller's thread.

The 4th commit is big and scary, but it's mostly moving existing code to new functions and structs and making the requests thread-safe.  The actual request processing in `OstreeFetcher` is unchanged -- at least that's the intent.

The last commit reverts the non-error-handling parts of https://git.gnome.org/browse/ostree/commit/?id=84fe2ffb2bcc788916beb563b77cd6fe2c836456 to once again push a temporary main context.

The installed tests pass and it seems to be working well for me.  But given how critical this code is, I'd advise not merging this immediately before an upstream release.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -33,12 +33,37 @@ lib_LTLIBRARIES += libostree-1.la
 libostreeincludedir = $(includedir)/ostree-1
 libostreeinclude_HEADERS = $(libostree_public_headers)
 
+ENUM_TYPES = \
+	src/libostree/ostree-fetcher.h \
+	$(NULL)
+
+src/libostree/ostree-enumtypes.h: src/libostree/ostree-enumtypes.h.template $(ENUM_TYPES)
+	$(AM_V_GEN) $(GLIB_MKENUMS) \
+	--template src/libostree/ostree-enumtypes.h.template \
+	$(ENUM_TYPES) > $@
+
+src/libostree/ostree-enumtypes.c: src/libostree/ostree-enumtypes.c.template $(ENUM_TYPES)
+	$(AM_V_GEN) $(GLIB_MKENUMS) \
+	--template src/libostree/ostree-enumtypes.c.template \
+	--fhead "#include \"ostree-enumtypes.h\"" \
+	$(ENUM_TYPES) > $@
+
+ENUM_GENERATED = \
+	src/libostree/ostree-enumtypes.h \
+	src/libostree/ostree-enumtypes.c \
+	$(NULL)
+
+BUILT_SOURCES += $(ENUM_GENERATED)
+
+CLEANFILES += $(BUILT_SOURCES)
+
 libbupsplit_la_SOURCES = \
 	src/libostree/bupsplit.h \
 	src/libostree/bupsplit.c \
 	$(NULL)
 
 libostree_1_la_SOURCES = \
+	$(ENUM_GENERATED) \
 	src/libostree/ostree-async-progress.c \
 	src/libostree/ostree-cmdprivate.h \
 	src/libostree/ostree-cmdprivate.c \
--- a/configure.ac
+++ b/configure.ac
@@ -43,6 +43,8 @@ AC_CHECK_HEADER([sys/xattr.h],,[AC_MSG_ERROR([You must have sys/xattr.h from gli
 
 PKG_PROG_PKG_CONFIG
 
+AM_PATH_GLIB_2_0
+
 GIO_DEPENDENCY="gio-unix-2.0 >= 2.40.0 libgsystem >= 2015.1"
 PKG_CHECK_MODULES(OT_DEP_GIO_UNIX, $GIO_DEPENDENCY)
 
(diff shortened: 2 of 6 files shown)


cgwalters:
Haven't looked at the code in detail, but how does this support progress notifications?   At the moment we're relying on the caller's mainloop iterating for that.  Which is a design bug I admit, but we're locked into backcompat now.  (We can introduce new API of course)

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1148975

mbarnes:
> Haven't looked at the code in detail, but how does this support progress notifications? At the moment we're relying on the caller's mainloop iterating for that. Which is a design bug I admit, but we're locked into backcompat now. (We can introduce new API of course)

The problem I fixed here is pushing a thread-default main context *after* creating an `OstreeFetcher` instance (and, internally, a `SoupSession`) will lock up the `SoupSession`, since it's using the now-buried main context.

That issue hits `_ostree_fetcher_request_uri_to_membuf()` and `_ostree_metalink_request_sync()`.

`ostree_repo_pull_with_options()` is a different beast, which I apparently overlooked.  So progress reporting is unchanged.  It looks like *technically* we could push our own main context there now.  Unless that's the backward-compat issue, in which case yeah we're screwed and need new API.

I think a new wrapper function that just pushes and pops a main context around calling `ostree_repo_pull_with_options()` would do the trick, no?  I'm at a loss for a good name though.

mbarnes:
I merged the first two commits (https://github.com/GNOME/ostree/commit/97efe12ac6b1801d283fb3391656bcf00895afc4 and https://github.com/GNOME/ostree/commit/581b7d61831ccacdc4678127301f86c81db30ccc) since they were just cleanups and @giuseppe reviewed them.

Of the remaining commits, the first one is also cleanup but introduces the use of `glib-mkenums` and still needs review.

cgwalters:
One thing I really want to do at some point is move to an unprivileged[1] separate process for fetches, and do GPG verification in the privileged process.   This patch kind of moves us closer in that direction by us communicating with a thread.  So that is good.

cgwalters:
Ok, I think this looks pretty good.  I have a little bit of nervousness about it because the HTTP code is nontrivial, but you did an excellent job.

Just one minor thing, `g_main_context_invoke()` was created for this, how about squashing in:

```
diff --git a/src/libostree/ostree-fetcher.c b/src/libostree/ostree-fetcher.c
index ee0db5b..affcdb3 100644
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -195,7 +195,7 @@ session_thread_idle_dispatch (gpointer data)
   return G_SOURCE_REMOVE;
 }
 
-static guint
+static void
 session_thread_idle_add (ThreadClosure *thread_closure,
                          SessionThreadFunc function,
                          gpointer data,
@@ -204,8 +204,8 @@ session_thread_idle_add (ThreadClosure *thread_closure,
   IdleClosure *idle_closure;
   g_autoptr(GSource) source = NULL;
 
-  g_return_val_if_fail (thread_closure != NULL, 0);
-  g_return_val_if_fail (function != NULL, 0);
+  g_return_if_fail (thread_closure != NULL);
+  g_return_if_fail (function != NULL);
 
   idle_closure = g_slice_new (IdleClosure);
   idle_closure->thread_closure = thread_closure_ref (thread_closure);
@@ -213,13 +213,11 @@ session_thread_idle_add (ThreadClosure *thread_closure,
   idle_closure->data = data;
   idle_closure->notify = notify;
 
-  source = g_idle_source_new ();
-  g_source_set_callback (source,
-                         session_thread_idle_dispatch,
-                         idle_closure,  /* takes ownership */
-                         (GDestroyNotify) idle_closure_free);
-
-  return g_source_attach (source, thread_closure->main_context);
+  g_main_context_invoke_full (thread_closure->main_context,
+                              G_PRIORITY_DEFAULT,
+                              session_thread_idle_dispatch,
+                              idle_closure,
+                              (GDestroyNotify) idle_closure_free);
 }
 
 static void
```

?

cgwalters:
(And sorry about the delay in review)

mbarnes:
Good call, I didn't know about `g_main_context_invoke()`.  (That would have saved me some typing in past...)

Since you just did an upstream release, this is a good time to merge and let it simmer in `master` for awhile.

mbarnes:
I rebased this branch on @alexlarsson's recent tmpdir work.  There were a few rebase conflicts but nothing too serious.  Unfortunately I don't know of a good way to show the conflict changes here.

Also squashed in @cgwalters's suggested `g_main_context_invoke()` change.

Retested the branch and pushed to master.  Here's the tail of the commit chain:
https://git.gnome.org/browse/ostree/commit/?id=f0b143ca8ac689818a1fe4b1960a16b82a68a843

PR #158, opened 2015-11-17
Validate checksums have correct length

mbarnes:
Keeping my promise to return to a point discussed in 
https://bugzilla.gnome.org/show_bug.cgi?id=755224

`ostree_checksum_bytes_peek()` can return `NULL` if the checksum has an incorrect length (most likely from disk corruption) but most callers are not prepared to handle this and would most likely crash.

Use `ostree_checksum_bytes_peek_validate()` in most places, which sets a `GError` on an invalid checksum.

### DIFF
--- a/src/libostree/ostree-repo-file.c
+++ b/src/libostree/ostree-repo-file.c
@@ -869,9 +869,14 @@ ostree_repo_file_tree_query_child (OstreeRepoFile  *self,
   c = g_variant_n_children (files_variant);
   if (n < c)
     {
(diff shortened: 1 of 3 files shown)


cgwalters:
LGTM, with or without variable renaming.

mbarnes:
Merged with the variable rename:
https://git.gnome.org/browse/ostree/commit/?id=5307af5a7adc79af331c6da804c0ea75cb6adf22

PR #159, opened 2015-12-04
giuseppe/parse datetime

giuseppe:


cgwalters:
I have a bit of a concern about long-term maintenance of this code, but it is pretty useful.  Since we're mainly saying this is a convenience for command line users, I think the risk is low.

I do expect serious OSTree users to be primarily writing scripts using the API, or reusing higher level applications.

cgwalters:
Merged.

PR #160, opened 2015-12-04
cmdline: Fatally error if the timestamp in a commit is invalid

cgwalters:
Previously we were just ignoring this, which hid a bug in
an earlier commit that generated them.

Also change the `commit` program to use both APIs - this
involves extra code, but not too much.

This way, reverting the fix with this on top caused the test suite to
fail.  Adding an active test for this would need a custom test program
using the C API, or adding a cmdline flag to the client, neither of
which quite seemed worth it.

### DIFF
--- a/src/ostree/ot-builtin-commit.c
+++ b/src/ostree/ot-builtin-commit.c
@@ -497,6 +497,11 @@ ostree_builtin_commit (int argc, char **argv, GCancellable *cancellable, GError 
           GDateTime *now = g_date_time_new_now_utc ();
           timestamp = g_date_time_to_unix (now);
           g_date_time_unref (now);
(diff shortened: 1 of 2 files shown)


mbarnes:
Good idea.  :+1:

There's another place where I got bit by this: `ostree_sysroot_upgrader_check_timestamps()`

It can be worked around when the timestamps are bad (`--allow-downgrade`), but still maybe worth failing more gracefully than `g_assert()`?  I can do a separate PR for that if you'd prefer.

cgwalters:
Merged, thanks.  Yeah, I think we should avoid `g_assert()` for things that can come over the network/disk; `GError` is better for that.

PR #161, opened 2015-12-15
giuseppe/prune-static-deltas-only

giuseppe:


giuseppe:
Thanks for the review.  I've pushed a new fixup patch

cgwalters:
LGTM!

giuseppe:
thanks, pushed now.

PR #162, opened 2015-12-16
repo: Set tmp permission to 01777 on each open

mbarnes:
This is a first cut at @alexlarsson's suggestion to set a sticky bit on an ostree repo's `tmp` directory so that normal users can still write files there.  I'm doing this in `ostree_repo_open()` so existing repos can try to self-correct.

Normal users can now create and own `tmp` subdirectories and lock files, and since ostree tries to reuse existing `tmp` subdirectories, I believe it needs to check both the directory name prefix and the directory owner.  The 2nd commit does that.

These changes allow `ostree remote refs` to continue to work without superuser privilege.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2356,6 +2356,31 @@ ostree_repo_open (OstreeRepo    *self,
   if (!gs_file_open_dir_fd (self->tmp_dir, &self->tmp_dir_fd, cancellable, error))
     goto out;
 
+  /* Make sure the tmp directory has full write permissions + the sticky
+   * bit (01777) so normal users of a system repo can still write temporary
+   * files in the repo.  Note, failure here is not fatal since most likely
+   * the permissions are already set correctly. */
+  if (fstat (self->tmp_dir_fd, &stbuf) == 0)
+    {
+      mode_t tmp_mode = S_IRWXU | S_IRWXG | S_IRWXO | S_ISVTX;
+      if (stbuf.st_mode != tmp_mode)
+        {
+          if (fchmod (self->tmp_dir_fd, tmp_mode) != 0)
+            {
+              /* Ignore permission errors.  This is expected
+               * if a normal user is acting on a system repo. */
+              if (errno != EPERM)
+                {
+                  perror ("ostree: Unable to set tmp directory permission");
+                }
+            }
+        }
+    }
+  else
+    {
+      perror ("ostree: Unable to get tmp directory status");
+    }
+
   if (self->mode == OSTREE_REPO_MODE_ARCHIVE_Z2 && self->enable_uncompressed_cache)
     {
       if (!gs_file_ensure_directory (self->uncompressed_objects_dir, TRUE, cancellable, error))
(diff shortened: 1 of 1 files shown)


alexlarsson:
I think this is right, but it needs more. We must also ensure that the created tmp sub directories are not writable by the non-user. Right now they are created 0777, which means someone could create tmpfiles in them and have your user pick them up. Also, we should probably not reuse directories with such permissions.

Also, isnt't there in general some risks with having a shared tmp dir that everyone can write to?
For instance, it seems like checkout_object_for_uncompressed_cache() creates temporary files in the "tmp" dir. At the moment anyone can remove, rename, replace or write to this file, racing with the rename into the uncompressed object cache. With the sticky bit set on the tmpdir we disallow remove, rename and replace, which is good. But we still create files with the final ownership in the tmp dir, and if that involves world writable then anyone can modify it while we're writing it.

Of course, if such a permission is there they could modify it in the final object store too, so maybe thats not necessarily a tmpdir issue. And in practice, most files would be root owned, and not world-writable, so its maybe not a realistic problem.

cgwalters:
I'm uncertain about this versus teaching the code to do a `mkdtemp` and write there if `getuid()` != repo owner.

One reason is that any unprivileged user could write there and take up space, and only  `cleanup_tmpdir()` which only happens on commits will clean it up.  In contrast for `/tmp` and `/var/tmp` systemd comes out of the box with tmpfiles snippets to automatically prune, and system administrators/scripts know to look there for stuff to delete.

mbarnes:
I need to verify this, but it may be that we only really need more open permissions for `_ostree_fetcher_request_uri_to_membuf()`.  I don't think those callers do atomic renames.  If that's true, using `/tmp` or `/var/tmp` should be fine.

alexlarsson:
@cgwalters Well, if this used to work, then the user used to be able to the tmpdir already. That said, maybe that *is* a bad thing and we should use /tmp for temporary files that do not need to be moved into the object dirs.

cgwalters:
I don't think we ever allowed non-root users to write to `/ostree/repo/tmp`.

alexlarsson:
Hos then did it use to work?

mbarnes:
I reread the code to answer that.  In the case of `_ostree_fetcher_request_uri_to_membuf()`, the `tmp` directory isn't used; the downloaded data stays in memory buffers.  This is controlled by the `is_stream` flag in the `OstreeFetcherPendingURI` struct.

What's tripping us up is the call to `glnx_make_lock_file()` via `_ostree_repo_allocate_tmpdir()` in `_ostree_fetcher_new()`.

mbarnes:
Possible solution could be to split out an `OstreeAsyncFetcher` subclass and move the async fetch functions there.  The subclass could then create its tmp dir, leaving `OstreeFetcher` to just deal in memory buffers.  Callers could then choose the appropriate class.

alexlarsson:
Maybe we could make the lock and tmpdir allocation lazy?

mbarnes:
> Maybe we could make the lock and tmpdir allocation lazy?

Yeah, that sounds easier; wait til the first call to `_ostree_fetcher_request_uri_with_partial_async()`.

mbarnes:
@cgwalters, @alexlarsson:
I submitted PR https://github.com/GNOME/ostree/issues/164 as an alternative to this one.  If we like that better I'll close this PR.

cgwalters:
Closing in favor of #164

PR #164, opened 2015-12-18
fetcher: Lazily create tmp directory

mbarnes:
This is an alternative to PR https://github.com/GNOME/ostree/pull/162 based on subsequent discussion.

The tmpdir is now lazily created from the new `SoupSession` thread instead of at instance initialization, so `OstreeFetcher` instances that only need to use `_ostree_fetcher_request_uri_to_membuf()` as a normal user won't encounter permission issues creating a tmpdir it doesn't need.

### DIFF
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -48,6 +48,10 @@ typedef struct {
   GMainLoop *main_loop;
 
   int tmpdir_dfd;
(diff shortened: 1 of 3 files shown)


cgwalters:
Ok, LGTM.  I am still having a feeling down the line it'd be easiest to do per-user in this case, but this is fine too.

cgwalters:
Merged, thanks!

PR #165, opened 2015-12-24
grub2_generate: load sysroot before using it

jlebon:
The logic for checking which bootversion to use tries to access
sysroot->bootversion if the user didn't specify an explicit bootversion
on the command-line nor through the env var. However, at that point, the
sysroot object is not yet initialized, so it will always return 0, even
when it's 1.

This would cause e.g. `grub2-mkconfig` to have no output for the BLS
entries whenever the entries were under `/boot/loader.1`.

Related: RHBZ1293986


cgwalters:
LGTM, thanks!

I would add that basically this made the OSTree grub2 code work only when ostree itself was upgrading the bootloader config, one couldn't generally reuse `grub2-mkconfig`.

PR #168, opened 2016-01-13
sysroot: Don't individually fsync dirs in checkout, rely on syncfs

cgwalters:
Originally, a lot of the `fsync()` calls here were added for the
wrong reason - I was chasing a bug that ended up being the extlinux
bootloader not parsing 64 bit ext4 filesystems.  But since it looked
like corruption, I tried adding a lot more `fsync()` calls.

All we should have to do is use `syncfs()`.  If that doesn't work,
it's a kernel bug.

I'm making this change because skipping the individual fsyncs can be a
major performance win - it's easier for the FS to optimize, we do more
in parallel, etc.

https://bugzilla.gnome.org/show_bug.cgi?id=757117

### DIFF
--- a/doc/ostree-sections.txt
+++ b/doc/ostree-sections.txt
@@ -220,6 +220,7 @@ ostree_repo_new_for_sysroot_path
 ostree_repo_new_default
 ostree_repo_open
 ostree_repo_set_disable_fsync
+ostree_repo_get_disable_fsync
 ostree_repo_is_system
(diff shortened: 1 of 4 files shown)


mbarnes:
Minor gripe about single-letter variables (`t`), but otherwise LGTM.  :+1:

cgwalters:
Fair enough, I hesitated there too.  I made it more verbose and pushed, thanks!

PR #169, opened 2016-01-13
repo: Note global transaction resume is legacy

cgwalters:
See docs for details.


mbarnes:
:+1:

cgwalters:
Updated :arrow_up: to only print if resuming.

mbarnes:
Yeah, that makes more sense.  :+1:

gatispaeglis:
+1

PR #170, opened 2016-01-13
[merged] repo: Clean up staging directory for previous boot IDs

cgwalters:
We had a policy of cleaning up all files in `$repo/tmp` older
than one day, but we should really clean up previous bootid staging
directories too, as they can potentially take up a lot of disk space.

https://bugzilla.gnome.org/show_bug.cgi?id=760531

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2562,6 +2562,16 @@ ostree_repo_open (OstreeRepo    *self,
       ostree_repo_set_disable_fsync (self, TRUE);
   }
 
+  { g_autofree char *tmp_expiry_seconds = NULL;
+
+    /* 86400 secs = one day */
+    if (!ot_keyfile_get_value_with_default (self->config, "core", "tmp-expiry-secs", "86400",
+                                            &tmp_expiry_seconds, error))
+      goto out;
+
+    self->tmp_expiry_seconds = g_ascii_strtoull (tmp_expiry_seconds, NULL, 10);
+  }
+
   if (!append_remotes_d (self, cancellable, error))
     goto out;
 
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2432,12 +2432,20 @@ ostree_repo_open (OstreeRepo    *self,
   /* We use a per-boot identifier to keep track of which file contents
    * possibly haven't been sync'd to disk.
    */
-  if (!g_file_get_contents ("/proc/sys/kernel/random/boot_id",
-                           &self->boot_id,
-                           NULL,
-                           error))
-    goto out;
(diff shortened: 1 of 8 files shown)


cgwalters:
NOTE!  Only lightly tested with existing unit tests, still needs more testing.

cgwalters:
There might be a higher level pattern here that we might want to hoist possibly into libglnx of "bootid-suffixed directory with cleanup of previous ones", since other components would likely want fsync-safe management of that.

On the other hand, maybe it's most reasonable for an http cache to do a journal-style single file with checksums etc. like the journal?  I need to look at what Firefox does.  librepo seems to checksum the files each time which is fairly lame.

cgwalters:
Rebased.  I will work more on this.  We should extend what we're doing now with magic environment variables like `OSTREE_SYSROOT_DEBUG` and support e.g. erroring out halfway through a pull.

cgwalters:
:surfer: Rebased again. Still working on the debug environment variables and tests.

cgwalters:
No longer WIP.  I think this is good for review :eyes: although it needs more testing :hammer:

src/libostree/ostree-repo.c
```diff
@@ -2548,6 +2548,16 @@ ostree_repo_open (OstreeRepo    *self,
...
+  { g_autofree char *tmp_expiry_seconds = NULL;
+
+    /* 86400 secs = one day */
+    if (!ot_keyfile_get_value_with_default (self->config, "core", "tmp-expiry-secs", "86400",
+                                            &tmp_expiry_seconds, error))
+      goto out;
+
+    self->tmp_expiry_seconds = g_ascii_strtoull (tmp_expiry_seconds, NULL, 10);
+  }
+
```
jlebon: Nice!

src/libostree/ostree-repo.c
```diff
@@ -2415,22 +2415,26 @@ ostree_repo_open (OstreeRepo    *self,
   if (self->inited)
     return TRUE;
 
-  /* We use a per-boot identifier to keep track of which file contents
-   * possibly haven't been sync'd to disk.
+  /* We use a dirctory named `staging-${BOOT_ID}-${RANDOM}` identifier
```
jlebon: Nit: typo here

jlebon:
Nice straight-forward cleanups. Small typo in one of the comments, but go ahead if you just want to merge it.

cgwalters:
Fixup :arrow_up:

cgwalters:
Oops, nope.

cgwalters:
Fixed up the wrong commit, so I just went ahead and rebased/squashed.

jlebon:
@cgwalters-bot r+ eed667d

jlebon:
@cgwalters-bot r+ 4113080

PR #172, opened 2016-01-23
Add a checkout option to skip fsync

cgwalters:
This is a better followup to dc9239dd7b09ef5e104309b4dbf0e136889da274
since I wanted to do fsync-less checkouts in rpm-ostree too, and
replicating the "turn off fsync temporarily" was in retrospect just a
hack.

We can simply add a boolean to the checkout options.


jlebon:
Makes sense!

cgwalters:
Merged, thanks!

PR #173, opened 2016-01-27
repo: Port -refs.c to openat()

cgwalters:
I'd like to incrementally convert all of `ostree-repo*.c` to
fd-relative usage, so that we can sanely introduce
`ostree_repo_new_at()` which doesn't involve GFile.

This one is medium risk, but passes the test suite.


giuseppe:
:+1:

cgwalters:
Merged, thanks!

PR #174, opened 2016-01-27
tests: Use "bash strict mode"

cgwalters:
I noticed in the static deltas tests, there were some tests that
should have been under `-o pipefail` to ensure we properly propagate
errors.

There were a few places where we were referencing undefined variables.

Overall, this is clearly a good idea IMO.


giuseppe:
yes it is definitely a good idea.  :+1:

PR #176, opened 2016-01-28
static-delta: Add `show` subcommand

cgwalters:
Right now though, almost all of the details of deltas are private, so
we can't do the "honest thing" and have the command line just use the
shared library.

Eventually some of this should appear in the API, but for now add
command line which is useful for debugging.

### DIFF
--- a/src/libostree/ostree-repo-static-delta-core.c
+++ b/src/libostree/ostree-repo-static-delta-core.c
@@ -22,6 +22,7 @@
 
 #include "ostree-core-private.h"
 #include "ostree-repo-private.h"
+#include "ostree-cmdprivate.h"
 #include "ostree-repo-static-delta-private.h"
 #include "otutil.h"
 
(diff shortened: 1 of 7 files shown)


src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -404,3 +405,95 @@ ostree_repo_static_delta_execute_offline (OstreeRepo                    *self,
...
+    goto out;
+
+  g_print ("Delta: %s\n", delta_id);
+  { guint64 ts;
+    g_variant_get_child (delta_superblock, 1, "t", &ts);
+    g_print ("Timestamp: %" G_GUINT64_FORMAT "\n", GUINT64_FROM_BE (ts));
+  }
+  { g_autoptr(GVariant) recurse = NULL;
+    g_variant_get_child (delta_superblock, 5, "@ay", &recurse);
+    g_print ("Number of parents: %u\n", (guint)(g_variant_get_size (recurse) / (32 * 2)));
```
mbarnes: `(32 * 2)` <- Would it be correct to use `OSTREE_STATIC_DELTA_OBJTYPE_CSUM_LEN` here?  (which is 33, not 32)

I feel like that needs either a named constant or a comment.

mbarnes:
@giuseppe is more the expert here, but aside from the one readability thing above, LGTM.

src/libostree/ostree-repo-static-delta-core.c
cgwalters: It is confusing =)  The `33` below this was hardcoded, and I changed it to use the constant.  The 32 there though is right, because it doesn't have an object type.  Nevertheless I submitted a PR to create a constant: https://github.com/GNOME/ostree/pull/177

mbarnes:
:+1:

cgwalters:
Merged, thanks!

PR #177, opened 2016-01-28
lib: Add a #define OSTREE_SHA256_DIGEST_LEN 32

cgwalters:
And use it internally.  This way it's a bit less magical.


mbarnes:
Nice readability win.  :+1:

cgwalters:
Merged, thanks!

PR #178, opened 2016-01-29
build: 'make clean' removes parse-datetime.c

giuseppe:
and fix make dist while at it.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -182,7 +182,7 @@ pkgconfig_DATA += src/libostree/ostree-1.pc
 
 gpgreadme_DATA = src/libostree/README-gpg
 gpgreadmedir = $(pkgdatadir)/trusted.gpg.d
-EXTRA_DIST += src/libostree/README-gpg src/libostree/bupsplit.h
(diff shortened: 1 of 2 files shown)


Makefile-libostree.am
```diff
@@ -182,7 +182,7 @@ pkgconfig_DATA += src/libostree/ostree-1.pc
 
 gpgreadme_DATA = src/libostree/README-gpg
 gpgreadmedir = $(pkgdatadir)/trusted.gpg.d
-EXTRA_DIST += src/libostree/README-gpg src/libostree/bupsplit.h
+EXTRA_DIST += src/libostree/README-gpg src/libostree/bupsplit.h src/libostree/ostree-enumtypes.h.template src/libostree/ostree-enumtypes.c.template src/libostree/ostree-deployment-private.h
```
mbarnes: Minor, but maybe split this into multiple lines?

mbarnes:
I know we don't use `make dist` for releases but I see no harm in keeping it working.  :+1:

cgwalters:
Yeah I just sign git tags, but I don't mind changing `EXTRA_DIST` either.  Changing `CLEANFILES` makes sense for sure.  So agreed with @mbarnes :+1:

giuseppe:
thanks, I've split the long line and pushed the patch.

PR #179, opened 2016-01-30
Add an `export` builtin, and API to write to libarchive

cgwalters:
At the moment I'm looking at using rpm-ostree to manage RPM inputs
which can then be converted into Docker images.  It's most convenient
if we can stream directly out of libostree rather than doing a
checkout + tar combination.

There are also backup/debugging etc. reasons to implement `export` as
well.


cgwalters:
All comments addressed except I didn't change the `#ifdef` order as we do that elsewhere in the code.

jlebon:
LGTM!

cgwalters:
Merged, though I changed a few things:

 - Added a man page
 - Use GNU tar format instead of PAX, as well, GNU `/bin/tar` barfs on libarchive PAX xattrs

giuseppe:
the final commit 355f8438ef4fe2720d25e8241aa76797b7029522 is missing the file ```man/ostree-export.xml```

cgwalters:
Oops, fixed, thanks.

PR #180, opened 2016-02-06
Deltas are stable

cgwalters:
Some work on stabilizing deltas...actually just expanding `show` and cleaning up the internals.  I'm trying to do another audit of the format.


giuseppe:
:+1:  looks fine to me.  Nevermind my comment, I changed idea about that and it makes sense to not change the configure output format.

cgwalters:
This was merged.

PR #181, opened 2016-02-09
fetcher: Fix hung GTlsInteraction

mbarnes:
The `GTlsInteraction` instance must be created in the session thread so it uses the correct `GMainContext`.

@miabbott reported this.  Rather urgent.


cgwalters:
:+1:

miabbott:
Woo!  Thanks for the quick turnaround @mbarnes!

PR #182, opened 2016-02-10
Import rofiles-fuse

cgwalters:
While it's not strictly tied to OSTree, let's move
https://github.com/cgwalters/rofiles-fuse in here because:

 - It's *very* useful in concert with OSTree
 - It's tiny
 - We can reuse OSTree's test, documentation, etc. infrastructure

One thing to consider also is that at some point we could experiment
with writing a FUSE filesystem for OSTree.  This could internalize a
better equivalent of `--link-checkout-speedup`, but on the other hand,
the cost of walking filesystem trees for these types of operations is
really quite small.

But if we did decide to do more FUSE things in OSTree, this is a step
towards that too.


jlebon:
The integration itself LGTM.

Any reason why you didn't go with a submodule? Are you planning to delete the original repo?

cgwalters:
Yeah, planning to delete the original repo.  Ultimately, this code should live in the kernel I think.  It's like a more restricted overlayfs.  See also http://www.spinics.net/lists/linux-fsdevel/msg75085.html

PR #183, opened 2016-02-11
pull: Support specifying exact commit to pull

cgwalters:
I don't know why we didn't do this a long time ago.  This extends the
pull API to allow grabbing a specific commit, and will set the branch
to it.  There's some support for this in the deploy engine, but there
are a lot of reasons to support it for raw pulls (such as subset
mirroring cases).

In fact I'm thinking we should also have the override-version logic
here too.

NOTE: One thing I debated here is inventing a new syntax on the
command line; you do: `ostree pull remotename branch:commitid`.  We
already use the colon `:` character to separate `remotename:branch`
too.  But a new character would be weird too.

Anyways, I wanted this for some other test cases.  Without this,
writing tests that go between different commits is more awkward as one
must generate the content in one repo, then pull downstream, then
generate more content, then pull again.  But now I can just keep track
of commit IDs and do exactly what I want without synchronizing the
tests.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1762,6 +1762,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   OstreeRepoPullFlags flags = 0;
   const char *dir_to_pull = NULL;
   char **refs_to_fetch = NULL;
+  char **override_commit_ids = NULL;
   GSource *update_timeout = NULL;
   gboolean disable_static_deltas = FALSE;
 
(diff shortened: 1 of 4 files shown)


jlebon:
This is mostly LGTM, although I wonder about this:

> NOTE: One thing I debated here is inventing a new syntax on the command line; you do: `ostree pull remotename branch:commitid`. We already use the colon `:` character to separate `remotename:branch` too. But a new character would be weird too.

In which situations do we use the `remotename:branch` syntax? Could we foresee a situation where a command argument could accept both `remotename:branch` and `branch:commitid` formats? E.g. when a ref is wanted, a `remotename:branch` should probably refer to the HEAD of that branch. But now you have the overhead of checking whether it could be interpreted the other way.

I wonder if using `@` would work better?

cgwalters:
Merged with a change to use '@', thanks for the suggestion!

PR #185, opened 2016-02-14
Support Docker-style whiteouts

cgwalters:
This is to enable importing Docker layers as ostree commits, then
checking them out in a union.


cgwalters:
Fixed that and some other bits in :arrow_up:.

cgwalters:
Merged, thanks!

PR #187, opened 2016-02-19
ostree-sysroot: add debug option to help testing

jlebon:
If ostree is run in a test setup where it operates as root in a tmp
directory, it might cause issues to flag the deployments as immutable.
The test harness might simply be doing an `rm -rf` (effectively the case
for gnome-desktop-testing-runner), which will then fail.

We add a new debug option to the ostree_sysroot object using GLib's
GDebugKey functionality to allow our tests to communicate to ostree that
we don't want immutable deployments.


jlebon:
This is my first foray into GLib and the GObject system, so don't hesitate to let me know if there's a better way to do something.

cgwalters:
LGTM!

PR #189, opened 2016-02-22
libarchive: Make autocreate_parents imply autocreating root dir

cgwalters:
Some Docker layers are just metadata in the `layer.json`.  If one is
mapping Docker layers to OSTree commits, one needs to create a dummy
root directory, because OSTree doesn't support metadata-only commits.

Let's just push that logic down here because it's easier than special
casing it in higher levels.


cgwalters:
This was merged.

PR #190, opened 2016-02-22
Gpg threadsafe

cgwalters:
Import of https://bugzilla.gnome.org/show_bug.cgi?id=761863 with a patch on top.


mbarnes:
Gah, meant to follow up on the `gpg_strerror_r` thing myself but forgot.

:+1:

cgwalters:
Merged, thanks.

PR #191, opened 2016-02-23
deltas: Include an endianness marker

cgwalters:
We screwed up and had delta integers use host endianness.  Start
digging out by at least annotating the endianness.

https://bugzilla.gnome.org/show_bug.cgi?id=762515


alexlarsson:
Why lowercase 'l' but uppercase 'B' ?

cgwalters:
I think the idea is it's little lowercase, and a big uppercase.  Anyways, it's what DBus uses:
https://dbus.freedesktop.org/doc/dbus-specification.html

```
1st BYTE	Endianness flag; ASCII 'l' for little-endian or ASCII 'B' for big-endian. Both header and body are in this endianness.
```

cgwalters:
This is one of those times where I kind of regret not using Protocol Buffers for this.  Oh well...

alexlarsson:
Anyway, that looks good to me, although we now need to actually use it :)

cgwalters:
Using it in :arrow_up:

alexlarsson:
Looks good to me, although you could handle the "unknown" case by looking at all sizes, and if say a majority is > 32bit, then assume byte-swapping is needed.

cgwalters:
Yeah, about to do autodetection.

cgwalters:
Now with heuristics :game_die: based on object size in :arrow_up:

alexlarsson:
Another heuristics is to look at the compressed and uncompressed size, and if the compressed is larger, then you probably need to byteswap them.

cgwalters:
For some small files, it's possible that compression makes them larger, as you still need to add a compression header.  But yeah...for these deltas it should really never happen.  Added an implementation in :arrow_up:

alexlarsson:
LGTM

cgwalters:
Merged, thanks!

PR #192, opened 2016-02-29
giuseppe/multiple refs

giuseppe:
I am not 100% sure about the change `ostree_repo_list_refs: include the prefix in the refs listing` and what down sides it can have, but that seems to be like the place where the issue described in the its commit message should be fixed.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -61,6 +61,7 @@ testfiles = test-basic \
 	test-xattrs \
 	test-auto-summary \
 	test-prune \
+	test-refs \
 	$(NULL)
 
 if BUILDOPT_FUSE
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -260,6 +260,7 @@ ostree_repo_write_content_async
 ostree_repo_write_content_finish
 ostree_repo_resolve_rev
 ostree_repo_list_refs
+ostree_repo_list_refs_ext
 ostree_repo_remote_list_refs
 ostree_repo_load_variant
 ostree_repo_load_commit
--- a/man/ostree-refs.xml
+++ b/man/ostree-refs.xml
@@ -57,7 +57,7 @@ Boston, MA 02111-1307, USA.
     <refsect1>
         <title>Description</title>
         <para>
-            Lists all refs available on the host.  If pecified, PREFIX assigns the refspec prefix; default prefix is null, which lists all refs.
+            Lists all refs available on the host.  If specified, PREFIX assigns the refspec prefix; default prefix is null, which lists all refs.
         </para>
     </refsect1>
 
--- a/man/ostree-refs.xml
+++ b/man/ostree-refs.xml
@@ -66,6 +66,13 @@ Boston, MA 02111-1307, USA.
 
         <variablelist>
             <varlistentry>
+                <term><option>--list</option></term>
+
+                <listitem><para>
+                  If a refspec is specified, do not remove it from the output.
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
                 <term><option>--delete</option></term>
 
                 <listitem><para>
--- a/src/libostree/libostree.sym
+++ b/src/libostree/libostree.sym
@@ -181,6 +181,7 @@ global:
         ostree_repo_list_commit_objects_starting_with;
         ostree_repo_list_objects;
         ostree_repo_list_refs;
+        ostree_repo_list_refs_ext;
         ostree_repo_list_static_delta_names;
         ostree_repo_load_commit;
         ostree_repo_load_file;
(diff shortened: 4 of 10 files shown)


mbarnes:
All looks good to me.  :+1:

mbarnes:
This could be a good follow-up: https://bugzilla.gnome.org/show_bug.cgi?id=746049

giuseppe:
I am mostly scared by https://github.com/giuseppe/ostree/commit/a34b95aedd4123e73f8393d6e8b5114947f67f55 and if it can break anything in the deploy phase.

giuseppe:
I've tested it and it seems to work, good to merge?

cgwalters:
I thought this code was fishy when I was doing https://github.com/GNOME/ostree/pull/173 - in fact I almost made a similar change then.  I'm a bit nervous though about changing things...let me see.

cgwalters:
I pushed the last two cleanup commits.

cgwalters:
The original code is crummy and your patches improve it...but my gut instinct is to create a new fixed API (and command line switch?)  The problem is I could imagine someone is relying on this.

While we're here let's make it extensible with a `flags` argument or so, something like:

```
ostree_repo_list_refs_ex (OstreeRepo       *self,
                       const char       *refspec_prefix,
                       guint flags,  /* Unused for now */
                       GHashTable      **out_all_refs,
                       GError          **error)
```

Then the command line is something like:

```
  ostree refs --list-all
  ostree refs --delete foo
  ostree refs --delete-matching-regexp foo.*
```

?

giuseppe:
my  concern is that the API can be used in the same bad way as it is currently used by `ostree refs --delete`.  Iterating the current refs and then delete some of them looks like a common scenario, e.g. implement a higher level logic for branches pruning.

`ostree_repo_list_refs` will return the same value `foo` both in case there is a single branch called `foo` or in the case there are two branches `foo` and `foo/foo` (a bit unrealistic but possible :))  In this case the user should first check if `foo/foo` exists and if it doesn't then assume it is the branch `ref`.  As of this not deterministic behavior, any usage of the API should be audited, so IMHO, I would rather break it clearly in the cases where this behavior is currently used.

Said so, I can make `ostree_repo_list_refs` a wrapper around a new `ostree_repo_list_refs_ex` function using a flag like: `OSTREE_REPO_LIST_OPTION_PRUNE_REFSPEC`.  What do you think?

cgwalters:
An API user can avoid breakage by simply not using the filtering inside of `list_refs` right?  I.e. pass `NULL` for `refspec_prefix`, then use `g_str_has_prefix()` ?

But clearly that's a bit lame...so we should have a new API.  I don't think we need a new flag, let's just fix it to not return substrings.

BTW, the reason your patch breaks the `ostree/` ref cleanup handling is simple; look at `cleanup_ref_prefix()`, it does:

```
      g_autofree char *ref = g_strconcat (prefix, "/", suffix, NULL);
      ostree_repo_transaction_set_refspec (repo, ref, NULL);
```

But that will now become `ostree/1/1/ostree/1/1/foo` and then do nothing.

So yeah, I think new API.

cgwalters:
(And port `cleanup_ref_prefix()` to use it)

giuseppe:
rebased and reworked the commits :arrow_up:

giuseppe:
I keep forgetting about `ostree-sections.txt` as well :)  I had to push the whole series again as I had some merge conflicts that I could not address in fixup commits.

About the tests, I have added `--canonicalize` and a test for it, it uses the new API.  Is it fine for now?  I can add more tests in C

cgwalters:
This looks a lot better if we're keeping the previous cmdline the same as before as well, just with new additions.  I may be being overly conservative here, but it's just so hard to know what we would break with a subtle change like this.

Let's at least briefly think about terms...`--canonicalize` isn't bad.  But what did you think about my `--list` suggestion?  It's fewer characters...and we can basically get people to ignore plain `ostree refs`.

cgwalters:
Hold off on any major test changes...I'm working on that now.

cgwalters:
Got it, OK.  Makes sense to me then, :+1:

giuseppe:
would `--list-all` work in the same way as `--canonicalize`?

cgwalters:
Yeah, that's what I was thinking.  What do you think?

giuseppe:
I am personally more used to/inclined towards `--canonicalize` but I type it wrong 2/3 of the times so I have no strong opinion about it, I will change it to `--list-all`

cgwalters:
`--canonicalize` is a better description of what it does, but what I'm thinking is we basically un-document/promote `ostree refs` and create a symmetry where you have `--list` and `--delete` as operations.  So someone new doesn't know that `ostree refs` is a hidden version that strips off the prefix.

(Should probably be `--list` and not `--list-all` since we do want to support sub-listings with this)

giuseppe:
renamed here :arrow_up:

cgwalters:
One issue; the new symbol actually needs to be in the `2016.4` list.  I'll fix that.

cgwalters:
Merged, thanks!

PR #194, opened 2016-03-02
tests: Port to glib-tap.mk, make `make check` run all of the tests 

cgwalters:
OSTree's code for testing predates the `glib-tap.mk` making its
way into GLib.  Let's switch to it, as it provides a number
of advantages.

By far the biggest advantage is that `make check` can start to run
most of the tests *in addition* to having them work installed.

This commit keeps the installed tests working, but `make check` turns
out to be really broken because...our TAP usage has bitrotted to say
the least.  Fix that all up.

Do some hacks so that the tests work uninstalled as well - in
particular, `glib-tap.mk` and the bits encoded into
`g_test_build_filename()` assume *recursive* Automake (blah).  Work
around that by creating a symlink when installed to loop back.


jlebon:
This needs to be rebased.

cgwalters:
Rebased.

jlebon:
Does `make check` pass on your end?

I get an error in `test-libarchive-import`:

```
FAIL: tests/test-libarchive-import
==================================

/libarchive/noautocreate-empty: OK
/libarchive/autocreate-empty: OK
/libarchive/error-device-file: OK
/libarchive/ignore-device-file: **
ERROR:tests/test-libarchive-import.c:234:test_libarchive_ignore_device_file: assertion failed (error == NULL): Failed to execute child process "ostree" (No such file or directory) (g-exec-error-quark, 8)
FAIL tests/test-libarchive-import (exit status: 134)
```

cgwalters:
Oh, right...it's currently assuming the binary is already installed.  Will look at handling uninstalled too.

cgwalters:
Fixup to add to $PATH.

giuseppe:
:+1:  from me with this amended:

```
diff --git a/cfg.mk b/cfg.mk
index a0ce8fa..5c3f0b4 100644
--- a/cfg.mk
+++ b/cfg.mk
@@ -1,4 +1,4 @@
-export VC_LIST_EXCEPT_DEFAULT=^(lib/.*|m4/.*|md5/.*|build-aux/.*|src/gettext\.h|.*ChangeLog)$$
+export VC_LIST_EXCEPT_DEFAULT=^(lib/.*|m4/.*|md5/.*|build-aux/.*|src/gettext\.h|.*ChangeLog|buildutil/.*)$$
 
 local-checks-to-skip = \
     sc_const_long_option \
diff --git a/tests/test-commit-sign.sh b/tests/test-commit-sign.sh
index b7d3fc6..8d52bca 100755
--- a/tests/test-commit-sign.sh
+++ b/tests/test-commit-sign.sh
@@ -20,7 +20,8 @@
 set -euo pipefail
 
 if ! ostree --version | grep -q -e '\+gpgme'; then
-    exit 77
+    echo "1..0 #SKIP no gpg support compiled in"
+    exit 0
 fi
 
 . $(dirname $0)/libtest.sh
diff --git a/tests/test-gpg-signed-commit.sh b/tests/test-gpg-signed-commit.sh
index 9068ea5..a0bf983 100755
--- a/tests/test-gpg-signed-commit.sh
+++ b/tests/test-gpg-signed-commit.sh
@@ -21,7 +21,8 @@
 set -euo pipefail
 
 if ! ostree --version | grep -q -e '\+gpgme'; then
-    exit 77
+    echo "1..0 #SKIP no gpgme support compiled in"
+    exit 0
 fi
 
 . $(dirname $0)/libtest.sh
diff --git a/tests/test-libarchive.sh b/tests/test-libarchive.sh
index 98e31c8..7309ffd 100755
--- a/tests/test-libarchive.sh
+++ b/tests/test-libarchive.sh
@@ -20,7 +20,8 @@
 set -euo pipefail
 
 if ! ostree --version | grep -q -e '\+libarchive'; then
-    exit 77
+    echo "1..0 #SKIP no libarchive support compiled in"
+    exit 0
 fi
 ```

**Comment by jlebon** [comment]
I seem to be getting LD_PRELOAD issues with a couple of them: [output](http://paste.fedoraproject.org/332995/01755614/) & [test-suite.log](http://paste.fedoraproject.org/332996/01759514/).

That said, since this is "opening" up a whole new slew of tests in a different environment, I don't mind having those issues fixed up separately.

**Comment by cgwalters** [comment]
This was merged, but will keep looking at issues here.

**Comment by cgwalters** [comment]
It looks like most of these issues are also of the form @giuseppe identified - we need to switch from `exit 77` to `echo... # SKIP`.

**Comment by cgwalters** [comment]
There's also one where you're missing `setfattr` - I'd just install that.

**Comment by cgwalters** [comment]
For the `LD_PRELOAD`...yeah, I have seen that too, haven't debugged it.  I'm tempted to go with a FUSE mount instead.

**Comment by cgwalters** [comment]
Pushed :arrow_up: which should fix two of them.

**Comment by jlebon** [comment]
Hmm, did you forget an `exit 0` in there? :)

**Comment by jlebon** [comment]
After installing `attr` and adding `exit 0` to that commit, the only remaining failure is:

```
ERROR: tests/test-basic.sh - too few tests run (expected 49, got 48)
```

Looking at `basic-test.sh`, I see at the very end

```bash
# Run this test only as non-root user.  When run as root, the chmod
# won't have any effect.
if test "$(id -u)" != "0"; then
    <snip>
    echo "ok unwritable repo was caught"
fi
```

Maybe we can just add an else to explicitly SKIP when root?

cgwalters:
Oops, yes.

PR #196, opened 2016-03-03
lib: Add ostree_sysroot_init_osname() API, bump mtime

cgwalters:
And change the command line to use it.  rpm-ostree had a copy
of this code, and thus there's a clear reason to have an API.

While we're moving this into API, ensure the mtime on deploy is bumped
after an osname is created, so that daemons like rpm-ostree can notice
changes.  (In reality, creating the directory should do this, but
let's be double sure)


mbarnes:
Looks nice.  Update `ostree-sections.txt` and :+1:

cgwalters:
Thanks, merged with that change.

PR #197, opened 2016-03-03
tests: Unify some tmpdir code, add ability for C to use libtest.sh

cgwalters:
I want to be able to easily test the C API on actual data in an OSTree
repo.  The shell `libtest.sh` has code to generate it.  Bridge the two
worlds by introducing a little `libostreetest` library which has a C
API which spawns a shell that runs things in `libtest.sh`.

Yes, this is about as beautiful as it sounds, which is to say, it's
not.  But it works!

Note while we were here, I realized we were actually now creating
*two* tmpdirs per test in `make check` because the tap driver was
already doing that.  Unify it so we know the C code can rely on it.


jlebon:
This is hilarious.

cgwalters:
Weird, I must have accidentally corrupted it before pushing.  Just fixed it in :arrow_up:

jlebon:
LGTM!

cgwalters:
Merged, thanks.

PR #198, opened 2016-03-03
tests: More TAP fixups

cgwalters:
Hopefully getting closer now.

### DIFF
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -425,6 +425,8 @@ if test "$(id -u)" != "0"; then
     assert_has_file expected-fail
     assert_file_has_content error-message "Permission denied"
     echo "ok unwritable repo was caught"
+else
+    echo "ok # SKIP not run when root"
 fi
 
 cd ${test_tmpdir}
--- a/tests/test-pull-corruption.sh
+++ b/tests/test-pull-corruption.sh
@@ -22,6 +22,7 @@ set -euo pipefail
 # If gjs is not installed, skip the test
 if ! gjs --help >/dev/null 2>&1; then
     echo "1..0 # SKIP no gjs"
+    exit 0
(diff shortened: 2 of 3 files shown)


cgwalters:
Oops, this was in my tree.  Review after commit if desired.

jlebon:
Yes, almost good now. Got a new failure now in `test-basic.sh`:

```
# output
<snip>
PASS: tests/test-basic.sh 48 whiteouts disabled
SKIP: tests/test-basic.sh 49 # SKIP not run when root
ERROR: tests/test-basic.sh - exited with status 1

# relevant test-suite.log section
++ stat '--printf=%Z\n' /var/tmp/tap-test.75JHzz/repo
++ cd ..
++ cmp timestamp-orig.txt timestamp-new.txt
881c324b7ec179fa954ce91470986a36e28c15d6d9034b0ffe3c84acc5fc2702
++ assert_not_reached 'failed to update mtime on repo'
++ echo failed to update mtime on repo
failed to update mtime on repo
++ exit 1
timestamp-orig.txt timestamp-new.txt differ: char 10, line 1
ERROR: tests/test-basic.sh - exited with status 1
```

giuseppe:
is there a way to skip the tests file altogether once the tests plans was printed?  We will probably need that on -o pipefail when the script terminates before all tests run

cgwalters:
I think the only option is to skip each individual test once the plan is printed =/  But can't we structure things so in this case we skip beforehand?

cgwalters:
See https://github.com/GNOME/ostree/pull/201

giuseppe:
I was just thinking when a test fails for any reason, then the error we got from the TAP driver is a bit unclear.  Should we run each test in a subshell so the process doesn't exit immediately?

PR #200, opened 2016-03-04
lib: Add ostree_sysroot_load_if_changed() API

cgwalters:
This will allow daemons like rpm-ostree to detect if there are any new
deployments efficiently, in combination with using inotify.  If there
are any changes, rpm-ostree wants publish them on DBus.

While we're here, add some changes to start doing unit C testing of
the sysroot API.


cgwalters:
Merged, thanks!

PR #201, opened 2016-03-04
tests/basic: Fix race in timestamp test

cgwalters:
%Z only uses seconds, so it's possible that we did the commit
in the same second, which made this test racy.

- Switch to full nanosecond precision using '%.Y' so it always differs
- Fix the inverted `cmp` usage
- Add a missing `ok`


jlebon:
Yup, that works for me!

PR #202, opened 2016-03-08
build: Don't install test data without --enable-installed-tests

cgwalters:
Otherwise $(installed_testdir) is empty so we try to put content in
`/`, which I noticed when trying to build an RPM (it works works fine
`sudo make install`).


mbarnes:
Makes sense.  :+1:

PR #203, opened 2016-03-11
summary cache

giuseppe:
store a cache of the summary file

https://bugzilla.gnome.org/show_bug.cgi?id=762973

### DIFF
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -316,5 +316,23 @@ _ostree_repo_read_bare_fd (OstreeRepo           *self,
 gboolean
 _ostree_repo_update_mtime (OstreeRepo        *self,
                            GError           **error);
-                           
+
+/* Load the summary from the cache if the provided .sig file is the same as the
+   cached version.  */
+gboolean
+_ostree_repo_load_cache_summary_if_same_sig (OstreeRepo        *self,
+                                             const char        *remote,
+                                             GBytes            *summary_sig,
+                                             GBytes            **summary,
+                                             GCancellable      *cancellable,
+                                             GError           **error);
+
+gboolean
+_ostree_repo_cache_summary (OstreeRepo        *self,
+                            const char        *remote,
+                            GBytes            *summary,
+                            GBytes            *summary_sig,
+                            GCancellable      *cancellable,
+                            GError           **error);
+
 G_END_DECLS
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -33,6 +33,8 @@ G_BEGIN_DECLS
 
 #define _OSTREE_OBJECT_SIZES_ENTRY_SIGNATURE "ay"
 
+#define _OSTREE_SUMMARY_CACHE_PATH "tmp/cache/summaries"
+
 /**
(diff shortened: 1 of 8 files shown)


mbarnes:
Overall LGTM.  Maybe wait for @cgwalters to have a look before merging.

cgwalters:
Overall, I'd say using the signature file as a modification hook is fine, but a better solution (can be in addition) is to properly do [ETag](https://en.wikipedia.org/wiki/HTTP_ETag) like @alexlarsson suggested.  I think this would require integration with fetcher/libsoup.

giuseppe:
fixup patches here :arrow_up:

giuseppe:
I need the file name without the extension `.sig` if present for the lookup in the self->remotes hash table.  If you prefer, I can rewrite it to use just a pointer to set '\0' and then set back its original value.  I used a named buffer as I thought it was more readable (at the expense of some extra copying)

giuseppe:
I can replace it with the following if you prefer:

```
diff --git a/src/libostree/ostree-repo-prune.c b/src/libostree/ostree-repo-prune.c
index c919d5b..3384883 100644
--- a/src/libostree/ostree-repo-prune.c
+++ b/src/libostree/ostree-repo-prune.c
@@ -118,7 +118,6 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
                         GError **error)
 {
   gboolean ret = FALSE;
-  gchar remote[NAME_MAX + 1];
   g_auto(GLnxDirFdIterator) dfd_iter = { 0, };
   glnx_fd_close int fd = -1;
 
@@ -137,7 +136,8 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
 
   while (TRUE)
     {
-      char *suffix;
+      size_t len;
+      gboolean has_sig_suffix = FALSE;
       struct dirent *dent;
 
       if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
@@ -146,13 +146,19 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
       if (dent == NULL)
         break;
 
-      g_strlcpy (remote, dent->d_name, sizeof (remote));
-      suffix = g_strstr_len (remote, -1, ".sig");
-      if (suffix)
-        suffix[0] = '\0';
+      len = strlen (dent->d_name);
+      if (len > 4 && g_strcmp0 (dent->d_name + len - 4, ".sig") == 0)
+        {
+          has_sig_suffix = TRUE;
+          dent->d_name[len - 4] = '\0';
+        }
 
-      if (!g_hash_table_contains (self->remotes, remote))
+      if (!g_hash_table_contains (self->remotes, dent->d_name))
         {
+          /* Restore the previous value to get the file name.  */
+          if (has_sig_suffix)
+            dent->d_name[len - 4] = '.';
+
           if (unlinkat (fd, dent->d_name, 0) < 0)
             {
               glnx_set_error_from_errno (error);
```

cgwalters:
Oh right.  Mmm...yeah, hard to make this beautiful.  I'm fine with either way.  Sorry about being overly picky initially.

giuseppe:
rebased and pushed to master.

alexlarsson:
This was broken: https://github.com/GNOME/ostree/pull/204

PR #206, opened 2016-03-16
contrib/golang: Initial golang bindings

cgwalters:
We were considering using this for Docker integration, but we may end
up going a different architectural path.  Anyways, it doesn't hurt to
have the bindings in here - they can do a few things.

I decided to fork some of the core code from
https://github.com/dradtke/gotk3 because...well, what we really need a
GIR-based core generator but I didn't want to start on the fully
correct thing until we knew we wanted it, and this was a quick hack.

Also, let's make a `contrib/` directory for things like this.

### DIFF
--- a/contrib/golang/COPYING
+++ b/contrib/golang/COPYING
@@ -1,0 +1,17 @@
+Portions of this code are derived from:
+
+https://github.com/dradtke/gotk3
+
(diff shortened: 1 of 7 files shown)


giuseppe:
:+1:

cgwalters:
Pushed, thanks!

PR #209, opened 2016-03-18
tests: Strengthen test tmpdir sanity check, be compat with ginsttest …

cgwalters:
…saving

GNOME Continuous uses ginstest-runner --report-directory, which causes
the tests to save their tmpdirs persistently.  This also means the
result directories didn't match the `/(var/)?tmp` regexp, which broke
the ostree tests in GContinuous.

Fix this by simply asserting that the tmpdir either has `.tmpdir` or
nothing.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -25,13 +25,14 @@ assert_not_reached () {
 
 test_tmpdir=$(pwd)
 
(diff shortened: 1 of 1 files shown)


jlebon:
LGTM!

PR #210, opened 2016-03-18
admin: Add a `dup` command

cgwalters:
I'm trying to improve the developer experience on OSTree-managed
systems, and I had an epiphany the other day - there's no reason we
have to be absolutely against mutating the current rootfs live.  The
key should be making it easy to rollback/reset to a known good state.

This new `dup` command isn't a full solution by any means, but it's
a step towards a workflow that would look like this:

```
ostree admin dup          # clone the current root
ostree admin dev-unlock   # remove the ro bind mount, and add rofiles-fuse
ostree admin dev-reset    # not sure of the name...but basically resync
                          # files in the current root from a checkout
```


jlebon:
Hmm, interesting idea.
Would this reset also delete new files created in the mount?

jlebon:
I feel like it would be safer to not support a "reset" operation and just force users to reboot to "relock" it. Otherwise you lose that nice property of full correspondence between the root and the repo.

Then we could combine all the ops to just one, i.e. `ostree admin unlock`, which would dup, then remount, then use rofiles. Then to "relock" , they just reboot.

cgwalters:
Hum, so the "dup" in this `unlock` would actually just be re-deploy, we wouldn't reorder it for rollback.  I see where you're going, but I'm a bit uncertain about equating reboot = revert.  If you think about this from a hotfix perspective, you might still need to reboot the machine but not lose your hotfix.

Yes, I'm aware that hotfix != development exactly...but if we were to have a hotfix workflow (and I think we should), it would at least be able to make use of the current `dup`.
 
On the other hand I definitely like the idea of one simple command for the `dev-unlock` workflow.  In fact maybe we have both `dev-unlock` and `hotfix-unlock` ?

cgwalters:
One interesting angle on this...if dev/hotfix is noted in the origin file, things like `rpm-ostree pkg-add` could know about it.  Then if you e.g. `pkg-add` in a `hotfix` or `dev` deployment, it automatically becomes live?

jlebon:
> Hum, so the "dup" in this unlock would actually just be re-deploy, we wouldn't reorder it for rollback.

Oh right, of course. So you wouldn't even need to do that.

> On the other hand I definitely like the idea of one simple command for the dev-unlock workflow. In fact maybe we have both dev-unlock and hotfix-unlock ?

How about if `unlock` just remounts rw and sets up rofiles, but `unlock --permanent` does the same but a dup before that (or reverse the flags maybe, e.g. `unlock --transient`, since the hotfix workflow is probably gonna be more common for end users)?

But I prefer this line of thinking yeah. That way you don't give a chance for users to shoot themselves in the foot if they forget to `dup` first.

> Then if you e.g. pkg-add in a hotfix or dev deployment, it automatically becomes live?

For that, I was thinking of a `rpm-ostree pkg-add --transient/--live` which just does a checkout overlay.

cgwalters:
Okay, I'm agreeing on a single `unlock`, but I think I'd vote for `unlock --hotfix` over `unlock --permanent`.  I'm envisioning a documentation section on "applying hotfixes", and I can't think of any other use case for `--permanent`.  Can you?

This is a minor quibble though and we can easily change later.  I'll revisit this next week and split out the API changes for sysroot for `NOT_DEFAULT` as a separate commit from a bigger `unlock` verb.

cgwalters:
Moving over to https://github.com/GNOME/ostree/pull/212

PR #211, opened 2016-03-18
Libglnx porting round 2

cgwalters:


src/ostree/ot-admin-builtin-switch.c
```diff
@@ -162,16 +162,11 @@ ot_admin_builtin_switch (int argc, char **argv, GCancellable *cancellable, GErro
...
-    if (opt_reboot && g_file_equal (ostree_sysroot_get_path (sysroot), real_sysroot))
-      {
-        gs_subprocess_simple_run_sync (NULL, GS_SUBPROCESS_STREAM_DISPOSITION_INHERIT,
-                                       cancellable, error,
-                                       "systemctl", "reboot", NULL);
-      }
-  }
+  if (opt_reboot)
+    {
+      if (!ot_admin_execve_reboot (sysroot, error))
```
cgwalters: I should mention here I switched to `execve`...since why not?

jlebon: There's a change in behaviour here, no? It wouldn't reboot before if you weren't in the real sysroot? Though I think this makes more sense (since the user specified `--reboot` explicitly), was there a reason this was the behaviour before?

jlebon: Ahh, there isn't a reboot flag actually. Looks like it was meant to be added.

jlebon:
This is LGTM, but I'll let someone else more familiar with libglnx & libgsystem have another check. Compiles and passes fine.

src/ostree/ot-admin-builtin-switch.c
cgwalters: The "don't reboot if not in sysroot" logic now lives in `ot_admin_execve_reboot()`.

cgwalters:
I think I'm going to self-r+ this since I think it's good and don't want to distract giuseppe/mbarnes too much at the moment.

cgwalters:
Merged.

PR #213, opened 2016-03-21
deploy: Handle a read-only /boot

cgwalters:
I'd like to encourage people to make OSTree-managed systems more
strictly read-only in multiple places.  Ideally everywhere is
read-only normally besides `/var/`, `/tmp/`, and `/run`.

`/boot` is a good example of something to make readonly.  Particularly
now that there's work on the `admin unlock` verb, we need to protect
the system better against things like `rpm -Uvh kernel.rpm` because
the RPM-packaged kernel won't understand how to do OSTree right.

In order to make this work of course, we *do* need to remount `/boot`
as writable when we're doing an upgrade that changes the kernel
configuration.  So the strategy is to detect whether it's read-only,
and if so, temporarily mount read-write, then remount read-only when
the upgrade is done.

We can generalize this in the future to also do `/etc` (and possibly
`/sysroot/ostree/` although that gets tricky).

One detail: In order to detect "is this path a mountpoint" is
nontrivial - I looked at copying the systemd code, but the right place
is to use `libmount` anyways.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -159,6 +159,11 @@ libostree_1_la_CFLAGS += $(OT_INTERNAL_SOUP_CFLAGS)
 libostree_1_la_LIBADD += $(OT_INTERNAL_SOUP_LIBS)
 endif
 
+if USE_LIBMOUNT
+libostree_1_la_CFLAGS += $(OT_DEP_LIBMOUNT_CFLAGS)
+libostree_1_la_LIBADD += $(OT_DEP_LIBMOUNT_LIBS)
+endif
+
 if USE_SELINUX
 libostree_1_la_CFLAGS += $(OT_DEP_SELINUX_CFLAGS)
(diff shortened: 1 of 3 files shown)


jlebon:
Straight-forward enough, LGTM!

Should we document this somewhere? Maybe somewhere on the [Atomic Upgrades](http://ostree.readthedocs.org/en/latest/manual/atomic-upgrades/) page.

jlebon:
BTW, unrelated, but it seems like the page title at http://ostree.readthedocs.org/ is "My Docs". Maybe it should be changed to "OSTree" ? (Or "OSTree Docs") ?

giuseppe:
:+1:

cgwalters:
Thanks, I added a bit to the docs in `atomic-upgrades.md`.  Merged.

cgwalters:
Also https://github.com/GNOME/ostree/commit/63e1c86dcd88e852702698f0f693de1962294198

PR #214, opened 2016-03-21
Upstreaming debian patches

cgwalters:
Alex pointed me at https://anonscm.debian.org/git/collab-maint/ostree.git/tree/debian/patches - they're all fine, and should be upstream.


giuseppe:
:+1:

PR #216, opened 2016-03-22
Introducing ostree-grub-generator script

gatispaeglis:
Why this alternative approach:

1) The current approach is less flexible than using a
   custom 'ostree-grub-generator' script. Each system can
   adjust this script for its needs, instead of using the
   hardcoded values from ostree-bootloader-grub2.c.

2) Too much overhead on embedded to generate grub.cfg
   via /etc/grub.d/ configuration files. It is still
   possible to do so, even with this patch applied.
   No need to install grub2 package on a target device.

3) The grub2-mkconfig code path has other issues:
   https://bugzilla.gnome.org/show_bug.cgi?id=761180

Task: https://bugzilla.gnome.org/show_bug.cgi?id=762220


gatispaeglis:
Using "libexec_" seems to work fine on Debian systems as well. By default "make install" simply creates the /usr/local/libexec/ even if Debian does not have this path.

Tested with default libexecdir:
./autogen.sh --with-grub2-mkconfig=no

And with explicitly set libexecdir:
./autogen.sh --with-grub2-mkconfig=no --libexecdir=/usr/local/lib/ostree

Both cases work fine.

cgwalters:
Awesome, looks good at a high level.  I'll get a chance to review/test this in more detail soon.  One thing I would say at a high level is that *some* OSTree users are still going to want on-target-system grub2-mkconfig...specifically anyone doing parallel installation with Windows or OS X.  (If anyone wonders "why are bootloaders hard" it's that problem crossed by the 5,000,000 ways to configure storage).

But I absolutely want to support this, so will take a look and even see if we can switch over Fedora Atomic Host.

gatispaeglis:
>  One thing I would say though is that some OSTree users are still going to want on-target-system grub2-mkconfig

That is also already possible with the ostree-grub-generator approach, where this script file would contain the following steps:

1) Read boot loader entries and generate new 15_ostree in /boot/loader
2) Run grub2-mkconfig from ostree-grub-generator

Where sysroot is has the following symlink setup:
/etc/grub.d/15_ostree -> /boot/loader/15_ostree

I have not tested this, but sounds like it should work.

cgwalters:
Mind rebasing?  Should be pretty easy.  Sorry I should have realized https://github.com/GNOME/ostree/pull/211 would conflict.

gatispaeglis:
I will do that on Tuesday when I will be back in the office. I realized that there is one issue in the current patch.. using the OSTREE_GRUB2_EXEC envvar for testing purposes would not work as expected for grub2-mkconfig code path if use_grub2_mkconfig=false as it would skip the:
if (use_grub2_mkconfig && ostree_sysroot_get_booted_deployment (self->sysroot) == NULL
       && g_file_has_parent (self->sysroot->path, NULL))

Need to rethink that part.

PR #217, opened 2016-03-23
admin-switch: Add missing reboot argument

cgwalters:
Spotted by jlebon in https://github.com/GNOME/ostree/pull/211


jlebon:
LGTM!

PR #218, opened 2016-03-23
tests: Add a test-abi

cgwalters:
This would have caught a potential ABI regression in
https://bugzilla.gnome.org/show_bug.cgi?id=764056
(If we ran this test while building `--without-libarchive`)


cgwalters:
@cgwalters-bot try

cgwalters:
@cgwalters-bot retry

jlebon:
@cgwalters-bot r+ e41594d

PR #219, opened 2016-03-23
[merged] OstreeSePolicy: add ostree_sepolicy_get_csum()

jlebon:
This can be used as a fingerprint to determine whether two
OstreeSePolicy objects are equivalent.

Also add documentation for ostree_sepolicy_get_name().

### DIFF
--- a/src/libostree/libostree.sym
+++ b/src/libostree/libostree.sym
@@ -331,4 +331,5 @@ global:
 LIBOSTREE_2016.5 {
 global:
         ostree_repo_import_object_from_with_trust;
+        ostree_sepolicy_get_csum;
 } LIBOSTREE_2016.4;
--- a/src/libostree/ostree-sepolicy.c
+++ b/src/libostree/ostree-sepolicy.c
@@ -163,80 +163,85 @@ ostree_sepolicy_class_init (OstreeSePolicyClass *klass)
 
 /* Find the latest policy file in our root and return its checksum. */
 static gboolean
-get_policy_checksum (char **out_csum,
+get_policy_checksum (char        **out_csum,
                      GCancellable *cancellable,
-                     GError **error)
+                     GError      **error)
 {
   gboolean ret = FALSE;
+
   const char *binary_policy_path = selinux_binary_policy_path ();
   g_autofree char *bindir_path = g_path_get_dirname (binary_policy_path);
-  g_autoptr(GFile) bindir = g_file_new_for_path (bindir_path);
-  g_autoptr(GFile) best_child = NULL;
+  g_autofree char *binfile_prefix = g_path_get_basename (binary_policy_path);
+
+  glnx_fd_close int bindir_dfd = -1;
+
+  g_autofree char *best_policy = NULL;
   int best_version = 0;
 
-  /* The binary_policy_path includes the prefix of the basename, but without the
-   * period, e.g. "policy". */
-  g_autofree char *binfile_prefix = g_path_get_basename (binary_policy_path);
-  g_autofree char *binfile_no_fnmatch = g_strdup_printf ("%s.*[!0-9]*",
-                                                         binfile_prefix);
+  static gsize regex_initialized;
+  static GRegex *regex;
 
-  g_autoptr(GFileEnumerator) direnum =
-    g_file_enumerate_children (bindir, OSTREE_GIO_FAST_QUERYINFO,
-                               G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
-                               cancellable, error);
+  g_auto(GLnxDirFdIterator) dfd_iter = { 0,};
 
-  if (!direnum)
+  if (g_once_init_enter (&regex_initialized))
+    {
+      g_autofree char *sregex = g_strdup_printf ("^\\Q%s\\E\\.[0-9]+$",
+                                                 binfile_prefix);
+      regex = g_regex_new (sregex, 0, 0, NULL);
+      g_assert (regex);
+      g_once_init_leave (&regex_initialized, 1);
+    }
+
+  if (!glnx_opendirat (AT_FDCWD, bindir_path, TRUE, &bindir_dfd, error))
+    goto out;
+
+  if (!glnx_dirfd_iterator_init_at (bindir_dfd, ".", FALSE, &dfd_iter, error))
     goto out;
 
   while (TRUE)
     {
-      GFile *child;
-      GFileInfo *fi;
-      GFileType ftype;
+      struct dirent *dent = NULL;
 
-      if (!gs_file_enumerator_iterate (direnum, &fi, &child, cancellable, error))
+      if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent,
+                                                       cancellable, error))
         goto out;
-      if (!fi)
+
+      if (dent == NULL)
         break;
 
-      ftype = g_file_info_get_file_type (fi);
-      if (ftype == G_FILE_TYPE_REGULAR)
+      if (dent->d_type == DT_REG)
         {
-          g_autofree char *name = g_file_get_basename (child);
-
-          /* poor man's regex */
-          if (g_str_has_prefix (name, binfile_prefix) &&
-              fnmatch (binfile_no_fnmatch, name, FNM_NOESCAPE))
+          /* we could use match groups to extract the version, but mehhh, we
+           * already have the prefix on hand */
+          if (g_regex_match (regex, dent->d_name, 0, NULL))
             {
               int version = /* do +1 for the period */
-                (int)g_ascii_strtoll (name + strlen (binfile_prefix) + 1,
+                (int)g_ascii_strtoll (dent->d_name + strlen (binfile_prefix)+1,
                                       NULL, 10);
               g_assert (version > 0);
 
               if (version > best_version)
                 {
                   best_version = version;
-
-                  /* NB: The docs for gs_file_enumerator_iterate() says we can't
-                   * unref child, but says nothing re. adding a ref (though it
-                   * should be fine judging by the source). That said, better to
-                   * just dup it. */
-                  g_object_unref (best_child);
-                  best_child = g_file_dup (child);
+                  g_free (best_policy);
+                  best_policy = g_strdup (dent->d_name);
                 }
             }
         }
     }
 
-  if (!best_child)
+  if (!best_policy)
     {
       g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
                    "Could not find binary policy file");
       goto out;
     }
 
-  *out_csum = ot_checksum_file (best_child, G_CHECKSUM_SHA256,
-                                cancellable, error);
+  {
+    g_autofree char *path = glnx_fdrel_abspath (bindir_dfd, best_policy);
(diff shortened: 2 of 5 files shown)


src/libostree/libostree.sym
```diff
@@ -321,4 +321,5 @@ global:
         ostree_sysroot_deployment_unlock;
         ostree_deployment_get_unlocked;
         ostree_deployment_unlocked_state_to_string;
+        ostree_sepolicy_get_csum;
```
cgwalters: This needs to be in the new 2016.5 section.

src/libostree/ostree-sepolicy.c
```diff
@@ -257,6 +260,18 @@ initable_init (GInitable     *initable,
         freecon (con);
       }
 
+      {
+        g_autofree guchar *csum_raw = NULL;
+        if (!ostree_checksum_file (etc_selinux_dir, OSTREE_OBJECT_TYPE_FILE,
```
cgwalters: Mmm...I would say use `ot_checksum_file` instead of this.  We don't actually care about uid/gid/perms or xattrs on the file I'd say...or more realistically the kernel and userspace selinux policy tools I'm almost sure don't.

As a bonus that function already returns an ASCII checksum.

src/libostree/libostree.sym
jlebon: Hmm, did you forget to merge the bump commit into master?

cgwalters: Weird...somehow, yes.  Anyways done now.

src/libostree/ostree-sepolicy.c
jlebon: Ahhh nice, will do.

cgwalters: Also I'm a bit confused...we're checksumming `etc_selinux_dir`, the directory??

Also regarding policy and checksums, see http://marc.info/?l=selinux&m=145677940526857&w=2

jlebon: Actually, `ot_checksum_file` only supports files, whereas `ostree_checksum_file` supports both.

jlebon: Ahhh, I thought it was smart enough to descend in the directory and checksum the files. Hmm, OK, will do some more reading.

jlebon: Right, it just did a checksum of the directory metadata, which is of course not what we want. Should have read more carefully.

I initially just thought checksumming the whole /etc/selinux dir was a safer & easier way to handle this, but it seems like that may be overkill and the only file that actually matters is the final compiled policy file, e.g. `policy.29`.

Should we just use `dirname(selinux_binary_policy_path())` and then checksum all the files under there (though there should only be one for the matching kernel it seems) ? Not sure yet if the path it returns respects `selinux_set_policy_root()` or if we'd have to build the full dirname ourselves.

cgwalters: I think we need to enumerate files named `policy.[0-9]+` and pick the biggest.

cgwalters:
Regarding `selinux_binary_policy_path()`, indeed it should respect the root.  If you look at https://github.com/SELinuxProject/selinux/blob/master/libselinux/src/selinux_config.c#L222 it prefixes the internal paths with the root.

jlebon:
:arrow_up: Updated. It a bit more long-winded than expected :).

src/libostree/ostree-sepolicy.c
```diff
@@ -155,6 +159,92 @@ ostree_sepolicy_class_init (OstreeSePolicyClass *klass)
...
+  int best_version = 0;
+
+  /* The binary_policy_path includes the prefix of the basename, but without the
+   * period, e.g. "policy". */
+  g_autofree char *binfile_prefix = g_path_get_basename (binary_policy_path);
+  g_autofree char *binfile_no_fnmatch = g_strdup_printf ("%s.*[!0-9]*",
+                                                         binfile_prefix);
+
+  g_autoptr(GFileEnumerator) direnum =
+    g_file_enumerate_children (bindir, OSTREE_GIO_FAST_QUERYINFO,
```
cgwalters: FWIW I've been trying to avoid using `GFile` for newer code but I know there's a lot of it left around here in the selinux bits.  I'm okay with this but we should keep in mind it should be ported later.

src/libostree/ostree-sepolicy.c
```diff
@@ -155,6 +159,92 @@ ostree_sepolicy_class_init (OstreeSePolicyClass *klass)
...
+  if (!direnum)
+    goto out;
+
+  while (TRUE)
+    {
+      GFile *child;
+      GFileInfo *fi;
+      GFileType ftype;
+
+      if (!gs_file_enumerator_iterate (direnum, &fi, &child, cancellable, error))
```
cgwalters: This can actually just be `g_file_enumerator_iterate` now.

src/libostree/ostree-sepolicy.c
```diff
@@ -155,6 +159,92 @@ ostree_sepolicy_class_init (OstreeSePolicyClass *klass)
...
+        goto out;
+      if (!fi)
+        break;
+
+      ftype = g_file_info_get_file_type (fi);
+      if (ftype == G_FILE_TYPE_REGULAR)
+        {
+          g_autofree char *name = g_file_get_basename (child);
+
+          /* poor man's regex */
```
cgwalters: Well we do have `GRegex`...

src/libostree/ostree-sepolicy.c
```diff
@@ -155,6 +159,92 @@ ostree_sepolicy_class_init (OstreeSePolicyClass *klass)
...
+          if (g_str_has_prefix (name, binfile_prefix) &&
+              fnmatch (binfile_no_fnmatch, name, FNM_NOESCAPE))
```
cgwalters: It confused me initially that `fnmatch` returns `0` if it matched.  Can we do `fnmatch (...) == FNM_NOMATCH` ?

cgwalters:
I'm okay with leaving the `GFile` stuff in (up to you) but it seems like using a regex would be better - we already have example usage of that in e.g. `_ostree_sysroot_parse_deploy_path_name()`

src/libostree/ostree-sepolicy.c
jlebon: Ahhh, will use.

src/libostree/ostree-sepolicy.c
jlebon: Yes, it tricked as well. I'll just switch over to regex to make it easier.

jlebon:
:arrow_up: Updated to use the glnx fd-based iterator.

src/libostree/ostree-sepolicy.c
```diff
@@ -163,80 +163,85 @@ ostree_sepolicy_class_init (OstreeSePolicyClass *klass)
...
-    g_file_enumerate_children (bindir, OSTREE_GIO_FAST_QUERYINFO,
-                               G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
-                               cancellable, error);
+  g_auto(GLnxDirFdIterator) dfd_iter = { 0,};
 
-  if (!direnum)
+  if (g_once_init_enter (&regex_initialized))
+    {
+      g_autofree char *sregex = g_strdup_printf ("^\\Q%s\\E\\.[0-9]+$",
+                                                 binfile_prefix);
```
cgwalters: This is a bit weird because we're basing a once-per-process value (the regex) on something that's determined dynamically.  Well actually it's always going to be `policy` right?  Eh.  Maybe this chunk would be better in `initable_init()` ?  Dunno.

It's okay as is too.

src/libostree/ostree-sepolicy.c
```diff
@@ -163,80 +163,85 @@ ostree_sepolicy_class_init (OstreeSePolicyClass *klass)
...
                    "Could not find binary policy file");
       goto out;
     }
 
-  *out_csum = ot_checksum_file (best_child, G_CHECKSUM_SHA256,
-                                cancellable, error);
+  {
+    g_autofree char *path = glnx_fdrel_abspath (bindir_dfd, best_policy);
+    g_autoptr(GFile) file = g_file_new_for_path (path);
+    *out_csum = ot_checksum_file (file, G_CHECKSUM_SHA256, cancellable, error);
```
cgwalters: Here I'd say we should just change `ot_checksum_file` to take fd + path.  `git grep` says there are no other callers right now.  Instead of `g_file_read()`, we'd use `openat()` + `glnx_fd_readall_bytes()` ?

src/libostree/ostree-sepolicy.c
```diff
@@ -163,80 +163,85 @@ ostree_sepolicy_class_init (OstreeSePolicyClass *klass)
...
-                     GError **error)
+                     GError      **error)
 {
   gboolean ret = FALSE;
+
   const char *binary_policy_path = selinux_binary_policy_path ();
   g_autofree char *bindir_path = g_path_get_dirname (binary_policy_path);
-  g_autoptr(GFile) bindir = g_file_new_for_path (bindir_path);
-  g_autoptr(GFile) best_child = NULL;
+  g_autofree char *binfile_prefix = g_path_get_basename (binary_policy_path);
```
cgwalters: Also fwiw there's a non-malloc-ing variant of this in `glnx_basename()`.

src/libostree/ostree-sepolicy.c
jlebon: No I agree. I was coding everything else as if the prefix could be different from `policy`, so might as well go all the way here and make it per-instantiation. Since this is called from the initable_init, we could just leave it here and just not make it static, no? (Or maybe just use `g_regex_match_simple()` ?)

src/libostree/ostree-sepolicy.c
jlebon: Agreed.

src/libostree/ostree-sepolicy.c
cgwalters: Yeah, `g_regex_match_simple()` is totally fine here IMO.

src/libostree/ostree-sepolicy.c
cgwalters: Thanks; I know this patch has gone through a review wringer, but I believe it's worth it for future changes elsewhere so we can continue the fdrel transition.

jlebon:
:arrow_up: Rebased (though I didn't squash the fixups for easier review -- the first two commits are essentially the same). Changed to use `GRegex` and adapted `ot_checksum_file`.

src/libostree/ostree-sepolicy.c
```diff
@@ -211,9 +197,18 @@ get_policy_checksum (char        **out_csum,
 
       if (dent->d_type == DT_REG)
         {
+          /* We could probably save a few hundred nanoseconds if we accept that
+           * the prefix will always be "policy" and hardcode that in a static
+           * compile-once GRegex... But picture how exciting it'd be if it *did*
+           * somehow change; there would be cheers & slow-mo high-fives at the
+           * sight of our code not breaking. Is that hope not worth a fraction
+           * of a millisecond? I believe it is... or maybe I'm just lazy. */
```
cgwalters: :laughing:

cgwalters:
@cgwalters-bot r+ 9f8b97f

PR #223, opened 2016-03-25
[merged] commit: Support generating commits with no parent, or a custom one

cgwalters:
When I'm doing local development builds, it's quite common for me not
to want to accumulate history.  There are also use cases for this on
build servers as well.

In particular, using this, one could write a build system that didn't
necessarily need to have access to (a copy of) the OSTree repository.
Instead, the build system would determine the last commit ID on the
branch, and pass that to a worker node, then sync the generated
content back.

The API supported generating custom commits that don't necessarily
reference the previous commit on the same branch, let's just expose
this in the command line for convenience.

I plan to also support this rpm-ostree.

### DIFF
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -19,7 +19,7 @@
 
 set -euo pipefail
 
-echo "1..50"
+echo "1..52"
 
 $OSTREE checkout test2 checkout-test2
 echo "ok checkout"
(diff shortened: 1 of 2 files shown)


tests/basic-test.sh
```diff
@@ -97,6 +97,23 @@ assert_file_has_content four '4'
...
+assert_streq $($OSTREE log test2-no-parent |grep '^commit' | wc -l) "1"
+$OSTREE commit -b test2-no-parent -s '' --parent=none $test_tmpdir/checkout-test2-4
+assert_streq $($OSTREE log test2-no-parent |grep '^commit' | wc -l) "1"
+echo "ok commit no parent"
+
+cd ${test_tmpdir}
+$OSTREE commit -b test2-custom-parent -s '' $test_tmpdir/checkout-test2-4
+$OSTREE commit -b test2-custom-parent -s '' $test_tmpdir/checkout-test2-4
+$OSTREE commit -b test2-custom-parent -s '' $test_tmpdir/checkout-test2-4
+assert_streq $($OSTREE log test2-custom-parent |grep '^commit' | wc -l) "3"
```
jlebon: This could probably be a test on its own, e.g. "ok commit default parent".

jlebon:
Code looks good. Though in that same line of thinking, I do wonder if we should also make specifying the branch optional. You might just want to store the commit directly and not tag a ref to it (and maybe combine that with a new e.g. `-q` option which just spits out the new csum only). That might make testing scripts simpler as well. WDYT?

I'm OK to just merge this in and do the above in a separate PR.

jlebon:
Also, a e.g. `--empty/--allow-empty` type of flag (a-la-git) would be nice so we can be more explicit than `-s ''`.

tests/basic-test.sh
cgwalters: The `"ok log output"` test has been covering this, right?

cgwalters:
`-s` is like git's `-m`, it's conceptually orthogonal to `--allow-empty` which is about the *content*.  (Though I see there's also `--allow-empty-message`).  That said we should do both, but I think I'd like to do things like this as separate PRs, sound ok?

tests/basic-test.sh
jlebon: Ahhh indeed.

jlebon:
Right, I meant `--allow-empty-message`. Yup, separate PR sounds fine, just thinking out loud :).
@cgwalters-bot r+ dcd4927

PR #226, opened 2016-03-28
[merged] packaging: fix bashism in dist-snapshot target

smcv:
On Debian and its derivatives, /bin/sh is a lightweight POSIX shell
(currently dash) which does not support the bash {foo,bar} syntax.


cgwalters:
I have a general preference for hardcoding bash rather than fixing to `/bin/sh` for things that aren't run "on target", but this is fine.  Thanks for the patch!

@cgwalters-bot r+ 13571b1

PR #227, opened 2016-03-29
[merged] commit: Support writing orphans

cgwalters:
The API supports this, and it's not hard for us to do in the command
line as well.  One possible use case is separating "content
generation" in a separate server.

Related: https://github.com/ostreedev/ostree/pull/223


jlebon:
Nice! Code LGTM, tests pass.

Ohh right, I didn't even realize that the default output of `ostree commit` was just the csum as I mentioned in the other PR.

@cgwalters-bot r+ 3f255

jlebon:
@cgwalters-bot r+ af3f255

PR #228, opened 2016-03-29
[merged] Introducing ostree-grub-generator

gatispaeglis:
ostree-grub-generator can be used to generate
an easy customizable grub.cfg file. Compile time
decision ostree-grub-generator vs grub2-mkconfig
can be overwritten with the OSTREE_GRUB2_EXEC
envvar - useful for auto tests and OS installers.

Why this alternative approach:

1) The current approach is less flexible than using a
   custom 'ostree-grub-generator' script. Each system can
   adjust this script for its needs, instead of using the
   hardcoded values from ostree-bootloader-grub2.c.

2) Too much overhead on embedded to generate grub.cfg
   via /etc/grub.d/ configuration files. It is still
   possible to do so, even with this patch applied.
   No need to install grub2 package on a target device.

3) The grub2-mkconfig code path has other issues:
   https://bugzilla.gnome.org/show_bug.cgi?id=761180

Task: https://bugzilla.gnome.org/show_bug.cgi?id=762220


cgwalters:
One thing that's worth noting here as well is that Peter Jones wrote this: http://pkgs.fedoraproject.org/cgit/rpms/grub2.git/tree/0029-blscfg-add-blscfg-module-to-parse-Boot-Loader-Specif.patch?id=dcaf8f68832f12983a03aa57275627c7c7167021

Which fits in a bit better with the original intention of the BLS...I forget why we're not using it in Fedora Atomic right now.  It might be as prosaic as it not being enabled by default?  That'd likely be easy to fix.  Possibly a larger issue is the uncertain state of the BLS (https://www.redhat.com/archives/anaconda-devel-list/2014-July/msg00002.html etc).

Anyways, this approach is also quite simple.  GRUB2 is critical path for me so I'll need to take some time to test this.  Doing so now.

src/boot/grub2/ostree-grub-generator
```diff
@@ -0,0 +1,104 @@
...
+    if [ -z "${title}" ]; then
+        title="(Untitled)"
+    fi
+}
+
+populate_menu()
+{
+    # Prepend "boot/" prefix when boot/ directory is on the same partition as usr/,
+    # if boot/ directory is on a separate partition, set boot_prefix=""
+    boot_prefix="/boot"
```
cgwalters: See https://github.com/ostreedev/ostree/pull/215 - I wonder if we could have OSTree pass this down via environment variable?  Though that only works on-target.  Perhaps anyone doing a cross system like OpenEmbedded could have that set the `OSTREE_BOOT_PARTITION` environment variable there?

gatispaeglis:
Having to set 1 extra envvar when building image not-on-target device sounds reasonable and maintainable to me.

jlebon:
bot, test pull request

gatispaeglis:
> One thing that's worth noting here as well is that Peter Jones wrote [1] 

I was not aware of this plugin. I can see how it could be handy on Desktop and if other operating systems / boot loaders actually followed the Boot Loader Specification. And as you already mentioned, state of BLS seems to be unclear.

But if we compare ostree-grub-generator vs GRUB BLS plugin [1] the advantage for ostree-grub-generator is:
- I guess GSubprocess is not limited to executing shell scripts only. Which means that ostree-grub-generator can be a compiled executable implemented in C, C++ (thus, can borrow the logic from [1] without the need for chroot-ing, grub dependencies and other complexities mentioned earlier) or a Python script for example. Whatever works better for a target environment.

[1] http://pkgs.fedoraproject.org/cgit/rpms/grub2.git/tree/0029-blscfg-add-blscfg-module-to-parse-Boot-Loader-Specif.patch?id=dcaf8f68832f12983a03aa57275627c7c7167021

src/boot/grub2/ostree-grub-generator
```diff
@@ -0,0 +1,101 @@
...
+    config_file=${entries_path}/${1}
+    title=""
+    initrd=""
+    options=""
+    linux=""
+
+    while read -r line
+    do
+        record=$(echo ${line} | cut -f 1 -d ' ')
+        value=$(echo ${line} | cut -s -f2- -d ' ')
```
cgwalters: This is a clear enough script, but it's tempting to move the logic into C, basically what we're doing in the syslinux/u-boot ones.  We don't have to do that now though, and shell script is convenient for customization.

src/libostree/ostree-bootloader-grub2.c
```diff
@@ -299,8 +299,25 @@ _ostree_bootloader_grub2_write_config (OstreeBootloader      *bootloader,
...
+
+#ifdef OSTREE_GRUB2_MKCONFIG
+  use_grub2_mkconfig = TRUE;
+#endif
+  /* Autotests can set this envvar to select which code path to test, useful for OS installers as well */
+  grub_exec = g_getenv ("OSTREE_GRUB2_EXEC");
+  if (grub_exec)
+    {
+      if (g_str_has_suffix(grub_exec, "-mkconfig"))
+        use_grub2_mkconfig = TRUE;
```
cgwalters: Curious about this conditional - wouldn't someone who wants `grub2-mkconfig` be fine just setting `OSTREE_GRUB2_EXEC=grub2-mkconfig` explicitly?  Do we need to magically interpret `OSTREE_GRUB2_EXEC=banana-mkconfig` as `grub2-mkconfig`?

cgwalters:
> You are thinking to introduce OSTREE_BOOT_PARTITION in #215?

Yeah.

src/libostree/ostree-bootloader-grub2.c
gatispaeglis: The reason I decided to check only for the "-mkconfig" suffix is because on Ubuntu there is no "grub2-mkconfig". On Ubuntu it is simply called grub-mkconfig, so "-mkconfig" would work for both distros then. I was thinking about adding a comment in the source code about this.

cgwalters: Or use `AC_PATH_PROG` at build time?

src/boot/grub2/ostree-grub-generator
```diff
@@ -0,0 +1,101 @@
...
+# be lost the next time file is regenerated. For more details refer to the ${script} script.
+EOF
+}
+
+populate_header()
+{
+cat >> ${new_grub2_cfg} <<EOF
+serial --unit=0 --speed=115200 --word=8 --parity=no --stop=1
+default=boot
+timeout=10
```
cgwalters: This is the sort of thing where we should probably have a config file at least for any general purpose use.  It's basically what `/etc/grub.d/00_header` is right?

I guess this is where things are a bit nicer with the state we have with syslinux where the ostree-syslinux code edits just the entries in there, otherwise admins can edit it by hand.  But that's only really possible because the syslinux format is easy to parse, which is definitely not true of grub2 :frowning:

cgwalters:
At a high level I'm inclined to take this patch with just some small cleanups - we don't need to rearchitect things to have a config file etc. initially...anyone who wants a general purpose bootloader can just use `grub2-mkconfig`, and we can keep this code path targeted for embedded.

gatispaeglis:
Cool, I will clean up the remaining parts tomorrow then :)

tests/libtest.sh
```diff
@@ -282,10 +296,13 @@ EOF
 
     case $bootmode in
         "syslinux")
-	    setup_os_boot_syslinux
+        setup_os_boot_syslinux
             ;;
```
gatispaeglis: Does OSTree repository has policy regarding spaces vs tabs? When I see tabs around lines that I am touching I feel like I have to convert everything to spaces :)

src/boot/grub2/ostree-grub-generator
gatispaeglis: >  It's basically what /etc/grub.d/00_header is right
Yeah, looks like it. This definitely can be improved further by Atomic project or whoever else decides that they need it first :) 

> I guess this is where things are a bit nicer with the state we have with syslinux where the ostree-syslinux code edits just the entries in there, otherwise admins can edit it by hand. 

I want to achieve similar funcionality in the u-boot backend by this patch:
https://github.com/ostreedev/ostree/pull/156/commits/5ee49772b001b9757d6cb21fcc587d5ddc66cdb7
Where admin can edit /uEnv.txt, deploy the edited version. Where during the deploy process OSTree prepends its env (loader/uEnv.txt) on top of the /uEnv.txt.
I am planing to revisit that change soon and add an auto tests that demonstrates the use case.

tests/libtest.sh
cgwalters: C code is GNU style with no tabs, but we don't have a policy for shell yet.  I think it's mostly just been whatever Emacs does by default which is 4 spaces?

I'd say no tabs, and try match what exists, and maybe we'll revisit consistent indentation for shell later?

gatispaeglis:
Ok, if I am not missing anything, then this patch can be considered complete. Unless you find something more than needs to be fixed up.

cgwalters:
bot, retest this please

cgwalters:
@cgwalters-bot try

cgwalters:
I think the PR tester has been failing since you based on a commit before we got it all working.  Can you rebase again?  But let's see what happens with a "try" merge.

jlebon:
Actually, the logs say `INFO: using MERGE of PR`, which means that it was testing it with the latest changes.

Searching for "FAIL:" shows that only one test failed:

```
Running test: ostree/test-admin-deploy-grub2.sh.test
Copying gpghome to /tmp/test-tmp-ostree_test-admin-deploy-grub2.sh.test-U2W4EY
1db2ee44eb4bd7eb63ec641c657f8378d593394d129e931f53267524c3ae3ee9
42ea980590b99edb1e98b416480a6cb4272709296096c4773709a4bf5485f4a0
3fb815e88cd96f2e6a85211f31d2995c2057d96ff41876626dfd1ec148d8a69c
1/32 objects
4/32 objects
7/32 objects
10/32 objects
13/32 objects
16/32 objects
19/32 objects
22/32 objects
25/32 objects
28/32 objects
31/32 objects
ostree/deploy/testos initialized as OSTree root
cp: cannot stat '/usr/libexec/installed-tests/ostree/../src/boot/grub2/ostree-grub-generator': No such file or directory
FAIL: ostree/test-admin-deploy-grub2.sh.test (Child process exited with code 1)
```

gatispaeglis:
Right, I tested with "make check", then it succeeds. Probably need to install ostree-grub-generator in usr/libexec/installed-tests/ostree/ for it to pass.

cgwalters:
The tests are passing here on my laptop (both `make check` and installed).  Hm, I wonder though if this has to do with the fact the PR tester is running the installed tests as root?  I'll look at that.

In the meantime reviewing this patch more, I found it made a lot more sense to me if the configure option was `--with-builtin-grub-mkconfig`, rather than "use the system" be a default yes.  And the naming in the code IMO is clearer too.

If you agree can you cherry-pick https://github.com/gatispaeglis/ostree/pull/1 into your PR branch?  (Yep, I made a PR for your PR, it works :smile:  )

In the big picture I am a bit sad that it's unlikely we can use this for Project Atomic yet, as I would really like to just ditch the grub2 binaries.  But it'd need a config file and other bits.

I'm also a bit uncertain about the long term maintenance of carrying this special case, and am wondering if we should redirect people to extlinux.  Was there a reason you chose grub2 over extlinux?

gatispaeglis:
I fixed the grub2 test after the last comment from jlebon so everything should be fine now.

> If you agree can you cherry-pick gatispaeglis#1 into your PR branch?
Patch seems fine to me.

>  and am wondering if we should redirect people to extlinux. Was there a reason you chose grub2 over extlinux?

Several reasons actually. I have been experimenting with Intel NUC board where grub2 (with UEFI firmware) is what Intel provides by default in http://git.yoctoproject.org/cgit/cgit.cgi/meta-intel so I didn't even consider other boot loader.
Other reason is that GRUB has better support for UEFI. UEFI support appears to be limited on syslinux according to https://wiki.archlinux.org/index.php/syslinux#UEFI_Systems . Particularly the following limitation:
"For Syslinux, the kernel and initramfs files need to be in the ESP, as Syslinux does not (currently) have the ability to access files outside its own partition (i.e. outside ESP in this case)." So syslinux + UEFI != ostree

cgwalters:
Oh right, I think we talked about the UEFI issue.  OK, let's go with this then.  At some point though I do want to revisit https://github.com/ostreedev/ostree/pull/228#issuecomment-203560635 - I didn't quite understand you reply there.  The GRUB2 builtin BLS loader wouldn't have any on-target dependencies either, but it would need some changes to be possibly more flexible/configurable?

@cgwalters-bot r+ 92c088e

cgwalters:
@cgwalters-bot r+ 29c088e

src/boot/grub2/ostree-grub-generator
gatispaeglis: >  At some point though I do want to revisit #228 (comment) - I didn't quite understand you reply there. The GRUB2 builtin BLS loader wouldn't have any on-target dependencies either, but it would need some changes to be possibly more flexible/configurable?

What I meant here is that this logic could be implemented entirely in ostree. Instead of working on making a GRUB2 builtin BLS loader more configurable, there could be a c++ binary (configurable) in ostree project that does what normally would be the job of "GRUB2 builtin BLS loader".

Then we could reuse the same code path what is currently used by ostree-grub-generator. Where ostree instead of executing ostree-grub-generator shell script would execute the above mentioned c++ binary.

Hopefully this clarifies it better.

PR #229, opened 2016-03-30
[merged] commit: support editor for orphan commits

jlebon:
This is a follow up to #227 to allow ostree to open the editor also for
orphan commits when no subject or body is given on the cmdline.


cgwalters:
@cgwalters-bot r+ 5b34d46

PR #231, opened 2016-03-30
tests: make them pass

jlebon:
WARNING: hacky solutions below.

### DIFF
--- a/tests/test-admin-locking.sh
+++ b/tests/test-admin-locking.sh
@@ -25,10 +25,10 @@ set -euo pipefail
 setup_os_repository "archive-z2" "syslinux"
 
 # If parallel is not installed, skip the test
-if ! parallel --help >/dev/null 2>&1; then
-    echo "1..0 # SKIP no /usr/bin/parallel"
+if ! which parallel >/dev/null 2>&1; then
+    echo "1..0 # SKIP no parallel"
     exit 0
-fi    
+fi
 
 echo "1..1"
 
--- a/tests/test-demo-buildsystem.sh
+++ b/tests/test-demo-buildsystem.sh
@@ -19,8 +19,11 @@
 
 set -euo pipefail
 
-if ! fusermount --version >/dev/null 2>&1; then
+if ! which fusermount >/dev/null 2>&1; then
     echo "1..0 # SKIP no fusermount"
+    exit 0
+elif ! capsh --print | grep ^Current | grep -q cap_sys_admin; then
+    echo "1..0 # SKIP no CAP_SYS_ADMIN"
     exit 0
 fi
 
--- a/tests/test-pull-corruption.sh
+++ b/tests/test-pull-corruption.sh
@@ -20,7 +20,7 @@
 set -euo pipefail
 
 # If gjs is not installed, skip the test
-if ! gjs --help >/dev/null 2>&1; then
+if ! which gjs >/dev/null 2>&1; then
     echo "1..0 # SKIP no gjs"
     exit 0
 fi
@@ -55,7 +55,16 @@ do_corrupt_pull_test() {
 
 # FIXME - ignore errors here since gjs in RHEL7 has the final
 # unrooting bug
-gjs $(dirname $0)/corrupt-repo-ref.js ${repopath} main || true
+
+# super hacky fix to support both installed and uninstalled tests. we'll
+# probably want to just enable it for one of them.
+if [ -f /lib64/libostree-1.so ]; then
+  gjs $(dirname $0)/corrupt-repo-ref.js ${repopath} main || true
+else
+  GI_TYPELIB_PATH=${BUILDDIR} LD_PRELOAD=${BUILDDIR}/.libs/libostree-1.so \
+    gjs $(dirname $0)/corrupt-repo-ref.js ${repopath} main || true
+fi
+
 assert_file_has_content corrupted-status.txt 'Changed byte'
 do_corrupt_pull_test
 echo "ok corruption"
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -114,7 +114,7 @@ assert_streq $($OSTREE log test2-custom-parent |grep '^commit' | wc -l) "3"
 echo "ok commit custom parent"
 
 cd ${test_tmpdir}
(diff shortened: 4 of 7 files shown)


jlebon:
bot, retest this please

jlebon:
bot, retest this please

jlebon:
Well, that was trickier to debug than expected.

For some reason `test-pull-untrusted.sh` would pass when run as root, but fail when run as testuser. It turned out that the way the files were stored in the repo when running as a testuser were different, which meant that a different `.file` object was chosen for corruption. Except that file turned out to be a symlink, so the `echo "broke"` actually just wrote to the `no_such_file` symlink target, thus keeping the actual symlink file's checksum the same and causing the pull-local to pass when it should have failed.

All green now! \o/
cc @cgwalters

tests/test-pull-corruption.sh
```diff
@@ -55,7 +55,16 @@ do_corrupt_pull_test() {
 
 # FIXME - ignore errors here since gjs in RHEL7 has the final
 # unrooting bug
-gjs $(dirname $0)/corrupt-repo-ref.js ${repopath} main || true
+
+# super hacky fix to support both installed and uninstalled tests. we'll
+# probably want to just enable it for one of them.
+if [ -f /lib64/libostree-1.so ]; then
```
jlebon: Actually, a better condition might be to just check if $BUILDDIR exists. Otherwise, we'll be using the installed SO even if it's not the one we just built.

smcv: I think a better approach to this is to set `LD_LIBRARY_PATH` for the uninstalled tests. I'm about to submit a branch that fixes that, among other things.

tests/test-admin-locking.sh
```diff
@@ -25,10 +25,10 @@ set -euo pipefail
 setup_os_repository "archive-z2" "syslinux"
 
 # If parallel is not installed, skip the test
-if ! parallel --help >/dev/null 2>&1; then
-    echo "1..0 # SKIP no /usr/bin/parallel"
+if ! which parallel >/dev/null 2>&1; then
```
smcv: Why is this use of `which` better?

This is a regression, at least for Debian: there are two at-least-semi-common `parallel` utilities (GNU parallel and moreutils parallel), and we specifically want GNU parallel.

tests/test-pull-corruption.sh
smcv: Also please note that this "super hacky fix" is going to fail on 32-bit, on Debian and its derivatives (it's `/usr/lib/x86_64/libostree-1.so` there), under jhbuild and so on. See #232 for the use of `LD_LIBRARY_PATH` that fixes this differently.

tests/test-admin-locking.sh
smcv: (The reason this would be a regression is that moreutils parallel doesn't have `--help`, so more or less by coincidence, the test is correctly skipped if you have moreutils parallel rather than GNU parallel)

tests/test-pull-corruption.sh
cgwalters: Yeah, agreed @smcv's fix here is better.  Jonathan, what do you think about rebasing your patch series on top of his?

I think now that we have the PR tester enabled we're going to have to merge to one PR.  @smcv maybe do any `fixup!`s on your PR, then Jonathan pulls it into this one?

tests/test-admin-locking.sh
jlebon: The reason is that parallel --help exited with 255, causing the condition to fail. I can replace that with --version though.

tests/test-pull-corruption.sh
jlebon: Yup, no complaints here. I knew this was gonna have to change, hence the "super hacky fix" warning. :)
Thanks @smcv for the proper fix.

tests/test-admin-locking.sh
jlebon: OK, well it seems like the moreutils one doesn't even have --version, and exits 1 when given. So we can just replace that --help with --version. @smcv, should I make a commit for you to cherry-pick, or do you want to apply that directly on your PR?

smcv: Maybe `parallel --gnu /bin/true </dev/null 2>/dev/null` would be more correct; moreutils parallel could conceivably grow `--help` and `--version`, but hopefully nothing incompatible with GNU parallel is going to gain a `--gnu` option.

smcv: > apply that directly on your PR

done

tests/test-demo-buildsystem.sh
```diff
@@ -19,9 +19,12 @@
 
 set -euo pipefail
 
-if ! fusermount --version >/dev/null 2>&1; then
+if ! which fusermount >/dev/null 2>&1; then
     echo "1..0 # SKIP no fusermount"
     exit 0
+elif ! capsh --print | grep ^Current | grep -q cap_sys_admin; then
```
smcv: I'm not sure this is correct? On my Debian system, I don't have cap_sys_admin in Current, but fusermount is setuid root, so this test doesn't need to be skipped (and indeed it passes).

Perhaps something like (pseudocode)

we have CAP_SYS_ADMIN || (fusermount is setuid root && securebits do not prevent us from taking advantage of that)

cgwalters: Oh right, this will turn off the tests when run inside a desktop system.  It's not securebits I think so much as the bounding set.  Changing "Current" to "Bounding set" should work in both cases I think.

smcv: That makes more sense to me! Any chance someone who sees this failure could commit it, after landing #232? (I don't currently have any container environments that would see this, only VMs, chroots and the real system)

jlebon: Actually, from quickly testing it, it seems like your skip check for fuse should be enough. It seems like it's not enough to run the container with `--cap-add=SYS_ADMIN`, but it actually needs `--privileged` to have a writable `/etc/fuse`

jlebon: I mean `/dev/fuse`

jlebon:
Replaced by #232.

PR #232, opened 2016-03-31
[merged] Fix various build-time test failures

smcv:
This branch makes the build-time tests for 2016.4 work in an autobuilder environment (Debian sbuild), and does most of the necessary cleanup for the build chroot to be deleted.

One remaining issue is that one test leaks several gpg-agent processes matching `gpg-agent --homedir /var/tmp/tap-test.*/repo/tmp/ostree-gpg-*` (where the first `*` is the same for all the processes). For the moment I haven't identified which one, so I'm working around that in the packaging for now.

I'm still getting two failures in git master that I didn't see in 2016.4, which are not fixed here:

* `tests/test-pull-untrusted.sh`: fails with `corrupted untrusted pull unexpectedly failed!` (test appears to be new since 2016.4)
* `tests/test-basic.sh`: fails at `assert_not_reached 'Found orphaned commit'` (regression since 2016.4)

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -216,6 +216,10 @@ EXTRA_DIST += \
 	tests/gpg-verify-data/trustdb.gpg \
 	tests/gpg-verify-data/gpg.conf
 
+tests-libreaddir-rand-so-symlink:
+	ln -fns ../.libs/libreaddir-rand.so tests
+ALL_LOCAL_RULES += tests-libreaddir-rand-so-symlink
+
 # Unfortunately the glib test data APIs don't actually handle
 # non-recursive Automake, so we change our code to canonically look
 # for tests/ which is just a symlink when installed.
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -131,6 +131,9 @@ libreaddir_rand_la_SOURCES = tests/readdir-rand.c
 libreaddir_rand_la_CFLAGS = $(OT_INTERNAL_GIO_UNIX_CFLAGS)
 libreaddir_rand_la_LIBADD = $(OT_INTERNAL_GIO_UNIX_LIBS)
 libreaddir_rand_la_LDFLAGS = -avoid-version
+if !ENABLE_INSTALLED_TESTS
+libreaddir_rand_la_LDFLAGS += -rpath $(abs_builddir)
+endif
 
 test_programs = tests/test-varint tests/test-ot-unix-utils tests/test-bsdiff tests/test-mutable-tree \
 	tests/test-keyfile-utils tests/test-ot-opt-utils tests/test-ot-tool-util \
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -25,7 +25,10 @@ include $(top_srcdir)/buildutil/glib-tap.mk
 TESTS_ENVIRONMENT += OT_TESTS_DEBUG=1 \
 	SRCDIR=$$(cd $(top_srcdir) && pwd) \
 	BUILDDIR=$$(cd $(top_builddir) && pwd) \
-	PATH=$$(cd $(top_builddir) && pwd):$${PATH}
+	GI_TYPELIB_PATH=$$(cd $(top_builddir) && pwd) \
+	LD_LIBRARY_PATH=$$(cd $(top_builddir)/.libs && pwd) \
+	PATH=$$(cd $(top_builddir) && pwd):$${PATH} \
+	$(NULL)
 
 uninstalled_test_scripts = tests/test-abi.sh
 
--- a/buildutil/tap-test
+++ b/buildutil/tap-test
@@ -13,7 +13,7 @@ touch ${tempdir}/.testtmp
 function cleanup () {
     if test -n "${TEST_SKIP_CLEANUP:-}"; then
 	echo "Skipping cleanup of ${tempdir}"
-    else if test -f ${tempdir}/.test; then
+    else if test -f ${tempdir}/.testtmp; then
 	rm "${tempdir}" -rf
     fi
     fi
--- a/tests/test-basic-user.sh
+++ b/tests/test-basic-user.sh
@@ -21,11 +21,7 @@ set -euo pipefail
 
 . $(dirname $0)/libtest.sh
 
-touch test-xattrs
-if ! setfattr -n user.testvalue -v somevalue test-xattrs; then
-    echo "1..0 # SKIP bare-user repository requires xattr support"
-    exit 0
-fi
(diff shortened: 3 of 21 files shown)


smcv:
> `tests/test-pull-untrusted.sh`: fails with `corrupted untrusted pull unexpectedly failed!` (test appears to be new since 2016.4)
> `tests/test-basic.sh`: fails at `assert_not_reached 'Found orphaned commit'` (regression since 2016.4)

Looks like these are among the issues that @jlebon fixed in #231.

I think setting the special `LD_LIBRARY_PATH` and `GI_TYPELIB_PATH` via `TESTS_ENVIRONMENT` is a better solution than doing it in `tests/test-pull-corruption.sh` as @jlebon did, so I've kept that bit even though it will conflict.

smcv:
Branch updated: redid one patch to always source `libtest.sh` before checking for xattr support, because otherwise the creation of `test-xattrs` breaks the check for an empty directory when doing installed-tests.

cgwalters:
bot, add author to whitelist

Makefile-tests.am
```diff
@@ -25,7 +25,10 @@ include $(top_srcdir)/buildutil/glib-tap.mk
 TESTS_ENVIRONMENT += OT_TESTS_DEBUG=1 \
 	SRCDIR=$$(cd $(top_srcdir) && pwd) \
 	BUILDDIR=$$(cd $(top_builddir) && pwd) \
-	PATH=$$(cd $(top_builddir) && pwd):$${PATH}
+	GI_TYPELIB_PATH=$$(cd $(top_builddir) && pwd) \
```
cgwalters: FWIW this sort of stuff is exactly the reason why I think InstalledTests is a generally more robust system.  But it is hard to beat the convenience and historical momentum behind `make check`, so we'll keep doing both.

It might be nice if there were a cargo-cultable `.am` fragment we could share for this?  Anyways this is obviously fine as is.

tests/test-basic-user.sh
```diff
@@ -19,10 +19,16 @@
 
 set -euo pipefail
 
-echo "1..1"
-
 . $(dirname $0)/libtest.sh
 
+touch test-xattrs
+if ! setfattr -n user.testvalue -v somevalue test-xattrs; then
```
cgwalters: Maybe extract this into a `libtest.sh` function like `skip_without_user_xattrs` ?

tests/test-delta.sh
```diff
@@ -156,6 +156,13 @@ ${CMD_PREFIX} ostree --repo=repo2 ls ${newrev} >/dev/null
 
 echo 'ok pull delta'
 
+touch test-xattrs
+if ! setfattr -n user.testvalue -v somevalue test-xattrs; then
+    echo "ok apply offline # SKIP bare-user repository requires xattr support"
+    echo "ok apply offline inline # SKIP bare-user repository requires xattr support"
+    exit 0
```
cgwalters: Doesn't this result in a truncated test case run?  It seems like we need to skip the whole file here unfortunately.

buildutil/tap-test
```diff
@@ -13,7 +13,7 @@ touch ${tempdir}/.testtmp
 function cleanup () {
     if test -n "${TEST_SKIP_CLEANUP:-}"; then
 	echo "Skipping cleanup of ${tempdir}"
-    else if test -f ${tempdir}/.test; then
+    else if test -f ${tempdir}/.testtmp; then
```
cgwalters: Oops =/

cgwalters: (In my defense I most often ended up debugging the tests locally, and on the autobuilders this all gets cleaned up anyways)

tests/test-delta.sh
smcv: I made sure to say `ok` the same number of times as if I hadn't skipped these.

smcv: I can hoist this to the top and skip the entire file if you prefer, though.

buildutil/tap-test
smcv: This mainly affects `gnome-desktop-testing-runner`, which is how I run the installed-tests (but not the build-time tests) for Debian.

tests/test-delta.sh
cgwalters: Oh, I see.  Hm, yeah that's probably going to be a maintenance pain over time.  I think I'd vote for skipping the whole file, assume we get test coverage from systems that aren't tmpfs for a while.

smcv: (Same thing happens in `test-prune`)

smcv:
Branch updated

* Port the tests that are entirely skipped if no xattrs to use a new convenience function as suggested (I haven't done the ones that are partially skipped yet)
* New patch to skip rofiles-fuse-related tests if we don't have /dev/fuse, which isn't available in my Debian schroot/sbuild autobuilder (they still pass on my laptop when outside a chroot)
* Split out the bits of #231 that are necessary to pass tests on my laptop, crediting @jlebon for those commits

jlebon:
Ahhh, I can close my PR then.

Could you also pull in the cap_sys_admin hunks? (Maybe in that new skip_without_fuse function). This is required when running from inside an unprivileged container. That said, I do intend to leave that cap on for the PR tester, but I'd rather not require it for everyone else.

smcv:
> Hm, yeah that's probably going to be a maintenance pain over time. I think I'd vote for skipping the whole file, assume we get test coverage from systems that aren't tmpfs for a while.

Done. I run the installed-tests in a virtual machine that does have an ext4 /var/tmp (the filesystem image is really on tmpfs, but that's outside the VM) so I have coverage both ways.

smcv:
> Could you also pull in the cap_sys_admin hunks?

I'm reluctant to do that, because I'm not sure they're completely correct (see my comments on your commit) and I don't experience this particular failure myself.

jlebon:
Nice, thanks for cleaning up that libreaddir-rand issue. It was making the test results hard to read.

jlebon:
I'm good to merge this, though some commits could be coalesced (e.g. all the parallel related stuff together, libreaddir, xattrs, etc...). I'll leave it up to you, @smcv and @cgwalters.

smcv:
> some commits could be coalesced

My intention was to squash the `fixup!` commits and nothing else (if I'm reading `CONTRIBUTING.md` correctly, the bot will do that); if this project's style dictates fewer/larger commits, you're welcome to squash more of them.

cgwalters:
Yeah, maybe a little too split but I'm fine with it here...some of the individual fixes have interesting rationales.

@cgwalters-bot r+ 2b75c7f

PR #233, opened 2016-03-31
[merged] tests: Make failing to kill the GPG agent non-fatal

cgwalters:
It's not working for me in `make check` on a RHEL 7 Workstation,
apparently because no GPG agent is spawned.  I'm guessing this has
something to do with the GPG version?

The downside of this is we will be less likely to notice if GPG
changes again and we start leaking agents like we're in The Matrix
Reloaded.  But the real solution to that is containers anyways.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -349,3 +349,7 @@ skip_without_fuse () {
         exit 0
     fi
 }
+
+libtest_cleanup_gpg () {
+    gpg-connect-agent --homedir ${test_tmpdir}/gpghome killagent /bye || true
+}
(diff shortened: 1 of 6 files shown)


cgwalters:
@cgwalters-bot delegate=smcv

smcv:
Seems fine to me. I'm on Debian unstable with GnuPG 2.1.11-6, for what it's worth.

I believe it's 2.1.0 (November 2014) that switched to this mode: <https://www.gnupg.org/faq/whats-new-in-2.1.html#autostart>, <https://www.gnupg.org/faq/whats-new-in-2.1.html#nosecring>

smcv:
@cgwalters-bot r+ a878d3f

cgwalters:
Yeah, and the PR tester is Fedora 23, but I have `gnupg2-2.0.22-3.el7.x86_64` here.

smcv:
> we will be less likely to notice if GPG changes again and we start leaking agents

My builds in schroot/sbuild will fail to unmount/discard their temporary chroots, so I'll notice at the next release.

PR #234, opened 2016-03-31
[merged] More test fixes

smcv:
Having fixed the build-time tests in #232, now it's the installed-tests' turn :-)

With these changes (minus the one for `test-abi` because that didn't exist in the 2016.4 release), the installed-tests pass on a Debian VM again.


cgwalters:
Can you review https://github.com/ostreedev/ostree/pull/233 ?  See http://escher.verbum.org:54856/ for Homu usage instructions.

cgwalters:
All looks good to me...I am uncertain right now why the first and last commits haven't been necessary for me yet, but not enough to dive in and try to debug why it's *not* failing.

@cgwalters-bot r+ 9b0a777

PR #236, opened 2016-04-01
[merged] Inherit options and remotes from parent repo

alexlarsson:
This makes parent repos much more useful. In particular I'd like to use this to make a user repo inherited from a system repo, pull from a http remote and then do a local pull into the system remote (via a privileged helper).

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -217,6 +217,27 @@ ost_repo_get_remote (OstreeRepo  *self,
   return remote;
 }
 
+static OstreeRemote *
+ost_repo_get_remote_inherited (OstreeRepo  *self,
+                               const char  *name,
+                               GError     **error)
+{
+  local_cleanup_remote OstreeRemote *remote = NULL;
+  g_autoptr(GError) temp_error = NULL;
+
+  remote = ost_repo_get_remote (self, name, &temp_error);
+  if (remote == NULL)
+    {
+      if (self->parent_repo != NULL)
+        return ost_repo_get_remote_inherited (self->parent_repo, name, error);
+
+      g_propagate_error (error, g_steal_pointer (&temp_error));
+      return NULL;
+    }
+
+  return g_steal_pointer (&remote);
+}
+
 static void
 ost_repo_add_remote (OstreeRepo   *self,
                      OstreeRemote *remote)
@@ -272,6 +293,8 @@ ostree_repo_get_remote_option (OstreeRepo  *self,
 {
   local_cleanup_remote OstreeRemote *remote = NULL;
   gboolean ret = FALSE;
+  g_autoptr(GError) temp_error = NULL;
+  g_autofree char *value = NULL;
 
   if (_ostree_repo_remote_name_is_file (remote_name))
     {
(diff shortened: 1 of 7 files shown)


src/libostree/ostree-repo.c
```diff
@@ -870,13 +870,30 @@ ostree_repo_get_boolean_option (OstreeRepo *self,
...
-                                            key, default_value,
-                                            result, error))
-    return FALSE;
+  ret_bool = g_key_file_get_boolean (self->config, section, key, &temp_error);
+  if (temp_error != NULL)
+    {
+      if (g_error_matches (temp_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND))
+        {
+          if (self->parent_repo != NULL)
+            return ostree_repo_get_boolean_option (self->parent_repo,
```
cgwalters: I'm uncertain about unilaterally inheriting *all* options from the parent repo.  For example, we obviously don't want `core/mode`.  This one wouldn't come in in practice since it'll always be set in a repo.

But another example is `core/fsync`.

Maybe the best example is `core/repo_version` (although interestingly that call doesn't use `ot_keyfile_get_boolean_with_default()` so it'd still work with this patch).

To flip this all around, are there *any* options we want to inherit other than the remotes?

alexlarsson:
I couldn't find a summary of all options, but for the options i found in the code you're right, we really don't want to inherit them. In particular many are tied to the repo mode which may be completely different in the child repo. Lemme just rebase this by dropping the core config inheriting.

src/libostree/ostree-repo.c
```diff
@@ -257,13 +257,18 @@ _ostree_repo_remote_name_is_file (const char *remote_name)
   return g_str_has_prefix (remote_name, "file://");
 }
 
+/**
+ * ostree_repo_get_remote_option:
+ * @self:
+ *
```
cgwalters: I think we need to do more than stub this out, otherwise e.g. introspection won't understand the `char **out_value` and it won't be accessible to scripts.

Something like:

```
@self: Repo
@remote_name: Name
@option_name: Option
@default_value: (allow-none): Value returned if @option_name is not present
@out_value: (out): Return location for value
@error: Error

OSTree remotes are represented by keyfile groups, formatted like:
`[remote "remotename"]`. This function returns a value named @option_name
underneath that group.
```

?

src/libostree/ostree-repo.c
```diff
@@ -315,29 +362,38 @@ ostree_repo_get_remote_list_option (OstreeRepo   *self,
...
     }
+  else if (self->parent_repo != NULL)
+    return ostree_repo_get_remote_list_option (self->parent_repo,
+                                               remote_name, option_name,
+                                               out_value,
+                                               error);
+  else
+    g_propagate_error (error, g_steal_pointer (&temp_error));
 
+  *out_value = g_steal_pointer (&value);
```
cgwalters: Looks like we're doing the transfer even on error cases, which in this function is probably fine, but one of the reasons I structure the code as:

```
  *out_foo =  steal
  *out_bar = steal
   ret = TRUE;
out:
  return ret;
```

is to avoid stealing on errors.  Anyways it's okay as is probably but something to keep in mind elsewhere.

tests/test-parent.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+#
+# Copyright (C) 2011 Colin Walters <walters@verbum.org>
```
cgwalters: I'd say this is copyright you (this year)

tests/test-parent.sh
```diff
@@ -0,0 +1,51 @@
...
+
+cd ${test_tmpdir}
+
+# Create a repo
+mkdir repo2
+${CMD_PREFIX} ostree --repo=repo2 init
+${CMD_PREFIX} ostree --repo=repo2 remote add --gpg-import=${test_tmpdir}/gpghome/trusted/pubring.gpg --set=gpg-verify=true aremote file://$(pwd)/repo test2
+
+# Create a repo with repo2 as parent
+${CMD_PREFIX} ostree init --repo=repo3 --mode=bare-user
```
cgwalters: Aren't we missing a `mkdir repo3`?  (Really ostree init should probably have a `--create` or something)

tests/test-parent.sh
```diff
@@ -0,0 +1,51 @@
...
+if ${CMD_PREFIX} ostree --repo=repo3 pull aremote; then
+    assert_not_reached "GPG verification unexpectedly succeeded"
+fi
+echo "ok unsigned pull w/parent"
+
+# Make a signed commit
+${OSTREE} commit -b test2 -s "A GPG signed commit" -m "Signed commit body" --gpg-sign=${TEST_GPG_KEYID_1} --gpg-homedir=${TEST_GPG_KEYHOME} --tree=dir=files
+
+# Make sure we pick up the remote url and gpg config from the parent
+echo "ok signed pull w/parent"
```
cgwalters: We make a signed commit but don't do another pull?

tests/test-parent.sh
alexlarsson: Actually its the other mkdir that is superfluous, ostree init *does* create the directory.

tests/test-parent.sh
alexlarsson: Ugh, seems like i removed to much of the testcase.

tests/test-parent.sh
cgwalters: Wha...hum.  That must have been a recent change, I swear we didn't use to do that.

cgwalters:
There's no notification for git pushes, so don't forget to write a comment after you push fixups and want re-review.  Plus it's an opportunity to use emoji!  :arrow_up: :arrow_right: :8ball: 

@cgwalters-bot r+ d5a9c34

PR #237, opened 2016-04-04
[merged] Pull local gpg

alexlarsson:
This adds support for enforcing gpg checks during pull-local.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1900,6 +1900,8 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   GSource *update_timeout = NULL;
   gboolean disable_static_deltas = FALSE;
   gboolean require_static_deltas = FALSE;
+  gboolean opt_gpg_verify = FALSE;
+  gboolean opt_gpg_verify_summary = FALSE;
 
   if (options)
     {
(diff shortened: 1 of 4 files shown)


alexlarsson:
I need this for the privileged helper approach. See mailing list for details about that.

giuseppe:
looks fine to me.  Waiting for @cgwalters opinion for the cmd changes

src/libostree/ostree-repo-pull.c
```diff
@@ -2107,12 +2107,14 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
       }
 
-    if (bytes_sig && !_ostree_repo_load_cache_summary_if_same_sig (self,
-                                                                   remote_name_or_baseurl,
-                                                                   bytes_sig,
-                                                                   &bytes_summary,
-                                                                   cancellable,
-                                                                   error))
+    if (bytes_sig &&
+        !_ostree_repo_remote_name_is_file (remote_name_or_baseurl) &&
```
cgwalters: Hm, can we call this function once and cache the result as `pull_data->is_local`?  Although looking again we already have `pull_data->remote_repo_local`.  Which should maybe be hoisted a lot earlier.

cgwalters:
Eh, we can do that later.  This looks good to me as is too, thanks!

@cgwalters-bot r+ 62ac503

cgwalters:
@alexlarsson Can you review followup in https://github.com/ostreedev/ostree/pull/239 ?

jlebon:
Was looking at the pull code and noticed that the new `gpg-verify` and `gpg-verify-summary` options that this PR added are actually only used for the file:// case. Shouldn't they override the config values in the configured remote case?

PR #239, opened 2016-04-05
[merged] pull: More consistently use remote_repo_local for local repos

cgwalters:
Check whether the URI is `file:///` once, then initialize a
child OstreeRepo early on, and use whether it's NULL or not
to know whether we're doing a local pull.

It's just cleaner I think rather than doing both.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1987,10 +1987,10 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   else
     {
       pull_data->remote_name = g_strdup (remote_name_or_baseurl);
-      if (!ostree_repo_remote_get_gpg_verify (self, remote_name_or_baseurl,
+      if (!ostree_repo_remote_get_gpg_verify (self, pull_data->remote_name,
                                               &pull_data->gpg_verify, error))
         goto out;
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -2066,14 +2076,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
 
-  if (strcmp (soup_uri_get_scheme (pull_data->base_uri), "file") == 0)
-    {
-      g_autoptr(GFile) remote_repo_path = g_file_new_for_path (soup_uri_get_path (pull_data->base_uri));
-      pull_data->remote_repo_local = ostree_repo_new (remote_repo_path);
-      if (!ostree_repo_open (pull_data->remote_repo_local, cancellable, error))
-        goto out;
-    }
-  else
+  if (!pull_data->remote_repo_local)
```
alexlarsson: These two are not the same.
The old code checked if the url (likely taken from the remote config) starts with file:
However, remote_repo_local is only set above if you pass in a direct url rather than a remote name.

cgwalters:
Ok, tweaked.  :wrench:

alexlarsson:
looks good to me.

cgwalters:
You should have r+ rights here.

alexlarsson:
@cgwalters-bot r+  d56c071

PR #240, opened 2016-04-06
[merged] build: Find grub2-mkconfig a bit more automagically

cgwalters:
If one happens to not have grub2 installed, the previous commit
would cause us to fall back to `grub-mkconfig` with bad results.

We should likely just do dynamic detection and avoid pain, but
there's no harm in also allowing this to be statically determined.

The automagic here is that if `/etc/os-release` says `ID_LIKE=fedora`,
we know to use `grub2-mkconfig`.  But distro scripts should set this
in the build configurations using `--with-grub2-mkconfig=`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -264,12 +264,8 @@ AC_ARG_WITH(grub2-mkconfig-path,
 AS_IF([test x$with_grub2_mkconfig_path = x], [
   dnl Otherwise, look for the path to the system generator.  On some
   dnl distributions GRUB2 *-mkconfig executable has 'grub2' prefix and
-  dnl on some 'grub'.
-  AC_CHECK_PROGS(GRUB2_MKCONFIG, [grub2-mkconfig grub-mkconfig])
-  dnl Now if we don't find it in the buildroot, hackily search for fedora and
-  dnl dnl use grub2- if found
-  AS_IF([test x$GRUB2_MKCONFIG = x], [
-    AS_IF([grep 'ID_LIKE.*fedora' /etc/os-release], [GRUB2_MKCONFIG=grub2-mkconfig], [GRUB2_MKCONFIG=grub-mkconfig])])
+  dnl on some 'grub'.  We default to grub2-mkconfig.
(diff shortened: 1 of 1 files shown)


jlebon:
Nice, thanks! This works for me.

jlebon:
BTW, before this commit, ostree wouldn't even compile if you passed the `--with-builtin-grub2-mkconfig`. I wonder if we should add a sort of matrix test for all the --with flags.

configure.ac
```diff
@@ -257,14 +257,21 @@ AC_ARG_WITH(builtin-grub2-mkconfig,
...
+                           [Path to grub2-mkconfig]))
+AS_IF([test x$with_grub2_mkconfig_path = x], [
+  dnl Otherwise, look for the path to the system generator.  On some
+  dnl distributions GRUB2 *-mkconfig executable has 'grub2' prefix and
+  dnl on some 'grub'.
+  AC_CHECK_PROGS(GRUB2_MKCONFIG, [grub2-mkconfig grub-mkconfig])
+  dnl Now if we don't find it in the buildroot, hackily search for fedora and
+  dnl dnl use grub2- if found
+  AS_IF([test x$GRUB2_MKCONFIG = x], [
+    AS_IF([grep 'ID_LIKE.*fedora' /etc/os-release], [GRUB2_MKCONFIG=grub2-mkconfig], [GRUB2_MKCONFIG=grub-mkconfig])])
```
cgwalters: Yeah, penalizes Debian a bit, but @smcv can just add it to the package.

cgwalters:
Fixup :arrow_up:

jlebon:
@cgwalters-bot r+ ad0daa9

configure.ac
smcv: Yes, anything with a `--with-`, `--without-`, `--enable-` or `--disable-` is easily dealt with in any reasonable distro (particularly if you mention this change in the release notes so I remember that it's changed when it's time to package the next release).

I'd vaguely lean towards defaulting to whatever it's called upstream.

jlebon:
@cgwalters Ahh, you didn't switch over homu to use the Jenkins status instead for this repo?

cgwalters:
Re homu/status: Not yet, will do tomorrow, don't let me forget.

PR #241, opened 2016-04-07
[merged] build: Make tests/libreaddir-rand.so rule use AM_V_GEN

cgwalters:
So non-verbose builds don't have a verbose rule smack in the middle.


giuseppe:
@cgwalters-bot +r 7bc82d5

giuseppe:
@cgwalters-bot r+ 7bc82d5

PR #243, opened 2016-04-07
WIP offline support

alexlarsson:
This patch adds a fallback-to-cache flags for the fetch_summary() (and list_remote_refs()) that lets the client specify that on network error, the local cache should be used to report data.

This is pretty important for xdg-app, as we want the gnome-software UI to work to some degree even when offline. 

Note, I've not actually tested this code, its more of a strawman for how to handle offline support in general. In particular, this only handles the completely offline case too. 

Another interesting feature we may want is to keep using the local cache for some small time without even checking the signature. For instance, if gnome-software causes a series of calls during the regular check for updates there is no need to update the cache for each call. For that we could use a ALWAYS_USE_CACHE and then do an initial call without that to prime the cache.

A completely different approach would be to put all these caches in xdg-app. 

Opinions?

### DIFF
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -319,6 +319,14 @@ gboolean
 _ostree_repo_update_mtime (OstreeRepo        *self,
                            GError           **error);
 
+gboolean
+_ostree_repo_load_cache_summary (OstreeRepo        *self,
+                                 const char        *remote,
+                                 GBytes            **summary,
+                                 GBytes            **summary_sig,
+                                 GCancellable      *cancellable,
+                                 GError           **error);
+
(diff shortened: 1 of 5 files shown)


cgwalters:
Humm...

First, shouldn't gnome-software use https://developer.gnome.org/gio/stable/GNetworkMonitor.html ?

Would it make sense to have `ostree_repo_set_offline()` ?

alexlarsson:
It could well do, but in general IF you get some kind of network error during startup of gnome-software due to flaky network or something it is almost always better to show stale data than to show nothing plus some error dialog. The later is really never useful.

cgwalters:
What about TLS verification errors?

alexlarsson:
Yeah. Those would be useful I guess. But some kind of timeout error for instance would never be interesting.

PR #244, opened 2016-04-07
[merged] Support pathnames for --subpath=...

alexlarsson:
This allows you to pull a single file, rather than just a directory.


cgwalters:
Shouldn't be hard to add a test to `tests/test-pull-subpath.sh` right?

cgwalters:
@cgwalters-bot r+ ec81905

PR #246, opened 2016-04-08
[merged] cfg.mk: ignore syntax-check for git.mk

giuseppe:
another trivial patch.  Is it fine to self-accept these trivial patches?


cgwalters:
Mmm...I think we have enough people to get to trivial patches too, but let's see over time.

@cgwalters-bot r+ 1820fa1

cgwalters:
Related, any reason for `syntax-check` to be separate from `check`?

giuseppe:
I know some GNU projects keep these separated but I don't see any reason why it should not be done automatically on `make check`

PR #247, opened 2016-04-08
[merged] Support custom version of /etc/ostree/remotes.d

alexlarsson:
This lets xdg-app configure this to look in /etc/xdg-app/remotes.d for the system xdg-app ostree repo.

Additionally we look in the remotes.d dir for gpg keys, so that systemwide remotes can add trusted keys that are used for that repo only.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -4513,12 +4513,22 @@ find_keyring (OstreeRepo          *self,
               OstreeRemote        *remote,
               GCancellable        *cancellable)
 {
+  g_autoptr(GFile) remotes_d = NULL;
   g_autoptr(GFile) file = NULL;
   file = g_file_get_child (self->repodir, remote->keyring);
 
   if (g_file_query_exists (file, cancellable))
     {
       return g_steal_pointer (&file);
+    }
+
+  remotes_d = get_remotes_d_dir (self);
+  if (remotes_d)
+    {
+      g_autoptr(GFile) file2 = g_file_get_child (remotes_d, remote->keyring);
+
+      if (g_file_query_exists (file2, cancellable))
+        return g_steal_pointer (&file2);
     }
 
   if (self->parent_repo)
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo.c
```diff
@@ -2312,19 +2327,35 @@ append_one_remote_config (OstreeRepo      *self,
  out:
   return ret;
 }
-                                 
+
+static const char *
+get_remotes_d_dir (OstreeRepo          *self)
+{
+  if (self->remotes_config_dir != NULL)
+    return self->remotes_config_dir;
```
cgwalters: Here we're *not* using the sysroot...

src/libostree/ostree-repo.c
```diff
@@ -4518,6 +4519,16 @@ find_keyring (OstreeRepo          *self,
       return g_steal_pointer (&file);
     }
 
+  remotes_d_dir = get_remotes_d_dir (self);
+  if (remotes_d_dir)
+    {
+      g_autoptr(GFile) remotes_d = g_file_resolve_relative_path (self->sysroot_dir, remotes_d_dir);
```
cgwalters: But here we are prepending sysroot.

I'd say we should do one or the other.  Probably consistently prepend.

cgwalters:
I like being able to have the keys next to the remote file.  Might be nice to even support having them as inline data.

(If only we were using ECC keys then they'd fit on a single line...)

cgwalters:
https://github.com/ostreedev/ostree/pull/258

alexlarsson:
You're right, its nicer if get_remotes_d_dir() prepend the sysroot.

alexlarsson:
(pushed fix for that)

cgwalters:
Careful, `fixup! fixup!` doesn't work.  You might like https://github.com/cgwalters/homegit/blob/master/bin/git-addfixup

cgwalters:
I think you'll need to force-push to fix this.

alexlarsson:
pushed squashed version

cgwalters:
@cgwalters-bot r+ e91741c

PR #248, opened 2016-04-08
[merged] small cleanups

jlebon:
- Revert 'cannot' --> 'can not' (it's the exception!)
- Remove duplicate function
- Squelch compiler warnings

### DIFF
--- a/docs/manual/related-projects.md
+++ b/docs/manual/related-projects.md
@@ -63,7 +63,7 @@ awareness of BTRFS in dpkg/rpm itself) will be required.
 The OSTree author believes that having total freedom at the block
 storage layer is better for general purpose operating systems. For
 example, with OSTree, one is free to use BTRFS in any way you like -
-you can use a subvolume for `/home`, or you can not.
+you may decide to use a subvolume for `/home`, or not.
 
 Furthermore, in its most basic incarnation, the rpm/dpkg + BTRFS
 doesn't solve the race conditions that happen when unpacking packages
--- a/docs/manual/related-projects.md
+++ b/docs/manual/related-projects.md
@@ -63,7 +63,7 @@ awareness of BTRFS in dpkg/rpm itself) will be required.
 The OSTree author believes that having total freedom at the block
 storage layer is better for general purpose operating systems. For
 example, with OSTree, one is free to use BTRFS in any way you like -
-you can use a subvolume for `/home`, or you cannot.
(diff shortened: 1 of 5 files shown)


src/libotutil/ot-variant-utils.c
```diff
@@ -319,7 +319,7 @@ ot_variant_bsearch_str (GVariant   *array,
                         int        *out_pos)
 {
   gsize imax, imin;
-  gsize imid;
+  gsize imid = -1;
```
jlebon: GCC is unfortunately not smart enough to figure out that the loop is guaranteed to run at least once. Maybe if we put it an the else block of the `n == 0` condition, but this is easier.

docs/manual/related-projects.md
```diff
@@ -63,7 +63,7 @@ awareness of BTRFS in dpkg/rpm itself) will be required.
 The OSTree author believes that having total freedom at the block
 storage layer is better for general purpose operating systems. For
 example, with OSTree, one is free to use BTRFS in any way you like -
-you can use a subvolume for `/home`, or you cannot.
+you can use a subvolume for `/home`, or you can not.
```
giuseppe: this will break make syntax-check.  Fix for me if you add this part:

diff --git a/cfg.mk b/cfg.mk
index 3806209..d6a5039 100644
--- a/cfg.mk
+++ b/cfg.mk
@@ -1,4 +1,4 @@
-export VC_LIST_EXCEPT_DEFAULT=^(git.mk|lib/.*|m4/.*|md5/.*|build-aux/.*|src/gettext\.h|.*ChangeLog|buildutil/.*)$$
+export VC_LIST_EXCEPT_DEFAULT=^(docs/.*|git.mk|lib/.*|m4/.*|md5/.*|build-aux/.*|src/gettext\.h|.*ChangeLog|buildutil/.*)$$
 

and we skip the syntax checks on the documentation.

jlebon:
Thanks @giuseppe, I should have realized that would have broken syntax check. I just reworded it. Hmm, maybe I'll add it to the PR tester.

cgwalters:
Why not also include the change to exclude `docs/` from syntax-check?

jlebon:
Because I thought we still wanted it to run on them. Looking at the source though, it seems like there aren't that many checks applicable. Added exclusion :arrow_up:.

cgwalters:
@cgwalters-bot r+ e3bb3e5

PR #249, opened 2016-04-08
[merged] build: Move grub2-15_ostree back to pkglibexecdir

cgwalters:
It's not quite namespaced enough to have
`/usr/libexec/grub2-15_ostree`, and the Fedora spec file expects
things in `/usr/libexec/ostree`.  Changing the spec file would be
annoying as we'd need conditionals.


cgwalters:

gatispaeglis:
@cgwalters-bot r+ 581c994

PR #250, opened 2016-04-11
[merged] Add support for separate cache-dir

alexlarsson:
This lets you specify a cache dir outside of the repo, which is useful for repeated operations on a system repository that you don't have write rights to. For instance, the system xdg-app repo.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1783,12 +1783,15 @@ _ostree_repo_load_cache_summary_if_same_sig (OstreeRepo        *self,
                                              GError           **error)
 {
   gboolean ret = FALSE;
-  const char *summary_cache_sig_file = glnx_strjoina (_OSTREE_SUMMARY_CACHE_PATH, "/", remote, ".sig");
+  const char *summary_cache_sig_file = glnx_strjoina (_OSTREE_SUMMARY_CACHE_DIR, "/", remote, ".sig");
 
   glnx_fd_close int prev_fd = -1;
   g_autoptr(GBytes) old_sig_contents = NULL;
 
-  if (!ot_openat_ignore_enoent (self->repo_dir_fd, summary_cache_sig_file, &prev_fd, error))
+  if (self->cache_dir_fd == -1)
+    return TRUE;
+
+  if (!ot_openat_ignore_enoent (self->cache_dir_fd, summary_cache_sig_file, &prev_fd, error))
     goto out;
 
   if (prev_fd < 0)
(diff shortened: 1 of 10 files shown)


alexlarsson:
Pushed a rebase that fixes the conflict

src/libostree/ostree-repo-pull.c
```diff
@@ -1783,12 +1783,15 @@ _ostree_repo_load_cache_summary_if_same_sig (OstreeRepo        *self,
...
   gboolean ret = FALSE;
-  const char *summary_cache_sig_file = glnx_strjoina (_OSTREE_SUMMARY_CACHE_PATH, "/", remote, ".sig");
+  const char *summary_cache_sig_file = glnx_strjoina (_OSTREE_SUMMARY_CACHE_DIR, "/", remote, ".sig");
 
   glnx_fd_close int prev_fd = -1;
   g_autoptr(GBytes) old_sig_contents = NULL;
 
-  if (!ot_openat_ignore_enoent (self->repo_dir_fd, summary_cache_sig_file, &prev_fd, error))
+  if (self->cache_dir_fd == -1)
+    return TRUE;
```
cgwalters: I think the direct return style is nicer, but currently 99% of ostree doesn't use it primarily because in order to do so one has to be 100% cleanup macros.

I guess let's try to migrate smaller functions, so this is OK.

src/libostree/ostree-repo.c
```diff
@@ -2574,6 +2574,34 @@ ostree_repo_set_disable_fsync (OstreeRepo    *self,
...
+ */
+gboolean
+ostree_repo_set_cache_dir (OstreeRepo    *self,
+                           const char    *cache_dir,
+                           GCancellable   *cancellable,
+                           GError        **error)
+{
+  int fd;
+
+  if (!glnx_opendirat (-1, cache_dir, TRUE, &fd, error))
```
cgwalters: I'd `s/-1/AT_FDCWD/` even though we support both.

src/libostree/ostree-repo.c
```diff
@@ -2574,6 +2574,34 @@ ostree_repo_set_disable_fsync (OstreeRepo    *self,
...
+ * @cache_dir: Path of location to store caches for remotes
+ *
+ * Set a custom location for the cache directory used for e.g.
+ * per-remote summary caches. Setting this manually is useful when
+ * doing operations on a system repo as a user because you don't have
+ * write permissions in the repo, where the cache is normally stored.
+ */
+gboolean
+ostree_repo_set_cache_dir (OstreeRepo    *self,
+                           const char    *cache_dir,
```
cgwalters: For new APIs please make them fd-relative by default, so add an `int dfd` here.

src/libostree/ostree-repo.c
cgwalters: (Or with the above change use dfd + path...or we could just have the caller pass a dfd directly since it's not clear to me why we need to do the opendir)

alexlarsson:
Made it take dfd + path. We want to do the opendir because most callers will not have a dfd already. In fact, all current calles in ostree doesn't.

cgwalters:
Before we do this...I just want to think about the architecture a bit more.  So we're doing unprivileged fetches, using the system repo as a reference.  Then we have updated data (both in cachedir and in the new repo), then we have the system repo do a pull from the new repo.  But how would the system repo pull find the data in the cachedir?

alexlarsson:
This is not really for the unprivileged fetches. Its more like if you do:
```
ostree remote list --repo=/system/repo some-remote
```
Then the summary cache file is not useful, as we can't write it after we've downloaded it.

This is important because gnome-software will regularly do similar operations (that need summary data) but we run gnome-software as a user so it will not be cached.

For a the unprivileged fetch we don't need to specify a separate cache dir at all.

cgwalters:
Is the cache data going to be kept or discarded when going from `remote list` to `pull` ?

alexlarsson:
For a pull operation you need write access to the repo, so you will be running as a privileged user. In that case you should not pass it the untrusted users cache-directory, that would be pretty risky, so yeah, it would re-download caches if needed then.

cgwalters:
Ok, thanks for walking me through this.  It feels to me like we could try to pull in the cache data assuming it passes GPG verification, but eh...something for another time.

@cgwalters-bot r+ bdd0bcf

PR #251, opened 2016-04-11
[merged] Fix the symbol versions for ostree_repo_get_remote_*option

alexlarsson:
These were accidentally added to 2016.4 instead of 2016.5


cgwalters:
@cgwalters-bot r+ 07530ce

PR #252, opened 2016-04-11
[merged] ostree-repo-pull: always initialize flags_i

jlebon:
Otherwise we get undefined behaviour if the client didn't explicitly set
any flags.

Also, add documentation for all the other options supported by
`ostree_repo_pull_with_options()`.


cgwalters:
Nice, thanks for the docs update!

@cgwalters-bot r+ 2022712

PR #254, opened 2016-04-13
[merged] Initialize GMutex

alexlarsson:
ostree pull-local crashed for me in thread_closure_unref () doing:
    g_mutex_clear (&thread_closure->output_stream_set_lock);

Seems like we never initialize this mutex.


cgwalters:
Really minor but can you give this a subject prefix like:

`fetcher: Initialize output_stream_set_lock mutex`

alexlarsson:
Updated subject

cgwalters:
@cgwalters-bot r+ 6fe3c56

PR #255, opened 2016-04-13
[merged] refs: Add g_prefix_error around opendir for easier debugging

cgwalters:
Addresses:
  https://github.com/projectatomic/rpm-ostree/issues/264

We should consider moving this down into `glnx_opendirat`, but for now
a quick fix.


jlebon:
LGTM! @cgwalters-bot r+ 5ec6dca

PR #256, opened 2016-04-13
[merged] commit: Fix crash if dfd_iter is NULL

alexlarsson:
in write_directory_content_to_mtree_internal dfd_iter can be NULL,
for instance if commiting from --tree=ref=FOO. Don't blindly de-ref
it to avoid crashing.


cgwalters:
Hmm, not immediately obvious why our existing uses of `commit --tree=ref` in `basic-test.sh` aren't failing.  Anyways, a mystery for later.

@cgwalters-bot r+ e3de788

PR #260, opened 2016-04-15
[merged] static-delta: Initialize read_source_fd to -1

alexlarsson:
If not, we'll get ESPIPE when seeking on fd 0.


cgwalters:
@cgwalters-bot r+ c61adb8

PR #264, opened 2016-04-17
[merged] libtest: show files' contents when assertions about them fail

smcv:
I've seen an intermittent test failure in an autobuilder (sbuild)
environment where logs from failed builds cannot be retrieved,
but I can no longer reproduce it. Put the contents of the offending
file in the test's failing output so that if the failure comes back,
it can be debugged.

Signed-off-by: Simon McVittie <smcv@debian.org>

---

FWIW, the test failure was that `ostree --repo=repo pull origin main` aborted during `test-pull-resume.sh`. As I said, I can't reproduce it any more; it might have been a bug in something that has subsequently been upgraded in Debian unstable.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -100,13 +100,17 @@ assert_has_dir () {
 
 assert_not_has_file () {
     if test -f "$1"; then
-	echo 1>&2 "File '$1' exists"; exit 1
+        sed -e 's/^/# /' < "$1" >&2
+        echo 1>&2 "File '$1' exists"
+        exit 1
     fi
 }
 
 assert_not_file_has_content () {
     if grep -q -e "$2" "$1"; then
-	echo 1>&2 "File '$1' incorrectly matches regexp '$2'"; exit 1
+        sed -e 's/^/# /' < "$1" >&2
+        echo 1>&2 "File '$1' incorrectly matches regexp '$2'"
+        exit 1
     fi
 }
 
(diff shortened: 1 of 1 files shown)


tests/libtest.sh
```diff
@@ -100,13 +100,17 @@ assert_has_dir () {
 
 assert_not_has_file () {
     if test -f "$1"; then
-	echo 1>&2 "File '$1' exists"; exit 1
+        sed -e 's/^/# /' < "$1" >&2
+        echo 1>&2 "File '$1' exists"
```
cgwalters: How about inverting the echo and sed, like:

```
echo 1>&2 "File '$1' exists, content follows:"
sed -e 's/^/# /' < "$1" >&2
exit 1
```

Feels like it'd flow better.

cgwalters:
I've also seen intermittent errors with a few of the pull tests, I suspect running them in a loop would help.  I'll start a run now.

tests/libtest.sh
smcv: I did it this way round so that the reason for failure ($1 exists) is closest to the notification that the build has failed (presumably at the end). After a failed build on an autobuilder, you're probably reading the log in reverse (or at least that's what I do).

If you'd strongly prefer the echo before the sed I can do that, though.

cgwalters:
Fair enough.

@cgwalters-bot r+ 0e891db

PR #265, opened 2016-04-18
[merged] Add --prefix and --subpath argument to ostree export

alexlarsson:
--prefix is already  supported in the CLI, and I need --prefix to do some prefix rewriting when exporting xdg-apps as OCI archives (i.e. convert files to usr).

### DIFF
--- a/src/libostree/ostree-repo-libarchive.c
+++ b/src/libostree/ostree-repo-libarchive.c
@@ -483,7 +483,7 @@ file_to_archive_entry_common (GFile         *root,
   if (opts->path_prefix && opts->path_prefix[0])
     {
       g_autofree char *old_pathstr = pathstr;
-      pathstr = g_build_filename (opts->path_prefix, old_pathstr, NULL);
+      pathstr = g_strconcat (opts->path_prefix, old_pathstr, NULL);
     }
 
   if (pathstr == NULL || !pathstr[0])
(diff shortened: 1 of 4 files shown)


cgwalters:
This PR hit the same thing I did earlier with `/dev/fuse` being randomly writable for some reason.  Will investigate.

cgwalters:
Would be nice to also test `--prefix` *and* `--subpath` together since I'm guessing that'll be the most common case.   But that's super minor and I'm mostly just waiting for #266

src/libostree/ostree-repo-libarchive.c
```diff
@@ -480,6 +480,12 @@ file_to_archive_entry_common (GFile         *root,
   g_autoptr(GVariant) xattrs = NULL;
   time_t ts = (time_t) opts->timestamp_secs;
 
+  if (opts->path_prefix && opts->path_prefix[0])
+    {
+      g_autofree char *old_pathstr = pathstr;
+      pathstr = g_build_filename (opts->path_prefix, old_pathstr, NULL);
```
cgwalters: One thing that is different from `git archive --prefix`  is that we're automatically inserting e.g. a `/` if the host platform is Unix.  In the unlikely event of a Windows port, we'll insert `\`.  Which in turn would make `--prefix` unusable on Windows to generate Unix archives.

So it's more annoying, but strictly more powerful, to use `g_strconcat()` instead.  The argument that it matches git is important for OSTree.

cgwalters:
I added another comment above - if you prefer to keep it as is though that's fine, I'm also OK to prefer automatic `/` over precluding Windows generating Unix archives for the moment.

alexlarsson:
I wasn't even aware of git archive --export. I guess great mind think alike.
Anyway, I changed it to behave like git, which uses the prefix as-is, i.e it never adds a /, i.e. you can do foo.txt => prefixfoo.txt.

cgwalters:
@cgwalters-bot r+ b9bdcf8

PR #266, opened 2016-04-18
[merged] libtest: Skip FUSE if CAP_SYS_ADMIN not in bounding set

cgwalters:
I'm not sure if something changed or there was a race condition , or
if extraterrestrials intervened to make the tests pass before, but
in the Docker container `/dev/fuse` does exist and since we appear
to be running as uid 0 it'll be writable.

We previously discussed that another requirement for FUSE (and most
setuid binaries) is having `CAP_SYS_ADMIN`, so let's check for that in
the bounding set too.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -370,6 +370,11 @@ skip_without_fuse () {
         exit 0
     fi
 
+    if ! capsh --print | grep -q 'Bounding set.*[^a-z]cap_sys_admin'; then
+	echo "1..0 # SKIP No cap_sys_admin in bounding set, can't use FUSE"
+        exit 0
+    fi
+
     if ! [ -w /dev/fuse ]; then
         echo "1..0 # SKIP no write access to /dev/fuse"
         exit 0


cgwalters:
@cgwalters-bot delegate=smcv

smcv:
Sorry, I don't speak fluent capabilities(7) yet... talk me through it. Is the reasoning here that the bounding set is the set of capabilities that we could possibly acquire (and in particular, that we would acquire by executing a setuid-root binary), therefore if we're in a container that omits `CAP_SYS_ADMIN` from its bounding set, there's no way we're going to be mounting anything?

(Also, how do I tell your bot I like a change again? Perhaps it could link to its own documentation when told to delegate control to someone?)

cgwalters:
Your rationale is exactly it!  For docs on Homu, see http://escher.verbum.org:54856/

FWIW I was more *adding* you as a reviewer since you'd worked on this previously, and not assigning.

cgwalters:
@jlebon WDYT?

jlebon:
This makes sense to me and is close to what I had originally IIRC. So r+ from me (feel free to r+ yourself if you're not waiting on anything else).

... /me quietly sweeps under the carpet the weird `/dev/fuse` behaviour and points the finger at Docker.

cgwalters:
Why not provide the r+ to Homu?

jlebon:
@cgwalters-bot r+ 1496683

PR #268, opened 2016-04-20
Updates on /boot

gatispaeglis:
Conversation history in https://github.com/ostreedev/ostree/pull/156

### DIFF
--- a/src/libostree/ostree-bootloader-uboot.c
+++ b/src/libostree/ostree-bootloader-uboot.c
@@ -29,6 +29,10 @@
 #include "otutil.h"
 
 #include <string.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <fcntl.h>
 
 struct _OstreeBootloaderUboot
 {
@@ -69,13 +73,17 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot     *self,
                                         GCancellable          *cancellable,
                                         GError               **error)
 {
+  gboolean ret = FALSE;
   g_autoptr(GPtrArray) boot_loader_configs = NULL;
   OstreeBootconfigParser *config;
   const char *val;
+  g_autofree char *boot_path = NULL;
+  g_autoptr(GFile) uenv_file = NULL;
+  char uenv_path[2048];
 
   if (!_ostree_sysroot_read_boot_loader_configs (self->sysroot, bootversion, &boot_loader_configs,
                                                  cancellable, error))
-    return FALSE;
+    goto out;
 
   /* U-Boot doesn't support a menu so just pick the first one since the list is ordered */
   config = boot_loader_configs->pdata[0];
@@ -85,9 +93,12 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot     *self,
     {
       g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
                    "No \"linux\" key in bootloader config");
-      return FALSE;
+      goto out;
     }
   g_ptr_array_add (new_lines, g_strdup_printf ("kernel_image=%s", val));
+
+  boot_path = strndup (val, strlen (val) - strlen ("/vmlinuz"));
+  g_ptr_array_add (new_lines, g_strdup_printf ("bootdir=%s", boot_path));
 
   val = ostree_bootconfig_parser_get (config, "initrd");
   if (val)
@@ -97,7 +108,32 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot     *self,
   if (val)
     g_ptr_array_add (new_lines, g_strdup_printf ("bootargs=%s", val));
 
-  return TRUE;
+  /* Merge with user's uEnv.txt if it exists */
+  snprintf (uenv_path, sizeof(uenv_path), "boot/%s/uEnv.txt", boot_path);
+  uenv_file = g_file_get_child (self->sysroot->path, uenv_path);
+  if (g_file_query_exists (uenv_file, cancellable))
+    {
+      g_autoptr(GInputStream) instream = NULL;
+      g_autoptr(GDataInputStream) datastream = NULL;
+      gsize len;
+
+      instream = (GInputStream*)g_file_read (uenv_file, cancellable, error);
+      if (!instream)
+        goto out;
+
+      datastream = g_data_input_stream_new (instream);
+      while (TRUE)
+        {
+          val = g_data_input_stream_read_line (datastream, &len, cancellable, error);
+          if (!val)
+            break;
+          g_ptr_array_add (new_lines, (char *)val);
+        }
+    }
+
+  ret = TRUE;
+out:
+  return ret;
 }
 
 static gboolean
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1366,6 +1366,59 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
         }
     }
 
+  {
+    /* Copy other files that are stored in the boot directory. Lets keep this simple:
+     *
+     * - Ignore subdirectories, only files in root of the boot directory are copied. There is
+     *   overhead of copying files to boot partition, therefore the amount of files in the boot/
+     *   should be kept to the minimum and subdirectories shouldn't really be necessary.
+     * - Files on the boot partition are updated only with major releases, when kernel/initramfs
+     *   bootcsum changes. Files that require frequent updates should not be stored here.
+     */
+    g_autoptr(GFileEnumerator) dir_enum = NULL;
+    g_autoptr(GFile) deployments_bootdir = g_file_get_child (deployment_dir, "boot");
+    dir_enum = g_file_enumerate_children (deployments_bootdir, OSTREE_GIO_FAST_QUERYINFO,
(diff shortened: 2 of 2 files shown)


gatispaeglis:
Other option which is quite flexible is to have a meta file (boot-include for example) as in /usr/lib/ostree-boot/boot-include. Where this file contains paths to files which to include into kernel/initramfs updates.

gatispaeglis:
For now updated to glnx_* APIs where it made sense and fixed the necessary parts to make it work with v2016.5.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1361,6 +1362,57 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+   *   bootcsum changes. Files that require frequent updates should not be stored here.
+   */
+  if (!glnx_dirfd_iterator_init_take_fd (dup (tree_boot_dfd), &dfd_iter, error))
+    goto out;
+
+  while (TRUE)
+    {
+      struct dirent *dent;
+
+      if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
```
cgwalters: FYI you can use `glnx_dirfd_iterator_next_dent_ensure_dtype` and look at `dent->d_type`, see e.g. https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-refs.c#L136

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1361,6 +1362,57 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+
+      if (fstatat (dfd_iter.fd, dent->d_name, &stbuf, 0) != 0)
+        {
+          if (errno == ENOENT)
+              continue;
+          glnx_set_error_from_errno (error);
+          goto out;
+        }
+
+      if (g_str_has_prefix (dent->d_name, "vmlinuz-") || g_str_has_prefix (dent->d_name, "initramfs-")
```
cgwalters: Wait, why is this code also copying the kernel?  That has to duplicate the existing code right?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1361,6 +1362,57 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+        continue;
+
+      if (fstatat (bootcsum_dfd, dent->d_name, &stbuf, 0) != 0)
+        {
+          if (errno != ENOENT)
+            {
+              glnx_set_prefix_error_from_errno (error, "fstat %s", dent->d_name);
+              goto out;
+            }
+          if (!glnx_file_copy_at (tree_boot_dfd, dent->d_name, &stbuf,
```
cgwalters: I think we have a utility function somewhere that's `cp -al` - because if `/boot` isn't a separate partition, we can just use hardlinks and avoid the space duplication.

src/libostree/ostree-bootloader-uboot.c
```diff
@@ -97,7 +108,31 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot     *self,
   if (val)
     g_ptr_array_add (new_lines, g_strdup_printf ("bootargs=%s", val));
 
-  return TRUE;
+  /* Append user's uEnv.txt if it exists */
+  snprintf (uenv_path, sizeof(uenv_path), "boot/%s/uEnv.txt", bootdir);
```
cgwalters: I'd use `glnx_strjoina()` instead.

cgwalters:
"Edit /usr/lib/ostree-boot/uEnv.txt"

Do you mean `/boot/uEnv.txt` ?  `/usr` should be a r-o bind mount with the system defaults, right?  Or is this s/admin/system builder/ ?

cgwalters:
Overall the code style is good, but this is going to have a lot of implications and I'm going to need some time to work through the matrix.

src/libostree/ostree-sysroot-deploy.c
cgwalters: Ah, we *used* to but this regressed in  1810de2b51680dbf35fdbd33d0a8d7e65eadc91f .  We switched from `gs_file_linkcopy_sync_data()` to `glnx_file_copy_at()`.

I'll look at a patch.

src/libostree/ostree-sysroot-deploy.c
gatispaeglis: It does not duplicate, kernel/initramfs is skipped - "continue".

src/libostree/ostree-sysroot-deploy.c
gatispaeglis: Here I used the same copy function as is used for kernel/initramfs, higher up in the function.

gatispaeglis:
> Edit /usr/lib/ostree-boot/uEnv.txt

I meant on the server side by admin/system-builder. This is not something that client devices would need to fiddle with.

I will take a look at the suggested alternative APIs. I am not familiar with the used libraries, so feel free to suggest when there is a better function to use.

> I'm going to need some time to work through the matrix.

No rush. For now I was interested to update my patches so I can use the latest release of OSTree. I am also interested in adding auto tests to prevent regressions, but before that we need to agree how to handle the copying of files from /usr/lib/ostree-boot. Dedicated sub paths? Meta data file? 

Or what I have right now? But with a conditional check:
if_exist (/usr/lib/ostree-boot/do-copy) {
  copy all top level files
} else {
  work as before
}

Then installer can just do:
touch /usr/lib/ostree-boot/do-copy to enable this feature.

This would be the same pattern as I have in firmware-convenience-symlinks feature, where OS installer enables feature by creating a symlink. Thus, decision to enable a certain features is based on paths being present.

cgwalters:
The u-boot change is independent of the other two, right?  Can you make it a separate PR?

cgwalters:
I think I'm roughly in favor of having magical filenames in `boot/` add extra semantics...as long as we still at least try to do things like have the different bootloaders/distributions behave the same way where we can.

gatispaeglis:
> The u-boot change is independent of the other two, right?

Its not independent, the u-boot change depends on having uEnv.txt in bootcsum dir:
`snprintf (uenv_path, sizeof(uenv_path), "boot/%s/uEnv.txt", bootdir);`

cgwalters:
How about using subdirectories of `/usr/lib/ostree-boot` like:

 - `ostree-bootdata`: Everything here is treated the same as the kernel/initramfs. So rpm-ostree would include the `config` and `ksyms` files etc.
 - `ostree-installonce`: Installers should pre-populate the `/boot` partition with this data on initial OS install, and it's never touched afterwards

Would the "firmware convenience" patch fit into `installonce`?

gatispaeglis:
> Would the "firmware convenience" patch fit into installonce?

Not really. The "ostree-installonce" would be unused in this case.

The generated symlink path looks as follows:

`6x_bootscript -> loader/6x_bootscript -> ../ostree/qt-os-1b10ce9f7bf26339e143a3282a6690d1cb34f0894f855cccbd36ba70628957fd/6x_bootscript`

After an update where bootcsum changes the symlink chain would be reset to:

`6x_bootscript -> loader/6x_bootscript -> ../ostree/qt-os-590db09c66551670019a487992f4dae9cb2067e241f7c7fefd6b3d35af55895b/6x_bootscript`

The 6x_bootscript file uses the symlink convenience feature and does require update when kernel version changes, therefore would need to be stored in ostree-bootdata.

I really would like to try an approach with a configuration files that enables/disables certain features in the sysroot, but not sure when I will get time to experiment with this.

cgwalters:
The high level here is I'm a bit worried about possibly breaking something in the Fedora/CentOS/ProjectAtomic space with this, but in the short term I'm open to either:

 - compile time option
 - runtime options

cgwalters:
And if there's anything small we can land here that would make maintaining patches easier, let's do that.

gatispaeglis:
- [ ] Needs autotests

gatispaeglis:
I will rebase this when https://github.com/ostreedev/ostree/pull/442 and https://github.com/ostreedev/ostree/pull/463 will go in.

Also I was thinking to rename `/usr/lib/ostree-boot/ostree-bootcsum` and make into a hidden file `.ostree-bootcsum-source`

gatispaeglis:
Will split this into several Pull Requests.

gatispaeglis:
After refactoring I realized that https://github.com/ostreedev/ostree/pull/268/commits/d76d6a620d001e1ac17843067735ad77375b7305 does not depend on "updates-on-boot" feature, but can grab system's uEnv.txt directly from the new default deployment (as it is a simple text file containing only u-boot env).

This is the patch set that I would like to get in before adding the final peace "the firmware convenience function".

vtolstov:
@cgwalters do you have some time to check this ?

vtolstov:
does it possible to fix merge conflict and merge this pr ?

davisRoman:
I'm working on a system that not only has a kernel and initramfs but also has a devicetree file that needs to be loaded by u-boot( which means that it needs to be made available in the uEnv.txt) This pull request seems to address this. Any updates on whether or not this will get merged? If not, can you suggest an alternate method that I can look at so that u-boot is able to load this additional devicetree file? Thank you.

gatispaeglis:
@vtolstov , @davisRoman , @cgwalters , Hi too all :) I am currently not working on the project for which I created these patches, but you guys are welcome to take over this PR.

PR #269, opened 2016-04-20
[merged] repo: Allow loading files staged in the transaction

alexlarsson:
Currently we can load metadata from the stage dir, but not file
data, which makes no sense.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3066,6 +3066,13 @@ ostree_repo_load_file (OstreeRepo         *self,
                                     error))
         goto out;
 
+      if (fd < 0 && self->commit_stagedir_fd != -1)
+        {
+          if (!ot_openat_ignore_enoent (self->commit_stagedir_fd, loose_path_buf, &fd,
+                                        error))
+            goto out;
+        }
+
       if (fd != -1)
         {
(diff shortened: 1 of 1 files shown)


cgwalters:
Hmm...what were you trying to do where you hit this?  Note that `ostree_repo_has_object()` which is used by `ostree-repo-pull.c` *does* look at the stagedir for content as well.

So we won't redownload content objects from the stagedir, right?

I'm not opposed to this patch...but I just want to understand what it fixes.

alexlarsson:
Oh, i just have some code that does a commit in a transaction and then looks at the commited files before closing the transaction. I don't think this is a "bug" as is, but if we can easily support that, why not?

cgwalters:
Okay, I'm fine with that.  In the future though please try to include the concrete use case in the commit message, something like "It'd be useful for xdg-app to be able to read committed files before closing the transaction, so let's support that".

@cgwalters-bot r+ ad1cea3

cgwalters:
Keep in mind though you can't depend on this behavior until a new release, so it's probably safest to change xdg-app to read the files after a commit anyways.

cgwalters:
BTW I submitted a followup in https://github.com/ostreedev/ostree/pull/272

PR #270, opened 2016-04-20
[merged] Remove unused variables from install_deployment_kernel

gatispaeglis:
I have seen more instance of this, but currently I am looking at this function.
Why doesn't ostree use compiler flags to automatically detect unused variables?


cgwalters:
It's because GCC treats variables with the cleanup attribute as used.  We should probably pull in an LLVM build since it doesn't do that.

@cgwalters-bot r+ 45a43bf

PR #271, opened 2016-04-20
[merged] Add URL override for fetching objects

krnowak:
This adds an `override-url` option to `ostree_repo_pull_with_options`, which, if specified, overrides the URL taken from repo config for a given remote.

This also adds a `--url` option to `ostree pull` which uses the above option.

### DIFF
--- a/src/ostree/ot-builtin-trivial-httpd.c
+++ b/src/ostree/ot-builtin-trivial-httpd.c
@@ -389,8 +389,15 @@ ostree_builtin_trivial_httpd (int argc, char **argv, GCancellable *cancellable, 
     {
       GOutputStream *stream = NULL;
 
-      if (g_strcmp0(opt_log, "-") == 0)
-        stream = G_OUTPUT_STREAM (g_unix_output_stream_new (STDOUT_FILENO, FALSE));
+      if (g_strcmp0 (opt_log, "-") == 0)
+        {
+          if (opt_daemonize)
+            {
+              ot_util_usage_error (context, "Cannot use --log-file=- and --daemonize at the same time", error);
+              goto out;
+            }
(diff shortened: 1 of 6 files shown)


cgwalters:
Can you add a test case?  Examples are in `tests/test-pull-commit-only.sh` and `tests/pull-test.sh` etc.

cgwalters:
One test case would be to have a repo that has an upstream remote, clone the remote into a different path, then rm -rf the original, then do a pull with `--url` from the mirror.

krnowak:
Sure, will get to it tomorrow.

krnowak:
Updated. I added a `--log-file` flag to trivial-httpd, because I couldn't get why the test was failing. Apparently I did not pass the `--mirror` flag to `ostree pull`, so fetching `refs/heads/main` from the mirror was returning 404. Not sure if you want it in a separate PR, please let me know.

src/ostree/ot-builtin-trivial-httpd.c
```diff
@@ -351,6 +385,30 @@ ostree_builtin_trivial_httpd (int argc, char **argv, GCancellable *cancellable,
 
   app->root = g_file_new_for_path (dirpath);
 
+  if (opt_log)
+    {
+      GOutputStream *stream = NULL;
+
+      if (g_strcmp0(opt_log, "-") == 0)
```
cgwalters: Missing space between identifier and paren here.

cgwalters:
Test LGTM.

krnowak:
Updated. Added the check for `--daemonize` and `--log-file=-` too.

cgwalters:
@cgwalters-bot r+ 27fa1ac

cgwalters:
In the future, I generally prefer test cases to come in the same commit as the code introducing them.  Not a big deal though.

krnowak:
Alright, will remember next time.

PR #272, opened 2016-04-21
[merged] repo: Simplify internal has_object() lookup code

cgwalters:
There was some leftover intermediate cruft here I noticed
while reviewing another patch:

 - We had an output `GFile*` for that was never used
 - We required the caller to allocate the loose pathbuf, but
   none of them ever reused it
 - We had an extra intermediate function

Also while looking at this, I'm now uncertain whether some of the
callers of `_ostree_repo_has_loose_object` should really be invoking
`ostree_repo_has_object()`, but let's leave that aside for now.


alexlarsson:
@cgwalters-bot r+ 41e067f

PR #273, opened 2016-04-22
[merged] repo: Fix temporary directory allocation

alexlarsson:
We looked for and locked old temporary directories so we can
reuse them if not in use. However, once we found one that
we can reuse we didn't stop iterating, and eventually we
reached the end. This means we can lock multiple dirs.


giuseppe:
@cgwalters-bot r+ 4addf60

PR #274, opened 2016-04-22
[merged] repo: Handle parent repos with different remote configuration

alexlarsson:
In the case we have a repo with a parent, and the child repo has a
remote called "foo", but some option is unset. Then when we look up
the parent repo for a value before using the default we will fail due
to the parent not having the "foo" remote. As soon as we find the
requested remote at some point in the hierarchy we need to ignore further
errors and use the default value.


giuseppe:
it looks good to me

@cgwalters-bot r+ 87c9aee

cgwalters:
Seems reasonable, but test case would be nice.

cgwalters:
@cgwalters-bot retry

PR #275, opened 2016-04-22
[merged] Rework libarchive import support

jlebon:
From the first commit message:

```
ostree-repo-libarchive.c: major refactor

- Make hardlink handling more generic. The previous strategy worked for
  tar archives, but not for cpio. It now works for both.
- Add support for SEL labeling (through the OstreeRepoCommitModifier)
- Add support for xattr_callback (through the OstreeRepoCommitModifier)
- Add support for filter (through the OstreeRepoCommitModifier)
- Add a use_ostree_convention option
```

The remaining commits add tests for the new features.

---

The main motivation for this work is https://github.com/projectatomic/rpm-ostree/pull/107. With these patches, `rpm-ostree` will be able to much more easily import RPMs into repos, with full support for filecaps and SELinux labeling.


src/libostree/ostree-repo-libarchive.c
```diff
@@ -45,269 +47,685 @@ propagate_libarchive_error (GError      **error,
                "%s", archive_error_string (a));
 }
 
+static inline const char *
+path_relative (const char *src)
+{
+  if (src[0] == '.' && src[1] == '/')
+    src += 2;
```
cgwalters: Doesn't this also need to be a loop so we canonicalize paths like `././foo` ?

jlebon: Good point.

jlebon:
:arrow_up: Fixup added!

cgwalters:
Only took a quick scan, but the high level looks really good.  Will look more next week.

src/libostree/ostree-repo-libarchive.c
```diff
@@ -50,10 +50,10 @@ propagate_libarchive_error (GError      **error,
...
 {
-  if (src[0] == '.' && src[1] == '/')
+  if (src[0] == '.' && src[1] == '\0')
+    return src + 1;
+  while (src[0] == '.' && src[1] == '/')
     src += 2;
-  else if (src[0] == '.' && src[1] == '\0')
-    src++;
   while (src[0] == '/')
     src++;
```
cgwalters: This still won't handle `/./foo` right?  I think the code needs to be more like:

```
while (TRUE)
  {
    if (src[0] == '.' && src[1] == '/')
      {
        src += 2;
        continue;
      }
   if (src[0] == '/')
     {
       src++;
       continue
     }
   break;
}
```

Or something?  There's got to be some LGPL-or-below code we can look at as a reference...okay, some quick `git grep relative` in libarchive led me to [strip_absolute_path](https://github.com/libarchive/libarchive/blob/e61afbd463d1b06fcf0d7c0e37feeb9334954af6/tar/util.c#L398) which is rather terrifying...we could likely drop the Windows code at least.

tests/test-libarchive-import.c
```diff
@@ -243,6 +301,166 @@ test_libarchive_ignore_device_file (gconstpointer data)
...
+  struct archive *a = archive_read_new ();
+  OstreeRepoImportArchiveOptions opts = { 0 };
+  OstreeRepoCommitModifier *modifier = NULL;
+  char buf[7] = { 0 };
+
+  GVariant* cb (OstreeRepo  *repo,
+                const char  *path,
+                GFileInfo   *file_info,
+                gpointer     user_data)
+  {
```
cgwalters: clang doesn't support GNU C nested functions, and I'm kind of with them there.  Instead it implements [blocks](https://en.wikipedia.org/wiki/Blocks_%28C_language_extension%29).

cgwalters: (To clarify I'm suggesting we use neither for now, i.e. just hoist this out)

src/libostree/ostree-repo-libarchive.c
```diff
@@ -45,269 +47,685 @@ propagate_libarchive_error (GError      **error,
...
+aic_get_final_entry_pathname (OstreeRepoArchiveImportContext *ctx)
+{
+  const char *pathname = archive_entry_pathname (ctx->entry);
+  g_autofree char *final = aic_get_final_path (ctx, pathname);
+
+  /* squash trailing slashes some archives append to dir names */
+  if (g_str_has_suffix (final, "/"))
+    {
+      g_autofree char *tmp = final;
+      final = g_strndup (final, strlen (final) - 1);
```
cgwalters: Also need to handle multiple trailing `/` here?  Would be a bit more efficient too to get the end pointer, then walk backwards and replace them with `NUL` in place.

src/libostree/ostree-repo-libarchive.c
```diff
@@ -45,269 +47,685 @@ propagate_libarchive_error (GError      **error,
...
+      g_autoptr(GVariantBuilder) builder =
+        ot_util_variant_builder_from_variant (xattrs, G_VARIANT_TYPE
+                                                        ("a(ayay)"));
+
+      if (!builder_add_label (builder, ctx->modifier->sepolicy, abspath, mode,
+                              cancellable, error))
+        return FALSE;
+
+      if (xattrs)
+        g_variant_unref (xattrs);
```
cgwalters: Doesn't this need to be above the `return`?

src/libostree/ostree-repo-libarchive.c
```diff
@@ -45,269 +47,685 @@ propagate_libarchive_error (GError      **error,
...
+static void
+aic_add_deferred_hardlink (OstreeRepoArchiveImportContext *ctx,
+                           const char        *hardlink,
+                           DeferredHardlink  *dh)
+{
+  gboolean new_slist;
+  GSList *slist;
+
+  slist = g_hash_table_lookup (ctx->deferred_hardlinks, hardlink);
+  new_slist = slist == NULL;
```
cgwalters: IMO parses more easily visually as: `new_slist = (slist == NULL)`

src/libostree/ostree-repo-libarchive.c
```diff
@@ -45,269 +47,685 @@ propagate_libarchive_error (GError      **error,
...
+   * */
+
+  g_autofree char *hardlink = aic_get_final_entry_hardlink (ctx);
+  guint64 size = g_file_info_get_attribute_uint64 (fi, "standard::size");
+
+  if (hardlink == NULL || size > 0)
+    if (!aic_import_file (ctx, parent, path, fi, cancellable, error))
+      return FALSE;
+
   if (hardlink)
```
cgwalters: This could be `else if (hardlink)` right?

src/libostree/ostree-repo-libarchive.c
```diff
@@ -45,269 +47,685 @@ propagate_libarchive_error (GError      **error,
...
+  while (g_hash_table_iter_next (&iter, &key, &value))
+    if (!aic_import_deferred_hardlinks_for (ctx, key, value, error))
+      return FALSE;
+
+  return TRUE;
 }
-#endif
+
+static void
+slist_free (void *data)
```
cgwalters: I'd call this `deferred_hardlinks_free` or so.

src/libostree/ostree-repo-libarchive.c
```diff
@@ -45,269 +47,685 @@ propagate_libarchive_error (GError      **error,
...
+{
+  void slist_free_item (void *data) {
+    DeferredHardlink *dh = data;
+    g_object_unref (dh->parent);
+    g_free (dh->path);
+    g_slice_free (DeferredHardlink, dh);
+  }
```
cgwalters: Another nested function.

src/ostree/ot-builtin-commit.c
```diff
@@ -147,13 +149,68 @@ parse_statoverride_file (GHashTable   **out_mode_add,
...
+    goto out;
+
+  ret_hash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+  lines = g_strsplit (contents, "\n", -1);
+
+  for (iter = lines; iter && *iter; iter++)
+    {
+      const char *line = *iter;
+
+      if (*line == '-')
```
cgwalters: My memory is a bit dim on why the statoverride files start with `+`...maybe so they could have comments with `#`?  Dunno.  If so I doubt it was used.

It seems like it'd be simpler to just have a newline-separated list.  (Anyone who complains about this should really be using the API IMO)

src/ostree/ot-builtin-commit.c
```diff
@@ -147,13 +149,68 @@ parse_statoverride_file (GHashTable   **out_mode_add,
...
+  g_autofree char *contents = NULL;
+  char **lines = NULL;
+
+  path = g_file_new_for_path (opt_skiplist_file);
+
+  if (!g_file_load_contents (path, cancellable, &contents, &len, NULL, error))
+    goto out;
+
+  ret_hash = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
+  lines = g_strsplit (contents, "\n", -1);
```
cgwalters: Could avoid lots of mallocs with iteration via `strchr (contents, '\n')`...doesn't matter for this case but just pointing it out as something to keep in mind for other more perf sensitive code.

src/libostree/ostree-repo-libarchive.c
jlebon: Ahhh OK, I will tweak this so we at least handle the non-expensive cases.

tests/test-libarchive-import.c
jlebon: Sounds good.

src/libostree/ostree-repo-libarchive.c
jlebon: Smart! I like that.

src/libostree/ostree-repo-libarchive.c
jlebon: Hmm, I don't think so. If `builder_add_label()` fails, the xattrs will be cleaned up by the cleanup func. If `builder_add_label()` succeeds, we need to drop the ref we have to the user's xattrs, if any, so that we can replace it with the variant ref we get from our builder (which already has the xattrs the user returned + the selabel).

src/libostree/ostree-repo-libarchive.c
jlebon: Will do.

src/libostree/ostree-repo-libarchive.c
jlebon: We need to also add to the list hardlinks which carry a payload so that they can be found later. E.g. in `aic_import_from_hardlink` we correct the initial non-hardlink file to point to the hardlink csum.

src/libostree/ostree-repo-libarchive.c
jlebon: Sounds good.

src/libostree/ostree-repo-libarchive.c
jlebon: Will pull out.

src/ostree/ot-builtin-commit.c
jlebon: Sure, I'll rework stat_override and this to be simpler and more memory efficient.

src/ostree/ot-builtin-commit.c
jlebon: I started working on doing this but felt silly doing it, so I went back to using `g_strsplit()`. It just felt like the trade-off in legibility was not worth it for something that people should probably use the API for if it somehow becomes a bottleneck.

jlebon:
Added fixups! :arrow_up:

src/ostree/ot-builtin-commit.c
cgwalters: Yeah sorry, I meant to make clear I wasn't saying we shouldn't change this code, agree it's not worth it.

src/ostree/ot-builtin-commit.c
cgwalters: Well I don't think we can change the statoverride format, it'd break gcontinuous (and maybe other people have discovered it)

jlebon: The leading + will still work actually. It goes through `g_strtod()` which allows it. It's just that we don't require it now.

src/ostree/ot-builtin-commit.c
jlebon: Right, makes sense now rereading your comment. :+1: I do like to "do it right" usually even when it's for a trivial case. But yeah, guess it wasn't worth it here.

src/ostree/ot-builtin-commit.c
```diff
@@ -97,104 +97,64 @@ static GOptionEntry options[] = {
...
-            }
-          
-          mode_add = (guint32)(gint32)g_ascii_strtod (line + 1, NULL);
-          g_hash_table_insert (ret_hash,
-                               g_strdup (spc + 1),
-                               GUINT_TO_POINTER((gint32)mode_add));
-        }
-    }
+  lines = g_strsplit (contents, "\n", -1);
+  for (char **iter = lines; iter && *iter; iter++)
```
cgwalters: I'd be a bit more comfortable if we had braces `{` for nested loops/conditionals...otherwise it's too easy down the line to add an additional statement and have it go in the wrong section.

src/ostree/ot-builtin-commit.c
```diff
@@ -97,104 +97,64 @@ static GOptionEntry options[] = {
...
+    if (**iter != '\0') /* skip empty lines at least */
+      if (!cb (*iter, cbdata, error))
```
cgwalters: Or alternatively here combine.

tests/basic-test.sh
```diff
@@ -181,15 +181,15 @@ echo "ok commit with no xattrs"
 
 cd ${test_tmpdir}
 cat > test-statoverride.txt <<EOF
-+2048 /a/nested/3
+2048 /a/nested/3
```
cgwalters: Can we add a new test, since we still need `+` to work?

src/ostree/ot-builtin-commit.c
cgwalters: (Although https://gnu.wildebeest.org/blog/mjw/2016/01/09/looking-forward-to-gcc6-nice-new-warnings/ will help with that)

jlebon:
I rebased and added fixups to address the last comments. I squashed all the previous fixups and only kept the new ones.

Since for now we're punting on the name:id issues some clients will face, I went a different way than we had originally talked about re. passing hashtables through the `OstreeRepoImportArchiveOptions`. I simply added a `callback_with_entry_pathname` flag which causes the path sent to the filter & xattr callbacks to match the original entry from the archive. This will be useful for e.g. `rpm-ostree` so that the unpacker can match the entry with rpmfi overrides.

For the name:id issues, we do not impose any restrictions in the ImportArchive API. Instead, client apps like `rpm-ostree` should do the checking themselves (e.g. in the filter callback) and report an error to the user if it does not yet support uid/gid != 0 (see also https://github.com/projectatomic/rpm-ostree/issues/49).

src/libostree/ostree-repo-libarchive.c
```diff
@@ -45,269 +47,715 @@ propagate_libarchive_error (GError      **error,
...
 
+static inline const char *
+path_relative (const char *src)
+{
+  /* One issue here is that some archives almost record the pathname as just a
+   * string and don't need to actually encode parent/child relationships in the
+   * archive. For us however, this will be important. So we do our best to deal
+   * with non-conventional paths (but note that we don't try to be sophisticated
+   * enough to handle e.g. ../ & ./ in the middle of normal dir comps, which
+   * would probably require a stack). Also important, we relativize the path. */
```
cgwalters: Note `ot_util_path_split_validate()` *does* intentionally explicitly error out if we find `..` as a path component.  See also `ot_util_filename_validate()`.

I think with the current logic it'd be possible to have `foo/../../../../bar`...and we wouldn't trip any assertions =/

We really need unit tests for this function.

I submitted https://github.com/ostreedev/ostree/pull/283

jlebon:
:arrow_up: Added a fixup to validate all paths using `ot_util_path_split_validate()`.

src/libostree/ostree-repo-libarchive.c
```diff
@@ -81,13 +82,23 @@ path_relative (const char *src)
   if (src[0] == '.' && src[1] == '\0')
     src += 1;
 
+  /* make sure that the final path is valid (no . or ..) */
+  if (!ot_util_path_split_validate (src, NULL, error))
```
cgwalters: A bit expensive to do it this way, we end up mallocing all the path components just to throw them away.  But eh...we can revisit this later.

cgwalters:
@cgwalters-bot r+ 37c09a3

PR #277, opened 2016-04-25
[merged] deploy: Start hardlinking kernel/initramfs on single-part again

cgwalters:
Commit
https://github.com/ostreedev/ostree/commit/1810de2b51680dbf35fdbd33d0a8d7e65eadc91f
lost an optimization where we would try hardlinks for the
kernel/initramfs in `/boot`.  This would be a noticeable space savings
on single-partition systems.


gatispaeglis:
Didn't test, but the change looks good to me.

cgwalters:

PR #278, opened 2016-04-25
[merged] test-parent: this test requires user xattrs

smcv:
---

Detected by Debian autobuilders, some of which place the entire build chroot on tmpfs.

### DIFF
--- a/tests/test-parent.sh
+++ b/tests/test-parent.sh
@@ -21,6 +21,8 @@ set -euo pipefail
 
 . $(dirname $0)/libtest.sh
 
+skip_without_user_xattrs
+
 echo '1..2'
 
 setup_test_repository "archive-z2"


cgwalters:
Right, because it uses `bare-user`.  This tmpfs issue is going to be a pain point, it really seems like offhand `mm/shmem.c` could basically treat xattrs as a special kind of file too.  An investigation for another day.

@cgwalters-bot r+ 87e17a0

smcv:
> This tmpfs issue is going to be a pain point

Indeed. I'm not delighted about the Debian build missing this much test coverage, but I don't think there's any way around it until tmpfs supports xattrs - we're keen to have all builds done in a chroot that is thrown away after use, and on a machine with enough RAM, doing that entirely in RAM is much faster than the alternatives (particularly since dpkg calls fsync a lot).

I believe the kernel developers' concern with xattrs on tmpfs is that they're a DoS risk, and would need to come out of the same size quota as normal file content to resolve that.

cgwalters:
I was thinking offhand that user xattrs could be represented just as hidden regular files effectively, so they'd be subject to the same quota limitations etc.  Anyways, maybe a weekend project later.

PR #279, opened 2016-04-28
[merged] build: Introduce --with-dracut=yesbutnoconf

cgwalters:
It's actually just easier for build systems (e.g. rpm-ostree)
using dracut to use `--add ostree` rather than indirecting
through the conf file.

This makes it easier for yum-managed systems to install ostree without
side effects.


gatispaeglis:
Makes sense, In my build scripts I use `--add ostree` directly as well.

cgwalters:

gatispaeglis:
@cgwalters-bot r+ 39293ed

PR #280, opened 2016-05-01
[merged] libglnx porting: Migrate from GSConsole

cgwalters:
To GLnxConsoleRef.  There were some subtleties here, for example we
used to reference `GSConsole` inside the progress changed function,
which at first seems like an ABI hazard, because e.g. rpm-ostree or
xdg-app could still be passing a `GSConsole` instance there.  Luckily,
it turns out to be compatible to just start calling libglnx here.

Another issue was that due to libglnx's use of the cleanup function,
we needed to ensure we always called `ostree_async_progress_finish()`
*before* the cleanup function was invoked.


giuseppe:
@cgwalters-bot r+ 17e97fd

PR #281, opened 2016-05-02
[merged] repo: Fix reading repos on read-only media

cgwalters:
I have a cache drive I often mount read-only, and the previous commit
for opening `tmp/cache` broke since `errno == EROFS`, not `EPERM`.

It turns out we already had the concept of a "writable" repo, so just
piggy back off that.


giuseppe:
@cgwalters-bot r+ e84e7ad

PR #282, opened 2016-05-03
[merged] libglnx porting: Migrate to glnx_stream_fstat()

cgwalters:
I ended up deciding to move this one into libglnx, seems like
something other libglnx-using software might want to do, even though
xdg-app doesn't right now.


jlebon:
@cgwalters-bot r+ abb57ae

PR #283, opened 2016-05-05
[merged] lib: Add more filename validations (no ., .. or /) in commit logic

cgwalters:
The filesystem commit code will never give us potentially hostile
filenames, and when importing from archives, we do some validation.

However, we should be extra paranoid and also add error messages in
the mtree in case someone tries to import a hostile
libarchive-supported format.


jlebon:
@cgwalters-bot r+ 6edb0be

PR #284, opened 2016-05-06
[merged] build: Use --std=gnu99

cgwalters:
It's actually very nice to be able to declare loop variables inside
the initializer.

Ideally we could turn off nested functions though.


jlebon:
@cgwalters-bot r+ 5d1f9e1

PR #285, opened 2016-05-07
trivial-httpd: Add __attribute__((format))

cgwalters:
I was briefly looking at building with clang mostly since it
detects unused variables with cleanup attributes, but then
I hit this fatal error.

It's a hard compile error with `-Wformat-nonliteral` since clang
doesn't know it's a format string.


jlebon:
Sanity checked that it compiles with no issues.
CI is down for the weekend, so I manually pushed it to master.

Issue #262, opened 2016-04-15
Refs that do not belong to ostree should not be automatically pruned


jlebon:
There are multiple apps now that also need to store data in the system repo. These apps might not want ostree to automatically prune their refs (see `ostree_admin_cleanup()`).

We need to make ostree smarter so that it only prunes the refs that belong to it.
See also PR #257.

cgwalters:
So...we ended up using a separate repo for this, but I have a suspicion this is still going to bite people down the line.

jlebon:
Yup, for this reason I had to add refs to the base layers for package layering (although to be fair, in that case the commit *does* "belong" to ostree, but we should probably still have some configurability here so we don't have to hack around it).

PR #286, opened 2016-05-09
[merged] OstreeRepo: add extensions directory

jlebon:
It's very useful for third-party applications to have someplace to store
their data guaranteed to be on the same device as the repo (thus
ensuring hardlinks) while still being shielded away from any of OSTree's
timely garbage collections.

We create a new "extensions/" subdirectory where apps can include
whatever they wish in "extensions/myapp/". This subdirectory is
completely unmanaged by ostree.

NB: I didn't bother making it a member of the OstreeRepo proper since we
don't really use it for anything else yet.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2210,6 +2210,13 @@ ostree_repo_create (OstreeRepo     *self,
   if (!g_file_make_directory (self->tmp_dir, cancellable, error))
     goto out;
 
+  {
+    g_autoptr(GFile) extensions_dir =
+      g_file_resolve_relative_path (self->repodir, "extensions");
+    if (!g_file_make_directory (extensions_dir, cancellable, error))
+      goto out;
+  }
+
   g_clear_object (&child);
   child = g_file_get_child (self->repodir, "refs");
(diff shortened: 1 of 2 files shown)


jlebon:
NB: This is the same as #257 (with an added test). Apparently, GitHub doesn't let you re-open PRs if you delete and then recreate the associated branch.

After #257, I had originally considered solving #262 as a better way to address my use case, but ended up going back to the extensions approach. In the end, it just seems cleaner to have the ability to be completely separated from the ostree repo for this.

cgwalters:
So related to this, I saw http://www.spinics.net/lists/linux-fsdevel/msg97709.html go by.  It would help us...but we would have to work on older kernels anyways.

It's worth reading at least to see a classic viro response :smile:

cgwalters:
Anyways, obviously pretty simple.  We don't actually have a good place in the docs for this right now, though we should create one at some point.

@cgwalters-bot r+ 41ac826

jlebon:
Thanks for the link. I had wondered why it wasn't supported in the first place, but never bothered to look more deeply into it. That cleared it up a bit (along with https://lkml.org/lkml/2007/12/18/323).

PR #287, opened 2016-05-09
[merged] sysroot: Plug some leaks

krnowak:
The deployments and booted deployment were never unreffed.

Also, remote (void) from the call to the unload function - it returns
nothing, so there is no need to fool the compiler.


cgwalters:
@cgwalters-bot r+ e4a81b1

How did you notice this?  Just code inspection?  A while ago I was trying to add some valgrind tests but it dropped off the radar...should look at updating those again.

krnowak:
Just code inspection.

PR #290, opened 2016-05-10
[merged] Allow passing options to summary fetching function

krnowak:
This adds a _with_options variant of the `ostree_repo_remote_fetch_summary` function, so I can pass a (familiar) `override-url` option. The rationale is that I'd like to fetch commits from several mirrors in LAN and then find out which of them actually have the latest commit. I guess that the only way to find this out is to fetch the summary and find the checksum of a commit for the specific ref and get the commit information from the repository. `ostree_repo_pull_with_options` does not seem to be able to tell me what are the checksums of the just fetched commits.


apidoc/ostree-sections.txt
```diff
@@ -246,6 +246,7 @@ ostree_repo_remote_get_gpg_verify
 ostree_repo_remote_get_gpg_verify_summary
 ostree_repo_remote_gpg_import
 ostree_repo_remote_fetch_summary
+ostree_repo_remote_fetch_summary_with_options
```
cgwalters: You'll also need to add this to the `libostree.sym` file, and in fact since this will be the first symbol since 2016.5, you'll need to introduce that version.

cgwalters: This should have failed `make check` right?

cgwalters:
@cgwalters-bot try

apidoc/ostree-sections.txt
krnowak: I didn't run it before, but yeah, it actually failed. Sorry about that.

Updated the file, hopefully I did that right - tests are passing for me now.

krnowak:
Updated.

cgwalters:
Ah,

```
    api docs (gtk-doc):                           no
```

Something to fix on the PR tester.

cgwalters:
@cgwalters-bot r+ 7950e61

cgwalters:
@cgwalters-bot retry

PR #291, opened 2016-05-11
[merged] libostree: Variant-related leak plugs and fixes

krnowak:
This tries to avoid leaking GVariantBuilders and GVariants in some
situations. The leaks were usually happening when some error occurred
or because of unclear variant ownership situation.

The former is mostly about making sure that g_variant_builder_clear is
called on builders that didn't finish their variant building process.

The latter is surely more work - sometimes the result of
g_variant_builder_end() should not be passed directly to a function,
but rather stored in a g_autoptr(GVariant), sunk and then passed to a
function. IMO, with an advent of g_autoptr, GVariants should be always
sunk instead of relying on some receiver function sinking it. This
would make an easy-to-follow policy of always sinking your
variants. Functions could then assume that the passed variant is
already sunk. These leaks are still happenning in commands, but they
are less harmful, since that code will not be used by some daemon as a
library routine.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -2384,7 +2384,7 @@ get_modified_xattrs (OstreeRepo                       *self,
 
       if (label)
         {
-          GVariantBuilder *builder;
+          g_autoptr(GVariantBuilder) builder = NULL;
 
           /* ret_xattrs may be NULL */
           builder = ot_util_variant_builder_from_variant (ret_xattrs,
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3510,13 +3510,16 @@ ostree_repo_delete_object (OstreeRepo           *self,
 
       if (tombstone_commits)
         {
-          g_autoptr(GVariantBuilder) builder = NULL;
-          builder = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
-          g_variant_builder_add (builder, "{sv}", "commit", g_variant_new_bytestring (sha256));
+          g_auto(GVariantBuilder) builder = {0,};
+          g_autoptr(GVariant) variant = NULL;
+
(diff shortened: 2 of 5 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -2384,7 +2384,7 @@ get_modified_xattrs (OstreeRepo                       *self,
 
       if (label)
         {
-          GVariantBuilder *builder;
+          g_autoptr(GVariantBuilder) builder;
 
           /* ret_xattrs may be NULL */
           builder = ot_util_variant_builder_from_variant (ret_xattrs,
```
cgwalters: Can you hoist up the initialization too?  I know it doesn't matter now, but I think autoptrs should *always* be initialized at a minimum as defense against code motion/refactoring.  I.e. if someone later adds code between the two that could throw we'd be freeing uninitialized memory.

src/libostree/ostree-repo.c
```diff
@@ -3510,13 +3510,16 @@ ostree_repo_delete_object (OstreeRepo           *self,
 
       if (tombstone_commits)
         {
-          g_autoptr(GVariantBuilder) builder = NULL;
-          builder = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
-          g_variant_builder_add (builder, "{sv}", "commit", g_variant_new_bytestring (sha256));
+          GVariantBuilder builder;
+          g_autoptr(GVariant) variant;
```
cgwalters: And same here, please init to `NULL`.

krnowak:
Updated.

cgwalters:
I don't mind the rebase, but in this case using `fixup!` would have been better to re-review.  (I know it may be painful to alternate workflow with other projects on github that don't use Homu or know about `fixup!`)

@cgwalters-bot r+ c6f2684

krnowak:
Whoops, I normally would remember that (I did that once recently for other ostree PR), but just switched from one project to another and my fingers typed the "git commit --amend --no-edit" themselves, I swear!

PR #292, opened 2016-05-11
[merged] Valgrind 2016 round1

cgwalters:


cgwalters:
@cgwalters-bot delegate=krnowak

tests/libtest.sh
```diff
@@ -77,11 +85,13 @@ if test -n "${OT_TESTS_DEBUG:-}"; then
 fi
 
 if test -n "${OT_TESTS_VALGRIND:-}"; then
-    CMD_PREFIX="env G_SLICE=always-malloc valgrind -q --leak-check=full --num-callers=30 --suppressions=${test_srcdir}/ostree-valgrind.supp"
+    CMD_PREFIX="env G_SLICE=always-malloc valgrind -q --leak-check=full --num-callers=30 --suppressions=${test_srcdir}/glib.supp --suppressions=${test_srcdir}/ostree.supp"
```
krnowak: Did you forget to add ostree.supp?

cgwalters: Oops, yes.  :heavy_plus_sign:

cgwalters: (It's empty right now because we're not yet testing possibly lost)

jlebon:
This is super awesome! :+1: 
I'm not very familiar with Valgrind so I'll let @krnowak do the final review. Would it be too early to incorporate this into the PR tester?

krnowak:
I tested in on test-pull-override-url.sh and it works like a charm. I only had to remove [this line](https://github.com/ostreedev/ostree/blob/68cbbec5777fe456636ab7b4d8c2b6d310844ceb/tests/libtest.sh#L255) (I have no idea what it is doing here) and do some `s/${CMD_PREFIX} ostree/${OSTREE}/`.

tests/libtest.sh
```diff
@@ -29,6 +29,14 @@ else
   test_builddir=$(dirname $0)
 fi
 
+# We need to get the real path in order to use valgrind
+if test -n "${OT_TESTS_UNINSTALLED:-}"; then
```
krnowak: You set this env var in `TESTS_ENVIRONMENT`, so maybe this `if` clause is not required, but instead set the `OSTREE` variable to that libtool output unconditionally? Unless we support overriding the `TESTS_ENVIRONMENT` variable.

Other idea could be creating a directory where we could put a symlink named `ostree` and point it to the `.libs/lt-ostree`, then prepend that directory to PATH env var. This would avoid the necessity of updating all the tests.

krnowak: @cgwalters: See this commit to get the idea. Seems to work.

https://github.com/kinvolk/ostree/commit/5700c7705f89119578a0e882343dca2e0bf58e15

cgwalters:
Very nice, thanks!  Incorporated your `fixup!`.

I don't want to merge this yet until I decide whether to workaround the syncfs bit for rhel7 or just only do these tests on a newer os, or backport valgrind, or something else.

cgwalters:
I am working on other CI/CD bits but will return to this PR hopefully next week.

cgwalters:
Ok, rebased :surfer: and added a `syncfs` workaround.  My vote is to land this now and iterate from master on leak fixes/improvements.

cgwalters:
@rh-atomic-bot delegate=krnowak

krnowak:
@rh-atomic-bot retry

krnowak:
@cgwalters Looks good in general, but you might want to update the commit message in the "tests: Modernize valgrind infrastructure" commit and do the `s/SUPRESS/SUPPRESS/g`.

Also, not sure if being a delegate gives me a power of telling the bot to retry the build.

cgwalters:
It's confusing but the github PR builder is distinct from Homu, or said another way the Homu state is distinct from the github status on the commit.  Your retry command would probably have worked but had no effect because the patch wasn't in the `r+` state.

The nice thing about Homu is you don't need to wait for the PR status - you can `r+` and it will be committed only if it passes the tests.

tests/libtest.sh
```diff
@@ -77,7 +77,7 @@ if test -n "${OT_TESTS_DEBUG:-}"; then
 fi
 
 if test -n "${OT_TESTS_VALGRIND:-}"; then
-    CMD_PREFIX="env G_SLICE=always-malloc valgrind -q --leak-check=full --num-callers=30 --suppressions=${test_srcdir}/ostree-valgrind.supp"
+    CMD_PREFIX="env G_SLICE=always-malloc OSTREE_SUPRESS_SYNCFS=1 valgrind -q --leak-check=full --num-callers=30 --suppressions=${test_srcdir}/glib.supp --suppressions=${test_srcdir}/ostree.supp"
```
krnowak: `s/SUPRESS/SUPPRESS/`

src/libostree/ostree-repo-commit.c
```diff
@@ -1473,10 +1473,16 @@ ostree_repo_commit_transaction (OstreeRepo                  *self,
       goto out;
     }
 
-  if (syncfs (self->tmp_dir_fd) < 0)
+  /* FIXME: Added since valgrind in el7 doesn't know about
+   * `syncfs`...we should delete this later.
+   */
+  if (g_getenv ("OSTREE_SUPRESS_SYNCFS") == NULL)
```
krnowak: `s/SUPRESS/SUPPRESS/`

cgwalters:
Argh, I thought you just meant the commit message, but now I see via the magic of Emacs `M-/` I propagated the typo everywhere :globe_with_meridians:   Fixup :arrow_up:

krnowak:
@rh-atomic-bot r+ 83d0842

PR #293, opened 2016-05-12
[merged] Makefile: Move to AM_DISTCHECK_CONFIGURE_FLAGS

cgwalters:
Since automake 1.11.2 it is recommended that packages
use AM_DISTCHECK_CONFIGURE_FLAGS instead of
DISTCHECK_CONFIGURE_FLAGS as the latter is intended
to be a user variable.

https://bugzilla.gnome.org/show_bug.cgi?id=766298


cgwalters:
@cgwalters-bot r+ efcdbb2

PR #294, opened 2016-05-12
[merged] Build fixes

gatispaeglis:


gatispaeglis:
The --enable-introspection configure switch also seems to need fixing. 

 ./configure --help | grep intro
  --enable-introspection=[no/auto/yes]
                          Enable introspection for this build

Trying to use it results in:
configure: WARNING: unrecognized options: --enable-introspection

Makefile-libostree.am
```diff
@@ -149,6 +148,7 @@ endif
 
 if USE_LIBSOUP
 libostree_1_la_SOURCES += \
+    $(ENUM_GENERATED) \
```
cgwalters: The problem is a bit messier than this...since other things may use generated enums.

And also, we don't want to generate enums for private headers in the first place, which is another bug.  I had a half-working patch to fix this...let me see if I can dig that up.

Are you actually planning to use `--with-soup=no`?

gatispaeglis: > Are you actually planning to use --with-soup=no? 

No, I discovered this issue only by accident, while trying to figure out why the build was failing for the issue fixed in the other patch.

cgwalters:
Ok, I have a more comprehensive variant of the second patch in https://github.com/ostreedev/ostree/pull/296

Can you review that and force push this PR to only have the first commit?

cgwalters:
@rh-atomic-bot r+ b3e39a0

PR #295, opened 2016-05-12
[merged] Add support for g_auto{,ptr}.

krnowak:
~~This bumps the dependency of GLib 2.44, which we already required because of g_steal_pointer.~~

~~I wasn't sure if the bumping should be done the way it is now, or just by bumping version of `gio-unix-2.0`.~~

This adds an opt-in support for g_autoptr for users with new enough glib.


configure.ac
```diff
@@ -45,7 +45,7 @@ PKG_PROG_PKG_CONFIG
 
 AM_PATH_GLIB_2_0
 
-GIO_DEPENDENCY="gio-unix-2.0 >= 2.40.0 libgsystem >= 2015.1"
+GIO_DEPENDENCY="glib-2.0 >= 2.44 gio-unix-2.0 >= 2.40.0 libgsystem >= 2015.1"
```
cgwalters: Also need to bump
`-DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_40 -DGLIB_VERSION_MAX_ALLOWED=GLIB_VERSION_2_40` in `Makefile.am`.

cgwalters: While you're there, can you add a comment saying to do so?

cgwalters:
Actually, we already auto-synthesize `g_steal_pointer` in libglnx: https://git.gnome.org/browse/libglnx/tree/glnx-backport-autoptr.h#n61

Now what makes this messy is we can't use libglnx in a public header.  But if this code used `GLIB_CHECK_VERSION` just like the polkit patch did, things work for consumers with new enough glib.

Talking with the Flatpak (the artist formerly known as xdg-app) maintainer, he'd like to be conservative with glib versions, so let's avoid bumping just yet.

krnowak:
That libglnx backports stuff confused me earlier. So, I removed the commit that bumped the glib-2.0 version and replaced it with the one adding comments about the MIN_REQUIRED and MAX_ALLOWED macros.

Also, I added some more ifdeffery to allow ostree to use the g_autoptr on its types (with an example) so maybe later it could replace glnx_unref_object with g_autoptr.

krnowak:
Ping?

krnowak:
Rebased to current master, fixed a typo in the commit message (copypasta from polkit commit message).

cgwalters:
Very nice work, thanks!  Looks good.

@rh-atomic-bot r+ 21e7248

PR #297, opened 2016-05-12
[merged] docs: fix swapped link syntax

miabbott:
A few links in the docs had the Markdown syntax swapped like:

(link title)[link url]

Just cleaned up those.  Verified via `mkdocs serve`


cgwalters:
Awesome, thanks!

@cgwalters-bot r+ bc3ab00

PR #304, opened 2016-05-20
[merged] tests: Fix build with old glib

alexlarsson:
test-checksum.c was using g_autofree without including libglnx.h which
has the backport for that.


cgwalters:
@rh-atomic-bot r+ 671be34

PR #305, opened 2016-05-23
[merged] Editor runs only when requested; otherwise, empty commits are performed,

Mathnerd314:
This is what I ended up with for https://bugzilla.gnome.org/show_bug.cgi?id=707069.

Sorry about the whitespace changes, I'm not sure how to avoid them.

With this patch, ostree commit will just do a commit, without an editor; you need to pass ``-e`` or ``--editor`` to run the editor.

### DIFF
--- a/src/ostree/ot-builtin-commit.c
+++ b/src/ostree/ot-builtin-commit.c
@@ -32,6 +32,7 @@
 
 static char *opt_subject;
 static char *opt_body;
+static gboolean opt_editor;
 static char *opt_parent;
 static gboolean opt_orphan;
 static char *opt_branch;
@@ -73,6 +74,7 @@ static GOptionEntry options[] = {
   { "parent", 0, 0, G_OPTION_ARG_STRING, &opt_parent, "Parent ref, or \"none\"", "REF" },
   { "subject", 's', 0, G_OPTION_ARG_STRING, &opt_subject, "One line subject", "SUBJECT" },
   { "body", 'm', 0, G_OPTION_ARG_STRING, &opt_body, "Full description", "BODY" },
+  { "editor", 'e', 0, G_OPTION_ARG_NONE, &opt_editor, "Use an editor to write the commit message", NULL },
   { "branch", 'b', 0, G_OPTION_ARG_STRING, &opt_branch, "Branch", "BRANCH" },
   { "orphan", 0, 0, G_OPTION_ARG_NONE, &opt_orphan, "Create a commit without writing a ref", NULL },
   { "tree", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_trees, "Overlay the given argument as a tree", "dir=PATH or tar=TARFILE or ref=COMMIT" },
@@ -425,7 +427,7 @@ ostree_builtin_commit (int argc, char **argv, GCancellable *cancellable, GError 
         goto out;
     }
 
-  if (!opt_subject && !opt_body)
+  if (opt_editor)
     {
       if (!commit_editor (repo, opt_branch, &opt_subject, &opt_body, cancellable, error))
         goto out;
--- a/src/ostree/ot-builtin-commit.c
+++ b/src/ostree/ot-builtin-commit.c
@@ -218,14 +218,18 @@ commit_editor (OstreeRepo     *repo,
   char **lines = NULL;
   int i;
 
-  *subject = NULL;
-  *body = NULL;
-
   input = g_strdup_printf ("\n"
       "# Please enter the commit message for your changes. The first line will\n"
       "# become the subject, and the remainder the body. Lines starting\n"
       "# with '#' will be ignored, and an empty message aborts the commit."
-      "%s%s\n", branch ? "\n#\n# Branch: " : "", branch ?: "");
+      "%s%s%s%s%s%s\n"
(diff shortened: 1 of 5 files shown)


giuseppe:
bot, test pull request

giuseppe:
thanks for working on it!  Whitespace changes are fine, especially if you are removing them :)
Please be sure to document the change in man/ostree-commit.xml as well, and add some tests as well.

While looking at the man file, I've noticed that some options are not documented, I have sent a PR here: https://github.com/ostreedev/ostree/pull/306/

cgwalters:
Your editor is likely configured to delete trailing whitespace in the whole file.  This is a bad idea in general I think, since it creates conflicts with any other patches that touch that file.

If we are going to strip whitespace I'd prefer the "one big commit" approach, or just doing it a little bit incrementally as code gets modified.

You may find https://github.com/cgwalters/homegit/blob/master/bin/git-un-diff-whitespace
useful.

cgwalters:
Commit message should match existing style.  See `git log` in this repo for examples as well as http://chris.beams.io/posts/git-commit/

I would personally prefer having two commits for this; suggestions for (git) :smile: commit messages:

``
commit: Add --editor command line argument

This allows users to conveniently override it from the command line.
```

```
commit: Allow omitting --subject, i.e. no log message

We originally required a message to match git,
but since OSTree is for binaries that are usually
made by automated systems, commit messages
don't always make sense.  Just relax
the restriction.
```

**Comment by Mathnerd314** [comment]
Updated to address comments.

I ended up removing the whitespace with
```
git show --ignore-space-at-eol -p HEAD > my.patch
git checkout HEAD~1
git apply my.patch
```
And then ``git commit -p`` split u the changes. The ``un-diff-whitespace`` script was helpful in design but didn't work (it claimed there were no changes to commit).

File: src/ostree/ot-builtin-commit.c [hunk 1]
```diff
@@ -218,14 +218,18 @@ commit_editor (OstreeRepo     *repo,
...
       "# with '#' will be ignored, and an empty message aborts the commit."
-      "%s%s\n", branch ? "\n#\n# Branch: " : "", branch ?: "");
+      "%s%s%s%s%s%s\n"
+              , branch ? "\n#\n# Branch: " : "", branch ? branch : ""
+              , *subject ? "\n" : "", *subject ? *subject : ""
+              , *body ? "\n" : "", *body ? *body : ""
+              );
+
+  *subject = NULL;
+  *body = NULL;
```
cgwalters: Why are we keeping these assignments at all now?

cgwalters:
Awesome, thanks!  These commits look exactly like what I was hoping!

src/ostree/ot-builtin-commit.c
Mathnerd314: ``*subject`` has to be null so that the loop [assigns](https://github.com/Mathnerd314/ostree/blob/bbf94d5a26e24286dc2f644e8f661f09f97911d6/src/ostree/ot-builtin-commit.c#L254-L256) the first non-comment line to the subject. ``*body`` is null mostly because it was easier; it could also have been set to null in an else condition [here](https://github.com/Mathnerd314/ostree/blob/bbf94d5a26e24286dc2f644e8f661f09f97911d6/src/ostree/ot-builtin-commit.c#L281).

Mathnerd314:
As far as the tests go, I was thinking I could copy Git's [editor tests](https://github.com/git/git/blob/c2c5f6b1e479f2c38e0e01345350620944e3527f/t/t7005-editor.sh), but Git is GPL and ostree is LGPL so I'm not sure that's the right approach.

cgwalters:
I'm not really concerned about testing interactive editing.  Not enough anyways to muddle things by having the GPL just for some tests just yet.

cgwalters:
I'd say let's just get a simple test in e.g. `basic-test.sh` that omits the `-s` argument, and squash that into 11106a4 ?

Mathnerd314:
Done.

tests/basic-test.sh
```diff
@@ -104,6 +104,12 @@ assert_streq $($OSTREE log test2-no-parent |grep '^commit' | wc -l) "1"
 echo "ok commit no parent"
 
 cd ${test_tmpdir}
+empty_rev=$($OSTREE commit -b test2-no-subject -s '' $test_tmpdir/checkout-test2-4)
+omitted_rev=$($OSTREE commit -b test2-no-subject-2 $test_tmpdir/checkout-test2-4)
+assert_streq $empty_rev $omitted_rev
```
cgwalters: That's going to be racy...ostree commits include timestamps, and I suspect the test is only passing for you when the commits happen to be made in the same second.  I'd pass `--timestamp="October 21 2015"` or so like some of the prune tests.

Mathnerd314: Good catch. Updated again. Although, it should be pretty rare for a small test commit to take a significant fraction of a second.

cgwalters:
Can you rebase/squash the test commit into 476445c to preserve bisectability?

Mathnerd314:
Updated again. 5 rebases in 3 days; probably not a record.

cgwalters:
Yeah I know I am picky about the commits, thanks for going through it :smile:

@rh-atomic-bot r+ 98289ee

PR #307, opened 2016-05-24
[merged] libotutil: Plug a leak

krnowak:
There's no need to allocate the variant builder on a heap, so allocate
it on the stack and avoid a memory leak at the same time.


cgwalters:
@rh-atomic-bot r+ 2fc38d7

PR #308, opened 2016-05-24
[merged] core: Add a function creating an archive-z2 content stream

krnowak:
It is quite similar to the already existing
ostree_raw_file_to_content_stream function, so I factored the common
part to a separate function. The difference is that we cannot report
the size of the resulting stream.

Can be useful for serving a "bare" repository as a faked "archive-z2"
repository.

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -459,7 +459,6 @@ header_and_input_to_stream (GVariant           *file_header,
  * @file_info: A file info
  * @xattrs: (allow-none): Optional extended attributes
  * @out_input: (out): Serialized object stream
- * @out_length: (out): Length of stream
  * @cancellable: Cancellable
  * @error: Error
  *
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -400,6 +400,99 @@ write_file_header_update_checksum (GOutputStream         *out,
   return ret;
 }
 
+/*
+ * header_and_input_to_stream:
+ * @file_header: A file header
+ * @input: File raw content stream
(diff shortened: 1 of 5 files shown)


giuseppe:
bot, test pull request

src/libostree/ostree-core.c
```diff
@@ -452,15 +444,92 @@ ostree_raw_file_to_content_stream (GInputStream       *input,
...
+  return TRUE;
+}
+
+/**
+ * ostree_raw_file_to_archive_z2_stream:
+ * @input: File raw content stream
+ * @file_info: A file info
+ * @xattrs: (allow-none): Optional extended attributes
+ * @out_input: (out): Serialized object stream
+ * @out_length: (out): Length of stream
```
cgwalters: This parameter doesn't exist.

cgwalters:
Looks good to me; only thing that has me hesitating is that we have no test coverage.  Which might argue for landing this as part of a larger commit to have e.g. `ostree trivial-httpd` support this.

cgwalters:
But I'd be okay merging this independently too.  A unit test that just did a roundtrip with this and `ostree_content_file_parse()` probably wouldn't be too hard.

krnowak:
Updated, added the test case.

src/libostree/ostree-core.c
krnowak: Uh, this is addressed, but apparently github moved your comment to the `ostree_raw_file_to_content_stream` function docs…

cgwalters:
Very nice!

@rh-atomic-bot r+ 8c09e24

jlebon:
@rh-atomic-bot retry

PR #309, opened 2016-05-25
[merged] Build on older versions of glib

alexlarsson:
Various places need to include libglnx.h for the autoptr backport
fallbacks to be there before ostree-autocleanups.h is included.

This fixes the build on centos7·

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -26,11 +26,11 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <gio/gfiledescriptorbased.h>
+#include "libglnx.h"
 #include "ostree.h"
 #include "ostree-core-private.h"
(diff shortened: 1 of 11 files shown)


giuseppe:
@rh-atomic-bot r+ ed476f9

jlebon:
Huh weird, not sure why it's requiring the CentOS CI status instead of Jenkins. @cgwalters, was homu switched over?

Actually looking at the settings for the master branch, it looks like the Jenkins status is no longer required to pass for master either.

cgwalters:
It's not instead of, currently it's both.

cgwalters:
Hm, looks like a possibly transient failure.

@rh-atomic-bot retry

cgwalters:
FWIW I've temporarily backed out the Homu change to gate on status-atomiccentosci.

PR #310, opened 2016-05-26
[merged] Add a function for veryfing any data with repo-specific GPG setup

krnowak:
With an addition of the extensions directory there is now a possibility to store some 3rd party data signed with the same keys as the summary file or commits. Would be great to make it easy to verify them too then.


giuseppe:
bot, test pull request

jlebon:
bot, retest this please

src/libostree/ostree-repo.c
```diff
@@ -4691,7 +4687,7 @@ _ostree_repo_gpg_verify_with_metadata (OstreeRepo          *self,
 
       if (!_ostree_gpg_verifier_add_keyring_dir (verifier, self->repodir,
                                                  cancellable, error))
-        goto out;
+        return NULL;
```
cgwalters: I'm not going to object to this, but so far in ostree we haven't been using "early returns" very much, mostly because when I was originally writing it the autocleanups weren't pervasive, so I ended up having some manual cleanup in `out:`.

So this is OK, and I'm in favor of slowly porting function by function.

src/libostree/ostree-repo.c
```diff
@@ -4911,6 +4911,96 @@ ostree_repo_verify_commit_ext (OstreeRepo    *self,
 }
 
 /**
+ * ostree_repo_verify_data:
```
cgwalters: Can we name this `ostree_repo_gpg_verify_data` or something for a bit of extra clarity?

src/libostree/ostree-repo.c
```diff
@@ -4911,6 +4911,96 @@ ostree_repo_verify_commit_ext (OstreeRepo    *self,
...
+  result = ostree_repo_verify_data_ext (self,
+                                        remote_name,
+                                        data,
+                                        signatures,
+                                        keyringdir,
+                                        extra_keyring,
+                                        cancellable,
+                                        error);
+
+  return ensure_valid_gpg_result (result, error);
```
cgwalters: Hmm.  What if we instead made this function public as `ostree_gpg_verify_result_require_one_signature()` or something?  Then we could have callers normally chain together the two of those and not have this wrapper function.

src/libostree/ostree-repo.c
krnowak: Sure. Will do.

src/libostree/ostree-repo.c
krnowak: Right, that's a good idea. So I'll drop the wrapper function and remove the `_ext` suffix from the wrapped function.

src/libostree/ostree-repo.c
krnowak: Yes, I was wondering how would you react to it. I changed it here, because the `out` label wasn't doing any cleanups at all.

krnowak:
Rebased and updated. I decided to use the `ostree_gpg_verify_result_require_valid_signature` name instead. I also put fixup commits inside the history to avoid conflicts (mostly in libostree.sym), but it might not look like that here, because github has its own idea about how to display the commits order.

cgwalters:
Looks great, thanks!

@rh-atomic-bot r+ 17cc3f3

PR #315, opened 2016-05-30
[merged] sysroot: Correct error handling path in previous libglnx port

cgwalters:
I happened to have the github page open with my commit after it was
merged, and then noticed a bug.

We still need to throw if we hit a non-`ENOENT` error.


giuseppe:
@rh-atomic-bot r+  	3112bf0

PR #318, opened 2016-06-01
[merged] repo: Display estimated time remaining when pulling

smspillaz:
Bug 765429 said that not having a time estimate can be annoying
when working with large pulls.

There isn't any complex time estimation logic here - we just take
the number of bytes remaining and do a linear projection of
the bytes per second rate at the current point in time.


giuseppe:
bot, test pull request

cgwalters:
Nice, thanks!

@rh-atomic-bot r+ 41aa034

PR #319, opened 2016-06-03
[merged] More glnx/gfile porting

cgwalters:
:train:

### DIFF
--- a/src/libostree/ostree-repo-static-delta-core.c
+++ b/src/libostree/ostree-repo-static-delta-core.c
@@ -159,7 +159,8 @@ ostree_repo_list_static_delta_names (OstreeRepo                  *self,
     }
 
   ret = TRUE;
-  gs_transfer_out_value (out_deltas, &ret_deltas);
+  if (out_deltas)
+    *out_deltas = g_steal_pointer (&ret_deltas);
  out:
   return ret;
 }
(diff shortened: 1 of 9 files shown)


src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -159,7 +159,8 @@ ostree_repo_list_static_delta_names (OstreeRepo                  *self,
     }
 
   ret = TRUE;
-  gs_transfer_out_value (out_deltas, &ret_deltas);
+  if (out_deltas)
+    *out_deltas = g_steal_pointer (&ret_deltas);
```
jlebon: Should we work towards phasing out `ot_transfer_out_value` as well in favour of this?

src/libotutil/ot-fs-utils.c
```diff
@@ -230,3 +230,24 @@ ot_openat_ignore_enoent (int dfd,
  out:
   return ret;
 }
+
+GBytes *
+ot_file_map_contents_bytes (int dfd,
```
jlebon: May  I suggest a `_at` suffix or maybe a s/map/mapat/ ?

src/libostree/ostree-repo-static-delta-core.c
cgwalters: Maybe though I'm less sure about that.  There are a lot of users.  It's tempting to introduce `glnx_transfer_out_value`.  Not sure.

cgwalters:
Sure, fixup :arrow_up:

cgwalters:
Must have missed a `git add` or something, pushed another fixup :arrow_up:

jlebon:
@rh-atomic-bot r+ 01fd095

PR #320, opened 2016-06-03
[merged] summary: Fix a typo

krnowak:
The --gpg-sign parameter will be used for signing the summary, not the
commit. This probably was copy-pasted from the ostree commit command.


cgwalters:
@rh-atomic-bot r+ 115e1d8

PR #321, opened 2016-06-06
[merged] tests: Use strict mode by default for C tests

cgwalters:
I was extending the C-based test suite to cover more, and noticed that
we weren't aborting if a command failed.  That made the tests somewhat
less useful, so let's fix it.


jlebon:
@rh-atomic-bot r+ b9e15dc

PR #322, opened 2016-06-07
[merged] tests/test-pull-c: New test that runs through the pull API via C

cgwalters:
We have had in the past issues with running `ostree_repo_pull()`
multiple times in the same process, embarassingly enough.  Nothing in
the current test suite covers this, so let's start.


jlebon:
This way of testing continues to entertain me.
@rh-atomic-bot r+ 01c3980

PR #325, opened 2016-06-07
[merged] Random fixes

jlebon:


cgwalters:
@rh-atomic-bot r+ 17b4e3b

configure.ac
```diff
@@ -38,8 +38,7 @@ GLIB_TESTS
 
 AC_CHECK_HEADER([sys/xattr.h],,[AC_MSG_ERROR([You must have sys/xattr.h from glibc])])
 
-AC_CHECK_PROGS(YACC, 'bison -y', :)
-AS_IF([test "$YACC" = :], [AC_MSG_ERROR([bison not found but required])])
+AS_IF([test "$YACC" != "bison -y"], [AC_MSG_ERROR([bison not found but required])])
```
giuseppe: I think this will break the case (you can try even with /usr/bin/bison as value):

```YACC="/path/to/bison -y" ./configure```

I wonder if we should use AC_PROG_YACC at all or simply use AC_CHECK_PROGS since we want to use bison and not another yacc

cgwalters: I have no issues hard depending on bison.

PR #327, opened 2016-06-09
[merged] Small patches

Mathnerd314:
These are some small and hopefully uncontroversial changes:
* Moving the pull code to repo-pull to make the footprint of libsoup smaller (the eventual goal is to run all networking / parsing code unprivileged in a fetcher process, similar to [OpenSSH](http://www.citi.umich.edu/u/provos/papers/privsep.pdf))
* Use 1 instead of 0 for file timestamps (borrowed from [Nix](https://github.com/NixOS/nix/commit/14bc3ce3d6d5745717fa19b8b43b5fdd117ff757) after I saw https://bugzilla.redhat.com/show_bug.cgi?id=1229160)
* Documentation changes / expansion

Of course, taken as a whole the diff is rather large.


giuseppe:
bot, test pull request

docs/manual/related-projects.md
```diff
@@ -147,37 +151,54 @@ See
...
+the auto-optimise-store option, but this is not on by default, and Nix does not
+guarantee that all of its files are in the content-addressed store. OSTree
+provides a git-like command line interface for browsing the content-addressed
+store, while Nix does not have this functionality.
+
+Nix used to use the immutable bit to prevent modifications to /nix/store, but
+now it uses a read-only bind mount. The bind mount can be privately remounted,
+allowing per-process privileged write access. OSTree still uses the immutable
+bit, which is awkward, since it has to be briefly removed to make a hard link
+when checking out.
```
cgwalters: OSTree only uses the immutable bit for `/` in deployment roots, so there's no real races here.

docs/manual/related-projects.md
```diff
@@ -147,37 +151,54 @@ See
...
+scale. OSTree supports using a build system that just rebuilds individual
+components (packages) as they change, without forcing a rebuild of their
+dependencies.
+
+Nix automatically detects runtime package dependencies by scanning content for
+hashes. OSTree only supports only system-level images, and doesn't do dependency
+management. Nix can store arbitrary files, using nix-store --add, but, more
+commonly, paths are added as the result of running a derivation file generated
+using the Nix language. OSTree is build-system agnostic; filesystem trees are
+committed using a simple C API, and this is the only way to commit files.
```
cgwalters:
```
...using a C API, or via a command line `ostree commit`, both
of which can store arbitrary data without semantics attached.
```

docs/manual/related-projects.md
```diff
@@ -147,37 +151,54 @@ See
...
+when checking out.
+
+NixOS supports switching OS images on-the-fly, by maintaining both booted-system
+and current-system roots. It is not clear how well this approach works. OSTree
+currently requries a reboot to switch images.
+
+Finally, NixOS supports installing user-specific packages from trusted
+repositories without requiring root, using a trusted daemon. OSTree could
+similarly support this with a setuid executable or trusted daemon, but there is
+still the problem of where to deploy user-specific programs.
```
cgwalters: http://flatpak.org/ has such a privileged helper, so let's link to it here.

docs/manual/related-projects.md
```diff
@@ -205,3 +226,39 @@ See
...
+## Git
+
+Although OSTree has been called "Git for Binaries", and the two share the idea
+of a hashed content store, the implementation details are quite different.
+OSTree supports extended attributes and uses SHA256 instead of Git's SHA1. It
+"checks out" files via hardlinks, rather than copying, and thus requires the
+files to be immutable. At the moment, OSTree commits may have at most one
+parent, as opposed to Git which allows an arbitrary number. Git uses a
+smart-delta protocol for updates, while OSTree uses (slow) HTTP requests or
+static deltas.
```
cgwalters: The `(slow)` here depends, I'd say "one http request per changed file", or "precomputed static deltas".

Remember for small incremental updates, base archive-z2 is really just fine.  Anyways I'd link to https://ostree.readthedocs.io/en/latest/manual/formats/ here rather than trying to elaborate too much.

docs/manual/related-projects.md
```diff
@@ -205,3 +226,39 @@ See
...
+
+## Conda
+
+Conda is an "OS-agnostic, system-level binary package manager and ecosystem";
```
cgwalters: Interesting, I hadn't really looked at this before.  Can you hyperlink it here too?

cgwalters:
The patches generally look good, but this PR is starting to stretch at least my definition of "small" :smile: 

I use https://github.com/cgwalters/homegit/blob/master/bin/git-hub-createpr to quickly create PRs from the command line, with a separate [topic branch](https://git-scm.com/book/en/v2/Git-Branching-Branching-Workflows#Topic-Branches) for each one.  It's not too much overhead once you get used to it I think.

cgwalters:
It looks like the pull and gtk-doc changes are going to be intertwined, but can you rebase out the mtime change into a separate PR at least?  And fix up the tests for it?

```
File 'cow-mtime' doesn't match regexp '0'
```

docs/manual/related-projects.md
Mathnerd314: Oh. Is there a reason for that? I don't think there's a race even if it was per-file, so long as OSTree retries removing the immutable bit if making the link fails with EPERM.

docs/manual/related-projects.md
Mathnerd314: The point here is that Nix's runtime references do have semantics - they affect pruning, garbage collection, and download behavior. There's a "related objects" field in OSTree commits which so far is unused, but could be adapted for this purpose. OSTree's lack of semantics here counts against it.

Storing arbitrary data is a separate task, I should probably mention the ``nix-support`` directory.

And I thought it was agreed that 'you should think of the OSTree command line as effectively a "demo" for the shared library' [1](https://github.com/ostreedev/ostree/blob/master/README.md). So I don't see why ``ostree commit`` is worth mentioning.

docs/manual/related-projects.md
cgwalters: Long term I think O_OBJECT is the right fix, and the bind mount was fine, so I didn't bother with the immutable bit.

docs/manual/related-projects.md
cgwalters: Hm, I guess that is unclear.  the `ostree admin` part is more of a command line demo, where as the `ostree` part (the "filesystem stuff") does generally call into the library but in some cases implements semantics too.  It is totally reasonable for a build system author to use the `ostree commit` commandline too, although advanced users should use the library there too.

docs/manual/related-projects.md
Mathnerd314: My impression from [the kernel discussion](http://www.spinics.net/lists/linux-fsdevel/msg75086.html) was that Theodore Ts'o didn't think O_OBJECT was worth implementing and it would never happen unless a new reason was found. In the kernel, implementation is not really the hard part; the hard part is convincing the maintainers. And waiting for them to change their mind is unlikely to work.

I'm not sure the bind mount alone is sufficient, since you might have a checkout in a non-standard location that's unprotected, but the immutable bit is probably sufficient and O_OBJECT is definitely overkill.

cgwalters: Let's not do an architecture discussion in the docs right now - can you fix the change to the docs to reflect the *current* reality (ostree uses the immutable bit only for `/`) and we can talk about future things elsewhere?

Mathnerd314: OK. But does OSTree use a bind mount currently? I'm not clear on that part.

cgwalters: Yes, just boot any ostree-using system and you'll see the `ro` bind mount over `/usr`.  It's true this doesn't protect against accidental damage from privileged programs traversing out to `/sysroot`, but then if you have malicious programs with uid 0 (and the SELinux perms to access the host fs) you're kind of screwed anyways.

docs/manual/related-projects.md
Mathnerd314: Oh, and I already mentioned the tools in the previous paragraph:
"OSTree provides a git-like command line interface for browsing the content-addressed
store, while Nix does not have this functionality."

Mathnerd314:
OK. I have split it up into 3 PR's. I used [hub](https://hub.github.com/), sorry but I still don't trust your dotfiles.

cgwalters:
@rh-atomic-bot r+ 915e230

PR #329, opened 2016-06-09
[merged] repo: Fix build on 32-bit systems

krnowak:
On 32-bit systems the modifier for printing 64bit values should be
%llu instead of %lu. Just use appriopriate macros that do the right
thing.


giuseppe:
bot, test pull request

giuseppe:
@rh-atomic-bot r+ c4f87a7

PR #330, opened 2016-06-09
[merged] repo: use OSTREE_TIMESTAMP (=1) for checked-out files

Mathnerd314:
This is split off from #327.

### DIFF
--- a/docs/manual/repo.md
+++ b/docs/manual/repo.md
@@ -47,6 +47,22 @@ payload sections.  The header contains uid, gid, mode, and symbolic
 link target (for symlinks), as well as extended attributes.  After the
 header, for regular files, the content follows.
 
+The OSTree data format intentionally does not contain timestamps. The reasoning
+is that data files may be downloaded at different times, and by different build
+systems, and so will have different timestamps but identical physical content.
+These files may be large, so most users would like them to be shared, both in
+the repository and between the repository and deployments.
+
+This could cause problems with programs that check if files are out-of-date by
+comparing timestamps. For Git, the logical choice is to not mess with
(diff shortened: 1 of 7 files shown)


cgwalters:
I have a worry that we'll break something that was relying on the previous zero value somehow...but I can't think of something concrete there.  So let's try it...

@rh-atomic-bot r+ 734f2d9

dbnicholson:
Would it be possible to have this reverted? mtime of 0 has been the semantics of ostree deployments from basically the beginning of the project. We (and others, see https://github.com/flatpak/flatpak/commit/b5204c908d3c43f329d146b41ae1cb6a7e6ffcfd) rely on that fact when generating trees.

In particular, this affects caches that use the mtime of the associated file or directory to determine if the cache is valid. By arbitrarily changing the mtime of the files to something else, all the caches we setup in the build are now invalidated. Preseeding caches is really important to the user experience as it avoids having the user wait while they're regenerated on first run.

Now, we could change our build infrastructure to preset all the mtimes to 1 to match this change, but what does that do for our existing users who are on an ostree that deploys with mtimes of 0? We could just revert this change at Endless (and the associated one in Flatpak), and that would be fine for our users. However, if we point non-Endless users to our apps, they'll have the great experience of waiting 10 seconds the first time they launch it while the fontconfig cache is rebuilt unnecessarily.

So, please consider reverting this and coordinating a revert with flatpak. IMO, this is not a backwards compatible change and the deploy semantics with mtime 0 were well established. I don't think working around GNU Tar's warning about files being "implausibly old" (which is absurd, 0 is a completely legitimate time) is worth this pain. Thanks.

PR #332, opened 2016-06-09
[merged] lib: Port some manual close() cleanups to be glnx_fd_close

cgwalters:
Just noticed this while reading some code, we didn't have many manual
`out: close()` bits left, this pushes us over the edge to autocleanup
almost everywhere.


krnowak:
In most of the cases you could also get rid of the now-empty `out` label in favor of explicit returns. But otherwise LGTM.

cgwalters:
Yeah, would like to do the conversion to explicit return but I think it should be a separate step.

@rh-atomic-bot delegate=krnowak

jlebon:
@rh-atomic-bot r+ 1898c61

PR #333, opened 2016-06-10
[merged] Fix various leaks

alexlarsson:


giuseppe:
@rh-atomic-bot r+ c86eb14

cgwalters:
LGTM too though just a reminder I like to see a bit of information in commit messages or PR headers for the motivation...could be something like "I was trying out the new valgrind in tests" or "I noticed gnome-software leaking in pulls".

It helps create cross-links so that if there happened to be a bug in these patches, someone else could say "oh let's look at what gnome-software is doing" etc.

alexlarsson:
Ftr i was trying to valgring the flatpak tests.

PR #334, opened 2016-06-10
[merged] tests: Exit valgrind tests if a leak is detected

alexlarsson:
This fails the test if OT_TESTS_VALGRIND is set and valgrind
detects a leak.

I noticed this was needed when I started looking at using valgrind in flatpak.

Note: This adds various failures when OT_TESTS_VALGRIND is set.


cgwalters:
Nice thanks!

@rh-atomic-bot r+ ee09cf9

PR #335, opened 2016-06-11
[merged] test-sysroot.js: set "strict mode" when sourcing libtest.sh

smcv:
As with the C tests in commit 08580118, this makes sure the test
fails as soon as something goes wrong.


giuseppe:
@rh-atomic-bot r+ 661c829

PR #336, opened 2016-06-12
[merged] Link libreaddir-rand to libdl

smcv:
It uses dlsym(). There's no point in being extra-portable here
because OSTree only targets Linux anyway.

---

I suspect this caused a test failure in Ubuntu which @alexlarsson worked around in his PPA; see <https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=826857>.


cgwalters:
@rh-atomic-bot r+ a91a6c5

PR #337, opened 2016-06-12
[merged] Get "make distcheck" to work

smcv:


cgwalters:
Okay, fixes for `make check` with `srcdir != builddir` is definitely good.  I can't guarantee I won't introduce new bugs in `make dist` - the canonical source representation should be considered to be git (+evtag), not a tarball.  The patches here look good though.

@rh-atomic-bot r+ 80d7f64

smcv:
> I can't guarantee I won't introduce new bugs in make dist - the canonical source representation should be considered to be git (+evtag), not a tarball.

The canonical source representation for Debian (and RPM, I think) must be a tarball of some sort, so I'm stuck with that, whatever you consider to be your canonical source. It can contain any directory tree of your choice, either a `git archive` variant that recurses through submodules or a `dist` archive or something else. At the moment I use `packaging/Makefile.dist-packaging` to get that tarball. If I switch to using `distcheck`, I'd use `distcheck` rather than `dist` (to ensure that the tarball does in fact work), and I'd probably also run a comparison with what's in git to check that nothing is missing.

cgwalters:
GContinuous is a git-native build system that understands git archives with submodules, and https://github.com/cgwalters/rpmdistro-gitoverlay injects this on top of RPM.

Basically it's not hard to do recursive git archive.

Issue #338, opened 2016-06-13
Issues with ref (tmp file cleanup, deleting folders, commit/create conflict with remote)


yuqi-zhang:
Currently there are 3 minor issues with refs:

1 ostree_repo_transaction_set_ref and ostree_repo_set_ref_immediate should clean up .tmp files on failure

Currently, if a ref "foo/bar" exists (foo is a folder), and a user tries to create foo, the API functions ostree_repo_transaction_set_ref and ostree_repo_set_ref_immediate in ostree-repo-commit.c would fail, and create a .tmp file (in the format of .tmpABCDE). ostree_repo_transaction_set_ref needs to abort the transaction or the .tmp file is not cleaned up. ostree_repo_set_ref_immediate will always create the .tmp file. This should be cleaned up on failures.

2 allow deleting of folders in ref

Currently, if a ref "foo/bar/baz" exists, and a user either "refs --delete foo" or "refs --delete foo/bar/baz", only the "baz" file gets deleted. All folders (both in local and remote) cannot be deleted unless the user goes into the refs folder and manually rm. The folders don't actually show up in "refs --list" and could cause much hassle. They should be deleted when a) the user specifies to delete or b) when the folder is empty of refs.

3 allow local refs with same name as remote refs/folders, for both commit and ref --create

Currently, if there is a remote ref "remote:foo/bar" or "remote:foo", and the user wants to create a local ref of the name "foo", either through "commit" or "refs --create" (new option), ostree errors saying that the folder "foo" exists, even though it is in the "remote" and not local. The user should be able to create a local ref called foo, which is not affected by the remote repo.

jlebon:
@yuqi-zhang I think all of these are fixed now, no?

yuqi-zhang:
I believe so, I'll close this now

PR #340, opened 2016-06-14
[merged] refs: add "ostree refs --create" and unit tests

yuqi-zhang:
Add the ability to create a ref (much like a git tag) for an
existing commit through "ostree refs EXISTING --create=NEWREF".
Previously the only way to create a new ref was by creating a new commit,
but refs --create allows multiple refs to point to the same commit.

The command will fail if:
 - None/more than one existing ref is specified
 - The specified EXISTING tag does not exist, or was not specified
 - The specified NEWREF already exists, or is the name of a folder

Add unit tests in tests-ref.sh to verify above functionality

### DIFF
--- a/src/ostree/ot-builtin-refs.c
+++ b/src/ostree/ot-builtin-refs.c
@@ -28,10 +28,12 @@
 
 static gboolean opt_delete;
 static gboolean opt_list;
+static char *opt_create;
 
 static GOptionEntry options[] = {
   { "delete", 0, 0, G_OPTION_ARG_NONE, &opt_delete, "Delete refs which match PREFIX, rather than listing them", NULL },
   { "list", 0, 0, G_OPTION_ARG_NONE, &opt_list, "Do not remove the prefix from the refs", NULL },
+  { "create", 0, 0, G_OPTION_ARG_STRING, &opt_create, "Create a new ref for an existing commit", "NEWREF" },
   { NULL }
 };
 
@@ -41,6 +43,8 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
   GHashTableIter hashiter;
   gpointer hashkey, hashvalue;
   gboolean ret = FALSE;
+  OstreeRepoTransactionStats stats;
+  g_autofree char *parent = NULL;
 
   if (opt_delete || opt_list)
     {
@@ -48,10 +52,16 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
                                       cancellable, error))
         goto out;
     }
+  else if(opt_create)
+    {
+      if (!ostree_repo_list_refs_ext (repo, NULL, &refs, OSTREE_REPO_LIST_REFS_EXT_NONE,
+                                      cancellable, error))
+        goto out;
+    }
   else if (!ostree_repo_list_refs (repo, refspec_prefix, &refs, cancellable, error))
     goto out;
 
-  if (!opt_delete)
+  if (!opt_delete && !opt_create)
     {
       g_hash_table_iter_init (&hashiter, refs);
       while (g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue))
@@ -60,7 +70,60 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
           g_print ("%s\n", ref);
         }
     }
+  else if(opt_create)
+    {
+      gboolean old_ref_exists = FALSE;
+      const char *checksum = NULL;
+
+      g_hash_table_iter_init (&hashiter, refs);
+      while (g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue))
+        {
+          g_autofree char *ref = NULL;
+          g_autofree char *remote = NULL;
+          const char *existing_ref= hashkey;
+
+          if (!ostree_parse_refspec (refspec_prefix, &remote, &ref, error))
+            goto out;
+
+          if(strcmp (opt_create, existing_ref) == 0)
+            {
+              g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                         "--create specified but ref %s already exists", existing_ref);
+              goto out;
+            }
+
+          if(strcmp(refspec_prefix, hashkey) == 0)
+            {
+              old_ref_exists = TRUE;
+              checksum = hashvalue;
+            }
+        }
+
+      if(old_ref_exists)
+        {
+          if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
+            goto out;
+
+          ostree_repo_transaction_set_ref (repo, NULL, opt_create, checksum);
+
+          if (!ostree_repo_resolve_rev (repo, opt_create, TRUE, &parent, error))
+            {
+              ostree_repo_abort_transaction (repo, cancellable, NULL);
+              goto out;
+            }
+
+          if (!ostree_repo_commit_transaction (repo, &stats, cancellable, error))
+            goto out;
+
+          ret = TRUE;
+          goto out;
+        }
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                       "Could not find existing ref %s", refspec_prefix);
+      goto out;
+    }
   else
+    /* delete */
     {
       g_hash_table_iter_init (&hashiter, refs);
       while (g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue))
@@ -97,6 +160,12 @@ ostree_builtin_refs (int argc, char **argv, GCancellable *cancellable, GError **
 
   if (argc >= 2)
     {
+      if(opt_create && argc > 2)
+        {
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                       "You must specify only 1 existing ref when creating a new ref");
+          goto out;
+        }
       for (i = 1; i < argc; i++)
         if (!do_ref (repo, argv[i], cancellable, error))
           goto out;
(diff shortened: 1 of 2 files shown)


src/ostree/ot-builtin-refs.c
```diff
@@ -41,17 +43,25 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
...
+  OstreeRepoTransactionStats stats;
+  g_autofree char *parent = NULL;
 
   if (opt_delete || opt_list)
     {
       if (!ostree_repo_list_refs_ext (repo, refspec_prefix, &refs, OSTREE_REPO_LIST_REFS_EXT_NONE,
                                       cancellable, error))
         goto out;
     }
+  else if(opt_create)
```
cgwalters: Space between `if` and `(` please.

cgwalters: (And everywhere else)

tests/test-refs.sh
```diff
@@ -65,4 +65,25 @@ ${CMD_PREFIX} ostree refs --repo=repo | wc -l > refscount.delete3
...
 assert_not_file_has_content reflist '^test-1$'
 
+#Add a few more commits, to test --create
+${CMD_PREFIX} ostree --repo=repo commit --branch=ctest -m ctest -s ctest tree
+${CMD_PREFIX} ostree --repo=repo commit --branch=foo/ctest -m ctest -s ctest tree
+
+${CMD_PREFIX} ostree --repo=repo refs | wc -l > refscount
+assert_file_has_content refscount "^5$"
+
+${CMD_PREFIX} ostree --repo=repo refs --create=ctest-new 2>/dev/null && (echo 1>&2 "refs --create unexpectedly succeeded without specifying an existing ref!"; exit 1)
```
cgwalters: This is OK, but elsewhere we use a lot the style:

```
if command-that-should-fail; then
  assert_not_reached "command-that-should-fail unexpectedly succeeded!"
fi
```

it's a bit more lines, but easier to understand I think.

src/ostree/ot-builtin-refs.c
```diff
@@ -60,7 +70,60 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
           g_print ("%s\n", ref);
         }
     }
+  else if(opt_create)
+    {
+      gboolean old_ref_exists = FALSE;
+      const char *checksum = NULL;
+
+      g_hash_table_iter_init (&hashiter, refs);
+      while (g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue))
```
cgwalters: Iterating over all of them to find one that exists is weird, can't we just use `ostree_repo_resolve_rev()` with `allow_noent = TRUE`?

src/ostree/ot-builtin-refs.c
```diff
@@ -60,7 +70,60 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
...
+        }
+
+      if(old_ref_exists)
+        {
+          if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
+            goto out;
+
+          ostree_repo_transaction_set_ref (repo, NULL, opt_create, checksum);
+
+          if (!ostree_repo_resolve_rev (repo, opt_create, TRUE, &parent, error))
```
cgwalters: Why is this code here?

src/ostree/ot-builtin-refs.c
```diff
@@ -60,7 +70,60 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
...
+          if(strcmp(refspec_prefix, hashkey) == 0)
+            {
+              old_ref_exists = TRUE;
+              checksum = hashvalue;
+            }
+        }
+
+      if(old_ref_exists)
+        {
+          if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
```
cgwalters: I'm not opposed to this but in practice I think it's saner to avoid the transactional ref bit - let's fix the bug that tempfiles stay around onerror first.

src/ostree/ot-builtin-refs.c
yuqi-zhang: Will try that, thanks

src/ostree/ot-builtin-refs.c
yuqi-zhang: This is because the code is checking both for the existing ref, as well as if the specified NEWREF is in use in the same iteration. Even if I find the existing ref from which I wanted to create a copy of, I still have to finish the iteration to see whether a duplicate of the NEWREF exists, which will cause the command to fail. Not a very clean way to do it I agree.

src/ostree/ot-builtin-refs.c
yuqi-zhang: Ok, should I fix that on the current origin/head, and then come back to this?

src/ostree/ot-builtin-refs.c
cgwalters: If I'm understanding correctly, I think we want to do another `ostree_repo_resolve_rev(..., opt_create, FALSE, ...)` above this, before we create a transaction.

cgwalters: Or perhaps more clearly, we can move this code up, change `allow_noent` from `TRUE -> FALSE`, and use it to determine `checksum`, right?

yuqi-zhang: Yes, I'll try changing to using resolve_rev to find conflicts

src/ostree/ot-builtin-refs.c
yuqi-zhang: Noted

tests/test-refs.sh
yuqi-zhang: I'll change to that instead, thanks!

src/ostree/ot-builtin-refs.c
yuqi-zhang: The reason I did that is because delete does it that way. Should I change that for delete as well?

src/ostree/ot-builtin-refs.c
cgwalters: It's OK as is too.

src/ostree/ot-builtin-refs.c
cgwalters: Delete needs to match *multiple* things so iterating over all of them makes sense, whereas here we only need to look at exactly two, right?

yuqi-zhang: Right, good point, I forgot about that, thanks

src/ostree/ot-builtin-refs.c
```diff
@@ -70,36 +70,25 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
...
+  else if (opt_create)
     {
-      gboolean old_ref_exists = FALSE;
-      const char *checksum = NULL;
+      g_autofree char *checksum = NULL;
+      g_autofree char *checksum_existing = NULL;
 
-      g_hash_table_iter_init (&hashiter, refs);
-      while (g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue))
+      if (ostree_repo_resolve_rev (repo, opt_create, TRUE, &checksum_existing, error))
```
cgwalters: This needs to be flattened, i.e.,

```
if (!ostree_repo_resolve_rev (...))
  goto out;
if (checksum_existing != NULL)
  set_error()
```

Otherwise if an unrelated error occurs (e.g. `EIO`, disk failing), we continue on.

cgwalters: More info at https://developer.gnome.org/glib/stable/glib-Error-Reporting.html

yuqi-zhang: Will fix that, sorry.

cgwalters:
LGTM!  However, our internal CI is busted which breaks merges :rage1: 

So let's wait a day and see if it comes back.

jlebon:
bot, retest this please

src/ostree/ot-builtin-refs.c
jlebon: Can't we remove this now that we already check it earlier above?

yuqi-zhang: Are you referencing the 3 resolve_rev's? In order, they check: if the NEW already exists (it shouldn't), if the EXISTING exists (it should), and whether the creation was successful (otherwise abort the transaction).

jlebon: Ahhh I see, I thought it was only `set_ref_immediate` that suffered from the tmp issue, but reading https://github.com/ostreedev/ostree/issues/338 says otherwise. Once that issue is fixed though, all this should probably be dropped in favour of a single `set_ref_immediate` call.

Also, shouldn't `allow_noent` be `FALSE` here?

yuqi-zhang: Actually closer inspection, the earlier check for whether NEW exists would also detect folder issues and error out, so this resolve_rev doesn't actually handle that case anymore, whoops.

So actually, it can be changed to set_ref_immediate right now I think, without the earlier issues. Let me just make sure of that.

src/ostree/ot-builtin-refs.c
```diff
@@ -43,7 +43,6 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
   GHashTableIter hashiter;
   gpointer hashkey, hashvalue;
   gboolean ret = FALSE;
-  OstreeRepoTransactionStats stats;
   g_autofree char *parent = NULL;
```
jlebon: Don't need this anymore, do we?

yuqi-zhang: Oh whoops, make didn't complain and I forgot about it, sorry, will remove

jlebon: Yeah, that's the cost of using `g_autofree`. It makes GCC think that it's used even when it's not (though technically it is... in the cleanup function).

jlebon:
Congrats on getting your first PR merged! :cake: 
@rh-atomic-bot r+ bdc8a74

Issue #339, opened 2016-06-14
remove hotfix from origin on upgrades


cgwalters:
Otherwise it shows up as incorrectly hotfixed.

cgwalters:
Also related to this...we need to consider whether to require some sort of flag to upgrade-and-drop-hotfixes.

`ostree admin upgrade --drop-hotfix` or something?

jlebon:
That sounds reasonable. It enforces the idea that hotfixes are only meant for dire situations.

jlebon:
To elaborate on what I meant: if you had to resort to a hotfix, then you probably want to be careful when doing an upgrade to ensure your hotfix won't be needed anymore.

cgwalters:
Downstream: https://bugzilla.redhat.com/show_bug.cgi?id=1346335

Mathnerd314:
Hotfix is a property of the deployment, not the remote. I'm not sure why OSTree stores unlocked=hotfix on the remote at all, since the deployment can just check for .usr-ovd-work.

As far as upgrades, could OSTree do a 3-way merge for hotfixes like with /etc?

cgwalters:
It's stored in the `.origin` file.  We certainly could investigate replaying the hotfix on top of upgrades.  The only ugly part would be extracting the data from overlayfs.

cgwalters:
This is really an rpm-ostree issue, submitted: https://github.com/projectatomic/rpm-ostree/pull/322

PR #343, opened 2016-06-15
sysroot/hotfix: Use the deployment as ground truth

Mathnerd314:
This is what I was thinking of for #339.

Untested at the moment.

### DIFF
--- a/src/ostree/ot-admin-builtin-upgrade.c
+++ b/src/ostree/ot-admin-builtin-upgrade.c
@@ -95,9 +95,6 @@ ot_admin_builtin_upgrade (int argc, char **argv, GCancellable *cancellable, GErr
                                                   "override-commit", NULL);
         }
 
-      /* Should we consider requiring --discard-hotfix here? */
-      origin_changed |= g_key_file_remove_key (origin, "origin", "unlocked", NULL);
-
       if (origin_changed)
         {
           /* XXX GCancellable parameter is not used. */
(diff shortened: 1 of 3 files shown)


src/ostree/ot-admin-builtin-upgrade.c
```diff
@@ -95,9 +95,6 @@ ot_admin_builtin_upgrade (int argc, char **argv, GCancellable *cancellable, GErr
                                                   "override-commit", NULL);
         }
 
-      /* Should we consider requiring --discard-hotfix here? */
-      origin_changed |= g_key_file_remove_key (origin, "origin", "unlocked", NULL);
```
cgwalters: I haven't thought about this a lot yet, but if we were to try this change, we'd still need to keep the code to delete it from the `.origin` file for backcompat.

jlebon:
bot, add author to whitelist

src/ostree/ot-admin-builtin-upgrade.c
Mathnerd314: The extra config options will be silently ignored, and preserve the option to downgrade to an older version of OSTree.  I'm not sure how deleting them helps backwards compatibility. 

I'm also not sure why backwards compatibility is a concern at all, at this point in time, but that's a separate issue.

cgwalters: Yeah, but in the situation where a user has a hotfixed system, then upgrades to this ostree version (discarding the hotfix), the `unlocked` bit in the origin will be carried forward forever.  Yeah it's just one line in the file but still, it's easy for us to keep this code (with a `/* legacy */` comment).

cgwalters: I think this PR would likely be OK with just this change.

cgwalters: Still happy to take this change with the tweak above.

PR #346, opened 2016-06-16
[merged] Add "archive" as an alias for "archive-z2"

cgwalters:
I find the "-z2" is really a long ago relic of the past when I changed
the format.  We no longer have anything to do with the original
`archive`, so let's start allowing people to type `--mode=archive`
which just looks saner.

At some point later I'll update the docs too, but it'll be an annoying
transition period as we'll have to say "On older OSTree, use -z2" etc.


giuseppe:
I like this! 

@rh-atomic-bot r+

PR #348, opened 2016-06-16
[merged] libglnx porting: delete temp files on failure of file creation

yuqi-zhang:
We noticed the temp files being left over in ostree when (mistakenly)
trying to create refs with names in use by folders. This fix removes
temp files created by glnx_file_replace_contents_at on failure.


cgwalters:
@rh-atomic-bot r+ 91d43ed

PR #349, opened 2016-06-17
[merged] repo: Avoid a possible divide by zero in progress

cgwalters:
I've actually hit the case in practice where they're the same, which
causes a divide by zero and hence a crash.  Computers are fast.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3870,7 +3870,10 @@ ostree_repo_pull_default_console_progress_changed (OstreeAsyncProgress *progress
       g_autofree char *formatted_bytes_sec = NULL;
       g_autofree char *formatted_est_time_remaining = NULL;
 
-      if ((current_time - start_time) < G_USEC_PER_SEC) // Ignore first second
+      /* Ignore the first second, or when we haven't transferred any
+       * data, since those could cause divide by zero below.
+       */
+      if ((current_time - start_time) < G_USEC_PER_SEC || bytes_transferred == 0)
         {
           formatted_bytes_sec = g_strdup ("-");
           formatted_est_time_remaining = g_strdup ("- ");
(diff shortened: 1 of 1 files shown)


Mathnerd314:
I had a different solution: https://github.com/Mathnerd314/ostree/commit/47d37b9e185667122b228351aead7704e37e3ece

cgwalters:
I actually didn't think about this hard enough, and just pushed an updated patch :arrow_up: 

It's not that they're equal, it's that the value is less than one.

cgwalters:
@rh-atomic-bot delegate=Mathnerd314

src/libostree/ostree-repo.c
```diff
@@ -3865,20 +3865,20 @@ ostree_repo_pull_default_console_progress_changed (OstreeAsyncProgress *progress
...
-      if (!bytes_sec) // Ignore first second
+      if ((current_time - start_time) < G_USEC_PER_SEC) // Ignore first second
         {
           formatted_bytes_sec = g_strdup ("-");
           formatted_est_time_remaining = g_strdup ("- ");
         }
       else
         {
+          guint64 bytes_sec = bytes_transferred / ((current_time - start_time) / G_USEC_PER_SEC);
+          guint64 est_time_remaining =  (total_delta_part_size - bytes_transferred) / bytes_sec;
```
Mathnerd314: You could still have bytes_sec be 0, if bytes_transferred is 0. So the conditional should be if (< first sec || bytes_transferred == 0)

Mathnerd314:
It might be easier to just convert everything over to floating-point, since that handles Infinity fairly naturally.

cgwalters:
Good point, not entirely sure it can happen right now offhand, but it's clearly right to check.  Pushed a fixup.

Mathnerd314:
Looks good to me. Let's see if this works...
@rh-atomic-bot r+ e2854ed

PR #353, opened 2016-06-18
More pull code changes

Mathnerd314:
As someone on #fedora-meeting told me, "Merge early, merge often."

The end goal is to get of synchronous fetches completely; this is a halfway step that fetches refs asynchronously but still does the summary on the main thread.

But there are also some global refactorings mixed in, namely OSTREE_SHA256_STRING_LEN and 
OSTREE_OBJECT_TYPE_COMMIT_META. Plus -Wempty-body and some confused logic in refs.


libglnx
```diff
@@ -1 +1 @@
-Subproject commit a6d08657aa868a0d5c7b5dd494e16f65415a148f
+Subproject commit afe3c3a86178c29ceaa3a5e46397ab2fa97202b3
```
jlebon: This is reverting #348.

Mathnerd314: Fixed in 3315b86.

Not sure why this happened in the first place... I guess git's submodule handling still sucks.

cgwalters:
Can you make separate PRs for:

 - SHA256 constants
 - refs logic error
 - pull: free fetch_data
 - Pull code aesthetics (first two commits)
 - Pull logic/API changes

Mathnerd314:
All the pull changes are interdependent; I don't think they can be meaningfully separated. But I split out the others.

Mathnerd314:
There's also #358.

cgwalters:
It looks like this fails with `git rebase -i --autosquash` locally at least due to ordering issues with the `fixup!`.  You may just need to rebase/squash and force push.

Mathnerd314:
OK. I squashed manually.

src/libostree/ostree-repo-pull.c
```diff
@@ -2133,6 +2195,65 @@ repo_remote_fetch_summary (OstreeRepo    *self,
...
+  FetchDeltaSuperBlockData *fetch_data = user_data;
+  OtPullData *pull_data = fetch_data->pull_data;
+  GError *local_error = NULL;
+  GError **error = &local_error;
+  GInputStream* input = _ostree_fetcher_stream_uri_finish (fetcher, result, error);
+  GBytes* delta_superblock_data = NULL;
+
+  if (input)
+    {
+      delta_superblock_data = g_input_stream_read_bytes ( input, OSTREE_MAX_METADATA_SIZE, pull_data->cancellable, error);
```
cgwalters: We need to catch and propagate this error.

src/libostree/ostree-repo-pull.c
```diff
@@ -2133,6 +2195,65 @@ repo_remote_fetch_summary (OstreeRepo    *self,
...
+    }
+  else
+    {
+      if (g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
```
cgwalters: `_ostree_fetcher_request_uri_to_membuf()` has an `allow_noent` flag which returns `NULL` but an unset error.  In some ways this can be cleaner than catching the `NOT_FOUND` error.

cgwalters:
I think the high level here is pretty good.  One thing I don't quite understand is why making the methods async makes it easier to do via a subprocess - we can communicate synchronously with a subprocess right?

That said it *is* ugly :japanese_ogre:  how we have a mix of sync/async code currently.  The new async code looks decent.

src/libostree/ostree-repo-pull.c
Mathnerd314: That error is a pointer to local_error, which in turn is propagated at the end of the function,, like in all of the async callbacks. Is that not sufficient?

I was more worried about using ``pull_data->cancellable``, but apparently that's thread-safe.

src/libostree/ostree-repo-pull.c
Mathnerd314: Right, I have a later sequence of commits which cleans this up. Should I just add those here or can it wait for another PR?

Mathnerd314:
> we can communicate synchronously with a subprocess right?

Yeah, I kind of realized that on Monday, after seeing https://github.com/ostreedev/ostree/commit/535033a4f01f772e696ad2201b34403203b98bea. What confused me was [this comment](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-pull.c#L2649-L2651), which AFAICT is completely misguided; there's no need to do anything to the fetcher to switch between synchronous and asynchronous requests.

cgwalters:
See the original commit: https://github.com/ostreedev/ostree/pull/157

Mathnerd314:
Ah, OK, there's this comment: https://bugzilla.gnome.org/show_bug.cgi?id=753336#c23
So the failure mode is not that the pull fails, but that it does all fetches sequentially. The question is how to test that...

src/libostree/ostree-repo-pull.c
cgwalters: It can wait I think.

src/libostree/ostree-repo-pull.c
cgwalters: What I mean is that if this sets the error, we shouldn't go on and call `delta_superblock_process`.  Although I see that ignores `NULL`...the rest of the async code uses `goto out` and I'd like to continue to match that style.

See for example `content_fetch_on_write_complete()`.  In this case we don't call any further processing and don't increment `pull_data->n_fetched[FETCH_CONTENT]`.

cgwalters: Actually in the error case here, we may then get into an overwriting-error situation `if (pull_data->require_static_deltas)` etc.

Say that the webserver returns 500 and we're requiring static deltas.

(Hm, we should probably have `ostree trivial-httpd` support a mode where it returns `500`s too).

cgwalters:
See also https://github.com/ostreedev/ostree/issues/373 which is caused by not doing fail fast on errors.

src/libostree/ostree-repo-pull.c
cgwalters: Are you going to be able to update this PR relatively soon for these error handling cleanups?

I'd like to at least land critical things like the memleak fixes and fix existing error handling soon without pulling in right away all of the pull code path reworks which are higher risk.

Mathnerd314: Ah, I didn't notice these comments before.

>What I mean is that if this sets the error, we shouldn't go on and call delta_superblock_process. Although I see that ignores NULL...the rest of the async code uses goto out and I'd like to continue to match that style.

This is what I have in my further branch: https://github.com/Mathnerd314/ostree/blob/4f84489/src/libostree/ostree-repo-pull.c#L1645-L1655. It's more reasonable than this code.

I'm starting to think it'll be easier to just keep all development in this PR, and split out the "useful" or less risky parts into separate PR's.

> I'd like to at least land critical things like the memleak fixes and fix existing error handling soon without pulling in right away all of the pull code path reworks which are higher risk.

I split out a subset of the "free fetch data" commit, it might help: #382.

cgwalters: Sorry about that, in retrospect I realize we need to be more formal about who owns what.  systemd uses a label for "reviewed, needs changes", I'll add one here too.

cgwalters:
Where are we on this?  I'd like to not leave your changes outstanding a long time.  Anything I can do to help?

Mathnerd314:
It seems from your comments that the general approach is good and all that's left is a few small nits, I've been ignoring those for my GSOC since I still have to finish the separate-process work. I'm almost done with that though; in a few days I will update the PR with all of that and it should be ready to re-review then.

PR #354, opened 2016-06-20
[merged] refs: allow overwrite of empty folders

yuqi-zhang:
We noticed that once a ref folder is created, there is no existing
command that can remove it. For example, once "foo/bar" is created,
even if the user deletes foo or all the refs under foo, the folder
will persist.

Now when the user attempts to create a ref "foo" either through commit
or refs --create, if a folder "foo" exists but is empty of refs, the
folder is removed and the new ref "foo" is created.

New unit tests in tests-ref.sh verify this functionality.

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -792,6 +792,25 @@ _ostree_repo_write_ref (OstreeRepo    *self,
     }
   else
     {
+      g_autoptr(GHashTable) refs = NULL;
+      GHashTableIter hashiter;
+      gpointer hashkey, hashvalue;
+
+      ostree_repo_list_refs (self, ref, &refs, cancellable, error);
+      g_hash_table_iter_init (&hashiter, refs);
+
+      while ((g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue)))
+        {
+          if (strcmp (ref, (char *)hashkey) != 0)
+            {
+              g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                           "Specified ref %s already exists as a folder", ref);
+              goto out;
+            }
+        }
+
+      glnx_shutil_rm_rf_at (dfd, ref, cancellable, error);
+
       if (!write_checksum_file_at (self, dfd, ref, rev, cancellable, error))
         goto out;
     }
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -107,7 +107,38 @@ write_checksum_file_at (OstreeRepo   *self,
 
     if (!_ostree_repo_file_replace_contents (self, dfd, name, (guint8*)bufnl, l + 1,
                                              cancellable, error))
-      goto out;
+      {
+        if (g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_IS_DIRECTORY))
+          {
+            g_autoptr(GHashTable) refs = NULL;
+            GHashTableIter hashiter;
+            gpointer hashkey, hashvalue;
+
+            g_clear_error (error);
+
+            if (!ostree_repo_list_refs (self, name, &refs, cancellable, error))
+              goto out;
+
+            g_hash_table_iter_init (&hashiter, refs);
+
+            while ((g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue)))
+              {
+                if (strcmp (name, (char *)hashkey) != 0)
+                  {
+                    g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                                 "Specified ref %s already exists as a folder", name);
+                    goto out;
+                  }
+              }
+
+            if (!glnx_shutil_rm_rf_at (dfd, name, cancellable, error))
+              goto out;
+
+            if (!_ostree_repo_file_replace_contents (self, dfd, name, (guint8*)bufnl, l + 1,
+                                                     cancellable, error))
+              goto out;
+          }
+      }
   }
 
   ret = TRUE;
@@ -792,25 +823,6 @@ _ostree_repo_write_ref (OstreeRepo    *self,
     }
   else
     {
-      g_autoptr(GHashTable) refs = NULL;
-      GHashTableIter hashiter;
-      gpointer hashkey, hashvalue;
-
-      ostree_repo_list_refs (self, ref, &refs, cancellable, error);
-      g_hash_table_iter_init (&hashiter, refs);
-
-      while ((g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue)))
-        {
-          if (strcmp (ref, (char *)hashkey) != 0)
-            {
-              g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                           "Specified ref %s already exists as a folder", ref);
-              goto out;
-            }
-        }
-
-      glnx_shutil_rm_rf_at (dfd, ref, cancellable, error);
-
       if (!write_checksum_file_at (self, dfd, ref, rev, cancellable, error))
         goto out;
     }
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -100,21 +100,22 @@ write_checksum_file_at (OstreeRepo   *self,
   {
     size_t l = strlen (sha256);
     char *bufnl = alloca (l + 2);
+    g_autoptr(GError) temp_error = NULL;
 
     memcpy (bufnl, sha256, l);
     bufnl[l] = '\n';
     bufnl[l+1] = '\0';
 
     if (!_ostree_repo_file_replace_contents (self, dfd, name, (guint8*)bufnl, l + 1,
-                                             cancellable, error))
+                                             cancellable, &temp_error))
       {
-        if (g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_IS_DIRECTORY))
+        if (g_error_matches (temp_error, G_IO_ERROR, G_IO_ERROR_IS_DIRECTORY))
           {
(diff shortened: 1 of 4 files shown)


src/ostree/ot-builtin-commit.c
```diff
@@ -428,7 +428,16 @@ ostree_builtin_commit (int argc, char **argv, GCancellable *cancellable, GError
...
     {
       if (!ostree_repo_resolve_rev (repo, opt_branch, TRUE, &parent, error))
-        goto out;
+        {
+          if (g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_IS_DIRECTORY))
+            {
+              /* A folder exists with the specified ref name,
+                 * which is handled by _ostree_repo_write_ref */
+              g_error_free (*error);
+              *error = NULL;
```
cgwalters: This is `g_clear_error (&error);`

src/libostree/ostree-repo-refs.c
```diff
@@ -792,6 +792,25 @@ _ostree_repo_write_ref (OstreeRepo    *self,
...
+
+      while ((g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue)))
+        {
+          if (strcmp (ref, (char *)hashkey) != 0)
+            {
+              g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                           "Specified ref %s already exists as a folder", ref);
+              goto out;
+            }
+        }
```
cgwalters: Enumerating everything when we're trying to write is a bit inefficient.  Can we instead catch `ENOTDIR` from the `write_checksum_file_at` below?

src/libostree/ostree-repo-refs.c
```diff
@@ -792,6 +792,25 @@ _ostree_repo_write_ref (OstreeRepo    *self,
...
+
+      glnx_shutil_rm_rf_at (dfd, ref, cancellable, error);
```
cgwalters: While it'd be a chunk of new code and possibly not worth it, I'd feel a bit more comfortable here if we instead had a function that was "remove empty directories only".

src/ostree/ot-builtin-commit.c
yuqi-zhang: Noted, will change, thanks!

src/libostree/ostree-repo-refs.c
yuqi-zhang: You mean catching EISDIR? Yes that would probably be better

src/libostree/ostree-repo-refs.c
```diff
@@ -792,6 +792,25 @@ _ostree_repo_write_ref (OstreeRepo    *self,
...
+
       if (!write_checksum_file_at (self, dfd, ref, rev, cancellable, error))
```
cgwalters: I took a quick sketch of this:

```
diff --git a/src/libostree/ostree-repo-refs.c b/src/libostree/ostree-repo-refs.c
index 29f0333..a836bc1 100644
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -59,6 +59,30 @@ add_ref_to_set (const char       *remote,
 }
 
 static gboolean
+recursive_clean_empty_dirs (int dfd, const char *origpath,
+                            GError **error)
+{
+  /* Generate a temp buffer */
+  char *path = strdupa (origpath);
+  char *p = strrchr (origpath, '/');
+  while (p)
+    {
+      if (unlinkat (dfd, path, AT_REMOVEDIR) != 0)
+        {
+          if (errno == ENOTDIR || errno == ENOTEMPTY)
+            return TRUE;
+        }
+      else
+        {
+          glnx_set_error_from_errno (error);
+          return FALSE;
+        }
+      *p = '\0';
+    }
+  return TRUE;
+}
+
+static gboolean
 write_checksum_file_at (OstreeRepo   *self,
                         int dfd,
                         const char *name,
@@ -88,13 +112,26 @@ write_checksum_file_at (OstreeRepo   *self,
 
   lastslash = strrchr (name, '/');
 
+ again:
+
   if (lastslash)
     {
       char *parent = strdupa (name);
+      g_autoptr(GError) temp_error = NULL;
       parent[lastslash - name] = '\0';
 
-      if (!glnx_shutil_mkdir_p_at (dfd, parent, 0777, cancellable, error))
-        goto out;
+      if (!glnx_shutil_mkdir_p_at (dfd, parent, 0777, cancellable, &temp_error))
+        {
+          if (g_error_matches (&temp_error, G_IO_ERROR, G_IO_ERROR_NOT_DIRECTORY))
+            {
+              g_clear_error (&temp_error);
+              if (!recursive_clean_empty_dirs (dfd, parent, cancellable, error))
+                goto out;
+              /* Note use of goto to loop back */
+              goto again;
+            }
+          goto out;
+        }
     }
 
   {
```

yuqi-zhang: Will take a look, thanks!

yuqi-zhang:
I also missed something in the tests, will take a look at that, sorry

src/libostree/ostree-repo-refs.c
yuqi-zhang: So after playing around with the code a bit, the problem is say, if I had a many levels of empty subfolders but wanted to create a new ref, unlinkat would catch the fact that there are existing subfolders (although empty) and return ENOTEMPTY. So if I wanted to delete all of them, I would have to look in each subfolder (recursively) and proceed to delete them one by one. That's why initially I iterated through the existing refs to check for the name instead and rm_rf all the folders if none of them are in use, and it would not require recursive check and deletion. Is it ok if I go back to that? Or would you rather I do recursive check and deletion?

cgwalters: I was trying to have the code above walk backwards, not saying the code works, but that's the idea.  It looks like it's missing at least a `p = strrchr (origpath, '/');` in the loop.

src/libostree/ostree-repo-refs.c
```diff
@@ -107,7 +107,38 @@ write_checksum_file_at (OstreeRepo   *self,
 
     if (!_ostree_repo_file_replace_contents (self, dfd, name, (guint8*)bufnl, l + 1,
                                              cancellable, error))
-      goto out;
+      {
+        if (g_error_matches (*error, G_IO_ERROR, G_IO_ERROR_IS_DIRECTORY))
```
cgwalters: The GError conventions allow the caller to pass `NULL`.  So anytime we want to do the equivalent of `try/catch`, we need to do:

```
g_autoptr(GError) temp_error = NULL;
if (!somefunction (..., &temp_error)) {
   if (g_error_matches (temp_error, ...)) {
     /* Handle error */
   } else {
     g_propagate_error (error, g_steal_pointer (&temp_error));
   }
}

**Comment by yuqi-zhang** [comment]
Changed the check to happen after _ostree_repo_file_replace_contents fails in write_checksum_file_at instead. Now the enumeration only happens when a user attempts to overwrite a ref folder, instead of everytime the user tries to write a ref.

File: src/libostree/ostree-repo-refs.c (hunk 5 above)
Comment (yuqi-zhang) [inline src/libostree/ostree-repo-refs.c]: Will look into

File: src/libostree/ostree-repo-refs.c [hunk 6]
```diff
@@ -107,7 +107,38 @@ write_checksum_file_at (OstreeRepo   *self,
...
+              goto out;
+
+            g_hash_table_iter_init (&hashiter, refs);
+
+            while ((g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue)))
+              {
+                if (strcmp (name, (char *)hashkey) != 0)
+                  {
+                    g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                                 "Specified ref %s already exists as a folder", name);
```
cgwalters: How about:  `_set_error (..., "Conflict: %s exists when writing %s", hashkey, name); ` ?

src/libostree/ostree-repo-refs.c
cgwalters: You can find other examples of this in the code via `git grep temp_error`

src/libostree/ostree-repo-refs.c
yuqi-zhang: Yeah, that would be more helpful. Will change, thanks!

src/libostree/ostree-repo-refs.c
```diff
@@ -107,7 +107,38 @@ write_checksum_file_at (OstreeRepo   *self,
...
+            g_clear_error (error);
+
+            if (!ostree_repo_list_refs (self, name, &refs, cancellable, error))
+              goto out;
+
+            g_hash_table_iter_init (&hashiter, refs);
+
+            while ((g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue)))
+              {
+                if (strcmp (name, (char *)hashkey) != 0)
```
cgwalters: Also...shouldn't this be: `if (g_str_has_prefix (name, hashkey))` or something like that?

yuqi-zhang: Hm, so how this works is, let's say our name is "foo/bar". When list_refs grabs all refs with prefix foo/bar, it will put into refs "foo/bar/..." AND "foo/bar" if it exists. It's not an error when we specify "foo/bar" to overwrite, but it would be if there is a "foo/bar/baz". That check basically says if it's not exactly "foo/bar", then we don't like it.

cgwalters: Oh I missed that we were only enumerating refs underneath name.  Yes, makes sense.  Actually that addresses some of my original concern as well - I thought at first we were enumerating everything.

cgwalters:
Thanks for persevering through the review here!  This one ended up being a bit tricky and hit on some subtle points.

@rh-atomic-bot r+ 2ab5f30

yuqi-zhang:
Thanks for guiding me along!

PR #355, opened 2016-06-20
fix bug in golang ostree bindings with incorrectly creating new repos

14rcole:
The function RepoNewOpen in contrib/golang/ostree.go would create a new repo struct, but wouldn't initialize the repo in the filesystem.  This change fixes that as well as moves the go files into contrib/golang/ostree so that the go command can resolve the packages correctly when they're being called by other projects.


cgwalters:
Hm...but we're not yet modifying the golang bindings to make use of this yet, right?  So the change is more like (with a commit message in the current style):

```repo: Add an ostree_repo_exists function

Many applications want to "idempotently" create a repository, i.e.
create if and only if it doesn't exist already yet.  Some of these
applications today check for the `objects` directory - we might
as well formalize this notion.
```

Right?

cgwalters:
Yeah, I'd prefer that.

PR #356, opened 2016-06-20
[merged] libostree.sym: Fix test-symbols

cgwalters:
The test isn't smart enough to ignore comments, so change the prefix.


jlebon:
@rh-atomic-bot r+ 76ab015

PR #357, opened 2016-06-21
add function to check if the repository exists on the filesystem

14rcole:
Add a function to check if a repository exists in the filesystem.  Can be used before attempting to create a repo with ostree_repo_create, which will fail if the repository already exists

### DIFF
--- a/src/libostree/libostree.sym
+++ b/src/libostree/libostree.sym
@@ -356,4 +356,5 @@ global:
         ostree_checksum_b64_to_bytes;
         ostree_checksum_b64_from_bytes;
         ostree_repo_checkout_at;
+        ostree_repo_is_created;
 } LIBOSTREE_2016.7;
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2094,6 +2094,26 @@ append_remotes_d (OstreeRepo          *self,
 }
 
 gboolean
+ostree_repo_is_created (OstreeRepo    *self,
+                        gboolean      *is_created,
+                        GCancellable  *cancellable,
+                        GError        **error)
+{
+  if (!ostree_repo_open (self, cancellable, error))
+    {
+      if (g_error_matches(error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        {
+          *is_created = FALSE;
+          return TRUE;
+        }
+      *is_created = FALSE;
+      return FALSE;
+    }
+  *is_created = TRUE;
+  return TRUE;
+}
+
+gboolean
 ostree_repo_open (OstreeRepo    *self,
                   GCancellable  *cancellable,
                   GError       **error)
--- a/tests/test-basic-c.c
+++ b/tests/test-basic-c.c
@@ -23,9 +23,12 @@
 #include <stdlib.h>
 #include <gio/gio.h>
 #include <string.h>
+#include <ostree.h>
 
 #include "libglnx.h"
 #include "libostreetest.h"
+
+#include <stdio.h>
 
 static void
 test_repo_is_not_system (gconstpointer data)
@@ -171,6 +174,37 @@ test_raw_file_to_archive_z2_stream (gconstpointer data)
   g_assert_cmpint (checks, >, 0);
 }
 
+static void
+test_repo_is_created (gconstpointer data)
+{
+  OstreeRepo *repo = OSTREE_REPO (data);
+  gboolean exists = 0;
+  gboolean created = 0;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GFile) path;
+  g_autoptr(GCancellable) cancellable;
+  int success;
+
+  /* Test a return value of TRUE on a repo that exists */
+  success = ostree_repo_is_created (repo, &exists, cancellable, &error);
+  g_assert (success);
+  g_assert(exists);
+  g_object_unref(repo);
+  /* Test a return value of FALSE on a repo that does not exist */
+  // Create an empty directory
+  mkdir ("not-a-repo", 0777);
+  // Create a repo struct associated with that directory
+  path = g_file_new_for_path ("not-a-repo");
+  repo = ostree_repo_new (path);
+  // Check to see if that repo has been initialized in the filesystem (it shouldn't have been)
+  error = NULL;
+  success = ostree_repo_is_created (repo, &exists, cancellable, &error);
+  if (error)
+    printf("%s\n", error->message);
+  g_assert (success);
+  g_assert (!exists);
+}
+
 int main (int argc, char **argv)
 {
   g_autoptr(GError) error = NULL;
@@ -184,6 +218,7 @@ int main (int argc, char **argv)
   
   g_test_add_data_func ("/repo-not-system", repo, test_repo_is_not_system);
   g_test_add_data_func ("/raw-file-to-archive-z2-stream", repo, test_raw_file_to_archive_z2_stream);
+  g_test_add_data_func ("/repo-created", repo, test_repo_is_created);
 
   return g_test_run();
  out:
--- a/src/libostree/ostree-repo.h
+++ b/src/libostree/ostree-repo.h
@@ -55,6 +55,12 @@ _OSTREE_PUBLIC
 OstreeRepo* ostree_repo_new_default (void);
 
 _OSTREE_PUBLIC
+gboolean      ostree_repo_is_created (OstreeRepo    *self,
+                                      gboolean      *does_exist,
+                                      GCancellable  *cancellable,
+                                      GError        **error);
+
+_OSTREE_PUBLIC
 gboolean      ostree_repo_open   (OstreeRepo     *self,
                                   GCancellable   *cancellable,
                                   GError        **error);


cgwalters:
Can you use my commit message from the other PR?  See `git log` in this repo for examples and http://chris.beams.io/posts/git-commit/ 

`git commit --amend` and `git push -f`

It'd also be nice to have a test (see `tests/test-repo-c.c`) but I can add one easily after.

14rcole:
I don't see the a file called `tests/test-repo-c.c` but I did amend the commit and push it up.  If you want me to add a test I can do that

cgwalters:
Er, I meant `tests/test-basic-c.c`.  Also can you remove the previous commit message?  It's now redundant I think.

14rcole:
I just did both of those. Let me know what you think

tests/test-basic-c.c
```diff
@@ -54,6 +55,25 @@ input_stream_to_bytes (GInputStream *input)
...
+  gboolean exists = 0;
+  GError *error = NULL;
+  /* Test a return value of TRUE on a repo that exists */
+  int success = ostree_repo_exists(repo, exists, error);
+  g_assert (success && exists);
+  /* Test a return value of FALSE on a repo that does not exist */
+  // Create an empty directory
+  mkdir ("not-a-repo",  S_IRUSR | S_IWUSR | S_IROTH | S_ IWOTH);
+  // Create a repo struct associated with that directory
+  GFile path = g_file_new_for_path ("not-a-repo");
```
cgwalters: You need a `*` here.  I'd also use `g_autoptr()` to free it, so:

`g_autoptr(GFile) path = g_file_new...`

tests/test-basic-c.c
```diff
@@ -54,6 +55,25 @@ input_stream_to_bytes (GInputStream *input)
...
+  repo = ostree_repo_new (path);
+  // Check to see if that repo has been initialized in the filesystem (it shouldn't have been)
+  success = ostree_repo_exists (repo, exists, error);
+  g_assert (success && exists);
```
cgwalters: And shouldn't this be `g_assert (success && !exists)` ?

Did you try: `make check TESTS=tests/test-basic-c` ?

14rcole:
I forgot to run `make check`

Once I made the g_autoptr change I the following error when I run `make check TESTS=tests/test-basic-c`

> In file included from /usr/lib64/glib-2.0/include/glibconfig.h:9:0,
                 from /usr/include/glib-2.0/glib/gtypes.h:32,
                 from /usr/include/glib-2.0/glib/galloca.h:32,
                 from /usr/include/glib-2.0/glib.h:30,
                 from /usr/include/glib-2.0/gobject/gbinding.h:28,
                 from /usr/include/glib-2.0/glib-object.h:23,
                 from /usr/include/glib-2.0/gio/gioenums.h:28,
                 from /usr/include/glib-2.0/gio/giotypes.h:28,
                 from /usr/include/glib-2.0/gio/gio.h:26,
                 from tests/test-basic-c.c:24:
/usr/include/glib-2.0/glib/gmacros.h:433:43: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
 #define _GLIB_CLEANUP(func)               __attribute__((cleanup(func)))


Any thoughts?

cgwalters:
You'll need to do:

```
g_autoptr(GFile) path = NULL;

 /* test code */

 path = g_file_new_for_path (...)
```

tests/test-basic-c.c
```diff
@@ -54,6 +55,25 @@ input_stream_to_bytes (GInputStream *input)
   return g_memory_output_stream_steal_as_bytes (G_MEMORY_OUTPUT_STREAM (mem_out_stream));
 }
 
+static void test_repo_exists (gconstpointer data)
+{
+  OstreeRepo *repo = OSTREE_REPO (data);
+  gboolean exists = 0;
+  GError *error = NULL;
+  /* Test a return value of TRUE on a repo that exists */
+  int success = ostree_repo_exists(repo, exists, error);
```
giuseppe: nitpick: space between 'ostree_repo_exists' and '('

14rcole:
The issues (and merge conflicts) should be fixed now

tests/test-basic-c.c
```diff
@@ -171,6 +172,30 @@ test_raw_file_to_archive_z2_stream (gconstpointer data)
...
+
+  /* Test a return value of FALSE on a repo that does not exist */
+  // Create an empty directory
+  mkdir ("not-a-repo", 777);
+  // Create a repo struct associated with that directory
+  path = g_file_new_for_path ("not-a-repo");
+  repo = ostree_repo_new (path);
+  // Check to see if that repo has been initialized in the filesystem (it shouldn't have been)
+  success = ostree_repo_exists (repo, &exists, error);
+  g_assert (success && !exists);
```
jlebon: I guess this is a bit of a nit since it's in a small test app anyway, but for good form: you need to unref the new repo here.

src/libostree/ostree-repo.c
```diff
@@ -2094,6 +2094,27 @@ append_remotes_d (OstreeRepo          *self,
 }
 
 gboolean
+ostree_repo_exists (OstreeRepo    *self,
+                    gboolean      *does_exist,
+                    GError        **error)
+{
+  struct stat stbuf;
+
+  if (lstat (gs_file_get_path_cached (self->objects_dir), &stbuf) < 0)
```
jlebon: Hmm, wouldn't it be a stronger test if we actually checked that it was a valid repo? E.g. call `ostree_repo_open()` on a separate reference and check for errors? Though I guess then, the function would be more appropriately called something like `ostree_repo_is_valid()`.

tests/test-basic-c.c
14rcole: How do I unref the repo?

src/libostree/ostree-repo.c
14rcole: I can add an ostree_repo_is_valid() function if you want or I can add that to this function, whichever you prefer

jlebon: I would rename this one to `is_valid()`. Actually, `is_created()` would be more appropriate I suppose since it's `create()` that does the work.

tests/test-basic-c.c
jlebon: You can do `g_object_unref (repo)`.

src/libostree/ostree-repo.c
cgwalters: `is_valid` feels a bit more `fsck` like.  I like `is_created()` a bit more.

src/libostree/libostree.sym
```diff
@@ -348,5 +348,6 @@ global:
 
 LIBOSTREE_2016.7 {
 global:
-        ostree_repo_resolve_rev_ext;
+        ostree_repo_resove_rev_ext;
```
cgwalters: Looks like we lost an `l`?

14rcole:
I messed that up when I resolved a merge conflict with that file.  I thought I had made a push to fix it though.  It should be fixed now

cgwalters:
Thanks for the patch!

@rh-atomic-bot r+ ae9b8e0

tests/test-basic-c.c
```diff
@@ -171,6 +172,30 @@ test_raw_file_to_archive_z2_stream (gconstpointer data)
...
+test_repo_is_created (gconstpointer data)
+{
+  OstreeRepo *repo = OSTREE_REPO (data);
+  gboolean exists = 0;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GFile) path;
+  int success
+
+  /* Test a return value of TRUE on a repo that exists */
+  success = ostree_repo_is_created (repo, &exists, error);
```
jlebon: I think you want to pass `&error` here. That's what's making the compiler cry :)

tests/test-basic-c.c
```diff
@@ -171,6 +172,30 @@ test_raw_file_to_archive_z2_stream (gconstpointer data)
...
+  g_assert (success && exists);
+  g_object_unref(repo);
+  /* Test a return value of FALSE on a repo that does not exist */
+  // Create an empty directory
+  mkdir ("not-a-repo", 777);
+  // Create a repo struct associated with that directory
+  path = g_file_new_for_path ("not-a-repo");
+  repo = ostree_repo_new (path);
+  // Check to see if that repo has been initialized in the filesystem (it shouldn't have been)
+  success = ostree_repo_is_created (repo, &exists, error);
```
jlebon: Here as well!

tests/test-basic-c.c
```diff
@@ -171,6 +172,30 @@ test_raw_file_to_archive_z2_stream (gconstpointer data)
   g_assert_cmpint (checks, >, 0);
 }
 
+static void
+test_repo_is_created (gconstpointer data)
+{
+  OstreeRepo *repo = OSTREE_REPO (data);
+  gboolean exists = 0;
+  g_autoptr(GError) error = NULL;
+  g_autoptr(GFile) path;
```
jlebon: Initialize to `NULL` ?

src/libostree/ostree-repo.c
```diff
@@ -2094,6 +2094,27 @@ append_remotes_d (OstreeRepo          *self,
 }
 
 gboolean
+ostree_repo_is_created (OstreeRepo    *self,
+                        gboolean      *does_exist,
+                        GError        **error)
+{
+  struct stat stbuf;
+
+  if (lstat (gs_file_get_path_cached (self->objects_dir), &stbuf) < 0)
```
jlebon: What I initially meant for this was to instead use `ostree_repo_open` on another reference to test if it's created, but we can add that later as well.

14rcole: So just to clarify, you meant to keep ostree_repo_exists() as is and add another function called ostree_repo_is_created() that checks if the repo exists and then checks to make sure it can be opened?

jlebon: No, I meant only have `ostree_repo_is_created()` as the new API, but inside, instead of stat'ing the objects dir, you would call `ostree_repo_open()`, since it pretty much goes through the whole process of verifying that the repo is created as it opens it. But of course, you don't want to change the `OstreeRepo` object that was passed to you, so you'd have to create a new reference from that one. But again, I don't mind getting something like that in later as well.

cgwalters: Hmm...we could change them to share more code over time, but as is now it's only ~10 lines in this function, I'm okay with it

cgwalters:
Hmm...looks like you have merges from master.  Can you `git rebase -i origin/master` and squash everything down into one commit?

src/libostree/ostree-repo.c
```diff
@@ -2094,6 +2094,26 @@ append_remotes_d (OstreeRepo          *self,
...
 
 gboolean
+ostree_repo_is_created (OstreeRepo    *self,
+                        gboolean      *is_created,
+                        GCancellable  *cancellable,
+                        GError        **error)
+{
+  if (!ostree_repo_open (self, cancellable, error))
+    {
+      if (errno == ENOENT)
```
cgwalters: So...this gets a bit confusing.  But we can't rely on the value of `errno` after `ostree_repo_open()`.  We've already translated from `errno -> GError`, so we'd need to look at that.

src/libostree/libostree.sym
```diff
@@ -344,6 +344,7 @@ global:
 LIBOSTREE_2016.7 {
 global:
         ostree_repo_resolve_rev_ext;
+        ostree_repo_is_created;
```
cgwalters: This needs to go in the `2016.8` section now.  Sorry for the churn here.

14rcole:
Those should be fixed not.  Let me know what you think

cgwalters:
diff --git a/src/libostree/ostree-repo.c b/src/libostree/ostree-repo.c
index fe7bbcf..80c254e 100644
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2093,24 +2093,41 @@ append_remotes_d (OstreeRepo          *self,
   return ret;
 }
 
+/**
+ * ostree_repo_try_open:
+ * @self: Repo
+ * @is_created: (out): Whether or not the repository exists
+ * @error: Error
+ *
+ * Determine whether or not an OSTree repository exists already at the
+ * provided location.  Any filesystem or other errors other than
+ * `ENOENT` will result in @error being set.
+ */
 gboolean
 ostree_repo_is_created (OstreeRepo    *self,
                         gboolean      *is_created,
-                        GCancellable  *cancellable,
                         GError        **error)
 {
-  if (!ostree_repo_open (self, cancellable, error))
+
+  if (!ostree_repo_open (self, cancellable, &local_error))
     {
-      if (g_error_matches(error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+      if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
         {
           *is_created = FALSE;
           return TRUE;
         }
-      *is_created = FALSE;
-      return FALSE;
+      else
+        {
+          return FALSE;
+        }
+    }
+  else
+    {
+      *is_created = TRUE;
+      return TRUE;
     }
-  *is_created = TRUE;
-  return TRUE;
 }
 
 gboolean
diff --git a/tests/test-basic-c.c b/tests/test-basic-c.c
index eb8b742..1bea62f 100644
--- a/tests/test-basic-c.c
+++ b/tests/test-basic-c.c
@@ -178,31 +178,27 @@ static void
 test_repo_is_created (gconstpointer data)
 {
   OstreeRepo *repo = OSTREE_REPO (data);
+  g_autoptr(OstreeRepo) nonexistent_repo = NULL;
   gboolean exists = 0;
-  gboolean created = 0;
   g_autoptr(GError) error = NULL;
   g_autoptr(GFile) path;
-  g_autoptr(GCancellable) cancellable;
   int success;
 
   /* Test a return value of TRUE on a repo that exists */
-  success = ostree_repo_is_created (repo, &exists, cancellable, &error);
-  g_assert (success);
-  g_assert(exists);
-  g_object_unref(repo);
+  success = ostree_repo_is_created (repo, &exists, NULL, &error);
+  g_assert_no_error (error);
+  g_assert (success && exists);
+
   /* Test a return value of FALSE on a repo that does not exist */
   // Create an empty directory
   mkdir ("not-a-repo", 0777);
   // Create a repo struct associated with that directory
   path = g_file_new_for_path ("not-a-repo");
-  repo = ostree_repo_new (path);
+  nonexistent_repo = ostree_repo_new (path);
   // Check to see if that repo has been initialized in the filesystem (it shouldn't have been)
-  error = NULL;
-  success = ostree_repo_is_created (repo, &exists, cancellable, &error);
-  if (error)
-    printf("%s\n", error->message);
-  g_assert (success);
-  g_assert (!exists);
+  success = ostree_repo_is_created (repo, &exists, NULL, &error);
+  g_assert_no_error (error);
+  g_assert (success && !exists);
 }
 
 int main (int argc, char **argv)

cgwalters:
Talked with Ryan in person, the https://github.com/ostreedev/ostree/pull/422 replaces this.

PR #358, opened 2016-06-22
[merged] refs: Fix a logic error

Mathnerd314:
I encountered the "Opening remotes/ dir" error while I was working with some broken pull code, and this fixed it. But it's an independent change.

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -772,8 +772,10 @@ _ostree_repo_write_ref (OstreeRepo    *self,
             goto out;
         }
 
-      if (!glnx_opendirat (refs_remotes_dfd, remote, TRUE, &dfd, error))
+      dfd = glnx_opendirat_with_errno (refs_remotes_dfd, remote, TRUE);
+      if (dfd < 0 && (errno != ENOENT || rev != NULL))
         {
+          glnx_set_error_from_errno (error);
           g_prefix_error (error, "Opening remotes/ dir %s: ", remote);
           goto out;
         }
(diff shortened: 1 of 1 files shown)


cgwalters:
What was the broken pull code?  I believe there's a bug here...I'm trying to better understand what it is though.
Is there a simple test case for this?

So previously we threw if the remote dir didn't exist.  Now if `errno == ENOENT && rev == NULL` (i.e. we're trying to delete a ref)...okay, I see now, we ignore trying to delete a ref for a nonexistent remote.

Ah, but confusingly...the `refs` builtin papers over this because it enumerates existing ones when doing a match, so we can't actually test this via the command line.

cgwalters:
@rh-atomic-bot r+ 6e66b8e

cgwalters:
For posterity the outstanding diff I had trying to write a test:

```
diff --git a/tests/basic-test.sh b/tests/basic-test.sh
index 003df89..f3bafae 100755
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -19,7 +19,7 @@
 
 set -euo pipefail
 
-echo "1..57"
+echo "1..58"
 
 $OSTREE checkout test2 checkout-test2
 echo "ok checkout"
@@ -265,6 +265,14 @@ ${CMD_PREFIX} ostree --repo=repo3 rev-parse aremote/test2
 echo "ok pull-local with --remote arg"
 
 cd ${test_tmpdir}
+${CMD_PREFIX} ostree --repo=repo3 rev-parse aremote:test2
+${CMD_PREFIX} ostree --repo=repo3 refs --delete aremote:test2
+${CMD_PREFIX} ostree --repo=repo3 refs --delete aremote:test2
+rm repo3/refs/remotes/aremote -rf
+${CMD_PREFIX} ostree --repo=repo3 refs --delete aremote:test2
+echo "ok refs --delete for a remote"
+
+cd ${test_tmpdir}
 ${CMD_PREFIX} ostree --repo=repo3 prune
 find repo3/objects -name '*.commit' > objlist-before-prune
 rm repo3/refs/heads/* repo3/refs/remotes/* -rf
```

Mathnerd314:
I guess there could be a C API test that imported ostree-repo-private.h. But that seems like overkill.

Mathnerd314:
For reference, the issue with the pull code was that I didn't increment the outstanding requests, so it finished early (before fetching the ref's commit) and thus tried to write a non-existent ref.

PR #359, opened 2016-06-22
[merged] Core changes

Mathnerd314:
These are the SHA256 changes, the configure option, and the COMMIT_META object type from #353.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -17,7 +17,7 @@ AC_PROG_YACC
 
 changequote(,)dnl
 if test "x$GCC" = "xyes"; then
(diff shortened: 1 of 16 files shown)


libglnx
```diff
@@ -1 +1 @@
-Subproject commit a6d08657aa868a0d5c7b5dd494e16f65415a148f
+Subproject commit afe3c3a86178c29ceaa3a5e46397ab2fa97202b3
```
giuseppe: does this change belong to this patch?

Mathnerd314: No, but there's a fixup commit so they cancel out.

cgwalters:
@rh-atomic-bot r+ e64813b

PR #360, opened 2016-06-22
[merged] core: Fix wrong return value docs

krnowak:
ostree_commit_get_parent() returns a string form of a checksum, not a
binary form.


cgwalters:
@rh-atomic-bot r+ 54dae2a

cgwalters:
bot, add author to whitelist

PR #362, opened 2016-06-23
[merged] pull: Correctly handle repo->parent_repo when applying static deltas

alexlarsson:
In flatpak i was using a parent repo, and it failed to update
with ENOENT when dispatching an set-read-source opcode, because the
object it referenced was in the parent repo.

This fixes that by making _ostree_repo_read_bare_fd look
at parent_repo.


cgwalters:
Makes sense.

@rh-atomic-bot r+ 559b612

PR #363, opened 2016-06-23
[merged] refs: resolve conflict between local/remote repos

yuqi-zhang:
Add the functionality to use the same name for refs in local and remote
repos. This helps users keep track of local refs of remote origin, much
like local and remote git branches.

Previously, when a local ref is specified, resolve_refspec would fall
back to searching through remote repos if the ref is not found locally.
This function now takes an extra flag to specify whether it should
search through remote repos. Additionally, ostree_repo_resove_rev_ext
was added to call resolve_refspec with fallback_remote being false, so
refs --create would no longer complain when trying to create a local
ref of the same name as a remote one.

Fix remote repo parsing not being handled correctly on refs --create.

### DIFF
--- a/src/libostree/libostree.sym
+++ b/src/libostree/libostree.sym
@@ -214,7 +214,6 @@ global:
         ostree_repo_remote_list;
         ostree_repo_remote_list_refs;
         ostree_repo_resolve_rev;
-        ostree_repo_resolve_rev_ext;
         ostree_repo_scan_hardlinks;
         ostree_repo_set_disable_fsync;
         ostree_repo_set_ref_immediate;
@@ -347,10 +346,7 @@ global:
  *                         NOTE NOTE NOTE
  */
 
-/* Uncomment this when adding a new symbol */
-/*
 LIBOSTREE_2016.7 {
 global:
-        someostree_some_new_symbol;
+        ostree_repo_resolve_rev_ext;
 } LIBOSTREE_2016.6;
-*/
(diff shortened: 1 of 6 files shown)


src/libostree/libostree.sym
```diff
@@ -214,6 +214,7 @@ global:
         ostree_repo_remote_list;
         ostree_repo_remote_list_refs;
         ostree_repo_resolve_rev;
+        ostree_repo_resolve_rev_ext;
```
cgwalters: This needs to go in the new 2016.7 section.

jlebon:
A test would be nice as well.

yuqi-zhang:
Sorry, will add test as well

src/libostree/ostree-repo-refs.c
```diff
@@ -216,8 +217,8 @@ resolve_refspec_fallback (OstreeRepo     *self,
 
   if (self->parent_repo)
     {
-      if (!resolve_refspec (self->parent_repo, remote, ref,
-                            allow_noent, &ret_rev, error))
+      if (!resolve_refspec (self->parent_repo, remote, ref, allow_noent,
+                            TRUE, &ret_rev, error))
```
jlebon: Shouldn't this lookup also inherit `fallback_remote`? Otherwise a ref can be found in a parent repo's remotes dir even though you originally specified `FALSE`.

yuqi-zhang: Hm, yeah I think I let that slip, let me change that and make sure that it doesn't affect anything else, thanks for pointing it out!

jlebon:
Good work. :+1: 
@rh-atomic-bot r+ 4d5ee1c

PR #366, opened 2016-06-26
[merged] entry_pathname_test_helper: these tests need extended attributes

smcv:
This is another test that fails when the entire filesystem is tmpfs, as is done on some Debian autobuilders.


cgwalters:
@rh-atomic-bot r+ 5bd1f44

PR #367, opened 2016-06-26
[merged] Fix tests' use of libtool

smcv:
libtoolize creates a version of libtool for the right architecture in $(top_builddir), which is guaranteed to be present, and is guaranteed to match what we are compiling (even during cross-compilation).

The libtool in $PATH might be for a different architecture, or even not exist at all.

Also stop the tests from proceeding (and later using the wrong ostree binary, or failing if ostree is not already installed) if a preparation step fails.


cgwalters:
@rh-atomic-bot r+ 7ab6199

PR #369, opened 2016-06-27
[merged] libglnx porting: Migrate to new tempfile code

cgwalters:
In general this is even cleaner now, though it was better after I
extracted a helper function for the "write tempfile with contents"
bits that were shared between metadata and regular file codepaths.

Depends on https://github.com/GNOME/libglnx/pull/17

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,7 +19,7 @@ include Makefile-decls.am
 
 shortened_sysconfdir = $$(echo "$(sysconfdir)" | sed -e 's|^$(prefix)||' -e 's|^/||')
 
-ACLOCAL_AMFLAGS = -I buildutil ${ACLOCAL_FLAGS}
+ACLOCAL_AMFLAGS = -I buildutil -I libglnx ${ACLOCAL_FLAGS}
 AM_CPPFLAGS += -DDATADIR='"$(datadir)"' -DLIBEXECDIR='"$(libexecdir)"' \
 	-DLOCALEDIR=\"$(datadir)/locale\" -DSYSCONFDIR=\"$(sysconfdir)\" \
 	-DSHORTENED_SYSCONFDIR=\"$(shortened_sysconfdir)\" \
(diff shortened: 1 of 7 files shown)


cgwalters:
Committed the libglnx dependency (though more review is welcome) and rebased :surfer:

cgwalters:
I saw locally `open(O_TMPFILE): Too many open files`

cgwalters:
In the big picture, I am now thinking that `O_TMPFILE` is going to be a good opportunity to re-pursue [O_OBJECT](http://www.spinics.net/lists/linux-fsdevel/msg75085.html) by instead having a special ioctl that at least XFS knows that can only be used for `O_TMPFILE`.  The model would be a lot clearer because the file wouldn't exist for other processes to see until the final link, so there'd be no visible-elsewhere mutable state.

cgwalters:
Hum interesting...`test-delta.sh` passes here.  I wonder whether this is a tempfile fallback issue.  Will look.

cgwalters:
This turns out to be that the untrusted-delta codepath explicitly tries to re-open the tempfile path, mmap it, and checksum it.  We can't do that with `O_TMPFILE`.

The right thing instead is to compute the checksum while writing, just like we do in `write_object()`.

cgwalters:
Depends on:

 - [ ] https://github.com/ostreedev/ostree/pull/392
 - [ ] https://github.com/ostreedev/ostree/pull/391

jlebon:
@cgwalters Can you rebase?

cgwalters:
:surfer:

cgwalters:
One risk of this is that now there are two pretty different ways to handle tmpfiles, and specifically CentOS 7.2 `xfs` doesn't do `O_TMPFILE`.  But I think ultimately the benefit is going to be worth it.

jlebon:
OK, these makes sense. I was a bit confused why libglnx needed the missing syscall tidbit, but I see the RHEL/CentOS kernels have `renameat2` backported.

@rh-atomic-bot r+ e105ee9

Overall, the O_TMPFILE approach is really nice. And this sets us up even better for when (hopefully) `linkat` can address the replace case too.

It might be interesting to add a new type in libglnx to encapsulate the pattern a bit more and have e.g. a variation on `glnx_fd_close` that also does the `linkat` bit. And maybe also another variation that opens an output stream for you.

hadess:
This regresses use of ostree on filesystems that don't support O_TMPFILE, such as ubifs, which is the root filesystem on quite a few eMMC/NAND block devices such as the ones used in ARM or "reduced functionality" Intel compatible platforms.

See https://bugzilla.gnome.org/show_bug.cgi?id=769453

PR #372, opened 2016-06-27
[merged] Upstream build fixes

dbnicholson:
Some fixes I found while rebasing our Endless version on 2016.5 and 2016.6. Our process uses `distcheck` to create a tarball from git to build a package from, so it hits some things that probably aren't normally seen upstream. We can keep some of these downstream, but I'd love to reduce our delta.

### DIFF
--- a/Makefile-boot.am
+++ b/Makefile-boot.am
@@ -38,6 +38,9 @@ endif
 if BUILDOPT_SYSTEMD
 systemdsystemunit_DATA = src/boot/ostree-prepare-root.service \
 	src/boot/ostree-remount.service
+
+# Allow the distcheck install under $prefix test to pass
+AM_DISTCHECK_CONFIGURE_FLAGS += --with-systemdsystemunitdir='$${libdir}/systemd/system'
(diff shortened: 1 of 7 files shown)


cgwalters:
Excellent patches.   Though I have to ask, is there a reason you don't just do a recursive git archive like [GContinuous](https://wiki.gnome.org/Projects/GnomeContinuous) and [rdgo](https://github.com/cgwalters/rpmdistro-gitoverlay)  and [Makefile.dist-packaging](https://github.com/ostreedev/ostree/blob/f4e92a1e06f80e1ad1d66eba6d060adbdbe4bd5b/packaging/Makefile.dist-packaging#L11) do?

@rh-atomic-bot r+ 2ed2d87

dbnicholson:
Our jenkins scripting is setup to work for all of our projects. It's pretty rigid, but `distcheck` is how it's setup now. Someday I'll probably try to rework it so that there are per-project recipes, but you might imagine this is not high on my priority queue. :smile:

PR #374, opened 2016-06-30
[merged] docs: Add a section on Docker

cgwalters:
This could have a lot more obviously, but just laying down my thoughts
as a starting point.


jlebon:
@rh-atomic-bot r+ bc2a51f

PR #375, opened 2016-07-01
[merged] delta: Add --if-not-exists option

cgwalters:
I often want to have "idempotent" systems that iterate to a known
state.  If after generating a commit, the system is interrupted, I'd
like the next run to still generate a delta.  But we don't want to
regenerate if one exists, hence this option.


jlebon:
@rh-atomic-bot r+ 0490b2a

PR #379, opened 2016-07-04
[merged] static-delta-core.c: squash unused var warning

jlebon:


cgwalters:
@rh-atomic-bot r+ 755c8ac

Issue #373, opened 2016-06-29
exception in non-404 error paths


cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1342168#c18

Mathnerd314:
I'm guessing that this is two problems:
* fetch data isn't freed properly (should be fixed by https://github.com/ostreedev/ostree/pull/353/commits/01fc763c69909b621843f445edbbb37d4e950b3a)
* An HTTP error (e.g. the 403) doesn't cancel the fetch, so it keeps going

cgwalters:
I managed to get the following trace:

```
(gdb) t a a bt

Thread 3 (Thread 0x7f1e7ebb6700 (LWP 24909)):
#0  0x00007f1e8258cbcd in poll () at /lib64/libc.so.6
#1  0x00007f1e82d6e0cc in g_main_context_iterate.isra.24 () at /lib64/libglib-2.0.so.0
#2  0x00007f1e82d6e1ec in g_main_context_iteration () at /lib64/libglib-2.0.so.0
#3  0x00007f1e82d6e229 in glib_worker_main () at /lib64/libglib-2.0.so.0
#4  0x00007f1e82d931b5 in g_thread_proxy () at /lib64/libglib-2.0.so.0
#5  0x00007f1e82869dc5 in start_thread () at /lib64/libpthread.so.0
#6  0x00007f1e8259721d in clone () at /lib64/libc.so.6

Thread 2 (Thread 0x7f1e7c0e5700 (LWP 24910)):
#0  0x00007f1e82591999 in syscall () at /lib64/libc.so.6
#1  0x00007f1e82db0b6e in g_cond_wait_until () at /lib64/libglib-2.0.so.0
#2  0x00007f1e82d43361 in g_async_queue_pop_intern_unlocked () at /lib64/libglib-2.0.so.0
#3  0x00007f1e82d93b12 in g_thread_pool_thread_proxy () at /lib64/libglib-2.0.so.0
#4  0x00007f1e82d931b5 in g_thread_proxy () at /lib64/libglib-2.0.so.0
#5  0x00007f1e82869dc5 in start_thread () at /lib64/libpthread.so.0
#6  0x00007f1e8259721d in clone () at /lib64/libc.so.6

Thread 1 (Thread 0x7f1e849cd840 (LWP 24907)):
#0  0x00007f1e82db08e5 in g_mutex_lock () at /lib64/libglib-2.0.so.0
[...]

PR #383, opened 2016-07-06
[merged] Fetcher deallocation cleanups

cgwalters:


mbarnes:
Looking at your stack trace in https://github.com/ostreedev/ostree/issues/373, it appears the session thread's `GMainContext` is somehow getting freed prematurely (based on `context=0xa0a0a0a0a0a0a0a0`). `OstreeFetcher` really doesn't do anything with it directly so I don't see any error there.

The test code looks good and works for me but the `finalize()` changes strike me as unnecessary on first impression since the shared structs are all ref-counted.  Thinking this could be a ref-counting bug but I'm still trying to reproduce it...

cgwalters:
Currently thinking this is a libsoup bug where in `idle_run_queue()` if we hit `priv->disposed` the source will be destroyed by the main loop, but left in `priv->run_queue_sources`.  Adding debug prints now to see.

mbarnes:
In addition to https://bugzilla.gnome.org/show_bug.cgi?id=768567, it might help to ensure the `SoupSession` is disposed in the same thread where it was created.  Otherwise there may yet be race conditions at play here.
```
diff --git a/src/libostree/ostree-fetcher.c b/src/libostree/ostree-fetcher.c
index 313df6a..10f361a 100644
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -44,7 +44,7 @@ typedef enum {
 typedef struct {
   volatile int ref_count;
 
-  SoupSession *session;
+  SoupSession *session;  /* not referenced */
   GMainContext *main_context;
   GMainLoop *main_loop;
 
@@ -140,7 +140,8 @@ thread_closure_unref (ThreadClosure *thread_closure)
 
   if (g_atomic_int_dec_and_test (&thread_closure->ref_count))
     {
-      g_clear_object (&thread_closure->session);
+      /* The session thread should have cleared this by now. */
+      g_warn_if_fail (thread_closure->session == NULL);
 
       g_clear_pointer (&thread_closure->main_context, g_main_context_unref);
       g_clear_pointer (&thread_closure->main_loop, g_main_loop_unref);
@@ -459,6 +460,7 @@ ostree_fetcher_session_thread (gpointer data)
    * context for this thread before creating the session. */
   g_main_context_push_thread_default (closure->main_context);
 
+  /* We retain ownership of the SoupSession reference. */
   closure->session = soup_session_async_new_with_options (SOUP_SESSION_USER_AGENT, "ostree ",
                                                           SOUP_SESSION_SSL_USE_SYSTEM_CA_FILE, TRUE,
                                                           SOUP_SESSION_USE_THREAD_CONTEXT, TRUE,
@@ -481,6 +483,11 @@ ostree_fetcher_session_thread (gpointer data)
 
   g_main_loop_run (closure->main_loop);
 
+  /* Since the ThreadClosure may be finalized from any thread we
+   * unreference the SoupSession ourself to ensure it's freed in
+   * the same thread where it was created. */
+  g_clear_pointer (&closure->session, g_object_unref);
+
   g_main_context_pop_thread_default (closure->main_context);
 
   thread_closure_unref (closure);
```

cgwalters:
Ok, I folded that change (with a bit more added) into the series and cleaned things up more.

cgwalters:
So far, this appears to fix things *without* the libsoup patch in https://bugzilla.gnome.org/show_bug.cgi?id=768567 which is a good sign too.

tests/libtest.sh
```diff
@@ -138,6 +138,7 @@ assert_file_has_content () {
     if ! grep -q -e "$2" "$1"; then
         sed -e 's/^/# /' < "$1" >&2
         echo 1>&2 "File '$1' doesn't match regexp '$2'"
+	cat $1
```
mbarnes: Is this a debugging leftover?

mbarnes:
Looks pretty bullet-proof to me.   :+1: with warm fuzzies.

Note the `libtest.sh` nit though.

mbarnes:
Something else you might want to sneak in here is the `GTask` in the `OstreeFetcherPendingURI` struct is unused.

tests/libtest.sh
cgwalters: Were you reading the "squashed" diff?  I split this into its own commit in e39f4188f644fba77888433a4b93c0da70a33eb3

mbarnes:
:+1:

mbarnes:
@rh-atomic-bot r+

PR #384, opened 2016-07-08
[merged] tests: Fix karg tests on ostree-booted system

cgwalters:
https://github.com/ostreedev/ostree/pull/372 caused these tests to
start failing when the host system is managed using ostree - since the
tests *do* replace the `ostree=` kernel argument.

### DIFF
--- a/tests/test-admin-deploy-karg.sh
+++ b/tests/test-admin-deploy-karg.sh
@@ -44,8 +44,12 @@ echo "ok deploy with --karg, but same config"
 
 ${CMD_PREFIX} ostree admin deploy --karg-proc-cmdline --os=testos testos:testos/buildmaster/x86_64-runtime
 for arg in $(cat /proc/cmdline); do
-    if test "${arg#ostree=*}" != ${arg}; then continue; fi
-    assert_file_has_content sysroot/boot/loader/entries/ostree-testos-0.conf "options.*$arg"
+    case "$arg" in
+	ostree=*) # Skip ostree arg that gets stripped out
+	   ;;
+	*) assert_file_has_content sysroot/boot/loader/entries/ostree-testos-0.conf "options.*$arg"
+	   ;;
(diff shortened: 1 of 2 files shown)


cgwalters:
@rh-atomic-bot delgate=dbnicholson

cgwalters:
@rh-atomic-bot delgate=dbnicholson

cgwalters:
@rh-atomic-bot delgate=dbnicholson

cgwalters:

tests/test-admin-deploy-karg.sh
```diff
@@ -44,6 +44,7 @@ echo "ok deploy with --karg, but same config"
 
 ${CMD_PREFIX} ostree admin deploy --karg-proc-cmdline --os=testos testos:testos/buildmaster/x86_64-runtime
 for arg in $(cat /proc/cmdline); do
+    if test "${arg#ostree=*}" != ${arg}; then continue; fi
```
dbnicholson: I would probably write this as a `case` statement. Something like
```
case "$arg" in
  ostree=*)
    # Skip ostree arg that gets stripped out
    ;;
  *)
    assert_file_has_content ...
    ;;
esac
```
But this seems fine, too. I'll try to give it a local test.

cgwalters:
More lines, but less obscure.  OK.  :arrow_up:

dbnicholson:
Only tested on a non-ostree system, but ack from me.

cgwalters:
You can use `@ rh-atomic-bot r+ d85b295` to approve, with no space between `@` and the name.

cgwalters:
@rh-atomic-bot r+ d85b295

dbnicholson:
> You can use @ rh-atomic-bot r+ d85b295 to approve, with no space between @ and the name.

Got it. Next time I'll try that.

PR #387, opened 2016-07-08
[merged] repo: Fix annotation for ostree_repo_add_gpg_signature_summary

dbnicholson:
GI needs to know this is a NULL terminated array of gchar or you'll
segfault using it from a binding.


cgwalters:
@rh-atomic-bot r+ 5e22402

PR #390, opened 2016-07-11
[merged] Revert "tests/libtest.sh: Print non-matching file on failure"

cgwalters:
This reverts commit 71301d18244a3a26a6ba6fd3934633ef94811a15.

I was confused by the fact that the non-matching content was
empty and forgot we already did print it.

Reported-by: smcv


jlebon:
@rh-atomic-bot r+ 98a77a5

PR #391, opened 2016-07-12
[merged] repo: Ensure we set mode for bare-user files before xattrs

cgwalters:
When trying to switch ostree to `O_TMPFILE`, I hit the fact that
by default it uses mode `000`.  It still works to write to the
open fd of course, but it *doesn't* work to set xattrs because
that code path for some reason in the kernel checks the mode bits.

This only broke for bare-user repos where we tried to set the xattr
before calling `fchmod()`, so just invert those two operations.


jlebon:
@rh-atomic-bot r+ 2381399

PR #392, opened 2016-07-12
[merged] deltas: Handle untrusted checksums faster and more robustly

cgwalters:
When reworking the ostree core [to use O_TMPFILE](https://github.com/ostreedev/ostree/pull/369),
I hit an issue in the way the untrusted delta codepath ends up trying
to re-open the file to checksum it.  That's not possible with
`O_TMPFILE` since the fd (which we opened `O_WRONLY`) is the only
accessible reference to the content.

Fix this by changing the delta processing code to update a checksum as
we're doing writes, which is also faster, and ends up simplifying the
code as well.

What would be an even larger simplification here is if we e.g. used a
separate thread calling `write_object()` or something like that; the
main issue I see there is somehow bridging the fact that function
wants a `GInputStream*` but the delta code is generating stream of
writes.


giuseppe:
the patch looks fine to me, I have done some testing as well.  Could `_ostree_repo_open_trusted_content_bare` be renamed since there is the same code path for both trusted and untrusted content?

cgwalters:
Fixup :arrow_up:

jlebon:
Yup, tested and working here as well!
@rh-atomic-bot r+ 6d1285b

PR #394, opened 2016-07-13
[merged] builtins/commit: Switch to using ostree_repo_write_dfd_to_mtree()

cgwalters:
Where we're referencing local files.  It's what
ostree_repo_write_directory_to_mtree() is doing internally anyways,
and I consider the `GFile*` based APIs to be generally deprecated for
the most part.


giuseppe:
@rh-atomic-bot r+ 37b1c56

PR #395, opened 2016-07-14
[merged] Makefile-tests.am: make check uses the built binaries

giuseppe:
The tests suite was failing locally as it was using the installed
version of rofiles-fuse, instead of the built one.  Create the needed
symlinks in tests/ as we are already doing for the "ostree" binary.

ostree-prepare-root and ostree-remount added for completeness.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -243,28 +243,15 @@ CLEANFILES += tests/libreaddir-rand.so tests/ostree-symlink-stamp \
 		tests/ostree-prepare-root-symlink-stamp tests/ostree-remount-symlink-stamp \
 		tests/rofiles-fuse-symlink-stamp tests/ostree
 
-tests/ostree-symlink-stamp: Makefile
+tests/%-symlink-stamp: Makefile
 	@set -e; \
-	real_bin=`cd $(top_builddir) && ./libtool --mode=execute echo ostree`; \
-	ln -sf "$${real_bin}" tests/ostree; \
-	touch $@
-
-tests/rofiles-fuse-symlink-stamp: Makefile
-	@set -e; \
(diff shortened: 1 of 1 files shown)


Makefile-tests.am
```diff
@@ -238,14 +239,34 @@ EXTRA_DIST += \
...
+		tests/ostree-prepare-root-symlink-stamp tests/ostree-remount-symlink-stamp \
+		tests/rofiles-fuse-symlink-stamp tests/ostree
 
 tests/ostree-symlink-stamp: Makefile
 	@set -e; \
 	real_bin=`cd $(top_builddir) && ./libtool --mode=execute echo ostree`; \
 	ln -sf "$${real_bin}" tests/ostree; \
 	touch $@
 
+tests/rofiles-fuse-symlink-stamp: Makefile
```
cgwalters: Couldn't we easily collapse these 4 with a [pattern rule](https://www.gnu.org/software/make/manual/html_node/Pattern-Rules.html)?

giuseppe: I could make for the new 3 ones, but the rule for ostree is using libtool, while the new ones point directly to the binary file

giuseppe: working on a new revision

giuseppe:
I pushed a fixup patch :arrow_up:

cgwalters:
That's some magic :black_joker: 

Looks OK to me though I have to say I'm a bit grumpy about the hacks involved in the uninstalled checks.  I get they're convenient but on the other hand [installed tests](https://wiki.gnome.org/Initiatives/GnomeGoals/InstalledTests) are fundamentally more reliable.

@rh-atomic-bot r+ 033f32c

PR #396, opened 2016-07-14
Checksum b64 for bindings

dbnicholson:
Provide checksum_b64 functions that can be used from GI. This allows external GI programs to use ostree's modified base64 encoding without having to reimplement it.


src/libostree/ostree-core.c
```diff
@@ -1273,6 +1273,20 @@ ostree_checksum_to_bytes_v (const char *checksum)
...
 /**
+ * ostree_checksum_b64_to_bytes:
+ * @checksum: An ASCII checksum
+ *
+ * Returns: (transfer full) (array fixed-size=32): Binary version of @checksum.
+ */
+guchar *
+ostree_checksum_b64_to_bytes (const char *checksum)
+{
+  guchar *ret = g_malloc (44);
```
cgwalters: This only needs to be 32, right?

cgwalters:
bot, add author to whitelist

src/libostree/ostree-core.c
dbnicholson: Duh. I fixed the annotation but didn't change the allocation.

dbnicholson:
Ugh. Accidentally deleted my branch and github closed this. Continue on #398, please.

PR #398, opened 2016-07-14
[merged] Checksum b64 for bindings

dbnicholson:
Provide checksum_b64 functions that can be used from GI. This allows external GI programs to use ostree's modified base64 encoding without having to reimplement it.

Sorry, I accidentally deleted my branch and github closed #396. This has the fix for the allocation in the `to_bytes` function.


src/libostree/libostree.sym
```diff
@@ -352,8 +352,8 @@ global:
  */
 
 /* UNCOMMENT WHEN ADDING THE FIRST NEW SYMBOL FOR 2016.8
-LIBOSTREE_2016.7 {
+LIBOSTREE_2016.8 {
```
cgwalters: I'm pretty sure I psyched myself out on this originally - I stared at it for 10 seconds and convinced myself it was wrong, then changed it when I shouldn't have.  Tip for future me: Don't overthink it.

cgwalters:
This is all good, but just for future reference I like to see a "real world use case" at least briefly described in the commit message.  "I'm using these b64 functions to parse static delta names in a script" or something.

@rh-atomic-bot r+ 10dd4cb

PR #399, opened 2016-07-15
Ostree will block out chown when run as non-root

mtohmaz:
@rhatdan @cgwalters This isn't mergeable yet, but I'd like feedback on the changes. Essentially in order to pull as non-root user, I blocked out any chown calls hat would require root privilege. Right now, the "privileged" flag is set to false, but I was wondering if we should create flag in ostree, or simply have it check if running as non-root.


cgwalters:
Ostree supports this natively, just use a `bare-user` repo.  See what xdg-app does.

PR #401, opened 2016-07-19
ostree commit: Fix combining trees with multiple --tree=ref arguments

wmanley:
You'd expect

```
ostree commit --tree=ref=A --tree=ref=B
```

to produce a commit with the union of the trees given.  Instead you'd get
a commit with the contents of just the latter commit.  This was due to an
optimisation where we'd skip filling out the `files` and `subdirs`
members of the mtree, just filling in the metadata instead.  This backfires
becuase this same code relies on checking the `files` and `subdirs` members
itself to work out whether the mtree is empty.

This commit removes the optimisation, fixing the bug.  Maybe there's a way
to keep the optimisation and still fix the bug but it's not obvious to
me.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -2593,16 +2593,6 @@ write_directory_to_mtree_internal (OstreeRepo                  *self,
 
       ostree_mutable_tree_set_metadata_checksum (mtree, ostree_repo_file_tree_get_metadata_checksum (repo_dir));
 
-      /* If the mtree was empty beforehand, the checksums on the mtree can simply
-       * become the checksums on the tree in the repo. Super simple. */
-      if (g_hash_table_size (ostree_mutable_tree_get_files (mtree)) == 0 &&
-          g_hash_table_size (ostree_mutable_tree_get_subdirs (mtree)) == 0)
-        {
-          ostree_mutable_tree_set_contents_checksum (mtree, ostree_repo_file_tree_get_contents_checksum (repo_dir));
-          ret = TRUE;
-          goto out;
-        }
-
       filter_result = OSTREE_REPO_COMMIT_FILTER_ALLOW;
     }
   else
--- a/tests/test-commit.sh
+++ b/tests/test-commit.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+#
+# Copyright (C) 2016 William Manley <will@williammanley.net>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
+
+set -euo pipefail
+
+. $(dirname $0)/libtest.sh
+
+skip_without_user_xattrs
+
+echo "1..1"
+
+setup_test_repository "bare"
+
+$OSTREE commit --tree=dir=$test_srcdir/tree-A -b A -m A
+$OSTREE commit --tree=dir=$test_srcdir/tree-B -b B -m B
+
+$OSTREE commit --tree=ref=A --tree=ref=B -b Combine -m Combine
+$OSTREE ls Combine / >contents
+assert_file_has_content contents "/file-a"
+assert_file_has_content contents "/file-b"
+echo "ok combining ref trees"
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -40,6 +40,7 @@ dist_uninstalled_test_scripts = tests/test-symbols.sh
 
 dist_test_scripts = \
 	tests/test-basic.sh \
+	tests/test-commit.sh \
 	tests/test-pull-subpath.sh \
 	tests/test-archivez.sh \
 	tests/test-remote-add.sh \


src/libostree/ostree-repo-commit.c
```diff
@@ -2593,16 +2593,6 @@ write_directory_to_mtree_internal (OstreeRepo                  *self,
 
       ostree_mutable_tree_set_metadata_checksum (mtree, ostree_repo_file_tree_get_metadata_checksum (repo_dir));
```
wmanley: I'm unsure whether this `set_metadata_checksum` should also be removed.  I'm a little confused what the metadata checksum is as I don't think there's a parallel in the git world.

giuseppe:
bot, test pull request

tests/test-commit.sh
```diff
@@ -0,0 +1,35 @@
...
+
+. $(dirname $0)/libtest.sh
+
+echo "1..1"
+
+setup_test_repository "bare"
+
+$OSTREE commit --tree=dir=$test_srcdir/tree-A -b A -m A
+$OSTREE commit --tree=dir=$test_srcdir/tree-B -b B -m B
+
```
giuseppe: does the commit miss some files?

Anyway, it would probably be cleaner to just do in the test itself:

mkdir tree-A tree-B
touch tree-A/file-a tree-B/file-b

wmanley: > does the commit miss some files?

Quite right.  I've force-pushed a fixed version.

> Anyway, it would probably be cleaner to just do in the test itself:

I considered doing this originally but decided against.  I think it's clearer to have test-artifacts static rather than generating them procedurally if possible.  The artefacts are in `$test_srcdir` so you know that they are going to appear to the test exactly as they are and are not going to be affected by side-effects from other tests.

This is particularly true when it starts to get more complicated, so less important when applied to this test and this test-directory.

I can change it if you like though.

wmanley:
Hmm, Jenkins isn't happy.  The test fails with:

    ERROR: tests/test-commit.sh
    ===========================

    ...
    + setup_test_repository bare
    ...
    + env LD_PRELOAD=/code/tests/libreaddir-rand.so ostree --repo=/var/tmp/tap-test.fyWGVM/repo commit --tree=dir=/code/tests/tree-A -b A -m A
    error: fsetxattr: Operation not supported

So it's a failure in the first line of my test:

    $OSTREE commit --tree=dir=$test_srcdir/tree-A -b A -m A

The test passes on my machine.

You might think that `fsetxattr` could fail because of a particular filesystem in use, but there are other places where we do commits in the tests.

Hmmm...

giuseppe:
you'll probably need `skip_without_user_xattrs` as it is done by some other tests as well

cgwalters:
Broadly speaking we haven't found a lot of use for the in-memory union approach because it's really common to need to regenerate derived cache files (`ldconfig`, `gtk-update-icon-cache`, etc.) which requires checking the trees out in the FS rather than in process.

Does your use case really not include those?

wmanley:
> Does your use case really not include those?

It doesn't currently.  I'm looking to use ostree deploying software to appliance-like devices which will run headless.  I'm currently trying get it working on a Nvidia Tegra TK1.  Currently I've committed the Linux 4 Tegra rootfs bundle that Nvidia provides into ostree and I want to combine that with the kernel and the ostree binaries I've built into a single tree for deployment.

It's just rough for now, but in the future I would expect to be using Makefiles to automate these steps whereby I would have a slowly changing base (maybe generated by apt-get inside a chroot) that I'll combine with fast-changing software from a git repo.  I'll want that combine step to be fast otherwise it'll slow down my development cycle.

wmanley:
> you'll probably need `skip_without_user_xattrs` as it is done by some other tests as well

Good idea, but it didn't seem to work:

```
+ skip_without_user_xattrs
+ touch test-xattrs
+ setfattr -n user.testvalue -v somevalue test-xattrs
+ echo 1..1
```

Seems to get passed fine.  I'll have to put some more debugging in.

cgwalters:
"I'll want that combine step to be fast otherwise it'll slow down my development cycle."

Agreed - and in fact, https://wiki.gnome.org/Projects/GnomeContinuous works very similarly to this - we use Yocto for the fixed base, and then a custom git+ostree build system on top for continuous delivery.

However, gcontinuous does use `--link-checkout-speedup` and reruns the triggers/post-scripts, since in general the higher layers *will* install shared libraries (so you need to rerun `ldconfig`), or install icons (so one needs to rerun `gtk-update-icon-cache`) etc.

I don't disagree this is a bug, however I tried to fix this before and got bogged down in some of the details - this bit of code is fragile.  And decided against trying to fix it right now given that in practice one does need the fs level checkouts.  Right?

src/libostree/ostree-repo-commit.c
cgwalters: Yeah...there isn't.  In git the metadata is inline in the tree object but for ostree I split it out since it contains xattrs and is potentially a bit larger.  In practice I'm not sure this was a win though.

cgwalters:
So git annotate points us to https://bugzilla.gnome.org/show_bug.cgi?id=707727 but that's just code cleanup...the end barrier is before when the code was in `ostree-repo.c`.  Some annotate searching leads us to:

https://github.com/ostreedev/ostree/commit/e9fd921a

And what we can see from this is it predates the idea of using `--link-checkout-speedup`...which we expect everyone to do now.

If that's the case then, perhaps it's simplest to not think too hard about this and just nuke the optimization as you've done here.

As far as the test then...I would agree with @giuseppe on this, we have lots and lots of other tests which procedurally generate files.  Speaking of other tests, I'd also prefer having this in e.g. `test-basic.sh` - not sure we need a whole new file just for this test yet.

src/libostree/ostree-repo-commit.c
```diff
@@ -2593,16 +2593,6 @@ write_directory_to_mtree_internal (OstreeRepo                  *self,
...
 
-      /* If the mtree was empty beforehand, the checksums on the mtree can simply
-       * become the checksums on the tree in the repo. Super simple. */
-      if (g_hash_table_size (ostree_mutable_tree_get_files (mtree)) == 0 &&
-          g_hash_table_size (ostree_mutable_tree_get_subdirs (mtree)) == 0)
-        {
-          ostree_mutable_tree_set_contents_checksum (mtree, ostree_repo_file_tree_get_contents_checksum (repo_dir));
-          ret = TRUE;
```
alexlarsson: I think we can still copy the checksum so we don't have to recalculate it. We invalidate this when something changes in the mtree anyway.

alexlarsson: Actually, that doesn't work, because we invalidate the checksum when we add the children.
So, lets just drop this optimization.

cgwalters:
Obsoleted by https://github.com/ostreedev/ostree/pull/581

PR #403, opened 2016-07-19
[merged] ostree-prepare-root as init

wmanley:
These commits generalise `ostree-prepare-root` for the purposes of using it to boot Linux for Tegra (Ubuntu 14.04) on the Nvidia Tegra TK1 dev-boards.  On the Tegra we don't use an initrd or systemd so the existing way that `ostree-prepare-root` is integrated is insufficient.

This has required 3 main generalisations.  They are mostly separable:

1. After setup `exec` a new init rather than `exit`ing.
2. Cope with `/proc` not having been mounted for the purposes of reading `/proc/cmdline`
3. Allow the eventual sysroot to already be mounted on `/` rather than `/sysroot` by using `pivot_root` where appropriate

I'm aware this has been [discussed on the mailing list](https://mail.gnome.org/archives/ostree-list/2015-October/msg00015.html) last October.

I do rather like this ostree model.  The main ostree binary itself takes advantage of various complex dependencies for it's advanced behaviour, but I can easily build `ostree-prepare-root` statically for boot integration.

These changes seem to work well on the device.  I've not completed the bootloader integration for this board but I think this work on `ostree-prepare-root` stands for itself.

I've opened this PR to gain some early feedback such as:

* Are these changes within the scope of the tool?
* Would they in principle be accepted?
* What other TODO items do I need to add to the list before this can be merged?

I'm not sure that this is the workflow that ostree follows but it seemed to be consistent with some of the other PRs open at the moment.

**TODO:**

- [x] Test that I haven't broken the existing systemd+dracut integration
- [x] Test that I haven't broken `overlayfs` integration
- [x] Test that we don't reintroduce the bug fixed in 4f75d4e where there are two filesystems mounted at `/sysroot`
- [x] Add automated tests

**For future PRs**

- Build `ostree-prepare-root` as a static binary

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -156,12 +156,10 @@ main(int argc, char *argv[])
   int orig_cwd_dfd;
 
   if (argc < 2)
-    {
-      fprintf (stderr, "usage: ostree-prepare-root SYSROOT\n");
-      exit (EXIT_FAILURE);
-    }
+    root_mountpoint = "/";
+  else
+    root_mountpoint = argv[1];
 
-  root_mountpoint = argv[1];
   deploy_path = resolve_deploy_path (root_mountpoint);
 
   /* Create a temporary target for our mounts in the initramfs; this will
@@ -303,6 +301,15 @@ main(int argc, char *argv[])
       perrorv ("failed to MS_MOVE %s to %s", deploy_path, root_mountpoint);
       exit (EXIT_FAILURE);
     }
-  
-  exit (EXIT_SUCCESS);
+
+  if (getpid() == 1)
+    {
+      execl ("/sbin/init", "/sbin/init", NULL);
+      perrorv ("failed to exec init inside ostree");
+      exit (EXIT_FAILURE);
+    }
+  else
+    {
+      exit (EXIT_SUCCESS);
+    }
 }
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -191,7 +191,6 @@ main(int argc, char *argv[])
   char *deploy_path = NULL;
   char srcpath[PATH_MAX];
   char destpath[PATH_MAX];
-  char newroot[PATH_MAX];
   struct stat stbuf;
   int orig_cwd_dfd;
 
@@ -202,16 +201,6 @@ main(int argc, char *argv[])
 
   deploy_path = resolve_deploy_path (root_mountpoint);
 
-  /* Create a temporary target for our mounts in the initramfs; this will
-   * be moved to the new system root below.
-   */
-  snprintf (newroot, sizeof(newroot), "%s.tmp", root_mountpoint);
-  if (mkdir (newroot, 0755) < 0)
-    {
-      perrorv ("Couldn't create temporary sysroot '%s': ", newroot);
-      exit (EXIT_FAILURE);
-    }
-  
   /* Work-around for a kernel bug: for some reason the kernel
    * refuses switching root if any file systems are mounted
    * MS_SHARED. Hence remount them MS_PRIVATE here as a
@@ -225,7 +214,7 @@ main(int argc, char *argv[])
     }
 
   /* Make deploy_path a bind mount, so we can move it later */
-  if (mount (deploy_path, newroot, NULL, MS_BIND, NULL) < 0)
+  if (mount (deploy_path, deploy_path, NULL, MS_BIND, NULL) < 0)
     {
       perrorv ("failed to initial bind mount %s", deploy_path);
       exit (EXIT_FAILURE);
@@ -233,7 +222,7 @@ main(int argc, char *argv[])
 
   /* Link to the deployment's /var */
   snprintf (srcpath, sizeof(srcpath), "%s/../../var", deploy_path);
-  snprintf (destpath, sizeof(destpath), "%s/var", newroot);
+  snprintf (destpath, sizeof(destpath), "%s/var", deploy_path);
   if (mount (srcpath, destpath, NULL, MS_MGC_VAL|MS_BIND, NULL) < 0)
     {
       perrorv ("failed to bind mount %s to %s", srcpath, destpath);
@@ -245,7 +234,7 @@ main(int argc, char *argv[])
   snprintf (srcpath, sizeof(srcpath), "%s/boot/loader", root_mountpoint);
   if (lstat (srcpath, &stbuf) == 0 && S_ISLNK (stbuf.st_mode))
     {
-      snprintf (destpath, sizeof(destpath), "%s/boot", newroot);
+      snprintf (destpath, sizeof(destpath), "%s/boot", deploy_path);
       if (lstat (destpath, &stbuf) == 0 && S_ISDIR (stbuf.st_mode))
         {
           snprintf (srcpath, sizeof(srcpath), "%s/boot", root_mountpoint);
@@ -257,7 +246,7 @@ main(int argc, char *argv[])
         }
     }
 
-  /* Here we do a dance to chdir to the newroot so that we can have
+  /* Here we do a dance to chdir to the deploy_path so that we can have
    * the potential overlayfs mount points not look ugly.  However...I
    * think we could do this a lot earlier and make all of the mounts
    * here just be relative.
@@ -269,9 +258,9 @@ main(int argc, char *argv[])
       exit (EXIT_FAILURE);
     }
 
-  if (chdir (newroot) < 0)
+  if (chdir (deploy_path) < 0)
     {
-      perrorv ("failed to chdir to newroot");
+      perrorv ("failed to chdir to deploy_path");
       exit (EXIT_FAILURE);
     }
 
@@ -324,19 +313,39 @@ main(int argc, char *argv[])
 
   touch_run_ostree ();
 
-  /* Move physical root to $deployment/sysroot */
-  snprintf (destpath, sizeof(destpath), "%s/sysroot", newroot);
-  if (mount (root_mountpoint, destpath, NULL, MS_MOVE, NULL) < 0)
+  /* In this instance typically we have our ready made-up up root at
+   * /sysroot/ostree/deploy/.../ (deploy_path) and the real rootfs at
+   * /sysroot (root_mountpoint).  We want to end up with our made-up root at
+   * /sysroot/ and the real rootfs under /sysroot/sysroot as systemd will be
+   * responsible for moving /sysroot to /.
+   *
+   * We need to do this in 3 moves to avoid trying to move /sysroot under
+   * itself:
+   *
+   * 1. /sysroot/ostree/deploy/... -> /sysroot.tmp
+   * 2. /sysroot -> /sysroot.tmp/sysroot
+   * 3. /sysroot.tmp -> /sysroot
+   */
+  if (mkdir ("/sysroot.tmp", 0755) < 0)
     {
-      perrorv ("Failed to MS_MOVE %s to '%s'", root_mountpoint, destpath);
+      perrorv ("couldn't create temporary sysroot /sysroot.tmp: ");
       exit (EXIT_FAILURE);
     }
 
-  /* Now that we've set up all the bind mounts in /sysroot.tmp which
-   * points to the deployment, move it /sysroot.  From there,
-   * systemd's initrd-switch-root.target will take over.
-   */
-  if (mount (newroot, root_mountpoint, NULL, MS_MOVE, NULL) < 0)
+  if (mount (deploy_path, "/sysroot.tmp", NULL, MS_MOVE, NULL) < 0)
+    {
+      perrorv ("failed to MS_MOVE '%s' to '/sysroot.tmp'", deploy_path);
+      exit (EXIT_FAILURE);
+    }
+
+  if (mount (root_mountpoint, "/sysroot.tmp/sysroot", NULL, MS_MOVE, NULL) < 0)
+    {
+      perrorv ("failed to MS_MOVE '%s' to '/sysroot.tmp/sysroot'",
+          root_mountpoint);
+      exit (EXIT_FAILURE);
+    }
+
+  if (mount ("/sysroot.tmp", root_mountpoint, NULL, MS_MOVE, NULL) < 0)
     {
       perrorv ("failed to MS_MOVE %s to %s", deploy_path, root_mountpoint);
       exit (EXIT_FAILURE);
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -33,6 +33,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/param.h>
+#include <sys/syscall.h>
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdarg.h>
@@ -184,6 +185,12 @@ resolve_deploy_path (const char * root_mountpoint)
   return deploy_path;
 }
 
+static int
+pivot_root(const char * new_root, const char * put_old)
+{
+  return syscall(__NR_pivot_root, new_root, put_old);
+}
+
 int
 main(int argc, char *argv[])
 {
@@ -197,6 +204,7 @@ main(int argc, char *argv[])
   else
     root_mountpoint = argv[1];
 
+  root_mountpoint = realpath (root_mountpoint, NULL);
   deploy_path = resolve_deploy_path (root_mountpoint);
 
   /* Work-around for a kernel bug: for some reason the kernel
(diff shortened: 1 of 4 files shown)


giuseppe:
bot, test pull request

gatispaeglis:
Awesome stuff. I have this item on my TODO list as well (https://bugreports.qt.io/browse/QTBUG-52758). I can help with testing the patch set on my reference devices. I should have some time for this during the next week.

src/switchroot/ostree-prepare-root.c
```diff
@@ -221,16 +220,6 @@ main(int argc, char *argv[])
 
   deploy_path = resolve_deploy_path (root_mountpoint);
 
-  /* Create a temporary target for our mounts in the initramfs; this will
-   * be moved to the new system root below.
```
cgwalters: This code was written for a reason in 4f75d4ea0b640dd69f13fd31976d6ddfb42c977a - are you sure it won't reintroduce that bug?

cgwalters:
Your commit messages are excellent and the code style/changes look really good at a high level.  What I'm a lot less certain about is whether this is (as you have on your checklist) going to break the initrd case.

I'll do some testing with this too myself.

(Semi-related; I'm guessing if you're not using an initrd you also don't have a `/boot` partition; have you hit the issues in https://github.com/ostreedev/ostree/pull/215 ?)

gatispaeglis:
The current patch set on systemd+dracut based images results in:
Resolved OSTree target..... 
Failed to MS_MOVE /sysroot to 'sysroot': Too many levels of symbolic links

cgwalters:
Maybe it's simplest in the end to basically fork the current `ostree-prepare-root.c` into `ostree-prepare-root-initramfs.c` and `ostree-prepare-root-bare.c`, and add `--enable-switchroot-noinitramfs` which selects between the two?

gatispaeglis:
I think that could result in too much duplicated code - kernel command line parsing, mounting /boot when single partition system, the overlay (ostree admin unlock) handling logic, and etc.

gatispaeglis:
Verified also the initramfs-less booting, worked seamlessly.

src/switchroot/ostree-prepare-root.c
```diff
@@ -108,63 +168,64 @@ touch_run_ostree (void)
...
+int
+main(int argc, char *argv[])
+{
+  const char *root_mountpoint = NULL;
+  char *deploy_path = NULL;
+  char srcpath[PATH_MAX];
+  struct stat stbuf;
+
+  if (argc < 2)
+    root_mountpoint = "/sysroot";
```
gatispaeglis: Would it make more sense to have `root_mountpoint = "/"` as the default value, when ostree-prepare-root called without args?

I think this kernel command line looks better:
`init=/ostree/boot.1/qt-os/aad0fb830ecea8c807dcdbbd93a88c65dd7489bb6532ef48eb3dc19cae3a7b22/0/usr/sbin/ostree-prepare-root`
then:
`init=/ostree/boot.1/qt-os/aad0fb830ecea8c807dcdbbd93a88c65dd7489bb6532ef48eb3dc19cae3a7b22/0/usr/sbin/ostree-prepare-root /`

src/switchroot/ostree-prepare-root.c
```diff
@@ -259,41 +308,58 @@ main(int argc, char *argv[])
...
+          die ();
+        }
 
-  touch_run_ostree ();
+      /* Now that we've set up all the bind mounts in the deployment, move it
+       * /sysroot.  From there, systemd's initrd-switch-root.target will take over.
+       */
+      if (mount (".", root_mountpoint, NULL, MS_MOVE, NULL) < 0)
+        {
+          perrorv ("failed to MS_MOVE %s to %s", deploy_path, root_mountpoint);
```
gatispaeglis: failed vs Failed inconsistency in perrorv() messages.

src/switchroot/ostree-prepare-root.c
```diff
@@ -259,41 +308,58 @@ main(int argc, char *argv[])
...
+      /* Move physical root to $deployment/sysroot */
+      if (mount (root_mountpoint, "sysroot", NULL, MS_MOVE, NULL) < 0)
+        {
+          perrorv ("Failed to MS_MOVE %s to 'sysroot'", root_mountpoint);
+          die ();
+        }
 
-  touch_run_ostree ();
+      /* Now that we've set up all the bind mounts in the deployment, move it
+       * /sysroot.  From there, systemd's initrd-switch-root.target will take over.
```
gatispaeglis: "move it /sysroot" => "move it to /sysroot" ?

src/switchroot/ostree-prepare-root.c
```diff
@@ -174,60 +235,48 @@ main(int argc, char *argv[])
...
       perrorv ("Failed to make \"/\" private mount: %m");
-      exit (EXIT_FAILURE);
+      die ();
     }
 
   /* Make deploy_path a bind mount, so we can move it later */
-  if (mount (deploy_path, newroot, NULL, MS_BIND, NULL) < 0)
+  if (mount (deploy_path, deploy_path, NULL, MS_BIND, NULL) < 0)
     {
       perrorv ("failed to initial bind mount %s", deploy_path);
```
gatispaeglis: Not introduced by this change, but:

"failed to initial bind mount" => "failed to create initial bind mount" ?

gatispaeglis:
Also for kernel to be able to use ostree-prepare-root as PID-1, it needs to be a static binary. As you have already mentioned. Further questions are:

1) Should this binary always be build as static?

   Sounds ok to me.

2) Have a command line arg "--with-static-switchroot"?

   Not ideal for installers. As I would prefer to leave it up to users to choose initramfs VS initramfs-less approach. Without requiring users to rebuild the binary, if the opposite approach was selected at compile time.

3) Build 2 binaries (from the same source code) by default - static and dynamic version of ostree-prepare-root:
    ostree-prepare-root (dynamic)
    ostree-prepare-root-init (static)

cgwalters:
@gatispaeglis Are you aware of a distribution which supports both initrd and no-initrd that doesn't build from source (i.e. isn't like OpenEmbedded/Buildroot)?  At least for my use case (Fedora/CentOS) we're pretty initrd-centric.  I guess one could manually set it up with our binaries, but it's not supported by the installer (Anaconda) for example.

@wmanley Do you think you can pick up debugging the initrd case?  If not, I can probably give it a shot sometime later this week...let me know so we don't conflict.

wmanley:
> Would it make more sense to have root_mountpoint = "/" as the default value, when ostree-prepare-root called without args?

Agreed, fixed.

src/switchroot/ostree-prepare-root.c
wmanley: > This code was written for a reason in 4f75d4e - are you sure it won't reintroduce that bug?

I've pushed a fixup that should fix this but haven't yet tested it explicitly.

wmanley: I've added a TODO list item to test this.

wmanley:
> have you hit the issues in #215 ?

I haven't because I haven't completed bootloader integration yet.  For playing with this I've been booting with init=/bin/bash and running `exec ostree-prepare-root /`.

The Tegra TK1 uses u-boot but is configured to boot from an extlinux configuration file in `/boot/extlinux/extlinux.conf` so I'll need to do a little more work on it before it'll "just work".  The TK1 uses a single ext4 partition for `/` including `/boot`, although IIRC the bootloader is configured to first try booting `/boot/extlinux/extlinux.conf` and then if that fails it falls back to booting `/extlinux/extlinux.conf`.

wmanley:
> Maybe it's simplest in the end to basically fork the current ostree-prepare-root.c into ostree-prepare-root-initramfs.c and ostree-prepare-root-bare.c, and add --enable-switchroot-noinitramfs which selects between the two?

I would disagree. There would be a lot of common code between the two, and this will be a pure generalisation.  This means that it's more likely to work or behave helpfully for other more exotic configurations, e.g. booting distros that use an initrd but not systemd/dracut.

wmanley:
> The current patch set on systemd+dracut based images results in:
> Resolved OSTree target..... 
> Failed to MS_MOVE /sysroot to 'sysroot': Too many levels of symbolic links

I suspected this might be the case.  I've reintroduced `/sysroot.tmp` in 	96801af, although I've not yet had a chance to test it as my connectivity is rather poor this week so can't afford to download a VM for testing.

I'll have to fiddle with the commit messages when squashing at the end.

src/switchroot/ostree-prepare-root.c
wmanley: Comments refactored in 	96801af.

src/switchroot/ostree-prepare-root.c
wmanley: Fixed in ac168422b03bb36dc8484fc1a37d15e4b01e7096 and various other fixups.

src/switchroot/ostree-prepare-root.c
```diff
@@ -259,41 +308,85 @@ main(int argc, char *argv[])
...
+          die ();
+        }
 
-  /* Move physical root to $deployment/sysroot */
-  snprintf (destpath, sizeof(destpath), "%s/sysroot", newroot);
-  if (mount (root_mountpoint, destpath, NULL, MS_MOVE, NULL) < 0)
-    {
-      perrorv ("Failed to MS_MOVE %s to '%s'", root_mountpoint, destpath);
-      exit (EXIT_FAILURE);
+      if (chdir ("/sysroot.tmp") < 0)
```
wmanley: I'm not sure if this `chdir` is necessary.  I haven't tried this yet.  I'd like to remove it if it works without.

src/switchroot/ostree-prepare-root.c
wmanley: Fixed in e68276f7cb939a4bd813632f9fa3be96ca11db3c

wmanley:
> Also for kernel to be able to use ostree-prepare-root as PID-1, it needs to be a static binary.

Some data on file sizes built for AMD64 with -O2 and stripped:

| configuration      | size |
|--------------------|-----|
| glibc static    |   724K |
| glibc shared    |   12K |
| musl static    |   30K |

wmanley:
> @wmanley Do you think you can pick up debugging the initrd case? If not, I can probably give it a shot sometime later this week...let me know so we don't conflict.

I've given it a crack, but I haven't had a chance to test it yet.  I'm going to have a play around with using `unshare` for testing these different use-cases, but I'm not sure how `pivot_root` will interact with mount namespaces.

src/switchroot/ostree-prepare-root.c
wmanley: My automated test passes without it so I've removed it.

wmanley:
All these fixups are getting a little silly so I'll do a rebase and push -f unless anyone has any objections? This will be necessary eventually to drop 3d935aa: "ostree-prepare-root: Don't bother creating /sysroot.tmp"

cgwalters:
Agreed it's force push :hammer: time.  (Also, note one can't `fixup! fixup!` anyways.  I use https://github.com/cgwalters/homegit/blob/master/bin/git-addfixup )

gatispaeglis:
> Test that I haven't broken the existing systemd+dracut integration

With the latest patch set this works as expected.

> Test that we don't reintroduce the bug fixed in 4f75d4e where there are two filesystems mounted at /sysroot

Checked this as well. The patch does not reintroduce the bug fixed in 4f75d4e.

>  Are you aware of a distribution which supports both initrd and no-initrd that doesn't build from source (i.e. isn't like OpenEmbedded/Buildroot)?

I had a specific (corner case) use case in mind, but probably this is not relevant for the upstream ostree project. I mentioned "an installer", but maybe my use case is better described as a "converter" - a script that converts OpemEmbedded distro into ostree-based distro. Therefore I wanted to provide statically and dynamically build binaries, but I start to think that I can use a static binary for both - initramfs and initramfs-less booting.

wmanley:
> Also, note one can't fixup! fixup! anyways

This was fixed in git 1.8.4.  See the [release notes](https://github.com/git/git/blob/master/Documentation/RelNotes/1.8.4.txt#L157-L158):

>  * Having multiple "fixup!" on a line in the rebase instruction sheet did not work very well with "git rebase -i --autosquash".

cgwalters:
Regarding static linking - this does seem like an obvious use case for musl.  I have absolutely no problem with whatever autoconf bits we need for that.

src/switchroot/ostree-prepare-root.c
wmanley: > Would it make more sense to have `root_mountpoint = "/"` as the default value, when ostree-prepare-root called without args?

I've remembered why this wasn't a problem for me.  I have a symlink `/sysroot -> .` on my rootfs, so this was being resolved to `/`.

gatispaeglis:
> What other TODO items do I need to add to the list before this can be merged?

I think one more item that needs to be done here to simplify integration with bootloaders includes updating boot/loader/entries/*.conf files. An example of what it currently contains:

`initrd /ostree/qt-os-../initramfs-*`
`options ostree=..`
`title ..`
`linux /ostree/qt-os-../vmlinuz-*`
 
When initramfs-* not provided in the tree, assume that booting with `ostree-prepare-root` as PID1. The `options` field should be extended to include:

`options ostree=.. inti=path/to/ostree-prepare-root/in/$ostree/deployment`

The current workaround with `ostree-grub-generator` would be: https://paste.kde.org/pafbskaqk
On u-boot shell this can not be done AFAIK, as the shell is very limited. Appending `init=` to `options` would help for bootloader integration code. 

Does this sound reasonable?

wmanley:
> I'm going to have a play around with using `unshare` for testing these different use-cases

I've written some basic test-cases for this.  They need to be run as root with:

    sudo make check TESTS=tests/test-switchroot.sh

wmanley:
> I think one more item that needs to be done here to simplify integration with bootloaders includes updating boot/loader/entries/*.conf files.
>
> ...
>
> Does this sound reasonable?

That sounds reasonable, but I'd like to get this PR with just changes to `ostree-prepare-root` merged first.  I've not yet had a play around with the bootloader integration code and I'll want to do a bunch of experimenting to explore the problem space before implementing a proper solution.

gatispaeglis:
Its fine by my, just wanted to point out that this might require a bit more work :+1:

wmanley:
Note to self: [Autotools: compiling two C++ files with two compilers](https://lists.gnu.org/archive/html/help-gnu-utils/2007-01/msg00007.html)

src/switchroot/ostree-prepare-root.c
```diff
@@ -44,6 +44,25 @@
 
 #include "ostree-mount-util.h"
 
+static void
+die (void)
+{
+  if (getpid() == 1)
+    {
+      /* We're init - Give the user a chance to recover */
+      fprintf (stderr, "init: Entering recovery shell\n");
```
cgwalters: I'm wondering if this should be a build-time option rather than the default.  My main concern is systems where the owner has set it up for restricted/limited local access.  Think PoS terminal, computer lab etc.  Basically the same rationale applies to this that applies to the logic for some systems asking for a root password for fsck.

But we're not in a position to ask for root, so I think we need to make this part reliable :smile:   And stay locked by default, but I see the value in this for debugging.

cgwalters:
Tested this locally by doing a build/rpm-ostree-compose and rebasing a [CAHC](https://wiki.centos.org/SpecialInterestGroup/Atomic/Devel) vagrant box to it.  All seems good, thanks a lot!

Happy to merge this now after resolving the shell question above, then we can do further PRs for things like static linking?

wmanley:
> Happy to merge this now after resolving the shell question above, then we can do further PRs for things like static linking?

Sounds good to me.

src/switchroot/ostree-prepare-root.c
wmanley: > I think we need to make this part reliable :smile: And stay locked by default

I see your point, but I think having to rebuild might be a little inconvenient if things have already gone wrong.  How about a kernel command line option to enable this?  Something like `ostree.shell-on-boot-failure=1`?  Perhaps something a little less verbose?

wmanley: Ah, but that would fail if we can't get to the point of mounting `/proc`.

cgwalters: I could imagine rather than offering an interactive shell, we do an "info dump" of things like the mount points and existing files.  I've always wished the kernel would show me `lsblk -o NAME,UUID` when failing to mount the rootfs for example - and I suspect that type of thing happens more often than this bit of ostree will break.

So it feels to me like the simplest thing for now is to rebase and drop this patch out - if you want we can keep it as a separate PR?

wmanley: Agreed.  I'll squash, drop this patch and force-push.

cgwalters:
Great, thanks for these patches!  Looking forward to iterating from here - I may get some time this week to look at #268 

@rh-atomic-bot r+ f3f80d8

gatispaeglis:
> I've not yet had a play around with the bootloader integration code and I'll want to do a bunch of experimenting to explore the problem space before implementing a proper solution.

@wmanley This is what works well for me. Let me know if you agree with this once your are done with your experimenting. I can do a PR for this, unless you disagree?

From fbfe0e7cae34d192e8c3739bea846ba005f55e48 Mon Sep 17 00:00:00 2001
From: Gatis Paeglis <gatis.paeglis@qt.io>
Date: Thu, 4 Aug 2016 14:11:10 +0200
Subject: [PATCH] Support for booting without initramfs

Upstream status: WIP
---
 src/libostree/ostree-sysroot-deploy.c | 17 ++++++++++++++---
 1 file changed, 14 insertions(+), 3 deletions(-)

diff --git a/src/libostree/ostree-sysroot-deploy.c b/src/libostree/ostree-sysroot-deploy.c
index 913c6e3..58413e2 100644
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1526,20 +1526,31 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
     ostree_bootconfig_parser_set (bootconfig, "linux", boot_relpath);
   }
 
+  val = ostree_bootconfig_parser_get (bootconfig, "options");
+  kargs = _ostree_kernel_args_from_string (val);
+
   if (dest_initramfs_name)
     {
       g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", dest_initramfs_name, NULL);
     }
-
+  else
+    {
+      g_autofree char *ostree_prepare_root = NULL;
+      /* A proper fix would be to add a define -DLIBSBINDIR='"$(libsbindir)"' and use it here */
+      ostree_prepare_root = g_strdup_printf ("init=/ostree/boot.%d/%s/%s/%d/usr/sbin/ostree-prepare-root /",
+                                             new_bootversion, osname, bootcsum,
+                                             ostree_deployment_get_bootserial (deployment));
+      _ostree_kernel_args_replace_take (kargs, ostree_prepare_root);
+      ostree_prepare_root = NULL;
+    }
 
   ostree_kernel_arg = g_strdup_printf ("ostree=/ostree/boot.%d/%s/%s/%d",
                                        new_bootversion, osname, bootcsum,
                                        ostree_deployment_get_bootserial (deployment));
-  kargs = _ostree_kernel_args_from_string (val);
   _ostree_kernel_args_replace_take (kargs, ostree_kernel_arg);
   ostree_kernel_arg = NULL;
+
   options_key = _ostree_kernel_args_to_string (kargs);
   ostree_bootconfig_parser_set (bootconfig, "options", options_key);
   
-- 
2.7.4

wmanley:
> @wmanley This is what works well for me. Let me know if you agree with this once your are done with your experimenting. I can do a PR for this, unless you disagree?

Go ahead and raise a PR.  I think it would be better to discuss this on it's own PR rather than this one :).

This particular patch won't work for me because of the hard-coded path to `/usr/sbin/ostree-prepare-root`, as you mention in your comment.  I have `ostree-prepare-root` installed into `/opt/ostree/usr/sbin/ostree-prepare-root`.  The tricky thing here is that using the value of `LIBSBINDIR` from the current ostree installation may be inappropriate: The image you're deploying may have `ostree-prepare-root` installed into a different location.

One fix might be to require that `ostree-prepare-root` exist at `$root/usr/sbin/ostree-prepare-root` as you have done here.  Another would be to install `ostree-prepare-root` into `/boot` at deploy time.  The former is certainly simpler.

I've built ostree and its deps into `/opt` because the Ubuntu 14.04 based system I'm trying to use has a glib that's too old to build ostree against.  Originally I tried to build ostree as a static binary but gave up on that after getting weird errors, but I didn't try very hard.

gatispaeglis:
> Another would be to install ostree-prepare-root into /boot at deploy time. The former is certainly simpler.

Another downside with this approach would be that whenever ostree-prepare-root is updated, a new /boot/ostree/$os-$bootcsum would need to be created.

Maybe a symlink in /usr/lib/ostree-boot/ostree-prepare-root -> path_to_ostree-prepare-root could work better. Have to think about a better solution, then I can open a new PR.

gatispaeglis:
https://github.com/ostreedev/ostree/pull/442

PR #406, opened 2016-07-22
[merged] libglnx: bump to latest

jlebon:
OSTree can sometimes print very long lines which lead to many empty
spaces getting printed if the output overflowed the line due to a bug in
libglnx.

Reported-by: Gatis Paeglis <gatis.paeglis@qt.io>


cgwalters:
Note https://github.com/ostreedev/ostree/pull/369 will conflict and is still outstanding.

@rh-atomic-bot r+ e301f68

PR #408, opened 2016-07-22
[merged] pull: Don't execute static deltas when mirroring

cgwalters:
We don't presently support this, since the static delta code assumes
it can just `mmap()` file objects.  We could at some point implement
this, but for now just skip executing deltas when doing
`archive -> archive` mirroring.

I noticed this when trying to mirror a repo in Jenkins in
[CAHC](https://wiki.centos.org/SpecialInterestGroup/Atomic/Devel).

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2705,7 +2705,8 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
                                     &from_revision, error))
         goto out;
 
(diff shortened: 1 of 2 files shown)


giuseppe:
@rh-atomic-bot r+ 584b9d9

alexlarsson:
Hmm, why exactly did this break? We use mirroring pull to a repo when pulling to a local user-owned repo in flatpak, which then will be pulled into the systemwide repo. This used to work fine, but now system installs in flatpak never use deltas...

PR #409, opened 2016-07-23
commit_meta refactoring

Mathnerd314:
This makes the codepath for commit_meta similar to all the other object types. IIRC it's a relatively self-contained change, compared to the others.

I also renamed `enqueue_object` to `fetch_object`, as it's more accurate.


src/libostree/ostree-repo-pull.c
```diff
@@ -1534,11 +1498,13 @@ process_one_static_delta_fallback (OtPullData   *pull_data,
...
           if (!g_hash_table_lookup (pull_data->requested_metadata, checksum))
             {
-              gboolean do_fetch_detached;
               g_hash_table_insert (pull_data->requested_metadata, checksum, checksum);
               
-              do_fetch_detached = (objtype == OSTREE_OBJECT_TYPE_COMMIT);
-              enqueue_one_object_request (pull_data, checksum, objtype, do_fetch_detached, FALSE);
+              /* Fetch metadata */
+              if (objtype == OSTREE_OBJECT_TYPE_COMMIT)
+                fetch_object (pull_data, g_strdup (checksum), OSTREE_OBJECT_TYPE_COMMIT_META);
```
cgwalters: Why the strdup here?  Isn't it a leak?

src/libostree/ostree-repo-pull.c
```diff
@@ -887,44 +875,25 @@ meta_fetch_on_complete (GObject           *object,
...
 
-  if (fetch_data->is_detached_meta)
-    {
-      if (!ot_util_variant_map_fd (fd, 0, G_VARIANT_TYPE ("a{sv}"),
-                                   FALSE, &metadata, error))
-        goto out;
-
-      /* Now delete it, see comment in corresponding content fetch path */
-      (void) unlinkat (_ostree_fetcher_get_dfd (fetcher), temp_path, 0);
+  if (!ot_util_variant_map_fd (fd, 0, ostree_metadata_variant_type (objtype),
```
cgwalters: Hm, it looks like we're unconditionally `mmap()ing` all objects?  That doesn't seem right.

src/libostree/ostree-repo-pull.c
```diff
@@ -840,35 +832,31 @@ meta_fetch_on_complete (GObject           *object,
...
+                pull_data->maxdepth != 0)
             {
               g_clear_error (&local_error);
               /* If the remote repo supports tombstone commits, check if the commit was intentionally
                  deleted.  */
               if (pull_data->has_tombstone_commits)
                 {
-                  enqueue_one_object_request (pull_data, checksum, OSTREE_OBJECT_TYPE_TOMBSTONE_COMMIT,
-                                              FALSE, FALSE);
+                  fetch_object (pull_data, g_strdup (checksum), OSTREE_OBJECT_TYPE_TOMBSTONE_COMMIT);
```
cgwalters: Same leak question here.

src/libostree/ostree-repo-pull.c
```diff
@@ -117,12 +117,6 @@ typedef struct {
 typedef struct {
   OtPullData  *pull_data;
   GVariant    *object;
-  gboolean     is_detached_meta;
-
-  /* Only relevant when is_detached_meta is TRUE.  Controls
-   * whether to fetch the primary object after fetching its
-   * detached metadata (no need if it's already stored). */
-  gboolean     object_is_stored;
```
cgwalters: By deleting this, it looks like we always redownload the commit object too now?  That's not the end of the world, but it does mean another HTTP request by default for "unchanged" state.  Can we still keep this?  Maybe call it `have_attached_object`?

src/libostree/ostree-repo-pull.c
cgwalters: Oh nevermind, I see this is just the metadata handler.

cgwalters:
Thanks for the patch! Looks like this is headed in the right direction and will be a good cleanup.

src/libostree/ostree-repo-pull.c
Mathnerd314: See the comment: "Only relevant when is_detached_meta is TRUE.". There were three combinations:
* is_detached_meta = FALSE: corresponds to a fetch of type COMMIT
* is_detached_meta = TRUE, object_is_stored = FALSE: corresponds to one fetch of type COMMIT and one of type COMMIT_META
* is_detached_meta = TRUE, object_is_stored = TRUE: corresponds to a fetch of type COMMIT_META

The code in scan_one_metadata_object_c prevents downloading commits multiple times, but commit_meta is always fetched. (There should probably be a test for that)

src/libostree/ostree-repo-pull.c
Mathnerd314: It's only a leak if fetch_object is not transfer-full. I have added a patch to make that so.

cgwalters: `(transfer full)` input functions are pretty unusual; I can't think of any other ones in the ostree source, i.e. `git grep '(transfer full)' |grep -v Returns |grep -v '(out)'` is empty.  There are some in GLib but still quite rare.

Why not just keep the memory owned by the hash table like it was originally and drop the strdup?

cgwalters: Does this make sense?  Anything blocking updating this patch?

PR #410, opened 2016-07-24
[merged] Valgrind admin deploy

cgwalters:
Decided to try `env OT_TESTS_VALGRIND=1 make -j 4 check TESTS=tests/test-admin-deploy-syslinux.sh`; there is more to do here, but this is most of it.


giuseppe:
@rh-atomic-bot r+

PR #411, opened 2016-07-24
Removing unused variables

ericcurtin:


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2135,8 +2131,7 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
         goto out;
     }
 
-  { ostree_cleanup_sepolicy_fscreatecon gpointer dummy = NULL;
-
+  {
```
cgwalters: This *is* used to unset the context state.   We likely need to annotate `G_GNUC_USED` or so.

cgwalters:
I'm guessing you're building with CLang?  This is likely going to conflict with some outstanding patches, but we can deal.

Can you [improve the commit message](http://chris.beams.io/posts/git-commit/) ?  See the existing bits in `git log`.  Something like this as a start:
```
build: Remove unused variables

I noticed this when building with CLang.
```

ericcurtin:
Done :)

cgwalters:
@rh-atomic-bot r+ 8cddf98

cgwalters:
@ericcurtin You're going to need to do a build with all options enabled for this (selinux, libarchive, etc.).

ericcurtin:
Ah yes I see there are things #ifdef'd out... I'll get to it this evening

cgwalters:
Fixed by https://github.com/ostreedev/ostree/pull/548

PR #412, opened 2016-07-26
Add introspectable versions of libarchive APIs

giuseppe:


src/libostree/ostree-repo.h
```diff
@@ -735,7 +735,7 @@ ostree_repo_checkout_tree (OstreeRepo               *self,
                            GError                  **error);
 
 /**
- * OstreeRepoCheckoutOptions:
+ * OstreeRepoCheckoutOptions: (skip)
```
cgwalters: Hmm...I have a worry that someone could have actually been using this *without* the bitfield parts.  But OK.

giuseppe: also  `subpath` and `devino_to_csum_cache` are not valid, as their offset is not correct.  Only `mode` and `overwrite_mode` could be used correctly.

src/libostree/ostree-repo-checkout.c
```diff
@@ -890,6 +890,39 @@ ostree_repo_checkout_tree_at (OstreeRepo                         *self,
...
+ * ostree_repo_checkout_tree_at_introspectable:
+ * @self: Repo
+ * @options: (allow-none): Options
+ * @destination_dfd: Directory FD for destination
+ * @destination_path: Directory for destination
+ * @commit: Checksum for commit
+ * @cancellable: Cancellable
+ * @error: Error
+ *
+ * Introspectable version of ostree_repo_checkout_tree_at.
```
cgwalters: Eeek.  That's a lot of new duplicate API to maintain.  I think if we're going to go this route we need to basically deprecate the earlier versions.  One way to do this would be to create e.g. `ostree-repo-deprecated.h` (still `#include'd` by `ostree-repo.h`).

Then we need to bikeshed on the name a bit...if this is the new canonical API then let's call it something other than "introspectable".  I'm blanking a bit...maybe just go all Microsoft and call it `ostree_repo_checkout_tree2`?  Or perhaps just `ostree_repo_checkout()`?

giuseppe: What about `ostree_repo_checkout_at`?

giuseppe:
I have worked on a different version, I also marked the existing functions as deprecated.

I tested separately the introspection as:
```
options = OSTree.RepoCheckoutAtOptions()
options.overwrite_mode = OSTree.RepoCheckoutOverwriteMode.UNION_FILES
options.process_whiteouts = True
options.disable_fsync = True
repo.checkout_at(options, rootfs_fd, rootfs, rev)
```

src/libostree/ostree-repo.h
```diff
@@ -659,6 +683,27 @@ gboolean ostree_repo_export_tree_to_archive (OstreeRepo                *self,
...
+  gboolean disable_xattrs;
+  guint64 timestamp_secs;
+  char *path_prefix;
+} OstreeRepoExportArchiveOptions2;
+
+_OSTREE_PUBLIC
+gboolean ostree_repo_export_tree_to_archive2 (OstreeRepo                *self,
+                                             OstreeRepoExportArchiveOptions2  *opts,
+                                             OstreeRepoFile            *root,
+                                             void                      *archive,  /* Really struct archive * */
```
cgwalters: Same issue here.  In general we're probably going to need something like `int fd` + "some form of archive options" to be introspectable.

cgwalters:
Thanks, this is looking like it's on the right track!  You could consider splitting out the "checkout" + deprecation as a separate PR to land now, then we can revisit how to make APIs that take `struct archive*` be introspectable on top of that?

giuseppe:
Thanks for the review.  I made a separate PR for the checkout part only.  I didn't mark the other two functions as deprecated, since there is no alternative yet, but I added a patch to hide them introspection

https://github.com/ostreedev/ostree/pull/417

PR #416, opened 2016-07-29
[merged] tests/delta-crosscheck: Add missing --from

cgwalters:
Otherwise we generate a static delta from the *previous* and get
confused.  To make doubly sure, add `--require-static-deltas` to pull.


jlebon:
@rh-atomic-bot r+ 6dff9e5

PR #418, opened 2016-07-29
[merged] Static delta minor fixes

jlebon:


cgwalters:
@rh-atomic-bot r+ e847446

Issue #419, opened 2016-07-29
Static deltas with no checksums


dbnicholson:
I was experimenting with static delta management and made a delta between 2 commits that were exactly the same but different revisions. The delta was created successfully, but later `_ostree_static_delta_parse_checksum_array` returned an error because there were no checksums.
```
$ ostree static-delta show 5824a0ce049681f17d4c7a40c51b8d1338b8d5131f79f29157066cf89f53b9bd-9617d9c15ba643ccf566b2cdf793d46f3b9bc42d392b1ac2940e9311e232e2c8
...
Delta: 5824a0ce049681f17d4c7a40c51b8d1338b8d5131f79f29157066cf89f53b9bd-9617d9c15ba643ccf566b2cdf793d46f3b9bc42d392b1ac2940e9311e232e2c8
Endianness: little
Timestamp: 1469740860
Number of parents: 0
Number of fallback entries: 0
Total Fallback Size: 0 (0 bytes)
Total Fallback Uncompressed Size: 0 (0 bytes)
Number of parts: 1
PartMeta0: nobjects=0 size=61 usize=0
PartPayload0: nmodes=0 nxattrs=0 blobsize=0 opsize=0
error: Invalid checksum array length 0
```
Should ostree refuse to generate these empty deltas or should the check be relaxed? It looks like `_ostree_static_delta_part_execute` has an assertion that there are more than 0 checksums, although `_ostree_repo_static_delta_part_have_all_objects` looks like it would succeed.

cgwalters:
I think I'd vote to:

 - Accept empty deltas since they may have been generated before
 - Refuse to generate new ones ?

dbnicholson:
That does sound like the best plan, but I worry about refusing to generate them now that I think about it. If that's going to become an error, then the caller has to somehow check ahead of time if the contents of the commits are the same. That seems annoying, and I'm not sure how you'd do it.

So, I think generating and accepting these useless deltas is probably the safest thing to do for users. It doesn't really hurt anything.

cgwalters:
Closed by https://github.com/ostreedev/ostree/pull/420

PR #420, opened 2016-07-30
[merged] Allow deltas with empty parts

dbnicholson:
OSTree will create deltas that have no objects if the 2 commits have no change in content. However, it currently refuses to process these valid deltas.

https://github.com/ostreedev/ostree/issues/419

### DIFF
--- a/tests/test-delta.sh
+++ b/tests/test-delta.sh
@@ -26,7 +26,7 @@ skip_without_user_xattrs
 bindatafiles="bash true ostree"
 morebindatafiles="false ls"
 
-echo '1..8'
+echo '1..10'
 
 mkdir repo
 ${CMD_PREFIX} ostree --repo=repo init --mode=archive-z2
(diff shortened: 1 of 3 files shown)


tests/test-delta.sh
```diff
@@ -208,3 +208,26 @@ ${CMD_PREFIX} ostree --repo=repo static-delta list | grep ^${origrev}-${newrev}$
...
+${CMD_PREFIX} ostree --repo=repo commit -b test -s test --tree=dir=files
+samerev=$(${CMD_PREFIX} ostree --repo=repo rev-parse test)
+${CMD_PREFIX} ostree --repo=repo static-delta generate --from=${newrev} --to=${samerev}
+${CMD_PREFIX} ostree --repo=repo static-delta show ${newrev}-${samerev} > show-empty.txt
+nparts_empty=$(grep '^Number of parts:' show-empty.txt)
+part0_meta_empty=$(grep '^PartMeta0:' show-empty.txt)
+totalsize_empty=$(grep '^Total Uncompressed Size:' show-empty.txt)
+assert_streq "${nparts_empty}" "Number of parts: 1"
+assert_str_match "${part0_meta_empty}" "nobjects=0"
+assert_streq "${totalsize_empty}" "Total Uncompressed Size: 0 (0 bytes)"
```
cgwalters: Hm, I'm realizing this is a minor lie because we're not accounting :ledger: for the inline commit object.  But that's something to fix later.

cgwalters:
@rh-atomic-bot r+ f391b19

tests/test-delta.sh
dbnicholson: Makes sense. The first two checks are sufficient, I think, and I only added this check later. It can be removed if it becomes a problem.

PR #422, opened 2016-08-01
[merged] repo: Make ostree_repo_create() nonfatal on existing repos

cgwalters:
In general we want to support "idempotentcy" or "state
synchronization" across interruption.  If a repo is only partially
created due to a crash or whatever, it's hard for a user to know that.
Let's just make `ostree_repo_create()` idempotent. Since all we're
doing is a set of `mkdirat()` invocations, it's quite simple.

This also involved porting to fd-relative, which IMO makes the
code a lot clearer.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -1842,8 +1842,15 @@ ostree_repo_mode_from_string (const char      *mode,
  * @cancellable: Cancellable
  * @error: Error
  *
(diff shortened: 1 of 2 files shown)


cgwalters:
@rh-atomic-bot delegate=14rcole

14rcole:
@rh-atomic-bot r+ 6c6b9ef

PR #423, opened 2016-08-01
[merged] Today's misc glnx porting patches

cgwalters:
We're getting fairly close, splitting these smaller ones out from a larger series I am working on.


giuseppe:
@rh-atomic-bot r+ 2607bb5

PR #425, opened 2016-08-03
[merged] repo: Flip the fsync default to off for new checkout API

cgwalters:
Since we're adding a new API, we have the opportunity to fix
the defaults.  We expect clients to do a `syncfs()` or equivalent
on their own now, since it's way more efficient.

Flip the checkout fsync default to off.


giuseppe:
@rh-atomic-bot r+

Issue #421, opened 2016-08-01
EBADF during static delta apply


cgwalters:
https://mail.gnome.org/archives/ostree-list/2016-July/msg00020.html

```
openat(3, "refs/remotes", O_RDONLY|O_NOCTTY|O_NONBLOCK|O_DIRECTORY|O_CLOEXEC) = 15
newfstatat(15, "qt-os", {st_mode=S_IFDIR|0755, st_size=4096, ...}, AT_SYMLINK_NOFOLLOW) = 0
openat(15, "qt-os", O_RDONLY|O_NOCTTY|O_NONBLOCK|O_DIRECTORY|O_CLOEXEC) = 19
close(15)                               = 0
newfstatat(19, "linux", {st_mode=S_IFDIR|0755, st_size=4096, ...}, AT_SYMLINK_NOFOLLOW) = 0
openat(19, "linux", O_WRONLY|O_DIRECTORY|O_CLOEXEC|0x400000) = 15
fallocate(15, 0, 0, 65)                 = -1 EOPNOTSUPP (Operation not supported)
fstat(15, {st_mode=S_IFREG, st_size=0, ...}) = 0
fstatfs(15, {f_type="EXT2_SUPER_MAGIC", f_bsize=4096, f_blocks=237798, f_bfree=44086, f_bavail=31798, f_files=61440, f_ffree=37789, f_fsid={297229479, -412021672}, f_namelen=255, f_frsize=4096}) = 0
pwrite(15, "\0", 1, 64)                 = 1
write(15, "6962c73eeebeedccb7fe05cdc57d5309"..., 65) = 65
newfstatat(19, "linux/qt", 0x7ffe6ea4b040, AT_SYMLINK_NOFOLLOW) = -1 EBADF (Bad file descriptor)
close(15)                               = 0
close(19)                               = 0
openat(8, ".", O_RDONLY|O_NOCTTY|O_NONBLOCK|O_DIRECTORY|O_CLOEXEC) = 15
fstat(15, {st_mode=S_IFDIR|0755, st_size=4096, ...}) = 0
fcntl(15, F_GETFL)                      = 0x18800 (flags O_RDONLY|O_NONBLOCK|O_LARGEFILE|O_DIRECTORY)
fcntl(15, F_SETFD, FD_CLOEXEC)          = 0
getdents(15, /* 8 entries */, 32768)    = 400
[...]

PR #426, opened 2016-08-03
[merged] Update libglnx for O_TMPFILE build time option

cgwalters:
See: https://github.com/ostreedev/ostree/issues/421

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -320,6 +320,7 @@ echo "
     libmount:                                     $with_libmount
     libarchive (parse tar files directly):        $with_libarchive
     static deltas:                                yes (always enabled now)
+    O_TMPFILE:                                    $enable_otmpfile
     man pages (xsltproc):                         $enable_man
     api docs (gtk-doc):                           $enable_gtk_doc
     gjs-based tests:                              $have_gjs


cgwalters:

gatispaeglis:
If I run ./autogen.sh, it does not contain "yes/no" for O_TMPFILE:

```
OSTree 2016.7
    ===============


    introspection:                                yes
    rofiles-fuse:                                 yes
    libsoup (retrieve remote HTTP repositories):  yes
    libsoup TLS client certs:                     yes
    SELinux:                                      yes
    systemd:                                      yes
    libmount:                                     yes
    libarchive (parse tar files directly):        yes
    static deltas:                                yes (always enabled now)
    O_TMPFILE:                                    
    man pages (xsltproc):                         yes
    api docs (gtk-doc):                           no
    gjs-based tests:                              yes
    dracut:                                       no
    mkinitcpio:                                   no
    grub2-mkconfig path:                          grub2-mkconfig
```

cgwalters:
Did you `git submodule update --init` ?

gatispaeglis:
I did:

git clone git@github.com:ostreedev/ostree.git 
cd ostree
curl https://patch-diff.githubusercontent.com/raw/ostreedev/ostree/pull/426.patch | git apply - 
./autogen.sh

Then git submodule update --init is not necessary.

cgwalters:
It looks like `git apply -` isn't updating the submodule for some reason.  Also I think things are a little strange if you do things in that order from a fresh clone, because `autogen.sh` will itself do `git submodule update --init`.  But we really want the submodule initialized before.

This works for me:

```
git clone git@github.com:ostreedev/ostree.git && cd ostree   # optional - you can use an existing checkout
git pull https://github.com/cgwalters/ostree.git update-libglnx-otmpfile
git submodule update --init
```

cgwalters:
In general with submodules I often use `git submodule status` and verify the commits are what I expect them to be.

gatispaeglis:
Ah, ok. Now it works :)

@rh-atomic-bot r+ aa0dbf7

PR #427, opened 2016-08-04
[merged] lib: Add padding booleans to OstreeRepoCheckoutAtOptions

cgwalters:
Since this is a new API, and adding booleans is the most likely thing
we'll do, let's stick some explicit padding for them in here now.

We could use the `unused_ints[]` but it'd be out of order, and this
will more clearly remind people about the padding.  The efficiency hit
versus bitfields is annoying, but oh well, not a real world problem.


giuseppe:
@rh-atomic-bot r+ d30a3c9

PR #429, opened 2016-08-04
[merged] lib: Use libglnx file replace API more consistently

cgwalters:
We have a better API now, drop use of the internal helper, which also
depended on libgsystem.

This required bumping libglnx to pull in a fix.


giuseppe:
@rh-atomic-bot r+ a6e1cf9

PR #430, opened 2016-08-04
[merged] lib: Fix a compiler warning introduced from earlier patch

cgwalters:
I think we'd end up closing stdin...yuck.  Need to investigate
having this be fatal, but CentOS 7 `gcc-4.8.5` doesn't understand
`-Werror=int-conversion`.


giuseppe:
@rh-atomic-bot r+ 85ebd9a

PR #433, opened 2016-08-05
[merged] repo: Drop more internally unused GFile members

cgwalters:
I forgot to actually remove `config_file` in the previous
commit, the txn lock hasn't been used in a long time, and
for the uncompressed cache, everything uses the fd already.


giuseppe:
@rh-atomic-bot r+ c00b282

Issue #415, opened 2016-07-28
libarchive: make it easy to canonicalize perms


cgwalters:
- Directories writable by user
 - No world writable files

anything else?

cgwalters:
https://github.com/projectatomic/rpm-ostree/blob/dee758e111318229915bda6abeebb6539b3c9c21/src/libpriv/rpmostree-unpacker.c#L505

PR #434, opened 2016-08-05
checkout: force writable directories with --user

giuseppe:
Otherwise an union checkout cannot create new files

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -651,7 +651,9 @@ checkout_tree_at (OstreeRepo                        *self,
   while (G_UNLIKELY (res == -1 && errno == EINTR));
   if (res == -1)
     {
-      if (errno == EEXIST && options->overwrite_mode == OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES)
+      if (errno == EEXIST
+          && options->overwrite_mode == OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES
+          && options->mode != OSTREE_REPO_CHECKOUT_MODE_USER)
         did_exist = TRUE;
       else
         {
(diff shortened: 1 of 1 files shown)


cgwalters:
See https://github.com/ostreedev/ostree/issues/415
Currently it's possible to override on import with a filter function. Maybe something like:

```
diff --git a/src/libostree/ostree-repo.h b/src/libostree/ostree-repo.h
index ff1e712..f4f457d 100644
--- a/src/libostree/ostree-repo.h
+++ b/src/libostree/ostree-repo.h
@@ -531,11 +531,13 @@ typedef OstreeRepoCommitFilterResult (*OstreeRepoCommitFilter) (OstreeRepo    *r
  * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_NONE: No special flags
  * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_SKIP_XATTRS: Do not process extended attributes
  * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_GENERATE_SIZES: Generate size information.
+ * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICALIZE_PERMS: Ensure directories are writable by user, avoid world-writable files
  */
 typedef enum {
   OSTREE_REPO_COMMIT_MODIFIER_FLAGS_NONE = 0,
   OSTREE_REPO_COMMIT_MODIFIER_FLAGS_SKIP_XATTRS = (1 << 0),
-  OSTREE_REPO_COMMIT_MODIFIER_FLAGS_GENERATE_SIZES = (1 << 1)
+  OSTREE_REPO_COMMIT_MODIFIER_FLAGS_GENERATE_SIZES = (1 << 1),
+  OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICALIZE_PERMS = (1 << 2)
 } OstreeRepoCommitModifierFlags;
 
 /**
```
?

giuseppe:
will that require a different commit/metadata for an user and a system image?  I thought about fixing the permissions here so that it will be possible to use the same data that is used for system images.

cgwalters:
I'd suggest we also change system images.

giuseppe:
something like this: https://github.com/projectatomic/atomic/pull/517 ?

I have tested it quickly here and it works for user checkouts as well

cgwalters:
The reason I'm driving this towards the generalized "canonicalize permissions" issue is I think we also need to avoid world-writable files for example.  The only use case for a world-writable directory is `/tmp` with the sticky bit, and I'd suggest there in user mode checkouts we disable that.

(Really in general for containers we should be redirecting `/tmp` to e.g. `/tmp/containers/$ID or so)

cgwalters:
Since that PR works, let's close this and possibly revisit #415 in another PR.

PR #436, opened 2016-08-05
[merged] Update libglnx: Add missing files to libglnx distribution

smcv:
This fixes distcheck.

PR #437, opened 2016-08-05
[merged]  Skip tests that use whiteouts under Docker/aufs 

smcv:
ostree's naming convention for whiteouts is similar to what is
done in aufs, which means we can't compose the trees to test this
feature under Docker with the aufs storage driver, as used on
travis-ci.

---

This might need rebasing to not be based on PR #436 but for now I'm pushing it just to have something visible.

### DIFF
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -452,38 +452,41 @@ echo "ok test error pre commit/bootid"
 # Whiteouts
 cd ${test_tmpdir}
 mkdir -p overlay/baz/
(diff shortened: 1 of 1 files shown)


cgwalters:
Rebase :surfer:  and drop the earlier patch please.  Also, :cry: at the need for this; see also https://github.com/opencontainers/image-spec/issues/24

cgwalters:
@rh-atomic-bot r+ 00b465c

PR #438, opened 2016-08-05
[merged] Run tests on travis-ci

smcv:
This is not suitable for merge yet, because it disables Colin's
Homu integration, so that I can run it against arbitrary branches.
We'll need to discuss how to deal with that.

The structure is loosely based on what I've done for dbus and OpenJK, with a small .travis.yml driving larger CI-agnostic install and build scripts. The new thing in this version (which I should try in dbus too) is using Docker containers to get environments other than a 2 year old copy of Ubuntu.

---

.travis.yml is obviously still Travis-specific, but tests/ci-* are
designed to be shareable with other CI environments if there is interest
in doing so.

At the moment I'm only testing on Debian and Ubuntu. In principle we
could try a non-Debian-derived Docker container such as Fedora or CentOS
inside travis-ci's Ubuntu environment, similar to what I'm doing
for Debian, but I don't know the correct setup commands to use there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: c
+dist: trusty
+sudo: required
+
+env:
+  - ci_distro=ubuntu ci_suite=trusty
+  - ci_docker=debian:jessie ci_distro=debian ci_suite=jessie
+  - ci_docker=debian:unstable ci_distro=debian ci_suite=unstable
+  - ci_docker=ubuntu:xenial ci_distro=ubuntu ci_suite=xenial
+
+script:
+  - tests/ci-install.sh
+  - ci_parallel=2 ci_sudo=yes tests/ci-build.sh
--- a/tests/ci-install.sh
+++ b/tests/ci-install.sh
@@ -1,0 +1,105 @@
+#!/bin/bash
+
+set -euo pipefail
+set -x
+
+NULL=
+: "${ci_docker:=}"
+: "${ci_in_docker:=}"
+: "${ci_suite:=jessie}"
+
+if [ $(id -u) = 0 ]; then
+    sudo=
+else
+    sudo=sudo
+fi
+
+if [ -n "$ci_docker" ]; then
+    sed \
+        -e "s/@ci_distro@/${ci_distro}/" \
+        -e "s/@ci_docker@/${ci_docker}/" \
+        -e "s/@ci_suite@/${ci_suite}/" \
+        < tests/ci-Dockerfile.in > Dockerfile
+    exec docker build -t ostree-ci .
+fi
+
+case "$ci_distro" in
+    (debian)
+        # Docker images use httpredir.debian.org but it seems to be
+        # unreliable; use a CDN instead
+        sed -i -e 's/httpredir\.debian\.org/deb.debian.org/g' /etc/apt/sources.list
+        ;;
+esac
+
+case "$ci_suite" in
+    (jessie)
+        # Add alexl's Debian 8 backport repository to get libgsystem
+        # TODO: remove this when libgsystem is no longer needed
+        $sudo apt-get -y update
+        $sudo apt-get -y install apt-transport-https wget
+        wget -O - https://sdk.gnome.org/apt/debian/conf/alexl.gpg.key | $sudo apt-key add -
+        echo "deb [arch=amd64] https://sdk.gnome.org/apt/debian/ jessie main" | $sudo tee /etc/apt/sources.list.d/flatpak.list
+        ;;
+
+    (trusty|xenial)
+        # Add alexl's Flatpak PPA, again to get libgsystem
+        # TODO: remove this when libgsystem is no longer needed
+        $sudo apt-get -y update
+        $sudo apt-get -y install software-properties-common
+        $sudo add-apt-repository --yes ppa:alexlarsson/flatpak
+        ;;
+esac
+
+case "$ci_distro" in
+    (debian|ubuntu)
+        # TODO: fetch this list from the Debian packaging git repository?
+        $sudo apt-get -y update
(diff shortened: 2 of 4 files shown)


tests/ci-install.sh
```diff
@@ -0,0 +1,97 @@
...
+        sed -i -e 's/httpredir\.debian\.org/deb.debian.org/g' /etc/apt/sources.list
+        ;;
+esac
+
+case "$ci_suite" in
+    (jessie)
+        $sudo apt-get -y update
+        $sudo apt-get -y install apt-transport-https wget
+        wget -O - https://sdk.gnome.org/apt/debian/conf/alexl.gpg.key | $sudo apt-key add -
+        echo "deb [arch=amd64] https://sdk.gnome.org/apt/debian/ jessie main" | $sudo tee /etc/apt/sources.list.d/flatpak.list
```
smcv: This deserves more comments, but what I'm doing here is adding a convenient source of libgsystem, namely @alexlarsson's repository for Flatpak (which contains OSTree and hence libgsystem). If I get the libgsystem/ostree/flatpak stack into Debian jessie-backports, I can use that instead.

tests/ci-install.sh
```diff
@@ -0,0 +1,97 @@
...
+        ;;
+
+    (trusty|xenial)
+        $sudo apt-get -y update
+        $sudo apt-get -y install software-properties-common
+        $sudo add-apt-repository --yes ppa:alexlarsson/flatpak
```
smcv: Similarly, what I'm doing here is downloading apt-add-repository, then using it to add an apt source where I can get libgsystem.

tests/ci-install.sh
```diff
@@ -0,0 +1,97 @@
...
+
+        if [ -n "$ci_in_docker" ]; then
+            adduser --disabled-password user </dev/null
+            apt-get -y install sudo systemd-sysv
+            echo "user ALL=(ALL) NOPASSWD: ALL" > /etc/sudoers.d/nopasswd
+            chmod 0440 /etc/sudoers.d/nopasswd
+        fi
+        ;;
+
+    (*)
```
smcv: Please add Fedora or CentOS support just above here :-)

smcv:
This branch also contains #436 and #437 in the hope that they are enough to make travis-ci pass.

tests/ci-install.sh
cgwalters: FWIW very shortly the libgsystem dep will be gone; I haven't submitted a PR yet for https://github.com/cgwalters/ostree/commits/glnx-the-final-round but I will after the other pending ones are done.

smcv:
<https://travis-ci.org/smcv/ostree/jobs/150142106> says
```
ERROR: files left in build directory after distclean:
./libswitchroot-mountutil.la
./ostree-remount
./ostree-prepare-root
```

So near and yet so far. Maybe when CI tests this stuff, it will work.

tests/ci-install.sh
smcv: > FWIW very shortly the libgsystem dep will be gone

I look forward to its demise :-)

.travis.yml
```diff
@@ -1,24 +1,25 @@
...
 
-branches:
-  only:
-    - auto
+script:
+  - tests/ci-install.sh
+  - ci_parallel=2 ci_sudo=yes tests/ci-build.sh
+
+# Disabled for now so that getting the tests working in Travis won't
+# interfere with Colin's infrastructure
```
cgwalters: Actually you can nuke this now, we use an internal Red Hat jenkins, and I have some work on using [CentOS CI](https://wiki.centos.org/QaWiki/CI).

tests/ci-install.sh
cgwalters: I am very strongly of the opinion that CI/packaging should be tightly integrated, and one should avoid things like hardcoded lists of package names in CI that duplicate the downstream deps.   So I maintain https://github.com/cgwalters/rpmdistro-gitoverlay which solves that problem for the Fedora/CentOS world.

(Right now our internal CI isn't using it, but I hope to convert)

cgwalters:
I'm OK with this...it won't be gating, but I think we can help maintain it.  For ostree at least we'll likely remain conservative with external dependency versions.

tests/ci-install.sh
cgwalters: (That said I don't have any objections to the dependency list here for Debian)

.travis.yml
smcv: If it's no longer needed, would you mind dropping it from master? I'm not hugely keen on removing functionality in a branch that's just meant to add more test coverage.

> we use an internal Red Hat jenkins

That's fine for Red Hat staff, but if my pull request fails on it, I can't see why or debug it :-)

smcv:
> For ostree at least we'll likely remain conservative with external dependency versions.

If one of the supported distros becomes too old, dropping it is fine - at least that way it will be a deliberate choice, not an accident.

.travis.yml
cgwalters: Like  https://github.com/ostreedev/ostree/pull/439  ?  Feel free to r+

tests/ci-install.sh
smcv: Doing some sort of CI on the Debian packaging by auto-merging the upstream code into it has been vaguely on my to-do list for a while, but I don't think that actually has the same coverage that this does (particularly if we care about the things that distcheck verifies).

smcv:
For now I've disabled distcheck, and I'm just doing: build, check, install, installcheck, installed-tests. I can add distcheck back in later, after checking that it works.

cgwalters:
Thanks for working on this!

@rh-atomic-bot r+ 7d2e0ea

PR #439, opened 2016-08-05
[merged] .travis.yml: Delete, we aren't using it right now

cgwalters:
@smcv will re-add a new version.


smcv:
@rh-atomic-bot r+

PR #442, opened 2016-08-08
Support for booting without initramfs

gatispaeglis:
WIP

I have tested that this patch works fine with u-boot and grub backends.

What is missing here is an automake rule that would create:

`/usr/lib/ostree-boot/ostree-prepare-root`-> `../../sbin/ostree-prepare-root`

or other way around:

`/usr/sbin/ostree-prepare-root`-> `../lib/ostree-boot/ostree-prepare-root`

@cgwalters You mentioned once that you want to move`ostree-prepare-root`and `ostree-remount` from `sbin` to `libexecdir`. I am thinking that maybe `/usr/lib/ostree-boot` is a better location? As this path would be the same on all systems and seems like a more natural place for all these "boot related" files (including `ostree-grub-generator`). Having this well known path would guarantee that this always works:

`init=/ostree/boot.%d/%s/%s/%d/usr/lib/ostree-boot/ostree-prepare-root /`

Related change: https://github.com/ostreedev/ostree/pull/268/commits/54168bc28be5079f5b23f9f996a15d38594ae202#diff-f0c889fb056dc4a713f8a87702fdd27eR1452
When copying `/usr/lib/ostree-boot,` exclude `ostree-{grub-generator, remount, prepare-root}`

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1458,20 +1458,28 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
     ostree_bootconfig_parser_set (bootconfig, "linux", boot_relpath);
   }
 
+  val = ostree_bootconfig_parser_get (bootconfig, "options");
+  kargs = _ostree_kernel_args_from_string (val);
+
   if (dest_initramfs_name)
     {
       g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", dest_initramfs_name, NULL);
       ostree_bootconfig_parser_set (bootconfig, "initrd", boot_relpath);
     }
-
-  val = ostree_bootconfig_parser_get (bootconfig, "options");
+  else
+    {
+      g_autofree char *prepare_root_arg = NULL;
+      prepare_root_arg = g_strdup_printf ("init=/ostree/boot.%d/%s/%s/%d/usr/lib/ostree/ostree-prepare-root",
+                                             new_bootversion, osname, bootcsum,
+                                             ostree_deployment_get_bootserial (deployment));
+      _ostree_kernel_args_replace_take (kargs, g_steal_pointer (&prepare_root_arg));
+    }
 
   ostree_kernel_arg = g_strdup_printf ("ostree=/ostree/boot.%d/%s/%s/%d",
                                        new_bootversion, osname, bootcsum,
                                        ostree_deployment_get_bootserial (deployment));
-  kargs = _ostree_kernel_args_from_string (val);
-  _ostree_kernel_args_replace_take (kargs, ostree_kernel_arg);
-  ostree_kernel_arg = NULL;
+  _ostree_kernel_args_replace_take (kargs, g_steal_pointer (&ostree_kernel_arg));
+
   options_key = _ostree_kernel_args_to_string (kargs);
   ostree_bootconfig_parser_set (bootconfig, "options", options_key);
   
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -90,6 +90,7 @@ dist_test_scripts = \
 	tests/test-switchroot.sh \
 	tests/test-pull-contenturl.sh \
 	tests/test-pull-mirrorlist.sh \
+	tests/test-no-initramfs.sh \
 	$(NULL)
 
 if BUILDOPT_FUSE
--- a/src/boot/grub2/ostree-grub-generator
+++ b/src/boot/grub2/ostree-grub-generator
@@ -1,7 +1,6 @@
 #!/bin/sh
 
-# To use a custrom script for generating grub.cfg, set the --with-grub2-mkconfig=no
-# configure switch when configuring and building OSTree.
+# The builtin grub.cfg generator.
 #
 # This script is called by ostree/src/libostree/ostree-bootloader-grub2.c whenever
 # boot loader configuration file needs to be updated. It can be used as a template
--- a/src/boot/grub2/ostree-grub-generator
+++ b/src/boot/grub2/ostree-grub-generator
@@ -28,7 +28,7 @@ entries_path=$(dirname $new_grub2_cfg)/entries
 
 read_config()
 {
-    config_file=${entries_path}/${1}
+    config_file=${1}
     title=""
     initrd=""
     options=""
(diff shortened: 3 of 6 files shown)


jlebon:
bot, retest this please

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1458,20 +1458,30 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
-
-  val = ostree_bootconfig_parser_get (bootconfig, "options");
+  else
+    {
+      g_autofree char *ostree_prepare_root = NULL;
+      ostree_prepare_root = g_strdup_printf ("init=/ostree/boot.%d/%s/%s/%d/usr/lib/ostree-boot/ostree-prepare-root /",
+                                             new_bootversion, osname, bootcsum,
+                                             ostree_deployment_get_bootserial (deployment));
+      _ostree_kernel_args_replace_take (kargs, ostree_prepare_root);
+      ostree_prepare_root = NULL;
```
cgwalters: Minor but this can now be: `_ostree_kernel_args_replace_take (kargs, g_steal_pointer (&ostree_prepare_root));`

cgwalters:
I *think* it'd be fine to just move the binaries unconditionally - systemd execs them via the `.service` files so the indirection should be OK.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1458,20 +1458,30 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
     {
       g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", dest_initramfs_name, NULL);
       ostree_bootconfig_parser_set (bootconfig, "initrd", boot_relpath);
     }
-
-  val = ostree_bootconfig_parser_get (bootconfig, "options");
+  else
+    {
+      g_autofree char *ostree_prepare_root = NULL;
+      ostree_prepare_root = g_strdup_printf ("init=/ostree/boot.%d/%s/%s/%d/usr/lib/ostree-boot/ostree-prepare-root /",
```
cgwalters: This will only work though if statically linked...which makes me think this should be part of an `#ifdef BUILDOPT_NO_INITRD` or something.  (And then fatally error out if we *do* find an initrd in that case?)

cgwalters:
Just some minor comments.  I think the next important step here is adding the build machinery to statically link the switchroot binary, right?

src/libostree/ostree-sysroot-deploy.c
gatispaeglis: For a better backward compatibility I suggest: `#ifdef  BUILDOPT_STATIC_PREPARE_ROOT`

Now when ostree is build with static `ostree-prepare-root` users

- can still use initramfs code path, if `initramfs-*` found (by using the static version inside of initramfs)
- when `initramfs-*` not found, append` init=` kernel arg.

and if there are any users out there that were using OSTree without initramfs already (by hacking kernel sources to include the ostree-prepare-root logic), they can continue doing that by not compiling `ostree-prepare-root` statically (`BUILDOPT_STATIC_PREPARE_ROOT` not defined.)

gatispaeglis:
>  I think the next important step here is adding the build machinery to statically link the switchroot binary, right?

@wmanley any success with the 'musl' patch?

src/libostree/ostree-sysroot-deploy.c
cgwalters: Let's set aside the prepare-root-in-kernel case since I'm having enough trouble keeping the support matrix in my head :smile:  and I doubt anyone is doing it.  The overhead of an additional exec in the startup path is small enough, and avoids having to carry kernel patches, etc.

So, I wouldn't disagree that statically linked users can still use an initramfs, but it'd be odd to have a static switchroot inside an initramfs.  I guess if it's small enough, I could imagine "general purpose" distros like Debian/CentOS always building it statically and allowing people to make initrd-less systems that use ostree, even if it's not going to be the default.  In my world (CentOS/Fedora) though we're not really focused on the embedded cases, so we basically always have an initrd.

Anyways, I'm ok with renaming the build option to `#ifdef BUILDOPT_STATIC_PREPARE_ROOT` which is I guess what all this boils down to :smile: 

So...aside from minor bits, are you thinking we merge this patch as is, or is it more of an RFC and we'll update it with build options later?

gatispaeglis:
> Let's set aside the prepare-root-in-kernel case since I'm having enough trouble keeping the support matrix in my head

Good :)

> So...aside from minor bits, are you thinking we merge this patch as is, or is it more of an RFC and we'll update it with build options later?

I pushed it here for the discussion purposes. I want to update this patch to include the configure switch: `--with-static-prepare-root`. Only my autotools knowledge is limited, so I will keep it simple. The "musl" version can be added as an improvement later on.

wmanley:
> @wmanley any success with the 'musl' patch?

Not yet.  I had patch to build manually with musl.  I haven't yet worked on the autotools bits.  I don't think it will be too difficult but I don't think I'm going to have a chance to work on it this week.

gatispaeglis:
This patch is still Work-In-Progress. There are few more issues that I need to resolve.

gatispaeglis:
Squashed the autotest into the relevant commit.

wmanley:
> @wmanley any success with the 'musl' patch?

musl patch merged as part of #477.

wmanley:
> musl patch merged as part of #477.

Also further musl fixes in #485

gatispaeglis:
FYI this is how I am using --karg-none : https://codereview.qt-project.org/#/c/174145/3/src/lib/qotaclientasync.cpp

cgwalters:
Is this still WIP, or is it ready for review?

wmanley:
@gatispaeglis: I was hoping to start using this soon.  What's your feeling on it?

gatispaeglis:
@cgwalters, The more accurate status of this PR would be "it is up from grabs" :) The patch in general is in good shape. I have been using it for a while. There is one issue that I noticed, but haven't had time to fix. The issue is in deployment_bootconfigs_equal. It would need to special case the "init=" command line argument (as it is already doing for "ostree=" ). Otherwise the bootloader configs are unnecessary rewritten, even when "inti=" is the same.

The latest snapshot of the documentation how I used this is in http://doc-snapshots.qt.io/qtota/index.html

@wmanley The above should answer your question too.

My focus recently has changed to other projects, but I can still help with reviewing.

cgwalters:
Ok cool, removing WIP.  Will take a look soon.

cgwalters:
bot, retest this please

cgwalters:
This was merged in https://github.com/ostreedev/ostree/pull/1401

PR #443, opened 2016-08-08
repo: Add a high level ostree_repo_import_oci_image_layer()

cgwalters:
I'm trying to solve a few issues with this.  First, the
current libarchive APIs are not introspectable (https://github.com/ostreedev/ostree/pull/412).
But at a higher level, with multiple programs like the atomic
command and os-exploder both importing Docker images into OSTree,
it makes sense to add a higher level API that "does the right thing"
rather than having both do it.


14rcole:
What's the reason behind having a directory file descriptor and a path?

cgwalters:
Internally ostree uses fd-relative a lot - it's more efficient and secure, and it helps to also do it externally.  You can just pass `-1` (or `AT_FDCWD` (which is really `-100`, see https://github.com/docker/docker/blob/master/pkg/system/utimes_linux.go#L12).

giuseppe:
this looks good, but I am wondering if we could also add the possibility to make directories writeable, as it will probably be duplicated by the users of this API.  What do you think?

cgwalters:
Ok, done :heavy_check_mark:

giuseppe:
LGTM, feel free to merge it, otherwise I will test it more tomorrow and see how it integrates in atomic

14rcole:
@rh-atomic-bot r+ eabd3bb

jlebon:
bot, retest this please

tests/test-libarchive.sh
```diff
@@ -154,3 +154,33 @@ $OSTREE checkout partial partial-checkout
 cd partial-checkout
 assert_file_has_content subdir/original "original"
 echo "ok tar partial commit contents"
+
+
+cd ${test_tmpdir}
+# We need a bare-user repo for this
+rm repo -rf
+ostree --repo=repo init --mode=bare-user
```
jlebon: `$OSTREE`

cgwalters:
Let's wait until people have tested this before merging.

- [ ] @14rcole 
- [ ] @giuseppe

giuseppe:
I tested it with atomic, it helps to simplify some code:

```
diff --git a/Atomic/syscontainers.py b/Atomic/syscontainers.py
index 8bfca1e..d9e829a 100644
--- a/Atomic/syscontainers.py
+++ b/Atomic/syscontainers.py
@@ -631,18 +631,12 @@ class SystemContainers(object):
     @staticmethod
     def _import_layers_into_ostree(repo, imagebranch, manifest, layers):
         repo.prepare_transaction()
+        modifier = OSTree.RepoCommitModifier(0, None, None)
         for layer, tar in layers.items():
             mtree = OSTree.MutableTree()
-            def filter_func(*args):
-                info = args[2]
-                if info.get_file_type() == Gio.FileType.DIRECTORY:
-                    info.set_attribute_uint32("unix::mode", info.get_attribute_uint32("unix::mode") | stat.S_IWUSR)
-                return OSTree.RepoCommitFilterResult.ALLOW
-
-            modifier = OSTree.RepoCommitModifier(0, filter_func, None)
-            repo.write_archive_to_mtree(Gio.File.new_for_path(tar), mtree, modifier, True)
-            root = repo.write_mtree(mtree)[1]
+            repo.import_oci_image_layer(None, -1, tar, mtree, modifier)
             metav = GLib.Variant("a{sv}", {'docker.layer': GLib.Variant('s', layer)})
+            root = repo.write_mtree(mtree)[1]
             csum = repo.write_commit(None, "", None, metav, root)[1]
             repo.transaction_set_ref(None, "%s%s" % (OSTREE_OCIIMAGE_PREFIX, layer), csum)
```

cgwalters:
You should just be able to pass `None` for modifier now too.

cgwalters:
Other things we may want to do:

 - Support mapping all uids to e.g. nobody
 - Strip all xattrs except security.capability?

cgwalters:
@willmtemple

cgwalters:
https://github.com/libarchive/libarchive/issues/752 is kind of a blocker for this.  @willmtemple worked around it by falling back to using gnu tar to explode to disk, then ostree commit from the filesystem, but this is a lot slower and obviously ugly.

PR #444, opened 2016-08-08
[merged] Final excision of libgsystem dependency

cgwalters:
Lots and lots of preparation led to this moment - when nothing
apparent changes for users!  Woo!

But seriously, having the extra dependency is a minor annoyance, and
in the big picture I think the libgsystem idea was wrong - we need to
land things in GLib, and use git submodules for API-unstable or
Linux-specific sharing.  For a lot of OSTree, the libgsystem `GFile*`
orientation was also wrong, we really want fd-relative.


jlebon:
Nice! Can we also drop it from `ostree.spec.in` ?
Also, there's a bunch of TODOs related to this in the new `ci-install.sh`.

jlebon:
@rh-atomic-bot r+ f3794e2

PR #445, opened 2016-08-08
[merged] prune: Retain the tip of each ref even with date pruning

cgwalters:
I hit an error with [CAHC](https://wiki.centos.org/SpecialInterestGroup/Atomic/Devel)
where we were doing time-based pruning.  `ostree summary -u` started failing,
and it took me a bit to realize it was because we were pruning
even the tip of old branches, which I was not at all expecting,
and I don't think users will too.

Perhaps in the future we could add some sort of --prune-ref-tips or
something if people wanted it, but I doubt it.


cgwalters:
Interesting that pull is flaking out on Travis.  I wonder if the more heavily loaded infrastructure is better at exposing races.  Will look at that.

jlebon:
Makes sense.
@rh-atomic-bot r+ 4d933cb

PR #446, opened 2016-08-09
[merged] build: Update .gitignore

cgwalters:
The patch to add more symlink bits also needed CLEANFILES updates.


jlebon:
@rh-atomic-bot r+ 0529d41

Issue #447, opened 2016-08-09
Disable static deltas by default for local pulls


dbnicholson:
When doing local pulls, you actually want to hardlink/copy the exact objects rather than generating them from static deltas.

PR #448, opened 2016-08-09
[merged] pull: Disable static deltas by default for local pulls

dbnicholson:
For local pulls there's no benefit pulling the static delta over the
individual object files since there's no HTTP overhead. Furthermore,
processing deltas always generates the objects whereas a standard pull
ensures that the exact object files are copied. Using deltas also
prevents hardlinking the objects if the repos exist on the same
filesystem.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2413,6 +2413,12 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
         }
     }
 
+  /* For local pulls, default to disabling static deltas so that the
+   * exact object files are copied.
+   */
+  if (pull_data->remote_repo_local && !require_static_deltas)
+    disable_static_deltas = TRUE;
+
   pull_data->static_delta_superblocks = g_ptr_array_new_with_free_func ((GDestroyNotify)g_variant_unref);
 
   {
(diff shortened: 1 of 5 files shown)


dbnicholson:
See https://github.com/ostreedev/ostree/issues/447.

src/libostree/ostree-repo-pull.c
```diff
@@ -2413,6 +2413,13 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
         }
     }
 
+  /* For local pulls, default to disabling static deltas so that the
+   * exact object files are copied.
+   */
+  if (pull_data->remote_repo_local &&
+      !(disable_static_deltas || require_static_deltas))
```
cgwalters: Out of curiosity, why not just `if (pull_data->remote_repo_local && !require_static_deltas)` ?

cgwalters:
Can you add a `Closes: #447` in the commit message to auto-link them?

A test case would be nice too but is optional.

src/libostree/ostree-repo-pull.c
dbnicholson: Oh, that would be better. I was worried about overriding the passed in disabled case, but I guess you'd just get the same thing.

dbnicholson:
Sure thing. I thought that was something you had some hook to do, but we do the same thing at Endless.

dbnicholson:
I reworked this now and fleshed it out a little more so that the existing delta tests actually work and you can make pull-local use deltas if you want. This is based on top of #455 so that the tests actually pass.

The test I added to see that deltas were not used by default in pull-local was to see if the objects were hardlinked. I'm not 100% sure that will work across all filesystems, though. I'm not sure how to detect if deltas weren't used otherwise.

cgwalters:
Excellent patches!

@rh-atomic-bot r+ e419f35

PR #449, opened 2016-08-09
[merged] Move ostree-* executables to /usr/lib/ostree

gatispaeglis:
The "ostree-boot" directory is for boot related files. Therefore is
a more appropriate place for ostree-{prepare-root,remount,grub-generator}
executibles and in addition saves us some typing in a console
when wanting to access the "ostree" cmd line.

Why not to use libexecdir instead? Because this directory does not
exist on some distros or paths between distros varies. There are several
reasons why a well known path is prefered (for example when passing
init=${ostree}/usr/lib/ostree-boot/ostree-prepare-root via kernel
command line).

### DIFF
--- a/Makefile-boot.am
+++ b/Makefile-boot.am
@@ -53,7 +53,7 @@ grub2configdir = $(sysconfdir)/grub.d
 INSTALL_DATA_HOOKS += install-grub2-config-hook
 else
 # We're using our internal generator
-libexec_SCRIPTS = src/boot/grub2/ostree-grub-generator
+ostree_boot_SCRIPTS = src/boot/grub2/ostree-grub-generator
 endif
 
 EXTRA_DIST += src/boot/dracut/module-setup.sh \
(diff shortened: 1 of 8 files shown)


gatispaeglis:
Still testing, but so far looks good. I assume this requires changing something in ostree/packaging/ostree.spec.in as well, but I am not sure how to test that.

cgwalters:
Hmm, this would break rpm-ostree at least which assumes that it can *create* `/usr/lib/ostree-boot` by moving the legacy `/boot` location.  I think we can teach it not to do that but...we never want this content to actually end up on `/boot`, right?

So it seems like it's easier to make it `/usr/lib/ostree` or so, which is what non-libexec distros like Debian typically use anyways.

gatispaeglis:
> we never want this content to actually end up on /boot, right?

Correct. 

> So it seems like it's easier to make it /usr/lib/ostree or so

Moving it to /usr/lib/ostree is a good solution too. I will update the patch accordingly.

cgwalters:
LGTM, thanks!

@rh-atomic-bot r+ 730aa2e

PR #451, opened 2016-08-09
[merged] tests: Port sysroot.js away from libgsystem

cgwalters:
My case-sensitive `git grep` didn't find this one originally,
but the Travis+Debian setup was configured to run it.


jlebon:
@rh-atomic-bot r+ 59552ed

PR #453, opened 2016-08-09
[merged] libostree.sym: Add 2016.9 section

cgwalters:
We should remember to do this in the commit updating configure.ac.


jlebon:
@rh-atomic-bot r+ 1946f01

Issue #454, opened 2016-08-09
Static delta generation broken


dbnicholson:
After f66906c, static delta generation is broken (on i386 at least). There are 2 issues:

* The call to `fcntl` with `F_DUPFD_CLOEXEC` when `filename` is not specified is missing the minimum fd number argument.
* In the out path, derefencing `part_temp_fds->len` may segfault if `part_temp_fds` is still `NULL`.

PR #455, opened 2016-08-09
Static delta generation broken

dbnicholson:
https://github.com/ostreedev/ostree/commit/f66906c40179ae9fe047be5412fe368f29a98290 broke some things.

Closes: #454


cgwalters:
This was included in https://github.com/ostreedev/ostree/pull/448

PR #458, opened 2016-08-11
[merged] repo: Really ignore progress changed user data

dbnicholson:
The documentation says this is ignored, implying that you should pass
NULL to it. However, the function immediately returns in this case even
though the argument isn't used anywhere.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3839,10 +3839,6 @@ ostree_repo_pull_default_console_progress_changed (OstreeAsyncProgress *progress
   guint fetched_delta_parts;
   guint total_delta_parts;
 
-  /* Historical note; we used to treat this as a GSConsole instance */
-  if (user_data == NULL)
-    return;
-
   buf = g_string_new ("");
 
   status = ostree_async_progress_get_status (progress);


cgwalters:
Hum...this came from https://github.com/ostreedev/ostree/commit/a6c731f6e7d1014aab5989c98a6432d28080905b originally.  It looks like I was trying to be maximally compatible.  Callers still end up passing data yes.

Why are you making this PR?  Did something break?  You just noticed it and are doing a cleanup?

Hopefully we didn't somehow break compatibility?

dbnicholson:
Yeah, I'm aware of a6c731f. I should have mentioned that in the commit message. Why would you pass any data when the documentation says it's ignored and the code does nothing with it? If you were writing new ostree code, you'd read that and pass in `NULL`.

Oh, I guess the compatibility you're mentioning is that if someone intentionally passed in `NULL` and expected that to mean no progress would be printed, then this would break. But why would you wire up the callback if you didn't want it to run?

But here's my actual issue. I'd like to use this function from python to have progress. However, because the arguments are `gpointer` and not proper typedefs (or maybe just missing annotation?), pygi can only pass integers or `None`. Now, I could pass `None` and the function would actually DTRT except for this check that doesn't actually make sense. If I try to pass in some garbage pointer or even a `GSystem.Console`, pygi won't do it.

cgwalters:
Ok.  "Trying to use this from pygobject" is the type of real-world usage that I like to see in commit messages.

But looking at things, I think it should be safe to just remove it.  In fact rpm-ostree no longer uses this, and flatpak's usage looks safe.

@rh-atomic-bot r+ 9b2c8fd

dbnicholson:
> Ok. "Trying to use this from pygobject" is the type of real-world usage that I like to see in commit messages.

I agree with that. However, in this case it seemed like a generic oversight when I first looked at it. It happens to make life more difficult with pygobject, but I thought any user of libostree could hit this.

PR #462, opened 2016-08-16
[merged] rofiles-fuse: Rework to be based on nlink

cgwalters:
Programs like `useradd` try to `open(/etc/passwd, O_RDWR)` to append,
which didn't work with rofiles-fuse.  Thinking about this, I realized
that there's a simpler algorithm for "can we write to this file" which
is "does it have a hardlink count <= 1"?

Switching to this both drops complexity (we no longer need to keep a
hash table of files we created), and also lets useradd work.

### DIFF
--- a/src/rofiles-fuse/main.c
+++ b/src/rofiles-fuse/main.c
@@ -39,7 +39,6 @@
 
 // Global to store our read-write path
 static int basefd = -1;
-static GHashTable *created_devino_hash = NULL;
 
 static inline const char *
 ENSURE_RELPATH (const char *path)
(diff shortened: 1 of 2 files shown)


jlebon:
Nice simplification. Hmm, I'm trying to make sure we wouldn't ever need the previous behaviour. This does make it less useful for the general purpose of protecting any directory. Would you consider making it into a switch instead? (Or make the previous behaviour a switch).

E.g. in distros we put this in `/usr/bin/`, so this technically breaks backwards compatibility. Though it's probably safe to assume no one else is using this, it's still something to consider.

cgwalters:
Who would want the previous behavior specifically though?  I can't think of how this behavior is *less* general purpose given that it permits more.

jlebon:
With the previous behaviour, you could protect any dir, not just those that were checkouts of hardlink farms, right? So e.g. what wasn't writable before now will be writable.

This is just something that came to mind when reviewing, I don't actually have a strong opinion on it (since, again, I highly doubt anyone else is using it, though that might point to putting it in libexec instead).

jlebon:
(Also, small typo :pencil: in the title of the first (but chronologically last) commit: `Acutally`)

cgwalters:
But protect against what?  I can't really think of a use case for rofiles-fuse that isn't hardlink farms.  If you want "pure read only" that's more like `mount --bind -o ro` right?

cgwalters:
Rebased with typo fix.

jlebon:
> But protect against what?

I mean, someone who still wants to allow deletion and creation, but disallow mutations.

> I can't really think of a use case for rofiles-fuse that isn't hardlink farms.

OK, that works for me. Just wanted to make sure we thought this through. :white_check_mark: 

@rh-atomic-bot r+ 8861c19

PR #463, opened 2016-08-17
Grub generator refactor

gatispaeglis:
WIP - need to do a bit more testing.


gatispaeglis:
I think nobody besides me has used the grub generator feature yet, but nevertheless this change is also backward compatible (with minor adjustments in build systems).

src/libostree/ostree-bootloader-grub2.c
```diff
@@ -315,7 +316,45 @@ _ostree_bootloader_grub2_write_config (OstreeBootloader      *bootloader,
...
+          config = boot_loader_configs->pdata[0];
+          ostree_arg = get_ostree_kernel_arg_from_config (config);
+          if (!ostree_arg)
+            {
+              g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                                   "No ostree= kernel argument found in boot loader configuration file");
+              goto out;
+            }
+
+          snprintf (generator_path, sizeof(generator_path), "%s/%s/usr/lib/ostree/ostree-grub-generator",
```
cgwalters: The problem with this in the general case is that it assumes the new tree's binaries work on the current one.  For say ELF binaries, if the new binary has a new dependency (e.g. versioned glibc symbol), or a new shared library, it's going to fail.  In this case it's a shell script so that'll work fine, but *everything else* it executes is going to be from the current tree...

There are going to be cases where we need to run code from the new tree - I have pending work on recompiling SELinux policy, and was going to look at doing that via a container (`bwrap`/`systemd-nspawn`) in the new root.

But what you're doing here isn't strictly necessary, right?  It's just trying to avoid the user having to upgrade twice to get the change, but at a risk of more incompatibilities.

gatispaeglis: >  In this case it's a shell script so that'll work fine, but everything else it executes is going to be from the current tree...

Right, I did not think about that.  

> But what you're doing here isn't strictly necessary, right? It's just trying to avoid the user having to upgrade twice to get the change, but at a risk of more incompatibilities.
 
Yes, it is not necessary. Just wanted to add an optimization, but now I see that it might not have been a good idea :) Will drop this change.

wmanley: > I have pending work on recompiling SELinux policy, and was going to look at doing that via a container (bwrap/systemd-nspawn) in the new root.

I have a script that I use for modifying ostree images by running a command in them: https://gist.github.com/wmanley/243cbcbe7c3cb938acc4cf5d689fde22 .  I use it with `qemu-user-static` so I can run commands on my ARMv7 images on my AMD64 PC.  It uses `runc`.  It may be useful to you.

PR #468, opened 2016-08-25
[merged] pull: Make .commitpartial files world readable

cgwalters:
In CentOS, these happened to appear in a repo that is served
via rsync, and having them not be world-readable caused mirroring
tools to fail.

They aren't secret, so don't make them so.


giuseppe:
@rh-atomic-bot r+ 7ea2caf

PR #469, opened 2016-08-25
[merged] Add contenturl and mirrorlist support

jlebon:
This PR adds the ability to specify a `contenturl` setting. If set, ostree will fetch static deltas and objects from `contenturl` but will use `url` for summaries & signatures. We also add mirrorlist support for both `url` and `contenturl` settings.

The overall goal here is to give content providers an easy way to make use of mirrors, while still making sure that files in the security chain are fetched from one (or maybe more) central location e.g. over pinned TLS if they so wish.

### DIFF
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -74,9 +74,9 @@ typedef struct {
   volatile int ref_count;
 
   ThreadClosure *thread_closure;
-  GSList *mirrorlist; /* list of base URIs */
+  GPtrArray *mirrorlist; /* list of base URIs */
   char *filename; /* relative name to fetch or NULL */
-  guint tried_mirrors;
+  guint mirrorlist_idx;
 
   OstreeFetcherState state;
 
@@ -206,6 +206,7 @@ pending_uri_unref (OstreeFetcherPendingURI *pending)
 
   g_clear_pointer (&pending->thread_closure, thread_closure_unref);
 
+  g_clear_pointer (&pending->mirrorlist, g_ptr_array_unref);
   g_free (pending->filename);
   g_clear_object (&pending->request);
   g_clear_object (&pending->request_body);
@@ -360,17 +361,21 @@ create_pending_soup_request (OstreeFetcherPendingURI  *pending,
                              GError                  **error)
 {
   g_autofree char *uristr = NULL;
+  SoupURI *next_mirror = NULL;
   SoupURI *uri = NULL;
 
   g_assert (pending->mirrorlist);
-  g_assert (pending->mirrorlist->data);
+  g_assert (pending->mirrorlist_idx < pending->mirrorlist->len);
 
-  uristr = g_build_filename (soup_uri_get_path (pending->mirrorlist->data),
+  next_mirror = g_ptr_array_index (pending->mirrorlist,
+                                   pending->mirrorlist_idx);
+  uristr = g_build_filename (soup_uri_get_path (next_mirror),
                              pending->filename /* may be NULL */, NULL);
-  uri = soup_uri_copy (pending->mirrorlist->data);
+  uri = soup_uri_copy (next_mirror);
   soup_uri_set_path (uri, uristr);
 
   g_clear_object (&pending->request);
+
   pending->request = soup_session_request_uri (pending->thread_closure->session,
                                                uri, error);
   soup_uri_free (uri);
@@ -961,12 +966,10 @@ on_request_sent (GObject        *object,
         }
       else if (!SOUP_STATUS_IS_SUCCESSFUL (msg->status_code))
         {
-          pending->tried_mirrors++;
-
           /* is there another mirror we can try? */
-          if (g_slist_next (pending->mirrorlist) != NULL)
+          if (pending->mirrorlist_idx + 1 < pending->mirrorlist->len)
             {
-              pending->mirrorlist = g_slist_next (pending->mirrorlist);
+              pending->mirrorlist_idx++;
               create_pending_soup_request (pending, &local_error);
               g_queue_insert_sorted (&pending->thread_closure->pending_queue,
                                      g_object_ref (task), pending_task_compare,
@@ -991,10 +994,10 @@ on_request_sent (GObject        *object,
                                          msg->status_code,
                                          soup_status_get_phrase (msg->status_code));
 
-              if (pending->tried_mirrors > 1)
+              if (pending->mirrorlist->len > 1)
                 g_prefix_error (&local_error,
                                 "All %u mirrors failed. Last error was: ",
-                                pending->tried_mirrors);
+                                pending->mirrorlist->len);
             }
           goto out;
         }
@@ -1059,7 +1062,7 @@ on_request_sent (GObject        *object,
 
 static void
 ostree_fetcher_mirrored_request_internal (OstreeFetcher         *self,
-                                          GSList                *mirrorlist,
+                                          GPtrArray             *mirrorlist,
                                           const char            *filename,
                                           gboolean               is_stream,
                                           guint64                max_size,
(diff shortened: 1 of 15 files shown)


jlebon:
I still need to add tests for this. Just wanted to post early to get some feedback as well as determine the conflict areas with @Mathnerd314's work.

jlebon:
:arrow_up: OK, now with tests!

jlebon:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;IVUvFd7O&#x2F;364-5-split-ostree-metadata-and-content-fetches"> (5) Split ostree metadata and content fetches</a>

giuseppe:
I had a quick look at it, and it looks fine.  @Mathnerd314, can you also take a look and see if/how this is going to cause any problem to you?

Mathnerd314:
Well, it conflicts in a lot of places. E.g.:
* I removed the synchronous fetch API completely, whereas this modifies it. 
* I moved all usage of suburi_new into a new file, this changes them to g_build_filename

But there are also existing commits I have to rebase over, so I'm not sure those count. The tricky parts are adding a new mirrorlist call and sending the GSList over D-Bus, which seem feasible, although not this weekend.

src/libostree/ostree-fetcher.c
```diff
@@ -937,20 +961,41 @@ on_request_sent (GObject        *object,
...
-              code = G_IO_ERROR_NOT_FOUND;
-              break;
-            default:
-              code = G_IO_ERROR_FAILED;
+              pending->mirrorlist = g_slist_next (pending->mirrorlist);
+              create_pending_soup_request (pending, &local_error);
+              g_queue_insert_sorted (&pending->thread_closure->pending_queue,
+                                     g_object_ref (task), pending_task_compare,
+                                     NULL);
+              remove_pending_rerun_queue (pending);
```
cgwalters: This looks wrong...we're inserting it :inbox_tray: in the pending queue, then immediately removing it :outbox_tray: ?

src/libostree/ostree-fetcher.h
```diff
@@ -70,20 +70,31 @@ void _ostree_fetcher_set_tls_database (OstreeFetcher *self,
...
-                                                    guint64                max_size,
-                                                    int                    priority,
-                                                    GCancellable          *cancellable,
-                                                    GAsyncReadyCallback    callback,
-                                                    gpointer               user_data);
-
-char *_ostree_fetcher_request_uri_with_partial_finish (OstreeFetcher *self,
-                                                       GAsyncResult  *result,
-                                                       GError       **error);
+void _ostree_fetcher_mirrorred_request_with_partial_async (OstreeFetcher         *self,
```
cgwalters: `s/mirrorred/mirrored/`

src/libostree/ostree-repo-pull.c
```diff
@@ -380,6 +344,21 @@ fetch_uri_contents_utf8_sync (OstreeFetcher  *fetcher,
...
+                              SoupURI        *uri,
+                              char          **out_contents,
+                              GCancellable   *cancellable,
+                              GError        **error)
+{
+  GSList *mirrorlist = g_slist_append (NULL, uri);
+  gboolean ret =
+    fetch_mirrorred_uri_contents_utf8_sync (fetcher, mirrorlist, NULL,
+                                            out_contents, cancellable, error);
+  g_slist_free (mirrorlist);
```
cgwalters: Could use `g_autoptr(GSList) mirrorlist` and this would all condense into a single line.

src/libostree/ostree-repo-pull.c
```diff
@@ -2006,6 +1972,105 @@ _ostree_preload_metadata_file (OstreeRepo    *self,
...
+      if (*mirror_uri_str == '\0' || *mirror_uri_str == '#')
+        continue;
+
+      mirror_uri = soup_uri_new (mirror_uri_str);
+      if (mirror_uri == NULL)
+        {
+          g_debug ("Can't parse mirrorlist line '%s'", mirror_uri_str);
+          continue;
+        }
+      else if (strcmp (soup_uri_get_scheme (mirror_uri), "file") == 0)
```
cgwalters: I would harden this specifically to only limit it to `http` and `https`.

src/libostree/ostree-repo-pull.c
```diff
@@ -2006,6 +1972,105 @@ _ostree_preload_metadata_file (OstreeRepo    *self,
...
+  if (!fetch_uri_contents_utf8_sync (fetcher, mirrorlist, &contents,
+                                     cancellable, error))
+    {
+      g_prefix_error (error, "While fetching mirrorlist '%s': ",
+                      mirrorlist_url);
+      goto out;
+    }
+
+  /* go through each mirror in mirrorlist and do a quick sanity check that it
+   * works so that we don't waste the fetcher's time when it goes through them.
```
cgwalters: Hum...but that's potentally a *lot* of new TCP connections in the happy path when the first mirror works.  I don't think this is a good idea by default.

src/libostree/ostree-repo-pull.c
```diff
@@ -2018,9 +2083,9 @@ repo_remote_fetch_summary (OstreeRepo    *self,
   glnx_unref_object OstreeFetcher *fetcher = NULL;
   g_autoptr(GMainContext) mainctx = NULL;
   gboolean ret = FALSE;
-  SoupURI *base_uri = NULL;
   gboolean from_cache = FALSE;
   g_autofree char *url_override = NULL;
+  GSList *mirrorlist = NULL;
```
cgwalters: Noticing you aren't using autoptr here reminds me...in general, there are really very few *good* use cases for linked lists; you almost always want a vector (`GPtrArray`) instead.  See e.g. [this](http://cglab.ca/~abeinges/blah/too-many-lists/book/#an-obligatory-public-service-announcement).

And with `GPtrArray` we can add a free-element function which makes using `autoptr` possible always.

src/libostree/ostree-fetcher.c
jlebon: We add to the pending queue, while `remove_pending_rerun_queue` removes from the outstanding queue since this is a new request. Maybe we should rename `remove_pending_rerun_queue` to something else. :abcd:

src/libostree/ostree-repo-pull.c
jlebon: Yeah, that's a good point. OTOH, if the first few mirrors are bad, then it's faster overall to do the check here. I fine-tuned this in the last fixup commit so that we only check up to the first happy mirror.

src/libostree/ostree-repo-pull.c
jlebon: Thanks for the article. I switched it over to `GPtrArray`.

jlebon:
Huh, weird. I can't reproduce those CI pull failures.

jlebon:
bot, retest this please

cgwalters:
There's a lot we could architecture :tokyo_tower: astronaut :rocket: here, but it looks pretty good, we can build on it after working on the multiprocess patches.

@rh-atomic-bot r+ c142d79

cgwalters:
```
(/usr/libexec/installed-tests/ostree/test-pull-c:20875): GLib-CRITICAL **: source_remove_from_context: assertion 'source_list != NULL' failed
FAIL: ostree/test-pull-c.test (Child process killed by signal 5)
```

jlebon:
Yup, there's something nasty lurking underneath. I've been debugging it on the PR tester itself since it's the only place I can reliably reproduce this. Not sure if I introduced it, or if it's a variation on what Travis has been running into for a while.

cgwalters:
Do you have a core dump/stack trace?

cgwalters:
FWIW in a past life I wrote [some code](https://git.gnome.org/browse/gnome-continuous/tree/src/js/tasks/testbase.js?id=d00998c1094fab5e0e00001667375a3a196c4177#n111) to watch the systemd journal from testing VMs and because coredumpctl does auto-stack extraction, we'd get stack traces collected on the build server.

jlebon:
Well, that took an embarrassingly long time to figure out. I had:

```c
  g_clear_pointer (&pull_data->meta_mirrorlist, (GDestroyNotify) g_ptr_array_unref);
  if (pull_data->content_mirrorlist != pull_data->meta_mirrorlist)
    g_clear_pointer (&pull_data->content_mirrorlist, (GDestroyNotify) g_ptr_array_unref);
```

Since we clear `meta_mirrorlist` before the comparison, the if condition will of course always be true, which meant that we ended up double unref'ing. Undefined behaviour ensued.

I also kept some error checking I added while debugging this.

cgwalters:
@rh-atomic-bot r+ 3cd17dd

PR #473, opened 2016-08-27
[merged] sysroot: Drop unnecessary `dup()` invocation

cgwalters:
It's close-on-exec, not close-on-fork.  I was clearly confused when
writing this; it works just fine to reference the fd in the child and
`fchdir()` before exec.  So drop the unnecessary duplication.

Just noticed this while reading the code for a random other reason.


giuseppe:
@rh-atomic-bot r+ 2d13763

giuseppe:
@rh-atomic-bot r+ ed795ea

cgwalters:
@giuseppe In this case retry like https://github.com/ostreedev/ostree/pull/473#issuecomment-242976196

PR #474, opened 2016-08-29
[merged] sysroot: Add a flag to suppress post-deploy cleanup

cgwalters:
I noticed seeing the output of `prune` twice in rpm-ostree,
and had always wondered why.  When reading the code for unrelated
reasons, I noticed the reason - we were pruning once here,
and then once after rpm-ostree regenerates its "base" refs.

There's no reason to clean twice, so let's add a flag so rpm-ostree
can suppress doing it inside libostree.


giuseppe:
@rh-atomic-bot r+ 523e1a3

PR #476, opened 2016-08-29
[merged] commit: Don't delete tmp/cache dir

cgwalters:
We hold a fd open on this, and it's basically now expected
to be immortal.  Confer that status.

This was showing up in flatpak crashers, because we'd get
an unexpected errno.

(I didn't test this fixes the crasher, but it's clearly right)

https://bugzilla.redhat.com/show_bug.cgi?id=1347293

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1219,6 +1219,12 @@ cleanup_tmpdir (OstreeRepo        *self,
       if (dent == NULL)
         break;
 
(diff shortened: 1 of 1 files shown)


alexlarsson:
@rh-atomic-bot r+ 3692f9f

cgwalters:
@rh-atomic-bot retry

alexlarsson:
@smcv Can you make sure this patch is in the debian packages?

smcv:
@alexlarsson: sure, I'll include this in my next ostree/flatpak update round.

PR #477, opened 2016-08-30
[merged] ostree-prepare-root: Allow building statically with musl

wmanley:
If the `--with-static-compiler=musl-gcc` configure flag is given.

ostree-prepare-root can be used as init in a system without a populated /lib.  To support this use case we need to link statically as we will be unable to locate libc.so at run time if it's not installed in /lib.

We support building ostree-prepare-root with a different compiler to the rest of ostree so we can use musl rather than glibc.  This reduces the size of the executable significantly: from ~700K -> ~30K.  We have to use `_SCRIPTS` here to get autotools to install this as an executable but without generating rules to make it itself which we have specified manually.

See https://lists.gnu.org/archive/html/help-gnu-utils/2007-01/msg00007.html for advice on using autotools in this manner.

File size measurements from #403:  File sizes built for AMD64 with -O2 and stripped:

| configuration      | size |
|--------------------|-----|
| glibc static    |   724K |
| glibc shared    |   12K |
| musl static    |   30K |

### DIFF
--- a/Makefile-switchroot.am
+++ b/Makefile-switchroot.am
@@ -17,7 +17,6 @@
 
 if BUILDOPT_SYSTEMD
 
-ostree_boot_PROGRAMS += ostree-prepare-root
 ostree_boot_PROGRAMS += ostree-remount
 
 noinst_LTLIBRARIES += libswitchroot-mountutil.la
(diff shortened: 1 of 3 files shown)


cgwalters:
Nice!  LGTM.

@rh-atomic-bot r+ 48db44a

wmanley:
I'm slightly in two minds about this.  There are actually two variables here:

1. Do you want to link ostree-prepare-root statically?
2. Do you want to use a custom compiler for `ostree-prepare-root`

The option I've added here conflates the two.  It's probably not a bad idea in practice as I can't imagine you'd want a dynamic musl build and the static glibc build build is just too fat, but I thought I'd point it out anyway.

Makefile-switchroot.am
```diff
@@ -28,9 +27,32 @@ libswitchroot_mountutil_la_SOURCES = \
...
+# executable significantly: from ~700K -> ~30K.  We have to use _SCRIPTS here
+# to get autotools to install this as an executable but without generating rules
+# to make it itself which we have specified manually.  See
+# https://lists.gnu.org/archive/html/help-gnu-utils/2007-01/msg00007.html
+ostree_boot_SCRIPTS = ostree-prepare-root
+
+ostree-prepare-root : $(ostree_prepare_root_SOURCES)
+	$(STATIC_COMPILER) -o $@ -static $(ostree_prepare_root_SOURCES) $(AM_CPPFLAGS) $(AM_CFLAGS)
+else
+ostree_boot_PROGRAMS += ostree-prepare-root
```
wmanley: Note: now `ostree-prepare-root` is built whether `BUILDOPT_SYSTEMD` is enabled or not.  This seems right as I've been using with both systemd based systems and non-systemd systems.

Makefile-switchroot.am
```diff
@@ -28,9 +27,32 @@ libswitchroot_mountutil_la_SOURCES = \
...
+
 ostree_prepare_root_CFLAGS = $(AM_CFLAGS) -Isrc/switchroot
+endif
 
 ostree_remount_SOURCES = src/switchroot/ostree-remount.c
 ostree_remount_LDADD = libswitchroot-mountutil.la
```
wmanley: For the sake of consistency I'm wondering if I should ditch libswitchroot-mountutil.la now that it's only linked into `ostree-remount`.

cgwalters: Probably.  One thing I meant to comment on - it looks like we can switch to `err` rather than `perrorv + exit()`?  That would clean up the code a lot, then we could likely move `path_is_on_readonly_fs` to a static inline or so.

cgwalters: (I'm a bit fan of the `err` function nowadays, it's not standard but it does look like musl has it too)

Makefile-switchroot.am
```diff
@@ -28,9 +27,32 @@ libswitchroot_mountutil_la_SOURCES = \
...
+# We support building ostree-prepare-root with a different compiler to the rest
+# of ostree so we can use musl rather than glibc.  This reduces the size of the
+# executable significantly: from ~700K -> ~30K.  We have to use _SCRIPTS here
+# to get autotools to install this as an executable but without generating rules
+# to make it itself which we have specified manually.  See
+# https://lists.gnu.org/archive/html/help-gnu-utils/2007-01/msg00007.html
+ostree_boot_SCRIPTS = ostree-prepare-root
+
+ostree-prepare-root : $(ostree_prepare_root_SOURCES)
+	$(STATIC_COMPILER) -o $@ -static $(ostree_prepare_root_SOURCES) $(AM_CPPFLAGS) $(AM_CFLAGS)
```
gatispaeglis: This fails on Ubuntu (tested both `--with-static-compiler=gcc` and `--with-static-compiler=gcc`) :

```
src/switchroot/ostree-prepare-root.c:30:20: fatal error: config.h: No such file or directory
compilation terminated.
```

With `$(DEFAULT_INCLUDES)` added to the above line, build succeeds.

wmanley:
`ostree-prepare-root` isn't working just yet as musl's `readlink` implementation depends on `/proc`: http://www.openwall.com/lists/musl/2016/06/08/2

Makefile-switchroot.am
wmanley: Fixed in #485

PR #478, opened 2016-08-30
[merged] switchroot: Simplify build system

wmanley:
by removing `libswitchroot-mountutil.la`.

This builds with both musl and glibc.  The man page says that the `err` family of functions are "nonstandard BSD extensions" so we should expect that they are available on BSD too.  I'd not come across `err`, etc. before @cgwalters mentioned them in #477, they're quite neat.  Thanks.


cgwalters:
Nice, that is as much better as I thought it'd be.  Thanks!

@rh-atomic-bot r+ f14f5a2

cgwalters:
bot, add author to whitelist

PR #481, opened 2016-08-31
[merged] Release 2016.9

cgwalters:


jlebon:
Could we wait until the mirrorlist PR gets in? Should be done testing very soon!

cgwalters:
Hmm...I generally like to have major changes sit in git master for a little bit before we put them in a release.

cgwalters:
We should be able to have a 2 week or so cadence for releases, so it shouldn't be long before we ship the mirrorlist bits.

cgwalters:
Basically, I'd like to have this show up in CAHC for some manual testing too before updating fedora etc.

jlebon:
Yup, that makes complete sense.
@rh-atomic-bot r+

Issue #482, opened 2016-08-31
Cannot cancel a pull


GeorgesStavracas:
While working with GNOME Software and Flatpak, it is noticeable that the pull ('Downloading...' phase of GNOME Software) doesn't stop when clicking the 'Cancel' button.

The PR below fixes this issue.

cgwalters:
This was fixed.

PR #483, opened 2016-08-31
[merged] repo-pull: properly store the cancellable

GeorgesStavracas:
OSTree function ostree_repo_pull_with_options starts a
series of operations that makes heavy use of the PullData's
cancellable.

This isn't effective, however, since nowhere in the code
the OtPullData.cancellable field is set. This is visible,
for example, when trying to cancel a Flatpak pull and nothing
happens, because the cancellable is not properly passed
to the pull data.

Fix that by setting the cancellable field of the pull data. It
owns a reference for safety reasons, and unreferences it at the
end of the operation.

ostreedev/ostree#482

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2252,6 +2252,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   pull_data->is_mirror = (flags & OSTREE_REPO_PULL_FLAGS_MIRROR) > 0;
   pull_data->is_commit_only = (flags & OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY) > 0;
   pull_data->is_untrusted = (flags & OSTREE_REPO_PULL_FLAGS_UNTRUSTED) > 0;
+  pull_data->cancellable = cancellable ? g_object_ref (cancellable) : NULL;
 
   if (error)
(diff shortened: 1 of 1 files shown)


cgwalters:
Hum.  We were clearly *intending* to do this.  Which makes me think it's somehow a regression.  Looking at `git log -S 'pull_data.*cancellable'` and others like `git log -S '->cancellable ='` I don't see us ever doing it.

src/libostree/ostree-repo-pull.c
```diff
@@ -2252,6 +2252,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   pull_data->is_mirror = (flags & OSTREE_REPO_PULL_FLAGS_MIRROR) > 0;
   pull_data->is_commit_only = (flags & OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY) > 0;
   pull_data->is_untrusted = (flags & OSTREE_REPO_PULL_FLAGS_UNTRUSTED) > 0;
+  pull_data->cancellable = g_object_ref (cancellable);
```
cgwalters: This needs to be: `cancellable ? g_object_ref (cancellable) : NULL`.

cgwalters:
@rh-atomic-bot r+ 2813b04

GeorgesStavracas:
This was failing on tests, but I just pushed a NULL-safe version of the patch.

PR #484, opened 2016-08-31
[merged] pull code: clean up mirrorlist hack

jlebon:
While converting the mirrorlist code from using `GSList` to `GPtrArray`, I
completely missed the fact that there is now a much cleaner way to do
this.


cgwalters:
@rh-atomic-bot r+ 4c16c63

cgwalters:
@rh-atomic-bot r+ 4c16c6e

PR #485, opened 2016-08-31
[merged] Fix running ostree-prepare-root with musl

wmanley:
#477 provided the ability to build `ostree-prepare-root` statically against musl, but it didn't run correctly.  This allows ostree-prepare-root to work with musl's unconventional approach to implementing `realpath`.

musl libc's implementation of `realpath` works by opening the path and then
doing a lookup in `/proc/self/fd` to find the canonical path.  This fails
if `/proc` is not mounted.  This causes problems for us if
`ostree-prepare-root` is `init` as `/proc` won't be mounted.

We have to mount `/proc` anyway for `/proc/cmdline` so this fix just
expands the scope over which `/proc` is mounted to include both our
`realpath` calls.

See also:

* http://www.openwall.com/lists/musl/2016/06/08/2 and
* http://git.musl-libc.org/cgit/musl/tree/src/misc/realpath.c?id=e738b8cbe64b6dd3ed9f47b6d4cd7eb2c422b38d

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -174,17 +174,19 @@ pivot_root(const char * new_root, const char * put_old)
 int
 main(int argc, char *argv[])
 {
-  const char *root_mountpoint = NULL;
+  const char *root_mountpoint = NULL, *root_arg = NULL;
   char *deploy_path = NULL;
   char srcpath[PATH_MAX];
   struct stat stbuf;
 
   if (argc < 2)
-    root_mountpoint = "/";
+    root_arg = "/";
   else
-    root_mountpoint = argv[1];
+    root_arg = argv[1];
 
-  root_mountpoint = realpath (root_mountpoint, NULL);
+  root_mountpoint = realpath (root_arg, NULL);
+  if (root_mountpoint == NULL)
+    err (EXIT_FAILURE, "realpath(\"%s\")", root_arg);
   deploy_path = resolve_deploy_path (root_mountpoint);
 
   /* Work-around for a kernel bug: for some reason the kernel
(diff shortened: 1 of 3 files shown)


wmanley:
bot, test pull request

src/switchroot/ostree-prepare-root.c
```diff
@@ -174,17 +174,19 @@ pivot_root(const char * new_root, const char * put_old)
...
 
   if (argc < 2)
-    root_mountpoint = "/";
+    root_arg = "/";
   else
-    root_mountpoint = argv[1];
+    root_arg = argv[1];
 
-  root_mountpoint = realpath (root_mountpoint, NULL);
+  root_mountpoint = realpath (root_arg, NULL);
```
cgwalters: Hmm...one thing I started wondering; why are we invoking `realpath` at all here?   We already had another invocation inside `resolve_deploy_path()` which dates to  bb6eedfb258d3001f61c42c7e920c03dae2bdc1a which I probably wrote on a netbook while riding a dinosaur or something.

Is it just cosmetic, or did I have an actual goal there?  I don't remember...might be interesting to try taking the realpath invocations out.

jlebon:
bot, retest this please

src/switchroot/ostree-prepare-root.c
wmanley: I added it originally because I was defaulting argv[1] to `/sysroot` and I had a symlink `sysroot -> .` in my rootfs.  `realpath` was to resolve this symlink.  I guess it's not necessary with the way `ostree-prepare-root` is invoked now.

wmanley: I would guess you need `realpath` above because `mount` probably won't like it being a symlink.  It would be worth trying though.

cgwalters:
The mount syscall will definitely follow links.  But regardless, this all looks good to me for now.

@rh-atomic-bot r+ 26adaef

PR #486, opened 2016-09-01
[merged] Distribute manual pages properly

dbnicholson:
If using a `make dist` tarball, there are a couple issues.

* The XML is not distributed, so after `make clean`, a subsequent build will fail.
* The man page distribution depends on `ENABLE_MAN`. If `xsltproc` is not installed, this may be false.

### DIFF
--- a/Makefile-man.am
+++ b/Makefile-man.am
@@ -40,7 +40,7 @@ man5_files = ostree.repo.5 ostree.repo-config.5
 man1_MANS = $(addprefix man/,$(man1_files))
 man5_MANS = $(addprefix man/,$(man5_files))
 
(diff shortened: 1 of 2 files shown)


cgwalters:
Yeah, this keeps happening because I don't use `make dist` but [recursive git archive](https://github.com/ostreedev/ostree/blob/master/packaging/Makefile.dist-packaging#L11).  But we can keep it going.

@rh-atomic-bot r+ 3a4e1dc

dbnicholson:
Yeah, I appreciate you keeping the `dist` stuff working, and I don't mind coming around to fix it up. I know it's not your usage.

dbnicholson:
Wait, don't merge this. Just found it's still broken.

PR #488, opened 2016-09-01
[merged] build: Actually distribute man page XML source

dbnicholson:
The make substitution pattern was wrong. The source files are
"ostree.xml", not "ostree.1.xml", for instance.


dbnicholson:
Sorry, this should have been in #486, but I caught it after you merged.

cgwalters:
bot, add author to whitelist

@rh-atomic-bot r+ 8a12ed6

jlebon:
> bot, add author to whitelist

The matching regex was a bit too strict. I loosened it up a bit so that the sentence just needs to be on its own line. Now let's see if it works.

bot, add author to whitelist

PR #489, opened 2016-09-01
[merged] repo: Only use mmap() for metadata > 16k

cgwalters:
See http://stackoverflow.com/questions/258091/when-should-i-use-mmap-for-file-access
and
https://lwn.net/Articles/591978/

I didn't really notice much performance difference in some small
tests, but I happened to be stracing and realized we were `mmap()`ing
even for 50 bytes which is not very useful, so let's not do it.


alexlarsson:
@rh-atomic-bot r+  	ecbb972

Issue #487, opened 2016-09-01
Summary file races


dbnicholson:
Originally from https://github.com/flatpak/flatpak/issues/149.

There appear to be multiple races with the summary files.

1. The previous summary signature file is not removed until the new summary is regenerated. A client could get the old signature and new summary in this window.

2. The summary signature can't be generated until after the new summary file is in place. In addition to the race from 1, if the client has `gpg-verify-summary=true`, then there's a window where no signature exists.

I think there should be a `ostree_repo_regenerate_summary` variant that also takes GPG keys so that ostree can generate the signature and summary in one pass.

To make the race smaller, this new variant would generate the signature and summary in the temporary directory and rename them. However, that's still racy on the server side, and it doesn't fix the race on the client side where the files may be regenerated between the requests for the summary and signature file.

Here's 2 thoughts about making this really atomic:

> The other idea I had to go full atomic would be a summaries directory with a current symlink that can be atomically swapped sure of like the bootloader directory in ostree.

> So, even better would be to have an atomically replaced summary.sig, which contained a metadata field with the unique name of the summary file (we could keep the "summary" name a symlink to the latest summary for backwards compat). Then we could keep old summary files around for a while. This would fix clients racing with the summary being replaced.

CC @alexlarsson

dbnicholson:
#490 does the simple thing of creating the summary file and signing it in a temporary directory and then renaming them. I think the added API is necessary whichever direction this goes, and the implementation could be changed later to become fully atomic.

ramcq:
Could we add support for parsing summary files with inline signatures and just doing it that way...? Or add a summary.insig or such, which we check for first?

cgwalters:
To clarify the status of this PR...I'm not sure it's worth adding API just to *narrow* a race.

This was discussed some on list:
https://mail.gnome.org/archives/ostree-list/2016-October/msg00002.html

I'm pushing basically to fetch the summary file over pinned TLS and not GPG sign it.

We could however investigate doing inline signatures on the summary - this is basically [what Debian did](http://www.chiark.greenend.org.uk/~cjwatson/blog/no-more-hash-sum-mismatch-errors.html).  I'm not opposed to that, but again I think TLS is basically a better match for this, since it helps provide freshness guarantees like Notary does.

dbnicholson:
[...]

PR #490, opened 2016-09-01
Narrow summary file race

dbnicholson:
Provide a new function, `ostree_repo_regenerate_summary_ext`, that creates and signs the summary file in a tmpdir before renaming them into place. This minimizes the race between summary generation and signing without going fully atomic. Hopefully this is good enough for now since being atomic on the client side is quite a bit more complicated.

Closes: #487

### DIFF
--- a/src/libostree/libostree.sym
+++ b/src/libostree/libostree.sym
@@ -360,9 +360,14 @@ global:
  *                         NOTE NOTE NOTE
  */
 
+LIBOSTREE_2016.10 {
+global:
+        ostree_repo_regenerate_summary_ext;
+} LIBOSTREE_2016.8;
+
 /* Remove comment when first new symbol is added
-LIBOSTREE_2016.10
+LIBOSTREE_2016.11 {
 global:
 	someostree_symbol_deleteme;
-} LIBOSTREE_2016.8;
+} LIBOSTREE_2016.10;
  * Remove comment when first new symbol is added */
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -381,6 +381,7 @@ ostree_repo_verify_commit
 ostree_repo_verify_commit_ext
 ostree_repo_verify_summary
 ostree_repo_regenerate_summary
+ostree_repo_regenerate_summary_ext
 <SUBSECTION Standard>
 OSTREE_REPO
 OSTREE_IS_REPO
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -4196,22 +4196,13 @@ ostree_repo_sign_delta (OstreeRepo     *self,
   return FALSE;
 }
 
-/**
- * ostree_repo_add_gpg_signature_summary:
- * @self: Self
- * @key_id: (array zero-terminated=1) (element-type utf8): NULL-terminated array of GPG keys.
- * @homedir: (allow-none): GPG home directory, or %NULL
- * @cancellable: A #GCancellable
- * @error: a #GError
- *
- * Add a GPG signature to the summary file.
- */
-gboolean
-ostree_repo_add_gpg_signature_summary (OstreeRepo     *self,
-                                       const gchar    **key_id,
-                                       const gchar    *homedir,
-                                       GCancellable   *cancellable,
-                                       GError        **error)
+static gboolean
+internal_repo_add_gpg_signature_summary (OstreeRepo     *self,
+                                         int             dir_fd,
+                                         const gchar    **key_id,
+                                         const gchar    *homedir,
+                                         GCancellable   *cancellable,
+                                         GError        **error)
 {
   gboolean ret = FALSE;
   g_autoptr(GBytes) summary_data = NULL;
@@ -4220,11 +4211,11 @@ ostree_repo_add_gpg_signature_summary (OstreeRepo     *self,
   g_autoptr(GVariant) normalized = NULL;
   guint i;
 
-  summary_data = ot_file_mapat_bytes (self->repo_dir_fd, "summary", error);
+  summary_data = ot_file_mapat_bytes (dir_fd, "summary", error);
   if (!summary_data)
     goto out;
 
-  if (!ot_util_variant_map_at (self->repo_dir_fd, "summary.sig",
+  if (!ot_util_variant_map_at (dir_fd, "summary.sig",
                                G_VARIANT_TYPE (OSTREE_SUMMARY_SIG_GVARIANT_STRING),
                                OT_VARIANT_MAP_ALLOW_NOENT, &existing_signatures, error))
     goto out;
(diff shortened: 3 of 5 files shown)


src/libostree/libostree.sym
```diff
@@ -360,9 +360,14 @@ global:
...
  */
 
+LIBOSTREE_2016.10 {
+global:
+        ostree_repo_regenerate_summary_ext;
+} LIBOSTREE_2016.8;
+
 /* Remove comment when first new symbol is added
-LIBOSTREE_2016.10
+LIBOSTREE_2016.11 {
```
alexlarsson: Maybe the patch is just weirdly rendered, but it looks like the new function gets called .10 and the old that .11.

alexlarsson:
Seems like a good first step to me. If at some point we make it atomic on the client side we need this function anyway.

src/libostree/libostree.sym
dbnicholson: It's just a little weirdness in the ostree symbols file plus weirdness in symbol version files in general. This part is in a comment with a fake symbol waiting for the next release. My addition is above in the .10 section that chains to the .8 section. There's a little extra noise in this patch because the commented out .10 section prior to this patch (there weren't any new symbols yet) was missing the trailing `{`.

cgwalters:
Debian's bug here is similar:

https://bugs.launchpad.net/launchpad/+bug/1430011
http://www.chiark.greenend.org.uk/~cjwatson/blog/no-more-hash-sum-mismatch-errors.html

A few other ostree bugs related here:

https://bugzilla.gnome.org/show_bug.cgi?id=739143
https://bugzilla.gnome.org/show_bug.cgi?id=729388

So I see two things here.  First, we could  introduce `summary-$summaryhash.sigs` which is named by the hash of the summary.  We'd need garbage collection for older ones (probably want to keep a few around to avoid issues with proxies that give stale results).

Second, we could support inline signatures for the summary file.  I suspect most users would be happy enough with that.  The problem is though it'd be a hard format change unless we basically doubled the size and had an inline-signed copy?

Related to all of this...where I'd like to go is separating metadata fetches from content fetches.  The idea here is you fetch metadata *atomically* from more centralized servers (with e.g. TLS pinning), then you can get content from anywhere.  One could pretty easily keep the ostree metadata here (refs/summary/signatures) entirely in memory and update it atomically.  The client side of this is https://github.com/ostreedev/ostree/pull/469

cgwalters:
Finally related to all of this...I'd like to consider backing away from storing the static delta data in the summary file.  It can get expensive quickly.  Flatpak is relying on signed commits in the deltas anyways, and I'd like to push people towards pinned TLS + signed commits as simplest.  Metadata is hard, particularly when one involves signing.

alexlarsson:
@cgwalters  `summary-$summaryhash.sigs` seems the wrong direction. Then you have to download the potentially large summary before the signature, which breaks the "same summary => no need to redownload" optimization. I.e. since we download the summary.sig first, we should have a mapping from that (could be a simple name of the summary file stored in the summary) to the full summary file.

alexlarsson:
I mean "same summary sig => no need to redownload"

cgwalters:
So...I am not opposed to this at all, and sorry for introducing the race in the first place (metadata is hard).  But, if we're making changes here I'd like to think a little bit harder about what we can do to fix the race vs narrowing the window.

I think what has me hesitating a bit is adding new API without solving the race.

dbnicholson:
I'd been meaning to follow up here, too. Here's my other idea that removes the race (I think), but would require new client code. Let me see if I can get a reasonable tree representation:
```
├── summaries
│   ├── 41af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3
│   │   ├── summary
│   │   └── summary.sig
│   ├── be53d2ef39bda0f458c5cad8590e162a54a44f7c9a45a5017e539fa7b90192b1
│   │   ├── summary
│   │   └── summary.sig
│   └── current -> be53d2ef39bda0f458c5cad8590e162a54a44f7c9a45a5017e539fa7b90192b1
├── summary -> summaries/current/summary
├── summary.sig -> summaries/current/summary.sig
```
When creating a new summary, use a temporary directory under summaries. Optionally sign in the temporary directory. Rename the temporary directory to the checksum of the summary file. Update the current symlink to point to checksum directory. This removes the server side race.

On new clients, they'd download `summaries/current/summary`, take a checksum of the summary file, and then download `summaries/<checksum>/summary.sig`. They might have an old summary at that point, but they'll get the right signature.

Old clients would still race trying to download the toplevel `summary.sig` after downloading the `summary` file, but they'd no longer be exposed to the server side race.

There would need to be some cleanup of old summary directories. Maybe you'd keep 3?

BTW, if we're adding the `_ext` API (which I think we should), it should probably get a `flags` parameter in case anything is wanted there later.

alexlarsson:
The proposal from @dbnicholson means we always have to start by downloading the entire summary, instead of starting with the sig, before detecting that there has been no chnges. If we do that, then we have to use an etag or something to avoid downloading the large metadata if nothing changed.

Alternatively we can make the subdirectory be named by the hash of the signature. Then you can start by downloading and verifying that, and use its checksum to find the summary file. And for unsigned summaries you can just use the toplevel symlinks, because there is no race.

dbnicholson:
Yeah, sorry. I'd never thought of that optimization. I thought ostree always downloaded the summary file. Another issue i thought of with that setup is that the symlinks require you to host on a UNIX server. I think the rest of ostree has always allowed it to be hosted on any dumb http server.

So, my other thought is that instead of a symlink that summaries/current is a ref file with the checksum of the summary file, which is also the directory name as above. Then you can download the current file and quickly see if the checksum for the summary file has changed. If you want more verification, you can get the signature file from the checksum directory as above.

Without the symlinks, the top level compat files wouldn't be able to avoid the server side race. The best you could do is narrow it as in the patches here.

alexlarsson:
Do we *need* it? I dunno. Without signed summaries we can't protect against certain things. For instance, someone can MITM you and claim the "org.some.app" app actually contains the commit from "org.other.app" from the same remote. Is this bad? I dunno, its somewhat limited as an attack vector.

We do require signed summaries today though.

What do you mean by split metadata/content, can you describe it in a bit more detail?

alexlarsson:
Also, we store some extra info in the summary files, such as app permissions/metadata and sizes in the summary file. Those would be fakeable if not signed, although we could re-verify them after download to catch issues.

cgwalters:
https://mail.gnome.org/archives/ostree-list/2016-September/msg00015.html
https://mail.gnome.org/archives/ostree-list/2016-October/msg00002.html

dbnicholson:
I'm gonna close this since it doesn't actually address the client side race. I moved a couple pieces into #1490 - a new API that does the generation and signing in one call and use of a temporary directory for the `summary` and `summary.sig` file so that you don't publish mismatched files.

pwithnall:
Do you mean #1490?

PR #491, opened 2016-09-02
ostree-union-fuse: new fuse filesystem to access multiple commits

giuseppe:
it allows to union mount several commits, it is useful with Docker
imported images that are made of several layers, as:

ostree-union-fuse -o whiteouts -o repo=/ostree/repo/ -o layers=layer-1:layer-2 mnt

Currently "atomic mount" is using "checkout --union" for mounting an
image stored in OSTree.  While this is the ideal solution for checking
out an image or in general files to the file system, it has some
problems to keep the transient mount/unmount semantic that it has with
Docker images.  To workaround these problems, we were using xattr's or a
text file simulating those when xattr's are not available.

ostree-union-fuse has a series of improvements over it:
- no files are checked out to the file system.
- appears in /proc/mounts
- umount can be used too (not only atomic umount)
- on a crash/reboot no files are left on the system.
- we don't need xattr's to store the settings of the mount.

Signed-off-by: Giuseppe Scrivano gscrivan@redhat.com

### DIFF
--- a/src/union-fuse/Makefile-inc.am
+++ b/src/union-fuse/Makefile-inc.am
@@ -1,0 +1,24 @@
+# Copyright (C) 2016 Colin Walters <walters@verbum.org>
+# Copyright (C) 2016 Red Hat, Inc.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
+
+bin_PROGRAMS += ostree-union-fuse
+
+ostree_union_fuse_SOURCES = src/union-fuse/main.c
+
+ostree_union_fuse_CFLAGS = $(AM_CFLAGS) -D_GNU_SOURCE -D_FILE_OFFSET_BITS=64 $(BUILDOPT_FUSE_CFLAGS) $(ostree_bin_shared_cflags) $(OT_INTERNAL_GIO_UNIX_CFLAGS) -I$(srcdir)/libglnx $(NULL)
+ostree_union_fuse_LDADD = libotutil.la $(ostree_bin_shared_ldadd) $(BUILDOPT_FUSE_LIBS) $(OT_INTERNAL_GIO_UNIX_LIBS) libglnx.la
--- a/src/union-fuse/main.c
+++ b/src/union-fuse/main.c
@@ -53,6 +53,7 @@ opt_config
   char *repo_location;
   char *layers;
   int  whiteouts;
+  int  memcache;
 };
 
 static struct opt_config config;
@@ -68,6 +69,7 @@ static GArray *layers;
 
 static int objects_basefd;
 
+static GHashTable *memcache_dir;
 
 /* path is a buffer at least (OSTREE_SHA256_STRING_LEN + 8) bytes.  */
 static void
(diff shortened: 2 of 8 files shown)


giuseppe:
@rhatdan, this should improve how "atomic mount" works with containers stored in OSTree

giuseppe:
bot, retry please

src/union-fuse/Makefile-inc.am
```diff
@@ -0,0 +1,23 @@
+# Copyright (C) 2016 Colin Walters <walters@verbum.org>
```
cgwalters: Should have your name

src/union-fuse/main.c
```diff
@@ -0,0 +1,752 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright (C) 2015,2016 Colin Walters <walters@verbum.org>
```
cgwalters: Your name (and maybe mine too if you feel it derives from rofiles-fuse enough)

src/union-fuse/main.c
```diff
@@ -0,0 +1,752 @@
...
+ out:
+  return ret;
+}
+
+/* Access directly the file in the repository. This is much better than dealing with seek on an GInputStream,
+   and we support only BARE repositories anyway.  */
+static gchar *
+ostree_object_path (const char *checksum)
+{
+  return g_strdup_printf ("%s/objects/%.2s/%s.file", config.repo_location, checksum, checksum + 2);
```
cgwalters: We shouldn't malloc this - I'd use fd-relative like is used in core ostree.

cgwalters:
I've thought about writing this a few times.  So...the thing is that FUSE really has a nontrivial performance hit.  It may be OK for basically just read-only binaries with one service.  A good example though of where it's probably a much worse hit is building software - if every access to `/usr/include/foo.h` requires round trips through FUSE...

You have a lot of use of malloc() here, and in particular we should really use fd-relative APIs - see what rofiles-fuse does.  Also, performance will require switching to the FUSE low-level API (gives us multithreading capability).

I'm inclined to add this, and I think if we improve the performance enough it could be pretty interesting.

src/union-fuse/main.c
giuseppe: I have based it all on rofiles-fuse (copied both Makefile-inc.am and main.c) that is why I left the original copyright line.  I added the copyright line to Red Hat, since I've worked on it during my working hours

giuseppe:
I have fixed a few things in the fixup patch, the original version didn't really work correctly with subpaths and expected the same directories to be present in every layer.

Instead of building every time the path for whiteouts files (this would be even more expensive now that I fixed it to work for each component of the path), I maintain a cache for each layer.

I am also using openat for ostree `.file` files, keeping a reference to the `repo/objects` directory, instead of building the full path as in the first revision.

It seems multithreading is working with the high level FUSE interface as well.  I have done a quick experiment  in readdir printing when the program enters and leaves the function with a sleep in between and I could easily see that there are two threads entering it and wait for its termination.

cgwalters:
How critical is this for syscontainers?  It's going to take be a bit to review.

giuseppe:
not really critical.  System containers will do fine for now with checking out the image/container and rm'ing it

cgwalters:
So the core thing we're trying to fix here is:

"""
...has some
problems to keep the transient mount/unmount semantic that it has with
Docker images.
"""

Can we dive into this a bit more?  Is it basically that the atomic command offers a `mount` that just doesn't make sense with ostree storage today, and we want to have something that feels "mountable"?

I guess the flip side of this - I personally really like that OSTree is "just files".  It's even easier to inspect/manipulate.

What are the use cases for mounting?

giuseppe:
I've just rebased the series.

The main use case for `atomic mount` is to quickly inspect an image or a container file system by mounting it to a local directory.  `atomic mount` is used by `atomic scan`, which internally mount the images/containers to a temporary directory and scan their content for vulnerabilities.

When we mount a container/image from OSTree we do now:

- for images: we do a checkout and store this information in an xattr (where possible, otherwise in a file), so that `atomic umount` knows how to deal with it and can `rm -rf` the checkout directory.
- for containers: we do a bind mount to the existing checkout, since the container is already checked out on the file system.

the problems are:

- It is not possible to use directly `umount` as it is possible for Docker images
- If the user forgets to `atomic umount` the image, then the checkout won't be cleaned up

giuseppe:
I've rebased it again.


@rh-atomic-bot retry

giuseppe:
is there anything more left to do before it could be reviewed?

cgwalters:
The thing that has me hesitating a bit on this is it feels like it kind of competes with the core ostree idea.  When do we use FUSE versus hardlink checkouts?  I think we can safely say it doesn't make much sense to use FUSE for ostree-on-host case.  Do we think all containers should really use FUSE?  I'm not certain...it is certainly faster to initialize and gives us "laziness", but the flip side is it penalizes every read operation.

It'd be possible to implement "checkout bound to process lifecycle" by having a daemon involved - when a client requires a checkout, the daemon holds a fd for the client, and if the client goes away, the checkout gets cleaned up.  Alternatively, do this via dbus and monitor for the requester's bus name.

cgwalters:
This also overlaps with https://github.com/ostreedev/ostree/pull/578 and https://github.com/ostreedev/ostree/pull/619 - Alex's initial patches were doing squashing when importing, so the "union" behavior of this FUSE approach wouldn't be used, right?

cgwalters:
We may want to have an ostree-fuse-mount regardless...it seems like it'd be quite useful for testing, and I could imagine using it real cases too; a good example is where rpm-ostree needs to read just a few files out of the target root (/etc/os-release), and we can't teach libdnf how to read from ostree.

What I'm just trying to figure out is whether we're doing this because we want "process lifecycle binding" or because it's a faster way to read just a few files (some process wants to just grab e.g. the rpmdb from a tree).

giuseppe:
I agree with you that it is not a good idea to use this FUSE filesystem for running any container, while the cache improves the situation, it is still slower than a real checkout to the file system.

The only case I had in mount was to fix the `atomic mount` command, which mount an image for inspection or `atomic scan` which internally uses mount.  In this case we don't care about the performance as anyway the checkout has also a cost.

giuseppe:
and to answer your question more directly: I think the biggest reason is to have "mount lifecycle binding", once the mount is gone, there should be no files left, where now we require a `rm -rf` if it is an image.  I'd really like to drop the hacks that are now in `atomic` to allow a mount of images/containers stored in OSTree.
A faster way to read just a few files is more of a side effect IMO.

giuseppe:
one user case where this could be helpful:

https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2017-February/msg00009.html

cgwalters:
Just thinking through options here.  Another approach would be to basically do what docker's overlayfs2 backend does - check out each layer individually, then use overlayfs - if mounted `ro`, it should give similar semantics.

The benefit would be avoiding another union filesystem implementation.  Conceptually then this "ostree-oci-storage" driver would be very similar to `overlay2`, except it'd dedup across independent images.

giuseppe:
yes, that would work as well.  Just we don't checkout a single layer, we will need to keep a separate storage for the layers, so that they can be used for inspection by mount.  We will also need to teach `atomic purge` or whoever is responsible to also drop this checkout when the layer is removed from the repository.  What I like of the ostree-union-fuse approach is that it won't require any additional checkout.

cgwalters:
I don't think FUSE is the right *long term* architecture though.  FUSE is good for doing "slow/metadata" type filesystems.  The way Flatpak uses fuse for the "document portal" makes sense to me - copying data between apps is unusual, we're likely to only be copying one file at a time, etc.  FUSE also makes some sense for sshfs - it's a non-critical path development/testing convenience hack.  I'm sure there are people relying on sshfs for "production" usage but I'd argue there are better choices.

In contrast, FUSE as the "base /usr" for containers...there's a lot of random access going on.  Using FUSE for this loses all of the concurrency (multicore, RCU) etc work in the kernel VFS layer.  It obscures access controls such as SELinux.

Yes, one could say "why does ostree have rofiles-fuse then?" and the answer is that I still want to drive this into the kernel, but haven't put together the time.  It's also somewhat off the critical path because we only use `rofiles-fuse` for *build* time but not runtime.

So...I think this circles back a bit to "where does the OSTree+OCI logic live and how does it work" discussions.  It's definitely a mess between flatpak/atomic/skopeo/containers-storage/ostree, since it's very hard to share useful code between all 3 of golang/C/python.

giuseppe:
yes, I agree with you, this should not be used at all for containers.  For atomic this is going to be used/useful only in the case an image is inspected (atomic scan).  Hopefully files (and I guess a subset of the tree) are going to be accessed by the scanner only once.
The alternative to the fuse is to checkout the entire image, as we do now, but that has a high initial cost, and the checkout is going to be immediately deleted after the inspection.
Or as you suggested we could maintain each layer separately and use overlayfs to mount the final image.  That could work, but it will be used only for the "atomic scan" scenario as the image checkout at the moment doesn't use it.  We could probably do that as a long term plan for image checkouts as well, but we will then have the problem of who is going to setup the overlay fs before the container starts.

cgwalters:
It looks to me like `atomic scan` checks out into `/run/atomic`...which is going to mean we don't do hardlinks, which loses most of the point of ostree.  I think atomic needs to use `/sysroot/ostree/tmp/atomic-scan` for ostree-backed images.

cgwalters:
Moved that to https://github.com/projectatomic/atomic/issues/916

giuseppe:
alternative solution proposed here: https://github.com/projectatomic/atomic/pull/979

giuseppe:
the alternative solution got merged, could this PR still be useful or do I close it?

PR #494, opened 2016-09-07
[merged] gpg: do not segfault when the algorithm name is not known

giuseppe:
Reported by: Patrick Uiterwijk


cgwalters:
@rh-atomic-bot r+ 3922d71

PR #497, opened 2016-09-08
[merged] delta: Add missing `goto out` for failure to mmap()

cgwalters:
This was hit in practice when generating a delta for a flatpak app on ARM
it looks like.


alexlarsson:
@rh-atomic-bot r+  	8c640c0

PR #498, opened 2016-09-08
[merged] delta: Unreference files we've processed

cgwalters:
This should help avoid address space exhaustion on 32 bit systems,
and in general is obviously going to improve efficiency.


alexlarsson:
I mean, it doesn't hurt to free early, but we still mmap() everything up-front, so the maximum memory use is the same.

alexlarsson:
That said, it seems to help my (parallel) usecase enough that if fixes the OOM issues i had.
So:
@rh-atomic-bot r+ 6241ffb

PR #499, opened 2016-09-08
[merged] fetcher: Fix another finalization deadlock

cgwalters:
If the current repo is already up to date (we have no content to
fetch), it's possible for the fetcher to not request any URIs.  So
create and then finalize it quickly.

Finalization involves calling `g_main_loop_quit()` +
`g_thread_wait()`.  However, if `g_main_loop_quit()` is run *before*
`g_main_loop_run()`, we'll deadlock because GMainLoop assumes in
`_run()` to start things.

This is a common trap - ideally, GMainLoop would record if `_quit()`
was called before `_run()` or something, but doing that now would
likely break people who are expecting quit() -> run() to restart.

In general, we've moved in various GLib-consuming apps to an
explicit "main context iteration with termination condition" model;
see `pull_termination_condition()` in the pull code.

This fixes this race condition.

I verified that an assertion in `_finalize` that more than
zero URIs were requested was hit in multiple test cases, and this patch
has survived a while of make check loops.

### DIFF
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -502,6 +502,10 @@ ostree_fetcher_session_thread (gpointer data)
     }
   closure->max_outstanding = 3 * max_conns;
 
+  /* This model ensures we don't hit a race using g_main_loop_quit();
+   * see also what pull_termination_condition() in ostree-repo-pull.c
+   * is doing.
+   */
   while (g_atomic_int_get (&closure->running))
     g_main_context_iteration (closure->main_context, TRUE);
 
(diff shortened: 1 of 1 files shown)


jlebon:
Nice catch!
This makes sense to me, but I'll let @mbarnes have a look as well.

src/libostree/ostree-fetcher.c
```diff
@@ -503,7 +502,8 @@ ostree_fetcher_session_thread (gpointer data)
     }
   closure->max_outstanding = 3 * max_conns;
 
-  g_main_loop_run (closure->main_loop);
+  while (g_atomic_int_get (&closure->running))
```
jlebon: Maybe a comment here about why we're using our own loop vs. Glib?

mbarnes:
Yeah, I've encountered this kind of startup race before.  At the time I solved it with a `GCond`+`GMutex` but I like your way better.  :+1:

jlebon:
@rh-atomic-bot r+

PR #500, opened 2016-09-08
[merged] sysroot: Avoid double cleanup, and ensure no cleanup if specified

cgwalters:
Since forever, we've been doing two cleanups.  In
https://github.com/ostreedev/ostree/commit/8ece4d6d51bdbe3e41ab318259276bb83e553aa0
I thought we were doing just one and wanted to go to zero (if specified),
but I actually just dropped one cleanup.

In https://github.com/projectatomic/rpm-ostree/pull/452
@jlebon pointed out the duplication.  Fix this by creating a new internal
deploy wrapper that takes cleanup flags.

(Since we already had the "piecemeal cleanup" API internally, let's
 frame it in terms of that, rather than passing down a boolean).

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1713,6 +1713,18 @@ ostree_sysroot_write_deployments (OstreeSysroot     *self,
                                   GCancellable      *cancellable,
                                   GError           **error)
 {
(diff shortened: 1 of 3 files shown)


jlebon:
LGTM!
@rh-atomic-bot r+ 56d1844

PR #501, opened 2016-09-08
[merged] core: Make OSTREE_TIMESTAMP public API

cgwalters:
This way e.g. flatpak can detect which timestamp it should use.
See `flatpak/common/flatpak-utils.c:flatpak_zero_mtime`.


cgwalters:
@rh-atomic-bot r+ c8d5f28

PR #506, opened 2016-09-12
[merged] pull: Do allow executing deltas when mirroring into bare{,-user}

cgwalters:
In https://github.com/ostreedev/ostree/pull/408 we fixed a
bug where we would crash when trying to execute deltas into
an archive repo (which isn't presently supported).

But that was overly aggressive - we obviously *can* execute deltas
when mirroring into a bare repo.  This should fix a regression with
the way flatpak uses mirroring to pull from a user repo into the
system.


alexlarsson:
@rh-atomic-bot r+  	6fc67f1

alexlarsson:
Seems homu failed somehow?

PR #507, opened 2016-09-13
[merged] File descriptor fixes

owtaylor:
Here are some fixes for closing file descriptors - the problem with cache_dir_fd in ostree-repo.c was found because I saw a close(-1) when hunting the second bug which, was causing problems for gnome-continuous.


cgwalters:
Ouch, thanks.

@rh-atomic-bot r+ dae3f88

PR #508, opened 2016-09-14
[merged] sysroot: Drop an unnecessary fsync

cgwalters:
While looking at a slow update issue (which I'm guessing is
unpredictable I/O latency in an OpenStack instance), I noticed
in one of the traces we were inside a fsync here.

Dropping the fsync here is just another of a long series of unwinding
them - we `syncfs()` the sysroot fd and `/boot` and we have a big
`sync()` anyways.


cgwalters:
Rebased :surfer:

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1827,9 +1827,10 @@ _ostree_sysroot_write_deployments_internal (OstreeSysroot     *self,
     {
       int new_bootversion = self->bootversion ? 0 : 1;
```
jlebon: Don't mind me, just testing the new review feature!

src/libotutil/ot-gio-utils.c
```diff
@@ -315,64 +315,28 @@ ot_gfile_ensure_unlinked (GFile         *path,
 }
 
 /**
- * ot_util_ensure_directory_and_fsync:
+ * ot_util_fsync_directory:
```
jlebon: Why not just remove it? Are we planning to make use of it in the future?

cgwalters: Yep, it was a bad merge.  Fixed.

jlebon:
I guess review feedback doesn't show up as comments in the API. 
@rh-atomic-bot r+ ab068df

PR #510, opened 2016-09-14
[merged] sysroot: Port some small cleanup code to fd-relative

cgwalters:
Just a quick patch since I saw this function scroll by in Emacs and it
was too ugly not to be rewritten.


giuseppe:
@rh-atomic-bot r+ be2eba5

PR #511, opened 2016-09-14
[merged] boot: Ensure we remount /var writable before systemd does journal flush

cgwalters:
Otherwise, we may not get a persistent journal for the first boot.
https://bugzilla.redhat.com/show_bug.cgi?id=1265295


jlebon:
@rh-atomic-bot r+ 1fa7cbc

PR #512, opened 2016-09-14
[merged] Trivial httpd no gfile

cgwalters:


jlebon:
Nice!
@rh-atomic-bot r+ a8b96f6

PR #513, opened 2016-09-20
[merged] ostree_sysroot_init_osname: also create /var/log

jlebon:
/var/log is another one of those core directories that should be made
available and properly labeled during early boot before tmpfiles.d
starts up.

Related: https://bugzilla.redhat.com/show_bug.cgi?id=1265295


cgwalters:
OK for now, though I think the real fix is probably in systemd.

@rh-atomic-bot r+ 7cff781

Issue #515, opened 2016-09-22
Allow ostree admin status unprivileged


dbnicholson:
Currently you can only run `ostree admin status` as root since it takes the sysroot lock. I think it's nice to allow unprivileged users to get the status of the system. After all, they could dig around in the repo and do it themselves, anyway.

PR coming.

PR #516, opened 2016-09-22
[merged] admin: Allow running status unlocked

dbnicholson:
It's useful to let non-root see the current system status.

Closes: #515

### DIFF
--- a/src/ostree/ot-admin-builtin-status.c
+++ b/src/ostree/ot-admin-builtin-status.c
@@ -97,7 +97,7 @@ ot_admin_builtin_status (int argc, char **argv, GCancellable *cancellable, GErro
   context = g_option_context_new ("List deployments");
 
   if (!ostree_admin_option_context_parse (context, options, &argc, &argv,
-                                          OSTREE_ADMIN_BUILTIN_FLAG_NONE,
+                                          OSTREE_ADMIN_BUILTIN_FLAG_UNLOCKED,
                                           &sysroot, cancellable, error))
     goto out;
 


dbnicholson:
Ping. Any thoughts on this? Is there any way to make the bots build again? I'm pretty confident the failure on travis is not because of this change.

src/ostree/ot-admin-builtin-status.c
```diff
@@ -97,7 +97,7 @@ ot_admin_builtin_status (int argc, char **argv, GCancellable *cancellable, GErro
   context = g_option_context_new ("List deployments");
 
   if (!ostree_admin_option_context_parse (context, options, &argc, &argv,
-                                          OSTREE_ADMIN_BUILTIN_FLAG_NONE,
+                                          OSTREE_ADMIN_BUILTIN_FLAG_NONE | OSTREE_ADMIN_BUILTIN_FLAG_UNLOCKED,
```
jlebon: You can just replace the argument here instead of ORing.

jlebon:
Yeah, not sure if there's a way to retrigger the Travis build. There's a known race condition that manifests itself much more easily on Travis. As long as the other CI status is green, we're OK :)

src/ostree/ot-admin-builtin-status.c
dbnicholson: How silly. I cargo culted that from another builtin, but clearly the `|` is useless. I'll fix that up.

dbnicholson:
Pushed the fix for that.

jlebon:
@rh-atomic-bot r+ 707a55f

cgwalters:
This is OK, though I'd say what we probably wanted here was a flag like "lock only if root" or something.  It's basically what RPM ends up doing with `rpm -qa` - unprivileged users can't grab a lock, so it just ignores it.  I'm sure dpkg/etc do the same.

In the bigger picture though, the fix for all things like this is to use a DBus API, which e.g. rpm-ostree already has.

PR #520, opened 2016-10-04
[merged] checkout: Fix fsync defaults for new API to be off for real

cgwalters:
My previous change in https://github.com/ostreedev/ostree/pull/425
actually broke things so we basically used the repository defaults =(

This is a subtle mess since we're only trying to flip things off
for the *new* API.

Clean this up so that the "default repo inheritance" lives only in one
place - in the compat layer for the old checkout API.  The new
checkout API defaults to off period, so the repository state is
irrelevant.


jlebon:
@rh-atomic-bot r+ 012b686

PR #522, opened 2016-10-05
[merged] libglnx: Update to latest

cgwalters:
In preparation for a future update.  Making this one now because
we need to add some includes since glnx-libcontainer went away,
and with it some includes for `sys/mount.h` etc.


cgwalters:
This one is small enough, I'd like it in the next release, so self-reviewing.

@rh-atomic-bot r+ b5442c4

PR #523, opened 2016-10-05
[merged] pull: Support multiple specifications of --subpath

alexlarsson:
I need this in flatpak to avoid doing multiple pulls when doing
locale subsetting.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -99,7 +99,7 @@ typedef struct {
   gboolean          is_commit_only;
   gboolean          is_untrusted;
 
-  char         *dir;
+  GPtrArray        *dirs;
   gboolean      commitpartial_exists;
 
   gboolean      have_previous_bytes;
@@ -117,6 +117,7 @@ typedef struct {
 typedef struct {
   OtPullData  *pull_data;
   GVariant    *object;
+  char        *path;
   gboolean     is_detached_meta;
 
   /* Only relevant when is_detached_meta is TRUE.  Controls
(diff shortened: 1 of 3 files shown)


cgwalters:
Overall looks sane, there is just some nontrivial code here I want to take a bit of time later to read through carefully.

alexlarsson:
I pushed a (trivially) rebased version that applies again.

src/libostree/ostree-repo-pull.c
```diff
@@ -1101,10 +1157,10 @@ scan_commit_object (OtPullData         *pull_data,
         goto out;
 
       queue_scan_one_metadata_object_c (pull_data, tree_contents_csum_bytes,
-                                        OSTREE_OBJECT_TYPE_DIR_TREE, recursion_depth + 1);
+                                        OSTREE_OBJECT_TYPE_DIR_TREE, g_strdup ("/"), recursion_depth + 1);
```
cgwalters: This is a leak, right?

src/libostree/ostree-repo-pull.c
```diff
@@ -381,12 +388,80 @@ static void
 enqueue_one_object_request (OtPullData        *pull_data,
                             const char        *checksum,
                             OstreeObjectType   objtype,
+                            const char        *path,
                             gboolean           is_detached_meta,
                             gboolean           object_is_stored);
 
 static gboolean
+matches_pull_dir (const char *current_file,
```
cgwalters: I had to read this twice at least, but it does make sense.

alexlarsson: If it was hard to write it should be hard to read...

Honestly, I don't know why things like this are so tricky. I wonder if there is a way to formulate it that is better.

cgwalters:
@rh-atomic-bot r+ 4084be1

PR #525, opened 2016-10-07
[merged] pull: Support inherit-transaction

alexlarsson:
If this is true, don't initiate, abort or commit a transaction, instead
it is assumed that the caller initiated the transaction, and that it
will eventually be commited.

This allows you to do multiple pulls or a combination of pulls and
commits in a single transaction.


cgwalters:
@rh-atomic-bot r+ 2931f7a

PR #529, opened 2016-10-14
[merged] Fix proxy authentication

jlebon:
Make sure to pass the proxy authentication if asked for it. Though libsoup should do that by default (see https://bugzilla.gnome.org/show_bug.cgi?id=772932).

Also includes a few patches I had lying around. :)

### DIFF
--- a/docs/manual/deployment.md
+++ b/docs/manual/deployment.md
@@ -43,14 +43,14 @@ deployment.  For short, we will call this the "tree", to
 distinguish it from the concept of a deployment.
 
 First, the tree must include a kernel stored as
-`/boot/vmlinuz(-.*)?-$checksum`.  The checksum should be a SHA256 hash
-of the kernel contents; it must be pre-computed before storing the
-kernel in the repository.  Optionally, the tree can contain an
-initramfs, stored as `/boot/initramfs(-.*)?-$checksum`.  If this
-exists, the checksum must include both the kernel and initramfs
-contents.  OSTree will use this to determine which kernels are shared.
-The rationale for this is to avoid computing checksums on the client
-by default.
+`vmlinuz(-.*)?-$checksum` in either `/boot` or `/usr/lib/ostree-boot`.
+The checksum should be a SHA256 hash of the kernel contents; it must be
+pre-computed before storing the kernel in the repository.  Optionally,
+the directory can also contain an initramfs, stored as
+`initramfs(-.*)?-$checksum`.  If this exists, the checksum must include
+both the kernel and initramfs contents.  OSTree will use this to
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-fetcher.c
```diff
@@ -277,6 +277,24 @@ session_thread_config_flags (ThreadClosure *thread_closure,
 }
 
 static void
+on_authenticate (SoupSession *session, SoupMessage *msg, SoupAuth *auth,
+                 gboolean retrying, gpointer user_data)
+{
+  if (msg->status_code == SOUP_STATUS_PROXY_UNAUTHORIZED) {
```
cgwalters: Brace should be on a newline here.

src/libostree/ostree-fetcher.c
```diff
@@ -277,6 +277,24 @@ session_thread_config_flags (ThreadClosure *thread_closure,
...
+    SoupURI *uri = NULL;
+    g_object_get (session, SOUP_SESSION_PROXY_URI, &uri, NULL);
+    if (retrying)
+      {
+        g_autofree char *s = soup_uri_to_string (uri, FALSE);
+        g_warning ("Invalid username or password for proxy '%s'", s);
```
cgwalters: I really don't like `g_warning()` as a general rule, particularly for things that can happen "normally" from things like incorrect passwords.

Do we really have no way to propagate this into a proper error back to the caller?  Maybe we have to do something like cache the error and return it later.

docs/manual/deployment.md
```diff
@@ -43,13 +43,14 @@ deployment.  For short, we will call this the "tree", to
...
-the kernel contents; it must be pre-computed before storing the kernel
-in the repository.  Optionally, the tree can contain an initramfs,
-stored as `/boot/initramfs-$checksum`.  If this exists, the checksum
-must include both the kernel and initramfs contents.  OSTree will use
-this to determine which kernels are shared.  The rationale for this is
-to avoid computing checksums on the client by default.
+`/boot/vmlinuz(-.*)?-$checksum`.  The checksum should be a SHA256 hash
+of the kernel contents; it must be pre-computed before storing the
+kernel in the repository.  Optionally, the tree can contain an
+initramfs, stored as `/boot/initramfs(-.*)?-$checksum`.  If this
```
cgwalters: While you're amending this, might as well mention all of this content can now be in either `/usr/lib/ostree-boot/` or `/boot`.

jlebon:
Now with less warnings and more errors! ^

cgwalters:
@rh-atomic-bot r+ ba97f3e

PR #531, opened 2016-10-17
[merged] Ostree cookies

sjoerdsimons:
Add support for adding cookies to the http requests sent to repositories to enable e.g. the usage of signed cookies with cloudfront for private repositories.

### DIFF
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -1057,6 +1057,7 @@ on_request_sent (GObject        *object,
               switch (msg->status_code)
                 {
                 case 404:
+                case 403:
                 case 410:
                   code = G_IO_ERROR_NOT_FOUND;
                   break;
--- a/src/libostree/ostree-fetcher.c
+++ b/src/libostree/ostree-fetcher.c
@@ -326,6 +326,16 @@ session_thread_set_proxy_cb (ThreadClosure *thread_closure,
     }
 }
 
+static void
+session_thread_set_cookie_jar_cb (ThreadClosure *thread_closure,
+                                  gpointer data)
+{
+  SoupCookieJar *jar = data;
+
+  soup_session_add_feature (thread_closure->session,
+                            SOUP_SESSION_FEATURE (jar));
+}
+
 #ifdef HAVE_LIBSOUP_CLIENT_CERTS
 static void
 session_thread_set_tls_interaction_cb (ThreadClosure *thread_closure,
@@ -744,6 +754,23 @@ _ostree_fetcher_set_proxy (OstreeFetcher *self,
                                proxy_uri,  /* takes ownership */
                                (GDestroyNotify) soup_uri_free);
     }
+}
+
+void
+_ostree_fetcher_set_cookie_jar (OstreeFetcher *self,
+                                const char    *jar_path)
+{
+  SoupCookieJar *jar;
+
+  g_return_if_fail (OSTREE_IS_FETCHER (self));
+  g_return_if_fail (jar_path != NULL);
+
+  jar = soup_cookie_jar_text_new (jar_path, TRUE);
+
+  session_thread_idle_add (self->thread_closure,
+                           session_thread_set_cookie_jar_cb,
+                           jar,  /* takes ownership */
+                           (GDestroyNotify) g_object_unref);
 }
 
 void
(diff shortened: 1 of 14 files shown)


src/libostree/ostree-fetcher.c
```diff
@@ -747,6 +757,23 @@ _ostree_fetcher_set_proxy (OstreeFetcher *self,
...
 void
+_ostree_fetcher_set_cookie_jar (OstreeFetcher *self,
+                                const char    *jar_path)
+{
+  SoupCookieJar *jar;
+
+  g_return_if_fail (OSTREE_IS_FETCHER (self));
+  g_return_if_fail (jar_path != NULL);
+
+  jar = soup_cookie_jar_text_new(jar_path, TRUE);
```
jlebon: Nit: missing a space here between before the parenthesis.

src/libostree/ostree-repo-pull.c
```diff
@@ -1969,6 +1969,19 @@ _ostree_repo_remote_new_fetcher (OstreeRepo  *self,
       _ostree_fetcher_set_proxy (fetcher, http_proxy);
   }
 
+  {
+    g_autofree char *jar_path = NULL;
+    g_autofree char *cookie_file = g_strdup_printf ("%s.cookies.txt",
+                                                    remote_name);
+
+    jar_path = g_build_filename (g_file_get_path (self->repodir), cookie_file,
+                                                  NULL);
```
jlebon: Nit: that indentation looks off.

src/ostree/ot-builtin-trivial-httpd.c
```diff
@@ -61,6 +63,7 @@ static GOptionEntry options[] = {
   { "random-500s", 0, 0, G_OPTION_ARG_INT, &opt_random_500s_percentage, "Generate random HTTP 500 errors approximately for PERCENTAGE requests", "PERCENTAGE" },
   { "random-500s-max", 0, 0, G_OPTION_ARG_INT, &opt_random_500s_max, "Limit HTTP 500 errors to MAX (default 100)", "MAX" },
   { "log-file", 0, 0, G_OPTION_ARG_FILENAME, &opt_log, "Put logs here", "PATH" },
+  { "expected-cookies", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_expected_cookies, "Put logs here", "KEY=VALUE" },
```
jlebon: I suppose you meant to update the description here? :)

sjoerdsimons:
@cgwalters I'll update the pull request for the style nitpicks you and jlebon found soonish. For the architecture stuff. 

Yes it's assumed another entity on the device will both manage the cookies (update as needed) and do the general management of trigger of updates etc. It can either do that through calling add-cookie or by updating the cookies file directly (Though the latter is a bit more fragile as libsoup is picky on the format).

Wrt. security it's all about how high a barrier you'd want to put in place. For this usage just putting the ostree repositories (and thus their device contents) wasn't acceptable, the potential that access to the distributed software is potentially obtained by direct access to the device is however accepted and understood.  Depending on the end-usage, secure boot should be in place with an option for an encrypted filesystem so even with direct access this is non-trivial, however the main goal there is to protect the end-user data not necessarily the system as such. Remote attestation would be interesting, but it's rather complex to achieve at the moment.

This is going a bit off-topic on this patchset (which really is just another authentication mechanism), but if you're thinking that level of security one thing I would like to see in ostree deployments would be the same sort of guarantees as dm-verity. Iotw that all data coming from the system partition is verified at load/read time.

Wrt. the usage of an external cookie jar rather then folding it into config, I partially did that because of the external cookie management, partially because that's easier with libsoup and partially because cookies can be arbitrarily long chunks of base64'd data which seemed ugly to have in the repo config :) (At some point i also considered using the libsoup sqlite cookie store, but for what is expected to be a small list of cookies the plain text variant seems more beneficial).

And for future changes, sure, happy to file issues before doing implementation and discussing the strategy a bit :)

Thanks for the feedback!

cgwalters:
Cookies are up to 4k, but I'd guess the cloudfront ones are a lot smaller than that.  The code seems like it'd be a lot smaller if we just embedded the cookies into the remote config to start, and that'd give a clearer path for data sync.

Though, the fact that we already have per-remote GPG keys that are out of band is also precedence for that approach.  On the other other hand, if we supported [ED25519](https://ed25519.cr.yp.to/) for signatures like the other cool kids, that'd easily fit in the remote config too...

sjoerdsimons:
fwiw cloudfronts biggest one seems to be around 350 bytes.

sjoerdsimons:
As remarked in another pull request, ping for branch update as github seems to fail to notify about update nicely :). All review notes should be covered (travis is unhappy due to what seems to be an infrasture issue)

cgwalters:
bot, add author to whitelist

cgwalters:
Pushed two fixups.  Thanks for the patch!

@rh-atomic-bot r+ ed7fbc7

PR #535, opened 2016-10-19
[merged] add .redhat-ci.yml and .redhat-ci.Dockerfile

jlebon:
Add a YAML file for the new Red Hat CI framework.

Rather than re-installing all the build deps everytime, which takes time, I added a Dockerfile that we can wire up to the Docker Hub. For now it lives at `jlebon/ostree-tester:rhci`, but we can move it under the `projectatomic` org (or a new `ostree` org).

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -1,0 +1,25 @@
+branches:
+    - master
+    - auto
+    - try
+
+container:
+    image: jlebon/ostree-tester:rhci
+
+tests:
+    - sh autogen.sh
+        --prefix=/usr
+        --libdir=/usr/lib64
+        --enable-installed-tests
+        --enable-gtk-doc
+    - make -j2
+    - make syntax-check
+    - make check
+    - make install
+    - gnome-desktop-testing-runner ostree
+    - sudo --user=testuser gnome-desktop-testing-runner ostree
+
+timeout: 30m
+
+artifacts:
+    - test-suite.log
--- a/.redhat-ci.Dockerfile
+++ b/.redhat-ci.Dockerfile
@@ -1,0 +1,21 @@
+FROM fedora:24
+MAINTAINER Jonathan Lebon <jlebon@redhat.com>
+
+RUN dnf install -y \
+        gcc \
+        sudo \
+        which \
+        attr \
+        fuse \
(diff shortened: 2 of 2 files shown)


.redhat-ci.yml
```diff
@@ -0,0 +1,25 @@
+branches:
+    - master
+    - auto
+    - try
+
+container:
+    image: jlebon/ostree-tester:rhci
+
+tests:
+    - sh autogen.sh
```
cgwalters: Why the explicit `sh`?

.redhat-ci.yml
```diff
@@ -0,0 +1,25 @@
...
+        --prefix=/usr
+        --libdir=/usr/lib64
+        --enable-installed-tests
+        --enable-gtk-doc
+    - make -j2
```
cgwalters: Not many cores eh? :smile:  Down the line this seems like something we'd want to inherit from the system, but fine for now.

.redhat-ci.yml
```diff
@@ -0,0 +1,25 @@
+branches:
+    - master
+    - auto
+    - try
+
+container:
+    image: jlebon/ostree-tester:rhci
```
cgwalters: The problem with this then is that if it's only updated manually, it's going to bitrot...

At some point of course, we should implement `rpm-ostree assemble-buildroot` or something, which would be fast *and* accurate.

jlebon: Right, this will definitely need to be set up as an automated build. Any preferences on which org? (Or I don't mind leaving it under my name either).

.redhat-ci.yml
jlebon: Heh, force of habit I suppose. :)

cgwalters:
Big picture, I'd like to down the line abstract this further so that we can optimize for the case of "CI on a single git repo that's *already* packaged" and use [rpmdistro-gitoverlay](https://github.com/cgwalters/rpmdistro-gitoverlay/) under the hood, and that tool should learn how to use rpm-ostree for buildroots.

(And likely be able to pull into an existing chain, a bit like where I was going with the CentOS CI bits)

But this is a good start.

@rh-atomic-bot r+ 374151b

Issue #537, opened 2016-10-21
Regression for flatpak


alexlarsson:
The commit https://github.com/ostreedev/ostree/commit/1d4f1b8878addd28059c3a3928640491755cd615 breaks flatpak for me. I keep getting permission errors when i run binaries. I'll look into it.

alexlarsson:
Yes, it breaks bare-user + checkout -U:

```
$ mkdir x
$ mkdir x/dir
$ ln -s dir x/symlink
$ ostree --repo=repo init --mode=bare-user
$ ostree --repo=repo commit  --branch=master x
abd3b36def5f10361cfcb73d57cfe117694e89b753fd47ad86558813c848bf39
$ ostree --repo=repo checkout -U master new_x
[jhbuild execute] ls -l x new_x/
new_x/:
total 4
drwxr-xr-x. 1 alex wheel 0  1 jan  1970 dir
-rw-r--r--. 2 alex wheel 4  1 jan  1970 symlink

x:
total 4
drwxr-xr-x. 1 alex wheel 0 21 okt 11.04 dir
lrwxrwxrwx. 1 alex wheel 3 21 okt 11.04 symlink -> dir
```

alexlarsson:
Ok. so the issue is that bare-user doesn't store symlink objects as symlinks in the repo:

```
repo/objects/7c:
-rw-r--r--. 2 alex wheel 4  1 jan  1970 04554307d6ceb331e699a4d0768d1a5062845d75df10a6d96d2c4995e979fd.file
```

vs
```
repo-bare/objects/7c:
lrwxrwxrwx. 1 alex wheel 3 21 okt 11.07 04554307d6ceb331e699a4d0768d1a5062845d75df10a6d96d2c4995e979fd.file -> dir
```

PR #538, opened 2016-10-21
[merged] Fix regression for symlinks in bare-user repos

alexlarsson:
Commit 1d4f1b8878addd28059c3a3928640491755cd615 started using hardlinks
checkouts of symlinks. However, symlinks are not stored as symlink in the
repo for bare-user repos, so this breaks user-mode checkouts of such repos.

We fix this by checking for !is_symlink in the bare-user case.

This fixes:
     https://github.com/ostreedev/ostree/issues/537


giuseppe:
@rh-atomic-bot r+ 2427356

Issue #526, opened 2016-10-11
Commitmeta read/write not part of transaction


alexlarsson:
ostree_repo_read/write_commit_detached_metadata never ever looks in the staging directory, so any update of this as part of e.g. a pull or signing something during a transaction will affect the repo directly, unaffected by the transaction state.

I'm not sure if this is by design or not, but it seems a bit iffy to me.

PR #539, opened 2016-10-21
[merged] Fix some issues with detached metadata handling

alexlarsson:
This fixes two issues:
1) Reading detached metadata from the parent repo failed (this breaks in some cases when updating the system repo with flatpak)
2) Detached metadata writes go directly to the repo, even if there is a staging directory because there is an active transaction.


alexlarsson:
Fixes https://github.com/ostreedev/ostree/issues/526

giuseppe:
it looks good to me

@rh-atomic-bot r+ e8291ad

Issue #541, opened 2016-10-22
ostree prune is problematic with partial commits


alexlarsson:
ostree prune calls ostree_repo_traverse_commit_union() on all commits, but if .commitpartial exists we ignore a NOT_FOUND error.

However, we still terminate the traverse at the first error, so we don't find later objects that *do* exist. Since these are not found ostree purge assumes they are unreached and removes them!

PR #542, opened 2016-10-24
[merged] Fix pruning of partial commits

alexlarsson:
If we have a partial commit it is not an error for a dirmeta to be
missing (in fact, that is likely), so instead of returning a not-found
error from ostree_repo_traverse_commit() we ignore the error and
continue.

In particular, this means we don't stop early at the first
missing dirmeta, which previously caused ostree_repo_prune() to
thing the dirmetas after that to be unreached and thus purged.

Also, we remove the special casing in ostree_repo_prune() to
not report errors for commitpartial, because these should not
be reported anymore.

This fixes https://github.com/ostreedev/ostree/issues/541

### DIFF
--- a/src/libostree/ostree-repo-prune.c
+++ b/src/libostree/ostree-repo-prune.c
@@ -311,29 +311,11 @@ ostree_repo_prune (OstreeRepo        *self,
       while (g_hash_table_iter_next (&hash_iter, &key, &value))
         {
           const char *checksum = value;
-          OstreeRepoCommitState commitstate;
-          GError *local_error = NULL;
-
-          if (!ostree_repo_load_commit (self, checksum, NULL, &commitstate,
-                                        error))
-            goto out;
 
           g_debug ("Finding objects to keep for commit %s", checksum);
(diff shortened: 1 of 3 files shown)


cgwalters:
Oh ugh! :japanese_goblin: 

Now if I'm understanding this correctly...the fallout from this bug would just be that we'd prune already downloaded objects for a partial commit, right?  IOW we're not doing anything really bad like unlinking objects that *are* referenced by other commits or so.

When filing issues/commits I really like to see at least a sentence for the real world situation; it could be something like:  "I noticed we were pruning already downloaded objects in flatpak" or something.  Is that the case?

alexlarsson:
Yes what happens is we pull a partial commit, deploy it and then afterwards we prune which causes the objects to be removed. We're not unlinking objects that are referenced by other non-partial commits.

However things get quite bad when we later update the branch due to https://github.com/ostreedev/ostree/issues/543 because the removed objects are likely needed when we update (if the change were to a different part of the app) and that issue causes us to not re-pull the pruned objects.

I have a workaround for this in flatpak though.

cgwalters:
@rh-atomic-bot r+ cae4d81

PR #544, opened 2016-10-24
[merged] docs: Link to releng-scripts

cgwalters:
Now that the repo starts to implement some of this stuff.


jlebon:
@rh-atomic-bot r+ f9c94c6

jlebon:
Not sure what's going on here. The right status is set on the auto commit. Seems like homu missed that request somehow.
@rh-atomic-bot retry

PR #546, opened 2016-10-25
[merged] tests: Skip libarchive/selinux tests if in container without SELinux

cgwalters:
I'm doing builds and `make check` inside a Docker container, with
selinux on as a build-time option, but no policy in the container.
This currently aborts.  Let's not do that.

(This type of thing is why installed tests are a better model)


jlebon:
@rh-atomic-bot r+ 9a9ca8a

PR #547, opened 2016-10-25
[merged] tree-wide: Use macros to initialize GVariant{Builder,Dict}

cgwalters:
With at least `gcc-6.2.1-2.fc25.x86_64` I was seeing:

```
src/libostree/ostree-repo.c:4688:43: warning: missing braces around initializer [-Wmissing-braces]
     g_auto(GVariantDict) deltas_builder = {{0,}};
```

etc.  GLib has macros to initialize things, let's just use them.

### DIFF
--- a/src/libotutil/otutil.h
+++ b/src/libotutil/otutil.h
@@ -26,6 +26,13 @@
 #include <string.h> /* Yeah...let's just do that here. */
 #include <libglnx.h>
 
+/* https://bugzilla.gnome.org/show_bug.cgi?id=766370 */
+#if !GLIB_CHECK_VERSION(2, 49, 3)
+#define OT_VARIANT_BUILDER_INITIALIZER {{0,}}
+#else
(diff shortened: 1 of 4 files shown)


cgwalters:
Ugh, requires 2.50, and it's going to be annoying to make it warning-free in both cases since the struct definition changed.

src/libotutil/otutil.h
```diff
@@ -26,6 +26,13 @@
 #include <string.h> /* Yeah...let's just do that here. */
 #include <libglnx.h>
 
+/* https://bugzilla.gnome.org/show_bug.cgi?id=766370 */
+#if !GLIB_CHECK_VERSION(2, 50, 0)
```
jlebon: Seems like this should be 2.49.3, no?

cgwalters: Fixed, though in practice I don't think anyone is going to be using random pre-release snapshots of GLib *after* the release.

jlebon:
@rh-atomic-bot r+ 3513924

PR #548, opened 2016-10-25
[merged] Clang fixes

cgwalters:


jlebon:
That's awesome! We should probably get a clang compiler going in the Red Hat CI.
@rh-atomic-bot r+ 272f829

cgwalters:
Yeah, that's what I was looking at.

PR #549, opened 2016-10-27
[merged] .redhat-ci.yml: add clang

jlebon:
Clang has better detection for unused vars when using auto cleanup
functions. We should eventually just fold this back into the first
testsuite. But let's just turn it on for now, at least until it's
satisfied with the whole codebase.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -23,3 +23,22 @@ timeout: 30m
 
 artifacts:
     - test-suite.log
+
+---
+
+inherit: true
+
+packages:
+  - clang
+
+tests:
+    - sh autogen.sh
+        --prefix=/usr
+        --libdir=/usr/lib64
+        --enable-installed-tests
+        --enable-gtk-doc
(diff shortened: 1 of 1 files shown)


.redhat-ci.yml
```diff
@@ -37,7 +37,7 @@ tests:
         --libdir=/usr/lib64
         --enable-installed-tests
         --enable-gtk-doc
-    - make -j2 CC=clang CFLAGS='-Wall'
+    - make -j2 CC=clang CFLAGS='-Werror'
```
cgwalters: I'd prefer just `-Werror=unused-variable`.  See https://mail.gnome.org/archives/desktop-devel-list/2012-July/msg00100.html

cgwalters:
Okay, yeah...need to figure out how to suppress that one.  Working on it.

jlebon:
Sure. Feel free to push on the same PR.

cgwalters:
https://github.com/ostreedev/ostree/pull/551

cgwalters:
@rh-atomic-bot r+ e25835d

Issue #550, opened 2016-10-27
'ostree static-delta show' ERROR code should not be reached


miabbott:
I created a mirror of the RHELAH 7.3 RC repo and used `wget` to retrieve the deltas.  When doing an `ostree static-delta show` on the single delta available, it returned the following:

```
OSTree:ERROR:src/libostree/ostree-repo-static-delta-core.c:915:_ostree_repo_static_delta_dump: code should not be reached
Aborted
```

This was done on the most recent CAHC compose. 

```
# rpm -q ostree
ostree-2016.12.1-d0e74cf3af73d6855eebe9c52f19105f7c84df0a.acde5df41a390384d9bdb8b1f939d842c24d0fcf.el7.centos.x86_64
# ostree --version
ostree 2016.12
  +libsoup +gpgme +libarchive +selinux
```

Full reproducer:

```
# mkdir -p /srv/rhelah-7.3-rc
# ostree init --repo=/srv/rhelah-7.3-rc --mode=archive-z2
# ostree remote add --repo=/srv/rhelah-7.3-rc --no-gpg-verify rcm <RHT internal URL>
# ostree pull --repo=/srv/rhelah-7.3-rc --mirror rcm rhel-atomic-host/7/x86_64/standard
# cd /srv/rhel-7.3-rc
# wget -R \*html\* --cut-dirs=5 -r --no-parent -nH <RHT internal URL>/deltas/
# ostree static-delta --repo=/srv/rhelah-7.3-rc list
347c3f5eb641e69fc602878c646cf42c4bcd5d9f36847a1f24ff8f3ec80f17b1-90c9735becfff1c55c8586ae0f2c904bc0928f042cd4d016e9e0e2edd16e5e97
# ostree static-delta --repo=/srv/rhelah-7.3-rc show 347c3f5eb641e69fc602878c646cf42c4bcd5d9f36847a1f24ff8f3ec80f17b1-90c9735becfff1c55c8586ae0f2c904bc0928f042cd4d016e9e0e2edd16e5e97
...
[...]

PR #553, opened 2016-10-27
[merged] delta: return valid enum member

jlebon:
If we can't figure out what endianness a delta is, we should just throw
ENDIAN_INVALID.

Resolves: #550


cgwalters:
Hmm yes, and it looks like the rest of the code does assume if we return `INVALID` that we should carry on with host endianness.

It'd probably be better to have added some sort of explicit
` OSTREE_DELTA_ENDIAN_UNKNOWN` that the caller could take to be `G_BYTE_ORDER` but...this is OK.

@rh-atomic-bot r+ 105b5fb

PR #555, opened 2016-10-28
deltas: Look for common alphanumeric prefixes

cgwalters:
Current versions of dracut generate reproducible initramfs files, but
we still fallback to the full file download if there is any minimal
change that causes a different checksum and file name.

This change extends that case to deal better with similar files that
have a different suffix.

Rewritten-by: Colin Walters <walters@verbum.org>

### DIFF
--- a/src/libostree/ostree-repo-static-delta-compilation-analysis.c
+++ b/src/libostree/ostree-repo-static-delta-compilation-analysis.c
@@ -187,19 +187,44 @@ build_content_sizenames_filtered (OstreeRepo              *repo,
   return ret;
 }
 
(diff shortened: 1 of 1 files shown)


giuseppe:
I've written the patch a while ago and IIRC I've done that way to prefer the same file name when in the `to_commit` set there are both a file with the same name and another with the same prefix.
e.g. trying to find a matching for "foo" -> in ("foo", "foo-2").  Wouldn't we match "foo-2" if it is before "foo" when they are sorted by size?

cgwalters:
Yeah, makes sense.

PR #556, opened 2016-10-29
[merged] Some valgrind memcheck fixes

smcv:
While trying to reproduce an intermittent test failure on Debian buildds (a segfault in `ostree pull`), I tried running the ostree tests under valgrind memcheck. I still can't reproduce the segfault, but I did find several memory leaks.

This is just a first round of fixes - I'm running with d258b124 reverted so I can fix leaks (and add new suppressions for OS libraries) incrementally.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2593,7 +2593,7 @@ load_metadata_internal (OstreeRepo       *self,
             }
           else
             {
-              GBytes *data = glnx_fd_readall_bytes (fd, cancellable, error);
+              g_autoptr(GBytes) data = glnx_fd_readall_bytes (fd, cancellable, error);
               if (!data)
                 goto out;
               ret_variant = g_variant_new_from_bytes (ostree_metadata_variant_type (objtype),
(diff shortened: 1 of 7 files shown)


src/libostree/ostree-repo.c
```diff
@@ -936,8 +936,9 @@ keyfile_set_from_vardict (GKeyFile     *keyfile,
       else if (g_variant_is_of_type (child, G_VARIANT_TYPE_STRING_ARRAY))
         {
           gsize len;
-          const char *const*strv_child = g_variant_get_strv (child, &len);
+          const gchar **strv_child = g_variant_get_strv (child, &len);
```
cgwalters: How about `g_autofree` here instead?

src/libostree/ostree-repo.c
```diff
@@ -2593,7 +2594,7 @@ load_metadata_internal (OstreeRepo       *self,
             }
           else
             {
-              GBytes *data = glnx_fd_readall_bytes (fd, cancellable, error);
+              g_autoptr (GBytes) data = glnx_fd_readall_bytes (fd, cancellable, error);
```
cgwalters: No space between autoptr and first paren.

smcv:
Thanks for reviewing. New version fixes your comments (and I have some more valgrind fixes, but I'll do those as a separate PR).

cgwalters:
@rh-atomic-bot r+ 9740ea4

cgwalters:
Hum, I don't see what failed.

@rh-atomic-bot retry

PR #557, opened 2016-10-29
[merged] Produce more complete dist tarballs

smcv:
Official ostree 2016.12 dist tarballs are missing some of the tests, which seems to be because they were built with gjs unavailable.


cgwalters:
Hm, I wonder if it'd work to unconditionally add them to `EXTRA_DIST`?  Anyways this is fine too.

@rh-atomic-bot r+ 1c0c576

PR #560, opened 2016-10-30
[merged] Filter bootloader supplied kernel cmdline options

sjoerdsimons:
Various bootloader add kernel commandline options dynamically, filter
these out when grabbing boot options from /proc/cmdline. Specifically
grub adds BOOT_IMAGE and systemd-boot adds initrd.

### DIFF
--- a/src/libostree/ostree-kernel-args.c
+++ b/src/libostree/ostree-kernel-args.c
@@ -53,6 +53,23 @@ split_keyeq (char *arg)
     }
 }
 
+static gboolean
+_arg_has_prefix (const char *arg,
+                 char      **prefixes)
+{
+  char **strviter;
+
+  for (strviter = prefixes; strviter && *strviter; strviter++)
+    {
+      const char *prefix = *strviter;
+
+      if (g_str_has_prefix (arg, prefix))
+        return TRUE;
+    }
+
+  return FALSE;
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-kernel-args.c
```diff
@@ -166,6 +193,9 @@ _ostree_kernel_args_append_proc_cmdline (OstreeKernelArgs *kargs,
   g_autofree char *proc_cmdline = NULL;
   gsize proc_cmdline_len = 0;
   g_auto(GStrv) proc_cmdline_args = NULL;
+  char *filtered_prefixes[] = { "BOOT_IMAGE=",
```
cgwalters: Can you add a comment here for each one, just like

```
{ "BOOT_IMAGE=", /* GRUB2 */
  "initrd=", /* sd-boot */
```

And also a comment at the top like:

```
/* If you update the list, also change the 
   lists in the tests; e.g.  tests/test-admin-deploy-karg.sh 
*/
```

cgwalters:
bot, add author to whitelist

cgwalters:
Thanks, don't forget to comment here if pushing a new commit, since github doesn't notify about that by default.

@rh-atomic-bot r+ 89587d1

Issue #543, opened 2016-10-24
commitpartial data shared between multiple pulls can break


alexlarsson:
There seem to be an issue with how pull handles partial commits.
If we're pulling some metadata, and its already locally available (is_stored in scan_one_metadata_object_c) then we don't scan it, assuming we have all the dependencies.
However, this is not right if the object stored locally is from a partial commit.

The code _does_ somewhat handle this case, by always scanning if pull_data->commitpartial_exists is set. However, that is only set if the commit we're pulling has a local commitpartial marker. However, that is not the case here, we're only just pulling this commit, instead the local metadata object is from _another_ partial commit.

cgwalters:
I think it'd fix this to do full scans for dirtree/dirmeta always, and just leave the "assume is complete" optimization for commits.

cgwalters:
Potential fix in https://github.com/ostreedev/ostree/pull/564

PR #564, opened 2016-11-01
[merged] pull: Redo logic for "scanning"

cgwalters:
What in the code is called "scanning" is ensuring (potentially
recursively) have an object, and if not, fetching it.  And then if
it's metadata, parsing it and finding new objects to fetch.

This logic has grown fairly complex.  What I'm trying to fix
right now is that if we're doing a pull-local to a remote repository
via `sshfs` (FUSE) we still end up scanning, which is inefficient.

We can take advantage of the "commitpartial" logic here - if a commit
isn't partial, it's complete, hence we don't need to scan it.

At the same time, I'm changing the logic here to *always* do scans for
dirtree objects.  This will fix cases where multiple commits share
dirtree objects.  We have "commitpartial" metadata, but no such concept
of partial/complete for dirtrees.

But, we'll only ever scan dirtrees if we scan commits, which is
what the section above fixes.

Closes: https://github.com/ostreedev/ostree/issues/543

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1033,11 +1033,17 @@ scan_commit_object (OtPullData         *pull_data,
                     GError            **error)
 {
   gboolean ret = FALSE;
+  /* If we found a legacy transaction flag, assume we have to scan.
+   * We always do a scan of dirtree objects; see
+   * https://github.com/ostreedev/ostree/issues/543
+   */
+  OstreeRepoCommitState commitstate;
   g_autoptr(GVariant) commit = NULL;
   g_autoptr(GVariant) parent_csum = NULL;
   const guchar *parent_csum_bytes = NULL;
   gpointer depthp;
   gint depth;
+  gboolean is_partial;
 
   if (recursion_depth > OSTREE_MAX_RECURSION)
     {
@@ -1085,6 +1091,13 @@ scan_commit_object (OtPullData         *pull_data,
           goto out;
         }
     }
+
+  if (!ostree_repo_load_commit (pull_data->repo, checksum, &commit, &commitstate, error))
+    goto out;
+
+  /* If we found a legacy transaction flag, assume all commits are partial */
+  is_partial = pull_data->legacy_transaction_resuming
+    || (commitstate & OSTREE_REPO_COMMIT_STATE_PARTIAL) > 0;
 
   if (!ostree_repo_load_variant (pull_data->repo, OSTREE_OBJECT_TYPE_COMMIT, checksum,
                                  &commit, error))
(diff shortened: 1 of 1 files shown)


cgwalters:
This passes unit tests for me, but needs some careful review.

src/libostree/ostree-repo-pull.c
```diff
@@ -1249,68 +1248,41 @@ scan_one_metadata_object_c (OtPullData         *pull_data,
...
-            {
-              do_scan = TRUE;
-              pull_data->commitpartial_exists = TRUE;
-            }
-          else if (pull_data->maxdepth != 0)
-            {
-              /* Not fully accurate, but the cost here of scanning all
-               * input commit objects if we're doing a depth fetch is
-               * pretty low.  We'll do more accurate handling of depth
-               * when parsing the actual commit.
```
alexlarsson: You're losing this comment, which seems to still apply.

alexlarsson:
I had a quick look, and it seems ok, but it needs more review. I'll try to look at it more tomorrow.

src/libostree/ostree-repo-pull.c
cgwalters: I decided to do one better than that and actually fix the FIXME.  I think the code ends up cleaner.

alexlarsson:
This seems like a good approach to me, but there seems to be some issues with pull (not necessarily related to this patch):

max_depth handling doesn't seem correct in the case of multiple pulled refs ending up with a shared commit.

For instance, consider this history:
```
a -> b -> c -> d
           \ -> e 
```

If you pull d and e with max_depth = 2, then you expect d to include b & c parents, and e to include b & a parents. However, what can happen is that as we scan commits (in order) d, c, b there are no values set in the commit_to_depth, so it gets initialized to the current depth (which at this point is from d). So, b will get depth "0".  When we later traverse from e and reach b it will already have a depth (0) set, so we will use that, which causes us to never look at a.

This is tricky to fix, because when you reach a commit object you don't know if its reachable via some other not-yet-scanned path, so we'll mark the commit as scanned, which implies that all required parents are scanned. Then when we later reach this with a different depth we'll not even rescan the commit due to it being in "scanned_metadata". I think we need to store the scanned depth in the scanned_metadata to fix this.

There seem to be a ordering issue with commitpartials in scan_one_metadata_object_c() when the remote is local. In this case we do ostree_repo_import_object_from_with_trust(), which (if trusted) writes directly to the repo, and *after* that we create the commitpartial files. If we're unlucky here and terminate before the commitpartial is written, then we have a global commit object which we didn't scan, but no commitpartial file.

cgwalters:
Hum...I think we should split those out as separate issues, I don't believe they're new with this PR, and I'm worried about conflating too many things at once.

alexlarsson:
@rh-atomic-bot r+ c979939

Issue #561, opened 2016-10-31
Don't put remote refs in summary file


cgwalters:
Otherwise no clients will be able to pull from it, since refspec != ref.

PR #565, opened 2016-11-01
[merged] repo: Don't put remote refs in the summary file

cgwalters:
I was doing a chain of mirroring like A -> B -> C

And repo B had A as a remote.  When I added B as
a remote to C, the summary file of B had a ref
upstream:foo/bar/baz, which caused all pulls from
B to C to fail, since the summary file is only
expected to have refs, not refspecs.

Closes: https://github.com/ostreedev/ostree/issues/561


jlebon:
@rh-atomic-bot r+ b0d6d47

Issue #528, opened 2016-10-12
'ostree pull' fetches deltas when '--commit-metadata-only' used


miabbott:
When using the `--commit-metadata-only` option with `ostree pull`, I expected to only get the commit data.

However, when there are deltas available on the server that correspond to the commits being fetched, those are also retrieved.

This is likely a corner-case, but here's the reproducer:
1.  Boot CentOS AH
2.  Add CAHC remote
3.  Edit origin file of booted commit and change refspec to point to CAHC
4.  Deploy HEAD^ of CAHC
5.  Reboot
6.  `ostree pull --commit-metadata-only --depth=1 centos-atomic-continuous:centos-atomic-host/7/x86_64/devel/continuous`
7. Observe deltas being fetched

PR #566, opened 2016-11-01
[merged] pull: Don't do deltas with --commit-metadata-only

cgwalters:
We should just download the commit objects directly, as it's
obviously a lot more efficient than deltas.

I had to generate a summary file in more places in the tests,
since once created, it needs to be updated.

Closes: https://github.com/ostreedev/ostree/issues/528

### DIFF
--- a/tests/pull-test.sh
+++ b/tests/pull-test.sh
@@ -35,7 +35,7 @@ function verify_initial_contents() {
     assert_file_has_content baz/cow '^moo$'
 }
 
-echo "1..12"
+echo "1..13"
 
 # Try both syntaxes
 repo_init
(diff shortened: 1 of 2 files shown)


tests/pull-test.sh
```diff
@@ -63,12 +63,28 @@ $OSTREE --repo=ostree-srv/gnomerepo checkout main checkout-origin-main
 echo moomoo > checkout-origin-main/baz/cow
 ${CMD_PREFIX} ostree --repo=ostree-srv/gnomerepo commit -b main -s "" --tree=dir=checkout-origin-main
 ${CMD_PREFIX} ostree --repo=ostree-srv/gnomerepo static-delta generate main
+${CMD_PREFIX} ostree --repo=ostree-srv/gnomerepo summary -u
```
jlebon: Is there a reason why this one needed to be added? Seems like the new logic here shouldn't affect it.

cgwalters: I think I wanted to ensure that the summary was consistently regenerated after doing any static deltas.

jlebon:
@rh-atomic-bot r+ 99e61c4

PR #567, opened 2016-11-03
[merged] deltas: Only keep one file open at a time during compilation

cgwalters:
Otherwise it's possible for us to exhaust available file descriptors
or (on 32 bit) run up against mmap limits.

In the rollsum case, we didn't need to hold open the "from" object
at all.  And in the bsdiff case, we weren't even looking at either of
the files until we started processing.

Also, while we have the patient open, switch to using O_TMPFILE
if available.


jlebon:
Looks sane overall, though I'm not very familiar with that part of the codebase. @giuseppe could you take a look as well?

giuseppe:
it looks good for me as well.

@rh-atomic-bot r+  	a212447

PR #569, opened 2016-11-09
[merged] .redhat-ci.yml: use new build key

jlebon:
This allows us to more concisely separate building from testing, which
in turn gives us a nicer inheritance pattern in our case.

See also: https://github.com/jlebon/redhat-ci/issues/11


cgwalters:
@rh-atomic-bot r+ 51a8932

cgwalters:
@rh-atomic-bot r+ 51a8932

PR #571, opened 2016-11-10
[merged] travis: Drop debian unstable since we can't fetch packages reliably

cgwalters:
I don't know what's going on, I suspect mirror churn.  Anyways,
it seems to be consistently failing now, so let's drop it.


jlebon:
@rh-atomic-bot r+ 6cc5120

PR #572, opened 2016-11-10
[merged] .redhat-ci.yml: no longer install libubsan & clang

jlebon:
Since they're now part of the auto-built image.


cgwalters:
@rh-atomic-bot r+ e4b4700

PR #574, opened 2016-11-16
[merged] pull: Add support for `http-headers` option

cgwalters:
Some deployments may want to gate access to content based on things
like OAuth.  In this model, the client system would normally compute a
token and pass it to the server via an API.

We could theoretically support this in the remote config too, but
that'd be a bit weird for OAuth as the information is dynamic.
Therefore this cleans up the code a little bit to more clearly handle
the case that the fetcher is initialized from both remote config
data plus pull options.


giuseppe:
@rh-atomic-bot r+ 566169f

PR #577, opened 2016-11-16
[merged] commit: Fix reading xattrs from OstreeRepoFile:s

alexlarsson:
When doing commit --tree=ref=XXX while at the same time applying some
form of modifier, ostree dies trying to read the xattrs using the
raw syscalls. We fix this by falling back to ostree_repo_file_get_xattrs()
in this case.

Also adds a testcase for this.


cgwalters:
@rh-atomic-bot r+ 8b9d14e

PR #578, opened 2016-11-16
commit: Support merging docker-style layers

alexlarsson:
This adds support to OstreeMutableTree for merging one mtree into another, using the mechanism for whiteouts that docker/OCI uses (i.e. .wh. whiteout files).

Since the spec doesn't guarantee the order of the whiteouts (i.e. a file from this layer can appear before a whiteout that affects it) we can't really apply layer in a streaming fashion. Thus we read the layer into an mtree end then we apply it.
    
This adds two operations to commit:
     --tree=layer-tar=TARFILE
 and
     --tree=layer-ref=COMMIT
 which are different ways to apply layers when commiting.

This PR contains two extra commits (from https://github.com/ostreedev/ostree/pull/401 and https://github.com/ostreedev/ostree/pull/577) that make the testsuite work. They should probably be merged first and then dropped from this PR before merging it.


alexlarsson:
Unclear why the CI failed here:
```
prohibit_empty_lines_at_EOF
tests/base.tar.gz
tests/layer.tar.gz
maint.mk: empty line(s) or no newline at EOF
```
Why does it look at the .tar.gz file for newlines at EOF?

giuseppe:
@alexlarsson I think it should be enough to add *.gz to `VC_LIST_ALWAYS_EXCLUDE_REGEX` in the `cfg.mk` file.  You can quickly test it locally with `make syntax-check`

alexlarsson:
@giuseppe Thanks!

giuseppe:
Once this gets merged, I'll probably have to rework https://github.com/ostreedev/ostree/pull/491 to use the new APIs

alexlarsson:
Seems to fail the test-commit.sh test in the installed variant. I think this really is an issue with the commit that i stole from https://github.com/ostreedev/ostree/pull/401

src/libostree/ostree-mutable-tree.c
```diff
@@ -304,6 +349,150 @@ ostree_mutable_tree_ensure_parent_dirs (OstreeMutableTree  *self,
...
+      else
+        {
+          g_autoptr(GError) local_error = NULL;
+          const char *whiteout_file_name =
+            file_name + strlen (OSTREE_WHITEOUT_PREFIX);
+
+          if (!ostree_mutable_tree_remove_child (self, whiteout_file_name, &local_error) &&
+              !g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+            {
+              g_propagate_error (error, g_steal_pointer (&local_error));
```
giuseppe: What do you think if we don't error out here?  I have tried with a manually crafted Docker image that has a whiteout for a file that doesn't exist in a lower layer.  The image could be imported and run without errors.  It'll probably turn to be useful for testing layers that are not built as part of the same image.

alexlarsson: What do you mean? It especially ignores the "not found" error and only reports if you have some other error. So, removing something from the base layers thats not there should work. Did it not?

alexlarsson:
Pushed a new version, based on updated patch for the --tree=ref issue from https://github.com/ostreedev/ostree/pull/581 
Hopefully this one passed the tests.

src/libostree/ostree-mutable-tree.c
giuseppe: ok sorry, I misread the condition, it is doing exactly what I meant :)

alexlarsson:
New version based on git master without any extra commits. Please review.

src/libostree/ostree-mutable-tree.c
```diff
@@ -151,6 +152,50 @@ set_error_noent (GError **error, const char *path)
...
+{
+  return TRUE;
+}
+
+gboolean
+ostree_mutable_tree_remove_all_children (OstreeMutableTree *self,
+                                         GError           **error)
+{
+  g_hash_table_foreach_remove (self->subdirs, remove_all, NULL);
+  g_hash_table_foreach_remove (self->files, remove_all, NULL);
```
alexlarsson: Yeah, forgot that.

alexlarsson: (fixed up)

jlebon:
Looks good to me. A bit bikeshed, though I wonder if the new APIs and CLI switches should use `oci_layer`/`oci-layer` since "layer" is a bit generic.

cgwalters:
I'd vote for `oci-layer` too.

Hum...so can you briefly describe the big picture here?  Would tools store both "layer refs" like the atomic/ostree integration is doing today, as well as generate "squashed commits" using this?

And the benefit here is...making it faster to check out?

alexlarsson:
This is more of a tool that can be used multiple ways. Either you can store layers as refs or you can combine refs and/or tarballs into a resolved branch.

In the case of flatpak, we never really use layers. Things are always fully resolved. However we run into layers as we start importing apps from oci images which may have multiple layers that need to be combined.

cgwalters:
https://mail.gnome.org/archives/ostree-list/2016-November/msg00012.html

PR #581, opened 2016-11-17
[merged] ostree commit: Fix combining trees with multiple --tree=ref arguments (alternative to https://github.com/ostreedev/ostree/pull/401)

alexlarsson:
This is a resubmit of the patches in https://github.com/ostreedev/ostree/pull/401 but with a rewritten test.

I dropped the old tests but kept author, commitlog  etc for the actual fix, then I added a new test in a separate commit, which creates the required data in the script and re-uses the basic test script for this.

I need this fixed for https://github.com/ostreedev/ostree/pull/578 which is why i'm resubmitting this.

Description from initial PR:

You'd expect
```
ostree commit --tree=ref=A --tree=ref=B
```
to produce a commit with the union of the trees given. Instead you'd get
a commit with the contents of just the latter commit. This was due to an
optimisation where we'd skip filling out the files and subdirs
members of the mtree, just filling in the metadata instead. This backfires
becuase this same code relies on checking the files and subdirs members
itself to work out whether the mtree is empty.

This commit removes the optimisation, fixing the bug. Maybe there's a way
to keep the optimisation and still fix the bug but it's not obvious to
me.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -2550,16 +2550,6 @@ write_directory_to_mtree_internal (OstreeRepo                  *self,
 
       ostree_mutable_tree_set_metadata_checksum (mtree, ostree_repo_file_tree_get_metadata_checksum (repo_dir));
 
-      /* If the mtree was empty beforehand, the checksums on the mtree can simply
-       * become the checksums on the tree in the repo. Super simple. */
-      if (g_hash_table_size (ostree_mutable_tree_get_files (mtree)) == 0 &&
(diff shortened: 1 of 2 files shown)


cgwalters:
Would you mind squashing the test in to preserve bisectability?

alexlarsson:
@cgwalters squashed

alexlarsson:
@rh-atomic-bot retest this please

alexlarsson:
bot, retest this please

cgwalters:
@rh-atomic-bot r+  	e24b0b3

PR #582, opened 2016-11-17
[merged] -fsanitize=thread

cgwalters:
Since we saw that random crash I thought I'd try `-fsanitize=thread`....whoo boy.  Some work to do, this is just the start.


jlebon:
@rh-atomic-bot r+ 6cb7a7f

PR #584, opened 2016-11-17
[merged] [UBSAN] deltas: Don't call memset(NULL, NULL, 0) with no xattrs

cgwalters:
This is actually fine in practice, but it triggers this
`-fsanitize=undefined` warning I saw in the test suite log:

```
src/libostree/ostree-repo-static-delta-compilation.c:160:10: runtime error: null pointer passed as argument 1, which is declared to never be null
```


jlebon:
@rh-atomic-bot r+ eef8a67

PR #588, opened 2016-11-18
[merged] tests: Support TEST_SKIP_CLEANUP=err

cgwalters:
I find myself often wanting to debug interactively failing tests.
This makes it more convenient to keep around the temporary directories
just for those tests, rather than accumulating tons of tempdirs from
the successful tests as well.


jlebon:
@rh-atomic-bot r+ b7d826b

PR #590, opened 2016-11-21
ci: Add an ASAN context

cgwalters:
Let's start doing this by default.  At some point, I'd like
to fold all of the sanitizers into one run or so.

### DIFF
--- a/.redhat-ci.Dockerfile
+++ b/.redhat-ci.Dockerfile
@@ -10,7 +10,7 @@ RUN dnf install -y \
         gjs \
         parallel \
         clang \
-        libubsan \
+        lib{ub,a,t}san \
         gnome-desktop-testing \
         redhat-rpm-config \
         elfutils \
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -39,3 +39,24 @@ env:
 
 tests:
 artifacts:
+
+---
+
+inherit: true
+context: ASAN
+# Until the container is regenerated
+packages:
+  - libasan
+
+env:
+    # Suppress leak detection in configure (and build, since
+    # g-ir-scanner uses glib), because sadly some of them leak
+    ASAN_OPTIONS: 'detect_leaks=false'
+    CFLAGS: '-fsanitize=address -fno-omit-frame-pointer'
+
+tests:
+    - env OT_SKIP_READDIR_RAND=1 ASAN_OPTIONS='detect_leaks=true:malloc_context_size=20' make check
+
+artifacts:
+    - config.log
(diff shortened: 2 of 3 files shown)


.redhat-ci.Dockerfile
```diff
@@ -10,7 +10,7 @@ RUN dnf install -y \
         gjs \
         parallel \
         clang \
-        libubsan \
+        lib{ub,a,t}san \
```
jlebon: This isn't going to take effect until it's part of the tree, so you'll also have to explicitly list the new packages in the YAML until then. I suppose that's probably why the `cc` sanity check is failing.

cgwalters:
:question: :bomb: `checking for GLIB - version >= 2.0.0... no`

cgwalters:
bot, retest this please

jlebon:
Yeah, clearly, we need to provide better feedback when validation fails. In this case, the issue is that the ASAN suite doesn't have a `build` section, so the Clang suite inherits neither a `build` nor a `tests` section, which is invalid.

cgwalters:
bot, retest this please

.redhat-ci.yml
```diff
@@ -52,9 +52,9 @@ packages:
 
 tests:
     - env NOCONFIGURE=1 ./autogen.sh
-    # Suppress leak detection in configure, because sadly some of them leak
+    # Suppress leak detection in configure (and build, since g-ir-scanner uses glib), because sadly some of them leak
     - env CFLAGS='-fsanitize=address' ASAN_OPTIONS=detect_leaks=false ./configure --prefix=/usr --libdir=/usr/lib64 --disable-gtk-doc
-    - make -j $(getconf _NPROCESSORS_ONLN)
+    - ASAN_OPTIONS=detect_leaks=false make -j $(getconf _NPROCESSORS_ONLN)
```
jlebon: So then can't we go back to the previous format and just override `env`?

jlebon:
Is this ready to merge, or did you mean to also use this PR to fix the leaks? Judging by the logs, that might take a while. :) (Should we open an issue to track this work, maybe?)

cgwalters:
I don't understand what's leaking.  I'm not reproducing it locally - I was originally using a f25 container with some `sudo make install` in it, then I tried a f24 container, now I'm trying to use the exact container being tested here.

cgwalters:
Okay, still getting leaks in CI env that I'm not seeing locally.  Will investigate in background.

cgwalters:
Obsoleted by https://github.com/ostreedev/ostree/pull/622

PR #593, opened 2016-11-23
[merged] Release 2016.14

cgwalters:


jlebon:
@rh-atomic-bot r+ 34ea65e

PR #595, opened 2016-11-28
[merged] pull: scan_commit_object() - don't load variant twice

alexlarsson:
ostree_repo_load_commit already loaded the object, no need
to load it twice.


alexlarsson:
Note: This fixes a leak, since the autoptr is written twice.

alexlarsson:
Also pushed another mem-leak fix

cgwalters:
@rh-atomic-bot r+ 94d8e34

PR #597, opened 2016-11-28
[merged] delta compilation: Fix leak

alexlarsson:
We need to ref-sik the new varian for g_autoptr to work


cgwalters:
@rh-atomic-bot r+ e6200c6

cgwalters:
Hum, actually I'm confused - how did this help?  `g_variant_unref (g_variant_new_string ("blah"))` doesn't leak, and that's all that should be going on here, right?

alexlarsson:
You're right (although it doesn't hurt to ref_sink it to make things play nice with automatic refcounting). However, I did see this in a valgrind log somehow, unless I misread it...

Issue #605, opened 2016-12-01
pull tests sometimes stall/hang


smcv:
* ostree v2016.14-27-g62b9463
* glib 2.51.0-38-g463a863
* libsoup 2.56.0-20-g1bb0157 (with or without patches from <https://bugzilla.gnome.org/show_bug.cgi?id=768567>)

With or without the patches from https://bugzilla.gnome.org/show_bug.cgi?id=768567 I intermittently see ostree processes hanging:

```
(gdb) set pagination off
(gdb) thread apply all bt

Thread 3 (Thread 0x7f6d95f44700 (LWP 32094)):
#0  0x00007f6d9d3d756d in poll () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007f6d9dbc8479 in g_poll (fds=fds@entry=0x55b01bd4ac40, nfds=nfds@entry=1, timeout=timeout@entry=-1) at /home/smcv/src/glib/glib/gpoll.c:124
#2  0x00007f6d9dbb6159 in g_main_context_poll (context=context@entry=0x55b01bd4a800, timeout=-1, priority=<optimized out>, fds=fds@entry=0x55b01bd4ac40, n_fds=n_fds@entry=1) at /home/smcv/src/glib/glib/gmain.c:4228
#3  0x00007f6d9dbb8976 in g_main_context_iterate (context=context@entry=0x55b01bd4a800, block=block@entry=1, dispatch=dispatch@entry=1, self=<optimized out>) at /home/smcv/src/glib/glib/gmain.c:3924
#4  0x00007f6d9dbb8a37 in g_main_context_iteration (context=0x55b01bd4a800, may_block=may_block@entry=1) at /home/smcv/src/glib/glib/gmain.c:3990
#5  0x00007f6d9dbb8a69 in glib_worker_main (data=<optimized out>) at /home/smcv/src/glib/glib/gmain.c:5783
#6  0x00007f6d9dbdf584 in g_thread_proxy (data=0x55b01bd41850) at /home/smcv/src/glib/glib/gthread.c:784
#7  0x00007f6d9d69d464 in start_thread (arg=0x7f6d95f44700) at pthread_create.c:333
#8  0x00007f6d9d3e09df in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:105

Thread 2 (Thread 0x7f6d96745700 (LWP 32096)):
[...]

Issue #601, opened 2016-11-29
large pull sometimes segfaults in g_mutex_lock(<some invalid pointer>)


smcv:
This is on Debian stable with backports: currently GLib 2.42, ostree 2016.13, flatpak 0.6.13, libsoup 2.48.

A large pull (downloading new Flatpak runtimes) sometimes segfaults. I was able to get this backtrace:

```
(gdb) thread apply all bt

Thread 11 (Thread 0x7fffe9ffb700 (LWP 11080)):
#0  syscall () at ../sysdeps/unix/sysv/linux/x86_64/syscall.S:38
#1  0x00007ffff6d9d8fc in g_mutex_unlock_slowpath (mutex=<optimized out>, 
    prev=<optimized out>)
    at /build/glib2.0-y6934K/glib2.0-2.42.1/./glib/gthread-posix.c:1330
#2  0x00007ffff6d9e30e in g_mutex_unlock (mutex=<optimized out>)
    at /build/glib2.0-y6934K/glib2.0-2.42.1/./glib/gthread-posix.c:1350
#3  0x00007ffff6d578a5 in g_source_attach (source=source@entry=0x7fffc8001950, 
    context=0x5555557e5390)
    at /build/glib2.0-y6934K/glib2.0-2.42.1/./glib/gmain.c:1146
#4  0x00007ffff72e47d0 in run_in_thread (job=<optimized out>, 
    c=<optimized out>, _data=0x55555580a4a0)
    at /build/glib2.0-y6934K/glib2.0-2.42.1/./gio/gsimpleasyncresult.c:867
#5  0x00007ffff72d15a6 in io_job_thread (task=<optimized out>, 
    source_object=<optimized out>, task_data=0x7fffec185880, 
    cancellable=<optimized out>)
    at /build/glib2.0-y6934K/glib2.0-2.42.1/./gio/gioscheduler.c:85
#6  0x00007ffff72f4c65 in g_task_thread_pool_thread (
    thread_data=0x7fffec0e2350, pool_data=<optimized out>)
    at /build/glib2.0-y6934K/glib2.0-2.42.1/./gio/gtask.c:1215
#7  0x00007ffff6d811d8 in g_thread_pool_thread_proxy (data=<optimized out>)
    at /build/glib2.0-y6934K/glib2.0-2.42.1/./glib/gthreadpool.c:307
---Type <return> to continue, or q <return> to quit---
[...]

Issue #583, opened 2016-11-17
pull: Racy assert: ostree_checksum_inplace_to_bytes: assertion failed: (checksum[j])


cgwalters:
I had to run quite a while to catch this:

```
+ . /home/walters/src/github/ostreedev/ostree.pull-race/tests/pull-test.sh
++ set -euo pipefail
++ echo 1..13
++ repo_init
++ cd /var/tmp/tap-test.lj2rm9
e9d4050dc755c023526c6b49a05ef2e4c127b4e1dc800cd6567d44ba159719d3
++ rm repo -rf
++ mkdir repo
++ env LD_PRELOAD=/home/walters/src/github/ostreedev/ostree.pull-race/tests/libreaddir-rand.so ostree --repo=repo init
+++ cat httpd-address
++ env LD_PRELOAD=/home/walters/src/github/ostreedev/ostree.pull-race/tests/libreaddir-rand.so ostree --repo=repo remote add --set=gpg-verify=false origin http://127.0.0.1:33749/ostree/gnomerepo
++ env LD_PRELOAD=/home/walters/src/github/ostreedev/ostree.pull-race/tests/libreaddir-rand.so ostree --repo=repo pull origin main
**
OSTree:ERROR:src/libostree/ostree-core.c:1234:ostree_checksum_inplace_to_bytes: assertion failed: (checksum[j])
```

cgwalters:
Running another loop overnight got me back in an infinite loop on a hash table, just like https://github.com/ostreedev/ostree/pull/383#issuecomment-231160278

Retrying with an ASAN build.

smcv:
```
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:58
#1  0x00007fbdf4dc640a in __GI_abort () at abort.c:89
#2  0x00007fbdf56787f6 in g_assertion_message (domain=domain@entry=0x7fbdf6277799 "OSTree", 
    file=file@entry=0x7fbdf62773d0 "/home/smcv/src/ostree/src/libostree/ostree-core.c", 
    line=line@entry=1234, 
    func=func@entry=0x7fbdf6277b40 <__func__.30225> "ostree_checksum_inplace_to_bytes", 
    message=message@entry=0x55d9acdd1c90 "assertion failed: (checksum[j])")
    at /home/smcv/src/glib/glib/gtestutils.c:2433
#3  0x00007fbdf5678846 in g_assertion_message_expr (
[...]

PR #606, opened 2016-12-01
[merged] tests: prepend to an existing LD_LIBRARY_PATH, GI_TYPELIB_PATH

smcv:
If we're using LD_LIBRARY_PATH for some locally-built library, we want
to search those after OSTree's own libraries.

---

My reproducer for #601, #605, #583 relies on this to pick up my locally-built (minimally optimized) GLib.


cgwalters:
@rh-atomic-bot r+ 0dd4ea3

PR #607, opened 2016-12-01
[merged] Terminate individual tests after 10 minutes

smcv:
While using the Automake parallel test harness, if a test hangs for
long enough for an external watchdog to kill the entire build process
(as happens in Debian sbuild after 150 minutes with no activity on
stdout/stderr), the logs will not be shown. If we make an individual
test time out sooner, logs are more likely to be shown.

We use SIGABRT so that the process(es) under test will dump core,
allowing the point at which ostree is blocking to be analyzed.
After 1 minute, if any have not died, we kill them again with SIGKILL.

---

This could be useful for reproducing #583 in an automated way, or doing automated testing that will continue to collect core dumps despite #583.

It is an improved version of the patch I've been using for a while in Debian, which waits 30 minutes and uses SIGTERM.

### DIFF
--- a/buildutil/tap-test
+++ b/buildutil/tap-test
@@ -19,7 +19,11 @@ function skip_cleanup() {
     echo "Skipping cleanup of ${tempdir}"
 }
 cd ${tempdir}
-${srcd}/${bn} -k --tap
+timeout \
+    --kill-after=60 \
+    --signal=ABRT \
+    $(( 600 * ${TEST_TIMEOUT_FACTOR:-1} )) \
+    ${srcd}/${bn} -k --tap
 rc=$?
 case "${TEST_SKIP_CLEANUP:-}" in
     no|"") cleanup ;;


cgwalters:
Eh.  I think this is really something that should be done in automake's test runner.

Also, it should be easily configurable by the external framework - if I want to run parallel tests on a slow device, it seems not implausible to hit 10 minutes.  So can you make this take input from an environment variable?

(If it was in Automake, the environment variable would be standardized)

Anyways in the short term, I'm OK to r+'ing this with the configurable bit.

smcv:
> Eh. I think this is really something that should be done in automake's test runner.

While that would be great, I think the perfect is the enemy of the good here. (Also, Automake upstream don't know how long is a reasonable upper bound for how long our tests take, but we do.)

> Also, it should be easily configurable by the external framework - if I want to run parallel tests on a slow device, it seems not implausible to hit 10 minutes.

Debian `autopkgtest` has a command-line option to multiply all arbitrary timeouts by a factor, which I think is a good approach to this. `$(( 600 * ${TEST_TIMEOUT_MULTIPLIER:-1.0} ))` or something?

smcv:
Or in fact `TEST_TIMEOUT_FACTOR` which mirrors autopkgtest's `--timeout-factor` more closely.

smcv:
Branch updated to multiply by TEST_TIMEOUT_FACTOR.

cgwalters:
@rh-atomic-bot r+ a9d22b8

PR #609, opened 2016-12-02
[merged] Libarchive autocleanup + ASAN tests

cgwalters:


cgwalters:
Turns out it's only the Debian travis coverage right now that's testing the "libglnx autocleanup since glib is too old" case.

jlebon:
Looks good!
@rh-atomic-bot r+ 7680a0e

PR #610, opened 2016-12-05
[merged] tests: Use G_DEBUG=fatal-warnings here too

cgwalters:
I am trying to track down a warning I see in `test-keyfile-utils`
which turned out to be the installed case only, but let's inject
this here too.

(The GLib default is broken, but it's hard to fix upstream without
 breaking the world)


jlebon:
@rh-atomic-bot r+ 1b7af4e

PR #611, opened 2016-12-05
[merged] tests/keyfile-utils: Drop tests covering preconditions

cgwalters:
The spam in stderr was bothering me, and further at some eventual
point in the future we want to annotate the functions with
`__attribute__((nonnull))` which would then cause tests like these to
become undefined behavior.

The coverage of this isn't worth the log spam basically.


jlebon:
@rh-atomic-bot r+ a7e8f5d

Issue #579, opened 2016-11-16
pull-local + commitpartial


cgwalters:
There seem to be a ordering issue with commitpartials in scan_one_metadata_object_c() when the remote is local. In this case we do ostree_repo_import_object_from_with_trust(), which (if trusted) writes directly to the repo, and after that we create the commitpartial files. If we're unlucky here and terminate before the commitpartial is written, then we have a global commit object which we didn't scan, but no commitpartial file.

cgwalters:
Split from https://github.com/ostreedev/ostree/pull/564

PR #613, opened 2016-12-06
[merged] pull: Write .commitpartial for local pulls first too

cgwalters:
This should help ensure we correctly handle being interrupted during
partial pulls.

Closes: https://github.com/ostreedev/ostree/issues/579


jlebon:
@rh-atomic-bot r+ c55edd7

PR #618, opened 2016-12-07
[merged] build: Add more default errors

cgwalters:
Newer gcc has `-Wincompatible-pointer-types`, hooray!
Add a few others that we pass today.


jlebon:
@rh-atomic-bot r+ 7878604

PR #619, opened 2016-12-07
WIP: OCI support for export and commit

alexlarsson:
This series adds support to libostree to handle OCI repositories, and then adds support for the --oci argument the export and commit commands.

This allows you to export an ostree commit to an OCI image, and import single-layer OCI images into an ostree commit. Additionally, if the imported OCI image was exported from ostree the commit info like summary, body, timestamp and metadata are re-imported.

I'm putting this here not really as a request to pull it as-is, but rather as a starting point for a discussion on this.

### DIFF
--- a/src/libostree/ostree-oci-registry.h
+++ b/src/libostree/ostree-oci-registry.h
@@ -1,0 +1,112 @@
+/*
+ * Copyright © 2016 Red Hat, Inc
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	 See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <http://www.gnu.org/licenses/>.
+ *
+ * Authors:
+ *       Alexander Larsson <alexl@redhat.com>
+ */
+
+#pragma once
+
+#include "libglnx/libglnx.h"
+
+#include <glib.h>
+#include <gio/gio.h>
+#include <archive.h>
+#include "ostree-json-oci.h"
+
+#define OSTREE_TYPE_OCI_REGISTRY ostree_oci_registry_get_type ()
+#define OSTREE_OCI_REGISTRY(obj) (G_TYPE_CHECK_INSTANCE_CAST ((obj), OSTREE_TYPE_OCI_REGISTRY, OstreeOciRegistry))
+#define OSTREE_IS_OCI_REGISTRY(obj) (G_TYPE_CHECK_INSTANCE_TYPE ((obj), OSTREE_TYPE_OCI_REGISTRY))
+
+_OSTREE_PUBLIC
+GType ostree_oci_registry_get_type (void);
+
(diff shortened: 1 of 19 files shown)


cgwalters:
Need something like this initially:
```
diff --git a/.redhat-ci.yml b/.redhat-ci.yml
index 6c170be..b2a2c10 100644
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -6,6 +6,9 @@ branches:
 container:
     image: projectatomic/ostree-tester
 
+packages:
+  - pkgconfig(json-glib-1.0)
+
 env:
     CFLAGS: '-fsanitize=undefined'
 
```

cgwalters:
Hum...do we need to expose json-glib as public API?  Previously when dealing with JSON I've found it way easier to convert it to gvariant first thing, then define accessors for the variant.

alexlarsson:
@cgwalters I don't see where it is exposed anywhere?

src/libostree/ostree-oci-registry.h
```diff
@@ -0,0 +1,112 @@
...
+                                                              GCancellable         *cancellable,
+                                                              GError              **error);
+_OSTREE_PUBLIC
+OstreeOciLayerWriter *ostree_oci_registry_write_layer        (OstreeOciRegistry    *self,
+                                                              GCancellable         *cancellable,
+                                                              GError              **error);
+
+
+_OSTREE_PUBLIC
+struct archive *ostree_oci_layer_writer_get_archive (OstreeOciLayerWriter  *self);
```
cgwalters: This will need to be `#ifndef __GI_SCANNER__` or so.

alexlarsson:
(I mean in a non -private header)

cgwalters:
Ok, right, I see json-glib is only included in the `-private.h` header.

The scope of this is a lot larger than I thought - I didn't realize we were going to do downloading via HTTP.  That said...it's clearly something we'll need to make this really usable.  

Ok, so...how about making this a build-time option?  That'd make it a bit easier to land in master while still allowing people who aren't using the functionality now to disable it, as well as avoiding committing to full API stability right away. Although, I guess flatpak will be doing what's in `ot-builtin-commit.c` internally, so it'd need this API?  Hmm.

src/ostree/ot-builtin-commit.c
```diff
@@ -322,12 +323,49 @@ parse_keyvalue_strings (char             **strings,
...
+{
+  DownloadLayerData data = { -1 };
+
+  data.fd = -1;
+  data.loop = g_main_loop_new (NULL, FALSE);
+  data.error = error;
+
+  ostree_oci_registry_download_blob (registry, digest, cancellable,
+                                     download_layer_cb, &data);
+  g_main_loop_run (data.loop);
```
cgwalters: See https://github.com/ostreedev/ostree/pull/499#issue-175807819

alexlarsson:
Initially I was hoping to be able to hide much more of this API in the private parts of ostree, which would allow it to change easier and would make it easier to directly expose things such as the json structs. However, the way ostree and libostree interact make this hard. Basically you would require a very highlevel "do everything oci" call in libostree, which doesn't fit very well with how things work in e.g. ot-builtins-commit.c.

Also, the OCI spec is still not 100% frozen which makes it tricky to freeze the API which refers to it.

So, I'm reconsidering where to put this code, and it seems like a better idea to put most of it in flatpak, maybe as a git submodule if something else wants to use it. Then we can put the minimal supporting code in libostree (for example ostree_repo_write_archive_fd_to_mtree, and maybe the mtree layer merging code).

cgwalters:
In the short term, that makes sense to me.  That sounds a bit like merging this approach with https://github.com/ostreedev/ostree/pull/443 plus the layer merging?

alexlarsson:
@cgwalters Yeah, sounds like a plan.

PR #620, opened 2016-12-08
[merged] build: Always do enum scanning now

cgwalters:
Since we stopped including the libsoup headers in `ostree-fetcher.h`,
we can now unconditionally do enum scanning, and drop a build time
conditional.

Prep for libcurl porting.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -34,10 +34,7 @@ libostreeincludedir = $(includedir)/ostree-1
 libostreeinclude_HEADERS = $(libostree_public_headers)
 
 ENUM_TYPES = $(NULL)
-
-if USE_LIBSOUP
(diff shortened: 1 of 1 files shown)


jlebon:
Can you explain the context a bit here? I looked into `glib-mkenums` and I understand the idea, but why do we need to generate these for `ostree-fetcher.h` at all, which is private anyway?

cgwalters:
Because the fetcher has a `config-flags` GObject property which uses `OSTREE_TYPE_FETCHER_CONFIG_FLAGS` which is defined by `glib-mkenums`.

cgwalters:
(We could just ditch that of course, it's not like we're really using it)

jlebon:
Ahh OK, I see now. It generates the equivalent of what we manually typed out for `ostree_sysroot_upgrader_flags_get_types ()`.

@rh-atomic-bot r+ b1965a3

PR #621, opened 2016-12-08
[merged] build: Make libsoup optional again

cgwalters:
The "remote cookies" code broke this.  While I'm not sure anyone is
actually using ostree-without-http, it isn't too hard to keep the
build time conditional going.  Further, this work is preparatory for
libcurl porting.


jlebon:
@rh-atomic-bot r+ 1399bbf

PR #622, opened 2016-12-08
[merged] ci: Combine UBSAN and ASAN by default

cgwalters:
I only recently realized this was possible.  While we're still seeing
leaks in the CI environment for some reason, adding ASAN gives us
use-after-free detection etc., which is obviously still very useful
even if we're not doing leak checking.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -8,8 +8,14 @@ required: true
 container:
     image: projectatomic/ostree-tester
 
+packages:
+  - libasan
+  - libubsan
+
 env:
-    CFLAGS: '-fsanitize=undefined'
+    CFLAGS: '-fsanitize=undefined -fsanitize=address'
+    ASAN_OPTIONS: 'detect_leaks=0'  # Right now we're not fully clean, but this gets us use-after-free etc
+    # TODO when we're doing leak checks: G_SLICE: "always-malloc"
 
 build:
     config-opts: >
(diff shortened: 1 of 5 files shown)


.redhat-ci.yml
```diff
@@ -8,8 +8,15 @@ required: true
...
 
+packages:
+  - libasan
+  - libubsan
+
 env:
-    CFLAGS: '-fsanitize=undefined'
+    CFLAGS: '-fsanitize=undefined -fsanitize=address'
+    OT_SKIP_READDIR_RAND: '1'  # Because LD_PRELOAD conflicts with asan
+    ASAN_OPTIONS: 'detect_leaks=false'  # Right now we're not fully clean, but this gets us use-after-free etc
```
jlebon: seems like it's `=0` to disable it?

cgwalters:
Ah!  It's the `sudo` run eating :fork_and_knife: environment variables.

cgwalters:
Okay, it's just travis running `make dist` now.   Will fix.

cgwalters:
Ok, travis is happy now too.

.redhat-ci.yml
```diff
@@ -8,8 +8,14 @@ required: true
 container:
     image: projectatomic/ostree-tester
 
+packages:
+  - libasan
+  - libubsan
```
jlebon: Can we add `libasan` to the Dockerfile as well? `libubsan` should already be in the image, so there shouldn't be any need to have it here.

jlebon:
@rh-atomic-bot r+ 9e6659e

PR #623, opened 2016-12-08
[merged] [ASAN] sysroot: Squash a leak in lockfile acquisition

cgwalters:
I installed `parallel` in my dev container, which got me
the sysroot locking tests, which caught this leak when
built with ASAN.

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -1323,7 +1323,7 @@ ostree_sysroot_try_lock (OstreeSysroot         *self,
                          GError               **error)
 {
   gboolean ret = FALSE;
-  GError *local_error = NULL;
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot.c
```diff
@@ -1338,7 +1338,7 @@ ostree_sysroot_try_lock (OstreeSysroot         *self,
         }
       else
         {
-          g_propagate_error (error, local_error);
+          g_propagate_error (error, g_steal_pointer (&local_error));
```
cgwalters: This is a pattern I think we should propagate to every other use of `g_propagate_error()` to avoid other errors like this.

jlebon:
@rh-atomic-bot r+ c787e4b

PR #624, opened 2016-12-08
[merged] tree-wide: Switch to autoptr for GOptionContext

cgwalters:
We were leaking in a few places that I noticed in an ASAN run.  Also,
this was one of the last non-autoptr cleanup sections we have in
`out:` cleanup sections, making us a lot closer to a potential
full-tree rewrite to `return FALSE`.


jlebon:
@rh-atomic-bot r+ ca36f65

PR #625, opened 2016-12-08
[merged] build: Error if glib isn't found

cgwalters:
This is a bit extracted from my work on ASAN.


jlebon:
@rh-atomic-bot r+ 1995292

Issue #626, opened 2016-12-08
Add unconfigured-state to remotes


cgwalters:
Like https://github.com/ostreedev/ostree/commit/b3ad113f788eaece9a4764ecafd3308154f59e8a but for remote configs instead of origin files.

See https://github.com/candlepin/subscription-manager/pull/1519

PR #627, opened 2016-12-08
[merged] repo: Add unconfigured-state to remote config options

cgwalters:
This is a migration from the origin version.  It's
nicer to have it in the remote, since that's what one
needs to change.  Then tools don't need to mess with
the origin file.o

In fact in this scenario one can keep the "media source" like
`file:///install/repo` or whatever, since conceptually that's where it
came from.  We're just providing a better error.

Closes: https://github.com/ostreedev/ostree/issues/626

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -179,6 +179,11 @@ Boston, MA 02111-1307, USA.
         <term><varname>tls-ca-path</varname></term>
         <listitem><para>Path to file containing trusted anchors instead of the system CA database.</para></listitem>
       </varlistentry>
+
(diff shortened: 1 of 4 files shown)


cgwalters:
Haven't done an end-to-end for this yet, but filing for review - the unit test is obviously trivial.

jlebon:
OK, that works. An unconfigured remote makes much more sense than an unconfigured origin. Let's wait on this until we confirm what will happen wrt candlepin/subscription-manager#1519.

jlebon:
@rh-atomic-bot r+ 3727d50

PR #628, opened 2016-12-09
[merged] lib: Squash last use of GFile deltas_dir

cgwalters:
I was having this thought today about making more of the OS readonly,
and ultimately if we got to the point where all ostree operations are
through the repo and sysroot dfds, we could have rpm-ostree be the
only process holding those fds open, and have a read-only bind mount
on top.

Anyways, we're not there, likely won't be soon, but this gets us
closer to being fully fd relative.


jlebon:
Nice! @rh-atomic-bot r+ fde4164

PR #629, opened 2016-12-09
[merged] repo: Fix annotations for remote_fetch_summary functions

dbnicholson:
These are out parameters, so add the (out) annotation and switch
(nullable) to (optional) since the latter is used for the purpose of
optional out parameters.


cgwalters:
@rh-atomic-bot r+  	4250a3e

PR #630, opened 2016-12-10
ci: Enable ASAN leak checking

cgwalters:
Let's see if we can track down the leaks in the CI env.


cgwalters:
bot, retest this please

PR #634, opened 2016-12-14
[merged] static-delta: Pretend that world unreadable objects are new objects

mariospr:
This will prevent including in the delta the bits to update files that
are not world readable, so that we don't run into a permissions problem
when applying the deltas from a bare-user repository that has a bare
repository set as its parent.

This is the case for Endless when updating flatpak runtimes, as the
temporary directory created in ~/.local/share/flatpak/system-cache will
be of type bare-user with its parent set to /var/lib/flatpak which is a
bare repository in EOS, as it's shared with the one at /ostree/repo.

### DIFF
--- a/src/libostree/ostree-repo-static-delta-compilation.c
+++ b/src/libostree/ostree-repo-static-delta-compilation.c
@@ -843,6 +843,25 @@ process_one_bsdiff (OstreeRepo                       *repo,
   return ret;
 }
 
+static gboolean
+check_object_world_readable (OstreeRepo   *repo,
+                             const char   *checksum,
+                             gboolean     *out_readable,
+                             GCancellable *cancellable,
+                             GError      **error)
+{
+  g_autoptr(GFileInfo) finfo = NULL;
+  guint32 mode;
+
+  if (!ostree_repo_load_file (repo, checksum, NULL, &finfo, NULL,
+                              cancellable, error))
+    return FALSE;
+
+  mode = g_file_info_get_attribute_uint32 (finfo, "unix::mode");
+  *out_readable = (mode & S_IROTH);
+  return TRUE;
+}
+
 static gboolean 
 generate_delta_lowlatency (OstreeRepo                       *repo,
                            const char                       *from,
@@ -973,6 +992,22 @@ generate_delta_lowlatency (OstreeRepo                       *repo,
       const char *from_checksum = value;
       ContentRollsum *rollsum;
       ContentBsdiff *bsdiff;
+      gboolean from_world_readable = FALSE;
+
+      /* We only want to include in the delta objects that we are sure will
+       * be readable by the client when applying the delta, regardless its
+       * access privileges, so that we don't run into permissions problems
+       * when the client is trying to update a bare-user repository with a
+       * bare repository defined as its parent.
+       */
+      if (!check_object_world_readable (repo, from_checksum, &from_world_readable, cancellable, error))
+        goto out;
+      if (!from_world_readable)
+        {
+          g_hash_table_iter_steal (&hashiter);
+          g_hash_table_add (new_reachable_regfile_content, (char*)from_checksum);
+          continue;
+        }
 
       if (!try_content_rollsum (repo, opts, from_checksum, to_checksum,
                                 &rollsum, cancellable, error))


mariospr:
The CI failure seems unrelated to this changeset. AFAICS, other PRs were failing exactly in the same test bot, so it's probably a failure in the CI system itself?

cgwalters:
Yeah...perhaps ostree should be more prescriptive about committing non-world-readable files?  Long term, I think we want such files like `/etc/shadow` to be generated by tmpfiles.d too, or alternatively created on first use. Then ostree conceptually doesn't have to care about secrets/non-world-readable data.

I think the patch should work, though we don't need to fall back to individual http fetch, I think we can just pretend it's a new object.   Something like this?

```
diff --git a/src/libostree/ostree-repo-static-delta-compilation.c b/src/libostree/ostree-repo-static-delta-compilation.c
index 1611d19..90325a6 100644
--- a/src/libostree/ostree-repo-static-delta-compilation.c
+++ b/src/libostree/ostree-repo-static-delta-compilation.c
@@ -974,6 +974,13 @@ generate_delta_lowlatency (OstreeRepo                       *repo,
       ContentRollsum *rollsum;
       ContentBsdiff *bsdiff;
 
+      if (!object_is_world_readable (key))
+        {
+          g_hash_table_iter_steal (&hashiter);
+          g_hash_table_insert (new_reachable_regfile_content, key, key);
+          continue;
+        }
+
       if (!try_content_rollsum (repo, opts, from_checksum, to_checksum,
                                 &rollsum, cancellable, error))
         goto out;
```

mariospr:
@cgwalters Thanks for the review. I've applied your suggestion and tested it and seems to fix the issue too, so I just updated the PR now. Feel free to review again, thanks!

src/libostree/ostree-repo-static-delta-compilation.c
```diff
@@ -973,6 +992,22 @@ generate_delta_lowlatency (OstreeRepo                       *repo,
...
+       * access privileges, so that we don't run into permissions problems
+       * when the client is trying to update a bare-user repository with a
+       * bare repository defined as its parent.
+       */
+      if (!check_object_world_readable (repo, from_checksum, &from_world_readable, cancellable, error))
+        goto out;
+      if (!from_world_readable)
+        {
+          g_hash_table_iter_steal (&hashiter);
+          g_hash_table_add (new_reachable_regfile_content, g_strdup (from_checksum));
```
cgwalters: Do we need to strdup here though?  Not having to do so was the idea behing using `_steal()` instead of `_remove()`.

mariospr: Oh! You're totally right, sorry about that. Will change it now

mariospr: PR updated again

cgwalters:
Thanks!  Now, our testing openstack instance may be down, but let's give this a shot:

@rh-atomic-bot r+ c2c2664

cgwalters:
@rh-atomic-bot retry

cgwalters:
Oh interesting, clang catches this:

```
src/libostree/ostree-repo-static-delta-compilation.c:1008:60: error: passing 'const char *' to parameter of type 'gpointer' (aka 'void *') discards qualifiers [-Werror,-Wincompatible-pointer-types-discards-qualifiers]
          g_hash_table_add (new_reachable_regfile_content, from_checksum);
                                                           ^~~~~~~~~~~~~
/usr/include/glib-2.0/glib/ghash.h:76:61: note: passing argument to parameter 'key' here
                                            gpointer        key);
```

In this case we need to express that we stole ownership by casting I think.

cgwalters:
i.e. `g_hash_table_add (new_reachable_regfile_content, (char*)from_checksum);`

mariospr:
Incredible, I'd swear there were no warnings con compiling. Will submit again

cgwalters:
Note it's clang in this case, not gcc.

mariospr:
Yes, sometimes I use clang but in this case I was compiling with gcc.

mariospr:
PR updated, btw. Let's see if we can get this simple patch in before the weekend :)

cgwalters:
@rh-atomic-bot r+ a3ff9fd

cgwalters:
@mariospr Can you look at the first commit in https://github.com/ostreedev/ostree/pull/678 ?

mariospr:
@cgwalters It looks good to me, sorry for introducing that issue (and thanks for fixing it, I'm cherry picking this one for Endless tomorrow first thing in the morning!).

PR #637, opened 2016-12-23
[merged] admin: Use execlp() to look for systemctl as the shell would

paulvt:
Fixes the problem that `ostree admin upgrade --reboot` actually doesn't reboot after a successful deployment. The error is:

    error: No such file or directory
because of:

    execve("systemctl", ["systemctl", "reboot"], [/* 11 vars */]) = -1 ENOENT (No such file or directory)


cgwalters:
Oh, embarrassing.  I'll just note I'm trying to make ostree more of a shared library, and hence the admin command lines are more of a demo.  But still, it should work :smile:  Thanks for the patch!

@rh-atomic-bot r+ 1d88baf

PR #638, opened 2016-12-23
[merged] docs: Fix ostree.version -> version

cgwalters:
I think originally I had envisioned this as `ostree.version`, but at the last
minute we changed it to just `version`. That's what all of the code uses, so
let's fix the docs.


cgwalters:
@rh-atomic-bot delegate=paulvt

paulvt:
Ah yes, then it shows up in `ostree admin status`. Thanks!

@rh-atomic-bot r+ a78c90e

Issue #640, opened 2017-01-04
question: ostree admin upgrade does it twice


johan-smits:
When we run a _ostree admin upgrade_ it deploys a new one (when there is one).
When you run it again (and you did not reboot yet to the new one) it does a redeploy of the already deployed again.
Why does it redeploy when the last one was successful?

We want to _cron_ the upgrade task on the desktops but we are limited to daily because we do not want to reboot automatically but we want to check for updates more frequently.

cgwalters:
You can do the "download but don't deploy" with a raw "ostree pull".  We don't have good porcelain for this right now though.

paulvt:
One would expect `ostree admin upgrade` downloads and decides not to deploy if it already has done so.
You do get this desired behaviour if you are already booted to the newest OS:

```
$ sudo ostree admin upgrade

1 metadata, 0 content objects fetched; 313 B transferred in 0 seconds                                                   
No update available.
```

cgwalters:
Yes - it's complex, one reason we do a redeploy is to re-do copying /etc, in case there were any changes.  See also https://github.com/projectatomic/rpm-ostree/issues/40

Big picture, at some point I'd like to merge the "pure ostree daemon" code from https://github.com/endlessm/eos-updater into ostree, and have it support things like "background pull of updates, deploy last pulled and reboot".

cgwalters:
Daemon aside, I think it'd be pretty easy to implement:

```
ostree admin upgrade --pull-only  # Run this in a cron job
ostree admin upgrade --deploy --reboot # Run this to actually do the deployment, just before rebooting
```

johan-smits:
We use it in a desktop environment and we want to prevent a reboot (users can be working on their desktop and do not appreciate a reboot in the middle of their work).

Will have a look at https://github.com/endlessm/eos-updater

cgwalters:
I took a quick crack at this in https://github.com/ostreedev/ostree/pull/642

cgwalters:
Going to close this as fixed by https://github.com/ostreedev/ostree/pull/642 - but I only lightly tested that, so feel free to reopen.  There's definitely bigger picture stuff here around having better tooling than raw `ostree admin upgrade` to fix, but I think the raw functionality is there now.

PR #642, opened 2017-01-05
upgrade: Add support for --pull-only and --deploy-only

cgwalters:
This makes it easier to script downloading updates in the background,
and only do deployments just before rebooting.

Partially addresses https://github.com/ostreedev/ostree/issues/640

### DIFF
--- a/src/libostree/ostree-sysroot-upgrader.h
+++ b/src/libostree/ostree-sysroot-upgrader.h
@@ -85,7 +85,8 @@ gboolean ostree_sysroot_upgrader_check_timestamps (OstreeRepo     *repo,
 
 typedef enum {
   OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_NONE = 0,
-  OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_ALLOW_OLDER = (1 << 0)
+  OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_ALLOW_OLDER = (1 << 0),
+  OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_SYNTHETIC = (1 << 1) /* Don't actually do a pull, just check timestamps/changed */
 } OstreeSysrootUpgraderPullFlags;
 
 _OSTREE_PUBLIC
--- a/src/ostree/ot-admin-builtin-upgrade.c
+++ b/src/ostree/ot-admin-builtin-upgrade.c
@@ -121,19 +121,7 @@ ot_admin_builtin_upgrade (int argc, char **argv, GCancellable *cancellable, GErr
         }
     }
 
-  /* In the pull-only case, we do a cleanup here to ensure that if multiple
-   * commits were pulled, we garbage collect any intermediate data before
-   * pulling more.  This is really a best practice in general, but for maximum
-   * compatiblity, we only do cleanup if a user specifies the new --pull-only
-   * option.  Otherwise, we would break the case of trying to deploy a commit
-   * that isn't directly referenced.
-   */
-  if (opt_pull_only)
-    {
-      if (!ostree_sysroot_cleanup (sysroot, cancellable, error))
-        goto out;
-    }
-  else if (opt_deploy_only)
+  if (opt_deploy_only)
     upgraderpullflags |= OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_SYNTHETIC;
 
   { g_auto(GLnxConsoleRef) console = { 0, };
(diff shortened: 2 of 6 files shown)


cgwalters:
I didn't test this locally yet, and would definitely need test suite additions.

cgwalters:
One major piece this is also missing is that `--deploy --reboot` will always do a deployment.

paulvt:
I'd say that `--reboot` conflicts with `--pull-only` (why would you do that?) and also with `--deploy-only`.
I see `--reboot` as a mechanism for rebooting conditionallly when something was pulled **and** then deployed. (Same as I wrongly considered `ostree admin upgrade` to pull and deploy if and only if something was pulled.)

cgwalters:
Okay, took a crack at fixing the gap by adding a "synthetic" pull flag that's used with `--deploy-only`.  This gets us the proper changed handling, so `ostree admin upgrade --deploy-only` will say "No changes".

(I still didn't test this much sorry, typing some code here while waiting for some tests to run)

cgwalters:
Rebased 🏄, now with docs 📚 and tests ✅.

cgwalters:
Removing WIP.

src/libostree/ostree-sysroot-upgrader.h
```diff
@@ -85,7 +85,8 @@ gboolean ostree_sysroot_upgrader_check_timestamps (OstreeRepo     *repo,
 
 typedef enum {
   OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_NONE = 0,
-  OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_ALLOW_OLDER = (1 << 0)
+  OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_ALLOW_OLDER = (1 << 0),
+  OSTREE_SYSROOT_UPGRADER_PULL_FLAGS_SYNTHETIC = (1 << 1) /* Don't actually do a pull, just check timestamps/changed */
```
jlebon: Do we need this flag though? Seems like it would be cleaner to just seed `new_revision` with the refspec on upgrader init (similar to what we do in rpm-ostree) and then just not call `ostree_sysroot_upgrader_pull()` at all, no?

src/ostree/ot-admin-builtin-upgrade.c
```diff
@@ -104,6 +121,21 @@ ot_admin_builtin_upgrade (int argc, char **argv, GCancellable *cancellable, GErr
         }
     }
 
+  /* In the pull-only case, we do a cleanup here to ensure that if multiple
+   * commits were pulled, we garbage collect any intermediate data before
+   * pulling more.  This is really a best practice in general, but for maximum
+   * compatiblity, we only do cleanup if a user specifies the new --pull-only
+   * option.  Otherwise, we would break the case of trying to deploy a commit
+   * that isn't directly referenced.
+   */
```
jlebon: I'm still not sure I understand this. Is there any risk to leaving the intermediate data until the cleanup that happens after deployment? Isn't this making us lose potential static delta jumps between consecutive `--pull-only` operations?

cgwalters: If a pull is successful, it will set the ref.  Let's call this the "new pulled base", as distinct from creating a new deployment.  `ostree_sysroot_cleanup` won't reset the ref.

Hence we should (though this isn't explicitly tested right now) keep deltas, since we still have the ref.  Now, that's all the argument for why this won't break anything.

Why I wanted to do this is - think about a scenario where due to a flaky internet connection we do a pull, it gets interrupted, do another pull, it gets interrupted, repeat.  If this happens a lot, we will endlessly accumulate partial new commits.

With this change, we'll trigger the one-day-timeout on staged commit data.  Now...perhaps we should really do the cleanup *after* an unsuccessful pull, to ensure that we will always reuse downloaded-but-not-consumed deltas?  Will look at that.

src/libostree/ostree-sysroot-upgrader.h
cgwalters: The equivalent of `RPMOSTREE_TRANSACTION_DEPLOY_FLAG_NO_PULL_BASE`?  So...yeah, I guess we could do that, but I like keeping the timestamp check in there, and I'd argue we should probably do the same in rpm-ostree.

What if e.g. someone did a raw `ostree pull --commit-metadata-only` or something, which *won't* do timestamp checks and hence could be more easily MITM'd - then later an admin does `ostree admin upgrade --deploy-only`, we'd be skipping the timestamp check.

src/ostree/ot-admin-builtin-upgrade.c
cgwalters: Moved some of that to https://github.com/ostreedev/ostree/issues/713

jlebon:
Looks good!
@rh-atomic-bot r+ 7e9cd67

PR #643, opened 2017-01-12
[merged] trivial-httpd: Daemonize better

cgwalters:
I was trying to debug `test-pull-c`, and typing `Ctrl-C` in gdb
ended up sending `SIGINT` to trivial-httpd as well, killing it.

Daemonize a bit more properly to avoid this. I also followed the standard
`/dev/null` guidelines.


jlebon:
@rh-atomic-bot r+ 17e9d68

PR #644, opened 2017-01-12
[merged] .dir-locals.el: Standard Emacs indentation config

cgwalters:
Better than having it per-file.


jlebon:
@rh-atomic-bot r+

PR #648, opened 2017-01-17
[merged] tests: Alias assert_not_reached() -> fatal()

cgwalters:
We had a lot of copies of the "echo something 1>&2; exit 1" code even though
`assert_not_reached()` was it.  Hence, I think we need a shorter alias for that.

Doing this particularly since I noticed a missing `1` in an `exit 1` call in the
rpm-ostree copy of this.


jlebon:
Boy, even Travis is happy for a change!
@rh-atomic-bot r+ e643560

PR #650, opened 2017-01-18
tests: Add a big (many objects) pull

cgwalters:
This would be more likely to tickle things like
https://github.com/ostreedev/ostree/issues/601
reliably.

Also, while working on the curl backend, I hit on the fact that curl doesn't
queue (by default, you can enable) and will happily create 20000+ concurrent TCP
connections if you try. Having this test would have made that more likely to
fail.


cgwalters:
Tangential fun fact: when testing this originally I was running nginx in a Docker container with the default bridge, and this killed `docker-proxy`.  Ended up running nginx with `--net=host` to avoid that.

tests/test-pull-many.sh
```diff
@@ -0,0 +1,98 @@
...
+cd ${test_tmpdir}
+rm main -rf
+mkdir main
+cd main
+ndirs=9443
+depth=0
+echo "$(date): Generating content..."
+while [ $ndirs -gt 0 ]; do
+    # 2/3 of the time, recurse a dir, up to a max of 9, otherwise back up
+    x=$(($RANDOM % 3))
```
giuseppe: should we avoid $RANDOM to make the input repeatable?

giuseppe:
tests are still failing, it seems that the `echo "ok ...` message is not present

jlebon:
bot, retest this please

giuseppe:
@rh-atomic-bot r+  	32d7de7

cgwalters:
@giuseppe try again

cgwalters:
@rh-atomic-bot r=giuseppe

PR #652, opened 2017-01-19
[merged] Executable status fixes for tests

smcv:
There was some confusion about which test files should or shouldn't be executable. This branch normalizes it:

* If a script makes sense to execute in its own right (not being sourced), chmod +x in git, install it +x in the build system and give it a shebang
* If a script is to be sourced, chmod it -x, install it -x and do not give it a shebang

This revealed that we have one test that isn't actually run, a `bare-user` equivalent of `tests/test-basic.sh`. Fix the syntax of its TAP output (it previously produced output as though it was two TAP tests) and enable it.


cgwalters:
Great, thanks!

@rh-atomic-bot r+ 7b23972

PR #654, opened 2017-01-24
Libcurl prep v2

cgwalters:


src/libostree/ostree-fetcher.c
```diff
@@ -63,7 +63,6 @@ typedef struct {
   int max_outstanding;
 
   /* Queue for libsoup, see bgo#708591 */
```
jlebon: Outdated comment?

src/libostree/ostree-fetcher.c
```diff
@@ -77,8 +76,10 @@ typedef struct {
 
 } ThreadClosure;
 
+
 static void
-session_thread_process_pending_queue (ThreadClosure *thread_closure);
+start_pending_request (ThreadClosure *thread_closure,
+                       GTask         *task);
```
jlebon: Do we actually need this prototype anymore?

src/libostree/ostree-fetcher.c
```diff
@@ -1110,10 +1081,8 @@ on_request_sent (GObject        *object,
                 goto out;
 
               (void) g_input_stream_close (pending->request_body, NULL, NULL);
-              g_queue_insert_sorted (&pending->thread_closure->pending_queue,
-                                     g_object_ref (task), pending_task_compare,
-                                     NULL);
-              remove_pending_rerun_queue (pending);
```
jlebon: We can probably delete this function now and remove the pending object right at the call sites. (Or at least, rename it to just `remove_pending`).

cgwalters:
Ok, comments addressed ⬆️

jlebon:
@rh-atomic-bot r+ 7d05650

PR #655, opened 2017-01-26
lib: Move the bupsplit selftest into our test framework

cgwalters:
We weren't running it before. Also I switched it to use GLib. Preparation for
some oxidation work (having an implementation of bupsplit in Rust).

I exported another function to do the raw rollsum operation which is what this
test suite uses.


jlebon:
I'm not familiar with the bup internals, but the changes here make sense.
@rh-atomic-bot r+ bfc62e3

PR #656, opened 2017-01-26
Oxidation

cgwalters:
This is an initial drop of "oxidation", or adding implementation
of components in Rust.  The bupsplit code is a good target - no
dependencies, just computation.

Translation into Rust had a few twists -

 - The C code relies a lot on overflowing unsigned ints, and
   also on the C promotion rules for e.g. `uint8_t -> int32_t`
 - There were some odd loops that I introduced bugs in while
   translating...in particular, the function always returns `len`,
   but I mistakenly translated to `len+1`, resulting in an OOB
   read on the C side, which was hard to debug.

On the plus side, an off-by-one array indexing in the Rust code paniced nicely.

In practice, we'll need a lot more build infrastructure to make this work, such
as using `cargo vendor` when producing build artifacts for example. Also, Cargo
is yet another thing we need to cache.

Where do we go with this? Well, I think we should merge this, it's not a lot of
code. We can just have it be an alternative CI target. Should we do a lot more
right now? Probably not immediately, but I find the medium/long term prospects
pretty exciting!

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -60,6 +60,21 @@ GIRS =
 TYPELIBS = $(GIRS:.gir=.typelib)
 endif
 
+# These bits based on gnome:librsvg/Makefile.am
+if ENABLE_RUST
+if RUST_DEBUG
+CARGO_RELEASE_ARGS=
+else
+CARGO_RELEASE_ARGS=--release
+endif
+
+check-local:
+	cd $(srcdir)/rust && CARGO_TARGET_DIR=@abs_top_builddir@/target cargo test
+
+clean-local:
+	cd $(srcdir)/rust && CARGO_TARGET_DIR=@abs_top_builddir@/target cargo clean
+endif # end ENABLE_RUST
+
 libglnx_srcpath := $(srcdir)/libglnx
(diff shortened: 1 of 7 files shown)


cgwalters:
Depends on https://github.com/ostreedev/ostree/pull/655

Makefile.am
```diff
@@ -60,6 +60,21 @@ GIRS =
...
+# These bits based on gnome:librsvg/Makefile.am
+if ENABLE_RUST
+if RUST_DEBUG
+CARGO_RELEASE_ARGS=
+else
+CARGO_RELEASE_ARGS=--release
+endif
+
+check-local:
+	cd $(srcdir)/rust && CARGO_TARGET_DIR=@abs_top_builddir@/target cargo test
```
jlebon: Do we actually need this target? Or are we planning to use the rust testing framework as well? Might want to rename it to `rust-check` or similar.

cgwalters: I think we'd definitely use the Rust unit testing framework, it's really nice.  In this case I explicitly chose not to because the idea is we could reuse the exact same tests that were done for the C implementation.

cgwalters:
🏄

jlebon:
OK cool. I won't pretend to understand all of the Rust code itself, but the integration itself looks nice!
@rh-atomic-bot r+ 5d51c7c

PR #658, opened 2017-01-29
tests: Add setup for more realistic repo, change pull-many to use

cgwalters:
As OSTree has evolved over time, the tests grew with it.  We
didn't start out with static deltas or a summary file, and the
tests reflect this.

What I really want to do is change more of the pull tests, from
corruption/proxying/mirroring etc. to use this more realistic
repo rather than the tiny one the other test creates.

We start by using some of the code from `test-pull-many.sh`, and change that
test to use `--disable-static-deltas` for pull, since the point of that test is
to *not* test deltas.

Still TODO is investigate changing other tests to use this.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -270,6 +270,102 @@ setup_fake_remote_repo1() {
     export OSTREE="${CMD_PREFIX} ostree --repo=repo"
 }
 
+# Set up a large repository for stress testing.
+# Something like the Fedora Atomic Workstation branch which has
+# objects: meta: 7497 content: 103541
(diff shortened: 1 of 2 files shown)


cgwalters:
bot, retest this please

jlebon:
Nice!
@rh-atomic-bot r+ 69f6816

jlebon:
(Why the reduction by 1/3 BTW? Maybe we can make `ndirs` a param.)

cgwalters:
Sorry, should have mentioned.  Just felt it was a bit overkill.

PR #659, opened 2017-01-30
Rename to libOSTree

cgwalters:
There are many motivating factors. The biggest is simply that at a practical
level, the command line is not sufficient to build a real system. The docs say
that it's a demo for the library. Let's make that more obvious, so people don't
try to use `ostree admin upgrade` for their real systems, and also don't use
e.g. `ostree commit` on the command line outside of test suites/quick hacking.

This change will also help clarify the role of rpm-ostree, which we will likely
be renamed to "nts". Then use of the term "ostree" will become much clearer. And
similarly for other people writing upgraders, they can say they use libostree.

I didn't try to change all of the docs and code at once, because it's going to
lead to conflicts.

The next big steps are:

  - Rename the github repo (github will inject a redirect)
  - Look at supporting a build where we don't do `ostree admin`, or at least
    it's only built for tests. We may want to split it off as a separate binary
    or so? That way people with their own upgraders don't need to ship it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,25 @@
-OSTree
+libOSTree
 ======
 
 New! See the docs online at [Read The Docs (OSTree)](https://ostree.readthedocs.org/en/latest/ )
 
 -----
 
-OSTree is a tool that combines a "git-like" model for committing and
-downloading bootable filesystem trees, along with a layer for
-deploying them and managing the bootloader configuration.
+This project is now known as "libOSTree", renamed from "OSTree"; the focus is on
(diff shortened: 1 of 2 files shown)


jlebon:
Let's do this!
@rh-atomic-bot r+ 536d5c3

jlebon:
bot, retest this please

jlebon:
Bah, getting confused with all the bots!
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #661, opened 2017-01-30
lib: Adjust comments in symbols section for last release

cgwalters:
2017.1 was released, move its section above the line.

Issue #404, opened 2016-07-21
Add API to refresh ostree config from filesystem


jlebon:
See https://github.com/projectatomic/rpm-ostree/issues/401.

miabbott:
Related downstream bug - https://bugzilla.redhat.com/show_bug.cgi?id=1362455

jlebon:
This is fixed in https://github.com/ostreedev/ostree/pull/662.

PR #662, opened 2017-01-30
Reload config

cgwalters:


cgwalters:
Prep patches in https://github.com/ostreedev/ostree/pull/661

jlebon:
Hmm, there are other things that we'll probably need to add reload support for (e.g. [the repo config itself](https://github.com/ostreedev/ostree/issues/404)). Maybe we should call the function `ostree_repo_reload()` and add flags?

cgwalters:
Rebased 🏄  And now reloads all of the config ✅

jlebon:
Ouuhh very nice!

@rh-atomic-bot r+ 4801f64

PR #664, opened 2017-01-30
travis: Trim flaky distros

cgwalters:
I went through the Travis history a bit, and these seem to be
the flaky ones, I'm pretty sure because they don't have the libsoup
patches backported.

As I say in comments, once we have libcurl, I'd like to enable that mostly
everywhere, which should (hopefully) be more reliable.

:phone: @smcv

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,9 @@ dist: trusty
 sudo: required
 
 env:
-  - ci_distro=ubuntu ci_suite=trusty
+  - ci_distro=ubuntu ci_suite=trusty ci_test=no  # TODO: use libcurl on this
   - ci_docker=debian:jessie-slim ci_distro=debian ci_suite=jessie
-  - ci_docker=debian:stretch-slim ci_distro=debian ci_suite=stretch
+  - ci_docker=debian:stretch-slim ci_distro=debian ci_suite=stretch ci_test=no  # TODO gpgme flake https://github.com/ostreedev/ostree/pull/664#issuecomment-276033383
   - ci_docker=ubuntu:xenial ci_distro=ubuntu ci_suite=xenial
 
 script:


smcv:
Debian stretch has had the libsoup patch since 2017-01-01 (https://tracker.debian.org/pkg/libsoup2.4 - currently "testing" refers to stretch). The test failure that I'm still seeing there is that `test-local-pull` sometimes hangs, which as far as I can tell is some initialization issue with gpgme, possibly related to Debian stretch having the "modern" gnupg 2.1 branch. I've wondered whether invoking tests like

```
GNUPGHOME=whatever gpg-agent --daemon ./tests/test-whatever.sh
```

(instead of relying on automatic starting of one gpg-agent per GNUPGHOME as we do now) might help, but so far haven't got round to testing that.

Debian jessie definitely does not have the libsoup patch.

If distributions of an age similar to jessie and trusty are something you are interested in supporting, would it be possible to leave those in the list, but set either `ci_test=no` or `ci_test_fatal=no`? That way we'd still be demonstrating that ostree *compiles* (and in particular that it hasn't accidentally grown newer dependencies), even if it doesn't always pass tests.

cgwalters:
Ah, `ci_test=no` it is, thanks!  Hm, Fedora 25 has the modern GPGME too, but I don't *think* we've seen flakes on that.  Anyways let's just do this for now so I can make progress on other things.

cgwalters:
Wait, hum...the jessie one has mostly been passing.  (That was confusing to me, since jessie is older than stretch, but it makes sense if it's GPGME and not libsoup)

smcv:
r+

smcv:
@rh-atomic-bot r+

smcv:
Sorry, I must have been misremembering whether I could r+ things here

smcv:
> the jessie one has mostly been passing. (That was confusing to me, since jessie is older than stretch, but it makes sense if it's GPGME and not libsoup)

The libsoup bug was a race condition. Perhaps it doesn't actually hit very often on Travis? I didn't see it very often on my laptop without special build options.

giuseppe:
@rh-atomic-bot delegate=smcv

@smcv could you try now?

smcv:
@rh-atomic-bot r+

cgwalters:
Hm, Homu seems to have lost this.

@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot r=smcv

cgwalters:
@rh-atomic-bot r+ 4da5f30

jlebon:
@rh-atomic-bot retest

jlebon:
@rh-atomic-bot retry

PR #665, opened 2017-01-30
libostree: Don't distribute generated enumtypes in tarballs

smcv:
They are built at "make" time and cleaned up by "make clean", so there
is no need to distribute them.


cgwalters:
Makes sense.  I think I cargo culted that code from somewhere else, but I forget where now.  Maybe wherever that is is fixed now too.

@rh-atomic-bot r+ e6cd0c4

PR #668, opened 2017-02-03
trusted.gpg.d: keep in the same location

jlebon:
With the package rename from ostree to libostree, the `trusted.gpg.d/` dir
changed install location from `/usr/share/ostree` to `/usr/share/libostree`.
Let's keep the same dir to remain compatible with existing installations
that may already have keys there.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -206,7 +206,7 @@ endif
 pkgconfig_DATA += src/libostree/ostree-1.pc
 
 gpgreadme_DATA = src/libostree/README-gpg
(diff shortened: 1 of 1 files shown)


cgwalters:
Oh man.  The problem is more widespread than this actually I think; anything which references one of the Automake `$(pkgXXXXX)` variables.

cgwalters:
Although, maybe it's OK if other things move.

jlebon:
Well, the other ones shouldn't be accessed by users though right? This is the only one I found that may contain user-provided data.

cgwalters:
@rh-atomic-bot r+  ce59ae3

PR #669, opened 2017-02-04
rust: Support `make dist` -> cargo vendor

cgwalters:
What we do here basically is set things up in a `dist-hook` so that our Rust
sources are vendored at `dist` time. This gives us a single tarball still, and
ideally should be transparent to downstream builders, as long as they have the
`cargo/rust` toolchain.


cgwalters:
With this, I did a build for Fedora rawhide, which includes many architectures, and [it built](https://koji.fedoraproject.org/koji/taskinfo?taskID=17594662).   We don't have an RPM `%check` section now, but at least we know we can do builds.

jlebon:
@rh-atomic-bot r+ c56e3e4

PR #670, opened 2017-02-05
packaging/: Delete

cgwalters:
This is obsolete for a few reasons. The spec file is out of date (mostly the
%files, but also the BRs), and further down the line we'll need to use `make
dist` so we pick up vendored Rust sources. So the "git archive" approach won't
work for much longer anyways.

This came up in https://mail.gnome.org/archives/ostree-list/2017-February/msg00005.html

Basically, anyone who wants to build packages should look at the upstream
dist-git - until such time as e.g. Fedora learns to support pulling spec files
from upstream.


giuseppe:
@rh-atomic-bot r+ 105023b

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #671, opened 2017-02-07
repo: Add archive/zlib-level option, drop default compression to 6

cgwalters:
The gzip default is 6.  When I was writing this code, I chose 9 under
the assumption that for long-term archival, the extra compression was
worth it.

Turns out level 9 is really, really not worth it.  Here's run at level 9
compressing the current Fedora Atomic Host into archive:

```
ostree --repo=repo pull-local repo-build fedora-atomic/25/x86_64/docker-host
real    2m38.115s
user    2m31.210s
sys     0m3.114s
617M    repo
```

And here's the new default level of 6:

```
ostree --repo=repo pull-local repo-build fedora-atomic/25/x86_64/docker-host
real    0m53.712s
user    0m43.727s
sys     0m3.601s
619M    repo
619M    total
```

As you can see, we run almost *three times* faster, and we take up *less
than one percent* more space.

Conclusion: Using level 9 is dumb.  And here's a run at compression level 1:

```
ostree --repo=repo pull-local repo-build fedora-atomic/25/x86_64/docker-host
real    0m24.073s
user    0m17.574s
sys     0m2.636s
643M    repo
643M    total
```

I would argue actually many people would prefer even this for "devel" repos.
For production repos, you want static deltas anyways.  (However, perhaps
we should support a model where generating a delta involves re-compressing
fallback objects with a bit stronger compression level).

Anyways, let's make everyone's life better and switch the default to 6.


src/libostree/ostree-repo.c
```diff
@@ -2219,6 +2219,19 @@ ostree_repo_open (OstreeRepo    *self,
...
 
+  { g_autofree char *compression_level_str = NULL;
+
+    /* gzip defaults to 6 */
+    (void)ot_keyfile_get_value_with_default (self->config, "archive", "zlib-level", NULL,
+                                            &compression_level_str, NULL);
+
+    if (compression_level_str)
+      /* Ensure level is in [1,9] */
+      self->zlib_compression_level = MAX (1, MIN (9, g_ascii_strtoull (compression_level_str, NULL, 10)));
```
jlebon: Check for successful conversion?

cgwalters: Eh...it's really annoying to do, we don't do it in various other places either.  It'd probably be worth adding a validating version of the API to GLib though.

jlebon:
@rh-atomic-bot r+ 7e3aa01

PR #673, opened 2017-02-08
diff: Add onwer-uid/gid arguments to diff command

ortogonal:
This makes it possible to use diff on local files were -owner-uid/gid
have been passed to commit, if not diff will say that all local files
are changed.


src/libostree/ostree-diff.h
```diff
@@ -69,7 +69,9 @@ gboolean ostree_diff_dirs (OstreeDiffFlags flags,
                            GPtrArray      *removed,
                            GPtrArray      *added,
                            GCancellable   *cancellable,
-                           GError        **error);
+                           GError        **error,
```
cgwalters: Unfortunately, this is public API, so we can't break it now.  We'd need to add a new version, like `ostree_diff_dirs_with_options()` or something.  Also, the `GError` parameter should be last.

cgwalters:
bot, add author to whitelist

src/libostree/ostree-diff.h
ortogonal: Ohh! Didn't think of that. I will fix that!

src/libostree/libostree.sym
```diff
@@ -89,6 +89,7 @@ global:
         ostree_deployment_set_index;
         ostree_deployment_set_origin;
         ostree_diff_dirs;
+        ostree_diff_dirs_with_options;
```
cgwalters: This will need to be in the `LIBOSTREE_2017.2` section down below - the section here is for symbols already released.

cgwalters:
The commit series here looks messy; can you rebase on master and squash your patches together?

PR #675, opened 2017-02-09
fetcher queue: also throttle on outstanding writes

jlebon:
When fetching over a fast enough connection, we can be receiving files
faster than we write them. This can then lead to `EMFILE` when we have
enough files open. This was made very easy to notice with the upcoming
libcurl backend, which makes use of pipelining.


cgwalters:
@rh-atomic-bot r+ c36859f

PR #676, opened 2017-02-09
ostree-repo: Clarify error behaviour of remote option getters

pwithnall:
Clarify the documentation for functions like
ostree_repo_get_remote_boolean_option(), stating what out_value will be
set to on error.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -284,7 +284,7 @@ _ostree_repo_remote_name_is_file (const char *remote_name)
  * OSTree remotes are represented by keyfile groups, formatted like:
  * `[remote "remotename"]`. This function returns a value named @option_name
  * underneath that group, or @default_value if the remote exists but not the
- * option name.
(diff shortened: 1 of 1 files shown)


cgwalters:
bot, add author to whitelist

cgwalters:
Hmm.  As a general rule, we don't usually touch the return values in error paths.  We *may* do so, and I'm OK documenting the cases where we do.

Really, what I'd say the rule in particular is "On error, you won't have to free the output value".  The combined best practice here is to use cleanup macros, initialize to NULL, and then things are robust even in the case where we *do* accidentally return an allocated out value in an error case.  (But again we shouldn't do this)

@rh-atomic-bot r+ df332d5

pwithnall:
The question which I was trying to answer when I was looking at the documentation was: “If ostree_repo_get_remote_boolean_option() fails, will it set out_value to default_value, or do I need to handle that myself?”. Normally I would assume that out values are not touched on failure — but the fact that a ‘default’ value exists here confuses things for me slightly.

PR #678, opened 2017-02-10
Delta progress

cgwalters:
One bugfix, one enhancement, and work on fixing delta progress.

### DIFF
--- a/src/libostree/ostree-repo-static-delta-compilation.c
+++ b/src/libostree/ostree-repo-static-delta-compilation.c
@@ -1003,11 +1003,7 @@ generate_delta_lowlatency (OstreeRepo                       *repo,
       if (!check_object_world_readable (repo, from_checksum, &from_world_readable, cancellable, error))
         goto out;
       if (!from_world_readable)
-        {
-          g_hash_table_iter_steal (&hashiter);
-          g_hash_table_add (new_reachable_regfile_content, (char*)from_checksum);
-          continue;
-        }
+        continue;
 
       if (!try_content_rollsum (repo, opts, from_checksum, to_checksum,
                                 &rollsum, cancellable, error))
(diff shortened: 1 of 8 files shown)


cgwalters:
Ok, pushed some more work here.  This clearly wants some tests, which is going to require a bit of (long overdue) work.  So I'll keep WIP until then.

src/libostree/ostree-repo-static-delta-compilation.c
```diff
@@ -1003,11 +1003,7 @@ generate_delta_lowlatency (OstreeRepo                       *repo,
       if (!check_object_world_readable (repo, from_checksum, &from_world_readable, cancellable, error))
         goto out;
       if (!from_world_readable)
-        {
-          g_hash_table_iter_steal (&hashiter);
-          g_hash_table_add (new_reachable_regfile_content, (char*)from_checksum);
-          continue;
-        }
+        continue;
```
dbnicholson: Took me a couple reads through this code again, but I agree this is right. There's no reason you'd want to include the old object in the delta. The rest of the code will DTRT as long as this unreadable object is kept out of the bsdiffs and rollsums to that the old object is not required on the pull side when applying the delta.

cgwalters:
Removing WIP.  It's clear to me now that we really need a "realistic repo" test which is what I was going for with `test-pull-many.sh`.  But running it currently is expensive; we'd need to refactor basically all of the tests to somehow derive from that - basically we only want to build the repo once.  Or maybe we move to having it be static, and stash it in S3, and simply require fetching it from the network?

giuseppe:
that sounds like a good idea, should be possible though to build it locally as well and run tests against it (I'd prefer to avoid network usage to S3 every time tests run)

cgwalters:
Ok, fixup ⬆️

giuseppe:
@rh-atomic-bot r+ 3465a8b

Issue #674, opened 2017-02-09
Add support for commit message file


paulvt:
We use a script to generate a automated commit message.  It would be a nice feature if the file could be passed to the commit (optionally in combination with `--editor`), similary to:
`git commit --file=<file> --edit`.
Currently, we have to pass the contents of the file as the contents of `--body="..."` which can lead to unwieldy command-lines.

N.B. If we currently do:
```
$ ostree commit ... --branch=test --subject="Some subject" --body="Line 1\nLine 2" --editor
```
It results into, in my opinion, awkward editor contents where the subject is put with the body below the explanation comments and without a newline separating the subject and body?

```
<cursor here>
# Please enter the commit message for your changes. The first line will
# become the subject, and the remainder the body. Lines starting
# with '#' will be ignored, and an empty message aborts the commit.
#
# Branch: test
Some subjet
Line 1
Line 2
```

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/681 - you can review/approve if you want.

PR #681, opened 2017-02-11
commit: Support -F/--body-file, like git

cgwalters:
This is more convenient to script for projects which haven't
yet made the leap to using the API.

Closes: https://github.com/ostreedev/ostree/issues/674


cgwalters:
@rh-atomic-bot delegate=paulvt

jlebon:
@rh-atomic-bot r+ 7ecde51

PR #682, opened 2017-02-11
build: Remove .PHONY for Rust shared library

cgwalters:
I have no idea why I made the lib `.PHONY` originally; it's clearly wrong, and I
noticed because when I was doing `sudo make install`, we were doing a rebuild,
which in turn triggered other things to be built, and they'd be owned by root.


jlebon:
@rh-atomic-bot r+ 7f828d3

PR #683, opened 2017-02-13
Fix segfault on empty remote string

OYTIS:
g_strconcat returns NULL if its first argument is null which makes 'strcmp' on line 110/116 fail with segfault.

 
Signed-off-by: Anton Gerasimov <anton@advancedtelematic.com>


cgwalters:
bot, add author to whitelist

cgwalters:
Thanks for the patch!  Code looks correct to me.  I just made two minor tweaks.  First I added a missing space between `if(remote)`.  Also, I added a bit more information to the commit message.

@rh-atomic-bot r+ 7094a3f

PR #684, opened 2017-02-14
rofiles-fuse: Support write/read_buf()

alexlarsson:
These allow us to avoid copying a lot of data around
in userspace. Instead we splice the data directly from
the fd to the destination fd.

### DIFF
--- a/src/rofiles-fuse/main.c
+++ b/src/rofiles-fuse/main.c
@@ -345,6 +345,26 @@ callback_create(const char *path, mode_t mode, struct fuse_file_info *finfo)
 }
 
 static int
+callback_read_buf (const char *path, struct fuse_bufvec **bufp,
+                   size_t size, off_t offset, struct fuse_file_info *finfo)
+{
+  struct fuse_bufvec *src;
+
+  src = malloc (sizeof (struct fuse_bufvec));
+  if (src == NULL)
+    return -ENOMEM;
+
+  *src = FUSE_BUFVEC_INIT (size);
+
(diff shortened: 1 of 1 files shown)


src/rofiles-fuse/main.c
```diff
@@ -345,6 +345,26 @@ callback_create(const char *path, mode_t mode, struct fuse_file_info *finfo)
 }
 
 static int
+callback_read_buf (const char *path, struct fuse_bufvec **bufp,
+                   size_t size, off_t offset, struct fuse_file_info *finfo)
+{
+  struct fuse_bufvec *src;
+
+  src = malloc (sizeof (struct fuse_bufvec));
```
cgwalters: Did fuse really add an interface for this that forces us to malloc for every read?   Reading the source, I guess so, but man, that's ugly.

cgwalters:
Reading the FUSE source, it looks like this interface was [added in 2010](https://github.com/libfuse/libfuse/commit/63322038) so I think we're safe to use this unconditionally.

@rh-atomic-bot r+  	009541e

cgwalters:
It looks like (unsurprisingly) the lowlevel API is more efficient.  So we should look at porting to that at some point.

alexlarsson:
That one is hard to back of a path-based backend though, it assumes a filesystem-like thing where you operate on individual inodes.

alexlarsson:
However, if you want performance, you should use -o kernel_cache,entry_timeout=60,attr_timeout=60 or something as arguments to rofiles-fuse.

PR #685, opened 2017-02-14
Release 2017.2

cgwalters:
We should get a release out to try to keep with at least a once-a-month cadence.
This one has some exciting stuff like libcurl and Rust, and various bugfixes.
Also importantly I want to cut this *before* we land some other bigger stuff, so
rpm-ostree can start using the reload_config API etc.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
 AC_PREREQ([2.63])
 dnl If incrementing the version here, remember to update libostree.sym too
-AC_INIT([libostree], [2017.1], [walters@verbum.org])
+AC_INIT([libostree], [2017.2], [walters@verbum.org])
 AC_CONFIG_HEADER([config.h])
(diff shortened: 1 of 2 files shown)


jlebon:
@rh-atomic-bot r+ fb34abf

```
+ ostree --repo=repo pull --disable-static-deltas --mirror origin exampleos/42/standard
timeout: the monitored command dumped core
Tue Feb 14 15:41:02 UTC 2017: Pulling content...
ERROR: tests/test-pull-many.sh - too few tests run (expected 1, got 0)
```

Huh, interesting, haven't seen this happen in a while on Red Hat CI. It failed in the last run of master too. Hard to tell without looking at the backtraces, though if it's due to libsoup, it might be time to backport those patches to Fedora as well.

jlebon:
The patches are in 2.57.1, though it's only been built for rawhide so far.

PR #689, opened 2017-02-16
contrib/golang: add more functions for commit

giuseppe:
I've added some more functions for being able to commits from Go
code.  Not all functions are exposed, but this is enough to commit a
directory.

### DIFF
--- a/contrib/golang/ostree.go
+++ b/contrib/golang/ostree.go
@@ -19,6 +19,29 @@ type Repo struct {
 	*GObject
 }
 
+type RepoFile struct {
+	ptr unsafe.Pointer
+}
+
+type MutableTree struct {
+	*GObject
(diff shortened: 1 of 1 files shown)


cgwalters:
Ah...I think we ended up moving this out into https://github.com/ostreedev/ostree-go right?  We should probably nuke `contrib/golang`.

cgwalters:
(Or am I missing something?)

giuseppe:
yes, I thought so.  Just when I tried ostree-go, I have not even managed to get it built so I was not sure about its status.  I'll try with that again.

giuseppe:
@cgwalters, I've prepared a PR to rm -r the contrib directory:

https://github.com/ostreedev/ostree/pull/690

PR #690, opened 2017-02-16
contrib/golang: rm directory

giuseppe:
rm -r the directory since we are keeping the Go bindings separately
under https://github.com/ostreedev/ostree-go


cgwalters:
@rh-atomic-bot r+ 9fe3957

PR #691, opened 2017-02-17
main: Make ostree --version output YAML (and add gitrev)

cgwalters:
I learned today that `docker version` does this and I really like
the idea.  While we have the patient open, also add the gitrev
with code taken from https://github.com/projectatomic/rpm-ostree/pull/584

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -19,6 +19,8 @@ include Makefile-decls.am
 
 shortened_sysconfdir = $$(echo "$(sysconfdir)" | sed -e 's|^$(prefix)||' -e 's|^/||')
 
+OSTREE_GITREV=$(shell if command -v git >/dev/null 2>&1 && test -d $(srcdir)/.git; then git describe --abbrev=42 --tags --always HEAD; fi)
+
 ACLOCAL_AMFLAGS = -I buildutil -I libglnx ${ACLOCAL_FLAGS}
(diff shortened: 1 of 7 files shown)


cgwalters:
```
$ ./ostree --version
libostree:
 Version: 2017.2
 Git: v2017.2-10-gd5a1222452e92c6f0c45696ef355d61193f54b17
 Features:
  - libcurl
  - libsoup
  - gpgme
  - libarchive
  - selinux
  - libmount
```

cgwalters:
Okay, so I have absolutely no idea right now why https://github.com/ostreedev/ostree/pull/691/commits/48ef1054dd82635e6f5a0d268261b90461782ea5 fixed the tests 😕  I added that to debug it, since it passed locally.

But anyways let's ignore that!  This should be good to go.

giuseppe:
I think "grep -q" exits as soon as it gets one occurrence as it doesn't really care how many there are, so causing an EPIPE in ostree.  It would probably work with "grep '- gpgme' > /dev/null".

Anyway, I think it is cleaner with a temporary file:

@rh-atomic-bot r+ 5df79df

cgwalters:
Ah right of course, and `-o pipefail` comes into play here.  That's like the second time I've forgotten about it and spent a while debugging in as many months.

PR #694, opened 2017-02-18
lib: Ensure an error is set in ensure_unlinked() if errno != ENOENT

cgwalters:
We hit this with:
```
27411 unlink("/boot/efi/EFI/fedora/grub.cfg.new") = -1 EROFS (Read-only file system)
```
from the grub2 code.

https://github.com/projectatomic/rpm-ostree/issues/633


giuseppe:
@rh-atomic-bot r+ edf1d14

PR #698, opened 2017-02-20
OSTree + smack

OYTIS:
Improves e.g. SMACK support. Motivated by a [bug](https://jira.automotivelinux.org/browse/SPEC-386) in AGL when using Cynara.
Depends on: https://github.com/GNOME/libglnx/pull/34
On RPi causes ENOENT error when temp_filename is NULL (no idea why glnx_dfd_name_set_all_xattrs doesn't give an error under the same conditions). Since temp_filename should not be NULL in this case, I attribute it to some problems with O_TMPFILE on Raspberry Pi and just compile ostree with '-DDISABLE_OTMPFILE'

Signed-off-by: Anton Gerasimov <anton@advancedtelematic.com>

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -273,6 +273,14 @@ AS_IF([ test x$with_selinux != xno ], [
 if test x$with_selinux != xno; then OSTREE_FEATURES="$OSTREE_FEATURES +selinux"; fi
 AM_CONDITIONAL(USE_SELINUX, test $with_selinux != no)
 
+AC_ARG_WITH(smack,
+AS_HELP_STRING([--with-smack], [Enable smack]),
+:, with_smack=no)
+AS_IF([ test x$with_smack = xyes], [
+  AC_DEFINE([WITH_SMACK], 1, [Define if we have smack.pc])
+])
+AM_CONDITIONAL(USE_SMACK, test $with_smack != no)
+
 dnl This is what is in RHEL7.2 right now, picking it arbitrarily
 LIBMOUNT_DEPENDENCY="mount >= 2.23.0"
 
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -110,6 +110,30 @@ write_file_metadata_to_xattr (int fd,
   return TRUE;
 }
 
+/* See https://github.com/ostreedev/ostree/pull/698 */
+#ifdef WITH_SMACK
+#define XATTR_NAME_SMACK "security.SMACK64"
+#endif
+
+static void
+ot_security_smack_reset_dfd_name (int dfd, const char *name)
+{
+#ifdef WITH_SMACK
+  char buf[PATH_MAX];
+  /* See glnx-xattrs.c */
+  snprintf (buf, sizeof (buf), "/proc/self/fd/%d/%s", dfd, name);
+  (void) lremovexattr (buf, XATTR_NAME_SMACK);
+#endif
+}
+
+static void
+ot_security_smack_reset_fd (int fd)
+{
+#ifdef WITH_SMACK
+  (void) fremovexattr (fd, XATTR_NAME_SMACK);
+#endif
+}
+
 gboolean
 _ostree_repo_commit_loose_final (OstreeRepo        *self,
                                  const char        *checksum,
@@ -221,6 +245,7 @@ commit_loose_object_trusted (OstreeRepo        *self,
 
       if (xattrs != NULL)
         {
+          ot_security_smack_reset_dfd_name (self->tmp_dir_fd, temp_filename);
           if (!glnx_dfd_name_set_all_xattrs (self->tmp_dir_fd, temp_filename,
                                                xattrs, cancellable, error))
             goto out;
@@ -252,6 +277,7 @@ commit_loose_object_trusted (OstreeRepo        *self,
 
           if (xattrs)
             {
+              ot_security_smack_reset_fd (fd);
               if (!glnx_fd_set_all_xattrs (fd, xattrs, cancellable, error))
                 goto out;
             }


cgwalters:
Hmm.  So in SELinux land there's a [setfscreatecon](https://linux.die.net/man/3/setfscreatecon) function that one can use for this.  I would assume SMACK has something similar...okay, just looked [but I don't see one](https://github.com/smack-team/smack/blob/master/libsmack/sys/smack.h).  Is SMACK even going to allow us to remove the label?  I think SELinux might allow it but then the kernel would mumble unhappily in `dmesg`.

Let me see if I can take a crack at doing this correctly in the SELinux case in libglnx, and then we can see if that model can be adapted to SMACK.  

Actually, I just remembered the best reference for this stuff is in the systemd source.  Let's [take a look](https://github.com/systemd/systemd/blob/baf327864a052e2fc063eaa83964b9b92edb0d21/src/basic/label.c).  Yes, so I think we need a variant of what systemd is doing here.  Let me take a quick crack at this.

OYTIS:
To tell the truth, I haven't gone deeply into SMACK. This patch definitely solves the problem with AGL using Cynara/SMACK.

cgwalters:
@OYTIS With your current system, are the SMACK xattrs defined on the build server, or are you defining some rule on the client side that sets the labels in `/sysroot/ostree` or something?  With [rpm-ostree](https://github.com/projectatomic/rpm-ostree) for [Atomic Host](http://www.projectatomic.io/download/) we're setting them on the server side.  libostree was in fact explicitly designed to support this model with SELinux.  We *don't* run the SELinux label regexps per client by default like rpm does.

While I'm not a SMACK expert, [this comment](https://jira.automotivelinux.org/browse/SPEC-386) looks wrong to me.  libostree is just writing the xattr which is part of the file/object metadata.  If that xattr is not correct, you need to fix your server-side build system I'd say.

To flesh out what rpm-ostree is doing a bit more, see:
https://github.com/projectatomic/rpm-ostree/blob/master/src/libpriv/rpmostree-postprocess.c#L1837
which is then just calling into [libostree's SELinux support](https://github.com/ostreedev/ostree/blob/f02dcc4997b228dcd551f2c3f01102e794599931/src/libostree/ostree-sepolicy.c#L450).

My impression is that your build system (rpm-ostree equivalent) should be doing something similar with SMACK.

OYTIS:
Xattrs are mostly assigned by the build side. The problem is that all files created by root seem to acquire security.SMACK64 xattr by default. So if you run 'ostree pull' as root, all newly created files in repo will get it. I'm not sure at the moment if it is just how SMACK works or there is some rule that can change this behaviour.

OYTIS:
My thought with this patch was that it is probably right for OSTree to assign exactly the same xattrs that were assigned on commit on build machine, no matter how target machine is configured. But maybe this error may be solved with fine-tuning SMACK as well.

cgwalters:
I sent a quick mail to Casey Schaufler and hopefully he'll enlighten us.  If we don't hear back I'll file a question issue on the [smack userspace](https://github.com/smack-team/smack/) project or see if they have a mailing list.

If we were to continue forward with this patch though, in the absence of more guidance/information, I'd say what we should do is remove *just* the smack xattr from newly created files before applying our own; something simple like: `(void) fremovexattr (fd, "security.smack")` or so.

cschaufler:
Smack (preferred over SMACK) always assigns a security.SMACK64 value when the inode is instantiated. If there isn't one on an existing file a default is taken from the superblock. There is no equivalent to setfscreatecon. Hopefully, your Smack configuration is simpler than your SELinux policy.

OYTIS:
@cgwalters As far as I understand how libostree works, any xattr that is present on a file in target repo, but not present on respective file in master repo will break target repo's consistency (wrt. how 'ostree fsck' understands it). It doesn't really matter much, or I my understanding is wrong?

cgwalters:
@cschaufler Thanks for joining in.

That's correct.  Now, we *could* change that.   But at least in the SELinux world, *every* file has a label.  So it hasn't come up there.   (Well, kind of...for flatpak at least the content isn't going to have labels, so we should fix that at some point)

But is the issue really no smack label?  It seems to me if your patch is *removing* the label, the label must be wrong to start with?  The systemd sources don't remove smack labels, they just overwrite them after.

cgwalters:
Oh hm...I just noticed smack has *multiple* labels, from the systemd git:

```
static const char* const smack_attr_table[_SMACK_ATTR_MAX] = {
        [SMACK_ATTR_ACCESS]     = "security.SMACK64",
        [SMACK_ATTR_EXEC]       = "security.SMACK64EXEC",
        [SMACK_ATTR_MMAP]       = "security.SMACK64MMAP",
        [SMACK_ATTR_TRANSMUTE]  = "security.SMACK64TRANSMUTE",
        [SMACK_ATTR_IPIN]       = "security.SMACK64IPIN",
        [SMACK_ATTR_IPOUT]      = "security.SMACK64IPOUT",
};
```

I wonder if what's happening is your build server is using smack and some of these labels are being assigned automatically server side, but not removed there?  Is the code you're using to generate commits to the ostree repository open source?

cschaufler:
All filesystem objects have SMACK64. Some program files may have SMACK64EXEC. Some directories will have SMACK64TRANSMUTE. SMACK64MMAP has not been used since the days of MeeGo. SMACK64IPIN and SMACK64IPOUT are only for sockets, and never show up on filesystems.

OYTIS:
@cgwalters 
The case is that the build system (Yocto) doesn't assign Smack attributes to created files, so they are committed without them. But once we pull them on the running system they are assigned automatically as @cschaufler said.
Maybe not assigning security.SMACK64 on build is the real bug, I'm not sure. At least there seems to be nothing in Smack support layers in Yocto that would care of this.
All [AGL](https://wiki.automotivelinux.org/agl-distro/source-code) code is open source, [this](https://git.automotivelinux.org/AGL/meta-agl-extra/tree/meta-sota/classes/image_types_ostree.bbclass) code is responsible for committing generated tree to an OSTree repo.

OYTIS:
But I think the problem is NOT in the commit stage. There are no Smack attributes in the original tree generated during normal build process.

cgwalters:
There not being a Yocto layer/tooling for this isn't surprising, because Yocto's default is to generate RPM/deb/etc packages, and the default is for (at least most) package managers to do this per-client.  (Well, at least for SELinux, I am not actually aware of the state of e.g. rpm + smack).

Anyways, I'm still confused how your patch helps.  It sounds like we're going to end up removing any smack xattrs that were auto-assigned by the kernel.   Which...[looking at the kernel source](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/security/smack/smack_lsm.c?id=9763dd6f8160dc9cc239fc2427c8173073204457#n1463), actually has the effect of resetting it to the default.

Maybe that's the bug here - the label we're inheriting from the superblock is wrong, and your patch ends up resetting it?

cgwalters:
So how about this instead?  Can you test it?

```
index 3448ab6..303ed6c 100644
--- a/configure.ac
+++ b/configure.ac
@@ -273,6 +273,14 @@ AS_IF([ test x$with_selinux != xno ], [
 if test x$with_selinux != xno; then OSTREE_FEATURES="$OSTREE_FEATURES +selinux"; fi
 AM_CONDITIONAL(USE_SELINUX, test $with_selinux != no)
 
+AC_ARG_WITH(smack,
+AS_HELP_STRING([--with-smack], [Enable smack]),
+:, with_smack=no)
+AS_IF([ test x$with_smack = xyes], [
+  AC_DEFINE([HAVE_SMACK], 1, [Define if we have smack.pc])
+])
+AM_CONDITIONAL(USE_SMACK, test $with_smack != no)
+
 dnl This is what is in RHEL7.2 right now, picking it arbitrarily
 LIBMOUNT_DEPENDENCY="mount >= 2.23.0"
 
index 07940f4..d67a726 100644
@@ -110,6 +110,30 @@ write_file_metadata_to_xattr (int fd,
   return TRUE;
 }
 
+/* See https://github.com/ostreedev/ostree/pull/698 */
+#ifdef WITH_SMACK
+#define XATTR_NAME_SMACK "security.SMACK64"
+#endif
+
+static void
+ot_security_smack_reset_dfd_name (int dfd, const char *name)
+{
+#ifdef WITH_SMACK
+  char buf[PATH_MAX];
+  /* See glnx-xattrs.c */
+  snprintf (buf, sizeof (buf), "/proc/self/fd/%d/%s", dfd, name);
+  (void) removexattr (path, XATTR_NAME_SMACK);
+#endif
+}
+
+static void
+ot_security_smack_reset_fd (int fd)
+{
+#ifdef WITH_SMACK
+  (void) fremovexattr (fd, XATTR_NAME_SMACK);
+#endif
+}
+
 gboolean
 _ostree_repo_commit_loose_final (OstreeRepo        *self,
                                  const char        *checksum,
@@ -221,6 +245,7 @@ commit_loose_object_trusted (OstreeRepo        *self,
 
       if (xattrs != NULL)
         {
+          ot_security_smack_reset_dfd_name (self->tmp_dir_fd, temp_filename);
           if (!glnx_dfd_name_set_all_xattrs (self->tmp_dir_fd, temp_filename,
                                                xattrs, cancellable, error))
             goto out;
@@ -252,6 +277,7 @@ commit_loose_object_trusted (OstreeRepo        *self,
 
           if (xattrs)
             {
+              ot_security_smack_reset_fd (fd);
               if (!glnx_fd_set_all_xattrs (fd, xattrs, cancellable, error))
                 goto out;
             }
```

cgwalters:
(Updated to `s,#ifdef HAVE_SMACK,#ifdef WITH_SMACK,` so the new code actually builds...)

If the approach seems OK to you I can push it into this PR.

OYTIS:
@cgwalters, thanks. 'removexattr' needs to be replaced with 'lremovexattr'. With this change everything works fine for me.

cgwalters:
@rh-atomic-bot delegate=OYTIS

Okay, I've updated this PR with that change.  Please [approve this PR](https://github.com/jlebon/redhat-ci/blob/master/README.md#interacting-with-redhat-ci) if it looks good to you.

OYTIS:
@rh-atomic-bot r+

cgwalters:
Hm, something got lost.

@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot r=OYTIS 4e553e2

cgwalters:
Possibly https://github.com/servo/homu/issues/82

PR #699, opened 2017-02-21
libtest: Re-enable quiet mode for building fs tree

cgwalters:
I think I commented this out while debugging something, and forgot to re-enable
it. Reading the log files should be a better again after this.


giuseppe:
@rh-atomic-bot r+ c943bb7

Issue #470, opened 2016-08-25
v0 of recursive deltas


cgwalters:
The deltas code/design supports recursion, but we didn't really follow through implementing this.  There's both client side and compiler work for this.

The goal should be that we support upgrading from both N-2 and N-1 to N reasonably.

gatispaeglis:
+1 This is a feature I would like to see in the upcoming ostree releases. I am wondering if this is something that could be available in a near future or is it low in the priority list?

cgwalters:
We can also do client-side delta chaining; the main issue is it would make the already-past-its-complexity-budget pull code even more complex.  But perhaps not unreasonably so, particularly if we started to spit `ostree-repo-pull.c` → `ostree-repo-pull-loose.c` and `ostree-repo-pull-deltas.c` or so.

In the short term, I'm thinking more https://github.com/ostreedev/ostree-releng-scripts/issues/10

sinnykumari:
Any plan to fully support applying recursive deltas on client side? This will be useful where we want to rely on deltas only repos (like FCOS mirroring https://github.com/fedora-infra/mirrormanager2/issues/258 ).

I was trying locally to pull ostree content using delta-only repo.
1. Created static delta from scratch for commit C1
2. client side pull works fine with option `--require-static-deltas` and pulls in C1
3. Created static delta from commit C1 to C2
4. Client side pull  to C2 works fine when client side repo is already on C1
5. Client side pull gives error if client wants to pull directly C2 from an empty repo . (error: Static deltas required, but none found for BRANCH to C2)
6. Similarly, consider more commits like C3 and C4 and we have created incremental delta from C2 to C3 and then C3 to C4. Client can't update from C2 to C4 because there is no direct static delta from C2 to C4

It will be nice if ostree can apply recursive delta on client side so that we don't have to really create static delta for all permutations (which will grow as the project grows).

PR #701, opened 2017-02-21
pull: Support require-deltas=upgrade

cgwalters:
Our general story for "production" repositories is that people use static
deltas. This is part of a story of better supporting that. In this model we
don't want clients to fall back to per-object fetch if due to some
config/systems failure deltas don't exist.

This will also make it easier for us to potentially introduce a "deltas only"
repository format in the future.

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -181,6 +181,14 @@ Boston, MA 02111-1307, USA.
       </varlistentry>
 
       <varlistentry>
+        <term><varname>require-deltas=upgrade</varname></term>
+        <listitem><para>Pull operations will require deltas for upgrades.
+        Ensures that client systems don't fall back to per-object fetch due to
+        server configuration error or badly sync'd data for
+        example.</para></listitem>
+      </varlistentry>
+
+      <varlistentry>
         <term><varname>unconfigured-state</varname></term>
         <listitem><para>If set, pulls from this remote will fail with the configured text.  This is intended for OS vendors which have a subscription process to access content.</para></listitem>
       </varlistentry>
--- a/tests/pull-test.sh
+++ b/tests/pull-test.sh
@@ -19,12 +19,17 @@
 
 set -euo pipefail
 
+function repo_init_origin() {
+    ${CMD_PREFIX} ostree --repo=repo remote delete origin 2>/dev/null || true
+    ${CMD_PREFIX} ostree --repo=repo remote add --set=gpg-verify=false "$@" origin $(cat httpd-address)/ostree/gnomerepo
+}
+
 function repo_init() {
     cd ${test_tmpdir}
     rm repo -rf
     mkdir repo
     ${CMD_PREFIX} ostree --repo=repo init
-    ${CMD_PREFIX} ostree --repo=repo remote add --set=gpg-verify=false origin $(cat httpd-address)/ostree/gnomerepo
+    repo_init_origin
 }
 
 function verify_initial_contents() {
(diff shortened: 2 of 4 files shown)


tests/pull-test.sh
```diff
@@ -192,12 +216,21 @@ cd ${test_tmpdir}
...
     assert_not_reached "--require-static-deltas unexpectedly succeeded"
 fi
 assert_file_has_content err.txt "deltas required, but none found"
 ${CMD_PREFIX} ostree --repo=repo fsck
 
+cd ${test_tmpdir}
+# And test it via the origin
+repo_init_origin --set=require-deltas=upgrade
+if ${CMD_PREFIX} ostree --repo=repo pull origin main 2>err.txt; then
+    assert_not_reached "--require-static-deltas unexpectedly succeeded"
```
jlebon: Nit: update the error msg.

man/ostree.repo-config.xml
```diff
@@ -181,6 +181,14 @@ Boston, MA 02111-1307, USA.
       </varlistentry>
 
       <varlistentry>
+        <term><varname>require-deltas=upgrade</varname></term>
```
jlebon: Are you planning for this to be a list of scenarios in the future? Would be cleaner & clearer to make it a boolean, e.g. `require-delta-upgrades` or `require-deltas-on-upgrade`, no?

cgwalters:
The recursive deltas issue is https://github.com/ostreedev/ostree/issues/470

As far as outdated clients, I'd say 2/3 of the burden here should fall on the OS vendor in providing a wealth of deltas.  Really, disk space for this is so cheap in the world of S3/GCS etc.

Another thing we could do is encourage having `--empty` deltas (i.e. "from none") for the latest tip, so if your system is really outdated, you fall back to redownloading everything (i.e. the status quo for OCI), and for really outdated systems, RPM/deb type package systems start to approach redownloading everything anyways.

Another idea is some metadata in the summary file which instructs the client to only fall back to loose fetches if it's sufficiently old.

miabbott:
Aplologies for not discussing the actual code here, but I wanted to throw in my thoughts on this approach.

> That makes sense. But unless we start supporting recursive deltas, it's a bit hard to guarantee. E.g. what if a client hasn't been updated in a while and is behind by N commits, but the content provider only generates static deltas for the last N - 1 commits?

I share @jlebon's concern in this scenario.  I don't think it would be much of a problem in a rigid, managed datacenter environment, but the community user that doesn't do upgrades frequently could get bit by this.  (Related to this, on one of our internal mailing lists, I saw a message from someone who upgraded from F21 to F25 in one fell swoop.  People like that user, while rare, would need a fallback.)

> Another idea is some metadata in the summary file which instructs the client to only fall back to loose fetches if it's sufficiently old.

This idea sounds promising.  Would like to hear more about how it could be accomplished.

cgwalters:
Presumably this is a desktop system; anyone who was actually *running* Fedora 21 (or in general is not keeping firefox + kernel up to date at a minimum) is an irresponsible threat to the information security of their employer and their coworkers.

The fallback here is easy, edit the remote config to disable the flag.

Note also for Fedora today with multiple branches per major, it's a rebase to upgrade, which wouldn't fall under this category right now.

Anyways, big picture for the "really ancient" scenario, having the OS vendor do either "from none" deltas or just fall back to loose fetches is going to work OK.

Maybe though the flag should look something like `require-deltas=upgrade=2 months` or something - so the repo operator would provide deltas for the last 2 months of releases, and the client fall back if due to server misconfiguration/transient sync error the delta didn't exist.

cgwalters:
Labeling as needs-rework for design.

Issue #696, opened 2017-02-20
ostree should avoid g_spawn usage


gatispaeglis:
.. as it hijacks the SIGCHLD handler for the entire process. Details in:

https://mail.gnome.org/archives/gtk-devel-list/2016-October/msg00015.html
https://bugreports.qt.io/browse/QTBUG-56338

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/702

PR #702, opened 2017-02-22
grub2: Use g_spawn_sync() rather than GSubprocess to avoid SIGCHLD

cgwalters:
Due to the async nature of `GSubprocess` it grabs `SIGCHLD` which
affects other software which might be using libostree, such as
QtOTA.

Closes: https://github.com/ostreedev/ostree/issues/696


cgwalters:

jlebon:
@rh-atomic-bot r+ 7043492

Issue #704, opened 2017-02-23
/boot readonly fails after txn if not altering bootloader config


cgwalters:
In https://github.com/ostreedev/ostree/pull/213 I added support for `/boot` being a readonly mount.

However,

```
# rpm-ostree install cowsay
...
Overlaying... done
Writing rpmdb... done
Writing OSTree commit... done
Copying /etc changes: 28 modified, 0 removed, 342 added
Transaction complete; bootconfig swap: no deployment count change: 0
error: Read-only file system
```

Note it crashes *after* the txn.  This is because we need to keep `/boot` writable for the cleanup phase as well.

PR #705, opened 2017-02-23
deploy: Correctly use libmount unref() calls rather than free()

cgwalters:
We saw a random ostree SEGV start popping up in our CI environment:
https://github.com/projectatomic/rpm-ostree/pull/641#issuecomment-281870424

Looking at this code more and comparing it to what util-linux does, I noticed we
had a write-after-free, since `mnt_unref_table()` will invoke
`mnt_unref_cache()` on its cache, and that function does:

```
	if (cache) {
		cache->rfcount--;
```

unconditionally.

Fix this by using `unref()`.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1692,8 +1692,8 @@ is_ro_mount (const char *path)
 
   fs = mnt_table_find_target(tb, path, MNT_ITER_BACKWARD);
   is_mount = fs && mnt_fs_get_target (fs);
-  mnt_free_cache (cache);
-  mnt_free_table (tb);
+  mnt_unref_cache (cache);
+  mnt_unref_table (tb);
 
   if (!is_mount)
     return FALSE;


cgwalters:
Though, perhaps based on https://github.com/ostreedev/ostree/issues/704 and https://github.com/ostreedev/ostree/issues/706 what we should really do is just revert https://github.com/ostreedev/ostree/pull/213/commits/32d43695d89f3c79d7a13b04e6b44cbbe45630a6 for now?

cgwalters:
Eh, it's probably safest to just go with this - a revert would be messy and less obviously safe.

jlebon:
Yeah, had no luck at all reproducing https://github.com/projectatomic/rpm-ostree/pull/641#issuecomment-281870424 locally. In re. to reverting https://github.com/ostreedev/ostree/pull/213/commits/32d43695d89f3c79d7a13b04e6b44cbbe45630a6, we don't want to break people's setups if they started relying on it. So I'm +1 on just getting this in.

@rh-atomic-bot r+ 4c3ef23

PR #709, opened 2017-02-23
man/repo-config: Document mirrorlist

cgwalters:
We should get more strict about docs. Add some text about `contenturl` and
mirrorlists.


jlebon:
Nice! @rh-atomic-bot r+ 43dfcb6

PR #710, opened 2017-02-24
pull: Use all available commits for delta sources

cgwalters:
The previous logic for static deltas was to use as a FROM
revision the current branch tip.  However, we want
to support deltas between branches in an automatic
fashion.

If a summary file is available, we already have an
enumerated list of deltas - so the logic introduced
here is to search it, and find the newest commit
we have locally that matches the TO revision target.

This builds on some thoughts from
https://github.com/ostreedev/ostree/pull/151#issuecomment-232390232

Closes: https://github.com/ostreedev/ostree/pull/151

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1866,6 +1866,86 @@ process_one_static_delta (OtPullData   *pull_data,
   return ret;
 }
 
+/* Loop over the static delta data we got from the summary,
+ * and find the newest commit for @out_from_revision that
+ * goes to @to_revision.
+ */
+static gboolean
+get_best_static_delta_start_for (OtPullData *pull_data,
+                                 const char *to_revision,
+                                 char      **out_from_revision,
+                                 GCancellable *cancellable,
+                                 GError      **error)
+{
+  GHashTableIter hiter;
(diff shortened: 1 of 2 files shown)


cgwalters:
Mostly WIP due to lack of tests.

src/libostree/ostree-repo-pull.c
```diff
@@ -1866,6 +1866,87 @@ process_one_static_delta (OtPullData   *pull_data,
...
+          candidate_ts > newest_candidate_timestamp)
+        {
+          newest_candidate = candidate;
+          newest_candidate_timestamp = candidate_ts;
+        }
+    }
+
+  if (newest_candidate)
+    *out_from_revision = g_strdup (newest_candidate);
+  return TRUE;
```
giuseppe: should `out_from_revision` be set to NULL if there is no `newest_candidate`?  Otherwise it won't be initialized in this case

cgwalters:
Now with tests ☑️.  I also verified that reverting the code but keeping the tests caused them to fail.

cgwalters:
(Removing WIP)

giuseppe:
@rh-atomic-bot r+ 83d8979

PR #711, opened 2017-02-24
tree-wide: Squash noncritical compiler warnings

cgwalters:
Should fix everything from
<https://kojipkgs.fedoraproject.org//packages/ostree/2017.2/3.fc25/data/logs/x86_64/build.log>

Anything that uses autocleanups should *always* be initialized directly I think,
even if a few lines down we directly assign, since this way it's more robust
against refactoring.

And the `freopen()` warnings are right - IMO we should *always* check return
values.

### DIFF
--- a/src/ostree/ostree-trivial-httpd.c
+++ b/src/ostree/ostree-trivial-httpd.c
@@ -520,7 +520,7 @@ run (int argc, char **argv, GCancellable *cancellable, GError **error)
         }
       else
         {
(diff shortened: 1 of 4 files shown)


jlebon:
@rh-atomic-bot r+ 0d454d1

jlebon:
I think the state for this got nixed when we restarted Homu.
@rh-atomic-bot retry

PR #712, opened 2017-02-24
deploy/libmount: Fix build with old util-linux 2.23 (CentOS7)

cgwalters:
https://github.com/ostreedev/ostree/pull/705 broke the build
on CentOS 7 which only has util-linux 2.23.

When I was thinking about this, I realized that there must really be a way to
make this safe even for older versions. Looking at that version of util-linux,
all we need to do is invert the order of frees so we `mnt_free_table()` *before*
`mnt_free_cache()`, like util-linux does:

https://github.com/karelzak/util-linux/blob/stable/v2.23/sys-utils/eject.c#L1131

We still use the `_unref()` versions if available.  I also fixed
the ordering there too for double plus redundant safety.


cgwalters:
Should fix https://ci.centos.org/view/Atomic/job/atomic-rdgo-centos7/13700/artifact/build-logs/failed/ostree/build.log

jlebon:
@rh-atomic-bot r+ a315603

PR #714, opened 2017-03-02
checkout: Support a "pure addition" mode

cgwalters:
I plan to use this for `rpm-ostree livefs`.
https://github.com/projectatomic/rpm-ostree/issues/639

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -641,6 +641,8 @@ checkout_tree_at (OstreeRepo                        *self,
   gboolean did_exist = FALSE;
   glnx_fd_close int destination_dfd = -1;
   int res;
+  struct stat repo_dfd_stat;
+  struct stat destination_stat;
   g_autoptr(GVariant) xattrs = NULL;
   g_autoptr(GFileEnumerator) dir_enum = NULL;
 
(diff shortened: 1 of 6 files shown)


cgwalters:
WIP due to lack of tests, cmdline wrapper.

cgwalters:
Now with tests ✅ - removing WIP.

src/libostree/ostree-repo-checkout.c
```diff
@@ -362,6 +375,12 @@ checkout_file_hardlink (OstreeRepo                          *self,
     {
       ret_was_supported = FALSE;
     }
+  else if (errno == EEXIST && options->overwrite_mode == OSTREE_REPO_CHECKOUT_OVERWRITE_ADD_FILES)
+    {
+      /* In this mode, we keep existing content.  Still need to mark the hardlink as supported.
+       */
+      ret_was_supported = TRUE;
```
jlebon: One concern with this is that the kernel checks for `EEXISTS` before checking whether hardlinking is supported. So we can't really assert this, right? Looking at `checkout_one_file_at()`, it relies on this info to update the devino cache, which will be wrong after this. Maybe add e.g. `out_exists` also so callers can distinguish this?

cgwalters: Oooh.  That's a good catch.  That would have been spectacularly painful to debug later.

cgwalters:
Ok, rebased 🏄, added a prep patch and a fixup.

src/libostree/ostree-repo-checkout.c
```diff
@@ -579,14 +581,14 @@ checkout_one_file_at (OstreeRepo                        *repo,
...
-                                   FALSE, &did_hardlink,
+                                   FALSE, &hardlink_res,
                                    cancellable, error))
         {
           g_prefix_error (error, "Using new cached uncompressed hardlink of %s to %s: ", checksum, destination_name);
           goto out;
         }
 
-      need_copy = !did_hardlink;
+      need_copy = hardlink_res == HARDLINK_RESULT_NOT_SUPPORTED;
```
jlebon: Parentheses on these, please!

cgwalters: Done ⬇️

jlebon:
Nice! Looks like some tests might need to be adjusted to grep for the new error message.

cgwalters:
Eh, my fixups went wrong, I just squashed to simplify.  One of the bugs was we need to initialize `hardlink_res`, since it's not set if we're not going to try to hardlink.

jlebon:
@rh-atomic-bot r+ 0188b3c

PR #715, opened 2017-03-02
pull: don't use static deltas if archive repo

jlebon:
In https://github.com/ostreedev/ostree/pull/408, we disabled the use of
static deltas when mirroring. Later,
https://github.com/ostreedev/ostree/pull/506 loosened this up again so
that we could use static deltas when mirroring into bare{-user} repos.

However, the issue which originally spurrred #408 is even more generic
than that: we want to avoid static deltas for any archive repo, not just
when doing a mirror pull. This patch tightens this up, and also
relocates the decision code to make it easier to read.


cgwalters:
@rh-atomic-bot r+  45eeab1

cgwalters:
@rh-atomic-bot r+ 6d4f6f7

PR #716, opened 2017-03-03
grub2: Use "linux16" only on x86/x86_64

cgwalters:
Got a report that a Fedora Atomic Host built for ppc64le didn't work with the
`linux16`, it needed `linux`. See the comments for more links.

### DIFF
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -29,6 +29,20 @@
 
 #include <string.h>
 
+/* I only did some cursory research here, but it appears
+ * that we only want to use "linux16" for x86 platforms.
+ * At least, I got a report that "linux16" is definitely wrong
+ * for ppc64.  See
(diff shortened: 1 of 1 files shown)


cgwalters:
@rh-atomic-bot delegate=baude

vtolstov:
On UEFI you need not linux16 and linux, but linuxefi. Does it possible to
add this?

3 Мар 2017 г. 5:32 пользователь "Colin Walters" <notifications@github.com>
написал:

cgwalters:
What distribution are you using?
My laptop is EFI and runs [Atomic Workstation](https://pagure.io/atomic-ws) which is basically Fedora 25+rpm-ostree, and I have `linuxefi` in my `/boot/efi/EFI/fedora/grub.cfg`, which definitely works.

I haven't dug into the GRUB world deeply here...one of the seriously messy things here is how many [patches Fedora has for grub2](http://pkgs.fedoraproject.org/cgit/rpms/grub2.git/tree/?h=f25).  So there may be some variance with upstream in this respect.

baude:
Tested on a handbuilt ppc64le atomic and this corrected the problem.  Perfect and thanks!

cgwalters:
@rh-atomic-bot r=vathpela 13ea685

PR #718, opened 2017-03-03
Allow and start using C99 declaration-after-statement

cgwalters:
I've seen code in a few places that I think on balance is definitely better this
way.  Some of our functions have huge variable declaration sections.
This change includes one small example where we could start using declarations
after statements.

A concern I had was - how does this interact with `__attribute__((cleanup))` and
early returns? I tested it, and AFAICS the behavior is what you'd expect - the
cleanup function isn't called if its variable isn't reachable.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -33,7 +33,6 @@ CC_CHECK_FLAGS_APPEND([WARN_CFLAGS], [CFLAGS], [\
         -Werror=incompatible-pointer-types \
         -Werror=misleading-indentation \
 	-Werror=missing-include-dirs -Werror=aggregate-return \
-	-Werror=declaration-after-statement \
 ])
 AC_SUBST(WARN_CFLAGS)
 
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -4309,16 +4309,14 @@ find_keyring (OstreeRepo          *self,
               OstreeRemote        *remote,
               GCancellable        *cancellable)
 {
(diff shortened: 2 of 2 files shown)


cgwalters:
My test program:
```
#define _GNU_SOURCE 1
#include <stdio.h>
#include <sys/xattr.h>
#include <sys/types.h>
#include <string.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <err.h>

static int c = 0;

static inline void
cleanup_test (void *p)
{
  void **pp = p;
  printf ("cleanup %u: %p\n", c, *pp);
  c++;
}

static void
testfunc (const char *argv1)
{
  __attribute__((cleanup(cleanup_test))) char *a = NULL;
  printf ("stage 0\n");

  if (strcmp (argv1, "err") == 0)
    return;

  /* The cleanup for this isn't called if we return */
  __attribute__((cleanup(cleanup_test))) char *b;

  printf ("stage 1\n");
}

int
main (int argc, char **argv)
{
  testfunc (argv[1]);
}
```

```
$ gcc -Wall -o foo foo.c && ./foo err
stage 0
cleanup 0: (nil)
$ gcc -Wall -o foo foo.c && ./foo x
stage 0
stage 1
cleanup 0: 0x1
cleanup 1: (nil)
```

jlebon:
Awesome. It just makes way more sense to declare variables where they'll be used.

@rh-atomic-bot r+ 5436fd1

cgwalters:
libsoup flake.

@rh-atomic-bot retry

PR #719, opened 2017-03-04
fetcher/curl: Fix a leak of timer source caught by ASAN

cgwalters:
I had to rebuild `glib` with `-fsanitize=address` in order to get a stack trace
to finally get this one.  However, *installing* that glib "system wide"
in my container breaks everything (including `rpm-ostree`, `dnf`, `pkg-config` etc.)
that wasn't built with ASAN.

So my test scenario right now is to extract the libs and do e.g.:
```
make && env LD_LIBRARY_PATH=$HOME/src/distgit/fedora/glib2/asan-libs make check TESTS=tests/test-basic.sh
```

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -416,11 +416,13 @@ timer_cb (gpointer data)
 {
   OstreeFetcher *fetcher = data;
   CURLMcode rc;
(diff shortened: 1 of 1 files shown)


jlebon:
Hooray for ASAN!
@rh-atomic-bot r+ 6d9c43f

PR #720, opened 2017-03-06
build: Fix disabling --enable-man if xsltproc is not available

pwithnall:
If --enable-man=maybe (the default), and xsltproc is not available, the
configuration code would check for it, correctly set enable_man=no, then
incorrectly overwrite that with enable_man=yes, which would result in
later trying to execute $(XSLTPROC) when it’s empty.


cgwalters:
@rh-atomic-bot r+  51cf65f

PR #723, opened 2017-03-07
Disable "ostree trivial-httpd" by default now

cgwalters:
This goes farther along the path of deprecating it; consumers should
have been ported at this point.  Though probably a lot of people
may still use `rpm-ostree-toolbox`.


jlebon:
@rh-atomic-bot r+ b6c798a

cgwalters:
@rh-atomic-bot retry

PR #724, opened 2017-03-08
repo/checkout: fix 32-bit builds

chergert:
__dev_t is 64-bit even on 32-bit Linux systems such as i386.


cgwalters:
Thanks!

@rh-atomic-bot r+ add3029

PR #725, opened 2017-03-08
Allow setting custom async progress callback timeout

GeorgesStavracas:
Flatpak is becoming a thing, and it makes heavy usage of OSTree, and GNOME Software allows managing Flatpak application. That exposed an issue where the progress is updated only once every 1 second, which works well for command line but not for so well for UIs.

This PR fixes that by adding an option to set the timeout to call the async progress callback.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2589,6 +2589,7 @@ reinitialize_fetcher (OtPullData *pull_data, const char *remote_name, GError **e
  *   * override-url (s): Fetch objects from this URL if remote specifies no metalink in options
  *   * inherit-transaction (b): Don't initiate, finish or abort a transaction, usefult to do mutliple pulls in one transaction.
  *   * http-headers (a(ss)): Additional headers to add to all HTTP requests
+ *   * update-frequency (u): Frequency to call the async progress callback in milliseconds, if any; only values higher than 0 are valid
  */
 gboolean
 ostree_repo_pull_with_options (OstreeRepo             *self,
@@ -2613,6 +2614,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   char **configured_branches = NULL;
   guint64 bytes_transferred;
   guint64 end_time;
+  guint update_frequency = 0;
   OstreeRepoPullFlags flags = 0;
   const char *dir_to_pull = NULL;
   g_autofree char **dirs_to_pull = NULL;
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -2589,6 +2589,7 @@ reinitialize_fetcher (OtPullData *pull_data, const char *remote_name, GError **e
  *   * override-url (s): Fetch objects from this URL if remote specifies no metalink in options
  *   * inherit-transaction (b): Don't initiate, finish or abort a transaction, usefult to do mutliple pulls in one transaction.
  *   * http-headers (a(ss)): Additional headers to add to all HTTP requests
+ *   * update-frequency (i): Frequency to call the async progress callback, if any; only values higher than 0 are valid
```
matthiasclasen: You probably want to say "in milliseconds" here, just to make things clear

And why make it "i" if you only allow nonnegative values ?

GeorgesStavracas:
I'm not sure how my changes are related to the test failures...

cgwalters:
bot, add author to whitelist

src/libostree/ostree-repo-pull.c
```diff
@@ -2613,6 +2614,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   char **configured_branches = NULL;
   guint64 bytes_transferred;
   guint64 end_time;
+  guint update_frequency;
```
jlebon: This needs to be initialized to 0, right? In case we don't unpack anything.

GeorgesStavracas:
Fixed the build errors (thanks @jlebon!)

src/ostree/ot-builtin-pull.c
```diff
@@ -252,6 +254,9 @@ ostree_builtin_pull (int argc, char **argv, GCancellable *cancellable, GError **
     g_variant_builder_add (&builder, "{s@v}", "depth",
                            g_variant_new_variant (g_variant_new_int32 (opt_depth)));
    
+    g_variant_builder_add (&builder, "{s@v}", "update-frequency",
+                           g_variant_new_variant (g_variant_new_int32 (opt_frequency)));
```
jlebon: s/int32/uint32/

jlebon:
Looks sane to me. Just one small nit, and then I think we can get this in!

> I'm not sure how my changes are related to the test failures...

Yeah, the Travis tester is a bit flaky unfortunately.

GeorgesStavracas:
Nice catch, @jlebon. Updated the PR

jlebon:
@rh-atomic-bot r+ 9d3aa95

PR #726, opened 2017-03-09
Release 2017.3

cgwalters:
It's been almost a month, I think the current git is working well and
not too risky.  We have some new API additions which I think the
various consumers of them are going to want.


jlebon:
🚂 All aboard! 🚂
@rh-atomic-bot r+ 9054e58

PR #730, opened 2017-03-10
Fix includes if built against musl

misery:
LOCK_* is defined in sys/file.h

http://git.musl-libc.org/cgit/musl/tree/include/sys/file.h


cgwalters:
I've never heard of `sys/file.h` before.  Must be something else in glibc pulling it in?  Anyways, seems fine to me.

@rh-atomic-bot r+ 7768f15

PR #732, opened 2017-03-10
Avoid unnecessary includes

misery:
"return" will do the same here. Also that style is used at the
end of the function.


jlebon:
Thanks!
@rh-atomic-bot r+ 42ab0ff

PR #733, opened 2017-03-13
builtin/show: Convert to direct return/decl-after-stmt style

cgwalters:
I just happened to be reading this code; still testing
the waters for the new code style.  If we decide to do more
of this it's likely worth trying out a [Coccinelle](http://coccinelle.lip6.fr/)
patch or something.

### DIFF
--- a/src/ostree/ot-builtin-show.c
+++ b/src/ostree/ot-builtin-show.c
@@ -52,17 +52,13 @@ do_print_variant_generic (const GVariantType *type,
                           const char *filename,
                           GError **error)
 {
-  gboolean ret = FALSE;
   g_autoptr(GVariant) variant = NULL;
 
   if (!ot_util_variant_map_at (AT_FDCWD, filename, type, TRUE, &variant, error))
-    goto out;
+    return FALSE;
 
   ot_dump_variant (variant);
-
-  ret = TRUE;
- out:
-  return ret;
+  return TRUE;
 }
 
 static gboolean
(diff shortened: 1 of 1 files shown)


jlebon:
It's nice that it's more compact too!

> If we decide to do more of this it's likely worth trying out a Coccinelle patch or something.

I don't have any strong opinion either way, though I think I'd vote for just doing those conversions piecewise as we touch them just to make `git blame` more useful (though I'll note that unlike e.g. indentation patches, the important logic lines are mostly untouched).

---

Interesting, the `curl` suite *also* failed at `test-pull-many.sh`, but while generating static delta:

```
+ env LD_PRELOAD=/var/tmp/checkout/tests/libreaddir-rand.so ostree --repo=ostree-srv/exampleos/repo static-delta generate '--from=exampleos/42/standard^^' '--to=exampleos/42/standard^'
  To:   e16208ef1844df7c3f84e6019f75336bfd190e43235cff05b3ead176633e6a8d
Generating static delta:
  From: e16208ef1844df7c3f84e6019f75336bfd190e43235cff05b3ead176633e6a8d
timeout: the monitored command dumped core
  To:   f36cecf42b1f21668b1ef1cd298a682bfc0e75b1561368f43d1618416cd49fe5
ERROR: tests/test-pull-many.sh - missing test plan
ERROR: tests/test-pull-many.sh - exited with status 124
```

Hmm, so there might be something lurking in ostree proper there.

jlebon:
Anyway, I don't think this is related to this PR.
@rh-atomic-bot r+ a642765

cgwalters:
Yep, let's track in https://github.com/ostreedev/ostree/issues/736

PR #735, opened 2017-03-13
core: Add runtime ostree_check_version()

cgwalters:
[Previously](https://github.com/ostreedev/ostree/pull/728) we added compile-time
checking for versions, but there are use cases for runtime checking as well,
because in a number of API calls we use `GVariant` as an API extension
mechanism.


jlebon:
@rh-atomic-bot r+ ae464c3

GeorgesStavracas:
Not that I'm pretending to be an official OSTree developer, but the commit looks good

giuseppe:
For a moment I thought it would make sense to move the logic we have in the macro into the C function and make the macro use it, but I see the benefits of having two versions one for runtime and one for build time: LGTM.

PR #738, opened 2017-03-15
Use OpenSSL for checksums

cgwalters:
Add an OpenSSL backend to the checksum input stream, which is where we do a lot
of checksumming (object commit, static deltas).

The raw OpenSSL performance is
[approximately double](https://gist.github.com/cgwalters/169349fd1c06fd4fb4d3a7ce33303222) on
my laptop; not only does OpenSSL have e.g. hand-tuned x86_64 assembly, the
current implementation uses the
[Intel SHA extensions](https://en.wikipedia.org/wiki/Intel_SHA_extensions).

Another reason to do this is I was idly thinking about adding
[Curve25519](https://en.wikipedia.org/wiki/Curve25519) signatures (like e.g.
Alpine does) instead of/in addition to GPG.  The rationale for that is
that GPG is pretty heavyweight, both in code footprint and the simple
fact that EC keys are way smaller.

I didn't benchmark ostree with this; we have bigger performance problems
really like the fact we just malloc way too much.  But, it's a step
in the right direction I think in combination with the libcurl work
where we're linking to openssl anyways.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -889,7 +889,10 @@ write_object (OstreeRepo         *self,
   g_mutex_unlock (&self->txn_stats_lock);
 
   if (checksum_input)
-    ret_csum = ot_checksum_instream_dup_digest (checksum_input, NULL);
+    {
+      g_assert (actual_checksum);
+      ret_csum = ostree_checksum_to_bytes (actual_checksum);
+    }
 
   ret = TRUE;
   ot_transfer_out_value(out_csum, &ret_csum);
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -619,12 +619,11 @@ write_object (OstreeRepo         *self,
   OstreeRepoMode repo_mode;
   g_autofree char *temp_filename = NULL;
   g_autofree guchar *ret_csum = NULL;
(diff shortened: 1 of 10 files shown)


cgwalters:
No urgency on this, just something I decided to try for experimentation.

src/libostree/ostree-repo-commit.c
```diff
@@ -889,7 +889,10 @@ write_object (OstreeRepo         *self,
   g_mutex_unlock (&self->txn_stats_lock);
 
   if (checksum_input)
-    ret_csum = ot_checksum_instream_dup_digest (checksum_input, NULL);
+    {
+      g_assert (actual_checksum);
+      ret_csum = ostree_checksum_to_bytes (actual_checksum);
```
cgwalters: This was necessary because unlike `GChecksum`, openssl apparently resets the context after calling `EVP_DigestFinal_ex()`.

cgwalters:
❤❤❤ ASAN - previously `AddressSanitizer: heap-buffer-overflow on address 0x6030000099f0 ... located 0 bytes to the right of 32-byte region`

jlebon:
This is really cool! The speedup here is a bit less dramatic at 1.79, though obviously still highly attractive. It'd be interesting to benchmark this for bare repos! (I suppose committing to an archive repo will still be dominated by compression).

Anyway, code looks good!
@rh-atomic-bot r+ 4ccb151

PR #739, opened 2017-03-15
pull: Also skip partial ½ commits for deltas if no summary file

cgwalters:
I was playing around in a FAH vagrant box, and hit:

```
Receiving delta parts: 3/4 453.2 kB/s 1.8 MB/145.8 MB
error: opcode set-read-source: No such file object b6e54ba3471b9c116ce6b9bfbf9e55fec60d35cfdb9ae5ae1ee219af02a591b7
```

This is because this host version doesn't yet have
https://github.com/ostreedev/ostree/pull/710

While Fedora is using summary files, some organizations may not. Let's copy the
logic very recently added in that commit to handle the legacy case too.

No new tests since this is a nice-to-have - we really do
expect people to be using summary files now.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1198,6 +1198,14 @@ gpg_verify_unwritten_commit (OtPullData         *pull_data,
 }
 
 static gboolean
+commitstate_is_partial (OtPullData   *pull_data,
+                        OstreeRepoCommitState commitstate)
+{
+  return pull_data->legacy_transaction_resuming
+    || (commitstate & OSTREE_REPO_COMMIT_STATE_PARTIAL) > 0;
+}
+
+static gboolean
 scan_commit_object (OtPullData         *pull_data,
                     const char         *checksum,
                     guint               recursion_depth,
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -3257,9 +3265,27 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
+            {
+              OstreeRepoCommitState from_commitstate;
+
+              if (!ostree_repo_load_commit (pull_data->repo, from_revision, NULL,
+                                            &from_commitstate, error))
+                goto out;
+
+              /* Was it partial?  OK, we can't use it. */
+              if (commitstate_is_partial (pull_data, from_commitstate))
+                have_valid_from_commit = FALSE;
```
jlebon: Should we instead do e.g.

```c
if (from_revision)
  g_clear_pointer (&from_revision, g_free);
```

Otherwise, we won't be using `from NULL` if we happen to
have a partial commit, right?

cgwalters: That topic is kind of...messy.  Which is mostly papered over by the fact we aren't making (let's call them) "scratch deltas" in most streams I know of.  Under what scenarios is it better to download a scratch delta versus doing the archive-individual fetches?

A useful heuristic for the ostree-host case might be to only enable the "scratch delta" case if the repo is empty.  That'd fix the anaconda case.

This isn't conceptually a regression though I think.

cgwalters: (If you look at an earlier version of this patch, I did exactly what you said but I realized that'd make it much more likely to trigger the from-scratch logic)

jlebon:
OK, it took me a bit to fully grok the issue here. The commit msg made it seem like #710 was the fix for this. Which it *is*, but that wasn't the original intent of #710. Just to be clear, can you make this more explicit and reword the commit msg to something like:

```
This is because this host version doesn't yet have
ostreedev#710, which incidentally fixed this for remotes
that make use of summary files (like Fedora). Though the bug
was still present on remotes that do not.

Let's copy the logic very recently added in that commit to
handle the legacy case too.
```

?

cgwalters:
OK, 🛫 rebased 🛬 and reworded 🖊.

jlebon:
@rh-atomic-bot r+ c460332

Issue #688, opened 2017-02-16
Support ostree_diff_dirs taking a OstreeRepoDevInoCache


alexlarsson:
In flatpak, i'm using an ostree repo to cache the previous module. All check outs of the cache use hardlinks back to the cache, and after each new commit i check out the new files again so that I always have hardlinks for files that are in the cache.

The checkouts use a OstreeRepoDevInoCache which makes each commit fast, which is important if the previous stages are large, such as when building a runtime.

However, I also use the cache to detect what got installed in this module. I do this by diffing the latest cache commit with the install dir. This is pretty slow though, as the comparison of the outstanding files have to calculate the checksum for all the files, even if they are hardlinks into the repo. To fix this i took ostree_diff_dirs() and made a copy of it that takes a OstreeRepoDevInoCache argument, which allows it to quickly find the checksum for unchanged (hardlinked) files.

Unfortunately there is no public OstreeRepoDevInoCache method to look up a checksum, so I had to hack it up by stealing the OstreeRepoDevInoCache definithins from the private headers. (I'm assuming this isn't going to change really, and if it does i'll have to handle it.)

Allowing the use of the cache for the diff seems generically useful though, for instance "ostree diff" could use it if only one of the dirs is a commit.

cgwalters:
There's actually another outstanding PR for extending the diff API: https://github.com/ostreedev/ostree/pull/673

It'd make sense to consolidate.

cgwalters:
See https://github.com/ostreedev/ostree/pull/740

PR #740, opened 2017-03-16
diff: Add onwer-uid/gid arguments to diff command

ortogonal:
This makes it possible to use diff on local files were -owner-uid/gid
have been passed to commit, if not diff will say that all local files
are changed.

### DIFF
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -172,6 +172,7 @@ OstreeDiffItem
 ostree_diff_item_ref
 ostree_diff_item_unref
 ostree_diff_dirs
+ostree_diff_dirs_with_options
 ostree_diff_print
 <SUBSECTION Standard>
 ostree_diff_item_get_type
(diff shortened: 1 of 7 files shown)


cgwalters:
Offhand the code here looks good.

However, there's another API addition request we should roll into this one: https://github.com/ostreedev/ostree/issues/688

For future extensibility we might make a struct now (as is done with `ostree_repo_checkout_at()` and the `struct OstreeRepoCheckoutAtOptions`.

I can do some of this if you prefer.

ortogonal:
Off course you can. Honestly I didn't really understand #688 fully so your help is welcome.

ortogonal:
I looked a bit more at this. Are you thinking about adding a struct called something like OstreeDiffDirsOptions which, at first, only contains uid/gid?

cgwalters:
Thanks, that was pretty close.  We need to fill in the struct though with dummy fields so it's actually extensible later.

I also fixed some minor indentation issues.

Thanks for the patch!

@rh-atomic-bot r+ 15cded5

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #742, opened 2017-03-18
sysroot/deploy: Some cleanup to decl-after-stmt/return FALSE style

cgwalters:
And fd-relative. I also introduced some helpers here which I'll use later in
more invasive patches.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -702,88 +702,61 @@ selinux_relabel_dir (OstreeSysroot                 *sysroot,
 }
 
 static gboolean
-selinux_relabel_file (OstreeSysroot                 *sysroot,
-                      OstreeSePolicy                *sepolicy,
-                      GFile                         *path,
-                      const char                    *prefix,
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2052,103 +2022,93 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
 
+  gint new_deployserial;
   if (!allocate_deployserial (self, osname, revision, &new_deployserial,
                               cancellable, error))
-    goto out;
+    return FALSE;
 
-  new_deployment = ostree_deployment_new (0, osname, revision, new_deployserial,
-                                          new_bootcsum, -1);
+  g_autofree char *new_bootcsum = NULL;
```
jlebon: Nit (since the goal here is to declare near the first use): there's no point in declaring `new_bootcsum` here if we can just pass `NULL` to `ostree_deployment_new`. Would make more sense to declare it down below before we call `checksum_from_kernel_src()`.

jlebon:
@rh-atomic-bot r+ 54f5e42

PR #743, opened 2017-03-19
sysroot: Have writing non-default deployments go after booted

cgwalters:
When I added this API, I hadn't considered the case where we
have a "pending" deployment.  the `_NOT_DEFAULT` case worked
fine without that.

However for `rpm-ostree ex livefs`, we want to push a rollback target which is
the snapshot of the current deployment before mutating.


cgwalters:
Eh, I decided for now to copy this code into rpm-ostree.  Will decide a bit later whether to keep this patch or not.

cgwalters:
I'm going to revisit this later.  Closing to clear off the board.

PR #744, opened 2017-03-20
sysroot: Prep refactoring of cleanup logic

cgwalters:
For future work I'm going to tweak how we handle cleanup, and
the private cleanup flags didn't really end up being used - we
only specify "prune repo or not".  So fold that into a boolean for now.

The sysroot deploy logic then has a single "do_postclean" boolean, which is all
I want to expose as public API.


jlebon:
@rh-atomic-bot r+ 067c685

PR #747, opened 2017-03-21
Bump libglnx, port a few callers to new error API

cgwalters:
Testing out the waters here. I think we should roll this into any future code
cleanup reworking we do.

Depends: https://github.com/GNOME/libglnx/pull/38


cgwalters:
Now that the libglnx PR has landed 🛬, updated with that.  Should be good to go.

jlebon:
@rh-atomic-bot r+ 87bafcf

PR #749, opened 2017-03-21
two build fixes

cgwalters:


cgwalters:
After:
```
eu-readelf -a  ./.libs/libostree-1.so|grep NEEDED.*crypto
  NEEDED            Shared library: [libcrypto.so.10]
```

jlebon:
@rh-atomic-bot r+ 4832db2

PR #750, opened 2017-03-22
Add bare-user-only mode that works on non-xattrs filesystems

alexlarsson:
This mode is similar to bare-user, but does not store the permission,
ownership (uid/gid) and xattrs in an xattr on the file objects in the
repo. Additionally it stores symlinks as symlinks rather than as
regular files+xattrs, like the bare mode. The later is needed because
we can't store the is-symlink in the xattr.
    
This means that some metadata is lost, such as the uid. When reading a
repo like this we always report uid, gid as 0, and no xattrs, so
unless this is true in the commit the resulting repository will
not fsck correctly.
    
However, it the main usecase of the repository is to check out with
--user-mode, then no information is lost, and the repository can
work on filesystems without xattrs (such as tmpfs).

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -306,8 +306,7 @@ commit_loose_object_trusted (OstreeRepo        *self,
             goto out;
         }
 
-      if (objtype == OSTREE_OBJECT_TYPE_FILE && (self->mode == OSTREE_REPO_MODE_BARE ||
-                                                 self->mode == OSTREE_REPO_MODE_BARE_USER))
+      if (objtype == OSTREE_OBJECT_TYPE_FILE && _ostree_repo_mode_is_bare (self->mode))
         {
           /* To satisfy tools such as guile which compare mtimes
            * to determine whether or not source files need to be compiled,
@@ -700,7 +699,7 @@ write_object (OstreeRepo         *self,
        * binary with trailing garbage, creating a window on the local
        * system where a malicious setuid binary exists.
        */
-      if ((repo_mode == OSTREE_REPO_MODE_BARE || repo_mode == OSTREE_REPO_MODE_BARE_USER) && temp_file_is_regular)
+      if ((_ostree_repo_mode_is_bare (repo_mode)) && temp_file_is_regular)
         {
           guint64 size = g_file_info_get_size (file_info);
 
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -227,7 +227,11 @@ commit_loose_object_trusted (OstreeRepo        *self,
     }
 
   /* Special handling for symlinks in bare repositories */
-  if (object_is_symlink && self->mode == OSTREE_REPO_MODE_BARE)
+  if (object_is_symlink && self->mode == OSTREE_REPO_MODE_BARE_USER_ONLY)
+    {
+      /* We don't store the metadata in bare-user-only, so we're done. */
+    }
+  else if (object_is_symlink && self->mode == OSTREE_REPO_MODE_BARE)
     {
       /* Now that we know the checksum is valid, apply uid/gid, mode bits,
        * and extended attributes.
@@ -283,7 +287,8 @@ commit_loose_object_trusted (OstreeRepo        *self,
             }
         }
 
-      if (objtype == OSTREE_OBJECT_TYPE_FILE && self->mode == OSTREE_REPO_MODE_BARE_USER)
+      if (objtype == OSTREE_OBJECT_TYPE_FILE &&
+          (self->mode == OSTREE_REPO_MODE_BARE_USER || self->mode == OSTREE_REPO_MODE_BARE_USER_ONLY))
         {
           if (!object_is_symlink)
             {
@@ -302,7 +307,8 @@ commit_loose_object_trusted (OstreeRepo        *self,
                 }
             }
 
-          if (!write_file_metadata_to_xattr (fd, uid, gid, mode, xattrs, error))
+          if (self->mode == OSTREE_REPO_MODE_BARE_USER &&
+              !write_file_metadata_to_xattr (fd, uid, gid, mode, xattrs, error))
             goto out;
         }
 
@@ -708,8 +714,10 @@ write_object (OstreeRepo         *self,
                                                              cancellable, error))
             goto out;
         }
-      else if (repo_mode == OSTREE_REPO_MODE_BARE && temp_file_is_symlink)
+      else if (_ostree_repo_mode_is_bare (repo_mode) && temp_file_is_symlink)
         {
+          /* Note: This will not be hit for bare-user mode because its converted to a
+             regular file and take the branch above */
           if (!_ostree_make_temporary_symlink_at (self->tmp_dir_fd,
                                                   g_file_info_get_symlink_target (file_info),
                                                   &temp_filename,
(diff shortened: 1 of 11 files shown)


alexlarsson:
I guess it needs some tests too

src/ostree/ot-builtin-commit.c
```diff
@@ -84,6 +85,7 @@ static GOptionEntry options[] = {
   { "add-detached-metadata-string", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_detached_metadata_strings, "Add a key/value pair to detached metadata", "KEY=VALUE" },
   { "owner-uid", 0, 0, G_OPTION_ARG_INT, &opt_owner_uid, "Set file ownership user id", "UID" },
   { "owner-gid", 0, 0, G_OPTION_ARG_INT, &opt_owner_gid, "Set file ownership group id", "GID" },
+  { "canonical-permissions", 0, 0, G_OPTION_ARG_NONE, &opt_canonical_permissions, "Canonicallize permissions in the same way bare-user does for hardlinked files", NULL },
```
cgwalters: s/Canonicallize/Canonicalize

src/ostree/ot-builtin-commit.c
```diff
@@ -196,6 +198,13 @@ commit_filter (OstreeRepo         *self,
       g_hash_table_remove (mode_adds, path);
     }
 
+  if (opt_canonical_permissions &&
+      g_file_info_get_file_type (file_info) == G_FILE_TYPE_REGULAR)
```
cgwalters: Why isn't this just the default if writing to `bare-user-only` though?

cgwalters:
This all looks pretty good to me!  One comment about the perms on commit; feels like if we do that it'd reduce the diff in the tests a lot too.

I also wonder if we should basically force on `-U` for checking out in this mode too?

src/ostree/ot-builtin-commit.c
alexlarsson: It could be, but we might want to commit this to an archive-z2 repo and then pull it into a bare-user-only one.

alexlarsson:
Force -U, or fail without it?

src/ostree/ot-builtin-commit.c
cgwalters: Yep, I understand the rationale for the option, but I'd still say it should be on by default for bare-user.  Among other things, every API caller will need to duplicate this logic if it's not.

alexlarsson:
Ah, i see what you mean, if its automatic we get less testcase noise.

cgwalters:
I'd vote force `-U`.  

And with both of these things, what I'm actually arguing is we do the defaulting not at the command line level, but at the libostree level.  For checkout this should be pretty easy, check the mode in `ostree_repo_checkout_at()` and override.

For commit it's likely to be trickier in the general case because we have `ostree_repo_write_content()` - we'd have to crack it open and mutate it.  We might be able to get away without doing that for now if we just change the bit in `write_directory_content_to_mtree_internal()` where we pull from the filesystem.

cgwalters:
To put this another way - I'd say we've succeeded if flatpak doesn't need any code changes to work with this new repo mode.  Right?

alexlarsson:
Well, some of the ostree test cases commit to archive-z2 repos, which are then pulled to bare-user-only, so forcing it on commit to bare-user-only is not enough.

cgwalters:
Right, I agree we still need the canonical permissions argument.  Along the API lines, what about making it `OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS` ?

alexlarsson:
flatpak doesn't technically *need* any changes (other than making repos of this type). However, it would be nice if flatpak build-export creates commits that are canonical so that fsck works on the flatpak local repo. However, build-export generally exports to archive-z2 repos, so any change we make to that will not affect much. The amount of changes needed are very small though.

src/libostree/ostree-repo-commit.c
```diff
@@ -2237,14 +2237,24 @@ _ostree_repo_commit_modifier_apply (OstreeRepo               *self,
...
     }
 
   modified_info = g_file_info_dup (file_info);
   result = modifier->filter (self, path, modified_info, modifier->user_data);
+
+  if ((modifier->flags & OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS) != 0 &&
+      g_file_info_get_file_type (file_info) == G_FILE_TYPE_REGULAR)
+    {
+      guint current_mode = g_file_info_get_attribute_uint32 (modified_info, "unix::mode");
+      g_file_info_set_attribute_uint32 (modified_info, "unix::mode", current_mode | 0744);
```
cgwalters: Don't we also need to set uid/gid to zero?

alexlarsson: Its already possible to do this with the separate uid/gid options, so i made it only do what we couldn't already do.

alexlarsson: We *could* make this automatically mean uid/gid 0 and no-xatts too, but that means you can't do the indiviual operations separately

cgwalters: If you're committing to a bare-user-only repo, those are going to be discarded.

I can't think of a scenario where one is committing to an archive repo and wants to use this flag, but *doesn't* also want the uid/gid/no-xattrs.

I get what you're saying about everything else being achievable, but the flip side is - if we're not aware of a use case, why add it?

(A small reason to go the flag = full canon route is that at some point I'd like to optimize commit so we don't `malloc()` like there's no tomorrow, and part of that is going to be to stop allocating `GFileInfo` in favor of a `struct stat` on the stack.  Which we can only do right now if we don't have a commit modifier)

cgwalters: That said this is a "lean against" feeling - if you really prefer it the way it is now, given flatpak would be the primary user that's OK by me.

alexlarsson:
I made --canonical-permissions imply everything.

alexlarsson:
I'm gonna skip modifying the commit deep internals to do this automatically for bare-user-only repos though. There is to much code that has to be modified for that, and its not the common usecase i imagine anyway.

cgwalters:
Re changing the commit core, agree that's probably not worth it right now.  I have a little concern we're going to hit corner cases without doing it, but OTOH if the tests pass, I feel fairly confident.

Maybe though when this shows up in the next release we mark it as experimental, until flatpak has been using it for a bit in its devel stream?

@rh-atomic-bot r+ a84ec48

cgwalters:
Hm, actually it looks like some of these test failures are not flakes, e.g.:

```
error: lsetxattrOperation not supported
```

alexlarsson:
I though that was some generic ci issue. Does other PRs not get it?

cgwalters:
With our Homu instance, every test there except Travis has to pass to be pushed.

alexlarsson:
But how my changes cause enosupp?

cgwalters:
Not sure yet; built your branch and `make check TESTS=tests/test-basic.sh` passes here.

alexlarsson:
All the tests work here, but its not only test-basic thats failing with ENOSUPP,  even tests/test-sysroot-c

alexlarsson:
I applied the fixups and rebased to master, fixing the (trivial) conflicts.

alexlarsson:
So, i really want to start using this for the flatpak runtime builds, so i'm ack:ing this again (since colin already ack:ed it above). Hope that is ok.
@rh-atomic-bot r+ 	ae279d0

cgwalters:
@rh-atomic-bot r+ ae279d0

PR #753, opened 2017-03-23
core: Convert some functions to new code style

cgwalters:
Not doing the whole file since it's big, these are some simple ones to keep
momentum up.

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -121,13 +121,8 @@ ostree_parse_refspec (const char   *refspec,
                       char        **out_ref,
                       GError      **error)
 {
-  gboolean ret = FALSE;
-  GMatchInfo *match = NULL;
-  g_autofree char *remote = NULL;
-
-  static gsize regex_initialized;
   static GRegex *regex;
-
+  static gsize regex_initialized;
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-core.c
```diff
@@ -830,22 +790,18 @@ ostree_checksum_file_from_input (GFileInfo        *file_info,
...
           if (!ot_gio_splice_update_checksum (NULL, in, checksum, cancellable, error))
-            goto out;
+            return FALSE;
         }
     }
 
-  ret_csum = ot_csum_from_gchecksum (checksum);
-
-  ret = TRUE;
+  g_autofree guchar *ret_csum = ot_csum_from_gchecksum (checksum);
```
jlebon: We can just assign it directly to the out var at this point, right?

cgwalters: In this case, yes.  As a general rule we have to be a bit careful since some functions have nullable out arguments.

jlebon:
@rh-atomic-bot r+ 57b6719

alexlarsson:
Check out FlatpakRepoTransaction inhttps://github.com/flatpak/flatpak/blob/master/common/flatpak-utils.h too. Very useful autoptr thing for ostree commits.

PR #754, opened 2017-03-23
Add Coccinelle 🐞 usage: one for blacklisting, one for patch collection

cgwalters:
This is inspired by the [Coccinelle](http://coccinelle.lip6.fr/) usage
in systemd.  I also took it a bit further and added infrastructure
to have spatches which should never apply.  This acts as a blacklist.

The reason to do the latter is that coccinelle is *way* more powerful than the
regular expresssions we have in `make syntax-check`.

I started with blacklisting `g_error_free()` directly. The reason that's bad is
it leaves a dangling pointer.

### DIFF
--- a/coccinelle/glnx-errors.cocci
+++ b/coccinelle/glnx-errors.cocci
@@ -1,0 +1,6 @@
+@@
+expression p;
+@@
+- glnx_set_error_from_errno (p);
+- goto out;
++ return  glnx_throw_errno (p);
--- a/coccinelle/glnx-errors.cocci
+++ b/coccinelle/glnx-errors.cocci
@@ -1,6 +1,0 @@
-@@
-expression p;
-@@
-- glnx_set_error_from_errno (p);
-- goto out;
-+ return  glnx_throw_errno (p);
--- a/tests/test-rollsum-cli.c
+++ b/tests/test-rollsum-cli.c
@@ -22,6 +22,8 @@
 
 #include "ostree-rollsum.h"
 
+#include "libglnx.h"
+
 int
 main (int argc, char **argv)
 {
--- a/tests/test-rollsum-cli.c
+++ b/tests/test-rollsum-cli.c
@@ -25,7 +25,7 @@
 int
 main (int argc, char **argv)
 {
-  GError *local_error = NULL;
+  g_autoptr(GError) local_error = NULL;
   GError **error = &local_error;
   GBytes *from_bytes = NULL;
   GBytes *to_bytes = NULL;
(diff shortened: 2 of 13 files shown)


coccinelle/glnx-errors.cocci
```diff
@@ -0,0 +1,6 @@
+@@
+expression p;
+@@
+- glnx_set_error_from_errno (p);
+- goto out;
++ return  glnx_throw_errno (p);
```
cgwalters: I haven't yet figured out enough Coccinelle to know if we can really do the "style rewrite" as a spatch, but this is a start.

tests/test-rollsum-cli.c
```diff
@@ -25,7 +25,7 @@
 int
 main (int argc, char **argv)
 {
-  GError *local_error = NULL;
+  g_autoptr(GError) local_error = NULL;
```
jlebon: Looks like some of the Travis runners need the glnx backports:

```
Selecting previously unselected package libglib2.0-0:amd64.
Preparing to unpack .../libglib2.0-0_2.42.1-1+b1_amd64.deb ...
Unpacking libglib2.0-0:amd64 (2.42.1-1+b1) ...
...
<snip>
...
../tests/test-rollsum-cli.c: In function ‘main’:
../tests/test-rollsum-cli.c:28:3: error: implicit declaration of function ‘g_autoptr’ [-Werror=implicit-function-declaration]
   g_autoptr(GError) local_error = NULL;
   ^
```

coccinelle/glnx-errors.cocci
cgwalters: Hm?  This is *semantic* patch, it applies to everything that matches the pattern.

cgwalters: (What that spatch *won't* do is rewrite `gboolean ret = FALSE` and the `out` label etc.  That may be possible, but I'd have to learn more spatch)

jlebon: What I mean is, this specific patch has no practical use for now, right? (Exactly because of what we mentioned above). I.e. this is something that we'd need to work on before we can apply it on the whole codebase.

cgwalters: Right, yep.  Though see ⬇️ - I pushed a rewrite of this which covers a lot more.  Here's [the diff it generates](https://paste.fedoraproject.org/paste/l6JxtC6y82HxCKx-NY8AKF5M1UNdIGYhyRLivL9gydE=):  `37 files changed, 461 insertions(+), 869 deletions(-)`

cgwalters: One problem obviously is that coccinelle seems to hardcode linux kernel style (I didn't find an option to change it), and in general doesn't try to preserve "pretty" indentation.

So I'm not actually sure we should use this one, but it demos what we *could* do.

jlebon:
Ahh OK, makes sense!
@rh-atomic-bot r+ d43c609

PR #755, opened 2017-03-23
build: Dist ostree-sepolicy-private.h

cgwalters:
Should fix the Travis builds which actually generate a legacy tarball via
Automake.

### DIFF
--- a/Makefile-libostree-defines.am
+++ b/Makefile-libostree-defines.am
@@ -31,6 +31,7 @@ libostree_public_headers = \
 	src/libostree/ostree-diff.h \
 	src/libostree/ostree-gpg-verify-result.h \
 	src/libostree/ostree-sepolicy.h \
+	src/libostree/ostree-sepolicy-private.h \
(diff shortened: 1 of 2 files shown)


jlebon:
Don't mind me, just investigating the CI flakes!
@rh-atomic-bot try

Makefile-libostree-defines.am
```diff
@@ -31,6 +31,7 @@ libostree_public_headers = \
 	src/libostree/ostree-diff.h \
 	src/libostree/ostree-gpg-verify-result.h \
 	src/libostree/ostree-sepolicy.h \
+	src/libostree/ostree-sepolicy-private.h \
```
jlebon: But we don't want this in `/usr/include` though, right?

cgwalters: Ooh yes, right you are!  👍

jlebon:
@rh-atomic-bot r+ 6f7f6a8

cgwalters:
@rh-atomic-bot retry

PR #756, opened 2017-03-23
build: Expose autocleanups ⎚ unconditionally, start using them

cgwalters:
I'd like to do this in rpm-ostree at least. Originally I was looking at porting
to `G_DECLARE_FINAL_TYPE` but eh, this is easier for now and won't bump our GLib
dependency which might matter for our embedded users.

For now I just did a few replacements in the `remote` command line. A full port
can come as we do other code cleanups.

This will actually break the flatpak build right now, but
that's easy to fix.  And we concluded in e.g.
https://bugs.freedesktop.org/show_bug.cgi?id=95065#c5
it's a bug for downstream projects to do that.

### DIFF
--- a/src/libostree/ostree-autocleanups.h
+++ b/src/libostree/ostree-autocleanups.h
@@ -26,14 +26,10 @@
 
 G_BEGIN_DECLS
 
-#ifndef OSTREE_WITH_AUTOCLEANUPS
-#define OSTREE_WITH_AUTOCLEANUPS 0
-#endif
-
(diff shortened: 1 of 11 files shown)


jlebon:
OK, makes sense. So to confirm, this will only break projects that currently compile without `OSTREE_WITH_AUTOCLEANUPS` and instead define the cleanup functions themselves right?

cgwalters:
Yep, flatpak patch in https://github.com/flatpak/flatpak/pull/667

jlebon:
@rh-atomic-bot r+ 4e94600

Issue #758, opened 2017-03-24
rework CI to work in overlayfs2 - security. xattr namespace handling as non-root


cgwalters:
The rhCI env was upgraded to F25AH/overlay2, and this breaks our test suite (specifically the latter).  I think we're picking up the selinux attrs, but overlay2 isn't letting us write them back.

So...this gets into a deep question - how valuable is it for us to pick up security attributes when running as non-root?  (And as root in Docker?)

We already today have `--no-xattrs` for commit...should our unit tests simply pass that unconditionally?  It'd be a fair amount of churn, so maybe we add an environment variable?

cgwalters:
It's useful to note though, that outside of overlayfs2, it does work to use ostree as non-root and capture SELinux labels.  Theoretically, one could use it for a backup system.  That said...it's theoretical.

I think the right thing to do now is bandaid this specific situation.  Working on that.

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/759

PR #759, opened 2017-03-24
Overlay2 ci

cgwalters:
(First commit was useful for me to debug)

There are a lot of things suboptimal about this approach, but
on the other hand we need to get our CI back up and running.

The basic approach is to - in the test suite, detect if we're on overlayfs. If
so, set a flag in the repo, which gets picked up by a few strategic places in
the core to turn on "ignore xattrs".

The core problem here is while overlayfs will let us read and
see the SELinux labels, it won't let us write them.

Down the line, we should improve this so that we can selectively ignore e.g.
`security.*` attributes but not `user.*` say.

Closes: #758

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -86,6 +86,17 @@ chmod -R u+w "${test_tmpdir}"
 export TEST_GPG_KEYHOME=${test_tmpdir}/gpghome
 export OSTREE_GPG_HOME=${test_tmpdir}/gpghome/trusted
 
+# See comment in ot-builtin-commit.c and https://github.com/ostreedev/ostree/issues/758
+echo evaluating for overlayfs...
+case $(stat -f --printf '%T' /) in
+    overlayfs)
+        echo "overlayfs found; enabling OSTREE_NO_XATTRS"
+        export OSTREE_SYSROOT_DEBUG="${OSTREE_SYSROOT_DEBUG},no-xattrs"
+        export OSTREE_NO_XATTRS=1 ;;
+    *) ;;
+esac
+echo done
+
 if test -n "${OT_TESTS_DEBUG:-}"; then
     set -x
 fi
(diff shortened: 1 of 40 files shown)


tests/libtest.sh
```diff
@@ -232,6 +240,16 @@ setup_test_repository () {
...
 }
 
+# A wrapper which also possibly disables xattrs for CI testing
+ostree_repo_init() {
+    repo=$1
+    shift
+    ${CMD_PREFIX} ostree --repo=${repo} init "$@"
+    if test -n "${OSTREE_NO_XATTRS:-}"; then
+        echo -e 'disable-xattrs=true\n' >> ${repo}/config
+    fi
```
cgwalters: This bridges from the test environment to the repo.

tests/libtest.sh
```diff
@@ -86,6 +86,16 @@ chmod -R u+w "${test_tmpdir}"
 export TEST_GPG_KEYHOME=${test_tmpdir}/gpghome
 export OSTREE_GPG_HOME=${test_tmpdir}/gpghome/trusted
 
+# See comment in ot-builtin-commit.c and https://github.com/ostreedev/ostree/issues/758
+echo evaluating for overlayfs...
+case $(stat -f --printf '%T' /) in
+    overlayfs)
+        echo "overlayfs found; enabling OSTREE_NO_XATTRS"
+        export OSTREE_NO_XATTRS=1 ;;
```
cgwalters: This part detects overlayfs.

jlebon:
Looks sane overall. One concern with this approach though is that we lose coverage of those code paths we're avoiding. But anyway, we don't actually have any SELinux tests right now. It's probably worth adding another test suite that runs in a separate non-overlay environment.

Now checking what's going on with the "Red Hat CI" suite.

jlebon:
Ahh shoot, this is a regression from https://github.com/jlebon/redhat-ci/commit/c7cf12cf6a737f4c969119f6e50033f7e1f4408f. Fixing.

jlebon:
bot, retest this please

cgwalters:
Okay, for test-sysroot-c we run the test script as a subprocess and won't pick up the new env.  But I am confused how it is working locally...will debug.

`tests/test-xattrs.sh` is just going to need to be skipped.

cgwalters:
> error: Writing object 78027900fac65350d8ea12625530b16471db6edf287d81b36a318d21782f3176.file: No space left on device

I did this for now on the node: `lvm lvextend atomicos/root -L +100G -r` until I work out where the provisioning is.

cgwalters:
I think this is good to go, we're just on flakes now.

jlebon:
Nice! Yeah, I'll codify all the config changes we did on the node next week.
@rh-atomic-bot r+ 03d0150

PR #760, opened 2017-03-27
Define TARGET_PREFIX to use with grub2 deployment

OYTIS:
Hardcoded path to ostree-grub-generator breaks workwlow with prefixes other that /usr.


Makefile.am
```diff
@@ -48,6 +48,8 @@ endif
 
 EXTRA_DIST += autogen.sh COPYING README.md
 
+DEFS += -DTARGET_PREFIX="\"$(prefix)\""
```
cgwalters: Huh, I'd never heard of `DEFS` before.  Let's instead add to `AM_CPPFLAGS`, since we use that above this for very similar things like `-DDATADIR='"$(datadir)"'`.

Makefile.am
```diff
@@ -24,7 +24,7 @@ OSTREE_GITREV=$(shell if command -v git >/dev/null 2>&1 && test -d $(srcdir)/.gi
 ACLOCAL_AMFLAGS = -I buildutil -I libglnx ${ACLOCAL_FLAGS}
 AM_CPPFLAGS += -DDATADIR='"$(datadir)"' -DLIBEXECDIR='"$(libexecdir)"' \
 	-DLOCALEDIR=\"$(datadir)/locale\" -DSYSCONFDIR=\"$(sysconfdir)\" \
-	-DSHORTENED_SYSCONFDIR=\"$(shortened_sysconfdir)\" \
+	-DTARGET_PREFIX='"$(prefix)' -DSHORTENED_SYSCONFDIR=\"$(shortened_sysconfdir)\" \
```
OYTIS: Closing parentheses is probably missing. I'll check if it works for me.

cgwalters: Yep, pushed another fixup ⬇️ - let me know if it works for you.

OYTIS:
@cgwalters Works fine, thank you.

cgwalters:
@rh-atomic-bot r+  	b56e8b3

PR #763, opened 2017-03-27
 Fix unlock temp dir label mounted over /usr 

rhatdan:
When an admin executes `ostree admin unlock`, the command creates
an overlay director in /var/tmp that is later mounted on /usr.  The
problem on an SELinux system you will end up with /usr being labeled
system_u:object_r:tmp_t:s0, when the original label was
system_u:object_r:usr_t:s0.  This is causing errors with atomic scan
on unlocked systems.  And could potentially cause other SELinux issues.

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -26,6 +26,7 @@
 #include <sys/wait.h>
 
 #include "ostree-core-private.h"
+#include "ostree-sepolicy-private.h"
 #include "ostree-sysroot-private.h"
 #include "ostree-deployment-private.h"
 #include "ostree-bootloader-uboot.h"
@@ -1691,6 +1692,7 @@ ostree_sysroot_deployment_unlock (OstreeSysroot     *self,
                                   GError           **error)
 {
   gboolean ret = FALSE;
+  glnx_unref_object OstreeSePolicy *sepolicy = NULL;
   OstreeDeploymentUnlockedState current_unlocked =
     ostree_deployment_get_unlocked (deployment); 
   glnx_unref_object OstreeDeployment *deployment_clone =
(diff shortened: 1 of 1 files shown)


rhatdan:
@miabbott PTAL

rhatdan:
Should fix https://github.com/projectatomic/atomic-host-tests/pull/119

rhatdan:
@miabbott I have not tested but this should fix the issue you are seeing.

cgwalters:
In ostree we have an (IMO) nice wrapper for libselinux context stuff; I pushed a rewritten version to use it above ⬆️.  Testing now.

cgwalters:
Ooh, in testing `self->sepolicy` was a trap.  This new version works for me.

@rh-atomic-bot delegate=rhatdan

src/libostree/ostree-sysroot.c
```diff
@@ -1765,8 +1771,16 @@ ostree_sysroot_deployment_unlock (OstreeSysroot     *self,
...
+        /* Ensure that the directory is created with the same label as `/usr` */
+        { g_auto(OstreeSepolicyFsCreatecon) con = { 0, };
+
+          if (!_ostree_sepolicy_preparefscreatecon (&con, sepolicy,
+                                                    "/usr", 0755, error))
+            goto out;
+
+          if (!glnx_mkdtempat (AT_FDCWD, development_ovldir, 0755, error))
+            goto out;
+        }
```
rhatdan: Should you set the fscreatecon back to default, in csae ostree creates any other files/directories?

cgwalters: That's exactly what the ` { g_auto(OstreeSepolicyFsCreatecon) con = { 0, };` bit does - we make extensive use of the [GCC cleanup attribute](https://blog.verbum.org/2012/05/09/__attribute__-cleanup-or-how-i-came-to-love-c-again/).

rhatdan: Excellent, Thanks.

rhatdan:
@rh-atomic-bot r+ ab1ac9d

miabbott:
Worked for me! 

See details here:  https://github.com/projectatomic/atomic/issues/951#issuecomment-289796585

PR #764, opened 2017-03-28
sysroot: Don't cache sepolicy

cgwalters:
In [this commit](https://github.com/ostreedev/ostree/commit/6ce80f9685e6273dbcb4731d6650a10976ea220a)
for some reason I added a `sepolicy` member to the sysroot.  I
have no idea why I did that, and it's conceptually wrong
since the policy is specific to a *deployment*.

This bit me when I was working on [a pull request](https://github.com/ostreedev/ostree/pull/763)
elsewhere, since at that point it was `NULL`.

We already pass around the sepolicy in the deployment code, so just stop caching
it.


jlebon:
@rh-atomic-bot r+ d4adb13

PR #765, opened 2017-03-28
ci: Enable -Werror=unused-result with -Wp,-D_FORTIFY_SOURCE=2

cgwalters:
(Also rename the other CI contexts to be more consistent)

We pass this right now. I just noticed an instance of this in bwrap, and I think
we should be trying to match the RPM build baseline.


jlebon:
@rh-atomic-bot r+ 47824e1

PR #766, opened 2017-03-28
lib: Fix OSTREE_CHECK_VERSION()

cgwalters:
Actually trying to use this in rpm-ostree, it kept returning successfully when I
didn't expect it to... The first conditional was always succeeding even when I
was asking for a newer minor.


jlebon:
Ouch! Thankfully we didn't create a release with this yet.
@rh-atomic-bot r+ 161c1c2

PR #768, opened 2017-03-29
sepolicy/commit: Make failure to find label fatal, add hack for /proc

cgwalters:
I was working on `rpm-ostree livefs` which does some ostree-based
filesystem diffs, and noticed that we were ending up with `/proc`
not being labeled in our base trees.

Reading the selinux-policy source, indeed we have:

```
/proc			-d	<<none>>
/proc/.*			<<none>>
```

This dates pretty far back.  We really don't want unlabeled
content in ostree.  In this case it's mostly OK since the kernel
will assign a label, but again *everything* should be labeled via
OSTree so that it's all consistent, which will fix `ostree diff`.

Notably, `/proc` is the *only* file path that isn't covered when composing a
Fedora Atomic Host. So I added a hack here to hardcode it (although I'm a bit
uncertain about whether it should really be `proc_t` on disk before systemd
mounts or not).


cgwalters:
http://marc.info/?l=selinux&m=149082134430052&w=2

cgwalters:
Removing WIP, though note this rolls up https://github.com/ostreedev/ostree/pull/769 so let's land that first.

cgwalters:
🚀 Rebased - note we landed the rpm-ostree PR which uses this.

jlebon:
@rh-atomic-bot r+ bdeebc1

jlebon:
Pending homu revival, let's just label approved PRs.

jlebon:
@rh-atomic-bot r+ bdeebc1

jlebon:
@rh-atomic-bot retry

PR #769, opened 2017-03-30
sepolicy: Fix regressions from introduction of sepolicy_new_at()

cgwalters:
Being bitten by lack of PR testing here.  There are two bugs:

- First and foremost, I forgot that GObject will call the property setters with
  the defaults.  This meant we were getting both path="/var/tmp/blah" and fd=-1,
  and we were accepting -1 as a fd, which then got converted into AT_FDCWD
  which was wrong.
- Since these properties are construct only and mutually exclusive, don't
  try to handle one resetting the other.  Assert that exactly one of them is set.

### DIFF
--- a/src/libostree/ostree-sepolicy.c
+++ b/src/libostree/ostree-sepolicy.c
@@ -112,15 +112,18 @@ ostree_sepolicy_set_property(GObject         *object,
           {
             /* Canonicalize */
             self->path = g_file_new_for_path (gs_file_get_path_cached (path));
+            g_assert_cmpint (self->rootfs_dfd, ==, -1);
           }
-        self->rootfs_dfd = -1;
       }
       break;
     case PROP_ROOTFS_DFD:
       {
-        self->rootfs_dfd = g_value_get_int (value);
-        g_clear_object (&self->path);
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sepolicy.c
```diff
@@ -300,6 +303,12 @@ initable_init (GInitable     *initable,
         goto out;
       self->rootfs_dfd = self->rootfs_dfd_owned;
     }
+  else if (!self->path)
```
jlebon: Ahh, OK that makes sense. It took me a bit to grok the logic here because they're just checking for the other one. Since we're sure now from the property setter that we have a proper XOR, can we rewrite these as:

```c
if (self->path)
  {
    ...
  }
else if (self->rootfs_dfd != -1)
  {
    ...
  }
else
  g_assert_not_reached ();
```

?

cgwalters: Not quite, because if we only have a rootfs dfd, we need to generate `self->path`.  I reworked the code ⬇️ though to `#if 0` out the bit we weren't using and changed the evaluation order.

jlebon:
@rh-atomic-bot r+ c915719

PR #693, opened 2017-02-17
Libglnx rebump + UBSAN

cgwalters:

PR #770, opened 2017-03-30
ci: Add a check that submodule changes include "Update submodule: "

cgwalters:
To prevent repeats of https://github.com/ostreedev/ostree/pull/693

I tested this script in https://github.com/cgwalters/playground/pull/48

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -11,6 +11,7 @@ container:
 
 packages:
   - libasan
+  - git
   - coccinelle
 
 env:
@@ -27,6 +28,7 @@ build:
 
 tests:
     - make syntax-check
+    - ./tests/ci-commitmessage-submodules.sh
     - make check
     - gnome-desktop-testing-runner -p 0 ostree
 
--- a/tests/ci-commitmessage-submodules.sh
+++ b/tests/ci-commitmessage-submodules.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+set -euo pipefail
+# Copyright 2017 Colin Walters <walters@verbum.org>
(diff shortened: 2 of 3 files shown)


.redhat-ci.yml
```diff
@@ -11,6 +11,7 @@ container:
 
 packages:
   - libasan
+  - git
```
jlebon: (And actually, we should be able to drop the other two -- we should try to keep it empty if it's already in the image, so that we don't even have to `makecache`)

tests/ci-commitmessage-submodules.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+set -euo pipefail
+
+BUMP_SUBMODULE_REGEXP="Update submodule: "
```
jlebon: This is a cool idea! Though we can make it even smarter/stronger 💪 , esp. to deal with multiple submodules (as we have here and in rpm-ostree). E.g.:

- Get the list of submodule subdirs present from `git submodule status` (ouhh actually there's an even nicer way: `git submodule foreach --quiet 'echo $path'`)
- Use e.g. `git diff --name-only` to check if a commit modifies a submodule path
- Grep for `^Update submodule: $submodule` in `git show -s --format=%B`

cgwalters: I agree that all sounds better...this is another case where I'm not totally sure it's was quite worth the time to polish, but I did anyways.  At least it's more likely to be reusable by others.  Update ⬇️

cgwalters:
bot, retest this please

jlebon:
Nice!

jlebon:
@rh-atomic-bot r+ b8709ac

PR #774, opened 2017-04-03
soup: Hold a ref to the pending URI during completion processing

cgwalters:
It was reported that in the range request handling, we called `remove_pending()`
twice, and this could be viewed as a use-after-free. Currently though, the task
object already holds a strong reference, so I observed the refcount was 2, hence
there is no use-after-free in practice.

Reported-By: "Siddharth Sharma" <siddharth@redhat.com>

### DIFF
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -1053,6 +1053,7 @@ on_request_sent (GObject        *object,
 
   if (!pending->request_body)
     goto out;
+  g_assert_no_error (local_error);
 
   if (SOUP_IS_REQUEST_HTTP (object))
     {
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -1040,13 +1040,12 @@ on_request_sent (GObject        *object,
                  gpointer        user_data) 
 {
   GTask *task = G_TASK (user_data);
-  OstreeFetcherPendingURI *pending;
-  GCancellable *cancellable;
+  /* Hold a ref to the pending across this function, since we remove
+   * it from the hash early in some cases, not in others. */
+  OstreeFetcherPendingURI *pending = pending_uri_ref (g_task_get_task_data (task));
+  GCancellable *cancellable = g_task_get_cancellable (task);
   GError *local_error = NULL;
   glnx_unref_object SoupMessage *msg = NULL;
-
-  pending = g_task_get_task_data (task);
-  cancellable = g_task_get_cancellable (task);
 
   pending->state = OSTREE_FETCHER_STATE_COMPLETE;
   pending->request_body = soup_request_send_finish ((SoupRequest*) object,
@@ -1054,7 +1053,7 @@ on_request_sent (GObject        *object,
 
   if (!pending->request_body)
     goto out;
-  
+
   if (SOUP_IS_REQUEST_HTTP (object))
     {
       msg = soup_request_http_get_message ((SoupRequestHTTP*) object);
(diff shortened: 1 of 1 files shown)


cgwalters:
Tested by applying this diff to ostree:
```
diff --git a/src/libostree/ostree-fetcher-soup.c b/src/libostree/ostree-fetcher-soup.c
index b130b48..561492d 100644
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -1067,6 +1067,7 @@ on_request_sent (GObject        *object,
           g_task_return_pointer (task,
                                  g_strdup (pending->out_tmpfile),
                                  (GDestroyNotify) g_free);
+          fprintf (stderr, "RR: Removing pending %p, refcnt %u\n", pending, pending->ref_count);
           remove_pending (pending);
           goto out;
         }
@@ -1180,6 +1181,7 @@ on_request_sent (GObject        *object,
       if (pending->request_body)
         (void) g_input_stream_close (pending->request_body, NULL, NULL);
       g_task_return_error (task, local_error);
+      fprintf (stderr, "ERR: Removing pending %p, refcnt %u\n", pending, pending->ref_count);
       remove_pending (pending);
     }
 
diff --git a/tests/test-pull-resume.sh b/tests/test-pull-resume.sh
index 51746b9..86070c7 100755
--- a/tests/test-pull-resume.sh
+++ b/tests/test-pull-resume.sh
@@ -42,6 +42,7 @@ do
   if ${CMD_PREFIX} ostree --repo=repo pull origin main 2>err.log; then
     break
   fi
+  cat err.log
   assert_file_has_content err.log 'error:.*\(Download incomplete\)\|\(Transferred a partial file\)'
 done
 if ${CMD_PREFIX} ostree --repo=repo fsck; then
```
Then, I ran this script:

```
$ while make check TESTS=tests/test-pull-resume.sh >/dev/null; do grep '^RR:' ./tests/test-pull-resume.sh.log; done
RR: Removing pending 0x11e6510, refcnt 2
RR: Removing pending 0x13d2b60, refcnt 2
RR: Removing pending 0xd34290, refcnt 2
RR: Removing pending 0x21764c0, refcnt 2
RR: Removing pending 0x214db70, refcnt 2
```

(I knew the pull-resume.sh test would trigger the path in question, though not every run)

src/libostree/ostree-fetcher-soup.c
```diff
@@ -1040,21 +1040,20 @@ on_request_sent (GObject        *object,
...
   glnx_unref_object SoupMessage *msg = NULL;
 
-  pending = g_task_get_task_data (task);
-  cancellable = g_task_get_cancellable (task);
-
   pending->state = OSTREE_FETCHER_STATE_COMPLETE;
   pending->request_body = soup_request_send_finish ((SoupRequest*) object,
                                                    result, &local_error);
 
   if (!pending->request_body)
```
jlebon: Hmm, just trying to understand. Under what circumstances do we call `remove_pending()` twice? The one in the `out` path is only executed if there was an error, right? And we check for that here by checking for the body. (Or does libsoup also set `local_error` for `REQUESTED_RANGE_NOT_SATISFIABLE`?)

cgwalters: That's absolutely right, and another reason this code path can't be reached.    I amended 🖊 the commit message 📓 below ⬇️

jlebon: Ahh OK, that makes sense now. So we don't actually need to add a ref, right? (Since as you mentioned the GTask already holds a ref to it). Though maybe a `g_assert (local_error == NULL)` would help it obvious they're exclusive.

cgwalters: OK, I added one ⬇️

jlebon:
@rh-atomic-bot r+ fc7f7da

PR #775, opened 2017-04-04
core: Fix default value of disable_xattrs

cgwalters:
Sigh.  Rather awful regression from https://github.com/ostreedev/ostree/pull/759


jlebon:
@rh-atomic-bot r+ 912926c

PR #779, opened 2017-04-05
checkout: Provide useful error with checkout -H and incompat mode

cgwalters:
Previously we'd assert and dump core if one used `checkout -H` without
`-U` on a bare-user repo, because we'd hit the bare-user symlink case.

Rework the code to handle this, and add tests. I hit this when I was going to
suggest to someone to use `-H` to ensure they were getting hardlinks.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -423,6 +423,7 @@ checkout_one_file_at (OstreeRepo                        *repo,
   gboolean ret = FALSE;
   const char *checksum;
   gboolean is_symlink;
+  gboolean is_bare_user_symlink = FALSE;
   gboolean can_cache;
   gboolean need_copy = TRUE;
   char loose_path_buf[_OSTREE_LOOSE_PATH_MAX];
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo-checkout.c
```diff
@@ -468,20 +467,38 @@ checkout_one_file_at (OstreeRepo                        *repo,
...
+          gboolean repo_is_usermode =
+            current_repo->mode == OSTREE_REPO_MODE_BARE_USER ||
+            current_repo->mode == OSTREE_REPO_MODE_BARE_USER_ONLY;
+          /* We're hardlinkable if the checkout mode matches the repo mode */
+          gboolean is_hardlinkable =
+            (current_repo->mode == OSTREE_REPO_MODE_BARE
+             && options->mode == OSTREE_REPO_CHECKOUT_MODE_NONE) ||
+            (repo_is_usermode && options->mode == OSTREE_REPO_CHECKOUT_MODE_USER);
+          /* NOTE: bare-user symlinks are not stored as symlinks */
+          gboolean is_bare_symlink = (repo_is_usermode && is_symlink);
```
jlebon: Should this be renamed `is_bare_user_symlink` ?

src/libostree/ostree-repo-checkout.c
```diff
@@ -596,7 +613,13 @@ checkout_one_file_at (OstreeRepo                        *repo,
   /* Fall back to copy if we couldn't hardlink */
   if (need_copy)
     {
-      g_assert (!options->no_copy_fallback);
+      /* Bare user mode can't hardlink symlinks, so we need to do a copy for
+       * those. (Although in the future we could hardlink inside checkouts) This
+       * assertion is intended to ensure that for regular files at least, we
+       * succeeded at hardlinking above.
+       */
+      if (!is_symlink)
```
jlebon: Would the intent here be better expressed as:

```c
if (options->no_copy_fallback)
  g_assert (is_bare_user_symlink);
```

?

cgwalters:
Fixup for comments ⬆️

jlebon:
@rh-atomic-bot r+ 054fd76

Issue #778, opened 2017-04-05
evaluate pipelining with http2/curl


cgwalters:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/JKSDVQ6NMWISZB7DX7UNBJSDSVZSWRAO/

>  I cannot see CURLOPT_PIPEWAIT or CURLMOPT_PIPELINING options in your code. They are needed if you want to take advantage of the core HTTP/2 features. Otherwise you can slightly benefit from the compressed HTTP headers but you are not using the HTTP/2 pipelining/multiplexing. I would suggest to have a look at the http2-download.c example: https://github.com/curl/curl/blob/curl-7_53_0/docs/examples/http2-download.c Kamil

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/780

PR #780, opened 2017-04-05
curl: Enable pipelining for HTTP/2

cgwalters:
Testing a fetch of `fedora-atomic/.../docker-host` from
an nginx instance over `https://127.0.0.1` using Fedora 25
versions.  Average over 3 runs:

Before: ~24.6 seconds
After: ~19 seconds

Speedup: ~30%

Closes: https://github.com/ostreedev/ostree/issues/778


jlebon:
Nice! @rh-atomic-bot r+ 7ae649f

Issue #783, opened 2017-04-07
pull via commit does not use static deltas 𝚫🛇


simao:
I am trying to pull some commits after generating a static delta:

``` 
  ostree --repo=small-bare --verbose remote summary origin
ostree.static-deltas: {'eaa75d75b9a868bb582e6660d90f0e666eecca8c6f00e16cd1f0176dabe88ca6-fcd3b9bfee0d609735da286657f11361bf85fecf9125c22e0d678ec07522b05f': <[byte 0xf0, 0x5c, 0x45, 0x42, 0x12, 0x65, 0x70, 0xb6, 0xf4, 0x7a, 0x83, 0x2c, 0xcb, 0xd2, 0x17, 0x0f, 0x22, 0x94, 0xf7, 0xc3, 0xcc, 0x43, 0x4b, 0x3b, 0xdd, 0xc1, 0xa8, 0x97, 0xe4, 0x2d, 0xd7, 0x9e]>}

 ostree --repo=small-bare --verbose pull --require-static-deltas --http-header='x-ats-namespace=simao' origin eaa75d75b9a868bb582e6660d90f0e666eecca8c6f00e16cd1f0176dabe88ca6
                                                                              
OT: starting fetch of eaa75d75b9a868bb582e6660d90f0e666eecca8c6f00e16cd1f0176dabe88ca6.commit (detached)
OT: fetch of eaa75d75b9a868bb582e6660d90f0e666eecca8c6f00e16cd1f0176dabe88ca6.commit (detached) complete
OT: starting fetch of eaa75d75b9a868bb582e6660d90f0e666eecca8c6f00e16cd1f0176dabe88ca6.commit
OT: fetch of eaa75d75b9a868bb582e6660d90f0e666eecca8c6f00e16cd1f0176dabe88ca6.commit complete
OT: write of eaa75d75b9a868bb582e6660d90f0e666eecca8c6f00e16cd1f0176dabe88ca6.commit complete
OT: starting fetch of eaa75d75b9a868bb582e6660d90f0e666eecca8c6f00e16cd1f0176dabe88ca6.commit (detached)
OT: starting fetch of 9ca9e91f41e271d51b8a47ac248652a40c501a9845610ed47e37077ac4aa2042.dirtree
OT: starting fetch of 794f9b8b314d4d57193fca37dab9149b075b598dadb39d33f77d6a815fca584c.dirmeta
[...]

PR #787, opened 2017-04-11
pull: Support deltas 𝚫 for explicit commits

cgwalters:
I think the majority of OSTree usage calls pull with refs, not
explicit commits.  We even added special "override syntax" with
`@` (e.g. `ostree pull foo@ab12c34`) as a hybrid.

However, some users may want to still pull explicit commits
for whatever reason.  The old static delta logic looked at
the previous commit of the ref.  However, in #710
we enhanced the logic to look at all local commits.

It's now a lot more natural to teach the delta logic
to support revisions, e.g. `ostree pull someorigin ab12c34`.

This also fixes the problem that before, `--require-static-deltas`
was completely ignored when processing revisions.

This is a nontrivial refactoring of the logic, but the end
result feels a lot more readable to me.

Closes: #783

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -68,6 +68,7 @@ typedef struct {
 
   gboolean          gpg_verify;
   gboolean          require_static_deltas;
+  gboolean          disable_static_deltas;
   gboolean          gpg_verify_summary;
   gboolean          has_tombstone_commits;
 
@@ -1879,10 +1880,15 @@ process_one_static_delta (OtPullData   *pull_data,
 /* Loop over the static delta data we got from the summary,
  * and find the newest commit for @out_from_revision that
  * goes to @to_revision.
+ *
+ * Additionally, @out_have_scratch_delta will be set to %TRUE
+ * if there is a %NULL → @to_revision delta, also known as
+ * a "from scratch" delta.
  */
 static gboolean
 get_best_static_delta_start_for (OtPullData *pull_data,
                                  const char *to_revision,
+                                 gboolean   *out_have_scratch_delta,
                                  char      **out_from_revision,
                                  GCancellable *cancellable,
                                  GError      **error)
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -2566,6 +2578,103 @@ reinitialize_fetcher (OtPullData *pull_data, const char *remote_name, GError **e
...
+
+              /* Was it partial?  Then we can't use it. */
+              if (commitstate_is_partial (pull_data, from_commitstate))
+                g_clear_pointer (&delta_from_revision, g_free);
+            }
+        }
+    }
+
+  /* If we found a delta, let's go */
+  if (delta_from_revision)
```
jlebon:
```
+ env LD_PRELOAD=/var/tmp/checkout/tests/libreaddir-rand.so ostree --repo=repo2 pull-local --require-static-deltas repo 37f59afd2f473636ecb96f434cfc5e50a52c7aacf66f6dd765876566b9ced467
error: Static deltas required, but none found for (null) to 37f59afd2f473636ecb96f434cfc5e50a52c7aacf66f6dd765876566b9ced467
ok heuristic endian detection
PASS: tests/test-delta.sh 4 heuristic endian detection
ERROR: tests/test-delta.sh - too few tests run (expected 12, got 4)
ERROR: tests/test-delta.sh - exited with status 1
```

I think this would fix the testsuite as well.

jlebon:
The refactor definitely makes things more legible overall.

src/libostree/ostree-repo-pull.c
cgwalters: I rewrote the code to make the logic clearer around when we'd do scratch deltas, but it turned out that wasn't the problem at all.

There wasn't actually a scratch delta at that point in the suite - we'd just been passing because of the previous semantics of `--require-static-deltas` being a noop.  Upon investigation, there was a scratch delta for `${origrev}`, so I used that.

jlebon:
Nice! @rh-atomic-bot r+ b4fece6

cgwalters:
@rh-atomic-bot r-

cgwalters:
@rh-atomic-bot force

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot r=jlebon

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot r+ b4fece6

PR #789, opened 2017-04-13
lib/core: Complete conversion to new code style

cgwalters:
No surprises here, all quite straightforward.


cgwalters:
(Note when I said "straightforward", of course it'd turn out my attempt to use `g_steal_pointer()` would backfire with a segfault since I didn't notice it was used twice in the call. This is something that Rust would have caught)

jlebon:
@rh-atomic-bot r+ cbf61ec

PR #790, opened 2017-04-13
lib/boot: Convert bootconfig parser to new code style

cgwalters:
This is a small one.


jlebon:
@rh-atomic-bot r+ 5c4d29e

PR #791, opened 2017-04-13
sysroot: Continue conversion of some simpler functions to new style

cgwalters:
This is only about 40%, and mostly simpler functions.  It's
nice to switch to `g_autoptr(GMatchInfo)` instead of our inline version.

I decided to add more usage of `ot_transfer_out_value()`, though it'd
be nice to try to have a copy of that in libglnx (or possibly glib).

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -284,47 +284,33 @@ ostree_sysroot_ensure_initialized (OstreeSysroot  *self,
                                    GCancellable   *cancellable,
                                    GError        **error)
 {
-  gboolean ret = FALSE;
-  struct stat stbuf;
-
   if (!ensure_sysroot_fd (self, error))
-    goto out;
+    return FALSE;
 
   if (!glnx_shutil_mkdir_p_at (self->sysroot_fd, "ostree/repo", 0755,
                                cancellable, error))
-    goto out;
+    return FALSE;
 
   if (!glnx_shutil_mkdir_p_at (self->sysroot_fd, "ostree/deploy", 0755,
                                cancellable, error))
-    goto out;
+    return FALSE;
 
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot.c
```diff
@@ -580,42 +515,29 @@ parse_origin (OstreeSysroot   *self,
...
+      g_autofree char *origin_contents =
+        glnx_file_get_contents_utf8_at (deployment_dfd, origin_path,
+                                        NULL, cancellable, error);
       if (!origin_contents)
-        goto out;
+        return FALSE;
 
       if (!g_key_file_load_from_data (ret_origin, origin_contents, -1, 0, error))
-        goto out;
+        return g_prefix_error (error, "Parsing %s: ", origin_path), FALSE;
```
jlebon: Hmm, maybe this deserves a `glnx_throw_prefix` or similar.

src/libostree/ostree-sysroot.c
```diff
@@ -451,74 +414,61 @@ _ostree_sysroot_read_boot_loader_configs (OstreeSysroot *self,
...
   /* Callers expect us to give them a sorted array */
   g_ptr_array_sort (ret_loader_configs, compare_loader_configs_for_sorting);
-
- done:
-  if (out_loader_configs)
-    *out_loader_configs = g_steal_pointer (&ret_loader_configs);
-  ret = TRUE;
- out:
-  return ret;
+  ot_transfer_out_value(out_loader_configs, &ret_loader_configs);
```
jlebon: Agreed, we should move this out somewhere so we can use it in rpm-ostree as well. I like that it's one line less of boilerplate.

jlebon:
Nice! @rh-atomic-bot r+ 5a5ba75

PR #793, opened 2017-04-14
cmdline: Start conversion to new code style

cgwalters:
This is just a few.  I'm tempted to try out the coccinelle
patch for this.

### DIFF
--- a/src/ostree/ot-main.c
+++ b/src/ostree/ot-main.c
@@ -223,7 +223,6 @@ ostree_option_context_parse (GOptionContext *context,
                              GError **error)
 {
   glnx_unref_object OstreeRepo *repo = NULL;
-  gboolean success = FALSE;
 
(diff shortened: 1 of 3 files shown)


src/ostree/ot-main.c
```diff
@@ -379,29 +368,22 @@ ostree_admin_option_context_parse (GOptionContext *context,
...
 
 gboolean
 ostree_ensure_repo_writable (OstreeRepo *repo,
                              GError **error)
 {
-  gboolean ret;
-
-  ret = ostree_repo_is_writable (repo, error);
-
-  g_prefix_error (error, "Cannot write to repository: ");
```
jlebon: Heh, had to do a double take here. That's a sly piece of code. Much prefer the rewritten version.

jlebon:
@rh-atomic-bot r+ daf63f5

PR #794, opened 2017-04-14
Rename "osname" → "stateroot"

cgwalters:
I never really liked the term "osname". I feel "stateroot" is a *lot* clearer,
since the osname/stateroot mostly just holds `/var`. Further it avoids the `os`
prefix which is already overloaded.

Some of the existing docs already talked about "operating system state", which
further reinforces this.

There's *lot* more things than this which reference the term "osname", but I
don't want to change *everything* yet in this patch in case we decide to do
something different - this just gets the highlights.


jlebon:
@rh-atomic-bot r+ 7343bcc

PR #792, opened 2017-04-14
repo/checkout: Finish conversion to new code style

cgwalters:
I plan to make some future changes here, and wanted to do this
first.

Random side note; how about converting the do/while loops for `EINTR` to
`TEMP_FAILURE_RETRY()`? We're very inconsistent about that...

PR #795, opened 2017-04-14
Checkout .dirmeta caching

cgwalters:
I was reading a strace the other day and noticed we were loading the same
`.dirmeta` object many times. Unlike the other object types, `.dirmeta` objects
don't accumulate much over time; there are only so many directory metadata types.
(Without SELinux involved it'd probably be 5-6 I'd guess offhand).

For `fedora-atomic/25/x86_64/docker-host` there are currently 34 `.dirmeta` in
the tree.

But how many times during a checkout did we load those 34 dirmeta objects?
With a quick strace:

```
$ strace -s 2048 -f -o strace.log ostree --repo=repo-build checkout -U fedora-atomic/25/x86_64/docker-host host-test-checkout
$ grep dirmeta strace.log | wc -l
7165
```

After, as you'd expect, we just loaded `34` from disk.  We do
6 system calls (`openat+fstat+fstat+read+read+close`) per dirmeta,
so we dropped a total of 42780 system calls - which is about 20% of the total
system calls made.

`perf record` tells me that we're spending ~40 of our time in the kernel during
a checkout, so reducing syscall traffic helps. Though most of that appears to be
in the VFS and XFS layers for `linkat` (which isn't surprising).

So how much did perf improve? Well, on my workstation, I get a lot of
fluctuation in timing, sometimes by 30%, so this was well within the noise. But
it's well worth speeding up checkout, and I think this optimization will shine
more as we improve performance elsewhere.

### DIFF
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -92,6 +92,9 @@ struct OstreeRepo {
   OstreeRepoTransactionStats txn_stats;
 
   GMutex cache_lock;
+  guint dirmeta_cache_refcount;
+  /* char * checksum → GVariant * for dirmeta objects, used in the checkout path */
+  GHashTable *dirmeta_cache;
 
   gboolean inited;
   gboolean writable;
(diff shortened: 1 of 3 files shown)


cgwalters:
Note this depends on https://github.com/ostreedev/ostree/pull/792

cgwalters:
:surfing_man: Rebased

src/libostree/ostree-repo-private.h
```diff
@@ -126,6 +129,24 @@ typedef struct {
...
 } OstreeDevIno;
 
+/* A MemoryCacheRef is an in-memory cache of objects (currently just DIRMETA).  This can
+ * be used when performing an operation that traverses a repository in someway.  Currently,
+ * the primary use case is ostree_repo_checkout_at() avoiding lots of duplicate dirmeta
+ * lookups.
+ */
+typedef struct {
+  OstreeRepo *repo;
+} OstreeRepoMemoryCacheRef;
```
jlebon: I'm confused why we need this, since the cache itself is part of `OstreeRepo`. It seems like the goal is to make the cache have a different lifetime from the `OstreeRepo` object? (Though it looks like the cache shouldn't be that large). Or was including it in `OstreeRepo` just a hack to avoid plumbing through a new set of arguments everywhere? Just making sure I understand the intent here.

I.e. can't we just always initialize the hash table and drop the secondary refcounting? This would have the advantage of reusing the cache across checkouts as well.

cgwalters: Right; I think `OstreeRepo` should boil down mostly to a directory fd.  Right now e.g. rpm-ostree will potentially hold on to the sysroot for a while (which I'd like to have hold a repo), and if we didn't do this, the cache would grow without bounds.

Including it in the repo itself is *partly* about avoiding argument plumbing - right now it'd be hugely painful since we'd have to rewrite `OstreeRepoFile`.  But it's also about the fact that in the future we want to do concurrent/mult-threaded checkout for example, and the repo acts as a shared cache there.

jlebon:
Ahh, understood.
@rh-atomic-bot r+ 3010157

PR #741, opened 2017-03-18
Sysroot merge ⛙ config

cgwalters:
Start of: https://github.com/ostreedev/ostree/issues/737

PR #797, opened 2017-04-14
 checkout: Add force_copy+SELinux options for checkout, use in deploy

cgwalters:
This is a variant of the efforts in #741
Working on `rpm-ostree livefs`, I realized though I needed to just
check out *new* files directly into the live `/etc` (and possibly
delete obsolete files).

The way the current `/etc` merge works is fundamentally different from
that.  So my plan currently is to probably do something like:

 - Compute diff
 - Check out each *new* file individually (as a copy)
 - Optionally delete obsolete files

Also, a few other things become more important - in the current deploy code, we
copy all of the files, then relabel them. But we shouldn't expose to *live*
systems the race conditions of doing that, plus we should only relabel files we
checked out.

By converting the deploy's /etc code to use this, we fix the same TODO item
there around atomically having the label set up as we create files. And further,
if we kill the `/var` relabeling which I think is unnecessary since Anaconda
does it, we could delete large chunks of code there.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -29,10 +29,16 @@
 #include "otutil.h"
 
 #include "ostree-repo-file.h"
+#include "ostree-sepolicy-private.h"
 #include "ostree-core-private.h"
 #include "ostree-repo-private.h"
 
 #define WHITEOUT_PREFIX ".wh."
+
+/* Per-checkout call state/caching */
+typedef struct {
+  GString *selabel_path_buf;
+} CheckoutState;
 
 static gboolean
 checkout_object_for_uncompressed_cache (OstreeRepo      *self,
@@ -160,6 +166,7 @@ write_regular_file_content (OstreeRepo            *self,
 static gboolean
 create_file_copy_from_input_at (OstreeRepo     *repo,
                                 OstreeRepoCheckoutAtOptions  *options,
+                                CheckoutState  *state,
                                 GFileInfo      *file_info,
                                 GVariant       *xattrs,
                                 GInputStream   *input,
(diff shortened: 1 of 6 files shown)


cgwalters:
WIP, needs more testing.  Also I note that we should really make more use of `setfscreatecon()` for dealing with the SELinux xattr... and basically have `glnx_set_xattrs_except_selinux()`.

cgwalters:
OK, removing WIP.  I've tested this manually on a F25AH host.  Note this rolls in https://github.com/ostreedev/ostree/pull/801

src/libostree/ostree-repo.h
```diff
@@ -767,14 +767,17 @@ typedef struct {
   gboolean enable_fsync;  /* Deprecated */
   gboolean process_whiteouts;
   gboolean no_copy_fallback;
-  gboolean unused_bools[7];
+  gboolean force_copy;
```
cgwalters: I may break this out as a separate PR too with exposure via the cmdline as well.

cgwalters:
:surfing_man: Rebased

This depends on https://github.com/ostreedev/ostree/pull/804 and https://github.com/ostreedev/ostree/pull/801

src/libostree/ostree-repo-checkout.c
```diff
@@ -217,6 +249,18 @@ create_file_copy_from_input_at (OstreeRepo     *repo,
...
+      if (sepolicy_enabled)
+        {
+          g_autofree char *label = NULL;
+          if (!ostree_sepolicy_get_label (options->sepolicy,
+                                          state->selabel_path_buf->str,
+                                          g_file_info_get_attribute_uint32 (file_info, "unix::mode"),
+                                          &label, cancellable, error))
+            return FALSE;
+          if (fsetxattr (temp_fd, "security.selinux", label, strlen (label), 0) < 0)
+            return glnx_throw_errno_prefix (error, "Setting security.selinux");
```
jlebon: Nit: would be helpful to include the name of the file in the error msg.

src/libostree/ostree-repo-checkout.c
```diff
@@ -688,6 +765,51 @@ checkout_tree_at (OstreeRepo                        *self,
...
+    {
+      GString *buf = g_string_new (options->sepolicy_prefix ?: options->subpath);
+      g_assert_cmpint (buf->len, >, 0);
+      // Ensure it ends with /
+      if (buf->str[buf->len-1] != '/')
+        g_string_append_c (buf, '/');
+      state.selabel_path_buf = buf;
+
+      /* Otherwise it'd just be corrupting things, and there's no use case */
+      g_assert (options->force_copy);
```
jlebon: Hmm, let's enforce this at the API entry point too.

src/libostree/ostree-repo-checkout.c
```diff
@@ -579,22 +652,15 @@ checkout_tree_at (OstreeRepo                        *self,
                        (guint64)repo_dfd_stat.st_dev, (guint64)destination_stat.st_dev);
 
   /* Set the xattrs now, so any derived labeling works */
```
jlebon: Can we update this comment as well since we might have done labeling already above?

src/libostree/ostree-repo-checkout.c
cgwalters: We already do that via the
`            return g_prefix_error (error, "Checkout of %s to %s: ", checksum, destination_name), FALSE;`
in the caller.

src/libostree/ostree-repo-checkout.c
cgwalters: This was already very close to the entrypoint, but OK.  Fixup :arrow_down:

jlebon:
Let's land #807 first before merging this one.

src/libostree/ostree-repo-checkout.c
jlebon: Ahh, got it.

jlebon:
🚄 @rh-atomic-bot r+ 8aefb93

jlebon:
Weird, the status did get reported.
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

jlebon:
Ohh I see.
@rh-atomic-bot r+ 2048b88

Issue #798, opened 2017-04-18
Latest ostree broke flatpak


alexlarsson:
The 2017.4 release broke flatpak: https://github.com/flatpak/flatpak/issues/710

It seems bare-user symlink checkouts are using hardlinks when they should not.
I bisected this to 
https://github.com/ostreedev/ostree/commit/d3385a3014e173299094133d644763e34e5ecd52

cgwalters:
Urg.  PR in https://github.com/ostreedev/ostree/pull/799

PR #799, opened 2017-04-18
checkout: Fix bare-user symlink checkouts

cgwalters:
Logic error introduced after refactoring; we hoisted the
`is_bare_user_symlink` variable to the top, but its computation
below.  But the `is_bare` symlink depended on it.

Closes: https://github.com/ostreedev/ostree/issues/798


alexlarsson:
Looks good to me.

alexlarsson:
Maybe we should have a symlink in our basic tests though?

cgwalters:
We did have one, we just weren't verifying it.  That's the addition to the test suite I made.  Right?

alexlarsson:
Oh, missed that part. Yeah, that looks good.

jlebon:
Ouch! Just to sanity check, I verified that the new symlink check would have failed without this fix.

@rh-atomic-bot r+ ee918d9

PR #800, opened 2017-04-18
Release 2017.5

cgwalters:
This is a bugfix release for 2017.4 to fix a regression
that broke flatpak: https://github.com/ostreedev/ostree/issues/798

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 AC_PREREQ([2.63])
 dnl If incrementing the version here, remember to update libostree.sym too
 m4_define([year_version], [2017])
-m4_define([release_version], [4])
+m4_define([release_version], [5])
 m4_define([package_version], [year_version.release_version])
 
 AC_INIT([libostree], [package_version], [walters@verbum.org])


jlebon:
Hmm, didn't get a notification for this for some reason.

This would pull in the recent static delta refactor too though, right? Could we make this a new `2017.4.1` branch & tag instead with just the regression fix?

cgwalters:
I thought about that, but the version stuff added to the header doesn't support minors right now...

cgwalters:
That said, we can ignore that I guess.  It'd just be a manual one-off.

cgwalters:
I'd prefer releasing the new code versus not to shortcutting the release infra right now.

cgwalters:
Maybe one thing that'd work is to make a one-off 2017.5, then push a commit bumping configure.ac but *not* a release to master?

cgwalters:
OK, I just reworded the commit message.

jlebon:
@rh-atomic-bot r+ d64c95d

PR #801, opened 2017-04-19
 checkout: Merge union/add logic for copies during checkout

cgwalters:
We really have an astonishing variety of similar functions which write files and
symlinks. I was working on a different PR and the duplication between the
union-mode and add-mode/none-mode checkout functions bothered me.

I realized that the "handle EEXIST" tri-state maps directly to the
`GLnxLinkTmpfileReplaceMode`, so deduping things makes even more sense.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -193,10 +193,17 @@ create_file_copy_from_input_at (OstreeRepo     *repo,
             return glnx_throw_errno (error);
         }
 
-      /* Process any xattrs now that we made the link */
-      if (xattrs && options->mode != OSTREE_REPO_CHECKOUT_MODE_USER)
+      /* Process ownership and xattrs now that we made the link */
+      if (options->mode != OSTREE_REPO_CHECKOUT_MODE_USER)
         {
-          if (!glnx_dfd_name_set_all_xattrs (destination_dfd, destination_name,
+          if (TEMP_FAILURE_RETRY (fchownat (destination_dfd, destination_name,
+                                            g_file_info_get_attribute_uint32 (file_info, "unix::uid"),
+                                            g_file_info_get_attribute_uint32 (file_info, "unix::gid"),
+                                            AT_SYMLINK_NOFOLLOW)) == -1)
+            return glnx_throw_errno_prefix (error, "fchownat");
+
+          if (xattrs != NULL &&
+              !glnx_dfd_name_set_all_xattrs (destination_dfd, destination_name,
                                              xattrs, cancellable, error))
             return FALSE;
         }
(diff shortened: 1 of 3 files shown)


cgwalters:
🛫 Rebased 🛬

src/libostree/ostree-repo-checkout.c
```diff
@@ -154,129 +154,57 @@ write_regular_file_content (OstreeRepo            *self,
...
-  if (g_file_info_get_file_type (file_info) == G_FILE_TYPE_SYMBOLIC_LINK)
-    {
-      if (!_ostree_make_temporary_symlink_at (destination_dfd,
-                                              g_file_info_get_symlink_target (file_info),
-                                              &temp_filename,
-                                              cancellable, error))
-        return FALSE;
-
+      /* Process any xattrs now that we made the link */
       if (xattrs)
```
jlebon: This is not the same behaviour though, right? I.e. if we're in `ADD_FILES` mode, then we should only apply xattrs if it didn't already exist, right? Also, it seems like the chmod bits conditional on user mode got snipped (not that it matters much for symlinks, though I suppose `ostree diff` would notice). Actually, even applying xattrs should be conditional on user mode.

But now, I'm surprised the testsuite didn't catch this. So either I'm missing something, or we should add a few more tests for these :)

cgwalters: Ah, good catch.  Yes.  In general, I'm not sure anyone uses xattrs on symlinks aside from SELinux (and possibly SMACK).  

Yeah, no test coverage for this bit right now.  Hmm.  Filed https://github.com/ostreedev/ostree/issues/806

jlebon: We still need the `chown` bit too, right? Otherwise, I see `ostree diff` immediately reporting something after a fresh checkout:

```
# ostree ls symlink-chown-test
d00755 0 0      0 /
l00777 1000 1000      0 /mylink -> mytarget
# mkdir tmp/copy-checkout && mount --bind tmp/copy-checkout tmp/copy-checkout
# ostree checkout symlink-chown-test tmp/copy-checkout/checkout
# ostree diff symlink-chown-test ./tmp/copy-checkout/checkout
M    /mylink
```

cgwalters: Yes, fixed.  I spent a little bit trying to add a test but it needs https://github.com/ostreedev/ostree/pull/797/commits/3a794eaa7791e2386cf77790d20ad449d43a75ab

We can add it after this all comes together.

cgwalters: i.e. fixup :arrow_down:

jlebon:
@rh-atomic-bot r+ 8afc474

PR #803, opened 2017-04-21
lib/cleanup: Port some of the cleanup code to fd-relative and new style

cgwalters:
There aren't many users of `g_file_enumerator_iterate()` left - those
remaining are usually good candidates for porting.  There's some more
porting to do in this file; a mix of trivial and harder.  This
one is a good candidate for an individual commit.


jlebon:
@rh-atomic-bot r+ aba1a2d

cgwalters:
@rh-atomic-bot retry

PR #804, opened 2017-04-21
repo: Add a "force copy" flag to checkout

cgwalters:
This is intended to be used for copying `/usr/etc` → `/etc` for
deployments.

A TODO here is to use `glnx_file_copy_at()` if the repo mode allows
it - then we'd use reflinks if available.

### DIFF
--- a/src/ostree/ot-builtin-checkout.c
+++ b/src/ostree/ot-builtin-checkout.c
@@ -42,6 +42,7 @@ static gboolean opt_from_stdin;
 static char *opt_from_file;
 static gboolean opt_disable_fsync;
 static gboolean opt_require_hardlinks;
+static gboolean opt_force_copy;
 
 static gboolean
 parse_fsync_cb (const char  *option_name,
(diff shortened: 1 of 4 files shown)


src/ostree/ot-builtin-checkout.c
```diff
@@ -102,6 +104,11 @@ process_one_checkout (OstreeRepo           *repo,
                        "Cannot specify both --union and --union-add");
           goto out;
         }
+      if (opt_require_hardlinks && opt_force_copy)
```
jlebon: Should  this be checked at the API level too?

tests/basic-test.sh
```diff
@@ -41,6 +41,7 @@ fi
 validate_checkout_basic() {
     (cd $1;
      assert_has_file firstfile
+     assert_not_streq $(stat -c '%h' firstfile) 1
```
jlebon: Interesting, this is failing in `test-basic-user.sh`. Surprisingly, not in the `-U -H` case, but just in the vanilla `$OSTREE checkout test2 checkout-test2` test. Is there a bug there? Seems like it's hardlinking from a bare-user repo even without `-U`.

jlebon: Wait, I read this backwards. The fact that it's failing means that it *is* copying rather than hardlinking, which is what we want in user-mode without `-U`. So I think this just needs some conditionals around it.

cgwalters: Yeah, this only works in bare right now.  This whole suite really needs rewriting.

src/ostree/ot-builtin-checkout.c
cgwalters: Done :arrow_down:

jlebon:
@rh-atomic-bot r+ aba5582

PR #807, opened 2017-04-24
tests/installed: New installed, privileged tests using Fedora AH

cgwalters:
Our container-driven tests can't e.g. test SELinux sanely, and
have to support being run as root *and* non-root too.

Use redhat-ci to provision a VM and run tests directly there. These are
installed tests too.

### DIFF
--- a/tests/installed/libinsttest.sh
+++ b/tests/installed/libinsttest.sh
@@ -1,0 +1,48 @@
+# Common definitions for installed, privileged tests
+#
+# Copyright (C) 2017 Colin Walters <walters@verbum.org>
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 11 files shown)


tests/installed/libinsttest.sh
```diff
@@ -0,0 +1,48 @@
...
+rpmostree=$(which rpm-ostree 2>/dev/null)
+if test -z "${rpmostree}"; then
+    skip "no rpm-ostree, at some point point this to raw ostree too"
+fi
+
+# We need to be root
+assert_streq $(id -u) 0
+
+PYTHON=
+for PYTHON in /usr/bin/python3 /usr/bin/python; do
```
jlebon: Maybe use a different placeholder var here, otherwise the `test -z` below will never fail.

jlebon:
This is really nice!
@rh-atomic-bot r+ ede8c14

PR #808, opened 2017-04-25
lib/util: Delete some leftover pre-libglnx directory opening functions

cgwalters:
These were migrated into libglnx; port the few callers to use that.


jlebon:
@rh-atomic-bot r+ 461b1d7

PR #809, opened 2017-04-25
 repo: More porting to new style

cgwalters:
I was planning to change some of the object loading code in the
future, so here's some porting.

Note that I rewrote `_ostree_repo_has_loose_object()` since it
used an error return across multiple functions.

Honestly I'm not sure about this `TEMP_FAILURE_RETRY()` business...
in reality we're going to end up with a ton of code linked in
process that doesn't do it.  Unix sucks =(  But I'm keeping
what was there out of consistency.


jlebon:
Nice! @rh-atomic-bot r+ 99fabed

PR #810, opened 2017-04-25
Minor fixes

pwithnall:
A few minor code fixes pulled out of PR #782 which should be self-contained enough to land now.


cgwalters:
Looks like the Travis(Ubuntu/Debian) builds have too old glib.  One thing I'll note here is the commit message in https://github.com/ostreedev/ostree/pull/810/commits/6c9639013620a51525a29251c2d052d7d290479f is a bit inaccurate in that libglnx carries backports, and we've been using that on older targets.

There are a few things we *don't* have backports for that I'd like to use, like `GMutexLocker` for example.

@alexlarsson - What's the min version of glib for flatpak?  I'm a bit surprised to see git master only does:
`PKG_CHECK_MODULES(BASE, [glib-2.0 gio-2.0 gio-unix-2.0])` without versions.

pwithnall:
> Looks like the Travis(Ubuntu/Debian) builds have too old glib.

Debian Jessie has [GLib 2.42.1](https://packages.debian.org/jessie/libglib2.0-0), which is a bit of a pain. How much do you care about supporting Jessie?

> One thing I'll note here is the commit message in 6c96390 is a bit inaccurate in that libglnx carries backports, and we've been using that on older targets.

Ah, I didn’t notice the backports. In that case, commit 6c96390 can probably be dumped. Given that OSTree currently builds fine on Jessie, our actual GLib dependency obviously can’t be above 2.42.

cgwalters:
There are two cases: flatpak and ostree-as-host.  For flatpak, looking at http://flatpak.org/getting.html it seems that we would indeed break the backports case, as ostree/flatpak are there, but newer glib isn't.

For ostree-as-host...I think anyone doing that with Debian is probably fine also pulling in a newer glib...hopefully?

But on the flip side if it's not *necessary* to do now...let's kick that can down the road a bit?

cgwalters:
I rebased this branch and dropped the GLib version bump.

@rh-atomic-bot r+ 9251b83

cgwalters:
Although on the `.gitignore` one...we're actually using [git.mk](https://github.com/behdad/git.mk) and offhand I think the right thing would be to add to `GITIGNOREFILES` in `Makefile.am`.  But eh, we can fix later.

alexlarsson:
@cgwalters flatpak has a glib-2.0 >= 2.44 requirement.

This is unfortunately not old enough for debian jessie, so @smcv has a back-port of some required G_DEFINE_TYPE header in the debian jessie backport.

cgwalters:
That's embedded in the flatpak package?  How about doing it in e.g. libglnx instead?

alexlarsson:
@cgwalters Its not at all in the upstream flatpak, debian is carrying a patch for it.
Adding it to libglnx sounds fine to me though. Its basically cut-and-pasting of some header file snippets.

smcv:
@cgwalters: Because Debian 9 'stretch' is in freeze and ostree doesn't have a stable-branch, any new upstream releases of ostree and 0.9.x development releases of flatpak are not going to get into Debian 9. New upstream releases of Flatpak 0.8.x do get into Debian 9.

Official stable-backports are always from the next Debian release into the previous, so from 9 into 8 at the moment; so I want to be able to compile Flatpak 0.8.x on Debian 8 'jessie' (GLib 2.42), but at this late stage in the release cycle I'm OK with patching it to make that work. I don't think it's worth introducing more backwards compat goo into libglnx just for my benefit.

For new ostree releases, and for Flatpak 0.9.x, the oldest thing I care about is Debian 9 (GLib 2.50).

smcv:
> New upstream releases of Flatpak 0.8.x do get into Debian 9

... as long as the release team let me get away with it. Some of the 0.8.x releases were verging on too big for a stable-branch IMO, I think they only got accepted because the release team trust me.

smcv:
Ubuntu LTS is the other main thing that matters in the Debian ecosystem: 14.04 has GLib 2.40 if you want to go that far back, and 16.04 has 2.48.

cgwalters:
I guess from my perspective, libostree as it stands today *is* a stable branch, and we certainly use it as such in the Fedora/CentOS ecosystem.  Yes, we will have a stream of new features, but we'll never break API/ABI.

smcv:
I don't mean a stable-branch like "GLib 2", I mean stable like "GLib 2.48.x" - no new features, no unnecessary refactoring/churn/diffstat, minimal regressions, every change justifiable as fixing a specific bug.

alexlarsson:
I don't plan to break ABI in flatpak 0.9 either, but thats not what "stable" means in the debian case.

cgwalters:
Well, it's fair to say we've had our regressions in libostree given 2017.4.  But I plan to step up our testing there.

smcv:
When Debian people say "stable" they primarily mean "doesn't change unexpectedly", more than they mean "doesn't crash".

Regardless of how much you test new ostree releases, I am never going to be able to integrate the latest upstream development release into stable Debian branches, for the same reasons we only follow point-release branches (if we take new upstream releases *at all*) for Linux, GLib, glibc, Perl, dbus and every other project in the distribution. In general, if the diffstat is larger than the release managers can reasonably review, then the answer is automatically no.

Targeted-bugfix-only stable-branches like Linux 4.9.x, GLib 2.50.x, dbus 1.10.x and so on are often OK, but if we took new "feature" releases, it would defeat the purpose of having a stable distro release. Rolling-release distros are available, and they have their place; Debian unstable is basically a rolling release. However, Debian stable is not one of those.

PR #811, opened 2017-04-25
repo: Fix incorrect use of errno() error throwing

cgwalters:
I happened to glance at the top of my most recent patch and
noticed that I used an `throw_errno()` function in a non-errno place.
I scanned the patch for other instances of this but didn't find one.


jlebon:
@rh-atomic-bot r+ 595ac27

PR #812, opened 2017-04-26
apidoc: Add missing enums to sections file

krnowak:
So they and their values show up in the documentation.


cgwalters:
@rh-atomic-bot r+ 48a67d7

PR #815, opened 2017-04-26
 sepolicy: Cache the value of is_selinux_enabled() to work around bug

cgwalters:
This fixes a regression from:
#797
which is really due to an underlying bug in libselinux which
we're working around:
http://marc.info/?l=selinux&m=149323809332417&w=2

We drop the per-policy instance variable, since the SELinux state
is *really* per-kernel.


jlebon:
Ahh, good catch.
@rh-atomic-bot r+ 9fa082a

PR #816, opened 2017-04-27
lib/checkout: Use TEMP_FAILURE_RETRY()

cgwalters:
I'm still not sure it's worth using, but it's easier on the eyes for sure.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -720,7 +720,7 @@ checkout_tree_at_recurse (OstreeRepo                        *self,
    */
   if (!did_exist)
     {
-      if (TEMP_FAILURE_RETRY (fchmod (destination_dfd, g_file_info_get_attribute_uint32 (source_info, "unix::mode")) < 0)
+      if (TEMP_FAILURE_RETRY (fchmod (destination_dfd, g_file_info_get_attribute_uint32 (source_info, "unix::mode"))) < 0)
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-checkout.c
```diff
@@ -126,19 +122,11 @@ write_regular_file_content (OstreeRepo            *self,
...
-      int res;
-      do
-        res = fchown (fd,
-                      g_file_info_get_attribute_uint32 (file_info, "unix::uid"),
-                      g_file_info_get_attribute_uint32 (file_info, "unix::gid"));
-      while (G_UNLIKELY (res == -1 && errno == EINTR));
-      if (G_UNLIKELY (res == -1))
+      if (TEMP_FAILURE_RETRY (fchown (fd, g_file_info_get_attribute_uint32 (file_info, "unix::uid"),
+                                      g_file_info_get_attribute_uint32 (file_info, "unix::gid"))) < 0)
         return glnx_throw_errno (error);
```
jlebon: This is more of a general comment rather than on this specific code, though after just reviewing the libglnx PR https://github.com/GNOME/libglnx/pull/43 (where more call sites were converted to throw errors prefixed with the syscall), it does seem like it would be nice to make it the standard coding style, I think? Might help debugging and diagnosing user-reported issues.

jlebon:
@rh-atomic-bot r+ 4c18e66

Issue #814, opened 2017-04-26
Error setting user password after package layering with rpm-ostree


mike-nguyen:
I am getting "Authentication token manipulation error" when changing user's password after using rpm-ostree package layering (I was able to change it before package layering) on Fedora Atomic Host Continuous.  The logs point to an selinux issue.  If I setenforce 0, I am able to change the password.  This might be related to https://github.com/ostreedev/ostree/commit/e8efd1c8dcaad8fbd3b05c400972d237406263e7

```
#  echo $pw | passwd atomic-user-2014 --stdin
Changing password for user atomic-user-2014.
passwd: Authentication token manipulation error
# setenforce 0
# echo $pw | passwd atomic-user-2014 --stdin
Changing password for user atomic-user-2014.
passwd: all authentication tokens updated successfully.
```

```
# rpm-ostree status
State: idle
Deployments:
● custom:fahc/25/x86_64/buildmaster
             Version: 25.175 (2017-04-25 17:27:37)
          BaseCommit: 0f9110627862710a40b2c701ee1a829dfadf7e62b3793bd3c9c2e1054f339622
     LayeredPackages: wget

  custom:fahc/25/x86_64/buildmaster
             Version: 25.175 (2017-04-25 17:27:37)
              Commit: 0f9110627862710a40b2c701ee1a829dfadf7e62b3793bd3c9c2e1054f339622
```

```
# rpm -q ostree
ostree-2017.4.20-e8efd1c8dcaad8fbd3b05c400972d237406263e7.d8b864b502c5853efda6e8f7d1b0ebff84b581a3.fc25.x86_64
```

Journal during error:
```
Apr 26 14:16:57 fedora.localdomain audit[2057]: AVC avc:  denied  { write } for  pid=2057 comm="passwd" name="etc" dev="dm-0" ino=16873451 scontext=unconfined_u:unconfined_r:passwd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:root_t:s0 tclass=dir permissive=0
[...]

PR #817, opened 2017-04-27
checkout/commit: Use glnx_regfile_copy_bytes() if possible

cgwalters:
Rather than `g_output_stream_splice()`, where the input is a regular
file.

See https://github.com/GNOME/libglnx/pull/44 for some more information.

I didn't try to measure the performance difference, but seeing the
read()/write() to/from userspace mixed in with the pointless `poll()` annoyed me
when reading strace.

As a bonus, we will again start using reflinks (if available) for `/etc`,
which is a regression from the https://github.com/ostreedev/ostree/pull/797
changes (which before used `glnx_file_copy_at()`).

Also, for the first time we'll use reflinks when doing commits from file-backed
content. This happens in `rpm-ostree compose tree` today for example.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -102,7 +102,7 @@ fsync_is_enabled (OstreeRepo   *self,
 static gboolean
 write_regular_file_content (OstreeRepo            *self,
                             OstreeRepoCheckoutAtOptions *options,
-                            GOutputStream         *output,
+                            int                    outfd,
                             GFileInfo             *file_info,
                             GVariant              *xattrs,
                             GInputStream          *input,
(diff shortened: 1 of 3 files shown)


cgwalters:
Couldn't help writing this after I'd been spending some time staring at `strace` output for https://github.com/ostreedev/ostree/issues/814 .

Before:
```
25056 openat(8, "ef/10923868386f564a1b63d0e261a8c3353ab470f69df2fe3b6298286655f09f.file", O_RDONLY|O_CLOEXEC) = 15
25056 lseek(15, 0, SEEK_CUR)            = 0
25056 openat(14, ".", O_WRONLY|O_DIRECTORY|O_CLOEXEC|O_TMPFILE, 0600) = -1 EOPNOTSUPP (Operation not supported)
25056 openat(14, "./tmp.y9lgUn", O_WRONLY|O_CREAT|O_EXCL|O_NOCTTY|O_NOFOLLOW|O_CLOEXEC, 0600) = 16
25056 lseek(16, 0, SEEK_CUR)            = 0
25056 poll([{fd=15, events=POLLIN}], 1, -1) = 1 ([{fd=15, revents=POLLIN}])
25056 read(15, "NAME=TestOS\nVERSION=42\nID=testos"..., 8192) = 71
25056 poll([{fd=16, events=POLLOUT}], 1, -1) = 1 ([{fd=16, revents=POLLOUT}])
25056 write(16, "NAME=TestOS\nVERSION=42\nID=testos"..., 71) = 71
25056 poll([{fd=15, events=POLLIN}], 1, -1) = 1 ([{fd=15, revents=POLLIN}])
25056 read(15, "", 8192)                = 0
25056 fchown(16, 1000, 1000)            = 0
25056 fchmod(16, 0100664)               = 0
25056 renameat2(14, "./tmp.y9lgUn", 14, "os-release", RENAME_NOREPLACE) = 0
25056 close(16)                         = 0
25056 close(15)      
```

After:
```
3778  newfstatat(8, "ef/10923868386f564a1b63d0e261a8c3353ab470f69df2fe3b6298286655f09f.file", {st_mode=S_IFREG|0664, st_size=71, ...}, AT_SYMLINK_NOFOLLOW) = 0
3778  openat(8, "ef/10923868386f564a1b63d0e261a8c3353ab470f69df2fe3b6298286655f09f.file", O_RDONLY|O_CLOEXEC) = 15
3778  lseek(15, 0, SEEK_CUR)            = 0
3778  openat(14, ".", O_WRONLY|O_DIRECTORY|O_CLOEXEC|O_TMPFILE, 0600) = -1 EOPNOTSUPP (Operation not supported)
3778  openat(14, "./tmp.QDjBhn", O_WRONLY|O_CREAT|O_EXCL|O_NOCTTY|O_NOFOLLOW|O_CLOEXEC, 0600) = 16
3778  sendfile(16, 15, NULL, 71)        = 71
3778  fchown(16, 1000, 1000)            = 0
3778  fchmod(16, 0100664)               = 0
3778  renameat2(14, "./tmp.QDjBhn", 14, "os-release", RENAME_NOREPLACE) = 0
3778  close(16)                         = 0
3778  close(15)   
```

cgwalters:
bot, retest this please

cgwalters:
Interesting; so the problem was that I think rhci was relying on github's auto-merge commit, rather than doing what homu would do via a rebase.  A bit of a messy corner case; rhci would have to know that the repo was homu managed.

But hopefully we can avoid that by relying on a FF-able PR, let's see.

jlebon:
bot, retest this please

src/libostree/ostree-repo-checkout.c
```diff
@@ -102,48 +102,62 @@ fsync_is_enabled (OstreeRepo   *self,
...
                             GCancellable          *cancellable,
                             GError               **error)
 {
   const OstreeRepoCheckoutMode mode = options->mode;
+  g_autoptr(GOutputStream) outstream = NULL;
 
-  if (g_output_stream_splice (output, input, 0,
-                              cancellable, error) < 0)
-    return FALSE;
+  if (G_IS_FILE_DESCRIPTOR_BASED (input))
```
jlebon: Will this condition ever fail?

src/libostree/ostree-repo-commit.c
```diff
@@ -570,28 +559,34 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
...
-  if (!fallocate_stream ((GFileDescriptorBased*)temp_out, length,
-                         cancellable, error))
+  if (!ot_fallocate (temp_fd, length, error))
     return FALSE;
 
-  if (g_output_stream_splice (temp_out, input, 0,
-                              cancellable, error) < 0)
-    return FALSE;
-  if (fchmod (temp_fd, 0644) < 0)
+  if (G_IS_FILE_DESCRIPTOR_BASED (input))
```
jlebon: Here too. Seems like they're always fd-based, no? I'm not very familiar with GIO.

src/libostree/ostree-repo-checkout.c
cgwalters: We expose as public API:
```
gboolean      ostree_repo_write_content (OstreeRepo       *self,
                                         const char       *expected_checksum,
                                         GInputStream     *object_input,
                                         guint64           length,
                                         guchar          **out_csum,
                                         GError          **error);
```

So we have to support it.  In practice using the command line tools, we will go through this code path when pulling from an `archive` repo to a `bare` one, since the content objects are zlib-compressed.

jlebon: Oh right, that makes sense.

jlebon:
@rh-atomic-bot r+ 4e09112

PR #818, opened 2017-04-28
ci: Add a context for testing flatpak

cgwalters:
This was part of the philosophy behind https://wiki.gnome.org/Initiatives/GnomeGoals/InstalledTests -
libraries like libostree don't need to replicate everything in unit tests, we
can use the tests from our dependencies directly too.

We'll also get API break coverage testing too.

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -155,3 +155,6 @@ host:
 
 tests:
   - docker run --rm --privileged -v $(pwd):/srv/code registry.fedoraproject.org/fedora:25 /bin/sh -c "cd /srv/code && ./ci/flatpak.sh"
+
+artifacts:
+  - test-suite.log
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -135,3 +135,23 @@ tests:
 
 artifacts:
   - test-suite.log
+
+---
+
+inherit: false
+branches:
+    - master
+    - auto
+    - try
+
+context: f25-flatpak
+required: false
+
+# This test case wants an "unprivileged container with bubblewrap",
+# which we don't have right now; so just provision a VM and do a
+# docker --privileged run.
+host:
+  distro: fedora/25/atomic
+
+tests:
+  - docker run --rm --privileged -v $(pwd):/srv/code registry.fedoraproject.org/fedora:25 /bin/sh -c "cd /srv/code && ./ci/flatpak.sh"
--- a/ci/flatpak.sh
+++ b/ci/flatpak.sh
@@ -9,6 +9,8 @@ build() {
     make -j 8
 }
 
+codedir=$(pwd)
+
 # Core prep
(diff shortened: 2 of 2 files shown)


cgwalters:
Although, this is less useful than it could be because
`SKIP: tests/test-run.sh - Cannot run bwrap`
which is https://github.com/jlebon/redhat-ci/issues/20

jlebon:
Can we just provision a VM for it for now?

cgwalters:
We could; that's what the flatpak CI is doing now.  I guess the thing that bothers me about that is that it's regressing to the old mindset of build-in-VM instead of build-in-container (and possibly run-in-VM after).

cgwalters:
That said, maybe it wouldn't be too hard to try to make a reusable script implementing that pattern, like I did for bubblewrap.  Then we could copy that around for a while, and it'd give us some more real-world experience for making it declarative in rhci.

cgwalters:
Oh right, the other problem with that is flatpak's tests are uninstalled.  So I'd have to port them.

jlebon:
Hmm, how about we provision a VM and run it all in a privileged container there?

> That said, maybe it wouldn't be too hard to try to make a reusable script implementing that pattern, like I did for bubblewrap.

I think that pattern will be obsoleted by adding RPM support to rhci, right? Then you have a bona fide tree with your bits already installed.

cgwalters:
Ok, done :arrow_up:

cgwalters:
Oh heh right, not quite:

```
SKIP: tests/test-run.sh - this test requires xattr support
```

Yep, the only sane thing to do is to change the flatpak tests to be installed.  I'd like to do that for ostree too.

.redhat-ci.yml
```diff
@@ -135,3 +135,23 @@ tests:
...
+required: false
+
+# This test case wants an "unprivileged container with bubblewrap",
+# which we don't have right now; so just provision a VM and do a
+# docker --privileged run.
+host:
+  distro: fedora/25/atomic
+
+tests:
+  - docker run --rm --privileged -v $(pwd):/srv/code registry.fedoraproject.org/fedora:25 /bin/sh -c "cd /srv/code && ./ci/flatpak.sh"
```
jlebon: Can we also upload the test results?

```yaml
artifacts:
  - test-suite.log
```

And in the `flatpak.sh` script, have something like:

```sh
trap '[ ! -f test-suite.log ] || cp test-suite.log /srv/code' ERR
```

?

jlebon:
That's weird. The default docker backend is still devmapper for F25, right?

Actually, looking at the output, I don't see `attr` getting installed at all. So could it just be `setfattr` missing? I suggested uploading the logs as well to help debug that.

jlebon:
Nice! The error is just flaky infra, so let's give it a try.
@rh-atomic-bot r+ 9ee612b

PR #821, opened 2017-04-28
repo: Delete the last use of GFile tmp_dir

cgwalters:
The keyring isn't large, so let's just fall back to copying it
rather than requiring `renameat()`.

Prep for `ostree_repo_open_at()`.


jlebon:
Nice!
@rh-atomic-bot r+ f115da3

cgwalters:
Looks like a pull flake.

@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #822, opened 2017-04-28
ci: Fix flatpak test pkg install

cgwalters:


jlebon:
Looks like some tests are failing with:

```
+ fusermount -u /var/tmp/test-flatpak-CIw3Db/runtime/doc
fusermount: failed to unmount /var/tmp/test-flatpak-CIw3Db/runtime/doc: No such file or directory
```

Though this is not a required suite yet, so we can fix it up later.
@rh-atomic-bot r+ 2a64610

PR #823, opened 2017-04-28
checkout: Dedup calls to memcache ref

cgwalters:
Minor, but I realized `checkout_tree_at()` is a better place to
do common setup before checkout.  Prep for
https://github.com/ostreedev/ostree/pull/813


jlebon:
@rh-atomic-bot r+ ed253bb

PR #824, opened 2017-05-01
ci: More flatpak ci fixes

cgwalters:
We need our `make install` to override the ostree RPM, so do it all in one txn.
This sort of thing is where a more rigorous model like rdgo/gcontinuous use
becomes better, but we'll hack it with shell for now.

### DIFF
--- a/ci/flatpak.sh
+++ b/ci/flatpak.sh
@@ -12,23 +12,27 @@ build() {
 codedir=$(pwd)
 
 # Core prep
-dnf -y install dnf-plugins-core
-dnf install -y @buildsys-build
-dnf install -y 'dnf-command(builddep)'
+yum -y install dnf-plugins-core @buildsys-build 'dnf-command(builddep)'
+# build+install ostree, and build deps for both, so that our
+# make install overrides the ostree via rpm
+dnf builddep -y ostree flatpak
+yum -y install flatpak && rpm -e flatpak
+# we use yaml below
+yum -y install python3-PyYAML
 
-# build+install ostree
-dnf builddep -y ostree
-build
-make install
+# Now get flatpak's deps from rhci file
 tmpd=$(mktemp -d)
 cd ${tmpd}
 # Frozen to a tag for now on general principle
 git clone --recursive --depth=1 -b 0.9.3 https://github.com/flatpak/flatpak
 cd flatpak
-dnf builddep -y flatpak
-# And runtime deps
-dnf install -y flatpak && rpm -e flatpak
-dnf install -y which attr fuse parallel # for the test suite
+python3 -c 'import yaml; y = list(yaml.load_all(open(".redhat-ci.yml")))[0]; print("\0".join(y["packages"]))' | xargs -0 yum install -y
+# back to ostree and build
+cd ${codedir}
+build
+make install
+cd ${tmpd}/flatpak
+patch -p1 < ${codedir}/ci/*.patch
 build
 # We want to capture automake results from flatpak
 cleanup() {
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -153,6 +153,8 @@ required: false
 # docker --privileged run.
 host:
   distro: fedora/25/atomic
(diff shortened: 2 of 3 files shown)


jlebon:
Yup, makes sense! Though there are still some failures it seems. 😢
Feel free to merge if you mean to investigate this in a separate PR.

jlebon:
```
ERROR: tests/test-run.sh
========================
...
+ /tmp/tmp.SrvEsn2GNT/flatpak/tests/test-webserver.sh repos
cmp: EOF on /var/tmp/tap-test.FIEWHd/httpd-port
++ cat httpd-port
+ port=
++ cat httpd-pid
+ FLATPAK_HTTP_PID=29889
+ flatpak remote-add --user --gpg-import=/var/tmp/test-flatpak-1CsUdX/gpghome/pubring.gpg test-repo http://127.0.0.1:/test
+ install_repo
+ flatpak --user install test-repo org.test.Platform master
error: Could not connect: Connection refused
```

Not sure what happened there. Seems like the web server couldn't get started?

All the other failures are variations of:

```
++ echo 'Couldn'\''t find '\''/var/tmp/test-flatpak-UQW869/system/exports/share/applications/mimeinfo.cache'\'''
Couldn't find '/var/tmp/test-flatpak-UQW869/system/exports/share/applications/mimeinfo.cache'
```

cgwalters:
It looks like this got a lot closer...we possibly got OOM killed?

bot, retest this please

cgwalters:
bot, retest this please

cgwalters:
I think this is https://github.com/flatpak/flatpak/pull/772

cgwalters:
bot, retest this please

cgwalters:
Ah yep, RAM was the problem.  This should be GTG.

ci/flatpak.sh
```diff
@@ -16,19 +16,20 @@ dnf -y install dnf-plugins-core
...
 dnf install -y 'dnf-command(builddep)'
 
-# build+install ostree
-dnf builddep -y ostree
+# build+install ostree, and build deps for both, so that our
+# make install overrides the ostree via rpm
+dnf builddep -y ostree flatpak
+dnf install -y flatpak && rpm -e flatpak
+# From the current flatpak builds
+dnf install -y gcc sudo which attr fuse gjs parallel clang libubsan gnome-desktop-testing 'pkgconfig(fuse)' 'pkgconfig(gio-unix-2.0)' 'pkgconfig(gobject-introspection-1.0)' 'pkgconfig(json-glib-1.0)' 'pkgconfig(libarchive)' 'pkgconfig(libelf)' 'pkgconfig(libsoup-2.4)' 'pkgconfig(ostree-1)' 'pkgconfig(polkit-gobject-1)' 'pkgconfig(libseccomp)' 'pkgconfig(xau)' 'bubblewrap' docbook-dtds docbook-style-xsl intltool libattr-devel libcap-devel libdwarf-devel elfutils systemd ostree gpgme-devel /usr/bin/xmlto /usr/bin/xsltproc redhat-rpm-config /usr/bin/update-mime-database /usr/bin/update-desktop-database /usr/bin/gtk-update-icon-cache
```
jlebon: How about we just fetch it from the YAML file after we clone it, e.g. something like:

```
python3 -c 'import yaml; y = list(yaml.load_all(open(".redhat-ci.yml")))[0]; print("\n".join(y["packages"]))' | xargs dnf install -y
```

?

cgwalters: OK, done (though I had to use NUL rather than newline separated to make e.g. `pkgconfig(gobject-introspection-1.0) >= 1.40.0` work).

That said the reason I didn't do this is I think at least for flatpak we should switch upstream to do privcontainer rather than host as we are here - and that circles back to enhancing rhci for this, or prototyping it outside perhaps.

cgwalters:
OK, rebased and squashed.

jlebon:
Nice!
@rh-atomic-bot r+ 8aa7556

PR #825, opened 2017-05-02
libostree: Add missing checks for invalid timestamps

pwithnall:
g_date_time_new_from_unix_utc() will not always return a valid GDateTime
— if the input timestamp is too big, GDateTime cannot represent it, and
the constructor returns NULL.

Add some missing checks for these situations. We don’t ever expect
timestamps to be this big, but they could be as a result of corruption
or a malicious repository.

### DIFF
--- a/src/libostree/ostree-gpg-verify-result.c
+++ b/src/libostree/ostree-gpg-verify-result.c
@@ -562,6 +562,14 @@ ostree_gpg_verify_result_describe_variant (GVariant *variant,
   key_id = (len > 16) ? fingerprint + len - 16 : fingerprint;
 
   date_time_utc = g_date_time_new_from_unix_utc (timestamp);
+  if (date_time_utc == NULL)
+    {
+      g_string_append_printf (output_buffer,
+                              "Can't check signature: timestamp %" G_GINT64_FORMAT " is invalid\n",
+                              timestamp);
+      return;
+    }
+
   date_time_local = g_date_time_to_local (date_time_utc);
   formatted_date_time = g_date_time_format (date_time_local, "%c");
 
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-gpg-verify-result.c
```diff
@@ -562,6 +562,14 @@ ostree_gpg_verify_result_describe_variant (GVariant *variant,
   key_id = (len > 16) ? fingerprint + len - 16 : fingerprint;
 
   date_time_utc = g_date_time_new_from_unix_utc (timestamp);
+  if (date_time_utc == NULL)
```
cgwalters: It looks like in this case we'd hit at least one or two `g_return_if_fail()s`, and at least on glibc systems just end up printing `(null)`.  So, probably not exploitable.

cgwalters:
Nice catch, thanks!  Did you happen to hit this elsewhere and just did a `git grep` to look for other instances?

It's too bad there's not (AFAIK) a GCC attribute to say "this function's return value must be checked against NULL", I guess it's the province of static analyzers.

@rh-atomic-bot r+ 8cc0968

pwithnall:
> Did you happen to hit this elsewhere and just did a git grep to look for other instances?

Yeah, I caught it by passing something invalid to `g_time_val_to_iso8601()` ([GLib bug](https://bugzilla.gnome.org/show_bug.cgi?id=782075)) which then made me look at `GDateTime` ([GLib bug](https://bugzilla.gnome.org/show_bug.cgi?id=782089)), then I grepped everything.

pwithnall:
> It's too bad there's not (AFAIK) a GCC attribute to say "this function's return value must be checked against NULL", I guess it's the province of static analyzers.

Yeah, this is one of the things which [Tartan](https://www.freedesktop.org/software/tartan/) can check, keying off the `(nullable)` annotation.

Clang does have [some attributes](https://clang.llvm.org/docs/AttributeReference.html#nullability-attributes) for this (some of which I think are also implemented in GCC?), but it’s far from a complete solution.

PR #826, opened 2017-05-02
libostree: Add some additional metadata to the summary file

pwithnall:
• Commit timestamps, so it’s easy to work out whether a given commit is
   newer than the one we have locally
 • Summary file timestamp, so it’s easy to work out whether the summary
   file is more up to date than another summary file
 • Summary file expiry time, so clients can work out when they should
   expect the summary file to next be updated, and hence can query for
   it at roughly the right time

The expiry time requires input from the user, so is currently never set
automatically. Programs using libostree can set it if they wish.

### DIFF
--- a/src/libostree/ostree-core.h
+++ b/src/libostree/ostree-core.h
@@ -158,6 +158,17 @@ typedef enum {
  * - a(s(taya{sv})) - Map of ref name -> (latest commit size, latest commit checksum, additional metadata), sorted by ref name
  * - a{sv} - Additional metadata, at the current time the following are defined:
  *   - key: "ostree.static-deltas", value: a{sv}, static delta name -> 32 bytes of checksum
+ *   - key: "ostree.summary.last-modified", value: t, timestamp (seconds since
+ *     the Unix epoch in UTC, big-endian) when the summary was last regenerated
+ *     (similar to the HTTP `Last-Modified` header)
+ *   - key: "ostree.summary.expires", value: t, timestamp (seconds since the
+ *     Unix epoch in UTC, big-endian) after which the summary is considered
+ *     stale and should be re-downloaded if possible (similar to the HTTP
+ *     `Expires` header)
+ *
+ * The currently defined keys for the `a{sv}` of additional metadata for each commit are:
+ *  - key: `ostree.commit.timestamp`, value: `t`, timestamp (seconds since the
+ *    Unix epoch in UTC, big-endian) when the commit was committed
  */
 #define OSTREE_SUMMARY_GVARIANT_STRING "(a(s(taya{sv}))a{sv})"
 #define OSTREE_SUMMARY_GVARIANT_FORMAT G_VARIANT_TYPE (OSTREE_SUMMARY_GVARIANT_STRING)
--- a/src/libostree/ostree-core.h
+++ b/src/libostree/ostree-core.h
@@ -104,9 +104,9 @@ typedef enum {
 /**
  * OSTREE_DIRMETA_GVARIANT_FORMAT:
  *
- * - u - uid
- * - u - gid
- * - u - mode
+ * - u - uid (big-endian)
+ * - u - gid (big-endian)
+ * - u - mode (big-endian)
  * - a(ayay) - xattrs
  */
 #define OSTREE_DIRMETA_GVARIANT_STRING "(uuua(ayay))"
@@ -120,9 +120,9 @@ typedef enum {
  * can't store in the real filesystem but we can still use a regular .file object
  * that we can hardlink to in the case of a user-mode checkout.
  *
- * - u - uid
- * - u - gid
- * - u - mode
+ * - u - uid (big-endian)
+ * - u - gid (big-endian)
+ * - u - mode (big-endian)
  * - a(ayay) - xattrs
  */
 #define OSTREE_FILEMETA_GVARIANT_STRING "(uuua(ayay))"
@@ -145,7 +145,7 @@ typedef enum {
  * - a(say) - Related objects
  * - s - subject
  * - s - body
- * - t - Timestamp in seconds since the epoch (UTC)
+ * - t - Timestamp in seconds since the epoch (UTC, big-endian)
  * - ay - Root tree contents
  * - ay - Root tree metadata
  */
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -4563,6 +4563,10 @@ ostree_repo_verify_summary (OstreeRepo    *self,
  * An OSTree repository can contain a high level "summary" file that
  * describes the available branches and other metadata.
  *
+ * If the timetable for making commits and updating the summary file is fairly
+ * regular, setting the `ostree.summary.expires` key in @additional_metadata
+ * will aid clients in working out when to check for updates.
+ *
  * It is regenerated automatically after a commit if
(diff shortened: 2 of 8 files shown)


src/libostree/ostree-core.h
```diff
@@ -158,6 +158,16 @@ typedef enum {
  * - a(s(taya{sv})) - Map of ref name -> (latest commit size, latest commit checksum, additional metadata), sorted by ref name
  * - a{sv} - Additional metadata, at the current time the following are defined:
  *   - key: "ostree.static-deltas", value: a{sv}, static delta name -> 32 bytes of checksum
+ *   - key: "ostree.summary.last-modified", value: t, timestamp (seconds since
+ *     the Unix epoch in UTC) when the summary was last regenerated (similar to
+ *     the HTTP `Last-Modified` header)
```
cgwalters: And we need this for cases where we're not fetching over HTTP?  At least in the HTTP case we could get the value from the headers and store it in an xattr or so.

But, I guess if we need to support non-HTTP, it's OK.

src/libostree/ostree-repo.c
```diff
@@ -4611,11 +4618,21 @@ ostree_repo_regenerate_summary (OstreeRepo     *self,
       if (!ostree_repo_load_variant (self, OSTREE_OBJECT_TYPE_COMMIT, commit, &commit_obj, error))
         goto out;
 
-      g_variant_builder_add_value (refs_builder, 
+      g_variant_dict_init (&commit_metadata_builder, NULL);
+
+      /* Forward the commit’s timestamp if it’s valid. */
+      g_variant_get_child (commit_obj, 5, "t", &commit_timestamp);
```
cgwalters: You want `ostree_commit_get_timestamp()` - otherwise we won't do the big-endian swapping.  (Keep this in mind for all ostree gvariant usage)

src/libostree/ostree-repo.c
```diff
@@ -4665,6 +4682,11 @@ ostree_repo_regenerate_summary (OstreeRepo     *self,
   }
 
   {
+    g_variant_dict_insert_value (&additional_metadata_builder, OSTREE_SUMMARY_LAST_MODIFIED,
+                                 g_variant_new_uint64 (g_get_real_time ()));
```
cgwalters: I thought the docs said seconds?  That's microseconds.

src/libostree/ostree-core.h
pwithnall: Yeah, for the cases where the summary file is stored as a file somewhere (mtime is a bit too fragile, and is not signed), or where it’s propagated through several Avahi peers (similarly), or where a HTTP server is used but configuring `Last-Modified` is a pain to do (I don’t know how likely this is, but from experience, setting HTTP headers on static content is not always trivial).

src/libostree/ostree-repo.c
pwithnall: Oops, will fix that.

src/libostree/ostree-repo.c
pwithnall: Well caught.

pwithnall:
Updated, now with additional bonus commits to allow viewing of a local summary file.

src/libostree/ostree-repo.c
```diff
@@ -4598,6 +4602,9 @@ ostree_repo_regenerate_summary (OstreeRepo     *self,
       const char *commit = g_hash_table_lookup (refs, ref);
       g_autofree char *remotename = NULL;
       g_autoptr(GVariant) commit_obj = NULL;
+      g_auto(GVariantDict) commit_metadata_builder = OT_VARIANT_BUILDER_INITIALIZER;
```
krnowak: You probably could use `G_VARIANT_DICT_INIT (NULL)` here, unless the GLib version requirement is too low.

src/ostree/ot-dump.c
```diff
@@ -251,11 +263,35 @@ ot_dump_summary_bytes (GBytes          *summary_bytes,
...
-      g_print ("%s: %s\n", key, string);
+      g_autofree gchar *value_str = NULL;
+      const gchar *pretty_key = NULL;
+
+      if (g_strcmp0 (key, "ostree.static-deltas") == 0)
+        {
+          pretty_key = "Static Deltas";
+          value_str = g_variant_print (value, FALSE);
+        }
+      else if (g_strcmp0 (key, "ostree.summary.last-modified") == 0)
```
krnowak: You could include the `ostree-repo-private.h` file here and use the macros instead of strings.

pwithnall: I deliberately decided to use the strings, since that seems to be the style in the rest of OSTree. I could be mistaken though; @cgwalters?

src/libostree/ostree-repo.c
pwithnall: Dependency is too low: OSTree depends on 2.40; `G_VARIANT_DICT_INIT()` requires 2.50.

src/ostree/ot-dump.c
cgwalters: Using defines seems nicer to me, I can't think of a reason we weren't elsewhere.

pwithnall: I’ve pushed a follow-up commit on the branch which switches to using all the existing `#define`s.

src/ostree/ot-builtin-summary.c
```diff
@@ -20,17 +20,21 @@
...
 
-static gboolean opt_update;
+static gboolean opt_update, opt_view, opt_raw;
 static char **opt_key_ids;
 static char *opt_gpg_homedir;
 
 static GOptionEntry options[] = {
   { "update", 'u', 0, G_OPTION_ARG_NONE, &opt_update, "Update the summary", NULL },
+  { "view", 'v', 0, G_OPTION_ARG_NONE, &opt_view, "View the local summary file", NULL },
+  { "raw", 0, 0, G_OPTION_ARG_NONE, &opt_raw, "View the raw bytes of the summary file", NULL },
```
cgwalters: Can you add at least a sanity test for this in `tests/`?

src/libostree/ostree-repo.c
```diff
@@ -4611,11 +4618,21 @@ ostree_repo_regenerate_summary (OstreeRepo     *self,
...
         goto out;
 
-      g_variant_builder_add_value (refs_builder, 
+      g_variant_dict_init (&commit_metadata_builder, NULL);
+
+      /* Forward the commit’s timestamp if it’s valid. */
+      commit_timestamp = ostree_commit_get_timestamp (commit_obj);
+      dt = g_date_time_new_from_unix_utc (commit_timestamp);
+
+      if (commit_timestamp > 0 && dt != NULL)
```
cgwalters: Why `if commit_timestamp > 0`?  What would make `0` invalid but not e.g. `1`?

pwithnall: I was treating `commit_timestamp == 0` as unset in general, but I guess that probably doesn’t apply here. Commits should always have timestamps. I’ll drop that clause.

src/ostree/ot-builtin-summary.c
pwithnall: Will do.

pwithnall: Done as an additional commit.

cgwalters:
@rh-atomic-bot r+ cf28bf0

PR #829, opened 2017-05-03
pull: Fix crash specifying override URL in summary fetch

dbnicholson:
The summary URL override is looked up with "&s", which directly
exchanges the data to a pointer without allocation. This was causing a
segfault calling ostree_repo_remote_fetch_summary_with_options from
pygobject.


jlebon:
Thanks!
@rh-atomic-bot r+ d601af7

PR #832, opened 2017-05-04
libostree: Make OstreeRemote a public and internal API

pwithnall:
Previously it was static to ostree-repo.c. Make it usable throughout
libostree so it can be used by an upcoming commit, but also expose the
typedef and reference counting functions so that opaque OstreeRemote
pointers can be used by user code, in anticipation of exposing more of
its API publicly in future.

---

This also includes a commit to introduce an `--enable-experimental-api` configure option, which the changes to the remotes API are hidden behind.

It also includes all the commits from PR #826 because I can’t figure out a way to make one PR depend on the other on GitHub.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -94,6 +94,8 @@ libostree_1_la_SOURCES = \
 	src/libostree/ostree-linuxfsutil.c \
 	src/libostree/ostree-diff.c \
 	src/libostree/ostree-mutable-tree.c \
+	src/libostree/ostree-remote.c \
+	src/libostree/ostree-remote-private.h \
 	src/libostree/ostree-repo.c \
 	src/libostree/ostree-repo-checkout.c \
 	src/libostree/ostree-repo-commit.c \
@@ -144,6 +146,11 @@ if HAVE_LIBSOUP_CLIENT_CERTS
 libostree_1_la_SOURCES += \
 	src/libostree/ostree-tls-cert-interaction.c \
 	src/libostree/ostree-tls-cert-interaction.h \
+	$(NULL)
+endif
+if !ENABLE_EXPERIMENTAL_API
+libostree_1_la_SOURCES += \
+	src/libostree/ostree-remote.h \
 	$(NULL)
 endif
 
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -156,7 +156,15 @@ libostree_1_la_LIBADD = libotutil.la libglnx.la libbsdiff.la libostree-kernel-ar
 libostree_1_la_LIBADD += $(bupsplitpath)
 EXTRA_libostree_1_la_DEPENDENCIES = $(top_srcdir)/src/libostree/libostree.sym
 
-EXTRA_DIST += src/libostree/libostree.sym
+EXTRA_DIST += \
+	src/libostree/libostree.sym \
+	src/libostree/libostree-experimental.sym \
+	$(NULL)
+
+if ENABLE_EXPERIMENTAL_API
+libostree_1_la_LDFLAGS += -Wl,--version-script=$(top_srcdir)/src/libostree/libostree-experimental.sym
+EXTRA_libostree_1_la_DEPENDENCIES += $(top_srcdir)/src/libostree/libostree-experimental.sym
+endif
 
 if USE_LIBARCHIVE
 libostree_1_la_CFLAGS += $(OT_DEP_LIBARCHIVE_CFLAGS)
--- a/apidoc/Makefile.am
+++ b/apidoc/Makefile.am
@@ -120,6 +120,7 @@ include $(top_srcdir)/gtk-doc.make
 EXTRA_DIST += \
 	version.xml \
 	ostree-sections.txt \
+	ostree-experimental-sections.txt \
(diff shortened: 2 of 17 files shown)


Makefile-libostree.am
```diff
@@ -156,7 +156,15 @@ libostree_1_la_LIBADD = libotutil.la libglnx.la libbsdiff.la libostree-kernel-ar
 libostree_1_la_LIBADD += $(bupsplitpath)
 EXTRA_libostree_1_la_DEPENDENCIES = $(top_srcdir)/src/libostree/libostree.sym
 
-EXTRA_DIST += src/libostree/libostree.sym
+EXTRA_DIST += \
+	src/libostree/libostree.sym \
+	src/libostree/libostree-experimental.sym \
```
cgwalters: I'm ok with this as is...but thinking 🤔 about it more - how about adding a `ostree-exp.pc`/`libostree-exp.so`?  If we were good about bumping the soname every time we changed API, I think I could use it for rpm-ostree too in Fedora.

Did we discuss the "separate .so" option?  I don't recall we did.

pwithnall: I’m not sure I want to commit to that level of stability for the unstable API. I don’t think we discussed a ‘separate `.so`’ option — I’m not sure how it would work, given that the experimental functions need access to the normal `libostree.so` internals.

cgwalters: Oh, true.  Theroetically we could use a variant of `OstreeCmdPrivateVTable` which is the same idea as [glib private call](https://git.gnome.org/browse/glib/commit/?id=9bf59d4a14b3fc4bb62f10c3665e99f9d8788bdc).

apidoc/Makefile.am
```diff
@@ -119,7 +119,19 @@ include $(top_srcdir)/gtk-doc.make
...
+	ostree-main-sections.txt \
+	ostree-experimental-sections.txt \
+	$(NULL)
+CLEANFILES += \
 	ostree-sections.txt \
 	$(NULL)
 
+if ENABLE_EXPERIMENTAL_API
+ostree-sections.txt: ostree-main-sections.txt ostree-experimental-sections.txt
+	cat $^ > $@
```
cgwalters: This is quite minor but I habitually make build rules write files atomically, otherwise you can end up with weird, hard to debug problems if a build is interrupted in the middle.  So:

```
cat  $^ > $@.tmp && mv $@{.tmp,}
```

(Perhaps in some future world Unix cmdline tools will learn `O_TMPFILE`)

src/libostree/libostree-experimental.sym
```diff
@@ -24,9 +24,8 @@
  * --enable-experimental-api. They are not stable or officially supported, and
  * might disappear or change in future releases. */
 
-/*
 LIBOSTREE_2017.6_EXPERIMENTAL {
 global:
-  some_symbol;
+  ostree_remote_ref;
+  ostree_remote_unref;
```
cgwalters: It looks like `tests/test-symbols.sh` will have to learn about this too.

src/libostree/ostree-autocleanups.h
```diff
@@ -59,6 +59,10 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (OstreeSysrootUpgrader, g_object_unref)
 
 G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC (OstreeRepoCommitTraverseIter, ostree_repo_commit_traverse_iter_clear)
 
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
```
cgwalters: There's something subtle here which I think is worth calling out; because this header file is public, this `#define` is used both for the build of ostree itself, and *also* clients will have to `#define` it.

That seems OK, but it'd probably be worth at least a comment in `configure.ac` that the name of the define is effectively API.

Or, we could use a different `#define` here for clients - `OSTREE_I_WANT_EXPERIMENTAL_API` ?  (And define it in our build if the automake conditional is set)

Makefile-libostree.am
pwithnall: I’d rather not put too much engineering time into this `--enable-experimental-api` flag, since it’s hopefully temporary. As long as it’s good enough for a few releases, I’ll be happy.

apidoc/Makefile.am
pwithnall: Fixed.

src/libostree/libostree-experimental.sym
pwithnall: Ah yes, fixed.

src/libostree/ostree-autocleanups.h
pwithnall: > That seems OK, but it'd probably be worth at least a comment in configure.ac that the name of the define is effectively API.

It’s deliberate — but you’re right, I should call it out in a comment. Will add one.

I could use a different `#define` if you want. For greppability, I kept them the same, but I’m not wedded to this idea.

Makefile-libostree.am
cgwalters: Yep, I think that's OK.

cgwalters:
@rh-atomic-bot r+  	16dc114

PR #834, opened 2017-05-04
Mark ostree_repo_transaction_set_ref* checksums nullable

dbnicholson:
So you can delete refs in a transaction from GI bindings. In order to test this, I added it to the JS tests, which only run installed. Which led me down the rabbit hole of why the installed tests were failing horribly.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1306,7 +1306,7 @@ ensure_txn_refs (OstreeRepo *self)
  * ostree_repo_transaction_set_refspec:
  * @self: An #OstreeRepo
  * @refspec: The refspec to write
- * @checksum: The checksum to point it to
+ * @checksum: (nullable): The checksum to point it to
  *
  * Like ostree_repo_transaction_set_ref(), but takes concatenated
  * @refspec format as input instead of separate remote and name
@@ -1329,7 +1329,7 @@ ostree_repo_transaction_set_refspec (OstreeRepo *self,
  * @self: An #OstreeRepo
  * @remote: (allow-none): A remote for the ref
  * @ref: The ref to write
- * @checksum: The checksum to point it to
+ * @checksum: (nullable): The checksum to point it to
  *
  * If @checksum is not %NULL, then record it as the target of ref named
  * @ref; if @remote is provided, the ref will appear to originate from that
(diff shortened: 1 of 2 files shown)


tests/test-basic-user-only.sh
```diff
@@ -21,5 +21,7 @@ set -euo pipefail
 
 . $(dirname $0)/libtest.sh
 
+skip_without_user_xattrs
```
cgwalters: Hmm...the whole idea of `bare-user-only` is we shouldn't need user xattrs.

cgwalters:
The CI tests here *do* run the installed tests, but inside a (Docker) container.  I assume you're doing development directly on a host system and hence the `/tmp` vs `/var/tmp` distinction?

InstalledTests certainly has the issue of not specifying much about container vs not.

cgwalters:
It's pretty tempting though to do a hard move to installed tests.

tests/test-basic-user-only.sh
dbnicholson: The problem is actually because both `setup_exampleos_com` in `libtest.sh` and `basic-test.sh` end up creating `bare-user` repos. The initial part of this one does succeed.

dbnicholson:
Yeah, I was just testing on my host with `gnome-desktop-testing-runner` so I could see if `test-core.js` succeeded. Indeed, setting `TMPDIR=/var/tmp` while running `gnome-desktop-testing-runner` does make the tests run correctly.

cgwalters:
I'd like to land the fix-installed-tests stuff first; broke that out into https://github.com/ostreedev/ostree/pull/837

tests/test-basic-user-only.sh
cgwalters: OK, let's fix that rather than skip, I'd like to keep as broad test coverage of flatpak as possible.

dbnicholson:
It looks like #837 is snowballing into something bigger. Should I remove all the tests changes and just leave the first commit here? I did test the change manually as well as through the installed tests and it does do the right thing.

cgwalters:
Yep, that's fine by me.

dbnicholson:
OK, PR updated.

cgwalters:
@rh-atomic-bot r+ ef2f5fb

PR #819, opened 2017-04-28
Rework OstreeAsyncProgress API to use GVariant

pwithnall:
I spent a little while today looking at different progress reporting APIs, and it turns out that despite @cgwalters’ reservations, `OstreeAsyncProgress` is pretty much ideal. I’ve pasted my analysis at the bottom of this PR.

This PR adds a few missing features to `OstreeAsyncProgress` which will allow it to be used by the LAN/USB work (PR #782) to give structured information about each ref it’s pulling, for example, by passing `GVariant` dicts in the progress information.

The PR also improves the atomicity of accesses to an `OstreeAsyncProgress` instance, which fixes some potential bugs in reporting inconsistent pull progress.

---

Options for progress reporting:
 * `EvPrintOperation`: have an object for the entire operation, report progress via its properties or methods
   - \+ Nicely encapsulates the lifetime of the operation (can be queried after the operation finishes)
   - − Properties aren’t threadsafe
   - \+ Easily extensible
   - \+ Typesafe
   - \+ Supports notification (notify signal)
   - − Not a standard interface (a new object has to be implemented for each operation)
   - − Queries are not atomic (different properties queried at different times)
 * `FlatpakProgressCallback`, `GFileProgressCallback`: caller provides a callback which has a fixed set of parameters
   - − Doesn’t encapsulate the lifetime of the operation (can’t be queried afterwards)
   - \+ Threadsafe
   - − Not extensible
   - \+ Supports notification (the callback itself)
   - \+ State can be queried atomically (all properties are returned atomically)
   - \+ Standard interface
   - \+ Typesafe
 * `OstreeAsyncProgress`: caller provides an object which has a notification callback associated with it
   - \+ Fairly well encapsulates the lifetime of the operation (can be queried after the operation finishes)
   - \+ Threadsafe
   - \+ Easily extensible (though more complex types are a bit of a pain)
   - − Not typesafe (relies on documentation), but subclasses can add type safety
   - \+ Supports notification (callback)
   - \+ Standard interface
   - − Queries are not atomic (different properties queried at different times), but this could be fixed easily

PR #835, opened 2017-05-05
Fix a few problems with notifications on OstreeAsyncProgress

pwithnall:
As spotted by @dbnicholson in PR #819.

### DIFF
--- a/src/libostree/ostree-async-progress.c
+++ b/src/libostree/ostree-async-progress.c
@@ -331,6 +331,8 @@ ostree_async_progress_set (OstreeAsyncProgress *self,
   if (self->dead)
     goto out;
 
+  changed = FALSE;
+
   va_start (ap, self);
 
   for (key = va_arg (ap, const char *), format_string = va_arg (ap, const char *);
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-async-progress.c
```diff
@@ -331,6 +331,8 @@ ostree_async_progress_set (OstreeAsyncProgress *self,
   if (self->dead)
     goto out;
 
+  changed = FALSE;
```
cgwalters: Why not init this at variable decl time instead?

pwithnall: I personally associate the assignment here more closely with the block below than with the actual function. Put another way, if there was a second block after this one which also uses `changed` (separately), then I wouldn’t want to rely on initialisation at declaration time for it.

But I’m not really fussed either way. I’ll change it if you want.

cgwalters:
OK.

@rh-atomic-bot r+ 0d521ca

cgwalters:
@rh-atomic-bot r+ 0d521ca

dbnicholson:
This fixed the couple issues I was having. Thanks!

PR #836, opened 2017-05-05
tree-wide: Convert to using autoptr(GString) vs g_string_free(...,TRUE)

cgwalters:
If we're freeing the segment, it's basically always better to use
`autoptr()`.  Fewer lines, more reliable, etc.

Noticed an instance of this in the pull code while reviewing a different PR,
decided to do a grep for it and fix it tree wide.


cgwalters:

cgwalters:

pwithnall:
@rh-atomic-bot r+ 515fea38c29f6023243972063a3e90da9ffbf37f

PR #837, opened 2017-05-08
Ci installed

cgwalters:


cgwalters:

dbnicholson:
This doesn't have the xattr fixes from 1037c03. Is that intentional? It sounds like you don't want to skip the tests, but nothing has been done to keep them from requiring xattrs.

dbnicholson:
BTW, is there any documentation on how to interact with rh-atomic-bot? I can look at some older PRs and copy/paste, but I don't really know what the commands do.

cgwalters:
Yep, https://github.com/jlebon/redhat-ci#interacting-with-redhat-ci and https://homu-projectatomic-ci-infra.e8ca.engint.openshiftapps.com/

cgwalters:
Right, not doing the xattrs here yet.

cgwalters:
Although I have some more pending commits here...let's hold off on merging.

cgwalters:
OK, bigger reworking :building_construction: of things :arrow_up:

Makefile-tests.am
```diff
@@ -287,6 +293,17 @@ tests/%-symlink-stamp: % Makefile
 	ln -sf "$${real_bin}" tests/$*; \
 	touch $@
 
+# See above comment on binding the tests to be either installed or not.
+if ENABLE_INSTALLED_TESTS
+installed_test_scripts = $(_installed_or_uninstalled_test_scripts)
+installed_test_programs = $(_installed_or_uninstalled_test_programs)
+check-local:
+	echo "NOTE: Installed tests are enabled; to run them, make install, then: gnome-desktop-testing-runner -p 0 libostree/"
```
dbnicholson: Sorry, I don't think this is right for a couple reasons.

1. The installed and in-tree tests are not replacements for each other. The former can run tests that the latter can't while the latter allows normal developer testing. I'd hate to have to build twice so I could run `make check` and `gnome-desktop-testing-runner` so I could run the tests not available in-tree.

2. If you want your CI system to not run tests twice, just don't run `make check`. Only run the installed tests.

cgwalters: Depends on your value of "normal developer testing" :smile:.  I use [atomicws](https://pagure.io/atomic-ws) where installing devel tools on the host is an anti-pattern; or to say it positively, I do build/tests in containers, and hence I `sudo make install` for ostree and other tools, and therefore installed tests work there too.

Anyways, I can add `--enable-installed-tests=exclusive` or so. 

I addressed your point `2)` in the commit message.  Some tests we have are *only* uninstalled and I don't have a sane way to run *just* those without introducing new infrastructure.

dbnicholson: I understand you're all in on containers, they definitely have appeal. But I doubt I'm the only person who uses `make check`. Otherwise, what was the purpose of #337? I personally would find that maintaining a container infrastructure just so I could make changes to ostree to be an impediment.

Does this change address the issue that some tests can only be run uninstalled? It doesn't seem like it since `make check` becomes a no-op if you build with `--enable-installed-tests` like you're trying to accomplish.

Anyways, just my 2 cents.

cgwalters: It's not a no op - again as I said in the commit message, e.g. `test-symbols.sh` is *always* uninstalled and we want to run it.

Your feedback is valuable and accurate!  I'm testing an updated version to add `--enable-installed-tests=exclusive` for use in CI.

cgwalters:
Ok, updated :arrow_up:

Makefile-tests.am
dbnicholson: OK, I got confused by `check-local`. I thought I remembered that it overrode `check`, but it's actually supplemental - http://www.gnu.org/software/automake/manual/html_node/Extending.html#Extending.

Makefile-tests.am
```diff
@@ -287,6 +294,17 @@ tests/%-symlink-stamp: % Makefile
 	ln -sf "$${real_bin}" tests/$*; \
 	touch $@
 
+# See above comment on binding the tests to be either installed or not.
+if ENABLE_INSTALLED_TESTS_EXCLUSIVE
+installed_test_scripts = $(_installed_or_uninstalled_test_scripts)
```
dbnicholson: I think we can probably assume no one is going to try to dist with --enable-installed-tests=exclusive, but this should probably be `dist_installed_test_scripts`.

cgwalters:
Well, we seem to have blown through our Travis budget for today or something.  Ah, actually looks like [there are issues](https://www.traviscistatus.com/) right now.

I think this patch should be good to go.

dbnicholson:
Yeah, seems OK to me.

@rh-atomic-bot r+ 0f38379

cgwalters:
Also, good catch on the `dist_` thing since it's actually quite likely I would do that.

PR #838, opened 2017-05-08
tests: Fix regex escaping in test-summary-view.sh

pwithnall:
There were some regex special characters in the pattern strings, which I
think were causing the test to fail on some Travis builds due to using
an invalid regex.

Fix that by matching using fixed strings instead. We don’t need regexes
here.

### DIFF
--- a/tests/test-summary-view.sh
+++ b/tests/test-summary-view.sh
@@ -47,20 +47,18 @@ ${CMD_PREFIX} ostree --repo=repo pull --mirror origin
 
 # Check the summary file exists in the checkout, and can be viewed.
 assert_has_file repo/summary
-output=$(${OSTREE} summary --view)
-echo "$output" | sed -e 's/^/# /'
-echo "$output" | grep --quiet --no-messages "* main"
-echo "$output" | grep --quiet --no-messages "* other"
-echo "$output" | grep --quiet --no-messages "ostree.summary.last-modified"
-echo "$output" | grep --quiet --no-messages "Static Deltas (ostree.static-deltas): {}"
+${OSTREE} summary --view > summary.txt
+assert_file_has_content_literal summary.txt "* main"
+assert_file_has_content_literal summary.txt "* other"
+assert_file_has_content_literal summary.txt "ostree.summary.last-modified"
+assert_file_has_content_literal summary.txt "Static Deltas (ostree.static-deltas): {}"
 echo "ok view summary"
 
 # Check the summary can be viewed raw too.
-raw_output=$(${OSTREE} summary --view --raw)
-echo "$raw_output" | sed -e 's/^/# /'
-echo "$raw_output" | grep --quiet --no-messages "('main', ("
-echo "$raw_output" | grep --quiet --no-messages "('other', ("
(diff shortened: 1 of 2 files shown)


tests/test-summary-view.sh
```diff
@@ -49,18 +49,18 @@ ${CMD_PREFIX} ostree --repo=repo pull --mirror origin
...
 output=$(${OSTREE} summary --view)
 echo "$output" | sed -e 's/^/# /'
-echo "$output" | grep --quiet --no-messages "* main"
-echo "$output" | grep --quiet --no-messages "* other"
-echo "$output" | grep --quiet --no-messages "ostree.summary.last-modified"
-echo "$output" | grep --quiet --no-messages "Static Deltas (ostree.static-deltas): {}"
+echo "$output" | grep --quiet --no-messages --fixed-strings "* main"
+echo "$output" | grep --quiet --no-messages --fixed-strings "* other"
+echo "$output" | grep --quiet --no-messages --fixed-strings "ostree.summary.last-modified"
+echo "$output" | grep --quiet --no-messages --fixed-strings "Static Deltas (ostree.static-deltas): {}"
```
cgwalters: I think this would be more nicely written with `libtest-core.sh` like:
```
$OSTREE summary -- view > out.txt
for value in "* main" "* other" ...; do
  assert_file_has_content_literal out.txt $value
done
```

(After adding `assert_file_has_content_literal` which would be a variant of `assert_file_has_content`)

cgwalters: Among other nice things, `libtest-core.sh`'s `assert_file_has_content` will print the non-matching value on failure which is what one wants for CI tests.

pwithnall: Ah, a magic bash library. Updated.

tests/test-summary-view.sh
```diff
@@ -47,20 +47,18 @@ ${CMD_PREFIX} ostree --repo=repo pull --mirror origin
...
 # Check the summary file exists in the checkout, and can be viewed.
 assert_has_file repo/summary
-output=$(${OSTREE} summary --view)
-echo "$output" | sed -e 's/^/# /'
-echo "$output" | grep --quiet --no-messages "* main"
-echo "$output" | grep --quiet --no-messages "* other"
-echo "$output" | grep --quiet --no-messages "ostree.summary.last-modified"
-echo "$output" | grep --quiet --no-messages "Static Deltas (ostree.static-deltas): {}"
+${OSTREE} summary --view > summary.txt
+assert_file_has_content summary.txt "* main"
```
cgwalters: Yep except you'll need to add `assert_file_has_content_`*literal* which doesn't exist at the moment.

pwithnall: Multitasking is obviously not working for me today. I’ll add that shortly.

pwithnall: OK, updated.

cgwalters:
Thanks!

@rh-atomic-bot r+  1a0d0bc

cgwalters:
@rh-atomic-bot retry

PR #839, opened 2017-05-08
pull: Allow additional HTTP headers for summary fetch

dbnicholson:
Read the http-headers (a(ss)) option in
ostree_repo_remote_fetch_summary_with_options like
ostree_repo_pull_with_options and add the headers to the fetcher. This
allows things like providing additional authorization headers to the
HTTP requests.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2460,10 +2460,14 @@ repo_remote_fetch_summary (OstreeRepo    *self,
   gboolean ret = FALSE;
   gboolean from_cache = FALSE;
   const char *url_override = NULL;
+  g_autoptr(GVariant) extra_headers = NULL;
   g_autoptr(GPtrArray) mirrorlist = NULL;
 
   if (options)
-    (void) g_variant_lookup (options, "override-url", "&s", &url_override);
+    {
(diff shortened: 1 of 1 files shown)


cgwalters:
@rh-atomic-bot r+ 981ba1d

mohanen:
I need to set some authorization headers 

how to use this extra_headers

dbnicholson:
Extra headers is an `a(ss)` GVariant, so it's just an array of `(header, value)` tuples. In python, the way I use it is like this:
```
    summary_opts = GLib.Variant('a{sv}', {
        'override-url': GLib.Variant('s', url),
        'http-headers': GLib.Variant('a(ss)', headers),
    })
    _, data, _ = repo.remote_fetch_summary_with_options(REMOTE_NAME, summary_opts)
```
In my case, `headers` is just a single element list with an `('Authorization', basic_auth)` tuple. `basic_auth` is `Basic ` + base64 encoded user:password in the normal HTTP way.

mohanen:
@dbnicholson  thank you , I am kinda new to glib. 

From your comment, I can understand a(ss) GVariant is array of key value pairs.   

I am using it in command line , is there anyway to give header in command line or do i have to edit the source.

PR #840, opened 2017-05-08
    tests: Migrate test-pull-many.sh to installed on FAH     

cgwalters:
`test-pull-many.sh` is was just too slow to be a unit test.  Generating
a bunch of files via shell is slow, the delta generation is slow, etc.
Every developer doesn't need to run it every time.
    
Somewhat address this by converting it into our installed test framework, which
moves it out of the developer fast paths.  Another advantage to this is
that we can simply reuse the FAH tree content rather than synthesizing
new bits each time.


cgwalters:
Hooray.  This should be good to go.

tests/installed/libinsttest.sh
```diff
@@ -20,6 +20,49 @@
...
+            if ! cmp ${test_tmpdir}/httpd-output.tmp ${test_tmpdir}/httpd-port 1>/dev/null; then
+                # If so, we've successfully extracted the port
+                break
+            fi
+        fi
+        sleep 0.1
+    done
+    port=$(cat ${test_tmpdir}/httpd-port)
+    echo "http://127.0.0.1:${port}" > ${test_tmpdir}/httpd-address
+    echo "$child_pid" > ${test_tmpdir}/httpd-pid
```
jlebon: This isn't torn down on exit, is it? I might be missing some context here, but are these installed tests meant to run on throwaway VMs once? (E.g. the redhat-ci path). Or can devs expect to be able to reuse the same local VM across runs?

cgwalters: Yeah, it's not torn down.  Broadly speaking installed tests that run as root should probably be assumed to be destructive, and hence we don't do VM reuse.  But as a convenience we do try to support re-running the tests to help debug in the local scenario.

Basically I did make a change to `rm repo -rf` at the start for this, but I'm not sure it's worth trying to invest more in this versus just having the local dev path use a VM-snapshotting framework like vagrant.

jlebon:
This is nice. I especially like the potential for better SELinux coverage.
@rh-atomic-bot r+ a4c5127

PR #843, opened 2017-05-09
ci: Move travis scripts from tests/ → ci/

cgwalters:
I think tests/ should be just that, ci/ is separate.  Also rename
the files to include "travis" since that's what we use them
for right now.


jlebon:
@rh-atomic-bot r+ 011fb78

PR #844, opened 2017-05-09
diff: Port some to new code style

cgwalters:
Continuing to chip away at this.  Using `g_file_enumerator_iterate()`
here helps notably.

I started on the much bigger `ostree_diff_dirs_with_options()` but
it's a lot messier - for later.


jlebon:
@rh-atomic-bot r+ 9b94b09

PR #845, opened 2017-05-09
sysroot: More porting to new code style

cgwalters:
This isn't all of this file yet, just doing another chunk.

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -587,60 +587,54 @@ parse_deployment (OstreeSysroot       *self,
                   GCancellable        *cancellable,
                   GError             **error)
 {
-  gboolean ret = FALSE;
-  const char *relative_boot_link;
-  glnx_unref_object OstreeDeployment *ret_deployment = NULL;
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot.c
```diff
@@ -1340,77 +1292,49 @@ ostree_sysroot_init_osname (OstreeSysroot       *self,
...
   if (!glnx_opendirat (self->sysroot_fd, deploydir, TRUE, &dfd, error))
-    goto out;
+    return FALSE;
 
   if (mkdirat (dfd, "var", 0777) < 0)
-    {
-      glnx_set_prefix_error_from_errno (error, "Creating %s", "var");
-      goto out;
-    }
+    return glnx_throw_errno_prefix (error, "Creating %s", "var");
```
jlebon: Nit: I think the `glnx_throw_*` functions don't suffer from that same arity issue, right? Though I'm guessing you probably used a macro in your editor to change these, so that's fine!

jlebon:
@rh-atomic-bot r+ 5ce745e

PR #847, opened 2017-05-10
lib/prune: Complete porting to new code style

cgwalters:
Only non-mechanical bit here was creating a local autoptr for a bit
where we'd previously done an unref for a struct member.


jlebon:
@rh-atomic-bot r+ 39b38bd

PR #850, opened 2017-05-11
checkout: Plug a memleak of the state stringbuf

cgwalters:
A struct without a cleanup macro is a struct likely to leak.


jlebon:
@rh-atomic-bot r+ 59acb91

PR #856, opened 2017-05-12
remount: Drop support for auto-tmpfs-on-var; use systemd.volatile=state

cgwalters:
In current systemd, there is:
[systemd-volatile-root](https://www.freedesktop.org/software/systemd/man/systemd-volatile-root.service.html)
which was introduced by [this commit](https://github.com/systemd/systemd/commit/91214a37ef4eb8042d2598aa89bae52b410d11a7).

I'd like to make further changes to how we handle `/var`, and I don't
want to reason about the interaction of our "tmpfs var" with too many
other things.

The comment about having "all /var handling in one place" was always inaccurate
given that we rely on systemd for mounting. And in general, I don't want to
duplicate too many things systemd does - it does them well, documents them, etc.

As far as I know, it was basically just Owen who was using this for the GNOME
hardware testing effort, and I'm sure he could easily switch over to
`systemd.volatile=state`.

### DIFF
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -38,24 +38,6 @@
 
 #include "ostree-mount-util.h"
 
-/* Having a writeable /var is necessary for full system functioning.
- * If /var isn't writeable, we mount tmpfs over it. While this is
- * somewhat outside of ostree's scope, having all /var twiddling
- * in one place will make future maintenance easier.
- */
-static void
-maybe_mount_tmpfs_on_var (void)
-{
-  if (!path_is_on_readonly_fs ("/var"))
-    return;
-
-  if (umount ("/var") < 0 && errno != EINVAL)
-    warn ("failed to unmount /var prior to mounting tmpfs, mounting over");
-
-  if (mount ("tmpfs", "/var", "tmpfs", 0, NULL) < 0)
-    err (EXIT_FAILURE, "failed to mount tmpfs on /var");
-}
-
 int
 main(int argc, char *argv[])
 {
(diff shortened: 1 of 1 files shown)


cgwalters:
I pinged @owtaylor for review.

owtaylor:
It certainly looks like systemd.volatile=state would have the same effect. It's hard for me to remember enough about what I was doing to be sure that systemd.volatile=state, but since it's not a current usage, that certainly shouldn't block anything! As far as I'm concerned you can go ahead.

(of course, someone else could be implicitly relying on this - without realizing it - so it's not something I'd push to a stable release without letting people have time to test.)

cgwalters:
@rh-atomic-bot delegate=owtaylor

Yeah, someone could be using it - but if they are, it's not like they're going to do an OS update and suddenly things break, because the model is inherently not using inplace updates.  They're going to be generating new images and booting those, and finding out things are broken.  Hopefully they'll come here.

For Fedora Atomic Host we did have a project at one point to do a "PXE to live" variant which would make use of this potentially, but it isn't active right now since our primary use cases are stateful-ish systems.

So my plan indeed is to get this in the next release, and if someone complains and for some reason has an ancient systemd we can revisit.

owtaylor:
Sounds good.

cgwalters:
@owtaylor You can approve this PR - see https://github.com/jlebon/redhat-ci#interacting-with-redhat-ci and https://homu-projectatomic-ci-infra.e8ca.engint.openshiftapps.com/

owtaylor:
@rh-atomic-bot r+

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot r=owtaylor c127838

PR #857, opened 2017-05-12
libglnx: bump and use new helper methods

jlebon:
Update submodule: libglnx


cgwalters:
@rh-atomic-bot r+  	86bae24

PR #859, opened 2017-05-15
 Switch to using a systemd generator for /var

cgwalters:
If one wants to set up a mount for `/var` in `/etc/fstab`, it
won't be mounted since `ostree-prepare-root` set up a bind mount for
`/var` to `/sysroot/ostree/$stateroot/var`, and systemd will take
the already extant mount over what's in `/etc/fstab`.

There are a few options to fix this, but what I settled on is parsing
`/etc/fstab` in a generator (exactly like `systemd-fstab-generator` does),
except here we look for an explicit mount for `/var`, and if one *isn't* found,
synthesize the default ostree mount to the stateroot. Another nice property is
that if an admin creates a `var.mount` unit in `/etc` for example, that will
also override our mount.

Note that today ostree doesn't hard depend on systemd, so this behavior only
kicks in if we're built with systemd *and* libmount support (for parsing
`/etc/fstab`).  I didn't really test that case though.

Initially I started writing this as a "pure libc" program, but at one point
decided to use `libostree.so` to find the booted deployment. That didn't work
out because `/boot` wasn't necessarily mounted and hence we couldn't find the
bootloader config. A leftover artifact from this is that the generator code
calls into libostree via the "cmd private" infrastructure. But it's an easy way
to share code, and doesn't hurt.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -378,10 +378,7 @@ dnl We have separate checks for libsystemd and the unit dir for historical reaso
 PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd], [have_libsystemd=yes], [have_libsystemd=no])
 AM_CONDITIONAL(BUILDOPT_LIBSYSTEMD, test x$have_libsystemd = xyes)
 AM_COND_IF(BUILDOPT_LIBSYSTEMD,
-           AC_DEFINE([HAVE_LIBSYSTEMD], 1, [Define if we have libsystemd])
-           systemdsystemgeneratordir=$($PKG_CONFIG --variable=systemdsystemgeneratordir systemd)
-           AC_SUBST(systemdsystemgeneratordir)
-           )
+           AC_DEFINE([HAVE_LIBSYSTEMD], 1, [Define if we have libsystemd]))
 
 AS_IF([test "x$have_libsystemd" = "xyes"], [
   with_systemd=yes
@@ -391,6 +388,13 @@ AS_IF([test "x$have_libsystemd" = "xyes"], [
               [with_systemdsystemunitdir=$($PKG_CONFIG --variable=systemdsystemunitdir systemd)])
   AS_IF([test "x$with_systemdsystemunitdir" != "xno"], [
     AC_SUBST([systemdsystemunitdir], [$with_systemdsystemunitdir])
+  ])
+  AC_ARG_WITH([systemdsystemgeneratordir],
+              AS_HELP_STRING([--with-systemdsystemgeneratordir=DIR], [Directory for systemd generators]),
+              [],
+              [with_systemdsystemgeneratordir=$($PKG_CONFIG --variable=systemdsystemgeneratordir systemd)])
+  AS_IF([test "x$with_systemdsystemgeneratordir" != "xno"], [
+    AC_SUBST([systemdsystemgeneratordir], [$with_systemdsystemgeneratordir])
   ])
 ])
 AM_CONDITIONAL(BUILDOPT_SYSTEMD, test x$with_systemd = xyes)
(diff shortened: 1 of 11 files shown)


src/libostree/ostree-impl-system-generator.c
```diff
@@ -0,0 +1,218 @@
...
+   * Note that our unit doesn't run if systemd.volatile is enabled;
+   * see https://github.com/ostreedev/ostree/pull/856
+   */
+  if (!g_output_stream_printf (outstream, &bytes_written, cancellable, error,
+                               "##\n# Automatically generated by ostree-system-generator\n##\n\n"
+                               "[Unit]\n"
+                               "Documentation=man:ostree(1)\n"
+                               "ConditionKernelCommandLine=!systemd.volatile\n"
+                               /* We need /sysroot mounted writable first */
+                               "After=ostree-remount.service\n"
```
jlebon: How about a `SourcePath=/etc/fstab` here?

jlebon: Don't we also need a `Before=local-fs.target` here?

configure.ac
```diff
@@ -378,7 +378,10 @@ dnl We have separate checks for libsystemd and the unit dir for historical reaso
 PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd], [have_libsystemd=yes], [have_libsystemd=no])
 AM_CONDITIONAL(BUILDOPT_LIBSYSTEMD, test x$have_libsystemd = xyes)
 AM_COND_IF(BUILDOPT_LIBSYSTEMD,
-           AC_DEFINE([HAVE_LIBSYSTEMD], 1, [Define if we have libsystemd]))
+           AC_DEFINE([HAVE_LIBSYSTEMD], 1, [Define if we have libsystemd])
+           systemdsystemgeneratordir=$($PKG_CONFIG --variable=systemdsystemgeneratordir systemd)
+           AC_SUBST(systemdsystemgeneratordir)
+           )
```
jlebon: Can you provide a `--with-systemdsystemgeneratordir` option here? So that I can do unprivileged installs without completely turning off systemd. I can add it as a follow-up PR too.

src/switchroot/ostree-remount.c
```diff
@@ -92,6 +100,7 @@ main(int argc, char *argv[])
           if (errno != EINVAL)
             err (EXIT_FAILURE, "failed to remount %s", target);
         }
+      printf ("Remounted: %s\n", target);
```
cgwalters: We call `err` and abort the process if it fails.

src/libostree/ostree-impl-system-generator.c
cgwalters: I actually had that originally since I was copying systemd, but then I realized the source *isn't* `/etc/fstab`.  Right?

Will add the `Before=`.

src/switchroot/ostree-remount.c
jlebon: But it'll still print it if we get `EINVAL`, right? That can be misleading. (Heck, maybe we should even print a different msg in the `EINVAL` case).

src/libostree/ostree-impl-system-generator.c
jlebon: Oh right, got confused. Yeah, there's not really an applicable `SourcePath` here.

src/switchroot/ostree-remount.c
cgwalters: True, fixed.  :arrow_down:

jlebon:
Verified working with `/var` as an fstab mount and not!
@rh-atomic-bot r+ 102ef0d

PR #861, opened 2017-05-15
 Update libglnx, port to new GLnxTmpfile

cgwalters:
See the rationale in: GNOME/libglnx#46
Basically we should now start unlinking the tmpfile in the non-`O_TMPFILE`
case in more places in the code.

Parts of the commit path get uglier though, but a refactoring there is
a bit overdue; the code is overly complex.

Update submodule: libglnx

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -99,8 +99,7 @@ struct FetcherRequest {
   OstreeFetcherRequestFlags flags;
   gboolean is_membuf;
   GError *caught_write_error;
-  char *out_tmpfile;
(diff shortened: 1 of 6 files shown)


jlebon:
Looks like this needs a rebase.

cgwalters:
Yeah...I think given the complexity of the ostree port, my vote is actually to land the libglnx change, update flatpak, and then I'll break out some prep work here and we'll keep libglnx at the current version for a bit while the prep work processes.

cgwalters:
Prep in https://github.com/ostreedev/ostree/pull/871

cgwalters:
Obsoleted by https://github.com/ostreedev/ostree/pull/958

PR #862, opened 2017-05-16
lib/repo: Fix double close()

cgwalters:
Should probably change `_take_fd()` to take a pointer and set to `-1`
at some point.

Regression from 8d58ab1002cbc4a1ecafe3d1a80984f8a60f41e9


jlebon:
@rh-atomic-bot r+ 2e07e58

PR #863, opened 2017-05-16
 lib: Add "open dfd iter handling noent" helper, port tree-wide

cgwalters:
Follow up to a previous patch that addressed a double-close; I
realized we already had a helper for doing "open dfd iter, do nothing
if we get ENOENT".  Raise it to libotuil, and port all consumers.

### DIFF
--- a/src/libostree/ostree-repo-static-delta-core.c
+++ b/src/libostree/ostree-repo-static-delta-core.c
@@ -73,99 +73,88 @@ ostree_repo_list_static_delta_names (OstreeRepo                  *self,
                                      GCancellable                *cancellable,
                                      GError                     **error)
 {
-  g_autoptr(GPtrArray) ret_deltas = NULL;
-  glnx_fd_close int dfd = -1;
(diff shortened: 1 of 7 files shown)


src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -73,99 +73,88 @@ ostree_repo_list_static_delta_names (OstreeRepo                  *self,
...
+    return FALSE;
+  if (!exists)
     {
-      if (errno != ENOENT)
-        {
-          glnx_set_error_from_errno (error);
-          return FALSE;
-        }
+      ot_transfer_out_value (out_deltas, &ret_deltas);
+      return TRUE;
```
jlebon: For consistency, can you comment as early return here?

cgwalters: Done :arrow_down:

jlebon:
@rh-atomic-bot r+ 1e780da

PR #866, opened 2017-05-16
tree-wide: Add a few missing O_CLOEXEC

cgwalters:
I noticed an instance of this while working on https://github.com/ostreedev/ostree/pull/861
Which apparently I cargo-culted into the new system generator bits.
Let's break this out as a small concise change.


jlebon:
@rh-atomic-bot r+ 827db74

PR #868, opened 2017-05-17
build: Use cd $(srcdir) instead of `git -C`

cgwalters:
Since the version in CentOS is too old, and we get a spam of warnings, plus
things like detecting the git repo break.

Fixes: 50f73cbac35be97fd5895531e295d05dabaa8ed9


jlebon:
@rh-atomic-bot r+ d19a3ea

PR #870, opened 2017-05-17
lib/pull: Port some functions to new code style

cgwalters:
Porting a lot of this file would be hard since in many cases we do processing in
the `out:` section, so let's do what we can.


cgwalters:
Looks like Travis just flaked on one of the builders.

jlebon:
@rh-atomic-bot r+ e413b44

PR #865, opened 2017-05-16
Release 2017.6

cgwalters:
There's already a lot queued.  In particular this brings some API
additions that rpm-ostree depends on.

PR #871, opened 2017-05-17
 Add stub for new libglnx tmpfile API, port simpler callers to it

cgwalters:
It's hard right now to do a full port to the new libglnx tmpfile
API since there are complex cases in the commit path which deal
with symlinks as well.

Let's make things more gradual by introducing the important part (struct with
autocleanup) here in libotutil, port what we can. This will make a future
complete port easier.

### DIFF
--- a/src/libotutil/ot-fs-utils.c
+++ b/src/libotutil/ot-fs-utils.c
@@ -25,6 +25,61 @@
 #include <sys/xattr.h>
 #include <gio/gunixinputstream.h>
 
+/* Before https://github.com/GNOME/libglnx/commit/9929adc, the libglnx
+ * tmpfile API made it hard to clean up tmpfiles in failure cases.
+ * it's API breaking. Carry the fix here until we're ready to fully port.
+ */
+void
+ot_tmpfile_clear (OtTmpfile *tmpf)
+{
+  if (!tmpf->initialized)
+    return;
+  if (tmpf->fd == -1)
+    return;
+  (void) close (tmpf->fd);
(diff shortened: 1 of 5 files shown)


cgwalters:
This can come after https://github.com/ostreedev/ostree/pull/865

jlebon:
So looking at the codebase, it seems like there's only 3 other instances where that API is used, right? It feels like it would be simpler than just do it all in one PR (even if it takes time) rather than carry a wrapper API and freeze libglnx? Or do you feel like those other 3 instances are especially complex?

cgwalters:
The problem is the commit codepath which uses `temp_filename` for symlinks, among other things. And the commit code is...way too complex for its own good.  It needs its own series of preps I think.

src/libotutil/ot-fs-utils.c
```diff
@@ -25,6 +25,54 @@
 #include <sys/xattr.h>
 #include <gio/gunixinputstream.h>
 
+/* Stub out old libglnx API to new one */
```
jlebon: Can we expand on this a bit? E.g.

```
Wrapper around libglnx API, which doesn't clean up properly.
This was fixed in https://github.com/GNOME/libglnx/commit/9929adc, though
it's API breaking. Carry the fix here until we're ready to fully port.
```

?

cgwalters: Did that, also rebased on the new version of the API and cleaned up.  :arrow_down:

jlebon:
@rh-atomic-bot r+ e4b8282

PR #872, opened 2017-05-17
lib/deploy: Remove code to SELinux relabel /var

cgwalters:
The history here is...long.  Basically at one point when I was
working on this, Fedora Atomic wasn't built using Anaconda, but instead
some code I took from gnome-continuous which used libguestfs.
That approach is long since dead.

Today, Anaconda has an `80-setfilecons.ks` which takes care of much of `/var`.
That should work.  And in general, I think taking care of the initial
labeling for `/var` should be the job of the installer.

We can delete *lot* of ugly code here in the deploy path.  The `/var`
labeling is the only user since `/etc` switched to label-on-checkout
in: e8efd1c8dcaad8fbd3b05c400972d237406263e7


cgwalters:
Haven't fully tested this yet, but filing it so I don't forget.

cgwalters:
This needs https://github.com/rhinstaller/anaconda/pull/1138

jlebon:
Should we wait until the anaconda patch gets in and percolates down? Otherwise the continuous streams will have those paths mislabeled right?

cgwalters:
We don't build ISOs in the continuous streams today, so the initial labeling won't be changed.  But there's no hurry on this anyways.

jlebon:
I think we can get this in now, right?

PR #873, opened 2017-05-17
pull: complete detached meta fetch before scanning

jlebon:
If somehow a repo has gpg verification on but doesn't have signatures
present for the existing commit, ostree would error out if it needs to
scan the commit object (e.g. if there are no updates available).

An instance of this is currently happening in Fedora AH, in which
signatures are not shipped in the ISO due to filesystem restrictions.
Another possible scenario is if a content provider switches from not
signing commits to signing them; even if older commits are retroactively
signed, clients' local commit objects would error out if they needed
scanning.

This patch adds a check to ensure that we always attempt to fetch the
detached metadata and wait for its result (whether it exists or not)
before moving on to scan their corresponding commit objects.

See also: https://github.com/projectatomic/rpm-ostree/issues/630

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -80,6 +80,7 @@ typedef struct {
   GHashTable       *expected_commit_sizes; /* Maps commit checksum to known size */
   GHashTable       *commit_to_depth; /* Maps commit checksum maximum depth */
   GHashTable       *scanned_metadata; /* Maps object name to itself */
+  GHashTable       *fetched_detached_metadata; /* Set<checksum> */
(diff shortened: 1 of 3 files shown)


cgwalters:
This looks sane to me - let's  do a release first though?   I'll update https://github.com/ostreedev/ostree/pull/865

cgwalters:
(If you feel strongly though I could be convinced to review carefully, land this and wait a day or two for soak, then release)

jlebon:
Nah, I think when it comes to the pull code 🐙, it's better to play it extra safe.

cgwalters:
Nice!

@rh-atomic-bot r+  	0f569e3

dustymabe:
can we get a release out with this in it?

PR #874, opened 2017-05-18
Allow commits to mark refs as EOL, replaced by others

dsd:
A commit can now include a "ostree.endoflife-rebase" metadata key
pointing to a new ref.

When updating, the sysroot upgrader will see this and proceed to
pull and deploy the new ref instead. The origin file in the new
deployment will point to the new ref.

This functionality is planned to be used in Endless OS. We will create
a lesser tested branch for brand new, cutting edge hardware support,
and ship that on hardware platforms that require the latest drivers.
However, once our slower-moving official release is later updated to
support the new hardware, we will use this functionality to migrate
those bleeding-edge users over to the official release.

### DIFF
--- a/src/libostree/ostree-sysroot-upgrader.c
+++ b/src/libostree/ostree-sysroot-upgrader.c
@@ -578,7 +578,12 @@ ostree_sysroot_upgrader_pull_one_dir (OstreeSysrootUpgrader  *self,
         g_free (self->origin_ref);
         self->origin_ref = g_strdup(new_ref);
         g_free (origin_refspec);
-        origin_refspec = g_strconcat (self->origin_remote, ":", new_ref, NULL);
+
+        if (self->origin_remote)
+          origin_refspec = g_strconcat (self->origin_remote, ":", new_ref, NULL);
+        else
+          origin_refspec = g_strdup (new_ref);
+
         g_key_file_set_string (self->origin, "origin", "refspec", origin_refspec);
     }
 
--- a/src/libostree/ostree-sysroot-upgrader.c
+++ b/src/libostree/ostree-sysroot-upgrader.c
@@ -512,7 +512,7 @@ ostree_sysroot_upgrader_pull_one_dir (OstreeSysrootUpgrader  *self,
   char *refs_to_fetch[] = { NULL, NULL };
   const char *from_revision = NULL;
   g_autofree char *origin_refspec = NULL;
-  g_autofree char *new_revision;
+  g_autofree char *new_revision = NULL;
   g_autoptr(GVariant) new_variant = NULL;
   g_autoptr(GVariant) new_metadata = NULL;
   g_autoptr(GVariant) rebase = NULL;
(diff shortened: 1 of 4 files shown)


cgwalters:
bot, add author to whitelist

src/libostree/ostree-repo.c
```diff
@@ -3526,7 +3526,7 @@ ostree_repo_read_commit (OstreeRepo   *self,
 gboolean
 ostree_repo_pull (OstreeRepo               *self,
                   const char               *remote_name,
-                  char                    **refs_to_fetch,
+                  const char              **refs_to_fetch,
```
cgwalters: We can't do this with a stable API; people turn these sorts of warnings into errors for good reason.

In the "GLib-and-above" stack, we're not very consistent about this.  The strictly correct thing is more `const char * const*`, which is long and annoying to type.  It's used by e.g. `g_variant_new_strv()`.    But not for example `g_strdupv()`.

Anyways, again we can't change our API, so please rebase without this.

cgwalters: (And we just have to eat the pain of casts between `char**` and `const char * const*`).

src/libostree/ostree-sysroot-upgrader.c
```diff
@@ -512,6 +512,10 @@ ostree_sysroot_upgrader_pull_one_dir (OstreeSysrootUpgrader  *self,
   const char *refs_to_fetch[] = { NULL, NULL };
   const char *from_revision = NULL;
   g_autofree char *origin_refspec = NULL;
+  g_autofree char *new_revision;
```
cgwalters: Init everything that uses autoptr to `NULL` for refactoring safety.

src/libostree/ostree-sysroot-upgrader.c
```diff
@@ -541,6 +545,42 @@ ostree_sysroot_upgrader_pull_one_dir (OstreeSysrootUpgrader  *self,
...
+  if (!ostree_repo_load_variant (repo,
+                                 OSTREE_OBJECT_TYPE_COMMIT,
+                                 new_revision,
+                                 &new_variant,
+                                 error))
+    return FALSE;
+
+  g_variant_get_child (new_variant, 0, "@a{sv}", &new_metadata);
+  rebase = g_variant_lookup_value (new_metadata, "ostree.endoflife-rebase", G_VARIANT_TYPE_STRING);
+  if (rebase) {
```
cgwalters: Braces on newline (GNU style).

cgwalters:
Code seems reasonable offhand.  One note to self is that `rpm-ostree` has a re-implementation of the upgrader, so we'd need to clone this logic there.

dsd:
I rebased to remove the char API change, and pushed a new commit also addressing the other review comments.

src/libostree/ostree-sysroot-upgrader.c
```diff
@@ -541,6 +545,43 @@ ostree_sysroot_upgrader_pull_one_dir (OstreeSysrootUpgrader  *self,
...
+          if (!ostree_repo_pull_one_dir (repo, self->origin_remote, dir_to_pull, refs_to_fetch,
+                                         flags, progress, cancellable, error))
+            return FALSE;
+        }
+
+        /* Use the new ref for the rest of the update process */
+        g_free (self->origin_ref);
+        self->origin_ref = g_strdup(new_ref);
+        g_free (origin_refspec);
+        origin_refspec = g_strconcat (self->origin_remote, ":", new_ref, NULL);
```
jlebon: We have to account for local branches here too, right?

E.g. something like

```c
if (self->origin_remote)
  origin_refspec = g_strconcat (self->origin_remote, ":", new_ref, NULL);
else
  origin_refspec = g_strdup (new_ref);
```

?

cgwalters: Or possibly just do the whole block under `if (rebase && self->origin_remote)`?  I can't think of a use case for *local* refs being redirected.  But there's also an argument that we should be symmetric in our handling of local/remote refs - your change is definitely right from that perspective.

jlebon: The symmetry argument definitely appeals to me, yeah. Even if only for making it easier to play with and test locally.

dsd:
I pushed another fixup commit which adds @jlebon's suggestion for handling local branches.

cgwalters:
Thanks!

@rh-atomic-bot r+  52858b5

PR #875, opened 2017-05-19
A few tweaks to the OstreeRemote API

pwithnall:
These are some more groundwork for https://github.com/pwithnall/ostree/tree/lan-and-usb. They’re mostly internal changes, apart from the addition of a new `ostree_remote_get_name()` public API (currently hidden behind `--enable--experimental-api`).

### DIFF
--- a/.redhat-ci.yml
+++ b/.redhat-ci.yml
@@ -79,6 +79,26 @@ tests:
 ---
 
 inherit: true
+
+context: f25-experimental-api
+
+build:
+    config-opts: >
+      --prefix=/usr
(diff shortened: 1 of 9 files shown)


cgwalters:
This all looks sane to me; sorry about breaking the build, I thought I'd tested it but apparently not.  I've pushed an additional commit here to add a CI context for this.

cgwalters:
Ah hah, and it failed.  Should be fixed now.

cgwalters:
@rh-atomic-bot r+ 6cf09c6

PR #877, opened 2017-05-19
tests/libtest-core: Copy rpm-ostree changes, clean up

cgwalters:
I want to keep this a "pure copy-able" file into various projects
like rpm-ostree, bwrap, and flatpak.  Pull in changes from rpm-ostree
to prep for that.

While we have the patient open, dedup the code for file matching a bit.

### DIFF
--- a/tests/libtest-core.sh
+++ b/tests/libtest-core.sh
@@ -1,4 +1,8 @@
-# Core source library for shell script tests
+# Core source library for shell script tests; this
+# file is intended to be the canonical source, which at
+# is copied at least into:
+#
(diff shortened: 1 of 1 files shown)


tests/libtest-core.sh
```diff
@@ -1,4 +1,8 @@
-# Core source library for shell script tests
+# Core source library for shell script tests; this
+# file is intended to be the canonical source, which at
+# is copied at least into:
```
jlebon: Nit: looks like you've got an extra "at" there. 🔡

jlebon:
@rh-atomic-bot r+ f0985e3

PR #880, opened 2017-05-23
build: Use AM_TESTS_ENVIRONMENT rather than TESTS_ENVIRONMENT

pwithnall:
TESTS_ENVIRONMENT is reserved for the user to be able to set when
running the tests. AM_TESTS_ENVIRONMENT is for the tests’ Makefile to
set itself.

https://www.gnu.org/software/automake/manual/html_node/Scripts_002dbased-Testsuites.html

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -27,7 +27,7 @@ EXTRA_DIST += \
 # We should probably consider flipping the default for DEBUG.  Also,
 # include the builddir in $PATH so we find our just-built ostree
 # binary.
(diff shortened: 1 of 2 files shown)


pwithnall:
I’ve uploaded a patch for a similar issue [in GLib](https://bugzilla.gnome.org/show_bug.cgi?id=782996).

pwithnall:
And thus concludes today’s 5-minute-anally-retentive-patch-for-something-which-doesn’t-matter-at-all.

cgwalters:
@rh-atomic-bot r+  	ecf8ea8

PR #881, opened 2017-05-23
 repo/commit: Split up metadata/content commit paths

cgwalters:
There was a lot of conditionals inside `write_object()` differentating
between metadata/content, and then for content, on the different repo
types.  Further, in the metadata path since the logic is simpler, can
present a non-streaming API, and further use `OtTmpfile`, etc.

Splitting them up helps drop a lot of conditionals. We introduce a small
`CleanupUnlinkat` that allows us to fully convert to the new code style in both
functions.

This itself is still prep for fully switching to `GLnxTmpfile`.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -836,31 +836,44 @@ write_metadata_object (OstreeRepo         *self,
    * tempfile until we compute the checksum. Some metadata like dirmeta is
    * commonly duplicated, and computing the checksum is going to be cheaper than
    * making a tempfile.
+   *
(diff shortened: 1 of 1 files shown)


cgwalters:
Prep for https://github.com/ostreedev/ostree/pull/861

cgwalters:
Medium risk, needs some careful review.  I'm likely to add some test cases here too to cover some of the API.  May split out some of the prep inside this as a separate PR.

jlebon:
Looks good! So I think this is sane. Just want to take a second quick look in the terminal.

jlebon:
@rh-atomic-bot r+ 4ca9ff2

Issue #802, opened 2017-04-20
Support Transfer-Encoding: gzip when downloading summaries


alexlarsson:
The summary can be pretty large, and it is not by itself compressed. We should indicate in the headers when getting this that we accept e.g. Transfer-Encoding: gzip, as that would help with the download size.

This isn't true for the rest of the files though, as they are generally already compressed.

PR #882, opened 2017-05-24
fetcher: Send Accept-Encoding: gzip when downloading summary

alexlarsson:
The summary file can get large, but it compresses well (something
which is not true of other files in the ostree repo which are
already compressed). By sending Accept-Encoding: gzip (and
handling the compressed results) we send a lot less data.

I set up the flathub repo (http://flathub.org/repo) to enable
gzip for the summary file (only), and the result is that the
331514 byte large summary was transferred in 122889 bytes.
On my (fast) network this decreased the time i took to do
"flatpak remote-ls flathub" by about 100msec.

This fixes https://github.com/ostreedev/ostree/issues/802

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -770,6 +770,9 @@ initiate_next_curl_request (FetcherRequest *req,
       curl_easy_setopt (req->easy, CURLOPT_SSLKEY, self->tls_client_key_path);
     }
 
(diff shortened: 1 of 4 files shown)


alexlarsson:
Oh, i tested both fetcher implementations by strace:ing the remote-ls flathub call and made sure they both send Accept-Encoding header on the summary request, and that the response was gzip:ed and everything worked.

It seems hard to have a real test of this without having a more complex http server in the tests.

cgwalters:
@rh-atomic-bot r+ 52ba5cb

PR #885, opened 2017-05-25
PAPR: migrate to the new name

jlebon:
The redhat-ci service has been renamed to PAPR. Previous values are
still supported though should be considered deprecated.


cgwalters:
@rh-atomic-bot r+ 5659da2

PR #820, opened 2017-04-28
Repo new at

cgwalters:
Just something I was hacking on on the side.  Turned out to be even uglier than I thought...there's a huge incestuous relationship between `OstreeRepo` and `OstreeSysroot` around the concept of a "system" repo and where I stalled out is in the remote reading/writing code.

May peel off the first patch though, since I think it's sane.

PR #886, opened 2017-05-25
[prep] ostree_repo_new_at()

cgwalters:
Prep for https://github.com/ostreedev/ostree/pull/820

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -729,6 +729,20 @@ ostree_sysroot_load (OstreeSysroot  *self,
   return ostree_sysroot_load_if_changed (self, NULL, cancellable, error);
 }
 
+static gboolean
+ensure_repo_opened (OstreeSysroot  *self,
+                    GError        **error)
+{
+  if (self->repo_opened)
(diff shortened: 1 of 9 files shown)


cgwalters:
This one just flaked, should be good to go.

src/libostree/ostree-sysroot.c
```diff
@@ -729,6 +729,20 @@ ostree_sysroot_load (OstreeSysroot  *self,
...
+static gboolean
+ensure_repo_opened (OstreeSysroot  *self,
+                    GError        **error)
+{
+  if (self->repo_opened)
+    return TRUE;
+  if (!ostree_repo_open (self->repo, NULL, error))
+    return FALSE;
+  /* And flag it as a system repo */
+  self->repo->is_system = TRUE;
```
jlebon: This should already be `TRUE` at the time the sysroot was constructed, right?

cgwalters: Yeah, I removed this.  It'll come back though in the later patches when this bit moves around.

jlebon:
@rh-atomic-bot r+ 5831800

PR #890, opened 2017-05-26
tree-wide: Add+run spatch to use glnx_throw()

cgwalters:
I had to run a sed job to add whitespace after, but otherwise this was easy.

### DIFF
--- a/coccinelle/new-error.cocci
+++ b/coccinelle/new-error.cocci
@@ -17,4 +17,3 @@ expression p, q;
 - g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED, p, q);
 - return FALSE;
 + return glnx_throw (error, p, q);
-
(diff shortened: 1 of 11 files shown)


coccinelle/new-error.cocci
```diff
@@ -0,0 +1,19 @@
+// Conversion for G_IO_ERROR_FAILED that could be glnx_throw()
+@@
+expression p;
+@@
+- g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED, p);
+- return FALSE;
++ return glnx_throw (error, "%s", p);
```
jlebon: Cool! Out of interest, would we be able to match the braces as well?

cgwalters: I tried but can't get it to work.  I don't understand why honestly.  It'd be nice, but I struggle to balance the time learning Coccinelle vs the benefit.

jlebon:
@rh-atomic-bot r+ e67af37

PR #891, opened 2017-05-26
cmd: Use autoptr for GKeyFile

cgwalters:
Prep for code style conversion.


jlebon:
@rh-atomic-bot r+ bb95fd0

PR #895, opened 2017-05-30
lib/util: Some style conversion

cgwalters:
I saw a few instances of `glnx_set_error_from_errno() + return FALSE`,
and fixed them and did a bit of style conversion.


jlebon:
@rh-atomic-bot r+ bb04754

Issue #892, opened 2017-05-27
remote: sysroot path handling


cgwalters:
The ostree remote change code looks up "sysroot path" + `/etc/ostree/remotes.d`.  This actually only makes sense if the sysroot path is a deployment.

For backwards compat, we need to keep supporting this.  See Anaconda:
https://github.com/rhinstaller/anaconda/blob/bf6219635df7963ee1fbb09ca68bb836cfd150c6/pyanaconda/payload/rpmostreepayload.py#L147
https://github.com/rhinstaller/anaconda/blob/bf6219635df7963ee1fbb09ca68bb836cfd150c6/pyanaconda/payload/rpmostreepayload.py#L307

As well as: https://github.com/rhinstaller/anaconda/commit/9c4977d0e38bb564854136679bcfbfc921fdc862

I think we need to strengthen the ostree remote → sysroot bridging to clearly distinguish between whether the sysroot is really a physical root or a deployment.  Possibly, we could key off whether or not `/etc/ostree/remotes.d` already exists?

This is necessary for https://github.com/ostreedev/ostree/pull/820

cgwalters:
A symptom of this issue is that on Fedora Atomic Host, `/sysroot/etc/ostree/remotes.d` exists (and is empty).

cgwalters:
We could likely introduce the concept of a "physical" sysroot vs a deployment one - perhaps just look for the existence of `/sysroot` as a dir?

PR #896, opened 2017-05-30
Add a notion of "physical" sysroot, use for remote writing

cgwalters:
Using `${sysroot}` to mean the physical storage root: We don't want to write to
`${sysroot}/etc/ostree/remotes.d`, since nothing will read it, and really
`${sysroot}` should just have `/ostree` (ideally). Today the Anaconda rpmostree
code ends up writing there. Fix this by adding a notion of "physical" sysroot.
We determine whether the path is physical by checking for `/sysroot`, which
exists in deployment roots (and there shouldn't be a `${sysroot}/sysroot`).

In order to unit test this, I added a `--sysroot` argument to `remote add`.
However, doing this better would require reworking the command line parsing for
the `remote` argument to support specifying `--repo` or `--sysroot`, and I
didn't quite want to do that yet in this patch.

Closes: https://github.com/ostreedev/ostree/issues/892

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -818,12 +818,21 @@ ostree_sysroot_load_if_changed (OstreeSysroot  *self,
   /* Determine whether we're "physical" or not, the first time we initialize */
   if (!self->loaded)
     {
-      if (fstatat (self->sysroot_fd, "sysroot", &stbuf, 0) < 0)
+      /* If we have a booted deployment, the sysroot is / and we're definitely
+       * not physical.
+       */
+      if (self->booted_deployment)
+        self->is_physical = FALSE;  /* (the default, but explicit for clarity) */
+      /* Otherwise - check for /sysroot which should only exist in a deployment,
+       * not in ${sysroot} (a metavariable for the real physical root).
+       */
+      else if (fstatat (self->sysroot_fd, "sysroot", &stbuf, 0) < 0)
         {
           if (errno != ENOENT)
             return glnx_throw_errno_prefix (error, "fstatat");
           self->is_physical = TRUE;
         }
+      /* Otherwise, the default is FALSE */
     }
 
   self->bootversion = bootversion;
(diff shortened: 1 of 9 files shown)


cgwalters:
This should be good to review/apply - PR status is travis infa flake.

src/libostree/ostree-sysroot.c
```diff
@@ -813,6 +815,17 @@ ostree_sysroot_load_if_changed (OstreeSysroot  *self,
                                cancellable, error))
     return FALSE;
 
+  /* Determine whether we're "physical" or not, the first time we initialize */
+  if (!self->loaded)
+    {
+      if (fstatat (self->sysroot_fd, "sysroot", &stbuf, 0) < 0)
```
jlebon: Hmm, this still feels a bit weak/arbitrary. Wouldn't it be safer to literally check if `self->path` corresponds to one of the picked up deployment dirs? E.g. by just cross-checking inodes? Or do we want to support initializing an `OstreeSysroot` to a deployment dir that's on a separate physical storage than `/`? (But then it'd seem more sane to create an `OstreeSysroot` to that physical sysroot storage instead anyway).

What I mean is, instead of checking that it's physical, it seems like it would be safer to check that it's *not* physical, since we're in pretty tight control of deployment dir layouts, etc..., right? (Whereas OTOH it seems like we shouldn't be affected by whatever else other than `/ostree` might be on the physical sysroot).

cgwalters: I'd say we are also testing for "not physical" too... if `/sysroot` exists that's a strong signal it's not physical.

Checking the inodes is interesting.  Hmm...maybe a simple one is to check whether we're in the booted deployment to mean not-physical. That would cover the 98% case.

cgwalters: Done :arrow_down:

cgwalters: (BTW I'm trying to avoid any path comparisons because part of the idea here is that one can create a sysroot from a directory fd)

jlebon:
@rh-atomic-bot r+ 4a63e20

PR #897, opened 2017-06-01
Add API function for getting last deployed sysroot

OYTIS:
While not used by OSTree, this function can be useful to other users of libostree (like https://github.com/advancedtelematic/aktualizr) as a good approximation of what deployment we're going to boot next after the reboot.


src/libostree/ostree-sysroot.c
```diff
@@ -1132,23 +1164,9 @@ ostree_sysroot_get_merge_deployment (OstreeSysroot     *self,
...
-          OstreeDeployment *deployment = self->deployments->pdata[i];
-
-          if (strcmp (ostree_deployment_get_osname (deployment), osname) != 0)
-            continue;
-
-          return g_object_ref (deployment);
-        }
-    }
-  return NULL;
+      return ostree_sysroot_get_last_deployed(self, osname);
```
jlebon: Minor nit: spacing between `(`.

src/libostree/ostree-sysroot.c
```diff
@@ -1108,6 +1108,38 @@ find_booted_deployment (OstreeSysroot       *self,
...
+ * ostree_sysroot_get_last_deployed:
+ * @self: Sysroot
+ * @osname: (allow-none): Operating system group
+ *
+ * Find the last deployed sysroot.
+ *
+ * Returns: (transfer full): The last deployment in OS or NULL if none
+ */
+OstreeDeployment *
+ostree_sysroot_get_last_deployed (OstreeSysroot     *self,
```
jlebon: I'd call this something like `ostree_sysroot_get_default_deployment`?

Actually, we have a similar version of this in rpm-ostree that returns not just the pending deployment, but also the rollback:

https://github.com/projectatomic/rpm-ostree/blob/95e9aa4/src/daemon/rpmostree-sysroot-core.c#L399

Maybe we should just upstream that to libostree.

cgwalters: Makes sense to me, can you try that @OYTIS  ?  `ostree_sysroot_query_deployments` ?

cgwalters:
Pushed a rewritten version for this :arrow_up:

cgwalters:
@rh-atomic-bot delegate=OYTIS

OYTIS:
Thanks, looks great.

OYTIS:
@rh-atomic-bot r+ b082119

PR #898, opened 2017-06-01
lib/repo: Delete unused private prototypes

cgwalters:
The implementations were removed in: 6ffcb24d227eae5a479caf45adb8037eceb6ae33
I noticed this while looking at the commit code.


jlebon:
@rh-atomic-bot r+ e555735

PR #900, opened 2017-06-02
lib/sysroot: Document the NO_CLEAN flag

krnowak:


jlebon:
Thanks!
@rh-atomic-bot r+ 41f723d

Issue #901, opened 2017-06-02
remotes are being deleted after 'rpm-ostree rebase'


miabbott:
In our CAHC/FAHC tests, we are seeing failures due to the remotes on the system going missing after the host is upgraded via `rpm-ostree rebase`.

This appears to have started happening with `ostree` at commit 1eff3e8, which maps to PR #896.

Example output below:

```
[root@mike-fedora ~]# rpm-ostree status
State: idle
Deployments:
● fedora-atomic:fedora-atomic/25/x86_64/docker-host
             Version: 25.127 (2017-05-19 21:57:03)
              Commit: cdd359911de49f3a8199ffd41a9894019562001d6cf9be66e1894c31b6fa1c66
[root@mike-fedora ~]# ostree remote list
fedora-atomic
[root@mike-fedora ~]# ostree remote add --no-gpg-verify custom  https://ci.centos.org/artifacts/sig-atomic/fahc/ostree/repo/
[root@mike-fedora ~]# rpm-ostree rebase custom:fahc/25/x86_64/buildmaster

838 metadata, 3399 content objects fetched; 185027 KiB transferred in 16 seconds                                                                                  
Copying /etc changes: 21 modified, 0 removed, 54 added
Transaction complete; bootconfig swap: yes deployment count change: 1
Upgraded:
  acl 2.2.52-11.fc24 -> 2.2.52-12.fc25
  atomic 1.17.1-4.fc25 -> 1.17.2-1.fc25
  atomic-devmode 0.3.6-1.fc25 -> 0.3.7.0-f42200dcff22700d582ee6f39182e5e6ca3e5e44.ff82423d5c92844e04dd0669cacd976740cd3393.fc25
  bubblewrap 0.1.7-1.fc25 -> 0.1.8.2-6ef45aae778b6ce68c39d97089479fdfe6b441b3.1b0f5b6ba6517b1ad426117beaa969547b8c648b.fc25
  container-selinux 2:2.10-1.fc25 -> 2:2.14-1.fc25
  dhcp-client 12:4.3.5-1.fc25 -> 12:4.3.5-2.fc25
  dhcp-common 12:4.3.5-1.fc25 -> 12:4.3.5-2.fc25
  dhcp-libs 12:4.3.5-1.fc25 -> 12:4.3.5-2.fc25
  freetype 2.6.5-7.fc25 -> 2.6.5-9.fc25
  glusterfs 3.10.1-1.fc25 -> 3.10.2-1.fc25
[...]

PR #902, opened 2017-06-02
Revert "Add a notion of "physical" sysroot, use for remote writing"

cgwalters:
This reverts commit 1eff3e8. There
are a few issues with it.  It's not a critical thing for now, so
let's ugly up the git history and revisit when we have time to
debug it and add more tests.

Besides the below issue, I noticed that the simple `ostree remote add`
now writes to `/ostree/repo/config` because we *aren't* using the
`--sysroot` argument.

Closes: #901

### DIFF
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -93,7 +93,6 @@ struct OstreeRepo {
   int objects_dir_fd;
   int uncompressed_objects_dir_fd;
   GFile *sysroot_dir;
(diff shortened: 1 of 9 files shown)


cgwalters:
This will reopen https://github.com/ostreedev/ostree/issues/892

cgwalters:
@rh-atomic-bot delegate=mike-nguyen

mike-nguyen:
@rh-atomic-bot r+ 8749c92

Issue #633, opened 2016-12-14
user-mode checkouts don't keep permissions when doing a copy fallback


sjoerdsimons:
When doing a repository checkout with user-mode and the checkout has to fallback to copying (e.g. checkout out on a different filesystem/btrfs subvolume), the file permissions aren't properly preserved as they should. e.g. some of the executable in the repository ended up being 0600 (read/write for the user) rather then the exported 0755 (rwx user, rx group/others)

gtristan:
I've been debugging this as I have run into the same issue, this actually seems to be a regression, and does not happen with the ostree v2016.14 which is installed via the package manager on my host.

Building ostree master shows in the version output:
```
libostree:
 Version: 2017.6
 Git: v2017.6-34-ga094879f3a65ff0b9909dcb7f436257f034d29cd
 Features:
  - libsoup
  - gpgme
  - libarchive
  - selinux
  - libmount
```

Following test case displays the problem with recent ostree, but does not show the problem with v2016.14

```
#!/bin/bash

# Test case demonstrating broken file permissions
#
# This problem turns out to be specific to checking out
# from archive-z2 repositories when the uncompressed cache
# is not used.
#
# Usage:
#
#  ./test.sh <work directory>
#
#
set -e
set -x

topdir=${1}
mkdir -p "${topdir}"
topdir="$(cd ${topdir} && pwd)"

parent_repo="${topdir}/parent_repo"
child_repo="${topdir}/child_repo"
checkin="${topdir}/checkin"
checkout_nocache="${topdir}/checkout_nocache"
checkout_cached="${topdir}/checkout_with_uncompressed_cache"

#
# Init the repo with compression, the problem seems to be specific
# to checking out from repositories that have compression enabled
# but uncompressed cache disabled at checkout time
#
ostree init --repo "${parent_repo}" --mode archive-z2

#
# Create test data with some permissions
#
mkdir -p "${checkin}/subdir"
echo "foo" > "${checkin}/foo"
echo "bar" > "${checkin}/bar"
chmod 0600 "${checkin}/foo"
chmod 0755 "${checkin}/bar"
echo "foo" > "${checkin}/subdir/foo"
echo "bar" > "${checkin}/subdir/bar"
chmod 0600 "${checkin}/subdir/foo"
chmod 0755 "${checkin}/subdir/bar"

#
# Create commit
#
ostree commit --repo "${parent_repo}" --branch "test" -s "test commit" "${checkin}"

#
# Checkout data from child repo
[...]

PR #903, opened 2017-06-02
checkout: also chmod in the user checkout case

jlebon:
When falling back to copying, we previously would only chmod checked out
files in the non-user-checkout mode. Fix this by always doing chmod.
The file_mode was being prepared but never actually applied.

Add a basic test in the archive-z2 --> usermode checkout case in which
we're guaranteed to always fall back to copy mode.

Closes: #633

### DIFF
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -557,6 +557,8 @@ ${CMD_PREFIX} ostree --repo=repo2 pull-local repo
 rm -rf test2-checkout
 ${CMD_PREFIX} ostree --repo=repo2 checkout -U --disable-cache test2 test2-checkout
 if test -d repo2/uncompressed-objects-cache; then
+    assert_file_has_mode test2-checkout/baz/cowro 600
+    assert_file_has_mode test2-checkout/baz/deeper/ohyeahx 755
     ls repo2/uncompressed-objects-cache > ls.txt
     if test -s ls.txt; then
 	assert_not_reached "repo has uncompressed objects"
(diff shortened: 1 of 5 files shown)


cgwalters:
Nice, thanks!

@rh-atomic-bot r+ 618091c

jlebon:
OK, hopefully this works now!

tests/basic-test.sh
```diff
@@ -557,8 +557,13 @@ ${CMD_PREFIX} ostree --repo=repo2 pull-local repo
 rm -rf test2-checkout
 ${CMD_PREFIX} ostree --repo=repo2 checkout -U --disable-cache test2 test2-checkout
 if test -d repo2/uncompressed-objects-cache; then
-    assert_file_has_mode test2-checkout/baz/cowro 600
-    assert_file_has_mode test2-checkout/baz/deeper/ohyeahx 755
+    # we're in archive mode, but the repo we pull-local from might be
```
cgwalters: Why is this underneath the `if test -d repo2/uncompressed-objects-cache; then` conditional?

cgwalters:
Ah yes, bare-user-only.  Thanks for the docs patch too.  Just one question I noticed, otherwise LGTM (still).

tests/basic-test.sh
jlebon: No good reason at all. :) I had somehow missed the fact this was an error path. I did push another commit to make this more obvious. ⬇️

cgwalters:
@rh-atomic-bot r+  e1c7336

Issue #899, opened 2017-06-01
--mirror --depth=-1 fails with error: Exceeded maximum recursion


bcl:
The source repo has 2741 commits in it. By experiment it seems I can only use up to --depth=247

cgwalters:
Yeah, the code was orirginally trying to protect against stack overflow.  There's a hardcoded `#define OSTREE_MAX_RECURSION (256)` right now.  However, since that was introduced, we switched to using a queue.  It's likely safe to lift this and make it basically arbitrary.

PR #904, opened 2017-06-02
Remove the OSTREE_MAX_RECURSION limit on metadata depth

bcl:
This was making it impossible to pull or mirror a large ostree repo, and
according to Colin is no longer necessary. It works fine with a test
against a repo with 2741 commit and 451468 objects in it.

Closes: #899


jlebon:
Thanks!

It looks like we don't need `recursion_depth` anymore. Though actually, it might be a helpful statistic to print out when users pass `--depth=-1`, so let's keep it around for now (I pushed a fixup with a comment to that effect).

@rh-atomic-bot r+ 663897d

PR #905, opened 2017-06-02
Don't install trivial-httpd man page if not enabled

cgwalters:
I just noticed this scroll by in a file listing.


jlebon:
@rh-atomic-bot r+ 89d9e0a

Issue #907, opened 2017-06-04
bare-user files are always executable (0755 perms)


gtristan:
With bare repository this is not an issue, but with bare-user and user mode checkouts, my expectation is that permissions are not munged, only file ownership is left to the active user.

This test case exhibits some strange behavior:

```
#!/bin/bash

# Test case demonstrating files becoming executable out of thin air
#
# This seems to occur with bare-user repository and checking out in user mode
#
# Usage:
#
#  ./test.sh <work directory>
#
#
set -e
set -x

topdir=${1}
mkdir -p "${topdir}"
topdir="$(cd ${topdir} && pwd)"

repo="${topdir}/repo"
checkin="${topdir}/checkin"
checkout="${topdir}/checkout"

#
# Create test data with some permissions
#
mkdir -p "${checkin}/subdir"
echo "foo" > "${checkin}/foo"
echo "bar" > "${checkin}/bar"
chmod 0600 "${checkin}/foo"
chmod 0755 "${checkin}/bar"
echo "foo" > "${checkin}/subdir/foo"
echo "bar" > "${checkin}/subdir/bar"
chmod 0600 "${checkin}/subdir/foo"
chmod 0755 "${checkin}/subdir/bar"
touch "${checkin}/emptyfile"

#
# Init repo with bare-user
#
ostree init --repo "${repo}" --mode bare-user

#
# Create commit
#
ostree commit --repo "${repo}" --branch "test" -s "test commit" "${checkin}"

#
# Checkout commit, just pass in --disable-cache here because it's default
# when using the API (but should not make a difference, here we are not
# using a compressed repo anyway).
#
ostree checkout --repo "${repo}" --user-mode --disable-cache "test" "${checkout}"
```

cgwalters:
Some discussion on list about this https://mail.gnome.org/archives/ostree-list/2017-June/msg00011.html

Honestly, I don't know why we use `0755` always.  Looking at a patch.

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/908 - does that make sense to you?

gtristan:
Well, sort of :)

First, yes now that I have understood a bit more the purpose of "bare-user" mode is for users to run software in these checkouts inside containers, it makes sense for this use case to strip the suid bits, and certainly this part was previously missing for hardlink user mode checkouts of "bare-user" repositories.

Second, I'm not sure it's necessary to force read-write for user onto the mode bits. Probably there are two opposing things to consider here:
[...]

PR #908, opened 2017-06-05
lib/repo: For bare-user, make repo modes mirror checkout user-mode

cgwalters:
Having every object in a bare-user repo (and checkouts) be executable
is ugly.  I can't think of a good reason to do that.  So make
the committed mode semantics mirror that for user-mode checkouts; we
strip suid/sgid bits.

However, we also do ensure that the written files are read/writable by the
owning user, since otherwise we couldn't do anything to them. I'm not aware of
any real use cases for non-readable/non-writable by owner files in ostree.

Closes: https://github.com/ostreedev/ostree/issues/907

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -283,19 +283,19 @@ commit_loose_object_trusted (OstreeRepo        *self,
       if (objtype == OSTREE_OBJECT_TYPE_FILE &&
           self->mode == OSTREE_REPO_MODE_BARE_USER)
         {
-          if (!object_is_symlink)
-            {
-              /* We need to apply at least some mode bits, because the repo file was created
-                 with mode 644, and we need e.g. exec bits to be right when we do a user-mode
-                 checkout. To make this work we apply all user bits and the read bits for
-                 group/other.  Furthermore, setting user xattrs requires write access, so
-                 this makes sure it's at least writable by us.  (O_TMPFILE uses mode 0 by default) */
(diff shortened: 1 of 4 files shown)


cgwalters:
Oh, interesting.  So this started failing in `bare-user-only` because the test suite does:

```
$OSTREE checkout test2-override checkout-test2-override
test -g checkout-test2-override/a/nested/2
```

For `bare-user`, because we weren't passing `-U` we ended up with a copy.  For `bare-user-only`, we suddenly started dropping suidness.

@alexlarsson - do you want ostree to preserve suid files in `bare-user-only`?  It seems like a bad idea, no?   Hmm, couldn't an app write a setuid file via the system helper today?

alexlarsson:
No that sounds pretty bad. We probably want to strip setuid in bare-user-only repos (i.e. lose fidelity for such objects).

cgwalters:
I wrote https://github.com/flatpak/flatpak/pull/837 - let's discuss this there.

cgwalters:
OK, updated this based on discussion in https://github.com/ostreedev/ostree/issues/907
Tests should pass now that we also fixed the suid bits in https://github.com/ostreedev/ostree/pull/909

cgwalters:
Argh, running into `umask` issues - at least in Fedora unprivileged users get `002`, but it looks like our Docker containers get `022.`  Added a fixup :arrow_up:

alexlarsson:
@rh-atomic-bot r+ 8fadd15

PR #909, opened 2017-06-06
Canonicalize bare-user-only perms with 0755 mask

cgwalters:
For the flatpak use case where bare-user-only was introduced, we actually
don't want to support s{u,g} id files in particular.

Actually, I can't think of a reason to have anything outside of the
`0755 i.e. (u=rwx,g=rx,o=rx)` mask, so that's what we do here.

This will have the effect of treating existing `bare-user-only` repositories as
corrupted if they have files outside of that mask, but I think we should do this
now; most of the flatpak users will still be on `bare-user`, and we haven't
changed the semantics of that mode yet.

Note that in this patch we will also *reject* file content that doesn't
match this.  This is somewhat asymmetric, since we aren't similarly rejecting
e.g. directory metadata.  But, this will close off the biggest source
of the problem for flatpak (setuid binaries).

See: https://github.com/ostreedev/ostree/pull/908
See: https://github.com/flatpak/flatpak/pull/837

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -281,7 +281,7 @@ commit_loose_object_trusted (OstreeRepo        *self,
         }
 
       if (objtype == OSTREE_OBJECT_TYPE_FILE &&
-          (self->mode == OSTREE_REPO_MODE_BARE_USER || self->mode == OSTREE_REPO_MODE_BARE_USER_ONLY))
+          self->mode == OSTREE_REPO_MODE_BARE_USER)
         {
           if (!object_is_symlink)
             {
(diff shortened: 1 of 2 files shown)


cgwalters:
I think we're going to need to scan flatpak content for files that don't match this.  I'm going to do some of that myself now.

alexlarsson:
LGTM

alexlarsson:
I'm not super worried about compatibility atm. We never use bare-user-only by default yet, its mostly a thing you can enable if you need to e.g. run on tmpfs or in docker. If anything is not working for that we can fix it over time by rebuilding with newer flatpak/ostree versions.

cgwalters:
OK, cool.  BTW you have homu rights, so feel free to drop a `r+ $checksum` here.

alexlarsson:
@rh-atomic-bot r+  	a8eaf0d

cgwalters:
@alexlarsson I sent you an invite link

alexlarsson:
@rh-atomic-bot r+ a8eaf0d

cgwalters:
It sounds like we may actually want to mask 0775 here based on https://github.com/flatpak/flatpak/pull/837#issuecomment-306737654 ?

That will transparently fix pulling into new `bare-user-only` repos some of that existing flatpak content.

PR #910, opened 2017-06-07
lib/repo: Don't copy xattrs when manipulating the GPG keyring

owtaylor:
Copying xattrs when manipulating the GPG keyring for a repository
causes errors when the underlying filesystem doesn't support writing
xattrs - overlayfs is a common example. It also causes the selinux
attributes of the keyring files to be copied from the temporary
location instead of properly inherited from the destination directory
(ending up, for example, as unconfined_u:object_r:user_tmp_t:s0, rather
than unconfined_u:object_r:data_home_t:s0)

NOTE: The probably with the selinux contexts could potentially considered that libglnx is not sophisticated enough about copying attrs - coreutils cp, when instructed to copy xattrs with 'cp --preserve=xattr' or 'cp -a' uses libattr routines to honor /etc/xattr.conf and has special handling of selinux - but in the end, I don't think there's any reason to believe that preserving xattrs is important in this location.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -1329,7 +1329,6 @@ ostree_repo_remote_gpg_import (OstreeRepo         *self,
   struct stat stbuf;
   gpgme_error_t gpg_error;
   gboolean ret = FALSE;
-  const GLnxFileCopyFlags copyflags = self->disable_xattrs ? GLNX_FILE_COPY_NOXATTRS : 0;
 
   g_return_val_if_fail (OSTREE_IS_REPO (self), FALSE);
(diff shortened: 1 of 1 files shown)


cgwalters:
This is a tricky topic, getting into the semantics of what exactly `glnx_copy_file()` should do.  For e.g. handling `/etc`, I think we *should* error out if somehow we have SELinux labels for the source files but fail to set in the destination.  (In practice this isn't going to happen since it's effectively required for them to be the same filesystem)

Anyways while I'm fine with this patch, down the line I'd like enhance `glnx_copy_file()` to have better semantics.  In particular, I think we should be copying ACLs, and that is possibly relevant in this context.

@rh-atomic-bot r+ 85cf530

PR #911, opened 2017-06-07
Minor fixes

pwithnall:
Various minor fixes: typo fixes, whitespace fixes, a minor optimisation in the `summary` file, and improvements in formatting for `ostree summary -v`.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2690,7 +2690,7 @@ initiate_request (OtPullData *pull_data,
  *   * override-commit-ids (as): Array of specific commit IDs to fetch for refs
  *   * dry-run (b): Only print information on what will be downloaded (requires static deltas)
  *   * override-url (s): Fetch objects from this URL if remote specifies no metalink in options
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -3392,7 +3392,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       g_autofree char *original_rev = NULL;
 
       if (pull_data->remote_name)
-        remote_ref = g_strdup_printf ("%s/%s", pull_data->remote_name, ref);
+        remote_ref = g_strdup_printf ("%s:%s", pull_data->remote_name, ref);
```
cgwalters: Hmm...yes, this looks weird/broken, but this change makes me a bit nervous.  I *think* this code originally came from the `pull-local` code to support `--remote=REMOTE             Add REMOTE to refspec` ?

And the use case for that...hm...I forget.  I think it was something like doing a pull without having to add a `file:///` remote?

Would you mind submitting this patch as a separate PR?

pwithnall: ⇒ PR #912.

cgwalters:
Thanks!

@rh-atomic-bot r+  b01ac08

PR #912, opened 2017-06-07
lib/pull: Fix construction of a refspec to use the correct separator

pwithnall:
This code looks like it was supposed to build a refspec, but it used a
slash as a separator rather than a colon. The following code does
recover by supporting prefix matching with slashes, but it seems like
this was perhaps not the intention.

---

Split out from PR #911.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3393,7 +3393,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       g_autofree char *original_rev = NULL;
           
       if (pull_data->remote_name)
-        remote_ref = g_strdup_printf ("%s/%s", pull_data->remote_name, ref);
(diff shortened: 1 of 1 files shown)


cgwalters:
Copying my comment here until I have time to dredge up some history and understand this code:

Hmm...yes, this looks weird/broken, but this change makes me a bit nervous. I think this code originally came from the pull-local code to support --remote=REMOTE Add REMOTE to refspec ?

And the use case for that...hm...I forget. I think it was something like doing a pull without having to add a file:/// remote?

cgwalters:
The original logic here is...old.  It appears to date to mostly be from https://github.com/ostreedev/ostree/commit/255b7d13090ee8d3957288e7d831d376292753ff (thanks for the great commit message me from 2012!) and there note that we had a "No changes" message which no longer exists.

That said it does look like the prefix matching works, so the logic to not set the ref if it's unchanged is working.  But your patch is clearly more correct.

@rh-atomic-bot r+ 921c0ad

PR #913, opened 2017-06-07
 repo/commit: Support group-writable files for bare-user-only

cgwalters:
These exist in the wild for flatpak, and aren't really a problem.
The canonical permissions are still `0{6,7}44`, we just support
the additional writable bit for the group now to avoid breaking
some flatpak content.

PR #914, opened 2017-06-07
 lib/checkout: Ignore world-writable dirs for bare-user-only checkout

cgwalters:
See #909 for more information on the
rationale. Basically there's no reason for flatpak (which uses `bare-user-only`)
to have world-writable dirs. Particularly with the presence of the system
helper.

An approach I considered instead was to parse and validate directory metadata
objects at commit time. We still may do that in addition; for file objects we *had*
to do it that way because the actual files would be laid down suid.  But directories
live only as inert `.dirmeta` objects until we do a checkout (i.e. `mkdir()`), so
we can solve the problem at checkout time.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -749,8 +749,18 @@ checkout_tree_at_recurse (OstreeRepo                        *self,
    */
   if (!did_exist)
     {
-      if (TEMP_FAILURE_RETRY (fchmod (destination_dfd, mode)) < 0)
-        return glnx_throw_errno (error);
+      guint32 canonical_mode;
+      /* Silently ignore world-writable directories (plus sticky, suid bits,
+       * etc.) when doing a checkout for bare-user-only repos. This is related
+       * to the logic in ostree-repo-commit.c for files.
+       * See also: https://github.com/ostreedev/ostree/pull/909 i.e. 0c4b3a2b6da950fd78e63f9afec602f6188f1ab0
+       */
+      if (self->mode == OSTREE_REPO_MODE_BARE_USER_ONLY)
+        canonical_mode = (mode & 0775) | S_IFDIR;
+      else
+        canonical_mode = mode;
+      if (TEMP_FAILURE_RETRY (fchmod (destination_dfd, canonical_mode)) < 0)
+        return glnx_throw_errno_prefix (error, "fchmod");
     }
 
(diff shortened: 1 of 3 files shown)


cgwalters:
This depends on https://github.com/ostreedev/ostree/pull/913

alexlarsson:
So, flatpak currently uses bare-user, although the long term plan when everything is working is to use bare-user-only. 

Does it ever make sense to have writable directories when you do a checkout --user, independent of the repo format? I mean, I understand the system container case where you need 100% fidelity and check out as root. But once you're changing the uid/gid of the files you check out, then the permissions recorded in the repo does not mean the same thing as they do in the repo.

cgwalters:
Probably not.  But I was trying to be conservative initially; changing `bare-user`  affects a lot more things (gcontinuous, buildstream, etc.), plus of course existing flatpak installs.

I was considering introducing a new flag for this (`checkout ---user-safe` or something?).  But...there's definitely an argument we should just do the same for `bare-user`.  If you're fine with that for flatpak, then I'd agree let's do it.

alexlarsson:
Its not just about bare-user though. I mean, if you have an archive-z2 repo, and you do a checkout --user from that, do you want a world-writable directory in your homedir? I mean, you might want to if you want 100% metadata fidelity, but then you would not use --user, which breaks that anyway. 

I dunno. Maybe its too magic and should be a separate flag like --user-safe.

cgwalters:
I posted to the list: https://mail.gnome.org/archives/ostree-list/2017-June/msg00026.html

However, I realized there is a problem; when you want to do a *commit* based on content you checked out.  And we'll definitely pick back up directory modes from there.

The only solution to this is to use a commit modifier (e.g. `ostree commit --statoverride`) for `/tmp`.

I think though most libostree users should start using `0700` directories and doing checkouts underneath there.  See:

 - https://github.com/flatpak/flatpak/pull/843
 - https://github.com/projectatomic/rpm-ostree/pull/821

The special case here really is flatpak system wide installs, since there the directory obviously does need to be accessible by unprivileged users.

So...I think we should go with this patch as is for now.

cgwalters:
Rebased :surfing_man:

cgwalters:
Or to elaborate, we could do this for `bare-user` behind a new flag, making the distinction between flatpak's use where we never want to have suid/worldwritable, and between "os buildroots" which need to keep around `/tmp` potentially.

But then again, given the impending migration to `bare-user-only`, is it worth it?

PR #916, opened 2017-06-08
OstreeRepo: Always look in staging directory for objects

alexlarsson:
Its often the case that we want to look at objects inside a commit,
before the objects the transaction is finished. For instance:
  https://github.com/flatpak/flatpak/pull/837
Which tries to verify the file permissions before committing the
transaction.

And:
  https://github.com/flatpak/flatpak/commit/1e5ffa926a25acb655af7889b679b140bf44870b
Which collects the storage size of the objects so that we can
put the total download size in the commit metadata.

I tried to find all the places where we did reads from the
object directories, and in particular this fixes:

 ostree_repo_load_file() for bare repos (archive-z2 was already working).
 ostree_repo_query_object_storage_size()
 Applying deltas that references not-yet-commited objects.

Note: Its somewhat tricky to test these changes as we don't really expose non-commited transaction states in the CLI, so it will only ever be hit by library users. So, this needs careful review.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2501,19 +2501,33 @@ load_metadata_internal (OstreeRepo       *self,
   return TRUE;
 }
 
+/* Basically fstatat(), but also looks in both the committed and staging
+ * directories, and returns *out_dfd for where we found the object.
+ */
 static gboolean
-query_info_for_bare_content_object (OstreeRepo      *self,
-                                    const char      *loose_path_buf,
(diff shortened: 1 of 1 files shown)


cgwalters:
We do have C API tests in `test-basic-c.c`.

cgwalters:
I simplified this patch a lot by having `query_info_for_bare_content_object()` return the dfd where it found the object.

cgwalters:
@rh-atomic-bot r+ 5253f54

PR #917, opened 2017-06-08
ci: Update to match current rpm-ostree

cgwalters:
This copies the `ci/` directory from rpm-ostree, with much the same rationale;
among other things we don't want to depend on the Docker hub.

The specific reason I'm doing this is that I want to add a CentOS7 build, but
that means we can't use `projectatomic/ostree-tester`, and at that point we
might as well unwind it all.


ci/build.sh
```diff
@@ -0,0 +1,15 @@
...
+dn=$(dirname $0)
+. ${dn}/libbuild.sh
+
+install_builddeps ostree
+
+dnf install -y sudo which attr fuse gjs parallel coccinelle clang \
+    libubsan libasan libtsan PyYAML gnome-desktop-testing redhat-rpm-config \
+    elfutils ${CI_PACKAGES:-}
+
+build --enable-gtk-doc --enable-installed-tests ${CONFIGOPTS}
```
jlebon: This should probably be `${CONFIGOPTS:-}`.

.papr.yml
```diff
@@ -4,128 +4,62 @@ branches:
...
-build:
-    config-opts: >
-      --prefix=/usr
-      --libdir=/usr/lib64
-      --enable-installed-tests=exclusive
-      --enable-gtk-doc
-      --with-curl
-      --with-openssl
+env:
+  CI_PACKAGES: "'pkgconfig(libcurl)' 'pkgconfig(openssl)'"
```
jlebon: I think it should just work without the single quotes around each.

cgwalters:
This should be good to go.

.papr.yml
```diff
@@ -4,128 +4,64 @@ branches:
...
-# The g-d-t-r timeout is for test-pull-many.sh; if tweaking this,
-# also be sure to change the other cases below
 tests:
-    - make syntax-check
-    - ./tests/ci-commitmessage-submodules.sh
-    - make check
-    - /bin/sh -c 'gnome-desktop-testing-runner -p 0 --timeout $((10 * 60)) libostree/'
+  - ci/ci-commitmessage-submodules.sh
+  - ci/build-check.sh
+  - make syntax-check
```
jlebon: Can we move `syntax-check` to run before `build-check.sh`?

.papr.yml
```diff
@@ -4,128 +4,64 @@ branches:
...
 env:
     CFLAGS: '-fsanitize=undefined -fsanitize-undefined-trap-on-error -fsanitize=address -O2 -Wp,-D_FORTIFY_SOURCE=2'
     ASAN_OPTIONS: 'detect_leaks=0'  # Right now we're not fully clean, but this gets us use-after-free etc
     # TODO when we're doing leak checks: G_SLICE: "always-malloc"
 
-build:
-    config-opts: >
-      --prefix=/usr
-      --libdir=/usr/lib64
-      --enable-installed-tests=exclusive
```
jlebon: We're losing this tidbit, so the `g-d-t-r` invocation in `build-check.sh` won't actually run anything right?

ci/build.sh
```diff
@@ -0,0 +1,15 @@
...
+set -xeuo pipefail
+
+dn=$(dirname $0)
+. ${dn}/libbuild.sh
+
+install_builddeps ostree
+
+dnf install -y sudo which attr fuse gjs parallel coccinelle clang \
+    libubsan libasan libtsan PyYAML gnome-desktop-testing redhat-rpm-config \
+    elfutils ${CI_PACKAGES:-}
```
jlebon: Any reason for using `CI_PACKAGES` instead of just plopping the extra ones directly in the YAML?

.papr.yml
cgwalters: Yes, good catch!

.papr.yml
cgwalters: Unfortunately `syntax-check` requires us to have run `configure` (though glancing at the code, not for good reason).

It's tempting to hook up `syntax-check` to `check`.

cgwalters:
Fixed up for comments :arrow_up:

jlebon:
Nice! /me makes note to delete the ostree-tester image.
@rh-atomic-bot r+ 98e3d5c

PR #918, opened 2017-06-08
repo: After renaming in all loose objects, ensure metadata is stable

cgwalters:
When a transaction is finished and we have moved all the staged loose
objects into the repo we fsync all the object directory, to ensure the
filenames are stable before we update the refs files to point to the
new commits.

With out this an unclean shutdown after the transaction is finished
could result in a refs file that points to an incomplete commit.

https://bugzilla.gnome.org/show_bug.cgi?id=759442

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1202,7 +1202,9 @@ rename_pending_loose_objects (OstreeRepo        *self,
   while (TRUE)
     {
       struct dirent *dent;
+      gboolean renamed_some_object = FALSE;
       g_auto(GLnxDirFdIterator) child_dfd_iter = { 0, };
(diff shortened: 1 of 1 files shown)


cgwalters:
Moving this out from https://github.com/ostreedev/ostree/pull/813

cgwalters:
This seems accurate; I worry a bit about doing a bunch of `fsync()` on the directories serially though.  Down the line perhaps what we really want at least for the ostree-as-host case is to combine the final commit with the deployment `syncfs()`.  Not entirely sure right now how to design that though.  Perhaps extend our "boot-id directory" model so that one has refs that are only valid for this boot?

cgwalters:
@rh-atomic-bot r+  	d4199ec

PR #922, opened 2017-06-12
Hardlink all local metadata pulls, support most hardlinking for bare-user{,-only} conversions

cgwalters:
See patches.


jlebon:
This looks good to me. Will let @alexlarsson have a look over as well.

jlebon:
@rh-atomic-bot delegate=alexlarsson

alexlarsson:
@rh-atomic-bot r+ b427a1a

PR #923, opened 2017-06-12
lib/repo: Refactor object copy import function

cgwalters:
This came up in: https://github.com/ostreedev/ostree/pull/881

Basically doing streaming for metadata is dumb. Split up the metadata/content
paths so we pass metadata around as `GVariant`. This drops the last internal
caller of `ostree_repo_write_metadata_stream_trusted()` which was the dumb
function mentioned.


jlebon:
@rh-atomic-bot r+ 8b997a5

PR #924, opened 2017-06-12
Support peer-to-peer discovery and fetching of refs (LAN and USB)

pwithnall:
This is a mostly-final version of my work to add support for peer-to-peer discovery and fetching of refs to libostree. The bits which are not finalised at the moment are:
 * `ostree-repo-finder-mount.[ch]`
 * `ostree-repo-finder-avahi.[ch]`
 * The documentation for `OstreeBloom`
 * Missing a man page for `ostree find-remotes`

Apart from that, everything is ready to review. I’ll be adding some integration tests next.

This does not include any of the work we discussed on unsigned summary files; I decided to keep that separate to ease review and merging. The two features do work separately; unsigned summaries are only needed once we start sharing refs from multiple collections from a single peer (which will be a very common occurrence, but not necessary to get things reviewed, compiled and tested). I’ll open a separate PR for unsigned summary file support once this one is on its way.

I’ve tried to split the commits up a bit and hide tool or API changes behind `OSTREE_ENABLE_EXPERIMENTAL_API`. Some of the core code changes are *not* hidden behind this, but should only take effect when used with repositories and remotes with `collection-id` keys in their configuration. If anything is unclear, please say so, as I can add more comments.

This branch doesn’t attempt any significant refactoring of the pull code, even though that might be helpful. It can follow in future to improve performance. I am not aiming for high performance in this first implementation; no profiling has been done. No significant integration testing has been done either, aside from the SH tests which I will be adding. This is not ready to have the `OSTREE_ENABLE_EXPERIMENTAL_API` guards removed.

This is an updated version of PR #782.


src/libostree/ostree-bloom.c
```diff
@@ -0,0 +1,556 @@
...
+ */
+OstreeBloomHashFunc
+ostree_bloom_get_hash_func (OstreeBloom *bloom)
+{
+  g_return_val_if_fail (bloom != NULL, NULL);
+
+  return bloom->hash_func;
+}
+
+/* SipHash code adapted from https://github.com/veorq/SipHash/blob/master/siphash.c */
```
cgwalters: Maybe copy literally from systemd's version?  If we can do literal copy/paste it'll be easier to pick up any bugfixes.

pwithnall: The ‘adapted’ here basically just means copied from a few files and the headers put in the right place. It should be trivial to pick up bugfixes for this. Would there be any advantage in using systemd’s code?

cgwalters: In this specific case probably not, but just as a general rule I use systemd as a copypaste/reference for various things.  Anyways it's fine as is.

pwithnall: OK. If you would prefer to go with systemd, that’s not a problem — but since the hash implementation is API (needs to be the same on the client and server sides of a peer-to-peer resolve operation), it needs to change now or never.

pwithnall:
Finally got all the automated checks to succeed, apart from f25ah-insttest, which seems to have an unrelated failure (I think?).

cgwalters:
I think my biggest concern is that we're not breaking functionality/security of non-collection pulls and build of libostree without `--enable-experimental-api`.  The fact that the tests pass is obviously a good indicator, but still the pull code is...complex and fragile.

Did you have any outstanding concerns in that respect?  Anything reviewers should look for specifically?

Other than that...it's a bit up to you how much review we do before landing.  I'd like to get a chance to at least superfically scan, that's going to be a day or two.  More detailed review would take longer but OTOH I'm also fine landing things and doing any further fixes post merge.

Also, there's an impending ostree release soon for https://github.com/flatpak/flatpak/issues/845 and I'd like to do that first.

pwithnall:
> I think my biggest concern is that we're not breaking functionality/security of non-collection pulls and build of libostree without `--enable-experimental-api`. The fact that the tests pass is obviously a good indicator, but still the pull code is...complex and fragile.
> 
> Did you have any outstanding concerns in that respect? Anything reviewers should look for specifically?

Indeed. I’ve kept the changes to the existing pull code to a minimum — they are all in the “lib/pull: Add collection support to ostree_repo_pull_with_options()” commit, adding support for a new `collection-refs` option. So that’s the commit to look at closely. At the very least, we need to verify that `ostree_repo_pull_with_options()` doesn’t change behaviour if the `collection-refs` option is *not* provided.

The other major concern I have is API stability: a lot of work in flatpak, gnome-software and eos-updater is based on this new libostree stuff, so I’d like to be 99% sure that you’re happy with the API. API obviously includes the new options, metadata schemas, config file schemas, etc.

> Other than that...it's a bit up to you how much review we do before landing. I'd like to get a chance to at least superfically scan, that's going to be a day or two. More detailed review would take longer but OTOH I'm also fine landing things and doing any further fixes post merge.

I’d be OK with a superficial scan of the whole thing, and an in-depth look at the pull code and the API additions (as discussed above). The fact that it’s all hidden behind `OSTREE_ENABLE_EXPERIMENTAL_API` means we can push in fixes later on.

> Also, there's an impending ostree release soon for flatpak/flatpak#845 and I'd like to do that first.

:+1:

src/libostree/libostree.sym
```diff
@@ -301,6 +301,7 @@ global:
         ostree_sysroot_write_origin_file;
         ostree_validate_checksum_string;
         ostree_validate_rev;
+        ostree_validate_remote_name;
```
cgwalters: This is the wrong place; you're not the first to do this though.  I submitted https://github.com/ostreedev/ostree/pull/931

pwithnall: Ufff, this was a definite thinko.

cgwalters:
Will flatpak also gain a `--enable-ostree-experimental` build flag?  Has that work been started?

How much ability/latitude do we have to change the experimental API and the consumers of it once this lands?

The fact that we're using remotes for install but collections for update seems like it's going to be weird, but that's inherent with the model I guess.

pwithnall:
> Will flatpak also gain a `--enable-ostree-experimental` build flag? Has that work been started?

That work has been started, but the branch hasn’t been rebased and updated since we renamed from ‘origin ID’ to ‘collection ID’. I’d like some certainty on the shape of the new libostree API before I start polishing it.

It looks like flatpak won’t need any additional API, so its changes could be kept conditional on whether it’s built against a libostree which has the experimental API. However, I haven’t looked at that part of the build goop in detail yet.

> How much ability/latitude do we have to change the experimental API and the consumers of it once this lands?

A bit (otherwise why did we bother with the experimental API flag?), but I’d like the general shape of things to be considered relatively stable. We can tweak function arguments, naming, flags, etc. though without too much trouble.

> The fact that we're using remotes for install but collections for update seems like it's going to be weird, but that's inherent with the model I guess.

I don’t think that’s quite right, but I’m not entirely sure what you mean, so I can’t concretely refute it. The remotes get used for both install and update: they provide the keyring configuration. Collections are also used for both install and update: they provide the global naming resolution for refs.

src/libostree/ostree-repo-refs.c
```diff
@@ -820,15 +862,133 @@ _ostree_repo_update_refs (OstreeRepo        *self,
...
+
+      if (!enumerate_refs_recurse (self, NULL, main_collection_id, refs_heads_dfd, base_path,
+                                   refs_heads_dfd, ".",
+                                   ret_all_refs, cancellable, error))
+        return FALSE;
+    }
+
+  g_string_truncate (base_path, 0);
+
+  if (!glnx_dirfd_iterator_init_at (self->repo_dir_fd, "refs/mirrors", TRUE, &dfd_iter, error))
```
cgwalters: This needs to handle `ENOENT` for existing repositories.  This is why the installed tests are failing, since they are using a repository generated with an older libostree.

cgwalters: (See `ot_dfd_iter_init_allow_noent()`)

src/libostree/libostree-devel.sym
```diff
@@ -19,7 +19,9 @@
 
 /* Add new symbols here.  Release commits should copy this section into -released.sym. */
 LIBOSTREE_2017.8 {
-};
+global:
+  ostree_validate_remote_name;
+} LIBOSTREE_2017.7;
```
cgwalters: This version number is wrong now.  Can you split this out into a separate PR?

src/libostree/ostree-core.c
```diff
@@ -183,6 +184,35 @@ ostree_validate_rev (const char *rev,
 }
 
 /**
+ * ostree_validate_remote_name:
+ * @remote_name: A remote name
+ * @error: Error
+ *
+ * Returns: %TRUE if @remote_name is a valid remote name
+ * Since: 2017.7
```
cgwalters: This one is `Since 2017.8` though.

But yeah, since this patch is nicely isolated, would you mind splitting it to a separate PR?

tests/test-symbols.sh
```diff
@@ -22,7 +22,7 @@ set -xeuo pipefail
 echo '1..3'
 
 released_syms=${G_TEST_SRCDIR}/src/libostree/libostree-released.sym
-if echo "$OSTREE_FEATURES" | grep --quiet --no-messages "DevelBuild"; then
+if echo "$OSTREE_FEATURES" | grep --quiet --no-messages "devel"; then
```
cgwalters: Oh duh, and the change still passed because at the time there were no `-devel` symbols.

Can you split out this one too?

cgwalters:
I pulled those two patches and did some minor fixups, also added tests for `ostree_validate_remote_name()` in https://github.com/ostreedev/ostree/pull/948

src/libostree/ostree-core.c
pwithnall: (For posterity, Colin split this out as PR #948.)

pwithnall: (I’ve opened PR #950 to fix the `Since` line.)

tests/test-symbols.sh
pwithnall: (For posterity, Colin split this out as PR #948.)

pwithnall:
I’ve fixed the `Since` lines throughout, fixed the `ENOENT` problem, and rebased on PR #950 to avoid conflicts with the docs.

pwithnall:
bot, retest this please

tests/test-find-remotes.sh
```diff
@@ -0,0 +1,225 @@
...
+
+${CMD_PREFIX} ostree --repo=local refs > refs
+assert_file_has_content refs "^apps-remote:app1$"
+assert_file_has_content refs "^os-remote:os/amd64/master$"
+
+${CMD_PREFIX} ostree --repo=local refs --collections | wc -l > refscount
+assert_file_has_content refscount "^0$"
+
+# Create a local mirror repository where we pull the branches *in mirror mode* from the two remotes.
+# This should pull them into refs/mirrors, since the remotes advertise a collection ID.
```
cgwalters: I think we may need a way to mirror into `refs/heads` as before - otherwise this will be a backcompat issue since older versions of libostree won't know how to pull from `refs/mirrors` right?

Although I guess this requires using `remote add --collection-id`...so if I want to run a "transparent mirror", I'd omit the collection ID?

tests/test-find-remotes.sh
```diff
@@ -0,0 +1,225 @@
...
+mkdir -p files
+pushd files
+${CMD_PREFIX} ostree --repo=../os-collection commit -s "Test os-collection commit 1" -b os/amd64/master --gpg-homedir=${TEST_GPG_KEYHOME} --gpg-sign=${TEST_GPG_KEYID_2} > ../os-checksum
+popd
+${CMD_PREFIX} ostree --repo=os-collection summary --update --gpg-homedir=${TEST_GPG_KEYHOME} --gpg-sign=${TEST_GPG_KEYID_2}
+
+# Create a local repository where we pull the branches from the two remotes as normal, using GPG.
+mkdir local
+ostree_repo_init local
+${CMD_PREFIX} ostree --repo=local remote add apps-remote file://$(pwd)/apps-collection --collection-id apps-collection --gpg-import=${test_tmpdir}/gpghome/key1.asc
```
cgwalters: Can we define and use best-practice for collection names?  Is it kebab case like current remote names?  DNS?  So far I've been trying to use e.g. `exampleos` for newer ostree tests.

Is there a reason to have the remote name differ from the collection ID by default?

Continuing with the kebab case, this would then be `exampleos-apps` for both the remote and the collection?

tests/test-find-remotes.sh
pwithnall: If we have a setup where the upstream repository exposes two collection IDs (C1, C2) each with two refs (R1, R2), it could advertise them in its `summary` file in two different ways.

Case A: refs list empty, collection ID is empty, collection map contains `{ C1: [R1, R2], C2: [R1, R2] }`.

Case B: refs list contains [R1, R2], collection ID is C1, collection map contains `{ C2: [R1, R2] }`.

The difference between these cases is a result of how the owner of the repository sets it up (libostree and the command line tools support both configurations).

If the libostree which is pulling from this upstream repository (to create a mirror, or as a normal client) is old, case A will fail, and case B will result in `refs/heads/R1` and `refs/heads/R2` (which are the C1 refs).

If the libostree which is pulling is new, both cases will result in `refs/mirrors/{C1,C2}/{R1,R2}`. We can’t mirror it to create a repository identical to case B, since the collection ID in the summary is meant to be used for upstream repositories only.

So I think this is a configuration thing. Generally, people deploying repositories should configure them as for case B, with the refs the repository lists in `refs/heads` being those it’s always listed; and the refs in `refs/mirrors` being new, additional ones.

The case we can’t support is an old libostree client pulling from a mirror created with a new libostree (i.e. the `refs/mirrors/{C1,C2}/{R1,R2}` mirror). I don’t think this is a problem in the setups I’ve been thinking about — mirroring would be used over P2P, and in order to access repositories over P2P, the client would have to be using a new libostree anyway.

Does that cover what you had in mind?

> Although I guess this requires using `remote add --collection-id`...so if I want to run a "transparent mirror", I'd omit the collection ID?

If `--collection-id` is omitted from `remote add` when setting up the mirror, I don’t think it’ll make a difference to the `pull --mirror`, since the code in `ostree-repo-pull.c` to process the refs bases it all off the downloaded `summary` file, rather than the remote’s configuration. That `--collection-id` argument is there to allow `find-remotes` to use that remote configuration for P2P resolution of refs in the `find-remotes apps-collection app1` invocation (and others) lower down in the test.

tests/test-find-remotes.sh
cgwalters: I guess reverse DNS is more obviously globally unique, so yeah, let's do that.

pwithnall: Changed in a series of fixups. Rebased to today’s master just before that and squashed in some prior fixups.

cgwalters:
I'm looking at `git diff origin/master -- src/libostree/ostree-repo-pull.c`.  I know we talked about this before some but I'm uncertain about `ostree_repo_pull_from_remotes_async()`.  It's basically trying to be a superset of the remote-based pull, right?

One way I was modeling this mentally at least before is that collections/remote finding is an *optimization* or an option for remote-based pulls.  We could just have an option like `enable-collection` or something, and transparently use the repo finders, etc.  The more we go down this route the less collections really need to be exposed, right?  Clearly it's nice to be able to see in advance (and potentially control) the remote set, but is that worth all of the new API surface?

A problem we talked about was progress reporting.  My vote here is that we only pull from at one most "non local HTTP" remote at a time.  And we don't do fine grained progress for files and avahi sources.  My intuition here is:

 - They're going to be fast enough it's not worth the trouble.   Perhaps for avahi we may want to turn on libcurl's `CURLOPTLOWSPEEDLIMIT` just as a sanity check and to be robust against broken peers. 
 - These sources are going to be enabled "in the background"; particularly in an automatic updates scenario, the user wouldn't have explicitly clicked to update, so there's no real reason to do gymnastics trying to present progress that they won't see.

pwithnall:
> I'm looking at `git diff origin/master -- src/libostree/ostree-repo-pull.c`. I know we talked about this before some but I'm uncertain about `ostree_repo_pull_from_remotes_async()`. It's basically trying to be a superset of the remote-based pull, right?

Yes, plus to roll in some other changes like `GAsyncResult`-based async support, and from-the-start support for splitting the metadata and content downloads by putting the resolved metadata in the `OstreeRepoFinderResult` array.

> One way I was modeling this mentally at least before is that collections/remote finding is an optimization or an option for remote-based pulls. We could just have an option like `enable-collection` or something, and transparently use the repo finders, etc. The more we go down this route the less collections really need to be exposed, right? Clearly it's nice to be able to see in advance (and potentially control) the remote set, but is that worth all of the new API surface?

We [discussed this before](https://public.etherpad-mozilla.org/p/ostree-avahi) and the decision was that the new API was OK, since it gives us the advantages:
 - split between remote finding+metadata, and content download
 - better visibility of which remotes are being pulled from
 - proper async support
 - potentially improved progress reporting (although I then looked at the different options for progress reporting, and decided `OstreeAsyncProgress` was already pretty optimal)

What’s changed? 

We could reduce the API surface area by not making `OstreeRepoFinder` (and its subclasses) public, but then we lose the ability to configure per-finder properties, like the `GMainContext` for the Avahi finder, or the `GVolumeMonitor` instance for the mount finder. They would still have to be exposed in the API as keyword arguments to `ostree_repo_find_remotes_async()` so that individual finders can be enabled or disabled — so we’d be trading C ABI for keyword ABI.

> A problem we talked about was progress reporting. A strawman here is that we only pull from at one most "non local HTTP" remote at a time (what I think of as "canonical remote"). And we don't do fine grained progress for files and avahi sources.

Why not do fine-grained progress for local/Avahi pulls? The code currently pulls from the resolved URIs one at a time, although this could be optimised in future. At the moment, the progress reporting is wrong (it will go from 0 to 100% for each URI being pulled from), but that’s something I’d like to look at in a follow-up commit once the flatpak side of things is done. For the moment, I’m happy in the knowledge that `OstreeAsyncResult` can support whatever progress reporting regime we want.

> My intuition here is:
>  - They're going to be fast enough it's not worth the trouble. Perhaps for avahi we may want to turn on libcurl's `CURLOPTLOWSPEEDLIMIT` just as a sanity check and to be robust against broken peers.

That looks like a useful option. I’d like to do that as a follow-up commit though, once this lot is sorted and I’ve got the flatpak side of things done.

> - These sources are going to be enabled "in the background"; particularly in an automatic updates scenario, the user wouldn't have explicitly clicked to update, so there's no real reason to do gymnastics trying to present progress that they won't see.

They would also be enabled when the user explicitly pressed ‘update’. For example, definitely in the case where the user mounts a USB stick containing updates they want to install. They will also be enabled in the background.

cgwalters:
You're right, sorry I should have referenced that etherpad.  I remembered we'd discussed it, but not where we ended up.  OK so I think I'm back to landing this and trying it out.

pwithnall:
> You're right, sorry I should have referenced that etherpad. I remembered we'd discussed it, but not where we ended up. OK so I think I'm back to landing this and trying it out.

OK. I’ve just rebased it on the latest master to fix the conflict with your recent fsck cleanup.

cgwalters:
@rh-atomic-bot r+ de43eb5

pwithnall:
Fantastic, thanks!

PR #926, opened 2017-06-13
lib/pull: Add OSTREE_REPO_PULL_FLAGS_BAREUSERONLY_FILES

cgwalters:
This is an option which is intended mostly for flatpak;
see: https://github.com/flatpak/flatpak/issues/845

We're adding an option for pulling into *all*
repo modes that has an effect similar to the `bare-user-only`
change from https://github.com/ostreedev/ostree/pull/909

This way one can pull content into e.g. a root-owned `bare` repository and
ensure that there aren't any setuid or world-writable files.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -111,6 +111,7 @@ typedef struct {
   gboolean          is_mirror;
   gboolean          is_commit_only;
   gboolean          is_untrusted;
+  gboolean          is_bareuseronly_files;
 
   GPtrArray        *dirs;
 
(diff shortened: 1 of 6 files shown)


cgwalters:
Quick self-review: Hm, we aren't handling the HTTP case.  Should either make than an error or make it work.

cgwalters:
And maybe this is really better as a repository option that does both files and dirs?  Then flatpak could edit the repo config to turn it on.

alexlarsson:
LGTM

@rh-atomic-bot r+ be9f181

PR #927, opened 2017-06-13
lib/checkout: Add bareuseronly_dirs option

cgwalters:
This is a continuation of https://github.com/ostreedev/ostree/pull/926
for directories instead of files.

See: https://github.com/flatpak/flatpak/issues/845

This option suppresses mode bits outside of `0775` for directory
checkouts.  I think most people should start doing this by default,
and use explicit overrides for e.g. `/tmp` if doing a recommit based
on a checkout.


alexlarsson:
@rh-atomic-bot r+ f0db53b

cgwalters:
Rebased :surfing_man: 

@rh-atomic-bot r=alexlarsson

alexlarsson:
@rh-atomic-bot r+ cc447dc

PR #928, opened 2017-06-13
build-sys: post-release version bump

cgwalters:
Per request by flatpak.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,7 +1,7 @@
 AC_PREREQ([2.63])
 dnl If incrementing the version here, remember to update libostree.sym too
 m4_define([year_version], [2017])
-m4_define([release_version], [6])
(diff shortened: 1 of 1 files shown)


dustymabe:
is there some way to know that this is basically a 'dev version' since we havn't released `.7` yet?

cgwalters:
(I mean yes, kind of `ostree --version` might help with that but the exact semantics there get murky and I'd like to know exactly what would be using that and why)

alexlarsson:
I don't think any released software should rely on an unreleased ostree version.
However, its very nice to have a check during development that at least you need an unstable version.
Obviously the ostree unstable version could change abi or we add api several times, but at least its better than nothing.

alexlarsson:
@rh-atomic-bot r+ 70118ed

PR #930, opened 2017-06-14
lib/pull: Extend BAREUSERONLY_FILES flag to HTTP requests

cgwalters:
Working on the flatpak PR: https://github.com/flatpak/flatpak/pull/849 It's
really more convenient if this works for HTTP pulls as well, since flatpak does
various types of pulling, and we can just set the flag everywhere.

Further, we might as well reject the content as early as possible.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -871,7 +871,12 @@ content_fetch_on_complete (GObject        *object,
   checksum_obj = ostree_object_to_string (checksum, objtype);
   g_debug ("fetch of %s complete", checksum_obj);
 
-  if (pull_data->is_mirror && pull_data->repo->mode == OSTREE_REPO_MODE_ARCHIVE_Z2)
+  /* If we're mirroring and writing into an archive repo, we can directly copy
+   * the content rather than paying the cost of exploding it, checksumming, and
+   * re-gzip.
+   */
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -871,7 +871,12 @@ content_fetch_on_complete (GObject        *object,
   checksum_obj = ostree_object_to_string (checksum, objtype);
   g_debug ("fetch of %s complete", checksum_obj);
 
-  if (pull_data->is_mirror && pull_data->repo->mode == OSTREE_REPO_MODE_ARCHIVE_Z2)
+  /* If we're mirroring and writing into an archive repo, we can directly copy
+   * the content rather than paying the cost of exploding it, checksumming, and
+   * re-gzip.
+   */
```
alexlarsson: Does this mean that pull --mirror is not validating checksums?

alexlarsson:
@rh-atomic-bot r+ 19b6a44

src/libostree/ostree-repo-pull.c
cgwalters: `pull --mirror` into `archive` repos, yes.  That's currently intentional, since if we want to do mirroring of static deltas we'll basically need that.  Though if someone wanted to change that we could try to validate, but it'd get very expensive for deltas at least.

PR #931, opened 2017-06-15
lib: Split symbol versioning into -released and -devel

cgwalters:
So far a lot of submitted PR have added symbols into the first
section.  Split the file into `-released` and `-devel` to make
this more obvious.

To further enforce things, we hardcode a checksum of the `-released`
file in `test-symbols.sh`.  Only release commits should update that
checksum.

Did you notice I like checksums?


cgwalters:

pwithnall:
Looks good in theory, but in practice the CI is not happy:
```
Verifying all public symbols are documented:
+ sha256sum -c released-sha256.txt
ok documented symbols
sha256sum: WARNING: 1 computed checksum did NOT match
```

cgwalters:
Hm, not quite sure how that happened.  Updated :arrow_up:

pwithnall:
@rh-atomic-bot r+ a295280c1aa88d3ec26ce3d4fe2133ac434ec474

PR #932, opened 2017-06-15
ostreee-version.h.in: Added Since: version annotations

gtristan:
This is especially interesting for the versioning symbols themselves,
as it is an indicator of when applications using introspection information
can start to use a symbol in the library to check if they have a
recent enough version of OSTree to use.


cgwalters:
bot, add author to whitelist

cgwalters:
@rh-atomic-bot r+  	9bcc9f7

Issue #846, opened 2017-05-10
Partial pull --mirror overwrites summary


alexlarsson:
If i mirror a specific ref (or sets of refs) and the source repo has a summary file, then the target summary file is replaced. I don't really think this is the right behaviour unless you're mirroring the entire source repo.

cgwalters:
Yeah, agree we shouldn't do this.  In the short term one can work around this by regenerating the summary after, though it does introduce a race if clients are going to pull from it directly.

alexlarsson:
I'm hitting this when mirroring some of the gnome runtimes into the flathub repo. Will look at it.

PR #935, opened 2017-06-16
pull: When mirroring, only replace summary if we're doing a full mirror

alexlarsson:
We're hitting this in flathub, where we have a bunch of local builds,
but we also mirror a few refs from the gnome runtime repo into it.
Its fixable by re-doing the summary, but for a short time the
wrong version is visible.

Fixes https://github.com/ostreedev/ostree/issues/846


cgwalters:
Added test cases.

@rh-atomic-bot r+ 81f71e8

PR #936, opened 2017-06-16
checkout: Fix SELinux policy labeling when recursing

cgwalters:
The code here tried to truncate the string to the previous length,
but that doesn't work when recursing, since further calls change the
length.

What actually ended up happening was the string would get corrupted
after the first level of recursion.


jlebon:
Ouch, tricksy business!
@rh-atomic-bot r+ a263462

PR #937, opened 2017-06-16
tests: Fix previous commit for selinux testing

cgwalters:
I only checked the test passed, I didn't read the output closely, and
made it succeed without testing anything.

Fix the absolute/relative `/etc` references.

### DIFF
--- a/tests/installed/itest-deploy-selinux.sh
+++ b/tests/installed/itest-deploy-selinux.sh
@@ -12,9 +12,9 @@ ostree admin deploy --karg-proc-cmdline ${host_refspec}
 new_deployment_path=/ostree/deploy/${host_osname}/deploy/${host_commit}.1
 
 # A set of files that have a variety of security contexts
-for file in fstab passwd exports hostname sysctl.conf /etc/yum.repos.d \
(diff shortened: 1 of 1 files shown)


tests/installed/itest-deploy-selinux.sh
```diff
@@ -12,9 +12,9 @@ ostree admin deploy --karg-proc-cmdline ${host_refspec}
 new_deployment_path=/ostree/deploy/${host_osname}/deploy/${host_commit}.1
 
 # A set of files that have a variety of security contexts
-for file in fstab passwd exports hostname sysctl.conf /etc/yum.repos.d \
-            /etc/NetworkManager/dispatcher.d/hook-network-manager; do
-    if ! test -e ${file}; then
+for file in fstab passwd exports hostname sysctl.conf yum.repos.d \
```
jlebon: Heh, because `yum.repos.d` is a repo, the `ls` calls below actually output a full dir listing. Though since the output is stable, it actually compares fine.

jlebon:
@rh-atomic-bot r+ c7aadda

PR #939, opened 2017-06-17
lib/repo: Fix annotations for out parameters

dashea:
Change the annotation of the out parameters on ostree_repo_load_file
from `(allow-none)` to `(optional) (nullable)`. `allow-none` is
ambiguous, since these parameters can be both NULL on input and set to
NULL on return.

This issue was encountered using the Haskell bindings, see https://github.com/haskell-gi/haskell-gi/issues/106


cgwalters:
@rh-atomic-bot r+ 3749919

cgwalters:
@rh-atomic-bot retry

PR #940, opened 2017-06-19
static delta apply: Work on bare-user-only repos

alexlarsson:
Flatpak make check is failing when applying a static delta
to a bare-user-only repo due to an assert. The fix is to add
bare-user-only to the assert check.


giuseppe:
@rh-atomic-bot r+ 4f21edb

PR #942, opened 2017-06-19
Release 2017.7

cgwalters:


cgwalters:
Here's the postrelease, which I'll do as a separate PR to make sure CI tests the release commit:

```
From 94a5aaccabd29279555b2e0435280a0914a711c6 Mon Sep 17 00:00:00 2001
Date: Mon, 19 Jun 2017 10:37:19 -0400
Subject: [PATCH] build-sys: Post-release version bump

---
 configure.ac                      | 4 ++--
 src/libostree/libostree-devel.sym | 2 ++
 2 files changed, 4 insertions(+), 2 deletions(-)

index 7946f4f..9366844 100644
--- a/configure.ac
+++ b/configure.ac
@@ -4,10 +4,10 @@ dnl update libostree-released.sym from libostree-devel.sym, and update the check
 dnl in test-symbols.sh, and also set is_release_build=yes below.  Then make
 dnl another post-release commit to bump the version, and set is_release_build=no.
 m4_define([year_version], [2017])
-m4_define([release_version], [7])
+m4_define([release_version], [8])
 m4_define([package_version], [year_version.release_version])
 AC_INIT([libostree], [package_version], [walters@verbum.org])
-is_release_build=yes
+is_release_build=no
 AC_CONFIG_HEADER([config.h])
 AC_CONFIG_MACRO_DIR([buildutil])
 AC_CONFIG_AUX_DIR([build-aux])
diff --git a/src/libostree/libostree-devel.sym b/src/libostree/libostree-devel.sym
index 01f182f..a9c3f61 100644
--- a/src/libostree/libostree-devel.sym
+++ b/src/libostree/libostree-devel.sym
@@ -18,6 +18,8 @@
 ***/
 
+LIBOSTREE_2017.8 {
+};
 
 /* Stub section for the stable release *after* this development one; don't
  * edit this other than to update the last number.  This is just a copy/paste
-- 
2.9.4

```

jlebon:
Hmm, I think it's time for a basic `RELEASE` SOP file. We can do that separately though.

We should also add a sanity check in the CI that:
`is_release_build=yes` iff commit title matches "Release $MAJOR.$MINOR"

That said, let's do this! 🚄
@rh-atomic-bot r+ 3973198

cgwalters:
> ERROR: tests/test-symbols.sh - too few tests run (expected 3, got 0)

Hm, travis fell over, and it's the only suite doing `distcheck` right now.  Will look at that.

cgwalters:
https://github.com/ostreedev/ostree/pull/944 fixes that.

Though I'm going to just point the `2017.7` tag at the release commit and add a known issue for `test-symbols.sh`.

PR #943, opened 2017-06-19
 Repouninit warning

cgwalters:
It's spurious, but unfortunately GCC doesn't currently understand that it will
always be set.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -12,7 +12,9 @@ packages:
   - git
 
 env:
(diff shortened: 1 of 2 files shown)


cgwalters:

cgwalters:
Or @jlebon if you're around, this is an easy one.

jlebon:
@rh-atomic-bot r+ 07db309

PR #944, opened 2017-06-19
Postrelease verbump

cgwalters:


jlebon:
Would you like to rebase on top of https://github.com/ostreedev/ostree/pull/945 first so we test that too at the same time? (We should see it fail there as expected since `is_release_build` is still `yes`, so we won't be able to merge it).

jlebon:
Actually, since it did fail correctly, we can merge this one now and rebase the other one.
@rh-atomic-bot r+ 5dd6da7

PR #945, opened 2017-06-19
ci: add ci-release-build.sh

jlebon:
Add a check that verifies that `is_release_build` is `yes` only for
release commits. And also verify that the commit message has the correct
version.


cgwalters:
@rh-atomic-bot r+  84f91c6

PR #946, opened 2017-06-19
Support verifying inline summary signatures

pwithnall:
This adds support for verifying inline signatures in `summary` files, including defining two new keys for containing such inline signatures (`OSTREE_SUMMARY_REFS_SIGNATURES` and `OSTREE_SUMMARY_ADDITIONAL_SIGNATURES`). It does not add support for generating inline signatures yet — the current plan is to modify `ostree_repo_add_gpg_signature_summary()` to read in, sign, and rewrite, the `summary` file. However, I wanted to get the schema of additional metadata keys sorted before implementing that.

Thoughts about the approach? This code is ready to merge apart from the fact it’s based on the lan-and-usb branch, which is still under review. Hence, please only look at the top 7 commits. Please also think about whether having `ostree_repo_add_gpg_signature_summary()` transparently rewrite the `summary` file to include inline signatures is your preferred approach for the write side of things.

Additional tests need to be added for the inline signature support, but they can only be added once libostree supports writing inline signatures; so they will be added with that. All the existing tests (for external `summary.sig` files) still pass.

---

This is based on the lan-and-usb branch and can’t be merged until that one is, as it depends on some of the new summary fields it introduces. The top 7 commits are relevant here.

### DIFF
--- a/apidoc/ostree-experimental-sections.txt
+++ b/apidoc/ostree-experimental-sections.txt
@@ -28,6 +28,7 @@ ostree_repo_find_remotes_finish
 ostree_repo_pull_from_remotes_async
 ostree_repo_pull_from_remotes_finish
 ostree_repo_resolve_keyring_for_collection
+OSTREE_REPO_METADATA_REF
 </SECTION>
 
 <SECTION>
--- a/src/libostree/ostree-repo.h
+++ b/src/libostree/ostree-repo.h
@@ -1244,6 +1244,32 @@ gboolean ostree_repo_regenerate_summary (OstreeRepo     *self,
                                          GCancellable   *cancellable,
                                          GError        **error);
 
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+
+/**
+ * OSTREE_REPO_METADATA_REF:
+ *
+ * The name of a ref which is used to store metadata for the entire repository,
+ * such as its expected update time (`ostree.summary.expires`), name, or new
+ * GPG keys. Metadata is stored on contentless commits in the ref, and hence is
+ * signed with the commits.
+ *
+ * This supersedes the additional metadata dictionary in the `summary` file
+ * (see ostree_repo_regenerate_summary()), as the use of a ref means that the
+ * metadata for multiple upstream repositories can be included in a single mirror
+ * repository, disambiguating the refs using collection IDs. In order to support
+ * peer to peer redistribution of repository metadata, repositories must set a
+ * collection ID (ostree_repo_set_collection_id()).
+ *
+ * Users of OSTree may place arbitrary metadata in commits on this ref, but the
+ * keys must be namespaced by product or developer. For example,
+ * `exampleos.end-of-life`. The `ostree.` prefix is reserved.
+ *
+ * Since: 2017.8
+ */
+#define OSTREE_REPO_METADATA_REF "ostree-metadata"
+
+#endif  /* OSTREE_ENABLE_EXPERIMENTAL_API */
 
 G_END_DECLS
 


cgwalters:
I haven't looked at this in detail yet, but one thing I'm still thinking about is the discussion on the list about metadata in the summary vs metadata in the commits.

Didn't we conclude that it was necessary to have some metadata in the commits as well, like the existing flatpak `xa.ref`?   We already need to fetch the commits anyways, right?  If we also include the collection ID in commits, do we need summary signatures?

pwithnall:
> I haven't looked at this in detail yet, but one thing I'm still thinking about is the discussion on the list about metadata in the summary vs metadata in the commits.
> 
> Didn't we conclude that it was necessary to have some metadata in the commits as well, like the existing flatpak `xa.ref`?

Yup, @krnowak is working on that. He posted some [questions about it on the mailing list here](https://mail.gnome.org/archives/ostree-list/2017-June/msg00034.html), and has a [WIP branch here](https://github.com/kinvolk/ostree/tree/krnowak/collection-ref-in-commit-metadata-T17610).

> We already need to fetch the commits anyways, right?

Yup.

> If we also include the collection ID in commits, do we need summary signatures?

Yes, for two reasons.

# Signing the complete refs list in the summary file allows revocation of refs (by dropping them from the list then re-signing it). That’s not something which we immediately need, but seems like a sensible property to have. The only way I can think of to support revocation using detached metadata would be to have a TTL on the detached metadata, and require it to be re-signed every x days in order to remain valid.
# There could be additional metadata in the summary file which needs to be signed. Alex had some in mind for flatpak (download sizes, iirc). The `ostree.summary.expires` key is another one. Other metadata keys requiring signing may appear in future (well, anyone using additional metadata at the moment could be assuming that it’s signed as per `summary.sig`).

pwithnall:
While working on the tests for this I spotted a few API niggles, so please don’t review in detail yet. I might need to rework some of it.

cgwalters:
flatpak is already injecting sizes into the commit metadata; see `flatpak_builtin_build_export()`.

As far as revocation...I dunno.  Is there someone who wants that specifically?  Is this like the case where e.g. flathub detects malware and wants to ban the app?  But there's higher level issues there like whether or not the app should be uninstalled; you'd want to render that in the client prominently and likely with some explanatory text, and not just have signatures start to fail.

I'm really wary of introducing a new way to sign things.  If we can only sign commits then the whole model is significantly simpler.

pwithnall:
> flatpak is already injecting sizes into the commit metadata; see flatpak_builtin_build_export().

There is other metadata, like `xa.title` (which applies to the whole repository, and wouldn’t really sit well as per-commit metadata) or `xa.default-branch` (although that one could be put into per-commit metadata). `xa.gpg-keys` (for key rotation) wouldn’t really make much sense as per-commit data.

@alexlarsson, what do you think? Does flatpak have a need for signed per-repository data (in the `summary` file), or can it all be put into signed per-commit metadata?

> As far as revocation...I dunno. Is there someone who wants that specifically? Is this like the case where e.g. flathub detects malware and wants to ban the app? But there's higher level issues there like whether or not the app should be uninstalled; you'd want to render that in the client prominently and likely with some explanatory text, and not just have signatures start to fail.

There is no specific use case calling for revocation. I just think it’s generally a useful principle to keep — but if it’s the only reason to want signatures in the `summary` file, that’s not compelling.

> I'm really wary of introducing a new way to sign things. If we can only sign commits then the whole model is significantly simpler.

Fair. There still seems to me to be a need to have *some* signed per-repository data though.

pwithnall:
Just pushed an updated version of the branch which reworks how the signing happens a bit (so it will now only sign metadata which relates to the collection ID returned by `ostree_repo_get_collection_id()`) and it also includes the collection ID in each plaintext so that it’s bound into the signature. The new branch also adds some documentation, some tests, and support for writing inline signatures too (bonded into `ostree_repo_add_gpg_signature_summary()`).

alexlarsson:
We have some repo global things we want to sign, such as the above title and default-branch. But also for migration to new url and gpg keys. I don't want these to be in each commit in the repo. However, moving them to a separate (content-less) branch, called say `flatpak-info` or something sounds like a fine approach.

pwithnall:
> We have some repo global things we want to sign, such as the above title and default-branch. But also for migration to new url and gpg keys. I don't want these to be in each commit in the repo. However, moving them to a separate (content-less) branch, called say `flatpak-info` or something sounds like a fine approach.

Hmm, I think that would work, and would give the benefits of not inventing a new way of signing things, and binding all the properties being signed together with a commit timestamp which gives an indication of freshness.

So, to be clear, are you suggesting:
 - Keep `summary` and list refs in it. Additional metadata may be stored in it, but should only be considered as a cache.
 - Continue to generate `summary.sig` (for backwards compatibility), but newer versions of libostree should only use it to verify things if ref names aren’t listed in commit metadata for verification there.
 - Older versions of libostree would be unaware of the ref names in commit metadata, and would continue to verify and use `summary{,.sig}`.
 - Per-repository metadata moves to a `ostree-metadata` ref (I think that dropping ‘flatpak’ from the name makes sense, since metadata from other projects could also be stored there). Metadata keys in the commits on that branch should be namespaced (`xa.blah`, etc.) as normal.
 - In order to allow P2P redistribution of that metadata, a repository must have ref names in commit metadata, and any per-repo metadata must be put on an `ostree-metadata` ref. That ref is not mandatory, since there might be no per-repo metadata. Redistribution of that ref would be like any other: it would require the repository to have a collection ID.
 - We don’t add support for inline signatures in `summary` files, since the summary file is now only treated as a cache and all the verification happens on commit metadata instead.

I guess that introduces one subtlety about the metadata for migration to a new URL: it applies to the canonical upstream URL for the collection *only*, and not to any mirrors of the collection, even if those are the ones which are actually serving the content to a peer. That should be fine. It’s just something to be careful to document.

---

@cgwalters, what do you think?

alexlarsson:
Not sure about this:

> but newer versions of libostree should only use it to verify things if ref names aren’t listed in commit metadata for verification there.

There is already a separate boolean property for whether to verify the summary, Maybe we can just stop setting that to true everywhere. Then users can decide what they want?

pwithnall:
> Not sure about this:
> 
>     but newer versions of libostree should only use it to verify things if ref names aren’t listed in commit metadata for verification there.
> 
> There is already a separate boolean property for whether to verify the summary, Maybe we can just stop setting that to true everywhere. Then users can decide what they want?

That wouldn’t interact well with P2P support, where we *can’t* re-sign the summary file for a repository which mirrors multiple collections. In terms of security, it would also be better to be clear about where the verification is happening in different situations, rather than ‘it could happen here, or maybe there, depending on X, Y, Z’·

cgwalters:
The `ostree-metadata` approach does indeed have a lot of nice properties.  I think though we should put it in `$repo/refs/metadata` or just `$repo/metadata` or something (not under `refs/heads`).  That way older clients won't mirror it, it won't show up under `ostree remote refs` etc.  Older clients won't interact with it at all, which is good becuase they couldn't make sense of it.

I'd say we make it an actual filesystem tree as well.  Maybe something like `/v0` is a GVariant (could be a tuple, then we introduce `v1` for new data?).  Or we could make `/v0` an `a{sv}` and use other file names for larger content.  (Acutally, how does appstream and things like screenshots fit in here potentially?)

alexlarsson:
True, it does not necessarily have to be commit metadata. As long as the commit is signed we can trust the entire content.

Flatpak stores appstream data in per-arch branches:
```
$ ostree --repo=/var/lib/flatpak/repo remote refs gnome | grep appstream
gnome:appstream/aarch64
gnome:appstream/arm
gnome:appstream/i386
gnome:appstream/x86_64

$ ls /var/lib/flatpak/appstream/flathub/x86_64/active/
appstream.xml.gz  icons
$ ls /var/lib/flatpak/appstream/flathub/x86_64/active/icons/
128x128  64x64
$ ls /var/lib/flatpak/appstream/flathub/x86_64/active/icons/64x64/
com.github.dahenson.agenda.png        net.blockout.BlockOutII.png              org.audacityteam.Audacity.png        org.inkscape.Inkscape.png
com.github.needleandthread.vocal.png  net.minetest.Minetest.png                org.baedert.corebird.png             org.libretro.RetroArch.png
com.teeworlds.Teeworlds.png           net.olofson.KoboDeluxe.png               org.darktable.Darktable.png          org.megaglest.MegaGlest.png
com.tux4kids.tuxmath.png              net.sourceforge.atanks.png               org.debian.TuxPuck.png               org.musicbrainz.Picard.png
com.tux4kids.tuxtype.png              net.sourceforge.btanks.png               org.DolphinEmu.dolphin-emu.png       org.mypaint.MyPaint.png
com.uploadedlobster.peek.png          net.sourceforge.chromium-bsu.png         org.freeciv.Freeciv.png              org.ppsspp.PPSSPP.png
com.valvesoftware.Steam.png           net.sourceforge.ExtremeTuxRacer.png      org.frozen_bubble.frozen-bubble.png  org.seul.pingus.png
com.vinszent.GnomeTwitch.png          net.sourceforge.mars-game.png            org.gna.Warmux.png                   org.supertuxproject.SuperTux.png
io.github.GnomeMpv.png                net.sourceforge.Ri-li.png                org.gnome.Eolie.png                  org.tuxfamily.XMoto.png
io.github.Hexchat.png                 net.sourceforge.torcs.png                org.gnome.FeedReader.png             org.tuxpaint.Tuxpaint.png
io.github.Pithos.png                  net.sourceforge.TuxFootball.png          org.gnome.Lollypop.png               org.wesnoth.Wesnoth.png
io.gitlab.osslugaru.Lugaru.png        net.supertuxkart.SuperTuxKart.png        org.gnome.Nautilus.png               ws.openarena.OpenArena.png
io.mgba.mGBA.png                      nl.openoffice.bluefish.png               org.gnome.Polari.png
io.thp.numptyphysics.png              org.armagetronad.ArmagetronAdvanced.png  org.gnome.Recipes.png
```

alexlarsson:
On the client side you want it in refs/remotes/$remote though, because you want to be able to mirror this for multiple remotes.

cgwalters:
Hm, I hadn't realized that; so flatpak's `appstream` branch name already makes it impossible to use a single ostree repo as storage for multiple upstreams.

Eh.  Dunno.  Maybe things *are* simpler if people wanting to dedup content across multiple repos use something like the `parent` functionality.  Though there's a whole other pile of issues with that approach.

pwithnall:
> The ostree-metadata approach does indeed have a lot of nice properties. I think though we should put it in `$repo/refs/metadata` or just `$repo/metadata` or something (not under `refs/heads`). That way older clients won't mirror it, it won't show up under `ostree remote refs` etc. Older clients won't interact with it at all, which is good becuase they couldn't make sense of it.

Surely you *want* older clients to mirror it, since it’s important data about the repository. Even if the older clients can’t understand the branch, other clients which pull from the mirror might.

> I'd say we make it an actual filesystem tree as well. Maybe something like `/v0` is a GVariant (could be a tuple, then we introduce `v1` for new data?). Or we could make `/v0` an `a{sv}` and use other file names for larger content. (Acutally, how does appstream and things like screenshots fit in here potentially?)

What advantages would putting an `a{sv}` in a new file tree bring? The code for handling metadata already does that for us already; just that the metadata ends up in a `.commit` file rather than the file system. Putting the repository metadata in commit metadata has the advantage that `ostree pull --commit-metadata-only` will still pull the repository metadata, which you probably want even if you don’t want commit content.

> Hm, I hadn't realized that; so flatpak's `appstream` branch name already makes it impossible to use a single ostree repo as storage for multiple upstreams.

Correct. Hence collection IDs.

> Eh. Dunno. Maybe things are simpler if people wanting to dedup content across multiple repos use something like the parent functionality. Though there's a whole other pile of issues with that approach.

Collections! I’ve, like, just spent 2 months solving this problem!

cgwalters:
I don't understand how collections help with the dedup issue; I'm thinking of e.g. server-side dedup.  Like a corporate mirror.  The main issue I'm thinking about here is having GC/prune work.

cgwalters:
Sorry, wasn't thinking about this enough; `refs/mirrors` does indeed help with this.

pwithnall:
While we discuss the design for the summary signing stuff, I’ve split out some of the less controversial changes from this branch into PR #961.

cgwalters:
`ostree-metadata` sounds nice to me, though there are a few reasons to store the metadata as content files rather than in the commit itself:

 - We get natural dedup; the gpg keys aren't large, but they're not tiny either
 - It's natually extensible, and also introspectable via e.g. `ostree ls`
 - A commit has to reference a tree, even if it's empty

But this isn't a really strong feeling; we can always add content in the metadata later too.

pwithnall:
> `ostree-metadata` sounds nice to me, though there are a few reasons to store the metadata as content files rather than in the commit itself:
> 
>     We get natural dedup; the gpg keys aren't large, but they're not tiny either
>     It's natually extensible, and also introspectable via e.g. `ostree ls`
>     A commit has to reference a tree, even if it's empty
> 
> But this isn't a really strong feeling; we can always add content in the metadata later too.

OK, I’ll rework this branch to use an `ostree-metadata` branch for per-repo metadata. To start with, I’ll put the metadata in the commit metadata `a{sv}`, rather than as content. Given the keys we expect to be in the per-repo metadata (listed below); the fact that none of them are going to be updated very often,;and the fact that if the `ostree-metadata` branches from multiple collections are mirrored in a single repository, none of them are going to share significant amounts of data for deduplication anyway; I think that putting the `a{sv}` in the commit content would be a premature optimisation. As you say, we can always break data out into the content later if needed.

Keys we expect to potentially see in the per-repo metadata at the moment:
 - `ostree.summary.expires`
 - `xa.title`
 - `xa.default-branch`
 - `xa.gpg-keys`
 - `xa.redirect-url`

pwithnall:
Hmm, I don’t see how the existing API (`ostree_repo_regenerate_summary()`, `ostree_repo_verify_summary()`, `ostree_repo_add_gpg_signature_summary()`) can be modified to handle the `ostree-metadata` branch, since the `summary` file and commits are completely different (for example, those functions don’t know about transactions). I think the approach to take might be to implement this in flatpak, and then see if it’s sensible to bring any convenience API back into libostree — perhaps for loading metadata from the `ostree-metadata` branch. The most I could do in libostree at te moment would be to change the documentation for various functions to mention the new approach, and that `additional_metadata` in `ostree_repo_regenerate_summary()` should be treated as an unsigned cache.

Thinking about the interaction between `summary.sig` and refs in commit metadata, I think it might be difficult to guarantee that exactly one of them is verified, since they’re handled at quite different places in the code. How about introducing a new `verify-commit-refs` config option per remote which should be set instead of `gpg-verify-summary`? If it’s set, and a commit doesn’t list a ref in its metadata when it’s supposed to (i.e. when pulling that ref), verification fails and hence the pull fails. That would be something for @krnowak’s work in https://github.com/kinvolk/ostree/tree/krnowak/collection-ref-in-commit-metadata-T17610 though; slightly decoupled from this PR.

pwithnall:
> How about introducing a new `verify-commit-refs` config option per remote

@krnowak points out that we can use the `collection-id` option — if it’s set, refs in commit metadata will be required.

cgwalters:
There's a bigger picture question here if we go this way about the existence of the summary file at all right?  In other words, we could move the list of refs, deltas etc. into `ostree-metadata` too.

Clearly we'll have a backwards compat period where we'll need to keep the summary around for a while, but given that it's already optional, I don't think it'll be too painful.   While you're right that we'd likely need a new `ostree_repo_update_metadata()` API, I suspect if we made `ostree summary -u` also write `ostree-metadata` we'd cover a lot of build systems already.

alexlarsson:
@cgwalters NO! The whole point with inline signing is to disconnect the two so that you can update the summary (say when you add a branch) without having to re-sign the entire thing with the gpg key.

cgwalters:
What's the problem with "re-sign the entire thing"?  We're talking about signing a commit right?

alexlarsson:
Yeah, we want to require you to sign if you change the global metadata, say the xa.gpg-keys. However, if we built an app on some other machine (including signing its ref in its commit) we want to be able to import it into the repo without having to have the signing key on the machine with the entire repo.

Having a separate unsigned commit for the summary would *work*, but how is that different from the summary file?

cgwalters:
I don't think signing requires "having to have the signing key on the machine with the entire repo".  Fedora for example has dedicated signing system that's separate from the other servers - the ostree commit (and *just* the commit) is sent to the signing server to be signed.  Supporting that is a reason there are APIs in libostree.

(In contrast for example, RPMs are inline signed, so the *entire* RPM is sent to the signing server, which gets slow for some of those gigabyte RPMs).

cgwalters:
Now, I think your higher level argument makes sense - we introduce `ostree-metadata` that's signed, and keep the summary file as is today as an optionally-signed index.  But it is interesting to think about the case where the server is set up to sign all metadata - then we could add the refs etc. into `ostree-metadata` and drop the summary.

alexlarsson:
The signing doesn't have to be *on* the machine, but it has to be able to somehow send the request to the signing machine.

What i want is a situation where the built app can be signed in the completely internal networked build machine + signing server, and then integrated to the repo of all apps running on a machine that is not connected to the internal network.

Compare it to signing rpms (signing app commit) to signing the yum metadata (signing summary)

pwithnall:
If we move the refs map into `ostree-metadata`, how does a client work out the checksum of the latest commit on the `ostree-metadata` branch? Via `refs/mirrors/$collection_id/ostree-metadata`?

cgwalters:
I get what you're saying and I'm OK with supporting it, but I think a better architecture is to have 3 levels:

CDN -> canonical server (not accessible to internet at large, CDN can make requests to it, and just ssh from a whitelisted set of IPs, 2FA etc.) -> signing server (even more hardened)

That way one isn't creating an (IMO) artificial distinction between signing different types of things.  That said I know this has been hashed out on the list before and so I'm OK with separating signed/unsigned metadata.

cgwalters:
@pwithnall I was thinking it's literally `refs/ostree-metadata`.  So a client just does a `GET` for that.

pwithnall:
> @pwithnall I was thinking it's literally `refs/ostree-metadata`. So a client just does a `GET` for that.

That won’t work when mirroring from multiple upstream collections; the collection ID needs to be included in there. And then you essentially have `refs/mirrors/$collection_id/ostree-metadata`. It seems a bit redundant to require parallel code paths for handling `ostree-metadata` in almost exactly the same way as any other ref.

pwithnall:
@cgwalters, any thoughts about this one? Do you want to gate merging it on #972, or something else?

cgwalters:
Feels like we should do #972 first, no?  I'm OK either way though.

pwithnall:
#972 is in. This one’s ready for review whenever.

pwithnall:
(It would be great if this could get into the 2017.8 release, since that would make our downstream packaging a bit easier!)

cgwalters:
Sorry for injecting uncertainty here...I know we agreed to inline summary signatures earlier.  I'm hoping this plan for the metadata ref will end up simpler and easier to reason about.

So...this PR is now just adding a well-known `#define`.  That seems sane to me, but should really also deprecate signing the summary in the docs at the same time?  We haven't yet implemented updating that metadata ref.  The data format of `ostree-metadata` and the transition to it is as yet a little unclear to me.

I think basically what I'm saying is let's just add the `#define` but not the docs?

pwithnall:
> Sorry for injecting uncertainty here...I know we agreed to inline summary signatures earlier. I'm hoping this plan for the metadata ref will end up simpler, easier to reason about, and align better with the collections work.

Agreed, I think it’s a better approach and am happy to be going with it.

> So...this PR is now just adding a well-known `#define`. That seems sane to me, but should really also deprecate signing the summary in the docs at the same time? We haven't yet implemented updating that metadata ref. The data format of `ostree-metadata` and the transition to it is as yet a little unclear to me.

The thing is that I can’t find a sensible place to implement the new `ostree-metadata` stuff in libostree without introducing a bunch of new API to get/set repository metadata. libostree already provides all the necessary low-level API (ref and commit support) for clients to handle `ostree-metadata` themselves. I was planning on implementing support for it in flatpak (done here: https://github.com/flatpak/flatpak/pull/884) and later codifying the sensible-looking bits in libostree.

> I think basically what I'm saying is let's just add the `#define` but not the docs?

Works for me. It does seem a bit silly to deprecate the summary support without a documented replacement.

pwithnall:
Updated. I’ll keep the documentation changes locally and propose them at a future date, when we’re looking at moving `OSTREE_REPO_METADATA_REF` out from behind `--enable-experimental-api`.

cgwalters:
Ah, OK I'll see if I can look at that flatpak PR too.

@rh-atomic-bot r+ 718e581

cgwalters:
@rh-atomic-bot r+ 718e851

PR #948, opened 2017-06-20
Lan and usb cherrypicks

cgwalters:


cgwalters:
@rh-atomic-bot r+ a181be8

pwithnall:
The `Since` line was wrong. I’ve opened up PR #950 to fix that.

PR #949, opened 2017-06-21
build: Always include ostree-trivial-httpd.xml in tarballs

smcv:
If we `make dist` after configuring with `--disable-trivial-httpd-cmdline` (the default, as was done for 2017.7), the resulting tarball would not include the source code for this man page, making it fail to build from source when subsequently configured with `--enable-trivial-httpd-cmdline`.

(This could be detected by using `--enable-trivial-httpd-cmdline` in the `AM_DISTCHECK_CONFIGURE_FLAGS`, but I can't test that yet, because distcheck has other failures for me at the moment.)


cgwalters:
@rh-atomic-bot r+  	7b2b3a6

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #951, opened 2017-06-22
 lib/repo: Split archive/bare file parsing

cgwalters:
Prep for future cleanup patches (in particular I want an internal-only
version at first that uses a fd+`struct stat`) to avoid allocations.

The new version avoids lots of deep nesting of conditionals as well
by hoisting the "not found" handling to an early return.

There's a bit of code duplication between the two cases but it's
quite worth the result.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2650,20 +2650,17 @@ repo_load_file_archive (OstreeRepo *self,
                                           out_input, out_file_info, out_xattrs,
                                           cancellable, error);
     }
+  else if (self->parent_repo)
+    {
+      return ostree_repo_load_file (self->parent_repo, checksum,
+                                    out_input, out_file_info, out_xattrs,
+                                    cancellable, error);
+    }
   else
     {
-      if (self->parent_repo)
-        {
-          return ostree_repo_load_file (self->parent_repo, checksum,
-                                        out_input, out_file_info, out_xattrs,
-                                        cancellable, error);
-        }
-      else
-        {
-          g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
-                       "Couldn't find file object '%s'", checksum);
-          return FALSE;
-        }
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo.c
```diff
@@ -2614,225 +2614,199 @@ _ostree_repo_read_bare_fd (OstreeRepo           *self,
...
+      if (!glnx_fstat (fd, &stbuf, error))
+        return FALSE;
+
+      g_autoptr(GInputStream) tmp_stream = g_unix_input_stream_new (glnx_steal_fd (&fd), TRUE);
+      /* Note return here */
+      return ostree_content_stream_parse (TRUE, tmp_stream, stbuf.st_size, TRUE,
+                                          out_input, out_file_info, out_xattrs,
+                                          cancellable, error);
+    }
+  else
```
jlebon: Looks like you can avoid another indentation level by taking out this else.

src/libostree/ostree-repo.c
```diff
@@ -2614,225 +2614,199 @@ _ostree_repo_read_bare_fd (OstreeRepo           *self,
...
+                                        out_input, out_file_info, out_xattrs,
+                                        cancellable, error);
         }
-
-      if (fd != -1)
+      else
         {
-          tmp_stream = g_unix_input_stream_new (fd, TRUE);
-          fd = -1; /* Transfer ownership */
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
```
jlebon: `return glnx_throw` ?

(Bis for a few others below).

cgwalters: Can't, we use `G_IO_ERROR_NOT_FOUND`.  (Though, maybe we could set `errno = ENOENT; return glnx_throw_errno()` but let's not scope that in now?)

src/libostree/ostree-repo.c
cgwalters: Sure, fixup :arrow_down:

src/libostree/ostree-repo.c
jlebon: Ahh, right!

jlebon:
@rh-atomic-bot r+ 47f9292

PR #952, opened 2017-06-23
lib/repo: More cleanup of load_file() internals

cgwalters:
This is followon work from previous cleanups.  Basically
`stat_bare_content_object()` was the `fstatat()` logic
and `ostree_repo_read_bare_fd()` was the `openat()` implementation;
they duplicated some bits to find the object in staging, recurse
into parent etc.

Further, I wanted an internal-only version of this API which didn't allocate
`GFileInfo`/`GInputStream` but used a plain `fd` and `struct stat` to avoid
mallocs.

The end version here I think looks a lot nicer, since we deduplicate the various
`open()` calls in the different cases for example.

### DIFF
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -321,11 +321,14 @@ _ostree_repo_commit_trusted_content_bare (OstreeRepo          *self,
                                           GError             **error);
 
 gboolean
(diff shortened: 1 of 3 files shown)


jlebon:
Pheww, that was a tricky one to review! It does look better though.
Are you planning to provide an `fd` version of `ostree_repo_load_file`?
@rh-atomic-bot r+ 99ec3ec

PR #953, opened 2017-06-23
lib/deltas: Port to more to new code style

cgwalters:
Looking at the uses of `ostree_repo_load_file()` here.


jlebon:
Nice! @rh-atomic-bot r+ 8e1285b

PR #956, opened 2017-06-23
tests: add a syntax-check rule for glnx_prefix_error()

cgwalters:
Same as the errno variant; the colon-space `: ` thing got me in a different
patch.


jlebon:
@rh-atomic-bot r+ 66e2a1a

PR #957, opened 2017-06-23
commit tmpfile prep

cgwalters:
A few patches in prep for #958.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -590,6 +590,8 @@ write_content_object (OstreeRepo         *self,
       return glnx_throw (error, "Unsupported file type %u", object_file_type);
     }
 
+  guint64 size;
+
   /* For bare-user, convert the symlink target to the input stream */
   if (repo_mode == OSTREE_REPO_MODE_BARE_USER && object_file_type == G_FILE_TYPE_SYMBOLIC_LINK)
     {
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -614,30 +614,32 @@ write_content_object (OstreeRepo         *self,
...
 
-      /* For bare-user we can't store symlinks as symlinks, as symlinks don't
-         support user xattrs to store the ownership. So, instead store them
-         as regular files */
-      temp_file_is_regular = TRUE;
-      temp_file_is_symlink = FALSE;
-
       if (file_input != NULL)
         g_object_unref (file_input);
       /* Include the terminating zero so we can e.g. mmap this file */
```
jlebon: One thing I just noticed here. We're creating a new stream that has just the target name + `nul`, but down there where we create the tmpfile for it, we're still using the `size` from the `GFileInfo`, so we don't actually end up writing the trailing zero, right?

src/libostree/ostree-repo-pull.c
```diff
@@ -1049,22 +1047,24 @@ meta_fetch_on_complete (GObject           *object,
...
 
-  fd = openat (_ostree_fetcher_get_dfd (fetcher), temp_path, O_RDONLY | O_CLOEXEC);
+  fd = openat (_ostree_fetcher_get_dfd (fetcher), tmp_unlinker.path, O_RDONLY | O_CLOEXEC);
   if (fd == -1)
     {
       glnx_set_error_from_errno (error);
       goto out;
     }
 
+  /* Now delete it, keeping the fd open as the last reference); see comment in
```
jlebon: Nit: unbalanced parenthesis.

src/libostree/ostree-repo-commit.c
cgwalters: We do write it actually, because we drop into the `g_output_stream_splice()` path.  But our `fallocate()` is wrong.

I'm actually kind of dubious about the value of writing it, since it doesn't make sense to mmap small files.  But let's keep it.  Fixup and new commit :arrow_down:

cgwalters: (Also, nice catch!)

jlebon:
@rh-atomic-bot r+ b2d8253

cgwalters:
Weird, github failed to close this.

PR #958, opened 2017-06-24
 Port to GLnxTmpfile

cgwalters:
There's lots of mechanically replacing `OtTmpFile` with `GLnxTmpfile`;
the biggest changes are in the commit path.  Symlink commits are now
very clearly separated from regular files.  Symlinks are `OtCleanupUnlinkat`,
and regular files are `GLnxTmpfile`.

The commit codepath separates those as `_ostree_repo_commit_path_final()` and
`_ostree_repo_commit_tmpf_final()`. A nice aspect of all of this is that they
both *consume* the temporary on success. This avoids an extra spurious
`unlink()` call.

One of the biggest bits of code motion is in `commit_loose_regfile_object()`,
which no longer needs to care about symlinks. For the most parth though it's
just removing conditionals.

Update submodule: libglnx

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -600,7 +600,7 @@ write_content_object (OstreeRepo         *self,
   if (phys_object_is_symlink)
     {
       /* This will not be hit for bare-user or archive */
-      g_assert (self->mode == OSTREE_REPO_MODE_BARE || self->mode == OSTREE_REPO_MODE_BARE_USER);
+      g_assert (self->mode == OSTREE_REPO_MODE_BARE || self->mode == OSTREE_REPO_MODE_BARE_USER_ONLY);
       if (!_ostree_make_temporary_symlink_at (self->tmp_dir_fd,
                                               g_file_info_get_symlink_target (file_info),
                                               &tmp_unlinker.path,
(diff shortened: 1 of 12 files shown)


cgwalters:
An updated version of https://github.com/ostreedev/ostree/pull/861

cgwalters:
Prep in: https://github.com/ostreedev/ostree/pull/957

cgwalters:
Rebased :surfing_man: and fixed up commit message.  No longer WIP.

cgwalters:
(Though for some reason at least here github is failing to rerender this PR update)

src/libostree/ostree-repo-commit.c
```diff
@@ -615,24 +589,17 @@ write_content_object (OstreeRepo         *self,
...
-                                                         cancellable, error))
-        return FALSE;
-    }
-  else if (_ostree_repo_mode_is_bare (repo_mode) && phys_object_is_symlink)
+  /* Is it a symlink physically? */
+  if (phys_object_is_symlink)
     {
+      g_assert (_ostree_repo_mode_is_bare (repo_mode));
       /* Note: This will not be hit for bare-user mode because its converted to a
          regular file and take the branch above */
```
jlebon: Nit: this comment needs to be updated.

jlebon:
Hmm, we should set up testing against a CentOS kernel.

src/libostree/ostree-repo-commit.c
cgwalters: Done :arrow_down:

jlebon:
Let's do https://github.com/ostreedev/ostree/pull/968 first?

jlebon:
@rh-atomic-bot r+ 912adc6

PR #961, opened 2017-06-26
Summary signature cleanups

pwithnall:
Various cleanups to the `summary.sig` code paths, split out from #946. These should be suitable to apply to master without needing discussion about the design.

### DIFF
--- a/man/ostree-summary.xml
+++ b/man/ostree-summary.xml
@@ -73,7 +73,14 @@ Boston, MA 02111-1307, USA.
                 <term><option>-u</option></term>
 
                 <listitem><para>
-		  Update the summary file.
+                  Update the summary file.
+                </para><para>
+                  Any additional arguments to the command
+                  are treated as additional key–value pairs to be added to the
+                  summary file as additional metadata. They must be in the format
+                  <command><replaceable>KEY</replaceable>=<replaceable>VALUE</replaceable></command>
+                  or as two separate arguments. The values must be in GVariant
+                  text format.
                 </para></listitem>
             </varlistentry>
 
--- a/man/ostree-summary.xml
+++ b/man/ostree-summary.xml
@@ -79,8 +79,10 @@ Boston, MA 02111-1307, USA.
                   are treated as additional key–value pairs to be added to the
                   summary file as additional metadata. They must be in the format
                   <command><replaceable>KEY</replaceable>=<replaceable>VALUE</replaceable></command>
-                  or as two separate arguments. The values must be in GVariant
-                  text format.
+                  or as two separate arguments. The keys must be namespaced for
+                  your organisation or repository using a dot prefix. The values
+                  must be in GVariant text format. For example,
+                  <command>exampleos.end-of-life "@t 1445385600"</command>.
                 </para></listitem>
             </varlistentry>
 
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3322,7 +3322,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
 
     if (pull_data->gpg_verify_summary && bytes_summary && bytes_sig)
       {
-        glnx_unref_object OstreeGpgVerifyResult *result = NULL;
+        g_autoptr(OstreeGpgVerifyResult) result = NULL;
 
         result = ostree_repo_verify_summary (self, pull_data->remote_name,
                                              bytes_summary, bytes_sig,
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3343,6 +3343,20 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
         pull_data->summary_data = g_bytes_ref (bytes_summary);
         pull_data->summary = g_variant_new_from_bytes (OSTREE_SUMMARY_GVARIANT_FORMAT, bytes_summary, FALSE);
 
+        if (!g_variant_is_normal_form (pull_data->summary))
+          {
+            g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                                 "Not normal form");
+            goto out;
+          }
+        if (!g_variant_is_of_type (pull_data->summary, OSTREE_SUMMARY_GVARIANT_FORMAT))
+          {
+            g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                         "Doesn't match variant type '%s'",
+                         (char *)OSTREE_SUMMARY_GVARIANT_FORMAT);
+            goto out;
+          }
+
         if (bytes_sig)
           pull_data->summary_data_sig = g_bytes_ref (bytes_sig);
       }
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3322,18 +3322,11 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
 
     if (pull_data->gpg_verify_summary && bytes_summary && bytes_sig)
       {
-        g_autoptr(GVariant) sig_variant = NULL;
         glnx_unref_object OstreeGpgVerifyResult *result = NULL;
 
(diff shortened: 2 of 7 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -3320,6 +3320,24 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
         goto out;
       }
 
+    if (pull_data->gpg_verify_summary && bytes_summary && bytes_sig)
+      {
+        g_autoptr(GVariant) sig_variant = NULL;
+        glnx_unref_object OstreeGpgVerifyResult *result = NULL;
```
cgwalters: This can be `g_autoptr()` now too.

src/libostree/ostree-repo-pull.c
```diff
@@ -3320,6 +3320,24 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
+
+        sig_variant = g_variant_new_from_bytes (OSTREE_SUMMARY_SIG_GVARIANT_FORMAT, bytes_sig, FALSE);
+        result = _ostree_repo_gpg_verify_with_metadata (self,
```
cgwalters: This is optional now, but more and more of the code in libostree is being ported to decl-after-stmt, and this would be an easy one.

src/ostree/ot-builtin-summary.c
```diff
@@ -40,6 +40,63 @@ static GOptionEntry options[] = {
...
+argv_to_additional_metadata (int      *argc,
+                             char   ***argv,
+                             GError  **error)
+{
+  g_autoptr(GVariantBuilder) builder = NULL;
+  int i;
+
+  builder = g_variant_builder_new (G_VARIANT_TYPE_VARDICT);
+
+  for (i = 1; i < *argc; i++)
```
cgwalters: decl-after-stmt style applies here too.

src/ostree/ot-builtin-summary.c
```diff
@@ -40,6 +40,63 @@ static GOptionEntry options[] = {
...
+        }
+      else
+        {
+          key = g_strdup ((*argv)[i]);
+
+          if (i + 1 >= *argc)
+            {
+              g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                           "Invalid additional metadata list: not an even number of arguments");
+              return NULL;
```
cgwalters: This is also `glnx_null_throw()`

src/ostree/ot-builtin-summary.c
```diff
@@ -40,6 +40,63 @@ static GOptionEntry options[] = {
...
+
+          value_str = (*argv)[i + 1];
+          i++;
+        }
+
+      value = g_variant_parse (NULL, value_str, NULL, NULL, error);
+      if (value == NULL)
+        {
+          g_prefix_error (error, "Error parsing variant ‘%s’: ", value_str);
+          return NULL;
```
cgwalters: `glnx_prefix_error_null`

src/ostree/ot-builtin-summary.c
```diff
@@ -40,6 +40,63 @@ static GOptionEntry options[] = {
...
+        }
+
+      g_variant_builder_add (builder, "{sv}", key, value);
+    }
+
+  /* Update the argument list. */
+  for (; i > 1; i--)
+    (*argv)[i] = NULL;
+
+  return g_variant_builder_end (builder);
```
cgwalters: Generally we `ref_sink()` variants; with autocleanups it's less useful to have things floating, and can be actually a danger :skull_and_crossbones:  if one was using an autocleanup and happened to later pass a floating variant to a function which sinks it.

src/ostree/ot-builtin-summary.c
```diff
@@ -40,6 +40,63 @@ static GOptionEntry options[] = {
...
+                             GError  **error)
+{
+  g_autoptr(GVariantBuilder) builder = NULL;
+  int i;
+
+  builder = g_variant_builder_new (G_VARIANT_TYPE_VARDICT);
+
+  for (i = 1; i < *argc; i++)
+    {
+      const gchar *equals = strchr ((*argv)[i], '=');
```
cgwalters: How about `char **args = *argv` and then substitute `(*argv)` → `args` below?

src/ostree/ot-builtin-summary.c
```diff
@@ -40,6 +40,63 @@ static GOptionEntry options[] = {
   { NULL }
 };
 
+/* Take arguments of the form [key=value] or [key, value] out of @argv and put
+ * them into an a{sv} variant. The value arguments must be parsable using
+ * g_variant_parse(). */
```
cgwalters: Hmm.  A bit ad-hoc.  If we're taking `g_variant_parse()`, is there a reason not to just require the user to pass the whole `a{sv}` that way?  Presumably this will be used in scripts where that'll be...well, scripted and not impact the interactive experience.

We have some prior art here for `ostree commit --add-metadata-string`; do we expect non-string keys here?

man/ostree-summary.xml
```diff
@@ -73,7 +73,14 @@ Boston, MA 02111-1307, USA.
...
                 <listitem><para>
-		  Update the summary file.
+                  Update the summary file.
+                </para><para>
+                  Any additional arguments to the command
+                  are treated as additional key–value pairs to be added to the
+                  summary file as additional metadata. They must be in the format
+                  <command><replaceable>KEY</replaceable>=<replaceable>VALUE</replaceable></command>
+                  or as two separate arguments. The values must be in GVariant
+                  text format.
```
cgwalters: Suggest prefixing custom values with something like `exampleos.` to avoid clashes with the `ostree.` prefix?

src/ostree/ot-builtin-summary.c
pwithnall: > is there a reason not to just require the user to pass the whole `a{sv}` that way?

I imagine that would get a bit faffy with keys which are optional (for example, the caller always wants to pass in `key1` and `key2`, but `key3` and `key4` are conditional on different things).

> We have some prior art here for `ostree commit --add-metadata-string`; do we expect non-string keys here?

Yeah, I think integers (timestamps) will be fairly common. `ostree.summary.expiry`, for example.

src/libostree/ostree-repo-pull.c
pwithnall: One of the later commits overwrites this anyway.

src/ostree/ot-builtin-summary.c
pwithnall: For `i`? The declaration is deliberately in a higher scope, since there’s a second loop which counts down afterwards.

man/ostree-summary.xml
pwithnall: Oh yes, definitely.

src/ostree/ot-builtin-summary.c
```diff
@@ -55,10 +109,16 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
 
   if (opt_update)
     {
+      g_autoptr(GVariant) additional_metadata = NULL;
+
       if (!ostree_ensure_repo_writable (repo, error))
         goto out;
 
-      if (!ostree_repo_regenerate_summary (repo, NULL, cancellable, error))
+      additional_metadata = argv_to_additional_metadata (&argc, &argv, error);
```
jlebon: A bit UX bikeshed, though any reason to use additional args for this? Seems random vs. a more explicit e.g. `--add-metadata` (repeatable) option.

pwithnall: A legitimate bikeshed. I went with positional args because they need less typing to use, but would be just as happy with a named arg if you think the positional ones are best saved for something else, or are inconsistent with other commands, or something. Shall I change it?

jlebon: Yeah, I'd vote for making it into long & short opts, if at least to be consistent with `ostree commit`'s `--add-metadata-string`. Compare to e.g. `ostree ls`, where the user more intuitively knows what the extra args do.

/bikeshed :)

pwithnall: Done.

cgwalters:
@rh-atomic-bot r+  	4ee6f27

PR #964, opened 2017-06-26
lib/deltas: Some style porting

cgwalters:
Just a few functions to keep up momentum.


jlebon:
@rh-atomic-bot r+ 026639e

Issue #966, opened 2017-06-27
ostree_remote_get_name is breaking g-i bindings


clumens:
This new function is listed as part of an experimental API, so it doesn't get compiled into the library unless you pass a configure option which is disabled by default.  However, it is still ending up in the .gir file somehow so attempting to build anything that uses g-i bindings results in an error, like so:

<command line>: can't load .so/.DLL for: /root/.cabal/lib/x86_64-linux-ghc-8.0.2/libHSgi-ostree-1.0.3-BeC4yfCrh7f4ZRVzVZN1er-ghc8.0.2.so (/root/.cabal/lib/x86_64-linux-ghc-8.0.2/libHSgi-ostree-1.0.3-BeC4yfCrh7f4ZRVzVZN1er-ghc8.0.2.so: undefined symbol: ostree_remote_get_name)

cgwalters:
Thanks for the report, PR in https://github.com/ostreedev/ostree/pull/967

PR #967, opened 2017-06-27
build: Don't scan ostree-remote.h for introspection if !experimental-api

cgwalters:
It's a bit unfortunate that the scanner doesn't error on this, but
anyways

Closes: https://github.com/ostreedev/ostree/issues/966


cgwalters:

jlebon:
Looks like infra flake.
@rh-atomic-bot r+ e04f48f

PR #968, opened 2017-06-27
papr: build and test on c7

jlebon:
Start testing on CentOS 7 as well to cover kernel differences (e.g.
`O_TMPFILE` support).

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -31,18 +31,22 @@ artifacts:
   - test-suite.log
 ---
 
-context: c7-build
+context: c7-primary
 inherit: true
 required: true
 
-container:
-  image: registry.centos.org/centos/centos:7
+packages:
+
+host:
+  distro: centos/7/atomic
 
 env:
     CFLAGS: ''
 
 tests:
-  - ci/build-check.sh
+  - docker run --privileged -v $PWD:$PWD --workdir $PWD
+    registry.centos.org/centos/centos:7 sh -c
+    'yum install -y git && ci/build-check.sh'
 
 ---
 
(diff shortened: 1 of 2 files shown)


jlebon:
Hmm

```
cp: cannot stat '/bin/busybox': No such file or directory
1..3
ERROR: tests/test-switchroot.sh - too few tests run (expected 3, got 0)
ERROR: tests/test-switchroot.sh - exited with status 1
```

Coming from:

```sh
	cp /bin/busybox "$1/bin"
	busybox --list | xargs -n1 -I '{}' ln -s busybox "$1/bin/{}"
```

But I'm confused how that ever worked. Looking at recent tests, this test is skipped because of:

```sh
	unshare -m true || \
	    skip "this test needs to set up mount namespaces, rerun as root"
```

Seems like we should make it conditional on having `/bin/busybox`?

cgwalters:
Yes, but there's an issue before that; `test-switchroot.sh` requires `CAP_SYS_ADMIN`, which `f25-primary` doesn't have, but you do have since you're running the container with `--privileged`.

It looks like we should just skip this test if we don't have busybox, but also try to install it?

Hrm, also this test doesn't quite work installed yet :frowning:

jlebon:
Right, it's only showing up now because `unshare` doesn't fail. EPEL doesn't seem to have it unfortunately. :(

> It looks like we should just skip this test if we don't have busybox, but also try to install it?

Yup, done above! ⬆️

cgwalters:
`unshare` is part of `util-linux`.

cgwalters:
@rh-atomic-bot r+  	61a63cd

jlebon:
Sorry, the "it" in my second sentence was referring to busybox, not unshare (in response to your suggestion to try to install it). I didn't look too hard if we could get it easily from somewhere else.

PR #969, opened 2017-06-27
ci: Actually run installed tests again

cgwalters:
We did it *again* 😭


jlebon:
Ahh, that's what went wrong. I was looking at the test outputs in horror. 😱

jlebon:
Looks like they're properly running now!
@rh-atomic-bot r+ d576411

PR #972, opened 2017-06-28
collection ref in commit metadata

krnowak:
This is related to #946. This implements adding collection ref information to commit metadata and verification of it during pulls. Includes tests.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -122,6 +122,7 @@ _installed_or_uninstalled_test_scripts += \
 	tests/test-refs-collections.sh \
 	tests/test-remote-add-collections.sh \
 	tests/test-summary-collections.sh \
+	tests/test-pull-collections.sh \
 	$(NULL)
 endif
 
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -56,6 +56,10 @@ G_BEGIN_DECLS
  * in a summary file. */
 #define OSTREE_COMMIT_TIMESTAMP "ostree.commit.timestamp"
 
+/* Well-known keys for the commit metadata */
+#define OSTREE_REF_BINDING "ostree.ref-binding"
+#define OSTREE_COLLECTION_BINDING "ostree.collection-binding"
+
 typedef enum {
   OSTREE_REPO_TEST_ERROR_PRE_COMMIT = (1 << 0)
 } OstreeRepoTestErrorFlags;
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1369,6 +1369,141 @@ commitstate_is_partial (OtPullData   *pull_data,
     || (commitstate & OSTREE_REPO_COMMIT_STATE_PARTIAL) > 0;
 }
 
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+/* Reads the collection-id of a given remote from the repo
+ * configuration.
+ */
+static char *
+get_real_remote_repo_collection_id (OstreeRepo  *repo,
+                                    const gchar *remote_name)
+{
+  g_autofree gchar *remote_collection_id = NULL;
+  if (!ostree_repo_get_remote_option (repo, remote_name, "collection-id", NULL,
+                                      &remote_collection_id, NULL) ||
+      (remote_collection_id == NULL) ||
+      (remote_collection_id[0] == '\0'))
+    return NULL;
+
+  return g_steal_pointer (&remote_collection_id);
+}
+
+/* Reads the collection-id of the remote repo. Where it will be read
+ * from depends on whether we pull from the "local" remote repo (the
+ * "file://" URL) or "remote" remote repo (likely the "http(s)://"
+ * URL).
+ */
+static char *
+get_remote_repo_collection_id (OtPullData *pull_data)
+{
+  if (pull_data->remote_repo_local != NULL)
+    {
+      const char *remote_collection_id =
+        ostree_repo_get_collection_id (pull_data->remote_repo_local);
+      if ((remote_collection_id == NULL) ||
+          (remote_collection_id[0] == '\0'))
+        return NULL;
+      return g_strdup (remote_collection_id);
+    }
+
+  return get_real_remote_repo_collection_id (pull_data->repo,
+                                             pull_data->remote_name);
+}
+#endif  /* OSTREE_ENABLE_EXPERIMENTAL_API */
+
+/* Verify the ref and collection bindings.
+ *
+ * The ref binding is verified only if it exists. But if we have the
+ * collection ID specified in the remote configuration then the ref
+ * binding must exist, otherwise the verification will fail. Parts of
+ * the verification can be skipped by passing NULL to requested_ref or
+ * checksum parameters.
+ *
+ * The collection binding is verified only when we have collection ID
+ * specified in the remote configuration. If it is specified, then the
+ * binding must exist and must be equal to the remote repository
+ * collection ID.
+ */
+static gboolean
+verify_bindings (OtPullData                 *pull_data,
+                 GVariant                   *commit,
+                 const OstreeCollectionRef  *requested_ref,
+                 GError                    **error)
+{
+  g_autofree char *remote_collection_id = NULL;
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+  remote_collection_id = get_remote_repo_collection_id (pull_data);
+#endif  /* OSTREE_ENABLE_EXPERIMENTAL_API */
+  g_autoptr(GVariant) metadata = g_variant_get_child_value (commit, 0);
+  g_autofree const char **refs = NULL;
+  if (!g_variant_lookup (metadata,
+                         OSTREE_REF_BINDING,
+                         "^a&s",
+                         &refs))
+    {
+      /* Early return here - if the remote collection ID is NULL, then
+       * we certainly will verify the collection binding in the
+       * commit.
+       */
+      if (remote_collection_id == NULL)
+        return TRUE;
+
+      return glnx_throw (error,
+                         "expected commit metadata to have ref "
+                         "binding information, found none");
+    }
+
+  if (requested_ref != NULL)
+    {
+      if (!g_strv_contains ((const char *const *) refs, requested_ref->ref_name))
+        {
+          g_autoptr(GString) refs_dump = g_string_new (NULL);
+          const char *refs_str;
+
+          if (refs != NULL && (*refs) != NULL)
+            {
+              for (const char **iter = refs; *iter != NULL; ++iter)
+                {
+                  const char *ref = *iter;
+
+                  if (refs_dump->len > 0)
+                    g_string_append (refs_dump, ", ");
+                  g_string_append_printf (refs_dump, "‘%s’", ref);
+                }
+
+              refs_str = refs_dump->str;
+            }
+          else
+            {
+              refs_str = "no refs";
+            }
+
+          return glnx_throw (error, "commit has no requested ref ‘%s’ "
+                             "in ref binding metadata (%s)",
+                             requested_ref->ref_name, refs_str);
+        }
+    }
+
+  if (remote_collection_id != NULL)
+    {
+      const char *collection_id;
+      if (!g_variant_lookup (metadata,
+                             OSTREE_COLLECTION_BINDING,
+                             "&s",
+                             &collection_id))
+        return glnx_throw (error,
+                           "expected commit metadata to have collection ID "
+                           "binding information, found none");
+      if (!g_str_equal (collection_id, remote_collection_id))
+        return glnx_throw (error,
+                           "commit has collection ID ‘%s’ in collection binding "
+                           "metadata, while the remote it came from has "
+                           "collection ID ‘%s’",
+                           collection_id, remote_collection_id);
+    }
+
+  return TRUE;
+}
+
 static gboolean
 scan_commit_object (OtPullData                 *pull_data,
                     const char                 *checksum,
@@ -1417,6 +1552,15 @@ scan_commit_object (OtPullData                 *pull_data,
 
   if (!ostree_repo_load_commit (pull_data->repo, checksum, &commit, &commitstate, error))
     goto out;
+
+  /* If ref is non-NULL then the commit we fetched was requested through the
+   * branch, otherwise we requested a commit checksum without specifying a branch.
+   */
+  if (!verify_bindings (pull_data, commit, ref, error))
+    {
+      g_prefix_error (error, "Commit %s: ", checksum);
+      goto out;
+    }
 
   /* If we found a legacy transaction flag, assume all commits are partial */
   is_partial = commitstate_is_partial (pull_data, commitstate);
@@ -5124,9 +5268,8 @@ check_remote_matches_collection_id (OstreeRepo  *repo,
 {
   g_autofree gchar *remote_collection_id = NULL;
 
-  if (!ostree_repo_get_remote_option (repo, remote_name, "collection-id", NULL,
-                                      &remote_collection_id, NULL) ||
-      remote_collection_id == NULL)
+  remote_collection_id = get_real_remote_repo_collection_id (repo, remote_name);
+  if (remote_collection_id == NULL)
     return FALSE;
 
   return g_str_equal (remote_collection_id, collection_id);
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -155,6 +155,8 @@ typedef struct {
    * whether to fetch the primary object after fetching its
    * detached metadata (no need if it's already stored). */
   gboolean     object_is_stored;
+
+  OstreeCollectionRef *requested_ref;  /* (nullable) */
 } FetchObjectData;
 
 typedef struct {
@@ -172,31 +174,35 @@ typedef struct {
   char *path;
   OstreeObjectType objtype;
   guint recursion_depth; /* NB: not used anymore, though might be nice to print */
+  OstreeCollectionRef *requested_ref;  /* (nullable) */
 } ScanObjectQueueData;
 
 static void start_fetch (OtPullData *pull_data, FetchObjectData *fetch);
 static void start_fetch_deltapart (OtPullData *pull_data,
                                    FetchStaticDeltaData *fetch);
 static gboolean fetcher_queue_is_full (OtPullData *pull_data);
-static void queue_scan_one_metadata_object (OtPullData         *pull_data,
-                                            const char         *csum,
-                                            OstreeObjectType    objtype,
-                                            const char         *path,
-                                            guint               recursion_depth);
+static void queue_scan_one_metadata_object (OtPullData                *pull_data,
+                                            const char                *csum,
+                                            OstreeObjectType           objtype,
+                                            const char                *path,
+                                            guint                      recursion_depth,
+                                            const OstreeCollectionRef *ref);
 
-static void queue_scan_one_metadata_object_c (OtPullData         *pull_data,
-                                              const guchar       *csum,
-                                              OstreeObjectType    objtype,
-                                              const char         *path,
-                                              guint               recursion_depth);
+static void queue_scan_one_metadata_object_c (OtPullData                *pull_data,
+                                              const guchar              *csum,
+                                              OstreeObjectType           objtype,
+                                              const char                *path,
+                                              guint                      recursion_depth,
+                                              const OstreeCollectionRef *ref);
 
-static gboolean scan_one_metadata_object_c (OtPullData         *pull_data,
-                                            const guchar       *csum,
-                                            OstreeObjectType    objtype,
-                                            const char         *path,
-                                            guint               recursion_depth,
-                                            GCancellable       *cancellable,
-                                            GError            **error);
+static gboolean scan_one_metadata_object_c (OtPullData                 *pull_data,
+                                            const guchar               *csum,
+                                            OstreeObjectType            objtype,
+                                            const char                 *path,
+                                            guint                       recursion_depth,
+                                            const OstreeCollectionRef  *ref,
+                                            GCancellable               *cancellable,
+                                            GError                    **error);
 
 static gboolean
 update_progress (gpointer user_data)
@@ -422,11 +428,14 @@ idle_worker (gpointer user_data)
                               scan_data->objtype,
                               scan_data->path,
                               scan_data->recursion_depth,
+                              scan_data->requested_ref,
                               pull_data->cancellable,
                               &error);
   check_outstanding_requests_handle_error (pull_data, &error);
 
   g_free (scan_data->path);
+  if (scan_data->requested_ref != NULL)
+    ostree_collection_ref_free (scan_data->requested_ref);
   g_free (scan_data);
   return G_SOURCE_CONTINUE;
 }
@@ -507,12 +516,13 @@ write_commitpartial_for (OtPullData *pull_data,
 }
 
 static void
-enqueue_one_object_request (OtPullData        *pull_data,
-                            const char        *checksum,
-                            OstreeObjectType   objtype,
-                            const char        *path,
-                            gboolean           is_detached_meta,
-                            gboolean           object_is_stored);
+enqueue_one_object_request (OtPullData                *pull_data,
+                            const char                *checksum,
+                            OstreeObjectType           objtype,
+                            const char                *path,
+                            gboolean                   is_detached_meta,
+                            gboolean                   object_is_stored,
+                            const OstreeCollectionRef *ref);
 
 static gboolean
 matches_pull_dir (const char *current_file,
@@ -749,7 +759,7 @@ scan_dirtree_object (OtPullData   *pull_data,
 
       /* Not available locally, queue a HTTP request */
       g_hash_table_add (pull_data->requested_content, file_checksum);
-      enqueue_one_object_request (pull_data, file_checksum, OSTREE_OBJECT_TYPE_FILE, path, FALSE, FALSE);
+      enqueue_one_object_request (pull_data, file_checksum, OSTREE_OBJECT_TYPE_FILE, path, FALSE, FALSE, NULL);
       file_checksum = NULL;  /* Transfer ownership */
     }
 
@@ -779,9 +789,9 @@ scan_dirtree_object (OtPullData   *pull_data,
 
       g_autofree char *subpath = g_strconcat (path, dirname, "/", NULL);
       queue_scan_one_metadata_object_c (pull_data, tree_csum_bytes,
-                                        OSTREE_OBJECT_TYPE_DIR_TREE, subpath, recursion_depth + 1);
+                                        OSTREE_OBJECT_TYPE_DIR_TREE, subpath, recursion_depth + 1, NULL);
       queue_scan_one_metadata_object_c (pull_data, meta_csum_bytes,
-                                        OSTREE_OBJECT_TYPE_DIR_META, subpath, recursion_depth + 1);
+                                        OSTREE_OBJECT_TYPE_DIR_META, subpath, recursion_depth + 1, NULL);
     }
 
   return TRUE;
@@ -883,6 +893,8 @@ fetch_object_data_free (FetchObjectData *fetch_data)
 {
   g_variant_unref (fetch_data->object);
   g_free (fetch_data->path);
+  if (fetch_data->requested_ref)
+    ostree_collection_ref_free (fetch_data->requested_ref);
   g_free (fetch_data);
 }
 
@@ -1062,7 +1074,7 @@ on_metadata_written (GObject           *object,
       goto out;
     }
 
-  queue_scan_one_metadata_object_c (pull_data, csum, objtype, fetch_data->path, 0);
+  queue_scan_one_metadata_object_c (pull_data, csum, objtype, fetch_data->path, 0, fetch_data->requested_ref);
 
  out:
   pull_data->n_outstanding_metadata_write_requests--;
@@ -1108,9 +1120,9 @@ meta_fetch_on_complete (GObject           *object,
               g_hash_table_add (pull_data->fetched_detached_metadata, g_strdup (checksum));
 
               if (!fetch_data->object_is_stored)
-                enqueue_one_object_request (pull_data, checksum, objtype, fetch_data->path, FALSE, FALSE);
+                enqueue_one_object_request (pull_data, checksum, objtype, fetch_data->path, FALSE, FALSE, fetch_data->requested_ref);
               else
-                queue_scan_one_metadata_object (pull_data, checksum, objtype, fetch_data->path, 0);
+                queue_scan_one_metadata_object (pull_data, checksum, objtype, fetch_data->path, 0, fetch_data->requested_ref);
             }
 
           /* When traversing parents, do not fail on a missing commit.
@@ -1125,7 +1137,7 @@ meta_fetch_on_complete (GObject           *object,
               if (pull_data->has_tombstone_commits)
                 {
                   enqueue_one_object_request (pull_data, checksum, OSTREE_OBJECT_TYPE_TOMBSTONE_COMMIT,
-                                              fetch_data->path, FALSE, FALSE);
+                                              fetch_data->path, FALSE, FALSE, NULL);
                 }
             }
         }
@@ -1162,9 +1174,9 @@ meta_fetch_on_complete (GObject           *object,
       g_hash_table_add (pull_data->fetched_detached_metadata, g_strdup (checksum));
 
       if (!fetch_data->object_is_stored)
-        enqueue_one_object_request (pull_data, checksum, objtype, fetch_data->path, FALSE, FALSE);
+        enqueue_one_object_request (pull_data, checksum, objtype, fetch_data->path, FALSE, FALSE, fetch_data->requested_ref);
       else
-        queue_scan_one_metadata_object (pull_data, checksum, objtype, fetch_data->path, 0);
+        queue_scan_one_metadata_object (pull_data, checksum, objtype, fetch_data->path, 0, fetch_data->requested_ref);
     }
   else
     {
@@ -1358,11 +1370,12 @@ commitstate_is_partial (OtPullData   *pull_data,
 }
 
 static gboolean
-scan_commit_object (OtPullData         *pull_data,
-                    const char         *checksum,
-                    guint               recursion_depth,
-                    GCancellable       *cancellable,
-                    GError            **error)
+scan_commit_object (OtPullData                 *pull_data,
+                    const char                 *checksum,
+                    guint                       recursion_depth,
+                    const OstreeCollectionRef  *ref,
+                    GCancellable               *cancellable,
+                    GError                    **error)
 {
   gboolean ret = FALSE;
   /* If we found a legacy transaction flag, assume we have to scan.
@@ -1421,7 +1434,7 @@ scan_commit_object (OtPullData         *pull_data,
     {
       queue_scan_one_metadata_object_c (pull_data, parent_csum_bytes,
                                         OSTREE_OBJECT_TYPE_COMMIT, NULL,
-                                        recursion_depth + 1);
+                                        recursion_depth + 1, NULL);
     }
   else if (parent_csum_bytes != NULL && depth > 0)
     {
@@ -1448,7 +1461,8 @@ scan_commit_object (OtPullData         *pull_data,
           queue_scan_one_metadata_object_c (pull_data, parent_csum_bytes,
                                             OSTREE_OBJECT_TYPE_COMMIT,
                                             NULL,
-                                            recursion_depth + 1);
+                                            recursion_depth + 1,
+                                            NULL);
         }
     }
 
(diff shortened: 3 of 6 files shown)


src/ostree/ot-builtin-commit.c
```diff
@@ -328,6 +334,55 @@ parse_keyvalue_strings (char             **strings,
...
+    }
+
+  g_assert_nonnull (opt_branch);
+
+  const char *collection_id = opt_collection_id;
+  const char *repo_collection_id = ostree_repo_get_collection_id (repo);
+  if (collection_id == NULL)
+    collection_id = repo_collection_id;
+  else if (g_strcmp0 (collection_id, repo_collection_id) != 0)
+    g_printerr ("Creating commit with different collection ID (%s) than the repo's one (%s)\n", collection_id, repo_collection_id);
```
cgwalters: Why not make this fatal?   (Well hm, I guess other than for writing test cases; we could have `env OSTREE_COMMIT_ALLOW_DIFFERENT_COLLECTION ostree commit ...`)

src/ostree/ot-builtin-commit.c
```diff
@@ -328,6 +334,55 @@ parse_keyvalue_strings (char             **strings,
...
+  g_autoptr(GVariantBuilder) builder = NULL;
+  builder = ot_util_variant_builder_from_variant (metadata,
+                                                  G_VARIANT_TYPE_VARDICT);
+  g_variant_builder_add (builder, "{s@v}", "ostree.commit.collection-ref",
+                         g_variant_new_variant (g_variant_new ("(ss)",
+                                                               collection_id,
+                                                               opt_branch)));
+
+  *out_metadata = g_variant_builder_end (builder);
+  g_variant_ref_sink (*out_metadata);
```
cgwalters: I tend to do `*out_metadata = g_variant_ref_sink (g_variant_builder_end (builder)));`

src/ostree/ot-builtin-commit.c
```diff
@@ -328,6 +334,55 @@ parse_keyvalue_strings (char             **strings,
...
+  if (collection_id == NULL || opt_key_ids == NULL)
+    {
+      *out_metadata = (metadata != NULL) ? g_variant_ref (metadata) : NULL;
+      return TRUE;
+    }
+
+  g_autoptr(GVariantBuilder) builder = NULL;
+  builder = ot_util_variant_builder_from_variant (metadata,
+                                                  G_VARIANT_TYPE_VARDICT);
+  g_variant_builder_add (builder, "{s@v}", "ostree.commit.collection-ref",
```
cgwalters: How about `ostree.ref-collection-binding`?

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
 static gboolean
+ensure_collection_ref_in_summary (GVariant   *summary,
+                                  const char *collection_id,
+                                  const char *ref)
+{
+  g_assert_nonnull (collection_id);
+  g_assert_nonnull (ref);
+
+  g_autoptr(GVariant) additional_metadata = NULL;
+  additional_metadata = g_variant_get_child_value (summary, 1);
```
cgwalters: New ostree code style is to init variables with declaration.  (Applies to other places below)

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+    return TRUE;
+
+  g_autoptr(GVariant) metadata = NULL;
+  metadata = g_variant_get_child_value (commit, 0);
+
+  gboolean unpacked;
+  const char *collection_id;
+  const char *ref;
+  unpacked = g_variant_lookup (metadata,
+                               "ostree.commit.collection-ref",
```
cgwalters: @pwithnall earlier made `#define`s for these types of things in `ostree-repo-private.h`.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+
+  if (!unpacked)
+    return glnx_throw (error,
+                       "expected commit metadata to have collection "
+                       "ref information, found none");
+
+  if ((collection_id == NULL) ||
+      (collection_id[0] == '\0') ||
+      (ref == NULL) ||
+      (ref[0] == '\0'))
```
cgwalters: `g_variant_lookup()` can't set them to `NULL` right?  As far as checking for the empty string...how about having `get_remote_collection_id()` check that?  Then we can just rely on the `g_str_equal()`.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+    return glnx_throw (error,
+                       "bad collection ref information in commit metadata");
+
+  if (!g_str_equal (collection_id, remote_collection_id))
+    return glnx_throw (error,
+                       "commit has collection id %s in metadata, "
+                       "while the remote it came from has collection id %s",
```
cgwalters: I'd add single quotes like `'%s'` around these just in case they are empty.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+}
+
+static char *
+get_remote_collection_id (OstreeRepo  *repo,
+                          const gchar *remote_name)
+{
+  g_autofree gchar *remote_collection_id = NULL;
+
+  /* TODO: This happens for local pulls, I guess we should assume no
+   * collection id for this case?
```
cgwalters: Can't we parse it out of the remote config?

Makefile-tests.am
```diff
@@ -121,6 +121,7 @@ _installed_or_uninstalled_test_scripts += \
 	tests/test-refs-collections.sh \
 	tests/test-remote-add-collections.sh \
 	tests/test-summary-collections.sh \
+	tests/test-pull-collections.sh \
```
cgwalters: Can you squash tests in with the commit that introduced the functionality?  Not sure why one would want them separate.

src/ostree/ot-builtin-commit.c
krnowak: I was rather thinking about removing the flag and adding the `OSTREE_TEST_COMMIT_OVERRIDE_COLLECTION_ID` env var for this purpose.

src/ostree/ot-builtin-commit.c
krnowak: Sure.

src/libostree/ostree-repo-pull.c
pwithnall: Better yet, `‘Unicode single quotes’`. 😁

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
 }
 
 static gboolean
+ensure_collection_ref_in_summary (GVariant   *summary,
+                                  const char *collection_id,
+                                  const char *ref)
```
pwithnall: It would be a bit more consistent to take in an `OstreeCollectionRef` here, rather than separate `char*`s.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+  gboolean check_main_collection;
+  check_main_collection = (g_strcmp0 (main_collection_id, collection_id) == 0);
+
+  if (check_main_collection)
+    {
+      g_autoptr(GVariant) refs = NULL;
+      gsize i, n;
+
+      refs = g_variant_get_child_value (summary, 0);
+      for (i = 0, n = g_variant_n_children (refs); i < n; i++)
```
pwithnall: This can be optimised using a binary search, since the refs are guaranteed to be in alphabetical order. Use `ot_variant_bsearch_str()`.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+    }
+
+  g_autoptr(GVariant) collection_map = NULL;
+  collection_map = g_variant_lookup_value (additional_metadata,
+                                           OSTREE_SUMMARY_COLLECTION_MAP,
+                                           G_VARIANT_TYPE ("a{sa(s(taya{sv}))}"));
+  if (collection_map != NULL)
+    {
+      GVariantIter collection_map_iter;
+      char *unpack_summary_collection_id;
```
pwithnall: `const`

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+      GVariant *unpack_refs;
```
pwithnall: `g_autoptr(GVariant)`

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+
+      g_variant_iter_init (&collection_map_iter, collection_map);
+
+      while (g_variant_iter_next (&collection_map_iter,
```
pwithnall: Use `g_variant_iter_loop()` to get fancy memory handling of `unpack_summary_collection_id` and `unpack_refs`.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+                                  "{&s@a(s(taya{sv}))}",
+                                  &unpack_summary_collection_id,
+                                  &unpack_refs))
+        {
+          gsize i, n;
+          g_autofree char *summary_collection_id = unpack_summary_collection_id;
+          g_autoptr(GVariant) refs = unpack_refs;
```
pwithnall: No need for these if you apply the three comments above.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+
+          if (g_strcmp0 (summary_collection_id, collection_id) != 0)
+            continue;
+
+          for (i = 0, n = g_variant_n_children (refs); i < n; i++)
```
pwithnall: Similarly, it should be possible to use `ot_variant_bsearch_str()` here too.

src/libostree/ostree-repo-pull.c
pwithnall: Isn’t that exactly what the `ostree_repo_get_remote_option()` call below does, or am I misunderstanding your comment?

src/libostree/ostree-repo-pull.c
pwithnall: ⇒ Add a new one for `ostree.commit.collection-ref`.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+  if (!ostree_repo_get_remote_option (repo, remote_name, "collection-id", NULL,
+                                      &remote_collection_id, NULL) ||
+      remote_collection_id == NULL)
+    return NULL;
+
+  return g_steal_pointer (&remote_collection_id);
+}
+
+static gboolean
+verify_collection_ref (OtPullData  *pull_data,
```
pwithnall: It would be useful to add a short doc comment for this function explaining what’s being verified and when verification is skipped.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+  gboolean unpacked;
+  const char *collection_id;
+  const char *ref;
+  unpacked = g_variant_lookup (metadata,
+                               "ostree.commit.collection-ref",
+                               "(&s&s)",
+                               &collection_id,
+                               &ref);
+
+  if (!unpacked)
```
pwithnall: You could eliminate the `unpacked` variable and just move the `g_variant_lookup()` call into the `if` condition to make things a little clearer.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+
+static gboolean
+verify_collection_ref (OtPullData  *pull_data,
+                       GVariant    *commit,
+                       GError     **error)
+{
+  g_autofree char *remote_collection_id;
+  remote_collection_id = get_remote_collection_id (pull_data->repo,
+                                                   pull_data->remote_name);
+  if (remote_collection_id == NULL)
```
pwithnall: I’m not sure we want to bail at this point. I think we should *always* verify the `ostree.commit.collection-ref` if it exists, and additionally fail verification if it doesn’t exist and `remote_collection_id != NULL`.

If `(remote_collection_id == NULL)`, the check for an `ostree.commit.collection-ref` which exists would only look at its ref name, not its collection ID.

Makefile-tests.am
krnowak: Want me to squash all three commits into one? The test is using features added in both previous commits.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+                       "while the remote it came from has collection id %s",
+                       collection_id, remote_collection_id);
+
+  g_autoptr(OstreeCollectionRef) collection_ref;
+  collection_ref = ostree_collection_ref_new (collection_id, ref);
+  if (!g_hash_table_contains (pull_data->requested_refs_to_fetch, collection_ref))
+    return glnx_throw (error,
+                       "commit has collection id %s and ref %s, "
+                       "which was not requested",
+                       collection_id, ref);
```
pwithnall: Single quotes here too, for consistency.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+
+  /* TODO: not sure about this step, if the summary is meant to be unsigned…
```
pwithnall: I’m not sure about this step either, but not because the `summary` is unsigned. I’m unsure what value it brings, given that the commit checksums we are pulling have been looked up from the `summary` file in the first place (at the start of the pull operation), so this is just verifying something we already know.

src/libostree/ostree-repo-pull.c
krnowak: Thanks for providing them at the same time. :)

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+
+  if (!g_str_equal (collection_id, remote_collection_id))
+    return glnx_throw (error,
+                       "commit has collection id %s in metadata, "
+                       "while the remote it came from has collection id %s",
+                       collection_id, remote_collection_id);
+
+  g_autoptr(OstreeCollectionRef) collection_ref;
+  collection_ref = ostree_collection_ref_new (collection_id, ref);
+  if (!g_hash_table_contains (pull_data->requested_refs_to_fetch, collection_ref))
```
pwithnall: I think this should also check that the checksum from `requested_refs_to_fetch` matches the `commit`’s checksum. Otherwise we could have the situation where I request to pull `refA` and `refB`, and a man in the middle modifies the `summary` file to return `checksum1` for both of them, which points to a commit on `refA` (but *no* commit for `refB`). This check will succeed for the assumed mappings `refA` → `checksum1` and `refB` → `checksum1`. It should fail for `refB` → `checksum1`.

Actually, adding a check on the value of `requested_refs_to_fetch` wouldn’t fix that case. But it would fix the case where the `summary` file contained `refA` → `checksum1` and `refB` → `checksum2`, but `checksum1` had its `ostree.commit.collection-ref` set to `refB` and `checksum2` had it set to `refA`.

src/libostree/ostree-repo-pull.c
```diff
@@ -1322,6 +1324,165 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+    return glnx_throw (error,
+                       "bad collection ref information in commit metadata");
+
+  if (!g_str_equal (collection_id, remote_collection_id))
+    return glnx_throw (error,
+                       "commit has collection id %s in metadata, "
+                       "while the remote it came from has collection id %s",
+                       collection_id, remote_collection_id);
+
+  g_autoptr(OstreeCollectionRef) collection_ref;
```
pwithnall: This could be defined on the stack.

src/libostree/ostree-repo-pull.c
```diff
@@ -1369,6 +1530,15 @@ scan_commit_object (OtPullData         *pull_data,
   if (!ostree_repo_load_commit (pull_data->repo, checksum, &commit, &commitstate, error))
     goto out;
 
+  if (pull_data->summary)
```
pwithnall: This `pull_data->summary` condition is unnecessary if we drop the `ensure_collection_ref_in_summary` check from above.

src/libostree/ostree-repo-pull.c
```diff
@@ -3619,6 +3789,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
 
   if (pull_data->legacy_transaction_resuming)
     g_debug ("resuming legacy transaction");
+  pull_data->requested_refs_to_fetch = requested_refs_to_fetch;
```
pwithnall: Should this have a ref added?

src/ostree/ot-builtin-commit.c
```diff
@@ -78,6 +81,9 @@ static GOptionEntry options[] = {
   { "body", 'm', 0, G_OPTION_ARG_STRING, &opt_body, "Full description", "BODY" },
   { "body-file", 'F', 0, G_OPTION_ARG_FILENAME, &opt_body_file, "Commit message from FILE path", "FILE" },
   { "editor", 'e', 0, G_OPTION_ARG_NONE, &opt_editor, "Use an editor to write the commit message", NULL },
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+  { "collection-id", 'c', 0, G_OPTION_ARG_STRING, &opt_collection_id, "Collection ID", "ID" },
```
pwithnall: I think the other commands have set the placeholder (last element) to `COLLECTION-ID` rather than `ID`.

src/ostree/ot-builtin-commit.c
```diff
@@ -328,6 +334,55 @@ parse_keyvalue_strings (char             **strings,
...
+                    GCancellable  *cancellable,
+                    GError       **error)
+{
+  if (opt_orphan)
+    {
+      if (opt_collection_id)
+        {
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                       "Can't use --collection-id together with --orphan");
+          return FALSE;
```
pwithnall: Use `glnx_throw()`.

src/ostree/ot-builtin-commit.c
pwithnall: If you make this fatal, what’s the point in having a `--collection-id` argument? In the case it’s not set, it’s set to `repo_collection_id`. In the case it’s set, it has to be set to `repo_collection_id` or there’s a fatal error.

pwithnall: Do we need to test the case where `collection_id != repo_collection_id` if it’s going to be disallowed?

src/libostree/ostree-repo-pull.c
krnowak: I guess I will need to remove the "continue" keyword, otherwise we will get leaks if I move `g_autoptr` outside the loop.

src/libostree/ostree-repo-pull.c
krnowak: I already did that, but I haven't yet pushed.

src/libostree/ostree-repo-pull.c
krnowak: Here I meant the case when we do `ostree pull-local /foo/repo ref`. In this case the `remote_name` variable is set to NULL, so we need to get the information from `pull_data->remote_repo_local`.

src/libostree/ostree-repo-pull.c
pwithnall: Yeah, you could turn it into an `if` block. That should be simpler once you’re doing the bsearch.

src/libostree/ostree-repo-pull.c
krnowak: I was thinking about it, but `pull_data` will not outlive the `requested_refs_to_fetch` variable. I can add it, for good measure.

src/libostree/ostree-repo-pull.c
krnowak: Got:
```
src/libostree/ostree-repo-pull.c:1466:70: warning: initialization discards 'const' qualifier from pointer target type [-Wdiscarded-qualifiers]
   const OstreeCollectionRef collection_ref = { remote_collection_id, ref };
                                                                      ^~~
```

So I decided to keep it as is to avoid casting the constness away.

pwithnall: I’ve been using
```
const OstreeCollectionRef collection_ref = { remote_collection_id, (char *) ref };
```
and it feels fairly satisfying to me. The `const`ness is retained through the use of `const` on the `OstreeCollectionRef`.

krnowak: Alright, update coming up then.

cgwalters:
Some further discussion on this in https://github.com/ostreedev/ostree/issues/983

src/libostree/ostree-repo-private.h
```diff
@@ -55,6 +55,9 @@ G_BEGIN_DECLS
  * in a summary file. */
 #define OSTREE_COMMIT_TIMESTAMP "ostree.commit.timestamp"
 
+/* Well-known keys for the commit metadata */
+#define OSTREE_REF_COLLECTION_BINDING "ostree.ref-collection-binding"
```
pwithnall: As discussed in issue #983, it looks like this needs to be split up into `ostree.ref-binding` (type `as`) and `ostree.collection-binding` (type `s`), with `ostree.collection-binding` only being used if `OSTREE_ENABLE_EXPERIMENTAL_API` is defined.

krnowak:
I took a bit different approach with this one, so excuse me for rebasing the PR on top of current master and squashing all the fixups.

Now, I'm passing the requested ref (the `OstreeCollectionRef` instance) throughout the pull code, so we can use this for verifying if the commit we have requested to fetch is indeed from the given collection-ref. The passed requested ref is also a key in the requested_refs_to_fetch, so we can use it to also check the if the checksum of the commit we just fetched is what we got.

There's a catch though - we only do the verification of the "HEAD" commit - I'm assuming that if this commit passes the verification, then we can also trust all the parent commits we fetch. We also skip the verification of commits for which we explicitly specified the checksum we want to pull. I'm not sure if that's right, though. In general there are four scenarios:
- HEAD commit from the requested ref - we verify it
- parent of the HEAD commit of the requested ref - we don't verify it, but we probably could check if the collection ref stored in the metadata is also in the `requested_refs_to_fetch` hash table.
- commit with no requested ref (basically happens only when we pass the commit checksum in the `refs` option to `ostree_repo_pull_with_options`) - no verification and I'm not sure what to do here. We probably could check if the collection id of the commit matches the collection id of the remote repo.
- parent of the commit with no requested ref - same as above I guess.

On the command line side, I dropped the `--collection-id` flag, and replaced it with `OSTREE_TEST_OVERRIDE_COLLECTION_ID` env var for testing purposes. I also added a string array flag `--bind-ref` for adding more refs to the ref binding metadata. The value of the `--branch` flag is added to the ref binding too. The refs in the metadata are sorted.

I decided against squashing the commits into one for now - I guess it will be easier to review the code commit by commit.

krnowak:
Updated again, this time we try to verify all the commits we fetch as much as possible.

pwithnall:
> Now, I'm passing the requested ref (the `OstreeCollectionRef` instance) throughout the pull code, so we can use this for verifying if the commit we have requested to fetch is indeed from the given collection-ref. The passed requested ref is also a key in the `requested_refs_to_fetch`, so we can use it to also check the if the checksum of the commit we just fetched is what we got.

Sounds like a sensible approach.

> There's a catch though - we only do the verification of the "HEAD" commit - I'm assuming that if this commit passes the verification, then we can also trust all the parent commits we fetch.

I think that’s reasonable, since the head commit includes the checksum of its parent in its signed data, which implies that the person who signed the head commit trusted the parent commit.

> We also skip the verification of commits for which we explicitly specified the checksum we want to pull. I'm not sure if that's right, though. In general there are four scenarios:
> 
>  * HEAD commit from the requested ref - we verify it
>  * parent of the HEAD commit of the requested ref - we don't verify it, but we probably could check if the collection ref stored in the metadata is also in the `requested_refs_to_fetch` hash table.

I don’t think there’s a need to verify any of the parent’s metadata, assuming we trust the checksum and the signature of it in the child commit.

>  * commit with no requested ref (basically happens only when we pass the commit checksum in the `refs` option to `ostree_repo_pull_with_options`) - no verification and I'm not sure what to do here. We probably could check if the collection id of the commit matches the collection id of the remote repo.

This is quite a corner case. Checking the collection ID makes sense. I can’t think of what else we could verify. The whole point of verifying the `ref-binding` is to verify the unsigned ref → checksum map from the `summary` file. If a checksum is provided without a ref, it’s probably come from somewhere other than the `summary` file, so there is no such binding to verify.

>  * parent of the commit with no requested ref - same as above I guess.

Similarly, I don’t think we need to check the parent.

src/libostree/ostree-repo-pull.c
```diff
@@ -1369,6 +1537,44 @@ scan_commit_object (OtPullData         *pull_data,
   if (!ostree_repo_load_commit (pull_data->repo, checksum, &commit, &commitstate, error))
     goto out;
 
+  /* If ref is NULL it means that the commit we fetched is the commit
```
cgwalters: "ref is not NULL" right?

tests/test-pull-collections.sh
```diff
@@ -0,0 +1,247 @@
...
+
+ensure_collection_ref() {
+    local repo=$1
+    local branch=$2
+    local collection_id=$3
+    shift 3
+    local refs=$(do_join "$@")
+
+    do_collection_ref_show "${repo}" "${branch}" >"${branch}-meta"
+    cat "${branch}-meta"
```
pwithnall: You can drop this `cat`, since `assert_file_has_content` will print out the content on failure.

src/libostree/ostree-repo-pull.c
```diff
@@ -1321,12 +1335,166 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+            {
+              refs_str = g_strdup ("no refs");
+            }
+
+          return glnx_throw (error, "commit has no requested ref ‘%s’ "
+                             "in ref binding metadata (%s)",
+                             requested_ref->ref_name, refs_str);
+        }
+
+      if (checksum != NULL)
```
cgwalters: I am actually a bit confused as to what this is checking.  The only paths where the checksum is filled in here is the summary file, or a commit ID override, or `opt_collection_refs_set`.

AFAICS in all of those cases, we'll be processing the provided commit hash, so this test should always succeed.  Is there something security-relevant here I'm missing?

If that's true, the code below gets a lot simpler since we can stop differentating based on `recursion_depth`.

src/libostree/ostree-repo-pull.c
```diff
@@ -151,6 +153,8 @@ typedef struct {
    * whether to fetch the primary object after fetching its
    * detached metadata (no need if it's already stored). */
   gboolean     object_is_stored;
+
+  OstreeCollectionRef *requested_ref;
```
pwithnall: Should this be `/* (nullable) */` for if we’re just pulling a checksum?

src/libostree/ostree-repo-pull.c
```diff
@@ -168,31 +172,35 @@ typedef struct {
   char *path;
   OstreeObjectType objtype;
   guint recursion_depth; /* NB: not used anymore, though might be nice to print */
+  OstreeCollectionRef *requested_ref;
```
pwithnall: Should this be `/* (nullable) */` for if we’re just pulling a checksum?

src/libostree/ostree-repo-pull.c
```diff
@@ -1321,12 +1335,166 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+  if (!g_variant_lookup (metadata,
+                         OSTREE_REF_BINDING,
+                         "^a&s",
+                         &refs))
+    {
+      /* Early return here - if the remote collection ID is NULL, then
+       * we certainly will verify the collection binding in the
+       * commit.
+       */
+      if (remote_collection_id == NULL)
```
pwithnall: > This is making collection setups require commits to have this binding; I'm OK with that if you are, but just noting it.

Yup, that’s deliberate.

> On the flip side...I could imagine a remote config option to require that this value exist, but eh...it's probably not worth it.

We’re treating the `collection-id` configuration option as that option. If we added another option, it would have to be set whenever `collection-id` was set (otherwise P2P support could not be safe, due to the need for P2P to splice together `summary` files from multiple collections). I can’t see much value in setting it when `collection-id` is not set. Hence they are equivalent.

src/libostree/ostree-repo-pull.c
```diff
@@ -1321,12 +1335,166 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+        return TRUE;
+
+      return glnx_throw (error,
+                         "expected commit metadata to have ref "
+                         "binding information, found none");
```
pwithnall: Would be useful to put the commit checksum and collection–ref information in the error message.

src/libostree/ostree-repo-pull.c
```diff
@@ -1321,12 +1335,166 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+              for (const char **iter = refs; *iter != NULL; ++iter)
+                {
+                  const char *ref = *iter;
+
+                  g_string_append_printf (refs_dump, "‘%s’, ", ref);
+                }
+              g_assert_cmpint (refs_dump->len, >=, 2);
+              gsize stray_comma = refs_dump->len - 2;
+              refs_str = g_string_free (g_steal_pointer (&refs_dump), FALSE);
+              refs_str[stray_comma] = '\0';
```
pwithnall: You can even eliminate `gboolean first` by checking `refs_str->len == 0` instead.

src/libostree/ostree-repo-pull.c
```diff
@@ -1321,12 +1335,166 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+            }
+          else
+            {
+              refs_str = g_strdup ("no refs");
+            }
+
+          return glnx_throw (error, "commit has no requested ref ‘%s’ "
+                             "in ref binding metadata (%s)",
+                             requested_ref->ref_name, refs_str);
```
pwithnall: Would be useful to also put the commit checksum in this error message.

pwithnall: (Similarly for the other error messages in this function.)

src/ostree/ot-builtin-commit.c
```diff
@@ -582,6 +651,10 @@ ostree_builtin_commit (int argc, char **argv, GCancellable *cancellable, GError
     {
       gboolean update_summary;
       guint64 timestamp;
+      g_autoptr(GVariant) old_metadata = g_steal_pointer (&metadata);
+
+      fill_bindings (repo, old_metadata, &metadata);
```
pwithnall: I’ve made these changes locally (Krzesimir’s contract with Endless has expired for now).

> The handling here is special enough (and we have enough non-commandline commit writing programs) though that I'm wondering if we should actually lower this logic into `ostree_repo_write_commit_with_opts()` or something and have one of those options be "refs"?

Maybe. I would prefer to do that as a follow-up commit though, rather than introducing more new API here.

src/libostree/ostree-repo-pull.c
pwithnall: (Already the case due to the `g_prefix_error()` calls in the caller.)

src/libostree/ostree-repo-pull.c
cgwalters: Yep, makes sense to me!

src/ostree/ot-builtin-commit.c
cgwalters: That's OK, let's just be sure we keep it in mind, since there are a lot of things that use the API directly and not the cmdline.

src/libostree/ostree-repo-pull.c
krnowak: Ah, no. Maybe it should be rephrased like "if ref is not NULL then the commit we fetched was requested through the branch, otherwise we requested a commit checksum without specifying a branch".

tests/test-pull-collections.sh
krnowak: Ah, that was a debugging leftover.

src/libostree/ostree-repo-pull.c
krnowak: Yeah. It can be NULL when pulling just a checksum or when pulling a non-commit object.

krnowak:
@pwithnall, thanks for picking this up.

pwithnall:
I’m pushing fixups at https://github.com/pwithnall/ostree/tree/collection-ref-in-commit-metadata. Not quite sure how to get those into this PR.

krnowak:
@pwithnall: I'm afraid that you will need to create a new PR and close this one.

pwithnall:
> @pwithnall: I'm afraid that you will need to create a new PR and close this one.

Nope, I’ve now got superpowers and can push fixups to PRs. 🔥

krnowak:
I'm surprised - that basically means you pushed the code to my branch here - https://github.com/kinvolk/ostree/tree/krnowak/collection-ref-in-commit-metadata

pwithnall:
> I'm surprised - that basically means you pushed the code to my branch here - https://github.com/kinvolk/ostree/tree/krnowak/collection-ref-in-commit-metadata

I think there’s a tickbox when you open a PR (which you must have ticked) about allowing fixups from project maintainers.

src/libostree/ostree-repo-pull.c
pwithnall: @krnowak?

I can’t think of a situation where this check would help, now, given that you’re tightly binding the `requested_ref` to each fetch operation.

krnowak:
Ah yes, there is one even on this page, in the labels column.

TIL.

src/libostree/ostree-repo-pull.c
krnowak: I'm just wondering if there is a possible situation that you pull `ref1` and `ref2` from a remote, so you fetch two commits and each commit has both `ref1` and `ref2` in its binding-ref metadata but somehow are swapped (so the `if (!g_strv_contains ((const char *const *) refs, requested_ref->ref_name))` check would pass) , so instead of getting `ref1 -> commit_for_ref1, ref2 -> commit_for_ref2` we get `ref1 -> commit_for_ref2, ref2 -> commit_for_ref1`.

pwithnall: I’m going to leave this code in place until this question is answered.

pwithnall: That makes sense to me. I guess the refs could get swapped if `opt_collection_refs_set` were used incorrectly, so this check is relevant. I’m not sure I can see how it could affect the results of a pull which uses the ref → checksum mapping from the `summary` file, since the refs can’t get swapped that way. I’ll add a comment to the code.

pwithnall: Hang on, no, the refs can’t get swapped if `opt_collection_refs_set` is used incorrectly, because:
 a) `pull_data->requested_refs_to_fetch` is set from `opt_collection_refs`, so the check would not fail; and
 b) if `[ref1, ref2]` is in the refs-binding for both commits, the situation is not invalid and the end result is correct anyway.

pwithnall: @krnowak?

krnowak: Oh, sorry, I thought that you already decided that checking the checksum does make any sense. I pored a bit over the pulling code that fills the `requested_refs_to_fetch` hash table and I didn't find a way to cause the checksum checking to fail by doing some nasty swapping or else.

So yeah, I think you can drop it. I guess I cargo-culted the check from the version where I wasn't forwarding the requested ref, but instead I was doing checks against the `requested_refs_to_fetch` hash map. So you also could drop the hash map from pull data too.

pwithnall: Thanks for taking the time to take another look through the code!

Makefile-tests.am
pwithnall: I think it makes sense to keep the tests as a separate commit, since they test all the functionality added in the previous commits.

pwithnall:
Updated version of the branch rebased and force-pushed. It should address all the review comments. I’ve squashed in all the fixups so far. @cgwalters, how does this look?

cgwalters:
It looks good to me!  If you don't mind let's do a release first though and land this right after that?

src/libostree/ostree-repo-pull.c
```diff
@@ -1357,12 +1369,148 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+}
+#endif  /* OSTREE_ENABLE_EXPERIMENTAL_API */
+
+/* Verify the ref and collection bindings.
+ *
+ * The ref binding is verified only if it exists. But if we have the
+ * collection ID specified in the remote configuration then the ref
+ * binding must exist, otherwise the verification will fail. Parts of
+ * the verification can be skipped by passing NULL to requested_ref or
+ * checksum parameters.
```
krnowak: There is no checksum parameter anymore.

src/libostree/ostree-repo-pull.c
```diff
@@ -1357,12 +1369,148 @@ commitstate_is_partial (OtPullData   *pull_data,
...
+#endif  /* OSTREE_ENABLE_EXPERIMENTAL_API */
+  g_autoptr(GVariant) metadata = g_variant_get_child_value (commit, 0);
+  g_autofree const char **refs = NULL;
+  if (!g_variant_lookup (metadata,
+                         OSTREE_REF_BINDING,
+                         "^a&s",
+                         &refs))
+    {
+      /* Early return here - if the remote collection ID is NULL, then
+       * we certainly will verify the collection binding in the
```
krnowak: Should be "will not".

cgwalters:
@rh-atomic-bot r+ d1e1908

PR #973, opened 2017-06-28
A few misc patches

cgwalters:


jlebon:
@rh-atomic-bot r+ af9cf2d

PR #974, opened 2017-06-28
lib: Add a helper to convert struct stat → GFileInfo

cgwalters:
It's more natural for a few calling places. Prep for patches to go the other
way, which in turn are prep for adding a commit filter v2 that takes `struct
stat`.

`ot_gfile_type_for_mode()` was only used in this function, so inline it here.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2791,18 +2791,14 @@ ostree_repo_load_file (OstreeRepo         *self,
         }
       if (out_file_info)
         {
-          *out_file_info = _ostree_header_gfile_info_new (stbuf.st_mode, stbuf.st_uid, stbuf.st_gid);
-          if (S_ISREG (stbuf.st_mode))
-            {
-              g_file_info_set_size (*out_file_info, stbuf.st_size);
-            }
-          else if (S_ISLNK (stbuf.st_mode))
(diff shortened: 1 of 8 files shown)


src/libostree/ostree-repo.c
```diff
@@ -2791,18 +2791,14 @@ ostree_repo_load_file (OstreeRepo         *self,
...
-          *out_file_info = _ostree_header_gfile_info_new (stbuf.st_mode, stbuf.st_uid, stbuf.st_gid);
-          if (S_ISREG (stbuf.st_mode))
-            {
-              g_file_info_set_size (*out_file_info, stbuf.st_size);
-            }
-          else if (S_ISLNK (stbuf.st_mode))
+          *out_file_info = _ostree_stbuf_to_gfileinfo (&stbuf);
+          if (S_ISLNK (stbuf.st_mode))
             {
               g_file_info_set_size (*out_file_info, 0);
```
jlebon: We don't need this anymore, right? (Or if it's not 0 by default, then should we make this part of `_ostree_stbuf_to_gfileinfo()` ?)

cgwalters: Fixed :arrow_down:

jlebon:
@rh-atomic-bot r+ d7bf212

PR #976, opened 2017-06-29
tests: Fix assert_files_hardlinked

cgwalters:
It was always succeeding because we were trying to stat the inode number, and
failing, and thus getting the empty string for both, which compared as true.

Regression from:
<https://github.com/ostreedev/ostree/commit/74e3581e>

Noticed this while working on
<https://github.com/ostreedev/ostree/pull/974>
and looking at the test results.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -123,9 +123,7 @@ files_are_hardlinked() {
 }
 
 assert_files_hardlinked() {
-    f1=$(stat -c %i $1)
-    f2=$(stat -c %i $2)
-    if ! files_are_hardlinked "$f1" "$f2"; then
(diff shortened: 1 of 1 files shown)


tests/libtest.sh
```diff
@@ -123,9 +123,7 @@ files_are_hardlinked() {
 }
```
jlebon: Would be good to add a check in `files_are_hardlinked()` that `f1` & `f2` are non-empty. Otherwise, we'll succeed even if the files don't exist at all.

cgwalters: Done :arrow_down:

jlebon:
@rh-atomic-bot r+ 7639672

PR #977, opened 2017-06-29
lib/repo: Port bareuser-conversion stat to bare load

cgwalters:
I noticed this is a simple call that's useful to port to the new internal-only
non-allocating API.


jlebon:
@rh-atomic-bot r+ 44a2937

PR #978, opened 2017-06-29
lib/pull: Don't fetch detached metadata twice for local pulls

cgwalters:
Obviously very minor, but I noticed this while working on `pull --reference`. If
we have a local repo, we'll have already done a hardlink and copied the detached
metadata too, so there's no reason to request it again via the fetcher path.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1514,6 +1514,11 @@ scan_one_metadata_object_c (OtPullData         *pull_data,
                                                           objtype, tmp_checksum, !pull_data->is_untrusted,
                                                           cancellable, error))
             return FALSE;
+          /* The import API will fetch both the commit and detached metadata, so
+           * add it to the hash to avoid re-fetching it below.
+           */
+          if (objtype == OSTREE_OBJECT_TYPE_COMMIT)
+            g_hash_table_add (pull_data->fetched_detached_metadata, g_strdup (tmp_checksum));
         }
       is_stored = TRUE;
       is_requested = TRUE;


src/libostree/ostree-repo-pull.c
```diff
@@ -1538,7 +1538,7 @@ scan_one_metadata_object_c (OtPullData         *pull_data,
       /* Even though we already have the commit, we always try to (re)fetch the
        * detached metadata before scanning it, in case new signatures appear.
        * https://github.com/projectatomic/rpm-ostree/issues/630 */
-      if (!g_hash_table_contains (pull_data->fetched_detached_metadata, tmp_checksum))
+      if (!pull_data->remote_repo_local && !g_hash_table_contains (pull_data->fetched_detached_metadata, tmp_checksum))
```
jlebon: If the `.commit` was already stored though (even before importing it just up there), we still need to fetch the detached metadata, no? In that situation (where we already have the `.commit` object from the start), we should just try to fetch the metadata regardless of whether we already have the detached metadata or not, in case new signatures appear.

Maybe something like this instead?

```diff
    if (pull_data->remote_repo_local)
      {
        if (!is_stored)
          {
            if (objtype == OSTREE_OBJECT_TYPE_COMMIT)
              {
                if (!write_commitpartial_for (pull_data, tmp_checksum, error))
                  goto out;
              }
            if (!ostree_repo_import_object_from_with_trust (pull_data->repo, pull_data->remote_repo_local,
                                                            objtype, tmp_checksum, !pull_data->is_untrusted,
                                                            cancellable, error))
              goto out;
+           g_hash_table_add (pull_data->fetched_detached_metadata, tmp_checksum);
          }
        is_stored = TRUE;
        is_requested = TRUE;
      }
```

Then we won't fetch the metadata if it's freshly imported, and we also make sure that the hash table is accurate.

cgwalters: Good idea! Though it needed a conditional for the object type (and a strdup).  Done below :arrow_down:

jlebon:
@rh-atomic-bot r+ ae384b5

cgwalters:
Rebased :surfing_man:

jlebon:
@rh-atomic-bot r+ 5cac4ab

PR #979, opened 2017-06-29
lib/pull: Some small style porting

cgwalters:
I'd mostly been avoiding this file since there's always patches outstanding, but
these few functions shouldn't conflict much.


jlebon:
@rh-atomic-bot r+ aaa1314

PR #980, opened 2017-06-29
tree-wide: Replace various uses of `archive-z2` → `archive`

cgwalters:
The `-z2` is annoying now since it's really a legacy; we've long
since supported typing `archive`.  Convert the docs fully and
explain that.

Also do some (but not all) of the tests just to encourage newer tests to use
`archive` too.


jlebon:
@rh-atomic-bot r+ 3e83d23

PR #981, opened 2017-06-29
cmdline/pull: Print final status even if noninteractive

cgwalters:
Previously, `ostree pull` was silent if not on a tty.  I don't
see a reason not to print the final status line at least.  This
is prep for more work in the test suite, so I can write assertions
on the output.

But it should also be nicer for people who e.g. do an `ostree pull` in a Jenkins
job or whatever.


jlebon:
Nice!
@rh-atomic-bot r+ cde21be

PR #982, opened 2017-06-29
 Add "pull --localcache-repo"

cgwalters:
The main use case I'm targeting this for is the Fedora Atomic Host installer
case where we embed the repo content in the installer, but we may want to
kickstart and download newer content. There, while we want to get a newer ref,
we can still use the local repo as an object cache, since we have it sitting
there in memory anyways.

Another case is where one has a host ostree (say e.g. Fedora Atomic
Workstation), and one wants to create a local archive mirror of FAH. Then one
can use `pull --reference /ostree/repo` and pull the common objects (e.g.
contents of `bash.rpm` etc.)

Closes: ostreedev#975

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -702,13 +702,17 @@ scan_dirtree_object (OtPullData   *pull_data,
       if (file_is_stored)
         continue;
 
+      /* Already have a request pending?  If so, move on to the next */
+      if (g_hash_table_lookup (pull_data->requested_content, file_checksum))
+        continue;
+
       /* Is this a local repo? */
       if (pull_data->remote_repo_local)
         {
           if (!import_one_local_content_object (pull_data, file_checksum, cancellable, error))
             return FALSE;
         }
-      else if (!g_hash_table_lookup (pull_data->requested_content, file_checksum))
+      else
         {
           /* In this case we're doing HTTP pulls */
           g_hash_table_add (pull_data->requested_content, file_checksum);
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -64,6 +64,7 @@ typedef struct {
   GPtrArray     *meta_mirrorlist;    /* List of base URIs for fetching metadata */
   GPtrArray     *content_mirrorlist; /* List of base URIs for fetching content */
   OstreeRepo   *remote_repo_local;
+  GPtrArray    *localcache_repos; /* Array<OstreeRepo> */
 
   GMainContext    *main_context;
   GCancellable *cancellable;
@@ -118,6 +119,9 @@ typedef struct {
   guint             n_fetched_deltapart_fallbacks;
   guint             n_fetched_metadata;
   guint             n_fetched_content;
+  /* Objects from pull --reference */
+  guint             n_fetched_localcache_metadata;
+  guint             n_fetched_localcache_content;
 
   int               maxdepth;
   guint64           start_time;
@@ -236,6 +240,8 @@ update_progress (gpointer user_data)
                              "scanned-metadata", "u", n_scanned_metadata,
                              "bytes-transferred", "t", bytes_transferred,
                              "start-time", "t", start_time,
+                             "metadata-fetched-localcache", "u", pull_data->n_fetched_localcache_metadata,
+                             "content-fetched-localcache", "u", pull_data->n_fetched_localcache_content,
                              /* Deltas */
                              "fetched-delta-parts",
                                   "u", pull_data->n_fetched_deltaparts,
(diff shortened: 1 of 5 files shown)


cgwalters:
Depends: https://github.com/ostreedev/ostree/pull/981

cgwalters:
And oh yeah, this works *really* well.  Check this out, as mentioned in the commit msg, I'm doing a fresh mirror of FAH but reusing my host (Atomic WS) repo:

```
[root@localhost repomirror]# ostree --repo=repo-test-reference pull --mirror -R /srv/host/ostree fah-25 fedora-atomic/25/x86_64/docker-host



GPG: Verification enabled, found 1 signature:

  Signature made Sun 25 Jun 2017 09:08:34 PM UTC using RSA key ID 4089D8F2FDB19C98
  Good signature from "Fedora 25 Primary <fedora-25-primary@fedoraproject.org>"

459 metadata, 1445 content objects fetched (6433 meta, 20181 content local); 204282 KiB transferred in 114 seconds
```

25000 objects in common!

src/libostree/ostree-repo-pull.c
```diff
@@ -64,6 +64,7 @@ typedef struct {
   GPtrArray     *meta_mirrorlist;    /* List of base URIs for fetching metadata */
   GPtrArray     *content_mirrorlist; /* List of base URIs for fetching content */
   OstreeRepo   *remote_repo_local;
+  GPtrArray    *repo_refs; /* Array<OstreeRepo> */
```
jlebon: A bit bikeshed, but can we call this the full `repo_references`? (Or maybe even `repo_local_references`). Given that "refs" already means something else in this code.

src/libostree/ostree-repo-pull.c
```diff
@@ -705,15 +710,43 @@ scan_dirtree_object (OtPullData   *pull_data,
       /* Is this a local repo? */
       if (pull_data->remote_repo_local)
         {
-          if (!import_one_local_content_object (pull_data, file_checksum, cancellable, error))
+          if (!import_one_local_content_object (pull_data, pull_data->remote_repo_local,
+                                                file_checksum, cancellable, error))
             return FALSE;
         }
-      else if (!g_hash_table_lookup (pull_data->requested_content, file_checksum))
+      else
```
jlebon: Shouldn't this still be `else if (!g_hash_table_lookup (pull_data->requested_content, file_checksum))`? If the request has already been made somehow, then we should commit to that, no?

src/libostree/ostree-repo-pull.c
```diff
@@ -1523,6 +1556,30 @@ scan_one_metadata_object_c (OtPullData         *pull_data,
       is_stored = TRUE;
       is_requested = TRUE;
     }
+  /* Do we have any --reference local repos? */
+  else if (pull_data->repo_refs)
```
jlebon: Should this be `else if (!is_stored && pull_data->repo_refs)` ?

cgwalters:
https://bugzilla.gnome.org/show_bug.cgi?id=764735 is related here.

cgwalters:
In testing a bit more, one downside of this is that we're doing the reads/writes in the main thread.  It'd be better to make this async.

src/libostree/ostree-repo-pull.c
cgwalters: Yeah.  Or maybe skip the git naming and call it `--localcache-repo` in the cmdline and `localcache_repos` in variables?

jlebon: Yeah, that sounds much better.

cgwalters:
Rebased and comments addressed :arrow_up:

cgwalters:
The async thing can come later.

src/libostree/ostree-repo-pull.c
```diff
@@ -1523,6 +1559,30 @@ scan_one_metadata_object_c (OtPullData         *pull_data,
       is_stored = TRUE;
       is_requested = TRUE;
     }
+  /* Do we have any --reference local repos? */
```
jlebon: Nit: outdated comment.

src/libostree/ostree-repo-pull.c
```diff
@@ -1523,6 +1559,30 @@ scan_one_metadata_object_c (OtPullData         *pull_data,
...
+          if (!localcache_repo_has_obj)
+            continue;
+          if (!ostree_repo_import_object_from_with_trust (pull_data->repo, refd_repo,
+                                                          objtype, tmp_checksum,
+                                                          !pull_data->is_untrusted,
+                                                          cancellable, error))
+            return FALSE;
+          is_stored = TRUE;
+          is_requested = TRUE;
+          pull_data->n_fetched_localcache_metadata++;
```
jlebon: I was going to suggest adding:

```c
  g_hash_table_add (pull_data->fetched_detached_metadata, g_strdup (tmp_checksum));
```

here as well... but I think it's better without it, right?

It seems safer to always fetch it from the actual remote. Not to mention, we might not be embedding detached metadata in the local cache repo for some reason (like in the Fedora ISO).

cgwalters: We are embedding it, but not necessarily for the to-be-fetched commit obviously.

cgwalters: Fixup :arrow_down:

jlebon:
@rh-atomic-bot r+ a39f35e

PR #984, opened 2017-06-29
lib/commit: Port a few minor functions to new style

cgwalters:
Not sure why these weren't converted before.


jlebon:
@rh-atomic-bot r+ 0decd5f

cgwalters:
@rh-atomic-bot retry

PR #985, opened 2017-06-29
lib/repo: Fix repo-finder deleting remote configs when run

pwithnall:
An inverted condition in _ostree_repo_add_remote() was causing the
OstreeRepoFinder to delete precisely the wrong remote
configurations from memory once it was finished. It’s supposed to delete
the ones which it transiently added; but was instead deleting all the
existing remote configurations.


cgwalters:
@rh-atomic-bot r+ 18df2d7

PR #986, opened 2017-06-30
lib/commit: Fix a tmpfile fd leak in static delta processing

cgwalters:
I had thought `glnx_link_tmpfile_at()` actually consumed the tmpfile;
it does consume the *path* but not the fd.  In the non-delta path
things were fine since we used the autocleanup.

But the delta code had a tmpfile allocated in its struct that got reused, and
hence leaked the fd. Fix this by making the commit API actually consume the
tmpfile fully, just like the path path.


jlebon:
@rh-atomic-bot r+ 95c524d

PR #987, opened 2017-06-30
 lib/repo: Add min-free-space-percent option, default 3%

cgwalters:
For ostree-as-host, we're the superuser, so we'll blow past
any reserved free space by default.  While deltas have size
metadata, if one happens to do a loose fetch, we can fill
up the disk.

Another case is flatpak: the system helper has similar concerns
here as ostree-as-host, and for `flatpak --user`, we also
want to be nice and avoid filling up the user's quota.

Closes: ostreedev#962

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -23,6 +23,7 @@
 #include "config.h"
 
 #include <glib-unix.h>
+#include <sys/statvfs.h>
 #include <gio/gfiledescriptorbased.h>
 #include <gio/gunixinputstream.h>
 #include <gio/gunixoutputstream.h>
@@ -574,6 +575,24 @@ write_content_object (OstreeRepo         *self,
   else
     size = 0;
 
+  /* Free space check; only applies during transactions */
+  if (self->min_free_space_percent > 0 && self->in_transaction)
+    {
+      g_mutex_lock (&self->txn_stats_lock);
+      g_assert_cmpint (self->txn_blocksize, >, 0);
+      const fsblkcnt_t object_blocks = (size / self->txn_blocksize) + 1;
+      if (object_blocks > self->max_txn_blocks)
+        {
+          g_mutex_unlock (&self->txn_stats_lock);
+          g_autofree char *formatted_required = g_format_size ((guint64)object_blocks * self->txn_blocksize);
+          return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
+                             self->min_free_space_percent, formatted_required);
+        }
+      /* This is the main bit that needs mutex protection */
+      self->max_txn_blocks -= object_blocks;
+      g_mutex_unlock (&self->txn_stats_lock);
+    }
+
   /* For regular files, we create them with default mode, and only
    * later apply any xattrs and setuid bits.  The rationale here
    * is that an attacker on the network with the ability to MITM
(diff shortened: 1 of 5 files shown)


cgwalters:
OK, this works pretty well in my testing, though I haven't added a unit or integration test yet.   Debating a bit which to do.

src/libostree/ostree-repo-commit.c
```diff
@@ -1057,6 +1093,29 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
...
   self->in_transaction = TRUE;
+  if (self->min_free_space_percent > 0)
+    {
+      struct statvfs stvfsbuf;
+      if (TEMP_FAILURE_RETRY (fstatvfs (self->repo_dir_fd, &stvfsbuf)) < 0)
+        return glnx_throw_errno_prefix (error, "fstatvfs");
+      g_mutex_lock (&self->txn_stats_lock);
+      self->txn_blocksize = stvfsbuf.f_bsize;
+      /* Convert fragment to blocks to compute the total */
+      guint64 total_blocks = (stvfsbuf.f_frsize * stvfsbuf.f_blocks) / stvfsbuf.f_bsize;
```
jlebon: Doesn't this multiplication yield the total size of the filesystem in bytes already? `statvfs(2)` says:

```
fsblkcnt_t     f_blocks;   /* Size of fs in f_frsize units */
```

Writing a quick program to check this:

```
$ cat tmp.c                    
#include <sys/statvfs.h>                         
#include <stdio.h>      

int main(void)          
{                       
    struct statvfs stat;                         
    if (statvfs("/", &stat) < 0)                 
        return 1;       

    unsigned long long total_bytes = stat.f_blocks * stat.f_frsize;                                
    printf("total size in bytes: %llu\n", total_bytes);                                            
    printf("total size in GB: %llu\n", total_bytes >> 30);                                         

    return 0;           
}                       
$ ./tmp                                                                          
total size in bytes: 95133081600                 
total size in GB: 88    
$ df -hT /                                                                       
Filesystem              Type  Size  Used Avail Use% Mounted on                                     
/dev/mapper/fedora-root ext4   89G   63G   22G  75% /          
```

So then, we need to divide this by `f_bsize` to be made comparable to `bfree`, right?

jlebon:
> OK, this works pretty well in my testing, though I haven't added a unit or integration test yet. Debating a bit which to do.

Hacky, but one way would be to add a test to the `f25ah-insttest` suite that uses loopback devices. The advantage is that we can easily try out different fs sizes and block sizes.

src/libostree/ostree-repo-commit.c
cgwalters: The first multiplication indeed yield size in bytes.  The intent is to do all calculations in terms of blocks, not bytes.  That's how the filesystem works ultimately.

When we're looking at object sizes, we divide it by the block size (and add one to be conservative, though obviously in practice an object could land exactly on a block boundary).

src/libostree/ostree-repo-commit.c
```diff
@@ -551,6 +569,24 @@ write_content_object (OstreeRepo         *self,
   else
     size = 0;
 
+  /* Free space check; only applies during transactions */
+  if (self->min_free_space_percent > 0 && self->in_transaction)
+    {
+      g_mutex_lock (&self->txn_stats_lock);
+      g_assert_cmpint (self->txn_blocksize, >, 0);
+      const fsblkcnt_t object_blocks = (size / self->txn_blocksize) + 1;
```
cgwalters: (Here's where we convert objects into blocks)

src/libostree/ostree-repo-commit.c
jlebon: Oh goodness, I completely missed the `/ stvfsbuf.f_bsize;` chunk of that line! (I really like the split view when looking at diffs, but this chops up lines much more when looking on my laptop).

cgwalters:
:surfing_man: Rebased and now with an installed :building_construction: test :white_check_mark:

jlebon:
@rh-atomic-bot r+ 4447145

PR #988, opened 2017-06-30
bin/commit: Port helper functions to new style

cgwalters:
Prep for more work here.  Can't yet port the main function
without a cleanup for transactions.


jlebon:
@rh-atomic-bot r+ 50f65d8

PR #989, opened 2017-06-30
 lib/commit: Ensure bare-user objects are always user-readable

cgwalters:
Some of the Jenkins jobs for Fedora Atomic Host broke after updating
to 2017.7, and it turns out that we regressed handling unreadable
files in `bare-user` mode.  An example of this is `/etc/shadow`, which
ends up in the ostree-as-host content as `/usr/etc/shadow`.

Now there are better fixes here; we should probably delete it and create it
during the config merge if it doesn't exist.  In general, having secret files in
ostree really isn't supported, so it doesn't make sense to include them.

But let's fix this regression - when operating as an unprivileged user we don't
have `CAP_DAC_OVERRIDE` and hence will fail to open un-user-readable objects.

(We still preserve the actual `0` mode of course in the xattr and will
 apply it in `bare`)

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -253,13 +253,15 @@ commit_loose_regfile_object (OstreeRepo        *self,
       /* Note that previously this path added `| 0755` which made every
        * file executable, see
        * https://github.com/ostreedev/ostree/issues/907
+       * We then changed it to mask by 0775, but we always need at least read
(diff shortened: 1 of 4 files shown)


cgwalters:
Depends: https://github.com/ostreedev/ostree/pull/988

cgwalters:
:surfing_man: Rebased

jlebon:
@rh-atomic-bot r+ d6a7a3b

PR #971, opened 2017-06-28
codebase: start using GLNX_HASH_TABLE_FOREACH macros

jlebon:
Use the new macros introduced recently in libglnx to make iterating over
hash tables cleaner. This is just a start, it does not migrate the whole
tree.

Update submodule: libglnx

Requires: https://github.com/GNOME/libglnx/pull/58

PR #990, opened 2017-06-30
pull: fix GLNX_HASH_TABLE_FOREACH_KV regression

jlebon:
This is a regression from #971. We were stuffing a pointer size inside a
variable of integer size. So the assignment was spilling over into other
variables' storage space. Actually use a gpointer and `GPOINTER_TO_INT` as
was done originally.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -4948,8 +4948,10 @@ ostree_repo_pull_from_remotes_async (OstreeRepo                           *self,
 
   /* Any refs left un-downloaded? If so, we’ve failed. */
   GLNX_HASH_TABLE_FOREACH_KV (refs_pulled, const OstreeCollectionRef*, ref,
-                                           gboolean, is_pulled)
+                                           gpointer, is_pulled_pointer)
     {
+      gboolean is_pulled = GPOINTER_TO_INT (is_pulled_pointer);
+
       if (is_pulled)
         continue;
 
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -1091,9 +1091,9 @@ ostree_repo_checkout_gc (OstreeRepo        *self,
   if (!to_clean_dirs)
     return TRUE; /* Note early return */
 
-  GLNX_HASH_TABLE_FOREACH (to_clean_dirs, guint, prefix)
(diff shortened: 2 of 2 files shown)


jlebon:

src/libostree/ostree-repo-pull.c
```diff
@@ -4948,8 +4948,10 @@ ostree_repo_pull_from_remotes_async (OstreeRepo                           *self,
 
   /* Any refs left un-downloaded? If so, we’ve failed. */
   GLNX_HASH_TABLE_FOREACH_KV (refs_pulled, const OstreeCollectionRef*, ref,
-                                           gboolean, is_pulled)
+                                           gpointer, is_pulled_pointer)
```
cgwalters: Oooh, evil! :japanese_goblin: 

How about https://github.com/GNOME/libglnx/pull/59 too?

cgwalters:
Compiling with that on also reveals we need at least:

```
diff --git a/src/libostree/ostree-repo-checkout.c b/src/libostree/ostree-repo-checkout.c
index 8729b81..a7af9f0 100644
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -1091,9 +1091,9 @@ ostree_repo_checkout_gc (OstreeRepo        *self,
   if (!to_clean_dirs)
     return TRUE; /* Note early return */
 
-  GLNX_HASH_TABLE_FOREACH (to_clean_dirs, guint, prefix)
+  GLNX_HASH_TABLE_FOREACH (to_clean_dirs, gsize, prefix)
     {
-      g_autofree char *objdir_name = g_strdup_printf ("%02x", prefix);
+      g_autofree char *objdir_name = g_strdup_printf ("%02x", (guint)prefix);
       g_auto(GLnxDirFdIterator) dfd_iter = { 0, };
 
       if (!glnx_dirfd_iterator_init_at (self->uncompressed_objects_dir_fd, objdir_name, FALSE,
```

Maybe this is something where we can have some magic that checks whether the provided argument is an integer...dunno.

cgwalters:
This looks fine to me...but maybe change this to update libglnx and fix everything in one go?

jlebon:
Bumped libglnx and updated commit msg!

cgwalters:
@rh-atomic-bot r+ f2986f0

pwithnall:
Excellent, I can confirm this fixes the crash from #971, thanks.

PR #992, opened 2017-07-04
pull: Cleanly error when doing local pulls of remote-prefixed refs

cgwalters:
In the storage PR I was trying to do a `pull-local` of the whole
`/ostree/repo` on the system, which ended up triggering a `g_critical()`
in the collections code, since we tried to parse a remote-prefixed ref
`fedora:fedora/26/x86_64/atomic-host` as a ref.

I'm not sure offhand what our behavior in this case *should* be.  I
think git only clones local refs, but I need to check.

This corner case arises only with `pull-local`.  But in any case,
while we were previously saying this is programmer error, since it's
so easy to pass various unchecked input into the pull machinery,
make invalid refs an explicit error.

### DIFF
--- a/tests/pull-test.sh
+++ b/tests/pull-test.sh
@@ -176,7 +176,10 @@ if ${CMD_PREFIX} ostree --repo=mirrorrepo-local pull-local otherrepo 2>err.txt; 
     fatal "pull with mixed refs succeeded?"
 fi
 assert_file_has_content err.txt "error: Invalid ref name origin:main"
-echo "ok pull local mirror error (mixed refs)"
+${CMD_PREFIX} ostree --repo=mirrorrepo-local pull-local otherrepo localbranch
+${CMD_PREFIX} ostree --repo=mirrorrepo-local rev-parse localbranch
+${CMD_PREFIX} ostree --repo=mirrorrepo-local fsck
+echo "ok pull-local mirror errors with mixed refs"
 
 cd ${test_tmpdir}
 ${CMD_PREFIX} ostree --repo=ostree-srv/gnomerepo commit -b main -s "Metadata string" --add-detached-metadata-string=SIGNATURE=HANCOCK --tree=ref=main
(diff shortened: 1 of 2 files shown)


tests/pull-test.sh
```diff
@@ -162,6 +162,23 @@ $OSTREE show main >/dev/null
...
+rm checkout-origin-main -rf
+${CMD_PREFIX} ostree --repo=ostree-srv/gnomerepo checkout main checkout-origin-main
+${CMD_PREFIX} ostree --repo=otherrepo commit -b localbranch --tree=dir=checkout-origin-main
+${CMD_PREFIX} ostree --repo=otherrepo remote add --set=gpg-verify=false origin file://$(pwd)/ostree-srv/gnomerepo
+${CMD_PREFIX} ostree --repo=otherrepo pull origin main
+rm mirrorrepo-local -rf
+ostree_repo_init mirrorrepo-local --mode=archive
+if ${CMD_PREFIX} ostree --repo=mirrorrepo-local pull-local otherrepo 2>err.txt; then
+    fatal "pull with mixed refs succeeded?"
+fi
```
jlebon: Can we add a sanity check here that

```
${CMD_PREFIX} ostree --repo=mirrorrepo-local pull-local otherrepo localbranch
```

still works even though the repo has mixed refs? Just to make sure this is only an issue in the unspecified case.

tests/pull-test.sh
```diff
@@ -162,6 +162,23 @@ $OSTREE show main >/dev/null
...
+assert_file_has_content err.txt "error: Invalid ref name origin:main"
+echo "ok pull local mirror error (mixed refs)"
```
jlebon: Nit: both this test and the previous test use as ok strings "ok pull local mirror ..." even though in the former, it's talking about `pull --mirror` from a local repo, and here it's talking about `pull-local` from a local repo. Maybe change this one to just e.g. `ok pull-local error (mixed refs)` ?

jlebon:
@rh-atomic-bot r+ 2f48856

Issue #991, opened 2017-07-02
Delta update fails with EPERM (O_TMPFILE + old glibc?)


alexlarsson:
In flathub i got: https://flathub.org/builds/#/builders/4/builds/446/steps/9/logs/stdio
```
Updating: org.gnome.Platform/x86_64/3.24 from flathub
Error: Failed to update org.gnome.Platform/x86_64/3.24: While pulling runtime/org.gnome.Platform/x86_64/3.24 from remote flathub: opcode open-splice-and-close: opcode open-splice-and-close: Writing object 579128a0d09b4fad3ed5156eef81e495e2da0d145e26002199553c956b5954be.file: Unable to set xattr: Permission denied
```
Which is a per-user repo that is type bare-user. Ostree version is 2017.7.
Updating with --no-static-deltas worked.

cgwalters:
The `Unable to set xattr` string comes from `write_file_metadata_to_xattr()` - i.e. we failed to write the `bare-user` `user.ostreemeta` xattr.

Did you expect to have a `bare-user-only` repo?  

Although I can't imagine though why `--no-static-deltas` would work.  The code paths are really the same between the two.

cgwalters:
I'm going to operate on the theory here that what happened is when you did a pull manually, you actually used the system repo instead of the per-user one or something like that.

This PR should make things a lot more clear: https://github.com/ostreedev/ostree/pull/993

cgwalters:
That build log also has an assertion error, urg.

alexlarsson:
I did no expect a bare-user-only pull. But I also did not expect an xattr error, because it was updating the repo in a location that definately supported xattrs, because we've installed a lot of things there before, and the update wih --non-static-deltas worked.

I don't know what could have caused this, and since this was the live flathub build machine i no longer have this state. 

One guess is that some codepath writes a file in /tmp rather than the ostree repo tmp dir.

cgwalters:
Is it continuing to happen?  Offhand...oh wait, I see a potential issue; in the delta path it looks like we're not doing a `fchmod()` before we try to set the xattr.  That's possibly why you were seeing `EPERM`.  This isn't an issue with `EOPNOTSUPP` as we were thinking at first.

cgwalters:
[...]

PR #993, opened 2017-07-05
lib/repo: Immediately error creating bare-user repo on tmpfs

cgwalters:
And in general, if for some reason we can't write `user.` xattrs, provide an
error immediately rather than doing it during a later pull. This way the failure
cause is a lot more obvious.

Related: https://github.com/ostreedev/ostree/issues/991

### DIFF
--- a/tests/installed/itest-bareuser-nouserxattrs.sh
+++ b/tests/installed/itest-bareuser-nouserxattrs.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+# Test that initializing a bare-user repo on tmpfs fails
+# Maybe at some point this will be fixed in the kernel
+# but I doubt it'll be soon
+# https://www.spinics.net/lists/linux-mm/msg109775.html
+
+set -xeuo pipefail
+
+dn=$(dirname $0)
+. ${dn}/libinsttest.sh
+
+test_tmpdir=$(prepare_tmpdir)
+trap _tmpdir_cleanup EXIT
+cd ${test_tmpdir}
+
+mkdir mnt
+mount -t tmpfs tmpfs mnt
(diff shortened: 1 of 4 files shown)


tests/installed/itest-bareuser-nouserxattrs.sh
```diff
@@ -0,0 +1,23 @@
...
+trap _tmpdir_cleanup EXIT
+cd ${test_tmpdir}
+
+mkdir mnt
+mount -t tmpfs tmpfs mnt
+if ostree --repo=mnt/repo init --mode=bare-user 2>err.txt; then
+    assert_not_reached "bare-user on tmpfs worked?"
+fi
+assert_file_has_content err.txt "Operation not supported"
+umount mnt
```
jlebon: Maybe move this to within the if statement and right after as well? Or alternatively, put something like

```sh
trap 'umount mnt && _tmpdir_cleanup' EXIT
```

right after mounting it.

cgwalters: The reason I didn't do that is they don't compose: https://stackoverflow.com/questions/3338030/multiple-bash-traps-for-the-same-signal

I pushed a fixup.

jlebon: Yeah, it's tricky. I ran into this in PAPR as well. That's why I still had `_tmpdir_cleanup` in my suggestion. Though this way looks nicer anyway!

jlebon:
@rh-atomic-bot r+ 955cd55

jlebon:
@rh-atomic-bot r+ 955cd55

jlebon:
Let's see if this does the trick.
@rh-atomic-bot retry

PR #994, opened 2017-07-05
Release 2017.8

cgwalters:
I think we should do this before we land a lot more stuff; in particular I'd like to release:

 - https://github.com/ostreedev/ostree/pull/989
 - https://github.com/ostreedev/ostree/pull/987

### DIFF
--- a/src/libostree/libostree-devel.sym
+++ b/src/libostree/libostree-devel.sym
@@ -18,6 +18,8 @@
 ***/
 
 /* Add new symbols here.  Release commits should copy this section into -released.sym. */
+LIBOSTREE_2017.9 {
+};
 
 /* Stub section for the stable release *after* this development one; don't
  * edit this other than to update the last number.  This is just a copy/paste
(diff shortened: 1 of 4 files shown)


jlebon:
And maybe https://github.com/ostreedev/ostree/pull/993 as well? It looks like a straightforward useful fix.

src/libostree/libostree-devel.sym
```diff
@@ -18,10 +18,6 @@
 ***/
 
 /* Add new symbols here.  Release commits should copy this section into -released.sym. */
-LIBOSTREE_2017.8 {
-global:
-  ostree_validate_remote_name;
-} LIBOSTREE_2017.7;
```
jlebon: It looks like you may also have to line up the hunk for the next release as well.

cgwalters: Done :arrow_down:

jlebon:
This looks 👍. Just waiting for Homu to be back online and get #993 in.

jlebon:
@rh-atomic-bot r+ 5b99347

PR #997, opened 2017-07-07
checkout: Don't set dir mtime to 0 when doing a force copy checkout

cgwalters:
When we [switched to using checkout + force_copy](https://github.com/ostreedev/ostree/commit/e8efd1c8dcaad8fbd3b05c400972d237406263e7),
a side effect that went unnoticed at the time is that we started
setting directory mtimes to zero.

See the below bug where we long ago set the file times to zero, which got fixed,
so let's not regress things by setting the directory times to zero either. (Even
though AFAICS GNU tar doesn't complain about those)

https://bugzilla.redhat.com/show_bug.cgi?id=1229160

### DIFF
--- a/tests/installed/itest-deploy-selinux.sh
+++ b/tests/installed/itest-deploy-selinux.sh
@@ -13,7 +13,7 @@ new_deployment_path=/ostree/deploy/${host_osname}/deploy/${host_commit}.1
 
 # Test /etc directory mtime
 if ! test ${new_deployment_path}/etc/NetworkManager -nt /etc/NetworkManager; then
-    ls -al ${new_deployment_path}/etc/NetworkManager -nt /etc/NetworkManager
+    ls -al ${new_deployment_path}/etc/NetworkManager /etc/NetworkManager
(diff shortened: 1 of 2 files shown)


tests/installed/itest-deploy-selinux.sh
```diff
@@ -11,6 +11,12 @@ dn=$(dirname $0)
 ostree admin deploy --karg-proc-cmdline ${host_refspec}
 new_deployment_path=/ostree/deploy/${host_osname}/deploy/${host_commit}.1
 
+# Test /etc directory mtime
+if ! test ${new_deployment_path}/etc/NetworkManager -nt /etc/NetworkManager; then
+    ls -al ${new_deployment_path}/etc/NetworkManager -nt /etc/NetworkManager
```
jlebon: Nit: extra `-nt` here.

cgwalters: Good catch, fixup :arrow_down:

jlebon:
@rh-atomic-bot r+ 0ece5f9

PR #999, opened 2017-07-10
lib/repo-commit: Fix types of content size cache entries

pwithnall:
Use goffset rather than gsize for file sizes. More importantly, get the
unpacked_size from g_file_info_get_size() (goffset) rather than from the
splice return value, which has type gssize.

This will make a difference on 32-bit systems, where goffset is defined
as off64_t, but gsize is 32 bits.


cgwalters:
@rh-atomic-bot r+ dcec3b5

PR #1004, opened 2017-07-13
lib/pull: Avoid journaling 404s for optional content

cgwalters:
Currently in Fedora we don't sign summaries, and every use of
`rpm-ostree` would emit to the journal an error when we failed
to fetch it.

Fix this by having `OSTREE_FETCHER_REQUEST_OPTIONAL_CONTENT` tell the fetcher
not to journal 404 errors. While fixing this, we had a mix of two booleans vs
the flags; fix things so we consistently use the flags in the fetcher and pull
code.


cgwalters:
@rh-atomic-bot delegate=mbarnes

mbarnes:
@rh-atomic-bot r+

mbarnes:
@rh-atomic-bot r+ https://github.com/ostreedev/ostree/pull/1004/commits/85ee583084432d1ee0af1644d185ab2cfdf9879f

cgwalters:
Weird...homu went nuts.

cgwalters:
@rh-atomic-bot retry

cgwalters:
Ah...I think this is us hitting bugs with the non-checksum-suffixed (or "bare") `r+` path.

@rh-atomic-bot r=mbarnes 85ee583

mbarnes:
I think I've already strayed from the normal workflow, but all tests have passed so I'm merging manually.

PR #1006, opened 2017-07-14
lib/pull: Do local content imports async too

cgwalters:
This came up in <https://github.com/ostreedev/ostree/pull/982>; when
we added more direct local importing, we did it synchronously.

This was actually quite a regression when doing local pulls between different
modes; in particular between a bare mode and `archive`, as we were suddenly
doing gzip {de,}compression in the main thread.

Down the line actually...a simpler fix is probably to change things so that the
local path is really only used when we know we can hardlink; everything else
would go though the fetcher codepath but with `file://`.

But this isn't a lot more code, and the speed/interactivity win is large.

Note we're only doing content async with this patch. We could do metadata as
well; we have the object already local. But the metadata code path is messier,
and metadata objects are smaller.

Another area where this comes up is that in e.g. Fedora releng, most operations
talk to a NetApp via NFS. So this has the classic network filesystem problem
that operations that are normally cheap like `stat()` can actually have
nontrivial latency. Doing as much as possible in threads is better there too.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -614,8 +614,9 @@ validate_bareuseronly_mode (OtPullData *pull_data,
   return TRUE;
 }
 
-/* Import a single content object in the case where
- * we have pull_data->remote_repo_local.
+/* Synchronously import a single content object; this is used async for content,
+ * or synchronously for metadata. @src_repo is either
+ * pull_data->remote_repo_local or one of pull_data->localcache_repos.
  *
  * One important special case here is handling the
  * OSTREE_REPO_PULL_FLAGS_BAREUSERONLY_FILES flag.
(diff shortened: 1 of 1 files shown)


cgwalters:
Before:
```
[root@localhost repomirror]# rm repo-bare-user/ -rf && ostree --repo=repo-bare-user init --mode=bare-user && time ostree --repo=repo-bare-user pull-local repo fedora/26/x86_64/atomic-host

Scanning metadata: 2378

real    0m24.139s
user    0m6.213s
sys     0m6.273s
```
After:
```
[root@localhost repomirror]# rm repo-bare-user/ -rf && ostree --repo=repo-bare-user init --mode=bare-user && time ostree --repo=repo-bare-user pull-local repo fedora/26/x86_64/atomic-host   

Writing objects: 10124

real    0m13.575s
user    0m9.143s
sys     0m10.266s
```

src/libostree/ostree-repo-pull.c
```diff
@@ -674,6 +674,82 @@ import_one_local_content_object (OtPullData *pull_data,
...
+                                             iataskdata->checksum,
+                                             cancellable,
+                                             &local_error))
+    g_task_return_error (task, g_steal_pointer (&local_error));
+  else
+    g_task_return_boolean (task, TRUE);
+}
+
+/* Import a single content object in the case where
+ * we have pull_data->remote_repo_local.
```
jlebon: Nit: also mention `pull_data->localcache_repos` here?

cgwalters: Ok, updated :arrow_down:

jlebon:
@rh-atomic-bot r+ 51a8a1a

PR #1007, opened 2017-07-17
Move the include directive to the enum template

ebassi:
There is no actual written guarantee in glib-mkenums that the template
line specified using --fhead will be added after the templates specified
inside the template file. Since the template file is only used once, we
can simply move the `#include` directive inside the template, so that it
is guaranteed to be in the right place.


cgwalters:
This changed in https://bugzilla.gnome.org/show_bug.cgi?id=779332

cgwalters:
bot, add author to whitelist

cgwalters:
I spent only a brief amount of time looking at the glib-mkenums bits for this; it makes sense, but I'm worried that we're going to be breaking the build of more projects.

@rh-atomic-bot r+ 8b26685

PR #1008, opened 2017-07-17
Add a notion of "physical" sysroot, use for remote writing

cgwalters:
(Note this PR was reverted in <https://github.com/ostreedev/ostree/pull/902>;
 this version should be better)

Using `${sysroot}` to mean the physical storage root: We don't want to write to
`${sysroot}/etc/ostree/remotes.d`, since nothing will read it, and really
`${sysroot}` should just have `/ostree` (ideally). Today the Anaconda rpmostree
code ends up writing there. Fix this by adding a notion of "physical" sysroot.
We determine whether the path is physical by checking for `/sysroot`, which
exists in deployment roots (and there shouldn't be a `${sysroot}/sysroot`).

In order to unit test this, I added a `--sysroot` argument to `remote add`.
However, doing this better would require reworking the command line parsing for
the `remote` argument to support specifying `--repo` or `--sysroot`, and I
didn't quite want to do that yet in this patch.

This second iteration of this patch fixes the bug we hit the first time;
embarassingly enough I broke `ostree remote list` finding system remotes.
The fix is to have `ostree_repo_open()` figure out whether it's the same
as `/ostree/repo` for now.

Down the line...we might consider having the `ostree remote` command line itself
instatiate an `OstreeSysroot` by default, but this maximizes compatibility; we
just have to pay a small cost that `ostree` usage outside of that case like
`ostree static-delta` in a releng Jenkins job or whatever will do this `stat()`
too.

Closes: https://github.com/ostreedev/ostree/issues/892

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -32,6 +32,7 @@
 #include <glnx-console.h>
 
 #include "ostree-core-private.h"
+#include "ostree-sysroot-private.h"
 #include "ostree-remote-private.h"
 #include "ostree-repo-private.h"
 #include "ostree-repo-file.h"
@@ -124,6 +125,10 @@ static guint signals[LAST_SIGNAL] = { 0 };
 G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
 
 #define SYSCONF_REMOTES SHORTENED_SYSCONFDIR "/ostree/remotes.d"
+
+static GFile *
+get_remotes_d_dir (OstreeRepo          *self,
+                   GFile               *sysroot);
 
 OstreeRemote *
 _ostree_repo_get_remote (OstreeRepo  *self,
@@ -467,6 +472,7 @@ ostree_repo_finalize (GObject *object)
   if (self->uncompressed_objects_dir_fd != -1)
     (void) close (self->uncompressed_objects_dir_fd);
   g_clear_object (&self->sysroot_dir);
+  g_weak_ref_clear (&self->sysroot);
   g_free (self->remotes_config_dir);
 
   if (self->loose_object_devino_hash)
@@ -546,10 +552,6 @@ ostree_repo_constructed (GObject *object)
   OstreeRepo *self = OSTREE_REPO (object);
 
   g_assert (self->repodir != NULL);
-
-  /* Ensure the "sysroot-path" property is set. */
-  if (self->sysroot_dir == NULL)
-    self->sysroot_dir = g_object_ref (_ostree_get_default_sysroot_path ());
 
   G_OBJECT_CLASS (ostree_repo_parent_class)->constructed (object);
 }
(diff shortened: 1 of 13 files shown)


cgwalters:
Though, maybe we could have anything that needs to know this "is a system repo" property like `ostree_remote_change()` compute it on demand.  We also have `ostree_repo_is_system()` but if we did it lazily in both places it'd probably work well.

src/ostree/ot-remote-builtin-add.c
```diff
@@ -66,6 +69,20 @@ ot_remote_builtin_add (int argc, char **argv, GCancellable *cancellable, GError
                                     OSTREE_BUILTIN_FLAG_NONE, &repo, cancellable, error))
     goto out;
 
+  /* As a special case, we can take a --sysroot argument. Currently we also
+   * require --repo because fixing that needs more cmdline rework.
+   */
```
mbarnes: You could pass `OSTREE_BUILTIN_FLAG_NO_REPO` to `ostree_option_context_parse` and define your own `--repo` option here so it's not required in the presence of `--sysroot`.

src/libostree/ostree-repo.c
```diff
@@ -1874,14 +1866,57 @@ append_one_remote_config (OstreeRepo      *self,
...
+        case OSTREE_REPO_SYSROOT_KIND_IS_SYSROOT_OSTREE:
+          sysroot = sysroot_owned = g_file_new_for_path ("/");
+          break;
+        case OSTREE_REPO_SYSROOT_KIND_VIA_SYSROOT:
+          {
+            sysroot_ref = (OstreeSysroot*)g_weak_ref_get (&self->sysroot);
+            /* Only write to /etc/ostree/remotes.d if we are pointed at a deployment */
+            if (sysroot_ref != NULL && !sysroot_ref->is_physical)
+              sysroot = ostree_sysroot_get_path (sysroot_ref);
+          }
```
mbarnes: nit: Brackets not needed here?

src/libostree/ostree-repo.c
```diff
@@ -1874,14 +1866,57 @@ append_one_remote_config (OstreeRepo      *self,
...
+          break;
+        }
+    }
+  /* For backwards compat, also fall back to the sysroot-path variable, which we
+   * don't set anymore internally, and I hope no one else uses.
+   */
+  if (sysroot == NULL && sysroot_ref == NULL)
+    sysroot = self->sysroot_dir;
```
cgwalters: Mmm...maybe?  We don't have a lot of precedent for deprecation-via-warning here.  The code will still use it until later PRs; maybe later we can remember to deprecate it fully?

mbarnes: Okay, fair enough.

src/ostree/ot-remote-builtin-add.c
cgwalters: Indeed, good idea.  It turned out to be somewhat more subtle to implement than I thought; my original version of the below :arrow_down: just returned an `OstreeRepo*` but in fact the way the repo holds a weak ref to the sysroot meant that weak ref went away (since we were only transiently holding the sysroot), and hence the "is physical" state (plus the sysroot fd) were no longer accessible.

Fixed that by holding onto the `OstreeSysroot` on the toplevel.  Perhaps what we should really do is just have every ostree command also accept `--sysroot` or so.  Maybe later.

mbarnes:
@rh-atomic-bot r+ 5b593ec

PR #1011, opened 2017-07-18
Update libglnx, port various bits to new API

cgwalters:
Using the error prefixing in the delta processing allows us to
do new code style.  Also strip trailing whitespace.

Use error prefixing in a few other random places.  I didn't
hunt for all of them, just testing out the new API.

Use `glnx_fchmod()`.  Also note I dropped one `fchmod (tmpf, 0600)`
which is no longer necessary.

Update submodule: libglnx

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -26,6 +26,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <gio/gfiledescriptorbased.h>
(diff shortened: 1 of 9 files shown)


jlebon:
Nice! @rh-atomic-bot r+ fa07cf4

jlebon:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #1012, opened 2017-07-18
ci/papr: Update to F26

cgwalters:
In particular F25AH will stop getting updates.


jlebon:
@rh-atomic-bot r+ d302e84

PR #1013, opened 2017-07-18
build: Don't distribute generated man pages

smcv:
We build them in "make" and clean them in "make clean", so there
doesn't seem much point in shipping them pre-generated in the tarball.


cgwalters:
@rh-atomic-bot r+  95c7e68

PR #1015, opened 2017-07-19
core: Sanitize error text validating refs (e.g. against HTML)

cgwalters:
See: https://github.com/projectatomic/rpm-ostree/issues/885

If we get a successful Apache directory listing HTML when fetching what we
intend to be a ref, we'd dump the HTML into the error.

I did some scanning of the pull code, and this was the only case
I saw offhand where we were dumping text out into an error.  Which
makes sense, since most of our formats are binary, the exeptions I
think are just `repo/config` and `repo/refs/`.

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -41,9 +41,11 @@
  */
 static char *
 quash_string_for_error_message (const char *input,
-                                int         maxlen)
+                                ssize_t     len,
+                                ssize_t     maxlen)
 {
-  size_t len = strlen (input);
+  if (len == -1)
+    len = strlen (input);
   if (maxlen != -1 && maxlen < len)
     len = maxlen;
 #if GLIB_CHECK_VERSION(2, 52, 0)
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-core.c
```diff
@@ -35,6 +35,38 @@
 #define ALIGN_VALUE(this, boundary) \
   (( ((unsigned long)(this)) + (((unsigned long)(boundary)) -1)) & (~(((unsigned long)(boundary))-1)))
 
+/* Return a copy of @input suitable for addition to
+ * a GError message; newlines are quashed, the value
+ * is forced to be UTF-8, is truncated to @maxlen (if maxlen != -1).
+ */
+static char *
+quash_string_for_error_message (const char *input,
+                                int         maxlen)
```
mbarnes: Function could take a `len` argument, since the caller has already calculated `strlen(input)`.

mbarnes:
@rh-atomic-bot r+ f95818f

PR #1016, opened 2017-07-19
Fix sizes

cgwalters:


.papr.yml
```diff
@@ -91,6 +91,21 @@ tests:
 
 ---
 
+inherit: true
+required: true
+
+context: f26-introspection-tests
+
+env:
+    CFLAGS: ''  # Reset ASAN
```
jlebon: By respecifying `env` it effectively replaces the previous dict, so you shouldn't need that. Might want to leave it to be more explicit though.

src/libostree/ostree-repo-commit.c
```diff
@@ -734,6 +734,17 @@ write_content_object (OstreeRepo         *self,
     }
   else
     {
+      /* Update size metadata if configured */
```
jlebon: Can you elaborate on how this makes a difference? Aren't `phys_object_is_symlink` and `object_file_type == G_FILE_TYPE_REGULAR` already mutually exclusive?

.papr.yml
```diff
@@ -79,14 +81,25 @@ tests:
...
+inherit: true
+required: true
+
+context: f26-introspection-tests
 
 env:
-  CONFIGOPTS: "--with-curl --with-openssl"
+    CFLAGS: ''  # Reset ASAN
+    CONFIGOPTS: "--without-curl --without-openssl"
+    INSTALLED_TESTS_PATTERN: "tests/test-sizes.js"
```
smcv: Also tests/test-corruption.sh tests/test-pull-corruption.sh tests/test-core.js and tests/test-sysroot.js?

Otherwise looks good to me - the code that @jlebon commented on has just been hoisted higher up, not modified, so whether it actually needs to be testing for G_FILE_TYPE_REGULAR seems orthogonal to this bug.

src/libostree/ostree-repo-commit.c
smcv: Surely the important thing, rather than that they are mutually exclusive, would be whether they are exhaustive? (Mutually exclusive: !(a && b) always; exhaustive: (a || b) always)

There are three cases:

* G_FILE_TYPE_REGULAR (phys_object_is_symlink is false)
* G_FILE_TYPE_SYMBOLIC_LINK && mode in (BARE, BARE_USER_ONLY) (phys_object_is_symlink is true)
* G_FILE_TYPE_SYMBOLIC_LINK && mode not in (BARE, BARE_USER_ONLY) (phys_object_is_symlink is false)

By getting here we know we are in one of the two !phys_object_is_symlink cases, so not the second one; but we don't know whether it is the first or the third. So the check seems non-redundant to me.

cgwalters: Yeah, bare-user symlinks are the tricky case, for those we have `phys_object_is_symlink = FALSE, object_file_type = G_FILE_TYPE_SYMBOLIC_LINK`.  Although actually now that I look at this, the whole thing can only apply for `archive` repos:

```
      g_assert (repo_mode == OSTREE_REPO_MODE_ARCHIVE_Z2);

      if (self->generate_sizes)
        indexable = TRUE;
```

Anyways given the bare-user bits I think we should keep the conditional.

cgwalters:
@rh-atomic-bot delegate=smcv

src/libostree/ostree-repo-commit.c
jlebon: Ahh right, I missed the bare-user case. Thanks for clarifying!

smcv:
Still looks good to me. Some of the CI is failing at the setup stage; probably orthogonal, but I'd feel more confident about it if someone could retry `f26-introspection-tests`?

cgwalters:
bot, retest this please

cgwalters:
You have `r+` on this PR; note at least I don't always wait until the PR-per status says green, since the test are rerun for the merge anyways.  Though in this particular case it is useful to sanity check the test suite has the results we expect.

jlebon:
@rh-atomic-bot r+ d8d3c10

PR #1021, opened 2017-07-19
grub2: Handle aarch64 (at least on Fedora/RHEL)

cgwalters:
See: https://github.com/projectatomic/rpm-ostree-toolbox/issues/102#issuecomment-316483554

The logic here should now match what grubby does.  I have no
idea whether this applies to other distributions though.

### DIFF
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -38,9 +38,19 @@
  * among others.
  */
 #if defined(__i386__) || defined(__x86_64__)
-#define GRUB2_USES_16 1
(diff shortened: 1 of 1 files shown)


jlebon:
This looks sane to me. I don't have an aarch64 machine available to test this, though it should be fine to get it in, seeing as it makes no practical changes on non-aarch64 systems and it can't make the aarch64 situation worse than it already is. Not sure if you're waiting for someone that can test this, but feel free to `r=jlebon` otherwise.

cgwalters:
Yeah, if it doesn't work we can try again.

@rh-atomic-bot r=jlebon ae771fb

PR #1022, opened 2017-07-20
 tests: More fixes for gjs tests

cgwalters:
Previous to this commit, the gjs tests were installed-only; and our
logic for handling the "--enable-installed-tests=exclusive" logic
actually also meant they weren't installed.

It did work for me locally with `--enable-installed-tests`.

However, to make things fully symmetric, let's enable the js tests to also be
run under `make check`.

Also remove `corrupt-repo-ref.js` from the PAPR invocation since it's not
actually a unit test, it's a utility helper.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -191,7 +191,7 @@ js_installed_tests = \
 	$(NULL)
 
 if BUILDOPT_GJS
-dist_installed_test_scripts = $(js_installed_tests)
+_installed_or_uninstalled_test_scripts += $(js_installed_tests)
 else
 EXTRA_DIST += $(js_installed_tests)
 endif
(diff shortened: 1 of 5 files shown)


Makefile-tests.am
```diff
@@ -191,7 +191,7 @@ js_installed_tests = \
 	$(NULL)
 
 if BUILDOPT_GJS
-dist_installed_test_scripts = $(js_installed_tests)
```
cgwalters: The bug was that this line silently *replaced* a previous value.

cgwalters:
Blah, of course the introspection tests context had to flake. 

bot, retest this please

jlebon:
Our GHPRB is old and hard of hearing unfortunately, so the command has to be as part of its own comment.

jlebon:
bot, retest this please

jlebon:
Hmm, more flake. I suspected the mirror we inject, though it seems fine when I query it myself. Let's try it again and if it fails, I'll disable it.

@rh-atomic-bot r+ ed7e8c4

jlebon:
@rh-atomic-bot retry

PR #1024, opened 2017-07-21
lib: Add #defines for current well-known metadata keys

cgwalters:
This came up in https://github.com/projectatomic/rpm-ostree/issues/142

Let's add `#define`s for our metadata keys, with documentation so
that, well, they're documented.


cgwalters:

peterbaouoft:
LGTM

peterbaouoft:
@rh-atomic-bot r+ 9476969

jlebon:
@rh-atomic-bot retry

PR #1025, opened 2017-07-21
lib: Add #define for endoflife metadata key

peterbaouoft:
It is a continuation of https://github.com/ostreedev/ostree/pull/1024
It adds documentation for endoflife metadata key.


cgwalters:
LGTM!

@rh-atomic-bot r+  e61b9cd

PR #1028, opened 2017-07-25
lib/core: Add #defines for ref/collection binding

cgwalters:
These were previously private, but since we expect people to use them, let's add
`#define`s like we did for some of the other commit metadata.

### DIFF
--- a/src/libostree/ostree-core.h
+++ b/src/libostree/ostree-core.h
@@ -243,7 +243,7 @@ typedef enum {
  * OSTREE_COMMIT_META_KEY_COLLECTION_BINDING:
  *
  * GVariant type `s`.  If this is added to a commit, `ostree_repo_pull()`
- * will enforce that the commit was retrieved from repository which has
+ * will enforce that the commit was retrieved from a repository which has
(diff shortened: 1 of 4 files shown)


cgwalters:
@jlebon Can you review this one?

src/libostree/ostree-core.h
```diff
@@ -226,6 +226,33 @@ typedef enum {
...
+ * and production branches.
+ *
+ * Since: 2017.9
+ */
+#define OSTREE_COMMIT_META_KEY_REF_BINDING "ostree.ref-binding"
+/**
+ * OSTREE_COMMIT_META_KEY_COLLECTION_BINDING:
+ *
+ * GVariant type `s`.  If this is added to a commit, `ostree_repo_pull()`
+ * will enforce that the commit was retrieved from repository which has
```
jlebon: Nit: from **a** repository ?

jlebon:
@rh-atomic-bot r+ 462decd

pwithnall:
Post-hoc review: looks good to me. Thanks for handling this. :)

PR #1030, opened 2017-07-26
ci: Enable libcurl by default on Fedora

cgwalters:
The insttest fell over since its build used libsoup, but that just dropped out
of FAH.


jlebon:
@rh-atomic-bot r+ 691617a

PR #708, opened 2017-02-23
fetcher: Log failures into journal

cgwalters:
Particularly when HTTP requests fail, I really want a lot more information.
We could theoretically stuff it into the `GError` message field, but
that gets ugly *fast*.

Using the systemd journal allows us to log things in a structured fashion.
Right now e.g. rpm-ostree won't be aware of this additional information,
but I think we could teach it to be down the line.

In the short term, users can learn to find it from `systemctl status rpm-ostreed`
or `journalctl -b -r -u rpm-ostreed`, etc.

One thing I'd like to do next is log successful fetches of e.g. commit objects
as well with more information about the originating server (things like the
final URL if we were redirected, did we use TLS pinning, what was the negotiated
TLS version+cipher, etc).

PR #1032, opened 2017-07-27
lib/pull: Journal fetch results

cgwalters:
This is a continuation of addition of journaling to libostree; see
e.g. <https://github.com/ostreedev/ostree/pull/708>.

I wanted more information at the end of fetches; in particular
some details about the delta execution (what opcodes etc.), but
this is a first step: we log things like the transferred data
as well as whether or not GPG was enabled, etc.

One awkward thing about this is how we map the fetcher options like
`tls-ca-path` back out into an enum for the code to log. But eh, hard to fix
without a bigger refactoring.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -50,9 +50,20 @@
 
 #include <gio/gunixinputstream.h>
 #include <sys/statvfs.h>
+#ifdef HAVE_LIBSYSTEMD
+#include <systemd/sd-journal.h>
+#endif
+
+#define OSTREE_MESSAGE_FETCH_COMPLETE_ID SD_ID128_MAKE(75,ba,3d,eb,0a,f0,41,a9,a4,62,72,ff,85,d9,e7,3e)
 
 #define OSTREE_REPO_PULL_CONTENT_PRIORITY  (OSTREE_FETCHER_DEFAULT_PRIORITY)
 #define OSTREE_REPO_PULL_METADATA_PRIORITY (OSTREE_REPO_PULL_CONTENT_PRIORITY - 100)
+
+typedef enum {
+  OSTREE_FETCHER_SECURITY_STATE_CA_PINNED,
+  OSTREE_FETCHER_SECURITY_STATE_TLS,
+  OSTREE_FETCHER_SECURITY_STATE_INSECURE,
+} OstreeFetcherSecurityState;
 
 typedef struct {
   OstreeRepo   *repo;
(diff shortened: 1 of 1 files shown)


cgwalters:
```
Jul 27 19:54:42 localhost.localdomain rpm-ostree[6183]: libostree pull from 'fedora-atomic' for fedora/26/x86_64/updates/atomic-host complete
                                                        security: GPG: yes http: TLS
                                                        non-delta: meta:74 content: 110
                                                        transfer: secs: 42 size: 103.7 MB
```

src/libostree/ostree-repo-pull.c
```diff
@@ -4060,6 +4101,62 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       ostree_async_progress_set_status (pull_data->progress, buf->str);
     }
 
+  if (bytes_transferred > 0)
+    {
+#ifdef HAVE_LIBSYSTEMD
```
jlebon: This will fail to compile on non-systemd platforms, no? Maybe just wrap the whole if-statement instead?

src/libostree/ostree-repo-pull.c
```diff
@@ -3892,6 +3923,16 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
 
   g_hash_table_unref (requested_refs_to_fetch);
   requested_refs_to_fetch = g_steal_pointer (&updated_requested_refs_to_fetch);
+  if (g_hash_table_size (requested_refs_to_fetch) == 1)
+    {
+      GLNX_HASH_TABLE_FOREACH_KV (requested_refs_to_fetch,
```
jlebon: Nit: you can just use `GLNX_HASH_TABLE_FOREACH` here since we don't need the value.

src/libostree/ostree-repo-pull.c
```diff
@@ -4060,6 +4101,62 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
+                }
+            }
+          g_string_append (msg, "\n");
+
+          if (pull_data->n_fetched_deltaparts > 0)
+            g_string_append_printf (msg, "delta: parts: %u loose: %u",
+                                    pull_data->n_fetched_deltaparts,
+                                    pull_data->n_fetched_metadata + pull_data->n_fetched_content);
+          else
+            g_string_append_printf (msg, "non-delta: meta:%u content: %u",
```
jlebon: Nit: missing space after `meta:`.

src/libostree/ostree-repo-pull.c
cgwalters: It'd just be an empty if block, no?  But yeah, it's cleaner to have the conditional enclose all of it.  I also folded in the conditional for `remote_name` and de-indented.

cgwalters: Updated :arrow_down:

jlebon: Sorry, had a brain fart and forgot for a moment that empty blocks are allowed! Though this looks cleaner anyway.

jlebon:
@rh-atomic-bot r+ 668867f

PR #1033, opened 2017-07-28
lib/repo: Add API to create and list ref aliases

cgwalters:
There are multiple use cases where we'd like to alias refs.

First, having a "stable" alias which gets swapped across major
versions: https://pagure.io/atomic-wg/issue/228

Another case is when a ref is obsoleted;
<https://pagure.io/atomic-wg/issue/303>
This second one could be done with endoflife rebase, but I think
this case is better on the server side, as we might later change
our minds and do actual releases there.

I initially just added some test cases for symlinks in the `refs/heads` dir to
ensure this actually works (and it did), but I think it's worth having APIs.

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -784,7 +784,7 @@ relative_symlink_to (const char *relpath,
                      const char *target)
 {
   g_assert (*relpath);
-  g_assert (*target);
+  g_assert (*target && *target != '/');
 
   g_autoptr(GString) buf = g_string_new ("");
 
@@ -797,7 +797,6 @@ relative_symlink_to (const char *relpath,
       g_string_append (buf, "../");
     }
 
-  g_assert (*target != '/');
   g_string_append (buf, target);
 
   return g_string_free (g_steal_pointer (&buf), FALSE);
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -472,6 +472,7 @@ ostree_repo_resolve_rev_ext (OstreeRepo                    *self,
 static gboolean
 enumerate_refs_recurse (OstreeRepo    *repo,
                         const char    *remote,
+                        OstreeRepoListRefsExtFlags flags,
                         const char    *collection_id,
                         int            base_dfd,
                         GString       *base_path,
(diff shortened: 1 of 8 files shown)


cgwalters:
Thought about this some more, and I think we want APIs to read and write alias refs.

cgwalters:
Lifting WIP.

src/ostree/ot-builtin-refs.c
```diff
@@ -183,22 +207,35 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
           else goto out;
         }
 
-      if (checksum_existing != NULL)
+      const gboolean creating_alias = opt_alias && opt_create;
+      if (!creating_alias || !g_hash_table_contains (ref_aliases, opt_create))
```
jlebon: We don't need `creating_alias` here, right? We're already in the `opt_create` branch, so the **if** condition can just be `(!opt_alias || !g_hash_table_contains (...))`, right?

Also, can we add a comment here re. `g_hash_table_contains`? It's a bit tricky to figure out the logic. E.g. even something simple like: `allow aliases to silently replace other aliases, but not regular refs`

src/ostree/ot-builtin-refs.c
```diff
@@ -156,13 +176,17 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
   else if (!ostree_repo_list_refs (repo, refspec_prefix, &refs, cancellable, error))
     goto out;
 
-  if (!opt_delete && !opt_create)
+  if (is_list)
```
jlebon: I guess this is more personal preference, though I find the former expression easier to see it's obviously correct just by inspecting the other `else if` and `else` in the tree.

src/libostree/ostree-repo-refs.c
```diff
@@ -757,17 +779,44 @@ ostree_repo_remote_list_refs (OstreeRepo       *self,
...
+  while (TRUE)
+    {
+      const char *slash = strchr (relpath, '/');
+      if (!slash)
+        break;
+      relpath = slash + 1;
+      g_string_append (buf, "../");
+    }
+
+  g_assert (*target != '/');
```
jlebon: Might as well move this up to the entry point and replace the `g_assert (*target);` there?

src/ostree/ot-builtin-refs.c
cgwalters: It feels messy either way...I kept as is.  Everything else done :arrow_down:

jlebon:
@rh-atomic-bot r+ b592bd3

Issue #608, opened 2016-12-01
signature duplicate checking not working


alexlarsson:
Check this out:

```
$ ostree --repo=repo show app/org.gnome.gedit/x86_64/master
commit 39095f029c8c1ebbcff1676e4daf6714d74dc0b01a8f6c1665223938a402d633
Date:  2016-12-01 13:50:02 +0000

    Export org.gnome.gedit

$ ostree gpg-sign --repo=repo app/org.gnome.gedit/x86_64/master 6A2B067FB5E17A1A3FC8A0DAEB6216DDB76C70E9
$ ostree --repo=repo show app/org.gnome.gedit/x86_64/mastercommit 39095f029c8c1ebbcff1676e4daf6714d74dc0b01a8f6c1665223938a402d633
Date:  2016-12-01 13:50:02 +0000

    Export org.gnome.gedit

Found 1 signature:

  Signature made tor  1 dec 2016 15:59:42 using DSA key ID EB6216DDB76C70E9
  Can't check signature: public key not found
$ ostree gpg-sign --repo=repo app/org.gnome.gedit/x86_64/master 6A2B067FB5E17A1A3FC8A0DAEB6216DDB76C70E9
$ ostree --repo=repo show app/org.gnome.gedit/x86_64/mastercommit 39095f029c8c1ebbcff1676e4daf6714d74dc0b01a8f6c1665223938a402d633
Date:  2016-12-01 13:50:02 +0000

    Export org.gnome.gedit

Found 2 signatures:

  Signature made tor  1 dec 2016 15:59:42 using DSA key ID EB6216DDB76C70E9
  Can't check signature: public key not found

  Signature made tor  1 dec 2016 15:59:49 using DSA key ID EB6216DDB76C70E9
  Can't check signature: public key not found
```
See how it duplicated the signature? However if I use a shorter version of the key id it works:
```
$ ostree gpg-sign --repo=repo app/org.gnome.gedit/x86_64/master EB6216DDB76C70E9
error: Commit is already signed with GPG key EB6216DDB76C70E9
```
It ends up checking for the matching key in `ostree_gpg_verify_result_lookup` where it does `g_str_has_suffix (signature->fpr, key_id_upper)`, but in my case key_id_upper is longer than the fingerprint, so it doesn't match. Is the suffix match supposed to be the other way? Or are you supposed to be able to use an even shorter form of the fingerprint?

ramcq:
[...]

PR #1036, opened 2017-07-30
gpg-verify-result: fix _lookup and _get to deal with subkeys

ramcq:
Both of these functions assumed that signature->fpr would correspond to the normal user-provided / user-visible IDs of GPG keys. In the case of keys with signing subkeys, the signature fingerprint is not the normal one that people use to refer to the key, so use gpgme_get_key to canonicalise to the primary key ID when getting or looking up the GPG keys from signatures.

Has the side-effect of also making _lookup more permissive at what kind of key ID it will accept when looking up - anything that works with GPG will work with _lookup now, rather than inventing our own matching semantics with substrings.

https://github.com/ostreedev/ostree/issues/608


cgwalters:
bot, add author to whitelist

cgwalters:
LGTM, thanks!

@rh-atomic-bot r+ 9f7b251

PR #1037, opened 2017-07-31
Coverity fixes: the good and the ugly

cgwalters:
There are a few real issues coverity found; these are the non-critical ones.


jlebon:
Awesome!

@rh-atomic-bot r+ 455024a

jlebon:
Awesome!

@rh-atomic-bot r+ 455024a

Issue #1035, opened 2017-07-29
Assert in curl callback makes gnome-software crash on rawhide


hughsie:
We're debugging Andreas Nilsson's rawhide laptop here and he's running into a gnome-software crash in ostree-fetcher-curl.c in event_cb. Looks like there's an assert in curl networking code callback that curl returns OK; this is clearly not the case here.

`rc = curl_multi_socket_action ()`
`g_assert (rc = CURLM_OK)`

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1038

cgwalters:
Was this debugged to any root cause?  e.g. incorrect proxy settings, connection being interrupted, DNS resolution failure, etc?

kalev:
No, we didn't debug it further, but a number of ports are blocked here on the network. Could that be the cause?

Thanks for the quick fix!

PR #1038, opened 2017-07-31
lib/curl: Only check individual request errors

cgwalters:
It looks like `curl_multi_socket_action()` will return an error
if *one* of the requests has an error, but we already check
for that explicitly by iterating over each handle.

In libcurl, the "easy" layer doesn't really make use of this
return value.  I did a bit of looking elsewhere; systemd
does check it as a runtime error, not an assertion.  librepo
doesn't use the multi interface.

Closes: https://github.com/ostreedev/ostree/issues/1035


jlebon:
@rh-atomic-bot r+ 604ff83

PR #1039, opened 2017-07-31
lib: Define and use an autoptr cleanup for gpgme_key_t

cgwalters:
Followup for previous patch, allows porting a bit to new code style.


src/libostree/ostree-gpg-verify-result.c
```diff
@@ -274,18 +273,13 @@ ostree_gpg_verify_result_lookup (OstreeGpgVerifyResult *result,
         {
           if (out_signature_index != NULL)
             *out_signature_index = signature_index;
-          ret = TRUE;
+          /* Not early return */
```
jlebon: `s/not/note/`

jlebon:
@rh-atomic-bot r+ aebd549

PR #1041, opened 2017-07-31
Coverity fixes: the bad

cgwalters:


src/ostree/ot-builtin-pull.c
```diff
@@ -212,7 +212,7 @@ ostree_builtin_pull (int argc, char **argv, GCancellable *cancellable, GError **
                     override_commit_ids = g_ptr_array_new_with_free_func (g_free);
 
                   /* Backfill */
-                  for (j = 2; j < i; i++)
+                  for (j = 2; j < i; j++)
```
jlebon: Ouch, that's a pretty bad one. Let's add a small test case for this as well in `pull-test.sh`? E.g. basically something like this should trigger it, right?

```
${CMD_PREFIX} ostree --repo=parentpullrepo pull origin main@${parent_rev} main@${rev}
```

cgwalters: Oh yeah, it was broken.  Needed more work, done in :arrow_down:

pwithnall:
The test case looks good to me.

cgwalters:
@rh-atomic-bot r=pwithnall 5c90602

PR #1044, opened 2017-08-01
 lib/deploy: Add structured logging info for syncfs() times

cgwalters:
I plan to at some point change rpm-ostree to read the journal messages from
libostree and render things like the time we spent in syncfs().


jlebon:
@rh-atomic-bot r+ e9d80ee

> I plan to at some point change rpm-ostree to read the journal messages from
libostree and render things like the time we spent in syncfs().

Not opposed, though this sounds error-prone. Wouldn't it be less painful to extend the API? Are there other projects that do something like this? (Fish out info about the underlying library from the journal).

PR #1046, opened 2017-08-02
lib/pull: Log state of summary/commit GPG verification

cgwalters:
Since we have both, we should clearly log the state of both of
them.  Split this out of a larger patch.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -4111,7 +4111,17 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
         g_string_append_printf (msg, "libostree pull from '%s' for %u refs complete",
                                 pull_data->remote_name, g_hash_table_size (requested_refs_to_fetch));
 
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -4111,7 +4111,17 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
                                 pull_data->remote_name, g_hash_table_size (requested_refs_to_fetch));
 
-      g_string_append_printf (msg, "\nsecurity: GPG: %s ", pull_data->gpg_verify ? "yes" : "no");
+      const char *gpg_verify_state;
+      if (pull_data->gpg_verify_summary)
+        {
+          if (pull_data->gpg_verify)
+            gpg_verify_state = "summary+commit";
+          else
+            gpg_verify_state = "summary-only";
```
jlebon: Nit: can we call this `summary`, to be consistent with `commit` and `summary+commit`?

jlebon: Ahh, gotcha. Fair enough.

jlebon:
@rh-atomic-bot r+ 7d5e7c7

PR #1047, opened 2017-08-02
tests/libtest-core.sh: Add a comment that this copy is canonical

cgwalters:
Ref: https://github.com/projectatomic/bubblewrap/pull/203


jlebon:
@rh-atomic-bot r+ 88caf5f

Issue #876, opened 2017-05-19
boot: Tracker for single partition journaling+GRUB issue


cgwalters:
See: http://marc.info/?l=linux-fsdevel&m=149520244919284&w=2

TL;DR: XFS doesn't flush the journal on `syncfs()` (which is what ostree uses).  GRUB doesn't know how to follow the XFS journal, so if the filesystem is in a dirty state (possible with xfs `/boot`, extremely likely with `/`, if the journaled data includes content for `/boot`, the system may be unbootable if a system crash occurs.

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1049

PR #1049, opened 2017-08-03
lib/deploy: Use a FIFREEZE/FITHAW cycle for /boot

cgwalters:
See: http://marc.info/?l=linux-fsdevel&m=149520244919284&w=2

XFS doesn't flush the journal on `syncfs()`. GRUB doesn't know how to follow the
XFS journal, so if the filesystem is in a dirty state (possible with xfs
`/boot`, extremely likely with `/`, if the journaled data includes content for
`/boot`, the system may be unbootable if a system crash occurs.

Fix this by doing a `FIFREEZE`+`FITHAW` cycle.  Now, most people
probably would have replaced the `syncfs()` invocation with those two
ioctls.  But this would have become (I believe) the *only* place in
libostree where we weren't safe against interruption.  The failure
mode would be ugly; nothing else would be able to write to the filesystem
until manual intervention.

The real fix here I think is to land an atomic `FIFREEZETHAW` ioctl
in the kernel.  I might try a patch.

In the meantime though, let's jump through some hoops and set up
a "watchdog" child process that acts as a fallback unfreezer.

Closes: https://github.com/ostreedev/ostree/issues/876

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1045,7 +1045,7 @@ fsfreeze_thaw_cycle (OstreeSysroot *self,
         {
           if (TEMP_FAILURE_RETRY (ioctl (rootfs_dfd, FITHAW, 0)) != 0)
             {
-              if (errno == EPERM && getuid () != 0)
+              if (errno == EPERM)
                 ; /* Ignore this for the test suite */
               else
                 err (1, "FITHAW");
@@ -1081,7 +1081,7 @@ fsfreeze_thaw_cycle (OstreeSysroot *self,
           /* Not supported, or we're running in the unit tests (as non-root)?
            * OK, let's just do a syncfs.
            */
-          if (errno == EOPNOTSUPP || (errno == EPERM && getuid () != 0))
+          if (G_IN_SET (errno, EOPNOTSUPP, EPERM))
             {
               if (TEMP_FAILURE_RETRY (syncfs (rootfs_dfd)) != 0)
                 return glnx_throw_errno_prefix (error, "syncfs");
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -22,8 +22,14 @@
 
 #include <gio/gunixinputstream.h>
 #include <gio/gunixoutputstream.h>
+#include <glib-unix.h>
 #include <sys/mount.h>
 #include <sys/statvfs.h>
+#include <sys/socket.h>
+#include <sys/ioctl.h>
+#include <sys/poll.h>
+#include <linux/fs.h>
+#include <err.h>
 
 #ifdef HAVE_LIBMOUNT
 #include <libmount.h>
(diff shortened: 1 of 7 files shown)


vtolstov:
@cgwalters thanks, sorry i don't understand about 
```
The failure
mode would be ugly; nothing else would be able to write to the filesystem
until manual intervention.
```

as i know, that kernel by default supports automatic unfreeze via:

```
 A value of 0 for the argument also means there is no timeout, but any other value is treated as a pointer to a timeout value in seconds
```

cgwalters:
You're quoting <https://lwn.net/Articles/287435/>?  As best I can tell, that timeout feature never landed.  See https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=fcccf502540e3d752d33b2d8e976034dee81f9f7

vtolstov:
@cgwalters thanks =). i'm use this feature is more then 3 years and think that it worked =)).
I'm lucky that i need to unfreeze by hand only two times =)

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -973,17 +978,88 @@ checksum_from_kernel_src (const char   *name,
 }
 
 static gboolean
-syncfs_dir_at (int            dfd,
-               const char    *path,
-               GCancellable  *cancellable,
-               GError       **error)
+fsfreeze_thaw_cycle (int            rootfs_dfd,
```
jlebon: Can we have a small comment here with some of the rationale from the commit msg? E.g. why we're doing `FIFREEZE/FITHAW` instead of just `syncfs`.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -973,17 +978,88 @@ checksum_from_kernel_src (const char   *name,
...
+                return glnx_throw_errno_prefix (error, "syncfs");
+              /* Write the completion, and return */
+              if (TEMP_FAILURE_RETRY (write (pipe_write, &c, sizeof (c))) < 0)
+                return glnx_throw_errno_prefix (error, "write(watchdog syncfs complete)");
+              return TRUE;
+            }
+          else
+            return glnx_throw_errno_prefix (error, "ioctl(FIFREEZE)");
+        }
+      if (ioctl (rootfs_dfd, FITHAW, 0) != 0)
```
jlebon: Maybe a shortcircuit here based on some env var so we can add an insttest that the watchdog kicks in as predicted?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -973,17 +978,88 @@ checksum_from_kernel_src (const char   *name,
...
+      int sigs[] = { SIGINT, SIGTERM };
+      for (guint i = 0; i < G_N_ELEMENTS (sigs); i++)
+        {
+          if (signal (sigs[i], SIG_IGN) == SIG_ERR)
+            err (1, "signal");
+        }
+      if (TEMP_FAILURE_RETRY (write (pipe_write, &c, sizeof (c))) < 0)
+        err (1, "write");
+      /* Wait for the parent to say it's going to freeze. */
+      (void) TEMP_FAILURE_RETRY (read (pipefds[0], &c, sizeof (c)));
```
jlebon: Wait, this is referring to the same pipe, though right? Won't we just pick up the data we just wrote up above? If we want bidirectional comms, I think we need two pipes, right? (So 4 fds total).

cgwalters: Yes, oops :smile:  Fixed this by using a `socketpair()`.

jlebon:
@rh-atomic-bot r+ 84f064e

cmurf:
While the original bug that brought this up only reproduces on XFS, the problem could happen on ext4 or Btrfs. So they all probably should get freeze/thaw after bootloader related things change.

PR #1050, opened 2017-08-03
Documentation: README: Remove deprecated wiki link

guyshapiro:
The old wiki only contains link back to readthedocs.
The link to is useless.


cgwalters:
@rh-atomic-bot r+ 4d42296

PR #1040, opened 2017-07-31
tree-wide: Use g_autoptr(Ostree*)

cgwalters:
Part of cleaning up our usage of libglnx; we want to use what's in GLib where we
can.

Had to change a few .c files to `#include ostree.h` early on to pick up
autoptrs for the core types.

PR #1051, opened 2017-08-03
tree-wide: Fix the build with old glib (Ubuntu Trusty etc.)

cgwalters:
This regressed with <https://github.com/ostreedev/ostree/pull/1040>
but currently the Travis builds aren't gating.


jlebon:
@rh-atomic-bot r+ b89f3f1

PR #1052, opened 2017-08-04
lib/sysroot: Add journal-msg signal

cgwalters:
This will allow us to drop the awful hack in rpm-ostree where we watch our own
stdout. In general, libraries shouldn't write to stdout.

Also we can kill the systemd journal wrapper code. There's some duplication at
each call site now...but it's easier than trying to write a `sd_journal_send()`
wrapper.

I was originally going to have this emit all of the structured data too as a
`GVariant` but decided it wasn't worth it right now.


jlebon:
OK, that works. We'll need to make sure to highlight this change in the release notes since it does modify the UX of ostree-for-host apps.

@rh-atomic-bot r+ 08dea2a

Was curious and ran:

```
$ git grep g_print src/libostree | wc -l
32
```

Though the large majority is from the static-delta generation & display code.

cgwalters:
Fell victim to https://github.com/ostreedev/ostree/pull/1070

@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1053, opened 2017-08-04
 lib/sysroot-deploy: Refactor kernel layout parsing

cgwalters:
I'd like to move the new canonical kernel directory to `/usr/lib/modules/$kver`,
as Fedora has done. The `get_kernel_from_tree()` function now abstracts over
parsing the data (src vs destination filenames, as well as checksum) in
preparation for adding the new case.

In preparation for this, let's change the current test suite to use the
*current* directory of `/usr/lib/ostree-boot`, and also add coverage of `/boot`.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -366,12 +366,7 @@ setup_os_repository () {
     shift
     bootmode=$1
     shift
-    if test -n "${1:-}"; then
-        bootdir=${1}
-        shift
-    else
-        bootdir=usr/lib/ostree-boot
-    fi
+    bootdir=${1:-usr/lib/ostree-boot}
 
     oldpwd=`pwd`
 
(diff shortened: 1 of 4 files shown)


tests/libtest.sh
```diff
@@ -363,8 +363,15 @@ setup_os_boot_grub2() {
...
-    bootmode=$2
     shift
+    bootmode=$1
+    shift
+    if test -n "${1:-}"; then
+        bootdir=${1}
+        shift
+    else
+        bootdir=usr/lib/ostree-boot
+    fi
```
cgwalters: I think I didn't do that since we'd lose the `shift`, but eh, we can deal with that if we start adding more arguments.

tests/test-admin-deploy-syslinux.sh
```diff
@@ -19,11 +19,31 @@
 
 set -euo pipefail
 
-echo "1..18"
+echo "1..19"
```
jlebon: This needs to be bumped to 20 since the FIFREEZE change. I'm guessing GitHub doesn't complain about it because it's the same hunk change.

cgwalters: Yep thanks, done.

jlebon:
@rh-atomic-bot r+ 503f556

PR #1054, opened 2017-08-04
main: Fix subcommand usage output

mwleeds:
This commit sets prgname correctly so that the "ostree subcommand
--help" output prints the subcommand rather than just "ostree".

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -59,6 +59,15 @@ else
 fi
 AM_CONDITIONAL(BUILDOPT_ASAN, [test x$using_asan = xyes])
 
+AC_MSG_CHECKING([for -fsanitize=thread in CFLAGS])
+if echo $CFLAGS | grep -q -e -fsanitize=thread; then
+  AC_MSG_RESULT([yes])
+  using_tsan=yes
+else
+  AC_MSG_RESULT([no])
+fi
+AM_CONDITIONAL(BUILDOPT_TSAN, [test x$using_tsan = xyes])
+
 # Initialize libtool
 LT_PREREQ([2.2.4])
 LT_INIT([disable-static])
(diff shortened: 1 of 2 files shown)


cgwalters:
See https://github.com/ostreedev/ostree/commit/f0519e541f296fc277db636567bb0f0e4d9fc62d

cgwalters:
bot, add author to whitelist

cgwalters:
Now, honestly my attempts at using TSAN stalled out a bit; I submitted https://bugzilla.gnome.org/show_bug.cgi?id=774645 but no one reviewed.

I dunno, we could make this bit conditional, like detect `-fsanitize=thread` in CFLAGS like we do with `-fsanitize=address` and do e.g.:

```
#ifndef BUILDOPT_TSAN
  g_set_prgname (...)
#endif
```
?

mwleeds:
Conditional compilation seems reasonable to me. I suppose another way would be to allow prgname to be overriden in `g_option_context_get_help` or another `GOptionContext` API.

mwleeds:
Pushed a fixed commit. Interesting that the use of `g_set_prgname` in subcommands didn't trip the sanitizer, or perhaps they were just missed.

configure.ac
```diff
@@ -59,6 +59,15 @@ else
...
 AM_CONDITIONAL(BUILDOPT_ASAN, [test x$using_asan = xyes])
 
+AC_MSG_CHECKING([for -fsanitize=thread in CFLAGS])
+if echo $CFLAGS | grep -q -e -fsanitize=thread; then
+  AC_MSG_RESULT([yes])
+  using_tsan=yes
+else
+  AC_MSG_RESULT([no])
+fi
+AM_CONDITIONAL(BUILDOPT_TSAN, [test x$using_tsan = xyes])
```
cgwalters: Almost; since we're doing the preprocessor here we need an `AM_COND_IF()` + `AC_DEFINE()`.

cgwalters:
Pushed a fixup.  Thanks for the patch!

@rh-atomic-bot r+ ec68a65

Issue #687, opened 2017-02-16
verify commit timestamp during pull


cgwalters:
Migrating from https://github.com/projectatomic/rpm-ostree/issues/623

If we aren't going to end up doing the upgrade because the remote latest commit is actually "older" than the local commit then let's do the check before we download any content objects. 

In the example below I downloaded a couple hundred mebibytes of content before we errored out. 

```
[root@localhost ~]# rpm-ostree upgrade 
Updating from: fedora-atomic:fedora-atomic/25/x86_64/docker-host

761 metadata, 3096 content objects fetched; 198587 KiB transferred in 63 seconds
error: Upgrade target revision '135014362783f962281290e02ea2f4ef74211aaa036f98a072ee10e2a5b3be58' with timestamp 'Tue Feb  7 00:10:30 2017' is chronologically older than current revision '8437a55b2bf20613797ed779a591f06b2ef92467b1a00c2ef519c0a3d8c9d623' with timestamp 'Sun Feb 12 19:26:55 2017'; use --allow-downgrade to permit
```

dustymabe:
hit this again today. downloaded `161309 KiB` just so I could get notified that I didn't want to upgrade :) 

I know this is not something a normal user would hit, but would be nice to cut down on "wasted data transfer" that I hit from time to time.

```
[root@f25a-DO ~]# rpm-ostree upgrade 

780 metadata, 3397 content objects fetched; 161309 KiB transferred in 101 seconds                                                                                                                                                           
error: Upgrade target revision '6a71adb06bc296c19839e951c38dc0b71ee5d7a82262fef9612f256f0c2a70da' with timestamp 'Sun 26 Mar 2017 09:04:50 PM UTC' is chronologically older than current revision 'bf3e9caf29e8980edd2548d8a7e63c43ca02de2588c07a07e69e75809090f60f' with timestamp 'Thu 30 Mar 2017 12:52:25 AM UTC'; use --allow-downgrade to permit
```

dustymabe:
I'm not seeing any improvement from the original issue:

```
[root@cloudhost ~]# rpm-ostree status
State: idle
Deployments:
● fedora-atomic:fedora/27/x86_64/atomic-host
                   Version: 27.14 (2017-11-21 21:32:48)
                    Commit: 88bf5b2187343da43ee1c689085e5636fca64c27b2f0facc76849e91b7faad52
              GPGSignature: Valid signature by 860E19B0AFA800A1751881A6F55E7430F5282EE4
[...]

PR #1055, opened 2017-08-05
lib/pull: Add support for timestamp-check option, use in upgrader

cgwalters:
For both flatpak and ostree-as-host, we really want to verify up front during
pulls that we're not being downgraded. Currently both flatpak and
`OstreeSysrootUpgrader` do this before deployments, but at that point we've
already downloaded all the data, which is annoying.

Closes: https://github.com/ostreedev/ostree/issues/687

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -2086,6 +2086,38 @@ ostree_commit_get_timestamp (GVariant  *commit_variant)
   return GUINT64_FROM_BE (ret);
 }
 
+/* Used in pull/deploy to validate we're not being downgraded */
+gboolean
+_ostree_compare_timestamps (const char   *current_rev,
+                            guint64       current_ts,
+                            const char   *new_rev,
+                            guint64       new_ts,
+                            GError      **error)
+{
+  /* Newer timestamp is OK */
+  if (new_ts > current_ts)
+    return TRUE;
+  /* If they're equal, ensure they're the same rev */
+  if (new_ts == current_ts || strcmp (current_rev, new_rev) == 0)
+    return TRUE;
+
+  /* Looks like a downgrade, format an error message */
+  g_autoptr(GDateTime) current_dt = g_date_time_new_from_unix_utc (current_ts);
+  g_autoptr(GDateTime) new_dt = g_date_time_new_from_unix_utc (new_ts);
+
+  if (current_dt == NULL || new_dt == NULL)
+    return glnx_throw (error, "Upgrade target revision '%s' timestamp (%" G_GINT64_FORMAT ") or current revision '%s' timestamp (%" G_GINT64_FORMAT ") is invalid",
+                       new_rev, new_ts,
+                       current_rev, current_ts);
+
+  g_autofree char *current_ts_str = g_date_time_format (current_dt, "%c");
+  g_autofree char *new_ts_str = g_date_time_format (new_dt, "%c");
+
+  return glnx_throw (error, "Upgrade target revision '%s' with timestamp '%s' is chronologically older than current revision '%s' with timestamp '%s'",
+                     new_rev, new_ts_str, current_rev, current_ts_str);
+}
+
+
 GVariant *
 _ostree_detached_metadata_append_gpg_sig (GVariant   *existing_metadata,
                                           GBytes     *signature_bytes)
--- a/tests/pull-test.sh
+++ b/tests/pull-test.sh
@@ -35,7 +35,7 @@ function verify_initial_contents() {
     assert_file_has_content baz/cow '^moo$'
 }
 
-echo "1..28"
+echo "1..29"
 
 # Try both syntaxes
 repo_init --no-gpg-verify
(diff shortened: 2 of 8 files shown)


tests/pull-test.sh
```diff
@@ -35,7 +35,7 @@ function verify_initial_contents() {
     assert_file_has_content baz/cow '^moo$'
 }
 
-echo "1..27"
+echo "1..28"
```
jlebon: Can we also have an `ostree admin upgrade` test in `admin-test.sh` to validate the sysroot upgrader changes as well? (E.g. just checking that file objects were not fetched?).

src/libostree/ostree-core.c
```diff
@@ -2086,6 +2086,38 @@ ostree_commit_get_timestamp (GVariant  *commit_variant)
...
+
+  if (current_dt == NULL || new_dt == NULL)
+    return glnx_throw (error, "Upgrade target revision '%s' timestamp (%" G_GINT64_FORMAT ") or current revision '%s' timestamp (%" G_GINT64_FORMAT ") is invalid",
+                       new_rev, new_ts,
+                       current_rev, current_ts);
+
+  g_autofree char *current_ts_str = g_date_time_format (current_dt, "%c");
+  g_autofree char *new_ts_str = g_date_time_format (new_dt, "%c");
+
+  return glnx_throw (error, "Upgrade target revision '%s' with timestamp '%s' is chronologically older than current revision '%s' with timestamp '%s'; use --allow-downgrade to permit",
```
jlebon: Maybe drop the `use --allow-downgrade to permit`, since we haven't necessarily gotten here through `ostree admin upgrade`, right?

Hmm, would be nice to have it appended in the sysroot upgrader instead, though we'd need to create a new error domain, it seems? (I don't see any error codes appropriate in the `G_IO_ERROR` domain at least, though maybe there's another glib domain that has something).

tests/pull-test.sh
```diff
@@ -205,6 +205,30 @@ ${CMD_PREFIX} ostree --repo=parentpullrepo rev-parse origin:main > main.txt
...
+assert_not_streq "${origrev}" "${newrev}"
+assert_streq ${newrev} "$(${CMD_PREFIX} ostree --repo=repo rev-parse main)"
+newrev2=$(${CMD_PREFIX} ostree --timestamp="October 25 1985" --repo=ostree-srv/gnomerepo commit -b main --tree=ref=main)
+${CMD_PREFIX} ostree --repo=ostree-srv/gnomerepo summary -u
+if ${CMD_PREFIX} ostree --repo=repo pull -T origin main 2>err.txt; then
+    fatal "pulled older commit with timestamp checking enabled?"
+fi
+assert_file_has_content err.txt "Upgrade.*is chronologically older"
+assert_streq ${newrev} "$(${CMD_PREFIX} ostree --repo=repo rev-parse main)"
+# But we can pull it without timestamp checking
```
jlebon: Seems like we're missing a final `ostree pull` here without `-T` ?

src/libostree/ostree-core.c
jlebon: > Hmm, would be nice to have it appended in the sysroot upgrader instead, though we'd need to create a new error domain, it seems?

By chance, it sounds like exactly that is being discussed in https://github.com/ostreedev/ostree/issues/1064.

pwithnall: I was planning on creating a special domain just for GPG-related errors; you’d probably want to create another separate one for this PR>

cgwalters:
OK, rebased :surfing_man: and reworked for all comments except the error domain one; not sure I want to dive into an error domain for pulls right now.  Dropped the text in the error message referring to a command line option.

jlebon:
I added a fixup to simplify something in the test, which was most likely what you meant.

@rh-atomic-bot r+ 1df0034

As an experiment, I tried out using GitHub directly, which worked but it was as painful, if not more so, than dropping down the the CLI. You have to get to the file, do your edit, then Ctrl-C Ctrl-V the commit message prepended by `fixup! `.

tests/test-admin-upgrade-not-backwards.sh
```diff
@@ -61,7 +61,7 @@ assert_file_has_content upgrade-err.txt 'chronologically older'
 currev=$(ostree --repo=sysroot/ostree/repo rev-parse testos:${ref})
 assert_not_streq ${newrev} ${currev}
 assert_streq ${origrev} ${currev}
-assert_not_has_file sysroot/ostree/repo/$(ostree_checksum_to_relative_object_path sysroot/ostree/repo ${tscheck_checksum})
+assert_not_has_file sysroot/ostree/repo/$tscheck_fileobjpath
```
cgwalters: Hmm actually no, the repos have different modes - the first one is `.filez`, the second is `.file`.  That's why I explicitly added the code to do path to checksum and used it twice.  Right?

jlebon: Ahh, sorry! Yes, that makes sense.
https://github.com/ostreedev/ostree/pull/1098

PR #1057, opened 2017-08-06
Documentation: static delta default from

guyshapiro:
Document that the default behavior of `ostree static-delta generate` if to generate the delta from the parent.

Note: I didn't try to compile after this change.
It was easier to suggest an edit than to create a bug report describing the issue.


cgwalters:
@rh-atomic-bot r+ 7bb8c56

cgwalters:
@rh-atomic-bot retry

PR #1058, opened 2017-08-07
lib/repo-pull: Fix counting of latest commits when finding repos

pwithnall:
The intended behaviour of ostree_repo_find_remotes() is to return
results which have the latest version of at least one of the requested
refs. Results which have some of the requested refs, but don’t have the
latest version of any of them, should be ignored. The logic to do this
was broken in the case that a result contained a positive number of the
requested refs, but none of them were the latest version. (It previously
worked when the result contained none of the requested refs.)

Fix the counting to work correctly in both cases.

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -779,6 +779,126 @@ ostree_repo_remote_list_refs (OstreeRepo       *self,
   return TRUE;
 }
 
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+static gboolean
+remote_list_collection_refs_process_refs (OstreeRepo   *self,
+                                          const gchar  *remote_name,
+                                          const gchar  *summary_collection_id,
+                                          GVariant     *summary_refs,
+                                          GHashTable   *ret_all_refs,
+                                          GError      **error)
+{
+  gsize j, n;
+
+  for (j = 0, n = g_variant_n_children (summary_refs); j < n; j++)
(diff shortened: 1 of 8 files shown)


cgwalters:
> FAIL: libostree/test-find-remotes.sh.test (Child process exited with code 133)

cgwalters:
The code seems reasonable to me, not sure what's wrong with the test offhand.

pwithnall:
> The code seems reasonable to me, not sure what's wrong with the test offhand.

I’m taking a look and it’s uncovering a few more problems. PR incoming in an hour or two.

src/libostree/ostree-repo-refs.c
```diff
@@ -779,6 +779,126 @@ ostree_repo_remote_list_refs (OstreeRepo       *self,
...
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+static gboolean
+remote_list_collection_refs_process_refs (OstreeRepo   *self,
+                                          const gchar  *remote_name,
+                                          const gchar  *summary_collection_id,
+                                          GVariant     *summary_refs,
+                                          GHashTable   *ret_all_refs,
+                                          GError      **error)
+{
+  gsize j, n;
```
cgwalters: I'm OK with this as is, but a lot of the libostree codebase continues to move to decl+init inline, so it'll probably be converted at some point.

cgwalters:
@rh-atomic-bot r+ 2215d00

PR #1056, opened 2017-08-05
tree-wide: Remove trailing semicolon from autoptr declarations

cgwalters:
It confuses `g-ir-scanner`, and isn't necessary.

PR #1060, opened 2017-08-07
libotutil: Drop trailing semicolons from GObject definitions

pwithnall:
g-ir-scanner now emits a warning about them. It’s generally accepted
convention to *not* have a trailing semicolon after G_DEFINE_*() macro
invocations.


cgwalters:
Dup of https://github.com/ostreedev/ostree/pull/1056

PR #1062, opened 2017-08-08
man: The min-free-space-percent item goes in [core] section

stefwalter:
The documentation incorrectly indicates that ```min-free-space-percent```
goes in the ```[remote "name"]``` section. It should go in ```[core]``` instead.


stefwalter:
FYI we had to use ```min-free-space-percent=0``` to make Cockpit CI tests work work: https://github.com/cockpit-project/cockpit/pull/7444

cgwalters:
@rh-atomic-bot r+ 81d26ea

PR #1063, opened 2017-08-08
autogen.sh: Fix running out of tree

stefwalter:
The autogen.sh script should be runnable out of tree. It's mostly
already the case, just one little tweak to make it work.

    $ mkdir build
    $ cd build
    $ ../autogen.sh --prefix=/usr

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -28,7 +28,6 @@ else
         gtkdocize
 fi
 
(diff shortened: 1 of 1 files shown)


cgwalters:
@rh-atomic-bot r+ 52a4a70

Issue #1061, opened 2017-08-07
Invalid assumption leads to assertion failure


mwleeds:
If fetching a summary file 404's, this leads to an assertion failure. This is because `ostree_repo_remote_fetch_summary_with_options` returns `TRUE` without putting anything in `out_summary`, and then that NULL is passed to `g_variant_new_from_bytes` [here](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-pull.c#L4839).

mwleeds:
It seems to be caused by clearing the `G_IO_ERROR_NOT_FOUND` [here](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-fetcher-util.c#L91). I tried looking at the git history but couldn't find why that behavior is there.

pwithnall:
Ah, that behaviour is documented in the documentation for `ostree_repo_remote_fetch_summary()`:
```
 * If no summary file exists on the remote server, @out_summary is set to
 * @NULL.  Likewise if the summary file is not signed, @out_signatures is
 * set to @NULL.  In either case the function still returns %TRUE.
```

The same expected behaviour is true of `fetch_summary_with_options()`, so the bug here is that the caller of that function is not correctly handling a `NULL` returned `summary`.

PR #1065, opened 2017-08-08
lib/repo: Fix handling of missing summary files when downloading

pwithnall:
The API for downloading a summary file can legitimately return NULL for
the summary file contents when it returns TRUE (success). This indicates
an error 404 — the summary file was not found.

Two call sites were not handling that correctly, which was causing later
assertion failures.

https://github.com/ostreedev/ostree/issues/1061


cgwalters:
Should we do `Closes: #1061`?

pwithnall:
> Should we do `Closes: #1061`?

Ah, those are the runes. Yes, will do.

cgwalters:
@rh-atomic-bot r+ 11a6205

PR #1066, opened 2017-08-08
lib/repo-finder: Emit gpg-verify-summary=false in dynamic remote config

pwithnall:
When returning results from finding repos, set gpg-verify-summary=false
in their configs, since any pulls from such remotes will necessarily
involve collection IDs, and hence should be using the unsigned summary
support. In the intended deployment mode for P2P transmission of OSTree
refs, summaries *cannot* be signed, so setting gpg-verify-summary=true
would cause all the pulls to fail.

The unsigned summary support is the move of repository metadata from
the summary file (not spliceable) to the well-known ostree-metadata ref
(spliceable, as it can exist for multiple collection IDs in the same
repository).

### DIFF
--- a/src/libostree/ostree-repo-finder-avahi.c
+++ b/src/libostree/ostree-repo-finder-avahi.c
@@ -824,9 +824,10 @@ ostree_avahi_service_build_repo_finder_result (OstreeAvahiService               
       g_clear_pointer (&remote->keyring, g_free);
       remote->keyring = g_strdup (repo->keyring);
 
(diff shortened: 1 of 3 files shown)


pwithnall:
This is probably going to cause test failures.

cgwalters:
@rh-atomic-bot r+ 1c08e6d

pwithnall:
> This is probably going to cause test failures.

Apparently not. I’ve been getting a lot of spurious failures recently due to the httpd processes in the unit tests not running properly. Maybe it’s time I restarted my machine.

PR #1069, opened 2017-08-09
Improve API for listing refs

mwleeds:
Based on the manpage and the commit message on b7b79fa78, the expected
behavior for "ostree refs -c" is to list all the refs that have
collection IDs, but instead only the local and mirrored refs are being
shown. This commit makes the refs command also print refs from
configured remotes, as it does without the -c option.

### DIFF
--- a/src/libostree/ostree-repo-prune.c
+++ b/src/libostree/ostree-repo-prune.c
@@ -335,7 +335,7 @@ ostree_repo_prune (OstreeRepo        *self,
       g_autoptr(GHashTable) all_collection_refs = NULL;  /* (element-type OstreeChecksumRef utf8) */
 
       if (!ostree_repo_list_collection_refs (self, NULL, &all_collection_refs,
-                                             cancellable, error))
+                                             OSTREE_REPO_LIST_REFS_EXT_EXCLUDE_REMOTES, cancellable, error))
         return FALSE;
 
       GLNX_HASH_TABLE_FOREACH_V (all_collection_refs, const char*, checksum)
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -560,7 +560,7 @@ _ostree_repo_list_refs_internal (OstreeRepo       *self,
       if (!ostree_parse_refspec (refspec_prefix, &remote, &ref_prefix, error))
         return FALSE;
 
-      if (remote)
+      if (!(flags & OSTREE_REPO_LIST_REFS_EXT_EXCLUDE_REMOTES) && remote)
         {
           prefix_path = glnx_strjoina ("refs/remotes/", remote, "/");
           path = glnx_strjoina (prefix_path, ref_prefix);
@@ -620,32 +620,35 @@ _ostree_repo_list_refs_internal (OstreeRepo       *self,
                                    ret_all_refs, cancellable, error))
         return FALSE;
 
-      g_string_truncate (base_path, 0);
-
-      if (!glnx_dirfd_iterator_init_at (self->repo_dir_fd, "refs/remotes", TRUE, &dfd_iter, error))
-        return FALSE;
-
-      while (TRUE)
+      if (!(flags & OSTREE_REPO_LIST_REFS_EXT_EXCLUDE_REMOTES))
         {
-          struct dirent *dent;
-          glnx_fd_close int remote_dfd = -1;
+          g_string_truncate (base_path, 0);
 
-          if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent, cancellable, error))
-            return FALSE;
-          if (!dent)
-            break;
-
-          if (dent->d_type != DT_DIR)
-            continue;
-
-          if (!glnx_opendirat (dfd_iter.fd, dent->d_name, TRUE, &remote_dfd, error))
+          if (!glnx_dirfd_iterator_init_at (self->repo_dir_fd, "refs/remotes", TRUE, &dfd_iter, error))
             return FALSE;
 
-          if (!enumerate_refs_recurse (self, dent->d_name, flags, NULL, remote_dfd, base_path,
-                                       remote_dfd, ".",
-                                       ret_all_refs,
-                                       cancellable, error))
-            return FALSE;
+          while (TRUE)
+            {
+              struct dirent *dent;
+              glnx_fd_close int remote_dfd = -1;
+
+              if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent, cancellable, error))
+                return FALSE;
+              if (!dent)
+                break;
+
+              if (dent->d_type != DT_DIR)
+                continue;
+
+              if (!glnx_opendirat (dfd_iter.fd, dent->d_name, TRUE, &remote_dfd, error))
+                return FALSE;
+
+              if (!enumerate_refs_recurse (self, dent->d_name, flags, NULL, remote_dfd, base_path,
+                                           remote_dfd, ".",
+                                           ret_all_refs,
+                                           cancellable, error))
+                return FALSE;
+            }
         }
     }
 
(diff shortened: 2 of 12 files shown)


mwleeds:
This is based on how eos-updater uses the `ostree_repo_list_collection_refs` API, so I'm pretty sure it's right, but we should get @pwithnall's opinion before merging.

pwithnall:
Hmm. I did deliberately omit support for `refs/remotes` from `ostree_repo_list_collection_refs()`, as it’s supposed to only list local refs (mirrors count as local). However, looking back at this, there is little reason to limit it to local refs, and this does make its semantics differ from `ostree_repo_list_refs()`.

If this change is made, `ostree_repo_regenerate_summary()` will change behaviour to include remote refs (which have a collection ID) in generated summary files. That should be OK, and matches what the existing non-collection--supporting code does (it calls `ostree_repo_list_refs()`).

This change introduces the possibility of duplicates in the results: I might have `refs/mirrors/CID/ref1` and `refs/remotes/remoteA/ref1`, and have `collection-id=CID` in my remote config for `remoteA`. What do we do in that situation?

@cgwalters, just to check, how would you prefer the `ostree_repo_list_*_refs()` APIs to work? Consistently return remote refs, or consistently not return them, or only return them if a match key is specified?

pwithnall:
I’m also a bit uncomfortable with the idea of checking the remote config from within this function, though there’s no rationale for that discomfort.

mwleeds:
@pwithnall To clarify, is the current behavior of "ostree refs -c" also deliberate?

pwithnall:
> @pwithnall To clarify, is the current behavior of "ostree refs -c" also deliberate?

Less so. It deliberately doesn’t list refs for which there is no collection ID. I think the fact it doesn’t list refs from `refs/remotes/*/*` which could have their collection ID looked up from the remote config is just an omission. Again, it would be good if @cgwalters could clarify what he prefers the behaviour to be for this command line tool.

mwleeds:
@rh-atomic-bot retry

src/libostree/ostree-repo-refs.c
```diff
@@ -1159,36 +1161,64 @@ ostree_repo_list_collection_refs (OstreeRepo    *self,
...
-
-  while (refs_mirrors_exists)
+  for (char **refs_dir = refs_dirs; refs_dir && *refs_dir; refs_dir++)
     {
-      struct dirent *dent;
-      glnx_fd_close int collection_dfd = -1;
-
-      if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent, cancellable, error))
+      gboolean refs_dir_exists = FALSE;
+      if (!ot_dfd_iter_init_allow_noent (self->repo_dir_fd, *refs_dir,
```
cgwalters: I tend to prefer naming like `char **iter = refs_dirs; ` and then `const char *refdir = *iter`.

src/libostree/ostree-repo-refs.c
```diff
@@ -1159,36 +1161,64 @@ ostree_repo_list_collection_refs (OstreeRepo    *self,
...
+              else
+                current_collection_id = g_strdup (remote_collection_id);
+            }
+
+          if (!glnx_opendirat (dfd_iter.fd, dent->d_name, TRUE, &collection_dfd, error))
+            return FALSE;
+
+          if (!enumerate_refs_recurse (self, NULL, OSTREE_REPO_LIST_REFS_EXT_NONE,
+                                       current_collection_id, collection_dfd, base_path,
+                                       collection_dfd, ".",
```
cgwalters: We can just pass `dfd_iter.fd, dent->d_name` here and drop the local fd right?

cgwalters:
Hmm...including remote refs in the summary seems wrong-ish to me.  In some of our CI jobs for example we pull from another repo to e.g. generate deltas, but we're not really intending to be a source of the original content necessarily.  On the other hand, we *do* have the content...and being able to do that kind of thing opportunistically is the idea behind collections.

cgwalters:
But what we do in the API and command line doesn't need to match what we do for the summary necessarily.  I can definitely understand wanting `refs -c` to show remote refs with collections.  So perhaps we land this but tweak it to not enumerate remotes when generating the summary?  

Also would be good to have a test case for this regardless.

pwithnall:
My original question was about the semantics you want for the `ostree_repo_list_refs_*()` APIs, @cgwalters. I guess that means the questions here are actually threefold:
 * What behaviour do we want for the APIs?
 * What behaviour do we want for the `summary` file?
 * What behaviour do we want for the `refs` and `refs -c` command line tools?

> So perhaps we land this but tweak it to not enumerate remotes when generating the summary?

In order to avoid a combinatorial explosion of different `ostree_repo_*_list_*_refs_*()` methods, perhaps this is where we should start using `OstreeRepoListRefsExtFlags` a bit more? Since you’ve broken the ice by adding a flag for aliases.

We could add a flag for including `refs/remotes/*/*` in the results. When passed to `ostree_repo_list_refs_ext()` it would switch between listing `refs/heads/*`; and listing `refs/heads/*` *and* `refs/remotes/*/*`. When passed to `ostree_repo_list_collection_refs()`*, it would switch between listing `refs/heads/*` and `refs/mirrors/*/*`; and listing `refs/*/*/*`.

*We would need to break experimental API and add a `OstreeRepoListRefsExtFlags` argument to `ostree_repo_list_collection_refs()`. I’m fine with that.

---

For the `summary` file, I think it would be OK to list remote refs which have a collection ID, since we’re not claiming to be the canonical source for them (unless they have the same collection ID as we’ve set as the main one in our `summary` file). This could be an option in `repo/config` though.

---

For the command line tools, I have no real preferences as long as: a) it’s possible to list all the refs somehow (potentially by passing several different options to `ostree refs`); and b) it’s easy to work out what kind of ref each one is (heads, mirrors or remotes; or an alias). Otherwise debugging is a complete pain.

cgwalters:
> <walters> ah i didn't understand that was a blocker
> <pwithnall[m]> yeah, the outcome of the thinking about which refs to put in the `summary` file affects how we implement our OSTree server to serve stuff over the LAN
> <pwithnall[m]> if the `summary` file does include mirrored refs, we can serve it up straight from the local repository
> <pwithnall[m]> if the `summary` file _doesn’t_ include mirrored refs, we need to rewrite it when serving

cgwalters:
I think I agree with basically everything you said above; what feels unclear to me is the `refs` command...honestly I really dislike it as is today, I find it confusing.  So I have a hard time saying what the behavior should be :smile:.  

The commit message is talking about the command line, but this is changing the *API* and hence everything that uses it.  It sounds like based on your proposal, this API change would cause the desired behavior changes, so perhaps we just need to change the commit message?

Should definitely have test cases for this though that cover the summary and `ostree refs` behavior.

pwithnall:
> Philip Withnall says:
> walters: cool, thanks, that makes sense
> regarding the refs command, how much do you care about backwards compatibility of its: a) arguments; b) output format?
> We could nuke it and redesign it from scratch to be less confusing; or change its output format for the existing options; or keep full compatibility and just add more options to do sensible things
> I guess if compatibility were broken, all the unit tests would need to be updated

> walters (IRC) says:
> offhand wary of completely changing output by default, but probably best to chat in an issue about this
> we could pull a trick like changing the output if stdout is a tty though that has its own set of downsides

> Philip Withnall says:
> ouch, that sounds a little evil
> I think I should probably take a look at what the delta is between the current output and some ‘ideal’ perfect output, and see what changes are feasible

pwithnall:
Ideally, I think the command line output from `refs` should:
 - indicate the type of each ref (heads, mirrors, remote)
 - potentially allow filtering by type
 - allow filtering to only show the refs which have a collection ID (i.e. mirrors, plus heads if the repository has a collection ID set)
 - indicate which refs are aliases
 - show the checksum for each ref (potentially not show the checksum for aliases, or show it alongside the name of the target of the alias)

However, the current output format is consumed by quite a few things (probably more than just our unit tests), so we probably can’t change it. We do have the differentiator of the `--list` argument — perhaps we could keep the existing output format when `--list` is not passed, and change the output format when it is passed? Otherwise we’re going to have to introduce a new argument (`--pretty`?) to modify the behaviour of `refs` to include more things in the output.

I suggest an output format like the following for non-collection refs, with fields (type, ref name/details, checksum):
```
remote remote1:my-favourite-ref 0123abcd
local  my-local-ref abcd0123
local  my-alias → my-local-ref abcd0123
```

And the following for `-c` output:
```
local  (localhost.TmpCollection, my-local-ref) abcd0123
local  (localhost.TmpCollection, my-alias) → (localhost.TmpCollection, my-local-ref) abcd0123
mirror (org.test.Collection, another-ref) 01230123
remote (org.test.Collection2, remote-ref-where-remote-has-collection-configured) ab01cd23
```

We could add a `--type {local,mirror,remote}` argument to filter by type. It could be specified multiple times to allow more types in the output.

It would be cool if we could [use coloured output to enhance readability](https://the.exa.website/), though that’s not something to handle in this task.

mwleeds:
My plan is to rework this PR to:

- implement the API changes proposed in https://github.com/ostreedev/ostree/pull/1069#issuecomment-322524151
- not affect `ostree refs` output
- include remote refs in `ostree refs -c`
- include remote refs in `summary` files
- have unit tests for the above

mwleeds:
Pushed a new version. @cgwalters if you could put reviewing this near the top of your todo list it would be much appreciated, because @pwithnall is going on PTO tomorrow and a few things are depending on this.

tests/test-find-remotes.sh
```diff
@@ -57,7 +57,7 @@ assert_file_has_content refs "^apps-remote:app1$"
 assert_file_has_content refs "^os-remote:os/amd64/master$"
 
 ${CMD_PREFIX} ostree --repo=local refs --collections | wc -l > refscount
-assert_file_has_content refscount "^0$"
+assert_file_has_content refscount "^2$"
```
pwithnall: It would be good to check what those 2 refs actually are, for paranoia reasons.

pwithnall: I’ve pushed a fixup.

src/libostree/ostree-repo-prune.c
```diff
@@ -335,7 +335,7 @@ ostree_repo_prune (OstreeRepo        *self,
       g_autoptr(GHashTable) all_collection_refs = NULL;  /* (element-type OstreeChecksumRef utf8) */
 
       if (!ostree_repo_list_collection_refs (self, NULL, &all_collection_refs,
-                                             cancellable, error))
+                                             OSTREE_REPO_LIST_REFS_EXT_EXCLUDE_REMOTES, cancellable, error))
```
cgwalters: Why?  Won't this break pruning for cases where we don't have collections but we use a remote, like the default ostree-as-host case?

src/ostree/ot-builtin-fsck.c
```diff
@@ -252,6 +252,7 @@ ostree_builtin_fsck (int argc, char **argv, GCancellable *cancellable, GError **
 
   g_autoptr(GHashTable) all_collection_refs = NULL;  /* (element-type OstreeCollectionRef utf8) */
   if (!ostree_repo_list_collection_refs (repo, NULL, &all_collection_refs,
+                                         OSTREE_REPO_LIST_REFS_EXT_EXCLUDE_REMOTES,
```
cgwalters: Similar question here as for prune.

src/libostree/ostree-repo-prune.c
mwleeds: Doesn't the `ostree_repo_list_refs` call a few lines above this take care of the remote refs?

cgwalters: Ah, yes.  Thanks.

cgwalters:
@rh-atomic-bot r+  2431103

cgwalters:
@rh-atomic-bot retry

PR #1070, opened 2017-08-09
lib/deploy: Close test suite race condition

cgwalters:
Saw this in a PR result; we need to wait for the child to have written its
result to stderr before we exit, otherwise the test suite may not read it in
time.


jlebon:
@rh-atomic-bot r+ 33af500

jlebon:
@rh-atomic-bot retry

Issue #1074, opened 2017-08-11
anaconda install of same commit that is embedded in ISO local cache fails


dustymabe:
If I try to install a system to the exact same ref/commit that is in the local cache on an ISO then I get an error during install:

```
Installing.
Starting pull of fedora/26/x86_64/atomic-host from fedora-atomic
Preparing deployment of fedora/26/x86_64/atomic-host
Deployment starting: fedora/26/x86_64/atomic-host
================================================================================
================================================================================
Error

  The following error occurred while installing.  This is a fatal error and
  installation will be aborted.

  ostree ['admin', '--sysroot=/mnt/sysimage', 'deploy', '--os=fedora-atomic',
  'fedora-atomic:fedora/26/x86_64/atomic-host'] exited with code 1
```

Here is the error from the program.log:

```
00:34:33,696 INFO program: Running... ostree admin --sysroot=/mnt/sysimage init-fs /mnt/sysimage
00:34:33,723 DEBUG program: Return code: 0
00:34:35,231 INFO program: Running... ostree admin --sysroot=/mnt/sysimage os-init fedora-atomic
00:34:35,261 INFO program: ostree/deploy/fedora-atomic initialized as OSTree root
00:34:35,261 DEBUG program: Return code: 0
00:34:35,262 INFO program: Running... ostree admin --sysroot=/mnt/sysimage deploy --os=fedora-atomic fedora-atomic:fedora/26/x86_64/atomic-host
00:34:35,296 INFO program: error: Checking out tree: No su945.dirtree object 85e4d486783fd39c71cc92a5c7fdc9014da518658239ddbcf401b7903e788
00:34:35,297 DEBUG program: Return code: 1
```

**Note that this is NOT an issue if I pull from a ref that has changed (i.e. a newer commit)** 

This was with a kickstart file that had the following ostreesetup line:

```
ostreesetup --osname="fedora-atomic" --remote="fedora-atomic" --url="https://kojipkgs.fedoraproject.org/atomic/26/" --ref="fedora/26/x86_64/atomic-host" --nogpg
```

with `ostree-2017.9-2.fc26.x86_64`, `rpm-ostree-2017.6-3.fc26.x86_64` and `anaconda-26.21.12-1.fc26.x86_64` embedded in the ISO.

The ISO I used is [here](https://dustymabe.fedorapeople.org/test-ostree-anaconda-productimg.iso).

jlebon:
Fix in https://github.com/ostreedev/ostree/pull/1076.

PR #1076, opened 2017-08-11
pull: mark commits from local cache as partial

jlebon:
If one of the localcache repos has the exact same commit we resolved
from the remote, then we need to make sure to mark it as partial so that
we download the full tree.

Closes: #1074


cgwalters:
@rh-atomic-bot r+ 5ab9e31

jlebon:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1077, opened 2017-08-14
Add bash completion

### LINKED ISSUE #933: autocompletion
I must confess I understand the "git idea" under ostree but in general all seems like black magic

Provide autocompletion will help to demistify it (as people can list contents and understand what's under the hood)

For example I recently learn how to use `ls` or `cat` command after realize that it list contents under repo/refs, like (for flatpak)

```
$ ostree --repo=/var/lib/flatpak/repo remote list
gnome
gnome-apps-nightly
gnome-nightly

$ ostree --repo=/var/lib/flatpak/repo remote show-url gnome-nightly
https://sdk.gnome.org/nightly/repo/

$ ls -1 /var/lib/flatpak/repo/refs/remotes/gnome-nightly/runtime/
org.gnome.Platform
org.gnome.Platform.Locale
org.gnome.Sdk
org.gnome.Sdk.Debug
org.gnome.Sdk.Locale
```

So you can complete this useful commands with ease:

- Contents of a repo:
    ```
    $ ostree --repo=/var/lib/flatpak/repo ls -R gnome-nightly/runtime/org.gnome.Sdk.Debug/ | less
    ```

- Contents of a file in a repo:
    ```
    $ ostree --repo=/var/lib/flatpak/repo cat gnome-nightly/runtime/org.gnome.Sdk.Debug/x86_64/master /files/source/glib/glib/gvariant-parser.c | less
    ```

Do you find useful an autocompletion script for bash-completion, is there any ongoing effort for that? I can provide a first implementation if none exist yet

mbarnes:
Scratching an old itch with ostree.

This completes not only command and option names, but also commit checksums, ref names, remote names, os names, and file paths.

Heavily based on how [atomic does it](https://github.com/projectatomic/atomic/blob/master/bash/atomic).

Closes #933

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -88,6 +88,17 @@ AS_IF([test "$YACC" != "bison -y"], [AC_MSG_ERROR([bison not found but required]
 
 PKG_PROG_PKG_CONFIG
 
+# PKG_CHECK_VAR added to pkg-config 0.28
+m4_define_default(
+    [PKG_CHECK_VAR],
+    [AC_ARG_VAR([$1], [value of $3 for $2, overriding pkg-config])
+     AS_IF([test -z "$$1"], [$1=`$PKG_CONFIG --variable="$3" "$2"`])
+     AS_IF([test -n "$$1"], [$4], [$5])])
+
+PKG_CHECK_VAR(bashcompdir, [bash-completion], [completionsdir], ,
+  bashcompdir="${sysconfdir}/bash_completion.d")
+AC_SUBST(bashcompdir)
+
 AM_PATH_GLIB_2_0(,,AC_MSG_ERROR([GLib not found]))
 
 dnl When bumping the gio-unix-2.0 dependency (or glib-2.0 in general),
--- a/configure.ac
+++ b/configure.ac
@@ -95,9 +95,9 @@ m4_define_default(
      AS_IF([test -z "$$1"], [$1=`$PKG_CONFIG --variable="$3" "$2"`])
      AS_IF([test -n "$$1"], [$4], [$5])])
 
-PKG_CHECK_VAR(bashcompdir, [bash-completion], [completionsdir], ,
-  bashcompdir="${sysconfdir}/bash_completion.d")
-AC_SUBST(bashcompdir)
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -88,6 +88,17 @@ AS_IF([test "$YACC" != "bison -y"], [AC_MSG_ERROR([bison not found but required]
...
+# PKG_CHECK_VAR added to pkg-config 0.28
+m4_define_default(
+    [PKG_CHECK_VAR],
+    [AC_ARG_VAR([$1], [value of $3 for $2, overriding pkg-config])
+     AS_IF([test -z "$$1"], [$1=`$PKG_CONFIG --variable="$3" "$2"`])
+     AS_IF([test -n "$$1"], [$4], [$5])])
+
+PKG_CHECK_VAR(bashcompdir, [bash-completion], [completionsdir], ,
+  bashcompdir="${sysconfdir}/bash_completion.d")
+AC_SUBST(bashcompdir)
```
jlebon: Can we add a `--with-bashcompdir` here to override this? I'd like to be able to do unprivileged builds.

jlebon:
Not sure what happened with the CentOS tester:

```
+ test -x /usr/bin/clang
### EXITED WITH CODE 1 AFTER 276s
```

That shouldn't be fatal, it's used in an if statement. We can investigate more if it happens again.

configure.ac
mbarnes: Actually, if you do `configure --help` you'll see `bashcompdir` listed as an environment variable you can use to override the `completionsdir` variable from `bash-completion.pc`.
```
$ ./configure --help
...
Some influential environment variables:
...
  bashcompdir  value of completionsdir for bash-completion, overriding pkg-config
...
```
Would that suffice for you?


(I'll capitalize/rename the variable so it's consistent with the other environment variables listed there.)

mbarnes:
:arrow_up:  `configure --help` now reads:
```
Some influential environment variables:
  ...
  BASH_COMPLETIONSDIR
              value of completionsdir for bash-completion, overriding
              pkg-config
  ...
```

jlebon:
@rh-atomic-bot r+ 58c6670

PR #1079, opened 2017-08-15
lib/sysroot: Support /usr/lib/modules/$kver for kernel/initramfs

cgwalters:
This is the new Fedora kernel standard layout; it has the advantage
of being in `/usr` like `/usr/lib/ostree-boot`, but it's not OSTree
specific.

Further, I think in practice forcing tree builders to compute the checksum is an
annoying stumbling block; since we already switched to e.g. computing checksums
always when doing pulls, the cost of doing another checksum for the
kernel/initramfs is tiny. The "bootcsum" becomes more of an internal
implementation detail.

Now, there is a transition; my current thought for this is that rpm-ostree will
change to default to injecting into both `/usr/lib/ostree-boot` and
`/usr/lib/modules`, and stop doing `/boot`, then maybe next year say we drop the
`/usr/lib/ostree-boot` by default.

### DIFF
--- a/docs/manual/deployment.md
+++ b/docs/manual/deployment.md
@@ -43,8 +43,13 @@ to a filesystem tree that represents the underlying basis of a
 deployment.  For short, we will call this the "tree", to
 distinguish it from the concept of a deployment.
 
-First, the tree must include a kernel stored as
-`vmlinuz(-.*)?-$checksum` in either `/boot` or `/usr/lib/ostree-boot`.
+First, the tree must include a kernel (and optionally an initramfs).  The
+current standard locations for these are `/usr/lib/modules/$kver/vmlinuz` and
+`/usr/lib/modules/$kver/initramfs`.  The "boot checksum" will be computed
+automatically.  This follows the current Fedora kernel layout, and is
+the current recommended path.  However, older versions of libostree don't
+support this; you may need to also put kernels in the previous (legacy)
+paths, which are `vmlinuz(-.*)?-$checksum` in either `/boot` or `/usr/lib/ostree-boot`.
 The checksum should be a SHA256 hash of the kernel contents; it must be
 pre-computed before storing the kernel in the repository.  Optionally,
 the directory can also contain an initramfs, stored as
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -882,46 +882,173 @@ ostree_sysroot_write_origin_file (OstreeSysroot         *sysroot,
                                      cancellable, error);
 }
 
-/* Originally OSTree defined kernels to be found underneath /boot
- * in the tree.  But that means when mounting /boot at runtime
- * we end up masking the content underneath, triggering a warning.
- *
- * For that reason, and also consistency with the "/usr defines the OS" model we
- * later switched to defining the in-tree kernels to be found under
- * /usr/lib/ostree-boot.
- */
-static gboolean
-get_kernel_from_tree (int             deployment_dfd,
-                      int            *out_boot_dfd,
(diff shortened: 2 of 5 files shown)


cgwalters:
Oh ug, the uboot code looks in `/usr/lib/ostree-boot` explicitly.

cgwalters:
Reworked this.

docs/manual/deployment.md
```diff
@@ -43,8 +43,13 @@ to a filesystem tree that represents the underlying basis of a
 deployment.  For short, we will call this the "tree", to
 distinguish it from the concept of a deployment.
 
-First, the tree must include a kernel stored as
-`vmlinuz(-.*)?-$checksum` in either `/boot` or `/usr/lib/ostree-boot`.
+First, the tree must include a kernel (and optionally an initramfs).  The
+current standard locations for this is `/usr/lib/modules/$kver/vmlinuz` and
```
jlebon: Nit: s/this is/these are/ ?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -882,46 +882,190 @@ ostree_sysroot_write_origin_file (OstreeSysroot         *sysroot,
...
+  /* First, gather from usr/lib/modules/$kver if it exists */
+  if (!get_kernel_from_tree_usrlib_modules (deployment_dfd, &usrlib_modules_layout, cancellable, error))
+    return FALSE;
+
+  /* Gather the legacy layout; if one exists, and it has an initramfs
+   * but the module layout doesn't, the legacy layout wins.  This is what happens
+   * with rpm-ostree with Fedora today, until rpm-ostree learns the new layout.
+   */
+  g_autoptr(OstreeKernelLayout) legacy_layout = g_new0 (OstreeKernelLayout, 1);
+  legacy_layout->boot_dfd = -1;
```
jlebon: */me puts on Coverity hat*
This feels like it deserves an `_ostree_kernel_layout_new ()` to be safe, though up to you!

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -882,46 +882,190 @@ ostree_sysroot_write_origin_file (OstreeSysroot         *sysroot,
...
       else
         {
-          if (!glnx_opendirat (deployment_dfd, "boot", TRUE, &ret_boot_dfd, error))
-            return FALSE;
+          legacy_layout->initialized = TRUE;
+          break;
         }
     }
 
+  /* ret_boot_dfd will point to either /usr/lib/ostree-boot or /boot, let's
```
jlebon: `s/ret_boot_dfd/legacy_layout->boot_dfd/`

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -882,46 +882,190 @@ ostree_sysroot_write_origin_file (OstreeSysroot         *sysroot,
...
+   * inspect it.
+   */
   g_auto(GLnxDirFdIterator) dfditer = { 0, };
-  if (!glnx_dirfd_iterator_init_at (ret_boot_dfd, ".", FALSE, &dfditer, error))
+  if (!glnx_dirfd_iterator_init_at (legacy_layout->boot_dfd, ".", FALSE, &dfditer, error))
     return FALSE;
 
   while (TRUE)
```
jlebon: Could we factor this out as well into a separate function similar to `get_kernel_from_tree_usrlib_modules` that spits out the `legacy_layout`? Would make this function much nicer to read if it just has the business logic.

cgwalters: Yep, definitely looks nicer.  Update :arrow_down:

jlebon:
Nice! @rh-atomic-bot r+ ae00359

PR #1080, opened 2017-08-15
Add naggy comments to help keep options in sync

mbarnes:
I meant to tack this on to #1077...

CLI options have to be kept synchronized in three different places now (code, man pages, bash completions).  I don't know how to automatically enforce this, so I just added a bunch of comments as a reminder to developers.

No code changes here.


cgwalters:
Hm, one of the contexts flaked on
> File 'err.txt' doesn't match regexp 'fifreeze watchdog was run'
I'll look at that.

Thanks for the patch!

@rh-atomic-bot r+ f6d9a2d

PR #1082, opened 2017-08-16
Miscellaneous patches

jlebon:
Fix signal callback signature. And other minor nits I've accumulated.


cgwalters:
@rh-atomic-bot r+ 65aec35

PR #1083, opened 2017-08-16
build-sys: Move bash completions to /usr/share/ by default

cgwalters:
This is in line with the "/etc is for sysadmins", "/usr is OS" model;
e.g. systemd's bash completions go there.

Making this change since I was looking at the required spec file changes.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -96,7 +96,7 @@ m4_define_default(
      AS_IF([test -n "$$1"], [$4], [$5])])
 
 PKG_CHECK_VAR(BASH_COMPLETIONSDIR, [bash-completion], [completionsdir], ,
-  BASH_COMPLETIONSDIR="${sysconfdir}/bash_completion.d")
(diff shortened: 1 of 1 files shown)


cgwalters:
Hum, I swear I'd tested this...you're clearly right though!

cgwalters:
Fixup :arrow_up:

mbarnes:
CI failure looks like a flake.  How do we rerun these?

jlebon:
You can just plow right through and r+ anyway. It will rerun the test on merge. In the future if you really do want to rerun before r+, you can say (in its own comment): `bot, retest this please`. See also https://github.com/projectatomic/papr#more-details-about-project-atomic-ci-services.

cgwalters:
Yeah, one of the nice things about using homu is you can just fire off an `r+` without waiting for the PR status.

mbarnes:
@rh-atomic-bot r+ b773573

PR #1084, opened 2017-08-16
lib/deploy: Really close testing race condition

cgwalters:
I added `waitpid()`, but that didn't actually help because we were
`daemon()`izing. Don't daemonize if we're testing so that we can `waitpid()`.

Note I still haven't reproduced this race locally, but I'm pretty sure this will
fix it.

While here, actually check the return value from `waitpid()` just in case
something goes wrong there.


jlebon:
@rh-atomic-bot r+ de4a226

PR #1085, opened 2017-08-16
lib/sysroot: Port a few functions to new style

cgwalters:
Not sure why we didn't do this earlier. Just noticed them when looking at the
code for a different reason.


jlebon:
@rh-atomic-bot r+ aa600e7

Issue #1087, opened 2017-08-17
v2017.9 symbol has no parent


cgwalters:
While prepping a release, I noticed this:

```
diff --git a/src/libostree/libostree-released.sym b/src/libostree/libostree-released.sym
index 0a138966..244d6f53 100644
--- a/src/libostree/libostree-released.sym
+++ b/src/libostree/libostree-released.sym
@@ -414,7 +414,7 @@ global:
 } LIBOSTREE_2017.7;
 
 LIBOSTREE_2017.9 {
-};
+} LIBOSTREE_2017.8;
 
 /* NOTE: Only add more content here in release commits!  See the
  * comments at the top of this file.
```

But...is that a compatible change to make?  What effect in practice does the symbol versioning parent have?  Since this section was empty, presumably it's irrelevant, but what would happen if we had a symbol in there?

cgwalters:
AFAICS, testing with the `2017.10` in `-devel.sym` and an rpm-ostree that links to some of those new symbols, it doesn't matter if we have a parent or not.

PR #1088, opened 2017-08-17
lib: Fix v2017.10 symbols to inherit from v2017.8

cgwalters:
I actually don't quite know what the version inheritance really does, but let's
be safe and fix this. I'm being conservative here and fixing it to inherit from
2017.8, skipping .9 since that doesn't have a parent.

Related: https://github.com/ostreedev/ostree/issues/1087


jlebon:
@rh-atomic-bot r+ 1ae0330

PR #1089, opened 2017-08-17
Release

cgwalters:


src/libostree/libostree-released.sym
```diff
@@ -416,6 +416,17 @@ global:
 LIBOSTREE_2017.9 {
 };
```
cgwalters: I think I'd prefer to not touch it; I worry that it will show up as a red flag in [ABI checking programs](https://lvc.github.io/abi-compliance-checker/). etc.

jlebon:
@rh-atomic-bot r+ 11a4f08

Issue #393, opened 2016-07-12
OSTree should clean up old ovl dirs in /var/tmp


jlebon:
Not sure when would be a good time to do this garbage collection. Maybe making it a new `OstreeSysrootCleanupFlag` ?

cgwalters:
We could potentially put them in the deployment root, and have some service that garbage collects on boot.

That said...I think the systemd default of `q /var/tmp 1777 root root 30d` tends towards the slow side.  It might also be interesting to teach systemd about the concept of "boot-id scoped tmpdirs", basically like ostree is doing internally.  Basically introduce `/run/boottmp -> /var/run/.systemdboot-1079bef-c0e4-407e-a09c-e8e147579932`, and teach ostree to write to `/run/boottmp`, and `systemd-tmpfiles` to delete any directories starting with `.systemdboot` or so on bootup.

cgwalters:
Or of course we just start putting them in `/tmp` which is (hopefully) a `tmpfs` except when it isn't.

PR #1090, opened 2017-08-17
boot: Add a tmpfiles.d snippet to clean up /var/tmp/ostree-ovl.XXX

cgwalters:
This is simplest for now.  Compare with similar logic from
`/usr/lib/tmpfiles.d/tmp.conf`:
```
R! /tmp/systemd-private-*
```

Closes: https://github.com/ostreedev/ostree/issues/393


jlebon:
Nice! @rh-atomic-bot r+ 5ac0524

PR #1092, opened 2017-08-18
fix up verification and presentation of GPG subkey signatures

ramcq:
Another pair of fixes on the path to https://github.com/ostreedev/ostree/issues/608 - firstly rework the previous decision to provide the primary key ID as the _ATTR_FINGERPRINT and instead add a new _FINGERPRINT_PRIMARY which is available if the key is present, and printed if it's different. The other is to add the signing GPG homedir as a verifying keyring when doing repo signing, which means that the signing key that we want to find duplicates for is available in the verifying GPGME context, so the previous work to make _lookup match primary key IDs is actually able to succeed.


cgwalters:
> /gpg-verify-result/attribute-basics: **
> OSTree:ERROR:tests/test-gpg-verify-result.c:176:test_attribute_basics: assertion failed (type_string == "(bbbbbsxxssss)"): ("(bbbbbsxxsssss)" == "(bbbbbsxxssss)")

ramcq:
Oops. Updated, should be OK now.

cgwalters:
I tweaked the commit descriptions; I much prefer having topic prefixes etc.

src/libostree/ostree-gpg-verify-result.c
```diff
@@ -418,6 +415,14 @@ ostree_gpg_verify_result_get (OstreeGpgVerifyResult *result,
             child = g_variant_new_string (v_string);
             break;
 
+          case OSTREE_GPG_SIGNATURE_ATTR_FINGERPRINT_PRIMARY:
+            if (key != NULL && key->subkeys != NULL)
+              v_string = key->subkeys->fpr;
+            if (v_string == NULL)
+              v_string = "[unknown fingerprint]";
```
cgwalters: This bit feels magical; how about we make this the empty string, and then clients can render that how they prefer (e.g. translated)?

src/libostree/ostree-gpg-verify-result.h
```diff
@@ -64,6 +64,10 @@ typedef struct OstreeGpgVerifyResult OstreeGpgVerifyResult;
  * @OSTREE_GPG_SIGNATURE_ATTR_USER_EMAIL:
  *   [#G_VARIANT_TYPE_STRING] The email address of the signing key's primary
  *   user
+ * @OSTREE_GPG_SIGNATURE_ATTR_FINGERPRINT_PRIMARY:
+ *   [#G_VARIANT_TYPE_STRING] Fingerprint of the signing key's primary key
+ *   (will be the same as OSTREE_GPG_SIGNATURE_ATTR_FINGERPRINT if the
+ *   the signature is already from the primary key rather than a subkey)
```
cgwalters: "If the primary key is not found, this value will be the empty string"?

src/libostree/ostree-gpg-verify-result.c
ramcq: I was just aiming for consistency with _USER_NAME and _USER_EMAIL, which suffers the same problem - but I suppose that behaviour (of having a baked "empty string") is now ABI. I guess we can do better with the fingerprint on the basis it's more likely to be used as a lookup key, but arguably accessing/using any of these strings if key_missing is true is a programmer error anyway. I don't mind making the default "" if you prefer supporting this programmer error even if it's inconsistent with the others... :)

ramcq:
Updated to make _FINGERPRINT_PRIMARY return "" if key_missing.

cgwalters:
Hm, your force-push lost my force-push to edit the commit messages to match style 😦.

cgwalters:
(Ah hah, I still have the commit messages on my local branch.  Will rebase on yours)

cgwalters:
I suppose to test this nicely we'd need to rework our test suite to have subkeys.  Eh, maybe later.  Otherwise nice work, thanks!

@rh-atomic-bot r+ 9142d84

ramcq:
@cgwalters Let's open another ticket for that! I still have one last tweak on _lookup_sigs so it only canonicalises the signatures, not the key you're trying to match - otherwise if you are trying to find/match/delete a specific subkey sig, you will match any other subkey from the same primary, which is weird/surprising. For me it means ostree gpg-sign --delete doesn't _quite_ behave as expected.

ramcq:
@cgwalters Maybe we should open one last ticket for the test case stuff! I still have one last tweak on _lookup_sigs so it only canonicalises the signature keys, not the key you're trying to lookup - otherwise if you are trying to find/match/delete a specific subkey sig, you will match any other subkey from the same primary, which is weird/surprising. For me it means ostree gpg-sign --delete doesn't _quite_ behave as expected.

cgwalters:
Sure, done in https://github.com/ostreedev/ostree/issues/1152

PR #1093, opened 2017-08-18
docs/build: s/libOSTree/libostree/

cgwalters:
I find "libOSTree" awkward to type and really to look at. Let's be nicer on
people's pinky fingers and eyes and drop it all down to lowercase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,19 +1,22 @@
-libOSTree
-======
+libostree
+---------
 
 New! See the docs online at [Read The Docs (OSTree)](https://ostree.readthedocs.org/en/latest/ )
 
 -----
 
-This project is now known as "libOSTree", renamed from "OSTree"; the focus is on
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -87,6 +88,7 @@ Once you have a git clone or recursive archive, building is the
 same as almost every autotools project:
 
 ```
+git submodule update --init
```
jlebon: `autogen.sh` takes care of that for us, no? Or is there something subtle going on here?

jlebon:
LGTM, just had one minor question, but feel free to `r=jlebon` if I'm missing something.

README.md
cgwalters: It does, but I think it's worth calling out explicitly that we use submodules.

jlebon:
@rh-atomic-bot r+ 2508814

PR #1095, opened 2017-08-19
lib/deploy: Add .img to end of initramfs in /usr/lib/modules

cgwalters:
Follow up to <https://github.com/ostreedev/ostree/pull/1079>; I was working on
the rpm-ostree updates for this, and I think it's more consistent if we have
`.img` here, since that's a closer match to the "remove $kver" that results in
`vmlinuz`. Also just best practice to have file suffix types where they make
sense.

The astute reader might notice this sneaks in a change where we'd crash if the
legacy bootdir didn't have an initramfs...yeah, should probably have test
coverage of that.


jlebon:
@rh-atomic-bot r+ 97a7c78

jlebon:
@rh-atomic-bot retry

PR #1097, opened 2017-08-21
lib/sysroot: fix placement for not-default deployment

guyshapiro:
When using the
OSTREE_SYSROOT_SIMPLE_WRITE_DEPLOYMENT_FLAGS_NOT_DEFAULT flag, the
deployment is said to be added after the booted or merge deployment.
Fix the condition to do so instead of adding it in the second place.

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -1545,7 +1545,7 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
           g_ptr_array_add (new_deployments, g_object_ref (deployment));
         }
 
-      if (!added_new)
+      if ((!added_new) && is_merge_or_booted)
         {
           g_ptr_array_add (new_deployments, g_object_ref (new_deployment));
           added_new = TRUE;


cgwalters:
Hm, yeah...rpm-ostree used this at one point but no longer does so since it carries the "livefs" logic there which needs to interact: https://github.com/projectatomic/rpm-ostree/blob/f51611dc74e64df0b0a3fb95383ba7bebff53272/src/daemon/rpmostree-sysroot-core.c#L316

@rh-atomic-bot r+ 6032b10

src/libostree/ostree-sysroot.c
```diff
@@ -1545,7 +1545,7 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
           g_ptr_array_add (new_deployments, g_object_ref (deployment));
         }
 
-      if (!added_new)
+      if ((!added_new) && is_merge_or_booted)
```
jlebon: But if we're on the merge deployment, and the next one is the booted deployment, we'll still be inserting between the merge and booted deployment, right? It seems like we would need e.g. `met_merge` and `met_booted` vars to keep track.

cgwalters: Yeah, I think you're right; probably best to lift the rpm-ostree logic entirely?

cgwalters: (And expose this on the cmdline so we can unit test it)

guyshapiro: To be honest, I don't plan to use this flag. I just stumbled upon the inconsistency between the code and the comment.
This suggested change makes the code behave as described in the comment, the way I understand it.

I don't have an opinion about the desirable behavior.

PR #1098, opened 2017-08-21
tests: fix admin upgrade timestamp file check

jlebon:
In #1055, I tried to be helpful but that didn't work out all too well.
We need to recompute the file path since one is in archive mode and the
other in bare mode.


cgwalters:
@rh-atomic-bot r+ b07b2bb

cgwalters:
FWIW I did manually test that the tests *fail* when I commented out the `timestamp-check` bit in the upgrader.  I do find myself wondering a lot about solving the "test the tests" problem...

jlebon:
And of course, my fix itself was wrong. Tested to work this time. ⬆️
Apologies for the kerfuffle.

cgwalters:
@rh-atomic-bot r+ 9b49075

jlebon:
@rh-atomic-bot retry

PR #1099, opened 2017-08-21
pull: better description for --mirror

jlebon:
Describe the behaviour of `--mirror` a bit better.

### DIFF
--- a/man/ostree-pull.xml
+++ b/man/ostree-pull.xml
@@ -103,7 +103,12 @@ Boston, MA 02111-1307, USA.
                 <term><option>--mirror</option></term>
 
                 <listitem><para>
-                    Write refs suitable for a mirror.
+                    Write refs suitable for a mirror, i.e. refs are stored in the
+                    <literal>heads/</literal> directory rather than the
+                    <literal>remotes/</literal> directory. This makes the target repo
+                    suitable to be exported for other clients to pull from as an ostree
+                    remote. If no specific refs are specified, all refs will be fetched (the
+                    remote must have a summary file present).
                 </para></listitem>
             </varlistentry>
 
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-repo.h
```diff
@@ -1081,7 +1081,7 @@ gboolean ostree_repo_prune_from_reachable (OstreeRepo             *self,
 /**
  * OstreeRepoPullFlags:
  * @OSTREE_REPO_PULL_FLAGS_NONE: No special options for pull
- * @OSTREE_REPO_PULL_FLAGS_MIRROR: Write out refs suitable for mirrors
+ * @OSTREE_REPO_PULL_FLAGS_MIRROR: Write out refs suitable for mirrors and fetch all refs if none requested
```
dustymabe: I think we need even more detail as noted in https://github.com/ostreedev/ostree/issues/1100. Is this the actual text that shows up in the man page or is this just a code comment?

jlebon: I added a corresponding blurb to the man page.

man/ostree-pull.xml
```diff
@@ -103,7 +103,12 @@ Boston, MA 02111-1307, USA.
                 <term><option>--mirror</option></term>
 
                 <listitem><para>
-                    Write refs suitable for a mirror.
+                    Write refs suitable for a mirror, i.e. refs are stored in the
+                    <literal>heads/</literal> directory rather than the
+                    <literal>remotes/</literal> directory. This makes the target repo
+                    suitable to be exported for other clients to pull from. If no specific
+                    refs are specified, all refs will be fetched (the remote must have a
+                    summary file present).
```
dustymabe: this is much better. I had sat down to write my own version and this almost matched exactly! Only thing I would change is maybe a little more detail here:

`This makes the target repo suitable to be a server for other clients to pull from as an ostree remote.`

jlebon:
@rh-atomic-bot delegate=dustymabe

dustymabe:
i'm guessing we are opposed to updating the `--help` text to give as much information as the man page?

jlebon:
Updated to mention #1100! ⬆️ 

> i'm guessing we are opposed to updating the --help text to give as much information as the man page?

Seems like the `--help` output should be more brief than the man page by nature, no? I hope folks maintaining build/distribution systems will at least make the effort to read the man page. :)

dustymabe:
@rh-atomic-bot r+

PR #1103, opened 2017-08-22
Distcheck fixes

pwithnall:
These are the fixes which I should have made in my earlier half-hearted attempt to ‘fix’ the bash-completion build machinery.

The fixes for systemd and bash-completion affect the 2017.10 release, as far as I can tell. The tmpfiles stuff landed after 2017.10 so is unaffected.


cgwalters:
@rh-atomic-bot r+ e585e74

PR #1105, opened 2017-08-23
libarchive: Add support for translating paths during commit

cgwalters:
For rpm-ostree, I want to move RPM files in `/boot` to `/usr/lib/ostree-boot`.
This is currently impossible without forking the libarchive code.  Supporting
this is pretty straightforward; we already had pathname translation in
the libarchive code, we just need to expose it as an option.

On the command line side, I chose to wrap this as a regexp. That should be good
enough for a lot of use cases; sophisticated users should as always be making
use of the API. Note that this required some new `#ifdef LIBARCHIVE` bits to use
the new API. Following previous patterns here, we use the new API only if a
relevant option is enabled, ensuring unit test coverage of both paths.

For the test cases, I ended up changing the accounting to avoid having to
multiply the test count.

### DIFF
--- a/src/libostree/ostree-repo-libarchive.c
+++ b/src/libostree/ostree-repo-libarchive.c
@@ -123,24 +123,30 @@ squash_trailing_slashes (char *path)
     *endp = '\0';
 }
 
+/* Like archive_entry_stat(), but since some archives only store the permission
+ * mode bits in hardlink entries, so let's just make it into a regular file.
+ * Yes, this hack will work even if it's a hardlink to a symlink.
+ */
+static void
+read_archive_entry_stat (struct archive_entry *entry,
+                         struct stat          *stbuf)
+{
+  const struct stat *st = archive_entry_stat (entry);
+
+  *stbuf = *st;
+  if (archive_entry_hardlink (entry))
+    stbuf->st_mode |= S_IFREG;
+}
+
+/* Create a GFileInfo from archive_entry_stat() */
 static GFileInfo *
 file_info_from_archive_entry (struct archive_entry *entry)
 {
-  const struct stat *st = archive_entry_stat (entry);
-  struct stat st_copy;
+  struct stat stbuf;
+  read_archive_entry_stat (entry, &stbuf);
 
-  /* Some archives only store the permission mode bits in hardlink entries, so
-   * let's just make it into a regular file. Yes, this hack will work even if
-   * it's a hardlink to a symlink. */
-  if (archive_entry_hardlink (entry))
-    {
-      st_copy = *st;
-      st_copy.st_mode |= S_IFREG;
-      st = &st_copy;
-    }
-
-  g_autoptr(GFileInfo) info = _ostree_stbuf_to_gfileinfo (st);
-  if (S_ISLNK (st->st_mode))
+  g_autoptr(GFileInfo) info = _ostree_stbuf_to_gfileinfo (&stbuf);
+  if (S_ISLNK (stbuf.st_mode))
     g_file_info_set_attribute_byte_string (info, "standard::symlink-target",
                                            archive_entry_symlink (entry));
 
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-repo.h
```diff
@@ -687,6 +687,28 @@ gboolean      ostree_repo_write_archive_to_mtree (OstreeRepo                   *
...
+ * will be freed via `g_free()`.
+ *
+ * This pathname translation will be performed *before* any processing from an
+ * active `OstreeRepoCommitModifier`. Will be invoked for all directory and file
+ * types, first with outer directories, then their sub-files and directories.
+ *
+ * Since: 2017.11
+ */
+typedef char *(*OstreeRepoImportArchiveTranslatePathname) (OstreeRepo     *repo,
+                                                           GFileType       file_type,
```
jlebon: Let's just pass the stat buffer directly?

src/libostree/ostree-repo.h
```diff
@@ -687,6 +687,28 @@ gboolean      ostree_repo_write_archive_to_mtree (OstreeRepo                   *
...
+ * @file_type: File type
+ * @src_path: Path in the archive
+ * @user_data: User data
+ *
+ * Possibly change a pathname while importing an archive. If %NULL is returned,
+ * then @src_path will be used unchanged.  Otherwise, return a new pathname which
+ * will be freed via `g_free()`.
+ *
+ * This pathname translation will be performed *before* any processing from an
+ * active `OstreeRepoCommitModifier`. Will be invoked for all directory and file
```
jlebon: We should document how this relates to `use_ostree_convention` here.

src/ostree/ot-builtin-commit.c
```diff
@@ -95,6 +97,7 @@ static GOptionEntry options[] = {
   { "no-xattrs", 0, 0, G_OPTION_ARG_NONE, &opt_no_xattrs, "Do not import extended attributes", NULL },
   { "link-checkout-speedup", 0, 0, G_OPTION_ARG_NONE, &opt_link_checkout_speedup, "Optimize for commits of trees composed of hardlinks into the repository", NULL },
   { "tar-autocreate-parents", 0, 0, G_OPTION_ARG_NONE, &opt_tar_autocreate_parents, "When loading tar archives, automatically create parent directories as needed", NULL },
+  { "tar-pathname-filter", 0, 0, G_OPTION_ARG_STRING, &opt_tar_pathname_filter, "When loading tar archives, use REGEX,REPLACEMENT against path names", "REGEX,REPLACEMENT" },
```
jlebon: This feels like it should be a `G_OPTION_ARG_STRING_ARRAY` instead, where the first filter that applies wins. Though since it's fully backwards compatible, we can do that later if we feel there's a use case for having CLI support for it.

src/ostree/ot-builtin-commit.c
```diff
@@ -219,6 +222,31 @@ commit_filter (OstreeRepo         *self,
...
+  TranslatePathnameData *tpdata = user_data;
+  g_autoptr(GError) tmp_error = NULL;
+  char *ret =
+    g_regex_replace (tpdata->regex, path, -1, 0,
+                     tpdata->replacement, 0, &tmp_error);
+  if (!ret)
+    {
+      g_assert_no_error (tmp_error);
+      g_assert_not_reached ();
+    }
```
jlebon: If the replacement didn't make any changes, we should return `NULL`, no? At least so we test the fallback behaviour.

src/libostree/ostree-repo-libarchive.c
```diff
@@ -247,6 +253,18 @@ aic_get_final_path (OstreeRepoArchiveImportContext *ctx,
...
+      struct stat stbuf;
+      path = path_relative (path, error);
+      read_archive_entry_stat (ctx->entry, &stbuf);
+      GFileType ftype = ot_gfiletype_from_stbuf (&stbuf);
+      char *ret = ctx->opts->translate_pathname (ctx->repo, ftype, path,
+                                                 ctx->opts->translate_pathname_user_data);
+      if (ret)
+        return ret;
+      /* Fall through */
+    }
```
jlebon: I'd say we add an `else` here. I.e. if `translate_pathname` is active, `use_ostree_convention` has no effect. Otherwise, it's odd thinking that there's a difference between returning `NULL` and returning the exact same string.

src/libostree/ostree-repo.h
cgwalters: I did that originally, but then we'd need to `#include <sys/stat.h>` in `ostree.h`.  Which...is probably fine but takes us a bit farther down the unix-only path.  OK, so libglnx already makes us linux-specific...yeah, I'll make that change back.

src/ostree/ot-builtin-commit.c
cgwalters: Agreed; let's consider this later, in the end I think we want to encourage use of the API.

cgwalters:
Updated for comments :arrow_up:

jlebon:
@rh-atomic-bot r+ 49f55c7

Issue #1096, opened 2017-08-19
print the "Usage" two times for any unkown ostree command


isula-xufeng:
Hi,

I found ostree will print "Usage" two times for any unknown ostree command, and after debugging,
I found "Usage" printed in two places both in ostree_run() function:
```
186           else
187             {
188               g_set_error (&error, G_IO_ERROR, G_IO_ERROR_FAILED,
189                            "Unknown command '%s'", command_name);
190               ostree_usage (commands, TRUE);                <-------- the first time printing
191             }
192         }
193
194       help = g_option_context_get_help (context, FALSE, NULL);
195       g_printerr ("%s", help);                                        <------- the second time printing                          
196
197       goto out;
198     }
```
So I want to know is this a issue? could we remove one of them?

Thanks a lot!
Xufeng

cgwalters:
Hi, thanks for the report!

PR in https://github.com/ostreedev/ostree/pull/1106

PR #1106, opened 2017-08-23
bin/main: Remove duplicated usage output on unknown commands

cgwalters:
It's been this way for a long time, though not forever; I went
back to v2014.11 as a random choice and it worked then.  Not
going to bother doing a full archive search for this though.
Anyone who wants more context can help themselves.

Closes: https://github.com/ostreedev/ostree/issues/1096


jlebon:
@rh-atomic-bot r+ 301fbad

Issue #1101, opened 2017-08-21
maint.mk injects LC_ALL=C which is not neccessarily UTF-8 


heftig:
```
  DOC   Building XML
./ostree-decl.txt:1209: warning: OstreeRemote has multiple definitions.
./ostree-decl.txt:3064: warning: OstreeRemote has multiple definitions.
../src/libostree/ostree-bloom.c:75: warning: Section bloom is not defined in the ostree-sections.txt file.
../src/libostree/ostree-remote.c:53: warning: Section remote is not defined in the ostree-sections.txt file.
../src/libostree/ostree-repo-finder-avahi.c:101: warning: Section ostree-repo-finder-avahi is not defined in the ostree-sections.txt file.
../src/libostree/ostree-repo-finder-config.c:58: warning: Section ostree-repo-finder-config is not defined in the ostree-sections.txt file.
../src/libostree/ostree-repo-finder-mount.c:68: warning: Section ostree-repo-finder-mount is not defined in the ostree-sections.txt file.
../src/libostree/ostree-soup-form.c:28: warning: Section soup-form is not defined in the ostree-sections.txt file.
../src/libostree/ostree-soup-uri.c:125: warning: Section soup-uri is not defined in the ostree-sections.txt file.
Traceback (most recent call last):
  File "/usr/bin/gtkdoc-mkdb", line 61, in <module>
    mkdb.Run(options)
  File "/usr/share/gtk-doc/python/gtkdoc/mkdb.py", line 284, in Run
    changed, book_top, book_bottom = OutputDB(os.path.join(ROOT_DIR, MODULE + "-sections.txt"), options)
  File "/usr/share/gtk-doc/python/gtkdoc/mkdb.py", line 671, in OutputDB
    file_objects)
  File "/usr/share/gtk-doc/python/gtkdoc/mkdb.py", line 2203, in OutputDBFile
    'MODULE': MODULE.upper(),
UnicodeEncodeError: 'ascii' codec can't encode character '\u2026' in position 12932: ordinal not in range(128)
make[2]: *** [Makefile:821: sgml-build.stamp] Error 1
make[2]: Leaving directory '/build/ostree/src/ostree/apidoc'
```

`src/libostree/ostree-async-progress.c` contains two annotations with an ellipsis (`…`).
The error vanishes after replacing `LC_ALL = C` in `maint.mk` with `LC_ALL = en_US.utf8`.

ostree v2017.10
gtk-doc master (01844e7)
[...]

PR #1107, opened 2017-08-23
build/maint.mk: Comment out setting of LC_ALL

cgwalters:
This triggers obscure bugs; really we shouldn't be overriding
the global locale here.  In practice, production build systems
will be using fixed locales anyways.

Also, we only use a small subset of this file (`make syntax-check`),
which appears to work OK without this.

I will probably try to work out where to submit this as at
least an issue report for upstream gnulib.

Closes: https://github.com/ostreedev/ostree/issues/1101


cgwalters:
@rh-atomic-bot delegate=heftig

heftig:
@rh-atomic-bot r+

PR #1108, opened 2017-08-23
lib/repo: Add some assertions for ABI sizes

cgwalters:
Things like https://sourceware.org/libabigail/manual/abidiff.html
look interesting but in a brief look I couldn't work out
how to conveniently use them for quick ABI sanity checking without
doing a diff from a previous build (which we could do but would be
more involved).

This way will at least catch struct ABI breaks on x86_64 which
I think we'd be most likely to do accidentally when trying
to use one of the previous unused values.

I found the hole values via gdb's `pahole` command.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -47,6 +47,33 @@
 #include <sys/file.h>
 #include <sys/statvfs.h>
 
+/* ABI Size checks for ostree-repo.h, only for LP64 systems;
+ * https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models
+ */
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo.c
```diff
@@ -607,6 +607,34 @@ ostree_repo_class_init (OstreeRepoClass *klass)
 static void
 ostree_repo_init (OstreeRepo *self)
 {
+  /* ABI Size checks for ostree-repo.h, only for LP64 systems;
```
jlebon: This doesn't have to be in `ostree_repo_init` right? Can we just put it at the top of the file?

jlebon:
Interesting stuff! @rh-atomic-bot r+ 346575c

PR #1110, opened 2017-08-23
ostree-sysroot: make simple_write_deployment smarter

jlebon:
This is a follow-up to #1097.
We make `simple_write_deployment` smart enough so that it can be used for
rpm-ostree's purposes. This is mostly an upstreaming of logic that
already existed there.

Notably we correctly append `NOT_DEFAULT` deployments *after* the booted
deployment and we now support `RETAIN_PENDING` and `RETAIN_ROLLBACK` flags
to have more granularity on deployment pruning.

Expose these new flags on the CLI using new options (as well as expose
the previously existing `NOT_DEFAULT` flag as `--not-as-default`).

I couldn't add tests for `--retain-pending` because the merge deployment
is always the topmost one. Though I did check that it worked in a VM.


cgwalters:
I only glanced at this quickly so far (though first patch looks obviously correct); but am I correct in thinking that we can't quite use this in rpm-ostree yet since we also have the livefs logic?

cgwalters:
Usage in https://github.com/projectatomic/rpm-ostree/pull/942

src/libostree/ostree-sysroot.c
```diff
@@ -1526,39 +1536,55 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
...
           g_ptr_array_add (new_deployments, g_object_ref (new_deployment));
           added_new = TRUE;
         }
     }
 
-  /* In this non-default case , an improvement in the future would be
-   * to put the new deployment right after the current default in the
-   * order.
-   */
-  if (!added_new)
```
cgwalters: Hm, don't we need this to handle the initial-deployment case?

jlebon: Good catch. Fixed in ⬇️!

cgwalters:
I'd wondered why it didn't break the tests, it turns out this API isn't called for initial deployments right now.  Which is OK, but I'd still say worth supporting.

@rh-atomic-bot r+ d3dc8e7

PR #1112, opened 2017-08-24
lib/repo-refs: Fix typos in last commit

mwleeds:
This fixes a whitespace error and a mistake that made it into the last
commit, 7ed881baa, at the last minute.

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -1215,10 +1215,10 @@ ostree_repo_list_collection_refs (OstreeRepo                 *self,
                   continue;
                 }
 
-              if (match_collection_id != NULL && g_strcmp0 (match_collection_id, current_collection_id) != 0)
+              if (match_collection_id != NULL && g_strcmp0 (match_collection_id, remote_collection_id) != 0)
                 continue;
               else
-                  current_collection_id = remote_collection_id;
+                current_collection_id = remote_collection_id;
             }
 
           if (!glnx_opendirat (dfd_iter.fd, dent->d_name, TRUE, &subdir_fd, error))


mwleeds:
@cgwalters sorry about pushing broken code. Fortunately it was quickly caught by @dbnicholson

src/libostree/ostree-repo-refs.c
```diff
@@ -1215,10 +1215,10 @@ ostree_repo_list_collection_refs (OstreeRepo                 *self,
                   continue;
                 }
 
-              if (match_collection_id != NULL && g_strcmp0 (match_collection_id, current_collection_id) != 0)
+              if (match_collection_id != NULL && g_strcmp0 (match_collection_id, remote_collection_id) != 0)
```
cgwalters: Hm, so before in that case `current_collection_id` looks like it was actually an uninitialized pointer.  It seems weird to me that the tests pass both ways.  I haven't looked closely at the tests - is there something wrong there that causes this to spuriously pass before?

dbnicholson: I didn't try very hard, but the test reliably segfaulted for me without initializing the pointer. However, if it's uninitialized, I suppose it could point to a random address that had something strcmp could handle, even 0.

mwleeds: I don't think it passed. I think the CI wasn't using `--enable-experimental-api`, and it only passed for me before I put in the last minute change to the handling of `current_collection_id`.

cgwalters:
We do have [a context](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/128663bd470841ac2eed2ff3855bf27d1ae1dbdd.3.1503612081126518811/output.log) for experimental API, and it passed for both.

dbnicholson:
That's weird. If I'm not mistaken, the log from that PR (#1069) is https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/7ed881baa75192521fa972fa702c66470dc0a225.3.1503604690192925704/output.log.

In my local build, I was getting an uninitialized warning from `src/libostree/ostree-repo-refs.c` and then a segfault in `tests/test-refs-collections.sh`. One thing I notice that's different but may be nothing is that I run the tests through `make check` whereas the CI build seems to use  `gnome-desktop-testing-runner` with installed tests. I don't see either of those problems in that build log.

cgwalters:
OK, this patch looks right.  I honestly still haven't yet fully grokked the test cases/logic here though.

@rh-atomic-bot r+ 128663b

cgwalters:
Sorry, it's super embarrassing our own CI system sucks at doing packages...

@rh-atomic-bot retry

PR #1116, opened 2017-08-25
checkout: add an extra checkout_overwrite mode

peterbaouoft:
This is for issue https://github.com/projectatomic/rpm-ostree/issues/365,
an extra option of overwrite is added so that when there is "non-directory"
file already exist during checkout, the error will be handled.

The extra overwrite mode will later be used in rpm-ostree `checkout_package` function to let the error handling happen before writing to rpmdb.

Currently, this is how it looks 
```
[fedora@atomichosttesting entries]$ sudo rpm-ostree install origin-clients
Checking out tree bc67ace... done
Enabled rpm-md repositories: updates fedora fahc

Updating metadata for 'updates': [===========================================================================================================================================================================] 100%
rpm-md repo 'updates'; generated: 2017-08-24 14:54:00

rpm-md repo 'fedora' (cached); generated: 2017-07-05 20:31:38

rpm-md repo 'fahc' (cached); generated: 2017-08-24 21:47:14


Importing metadata [=========================================================================================================================================================================================] 100%
Resolving dependencies... done
Applying 46 overlays... error: Checking out origin-clients-1.5.1-2.fc26.x86_64: Hardlinking 01/d9c29df59df7a8389aac5e91f4ef23303ae18c7bf57b285886428359d10673.file to kubectl: File exists
```


src/libostree/ostree-repo.h
```diff
@@ -808,6 +808,7 @@ typedef enum {
   OSTREE_REPO_CHECKOUT_OVERWRITE_NONE = 0,
   OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES = 1,
   OSTREE_REPO_CHECKOUT_OVERWRITE_ADD_FILES = 2, /* Since: 2017.3 */
+  OSTREE_REPO_CHECKOUT_OVERWRITE_STRICT = 3 /* Since: 2017.8 */
```
cgwalters: This should be `2017.11`.

I had to re-read the docs/code for `UNION` and `ADD`...so basically `ADD` will *ignore* existing files, but we want to error out.  But how would your new `_STRICT` be different from the existing `_NONE`?  I think I see...we'd error out for existing *directories*.  So maybe let's name this something like `_ADD_DIRS` ?

Also a test case would be great; you'd need to wire this up to the command line like `ostree checkout --add-dirs`, like we already have `ostree checkout --union` etc.


Hmm...so `_ADD_FILES`

peterbaouoft: yup, sure :)

tests/basic-test.sh
```diff
@@ -459,6 +459,25 @@ assert_file_has_content checkout-test-union-add/union-add-test 'existing file fo
...
+$OSTREE checkout --add-dirs test-add-dirs checkout-test-add-dirs
+echo "ok adding new directories and new file"
+# Make a new file, and try the checkout again
+touch add-dirs-test/test_one/another_test_file
+$OSTREE commit ${COMMIT_ARGS} -b test-add-dirs --tree=dir=add-dirs-test
+# Check out the latest commit, should fail due to existing files
+if $OSTREE checkout --add-dirs test-add-dirs checkout-test-add-dirs 2> err.txt; then
+    assert_not_reached "checking out files unexpectedly succedded!"
+fi
+assert_file_has_content err.txt 'File exists'
```
cgwalters: Let's also do a `--union` checkout here and assert that that works?

Also can you squash these two commits together?

peterbaouoft: yup, changed based on the comment, but there are something that I noticed while adding the check for union.

First, when you do a touch for multiple files, it seems that the content between those files were shared, ( I did a test with `touch a` and `touch b`, then whenever I change a, b's content also get changed)

Because of that,  I removed touch and replaced it with echo( which I have seen it working fine so far) 

Second, when I do a modification on the checked out files, it seems like when you do union, the file content will not be overwritten because of the hard link. Therefore, I have to manually remove the file to break the hardlink. Is that wanted behavior?

If I have done something wrong in the code or some of my observations were not suppose to happen, feel free to point it out. :)

cgwalters: Yeah, generally a good idea to use separate content.

As far as unioning....hmm, I'm not sure exactly what you mean there; does it only happen when the files are hardlinked?  There is this comment https://github.com/ostreedev/ostree/blob/355e8516b04801caf50b0b8a218fc8f6321d8d05/src/libostree/ostree-repo-checkout.c#L341
which I think bit me before.

peterbaouoft: yea, I need to remove the hardlinked files and create the one with same name to make file overwrite work.
which is the code section below: 
```
# First 'break' the hard link for the previous files
rm checkout-test-add-dirs/test_one/test_file
echo 'file for testing union mode alongwith add-dirs mode' > checkout-test-add-dirs/test_one/test_file```

File: tests/basic-test.sh [hunk 3]
```diff
@@ -459,6 +459,32 @@ assert_file_has_content checkout-test-union-add/union-add-test 'existing file fo
...
+echo 'second test file' >  add-dirs-test/test_one/test_second_file
+$OSTREE commit ${COMMIT_ARGS} -b test-add-dirs --tree=dir=add-dirs-test
+# Check out the latest commit, should fail due to existing files
+if $OSTREE checkout --add-dirs test-add-dirs checkout-test-add-dirs 2> err.txt; then
+    assert_not_reached "checking out files unexpectedly succedded!"
+fi
+assert_file_has_content err.txt 'File exists'
+# Verify that Union mode still functions properly
+# First 'break' the hard link for the previous files
+rm checkout-test-add-dirs/test_one/test_file
```
peterbaouoft: without this removing line, the `test_file` content will not be affected when union is performed

src/libostree/ostree-repo.h
```diff
@@ -808,6 +808,7 @@ typedef enum {
   OSTREE_REPO_CHECKOUT_OVERWRITE_NONE = 0,
   OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES = 1,
   OSTREE_REPO_CHECKOUT_OVERWRITE_ADD_FILES = 2, /* Since: 2017.3 */
+  OSTREE_REPO_CHECKOUT_OVERWRITE_ADD_DIRS = 3 /* Since: 2017.11 */
```
jlebon: This might be a bit bikeshed, though `ADD_DIRS` seems funny to me, since it also adds files in already existing dirs. How about something like `DISJOINT_UNION_FILES` instead? E.g. description could be: `When layering checkouts, error out on exiting files, and do not modify existing directories`.

peterbaouoft: Hey, thanks for the comment :). I am fine with both names, but if I am a bit unsure I am getting the description correctly. Does it mean that we dont want to modify the existing directories at all upon failure? ( we need to clean up the old files checked out upon failures) 
 because ... we current still do not revert changes done to existing directories upon failure for checkout. 

 Since checkout is done recursively, it is likely that before the checkout errors out, some files have been already checked out into the file system. In that case, we kind of still "modify" existing directories.  Here is a small example of the case I mentioned:
```rm -rf ttt/
[vagrant@localhost ~]$ ls
atomic  CLearning  fedora-atomic  GLibTrial  libdnf  ostree  PythonTesting  repo  rpm-ostree  test  yum
[vagrant@localhost ~]$ sudo ostree checkout test_branch ttt
[vagrant@localhost ~]$ ls
atomic  CLearning  fedora-atomic  GLibTrial  libdnf  ostree  PythonTesting  repo  rpm-ostree  test  ttt  yum
[vagrant@localhost ~]$ cd ttt
[vagrant@localhost ttt]$ ls
a  b
[vagrant@localhost ttt]$ rm a
[vagrant@localhost ttt]$ cd
[vagrant@localhost ~]$ sudo ostree checkout --add-dirs test_branch ttt
error: Hardlinking fb/0029f0a483777d541db42de53320e46e67ffd7df21203087766d966509f14b.file to b: File exists
[vagrant@localhost ~]$ ls
atomic  CLearning  fedora-atomic  GLibTrial  libdnf  ostree  PythonTesting  repo  rpm-ostree  test  ttt  yum
[vagrant@localhost ~]$ cd ttt
[vagrant@localhost ttt]$ ls
a  b
```

I did not touch any other error handling part for original code, so I am guessing this problem( the files remaining) was there before? If that is not the case, please let me know, so I can go ahead and fix it, I am also willing to add the cleanup if needed. :)

jlebon: Yeah, the "do not modify existing directories" could be reworded. In this case, it essentially means "don't error out if the directory already exists"; so changing the set of files in the directory itself is fine.

Anyway, re. the new flag name itself, I think it's OK either way. As long as it's properly documented!

peterbaouoft: yup, agreed, will change the comment to make it more understandable :)

cgwalters: Right, we don't try at all to handle partial checkout errors; the way libostree programs all tend to work is to make a *new* root and throw the entire thing away if something failed.

cgwalters: As far as the name...I spent some time on https://en.wikipedia.org/wiki/Disjoint_union and was confused until I realized it meant the second definition - simply that the sets of files are disjoint and we're unioning them, right?

I suspect either way the doc text is most important.  Maybe:

> When layering checkouts, error out if a file would be replaced, but add new files and directories

?

peterbaouoft: yup sure, sounds good to me, doing it. I was not too sure about disjoint_union earlier too, so thanks for the clarification as well. :)

src/libostree/ostree-repo.h
```diff
@@ -808,6 +808,7 @@ typedef enum {
   OSTREE_REPO_CHECKOUT_OVERWRITE_NONE = 0,
   OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES = 1,
   OSTREE_REPO_CHECKOUT_OVERWRITE_ADD_FILES = 2, /* Since: 2017.3 */
+  OSTREE_REPO_CHECKOUT_OVERWRITE_DISJOINT_UNION_FILES = 3 /* Since: 2017.11 */
```
jlebon: Let's also add a docstring to the comment block just above this enum!

peterbaouoft: yup, thanks for that, did not notice that at all :p,  and sorry for the late response,... as today was a intern lunch happening ( farewell to the old interns because it's their last day :'( ), pushed new one ⬆️

peterbaouoft:
One sec.. this is still WIP..  Let me add the option to the bash/ostree completion as well as man page, (but may take a while .. cuz never done it earlier..  :) )

cgwalters:
Nice work, thank you!

@rh-atomic-bot r+ f96f11d

peterbaouoft:
Thank you all for the comments! :)

PR #1119, opened 2017-08-29
Remove shebang from bash completions

smcv:
bash completions are to be sourced. It makes little sense to
execute them.

Detected by Debian's Lintian tool, which warns about non-executable
files that appear to be #! scripts.

### DIFF
--- a/bash/ostree
+++ b/bash/ostree
@@ -1,5 +1,3 @@
-#!/bin/bash
-#
 # bash completion file for ostree commands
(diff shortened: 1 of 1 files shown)


cgwalters:
And so it begins not with a bang, but with a <something clever>.

@rh-atomic-bot r+ 327fd2d

jlebon:
@rh-atomic-bot retry

PR #1120, opened 2017-08-29
Fix build-time tests when /var/tmp and/or /tmp do not support user xattrs

smcv:
I'm sure I fixed this at least once before, but I had to ignore the build-time tests' results in Debian due to intermittent failures (which are hopefully now resolved), and it regressed.

Some (all?) Debian autobuilders work by unpacking a cached OS tarball onto a tmpfs and building in that chroot, so there is no location visible to the builder where xattrs work.


cgwalters:
@rh-atomic-bot r+ 1017ed2

jlebon:
@rh-atomic-bot retry

PR #1114, opened 2017-08-25
bin/commit: Add --selinux-policy option

cgwalters:
This was really straightforward to implement, and is useful
for dev/test scenarios mainly like we have in rpm-ostree at least.

Closes: https://github.com/ostreedev/ostree/issues/1113

PR #1121, opened 2017-08-30
commit: filter out selinux label before commit

jlebon:
The new --selinux-policy added in [0] exposed a subtle issue in the way
we handle labeling during commit. The CI system in rpm-ostree hit this
when trying to make use of it[1].

Basically, because of the way we use a GVariant to represent xattrs, if
a file to be committed already has an SELinux label, the xattr object
ends up with *two* label entries. This of course throws off fsck later
on, since the checksum will have gone over both entries, even though the
on-disk file will only have a single label (in which the second entry
wins).

I confirmed that the `fsck` added in the installed test fails without
the rest of this patch.

[0] https://github.com/ostreedev/ostree/pull/1114
[1] https://github.com/projectatomic/rpm-ostree/pull/953

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -34,6 +34,7 @@
 #include "ostree.h"
 #include "ostree-core-private.h"
 #include "ostree-repo-private.h"
+#include "ostree-sepolicy-private.h"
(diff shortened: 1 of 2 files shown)


cgwalters:
OK, and the test didn't actually do `chcon` in place.

@rh-atomic-bot r+ 43e4ed8

PR #1122, opened 2017-08-30
tests: Make the deployment mutable in test-sysroot.js

mwleeds:
See the commit message for the details. This issue was preventing the Endless Jenkins instance from building ostree.

### DIFF
--- a/tests/test-sysroot.js
+++ b/tests/test-sysroot.js
@@ -42,6 +42,8 @@ print('1..1')
 
 libtestExec('setup_os_repository archive-z2 syslinux');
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Yeah, this is a super messy area - I tend to run the tests as non-root, but our CI jobs run as root in Docker so should have hit this too...ah, it's because Docker doesn't grant `CAP_LINUX_IMMUTABLE` by default.

@rh-atomic-bot r+ f3218f9

cgwalters:
@rh-atomic-bot r+ f321859

PR #1125, opened 2017-09-01
tree-wide: Replace archive-z2 with archive

cgwalters:
In almost all places. There are just a few exceptions; one tricky bit for
example is that the repo config must still have `mode=archive-z2`, since
`archive` used to mean something else. (We could very likely just get rid of
that check, but eh, later).

I also added a test that one can still do `ostree repo init --mode=archive-z2`.

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -76,7 +76,7 @@ Boston, MA 02111-1307, USA.
     <variablelist>
       <varlistentry>
         <term><varname>mode</varname></term>
-        <listitem><para>One of <literal>bare</literal>, <literal>bare-user</literal> or <literal>archive-z2</literal> (Note that <literal>archive</literal> is used everywhere else.)</para></listitem>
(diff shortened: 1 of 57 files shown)


man/ostree.repo-config.xml
```diff
@@ -76,7 +76,7 @@ Boston, MA 02111-1307, USA.
     <variablelist>
       <varlistentry>
         <term><varname>mode</varname></term>
-        <listitem><para>One of <literal>bare</literal>, <literal>bare-user</literal> or <literal>archive-z2</literal>.  </para></listitem>
+        <listitem><para>One of <literal>bare</literal>, <literal>bare-user</literal> or <literal>archive-z2</literal> (Note that <literal>archive</literal> is used everywhere else.)</para></listitem>
```
jlebon: Nit: shouldn't that 'N' be lower case?

Also, let's put that notice in `man/ostree-init.xml` as well?

cgwalters: Sure, done ⬇️

jlebon:
@rh-atomic-bot r+ 34b6718

PR #1123, opened 2017-09-01
Prep for ostree_sysroot_new_at()

cgwalters:

PR #1126, opened 2017-09-01
main: print usage when no command given

jlebon:
Minor regression from #1106. We
want to print the usage text both when unknown commands are passed, as
well as when no commands are passed at all.

### DIFF
--- a/src/ostree/ot-admin-builtin-instutil.c
+++ b/src/ostree/ot-admin-builtin-instutil.c
@@ -106,7 +106,7 @@ ot_admin_builtin_instutil (int argc, char **argv, GCancellable *cancellable, GEr
 
       /* This will not return for some options (e.g. --version). */
       if (ostree_admin_option_context_parse (context, NULL, &argc, &argv,
-                                             OSTREE_ADMIN_BUILTIN_FLAG_UNLOCKED,
+                                             OSTREE_ADMIN_BUILTIN_FLAG_NO_SYSROOT,
                                              NULL, cancellable, error))
         {
           if (subcommand_name == NULL)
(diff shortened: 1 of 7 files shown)


cgwalters:
At this point let's add a test case?

jlebon:
Fixed an existing related to this and added a check for it. ⬆️

cgwalters:
Not worth rejigging for this, but WDYT about the commit topic style I've been using like `bin/commit` and `lib/commit` to distinguish?  In this case it'd be e.g. `bin/config: Tweak parameter string`.  (Also note at least I tend to capitalize the first word...though that is quite inconsistent even in the linux kernel commits)

@rh-atomic-bot r+ 41c6f98

jlebon:
Yup, that makes sense to me re. `bin` vs `lib`. I usually don't capitalize messages, though I'll make an effort for this repo! :) Now let's see if the CI will play nice today.

@rh-atomic-bot retry

cgwalters:
Looks like this fails `make syntax-check`

jlebon:
OK, rebased and tweaked to adjust for https://github.com/ostreedev/ostree/pull/1123! ⬆️

jlebon:
And reworded commit titles to match convention!

jlebon:
>```
>++ grep --count '^Usage' err
>+ '[' 2 = 1 ']'
>```

Odd, that's definitely not the case.

```sh
$ ostree remote |& grep --count '^Usage'
1
```

Wonder if there's some weird race condition going on.

cgwalters:
I've seen a race or two recently in at least the admin tests that I haven't tracked down.  But this one looks valid:

https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/7f97c7440353ad5c955c13cf4e966594c462b196.3.1504903473009034131/output.log

jlebon:
Heh, that was tricky. `ostree remote` only prints the help output twice when there's no system repo, which explains why I had trouble reproducing it. Fixup above! ⬆️

cgwalters:
Ah, nice work!

@rh-atomic-bot r+ cc7fcda

cgwalters:
Hmm...two new flakes?

jlebon:
@rh-atomic-bot retry

PR #1124, opened 2017-09-01
bin/prune: Port to new style

cgwalters:
No functional changes, all straightforward.  Prep for
https://github.com/ostreedev/ostree/issues/1115

PR #1127, opened 2017-09-01
Prune only branch

cgwalters:


cgwalters:
Depends: https://github.com/ostreedev/ostree/pull/1124

jlebon:
This needs a rebase.

cgwalters:
Rebased! 🏄

tests/test-prune.sh
```diff
@@ -44,16 +44,23 @@ done
 
 ${CMD_PREFIX} ostree --repo=repo pull --depth=-1 origin test
 
+assert_repo_has_n_commits() {
+    repo=$1
+    count=$2
+
+    find ${repo}/objects -name '*.commit' | wc -l > commitcount
+    assert_file_has_content commitcount '^'${count}'$'
+    rm -f commitcount
```
jlebon: How about:

```sh
[[ $(find ... | wc -l) == ${count} ]]
```

?

src/ostree/ot-builtin-prune.c
```diff
@@ -239,6 +242,41 @@ ostree_builtin_prune (int argc, char **argv, GCancellable *cancellable, GError *
...
+              if (!ostree_repo_resolve_rev (repo, ref, FALSE, &commit, error))
+                return FALSE;
+              g_hash_table_add (only_branches_set, *iter);
+            }
+
+          /* Iterate over all refs, add equivalent of --retain-branch-depth=$ref=-1
+           * if the ref isn't in --only-branch set.
+           */
+          g_hash_table_iter_init (&hash_iter, all_refs);
+          while (g_hash_table_iter_next (&hash_iter, &key, &value))
```
jlebon: `GLNX_HASH_TABLE_FOREACH` ?

src/ostree/ot-builtin-prune.c
```diff
@@ -239,6 +242,41 @@ ostree_builtin_prune (int argc, char **argv, GCancellable *cancellable, GError *
...
+          /* Turn --only-branch into a set */
+          g_autoptr(GHashTable) only_branches_set = g_hash_table_new (g_str_hash, g_str_equal);
+          for (char **iter = opt_only_branches; iter && *iter; iter++)
+            {
+              const char *ref = *iter;
+              g_autofree char *commit = NULL;
+              /* Ensure the specified branch exists */
+              if (!ostree_repo_resolve_rev (repo, ref, FALSE, &commit, error))
+                return FALSE;
+              g_hash_table_add (only_branches_set, *iter);
```
jlebon: `s/*iter/ref/` ?

src/ostree/ot-builtin-prune.c
```diff
@@ -239,6 +242,41 @@ ostree_builtin_prune (int argc, char **argv, GCancellable *cancellable, GError *
...
+       * to prune exampleos/x86_64/foo to just the latest commit.
+       */
+      if (opt_only_branches)
+        {
+          /* Turn --only-branch into a set */
+          g_autoptr(GHashTable) only_branches_set = g_hash_table_new (g_str_hash, g_str_equal);
+          for (char **iter = opt_only_branches; iter && *iter; iter++)
+            {
+              const char *ref = *iter;
+              g_autofree char *commit = NULL;
```
jlebon: We can just pass `NULL` right?

src/ostree/ot-builtin-prune.c
```diff
@@ -49,6 +50,7 @@ static GOptionEntry options[] = {
   { "keep-younger-than", 0, 0, G_OPTION_ARG_STRING, &opt_keep_younger_than, "Prune all commits older than the specified date", "DATE" },
   { "static-deltas-only", 0, 0, G_OPTION_ARG_NONE, &opt_static_deltas_only, "Change the behavior of delete-commit and keep-younger-than to prune only static deltas" },
   { "retain-branch-depth", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_retain_branch_depth, "Additionally retain BRANCH=DEPTH commits", "BRANCH=DEPTH" },
+  { "only-branch", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_only_branches, "Only prune BRANCH (may be specified multiple times)", "BRANCH" },
```
jlebon: Have you thought about making these positional args, e.g. just `ostree prune foo bar`? Seems more natural.

Hmm, although `git prune foo bar` has the opposite meaning:

```
<head>...
    In addition to objects reachable from any of our references, keep objects reachable from listed <head>s.
```

which I find odd, but if in case people are used to that, I'm fine with leaving it as a switch!

src/ostree/ot-builtin-prune.c
```diff
@@ -239,6 +242,41 @@ ostree_builtin_prune (int argc, char **argv, GCancellable *cancellable, GError *
...
+
+          /* Iterate over all refs, add equivalent of --retain-branch-depth=$ref=-1
+           * if the ref isn't in --only-branch set.
+           */
+          g_hash_table_iter_init (&hash_iter, all_refs);
+          while (g_hash_table_iter_next (&hash_iter, &key, &value))
+            {
+              const char *ref = key;
+              if (!g_hash_table_contains (only_branches_set, ref))
+                g_hash_table_insert (retain_branch_depth, g_strdup (ref), GINT_TO_POINTER ((int)-1));
```
jlebon: Should we be erroring in case the ref is already in the hash table? E.g. someone  typing `--only-branch foo --retain-branch-depth=bar=5` might not understand what the command is doing.

src/ostree/ot-builtin-prune.c
cgwalters: Mmm...not sure about promoting this to the toplevel quite yet.

cgwalters:
Fixups ⬆️

src/ostree/ot-builtin-prune.c
```diff
@@ -255,20 +255,19 @@ ostree_builtin_prune (int argc, char **argv, GCancellable *cancellable, GError *
...
             {
               const char *ref = *iter;
-              g_autofree char *commit = NULL;
               /* Ensure the specified branch exists */
-              if (!ostree_repo_resolve_rev (repo, ref, FALSE, &commit, error))
+              if (!ostree_repo_resolve_rev (repo, ref, FALSE, NULL, error))
                 return FALSE;
-              g_hash_table_add (only_branches_set, *iter);
+              if (g_hash_table_contains (retain_branch_depth, ref))
+                return glnx_throw (error, "--retain-branch-depth conflicts with --only-branch");
```
jlebon: We want this bit down there no? I.e. we do want to be able to specify `--retain-branch-depth=my-ref=0 --only-branch my-ref`, right?

I was targeting the `g_hash_table_insert` below instead:

```c
GLNX_HASH_TABLE_FOREACH (all_refs, const char *, ref)
  {
    if (!g_hash_table_contains (only_branches_set, ref))
      {
        if (!g_hash_table_insert (retain_branch_depth, g_strdup (ref), GINT_TO_POINTER ((int)-1)))
          return glnx_throw (error, "--retain-branch-depth conflicts with --only-branch");
      }
  }
```

cgwalters: Hum...so a high level design I had here was that with `--only-branch` one just uses the global `--depth`.  In other words: `--retain-branch-depth=my-ref=0 --only-branch my-ref` == `--refs-only --only-branch my-ref` right?

I'm not quite sure when one would want to combine `--only-branch` with different depths.

jlebon: Don't we want to support e.g. `--only-branch foo --only-branch bar` ? So then you have to resort to `--retain-branch-depth` if you want them pruned at different depths, no? Well, they can just use separate invocations I suppose (or use the API directly). Though it seems easy enough to support. Otherwise, we should just error out if any `--retain-branch-depth` is passed at all when `--only-branch` is given I'd say. And also, remove the related comment block:

```diff
+      /* Process --only-branch. Note this combines with --retain-branch-depth; one
+       * could do e.g.:
+       *  * --only-branch exampleos/x86_64/foo
+       *  * --retain-branch-depth exampleos/x86_64/foo=0
+       * to prune exampleos/x86_64/foo to just the latest commit.
+       */
```

cgwalters: Yeah, I think your original comment here was right - I don't see a reason not to support combining the two, even with different branches.  Thanks for sticking this out...another fixup ⬇️

PR #1130, opened 2017-09-05
bin/local-pull: Clarify docs, add more tests for corrupted  local pulls

cgwalters:
I was reading the pull-local command docs and realized it was somewhat unclear
that `--untrusted` *only* applied to local repo pulls; in other words that we
always treat non-local pulls as untrusted.

Tweak the docstring, and add tests that verify this explicitly.


jlebon:
@rh-atomic-bot r+ dd78238

jlebon:
@rh-atomic-bot retry

PR #1132, opened 2017-09-05
lib: Add a private helper to abort txns, use in sysroot cleanup

cgwalters:
Steal some code from flatpak for this, which allows porting a few more things to
new style. I started on a public API version of this but was trying to roll some
other things into it and it snowballed.

Let's do this version since it's easy for now.

Note that while here I changed things so that `generate_deployment_refs()` now
uses a single txn rather than 4.

### DIFF
--- a/src/libostree/ostree-sysroot-cleanup.c
+++ b/src/libostree/ostree-sysroot-cleanup.c
@@ -21,6 +21,7 @@
 #include "config.h"
 
 #include "otutil.h"
+#include "ostree-repo-private.h"
 #include "ostree-linuxfsutil.h"
 
 #include "ostree-sysroot-private.h"
@@ -341,10 +342,7 @@ cleanup_old_deployments (OstreeSysroot       *self,
   return TRUE;
 }
 
-/* libostree holds a ref for each deployment's exact checksum to avoid it being
- * GC'd even if the origin ref changes.  This function resets those refs
- * to match active deployments.
- */
+/* Delete the ref bindings for a non-active boot version */
 static gboolean
 cleanup_ref_prefix (OstreeRepo         *repo,
                     int                 bootversion,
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -353,34 +354,18 @@ cleanup_ref_prefix (OstreeRepo         *repo,
                     GCancellable       *cancellable,
                     GError            **error)
 {
```
jlebon: Maybe a small note at the top of this function that we're called assuming an active transaction is underway?

src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -353,34 +354,18 @@ cleanup_ref_prefix (OstreeRepo         *repo,
...
-  if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
-    goto out;
+    return FALSE;
 
-  g_hash_table_iter_init (&hashiter, refs);
-  while (g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue))
+  GLNX_HASH_TABLE_FOREACH (refs, const char *, ref)
     {
-      const char *ref = hashkey;
       ostree_repo_transaction_set_refspec (repo, ref, NULL);
```
jlebon: Actually, can't we just use `ostree_repo_set_ref_immediate()` for these?

cgwalters: Yep, done.  Also added tests ⬇️

src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -353,34 +354,19 @@ cleanup_ref_prefix (OstreeRepo         *repo,
...
-    goto out;
-
-  ret = TRUE;
- out:
-  ostree_repo_abort_transaction (repo, cancellable, NULL);
-  return ret;
+  return TRUE;
 }
 
 static gboolean
```
jlebon: Am I dreaming, or did you have nice comments on top of these functions briefly explaining what they did in a previous version of this PR? I really liked them. :)

jlebon:
Nice! @rh-atomic-bot r+ 2f6ddaa

cgwalters:
Rebased 🏄

src/libostree/ostree-sysroot-cleanup.c
cgwalters: Not entirely sure, but I added a comment on each function for this updated PR

jlebon:
@rh-atomic-bot r+ b7c3d02

PR #1135, opened 2017-09-06
lib/gpg: Use nicer helper for gpg error messages

cgwalters:
The vast majority of invocations of `ot_gpgme_error_to_gio_error()` were paired
with `g_prefix_error()`; let's combine them for the same reason we do
`glnx_throw_errno_prefix()`. For the few cases that don't we might as well add
some prefix.

I also changed it to `return FALSE` in prep for more style porting.

### DIFF
--- a/src/libostree/ostree-gpg-verifier.c
+++ b/src/libostree/ostree-gpg-verifier.c
@@ -187,14 +187,14 @@ _ostree_gpg_verifier_check_signature (OstreeGpgVerifier  *self,
           gpg_error = gpgme_data_new_from_fd (&kdata, fd);
           if (gpg_error != GPG_ERR_NO_ERROR)
             {
-              ot_gpgme_error_to_gio_error (gpg_error, error);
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-gpg-verifier.c
```diff
@@ -187,14 +187,14 @@ _ostree_gpg_verifier_check_signature (OstreeGpgVerifier  *self,
...
-              ot_gpgme_error_to_gio_error (gpg_error, error);
+              ot_gpgme_throw (gpg_error, error, "Loading data from fd %i", fd);
               goto out;
             }
 
           gpg_error = gpgme_op_import (result->context, kdata);
           if (gpg_error != GPG_ERR_NO_ERROR)
             {
-              ot_gpgme_error_to_gio_error (gpg_error, error);
+              ot_gpgme_throw (gpg_error, error, "import");
```
jlebon: Nit: let's add a more descriptive prefix in line with the other ones? E.g. `Unable to import key` ?

jlebon:
@rh-atomic-bot r+ a971318

PR #1140, opened 2017-09-06
lib/repo: Add apidoc for repo properties

cgwalters:
However, they weren't showing up in the output HTML and I have
no idea why; I looked at what we're doing and it looks close enough
to what's going on in `GDBusConnection` that I was using as a reference.
I'm not going to spend a lot of time to debug it right now.


jlebon:
@rh-atomic-bot r+ 6bb8465

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #1141, opened 2017-09-06
tests/rofiles-fuse: Add tests for chmod/chown

cgwalters:
Prep for https://github.com/ostreedev/ostree/pull/1137 where
we were incorrectly handling `chown()` on symlinks.


jlebon:
@rh-atomic-bot r+ ec4716a

PR #1142, opened 2017-09-06
tree-wide: Use helpers for unlinkat()

cgwalters:
We have `ot_ensure_unlinked_at()` for the "ignore ENOENT" case, and
`glnx_unlinkat()` otherwise. Port all in-tree callers to one or the other as
appropriate.

Just noticed an unprefixed error in the refs case and decided to do a tree-wide
check.


jlebon:
@rh-atomic-bot r+ 4858918

Issue #1139, opened 2017-09-06
broken cookie path lookup


cgwalters:
Saw this in a strace:

```
[pid  4354] stat("/srv/repomirror/repo-corrupt-mirror/file:///srv/repomirror/repo-corrupt-bare.cookies.txt", 0x7ffdf9cd29f0) = -1 ENOENT (No such file or directory)
```

PR #1145, opened 2017-09-07
lib/pull: Only look for cookie files for non-local remotes

cgwalters:
Just noticed this while reading an strace.

Closes: https://github.com/ostreedev/ostree/issues/1139


jlebon:
@rh-atomic-bot r+ 453b258

PR #1146, opened 2017-09-07
lib/pull: A bit of new style porting

cgwalters:
A lof of the functions here are async and have nontrivial exits, but these ones
are all sync were straightforward ports.

Not prep for anything, just chipping away at porting.


jlebon:
@rh-atomic-bot r+ 900dfd6

PR #1147, opened 2017-09-07
lib/commit: Add some error prefixing for txn commit/tmpdir

cgwalters:
To help debug this: https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2017-September/msg00001.html

Currently we just get: `error: Commit: unlinkat: Directory not empty`


jlebon:
@rh-atomic-bot r+ 177a455

PR #1148, opened 2017-09-07
lib/sysroot: Use fd-relative acccess for bootversion cleanup

cgwalters:
I noticed this was an easy change.

### DIFF
--- a/src/libostree/ostree-sysroot-cleanup.c
+++ b/src/libostree/ostree-sysroot-cleanup.c
@@ -210,35 +210,29 @@ cleanup_other_bootversions (OstreeSysroot       *self,
                             GCancellable        *cancellable,
                             GError             **error)
 {
-  int cleanup_bootversion = self->bootversion == 0 ? 1 : 0;
-  int cleanup_subbootversion = self->subbootversion == 0 ? 1 : 0;
+  const int cleanup_bootversion = self->bootversion == 0 ? 1 : 0;
+  const int cleanup_subbootversion = self->subbootversion == 0 ? 1 : 0;
+  /* Reusable buffer for path */
+  g_autoptr(GString) buf = g_string_new ("");
 
-  g_autoptr(GFile) cleanup_boot_dir = NULL;
-  cleanup_boot_dir = ot_gfile_resolve_path_printf (self->path, "boot/loader.%d", cleanup_bootversion);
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -210,35 +210,29 @@ cleanup_other_bootversions (OstreeSysroot       *self,
...
+  g_string_truncate (buf, 0); g_string_append_printf (buf, "ostree/boot.%d.1", cleanup_bootversion);
+  if (!glnx_shutil_rm_rf_at (self->sysroot_fd, buf->str, cancellable, error))
     return FALSE;
-  g_clear_object (&cleanup_boot_dir);
 
-  cleanup_boot_dir = ot_gfile_resolve_path_printf (self->path, "ostree/boot.%d.%d", self->bootversion,
-                                                   cleanup_subbootversion);
-  if (!glnx_shutil_rm_rf_at (AT_FDCWD, gs_file_get_path_cached (cleanup_boot_dir), cancellable, error))
+  /* And finally the other subbootversion */
+  g_string_truncate (buf, 0); g_string_append_printf (buf, "ostree/boot.%d.%d", self->bootversion, cleanup_subbootversion);
```
peterbaouoft: hey, trying to learn how to do a review slowly, I am still not very familiar with the code base, so.. might ask questions often, hopefully you don't mind :). 

So I have a few doubts.. it will be great if you have time to answer them :-) , but this is kinda low priority, just want to familiarize with more code base,
1: what might be the reason that we have two statements in the single line here? 

2: if I am understanding this correctly, this patch seems to avoid the usage of using current working directory path (AT_FDWCD) , and will use relative path based on the sysroot_fd for cleaning up? 
 
Thanks in advance! :) And if the questions is worded poorly, let me know :).

cgwalters: Great, thanks for diving in!  You're absolutely right on point `1:`...it's unusual.  I went back and forth on this, and am fine changing it.  The reason I did it initially is to make the function a bit more compact and less "visually" duplicative so one focuses more on the string.

For point `2:` - yep, that's right!

peterbaouoft: Thanks for answering :)

cgwalters:

peterbaouoft:
@rh-atomic-bot r+ 3297dbb

PR #1149, opened 2017-09-07
tests: Check "refs -c PREFIX" behavior

mwleeds:
This commit adds tests to check the behavior of "refs -c PREFIX", where
prefix is interpreted as a collection ID.

@cgwalters this is the test that would have failed without 5f8e339e45314174dd49221fcf09197e8d1e7ba4.


cgwalters:
Nice, thanks for following up on this!

@rh-atomic-bot r+ 101e3b3

PR #1150, opened 2017-09-07
tree-wide: Add error prefixing for most remaining syscalls

cgwalters:
There were some important ones there like a random `syncfs()`. The remaining
users are mostly blocked on the "fstatat enoent" case, I'll wait to port those.


jlebon:
@rh-atomic-bot r+ 65c972e

Issue #1133, opened 2017-09-05
Cannot use remotes_config_dir without sysroot


dbnicholson:
I have a repository manager that runs on a server and pulls from build machines into a set of central repos. In order to pull, ostree needs a remote. Rather than adding every possible remote to every repo (both cluttering the repo config file and leaking internal details), I use a remotes configuration directory. Commit e0346c1 broke this since now the `remotes-config-dir` property is only respected if the `sysroot-path` property is set. That's not set since these are not system repos.

cgwalters:
Ah.  Hmm, makes sense I guess; I'd just seen the `remotes-config-dir` as part of the weird way we propagate bits from the sysroot to repo levels.

dbnicholson:
Right. I would say the the remotes config dir is typically used in the system repo scope, but it was helpful to use for a normal repo.

Alternatively, if you could pull without specifying a remote, I'd use that. The way I use the remotes configuration directory right now is that I just populate a single fake remote without a URL and then always override the URL when pulling and fetching the summary. This allows me to pull from semi-arbitrary URLs without managing remote configuration.

dbnicholson:
Also, `g_file_resolve_relative_path (self->sysroot_dir, self->remotes_config_dir);` is broken if `remotes_config_dir` is absolute and `sysroot_dir` is set (and you actually want to use the alternate sysroot).

dbnicholson:
This also breaks flatpak's use of `remotes-config-dir` - https://github.com/flatpak/flatpak/blob/master/common/flatpak-dir.c#L762. That was actually the reason this was originally added for #247.

cgwalters:
Ugh.  OK, I'll look at this.

dbnicholson:
Hmm, if I fix this by making `remotes-config-dir` respected again, it will certainly break flatpak (which is how I ended up here in the first place). Before e0346c1, only reading the config would respect `remotes-config-dir`. Adding remotes would only use it for a system repo (and then it was hardcoded to /etc/ostree/remotes.d). After e0346c1, adding remotes respects `remotes-config-dir` but only for system repos.

[...]

Issue #1134, opened 2017-09-05
Use of /etc/ostree/remotes.d should be explicit


dbnicholson:
Currently there's a lot of magic to decide if remotes should be added in /etc/ostree/remotes.d. While there are good reasons to put the remote configuration there, we'd like to avoid it for a couple reasons:

1. It's not coherent with `ostree_repo_write_config`, which is used by the `ostree config` CLI as well as `flatpak remote-modify`. You can easily end up with a broken configuration where you have a remote defined in both /etc/ostree/remotes.d and /ostree/repo/config. See also #1111 and https://bugzilla.gnome.org/show_bug.cgi?id=753373.

2. We get different behavior when we're constructing without sysroot set vs. modifying from within the system.

3. We really just want to define the remotes in /ostree/repo/config sometimes and there's currently no way to opt out of this behavior.

I think it would be better to make this behavior explicitly opt in since the API is inconsistent with its usage. You can easily just write a file in /etc/ostree/remotes.d if you like, but IMO it's much safer to have ostree itself not try to manage both locations.

dbnicholson:
I added a repo config option in #1155 which would allow us to opt out of adding remotes in the config dir in our shared ostree/flatpak system repo.

PR #1151, opened 2017-09-07
repo: Fix non-system remotes-config-dir usage

dbnicholson:
Before commit e0346c1, a non-system repo could specify
remotes-config-dir and have remotes read from there. However, adding
remotes would only be done in the config dir for a system repo. Restore
that by respecting remotes-config-dir when no sysroot is found and
adding back the ostree_repo_is_system() check when adding remotes.

Closes: #1133

### DIFF
--- a/tests/test-remotes-config-dir.js
+++ b/tests/test-remotes-config-dir.js
@@ -1,0 +1,73 @@
+#!/usr/bin/env gjs
+//
+// Copyright (C) 2013 Colin Walters <walters@verbum.org>
+// Copyright (C) 2017 Dan Nicholson <nicholson@endlessm.com>
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; either
+// version 2 of the License, or (at your option) any later version.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the
+// Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+// Boston, MA 02111-1307, USA.
+
+const GLib = imports.gi.GLib;
+const Gio = imports.gi.Gio;
+const OSTree = imports.gi.OSTree;
+
+function assertEquals(a, b) {
+    if (a != b)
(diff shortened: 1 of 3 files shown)


cgwalters:
I've only briefly thought about this so far...your patch sounds and looks reasonable, but I can't say I've fully worked out the intersection between all of the different cases.

For example, we talked about your repo tool (is that FOSS and posted publicly?).  Does this patch fix that or were you going to change it?

You outlined the implications for flatpak in the issue...are we avoiding changing its behavior here via  the `ostree_repo_is_system (self)` check?  (But this is where I get fuzzy around the implications for the EndlessOS case...do you want flatpak to read `/etc/ostree/remotes.d` or `/etc/flatpak/remotes.d` or both or neither?)

dbnicholson:
Yeah, this remotes.d saga has snowballed more than I thought. :)

1. This does fix my repo manager case. Now `get_remotes_d_dir` returns the directory I pass through `remotes-config-dir`. This is used when reading or deleting remotes. Since the tool does not manage system repos, it wouldn't add remotes in the config dir, although it never adds remotes, anyways.

2. For flatpak (in the standard non-system repo case), this restores the previous semantics. As mentioned above, the configured directory will be used for reading and deleting remotes. However, by adding back the `ostree_repo_is_system` check in the adding case, it will not add remotes in the config dir. That's important because the way flatpak adds or modifies remotes currently is to add a minimal remote (just name and URL), and then add all the options via `ostree_repo_write_config`. If a new remote gets added, it would go in the config dir via `remote_change`, and then `write_config` would add a duplicate in the config file. This is a hard error as soon as the repo is opened and the remote config is reloaded. This is how I got started down this issue and only just noticed it would break case 1 for my by inspection.

3. Our flatpak as system ostree repo is still broken as a corollary to 2 since `ostree_repo_is_system` will be true. But even ignoring flatpak, there are bugs with remotes being added in the config dir since there's no API to modify a remote and `write_config` would break that. So, now I'm looking at adding a repo option to disable adding remotes in the config dir for #1134. I'd then add that to our new deployments and write some update script that sets it for existing users. Until there's a coherent way to modify remotes in the config dir, I just want to avoid that completely.

tests/test-remotes-config-dir.js
```diff
@@ -0,0 +1,73 @@
...
+repo.create(OSTree.RepoMode.ARCHIVE_Z2, null);
+repo.open(null);
+
+// See if the remotes.d remote exists
+let remotes = repo.remote_list()
+assertNotEquals(remotes.indexOf('foo'), -1);
+
+print("ok read-remotes-config-dir");
+
+// Adding a remote should not go in the remotes.d dir
```
cgwalters: (for now...)

cgwalters:
@rh-atomic-bot r+ 46f5289

PR #1153, opened 2017-09-08
lib/grub2: Port some to new code style

cgwalters:
I resisted trying to do anything invasive here like fd-relative porting as our
coverage is weak. But this was all straightforward porting to decl-after-stmt
style.


jlebon:
@rh-atomic-bot r+ 9724651

PR #1154, opened 2017-09-08
lib/syslinux: Port to new code style

cgwalters:
There was only one tricky bit here around the ownership of the lines; I made use
of `g_steal_pointer()` to consistently track ownership, and converted to a `for`
loop while still preserving the loop logic around the last entry.

### DIFF
--- a/src/libostree/ostree-bootloader-syslinux.c
+++ b/src/libostree/ostree-bootloader-syslinux.c
@@ -59,44 +59,33 @@ _ostree_bootloader_syslinux_get_name (OstreeBootloader *bootloader)
 }
 
 static gboolean
-append_config_from_boostree_loader_entries (OstreeBootloaderSyslinux  *self,
-                                        gboolean               regenerate_default,
-                                        int                    bootversion,
-                                        GPtrArray             *new_lines,
-                                        GCancellable          *cancellable,
-                                        GError               **error)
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-bootloader-syslinux.c
```diff
@@ -119,61 +106,49 @@ _ostree_bootloader_syslinux_write_config (OstreeBootloader          *bootloader,
...
    * more time at the end where line = NULL to ensure we finish off
    * processing the last LABEL.
    */
-  iter = lines;
-  while (TRUE)
+  for (char **iter = lines; iter; iter++)
     {
-      char *line = *iter;
+      /* Take ownership of the string element */
+      g_autofree char *line = g_steal_pointer (iter);
```
jlebon: Hmm, there's a subtle issue here (which looks like it existed even before). `g_strfreev` stops at the first `NULL` it finds. If we break out of this loop earlier than expected, the remaining strings will never be freed, right? It seems like it's simpler to just `g_strdup` in that case?

cgwalters: Yeah, you're right.  Updated, also did some more re-indent while there ⬇️

jlebon:
@rh-atomic-bot r+ 4b2e738

PR #1155, opened 2017-09-08
repo: Add add-remotes-config-dir option

dbnicholson:
This option allows a repo to explicitly opt out of adding new remotes in
a remotes configuration directory. This currently defaults to true for
system repos and false for non-system repos to maintain legacy behavior
that non-system repos don't add remotes in a configuration directory.
That would be problematic for flatpak, which specifies a remotes config
dir but adds remotes in ways that are incompatible with it.

So, what this really does is allow system repos to control whether they
want to add remotes in the config dir or not. That's important if your
flatpak repo is the system repo like at Endless.

Closes: #1134

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -122,6 +122,27 @@ Boston, MA 02111-1307, USA.
         keep free. The default value is 3.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>add-remotes-config-dir</varname></term>
+        <listitem>
+          <para>
+            Boolean value controlling whether new remotes will be added
+            in the remotes configuration directory. Defaults to
+            <literal>true</literal> for system ostree repositories. When
+            this is <literal>false</literal>, remotes will be added in
+            the repository's <filename>config</filename> file.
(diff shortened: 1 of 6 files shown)


man/ostree.repo-config.xml
```diff
@@ -122,6 +122,27 @@ Boston, MA 02111-1307, USA.
         keep free. The default value is 3.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>add-remotes-config-dir</varname></term>
```
cgwalters: Minor bikeshed...what do you think about `write-remotes-config-dir`?

dbnicholson: Oh, that would be fine. I meant to add in the PR that I was open to suggestions on the name.

cgwalters:
No strong opinion here on the name, let's keep as is then.  Thanks for the patch!

@rh-atomic-bot r+ 4664c9f

dbnicholson:
Heh, I was going to change it this morning. I just saw the email over the weekend but wasn't going to be working on it until today. I actually liked your name better.

PR #1156, opened 2017-09-08
Change checkout disjoint to merge identical files

cgwalters:


cgwalters:
Posting for discussion; needs tests etc.

src/libostree/ostree-repo-checkout.c
```diff
@@ -329,25 +358,54 @@ checkout_file_hardlink (OstreeRepo                          *self,
...
+          struct stat src_stbuf;
+          if (!glnx_fstatat (srcfd, loose_path, &src_stbuf,
+                             AT_SYMLINK_NOFOLLOW, error))
+            return FALSE;
+          struct stat dest_stbuf;
+          if (!glnx_fstatat (destination_dfd, destination_name, &dest_stbuf,
+                             AT_SYMLINK_NOFOLLOW, error))
+            return FALSE;
+          const gboolean is_identical =
+            (src_stbuf.st_dev == dest_stbuf.st_dev &&
```
peterbaouoft: Hey, this line is checking for if file from the old src and the new destination is on the same device and having same file serial number right? Then if both are same, we can conclude two files are identical? But, it also seems a bit weird to me because to me identical also kinda applies when two files share the same name and exact same content. 

I am not sure if this case will happen in ostree checkout case, but let's say if we were checking out `/usr/share/xz/file` to the ostree and there is a file already existing at the same location (`/usr/share/xz/file`) , with the same content, will this implementation still catches that? 

Again, I am not very familiar with the codebase, so. this might be a misinterpretation of the code, sorry if that happens, and thanks in advance!  :)

jlebon: Hmm, I guess we could also just checksum the file to determine if it's identical. Though I'm not sure what the chances of them having the same content + file info but somehow *not* being from a checkout. (E.g. ostree canonicalizes timestamps to epoch, which would be odd in any other situations).

peterbaouoft: ah it seems my question was poorly worded :(, apologies. 

So, what I actually meant was that do we also account for the case that two files have same file content,path , but different st_ino. (Not sure if that is possible to happen for checking out files).

But if that situation exists, should we also consider those two files to be identical?

cgwalters: If checkouts aren't hardlinked, a lot (not all, but most) of the value of ostree drops away.  So I'm not sure it's worth thinking too much about the case where we're not hardlinking.  Right?

cgwalters: Latest fixup ⬇️  makes the non-hardlinking disjoint union path an explicit error.

peterbaouoft: Ah I believe I thought it wrong before. It makes more sense now. Thanks for the clarification :)

cgwalters:
Lifting WIP.  It's tempting to rename this...`UNION_MERGE_IDENTICAL`?  Dunno.

src/libostree/ostree-repo-checkout.c
```diff
@@ -218,23 +216,51 @@ create_file_copy_from_input_at (OstreeRepo     *repo,
...
+                                              cancellable, error);
+                    if (!dest_target)
+                      return FALSE;
+                    /* In theory we could also compare xattrs...but eh */
+                    if (g_str_equal (dest_target, target))
+                      return TRUE;
+                  }
+                errno = EEXIST;
+                return glnx_throw_errno_prefix (error, "symlinkat");
+              }
```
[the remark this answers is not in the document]
cgwalters: If a union match isn't exhaustive, GCC will make that a warning:

```
src/libostree/ostree-repo-checkout.c: In function ‘checkout_file_hardlink’:
src/libostree/ostree-repo-checkout.c:371:7: warning: enumeration value ‘OSTREE_REPO_CHECKOUT_OVERWRITE_LAUNCH_NETHACK’ not handled in switch [-Wswitch]
```

Which kind of forces auditing all callers when adding a case.  If you have a `default:` then one needs to hunt manually via `git grep` etc.

Hm, we should make that fatal.  I'll submit a separate PR.

src/libostree/ostree-repo-checkout.c
```diff
@@ -329,25 +365,56 @@ checkout_file_hardlink (OstreeRepo                          *self,
...
+                               AT_SYMLINK_NOFOLLOW, error))
+              return FALSE;
+            const gboolean is_identical =
+              (src_stbuf.st_dev == dest_stbuf.st_dev &&
+               src_stbuf.st_ino == dest_stbuf.st_ino);
+            if (is_identical)
+              ret_result = HARDLINK_RESULT_SKIP_EXISTED;
+            else
+              return glnx_throw_errno_prefix (error, "Hardlinking %s to %s", loose_path, destination_name);
+          }
```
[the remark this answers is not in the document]
cgwalters: `g_assert_not_reached()` will always terminate the process; it invokes `g_assertion_message_expr()` which is annotated as `G_GNUC_NORETURN` so the compiler knows this.  So having a `break` is optional - more of a style issue.  

I think i'd vote though for always having `break`.  Will add one.

src/ostree/ot-builtin-checkout.c
```diff
@@ -136,7 +136,15 @@ process_one_checkout (OstreeRepo           *repo,
       else if (opt_union_add)
         options.overwrite_mode = OSTREE_REPO_CHECKOUT_OVERWRITE_ADD_FILES;
       else if (opt_disjoint_union)
-        options.overwrite_mode = OSTREE_REPO_CHECKOUT_OVERWRITE_DISJOINT_UNION_FILES;
+        {
+          if (!opt_require_hardlinks)
```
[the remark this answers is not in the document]
cgwalters: If we didn't require it, we'd have to change `create_file_copy_from_input_at()` to also handle the merge (right now we abort), and it'd be harder since we'd actually need to checksum the files rather than comparing by device+inode.  Since I don't know of a good use case for that, I'd prefer not supporting it and only do so if we have to.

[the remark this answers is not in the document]
peterbaouoft: ah I see, it makes sense. Thanks for the explanation :)

cgwalters:
Before we merge this though...let's think again about names, since the sets are no longer disjoint (right?).  How about `UNION_IDENTICAL` ?

cgwalters:
@jlebon You came up with the last name, WDYT?

cgwalters:
For reference AFAICS the librpm code is here:

https://github.com/rpm-software-management/rpm/blob/e3cd2bc85e0578f158d14e6f9624eb955c32543b/lib/rpmfi.c#L921

jlebon:
Minor nit: there's a typo in the commit message: `s/exmaple/example/`!

Heh, yeah I don't think there's any short name that will fully describe the new behaviour. `UNION_IDENTICAL` seems appropriate to me! (We should also update the docstring for it).

cgwalters:
@peterbaouoft is `UNION_IDENTICAL` OK by you?

peterbaouoft:
ah Forgot to comment it in the review, and yup I am fine with it. Thanks for fixing it :)

cgwalters:
Updated ⬆️ 

Besides the mechanical rename, I also linked to the librpm code in a comment and changed the `mkdir` handling to also do an exhaustive `switch` for consistency.

jlebon:
@rh-atomic-bot r+ f449d93

peterbaouoft:
Hey, this patch will also affect rpm-ostree side(I think) https://github.com/projectatomic/rpm-ostree/pull/974, I can file a patch for it, is that fine? The patch was mainly to change in the rpm-ostree side  from `--disjoint union` to `--union-identical` and `OSTREE_REPO_CHECKOUT_OVERWRITE_DISJOINT_UNION_FILES` to `OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_IDENTICAL` . Thanks :)

cgwalters:
Yep, please do!

cgwalters:
@rh-atomic-bot retry

PR #1158, opened 2017-09-11
ostree/summary: Generate an ostree-metadata ref when updating summary

pwithnall:
This is the new way of publishing repository metadata, rather than as
additional-metadata in the summary file. The use of an ostree-metadata
ref means that the metadata from multiple upstream collections is not
conflated when doing P2P mirroring of many repositories.

The new ref is only generated if the repository has a collection ID set.
The old summary file continues to be generated for backwards
compatibility (and because it continues to be the canonical ref →
checksum map for the repository).

The new code is only used if configured with --enable-experimental-api.

Includes unit tests.

### DIFF
--- a/src/ostree/ot-builtin-summary.c
+++ b/src/ostree/ot-builtin-summary.c
@@ -83,7 +83,6 @@ build_additional_metadata (const char * const  *args,
 gboolean
 ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError **error)
 {
-  gboolean ret = FALSE;
   g_autoptr(GOptionContext) context = NULL;
   g_autoptr(OstreeRepo) repo = NULL;
   OstreeDumpFlags flags = OSTREE_DUMP_NONE;
@@ -91,24 +90,24 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
   context = g_option_context_new ("Manage summary metadata");
 
   if (!ostree_option_context_parse (context, options, &argc, &argv, OSTREE_BUILTIN_FLAG_NONE, &repo, cancellable, error))
-    goto out;
+    return FALSE;
 
   if (opt_update)
     {
       g_autoptr(GVariant) additional_metadata = NULL;
 
       if (!ostree_ensure_repo_writable (repo, error))
-        goto out;
+        return FALSE;
 
       if (opt_metadata != NULL)
         {
           additional_metadata = build_additional_metadata ((const char * const *) opt_metadata, error);
           if (additional_metadata == NULL)
-            goto out;
+            return FALSE;
         }
 
       if (!ostree_repo_regenerate_summary (repo, additional_metadata, cancellable, error))
-        goto out;
+        return FALSE;
 
       if (opt_key_ids)
         {
@@ -117,7 +116,7 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
                                                       opt_gpg_homedir,
                                                       cancellable,
                                                       error))
-            goto out;
+            return FALSE;
         }
     }
   else if (opt_view)
@@ -129,7 +128,7 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
 
       summary_data = ot_file_mapat_bytes (repo->repo_dir_fd, "summary", error);
       if (!summary_data)
-        goto out;
+        return FALSE;
 
       ot_dump_summary_bytes (summary_data, flags);
     }
@@ -137,10 +136,8 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
     {
       g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
                    "No option specified; use -u to update summary");
-      goto out;
+      return FALSE;
     }
 
-  ret = TRUE;
- out:
-  return ret;
+  return TRUE;
 }
--- a/src/ostree/ot-builtin-summary.c
+++ b/src/ostree/ot-builtin-summary.c
@@ -106,6 +106,97 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
             return FALSE;
         }
 
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+      const char *collection_id = ostree_repo_get_collection_id (repo);
+#else  /* if !OSTREE_ENABLE_EXPERIMENTAL_API */
+      const char *collection_id = NULL;
+#endif  /* OSTREE_ENABLE_EXPERIMENTAL_API */
+
+      /* Write out a new metadata commit for the repository. */
+      if (collection_id != NULL)
+        {
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+          OstreeCollectionRef collection_ref = { (gchar *) collection_id, (gchar *) OSTREE_REPO_METADATA_REF };
+          g_autofree char *old_ostree_metadata_checksum = NULL;
+          g_autofree gchar *new_ostree_metadata_checksum = NULL;
+          g_autoptr(OstreeMutableTree) mtree = NULL;
+          g_autoptr(OstreeRepoFile) repo_file = NULL;
+          g_autoptr(GVariantDict) new_summary_commit_dict = NULL;
+          g_autoptr(GVariant) new_summary_commit = NULL;
+
+          if (!ostree_repo_resolve_rev (repo, OSTREE_REPO_METADATA_REF,
+                                        TRUE, &old_ostree_metadata_checksum, error))
+            return FALSE;
+
+          /* Add bindings to the metadata. */
+          new_summary_commit_dict = g_variant_dict_new (additional_metadata);
+          g_variant_dict_insert (new_summary_commit_dict, OSTREE_COMMIT_META_KEY_COLLECTION_BINDING,
+                                 "s", collection_ref.collection_id);
+          g_variant_dict_insert_value (new_summary_commit_dict, OSTREE_COMMIT_META_KEY_REF_BINDING,
+                                       g_variant_new_strv ((const gchar * const *) &collection_ref.ref_name, 1));
+          new_summary_commit = g_variant_dict_end (new_summary_commit_dict);
+
+          if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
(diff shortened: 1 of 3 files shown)


pwithnall:
The original introduction of `ostree-metadata` was discussed in PR #946.

pwithnall:
bot, retest this please

pwithnall:
@rh-atomic-bot, retest this please

src/ostree/ot-builtin-summary.c
```diff
@@ -107,6 +107,80 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
...
+      if (collection_id != NULL)
+        {
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+          OstreeCollectionRef collection_ref = { (gchar *) collection_id, (gchar *) OSTREE_REPO_METADATA_REF };
+          g_autofree char *old_ostree_metadata_checksum = NULL;
+          g_autofree gchar *new_ostree_metadata_checksum = NULL;
+          g_autoptr(OstreeMutableTree) mtree = NULL;
+          g_autoptr(OstreeRepoFile) repo_file = NULL;
+          g_autoptr(GVariantDict) new_summary_commit_dict = NULL;
+          g_autoptr(GVariant) new_summary_commit = NULL;
```
cgwalters: We prefer declare-and-initialize for new code; doesn't have to be done in this patch but someone will probably do it later if not.

src/ostree/ot-builtin-summary.c
```diff
@@ -107,6 +107,80 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
...
+                                 "s", collection_ref.collection_id);
+          g_variant_dict_insert_value (new_summary_commit_dict, OSTREE_COMMIT_META_KEY_REF_BINDING,
+                                       g_variant_new_strv ((const gchar * const *) &collection_ref.ref_name, 1));
+          new_summary_commit = g_variant_dict_end (new_summary_commit_dict);
+
+          if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
+            return FALSE;
+
+          mtree = ostree_mutable_tree_new ();
+          if (!ostree_repo_write_dfd_to_mtree (repo, AT_FDCWD, ".", mtree, NULL, NULL, error))
```
cgwalters: Hmm?  I don't think we should be relying on the process' cwd here, right?  It looks like that cwd is the entire unit test directory at least from the tests.

Are you intending to include *any* content in the metadata ref?  If not, the libarchive code has some "create empty directory" code.

src/ostree/ot-builtin-summary.c
pwithnall: This code was copied from the almost-identical implementation I did in flatpak; so I’d prefer not to rewrite it on import.

src/ostree/ot-builtin-summary.c
pwithnall: Oops. The intention is to not include any content in the commit, correct. I can’t see anything immediately obvious for this in the libarchive code. Which code do you mean, exactly?

I need to fix flatpak to not use `AT_FDCWD, "."` too, once this is done.

cgwalters: It's this bit:

```
  /* If we didn't import anything at all, and autocreation of parents
   * is enabled, automatically create a root directory.  This is
   * useful primarily when importing Docker image layers, which can
   * just be metadata.
   */
  if (opts->autocreate_parents &&
      ostree_mutable_tree_get_metadata_checksum (mtree) == NULL)
    {
      glnx_unref_object GFileInfo *fi = g_file_info_new ();
      g_file_info_set_attribute_uint32 (fi, "unix::uid", 0);
      g_file_info_set_attribute_uint32 (fi, "unix::gid", 0);
      g_file_info_set_attribute_uint32 (fi, "unix::mode", DEFAULT_DIRMODE);

      g_autoptr(GFileInfo) mfi = NULL;
      (void)_ostree_repo_commit_modifier_apply (self, modifier, "/",
                                                fi, &mfi);

      if (!aic_ensure_parent_dir_with_file_info (&aictx, mtree, "/", mfi, NULL,
                                                 cancellable, error))
        goto out;
    }

```

pwithnall: Great, thanks. I’ve updated the patch and added a test.

pwithnall: I’ll fix flatpak tomorrow; need to head off now.

src/ostree/ot-builtin-summary.c
```diff
@@ -107,6 +107,80 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
...
+            return FALSE;
+          if (!ostree_repo_write_mtree (repo, mtree, (GFile **) &repo_file, NULL, error))
+            return FALSE;
+
+          if (!ostree_repo_write_commit (repo, old_ostree_metadata_checksum,
```
cgwalters: What's the rationale behind the commit having a parent?  In what cases would one want old metadata?

I guess with this administrators could do `ostree --repo=repo prune --only-branch=ostree-metadata --depth=0` when they wanted, but still.

src/ostree/ot-builtin-summary.c
```diff
@@ -107,6 +107,80 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
             return FALSE;
         }
 
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+      const char *collection_id = ostree_repo_get_collection_id (repo);
+#else  /* if !OSTREE_ENABLE_EXPERIMENTAL_API */
+      const char *collection_id = NULL;
+#endif  /* OSTREE_ENABLE_EXPERIMENTAL_API */
+
+      /* Write out a new metadata commit for the repository. */
```
cgwalters: Can you explain a bit *what* metadata is being included here?  It looks it's basically a stub, but callers could use `--add-metadata`?

cgwalters:
Is it OK if this waits till the next release cycle?

src/ostree/ot-builtin-summary.c
pwithnall: `additional_metadata` is passed into the vardict for `new_summary_commit`, so the stored metadata is identical to that in the `summary` file (plus the collection/ref binding keys). By default, that is indeed nothing, unless the user provides an `--add-metadata` option.

pwithnall:
> Is it OK if this waits till the next release cycle?

Sure, although we might carry it as a patch in our OSTree package until then, so if you have any architectural issues to raise with it, please do so now. I don’t want our OSTree package to end up diverging from upstream.

src/ostree/ot-builtin-summary.c
cgwalters: Hmm, but if the intention is to get rid of the summary file down the line, don't we want the list of refs/deltas?

cgwalters:
I'm going to mark this as WIP since it feels like it needs a bit of design.  Are we agreed that this replaces the whole summary file down the line?  Maybe we'll need infrastructure to re-synthesize something that looks summary-file like for the pull code?

One thing I'd hope we can achieve here is split up the metadata into separate chunks to scale better.

src/ostree/ot-builtin-summary.c
pwithnall: The intention is not to get rid of the summary file down the line, since it remains a useful cache for the list of refs/deltas. The `ostree-metadata` ref is only replacing the *additional metadata* from the `summary`.

pwithnall:
> Are we agreed that this replaces the whole summary file down the line?

Nope, the intention is that this only replaces the *additional metadata* in the `summary` file. The `summary` file remains present, unsigned, as a list of the refs.

The deltas are currently stored in the additional metadata map (right? I haven’t checked), so I guess they end up moving to `ostree-metadata`. I’m more on the fence about that one, because the list of deltas is content which could vary between a repository and a mirror of that repository; just like the list of refs. Maybe I need to look into that aspect of it further.

pwithnall:
(Would it help to have a call about this?)

cgwalters:
One thing that occurred to me from our discussion is that if we want to deprecate the summary file, perhaps we need an `ostree-summary-metadata` ref which is distinct from the `ostree-metadata` which sounds more static?

pwithnall:
> One thing that occurred to me from our discussion is that if we want to deprecate the summary file, perhaps we need an `ostree-summary-metadata` ref which is distinct from the `ostree-metadata` which sounds more static?

What are you suggesting this `ostree-summary-metadata` ref stores? The list of other refs in the repository? Currently, I think I’d rather keep the `summary` file for that. However, I don’t want to get too sucked in to discussing the long-term future of the `summary` file in *this* PR, given that it’s still a bit nebulous. Open a new issue and we can plan there (and pull in Alex and anyone else interested)?

pwithnall:
Rebased against master and squashed in the fixups. Ready for a second round of review.

src/ostree/ot-builtin-summary.c
cgwalters: @pwithnall Did you have an opinion on this issue?  (Ah, I see the unit tests rely on it)

pwithnall: The unit tests don’t rely on it inasmuch as they test that the code implements it.

The rationale for metadata commits having a parent is that it allows the metadata history of the repository to be represented, at fairly low cost. As you say, admins can always prune it later if they want.

src/ostree/ot-builtin-summary.c
```diff
@@ -106,6 +106,97 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
...
+            return FALSE;
+
+          if (opt_key_ids != NULL)
+            {
+              for (const char * const *iter = (const char * const *) opt_key_ids;
+                   iter != NULL && *iter != NULL; iter++)
+                {
+                  const char *key_id = *iter;
+
+                  if (!ostree_repo_sign_commit (repo,
```
cgwalters: Generating, signing, and committing in one go makes doing signing asynchronously harder.  At least in Fedora releng does signing async.  But this is something we can enhance later.

cgwalters:
@rh-atomic-bot r+ f126610

src/ostree/ot-builtin-summary.c
pwithnall: If `opt_key_ids == NULL` then signing won’t happen as part of this utility call.

PR #1159, opened 2017-09-12
repo: Ensure new config doesn't set remotes in separate file

dbnicholson:
If the new configuration passed to ostree_write_config () tries to
update options for a remote defined in a separate config file, return an
error. Without this, the full configuration would contain duplicate
remote specifications, which would raise an error the next time the repo
is opened.

Something I thought about when doing the other recent remotes work.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -936,21 +936,17 @@ ostree_repo_write_config (OstreeRepo *self,
   /* Ensure that any remotes in the new config aren't defined in a
    * separate config file.
    */
-  g_auto(GStrv) groups = NULL;
-  gsize num_groups, i;
-  groups = g_key_file_get_groups (new_config, &num_groups);
-  for (i = 0; i < num_groups; i++)
+  gsize num_groups;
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo.c
```diff
@@ -933,6 +933,44 @@ ostree_repo_write_config (OstreeRepo *self,
...
+              if (!g_error_matches (local_error, G_IO_ERROR,
+                                    G_IO_ERROR_NOT_FOUND))
+                {
+                  g_propagate_error (error, g_steal_pointer (&local_error));
+                  return FALSE;
+                }
+            }
+          else if (cur_remote->file != NULL)
+            {
+              g_set_error (error, G_IO_ERROR, G_IO_ERROR_EXISTS,
```
dbnicholson: I'm not sure `G_IO_ERROR_EXISTS` is the error that should be thrown here. Suggestions welcome.

cgwalters:
LGTM!  I pushed a fixup to port it more fully to declare-and-initialize style.

@rh-atomic-bot r+ 0e4f208

cgwalters:
@rh-atomic-bot retry

PR #1161, opened 2017-09-12
lib/sysroot: Use direct g_mkdtemp() for overlay tmpdir

cgwalters:
The new libglnx `glnx_mkdtempat()` uses autocleanups, which
is inconvenient for this use case where we *don't* want autocleanups.
Since we don't need it to be fd-relative, just directly invoke
`g_mkdtemp_full()` which is fine for this use case.

Prep for updating libglnx.


jlebon:
@rh-atomic-bot r+ 0dee59b

cgwalters:
@rh-atomic-bot retry

PR #1166, opened 2017-09-13
Replace remote change op

dbnicholson:
The current add and delete remote change ops don't handle the case of modification of existing remotes. There are 2 cases this PR is trying to cover:

1. Making modifications to an existing remote. This is the merge operation. It's exposed in the CLI through a new `ostree remote modify` subcommand.

2. Setting a remote whether it previously existed or not. This is the replace operation. It's exposed in the CLI through the new `--force` option for `ostree remote add`. Here I've allowed replace to succeed even if the remote doesn't exist since it really seemed like a force add is the only place it would be used. Let me know what you think.

One question I had is if the new `OstreeRepoRemoteChange` enum values should be documented with a Since or something like that.

This also fixes https://bugzilla.gnome.org/show_bug.cgi?id=753373 by merging when the remote exists or adding when it doesn't.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -1735,6 +1735,87 @@ ostree_repo_remote_delete (OstreeRepo     *self,
   return impl_repo_remote_delete (self, NULL, FALSE, name, cancellable, error);
 }
 
+
+static gboolean
+impl_repo_remote_replace (OstreeRepo     *self,
+                          GFile          *sysroot,
+                          const char     *name,
+                          const char     *url,
+                          GVariant       *options,
+                          GCancellable   *cancellable,
+                          GError        **error)
+{
+  g_return_val_if_fail (name != NULL, FALSE);
+  g_return_val_if_fail (url != NULL, FALSE);
+  g_return_val_if_fail (options == NULL || g_variant_is_of_type (options, G_VARIANT_TYPE ("a{sv}")), FALSE);
+
+  if (!ostree_validate_remote_name (name, error))
+    return FALSE;
+
+  g_autoptr(GError) local_error = NULL;
+  g_autoptr(OstreeRemote) remote = _ostree_repo_get_remote (self, name, &local_error);
+  if (remote == NULL)
+    {
+      if (!g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        {
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return FALSE;
+        }
+      if (!impl_repo_remote_add (self, sysroot, FALSE, name, url, options,
+                                 cancellable, error))
+        return FALSE;
+    }
+  else
+    {
+      /* Replace the entire option group */
+      if (!g_key_file_remove_group (remote->options, remote->group, error))
+        return FALSE;
+
+      if (g_str_has_prefix (url, "metalink="))
+        g_key_file_set_string (remote->options, remote->group, "metalink",
+                               url + strlen ("metalink="));
+      else
+        g_key_file_set_string (remote->options, remote->group, "url", url);
+
+      if (options != NULL)
+        keyfile_set_from_vardict (remote->options, remote->group, options);
+
+      /* Write out updated settings */
+      if (remote->file != NULL)
+        {
+          gsize length;
+          g_autofree char *data = g_key_file_to_data (remote->options, &length,
+                                                      NULL);
+
+          if (!g_file_replace_contents (remote->file, data, length,
+                                        NULL, FALSE, 0, NULL,
+                                        cancellable, error))
+            return FALSE;
+        }
+      else
+        {
+          g_autoptr(GKeyFile) config = ostree_repo_copy_config (self);
+
+          /* Remove the existing group if it exists */
+          if (!g_key_file_remove_group (config, remote->group, &local_error))
+            {
+              if (!g_error_matches (local_error, G_KEY_FILE_ERROR,
+                                    G_KEY_FILE_ERROR_GROUP_NOT_FOUND))
+                {
+                  g_propagate_error (error, g_steal_pointer (&local_error));
+                  return FALSE;
+                }
+            }
+
+          ot_keyfile_copy_group (remote->options, config, remote->group);
+          if (!ostree_repo_write_config (self, config, error))
+            return FALSE;
+        }
+    }
+
+  return TRUE;
+}
+
 /**
  * ostree_repo_remote_change:
  * @self: Repo
@@ -1776,6 +1857,9 @@ ostree_repo_remote_change (OstreeRepo     *self,
     case OSTREE_REPO_REMOTE_CHANGE_DELETE_IF_EXISTS:
       return impl_repo_remote_delete (self, sysroot, TRUE, name,
                                       cancellable, error);
+    case OSTREE_REPO_REMOTE_CHANGE_REPLACE:
+      return impl_repo_remote_replace (self, sysroot, name, url, options,
+                                       cancellable, error);
     }
   g_assert_not_reached ();
 }
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -1761,6 +1761,7 @@ impl_repo_remote_replace (OstreeRepo     *self,
           g_propagate_error (error, g_steal_pointer (&local_error));
           return FALSE;
         }
+      g_clear_error (&local_error);
       if (!impl_repo_remote_add (self, sysroot, FALSE, name, url, options,
                                  cancellable, error))
         return FALSE;
--- a/tests/test-remotes-config-dir.js
+++ b/tests/test-remotes-config-dir.js
@@ -34,7 +34,7 @@ function assertNotEquals(a, b) {
 	throw new Error("assertion failed " + JSON.stringify(a) + " != " + JSON.stringify(b));
 }
 
-print('1..6')
+print('1..9')
 
 let remotesDir = Gio.File.new_for_path('remotes.d');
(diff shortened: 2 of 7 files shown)


cgwalters:
Let me look at this after the next release if that's OK.

dbnicholson:
Sure, no rush. I was just scratching an itch since I was looking at all the remote stuff. That said, one of my goals with this is to convert flatpak to do reliable remote modifications, so having it in an ostree release is kind of important to me.

tests/test-remotes-config-dir.js
```diff
@@ -112,3 +112,42 @@ try {
...
+}
+
+print("ok merge-missing-remote-fails");
+
+// Test merging options to remote in config dir
+let opts = GLib.Variant.new('a{sv}', {
+    'branches': GLib.Variant.new('as', ['master'])
+});
+repo.remote_change(null, OSTree.RepoRemoteChange.MERGE, 'baz',
+                   'http://baz2', opts, null);
```
cgwalters: One issue this raises is that `* @url: URL for remote` is not currently `(nullable)`.  We could make it so...it'd be one of those subtle IABI breaks that only affects compiled languages.

(The reason to make it nullable is to make it more convenient to change the options without changing the URL)

src/libostree/ostree-repo.c
```diff
@@ -1266,6 +1266,86 @@ impl_repo_remote_merge (OstreeRepo     *self,
...
+
+  g_autoptr(GError) local_error = NULL;
+  g_autoptr(OstreeRemote) remote = _ostree_repo_get_remote (self, name, &local_error);
+  if (remote == NULL)
+    {
+      if (!g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
+        {
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return FALSE;
+        }
```
cgwalters: `g_clear_error (&local_error)` here.

src/libostree/ostree-repo.c
```diff
@@ -1266,6 +1266,86 @@ impl_repo_remote_merge (OstreeRepo     *self,
...
+    {
+      /* Replace the entire option group */
+      if (!g_key_file_remove_group (remote->options, remote->group, error))
+        return FALSE;
+
+      if (g_str_has_prefix (url, "metalink="))
+        g_key_file_set_string (remote->options, remote->group, "metalink",
+                               url + strlen ("metalink="));
+      else
+        g_key_file_set_string (remote->options, remote->group, "url", url);
```
cgwalters: Hm, at this point we should extract this to a helper I'd say.

src/ostree/ot-remote-builtin-modify.c
```diff
@@ -0,0 +1,149 @@
...
+gboolean
+ot_remote_builtin_modify (int argc, char **argv, GCancellable *cancellable, GError **error)
+{
+  g_autoptr(GOptionContext) context = NULL;
+  g_autoptr(OstreeSysroot) sysroot = NULL;
+  g_autoptr(OstreeRepo) repo = NULL;
+  const char *remote_name;
+  char **iter;
+  g_autoptr(GVariantBuilder) optbuilder = NULL;
+  g_autoptr(GVariant) options = NULL;
```
jlebon: This is minor, though note that we tend to declare things as they are assigned now. For iteration variables like `iter` and `i`, you can just declare in the for loop setup.

dbnicholson: I thought about that. I just copied remote-builtin-add, but I can make these more modern-y.

src/libostree/ostree-repo.c
dbnicholson: Right, I didn't spend a lot of time thinking about refactoring things. I mostly wanted to get the feature and tests in place to see what you thought. But I'll look through again and see if there are clear winners.

src/libostree/ostree-repo.c
dbnicholson: Originally I had that, but is it necessary here? It's declared with `g_autoptr` and never reused in this branch. But I can add it to be sure it doesn't leak.

tests/test-remotes-config-dir.js
dbnicholson: I hadn't thought of that. The `url` parameter is pretty odd in this operation. I had some more thoughts about this that I'll follow up on in the general PR conversation.

dbnicholson:
Thinking about the `(nullable)` `url` param for `MERGE` some more.

Maybe it would be best to move this out to it's own function like `ostree_repo_remote_modify` that either has a `(nullable)` URL parameter or just handles everything through the `options` `a{sv}` variant.

Furthermore, I later thought of another operation that would be useful - clearing an option so it's effectively reset to the default. That type of operation would require `url` to be `NULL`, and in that case how do you differentiate "I want to remove the `url` setting" (even though that would be silly) from "this is what the API requires".

So, one idea I had to tackle both of those is a separate function:
```
/**
 * ostree_repo_remote_modify:
 * @self: Repo
 * @remove: Remove the remote option
 * @name: Name of remote
 * @options: GVariant of type a{sv} when @remove is FALSE, type as when TRUE
 * @cancellable: Cancellable
 * @error: Error
 *
 * Set or remove options from an existing remote.
 *
 * Returns: %TRUE on success, %FALSE on failure
*/
gboolean
ostree_repo_remote_modify (OstreeRepo     *self,
                           gboolean        remove,
                           const char     *name,
                           GVariant       *options,
                           GCancellable   *cancellable,
                           GError        **error);
```
Any option you want to modify is in the variant. When removing options, the variant is simply `as` rather than `a{sv}` since there's no value needed. I'm not sure how I feel about the variant type being different depending on another parameter. What do you think?

src/libostree/ostree-repo.c
cgwalters: Yes, but at this point the error is set - so we must clear it for the invocation of `impl_repo_remote_add()` right below, otherwise get could get the overwriting-error error.

cgwalters:
Wouldn't one possibly want to both add/change and remove options at the same time?  So perhaps what we want is `a{smv}` or so?  (Can one do that?  Huh fun, looks like GLib supports it but the current pygobject bindings can't)

src/libostree/ostree-repo.c
dbnicholson: The call to `impl_repo_remote_add` uses `error`, not `local_error`. `local_error` is only used to check if the repo exists. The caller's `error` is used when adding and `local_error` will be cleared regardless of what happens there. I don't mind clearing the error, but I don't think it's really necessary.

cgwalters: Ah yes, you're totally right.  That said, I still think it's best practice to clear checked errors right after checking them.  Imagine if someone later comes and wants to add another checked call, sees there's already a `local_error` value and doesn't realize it may have been written earlier.

cgwalters:
Would definitely like to get this in!  Maybe it's simplest to punt on "modify" for now and have callers implement by using replace?

dbnicholson:
Sorry for leaving this so long. My ostree hacking time is at a minimum these days. This is something I'd really like to have since we've had a never ending stream of remote configuration issues in Endless.

Another thought I had later was to just add accessors for an already defined remote like flatpak has. I like their APIs:
```
FLATPAK_EXTERN FlatpakRemote        *flatpak_installation_get_remote_by_name (FlatpakInstallation *self,
                                                                              const gchar         *name,
                                                                              GCancellable        *cancellable,
                                                                              GError             **error);
FLATPAK_EXTERN FlatpakRemote * flatpak_remote_new (const char    *name);
FLATPAK_EXTERN char *        flatpak_remote_get_url (FlatpakRemote *self);
FLATPAK_EXTERN void          flatpak_remote_set_url (FlatpakRemote *self,
                                                     const char    *url);
```
I could see something very similar in ostree using the currently experimental `OstreeRemote`. In addition to some common accessors like `(get|set)_url()` and `(get|set)_gpg_verify()`, I'd add a generic option accessor:
```
gchar *ostree_remote_get_option (OstreeRemote *remote, const gchar *option);
void ostree_remote_set_option (OstreeRemote *remote, const gchar *option, const gchar *value);
```
I could see possibly mirroring some of the `GKeyFile` APIs for specific values like `(get|set)_boolean_option()`.

And finally an API to remove an option:
```
void ostree_remote_remove_option (OstreeRemote *remote, const gchar *option);
```

After that you'd have to make a call to persist the changes:
```
gboolean ostree_repo_remote_update (OstreeRepo *repo, OstreeRemote *remote, GCancellable cancellable, GError **error);
```

What do you think?

matthiasclasen:
This would still be useful for flatpak

dbnicholson:
> Would definitely like to get this in! Maybe it's simplest to punt on "modify" for now and have callers implement by using replace?

OK, so as you say maybe it's best to just add the `REPLACE` change operation and worry about modifying in another PR. That's actually what I really want, but I personally think it would be best to flesh out the `OstreeRemote` API similar to flatpak but that would require more work..

dbnicholson:
I repushed now with just the replace operation (and changed the PR title). I added a fixup for the `g_clear_error()` @cgwalters requested and added a `Since` in the enum comment (should have been a fixup, too, sorry). I removed all the modification stuff and will move that to a separate PR after I have a bit more thinking about what the best interface is.

ramcq:
@cgwalters @jlebon This API is causing much sadness in flatpak (see https://github.com/flatpak/flatpak/issues/1665 and https://github.com/flatpak/flatpak/issues/2508). Is it possible to give this PR a nudge forward so that Flatpak has at least one way to do remote modification in the config dir case?

cgwalters:
LGTM, thanks!

@rh-atomic-bot r+ d010d5f

PR #1168, opened 2017-09-13
lib/repo: Port tmpdir locking func to new style

cgwalters:
Prep for future work.


jlebon:
@rh-atomic-bot r+ d82f6da

PR #1169, opened 2017-09-13
tests: Port some bits of C to new style

cgwalters:
Where we can; perhaps after updating libglnx we should use the
new test error macro?


jlebon:
@rh-atomic-bot r+ db11185

Issue #1165, opened 2017-09-13
DevIno cache doesn't work when you're using a modifier on the commit


alexlarsson:
The way the dev-ino cache works is that if the source file matches the dev+ino, then we immediately use that file with that object as-is. However, that might not be correct if a OstreeRepoCommitModifier is in effect. The following demo demonstrates how this goes wrong:

```
# Create bare repo with a commit
mkdir to-commit
echo CONTENT > to-commit/file
ostree init --repo=repo --mode=bare
ostree commit --repo=repo --branch=master to-commit

# Check out, ensure we have hardlinks
ostree checkout --repo=repo -H master checked-out
# Check in again, but now with modified owner
sudo ostree commit --repo=repo --owner-uid=42 --branch=new-owner checked-out

# Look, these are different
ostree ls -RC --repo=repo master
ostree ls -RC --repo=repo new-owner

# Now try with --link-checkout-speedup
sudo ostree commit --repo=repo --owner-uid=42 --link-checkout-speedup --branch=new-owner-speedup checked-out

# This shows the wrong result, the owner of /file is not 42!
ostree ls -RC --repo=repo new-owner-speedup
```

I think we need to run the modifier on the actual file metadata, and only use the hardlinked file in case none of the metadata changes.

cgwalters:
(Why `sudo ostree commit`?)

Is this a regression?  Or something new you're trying to do?

Definitely a bug anyways.

jlebon:
Will take a look at this.

cgwalters:
I want to understand the use case here.  Is there one or is this theoretical?

Note: the command line is mostly just a demo, and `--link-checkout-speedup` in particular should be viewed as deprecated in favor of `OstreeRepoDevInoCache` (not accessible from cmdline since it requires state).  I'm *presuming* this is just a cmdline version of a flatpak issue though - is that correct?

Under what scenarios do you want to globally change the UID?

alexlarsson:
I do not have a particular usecase, I just noticed this when looking for ways to optimize flatpak-builder. 

In particular, it has a ostree repo for the build cache and a hardlinked checkout of that, and then we want to commit (a subset of) this directory to a different repo. I was hoping we could use something like a dev-ino cache to do map the files being commited to the cache repo and then avoiding to do the checksum of every file. That approach failed though, because the cache repo has different metadata than the final repo, so the cache checksums are wrong anyway. (The cache repo has files with your uid, but the final repo has canonical permissions.)

[...]

PR #1170, opened 2017-09-13
Pr/devino cache bug

jlebon:
Fixes https://github.com/ostreedev/ostree/issues/1165.


src/libostree/ostree-repo-commit.c
```diff
@@ -2858,16 +2858,6 @@ write_dfd_iter_to_mtree_internal (OstreeRepo                  *self,
       if (!glnx_fstatat (src_dfd_iter->fd, dent->d_name, &stbuf, AT_SYMLINK_NOFOLLOW, error))
         return FALSE;
 
-      const char *loose_checksum = devino_cache_lookup (self, modifier, stbuf.st_dev, stbuf.st_ino);
```
cgwalters: I think this patch should be included with the last commit since they're not
*really* independent in the sense that I think we were doing this early lookup
intentionally before.  It was intending to be a fast path.

src/libostree/ostree-repo-commit.c
```diff
@@ -2618,16 +2619,20 @@ write_directory_content_to_mtree_internal (OstreeRepo                  *self,
...
       g_autoptr(GInputStream) file_input = NULL;
       g_autoptr(GVariant) xattrs = NULL;
       g_autoptr(GInputStream) file_object_input = NULL;
       g_autofree guchar *child_file_csum = NULL;
       g_autofree char *tmp_checksum = NULL;
 
-      loose_checksum = devino_cache_lookup (self, modifier,
-                                            g_file_info_get_attribute_uint32 (child_info, "unix::device"),
-                                            g_file_info_get_attribute_uint64 (child_info, "unix::inode"));
+      /* only check the devino cache if the file info was not modified */
```
cgwalters: What about xattrs?  For the rpm-ostree use case it'd make commit notably more expensive as suddenly we'd have to load and compare the xattrs.

If we're explicitly not comparing xattrs then let's at least add a comment.

jlebon: Ahh of course, we need to check xattrs too. Will add that.

Hmm, do you mean in the rpm-ostree local assembly path? Though we don't provide an xattr modifier when checking out files there, so we can probably do it in a smart way so we don't even bother comparing in such cases.

cgwalters: Hmm...but actually in the layering case I think we *should* be setting the sepolicy to ensure we relabel any files created by scripts.  (In practice this probably doesn't matter much and really everything should mostly be `usr_t` but still).

jlebon:
OK, this now also checks for modifications to extended attributes. One thing I'm not sure of is:

> This patch does slightly change the semantics of the xattr callback.
Previously, returning NULL from the cb meant no xattrs at all. Now, it
means to default to the on-disk state. We might want to consider putting
that behind a flag instead. It seems like a more useful behaviour so
that callers can only override the files they want to without losing
original on-disk state (and if they don't want that, just return an
empty GVariant).

Without that change, really there's not much point in using a devino cache if also using an xattr cb.

cgwalters:
Pull out the first commit to a separate PR?

cgwalters:
I'd like to defer this until we've had a chance to discuss in https://github.com/ostreedev/ostree/issues/1165

src/libostree/ostree-repo-commit.c
```diff
@@ -2557,6 +2571,7 @@ write_directory_content_to_mtree_internal (OstreeRepo                  *self,
   g_autoptr(GFileInfo) modified_info = NULL;
   OstreeRepoCommitFilterResult filter_result =
     _ostree_repo_commit_modifier_apply (self, modifier, child_relpath, child_info, &modified_info);
+  const gboolean child_info_was_modified = (child_info != modified_info);
```
cgwalters: I think my main problem is this - if we have say a commit filter active and that filter just wants to change *some* paths like gnome-continuous does with suid binaries, this will still forcibly re-checksum everything since the info will be different.

I think we need to compare it structurally the same way we do xattrs.

jlebon:
Rebased and updated to also inspect the `GFileInfo`. ⬆️

src/libostree/ostree-core.c
```diff
@@ -1568,12 +1568,52 @@ _ostree_stbuf_to_gfileinfo (const struct stat *stbuf)
   g_file_info_set_attribute_uint32 (ret, "unix::uid", stbuf->st_uid);
   g_file_info_set_attribute_uint32 (ret, "unix::gid", stbuf->st_gid);
   g_file_info_set_attribute_uint32 (ret, "unix::mode", mode);
+
+  /* those aren't stored by ostree, but used by the devino cache */
+  g_file_info_set_attribute_uint32 (ret, "unix::device", stbuf->st_dev);
+  g_file_info_set_attribute_uint64 (ret, "unix::inode", stbuf->st_ino);
```
cgwalters: Ooh, tricky.

src/libostree/ostree-repo-commit.c
```diff
@@ -2363,58 +2363,67 @@ ptrarray_path_join (GPtrArray  *path)
...
+                  GCancellable                     *cancellable,
+                  GError                          **error)
+{
+  /* track whether the returned xattrs differ from the file on disk */
+  gboolean modified = TRUE;
+
+  /* fetch on-disk xattrs if needed & not disabled */
+  g_autoptr(GVariant) original_xattrs = NULL;
+  if (!(modifier && (modifier->flags & (OSTREE_REPO_COMMIT_MODIFIER_FLAGS_SKIP_XATTRS |
+                                        OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS)) > 0)
```
cgwalters: Not new, but maybe while we're here, host out the flags checking like:
```
const gboolean flags_skip_xattrs = (modifier->flags & (OSTREE_REPO_COMMIT_MODIFIER_FLAGS_SKIP_XATTRS |
 OREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS)) > 0;
if (!(modifier && flags_skip_xattrs && !self->disable_xattrs))
  {
    ...
```

tests/basic-test.sh
```diff
@@ -603,6 +603,24 @@ $OSTREE checkout test2 test2-checkout
...
+=$((0600)) /baz/cowro
+EOF
+$OSTREE commit ${COMMIT_ARGS} --statoverride=statoverride.txt \
+  --table-output --link-checkout-speedup -b test2-tmp test2-checkout > stats.txt
+$OSTREE diff test2 test2-tmp > diff-test2
+assert_file_has_content diff-test2 'M */baz/cow$'
+assert_not_file_has_content diff-test2 'M */baz/cowro$'
+assert_not_file_has_content diff-test2 'baz/saucer'
+# only /baz/cow is a cache miss
+assert_file_has_content stats.txt '^Content Written: 1$'
```
cgwalters: Ah, right I forgot we already had a way to unit test this.  Cool.

tests/test-basic-c.c
```diff
@@ -236,20 +236,140 @@ test_object_writes (gconstpointer data)
...
+  g_assert (ret);
+
+  gboolean on_overlay;
+  ret = ot_check_for_overlay (&on_overlay, &error);
+  g_assert_no_error (error);
+  g_assert (ret);
+
+  if (on_overlay)
+    {
+      g_test_skip ("overlayfs detected");
```
cgwalters: This is fine for now, though a more accurate fix would be porting the "can relabel" logic.  Though OTOH I agree it is perhaps easier for dev testing to set the directory.  Maybe we want TMPDIR_PREFIX though?

cgwalters:
Eh, might as well zap 🗲 this one; we can do my one minor style tweak later.

@rh-atomic-bot r+ 211e748

tests/libostreetest.c
```diff
@@ -72,29 +72,44 @@ ot_test_setup_repo (GCancellable *cancellable,
   return g_steal_pointer (&ret_repo);
 }
 
+gboolean
+ot_check_for_overlay (gboolean  *on_overlay,
+                      GError   **error)
+{
+  /* Keep this in sync with the overlayfs bits in libtest.sh */
+  struct statfs stbuf;
+  const char *pwd = g_getenv ("PWD");
```
smcv: Surely `g_get_current_dir()` (needs to be assigned to a `g_autofree gchar *`), or simply replace `pwd` with `"."` below?

I'm not completely sure of the startup sequence of this stuff, but if the test isn't guaranteed to have been started by a shell designed for interactive use, then I don't think `PWD` is guaranteed to be up-to-date.

According to its documentation, `g_get_current_dir()` returns `PWD`, but only after verifying that `PWD` is in fact the same directory as `"."`.

tests/libtest.sh
```diff
@@ -73,7 +73,7 @@ export OSTREE_GPG_HOME=${test_tmpdir}/gpghome/trusted
 # See comment in ot-builtin-commit.c and https://github.com/ostreedev/ostree/issues/758
 # Also keep this in sync with the bits in libostreetest.c
 echo evaluating for overlayfs...
-case $(stat -f --printf '%T' /) in
+case $(stat -f --printf '%T' $PWD) in
```
smcv: Same here: this could be `.` (although at least this is a bash script so `PWD` is guaranteed(?) to be set and up-to-date)

tests/test-basic-c.c
```diff
@@ -236,20 +236,140 @@ test_object_writes (gconstpointer data)
...
+
+  g_autoptr(GFile) repo_path = g_file_new_for_path ("repo");
+
+  /* re-initialize as bare */
+  ret = ot_test_run_libtest ("setup_test_repository bare", &error);
+  g_assert_no_error (error);
+  g_assert (ret);
+
+  gboolean on_overlay;
+  ret = ot_check_for_overlay (&on_overlay, &error);
```
smcv: If this is a proxy for "check for a filesystem where we can set xattrs", please probe for that instead - tmpfs can't have xattrs either, and for optimal speed (and lack of fsync slowdown), some Debian buildds use a chroot where the entire visible filesystem is on tmpfs.

tests/libostreetest.c
jlebon: Sure. I had started with `getenv ("TEST_TMPDIR")` at first, though changed it afterwards since it doesn't actually appear when running with gdtr.

tests/test-basic-c.c
jlebon: Yeah, @cgwalters echoed the same sentiments higher up. I'll rework it.

jlebon:
OK, I reworked the tests so that they explicitly check for relabeling support rather than overlayfs. It's a bit confusing because `OSTREE_NO_XATTRS` should be broken down further depending on what repo/checkout/commit modes we're trying to test. Anyway, I *think* I did the right thing for now in ⬆️.

jlebon:
And now we also run as an installed test to make sure the new xattr test actually runs! ⬆️

tests/libostreetest.c
```diff
@@ -72,29 +72,88 @@ ot_test_setup_repo (GCancellable *cancellable,
...
+{
+  g_auto(GLnxTmpfile) tmpf = { 0, };
+  if (!glnx_open_tmpfile_linkable_at (AT_FDCWD, ".", O_RDWR | O_CLOEXEC, &tmpf, error))
+    return FALSE;
+
+  g_autoptr(GError) local_error = NULL;
+  g_autoptr(GBytes) bytes = glnx_fgetxattr_bytes (tmpf.fd, "security.selinux", &local_error);
+  if (!bytes)
+    {
+      /* libglnx preserves errno */
```
cgwalters: Hmm...not explicitly.  I think it likely mostly works today, but I'm certain that e.g. `glnx_null_throw_errno_prefix()` might invoke `malloc()` which can definitely make system calls... if we want to make that guarantee I think we'll need something more like systemd's `_PROTECT_ERRNO`.

tests/libtest.sh
```diff
@@ -516,12 +544,15 @@ os_repository_new_commit ()
...
-have_user_xattrs() {
+assert_has_setfattr() {
     if ! which setfattr 2>/dev/null; then
         fatal "no setfattr available to determine xattr support"
     fi
+}
+
+_have_user_xattrs=''
+have_user_xattrs() {
+    assert_has_setfattr
```
cgwalters: Minor, but maybe move the assertion below the test so we don't call it every time?

tests/libostreetest.c
cgwalters: Nevermind, I forgot we do save it.

cgwalters:
OK, this is looking pretty good to me!  Let's get it in, we can do any followups after.  Maybe at some point we can stub out a context or case that uses tmpfs so @smvc doesn't have to come in and fix things up after every release.

@rh-atomic-bot r+ 9410a8f

PR #1171, opened 2017-09-13
lib/checkout: Do UNION_FILES via atomic renameat()

cgwalters:
I was looking at fixing an `rpm-ostree livefs` bug where we need to replace
`/usr/lib/passwd`. It's obviously bad if that temporarily disappears 😉. My plan
is to do a subpath checkout of just `/usr/lib/{passwd,group}`.

Make this atomic (i.e. file always exists) by changing the logic to create a
temporary link in repo/tmp, then rename() it into place.

A bonus here is we kill one of the very few (only?) non-error-cleanup i.e.
"non-linear" `goto`s in the ostree codebase.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -338,6 +338,41 @@ typedef enum {
   HARDLINK_RESULT_LINKED
 } HardlinkResult;
 
+/* Used for OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES.  In
+ * order to atomically replace a target, we add a new link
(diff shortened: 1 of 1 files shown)


cgwalters:
(For the next release)

jlebon:
Nice! @rh-atomic-bot r+ bb4e339

cgwalters:
@rh-atomic-bot retry

jlebon:
There's something funny going on with the new PAPR backend causing tests to time out. Investigating.

jlebon:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1172, opened 2017-09-14
Update libglnx

cgwalters:
Update libglnx, which is mostly port the repo stagedir code
to the new tmpdir API.  This turned out to require some
libglnx changes to support de-allocating the tmpdir ref while
still maintaining the on-disk dir.

Update submodule: libglnx


cgwalters:
For after the next release.

cgwalters:
Now depends on https://github.com/ostreedev/ostree/pull/1176

PR #1176, opened 2017-09-15
lib/pull: Drop partial fetch code from libsoup backend

cgwalters:
Doing this in prep for libglnx tmpdir porting, but I think we should also do
this because the partial fetch code IMO was never fully baked; among other
things it was never integrated into the scheme we came up with for "boot id
sync" that we use for complete/staged objects.

There's a lot of complexity here that while we have some coverage for, I think
we need to refocus on the core functionality. The libcurl backend doesn't have
an equivalent to this today.

In particular for small objects, this is simply overly complex. The downside is
clearly for large objects like FAH's 61MB initramfs; not being able to resume
fetches of those is unfortunate.

In practice though, I think most people should be using deltas, and we need to
make sure deltas work for large objects anyways.

Further ultimately the peer-to-peer work should help a lot for people
with truly unreliable connections.

### DIFF
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -909,6 +909,9 @@ on_stream_read (GObject        *object,
           if (!glnx_open_tmpfile_linkable_at (pending->thread_closure->base_tmpdir_dfd, ".",
                                               O_WRONLY | O_CLOEXEC, &pending->tmpf, &local_error))
             goto out;
+          /* This should match the libcurl chmod */
+          if (!glnx_fchmod (pending->tmpf.fd, 0644, &local_error))
+            goto out;
           pending->out_stream = g_unix_output_stream_new (pending->tmpf.fd, FALSE);
         }
       else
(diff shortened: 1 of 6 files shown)


cgwalters:
Of course another path would be to just delete the libsoup backend entirely.

cgwalters:
Dependency of https://github.com/ostreedev/ostree/pull/1172

src/libostree/ostree-fetcher-soup.c
```diff
@@ -990,14 +941,24 @@ on_stream_read (GObject        *object,
         }
       else
         {
-          g_task_return_pointer (task,
-                                 g_strdup (pending->out_tmpfile),
-                                 (GDestroyNotify) g_free);
+          g_autofree char *uristring
+            = soup_uri_to_string (soup_request_get_uri (pending->request), FALSE);
```
jlebon: Minor: I tend to prefer keeping the `=` on the first line.

src/libostree/ostree-fetcher-soup.c
```diff
@@ -990,14 +941,24 @@ on_stream_read (GObject        *object,
...
+          g_autofree char *tmpfile_path =
+            ostree_fetcher_generate_url_tmpname (uristring);
+          if (!glnx_link_tmpfile_at (&pending->tmpf, GLNX_LINK_TMPFILE_REPLACE,
```
jlebon: Looks like we need to `chmod` to `0666` here before linking it in place.

```
+ mode=600
+ test 600 = 600
+ assert_not_reached 'Content object unreadable by others: 600'
+ fatal 'Content object unreadable by others: 600'
+ echo Content object unreadable by others: 600
Content object unreadable by others: 600
```

src/libostree/ostree-fetcher-soup.c
cgwalters: Yep me too, thanks for catching that.

src/libostree/ostree-fetcher-soup.c
cgwalters: Yep, thanks for digging through the logs before me 😄  Fixup ⬇️

jlebon:
@rh-atomic-bot r+ 469ae9c

PR #1177, opened 2017-09-15
Fix test-help.sh for the case where ostree trivial-httpd is enabled

smcv:
It's deprecated, but as long as it's still supported at all, it
shouldn't make the tests fail.

$ ostree trivial-httpd --help
Usage:
  /usr/lib/libostree/ostree-trivial-httpd [OPTION…] [DIR] - Simple webserver

### DIFF
--- a/tests/test-help.sh
+++ b/tests/test-help.sh
@@ -75,7 +75,16 @@ test_recursive() {
         assert_file_empty out
 
         for subcmd in $builtins; do
-            test_recursive "$cmd $subcmd"
+            case "$subcmd" in
+                (trivial-httpd)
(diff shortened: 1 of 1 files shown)


jlebon:
Hmm, I feel like we should just special-case `trivial-httpd` and not recurse into it at all. Meh, it's fine either way really.

Thanks!
@rh-atomic-bot r+ 4cce115

smcv:
Actually, this won't work. `ostree trivial-httpd` runs `${libexecdir}/ostree-trivial-httpd` which won't exist in the case where we are doing a build on a minimal system. So we do have to just skip trivial-httpd.

jlebon:
@rh-atomic-bot r+ d94f2f4

PR #1178, opened 2017-09-15
Fix JavaScript tests with gjs 1.50.0

smcv:
In recent gjs, you can't declare a variable with "let" multiple times.

### DIFF
--- a/tests/corrupt-repo-ref.js
+++ b/tests/corrupt-repo-ref.js
@@ -47,7 +47,6 @@ function listObjectChecksumsRecurse(dir, allObjects) {
     e.close(null);
 } 
 
-let [,commit] = repo.resolve_rev(refToCorrupt, false);
 let [,root,commit] = repo.read_commit(refToCorrupt, null);
 let allObjects = {};
 allObjects[commit + '.commit'] = true;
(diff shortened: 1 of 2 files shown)


tests/corrupt-repo-ref.js
```diff
@@ -47,7 +47,7 @@ function listObjectChecksumsRecurse(dir, allObjects) {
     e.close(null);
 } 
 
-let [,commit] = repo.resolve_rev(refToCorrupt, false);
+repo.resolve_rev(refToCorrupt, false);
```
cgwalters: This is basically dead code now.

tests/test-sysroot.js
```diff
@@ -97,9 +97,9 @@ assertEquals(deploymentPath.query_exists(null), false);
 
 //// Ok, redeploy, then add a new revision upstream and pull it
 
-let [,deployment] = sysroot.deploy_tree('testos', rev, origin,
-					mergeDeployment, null,
-					null);
+[,deployment] = sysroot.deploy_tree('testos', rev, origin,
+				 mergeDeployment, null,
```
cgwalters: Hm, looks like we have space/tab issues here.  Ug.  Will try to do an untabify pass at some point.  (Don't need to in this patch)

smcv: I noticed that, and tried to stick to how it's currently done.

tests/corrupt-repo-ref.js
smcv: Ah, I'd assumed it was being called for some side-effect.

Issue #1174, opened 2017-09-14
Change schema for finding repositories on USB sticks in OstreeRepoFinderMount


pwithnall:
tl;dr: I want to make a backwards-incompatible change to how `OstreeRepoFinderMount` looks for repos on USB sticks. That should be fine though, since it’s experimental API.

I’ve been putting together a tool for pulling an OSTree onto a USB stick so that `OstreeRepoFinderMount` can find it (which will be a separate PR). While doing so, I started thinking about live USBs and how this will interact with those. Ideally, a live USB stick containing an OSTree-based OS could be found by `OstreeRepoFinderMount` without any additional work. The current schema for finding repositories which `OstreeRepoFinderMount` uses can’t support that — it requires a load of `$usb_stick/.ostree/repos/$collection_id/$ref_name` symlinks to `$usb_stick/ostree/repo`.

Thinking about that a bit more, I realised that the current scheme won’t pick up new refs from the live USB’s repository if (for example) someone boots from the live USB, pulls a new ref into the repository (for example, by installing a flatpak app), then plugs the USB stick into another computer and tries to find that ref on the stick using `OstreeRepoFinderMount`. The ref will exist in the live USB’s repository, but there will be no `$usb_stick/.ostree/repos/$collection_id/$ref_name` symlink for it.

I think that requiring a symlink for each collection–ref was a mistake. It was intended to allow multiple independent repositories to coexist on a single USB stick, but I think a better way of doing this is to have a `$usb_stick/.ostree/repos.d` directory containing symlinks to the repositories (or the repositories themselves), as children. The names of the directories/symlinks in `repos.d` doesn’t matter, though we could perhaps access them in lexical order for predictability.

In addition to this, I think `OstreeRepoFinderMount` should also always look for `$usb_stick/ostree/repo` as a repository, since that’s a well-known path for OSTree repositories, and sufficiently common that it seems a bit pointless to need well-known symlink to it from `$usb_stick/.ostree/repos.d`.

Two questions:
 - Is the experimental API break OK? I assume it will be.
 - Does the new schema sound reasonable?

cgwalters:
The old scheme seemed weird to me for sure, I doubt anyone else is using it.

[...]

PR #1179, opened 2017-09-15
lib/repo-finder-mount: Change the schema for finding repos on volumes

pwithnall:
See issue #1174 for the rationale behind this. In summary:
 • It required two lists of collection–refs to be maintained: one in the
   repository, and one pointing to the repository.
 • It didn’t automatically work for live USBs of OSs based on OSTree
   (where there’s always a repository at /ostree/repo).
 • It was unnecessarily complex.

The new scheme allows a list of repositories to be searched, but without
needing a layer of indirection through their collection–refs. It adds
/ostree/repo and /.ostree/repo as well-known repository locations which
are always checked on a mounted volume (if they exist).

Update the unit tests accordingly.

---

Plus a few other trivial fixes.

https://github.com/ostreedev/ostree/issues/1174

### DIFF
--- a/src/libostree/ostree-repo-finder-mount.c
+++ b/src/libostree/ostree-repo-finder-mount.c
@@ -433,6 +433,7 @@ ostree_repo_finder_mount_resolve_async (OstreeRepoFinder                  *finde
 
       for (i = 0; refs[i] != NULL; i++)
         {
+          const OstreeCollectionRef *ref = refs[i];
           g_autofree gchar *resolved_repo_uri = NULL;
           g_autofree gchar *keyring = NULL;
           g_autoptr(UriAndKeyring) resolved_repo = NULL;
@@ -444,24 +445,24 @@ ostree_repo_finder_mount_resolve_async (OstreeRepoFinder                  *finde
               GHashTable *repo_refs = repo_and_refs->refs;
               g_autofree char *repo_path = g_file_get_path (ostree_repo_get_path (repo));
 
-              const gchar *checksum = g_hash_table_lookup (repo_refs, refs[i]);
+              const gchar *checksum = g_hash_table_lookup (repo_refs, ref);
 
               if (checksum == NULL)
                 {
                   g_debug ("Ignoring repository ‘%s’ when looking for ref (%s, %s) on mount ‘%s’ as it doesn’t contain the ref.",
-                           repo_path, refs[i]->collection_id, refs[i]->ref_name, mount_name);
+                           repo_path, ref->collection_id, ref->ref_name, mount_name);
                   g_clear_error (&local_error);
                   continue;
                 }
 
               /* Finally, look up the GPG keyring for this ref. */
-              keyring = ostree_repo_resolve_keyring_for_collection (parent_repo, refs[i]->collection_id,
+              keyring = ostree_repo_resolve_keyring_for_collection (parent_repo, ref->collection_id,
                                                                     cancellable, &local_error);
 
               if (keyring == NULL)
                 {
                   g_debug ("Ignoring repository ‘%s’ when looking for ref (%s, %s) on mount ‘%s’ due to missing keyring: %s",
-                           repo_path, refs[i]->collection_id, refs[i]->ref_name, mount_name, local_error->message);
+                           repo_path, ref->collection_id, ref->ref_name, mount_name, local_error->message);
                   g_clear_error (&local_error);
                   continue;
                 }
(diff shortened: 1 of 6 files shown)


src/libostree/ostree-repo-finder-mount.c
```diff
@@ -256,122 +433,66 @@ ostree_repo_finder_mount_resolve_async (OstreeRepoFinder                  *finde
...
+                           repo_path, refs[i]->collection_id, refs[i]->ref_name, mount_name, local_error->message);
+                  g_clear_error (&local_error);
+                  continue;
+                }
+
+              /* There is a valid repo at (or pointed to by)
+               * $mount_root/.ostree/repos.d/$something.
+               * Add it to the results, keyed by the canonicalised repository URI
+               * to deduplicate the results. */
+              g_autofree char *canonical_repo_path = realpath (repo_path, NULL);
```
cgwalters: Not new in this patch but I don't quite get why we need to canonicalize; feels like a leftover from the previous symlinks schema?

src/libostree/ostree-repo.c
```diff
@@ -2615,6 +2615,31 @@ ostree_repo_get_dfd (OstreeRepo  *self)
...
+ * ostree_repo_equal:
+ * @a: an #OstreeRepo
+ * @b: an #OstreeRepo
+ *
+ * Check whether two opened repositories are the same on disk: if their root
+ * directories are the same inode. If @a or @b are not open yet (due to
+ * ostree_repo_open() not being called on them yet), %FALSE will be returned.
+ *
+ * Returns: %TRUE if @a and @b are the same repository on disk, %FALSE otherwise
+ * Since: 2017.11
```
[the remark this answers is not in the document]
pwithnall: Oops, fixup pushed.

src/libostree/ostree-repo-finder-mount.c
```diff
@@ -256,122 +433,66 @@ ostree_repo_finder_mount_resolve_async (OstreeRepoFinder                  *finde
...
-              g_debug ("Ignoring ref (%s, %s) on mount ‘%s’ as it is the same as the one we are resolving",
-                       refs[i]->collection_id, refs[i]->ref_name, mount_name);
-              g_clear_error (&local_error);
-              continue;
+              const RepoAndRefs *repo_and_refs = &g_array_index (repos_refs, RepoAndRefs, j);
+              OstreeRepo *repo = repo_and_refs->repo;
+              GHashTable *repo_refs = repo_and_refs->refs;
+              g_autofree char *repo_path = g_file_get_path (ostree_repo_get_path (repo));
+
+              const gchar *checksum = g_hash_table_lookup (repo_refs, refs[i]);
```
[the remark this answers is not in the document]
pwithnall: Sensible. Fixup pushed.

src/libostree/ostree-repo-finder-mount.c
pwithnall: There are still symlinks in this schema (although fewer of them); canonicalisation allows us to eliminate the duplicate arising from (for example):
 * `$usb_stick/ostree/repo` being a repository
 * `$usb_stick/.ostree/repos.d/alias` → `$usb_stick/ostree/repo` being a symlink
or the duplicate arising from two symlinks in `$usb_stick/.ostree/repos.d` pointing to the same repository elsewhere on the stick.

Is there any downside to canonicalising? I guess it’s not dirfd-relative, which is a a bit of a pain, but I don’t think `realpathat()` exists. I can write a libglnx version of `realpathat()` if you’re concerned about that.

cgwalters:
Ah sorry, I hadn't realized https://github.com/ostreedev/ostree/pull/1172 would conflict - can you rebase?

pwithnall:
> Ah sorry, I hadn't realized #1172 would conflict - can you rebase?

Rebased.

src/libostree/ostree-repo-finder-mount.c
cgwalters: We could canonicalize by making a hashset - maybe add `ostree_repo_hash` too since you just added `_equals()`?  Anyways, doesn't have to be done in this patch.

cgwalters:
@rh-atomic-bot r+ 4842bec

cgwalters:
(PR context failure is due to mirror desync with the base image, `glibc` gets downgraded since we found an older `glibc-devel`)

cgwalters:
@rh-atomic-bot retry

src/libostree/ostree-repo-finder-mount.c
pwithnall: I’ve opened issue #1191 as a follow-up.

pwithnall:
@cgwalters and I talked about this today and decided that it probably also makes sense to hard-code searches for `/var/lib/flatpak` and `/var/lib/containers/atomic` in `OstreeRepoFinderMount`. I’ll do that on Monday.

The use case for this is for finding remotes from a live USB which contains an OS which uses separate repos for the OS and system flatpaks; or which contains a non-OSTree OS and a system flatpak repo.

PR #1181, opened 2017-09-17
Fix build failure when libarchive is not used 

yceh:
A "}"  at line 641 is missing when HAVE_LIBARCHIVE is not defined (even though probably few will use ostree without libarchive).


jlebon:
Thanks! 👍
I did some minor tweaks and also dropped the libglnx bump.

@rh-atomic-bot r+ c5b8169

PR #1182, opened 2017-09-18
create-usb: Add a create-usb command to complement OstreeRepoFinderMount

pwithnall:
This can be used to put OSTree repositories on USB sticks in a format
recognised by OstreeRepoFinderMount.

---

This depends on PR #1179 (and hence contains the same initial commits, because github can’t understand dependencies between PRs), and **must be rebased and updated after PR #1179 is pushed**.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -856,24 +856,55 @@ fetch_ref_contents (OtPullData                 *pull_data,
                     GCancellable               *cancellable,
                     GError                    **error)
 {
-  g_autofree char *filename = NULL;
-
-  if (ref->collection_id == NULL || g_strcmp0 (ref->collection_id, main_collection_id) == 0)
-    filename = g_build_filename ("refs", "heads", ref->ref_name, NULL);
-  else
-    filename = g_build_filename ("refs", "mirrors", ref->collection_id, ref->ref_name, NULL);
-
   g_autofree char *ret_contents = NULL;
-  if (!fetch_mirrored_uri_contents_utf8_sync (pull_data->fetcher,
-                                              pull_data->meta_mirrorlist,
-                                              filename, &ret_contents,
-                                              cancellable, error))
-    return FALSE;
 
+  if (pull_data->remote_repo_local != NULL && ref->collection_id != NULL)
+    {
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+      g_autoptr(GHashTable) refs = NULL;  /* (element-type OstreeCollectionRef utf8) */
+      if (!ostree_repo_list_collection_refs (pull_data->remote_repo_local,
+                                             ref->collection_id, &refs,
+                                             OSTREE_REPO_LIST_REFS_EXT_NONE,
+                                             cancellable, error))
+        return FALSE;
+
+      ret_contents = g_strdup (g_hash_table_lookup (refs, ref));
+#else  /* if !OSTREE_ENABLE_EXPERIMENTAL_API */
+      g_assert_not_reached ();
+#endif  /* !OSTREE_ENABLE_EXPERIMENTAL_API */
+    }
+  else if (pull_data->remote_repo_local != NULL)
+    {
+      g_autoptr(GHashTable) refs = NULL;  /* (element-type utf8 utf8) */
+      if (!ostree_repo_list_refs (pull_data->remote_repo_local, NULL,
+                                  &refs, cancellable, error))
+        return FALSE;
+
+      ret_contents = g_strdup (g_hash_table_lookup (refs, ref->ref_name));
+    }
+  else
+    {
+      g_autofree char *filename = NULL;
+
+      if (ref->collection_id == NULL || g_strcmp0 (ref->collection_id, main_collection_id) == 0)
+        filename = g_build_filename ("refs", "heads", ref->ref_name, NULL);
+      else
+        filename = g_build_filename ("refs", "mirrors", ref->collection_id, ref->ref_name, NULL);
+
+      if (!fetch_mirrored_uri_contents_utf8_sync (pull_data->fetcher,
+                                                  pull_data->meta_mirrorlist,
+                                                  filename, &ret_contents,
+                                                  cancellable, error))
+        return FALSE;
+    }
+
+  /* Validate and return. */
   g_strchomp (ret_contents);
 
   if (!ostree_validate_checksum_string (ret_contents, error))
-    return glnx_prefix_error (error, "Fetching %s", filename);
+    return glnx_prefix_error (error, "Fetching checksum for ref (%s, %s)",
+                              ref->collection_id ? ref->collection_id : "(empty)",
+                              ref->ref_name);
 
   ot_transfer_out_value (out_contents, &ret_contents);
   return TRUE;
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -861,26 +861,22 @@ fetch_ref_contents (OtPullData                 *pull_data,
   if (pull_data->remote_repo_local != NULL && ref->collection_id != NULL)
     {
 #ifdef OSTREE_ENABLE_EXPERIMENTAL_API
-      g_autoptr(GHashTable) refs = NULL;  /* (element-type OstreeCollectionRef utf8) */
-      if (!ostree_repo_list_collection_refs (pull_data->remote_repo_local,
-                                             ref->collection_id, &refs,
-                                             OSTREE_REPO_LIST_REFS_EXT_NONE,
-                                             cancellable, error))
+      if (!ostree_repo_resolve_collection_ref (pull_data->remote_repo_local,
+                                               ref, TRUE  /* ignore enoent */,
+                                               OSTREE_REPO_RESOLVE_REV_EXT_NONE,
+                                               &ret_contents, cancellable, error))
         return FALSE;
-
-      ret_contents = g_strdup (g_hash_table_lookup (refs, ref));
 #else  /* if !OSTREE_ENABLE_EXPERIMENTAL_API */
       g_assert_not_reached ();
 #endif  /* !OSTREE_ENABLE_EXPERIMENTAL_API */
     }
   else if (pull_data->remote_repo_local != NULL)
     {
-      g_autoptr(GHashTable) refs = NULL;  /* (element-type utf8 utf8) */
-      if (!ostree_repo_list_refs (pull_data->remote_repo_local, NULL,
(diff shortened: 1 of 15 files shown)


pwithnall:
Thinking about file systems, this either needs to check the file system of the destination repository (and check that it supports xattrs), or use a repository mode which doesn’t need xattrs (I think the only option in this case is archive-z2; we can’t use bare-user-only because we can’t lose the permissions since the repository isn’t a leaf checkout).

Colin, would it make sense for this tool to potentially eventually grow a `--format` option which wipes the USB stick before creating the repository? Or would you prefer to keep that separate?

pwithnall:
> Thinking about file systems, …

I’ve pushed fixups which check whether the destination file system supports xattrs. If it does, `create-usb` uses a `bare-user` repository; otherwise, it uses `archive-z2`. If the destination repository already exists, its existing mode is used (which might cause pull failures if the existing mode is inappropriate, but I think that’s reasonable behaviour).

cgwalters:
Hm, need to think about this a bit more; big picture I'd been trying to slowly move away from the `ostree` command line as an end user thing (but obviously keep the non-`ostree admin` i.e. the repo management bits around). 

This feels like sort of an in-between; not opposed or anything, just need to think about it a bit.

pwithnall:
> ☔️ The latest upstream changes (presumably 64b23fd) made this pull request unmergeable. Please resolve the merge conflicts.

Rebased (same for the other dependent PRs).

src/libostree/ostree-repo-pull.c
```diff
@@ -856,24 +856,55 @@ fetch_ref_contents (OtPullData                 *pull_data,
...
+#endif  /* !OSTREE_ENABLE_EXPERIMENTAL_API */
+    }
+  else if (pull_data->remote_repo_local != NULL)
+    {
+      g_autoptr(GHashTable) refs = NULL;  /* (element-type utf8 utf8) */
+      if (!ostree_repo_list_refs (pull_data->remote_repo_local, NULL,
+                                  &refs, cancellable, error))
+        return FALSE;
+
+      ret_contents = g_strdup (g_hash_table_lookup (refs, ref->ref_name));
```
cgwalters: Can't we just `ostree_repo_resolve_rev()`?

src/libostree/ostree-repo-pull.c
```diff
@@ -856,24 +856,55 @@ fetch_ref_contents (OtPullData                 *pull_data,
...
+    {
+#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
+      g_autoptr(GHashTable) refs = NULL;  /* (element-type OstreeCollectionRef utf8) */
+      if (!ostree_repo_list_collection_refs (pull_data->remote_repo_local,
+                                             ref->collection_id, &refs,
+                                             OSTREE_REPO_LIST_REFS_EXT_NONE,
+                                             cancellable, error))
+        return FALSE;
+
+      ret_contents = g_strdup (g_hash_table_lookup (refs, ref));
```
cgwalters: (See below, this reminds me we need a collection equivalent of `ostree_repo_resolve_rev()`.)

src/ostree/ot-builtin-create-usb.c
```diff
@@ -0,0 +1,275 @@
...
+
+  /* Ensure a summary file is present to make it easier to look up commit checksums. */
+  /* FIXME: It should be possible to work without this, but find_remotes_cb() in
+   * ostree-repo-pull.c currently assumes a summary file (signed or unsigned) is
+   * present. */
+  struct stat stbuf;
+  if (!glnx_fstatat_allow_noent (ostree_repo_get_dfd (dest_repo), "summary", &stbuf, 0, error))
+    return FALSE;
+  if (errno == ENOENT &&
+      !ostree_repo_regenerate_summary (dest_repo, NULL, cancellable, error))
```
cgwalters: Need to regenerate the summary always after doing a pull, right?

src/ostree/ot-builtin-create-usb.c
```diff
@@ -0,0 +1,275 @@
...
+
+          for (i = 0; i < 100; i++)
+            {
+              g_autofree char *symlink_path = g_strdup_printf (".ostree/repos.d/%02u-generated", i);
+
+              int ret = TEMP_FAILURE_RETRY (symlinkat (relative_dest_repo_path, mount_root_dfd, symlink_path));
+              if (ret < 0 && errno != EEXIST)
+                return glnx_throw_errno_prefix (error, "symlinkat(%s → %s)", symlink_path, relative_dest_repo_path);
+              else if (ret >= 0)
+                break;
```
cgwalters: There's a private `_ostree_make_temporary_symlink_at()` though it's harder to use private bits from the cmdline.  It uses `glnx_gen_temp_name()` which is visible here.

pwithnall:
Summary of what @cgwalters and I discussed today:
 * Let’s keep the addition as `ostree create-usb` for now
 * I’ll talk to people within Endless about our future plans for whether to integrate creating a USB stick into the UI (which affects whether/when we should libraryise this)
 * I’ll also talk to @alexlarsson about whether this would make sense to add to the `flatpak` command line tool too
 * We need to ensure `ostree create-usb` remains behind `--enable-experimental-api` so we have the liberty to change/break/remove it at any point

src/ostree/ot-builtin-create-usb.c
pwithnall: Not really; only when publishing the `summary`. Although the end result is equivalent.

src/libostree/ostree-repo-pull.c
pwithnall: Indeed, yes.

src/libostree/ostree-repo-pull.c
pwithnall: I’ll add a basic version of that in this PR, but it’ll basically be a splitting out of the code which is already in the PR. It can be optimised in a follow-up issue.

src/ostree/ot-builtin-create-usb.c
pwithnall: Looks fairly similar to what I’ve got in `ot-builtin-create-usb.c`. Is there anything wrong with the code I’ve got already? I’d prefer to keep the template as `%02u` since that’s a bit more human-friendly than `XXXXXX`.

cgwalters: It's fine as is, just noting.

src/libostree/ostree-repo-pull.c
cgwalters: Yep, fine by me.

pwithnall:
I’ve pushed a few updates.

src/libostree/ostree-repo-refs.c
```diff
@@ -468,6 +468,69 @@ ostree_repo_resolve_rev_ext (OstreeRepo                    *self,
...
+ * There are currently no @flags which affect the behaviour of this function.
+ *
+ * Returns: %TRUE on success, %FALSE on failure
+ * Since: 2017.12
+ */
+gboolean
+ostree_repo_resolve_collection_ref (OstreeRepo                    *self,
+                                    const OstreeCollectionRef     *ref,
+                                    gboolean                       allow_noent,
+                                    OstreeRepoResolveRevExtFlags   flags,
```
cgwalters: Maybe take this opportunity to add `OSTREE_REPO_RESOLVE_REV_EXT_ALLOW_NOENT` and drop the boolean?

pwithnall:
Note that I’m still running through the tests on these updates, so don’t push anything yet! The code is good to review though.

src/libostree/ostree-repo-refs.c
cgwalters: (Although we already shipped `ostree_repo_resolve_rev_ext()` which has both...dunno)

cgwalters:
Looks like the test case error matching will have to be adjusted:

```
++ sed -e 's/^/# /'
# error: Fetching checksum for ref ((empty), main): Invalid rev lots of html here  lots of html here  lots of html here  lots of
++ fatal 'File '\''err.txt'\'' doesn'\''t match fixed string list '\''error: Fetching refs/heads/main: Invalid rev lots of html here  lots of html here  lots of html here  lots of'\'''
```

cgwalters:
LGTM aside from that!

pwithnall:
> Looks like the test case error matching will have to be adjusted:

Indeed. The rest of the tests pass. PR updated.

pwithnall:
> * I’ll talk to people within Endless about our future plans for whether to integrate creating a USB stick into the UI (which affects whether/when we should libraryise this)

We do eventually plan to create a UI for creating a USB stick, but it’s a very long-term plan, and it’s not currently clear what shape the UI would be, or where it would be integrated.

> * I’ll also talk to @alexlarsson about whether this would make sense to add to the flatpak command line tool too

I briefly asked Alex and he seemed favourable to the idea of some kind of `flatpak create-usb` command, so that may happen at some point. I’m not going to work on that right now though.

So eventually we may end up libraryising the code in `ostree create-usb`, but I don’t think it makes sense to do that before we have a concrete plan for whacking it in a UI, for example.

cgwalters:
@rh-atomic-bot r+ da2cf63

pwithnall:
@rh-atomic-bot, retest please

cgwalters:
@rh-atomic-bot retry

cgwalters:
Indeed seems to pass locally, and I can't see an obvious error (though unfortunately g-d-t-r makes it a pain...need to use `--report-directory`)

cgwalters:
Took a quick stab at a debugging aid in https://github.com/ostreedev/ostree/pull/1218

pwithnall:
Looking at [the log](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/da2cf634219d5d5fb4b8aba4cb4ec5283cb5178e.4.1506438898273866703/output.log):

```
ok 4 adding ref to an existing usb
… (intermingled test output from parallel tests)
Imported 1 GPG key to remote "remote1"
FAIL: libostree/test-create-usb.sh.test (Child process exited with code 127)
```

I think the problem is this line in `test-create-usb.sh`:
```
${test_builddir}/repo-finder-mount finder-repo dest-mount1 org.example.Collection1 test-1 org.example.Collection1 test-2 &> out
```

Exit status 127 normally means ‘command not found’, which would fit with this.

Also from the log, we have `test_builddir=/usr/libexec/installed-tests/libostree/tests`, so it looks like I failed to make this particular test work when run installed.

Fix pushed.

pwithnall:
> All checks have passed

🎊😁🎆🎶

cgwalters:
@rh-atomic-bot r+ d49a164

cgwalters:
@rh-atomic-bot retry

PR #1183, opened 2017-09-18
Add support for pkcs11 URIs in TLS client key/certificate

OYTIS:
Unfortunately I don't know how to do it for libsoup (or if it's even possible), so this part is missing.

This patch supports the respective [change to Aktualizr](https://github.com/advancedtelematic/aktualizr/pull/353).

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -747,6 +747,16 @@ initiate_next_curl_request (FetcherRequest *req,
 
   if (self->tls_client_cert_path)
     {
+      if (g_str_has_prefix (self->tls_client_key_path, "pkcs11:"))
+        {
+          curl_easy_setopt (req->easy, CURLOPT_SSLENGINE, "pkcs11");
+          curl_easy_setopt (req->easy, CURLOPT_SSLENGINE_DEFAULT, 1L);
+          curl_easy_setopt (req->easy, CURLOPT_SSLKEYTYPE, "ENG");
+        }
+
+      if (g_str_has_prefix (self->tls_client_cert_path, "pkcs11:"))
+        curl_easy_setopt (req->easy, CURLOPT_SSLCERTTYPE, "ENG");
+
       curl_easy_setopt (req->easy, CURLOPT_SSLCERT, self->tls_client_cert_path);
       curl_easy_setopt (req->easy, CURLOPT_SSLKEY, self->tls_client_key_path);
     }
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -747,13 +747,21 @@ initiate_next_curl_request (FetcherRequest *req,
 
   if (self->tls_client_cert_path)
     {
+      /* Support for pkcs11:
+       * https://github.com/ostreedev/ostree/pull/1183
+       * This will be used by https://github.com/advancedtelematic/aktualizr
+       * at least to fetch certificates.  No test coverage at the moment
+       * though. See https://gitlab.com/gnutls/gnutls/tree/master/tests/pkcs11
+       * and https://github.com/opendnssec/SoftHSMv2 and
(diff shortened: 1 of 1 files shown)


cgwalters:
Please match the existing GNU code style (brace on newline, space between identifier and paren etc.)

Offhand this seems reasonable - but do you have any links for more information about this?  Preferably patches to existing applications.  Once we have those links let's also add them in a code comment above.

Some quick searching found https://bugzilla.redhat.com/show_bug.cgi?id=1236526 that was part of a Fedora effort to enable these types of URIs; I'm not sure how far it got.

I'm not an expert in this area but my brief understanding was that gnutls was farthest ahead here...what TlS library did you build libcurl with?

I'm guessing offhand that you want to make use of either a hardware crypto or a distinct software enclave; if you can share that information I think it'd be useful.

Unit testing this is going to be interesting...maybe https://github.com/p11-glue/p11-kit/tree/master/p11-kit has some scaffolding we can use?

OYTIS:
@cgwalters 
> Please match the existing GNU code style (brace on newline, space between identifier and paren etc.)

Done (hopefully).

> Offhand this seems reasonable - but do you have any links for more information about this? Preferably patches to existing applications. Once we have those links let's also add them in a code comment above.

The patch for [aktualizr](https://github.com/advancedtelematic/aktualizr/) is work in progress, I'll make clean-ups and add the link to the PR here.

> I'm not an expert in this area but my brief understanding was that gnutls was farthest ahead here...what TlS library did you build libcurl with?

I was working with OpenSSL + engine_pkcs11, just because OpenSSL has their nice tool for debugging. But it shouldn't theoretically depend on TLS backend.

> I'm guessing offhand that you want to make use of either a hardware crypto or a distinct software enclave; if you can share that information I think it'd be useful.

Right now we're using [SoftHSM](https://github.com/opendnssec/SoftHSMv2) for testing. The actual hardware module is not yet decided, we hope to support all of them as long as they have pkcs11 interface. The whole end-to-end testing harness (apart from aktualizr we need to patch our yocto/OE layer, [meta-updater](https://github.com/advancedtelematic/meta-updater)) is not yet finalized as well. But it involves importing key/certificate pair into SoftHSM.

> Unit testing this is going to be interesting...maybe https://github.com/p11-glue/p11-kit/tree/master/p11-kit has some scaffolding we can use?

Good idea. I was using SoftHSM to test the whole stuff (aktualizr + libostree), but we can take a closer look at p11-kit

OYTIS:
@cgwalters 
I've added [the link](https://github.com/advancedtelematic/aktualizr/pull/353) to Aktualizr's PR to the initial comment.
I'm still not sure about unit testing. SoftHSM still seems to be the most straightforward approach, but I don't know how integrate this with OSTree's test harness. I'll take a closer look at the latter.

src/libostree/ostree-fetcher-curl.c
```diff
@@ -747,6 +747,16 @@ initiate_next_curl_request (FetcherRequest *req,
 
   if (self->tls_client_cert_path)
     {
+      if (!strncmp (self->tls_client_key_path, "pkcs11:", 7))
```
jlebon: Note there's a nice `g_str_has_prefix` function for this!

OYTIS: Thanks!

cgwalters:
I was looking at gnutls for a different PR and happened to see they have https://gitlab.com/gnutls/gnutls/tree/master/tests/pkcs11 - I only glanced at it though.

cgwalters:
OK, we don't need to block on figuring out unit testing for this yet.  I'll assume you've tested it locally though.

I added a fixup commit that adds a comment with more information.

@rh-atomic-bot r+ d00514a

Would appreciate though of course if you get a chance to drop in any more information about the test framework you're using.

BTW, at least for flatpak we started building and running a stable flatpak snapshot as a CI context; we could consider doing the same for aktualizr.   Offhand I don't know how involved that'd be, it involves some maintenance work but the obvious win is that we don't have to duplicate the tests.

Issue #1184, opened 2017-09-18
fsync on by default even with txn


cgwalters:
We've had lots of issues with this over time...anyways I was debugging a segfault from what I thought was some outstanding work and saw the stack below.

Basically we're doing `fsync()` on object commit even if we're in a txn by default still.   I think this is pretty broken.  Actually embarassing since we've done all this work on using `syncfs()` and the per-boot-id staging directory etc. that is much less useful if we're doing individual syncs.

<details>
        Signal: 11 (SEGV)
     Timestamp: Mon 2017-09-18 15:17:02 UTC (1min 56s ago)
  Command Line: ostree --repo=mnt/repo pull-local /ostree/repo f8595ef11ece381bea837d470fd5baf1a2b2301d8300029bd021d4b9b3a00da7
    Executable: /usr/bin/ostree
 Control Group: /user.slice/user-1000.slice/session-4.scope
          Unit: session-4.scope
         Slice: user-1000.slice
       Session: 4
     Owner UID: 1000 (vagrant)
       Boot ID: a2d1e14193434c3aada9135ed5694f4c
    Machine ID: 60cab8c8a6b84d178c3f175dfcf30e1b
      Hostname: localhost.localdomain
       Storage: /var/lib/systemd/coredump/core.ostree.0.a2d1e14193434c3aada9135ed5694f4c.3106.1505747822000000.lz4
       Message: Process 3106 (ostree) of user 0 dumped core.
                
                Stack trace of thread 3109:
                #0  0x00007f2dbd9ef88b import_via_hardlink_is_possible (libostree-1.so.1)
                #1  0x00007f2dbd9ef934 ostree_repo_import_object_from_with_trust (libostree-1.so.1)
                #2  0x00007f2dbda00aeb import_one_local_content_object_sync (libostree-1.so.1)
                #3  0x00007f2dbda00d37 async_import_in_thread (libostree-1.so.1)
                #4  0x00007f2dbc939086 g_task_thread_pool_thread (libgio-2.0.so.0)
                #5  0x00007f2dbc3bbf00 g_thread_pool_thread_proxy (libglib-2.0.so.0)
                #6  0x00007f2dbc3bb536 g_thread_proxy (libglib-2.0.so.0)
                #7  0x00007f2dbbe7f36d start_thread (libpthread.so.0)
                #8  0x00007f2dbbbb7bbf __clone (libc.so.6)
                
                Stack trace of thread 3107:
[...]

PR #1186, opened 2017-09-18
 lib/repo: Don't syncfs or fsync() dirs if fsync opt is disabled

cgwalters:
There are use cases for not syncing at all; think build cache repos, etc. Let's
be consistent here and make sure if fsync is disabled we do no sync at all.

I chose this opportunity to add tests using the shiny new strace fault
injection.  I can forsee using this for a lot more things, so I made
the support for detecting things generic.

Related: #1184

### DIFF
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -19,7 +19,7 @@
 
 set -euo pipefail
 
-echo "1..$((72 + ${extra_basic_tests:-0}))"
+echo "1..$((73 + ${extra_basic_tests:-0}))"
 
 $CMD_PREFIX ostree --version > version.yaml
 python -c 'import yaml; yaml.safe_load(open("version.yaml"))'
@@ -809,8 +809,11 @@ cd ${test_tmpdir}
 rm -rf test2-checkout
 mkdir -p test2-checkout
 cd test2-checkout
-touch should-not-be-fsynced
-$OSTREE commit ${COMMIT_ARGS} -b test2 -s "Unfsynced commit" --fsync=false
+echo 'should not be fsynced' > should-not-be-fsynced
+if ! skip_one_without_strace_fault_injection; then
+    ${CMD_PREFIX} strace -o /dev/null -f -e inject=syncfs,fsync,sync:error=EPERM ostree --repo=${test_tmpdir}/repo commit ${COMMIT_ARGS} -b test2 -s "Unfsynced commit" --fsync=false
+    echo "ok fsync disabled"
+fi
 
 # Run this test only as non-root user.  When run as root, the chmod
 # won't have any effect.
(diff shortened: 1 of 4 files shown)


cgwalters:
> strace: ptrace(PTRACE_TRACEME, ...): Operation not permitted

Ah, right 😦

tests/basic-test.sh
```diff
@@ -809,8 +809,11 @@ cd ${test_tmpdir}
 rm -rf test2-checkout
 mkdir -p test2-checkout
 cd test2-checkout
-touch should-not-be-fsynced
-$OSTREE commit ${COMMIT_ARGS} -b test2 -s "Unfsynced commit" --fsync=false
+echo 'should not be fsynced' > should-not-be-fsynced
+if ! skip_one_without_strace_fault_injection; then
+    ${CMD_PREFIX} strace -o /dev/null -f -e inject=syncfs,fsync,sync:error=EPERM ostree --repo=${test_tmpdir}/repo commit ${COMMIT_ARGS} -b test2 -s "Unfsynced commit" --fsync=false
+    echo "ok fsync disabled"
+fi
```
jlebon: Maybe another test here for the opposite? I.e. checking that without `--fsync=false`, we fail here with `EPERM`.

tests/libtest.sh
```diff
@@ -540,6 +540,29 @@ skip_without_user_xattrs () {
     fi
 }
 
+# https://brokenpi.pe/tools/strace-fault-injection
+_have_strace_fault_injection=''
+have_strace_fault_injection() {
+    if test "${_have_strace_fault_injection}" = ''; then
+        if strace -P ${test_srcdir}/libtest-core.sh -e inject=read:retval=0 cat ${test_srcdir}/libtest-core.sh >out.txt &&
```
jlebon: This is pretty cool!

tests/libtest.sh
```diff
@@ -540,6 +540,29 @@ skip_without_user_xattrs () {
...
+           test '!' -s out.txt; then
+            _have_strace_fault_injection=0
```
jlebon: I think I would change this to the less confusing `=1` and `=0` in the `else` branch, and then make the last line of the function `test "${_have_strace_fault_injection}" = '1'`.

tests/libtest.sh
```diff
@@ -512,9 +512,21 @@ os_repository_new_commit ()
 }
 
 # Usage: if ! skip_one_without_user_xattrs; then ... more tests ...; fi
+_have_user_xattrs=''
+have_user_xattrs() {
+    if test "${_have_user_xattrs}" = ''; then
+        touch test-xattrs
+        if setfattr -n user.testvalue -v somevalue test-xattrs 2>/dev/null; then
+            _have_user_xattrs=0
```
jlebon: And similarly here too (re. swapping `=1` and `=0`). It's just much easier to reason about I find.

cgwalters:
Fixups ⬆️

jlebon:
@rh-atomic-bot r+ f524af7

jlebon:
@rh-atomic-bot retry

PR #1187, opened 2017-09-18
lib/repo: Minor cleanup to object import function

cgwalters:
We have a lot of layers of abstraction here; let's fold in the `trusted`
conditional into the call, since that's all the public API we're using does
anyways.

Prep for a future patch around object copying during imports.


jlebon:
@rh-atomic-bot r+ 36aebd9

jlebon:
@rh-atomic-bot retry

PR #1188, opened 2017-09-19
tests: Explicitly unset LANGUAGE after setting LC_ALL

smcv:
As a GNU extension, LANGUAGE takes precedence over LC_ALL for
gettext(3) whenever the locale is not C, causing tests that grep for
specific English strings to fail when run in non-English locales.
The upstream glibc proposal for C.UTF-8 would give C.UTF-8 the same
special case as C here, but the implementation in Debian does not
currently have this, so we have to unset LANGUAGE too.

### DIFF
--- a/tests/libtest-core.sh
+++ b/tests/libtest-core.sh
@@ -40,6 +40,8 @@ if locale -a | grep C.UTF-8 >/dev/null; then
 else
     export LC_ALL=C
 fi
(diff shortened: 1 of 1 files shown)


jlebon:
This looks sane to me. Let's just wait for https://github.com/ostreedev/ostree/pull/1190 to get in.

jlebon:
@rh-atomic-bot r+ c0dff23

PR #1189, opened 2017-09-19
Add support for GnuTLS for hashing

jlaako:
Support for GnuTLS is also partially preparatory work for supporting PKCS#7 signatures using X.509 keys instead of GnuPG through gpgme.

There is also a separate build fix patch for misplaced #ifdef in the sources that incorrectly crosses control block.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -329,15 +329,29 @@ AS_IF([ test x$with_smack = xyes], [
 ])
 AM_CONDITIONAL(USE_SMACK, test $with_smack != no)
 
+dnl crypto
+AC_ARG_WITH(crypto,
+AS_HELP_STRING([--with-crypto], [Choose library for checksums, one of glib, openssl, gnutls (default: glib)]),
+:, with_crypto=glib)
+
+AS_IF([test $with_crypto = glib],
+      [],
+      [test $with_crypto = openssl],
+      [with_openssl=yes],
+      [test $with_crypto = gnutls],
+      [],
+      [AC_MSG_ERROR([Invalid --with-crypto $with_crypto])]
+      )
+
 dnl begin openssl (really just libcrypto right now)
+dnl Note this option is now deprecated in favor of --with-crypto=openssl
 OPENSSL_DEPENDENCY="libcrypto >= 1.0.1"
 AC_ARG_WITH(openssl,
-AS_HELP_STRING([--with-openssl], [Enable use of OpenSSL libcrypto (checksums)]),
-:, with_openssl=no)
-
+AS_HELP_STRING([--with-openssl], [Enable use of OpenSSL libcrypto (checksums)]),with_openssl=$withval,with_openssl=no)
 AS_IF([ test x$with_openssl != xno ], [
-      PKG_CHECK_MODULES(OT_DEP_OPENSSL, $OPENSSL_DEPENDENCY)
+      PKG_CHECK_MODULES(OT_DEP_CRYPTO, $OPENSSL_DEPENDENCY)
       AC_DEFINE([HAVE_OPENSSL], 1, [Define if we have openssl])
+      with_crypto=openssl
       with_openssl=yes
 ], [
       with_openssl=no
@@ -345,6 +359,17 @@ AS_IF([ test x$with_openssl != xno ], [
 if test x$with_openssl != xno; then OSTREE_FEATURES="$OSTREE_FEATURES openssl"; fi
 AM_CONDITIONAL(USE_OPENSSL, test $with_openssl != no)
 dnl end openssl
+
+dnl begin gnutls; in contrast to openssl this one only
+dnl supports --with-crypto=gnutls
+GNUTLS_DEPENDENCY="gnutls >= 3.5.0"
+AS_IF([ test $with_crypto = gnutls ], [
+      PKG_CHECK_MODULES(OT_DEP_CRYPTO, $GNUTLS_DEPENDENCY)
+      AC_DEFINE([HAVE_GNUTLS], 1, [Define if we have gnutls])
+      OSTREE_FEATURES="$OSTREE_FEATURES gnutls"
+])
+AM_CONDITIONAL(USE_GNUTLS, test $with_crypto = gnutls)
+dnl end gnutls
 
 dnl Avahi dependency for finding repos
 AVAHI_DEPENDENCY="avahi-client >= 0.6.31 avahi-glib >= 0.6.31"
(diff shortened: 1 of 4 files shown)


src/ostree/ot-builtin-commit.c
```diff
@@ -631,12 +631,12 @@ ostree_builtin_commit (int argc, char **argv, GCancellable *cancellable, GError
                   if (!ostree_repo_import_archive_to_mtree (repo, &opts, archive, mtree,
                                                             modifier, cancellable, error))
                     goto out;
-                }
 #else
               g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED,
                            "This version of ostree is not compiled with libarchive support");
               return FALSE;
 #endif
+                }
```
jlebon: Note that this was fixed in https://github.com/ostreedev/ostree/pull/1181, so you should be able to just rebase and drop that one out.

jlebon:
bot, test pull request

cgwalters:
What's the rationale behind gnutls *and* libgcrypt?  Which HTTP backend are you planning to use, and if it's libcurl, is it also configured for gnutls?

Can you include the PR header text in the commit message itself too?  Really as much real-world information in the commit messages the better; github PR conversations may not be around 10 years later.

We'll also want to add a CI context for this; I can take care of that though.

jlaako:
@jlebon I'll rebase the PR and drop the second patch.

@cgwalters rationale for these things is that ultimately we would like to use ostree without gpgme for both GPLv3 and performance reasons. As I wrote in a mailing list post a while ago, using X.509 certificates and PKCS#7 signatures is most straightforward and standard method. For that purpose gnutls has best support and I have code for doing signing and verification with these that I now need to adapt to ostree. I was also planning to implement the same using libgcrypt (which shares the cryptographic base with gpg), but it seems quite a bit more complex to do. I anyway implemented autotools parts and hashing for both, thinking that gnutls is good for X.509/PKCS#7 and libgcrypt is good for use when using gpgme - both support hardware accelerators like AES-NI. The X.509/PKCS#7 part could also be implemented with openssl in future, but licensing wise openssl is tricky to combine (L)GPL software due to license incompatibility.

I'm not planning to make changes regarding HTTP backend part. libcurl configuration is a separate distro specific question. I personally prefer libsoup which supports gnutls and at some point I could add support for that too (of course depending on how welcome such would be)...

src/libotutil/ot-checksum-instream.c
```diff
@@ -1,3 +1,4 @@
+/* vi: set et sw=2 ts=2 cino=t0,(0: */
```
cgwalters: Let's not do this, see instead https://github.com/ostreedev/ostree/pull/1206 - presumably there's a vim equivalent?  There's also `.editorconfig` which I think I used elsewhere.

jlebon: There is. I upstreamed my copy at https://github.com/ostreedev/ostree/pull/1208. @jlaako, feel free to suggest tweaks if there's things I'm missing.

configure.ac
```diff
@@ -346,6 +346,45 @@ if test x$with_openssl != xno; then OSTREE_FEATURES="$OSTREE_FEATURES openssl";
 AM_CONDITIONAL(USE_OPENSSL, test $with_openssl != no)
 dnl end openssl
 
+dnl begin gnutls
```
cgwalters: There's currently no reason to link to more than one of (openssl, gnutls, libgcrypt) right?  So maybe we should instead have `--with-crypto=` ?  And then the default value would be `glib`.

src/libotutil/ot-checksum-instream.c
```diff
@@ -68,6 +84,12 @@ ot_checksum_instream_class_init (OtChecksumInstreamClass *klass)
 
   object_class->finalize = ot_checksum_instream_finalize;
   stream_class->read_fn = ot_checksum_instream_read;
+
+# if defined(HAVE_GCRYPT)
+  g_assert (gcry_check_version (GCRYPT_VERSION));
+  gcry_control (GCRYCTL_DISABLE_SECMEM, 0);
+  gcry_control (GCRYCTL_INITIALIZATION_FINISHED, 0);
```
cgwalters: Not your fault - this bit is recommended by the docs, but I have to say this is ugly.  Do we really need to do this dance to just use the checksums API?

Also, doesn't it potentially conflict with using libgpgme in process to e.g. sign things?  Hopefully calling `GCRYCTL_DISABLE_SECMEM` doesn't like deliberately do a `open('/tmp/myprivategpgkey').write(...).chmod(0644)` or whatever but still...

cgwalters: It sounds like you are planning to use gnutls here and not libgcrypt; if that's the case can we just omit that for now?

jlaako: Yes, unfortunately it is necessary, otherwise some of the hash stuff fails on "make check" with GPG_ERR_NOT_OPERATIONAL...

I can remove the libgcrypt parts if it is seen best alternative. Initially gnutls seems to be most convenient way to do the certificate and signature functionality. And it is sort of supported by the fact that libsoup uses it too (when/if used with ostree).

I call DISABLE_SECMEM to avoid issues when doing containerized builds or builds on restricted environments that don't allow mlock() and similar. My thinking was that build servers are not high risk environments for key leakage and at client side public keys are not sensitive.

configure.ac
jlaako: I can try to look into this, but I'm not so great on autotools...

src/libotutil/ot-checksum-instream.c
jlaako: OK, I hope vim knows how to deal with it. I'll remove it.

configure.ac
cgwalters: No worries, I can do it.

cgwalters:
OK, I updated your PR.   Can you take a look to see if the changes still look good to you? Feel free to reword/rewrite/extend the commit message if you want, etc.

jlaako:
Looks good to me!

cgwalters:
Great, thanks for the patch!  Can you file an issue beforehand for the bigger work in https://github.com/ostreedev/ostree/pull/1189#issuecomment-330767879 if you're planning to do it anytime soon, so we can discuss architecture there?  I think I mentioned this somewhere else, but one plan might be to add hooks such that signature verification could be done by a higher level app/library too.  (There are advantages/disadvantages to this of course)

@rh-atomic-bot r+ f694c38

PR #1190, opened 2017-09-19
ci: Use distro-sync instead of upgrade

cgwalters:
This should help ensure that if installing `glibc-devel` downgrades `glibc`
we'll run through our hackround for
<https://bugzilla.redhat.com/show_bug.cgi?id=1394862>.


jlebon:
@rh-atomic-bot r+ 8a11b99

PR #1192, opened 2017-09-19
tests: Reset umask to 022 while creating test repository

smcv:
In test-basic-root.sh we make assertions about the permissions
of files like baz/cow, which were created without an explicit chmod.
We can't do that unless we control the permissions.

For some reason the "debomatic" autobuilder used to do some Debian
archive rebuilds does the entire build including build-time tests
as uid 0 with umask 002, which broke those assertions. This seems
a weird thing to do, and I've opened a bug, but it also seems
reasonable to fix this test.

This also lets us remove a couple of existing workarounds for the
same issue.

Bug-Debian: https://bugs.debian.org/876138

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -150,6 +150,8 @@ setup_test_repository () {
     export OSTREE="${CMD_PREFIX} ostree ${ot_repo}"
 
     cd ${test_tmpdir}
+    local oldumask="$(umask)"
+    umask 022
     mkdir files
     cd files
     ot_files=`pwd`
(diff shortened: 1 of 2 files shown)


smcv:
> This seems a weird thing to do, and I've opened a bug

https://github.com/debomatic/debomatic/issues/36

cgwalters:
Ah, yes.  Nice.

@rh-atomic-bot r+ bd8d666

jlebon:
@rh-atomic-bot retry

PR #1194, opened 2017-09-19
ci: Add an "all options disabled" context

cgwalters:
We keep occasionally regressing this so let's start covering it now. I'm
intentionally not running the tests since that would likely entail a lot more
conditionalizing pain.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -78,10 +78,10 @@ inherit: true
 
 context: f26-minimal
 env:
-  CONFIGOPTS: '--without-curl --without-soup --disable-gtk-doc --disable-man \
-               --disable-rust --without-libarchive --without-selinux --without-smack \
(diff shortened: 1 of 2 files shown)


jlebon:
bot, retest this please

jlebon:
> checking for OT_DEP_CURL... yes
> configure: error: Curl enabled, but libsoup is not; libsoup is needed for tests
> ### EXITED WITH CODE 1 AFTER 138s

Hmm, it doesn't seem to be picking up the `--without-curl` properly.

cgwalters:
OK, fixed ⬆️

jlebon:
bot, retest this please

jlebon:
@rh-atomic-bot r+ d3c0cbd

jlebon:
@rh-atomic-bot retry

PR #1193, opened 2017-09-19
A bit more import cleanup work

cgwalters:

PR #1197, opened 2017-09-20
Import copy fastpath

cgwalters:
Splitting the copy fastpath bit out from https://github.com/ostreedev/ostree/pull/1193

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -3184,19 +3184,15 @@ import_is_bareuser_only_conversion (OstreeRepo *src_repo,
 
 /* Returns TRUE if we can potentially just call link() to copy an object. */
 static gboolean
-import_via_hardlink_is_possible (OstreeRepo *src_repo,
-                                 OstreeRepo *dest_repo,
-                                 OstreeObjectType objtype)
+import_via_reflink_is_possible (OstreeRepo *src_repo,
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -3230,76 +3226,142 @@ copy_detached_metadata (OstreeRepo    *self,
...
           *out_was_supported = FALSE;
           return TRUE;
         }
-      else
-        g_assert_not_reached ();
-    }
 
-  if (!_ostree_repo_ensure_loose_objdir_at (self->objects_dir_fd, loose_path_buf, cancellable, error))
-    return FALSE;
+      /* This is yet another variation of glnx_file_copy_at()
```
jlebon: I'm confused, if it's a bare file, we *have* to `chown`, right? (Well, unless we're running as the id of the source object, which is probably an optimization we should teach to `glnx_file_copy_at`).

cgwalters: Yep, realized that after submitting. Will fix soon.

cgwalters: (To clarify we only need to chown for bare repos)

cgwalters:
bot, retest this please

jlebon:
@rh-atomic-bot r+ e9d7273

PR #1199, opened 2017-09-20
ci: don't run tests on overlayfs/tmpfs

jlebon:
It occurred to me that a simple way of getting coverage for tests that don't work on overlayfs/tmpfs is to simply run them in a directory mounted from the host.

To do this, we switch `f26-primary` to execute on a test node rather than in a container, and make use of the new `TEST_TMPDIR` var to point it at the mounted `/srv` dir.

We also have to tweak the overlayfs detection logic only check the filesystem in which we're actually running the test, rather than just `/`.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -6,8 +6,8 @@ branches:
 required: true
 context: f26-primary
 
-container:
-    image: registry.fedoraproject.org/fedora:26
+host:
+    distro: fedora/26/atomic
 
 env:
     # Enable all the sanitizers for this primary build.
@@ -17,11 +17,15 @@ env:
     GI_SCANNERFLAGS: '--warn-error'
     ASAN_OPTIONS: 'detect_leaks=0'  # Right now we're not fully clean, but this gets us use-after-free etc
     # TODO when we're doing leak checks: G_SLICE: "always-malloc"
+    TEST_TMPDIR: /srv
 
 tests:
-  - ci/ci-commitmessage-submodules.sh
-  - ci/build-check.sh
-  - ci/ci-release-build.sh
+  - docker run -e TEST_TMPDIR -e CFLAGS -e GI_SCANNERFLAGS -e ASAN_OPTIONS
+    --privileged -v $PWD:$PWD --workdir $PWD -v /srv:/srv
+    registry.fedoraproject.org/fedora:26 sh -c
+    'ci/ci-commitmessage-submodules.sh &&
+     ci/build-check.sh &&
+     ci/ci-release-build.sh'
 
 timeout: 30m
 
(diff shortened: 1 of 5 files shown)


.papr.yml
```diff
@@ -17,11 +17,15 @@ env:
...
     ASAN_OPTIONS: 'detect_leaks=0'  # Right now we're not fully clean, but this gets us use-after-free etc
     # TODO when we're doing leak checks: G_SLICE: "always-malloc"
+    TEST_TMPDIR: /srv
 
 tests:
-  - ci/ci-commitmessage-submodules.sh
-  - ci/build-check.sh
-  - ci/ci-release-build.sh
+  - docker run -e TEST_TMPDIR -e CFLAGS -e GI_SCANNERFLAGS -e ASAN_OPTIONS
+    --privileged -v $PWD:$PWD --workdir $PWD -v /srv:/srv
```
cgwalters: There's nothing for this change that requires `--privileged` right?  I suspect it's just SELinux labeling that we'd hit; relabeling with `chcon -R -h -t container_file_t $PWD` and doing the same with `TEST_TMPDIR` would let us avoid being privileged, and that in turn would help guarantee these tests can be run in Kube for example.

jlebon:
Woohoo!

Before:

```
+ gnome-desktop-testing-runner -p 0 libostree/
Running test: libostree/test-pull-contenturl.sh.test
Copying gpghome to /tmp/test-tmp-libostree_test-pull-contenturl.sh.test-7NW96Y
evaluating for overlayfs...
overlayfs found; enabling OSTREE_NO_XATTRS
done
```

After:

```
+ gnome-desktop-testing-runner -p 0 libostree/
Running test: libostree/test-refs.sh.test
Copying gpghome to /srv/test-tmp-libostree_test-refs.sh.test-L1CX6Y
evaluating for overlayfs...
done
```

jlebon:
Sure, split out https://github.com/ostreedev/ostree/pull/1207.

.papr.yml
jlebon: Good point, I added a fixup to try that out. ⬇️

jlebon: Ahh, hmm, we're still hitting

```
error: lsetxattr: Permission denied
```

when trying to commit files (with a `relabelfrom` SELinux violation).

>  I suspect it's just SELinux labeling that we'd hit; relabeling with `chcon -R -h -t container_file_t $PWD` and doing the same with `TEST_TMPDIR` would let us avoid being privileged,

But we'll still be trying to re-apply all the labels when committing to a bare repo, right? Unless we teach libglnx to only update `security.selinux` if the label isn't already set to the wanted value, though that's quite a performance hit. We could hide the behaviour behind an env var, I guess? We could also put back `--privileged` (or at least `--security-opt label:disable`) and add another testsuite to make sure the in-container path doesn't break?

cgwalters:
I'd say we should unconditionally disable xattrs for `bare` mode unit tests, and do the wiring to change `test-basic.sh` to run fully as an installed test (as root).

cgwalters:
To elaborate more concretely, I propose:  Our unit-style tests i.e. `make check` which is currently also supported as "InstalledTestsSpec" style assume non-root and focus on testing `bare-user-only` and `archive`, and *optionally* `bare-user` if supported.  (Mostly the status quo, minus `bare`).

Retain executing those tests in containers (which we should also make fully unprivileged and not just the docker default).

The tests that use `bare` (which is a lot of the key ones like `test-basic.sh` and `test-pull-bare.sh`) are migrated to instead be run via `tests/installed/test-bare-unit.sh`).

cgwalters:
If you agree, I can take care of doing this.

jlebon:
Agreed, that's a logical way to split the tests.

> If you agree, I can take care of doing this.

Sounds good!

cgwalters:
Moved to https://github.com/ostreedev/ostree/pull/1217

PR #1200, opened 2017-09-20
rofiles-fuse: also pass mode for O_RDONLY

jlebon:
In the `O_RDONLY` case, we were calling `openat` without a mode
argument. However, it's perfectly legal (albeit unusual) to do
`open(O_RDONLY|O_CREAT)`. One such application that makes use of this is
`flock(1)`.

This was actually caught by `_FORTIFY_SOURCE=2`, and once we run
`rofiles-fuse` with `-f`, the message is clear:

```
*** invalid openat64 call: O_CREAT or O_TMPFILE without mode ***:
rofiles-fuse terminated
======= Backtrace: =========
/lib64/libc.so.6(+0x7c8dc)[0x7f36d9f188dc]
/lib64/libc.so.6(__fortify_fail+0x37)[0x7f36d9fbfaa7]
/lib64/libc.so.6(+0x10019a)[0x7f36d9f9c19a]
rofiles-fuse[0x401768]
...
```

Without `_FORTIFY_SOURCE`, the file gets created, but its mode is
completely random.

I ran into this while investigating https://github.com/projectatomic/rpm-ostree/pull/1003.

### DIFF
--- a/src/rofiles-fuse/main.c
+++ b/src/rofiles-fuse/main.c
@@ -313,7 +313,7 @@ do_open (const char *path, mode_t mode, struct fuse_file_info *finfo)
   if ((finfo->flags & O_ACCMODE) == O_RDONLY)
     {
       /* Read */
-      fd = openat (basefd, path, finfo->flags);
+      fd = openat (basefd, path, finfo->flags, mode);
       if (fd == -1)
         return -errno;
     }
--- a/tests/test-rofiles-fuse.sh
+++ b/tests/test-rofiles-fuse.sh
@@ -26,7 +26,7 @@ skip_without_user_xattrs
 
 setup_test_repository "bare"
 
-echo "1..7"
+echo "1..8"
 
 cd ${test_tmpdir}
 mkdir mnt
@@ -114,3 +114,6 @@ fi
 assert_file_has_content err.txt "Unable to do hardlink checkout across devices"
 
 echo "ok checkout copy fallback"
+
+# check that O_RDONLY|O_CREAT is handled correctly; used by flock(1) at least
+flock mnt/nonexistent-file echo "ok create file in ro mode"


jlebon:
Requires #1199.

tests/libtest.sh
```diff
@@ -514,8 +514,15 @@ os_repository_new_commit ()
     cd ${test_tmpdir}
 }
 
+assert_has_setfattr () {
+    if ! which setfattr 2>/dev/null; then
+        assert_no_reached "no setfattr available to determine xattr support"
```
cgwalters: Missing a `t` in `not` right?  Could just use `fatal`.

.papr.yml
```diff
@@ -17,11 +17,15 @@ env:
     GI_SCANNERFLAGS: '--warn-error'
     ASAN_OPTIONS: 'detect_leaks=0'  # Right now we're not fully clean, but this gets us use-after-free etc
     # TODO when we're doing leak checks: G_SLICE: "always-malloc"
+    TEST_TMPDIR: /srv
 
 tests:
-  - ci/ci-commitmessage-submodules.sh
-  - ci/build-check.sh
-  - ci/ci-release-build.sh
+  - docker run -e TEST_TMPDIR -e CFLAGS -e GI_SCANNERFLAGS -e ASAN_OPTIONS
```
cgwalters: Hmm...the more we do this though the harder it is to migrate tests to run in e.g. Kubernetes.  It's tricky; my recent attempt to use `strace` fault injection was blocked by the default docker seccomp policy, and going down this route would allow us to also easily disable that seccomp policy.

OTOH, ah just found [this on Kubernetes seccomp](https://github.com/kubernetes/features/issues/135).
Also what we're doing here maps to Kubernetes `emptyDir` volumes right?  

So...dunno.

jlebon:
OK, now this PR is just about the `rofiles-fuse` bug!

.papr.yml
jlebon: Yeah, I understand. Though at the same time, we really should strive for maximum coverage in our CI, right? Given that the majority of use cases of ostree are *not* on overlay/tmpfs, it seems like a big one to get right.

I think due to the position in the stack of many of the projects hooked to PAPR, we can't expect great coverage from running inside unprivileged Docker containers. I don't expect the common pattern of provisioning a VM and running privileged code there to diminish after migrating to Kubernetes/OCP, right?

tests/libtest.sh
jlebon: Thank! This is fixed in #1199.

cgwalters:
@rh-atomic-bot r+ c922c34

PR #1202, opened 2017-09-21
lib/repo-pull: Fix remote names in refspecs from non-mirror P2P pulls

pwithnall:
A set of commits to fix the situation where doing a non-mirror pull using the P2P functionality (`ostree find-remotes --pull …`) would result in the generated refs having an invalid remote name. They would use the name of the dynamic remote which the pull operation used, rather than a remote in `repo/config` which matches the collection ID for those refs.

See the commit messages for more details.

This set of commits depends on PR #1182, although I could possibly rebase it against master to avoid blocking on that. (Both PRs touch `OstreeRepoFinderMount` non-trivially.)

### DIFF
--- a/src/libostree/ostree-remote-private.h
+++ b/src/libostree/ostree-remote-private.h
@@ -35,17 +35,24 @@
 
 G_BEGIN_DECLS
 
+/* @refspec_name is set if this is a dynamic remote. It’s the name of the static
+ * remote which this one inherits from, and is what should be used in refspecs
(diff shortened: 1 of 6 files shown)


pwithnall:
Rebased now that #1182 has been merged; ready for review.

cgwalters:
This makes sense to me and the code all looks good.  I have some lingering uncertainty which probably just boils down to me needing read more of the collections code and understand it more.

@rh-atomic-bot r+ d92a020

Issue #1196, opened 2017-09-20
(process:10533): OSTree-CRITICAL **: ostree_repo_commit_transaction: assertion 'self->in_transaction == TRUE' failed


miabbott:
In a recent FAHC compose (26.312), installing an `etcd` system container blew up and the trace seems to point to `ostree`

```
# rpm-ostree deploy 26776f17b576a7b6b4e71572850c2c0b643b859ff560036da0eda189e7bd7dbe                                                                                                   
Validating checksum '26776f17b576a7b6b4e71572850c2c0b643b859ff560036da0eda189e7bd7dbe'
2 metadata, 0 content objects fetched; 852 B transferred in 0 seconds
21 metadata, 0 content objects fetched; 8 KiB transferred in 0 seconds

779 metadata, 3193 content objects fetched; 161006 KiB transferred in 10 seconds                                                                                                                                   
Copying /etc changes: 24 modified, 0 removed, 78 added
Transaction complete; bootconfig swap: yes deployment count change: 0
Upgraded:
  bubblewrap 0.1.8.21-8ee3ca50e7c943ba4c99cdd5741857ecad244a60.b5b2429fefc4662c2cfedfb377a22e3480d261f6.fc26 -> 0.1.8.23-e98443065f9abc3c5e2ee52320ace88340a33108.b5b2429fefc4662c2cfedfb377a22e3480d261f6.fc26
  container-storage-setup 0.6.0-1.giteb688d4.fc26 -> 0.7.0-1.git4ca59c5.fc26
  emacs-filesystem 1:25.3-1.fc26 -> 1:25.3-3.fc26
  kernel 4.12.12-300.fc26 -> 4.12.13-300.fc26
  kernel-core 4.12.12-300.fc26 -> 4.12.13-300.fc26
  kernel-modules 4.12.12-300.fc26 -> 4.12.13-300.fc26
  libgcc 7.1.1-3.fc26 -> 7.2.1-2.fc26
  libgomp 7.1.1-3.fc26 -> 7.2.1-2.fc26
  libstdc++ 7.1.1-3.fc26 -> 7.2.1-2.fc26
  ostree 2017.11.6-2f22e5b3599a216f557c9c63eab5494ccd9e2a06.5f781f21a11ef652a476cc05c2923c1e08204b2b.fc26 -> 2017.11.20-e3c3ec5dd91492e82c79223052443d038c60f41c.5f781f21a11ef652a476cc05c2923c1e08204b2b.fc26
[...]

PR #1204, opened 2017-09-21
lib/repo: Use correct name for tmpdir lockfile

cgwalters:
Such an evil bug 🙈. I was just reading an strace trying to figure out what was
going on, and noticed we had the `XXXXXX` in the lockfile name. It was only
after that I realized that that this might *be* the cause of the skopeo issue.

This is another case where we definitely need more test coverage of things that
actually use the API multiple times in process; might look at dusting off the
work for the rpm-ostree test.

Closes: https://github.com/ostreedev/ostree/issues/1196


jlebon:
Hah, nice! @rh-atomic-bot r+ 0c7182d

Issue #1191, opened 2017-09-19
Add ostree_repo_hash() and use it to canonicalise sets of repos


pwithnall:
As suggested in https://github.com/ostreedev/ostree/pull/1179#discussion_r139683710, rather than using `realpath()` on the paths of a set of repositories to try and eliminate duplicates, it would be better to add an `ostree_repo_hash()` function which can be used with `ostree_repo_equal()` with a `GHashTable` to eliminate duplicates.

`ostree_repo_hash()` could return its hash value as `g_int_hash (repo->device) ^ g_int_hash (repo->inode)`, for example. If the repository is not yet open, we could either abort, or return a fixed hash value. If we do the latter, we’ll run into problems if a repo is added to a hash table, *then* opened. The former is a nuclear option for avoiding this.

pwithnall:
I’ve put a patch in PR #1205 which implements `ostree_repo_hash()` and adds tests for it. It *doesn’t* port `OstreeRepoFinderMount` to use the new function, since it’s significantly rewritten in PR #1179, so this issue shouldn’t be closed until 1179 is in and another PR can be opened to use `ostree_repo_hash()` in the rewritten version of `OstreeRepoFinderMount`.

In the meantime, though, PR #1205 can be pushed independently of #1179.

pwithnall:
Hah, actually, the `realpath()` in `OstreeRepoFinderMount` is more necessary than I thought at the moment: since we’re now using `ostree_repo_open_at()` there, the result of calling `ostree_repo_get_path()` is a file in `/proc/self/fd`, which is really not useful as the path in a URI to return to the user. The `realpath()` call is currently masking that problem.

@cgwalters, could we do better with `ostree_repo_open_at()` than this? Perhaps it could do the `realpath()` call internally so that it can return a more useful value for `ostree_repo_get_path()`. Or perhaps we could introduce a new `ostree_repo_get_uri()` function which returns a canonical `file://` URI for the repository if possible, or `NULL` otherwise (if, for example, its root DFD is not addressable).

pwithnall:
I’ve got a [WIP branch here](https://github.com/pwithnall/ostree/tree/repo-finder-mount-hashing) which plays around with using `ostree_repo_hash()` in `OstreeRepoFinderMount`. It works, but I can’t eliminate the `realpath()` call and it doesn’t simplify the code at all. I don’t think it’s worth pushing unless we can work out a decent way of dropping the existing `realpath()` call.

[...]

PR #1205, opened 2017-09-21
lib/repo: Add ostree_repo_hash() and tests

pwithnall:
Add a hash function for OstreeRepo instances, which relies on the repo
being open, and hence being able to hash the device and inode of its
root directory.

Add unit tests for this and ostree_repo_equal().

https://github.com/ostreedev/ostree/issues/1191

### DIFF
--- a/tests/test-repo.c
+++ b/tests/test-repo.c
@@ -1,0 +1,168 @@
+/* -*- mode: C; c-file-style: "gnu"; indent-tabs-mode: nil; -*-
+ *
+ * Copyright © 2017 Endless Mobile, Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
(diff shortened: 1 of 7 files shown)


tests/test-repo.c
```diff
@@ -0,0 +1,168 @@
...
+  GLnxTmpDir tmpdir;  /* (owned) */
+} Fixture;
+
+static void
+setup (Fixture       *fixture,
+       gconstpointer  test_data)
+{
+  g_autoptr(GError) error = NULL;
+
+  (void) glnx_mkdtemp ("test-repo-XXXXXX", 0700, &fixture->tmpdir, &error);
```
cgwalters: Side note; ostree's test infra automatically allocates a tmpdir (and so does gnome-desktop-testing-runner).  But this is OK, it does make it easier to run just one test (which is a whole issue in itself).

cgwalters:
Nice, thanks!

@rh-atomic-bot r+ eb2f454

PR #1206, opened 2017-09-21
tree-wide: Remove Emacs modelines

cgwalters:
We added a `.dir-locals.el` in commit: 9a77017d87b74c5e2895cdd64ad098018929403f
There's no need to have it per-file, with that people might think
to add other editors, which is the wrong direction.


jlebon:
@rh-atomic-bot r+ d32e00c

PR #1207, opened 2017-09-21
tests: allow specifying tmpdir

jlebon:
Allow developers to override the default `/var/tmp` dir, which e.g. might
be on overlayfs and thus produces reduced coverage.


cgwalters:
@rh-atomic-bot r+ f5a389a

PR #1208, opened 2017-09-21
tree: add .vimrc

jlebon:
This requires the `exrc` option to be turned on.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,6 @@
+[*.[ch]]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+indent_brace_style = gnu
+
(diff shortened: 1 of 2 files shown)


cgwalters:
I remembered where `.editorconfig` came from, see
https://github.com/projectatomic/bubblewrap/blob/master/.editorconfig

Mind if I squash a copy of that in and amend the commit message?  (or you can)

jlebon:
Oh that's neat! That's a much cleaner solution.
Sure, go ahead!

cgwalters:
@rh-atomic-bot r+ 0d10ccc

jlebon:
@rh-atomic-bot r=cgwalters aad62c8

Issue #1211, opened 2017-09-23
pull --mirror from HTTP into archive doesn't verify checksums by default


cgwalters:
I now think https://github.com/ostreedev/ostree/commit//fab1e113db558cb7d6754e243919558df92d4864 
was a mistake; because it breaks the mental model that at least I'd built up that "local repos don't have checksums verified, HTTP does".

For example, a problem with this is (with that mental model in place) it's easy for people who set up mirrors like this to then do local pulls, and at that point we've done a deployment with no checksum verification.

Further, since then we did https://github.com/ostreedev/ostree/pull/671 AKA https://github.com/ostreedev/ostree/commit/3d38f03e4fd4cc20f754bb787feb0d109387f4f8 which is really most of the speed hit. 

So let's switch the default even for this case to doing checksum verification, and add `ostree pull --http-trusted`.  People who are in situations where they know they want this can find it and turn it on.

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1212

cgwalters:
Hm, now that I'm thinking about this, there's no reason we couldn't compute the checksum, and take the object as is if it passes.  That'd still avoid paying the recompression penalty.

PR #1212, opened 2017-09-25
lib/pull: Default checksum for archive mirror, add TRUSTED_HTTP flag 

cgwalters:
I now think commit fab1e11 was a mistake;
because it breaks the mental model that at least I'd built up that "local repos
don't have checksums verified, HTTP does".

For example, a problem with this is (with that mental model in place) it's easy
for people who set up mirrors like this to then do local pulls, and at that
point we've done a deployment with no checksum verification.

Further, since then we did PR #671 AKA commit 3d38f03 which is really most of
the speed hit.

So let's switch the default even for this case to doing checksum verification,
and add `ostree pull --http-trusted`. People who are in situations where they
know they want this can find it and turn it on.

Closes: #1211

### DIFF
--- a/src/ostree/ot-builtin-pull.c
+++ b/src/ostree/ot-builtin-pull.c
@@ -56,7 +56,7 @@ static GOptionEntry options[] = {
    { "require-static-deltas", 0, 0, G_OPTION_ARG_NONE, &opt_require_static_deltas, "Require static deltas", NULL },
    { "mirror", 0, 0, G_OPTION_ARG_NONE, &opt_mirror, "Write refs suitable for a mirror and fetches all refs if none provided", NULL },
    { "subpath", 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &opt_subpaths, "Only pull the provided subpath(s)", NULL },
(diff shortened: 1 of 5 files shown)


src/ostree/ot-builtin-pull.c
```diff
@@ -57,6 +58,7 @@ static GOptionEntry options[] = {
    { "mirror", 0, 0, G_OPTION_ARG_NONE, &opt_mirror, "Write refs suitable for a mirror and fetches all refs if none provided", NULL },
    { "subpath", 0, 0, G_OPTION_ARG_FILENAME_ARRAY, &opt_subpaths, "Only pull the provided subpath(s)", NULL },
    { "untrusted", 0, 0, G_OPTION_ARG_NONE, &opt_untrusted, "Do not verify checksums of local sources (always enabled for HTTP pulls)", NULL },
```
jlebon: The symmetry between the two options made me realize the description for this is wrong. It should be "**Do** verify...".

cgwalters: Ooh yes, good catch!  Rebased 🏄 and added a prep commit for this ⬇️

jlebon:
@rh-atomic-bot r+ f9b26ef

Issue #1210, opened 2017-09-22
lib/repo-finder-mount: Support /var/lib/flatpak


pwithnall:
See https://github.com/ostreedev/ostree/pull/1179#issuecomment-331489547. Add support for searching for `/var/lib/flatpak` and `/var/lib/containers/atomic` (@cgwalters is that correct?) on a mounted USB stick to `OstreeRepoFinderMount`. Ideally also add that to the unit tests.

cgwalters:
This was fixed by https://github.com/ostreedev/ostree/commit/cd70aaac11c68888f56c753f8d1fda8a7895c75f

PR #1213, opened 2017-09-25
lib/repo-finder-mount: Add var/lib/flatpak as a well-known directory

pwithnall:
Add this as an additional well-known directory which is checked on
mounted removable drives to see if it contains OSTree repos we can pull
refs from.

https://github.com/ostreedev/ostree/issues/1210

### DIFF
--- a/src/libostree/ostree-repo-finder-mount.c
+++ b/src/libostree/ostree-repo-finder-mount.c
@@ -49,7 +49,8 @@
  * enumerated, and all OSTree repositories below it will be searched, in lexical
  * order, for the requested #OstreeCollectionRefs. The names of the directories
  * below `.ostree/repos.d` are irrelevant, apart from their lexical ordering.
- * The directory `ostree/repo` will be searched after the others, if it exists.
+ * The directories `.ostree/repo`, `ostree/repo` and `var/lib/flatpak`
+ * will be searched after the others, if they exist.
  * Non-removable volumes are ignored.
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-finder-mount.c
```diff
@@ -413,15 +414,20 @@ ostree_repo_finder_mount_resolve_async (OstreeRepoFinder                  *finde
...
-                         mount_name, &mount_root_stbuf,
-                         parent_repo, repos_refs, cancellable);
+      /* Also check the well-known special-case directories in the mount.
+       * Add them after sorting, so they’re always last. */
+      const gchar * const well_known_repos[] =
+        {
+          ".ostree/repo",
+          "ostree/repo",
+          "var/lib/flatpak",
+          "var/lib/containers/atomic",
```
cgwalters: Sorry I was wrong when I mentioned this; it uses that dir for checkouts but actually uses `/ostree/repo`.

cgwalters: (Otherwise LGTM, can you force-push an update that also drops that bit from the commit msg title?  Sorry again for the incorrect info originally)

pwithnall: Force-pushed an update.

cgwalters:
@rh-atomic-bot r+ 329325d

PR #1214, opened 2017-09-26
lib/commit: Some misc porting to decl-after-stmnt

cgwalters:
Just happened to have this file open.


jlebon:
@rh-atomic-bot r+ e03e9fa

PR #1216, opened 2017-09-26
lib/commit: fix using uninitialized var

jlebon:
Noticed this while reading the code. The `child` var hasn't been
initialized yet at the time we throw this error (and even then, it's
only conditionally initialized). To be nice, let's just always calculate
the child path and pass that along.

Also do some minor style porting to decl near use.

Split out from #1170.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -2496,23 +2496,16 @@ write_directory_content_to_mtree_internal (OstreeRepo                  *self,
                                            GCancellable                *cancellable,
                                            GError                     **error)
 {
-  g_autoptr(GFile) child = NULL;
-  g_autoptr(GFileInfo) modified_info = NULL;
-  g_autoptr(OstreeMutableTree) child_mtree = NULL;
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -2496,23 +2496,16 @@ write_directory_content_to_mtree_internal (OstreeRepo                  *self,
...
-
   g_assert (dir_enum != NULL || dfd_iter != NULL);
 
-  name = g_file_info_get_name (child_info);
+  const char *name = g_file_info_get_name (child_info);
   g_ptr_array_add (path, (char*)name);
 
-  if (modifier != NULL)
-    child_relpath = ptrarray_path_join (path);
+  g_autofree char *child_relpath = ptrarray_path_join (path);
```
cgwalters: IIRC I did this intentionally to avoid the `malloc` if we didn't have a commit modifier.  But eh, perhaps down the line we can tweak this to use a scratch `GString` buffer like we do in other places.

cgwalters:
@rh-atomic-bot r+ d33828e

PR #1217, opened 2017-09-26
tests,ci: Move "test-basic" (bare mode) to installed test

cgwalters:
Our CI uses default Docker, which has SELinux labeling but is rather
evil in returning `EOPNOTSUPP` to any attempts to set `security.selinux`,
even if to the same value.

The previous fire 🔥 for this was: https://github.com/ostreedev/ostree/pull/759

The `bare` repo mode really only makes sense as uid 0, so our installed
test framework is a good match for this.  However, the unit tests *do*
work in a privileged container even as non-root, and *also* should
work on SELinux-disabled systems.  So let's teach the test framework
how to skip in those situations.

I tested this both in a priv container (my default builder) and an unpriv
container (like our CI).

At the same time, start executing the `test-basic.sh` from an installed test,
so we get better coverage than before.

This is just the start - all of the sysroot tests really need the
same treatment.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -546,6 +546,29 @@ skip_without_user_xattrs () {
     fi
 }
 
+# Skip unless SELinux is disabled, or we can relabel.
+# Default Docker has security.selinux xattrs, but returns
+# EOPNOTSUPP when trying to set them, even to the existing value.
+# https://github.com/ostreedev/ostree/pull/1182
+skip_without_no_selinux_or_relabel () {
+    cd ${test_tmpdir}
+    echo testlabel > testlabel.txt
+    selinux_xattr=security.selinux
+    if getfattr --encoding=base64 -n ${selinux_xattr} testlabel.txt >label.txt 2>err.txt; then
+        label=$(grep -E -e "^${selinux_xattr}=" < label.txt |sed -e "s,${selinux_xattr}=,,")
+        if setfattr -n ${selinux_xattr} -v ${label} testlabel.txt 2>err.txt; then
+            echo "SELinux enabled in $(pwd), and have privileges to relabel"
+            return 0
+        else
+            sed -e 's/^/# /' < err.txt >&2
+            skip "Found SELinux label, but unable to set (Unprivileged Docker?)"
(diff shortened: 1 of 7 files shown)


cgwalters:
Migrated from https://github.com/ostreedev/ostree/pull/1199

cgwalters:
OK, this should be ready for review.

tests/libtest.sh
```diff
@@ -546,6 +546,29 @@ skip_without_user_xattrs () {
     fi
 }
 
+# Skip unless SELinux is disabled, or we can relabel.
+# Default Docker has security.selinux xattrs, but returns
+# EOPNOTSUPP when trying to set them, even to the existing value.
+# https://github.com/ostreedev/ostree/pull/1182
```
jlebon: Is this the right link? I don't see anything about that there.

tests/libtest.sh
```diff
@@ -546,6 +546,29 @@ skip_without_user_xattrs () {
...
+skip_without_no_selinux_or_relabel () {
+    cd ${test_tmpdir}
+    echo testlabel > testlabel.txt
+    selinux_xattr=security.selinux
+    if getfattr --encoding=base64 -n ${selinux_xattr} testlabel.txt >label.txt 2>err.txt; then
+        label=$(grep -E -e "^${selinux_xattr}=" < label.txt |sed -e "s,${selinux_xattr}=,,")
+        if setfattr -n ${selinux_xattr} -v ${label} testlabel.txt 2>err.txt; then
+            echo "SELinux enabled in $(pwd), and have privileges to relabel"
+            return 0
```
jlebon: We should cache the result of this here too just like the others, though we can do that later!

tests/libtest.sh
cgwalters: Fixup ⬇️

tests/libtest.sh
cgwalters: Possibly, though right now it's a file-global `skip` unlike the other `skip_one`s.

jlebon:
@rh-atomic-bot r+ 42e7137

PR #1218, opened 2017-09-26
ci: Record primary context results using g-d-t-r --report-directory

cgwalters:
So the output isn't all intermingled.  Yes, need to improve g-d-t-r.

### DIFF
--- a/ci/build-check.sh
+++ b/ci/build-check.sh
@@ -6,12 +6,26 @@ set -xeuo pipefail
 dn=$(dirname $0)
 . ${dn}/libbuild.sh
 ${dn}/build.sh
+topdir=$(git rev-parse --show-toplevel)
+resultsdir=$(mktemp -d)
 make check
 make syntax-check  # TODO: do syntax-check under check
+# See comment below
+for x in test-suite.log config.log; do
+    mv ${x} ${resultsdir}
+done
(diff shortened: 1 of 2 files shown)


cgwalters:
Blah, really just need to fix g-d-t-r to have a saner option to record the output even for successful tests.

cgwalters:
> Removing gdtr-results/

Ahhh right.

ci/build-check.sh
```diff
@@ -25,3 +39,11 @@ if test -x /usr/bin/clang; then
     export CC=clang
     build
 fi
+
+# Keep this in sync with papr.yml
+# TODO; Split the main/clang builds into separate build dirs
+for x in test-suite.log config.log gdtr-results; do
+    if test -e ${x}; then
```
jlebon: This should be `test -e ${resultsdir}/${x}` right?

jlebon:
Looks like it worked! Though that last fixup is still targeting the wrong commit.

cgwalters:
"I know, I'll use regular expressions!" ... https://github.com/cgwalters/homegit/commit/db05a911c0c73e051048c847f2155940a42dcac6

jlebon:
@rh-atomic-bot r+ ee61d31

PR #1219, opened 2017-09-26
lib/pull: Add status for imported objects

cgwalters:
Followup for recent work in commits:

 - 8a7a35970928331a028ccddd04590ac300dbd26e
 - 1a9a473580e9066f5d4a969932d703b5b104b2b7

Keep track of how many objects we imported, and print that for `ostree
pull-local` (also do this even if noninteractive, like we did for `pull`).


jlebon:
@rh-atomic-bot r+ bce9cb2

PR #1221, opened 2017-09-27
    pull: Add support for basic auth

cgwalters:
This has been requested a few times for people delivering
OSTree content and wanting to do access control.

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -261,9 +261,8 @@ destroy_and_unref_source (GSource *source)
 }
 
 static char *
-request_get_uri (FetcherRequest *req, guint idx)
+request_get_uri (FetcherRequest *req, SoupURI *baseuri)
 {
-  SoupURI *baseuri = req->mirrorlist->pdata[idx];
   if (!req->filename)
     return soup_uri_to_string (baseuri, FALSE);
   { g_autofree char *uristr = soup_uri_to_string (baseuri, FALSE);
@@ -715,7 +714,8 @@ initiate_next_curl_request (FetcherRequest *req,
 
   g_assert_cmpint (req->idx, <, req->mirrorlist->len);
 
-  { g_autofree char *uri = request_get_uri (req, req->idx);
+  SoupURI *baseuri = req->mirrorlist->pdata[req->idx];
+  { g_autofree char *uri = request_get_uri (req, baseuri);
     curl_easy_setopt (req->easy, CURLOPT_URL, uri);
   }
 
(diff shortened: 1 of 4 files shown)


kleest:
Any update on this?

Mydayyy:
Hello kleest,

we had the same problem. As the author of this commit fell asleep last september we searched for a workaround. Luckily, we just found one. 

You can just place the basic auth on the server inside the url for ostree 

Regards
Mydayyy ( Someone who actually finishes pull requests )

src/libostree/ostree-fetcher-curl.c
```diff
@@ -771,6 +771,9 @@ initiate_next_curl_request (FetcherRequest *req,
   if ((self->config_flags & OSTREE_FETCHER_FLAGS_TRANSFER_GZIP) > 0)
     curl_easy_setopt (req->easy, CURLOPT_ACCEPT_ENCODING, "");
 
+  /* If we have e.g. basic auth in the URL string, let's honor that */
+  curl_easy_setopt (req->easy, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
```
dnohales: Take note that when the URL is set to the curl object, at some point we call `soup_uri_to_string`, which on the [documentation](https://developer.gnome.org/libsoup/stable/SoupURI.html#soup-uri-to-string) specifically says:

> Note that the output will never contain a password, even if `uri` does.

I'd suggest to set the username and password manually like this:

```suggestion
  const char *username = soup_uri_get_user (req->mirrorlist->pdata[req->idx]);
  curl_easy_setopt (req->easy, CURLOPT_USERNAME, username);

  const char *password = soup_uri_get_password (req->mirrorlist->pdata[req->idx]);
  curl_easy_setopt (req->easy, CURLOPT_PASSWORD, password);
```

With this change it works for me for a protected repo.

dnohales:
Any updates?

src/libostree/ostree-fetcher-curl.c
palvarez89: This worked for me too, thanks!!

jlebon:
@cgwalters This code looks pretty good to me! WDYT about taking the suggested change and just getting this in?

src/libostree/ostree-fetcher-curl.c
cgwalters: Thanks a lot for taking the time to track that down!  I ended up using that suggestion but factored out a variable for the URI that we use once.

cgwalters:
OK rebased :surfing_man:  and updated with (tweaked) suggestion.

PR #1223, opened 2017-09-27
lib/diff: Add compile-time ABI check on 64 bit arches

cgwalters:
Like what was done for most of the `ostree-repo.h` values.  Prep
for adding a new option.

### DIFF
--- a/src/libostree/ostree-diff.c
+++ b/src/libostree/ostree-diff.c
@@ -26,6 +26,16 @@
 #include "ostree-repo-private.h"
 #include "otutil.h"
 
+/* See ostree-repo.c for a bit more info about these ABI checks */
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-diff.c
```diff
@@ -26,6 +26,16 @@
 #include "ostree-repo-private.h"
 #include "otutil.h"
 
+/* See ostree-repo.c for a bit more info about these ABI checks */
```
jlebon: Can you add a mention here and in `ostree-repo.c` that the holes were found with `pahole`?

cgwalters: I like adding the info but how about just canonically in one place?  Fixup ⬇️

jlebon:
Yup, that works!
@rh-atomic-bot r+ a976e3f

PR #1225, opened 2017-09-27
tree-wide: Bump libglnx, port to new lockfile init

cgwalters:
In particular I'd like to get the copy fix in, since it might affect users for
the keyring bits.

Update submodule: libglnx


jlebon:
@rh-atomic-bot r+ 2a2954c

PR #1227, opened 2017-09-28
libarchive: Fix assertion error if used with CANONICAL_PERMISSIONS

cgwalters:
I was trying to do a change for rpm-ostree to use
`OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS`
for container builds with `bare-user-only,` but hit an assertion here
ultimtely because we weren't setting `standard::type`.

Rather than hand-rolling `GFileInfo` creation, use the stat buffer conversion
code which is more robust and used in multiple places already.


jlebon:
@rh-atomic-bot r+ 5d05fd3

PR #1231, opened 2017-10-01
lib/bloom: Fix bloom hashing on 32-bit architectures

pwithnall:
There was an implicit cast from guint64 to gsize (which is 32-bit on
armhf, for example) before the modulus arithmetic which safely narrows
the index.

Fix that by using a guint64 intermediate variable and making the cast
explicit.


cgwalters:
@rh-atomic-bot r+ a289944

PR #1236, opened 2017-10-02

jlebon:


cgwalters:
@rh-atomic-bot r+ 620f815

PR #1237, opened 2017-10-02
Sysroot new at prep

cgwalters:


jlebon:
@rh-atomic-bot r+ 4c9cd57

PR #1238, opened 2017-10-02
lib/pull: Fix regression with pull-local for nonexistent refs

cgwalters:
I was reading the pull code for the last release, and spotted
a bug in commit f923c2e1eaebe0c781f07d34ae1a03f94357bccd - in
the case where the ref doesn't exist, we don't set an error,
tripping an assertion in the main code.

The previous code wanted the ref to always exist, so just flip back the boolean
for "ignore noent". I moved the `g_strchomp()` just into the HTTP path - if a
local repo is corrupted in this way it's something to fix in that repo.


pwithnall:
@rh-atomic-bot r+ d9d70ae6b6750ffed17e7bfe060b9ce89d2a82ff

PR #1239, opened 2017-10-02
lib/bloom: Add some missing preconditions on n_bytes

pwithnall:
These shouldn’t change the bloom filter’s behaviour at all, but make it
a bit more obvious what the programmatical limitations are on the sizes
it can deal with.

In reality, those sizes should never be reached because they won’t fit
in a DNS-SD record.


cgwalters:
@rh-atomic-bot r+ 042292b

PR #1240, opened 2017-10-02
lib/pull: Minor cleanup to metadata scanning function, add docs

cgwalters:
I'm regretting a bit having the `guint8*csum` variant of checksums
except for the serialized form.  Once we start doing processing
it's easier to just have it remain hex.

Do an on-stack conversion for the metadata scanning function; this
drops a malloc and also just looks nicer.

Also add some long-awaited function comments to the two.


jlebon:
More high-level comments = awesomeness.
@rh-atomic-bot r+ cd9889b

PR #1241, opened 2017-10-03
lib/gpg: Print debug info when reading GPG keys

mwleeds:
This commit adds debug output whenever libostree reads GPG keys, which
can come from different locations in the file system. This is especially
helpful in debugging "GPG signatures found, but none are in trusted
keyring" errors, which in my case was caused by OSTree looking in
/usr/local/share/ostree/trusted.gpg.d/ rather than
/usr/share/ostree/trusted.gpg.d/.

### DIFF
--- a/src/libostree/ostree-gpg-verifier.c
+++ b/src/libostree/ostree-gpg-verifier.c
@@ -272,6 +272,9 @@ _ostree_gpg_verifier_add_keyring_file (OstreeGpgVerifier  *self,
 {
   g_return_if_fail (G_IS_FILE (path));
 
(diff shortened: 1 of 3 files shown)


cgwalters:
OK, though might also be useful to add something like a counter (e.g. `(3 trusted keys)`) to the error output; presumably then you'd have seen zero and that might have helped.

(FWIW I do most of my ostree development in a dev container, and `sudo make install` there, so I never use things like `/usr/local` which are really prone to breakage exactly like this.  Sometimes I do an install on top of `ostree admin unlock` on my host too)

@rh-atomic-bot r+ 83a4622

PR #1242, opened 2017-10-03
lib/repo-commit: Import detached metadata even if hardlink exists

pwithnall:
Spotted while reading through the code, it looks like the
copy_detached_metadata() call is accidentally omitted if a hardlink
already exists for the .commit object.

---

@cgwalters, this change passes the tests, but you’d better take a look and see if it fits in with what you were thinking about in commit 8a7a35970928331a028ccddd04590ac300dbd26e (PR #1197).


cgwalters:
Yep, makes sense to me.  Down the line I think we should be more intelligent about detached metadata too (particularly determining whether it changed), but this is symmetric with what we do in HTTP pulls.

Thanks for the patch!

@rh-atomic-bot r+ 66dde7a

PR #1244, opened 2017-10-03
app/main: use HAVE_LIBCURL_OR_LIBSOUP for pull CLI

jlebon:
We want `pull` to be included as long as we have at least either
`libcurl` or `libsoup` to back it. Of course, this is a moot point for
now since `libsoup` is currently a build requirement.


cgwalters:
@rh-atomic-bot r+ 4139d5f

PR #1245, opened 2017-10-03
ci: really turn on warnings

jlebon:
We didn't have `-Wall` in our `CFLAGS`. It's normally injected by
`configure.ac`, but because we *did* have `-Werror`, it was skipped.

We just turn it on from very early on in `build-check.sh` so that both
the gcc and clang builds make use of it and install a self-test to make
sure we don't regress on it.


cgwalters:
Ouch 🤕

cgwalters:
@rh-atomic-bot r+ 4a67a23

PR #1247, opened 2017-10-04
lib/utils: Port a bit to decl-after-stmt style 

cgwalters:


cgwalters:
OK, this patch grew in scope.

jlebon:
Let's split out the style porting into a seperate PR?

cgwalters:
There's conflicts though.  I guess a more proper order would be the `mmap()` first, then style?

cgwalters:
OK, this is back to just style porting, but let's do `mmap()` first and I'll rebase this one on it?

src/libotutil/ot-variant-utils.c
```diff
@@ -36,45 +37,42 @@ ot_gvariant_new_empty_string_dict (void)
...
 {
-  gpointer data_copy;
-  GVariant *ret;
-
-  data_copy = g_memdup (data, len);
-  ret = g_variant_new_from_data (G_VARIANT_TYPE ("ay"), data_copy,
+  gpointer data_copy = g_memdup (data, len);
+  GVariant *ret = g_variant_new_from_data (G_VARIANT_TYPE ("ay"), data_copy,
                                  len, FALSE, g_free, data_copy);
   return ret;
```
jlebon: Minor: we can just `return` it directly right, right?

jlebon:
@rh-atomic-bot r+ f8adf9d

PR #1250, opened 2017-10-04
lib/commit: minor coverity fix

jlebon:
Appease Coverity by using the same condition for both the ternary check
and the if-condition later on. It should be smart enough to figure out
that `dir_enum == NULL` implies that `dfd_iter != NULL` from the
assertion at the top of the function.

Coverity CID: #1457318


cgwalters:
@rh-atomic-bot r+ bef6363

PR #1251, opened 2017-10-04
Deduplicate and fix up our use of mmap()

cgwalters:
Buried in this large patch is a logical fix:

```
-  if (!map)
-    return glnx_throw_errno_prefix (error, "mmap");
+  if (map == (void*)-1)
+    return glnx_null_throw_errno_prefix (error, "mmap");
```

Which would have helped me debug another patch I was working
on.  But it turns out that actually correctly checking for
errors from `mmap()` triggers lots of other bugs - basically
because we sometimes handle zero-length variants (in detached
metadata).  When we start actually returning errors due to
this, things break.  (It wasn't a problem in practice before
because most things looked at the zero size, not the data).

Anyways there's a bigger picture issue here - a while ago
we made a fix to only use `mmap()` for reading metadata from disk
only if it was large enough (i.e. `>16k`).  But that didn't
help various other paths in the pull code and others that were
directly doing the `mmap()`.

Fix this by having a proper low level fs helper that does "read all data from
fd+offset into GBytes", which handles the size check. Then the `GVariant` bits
are just a clean layer on top of this. (At the small cost of an additional
allocation)

Side note: I had to remind myself, but the reason we can't just use
`GMappedFile` here is it doesn't support passing an offset into `mmap()`.

### DIFF
--- a/src/libotutil/ot-fs-utils.c
+++ b/src/libotutil/ot-fs-utils.c
@@ -172,7 +172,7 @@ ot_fd_readall_or_mmap (int           fd,
 
   /* http://stackoverflow.com/questions/258091/when-should-i-use-mmap-for-file-access */
   if (start > stbuf.st_size)
-    start = stbuf.st_size;
+    return g_bytes_new_static (NULL, 0);
   const gsize len = stbuf.st_size - start;
   if (len > 16*1024)
     {
--- a/src/libotutil/ot-fs-utils.c
+++ b/src/libotutil/ot-fs-utils.c
@@ -22,6 +22,7 @@
 #include "ot-fs-utils.h"
 #include "libglnx.h"
 #include <sys/xattr.h>
+#include <sys/mman.h>
 #include <gio/gunixinputstream.h>
 #include <gio/gunixoutputstream.h>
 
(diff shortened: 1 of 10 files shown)


src/ostree/ot-builtin-summary.c
```diff
@@ -217,7 +217,10 @@ ostree_builtin_summary (int argc, char **argv, GCancellable *cancellable, GError
       if (opt_raw)
         flags |= OSTREE_DUMP_RAW;
 
-      summary_data = ot_file_mapat_bytes (repo->repo_dir_fd, "summary", error);
+      glnx_fd_close int fd = -1;
+      if (!glnx_openat_rdonly (repo->repo_dir_fd, "summary", TRUE, &fd, error))
+        return FALSE;
+      summary_data = ot_fd_readall_or_mmap (fd, 0, error);
```
jlebon: This seems like a common enough pattern to offer an equivalent wrapper in `ot-fs-util.c`, no?

src/libotutil/ot-fs-utils.c
```diff
@@ -144,22 +145,57 @@ ot_dfd_iter_init_allow_noent (int dfd,
...
+                       GError      **error)
+{
+  struct stat stbuf;
+  if (!glnx_fstat (fd, &stbuf, error))
     return FALSE;
 
-  return g_mapped_file_get_bytes (mfile);
+  /* http://stackoverflow.com/questions/258091/when-should-i-use-mmap-for-file-access */
+  if (start > stbuf.st_size)
+    start = stbuf.st_size;
```
jlebon: Let's just do:

```c
return g_bytes_new_take (NULL, 0);
```

here?

src/ostree/ot-builtin-summary.c
cgwalters: Yeah, there are a few places, but it's messy since the summary path wants `GBytes*`, the detached metadata wants `GVariant` right now, and I didn't quite feel like two more wrappers.

cgwalters: Er, sorry, I thought you meant something else.  Eh...there's just two instances of that right now right?

src/libotutil/ot-fs-utils.c
cgwalters: Fixup ⬇️

src/ostree/ot-builtin-summary.c
jlebon: Oh right sorry, I meant the other one (`openat` followed by `ot_variant_read_fd`). We can do that later on though.

jlebon:
Nice! @rh-atomic-bot r+ 0765b65

PR #1252, opened 2017-10-04
lib/pull: Change fetcher to return O_TMPFILE 

cgwalters:


cgwalters:
Eh, I squashed these; not worth having them separate.

jlebon:
Nice, definitely much nicer to follow!
@rh-atomic-bot r+ 0748be3

PR #1254, opened 2017-10-05
Two more misc utils/ patches

cgwalters:


src/libotutil/otutil.c
```diff
@@ -1,40 +0,0 @@
...
- */
-
-#include "config.h"
-
-#include "otutil.h"
-
-#include <string.h>
-
-void
-ot_ptrarray_add_many (GPtrArray  *a, ...)
```
jlebon: We have to remove the declaration from the header file too, right?

jlebon:
⚡️ @rh-atomic-bot r+ d1e3186

PR #1255, opened 2017-10-06
commit: Add _CONSUME modifier flag

cgwalters:
For many cases of commit, we can actually optimize things by simply "adopting"
the object rather than writing a new copy. For example, in rpm-ostree package
layering.

We can only make that optimization though if we take ownership of the file. This
commit hence adds an API where a caller tells us to do so. For now, that just
means we `unlink()` the files/dirs as we go, but we can now later add the
"adopt" optimization.

### DIFF
--- a/bash/ostree
+++ b/bash/ostree
@@ -773,6 +773,7 @@ _ostree_commit() {
         --link-checkout-speedup
         --no-xattrs
         --orphan
+        --consume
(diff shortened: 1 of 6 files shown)


jlebon:
Interesting! @rh-atomic-bot r+ 8c616ac

PR #1256, opened 2017-10-06
lib: Add a lighter weight internal checksum wrapper

cgwalters:
The faster (OpenSSL/GnuTLS) code lived in a `GInputStream` wrapper, and that
adds a lot of weight (GObject + vtable calls). Move it into a simple
autoptr-struct wrapper, and use it in the metadata path, so we're
now using the faster checksums there too.

This also drops a malloc there as the new API does hexdigest in place to a
buffer.

Prep for more work in the commit path to avoid `GInputStream` for local file
commits, and ["adopting" files](https://github.com/ostreedev/ostree/pull/1255).

### DIFF
--- a/src/libotutil/ot-checksum-utils.c
+++ b/src/libotutil/ot-checksum-utils.c
@@ -62,7 +62,7 @@ typedef struct {
   guint digest_len;
 } OtRealChecksum;
 
-G_STATIC_ASSERT (sizeof(OtChecksum) >= sizeof(OtRealChecksum));
+G_STATIC_ASSERT (sizeof (OtChecksum) >= sizeof (OtRealChecksum));
 
 void
 ot_checksum_init (OtChecksum *checksum)
@@ -210,7 +210,7 @@ ot_gio_splice_update_checksum (GOutputStream  *out,
       char buf[4096];
       do
         {
-          if (!g_input_stream_read_all (in, buf, sizeof(buf), &bytes_read, cancellable, error))
+          if (!g_input_stream_read_all (in, buf, sizeof (buf), &bytes_read, cancellable, error))
             return FALSE;
           if (!ot_gio_write_update_checksum (out, buf, bytes_read, &bytes_written, checksum,
                                              cancellable, error))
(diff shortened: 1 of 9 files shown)


src/libotutil/ot-checksum-utils.c
```diff
@@ -119,12 +237,16 @@ ot_gio_splice_get_checksum (GOutputStream  *out,
...
 
-  if (!ot_gio_splice_update_checksum (out, in, checksum, cancellable, error))
+  if (!ot_gio_splice_update_checksum (out, in, &checksum, cancellable, error))
     return FALSE;
 
-  g_autofree guchar *ret_csum = ot_csum_from_gchecksum (checksum);
+  guint8 digest[_OSTREE_SHA256_DIGEST_LEN];
+  ot_checksum_get_digest (&checksum, digest, sizeof(digest));
+  g_autofree guchar *ret_csum = g_malloc (sizeof (digest));
+  memcpy (ret_csum, digest, sizeof (digest));
```
jlebon: Minor nit: we're not very consistent with space between the `sizeof` operator and the opening parenthesis.

src/libotutil/ot-checksum-utils.h
```diff
@@ -21,19 +21,44 @@
...
 guchar *ot_csum_from_gchecksum (GChecksum *checksum);
 
+struct OtChecksum {
+  gboolean initialized;
+  guint uints[2];
+  gpointer data[2];
+};
+typedef struct OtChecksum OtChecksum;
+
+/* Same as OSTREE_SHA256_DIGEST_LEN, but this header can't depend on that */
```
jlebon: I think I missed the part about why we can't do that. Can you expand on this comment?

cgwalters: We could, but it feels like a layering violation since conceptually `src/libostree` depends on `src/libotutil` (which depends on `libglnx/`), etc.

cgwalters: I came up with something that'll at least error out if we somehow make them different :arrow_dow

jlebon:
Time for a ⚡️! @rh-atomic-bot r+ 36c8169

PR #1258, opened 2017-10-10
lib/checkout: fallback to checksum for UNION_IDENTICAL

jlebon:
There's a subtle issue going on with the way we use `UNION_IDENTICAL`
now in rpm-ostree. Basically, the crux of the issue is that we checkout
the whole tree from the system repo, but then overlay packages by
checking out from the pkgcache repo. This is an easy way to break the
assumption that we will be merging hardlinks from the same repo.

This ends up causing issues like:
projectatomic/rpm-ostree#1047

There, `vim-minimal` is already part of the host and has an object for
`/usr/share/man/man1/ex.1.gz`. `vim-common` has that same file, but
because it's unpacked in the pkgcache repo first, the hardlinks are not
the same.

There are a few ways we *could* work around this in rpm-ostree itself,
e.g. by re-establishing hardlinks when we do the content pull into the
system repo, but it still felt somewhat hacky. Let's just do this the
proper way and fall back to checksumming the target file if needed,
which is what librpm does as well in this case. Note that we only
checksum if they're not hard links, but they're the same size.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -374,7 +374,8 @@ hardlink_add_tmp_name (OstreeRepo              *self,
 
 static gboolean
 checkout_file_hardlink (OstreeRepo                          *self,
-                        OstreeRepoCheckoutAtOptions           *options,
+                        const char                          *checksum,
+                        OstreeRepoCheckoutAtOptions         *options,
                         const char                          *loose_path,
                         int                                  destination_dfd,
                         const char                          *destination_name,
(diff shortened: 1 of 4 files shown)


cgwalters:
Mmm....probably the best fix is to circle back and address https://github.com/ostreedev/ostree/issues/262 - we definitely want dedup, and unifying the repos would allow fixing a lot of ugly stuff.

src/libostree/ostree-repo-checkout.c
```diff
@@ -436,10 +437,25 @@ checkout_file_hardlink (OstreeRepo                          *self,
...
               (src_stbuf.st_dev == dest_stbuf.st_dev &&
                src_stbuf.st_ino == dest_stbuf.st_ino);
+            if (!is_identical && (src_stbuf.st_size == dest_stbuf.st_size))
+              {
+                /* resort to a checksum check */
+                g_autofree char *abspath =
+                  glnx_fdrel_abspath (destination_dfd, destination_name);
+                g_autoptr(GFile) dest_file = g_file_new_for_path (abspath);
+
+                /* XXX: need an fd-relative version of this API */
```
cgwalters: Not too hard but would conflict with the checksum PR; we can do this after.

src/libostree/ostree-repo-checkout.c
```diff
@@ -436,10 +437,25 @@ checkout_file_hardlink (OstreeRepo                          *self,
...
+                g_autofree guchar *csum = NULL;
+                if (!ostree_checksum_file (dest_file, OSTREE_OBJECT_TYPE_FILE,
+                                           &csum, cancellable, error))
+                  return FALSE;
 
+                g_autofree char *actual_checksum = ostree_checksum_from_bytes (csum);
```
cgwalters: (That said though there's no reason not to use `ostree_checksum_inplace_from_bytes()` and lose a malloc right?)

src/libostree/ostree-repo-checkout.c
```diff
@@ -436,10 +437,25 @@ checkout_file_hardlink (OstreeRepo                          *self,
             if (!glnx_fstatat (destination_dfd, destination_name, &dest_stbuf,
                                AT_SYMLINK_NOFOLLOW, error))
               return FALSE;
-            const gboolean is_identical =
+            gboolean is_identical =
               (src_stbuf.st_dev == dest_stbuf.st_dev &&
                src_stbuf.st_ino == dest_stbuf.st_ino);
+            if (!is_identical && (src_stbuf.st_size == dest_stbuf.st_size))
+              {
+                /* resort to a checksum check */
```
cgwalters: Can you add a little bit more info here; could just be circular with a link to this PR, which I tend to do since it saves people time in `git annotate` if the code moves around.

Like:
```
  /* It might happen that the file came from another repo, as is currently the
   * case in rpm-ostree package layering.  We resort to doing a full checksum.
   * https://github.com/ostreedev/ostree/pull/1258
   */
```

jlebon:
> Mmm....probably the best fix is to circle back and address #262 - we definitely want dedup, and unifying the repos would allow fixing a lot of ugly stuff.

Yeah definitely. Actually, now that I think of it, since we no longer maintain multiple commits in each pkg branch (that was needed when to avoid potential relabeling, but we always relabel now), we shouldn't hit pruning issues anymore. So I *think* we should be able to migrate to using the system repo without too much trouble.

Regardless, the fallback makes sense on its own anyway. Will look into the CI failures.

jlebon:
Rebased. This now also depends on #1263!

jlebon:
Rebased! ⬆️

jlebon:
If checksumming during checkout is too unexpected an operation, we can make this a new flag instead to make it a conscious choice. E.g. `UNION_IDENTICAL_CHECKSUM` ?

PR #1259, opened 2017-10-11
tree-wide: Update to new libglnx fd APIs

cgwalters:
This ends up a lot better IMO.  This commit is *mostly* just
`s/glnx_close_fd/glnx_autofd`, but there's also a number of hunks like:

```
-  if (self->sysroot_fd != -1)
-    {
-      (void) close (self->sysroot_fd);
-      self->sysroot_fd = -1;
-    }
+  glnx_close_fd (&self->sysroot_fd);
```

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -488,18 +488,13 @@ ostree_repo_finalize (GObject *object)
   g_free (self->stagedir_prefix);
   g_clear_object (&self->repodir_fdrel);
   g_clear_object (&self->repodir);
-  if (self->repo_dir_fd != -1)
-    (void) close (self->repo_dir_fd);
+  glnx_close_fd (&self->repo_dir_fd);
   glnx_tmpdir_unset (&self->commit_stagedir);
   glnx_release_lock_file (&self->commit_stagedir_lock);
-  if (self->tmp_dir_fd != -1)
-    (void) close (self->tmp_dir_fd);
-  if (self->cache_dir_fd != -1)
-    (void) close (self->cache_dir_fd);
-  if (self->objects_dir_fd != -1)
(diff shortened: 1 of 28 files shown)


src/libostree/ostree-repo.c
```diff
@@ -488,18 +488,13 @@ ostree_repo_finalize (GObject *object)
...
-  if (self->cache_dir_fd != -1)
-    (void) close (self->cache_dir_fd);
-  if (self->objects_dir_fd != -1)
-    (void) close (self->objects_dir_fd);
-  if (self->uncompressed_objects_dir_fd != -1)
-    (void) close (self->uncompressed_objects_dir_fd);
+  glnx_close_fd (&self->tmp_dir_fd);
+  glnx_close_fd (&self->cache_dir_fd);
+  glnx_close_fd (&self->objects_dir_fd);
+  glnx_close_fd (&self->uncompressed_objects_dir_fd);
```
jlebon: Nice!

tests/test-repo-finder-mount.c
```diff
@@ -464,7 +464,7 @@ test_repo_finder_mount_well_known (Fixture       *fixture,
   g_autoptr(GError) error = NULL;
   g_autoptr(GList) mounts = NULL;  /* (element-type OstreeMockMount)  */
   g_autoptr(GMount) mount = NULL;
-  glnx_fd_close int repos = -1;
+  glnx_autofd int repos = -1;
```
jlebon: After looking at the full diff, the obvious thing that comes to mind is; would it make sense to just absorb the `int` into the macro? Unlike e.g. `g_autofree`, there's only a single type we would ever use this with, right? Just a thought! 💡

jlebon:
@rh-atomic-bot r+ 843c740

PR #1260, opened 2017-10-11
lib/deploy: Ignore FIFREEZE/FITHAW errors when already in state

dbnicholson:
If the filesystem is already frozen, FIFREEZE returns EBUSY, and if the
filesystem is already thawed, FITHAW returns EINVAL. It's very unlikely
these issues would arise on a real ostree system since the sysroot would
be locked during the freeze/thaw cycle.

However, when multiple fake sysroots are used during the test suite (run
as root), the tests could race to run the freeze/thaw cycle without
locking. Furthermore, there's no reason why an independent process might
be trying to freeze the filesystem while ostree was deploying. Ignore
but warn for these errors since there's not much ostree can do about it,
anyways.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1332,11 +1332,15 @@ fsfreeze_thaw_cycle (OstreeSysroot *self,
       /* Do a freeze/thaw cycle; TODO add a FIFREEZETHAW ioctl */
       if (ioctl (rootfs_dfd, FIFREEZE, 0) != 0)
         {
-          /* Not supported, or we're running in the unit tests (as non-root)?
+          /* Not supported, we're running in the unit tests (as non-root), or
+           * the filesystem is already frozen (EBUSY).
            * OK, let's just do a syncfs.
            */
-          if (G_IN_SET (errno, EOPNOTSUPP, EPERM))
+          if (G_IN_SET (errno, EOPNOTSUPP, EPERM, EBUSY))
             {
+              /* Warn if the filesystem was already frozen */
+              if (errno == EBUSY)
+                g_debug ("Filesystem already frozen, falling back to syncfs");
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1332,11 +1332,15 @@ fsfreeze_thaw_cycle (OstreeSysroot *self,
...
+          /* Not supported, we're running in the unit tests (as non-root), or
+           * the filesystem is already frozen (EBUSY).
            * OK, let's just do a syncfs.
            */
-          if (G_IN_SET (errno, EOPNOTSUPP, EPERM))
+          if (G_IN_SET (errno, EOPNOTSUPP, EPERM, EBUSY))
             {
+              /* Warn if the filesystem was already frozen */
+              if (errno == EBUSY)
+                g_warning ("Filesystem already frozen, falling back to syncfs");
```
cgwalters: I'm not a big fan of random spew from libraries.  (What spew we do I'd prefer to use `sd_journal_print` and not stderr).   Let's drop the `g_warning()` here?

I'd been thinking to at some point gather stats on how long `syncfs()` takes etc (and eventually a signal so clients can know we're doing it), so we could also log whether we used `FIFREEZE` or `syncfs()` then too.

pwithnall: I suggest s/`g_warning`/`g_debug`/ instead of just dropping it.

dbnicholson: Sure. So, no message or add a journal ID for one? Also, same for the warning when thawing?

cgwalters: `g_debug` is OK by me for now.

dbnicholson: 👍 reworking

dbnicholson:
Ready again.

cgwalters:
@rh-atomic-bot r+ 01d9f17

PR #1261, opened 2017-10-11
libotutil/checksum-utils: fix openssl compilation

jlebon:


cgwalters:
@rh-atomic-bot r+ 4f3ffaa

PR #1263, opened 2017-10-11
lib/core: add ostree_checksum_file_at API

jlebon:
This is like `ostree_checksum_file` but fd-relative. This will be used
by #1258.

AFAICT, we actually didn't have any tests that check the `checksum` CLI.
Add a basic one here to test the old code as well as the new code.


src/libostree/ostree-core.c
```diff
@@ -832,6 +832,78 @@ ostree_checksum_file (GFile            *f,
...
+      if (!glnx_fstatat (dfd, path, stbuf, AT_SYMLINK_NOFOLLOW, error))
+        return FALSE;
+    }
+
+  g_autoptr(GFileInfo) file_info = _ostree_stbuf_to_gfileinfo (stbuf);
+
+  g_autoptr(GInputStream) in = NULL;
+  if (S_ISREG (stbuf->st_mode))
+    {
+      glnx_fd_close int fd = -1;
```
cgwalters: `glnx_autofd` now.

src/libostree/ostree-core.h
```diff
@@ -420,6 +421,24 @@ gboolean ostree_checksum_file (GFile             *f,
                                GCancellable      *cancellable,
                                GError           **error);
 
+/**
+ * OstreeDiffFlags:
```
cgwalters: `ChecksumFlags`.

Also `Since: 2017.13`

tests/basic-test.sh
```diff
@@ -239,6 +239,33 @@ assert_file_has_content err.txt "No such metadata object"
...
+    rm -rf checksum-test
+    $OSTREE checkout test2 checksum-test
+    find checksum-test/ -type f | while read fn; do
+        checksum=$($CMD_PREFIX ostree checksum $CHECKSUM_FLAG $fn)
+        objpath=repo/objects/${checksum::2}/${checksum:2}.file
+        assert_has_file $objpath
+        # running `ostree checksum` on the obj might not necessarily match, let's
+        # just check that they have the same content to confirm that it's
+        # (probably) the originating file
+        object_content_checksum=$(md5sum $objpath | cut -f1 -d' ')
```
cgwalters: Why not use `sha256sum`?

src/libostree/ostree-core.c
```diff
@@ -832,6 +832,78 @@ ostree_checksum_file (GFile            *f,
...
+ *
+ * Since: 2017.13
+ */
+gboolean
+ostree_checksum_file_at (int               dfd,
+                         const char       *path,
+                         struct stat      *stbuf,
+                         OstreeObjectType  objtype,
+                         OstreeChecksumFlags flags,
+                         guchar          **out_csum,
```
cgwalters: I've been thinking about using hex for all new checksum APIs.

jlebon:
Rebased and fixup! ⬆️

src/libostree/ostree-core.c
```diff
@@ -839,12 +839,13 @@ ostree_checksum_file (GFile            *f,
  * @stbuf (allow-none): Optional stat buffer
  * @objtype: Object type
  * @flags: Flags
- * @out_csum: (out) (array fixed-size=32): Return location for binary checksum
+ * @checksum_buf (out caller-allocates): Buffer to hold SHA256 hex checksum
```
cgwalters: Although, sorry to make you rev this again...this might not be introspectable.  I think we'd still need the `(array fixed-size=65)` at least.

So...eh, I guess let's just malloc here and basically use the out-buffer approach for more internal APIs - basically where it matters more is anything we do in bulk, so it's obviously fine to malloc a commit checksum, but we should try to reduce internal allocs around writing files.

jlebon: Updated below! ⬇️

src/libostree/ostree-core.c
```diff
@@ -839,7 +839,7 @@ ostree_checksum_file (GFile            *f,
  * @stbuf (allow-none): Optional stat buffer
  * @objtype: Object type
  * @flags: Flags
- * @checksum_buf (out caller-allocates): Buffer to hold SHA256 hex checksum
+ * @out_checksum (out) (tranfer full): Return location for hex checksum
```
cgwalters: `transfer`

src/ostree/ot-builtin-checksum.c
```diff
@@ -94,9 +94,9 @@ ostree_builtin_checksum (int argc, char **argv, GCancellable *cancellable, GErro
       return data.success;
     }
 
-  char csum[OSTREE_SHA256_STRING_LEN+1];
+  g_autofree char *csum = NULL;
```
cgwalters: Let's say `checksum` now too for consistency?  (I'm not sure I ever documented the naming scheme but it's *mostly* used consistently now)

cgwalters:
@rh-atomic-bot r+ ab2462b

Issue #1262, opened 2017-10-11
FIFREEZE watchdogs not exiting from test-sysroot.js


dbnicholson:
After running `tests/test-sysroot.js`, there are always 5 `gjs` processes hanging around. After some debugging, I can see that they're the `FIFREEZE` watchdog daemons. If I include `test-fifreeze` in `OSTREE_SYSROOT_DEBUG`, then the test just hangs until `tap-test` kills everything.

The weird thing is that I see it getting a successful `poll`, in which case it calls `exit (EXIT_SUCCESS)`. But `gjs` isn't exiting. Perhaps this is a `gjs` bug. Here's the debugging code I added to see what was going on:
```
diff --git a/src/libostree/ostree-sysroot-deploy.c b/src/libostree/ostree-sysroot-deploy.c
index 920c0eb4..11b7acbb 100644
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1249,6 +1249,7 @@ fsfreeze_thaw_cycle (OstreeSysroot *self,
   char c = '!';
   if (pid == 0) /* Child watchdog/unfreezer process. */
     {
+      g_printerr("watchdog pid: %d\n", getpid());
       (void) close (glnx_steal_fd (&sock_parent));
       /* Daemonize, and mask SIGINT/SIGTERM, so we're likely to survive e.g.
        * someone doing a `systemctl restart rpm-ostreed` or a Ctrl-C of
@@ -1257,8 +1258,12 @@ fsfreeze_thaw_cycle (OstreeSysroot *self,
        */
       if (!debug_fifreeze)
         {
+          int tmp_stderr = dup (2);
           if (daemon (0, 0) < 0)
             err (1, "daemon");
+          (void) dup2 (tmp_stderr, 2);
+          (void) close (tmp_stderr);
+          g_printerr ("watchdog daemon pid: %d\n", getpid());
         }
       int sigs[] = { SIGINT, SIGTERM };
       for (guint i = 0; i < G_N_ELEMENTS (sigs); i++)
@@ -1284,6 +1289,7 @@ fsfreeze_thaw_cycle (OstreeSysroot *self,
       struct pollfd pfds[1];
       pfds[0].fd = sock_watchdog;
       pfds[0].events = POLLIN | POLLHUP;
+      g_printerr ("Polling %d\n", sock_watchdog);
[...]

PR #1264, opened 2017-10-11
lib/deploy: Use _exit() for FIFREEZE watchdog

cgwalters:
This works around an (IMO) SpiderMonkey bug - it tries to
clean up in a shared library destructor, but doesn't install a
`pthread_atfork()` handler to unset its state.

Closes: https://github.com/ostreedev/ostree/issues/1262


cgwalters:

dbnicholson:
@rh-atomic-bot r+ 5a9a1ddeb4a

dbnicholson:
Thanks for finding that! Tested locally and it works.

PR #1266, opened 2017-10-12
lib/deltas: Check cancellable during processing

cgwalters:
Let's react to `Ctrl-C` faster here. Noticed while I was doing an update on my
desktop and playing with cancellation.


jlebon:
@rh-atomic-bot r+ d9cb33e

Issue #1163, opened 2017-09-12
Describe subcommands in help output


jlebon:
This is a clone of https://github.com/projectatomic/rpm-ostree/issues/806.

On "root" commands that have subcommands, we should have a short one-liner of what each command does in the -h/--help output, rather than just listing them.

Contrast with e.g. `oc`:

```
OpenShift Client

This client helps you develop, build, deploy, and run your applications on any
OpenShift or Kubernetes compatible platform. It also includes the administrative
commands for managing a cluster under the 'adm' subcommand.

Basic Commands:
  types           An introduction to concepts and types
  login           Log in to a server
  new-project     Request a new project
  new-app         Create a new application
  status          Show an overview of the current project
  project         Switch to another project
  projects        Display existing projects
  explain         Documentation of resources
  cluster         Start and stop OpenShift cluster

Build and Deploy Commands:
  rollout         Manage a Kubernetes deployment or OpenShift deployment config
  deploy          View, start, cancel, or retry a deployment
  rollback        Revert part of an application back to a previous deployment
  new-build       Create a new build configuration
  start-build     Start a new build
  cancel-build    Cancel running, pending, or new builds
  ...
```

jlebon:
This was done in rpm-ostree:

```
[cloud-user@f26-tmp ~]$ rpm-ostree --help
Usage:
  rpm-ostree [OPTION?] COMMAND

Builtin Commands:
  compose          Commands to compose a tree
  cleanup          Clear cached/pending data
  db               Commands to query the RPM database
  deploy           Deploy a specific commit
  rebase           Switch to a different tree
  rollback         Revert to the previously booted tree
  status           Get the version of the booted system
  upgrade          Perform a system upgrade
  reload           Reload configuration
  initramfs        Enable or disable local initramfs regeneration
  install          Download and install layered RPM packages
  uninstall        Remove one or more overlay packages

Help Options:
  -h, --help       Show help options

Application Options:
  --version        Print version information and exit
```

peterbaouoft:
May I take this issue and work on it? Might take some time tho due to unfamiliarity of the code base. Thanks  :)

jlebon:
Go for it! :)

peterbaouoft:
Hmm, after reading more into the code, it feels that there are many duplicate code for subcommand handling, and for `ostree_option_context_parse()`, we are not able to know which command calls the function, making it harder to move the command description. 

[...]

PR #1267, opened 2017-10-13
ostree: Describe subcommands in help output

peterbaouoft:
The added subcommands help will be provided to the following commands:
- ostree -h
- ostree admin -h
- ostree admin instutil -h
- ostree remote -h
- ostree static-delta -h

This is very similar to https://github.com/projectatomic/rpm-ostree/pull/916

Closes: https://github.com/ostreedev/ostree/issues/1163
Demo: 
```
[root@localhost ostree]# ostree -h
Usage:
  ostree [OPTION?] COMMAND

Builtin Commands:
  admin             Commands that needs admin privilege
  cat               Concatenate contents of files
  checkout          Check out a commit into a filesystem tree
  checksum          Checksum a file or directory
  commit            Commit a new revision
  config            Change repo configuration settings
  diff              Compare directory TARGETDIR against revision REV
  export            Stream COMMIT to stdout in tar format
  fsck              Check the repository for consistency
  gpg-sign          Sign a commit
  init              Initialize a new empty repository
  log               Show log starting at commit or ref
  ls                List file paths
  prune             Search for unreachable objects
  pull-local        Copy data from SRC_REPO
  pull              Download data from remote repository
  refs              List refs
  remote            Remote commands that may involve internet access
  reset             Reset a REF to a previous COMMIT
  rev-parse         Output the target of a rev
  show              Output a metadata object
  static-delta      Static delta related commands
  summary           Manage summary metadata

Help Options:
  -h, --help        Show help options

Application Options:
  -v, --verbose     Print debug information during command processing
  --version         Print version information and exit
```

```
[root@localhost ostree]# ostree admin -h
Usage:
  ostree admin [OPTION?] --print-current-dir|COMMAND

Builtin "admin" Commands:
  cleanup            Delete untagged deployments and repository objects
  config-diff        Diff current /etc configuration versus default
  deploy             Checkout revision REFSPEC as the new default deployment
  init-fs            Initialize a root filesystem
  instutil           Provide instutil commands, allow admin to change boot configuration and relabel selinux 
  os-init            Initialize empty state for given operating system
  set-origin         Set Origin and create a new origin file
  status             List deployments
  switch             Construct new tree from REF and deploy it
  undeploy           Delete deployment INDEX
  unlock             Make the current deployment mutable (as a hotfix or development)
  upgrade            Construct new tree from current origin and deploy it, if it changed

Help Options:
  -h, --help         Show help options

Application Options:
  --sysroot=PATH     Create a new OSTree sysroot at PATH
  -v, --verbose      Print debug information during command processing
  --version          Print version information and exit
```

```
[root@localhost ostree]# ostree admin instutil -h
Usage:
  ostree admin instutil [OPTION?] COMMAND

Builtin "admin instutil" Commands:
  selinux-ensure-labeled  relabel all or part of a deployment
  set-kargs               set new kernel command line arguments(Not stable) 
  grub2-generate          generate GRUB2 configuration from given BLS entries

Help Options:
  -h, --help         Show help options

Application Options:
  --sysroot=PATH     Create a new OSTree sysroot at PATH
  -v, --verbose      Print debug information during command processing
  --version          Print version information and exit
```

```
[root@localhost ostree]# ostree remote -h        
Usage:
  ostree remote [OPTION?] COMMAND

Builtin "remote" Commands:
  add               Add a remote repository
  delete            Delete a remote repository
  show-url          Show remote repository URL
  list              List remote repository names
  gpg-import        Import GPG keys
  add-cookie        Add a cookie to remote
  delete-cookie     Remote one cookie from remote
  list-cookies      Show remote repository cookies
  refs              List remote refs
  summary           Show remote summary

Help Options:
  -h, --help        Show help options

Application Options:
  -v, --verbose     Print debug information during command processing
  --version         Print version information and exit
```
```
[root@localhost ostree]# ostree static-delta -h
Usage:
  ostree static-delta [OPTION...] COMMAND

Builtin "static-delta" Commands:
  list             list static delta files
  show             Dump information on a delta
  delete           Remove a delta
  generate         Generate static delta files
  apply-offline    Apply static delta file
```


peterbaouoft:
Looks like... the failure was because... the test-help.sh file also reads description as one of the potential commands :(. 

> _fatal_print_file out 'File '\''out'\'' doesn'\''t match regexp '\''ostree admin Delete'\'''

:'(.

jlebon:
Ahh, heh. I think you might need to tweak the `sed` so it only picks up the first word of each line? Alternatively, you can also pipe into e.g. `cut -d' '`.

src/ostree/ot-admin-builtin-instutil.c
```diff
@@ -31,15 +31,19 @@
...
   const char *name;
   gboolean (*fn) (int argc, char **argv, GCancellable *cancellable, GError **error);
+  const char *description;
 } OstreeAdminInstUtilCommand;
 
 static OstreeAdminInstUtilCommand admin_instutil_subcommands[] = {
 #ifdef HAVE_SELINUX
-  { "selinux-ensure-labeled", ot_admin_instutil_builtin_selinux_ensure_labeled },
+  { "selinux-ensure-labeled", ot_admin_instutil_builtin_selinux_ensure_labeled,
+    "relabel all or part of a deployment" },
```
jlebon: Let's capitalize these too to be consistent.

src/ostree/ot-admin-builtin-instutil.c
```diff
@@ -31,15 +31,19 @@
...
 #endif
-  { "set-kargs", ot_admin_instutil_builtin_set_kargs },
-  { "grub2-generate", ot_admin_instutil_builtin_grub2_generate },
-  { NULL, NULL }
+  { "set-kargs", ot_admin_instutil_builtin_set_kargs,
+    "set new kernel command line arguments(Not stable) "  },
```
jlebon: Minor: trailing space there!

src/ostree/ot-remote-builtin-delete-cookie.c
```diff
@@ -41,7 +41,7 @@ gboolean
 ot_remote_builtin_delete_cookie (int argc, char **argv, GCancellable *cancellable, GError **error)
 {
   g_autoptr(OstreeRepo) repo = NULL;
-  g_autoptr(GOptionContext) context = g_option_context_new ("NAME DOMAIN PATH COOKIE_NAME- Remote one cookie from remote");
+  g_autoptr(GOptionContext) context = g_option_context_new ("NAME DOMAIN PATH COOKIE_NAME- Remove one cookie from remote");
```
jlebon: It looks good overall! The only issue I have is that the description string is now duplicated twice right? Once in the struct and once in the parameter string?

Maybe let's pass the struct to the command so that `ostree_option_context_parse` can set it as the summary? Similar to what we do in rpm-ostree.

src/ostree/ot-builtin-static-delta.c
```diff
@@ -50,12 +50,17 @@ BUILTINPROTO(apply_offline);
...
 
 static OstreeCommand static_delta_subcommands[] = {
-  { "list", ot_static_delta_builtin_list },
-  { "show", ot_static_delta_builtin_show },
-  { "delete", ot_static_delta_builtin_delete },
-  { "generate", ot_static_delta_builtin_generate },
-  { "apply-offline", ot_static_delta_builtin_apply_offline },
-  { NULL, NULL }
+  { "list", ot_static_delta_builtin_list,
+    "list static delta files" },
```
jlebon: s/list/List/

src/ostree/ot-remote-builtin-delete-cookie.c
peterbaouoft: yea, we could do like https://github.com/projectatomic/rpm-ostree/commit/12c34bb2491a07079c911ef26401fee939e5573c.

But it kinda becomes tricky when we have multiple type of commands included in Ostree. One for `OstreeAdmin`, `OstreeAdminInstutil`, `OstreeRemote`, and things become a bit messier when we have two different parsing method. `ostree_option_context_parse` and `ostree_admin_option_context_parse`. So, going through a similar `invocation` line as rpm-ostree may involve many changes/refactoring to the existing code.

But, it may be not that difficult if we can merge/change all type of different commands in ostree into one whole `OstreeCommand`, and then passing the struct along won't be that hard.  But that also involves changing a lot of code :(, since I am a newbie here, I am inexperienced about whether or not it is worth to do so :p. But I think for now will just go ahead and have it a try :)

peterbaouoft:
So Rebased... and addressed most of comments ⬆️. 

>  description string is now duplicated twice right? Once in the struct and once in the parameter string

 This patch will be the prep for that. Another patch will hopefully come soon :). 

There are actually a lot of files being changed in order to be able to "pass the command struct" through builtins. Also, it is my first time doing this type of refactoring, so I might make mistakes and please tell me if I did it wrong. :) 

So, for the most recent patch, there are a few things worth mentioning(I think): 

1: The built in flags for admin related commands are  'OSTREE_BUILTIN_FLAG_NO_REPO', and the OstreeAdminBuiltinFlags is unchanged and remained the same. The `ostree_admin_option_context_parse` 
will have one more parameter `invocation` to represent the command struct. This was just me .. trying to avoid changing too much of the code :p. 

2: The default builtin flag in `ostree_option_context_parse` is OSTREE_BUILTIN_FLAG_NO_REPO, when no command struct is passed in. 

3: The other changes are all very similar to that from rpm-ostree https://github.com/projectatomic/rpm-ostree/commit/12c34bb2491a07079c911ef26401fee939e5573c

peterbaouoft:
> Is this still WIP? Looks pretty good to me.

Not any more :), I just updated the title, but unable to remove the label. And thanks for review, will address the comments shortly :).

peterbaouoft:
> Maybe Like Unix cat, but reads data from an OSTree repository ?
Hum...perhaps we shouldn't be rewriting the descriptions in this commit. Nevermind, let's do that after.

Sure :), added fixup based comments above ⬆️

cgwalters:
Nice work, thanks a lot!

@rh-atomic-bot r+ 6c26e62

peterbaouoft:
Thanks for the review :)

PR #1269, opened 2017-10-13
Pull gpg stage

cgwalters:


src/libostree/ostree-repo-pull.c
```diff
@@ -1241,13 +1241,28 @@ meta_fetch_on_complete (GObject           *object,
...
           GVariant *detached_data = g_hash_table_lookup (pull_data->fetched_detached_metadata, checksum);
+          OtChecksum hasher = { 0, };
+          ot_checksum_init (&hasher);
+          { g_autoptr(GBytes) bytes = g_variant_get_data_as_bytes (metadata);
+            ot_checksum_update_bytes (&hasher, bytes);
+          }
+          char hexdigest[OSTREE_SHA256_STRING_LEN+1];
+          ot_checksum_get_hexdigest (&hasher, hexdigest, sizeof (hexdigest));
+
+          if (strcmp (checksum, hexdigest) != 0)
```
jlebon: Minor: I like to use `g_str_equal` instead when possible. It's just easier to write and verify.

src/libostree/ostree-repo-pull.c
```diff
@@ -1241,13 +1241,28 @@ meta_fetch_on_complete (GObject           *object,
...
+            {
+              (void) glnx_throw (error, "Corrupted %s object %s (actual checksum is %s)",
+                                 ostree_object_type_to_string (objtype),
```
jlebon: We already know we're handling a commit object, right? Let's just simplify this?

src/libostree/ostree-repo-pull.c
```diff
@@ -1241,13 +1241,28 @@ meta_fetch_on_complete (GObject           *object,
...
-      /* For commit objects, check the GPG signature before writing to the repo,
-       * and also write the .commitpartial to say that we're still processing
-       * this commit.
+      /* For commit objects, compute the hash and check the GPG signature before
+       * writing to the repo, and also write the .commitpartial to say that
+       * we're still processing this commit.
        */
       if (objtype == OSTREE_OBJECT_TYPE_COMMIT)
         {
           GVariant *detached_data = g_hash_table_lookup (pull_data->fetched_detached_metadata, checksum);
```
jlebon: Can we add a `/* NULL if it wasn't found */` somewhere here to clarify that `detached_data` can be `NULL`?

src/libostree/ostree-repo-pull.c
cgwalters: I'll do a followup to unify all the checksum checking?

src/libostree/ostree-repo-pull.c
cgwalters: Fixup for the first bit below, I'll fix this with a separate followup.

jlebon:
@rh-atomic-bot r+ c26e519

PR #1270, opened 2017-10-13
tests: Add test-pull-bareuseronly

cgwalters:
I was going to fix a bug in the static deltas code and I noticed
we were missing `pull-test.sh` coverage for bareuseronly 🙈.

Obviously fixing this requires duplicating some of the bits we have in
`basic-test.sh`; need to hoist that into `libtest.sh`. For now though let's get
the coverage.

### DIFF
--- a/tests/pull-test.sh
+++ b/tests/pull-test.sh
@@ -27,9 +27,26 @@ function repo_init() {
     ${CMD_PREFIX} ostree --repo=repo remote add origin $(cat httpd-address)/ostree/gnomerepo "$@"
 }
 
+repo_init --no-gpg-verify
+
+# See also the copy of this in basic-test.sh
+COMMIT_ARGS=""
+CHECKOUT_U_ARG=""
+CHECKOUT_H_ARGS="-H"
+if is_bare_user_only_repo repo; then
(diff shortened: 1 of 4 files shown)


tests/pull-test.sh
```diff
@@ -114,13 +131,14 @@ ${CMD_PREFIX} ostree --repo=mirrorrepo prune --refs-only
 ${CMD_PREFIX} ostree --repo=mirrorrepo pull --bareuseronly-files origin main
 echo "ok pull (bareuseronly, safe)"
 
+if ! is_bare_user_only_repo repo; then
```
jlebon: Wait, why doesn't this work in a bare-user-only repo?

Actually, looking at all the commands in this test, none of them actually use the `repo` repo, right? I.e. it's basically independent of `repo_mode` it seems. Might be missing something though.

cgwalters: Fixup ⬇️

jlebon:
@rh-atomic-bot r+ 102539b

jlebon:
@rh-atomic-bot retry

cgwalters:
bot, retest this please

cgwalters:
@rh-atomic-bot retry

PR #1271, opened 2017-10-13
lib/utils: Check for invalid UTF-8 in filenames

mwleeds:
In case a filename contains invalid UTF-8 characters, libostree will
pass it to g_variant_builder_add() in create_tree_variant_from_hashes()
anyway, which leads to a critical warning from glib and an invalid
commit. This commit makes ostree print a useful error and exit instead.


mwleeds:
Based on the docs for `g_variant_builder_add_value` (which say "It is an error to call this function in any way that would create an inconsistent value to be constructed") I'm pretty sure this is the right thing to do.

cgwalters:
Thanks!  I pushed a fixup that adds a test.

@rh-atomic-bot r+ b9dba20

cgwalters:
@rh-atomic-bot retry

PR #1272, opened 2017-10-14
lib/commit: Implement "adoption" with CONSUME flag

cgwalters:
For checkouts that are on the same device, for regular files we can simply
"adopt" existing files. This is useful in the "build from subtrees" pattern that
happens with e.g. `rpm-ostree install` as well as flatpak and gnome-continuous.

New files are things like an updated `ldconfig` cache, etc. And particularly for
`rpm-ostree` we always regenerate the rpmdb, which for e.g. this workstation is
`61MB`.

We probably should have done this from the start, and instead had a `--copy`
flag to commit, but obviously we have to be backwards compatible.

There's more to do here - the biggest gap is probably for `bare-user` repos,
which are often used with things like `rpm-ostree compose tree` for host
systems. But we can do that later.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -2837,6 +2837,7 @@ write_directory_content_to_mtree_internal (OstreeRepo                  *self,
              */
             can_adopt = FALSE;
         }
+      gboolean did_adopt = FALSE;
 
       /* The very fast path - we have a devino cache hit, nothing to write */
       if (loose_checksum)
@@ -2854,6 +2855,7 @@ write_directory_content_to_mtree_internal (OstreeRepo                  *self,
             return FALSE;
           if (!ostree_mutable_tree_replace_file (mtree, name, checksum, error))
             return FALSE;
+          did_adopt = TRUE;
         }
       else
         {
@@ -2893,7 +2895,7 @@ write_directory_content_to_mtree_internal (OstreeRepo                  *self,
               return FALSE;
         }
 
-      if (delete_after_commit && !can_adopt)
+      if (delete_after_commit && !did_adopt)
         {
           if (!glnx_unlinkat (dfd_iter->fd, name, 0, error))
             return FALSE;
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -38,6 +38,18 @@
 #include "ostree-checksum-input-stream.h"
 #include "ostree-varint.h"
 
+/* In most cases, we write into a staging dir for commit, but we also allow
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -2722,23 +2878,24 @@ write_directory_content_to_mtree_internal (OstreeRepo                  *self,
...
-            return FALSE;
+            char tmp_checksum[OSTREE_SHA256_STRING_LEN+1];
+            ostree_checksum_inplace_from_bytes (child_file_csum, tmp_checksum);
+            if (!ostree_mutable_tree_replace_file (mtree, name, tmp_checksum,
+                                                   error))
+              return FALSE;
         }
 
-      if (delete_after_commit)
+      if (delete_after_commit && !did_adopt)
```
jlebon: Maybe a small comment here to clarify that `adopt_and_commit_regfile()` did a `renameat`?

tests/basic-test.sh
```diff
@@ -181,12 +181,35 @@ cd ${test_tmpdir}
...
 assert_not_has_dir $test_tmpdir/checkout-test2-l
+$OSTREE fsck
 # Some of the later tests are sensitive to state
 $OSTREE reset test2 test2^
+$OSTREE prune --refs-only
 echo "ok consume (nom nom nom)"
 
+# Test adopt
+cd ${test_tmpdir}
+if grep -q '^mode=bare' repo/mode || is_bare_user_only_repo repo; then
```
jlebon: Did you mean `repo/config`?

This will match `bare-user` too, right? Shouldn't this be `grep -q '^mode=bare$' repo/config`? But really, it seems like we should just always run the test and only do the hardlink check in the non-`bare-user` case?

tests/basic-test.sh
```diff
@@ -181,12 +181,35 @@ cd ${test_tmpdir}
...
+$OSTREE fsck
+eatme_objpath=$(ostree_file_path_to_object_path repo test2 /eatme.txt)
+assert_files_hardlinked ${test_tmpdir}/eatme-savedlink.txt ${eatme_objpath}
+assert_not_has_dir $test_tmpdir/checkout-test2-l
+# Some of the later tests are sensitive to state
+$OSTREE reset test2 test2^
+$OSTREE prune --refs-only
+rm -f ${test_tmpdir}/eatme-savedlink.txt
+fi
+echo "ok adopt"
```
jlebon: Let's explicitly do a skip here in the `else` branch?

tests/basic-test.sh
cgwalters: Good idea 💡!  Done ⬇️

jlebon:
@rh-atomic-bot r+ 0b83759

PR #1274, opened 2017-10-16
Make sure *.am.inc are up to date before `make dist`

smcv:
v2017.12 didn't include test-libglnx-shutil.c, but if you re-run
autogen.sh (as we do in Debian, to update the Autotools build system)
it will try to build it.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -90,13 +90,18 @@ endif # end ENABLE_RUST
 libglnx_srcpath := $(srcdir)/libglnx
 libglnx_cflags := $(OT_DEP_GIO_UNIX_CFLAGS) "-I$(libglnx_srcpath)"
 libglnx_libs := $(OT_DEP_GIO_UNIX_LIBS)
+$(srcdir)/libglnx/Makefile-libglnx.am.inc: $(srcdir)/libglnx/Makefile-libglnx.am
+	sed -e 's,$(libglnx_srcpath),libglnx,g' < $< > $@
 include libglnx/Makefile-libglnx.am.inc
 EXTRA_DIST += libglnx/Makefile-libglnx.am
 noinst_LTLIBRARIES += libglnx.la
 
 libbsdiff_srcpath := $(srcdir)/bsdiff
+
 libbsdiff_cflags := $(OT_DEP_GIO_UNIX_CFLAGS) "-I$(bsdiff_srcpath)"
 libbsdiff_libs := $(OT_DEP_GIO_UNIX_LIBS)
+$(srcdir)/bsdiff/Makefile-bsdiff.am.inc: $(srcdir)/bsdiff/Makefile-bsdiff.am.inc
+	sed -e 's,$(libbsdiff_srcpath),bsdiff,g' < $< > $@
 include bsdiff/Makefile-bsdiff.am.inc
 EXTRA_DIST += bsdiff/Makefile-bsdiff.am
 noinst_LTLIBRARIES += libbsdiff.la
--- a/Makefile.am
+++ b/Makefile.am
@@ -90,6 +90,11 @@ endif # end ENABLE_RUST
 libglnx_srcpath := $(srcdir)/libglnx
 libglnx_cflags := $(OT_DEP_GIO_UNIX_CFLAGS) "-I$(libglnx_srcpath)"
 libglnx_libs := $(OT_DEP_GIO_UNIX_LIBS)
+# See also autogen.sh and https://github.com/ostreedev/ostree/pull/1274/
+#
+# v2017.12 didn't include test-libglnx-shutil.c, but if you re-run
+# autogen.sh (as we do in Debian, to update the Autotools build system)
+# it will try to build it.
 $(srcdir)/libglnx/Makefile-libglnx.am.inc: $(srcdir)/libglnx/Makefile-libglnx.am
 	sed -e 's,$(libglnx_srcpath),libglnx,g' < $< > $@
 include libglnx/Makefile-libglnx.am.inc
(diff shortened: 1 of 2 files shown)


Makefile.am
```diff
@@ -90,13 +90,18 @@ endif # end ENABLE_RUST
 libglnx_srcpath := $(srcdir)/libglnx
 libglnx_cflags := $(OT_DEP_GIO_UNIX_CFLAGS) "-I$(libglnx_srcpath)"
 libglnx_libs := $(OT_DEP_GIO_UNIX_LIBS)
+$(srcdir)/libglnx/Makefile-libglnx.am.inc: $(srcdir)/libglnx/Makefile-libglnx.am
```
cgwalters: Ah...wait, so you're only fixing a bug where you're using `./autogen.sh && make dist` and then building the resulting tarball?   This can't work for an `autogen.sh && make` build as we'll do the `include` *before* handling the rule, correct?

cgwalters: (Also I pushed a fixup to add some comments)

smcv:
Sorry, the PR message should have been clearer/more specific. I'm moving this up to being a top-level comment so that it won't go away when the commit is rebased out of existence to be replaced by one with better comments.

2017.12 was released with a missing file, presumably because of this workflow:

* `./autogen.sh`
* some development (including updating libglnx to add test-libglnx-shutil.c, but no re-runs of `./autogen.sh`)
* `make dist`
* release

The resulting tarball has an outdated `Makefile*.am.inc` that doesn't reference `test-libglnx-shutil.c`, with the result that `make dist` doesn't ship that file.

We re-run `autogen.sh` during the Debian build, because we want to re-generate as much as possible from its ultimate source code so we are not relying on the version that you built with a non-Debian toolchain - partly to prove that we can (GPL compliance), and partly so we can fix bugs in the generated files or in the tools that generate them if we have to, without editing generated files which is always fairly horrible. `autogen.sh` re-generates the `Makefile*.am.inc`. Now it correctly references `test-libglnx-shutil.c`, but we don't have that file (it wasn't in the tarball) so the build fails.

The dependencies added in this PR fix that: `dist` indirectly depends on `$(am__DIST_COMMON)` which includes all the `include`'d files. This means your `make dist` would have re-generated the outdated `Makefile*.am.inc` to be up to date, so your tarball would have included `test-libglnx-shutil.c`, and everything would work as expected.

As you've observed, it is still necessary to duplicate the `sed` bits in `autogen.sh` to bootstrap the project, because you can't run `make` until the edited `Makefile*.am.inc` are in place.

I'm not sure how best to distil that down into a comment in the code...

cgwalters:
> because we want to re-generate as much as possible from its ultimate source code

i.e. you should be using git directly 😄  Maybe get an exception somehow to the "prefer manual tarballs" policy?

Anyways...I'll probably try to wedge some sort of at least dist+build into the CI and use that.

@rh-atomic-bot r+ bb7d8ce

smcv:
> > because we want to re-generate as much as possible from its ultimate source code
>
> i.e. you should be using git directly 😄 Maybe get an exception somehow to the "prefer manual tarballs" policy?

There are two semi-conflicting policies going on here: if the upstream releases a canonical artefact containing source code (which you do), we are meant to use that one as our "original" tarball, with an accompanying signature if there is one; and if something is a derived file, we try to rebuild it at build-time. Autotools build systems have always been somewhat in conflict with at least one of those.

The Debian source package format is older than git winning the DVCS wars (and indeed older than DVCSs, or widespread use of VCSs at all), which is why it's based on the lowest-common-denominator format that every VCS and every project can hopefully emit somehow: a tarball with source code from upstream, and a series of patches to apply any local modifications that were necessary. The patch stack is separate because many licenses say we have to track the dates and/or content of our modifications, and because if we forget to send changes upstream, it's considered rude to have them in a format that an interested upstream couldn't pick out.

I can use whatever "original" tarball I can reasonably justify as being what you released, but when you're releasing `make dist` tarballs, I can't really justify anything else as being "the release". If you'd prefer to release some sort of recursive variant of `git archive` instead (like the ones you used to produce from a script in `packaging/`), I could use those. For upstreams that don't release non-git artefacts at all, I'd normally use plain `git archive`, or GitHub's automatic export based on `git archive`; but I can't do that here because ostree has submodules.

smcv:
> I can't really justify anything else as being "the release"

... unless you tell me that some specific thing should be considered to be the source release, in which case if anyone in Debian complains, I can point them to that.

Makefile.am
```diff
@@ -90,13 +90,24 @@ endif # end ENABLE_RUST
...
 include libglnx/Makefile-libglnx.am.inc
 EXTRA_DIST += libglnx/Makefile-libglnx.am
 noinst_LTLIBRARIES += libglnx.la
 
 libbsdiff_srcpath := $(srcdir)/bsdiff
+
 libbsdiff_cflags := $(OT_DEP_GIO_UNIX_CFLAGS) "-I$(bsdiff_srcpath)"
 libbsdiff_libs := $(OT_DEP_GIO_UNIX_LIBS)
+# See the comment for the similar libglnx bit above
+$(srcdir)/bsdiff/Makefile-bsdiff.am.inc: $(srcdir)/bsdiff/Makefile-bsdiff.am.inc
```
dbnicholson: This is causing `make` to issue a circular dependency warning:
```
make: Circular bsdiff/Makefile-bsdiff.am.inc <- bsdiff/Makefile-bsdiff.am.inc dependency dropped.
```
Presumably the prereq should be `$(srcdir)/bsdiff/Makefile-bsdiff.am` without the `.inc`.

dbnicholson: Filed #1276 with that change.

smcv: Your fix is clearly correct. Sorry about that...

cgwalters:
@smcv This closed PR isn't really the right place to discuss this but if you have any pull to convince Debian releng:

>  For upstreams that don't release non-git artefacts at all, I'd normally use plain git archive, or GitHub's automatic export based on git archive; but I can't do that here because ostree has submodules.

Anyone can create recursive archives; it's really not hard.  gnome-continuous handles submodules, so does [rpmdistro-gitoverlay](https://github.com/cgwalters/rpmdistro-gitoverlay/) for example.

Further, ostree tags are signed using [git-evtag](http://github.com/cgwalters/git-evtag) which is *specifically* designed to solve this problem - its checksum includes submodules, and is designed to be reproducible.

So again from my POV git *is* the thing, even with submodules.

PR #1275, opened 2017-10-16
Cope with xattr syscalls raising EOPNOTSUPP

smcv:
ENOTSUP and EOPNOTSUPP are numerically equal on most Linux ports,
but inexplicably differ on PA-RISC (hppa) and possibly other
rare architectures.


cgwalters:
Hm, maybe libglnx should have a `GLNX_ERRNO_IS_OPNOTSUP()` or something.  But anyways this looks good for now, thanks a lot!

@rh-atomic-bot r+ 4a5871d

PR #1276, opened 2017-10-16
build: Fix bsdiff Makefile circular dependency

dbnicholson:
The intended use was to have the .am.inc generated from the .am like the
libglnx one. Without this, make was detecting a circular dependency and
dropping the rule:

  make: Circular bsdiff/Makefile-bsdiff.am.inc <- bsdiff/Makefile-bsdiff.am.inc dependency dropped.


jlebon:
Thanks!
@rh-atomic-bot r+ d86a8a3

PR #1277, opened 2017-10-16
lib: Use a common helper function to compare checksums

cgwalters:
So we get a consistent error message; came up in a PR review.


jlebon:
@rh-atomic-bot r+ 8455e20

Issue #1273, opened 2017-10-15
unclear error when running ostree-admin-unlock without overlayfs


guyshapiro:
Tried to run `ostree admin unlock` on machine without overlayfs support (before I know it shouldn't work).
The result was the following message:
```
(ostree admin unlock:811): GLib-WARNING **: GError set over the top of a previous GError or uninitialized memory.
This indicates a bug in someone's code. You must ensure an error is NULL before it's set.
The overwriting error message was: No such file or directory
error: overlayfs mount helper: Child process exited with code 1
```
If I understand correctly, this message indicates a bug in the error handling code.

(Tested on commit fc33ae018dccbcab633b69474db0811987355212)

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1279 and you have `r+` rights.

PR #1279, opened 2017-10-16
lib/sysroot: Fix error handling when mounting overlayfs fails

cgwalters:
This isn't perfect, but at least we fix an error-overwrite error, and in
practice `ostree admin unlock` isn't wrapped by `rpm-ostree` yet, so spew to
stderr is OK.

Closes: https://github.com/ostreedev/ostree/issues/1273

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -22,6 +22,7 @@
 #include "otutil.h"
 #include <sys/file.h>
 #include <sys/mount.h>
+#include <err.h>
 #include <sys/wait.h>
 
 #include "ostree.h"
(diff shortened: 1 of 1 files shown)


cgwalters:
@rh-atomic-bot delegate=guyshapiro

cgwalters:
```
[root@localhost ~]# strace -o strace.log -f -e inject=mount:error=EINVAL ostree admin unlock
ostree: mount: Invalid argument
error: Failed overlayfs mount: Child process exited with code 1
```

guyshapiro:
The fix looks good. However, I won't be able to test it on the next few days, and I don't have experience with the GLib error handling  framework.
So, FWIW, @rh-atomic-bot r+ 159e44473449df802ef793bc1ad242b5645dd176

PR #1281, opened 2017-10-17
lib/repo-finder: Add OstreeRepoFinderOverride

pwithnall:
This is another OstreeRepoFinder implementation; it returns results from
a given set of URIs. It’s designed to be used for implementing user
overrides to other repo-finders, or for implementing unit tests.

### DIFF
--- a/src/libostree/ostree-repo-finder-override.c
+++ b/src/libostree/ostree-repo-finder-override.c
@@ -1,0 +1,314 @@
+/*
+ * Copyright © 2017 Endless Mobile, Inc.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ * Authors:
+ *  - Philip Withnall <withnall@endlessm.com>
+ */
+
+#include "config.h"
+
+#include <fcntl.h>
+#include <gio/gio.h>
+#include <glib.h>
+#include <glib-object.h>
+#include <libglnx.h>
+
+#include "ostree-autocleanups.h"
+#include "ostree-remote-private.h"
+#include "ostree-repo.h"
+#include "ostree-repo-private.h"
+#include "ostree-repo-finder.h"
+#include "ostree-repo-finder-override.h"
+
+/**
+ * SECTION:ostree-repo-finder-override
+ * @title: OstreeRepoFinderOverride
+ * @short_description: Finds remote repositories from a list of repository URIs
+ * @stability: Unstable
+ * @include: libostree/ostree-repo-finder-override.h
+ *
+ * #OstreeRepoFinderOverride is an implementation of #OstreeRepoFinder which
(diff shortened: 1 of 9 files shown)


src/libostree/ostree-repo-finder-override.c
```diff
@@ -0,0 +1,314 @@
...
+          g_clear_error (&local_error);
+          continue;
+        }
+
+      for (j = 0; refs[j] != NULL; j++)
+        {
+          g_autoptr(OstreeRemote) keyring_remote = NULL;
+
+          /* Look up the GPG keyring for this ref. */
+          keyring_remote = ostree_repo_resolve_keyring_for_collection (parent_repo,
```
mwleeds: If I understand correctly, this only succeeds if an existing remote is configured with the same collection ID and that remote has a GPG keyring. Maybe we should document that in the description of `OstreeRepoFinderOverride`?

src/libostree/ostree-repo-finder-override.c
```diff
@@ -0,0 +1,314 @@
...
+/* This must return a valid remote name (suitable for use in a refspec). */
+static gchar *
+uri_and_keyring_to_name (const gchar *uri,
+                         const gchar *keyring)
+{
+  g_autofree gchar *escaped_uri = g_uri_escape_string (uri, NULL, FALSE);
+  g_autofree gchar *escaped_keyring = g_uri_escape_string (keyring, NULL, FALSE);
+
+  /* FIXME: Need a better separator than `_`, since it’s not escaped in the input. */
+  g_autofree gchar *out = g_strdup_printf ("%s_%s", escaped_uri, escaped_keyring);
```
mwleeds: This is just for uniqueness right? It doesn't seem to get parsed anywhere.

src/libostree/ostree-repo-finder-override.c
```diff
@@ -0,0 +1,314 @@
...
+                                           const OstreeCollectionRef * const *refs,
+                                           OstreeRepo                        *parent_repo,
+                                           GCancellable                      *cancellable,
+                                           GAsyncReadyCallback                callback,
+                                           gpointer                           user_data)
+{
+  OstreeRepoFinderOverride *self = OSTREE_REPO_FINDER_OVERRIDE (finder);
+  g_autoptr(GTask) task = NULL;
+  g_autoptr(GPtrArray) results = NULL;
+  const gint priority = 100;  /* arbitrarily chosen; lower than the others */
```
mwleeds: I think it's a bit odd that the class is called "Override" but its results don't override the others. I understand that the use case is probably just passing `OstreeRepoFinderOverride` and none of the others to the finder, but maybe we should at least document the effect of the priority choice?

src/libostree/ostree-repo-finder-override.c
pwithnall: Sort of. It ends up as the name of the `OstreeRemote` in an `OstreeRepoFinderResult` returned by the finder, so could theoretically be parsed by the caller (although I really wouldn’t expect that). It’s primarily for uniqueness, yes.

src/libostree/ostree-repo-finder-override.c
pwithnall: Good point.

pwithnall: There is also the option to expose `priority` as a property of the `OstreeRepoFinderOverride` at some point in the future, and make it mutable.

pwithnall: It also appears I made a thinko here. I intended the results from `OstreeRepoFinderOverride` to be higher priority than others, but then never actually made that change.

src/libostree/ostree-repo-finder-override.c
pwithnall: True. I’ll add it to the docs.

pwithnall:
Fixups pushed containing fixes for your review comments, thanks.

I’m not opposed to renaming the class to something not involving the word ‘override’. For example, `OstreeRepoFinderStatic` is a possibility.

cgwalters:
@rh-atomic-bot r=mwleeds b5fad57

jlebon:
@rh-atomic-bot retry

PR #1283, opened 2017-10-17
lib/commit: (refactor) Clean up delta bare write API

cgwalters:
The way `_ostree_repo_open_content_bare()` did both looking for the object and
possibly creating a new fd was just weird and inconsistent with e.g. the pull
code where we always call `has_object()` first.

Just call `has_object()` in the delta paths that used this too, making the
implementation right now a thin wrapper around
`glnx_open_tmpfile_linkable_at()`, but this is prep for a later patch which does
more.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -412,23 +412,9 @@ _ostree_repo_open_content_bare (OstreeRepo          *self,
                                 const char          *checksum,
                                 guint64              content_len,
                                 GLnxTmpfile         *out_tmpf,
-                                gboolean            *out_have_object,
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -412,23 +412,9 @@ _ostree_repo_open_content_bare (OstreeRepo          *self,
                                 const char          *checksum,
```
jlebon: Can we update the preceding comment block too?

cgwalters: Yep done ⬇️

jlebon:
@rh-atomic-bot r+ 3fbd2fd

Issue #1285, opened 2017-10-18
Be robust against unexpected files in refs/, e.g. ".latest_rsync"


parasense:
I observed the error when updating the summary file:

```
$ sudo ostree summary -u --repo=/mnt/redhat/nightly/Atomic/ostree/repo/
[sudo] password for jdisnard: 
error: No such metadata object .commit
```

The ostree is an rsync copy of my production repo, but contains a lot of little files named .latest_rsync , which is an artefact of how my replication tool deploys the ostree repo to a CDN. I guess some scheme to avoid re-uploading contents. Anyways, after running the above cmd under strace,  I observed ostree attempting to open those files.... 

```
fcntl(12, F_GETFL)                      = 0x38800 (flags O_RDONLY|O_NONBLOCK|O_LARGEFILE|O_DIRECTORY|O_NOFOLLOW)
fcntl(12, F_SETFD, FD_CLOEXEC)          = 0
getdents(12, /* 4 entries */, 32768)    = 120
openat(8, "rhel-atomic-host/7/x86_64/.latest_rsync", O_RDONLY|O_NOCTTY|O_CLOEXEC) = 13
fstat(13, {st_mode=S_IFREG|0644, st_size=0, ...}) = 0
read(13, "", 16)                        = 0
close(13)                               = 0
openat(8, "rhel-atomic-host/7/x86_64/standard", O_RDONLY|O_NOCTTY|O_CLOEXEC) = 13
fstat(13, {st_mode=S_IFREG|0644, st_size=65, ...}) = 0
```

so on a whim I removed all the .latest_rsync files from my copy of the repo, and then the summary file update worked afterwards. Clearly my repo was corrupted by these tiny files, and I know the fix is to simply remove them all, but I figured ostree might be made more resilient against such things.

cgwalters:
Sure, easy enough.  PR in https://github.com/ostreedev/ostree/pull/1286

parasense:
Thanks, and meanwhile I'm filing a bug against my deployment tool, to ask they stop using dot files and perhaps use extended attributes or some other scheme to track uploads. Appreciate the quick turn around, thanks again.

PR #1286, opened 2017-10-18
Disallow refs starting with a non-letter or digit

cgwalters:
Change the regexp for validating refs to require at least one letter or digit
before allowing the other special chars in the set `[.-_]`. Names that start
with `.` are traditionally Unix hidden files; let's ignore them under the
assumption they're metadata for some other tool, and we don't want to
potentially conflict with the special `.` and `..` Unix directory entries.
Further, names starting with `-` are problematic for Unix cmdline option
processing; there's no good reason to support that. Finally, disallow `_` just
on general principle - it's simpler to say that ref identifiers must start with
a letter or digit.

We also ignore any existing files (that might be previously created refs) that
start with `.` in the `refs/` directory - there's a Red Hat tool for content
management that injects `.rsync` files, which is why this patch was first
written.

V1: Update to ban all refs starting with a non-letter/digit, and
    also add another call to `ostree_validate_rev` in the pull
    code.

Closes: https://github.com/ostreedev/ostree/issues/1285

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3853,6 +3853,8 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
 
       while (g_variant_iter_loop (collection_refs_iter, "(&s&s&s)", &collection_id, &ref_name, &checksum))
         {
+          if (!ostree_validate_rev (ref_name, error))
+            goto out;
           g_hash_table_insert (requested_refs_to_fetch,
                                ostree_collection_ref_new (collection_id, ref_name),
                                (*checksum != '\0') ? g_strdup (checksum) : NULL);
--- a/tests/test-refs.sh
+++ b/tests/test-refs.sh
@@ -23,7 +23,7 @@ set -euo pipefail
 
 setup_fake_remote_repo1 "archive"
 
-echo '1..2'
+echo '1..5'
 
 cd ${test_tmpdir}
 mkdir repo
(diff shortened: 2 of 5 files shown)


tests/test-refs.sh
```diff
@@ -88,6 +88,19 @@ if ${CMD_PREFIX} ostree --repo=repo refs foo/ctest --create=ctest; then
     assert_not_reached "refs --create unexpectedly succeeded in overwriting an existing prefix!"
 fi
 
+# https://github.com/ostreedev/ostree/issues/1285
+# One tool was creating .latest_rsync files in each dir, let's ignore stuff like
+# that.
+echo '👻' > repo/refs/heads/.spooky_and_hidden
+ostree --repo=repo refs > refs.txt
```
jlebon: Missing `${CMD_PREFIX}` here and below!

cgwalters:
We already do validate creating refs AFAICS, this is just another tool dropping files in there.

tests/test-refs.sh
cgwalters: Done ⬇️

jlebon:
What I mean is:

```
$ ostree --repo=. init --mode=bare-user
$ mkdir tmp2
$ echo "asdf" > tmp2/asdf
$ ostree commit -b .ref_that_starts_with_dot tmp2
f3c0a7d868d3293f922c3ed0d29b180831e17a8f331d659ed910a8593907640d
$ file refs/heads/.ref_that_starts_with_dot
refs/heads/.ref_that_starts_with_dot: ASCII text
```

We should have errored out there, right?

cgwalters:
Ah.  Yes...well that definitely increases the scope of this, but I think you're right.  Reworked in ⬆️

cgwalters:
Hum, we currently accept anything that GRegex allows as `\w\d` but...that's a whole lot of stuff in Unicode.  I wonder if we shouldn't also restrict to ASCII.  But eh...not right now?

src/libostree/ostree-repo-pull.c
```diff
@@ -3853,6 +3853,8 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
 
       while (g_variant_iter_loop (collection_refs_iter, "(&s&s&s)", &collection_id, &ref_name, &checksum))
         {
+          if (!ostree_validate_rev (ref_name, error))
```
jlebon: We also need this in the next `else if` for `refs_to_fetch`, right?

cgwalters: There's already one there, no?  Line 3879?

jlebon: Whoops! Indeed.

jlebon:
@rh-atomic-bot r+ 7266b15

alexlarsson:
This is breaking stuff in flatpak due to the use of "${remote}:." as argument to ostree_repo_list_refs to find all local refs pulled from a particular remote.

alexlarsson:
See https://github.com/flatpak/flatpak/issues/1444

jlebon:
Hmm, we should just be able to support `${remote}:` for that, right? Seems cleaner than `:.`.

jlebon:
#1500

PR #1287, opened 2017-10-18
More checksum cleanups

cgwalters:


jlebon:
@rh-atomic-bot r+ 1089da7

PR #1291, opened 2017-10-18
build: Do *.am.inc substitutions from make correctly

dbnicholson:
The string needs to be escaped with a $ since we want to replace the
literal string $(path). Without this make will run the command with the
value of the variable, which won't match anything in the input Makefile
stub.

CC @smcv


smcv:
Looks good to me.

cgwalters:
@rh-atomic-bot r=smcv db52d73

PR #1292, opened 2017-10-18
Repository locking

dbnicholson:
In #813, some simple repository locking was being considered to prevent prunes from running concurrently with checkouts and transactions. While I think that would be a great thing, I don't think that level of locking is rich enough for ostree. Since the locking isn't recursive, it would prevent both libostree and ostree programs from obtaining an exclusive lock if it would also happen somewhere else in libostree. The simple example I can think of is `ostree_repo_prune_static_deltas()`. You'd want to take an exclusive lock there as an entry point to libostree, but it's also called from `ostree_repo_prune()`, which you'd also want to take an exclusive lock for. That would simply deadlock.

The alternative I came up with here is to maintain the lock in thread local storage with a stack of lock states. Then each thread locks independently and the stack stays coherent. ddfa34b125365b4a0dd293c95c68e4fb2bc1c2e7 has that implementation with a large comment in `ostree-repo.c` describing how it works. The API is then a push/pop interface.

The drawback of having the lock in TLS is that it seems to prevent taking the lock asynchronously since that would likely involve having another thread acquire the lock. Once that thread completed, the lock would be dropped. So, the implementation here blocks synchronously when pushing or popping the lock. The lock timeout can be managed globally with a repo config option or from an API.

All the new APIs are experimental, but I think it would be best to make them public so ostree programs can do their own locking around a series of ostree operations.

I added a bunch of locking in places I thought it would be needed. One spot I didn't try to tackle was the core and remote config. Thinking about all the use cases got me confused, so I punted for now...

This closes https://bugzilla.gnome.org/show_bug.cgi?id=759442.

Let me know what you think! I know there's a lot here, but this issue has affected us quite a bit at Endless.


cgwalters:
High level looks sane at a brief glance, but there's lots of details here.  (One I have is how this intersects with worker threads we create internally)

How about we avoid lots of `#ifdefs` here by having a private version that's used internally, then the "experimental" bit is just exposing it publicly?   The private version could actually be a repo setting too so it doesn't depend on the experimental-API build?

dbnicholson:
> High level looks sane at a brief glance, but there's lots of details here. (One I have is how this intersects with worker threads we create internally)

Yeah, I tried to think about that a lot, but I couldn't totally come to grips with it. Basically, the worker threads are OK if you only ever acquire a shared lock. Which (I believe) is the case now. Otherwise, you have to arrange to take the lock before the worker is spawned and ensure that the worker doesn't go through a path that acquires locks.

I spent a lot of time thinking about that how to address that. I had another version that tried to detect if the lock was being acquired in the same thread by stashing the main context in use when the repo was opened and skipping if the main context in use when the lock was acquired wasn't the same. But that only tells you if the context is the same, not if the thread is the same. (I might be missing something there. I am definitely not the `GMainContext` guru.)

The only other thing I could think of was having a dedicated locking thread, but I then didn't know how to address the lock stack being updated out of order by multiple threads.

> How about we avoid lots of #ifdefs here by having a private version that's used internally, then the "experimental" bit is just exposing it publicly?

You mean in `src/ostree`? I could add a `cmdprivate` pointer if that's better. Most of `src/libostree` is clean except for the headers.

> The private version could actually be a repo setting too so it doesn't depend on the experimental-API build?

Not sure I follow here. What would the setting be?

cgwalters:
Yeah, I mean the binaries would use a cmdprivate API to do locking, which we could then enable by setting e.g.
```
[core]
locking=true
```
or so in the repo config.  (Which itself we could label as experimental I guess).   That'd allow us to land most of these patches and gain experience with it a bit, even on builds that don't `--enable-experimental-api`.

dbnicholson:
Oh, well I only added actual locking in `fsck`. The rest of the stuff in `src/ostree` is just turning the `--lock-timeout` option into a call to `ostree_repo_set_lock_timeout()`, which could be `cmdprivate` like you say.

However, the repo config option could be used to make `ostree_repo_lock_push()` and `ostree_repo_lock_pop()` into noops. That's a good idea in case this is broken...

I also see that the CI failed immediately because I haven't actually tested without `--enable-experimental-api` in a while. Whoops.

man/ostree.repo-config.xml
```diff
@@ -143,6 +143,18 @@ Boston, MA 02111-1307, USA.
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>lock-timeout</varname></term>
+        <listitem>
+          <para>
+            Integer value controlling the default repository locking
+            timeout in seconds. If the value is <literal>-1</literal>,
```
pwithnall: What if the value is 0 or <-1? Could go with “If the value is negative then ostree will block…”

man/ostree.xml
```diff
@@ -120,6 +120,23 @@ Boston, MA 02111-1307, USA.
...
+                 non-experimental and change the \-\- back to normal.
+            <varlistentry>
+              <term><option>\-\-lock-timeout</option>=TIMEOUT</term>
+
+              <listitem>
+                <para>
+                  Some <command>ostree</command> commands require the
+                  repository to be locked while operating. This option sets
+                  the repository lock timeout to TIMEOUT seconds. If TIMEOUT
+                  is <literal>-1</literal>, then the locking will block
```
pwithnall: Similarly here: what if it’s <-1 or 0?

src/libostree/ostree-repo-private.h
```diff
@@ -413,8 +418,36 @@ _ostree_repo_get_remote_inherited (OstreeRepo  *self,
...
+                                     gboolean       exclusive,
+                                     GCancellable  *cancellable,
+                                     GError       **error);
+gboolean      ostree_repo_lock_pop (OstreeRepo    *self,
+                                    GCancellable  *cancellable,
+                                    GError       **error);
+
+typedef OstreeRepo OstreeRepoAutoLock;
+OstreeRepoAutoLock * ostree_repo_auto_lock_push (OstreeRepo    *self,
+                                                 gboolean       exclusive,
```
pwithnall: Might make sense to use an enum or flags instead of a boolean here, since it’s typically not obvious what the sense of a boolean is. Does `TRUE` mean ‘make it exclusive’ or ‘make it shared’?

src/libostree/ostree-repo-private.h
```diff
@@ -37,6 +37,9 @@ G_BEGIN_DECLS
 #define _OSTREE_MAX_OUTSTANDING_FETCHER_REQUESTS 8
 #define _OSTREE_MAX_OUTSTANDING_DELTAPART_REQUESTS 2
 
+/* Keep this in sync with the man page for the core.lock-timeout option. */
```
pwithnall: Mention that it’s in seconds.

src/libostree/ostree-repo-private.h
```diff
@@ -153,6 +157,7 @@ struct OstreeRepo {
   guint64 tmp_expiry_seconds;
   gchar *collection_id;
   gboolean add_remotes_config_dir; /* Add new remotes in remotes.d dir */
+  gint lock_timeout;
```
pwithnall: Maybe name it `lock_timeout_seconds` to make the units more obvious? Same for the `#define` above.

src/libostree/ostree-repo-prune.c
```diff
@@ -401,6 +417,11 @@ ostree_repo_prune_from_reachable (OstreeRepo        *self,
                                   GCancellable      *cancellable,
                                   GError           **error)
 {
+  g_autoptr(OstreeRepoAutoLock) lock = NULL;
+  lock = ostree_repo_auto_lock_push (self, TRUE, cancellable, error);
+  if (!lock)
+    return FALSE;
```
pwithnall: One thought: if you changed the definition of `ostree_repo_auto_lock_push()` to:
`gboolean ostree_repo_auto_lock_push (OstreeRepo *self, OstreeRepoLockFlags flags, OstreeRepoAutoLock **out_lock, GCancellable *cancellable, GError **error)`, then all the uses of it could simplify to:
```
g_autoptr(OstreeRepoAutoLock) lock = NULL;
if (!ostree_repo_auto_lock_push (self, OSTREE_REPO_LOCK_EXCLUSIVE, &lock, cancellable, error))
  return FALSE;
```

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,508 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+              cur_state_name = (cur_state == LOCK_EX) ? "exclusive" : "shared";
+            }
+        }
+      else
+        {
+          stack_len = 0;
+          cur_state_name = "null";
+        }
+
+      g_debug("Free lock: state=%s, depth=%u", cur_state_name, stack_len);
```
pwithnall: Nitpick: Missing space before `(` (and in a couple of other places).

src/libostree/ostree-repo.c
```diff
@@ -5099,3 +5693,47 @@ ostree_repo_set_collection_id (OstreeRepo   *self,
...
+ * Set the lock timeout for the repository. A timeout of -1 indicates that the
+ * application will block until the lock is acquired. Otherwise, it represents
+ * the number of seconds the application will sleep attempting to acquire the
+ * repository lock.
+ *
+ * Note that if the repository configuration is reloaded, then a timeout
+ * configured in the core.lock-timeout option will replace this.
+ *
+ * Returns: %TRUE on success, %FALSE otherwise
+ * Since: 2017.8
```
pwithnall: Since: 2017.13

(probably wrong in a few other places)

src/libostree/ostree-repo.h
```diff
@@ -108,6 +108,40 @@ OstreeRepo *  ostree_repo_create_at (int             dfd,
...
+                                     GError       **error);
+_OSTREE_PUBLIC
+gboolean      ostree_repo_lock_pop (OstreeRepo    *self,
+                                    GCancellable  *cancellable,
+                                    GError       **error);
+
+/**
+ * OstreeRepoAutoLock: (skip)
+ *
+ * This is simply an alias tox #OstreeRepo used for automatic lock cleanup.
```
pwithnall: s/tox/to/

man/ostree.repo-config.xml
dbnicholson: 0 means don't wait - true non-blocking. I thought that was implied, but I could spell it out. < -1 is an error. I could just do all negative values, but I liked having the explicit value. It also allowed me to later use -2 as a sentinel.

man/ostree.xml
dbnicholson: OK.

src/libostree/ostree-repo-private.h
dbnicholson: Sure.

src/libostree/ostree-repo-private.h
dbnicholson: Sure

src/libostree/ostree-repo-private.h
dbnicholson: Yeah, I thought about that later after doing `TRUE`/`FALSE` so many times. It seemed sort of silly to have an enum that would only ever have 2 exclusive states. I guess:
```
typedef enum {
  OSTREE_REPO_LOCK_SHARED,
  OSTREE_REPO_LOCK_EXCLUSIVE
} OstreeRepoLockType;
```

src/libostree/ostree-repo-prune.c
dbnicholson: I could go either way on this and did consider that interface. I do generally like ostree's boolean return values with supplemental information in out params. In this case it seemed overkill to me, though, as both the return value and the out param would be transmitting the same information. Would the out param be a requirement? it would be useless to use the function otherwise. Not that we're counting cycles in ostree, but it's also less efficient to put another parameter on the stack and fill it in.

You can, of course, do the call and check in one line if you want with the current interface:
```
if (!(lock = ostree_repo_auto_lock_push(...)))
  return FALSE;
```
That's definitely not as aesthetically pleasing as the interface you've suggested, which is also why I prefer to write it on 2 lines.

src/libostree/ostree-repo.c
dbnicholson: Good catch.

src/libostree/ostree-repo-private.h
pwithnall: I think people (me included) should get out of the habit of thinking of two-element enums as ‘silly’, if they improve code readability. That said, there’s the possibility that this could be extended in future to include more flags or enum elements for more esoteric extensions to the locking.

src/libostree/ostree-repo.c
dbnicholson: Bah. I actually think I was pretty good about setting it to 2017.13, but I'll skim through for that. Thanks for catching that.

src/libostree/ostree-repo.h
dbnicholson: Thanks, I saw that myself later.

src/libostree/ostree-repo-private.h
dbnicholson: OK, I'll do it. That said, I would like this to remain a simple enum and not migrate into some flags setup later.

cgwalters:
Yeah...we should probably try to move fsck into the library too.  But doesn't need to block this patch.

Let's try to get this cleaned up and then I'm happy to merge it and we can do followup commits?

dbnicholson:
@cgwalters Had another thought about the threads.

* If the lock to be taken is shared, only do it when outside a transaction. Assuming that the transaction is started on a thread that will live throughout the life of the transaction, then you already have a shared lock.

* If the lock to be taken is exclusive, add internal _unlocked versions that would be called in any place that would be run from a worker thread. So far I've only encountered `ostree_repo_delete_object` removing tombstones when writing out commits.

I have a bunch of fixups that I'll push out shortly that should address the comments so far.

dbnicholson:
OK, I added a bunch of fixups that I'll squash soon, but I wanted to give a chance to see the intermediate state if anyone was interested. I believe I addressed most of the comments so far.

TODO:

* Add the `core.locking` option to globally enable or disable locking
* Investigate locking from worker threads more
* Look at locking for core and remote config

cgwalters:
Are you going to have a chance to rebase this soon?  If not I might be able to do it early this week.

dbnicholson:
Yeah, I'll rebase later today. Some of the commits should be reordered a bit.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+          g_debug ("Closing repo lock file");
+          (void) close (lock->fd);
+        }
+      g_free (lock);
+    }
+}
+
+/* Wrapper to handle flock vs OFD locking based on GLnxLockFile */
+static gboolean
+do_repo_lock(int fd,
```
pwithnall: Nitpick: Missing space before `(`. Same for various methods below.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+#else
+  res = -1;
+  errno = EINVAL;
+#endif
+
+  /* Fallback to flock when OFD locks not available */
+  if (res < 0)
+    {
+      if (errno == EINVAL)
+        res = flock (fd, flags);
```
pwithnall: This needs a `TEMP_FAILURE_RETRY` loop.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+static gboolean
+do_repo_lock(int fd,
+             int flags)
+{
+  int res;
+
+#ifdef F_OFD_SETLK
+  struct flock fl = {
+    .l_type = (flags & ~LOCK_NB) == LOCK_EX ? F_WRLCK : F_RDLCK,
+    .l_whence = SEEK_SET,
```
pwithnall: Might be clearer to explicitly initialise `.l_start` and `.l_len` to 0 here.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+  };
+
+  res = fcntl (fd, (flags & LOCK_NB) ? F_OFD_SETLK : F_OFD_SETLKW, &fl);
```
pwithnall: Needs a `TEMP_FAILURE_RETRY` loop.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+static gboolean
+do_repo_unlock(int fd,
+               int flags)
+{
+  int res;
+
+#ifdef F_OFD_SETLK
+  struct flock fl = {
+    .l_type = F_UNLCK,
+    .l_whence = SEEK_SET,
```
pwithnall: Same comments as above function about explicitly initialising `.l_start` and `.l_len`, and `TEMP_FAILURE_RETRY`.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+      g_debug ("Creating repo lock table");
+      lock_table = g_hash_table_new_full (NULL, NULL, NULL,
+                                          (GDestroyNotify)free_repo_lock);
+      g_private_set (&repo_lock_table, lock_table);
+    }
+
+  OstreeRepoLock *lock = g_hash_table_lookup (lock_table, self);
+  if (lock == NULL)
+    {
+      lock = g_new (OstreeRepoLock, 1);
```
pwithnall: Nitpick: I’d always use `g_new0()` for paranoia reasons.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+      lock->stack = g_queue_new ();
+      g_debug ("Opening repo lock file");
+      lock->fd = openat (self->repo_dir_fd, ".lock",
```
pwithnall: I think this also needs `TEMP_FAILURE_RETRY`.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+  const char *cur_state_name;
+  if (stack_len == 0)
+    {
+      cur_state = LOCK_UN;
+      cur_state_name = "unlocked";
+    }
+  else
+    {
+      cur_state = GPOINTER_TO_INT (g_queue_peek_head (lock->stack));
+      cur_state_name = (cur_state == LOCK_EX) ? "exclusive" : "shared";
```
pwithnall: The code for calculating `cur_state_name` and `cur_state` is present in a couple of functions and could be abstracted out.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+ * ostree_repo_lock_push
+ * @self: a #OstreeRepo
+ * @lock_type: the type of lock to acquire
+ * @cancellable: a #GCancellable
+ * @error: a #GError
+ *
+ * Takes a lock on the repository and adds it to the lock stack. If @lock_type
+ * is %OSTREE_REPO_LOCK_SHARED, a shared lock is taken. If @lock_type is
+ * %OSTREE_REPO_LOCK_EXCLUSIVE, an exclusive lock is taken. shared lock is
+ * used. The actual lock state is only changed when locking a previously
```
pwithnall: Something looks wrong with the ‘shared lock is used’ mini-sentence. What’s it supposed to mean?

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+ * Since: 2017.13
+ */
+gboolean
+ostree_repo_lock_push (OstreeRepo          *self,
+                       OstreeRepoLockType   lock_type,
+                       GCancellable        *cancellable,
+                       GError             **error)
+{
+  g_return_val_if_fail (self != NULL, FALSE);
+  g_return_val_if_fail (self->inited, FALSE);
```
pwithnall: Missing preconditions:
```
g_return_val_if_fail (OSTREE_IS_REPO (self), FALSE);
g_return_val_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable), FALSE);
g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+
+  if (!self->writable)
+    return TRUE;
+
+  if (self->lock_timeout_seconds == -1)
+    {
+      g_debug ("Pushing lock blocking");
+      return push_repo_lock (self, lock_type, TRUE, error);
+    }
+  else
```
pwithnall: Maybe add a `g_assert (self->lock_timeout_seconds >= 0)` here.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+    {
+      gint waited = 0;
```
pwithnall: `waited` could be unsigned, and you could convert the non-negative `lock_timeout_seconds` to unsigned within this block.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+      g_debug ("Pushing lock non-blocking with timeout %d",
+               self->lock_timeout_seconds);
+      for (;;)
```
pwithnall: `while (TRUE)` might make it a bit more explicit that this loop is potentially infinite.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+                       remaining, (remaining == 1) ? "" : "s");
+            }
+          waited++;
+          sleep (1);
+        }
+    }
+}
+
+/**
+ * ostree_repo_lock_pop
```
pwithnall: Missing a `:` from the end of the function name (same problem in other gtk-doc comments).

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+ * Returns: %TRUE on success, otherwise %FALSE with @error set
+ * Since: 2017.13
+ */
+gboolean
+ostree_repo_lock_pop (OstreeRepo    *self,
+                      GCancellable  *cancellable,
+                      GError       **error)
+{
+  g_return_val_if_fail (self != NULL, FALSE);
+  g_return_val_if_fail (self->inited, FALSE);
```
pwithnall: Preconditions missing as in previous function.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+
+  if (!self->writable)
+    return TRUE;
+
+  if (self->lock_timeout_seconds == -1)
+    {
+      g_debug ("Popping lock blocking");
+      return pop_repo_lock (self, TRUE, error);
+    }
+  else
```
pwithnall: I suggest adding a `g_assert()` here and converting `lock_timeout_seconds` to unsigned within this block, as in the previous function.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+    {
+      gint waited = 0;
+      g_debug ("Popping lock non-blocking with timeout %d",
+               self->lock_timeout_seconds);
+      for (;;)
```
pwithnall: Suggest `while (TRUE)` as in the previous function.

src/libostree/ostree-repo.c
```diff
@@ -2271,6 +2783,38 @@ reload_core_config (OstreeRepo          *self,
...
+            return FALSE;
+          }
+      }
+    else
+      {
+        char *endptr = NULL;
+        gint64 lock_timeout = g_ascii_strtoll (lock_timeout_str, &endptr, 0);
+        if ((lock_timeout < -1) ||
+            (lock_timeout > G_MAXINT) ||
+            (lock_timeout == 0 && endptr == lock_timeout_str))
```
pwithnall: This doesn’t check for the case where `*endptr != '\0'`, which could be caused by `lock_timeout_str == "123somestring"`.

src/libostree/ostree-repo.c
```diff
@@ -5099,3 +5702,47 @@ ostree_repo_set_collection_id (OstreeRepo   *self,
...
+/**
+ * ostree_repo_get_lock_timeout:
+ * @self: an #OstreeRepo
+ *
+ * Get the current lock timeout for the repository. A timeout of -1 indicates
+ * that the application will block until the lock is acquired. Otherwise, it
+ * represents the number of seconds the application will sleep attempting to
+ * acquire the repository lock.
+ *
+ * Returns: lock timeout for the repository
```
pwithnall: Nitpick: maybe add `, in seconds` to this?

src/libostree/ostree-repo.c
```diff
@@ -5099,3 +5702,47 @@ ostree_repo_set_collection_id (OstreeRepo   *self,
...
+
+/**
+ * ostree_repo_set_lock_timeout:
+ * @self: an #OstreeRepo
+ * @timeout: new lock timeout in seconds
+ *
+ * Set the lock timeout for the repository. A timeout of -1 indicates that the
+ * application will block until the lock is acquired. Otherwise, it represents
+ * the number of seconds the application will sleep attempting to acquire the
+ * repository lock.
```
pwithnall: Should mention that anything `< -1` is not allowed.

src/libostree/ostree-repo.h
```diff
@@ -107,6 +107,51 @@ OstreeRepo *  ostree_repo_create_at (int             dfd,
 
 #ifdef OSTREE_ENABLE_EXPERIMENTAL_API
 
+/**
+ * OstreeRepoLockType:
+ * The type of repository lock to acquire.
+ * @OSTREE_REPO_LOCK_SHARED: A shared lock
+ * @OSTREE_REPO_LOCK_EXCLUSIVE: An exclusive lock
```
pwithnall: Missing a `Since:` line.

src/ostree/ot-main.c
```diff
@@ -48,6 +50,28 @@ static GOptionEntry repo_entry[] = {
...
+parse_lock_timeout (const char  *option_name,
+                    const char  *value,
+                    gpointer     data,
+                    GError     **error)
+{
+  char *endptr = NULL;
+  gint64 lock_timeout = g_ascii_strtoll (value, &endptr, 0);
+  if ((lock_timeout < -1) ||
+      (lock_timeout > G_MAXINT) ||
+      (lock_timeout == 0 && endptr == value))
```
pwithnall: As with the other `g_ascii_strtoll()` call, this should check for `*endptr != '\0'` as another error condition.

Makefile-libostree.am
```diff
@@ -261,6 +261,10 @@ OSTree-1.0.gir: libostree-1.la Makefile
 OSTree_1_0_gir_EXPORT_PACKAGES = ostree-1
 OSTree_1_0_gir_INCLUDES = Gio-2.0
 OSTree_1_0_gir_CFLAGS = $(libostree_1_la_CFLAGS)
+if ENABLE_EXPERIMENTAL_API
+# When compiling this is set via config.h, but g-ir-scanner doesn't use that
```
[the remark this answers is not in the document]
dbnicholson: Interesting, I didn't know that. I'll try that locally.

src/libostree/ostree-cmdprivate.c
```diff
@@ -1,3 +1,4 @@
+
```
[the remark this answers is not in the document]
dbnicholson: Whoops...

src/libostree/ostree-cmdprivate.c
```diff
@@ -48,7 +49,14 @@ ostree_cmd__private__ (void)
     impl_ostree_generate_grub2_config,
     _ostree_repo_static_delta_dump,
     _ostree_repo_static_delta_query_exists,
-    _ostree_repo_static_delta_delete
+    _ostree_repo_static_delta_delete,
+    /* Remove this when ostree_repo_set_lock_timeout is no longer experimental */
```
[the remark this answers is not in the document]
dbnicholson: Sure

src/libostree/ostree-cmdprivate.h
```diff
@@ -20,17 +20,34 @@
 #pragma once
 
 #include "ostree-types.h"
+/* This is only needed to get the OstreeRepoLockType and OstreeRepoAutoLock
```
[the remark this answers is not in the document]
dbnicholson: Sure

src/libostree/ostree-repo-checkout.c
```diff
@@ -991,6 +992,16 @@ checkout_tree_at (OstreeRepo                        *self,
       g_assert (options->force_copy);
     }
 
+  /* Take a shared repo lock to try to ensure objects aren't deleted. If the
+   * repo is not writable, this will be a noop and we just hope for the
```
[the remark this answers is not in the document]
dbnicholson: The case is simply an unprivileged user trying to checkout from /ostree/repo or /var/lib/flatpak/repo or something like that. They should be allowed to do a checkout as they were before.

Like you say, if the repo is not writable by the current user, then they're not going to be deleting any objects. But if, say, flatpak is pruning /var/lib/flatpak/repo while you're trying to do a checkout, bad things could happen.

But that's already the case, and I don't think we should optimize for unprivileged users. I'm really just trying to make sure that multiple processes running as the repo owner can't screw each other since that's by far the common case. Since the unprivileged user won't even be able to open the lock file (it's opened 600 particularly so that only repo owners can take exclusive locks), then I don't think they'll be able to get the lock state.

An inotify watch is interesting. I'm not sure inotify can tell you when the lock state changes, though, only if it's been opened.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,513 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+ * each thread maintains its own lock stack. See the OstreeRepoLock type
+ * below.
+ *
+ * The actual locking is done using either open file descriptor locks or
+ * flock locks. This allows the locking to work with concurrent
+ * processes. The lock file is held on the ".lock" file within the
+ * repository.
+ *
+ * The intended usage is to take a shared lock when writing objects or
+ * reading objects in critical sections. Exclusive locks are taken when
```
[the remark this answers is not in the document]
dbnicholson: Yeah, it's something I punted on because the semantics are a little weird. Internally, ostree reads it into memory once and then it's accessed via `ostree_repo_get_config` or `ostree_repo_copy_config`. I don't think locking there would be helpful.

What I think you could do is take a shared lock during `ostree_repo_reload_config` (when the files are actually read) and take an exclusive lock during `ostree_repo_write_config` and `ostree_repo_remote_{add,change,delete}`. But then I think it can't really be managed all the way internally because the typical pattern is `ostree_repo_copy_config`, update keyfile, `ostree_repo_write_config`. You'd really want to take an exclusive lock around the whole set so someone else doesn't change it while you're making keyfile modifications. That's certainly something flatpak does. But I didn't think all the way through it so punted for now.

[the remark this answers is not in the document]
dbnicholson: Oh, and `ostree_repo_reload_config` is called during `ostree_repo_open`, so I'd be worried about locking before you could even open the repo. But I guess you could probably skip the locking there if `self->inited` is `FALSE`.

src/libostree/ostree-repo.c
dbnicholson: Blah. I thought I was good about that.

src/libostree/ostree-repo.c
dbnicholson: I think you're right. I copied this from `GLnxLockFile`, but I agree that explicitly setting the members is clearer.

src/libostree/ostree-repo.c
dbnicholson: Good call.

src/libostree/ostree-repo.c
dbnicholson: Yeah, this can also get `EINTR` when you're doing a blocking lock.

src/libostree/ostree-repo.c
dbnicholson: Heh. I actually did that intentionally for similar paranoid reasons. If `free_repo_lock` is called and `lock->fd == 0`, then you'd close `stdin`. So, I explicitly ordered the initialization below so you wouldn't do that. But I guess the compiler could do something different either way... I'll change it to `g_new0` so we don't accidentally free random memory.

src/libostree/ostree-repo.c
dbnicholson: Good call.

src/libostree/ostree-repo.c
dbnicholson: I suppose. It's so simple, but I guess that would be nicer.

src/libostree/ostree-repo.c
dbnicholson: Whoops. Probably a rebase mishap. I'll clean that up.

src/libostree/ostree-repo.c
dbnicholson: Sure. I think you could make that argument for most every API in ostree, though...

src/libostree/ostree-repo.c
dbnicholson: Since any setter for `self->lock_timeout_seconds` already checks that it's `>= -1`, I decided not to bother here. But I could add the assert to make sure we don't mess up somewhere.

src/libostree/ostree-repo.c
dbnicholson: What's the advantage of that?

src/libostree/ostree-repo.c
dbnicholson: Doesn't make a big difference to me. `for (;;)` is used elsewhere in the ostree code, so I used that. Both constructs imply potentially infinite looping to me.

src/libostree/ostree-repo.c
dbnicholson: Right. I'd seen it there sometimes and sometimes not within ostree, but I agree that's the canonical format.

src/libostree/ostree-repo.c
dbnicholson: Good catch. I hadn't thought of that.

src/libostree/ostree-repo.c
dbnicholson: Sure.

src/libostree/ostree-repo.c
dbnicholson: OK

src/libostree/ostree-repo.h
dbnicholson: Good catch.

src/libostree/ostree-repo-checkout.c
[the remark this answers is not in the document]
pwithnall: Yeah, having done a bit of reading, I don’t think it’s possible to monitor the status of a file lock using inotify. The only way of doing it is to poll with `fcntl(F_GETLK)`, but that’s horrific and racy.

The other approach I can think of is to use a world-writeable directory (somewhere in `/run`, probably), which can contain the locks for many repositories, each with a deterministic name based on the repository directory’s device ID and inode.

That said, allowing an unprivileged process to take a lock on a repository it can’t write to allows priority inversion attacks, since there’s no realistic way to stop it taking an exclusive lock.

Unless we put shared locks in `/run` and exclusive locks in `$repo/.lock`?

src/libostree/ostree-repo.c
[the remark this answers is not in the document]
pwithnall: Seems reasonable to me to punt on this for now, but please file a follow-up issue about it.

src/libostree/ostree-repo.c
pwithnall: I find such assertions help to document the code, and also help keep static analysis tools on the straight and narrow.

src/libostree/ostree-repo.c
pwithnall: Eliminates the question of ‘what if `waited` ends up being negative?’.

src/libostree/ostree-repo.c
pwithnall: *shrug* :-)

src/libostree/ostree-repo.c
pwithnall: The `g_ascii_strto*()` are an absolute nightmare for correct error handling.

GLib recently acquired `g_ascii_string_to_{un,}signed()`, which are a lot better, but unfortunately too recent for libostree to use yet. :-(

cgwalters:
Would be happy to start trying to take this - perhaps it's simplest to split this up into a PR that adds the internal locking API with just a few uses - maybe just prune and commit?  Then we can gradually trickle in changes to enable it in different places like fsck as we test?

dbnicholson:
Sorry, got busy with a bunch of other stuff but I should be getting back to this today or tomorrow. I'll split it up into 2 PRs like you say. Some other stuff I've thought of in the meantime:

* A repo wide locking enable boolean. I hope this wouldn't be necessary much, but it would allow programs to opt out of locking if we run into issues with this or if there's a sequence that's just too complex for the internal locking.

* A thread specific locking enable boolean. This is my idea for handling the worker threads. What you'd do is take a lock, spawn the thread, and immediately disable locking for that thread. When the thread completed, the spawner would pop the lock. That would allow (for example) all the pull threads to run without locking while the thread spawner has the repo locked and avoid the issues with

dbnicholson:
I opened #1343 with the slimmed down initial setup as requested. I still passes the test suite, although I haven't really used it in real applications.

Makefile-libostree.am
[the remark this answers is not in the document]
dbnicholson: This was separately fixed in #1322.

src/libostree/ostree-cmdprivate.c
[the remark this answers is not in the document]
dbnicholson: Will address after #1343.

src/libostree/ostree-cmdprivate.c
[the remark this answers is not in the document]
dbnicholson: Will address after #1343.

src/libostree/ostree-cmdprivate.h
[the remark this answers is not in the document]
dbnicholson: Will address after #1343.

src/libostree/ostree-repo-checkout.c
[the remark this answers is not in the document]
dbnicholson: I still believe the way this was written where only the repo owner can manage the lock is correct. I think any attempt to allow non-owners to take a lock would be very complex and likely mean that flock-type locking can't be used.

src/libostree/ostree-repo.c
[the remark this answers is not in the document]
dbnicholson: Will address after #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343. I broke out a little helper and struct that seems to work correctly.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343. I put an assert for `>= -1` above the entire block.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Stlll sticking with `for(;;)` unless there's a tie-breaker :)

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343.

src/libostree/ostree-repo.c
dbnicholson: Fixed in #1343. As before, added an assert for `>= -1` above both blocks.

src/libostree/ostree-repo.c
dbnicholson: Sticking with `for(;;)` as above.

src/libostree/ostree-repo.c
dbnicholson: Will fix after #1343. This part didn't make it into the slimmed down setup.

src/libostree/ostree-repo.c
dbnicholson: Will fix after #1343.

src/libostree/ostree-repo.c
dbnicholson: Will fix after #1343.

src/libostree/ostree-repo.h
dbnicholson: Fixed in #1343.

src/ostree/ot-main.c
dbnicholson: Will fix after #1343.

cgwalters:
How do you want to proceed with rebasing this?  Are you planning to work on it anytime soon?

dbnicholson:
Sorry, I had to take off my ostree developer hat for a while. Next week I'll get back on this and flesh it out. Does that timeline work or were you looking for something sooner?

cgwalters:
No pressure on the timeline!  Currently locking is disabled by default.  I am happy to help out a bit if we can figure out how to subdivide work.  Just trying to keep things moving!

PR #1293, opened 2017-10-19
lib/repo-finder-avahi: Fix a leak in a GVariantIter loop

pwithnall:
Use g_variant_iter_loop() rather than next(), since it automatically
handles freeing the child memory each iteration. Previously, we leaked
it for all but the last iteration.


cgwalters:
@rh-atomic-bot r+ 3f9ffc6

PR #1294, opened 2017-10-19
lib/checkout: For "process whiteouts" mode, replace directories too

cgwalters:
I'm playing around with some ostree ⇔ OCI/Docker bits, and ran
into this while importing an OCI image that built from the Fedora
base image where `/home` is a regular directory, and I added a layer
that did the ostree bits of moving it to `/var` and leaving a symlink.

OCI/Docker supports this.  Now since "process whiteouts" is really the
"enable OCI/Docker" mode, let's only replace dirs if that's enabled.
This leaves the `UNION_FILES` targeted for its original use case
which is unioning components/packages.  (Although that use case itself
is now a bit superceded by `UNION_IDENTICAL`, but eh).

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -228,8 +228,24 @@ create_file_copy_from_input_at (OstreeRepo     *repo,
               return glnx_throw_errno_prefix (error, "symlinkat");
             case OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES:
               {
-                /* Unioning?  Let's unlink and try again */
-                (void) unlinkat (destination_dfd, destination_name, 0);
+                /* For unioning, we further bifurcate a bit; for the "process whiteouts"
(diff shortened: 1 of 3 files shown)


jlebon:
That makes sense to me. Maybe let's add a small blurb to the docstring for `OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES` in `ostree-repo.h`? E.g.:

```diff
- * @OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES: When layering checkouts, unlink() and replace existing files, but do not modify existing directories
+ * @OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES: When layering checkouts, unlink() and replace existing files, but do not modify existing directories (except if processing whiteouts)
```

?

cgwalters:
Sure, fixup ⬆️

jlebon:
@rh-atomic-bot r+ 2e754c6

PR #1295, opened 2017-10-19
repo: Add wrapper function for setting devino cache on checkout opts

cgwalters:
I was trying to use this with pygobject for an OCI+ostree project, and pygobject
rejected simply assigning to the field (understandably, since it can't bind the
lifetime together).

Add a wrapper function, which is still unsafe, but hides that unsafety
where most people shouldn't find it.  And if they do...well, sorry,
Rust wasn't invented when ostree was started.


src/libostree/ostree-repo-checkout.c
```diff
@@ -1176,6 +1176,24 @@ ostree_repo_checkout_at (OstreeRepo                        *self,
...
+/**
+ * ostree_repo_checkout_at_options_set_devino:
+ * @opts: Checkout options
+ * @cache: (transfer none) (nullable): Devino cache
+ *
+ * This function simply assigns @cache to the `devino_to_csum_cache` member of
+ * @opts; it's only useful for introspection.
+ *
+ * Note that cache does *not* have its refcount incremented - the lifetime of
+ * @cache must be equal to or greater than that of @opts.
```
pwithnall: Missing `Since: 2017.13`

cgwalters: Fixup ⬇️

jlebon:
@rh-atomic-bot r=pwithnall 41b1b3d

PR #1296, opened 2017-10-19
core: Add standard SOURCE_TITLE metadata key

cgwalters:
This is a freeform string useful to track/display when a commit is "derived"
from some other format.  For example, in the rpm-ostree test we make a
`vmcheck` ref that conceptually overlays the default ref like
`fedora-atomic:fedora/26/x86_64/atomic-host`.

My current patch sets the source title to e.g.
"Dev overlay on fedora-atomic:fedora/26/x86_64/atomic-host".

Another case I'm working on now is importing OCI images to use
as host images.  For that case, the source title is
With this patch we could then set the original OCI image name + tag
as the source name, like:
"oci:cgwalters/demo-custom-fedora-atomic-host:26".

### DIFF
--- a/src/libostree/ostree-core.h
+++ b/src/libostree/ostree-core.h
@@ -229,6 +229,23 @@ typedef enum {
  */
 #define OSTREE_COMMIT_META_KEY_ENDOFLIFE "ostree.endoflife"
 /**
+ * OSTREE_COMMIT_META_KEY_SOURCE_TITLE:
+ *
+ * GVariant type `s`. This should hold a relatively short single line value
+ * containing a human-readable "source" for a commit, intended to be displayed
+ * near the origin ref.  This is particularly useful for systems that inject
+ * content into an OSTree commit from elsewhere - for example, generating from
+ * an OCI or qcow2 image. Or if generating from packages, the enabled repository
+ * names and their versions.
+ *
+ * Try to keep this key short (e.g. < 80 characters) and human-readable; if you
+ * desire machine readable data, consider injecting separate metadata keys.
+ *
+ * Since: 2017.13
+ */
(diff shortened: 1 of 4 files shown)


cgwalters:
rpm-ostree consumption patch: https://github.com/projectatomic/rpm-ostree/pull/1069

jlebon:
This is a cool idea! Should we also add a related `SOURCE_CHECKSUM` key? I.e. a hash that represents the state of the source at the time. E.g. in the vmcheck case, it'd be the ostree commit hash of the original `fedora-atomic` commit. In the OCI case, it'd be the checksum of the image. And in the "from a bag of packages" case, it could be the goal hash, or the git commit hash, etc... The advantage of standardizing it vs letting each case invent its own key is that it makes it easier to introspect and display regardless of the source.

cgwalters:
I'm not sure...for the OCI case the checksum means a different thing than ostree.

For `rpm-ostree compose`, I'm not totally sure how useful the goal hash would be.

I think my high level feeling here is - let's let the relevant tools add whatever metadata keys are specific to what they're doing, and if there's something to standardize down the line, we circle back here and do that?

cgwalters:
But hold off on merging, I'm going to update this PR to also render in `ostree admin status`.

cgwalters:
OK, rebased 🏄 and updated with some prep commits and adds testing.  Lifting WIP.

jlebon:
Looks good! Though the tests are complaining it seems.

cgwalters:
Whee, that was fun.  Provoked a uboot integration bug.  Fixup ⬆️

jlebon:
@rh-atomic-bot r+ 339a094

PR #1297, opened 2017-10-20
lib/commit: Fix hardlink checkout commit with bare-user + mod xattrs

cgwalters:
This is more subtle fallout from:
https://github.com/ostreedev/ostree/pull/1170
AKA commit: 8fe45362578a43260876134d6547ebd0bb2485c3

Before, if we found a devino cache hit, we'd use it unconditionally.

Recall that `bare-user` repositories are very special in that they're the only
mode where the on disk state ("physical state") is not the "real" state. The
latter is stored in the `user.ostreemeta` xattr. (`bare-user` repos are also
highly special in that symlinks are regular files physically, but that's not
immediately relevant here).

Since we now have `bare-user-only` for the "pure unprivileged container" case,
`bare-user` should just be used for "OS builds" which have nonzero uids (and
possibly SELinux labels etc.)

In an experimental tool I'm writing "skopeo2ostree" which imports OCI images
into refs, then squashes them together into a single final commit, we lost the
the `81` group ID for `/usr/libexec/dbus-1/dbus-daemon-launch-helper`.

This happened because the commit code was loading the "physical" disk state,
where the uid/gid are zero because that's the uid I happened to be using. We
didn't just directly do the link speedup because I was using `--selinux-policy`
which caused the xattrs to change, which caused us to re-commit objects from the
physical state.

The unit test I added actually doesn't quite trigger this, but I left
it because "why not".  Really testing this requires the installed test
which uses SELinux policy from `/`.

The behavior without this fix looks like:

```
-00755 0 0     12 { [(b'user.ostreemeta', [byte 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x51, 0x00, 0x00, 0x81, 0xed]), (b'security.selinux', b'system_u:object_r:lib_t:s0')] } /usr/lib/dbus-daemon-helper
```

which was obviously totally broken - we shouldn't be picking up the
`user.ostreemeta` xattr and actually committing it of course.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -2543,6 +2543,7 @@ get_final_xattrs (OstreeRepo                       *self,
                   GFile                            *path,
                   int                               dfd,
                   const char                       *dfd_subpath,
+                  GVariant                         *source_xattrs,
                   GVariant                        **out_xattrs,
                   gboolean                         *out_modified,
                   GCancellable                     *cancellable,
@@ -2558,7 +2559,9 @@ get_final_xattrs (OstreeRepo                       *self,
   g_autoptr(GVariant) original_xattrs = NULL;
   if (!skip_xattrs && !self->disable_xattrs)
     {
-      if (path && OSTREE_IS_REPO_FILE (path))
+      if (source_xattrs)
+        original_xattrs = g_variant_ref (source_xattrs);
+      else if (path && OSTREE_IS_REPO_FILE (path))
         {
           if (!ostree_repo_file_get_xattrs (OSTREE_REPO_FILE (path), &original_xattrs,
                                             cancellable, error))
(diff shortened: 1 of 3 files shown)


cgwalters:
I think down the line we'll need to better distinguish between new content and existing hardlinks to make this pattern work well.  For example doing unprivileged OS builds, we want to specify `ostree commit --owner-uid 0` but only have that apply to *new* files.  Which actually is what gnome-continuous has been doing for a long time, and that PR broke it too.

Maybe something like a flag in the callbacks that say whether or not the object existed, then callbacks can decide what to do?

jlebon:
Ouch, yeah that makes sense. Just a question about the commin message:

> The unit test I added actually doesn't quite trigger this, but I left
it because "why not".  Really testing this requires the installed test
which uses SELinux policy from `/`.

It looks like this was done now, right?

cgwalters:
Yep, that's done, just noting why I kept the unit test.

jlebon:
Oh I see. I misunderstood the "Really testing this requires..." to mean it wasn't part of this patch when it clearly was. But makes sense now!

@rh-atomic-bot r+ fc23e29

PR #1301, opened 2017-10-23
lib/repo-finder-avahi: Fix memory corruption of a GVariantIter

pwithnall:
A GVariantIter* was being passed to a GVariant format string varargs,
rather than a GVariantIter**. This resulted in memory corruption.

So we can continue to reuse ref_map throughout the function, make it a
GVariantIter* rather than a stack-allocated GVariantIter.


cgwalters:
@rh-atomic-bot r+ 1fcd1ab

PR #1303, opened 2017-10-24
lib/repo-finder-mount: Fix propagating NULL errors

pwithnall:
These two code paths tried to propagate errors which had never been set.
Set new errors instead.


cgwalters:
@rh-atomic-bot r+ a9bf3d2

PR #1304, opened 2017-10-24
ci: Update centos container to current again

cgwalters:
Repo sync issue should be cleared now, and for some reason the previous
container is already GC'd.


cgwalters:

peterbaouoft:
@rh-atomic-bot r+ b041656

PR #1305, opened 2017-10-24
lib/repo: Properly handle NULL homedir when signing commit

dbnicholson:
Without this, ostree_repo_sign_commit throws a critical message when no
homedir is provided:

(ostree gpg-sign:5034): GLib-GIO-CRITICAL **: g_file_new_for_path: assertion 'path != NULL' failed


cgwalters:
@rh-atomic-bot r+ 709ffa2

PR #1307, opened 2017-10-25
lib/repo-finder-mount: Ignore mounts which have a ‘system’ file system

pwithnall:
For example, tmpfs or a cgroup file system. This is basically an
optimisation of the list of file systems we check for repositories,
since we would never expect any of these file systems to be capable of
containing a repository.

Depends on the new API from
https://bugzilla.gnome.org/show_bug.cgi?id=788927.

---

This can’t be merged until https://bugzilla.gnome.org/show_bug.cgi?id=788927 is merged into GLib, since we don’t have a more fine-grained check for the new APIs other than the advertised GLib version. Note this will force anyone building with GLib git master ≥ 2.55 (but from an older commit than the new APIs) to update their GLib.

### DIFF
--- a/src/libostree/ostree-repo-finder-mount.c
+++ b/src/libostree/ostree-repo-finder-mount.c
@@ -23,6 +23,7 @@
 #include "config.h"
 
 #include <gio/gio.h>
+#include <gio/gunixmounts.h>
 #include <glib.h>
 #include <glib-object.h>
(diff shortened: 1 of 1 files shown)


cgwalters:
Marking as WIP until the GLib patch lands.

cgwalters:
@rh-atomic-bot r+ 90ac6e2

Issue #1306, opened 2017-10-25
Not reading commitpartial state from parent repo


alexlarsson:
When loading a commit with ostree_repo_load_commit() it can load the commit from ->parent_repo. However, it never reads the commitpartial state from the parent, even when the commit is from there.

This leads to issues when doing a delta update from a repo with a parent, if the commit is partial and in the parent it will still try to do a delta update which may reference files that are not locally available, as in https://github.com/flatpak/flatpak/pull/1126

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1308

PR #1308, opened 2017-10-25
lib/repo: Fix loading commitstate with parent repos

cgwalters:
This makes the code nicer too. Properly unit testing this though really wants
like a whole set of stuff around parent repos...but we do have coverage of the
non-parent path in the current pull tests.

Closes: https://github.com/ostreedev/ostree/issues/1306


alexlarsson:
Looks good to me.
@rh-atomic-bot r+ f1f714e

PR #1309, opened 2017-10-25
Streaming static deltas

alexlarsson:
This adds a new OtVariantBuilder helper that supports creating potentially large GVariants by directly writing them to a file descriptor, rather than building them up in memory. It contains some copied private GVariant code from glib which is used as the basis for the builder.

It then uses this builder to write the delta superblock. Additionally it changes the way the delta parts are written so that we don't have to keep all the deltas in memory, but can write them to a temporary file as they are done. For inlined parts these temporary files are then copied directly into the superblock fd.

This drops the requirement for generating static deltas a lot. In a test creating a (non-inlined) static delta for the gnome sdk went from 1.5 gig virt/1g res to 600 meg virt/300m res.

### DIFF
--- a/src/libotutil/ot-variant-builder.c
+++ b/src/libotutil/ot-variant-builder.c
@@ -1,0 +1,1219 @@
+/*
+ * Copyright (C) 2017 Alexander Larsson <alexl@redhat.com>.
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the
+ * Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+ * Boston, MA 02111-1307, USA.
+ *
+ * Author: Alexander Larsson <alexl@redhat.com>.
+ */
+
+#include "config.h"
+
+#include "ot-variant-builder.h"
+#include "libglnx/libglnx.h"
+
+/*****************************************************************************************
+ * This code is copied from gvariant in glib. With the following copyright:
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * Copyright © 2007, 2008 Ryan Lortie
+ * Copyright © 2010 Codethink Limited
+ *****************************************************************************************/
+
+typedef struct _GVariantTypeInfo GVariantTypeInfo;
+
+#define G_VARIANT_TYPE_INFO_CHAR_MAYBE      'm'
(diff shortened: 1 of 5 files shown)


alexlarsson:
Note, the deltas produced were identical, sans the timestamp.

alexlarsson:
This will help https://github.com/flatpak/flatpak/issues/483

alexlarsson:
Gah, it failed because of the permissions on the superblock. Honestly glnx_link_tmpfile_at() should probably take a mode argument to avoid issues like this in the future.

alexlarsson:
```
src/libotutil/ot-variant-builder.c:1028:23: error: unused variable 'keep_around_until_return' [-Werror,-Wunused-variable]
  g_autoptr(GVariant) keep_around_until_return = g_variant_ref_sink (value);
```
It is used, by unref:ing it when we go out of scope. @cgwalters Do you know how to fix this warning?

alexlarsson:
Finally it passed CI :)

src/libotutil/ot-variant-builder.c
```diff
@@ -0,0 +1,1219 @@
...
+  /* 'x' */ { fixed_aligned(8) },   /* int64 */
+  /* 'y' */ { fixed_aligned(1) },   /* byte */
+#undef fixed_aligned
+#undef not_a_type
+#undef unaligned
+#undef aligned
+};
+
+static GRecMutex g_variant_type_info_lock;
+static GHashTable *g_variant_type_info_table;
```
cgwalters: Hmmm...won't this possibly break if there are two copies of this in process?  (BTW last I measured this type info table was part of what made GVariant slow)

cgwalters: I'm hesistant a bit to do this without at least opening up a GLib bug and tracking this API addition there.

cgwalters:
(Just two quick comments...will look at this more today)

The other thing that occurs to me here is I think there's an argument that GVariant should only be used for metadata.  Basically have [metadata][payload] and the total size of the payload is in metadata. Which is how it works today for the content objects in ostree.  But...that'd be a format change.

src/libotutil/ot-variant-builder.c
alexlarsson: Why would it break? Its jus a local static variable which protects a hashtable. It doesn't affect anything outside this file.

alexlarsson:
The gvariant wire format depends on knowing the total size of the container, so you cannot append arbitrary stuff after it and still work, without also adding some kind of header size field external to the gvariant.

alexlarsson:
btw, i think the second patch in this series that adds the finish_part() code could probably go in (with minor changes) even without the streaming writes. It will be less of a gain, but still a gain.

cgwalters:
> The gvariant wire format depends on knowing the total size of the container, so you cannot append arbitrary stuff after it and still work, without also adding some kind of header size field external to the gvariant.

Yes, but for content objects (and we could make this true for delta parts), total size = size of file.

src/libotutil/ot-variant-builder.c
cgwalters: I was thinking the `GVariant` APIs might be relying on something like pointer identity of `GVariantType`  or something.    But yeah, it doesn't *look* like that's the case.  And in the end we're not returning a `GVariant`, so data transfer is just "one way" here.

cgwalters:
I haven't looked closely, but I have a feeling it wouldn't be hard to add this to GLib directly if rather than using `write()` we used `mmap()` on the fd, and `fallocate()` to extend it.

cgwalters:
Anyways I'm OK to try merging this, but before I do I'm playing with `manual-tests/static-delta-generate-crosscheck.sh` on some FAH content and seeing how things work.  That unearthed a few bugs (not apparently related to this)

alexlarsson:
Also, for that typeinfo hashtable, i copied the code as is. But it spends a lot of work building up some kind of offset table for quickly looking up members by offset, in generate_table(). If we only write stuff that code could be deleted (as well as some parts of GVariantMemberInfo.

I kept it there for now as I was thinking of possibly using this on the reader side when applying the static delta in streaming fashion.

alexlarsson:
@cgwalters if you mmap you run into issues on 32bit writing files > 2 GB.

cgwalters:
Yeah...but deltas were designed so that you'd generally make them on a "real" build server (they're not arch dependent).  That said I understand for making bundles one wants it to work on "small" hardware too.

Anyways I tracked down the deltas issue I hit when testing this which wasn't related: https://github.com/ostreedev/ostree/pull/1312

cgwalters:
OK, I verified this passes my FAH test case too with my 2 other PRs applied.

@rh-atomic-bot r+ b200880

PR #1311, opened 2017-10-27
lib: Minor static delta fixes

cgwalters:
First, the manual crosscheck script bitrotted; it got caught up
in the "use libtest repo creation wrapper" bit, and also it
seems like at some point `pull --require-static-deltas` changed
meaning when dealing with `file:///` repos.  I have more work to
unwind that.

Next, I'm seeing a delta failure which looks like a static delta
miscompilation with rollsums; change the compiler to print out
the source object too, which helped me debug this.

And finally in the processing code, fix incorrect error prefixing, which was
misleading.

### DIFF
--- a/manual-tests/static-delta-generate-crosscheck.sh
+++ b/manual-tests/static-delta-generate-crosscheck.sh
@@ -42,11 +42,11 @@ assert_streq() {
 }
 
 validate_delta_options() {
-    mkdir testrepo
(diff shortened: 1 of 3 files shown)


cgwalters:
@rh-atomic-bot r=ashcrow bbae2f3

cgwalters:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1313, opened 2017-10-27
lib/pull: When --require-static-deltas, use them even for file:/// repos

cgwalters:
I didn't fully spelunk this, but from what `static-delta-generate-crosscheck.sh`
had, we appeared to be doing this before, and it's clearly useful for local
testing rather than needing to spin up a HTTP server.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1438,6 +1438,10 @@ static char *
 get_real_remote_repo_collection_id (OstreeRepo  *repo,
                                     const gchar *remote_name)
 {
+  /* remote_name == NULL can happen for pull-local */
+  if (!remote_name)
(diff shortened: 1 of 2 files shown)


jlebon:
@rh-atomic-bot r+ 5a9f752

jlebon:
@rh-atomic-bot retry

jlebon:
```
FAIL: libostree/test-delta.sh.test (Child process exited with code 133)
```

Hmm, the same (and only) test failed both in the PR test and the merge test. Probably not a flake?

cgwalters:
Yep, triggered an `--enable-experimental-api` bug.  Rebased 🏄 and fixup ⬆️

jlebon:
@rh-atomic-bot r+ 08c5dea

PR #1317, opened 2017-11-01
static deltas: Set optional flag for superblock

taheris:
I believe static delta fetching should be optional. Following on from the changes made in [this commit](https://github.com/ostreedev/ostree/pull/1004/files), this PR sets the `OSTREE_FETCHER_REQUEST_OPTIONAL_CONTENT` flag for fetching the superblock as well.


cgwalters:
Yep, thanks!

@rh-atomic-bot r+ c313b16

PR #1318, opened 2017-11-01
Update libglnx

cgwalters:
Mostly for https://github.com/GNOME/libglnx/pull/94
since I keep seeing it in the terminal.

Update submodule: libglnx


jlebon:
@rh-atomic-bot r+ 0b140b9

PR #1320, opened 2017-11-02
build: Add -Werror=undef by default, fix fallout

cgwalters:
The main thing here is that a ton of stuff has happened in gnulib since we
imported `parse-datetime.y`. I cherry-picked a little bit of it, but that
upstream doesn't seem to build with `-Wundef`, so I just deleted some hunks.

(Note I reindented the warnings consistently)

Update submodule: libglnx


jlebon:
bot, retest this please

jlebon:
@rh-atomic-bot r+ 2538994

PR #1322, opened 2017-11-03
Enable bindings to experimental APIs

dbnicholson:
I'd like to use the experimental APIs from python, but `g-ir-scanner` isn't currently including them. Do that and fix up a couple other issues seen along the way.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -261,6 +261,10 @@ OSTree-1.0.gir: libostree-1.la Makefile
 OSTree_1_0_gir_EXPORT_PACKAGES = ostree-1
 OSTree_1_0_gir_INCLUDES = Gio-2.0
 OSTree_1_0_gir_CFLAGS = $(libostree_1_la_CFLAGS)
+if ENABLE_EXPERIMENTAL_API
+# When compiling this is set via config.h, but g-ir-scanner can't use that
+OSTree_1_0_gir_CFLAGS += -DOSTREE_ENABLE_EXPERIMENTAL_API=1
+endif
 OSTree_1_0_gir_LIBS = libostree-1.la
 OSTree_1_0_gir_SCANNERFLAGS = --warn-all --identifier-prefix=Ostree --symbol-prefix=ostree $(GI_SCANNERFLAGS)
 OSTree_1_0_gir_FILES = $(libostreeinclude_HEADERS) $(filter-out %-private.h %/ostree-soup-uri.h $(libostree_experimental_headers),$(libostree_1_la_SOURCES))
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -54,7 +54,7 @@ src/libostree/ostree-enumtypes.h: src/libostree/ostree-enumtypes.h.template $(EN
 	--template $< \
 	$(ENUM_TYPES) > $@.tmp && mv $@.tmp $@
 
-src/libostree/ostree-enumtypes.c: src/libostree/ostree-enumtypes.c.template $(ENUM_TYPES)
+src/libostree/ostree-enumtypes.c: src/libostree/ostree-enumtypes.c.template src/libostree/ostree-enumtypes.h $(ENUM_TYPES)
 	$(AM_V_GEN) $(GLIB_MKENUMS) \
 	--template $< \
 	$(ENUM_TYPES) > $@.tmp && mv $@.tmp $@
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -278,7 +278,7 @@ ostree_validate_remote_name (const char  *remote_name,
 
 /**
  * ostree_validate_collection_id:
- * @rev: (nullable): A collection ID
+ * @collection_id: (nullable): A collection ID
  * @error: Error
  *
  * Check whether the given @collection_id is valid. Return an error if it is
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -5539,8 +5539,11 @@ check_remote_matches_collection_id (OstreeRepo  *repo,
   return g_str_equal (remote_collection_id, collection_id);
 }
 
+/* FIXME: Export this to bindings once OstreeRemote is properly registered as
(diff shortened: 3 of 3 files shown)


dbnicholson:
CC @pwithnall.

Makefile-libostree.am
```diff
@@ -54,7 +54,7 @@ src/libostree/ostree-enumtypes.h: src/libostree/ostree-enumtypes.h.template $(EN
 	--template $< \
 	$(ENUM_TYPES) > $@.tmp && mv $@.tmp $@
 
-src/libostree/ostree-enumtypes.c: src/libostree/ostree-enumtypes.c.template $(ENUM_TYPES)
+src/libostree/ostree-enumtypes.c: src/libostree/ostree-enumtypes.c.template src/libostree/ostree-enumtypes.h $(ENUM_TYPES)
 	$(AM_V_GEN) $(GLIB_MKENUMS) \
 	--template $< \
 	$(ENUM_TYPES) > $@.tmp && mv $@.tmp $@
```
pwithnall: Note: We could also change this to use `--output $@` to avoid the `.tmp`-and-`mv` dance. The full recommended rules (from the [new documentation](https://bugzilla.gnome.org/show_bug.cgi?id=788948)) are:
```
# A list of headers to inspect
project_headers = \
        project-foo.h \
        project-bar.h \
        project-baz.h

enum-types.h: $(project_headers) enum-types.h.in
        $(AM_V_GEN)$(GLIB_MKENUMS) \
                --template=enum-types.h.in \
                --output=$@ \
                $(project_headers)

enum-types.c: $(project_headers) enum-types.c.in enum-types.h
        $(AM_V_GEN)$(GLIB_MKENUMS) \
                --template=enum-types.c.in \
                --output=$@ \
                $(project_headers)

BUILT_SOURCES += enum-types.h enum-types.c
CLEANFILES += enum-types.h enum-types.c
EXTRA_DIST += enum-types.h.in enum-types.c.in
```

Makefile-libostree.am
```diff
@@ -261,6 +261,10 @@ OSTree-1.0.gir: libostree-1.la Makefile
 OSTree_1_0_gir_EXPORT_PACKAGES = ostree-1
 OSTree_1_0_gir_INCLUDES = Gio-2.0
 OSTree_1_0_gir_CFLAGS = $(libostree_1_la_CFLAGS)
+if ENABLE_EXPERIMENTAL_API
+# When compiling this is set via config.h, but g-ir-scanner can't use that
+OSTree_1_0_gir_CFLAGS += -DOSTREE_ENABLE_EXPERIMENTAL_API=1
```
pwithnall: I can’t remember where I commented about this before, but I think it’s possible to just pass `config.h` to `g-ir-scanner` in its list of files to scan (as the first file), and then it will end up being `#include`d in the generated C file which it compiles to introspect the GTypes.

I haven’t actually tried that though. Does it work?

pwithnall: Yes, I did comment on this recently: https://github.com/ostreedev/ostree/pull/1292/files#r147126774

Did you try it and it didn’t work? :-(

pwithnall:
Other than those comments, these look like useful fixes. Well spotted, thanks. :-)

Makefile-libostree.am
dbnicholson: Sorry, meant to mention that. It's kind of in the commit message. I did try it and `g-ir-scanner` was not happy. It expects that all the macros are going to follow your namespace (i.e., `OSTREE_*`), but there are a ton of random macros in there like `LT_OBJDIR`. So, I think you can do it, but it will output a ton of warnings. I think for this one special case it's easier to just pass the flag on the command line.

cgwalters:
LGTM, will let @pwithnall do the `r+`.

Makefile-libostree.am
dbnicholson: If it's important to you, I can make this change, but it will only delay getting the ref binding stuff you need landed :)

Makefile-libostree.am
pwithnall: Cool, thanks for trying it. Could you [file a bug against gobject-introspection](https://bugzilla.gnome.org/enter_bug.cgi?product=gobject-introspection) about making this a bit nicer to use please? It’s a fairly common idiom, I think, and it would be nice if `g-ir-scanner` made it a bit easier to pass in `config.h`.

Makefile-libostree.am
pwithnall: I’ll do it as a follow-up.

pwithnall:
@rh-atomic-bot r+ e7c895feeaaa6c49e81487a1f85fa62f39aa13b9

Makefile-libostree.am
dbnicholson: Done - https://bugzilla.gnome.org/show_bug.cgi?id=789881. I honestly don't know if much can be done, though, since I don't know how g-ir-scanner would know the difference between random macros you want to ignore and macros you want parsed that don't follow your namespace.

dbnicholson:
@cgwalters Seems the atomic builders aren't actually running. Any way to get that going or merge this?

cgwalters:
@rh-atomic-bot r=pwithnall e7c895f

cgwalters:
@dbnicholson We hit the github API rate limit, should be cleared now.

dbnicholson:
@cgwalters Sorry, still looks to be stuck. BTW, can you remind me where the @rh-atomic-bot commands are documented?

cgwalters:
Here: https://github.com/projectatomic/papr#more-details-about-project-atomic-ci-services

Unfortunately it's currently backed by an internal Jenkins so it's hard to debug externally.  We have some background work on that.

@rh-atomic-bot retry

cgwalters:
It looks like we may have already went back up to the rate limit 😢
>[6] {'message': 'API rate limit exceeded for rh-atomic-bot.', 'documentation_url': 'https://developer.github.com/v3/#rate-limiting'}

Makefile-libostree.am
pwithnall: Done: https://github.com/ostreedev/ostree/pull/1329

Issue #1321, opened 2017-11-03
Error executing pull --require-static-deltas when up to date


abferm:
When executing a pull with static deltas required I get an error complaining about no delta from null to the current commit if my local branch is up to date with the remote. I would expect this not to be the case.

```
root@host:/var/system/home/user# ostree pull --url ${OSTREE_URL} --require-static-deltas sota ${OSTREE_BRANCH}

Receiving metadata objects: 0/(estimating) -/s 0 bytes                                                                                                                                                                              
error: Static deltas required, but none found for (null) to ec475e438571a6b74cf91df4f7734957155e446375e5e46667d3de226c38b42d
root@host:/var/system/home/user# ostree pull --url ${OSTREE_URL} sota ${OSTREE_BRANCH}

Scanning metadata: 1918
```

cgwalters:
Hmm, I'm not reproducing this immediately with git master.  Tried adding this to the tests:

```
diff --git a/tests/pull-test.sh b/tests/pull-test.sh
index 3f8030e0..e7653cc0 100644
--- a/tests/pull-test.sh
+++ b/tests/pull-test.sh
@@ -379,6 +379,8 @@ else
     ${CMD_PREFIX} ostree --repo=repo rev-parse ${delta_target}
 fi
 ${CMD_PREFIX} ostree --repo=repo fsck
+# And doing another pull with no changes shouldn't fail
+${CMD_PREFIX} ostree --repo=repo pull --require-static-deltas origin ${delta_target}
 done
 
 cd ${test_tmpdir}
```

which I expected to fail but didn't.  That said...the code is complex here, I'd have expected us to handle this in `initiate_request()` but I'm not seeing that.  Looking at this more...

cgwalters:
Oh wait, this was fixed by https://github.com/ostreedev/ostree/commit/5b860fba05651b075e10e19b295a1d00ebb5b8e2 which is in the latest release.  Can you try that if you're not already using v2017.13?

cgwalters:
Nope, reverting that doesn't introduce the bug.  I think I see the problem now...

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1323

abferm:
I've noticed that FROM and TO on a static delta can be the same commit. Should such a static delta be considered a good workaround for this issue?

PR #1323, opened 2017-11-03
lib/pull: Avoid error if current with --require-static-deltas

cgwalters:
A tricky thing here that caused this to go past a lot of our tests
is that the code was mostly OK if there was an available delta from
an older commit.  But this case broke if we e.g. had a new OS
deployment and did a `--require-static-deltas` pull, i.e. the initial
state.

I cleaned up our "find static delta state" function to return an enumeration,
and extended it with an "already have the commit" state.  A problem
I then hit is that we've historically fetched detached metadata for
non-delta pulls, even if the commit hasn't changed.  I decided not to
do that for `--require-static-deltas` pulls for now; otherwise the
code gets notably more complex.

Closes: https://github.com/ostreedev/ostree/issues/1321

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2325,19 +2325,40 @@ process_one_static_delta (OtPullData                 *pull_data,
   return ret;
 }
 
-/* Loop over the static delta data we got from the summary,
- * and find the newest commit for @out_from_revision that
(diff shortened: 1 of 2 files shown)


cgwalters:
Fixup ⬆️

jlebon:
@rh-atomic-bot r+ dfddf14

PR #1325, opened 2017-11-03
tests: Don't symlink rofiles-fuse if it's disabled

dbnicholson:
Creating the symlink will cause make to try to build rofiles-fuse, which
will fail if it's disabled. Normally I wouldn't disable rofiles-fuse,
but it's triggering a hang in our ARM Xenial builder's kernel in splice.
I'm sure that's fuse's fault, but for now I just need to disable
rofiles-fuse there and found --disable-rofiles-fuse didn't actually
work.


cgwalters:
@rh-atomic-bot r+ 9b4737c

cgwalters:
@rh-atomic-bot retry

PR #1326, opened 2017-11-06
grub-generator: If OSTREE_BOOT_PARTITION is not set, default to /boot

clopez:
## Summary of the issue
Unable to boot with grub after deploying a new image with ostree admin

## Description
I'm using Yocto/poky and [meta-updater](https://github.com/advancedtelematic/meta-updater).

After pulling the new generated ostree repository, I deploy it on the target device.

Everything goes fine but with the grub config generated. It generates a wrong config because `OSTREE_BOOT_PARTITION` is not set when i run the following from the terminal on the device:
```
ostree admin deploy --os=poky my-remote:genericx86-64
```

This generates a config where the kernel and initramfs files point to a non existent path at ```/ostree/poky-c24b509436e2f75d9e2f8b6491b7f6b3d1007ed8d67328fdf03508314c6bd5a1/vmlinuz ``` when the right path should be ```/boot/ostree/poky-c24b509436e2f75d9e2f8b6491b7f6b3d1007ed8d67328fdf03508314c6bd5a1/vmlinuz```

### DIFF
--- a/src/boot/grub2/ostree-grub-generator
+++ b/src/boot/grub2/ostree-grub-generator
@@ -61,7 +61,12 @@ read_config()
 
 populate_menu()
 {
-    boot_prefix="${OSTREE_BOOT_PARTITION}"
+    # Default to /boot if OSTREE_BOOT_PARTITION is not set
+    if [ -z ${OSTREE_BOOT_PARTITION+x} ]; then
+        boot_prefix="/boot"
+    else
+        boot_prefix="${OSTREE_BOOT_PARTITION}"
+    fi
     for config in $(ls ${entries_path}); do
         read_config ${config}
         menu="${menu}menuentry '${title}' {\n"
--- a/src/boot/grub2/ostree-grub-generator
+++ b/src/boot/grub2/ostree-grub-generator
@@ -61,7 +61,7 @@ read_config()
 
 populate_menu()
 {
-    # Default to /boot if OSTREE_BOOT_PARTITION is not set and /boot and / are on the save device
+    # Default to /boot if OSTREE_BOOT_PARTITION is not set and /boot and / are on the same device
     if [ -z ${OSTREE_BOOT_PARTITION+x} ] && [ $(df / /boot|awk '{print $1}'|uniq|wc -l) -eq 2 ]; then
         boot_prefix="/boot"
     else
--- a/src/boot/grub2/ostree-grub-generator
+++ b/src/boot/grub2/ostree-grub-generator
@@ -61,8 +61,8 @@ read_config()
 
 populate_menu()
 {
-    # Default to /boot if OSTREE_BOOT_PARTITION is not set
-    if [ -z ${OSTREE_BOOT_PARTITION+x} ]; then
+    # Default to /boot if OSTREE_BOOT_PARTITION is not set and /boot and / are on the save device
+    if [ -z ${OSTREE_BOOT_PARTITION+x} ] && [ $(df / /boot|awk '{print $1}'|uniq|wc -l) -eq 2 ]; then
         boot_prefix="/boot"
     else
         boot_prefix="${OSTREE_BOOT_PARTITION}"
--- a/src/boot/grub2/ostree-grub-generator
+++ b/src/boot/grub2/ostree-grub-generator
@@ -61,8 +61,8 @@ read_config()
 
 populate_menu()
 {
-    # Default to /boot if OSTREE_BOOT_PARTITION is not set and /boot and / are on the same device
-    if [ -z ${OSTREE_BOOT_PARTITION+x} ] && [ $(df / /boot|awk '{print $1}'|uniq|wc -l) -eq 2 ]; then
+    # Default to /boot if OSTREE_BOOT_PARTITION is not set and /boot is on the same device than ostree/repo
+    if [ -z ${OSTREE_BOOT_PARTITION+x} ] && [ -d /boot/ostree ] && [ -d /ostree/repo ] && [ $(stat -c '%d' /boot/ostree) -eq $(stat -c '%d' /ostree/repo) ]; then
         boot_prefix="/boot"
     else
         boot_prefix="${OSTREE_BOOT_PARTITION}"


cgwalters:
Urgh, yeah that's fallout from https://github.com/ostreedev/ostree/pull/1320 but just for older GLib, right now Travis isn't gating, and our gating CI doesn't build on old GLib yet.

cgwalters:
(I'll look at this)

src/boot/grub2/ostree-grub-generator
```diff
@@ -61,7 +61,12 @@ read_config()
 
 populate_menu()
 {
-    boot_prefix="${OSTREE_BOOT_PARTITION}"
+    # Default to /boot if OSTREE_BOOT_PARTITION is not set
```
cgwalters: Hmm...I'm worried that some people are using ostree without a `/boot` partition and may be relying on this.  We should probably try to detect this dynamically?  Something like `test $(($(findmnt /boot | wc -l) > 0)) = 1` or something?  Or actually we already detect this inside `ostree-sysroot-deploy.c`; we could pass that via environment variable to this script?

clopez: Actually I'm using OSTree without a ```/boot``` partition :)
So I understand this may break things for people that has ```/boot``` in a different partition, as in that case ```/boot/something/vmlinuz``` becomes just ```/something/vmlinuz``` when grub access the partition directly to load the kernel.

The suggestion of just checking if ```findmnt /boot``` reports something is not going to work because ```/boot``` is bind-mounted by [ostree-prepare-root in the initramfs](https://github.com/advancedtelematic/meta-updater/blob/afd18c7/recipes-sota/ostree-initrd/files/init.sh#L66) before doing the ```pivot_root``` and starting the actual system.

One possible idea is to check if ```/boot``` and ```/``` point to the same target device (```/dev/sda2``` in my case)

```
root@genericx86-64:~# findmnt /boot
TARGET SOURCE           FSTYPE OPTIONS
/boot  /dev/sda2[/boot] ext4   rw,relatime,data=ordered

root@genericx86-64:~# findmnt /    
TARGET SOURCE                                                                                                   FSTYPE OPTIONS
/      /dev/sda2[/ostree/deploy/poky/deploy/03dd817d1940a972a96256d952ba66dd8d5b7a43db5d19f048fd33d9c6c867b9.0] ext4   rw,relatime,data=ordered

root@genericx86-64:~# cat /proc/mounts | grep sda2
/dev/sda2 /sysroot ext4 rw,relatime,data=ordered 0 0
/dev/sda2 / ext4 rw,relatime,data=ordered 0 0
/dev/sda2 /boot ext4 rw,relatime,data=ordered 0 0
/dev/sda2 /usr ext4 ro,relatime,data=ordered 0 0
```

src/boot/grub2/ostree-grub-generator
```diff
@@ -61,7 +61,7 @@ read_config()
 
 populate_menu()
 {
-    # Default to /boot if OSTREE_BOOT_PARTITION is not set and /boot and / are on the save device
+    # Default to /boot if OSTREE_BOOT_PARTITION is not set and /boot and / are on the same device
     if [ -z ${OSTREE_BOOT_PARTITION+x} ] && [ $(df / /boot|awk '{print $1}'|uniq|wc -l) -eq 2 ]; then
```
cgwalters: How about comparing `stat -c '%d %i'`?

But that said I think I'd prefer to pass this down by env from the upgrader which already determined this, see the `is_ro_mount()` bits - we'd just need to record whether we found a mount or not, in addition to whether it's read-only.

(But I'd also merge the patch if we used the `stat` too)

src/boot/grub2/ostree-grub-generator
cgwalters: Ah, I wasn't aware of that initramfs implementation.  Interesting.

src/boot/grub2/ostree-grub-generator
clopez: The inode number is different in my use case:
```
root@genericx86-64:~# stat -c '%d %i' / /boot
2050 68658
2050 22
```
Will you be ok with just comparing the device number (%d)?

cgwalters: Hmm...but bind mounts should have the same underlying device/inode, right?

clopez: The bind-mount on ```/boot``` points to the inode of ```/sysroot/boot``` which is different from the inode of ```/```

```
root@genericx86-64:~# stat -c '%d %i' /boot /sysroot/boot/            
2050 22
2050 22
```

And I'm not sure if comparing against ```/sysroot/boot```rather than / is a good idea

cgwalters:
Urgh, that bind mount of `/boot` thing keeps confusing me.  Partly because due to needing to support dm-crypt etc. (as I'm using on the laptop where I'm typing this) I just don't use ostree without a `/boot` partition.

Also now that I look closely at this it looks like we'll also need to tweak this script for e.g. https://github.com/ostreedev/ostree/pull/1021

So...how about comparing subpaths like `stat -c '%D' /boot/ostree/ /ostree/repo`?  That should be pretty reliable.

cgwalters:
(That's roughly similar to what we do for installing the kernel; we try just doing `linkat()` from the repo, only falling back to a copy for cross device)

clopez:
I have just discovered that grub has an utility named ```grub-mkrelpath``` that gives you the real path of a file on its own partition. And the grub script that generate grub.cfg use this (at least on Debian).

For example I get this:
```
root@genericx86-64:~# grub-mkrelpath /boot/ostree/poky-c24b509436e2f75d9e2f8b6491b7f6b3d1007ed8d67328fdf03508314c6bd5a1/vmlinuz
/ostree/deploy/poky/deploy/8e4cac51d1f84c5a8273f7524710f7f7434fe332b1fd00b88dbb23c6d7b0e3bd.0/boot/ostree/poky-c24b509436e2f75d9e2f8b6491b7f6b3d1007ed8d67328fdf03508314c6bd5a1/vmlinuz

root@genericx86-64:~# grub-mkrelpath /boot/ostree/poky-56a177f413c6328789931fd2b6fd4ada001cb491e25f794669ff118534230110/initramfs
/ostree/deploy/poky/deploy/8e4cac51d1f84c5a8273f7524710f7f7434fe332b1fd00b88dbb23c6d7b0e3bd.0/boot/ostree/poky-56a177f413c6328789931fd2b6fd4ada001cb491e25f794669ff118534230110/initramfs

root@genericx86-64:~# grub-mkrelpath /boot/
/ostree/deploy/poky/deploy/8e4cac51d1f84c5a8273f7524710f7f7434fe332b1fd00b88dbb23c6d7b0e3bd.0/boot
```
So one possible alternative solution (that at first sight looks more robust) is to call this utility to get the real path for the kernel and the initramfs on the partition that boots.

However this would introduce a dependency on grub for OSTree, because it would need to run this  ```grub-mkrelpath``` on the target system when it deploys a new OStree commit to generate the grub config file.

Currently the system doesn't need to have grub installed even if it boots via grub (as is with the case of this test distro that i'm using built with Yocto)

cgwalters:
Right, the point of using this is you don't need to drag grub into the system.

cgwalters:
Sorry about the delay.  This looks pretty plausible to me.  Since we don't have much CI coverage for this, I just want to double check - you tested this and it works?

cgwalters:
One other detail is to double check it works with ash/dash etc.  That we could easily add to CI.

clopez:
I tested this with ash, dash and bash and it worked fine.

cgwalters:
@rh-atomic-bot r+ 7af23ca

PR #1327, opened 2017-11-06
build: Work around -Wundef with older GLib

cgwalters:
This previous PR caused our `-Werror=undef` to trigger when building
with older GLib; this should work around it.


jlebon:
@rh-atomic-bot r+ f2f2675

PR #1328, opened 2017-11-06
build: Also fix -Werror=undef for old libsoup

cgwalters:
Same change as before, but for libsoup.


cgwalters:
Let's check Travis this time too.

jlebon:
@rh-atomic-bot r+ 367da99

PR #1329, opened 2017-11-07
build: Use --output parameter for glib-mkenums to avoid tmp file switch

pwithnall:
Avoid renaming through a temporary file by using the --output parameter
for glib-mkenums, which handles this internally.

This introduces no functional changes.


cgwalters:
Target GLib for CentOS7 and the Debian versions in Travis aren't new enough.  Not sure we can do much other than close this for now?

pwithnall:
Oh bother. That’s not likely to change soon, is it. Yeah, best close it for now. The patch can always be resurrected later. (Do you have a github label for that? Or maybe a milestone?)

cgwalters:
Maybe rework the patch to just add a "# TODO GLIB_CHECK_VERSION > 2.5X: use --output" then we can later `git grep GLIB_CHECK_VERSION`?

Issue #1331, opened 2017-11-07
Crash in curl_cb when installing flatpaks through gnome-software


kalev:
```
GLib:ERROR:ghash.c:377:g_hash_table_lookup_node: assertion failed: (hash_table->ref_count > 0)

Thread 75 "pool" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffca7fc700 (LWP 14534)]
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	}
(gdb) bt
#0  0x00007ffff437469b in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff43763b1 in __GI_abort () at abort.c:79
#2  0x00007ffff745b69d in g_assertion_message (domain=domain@entry=0x7ffff747cfee "GLib", file=file@entry=0x7ffff7481a62 "ghash.c", line=line@entry=377, func=func@entry=0x7ffff7481d10 <__func__.10214> "g_hash_table_lookup_node", message=message@entry=0x7fffbc060590 "assertion failed: (hash_table->ref_count > 0)") at gtestutils.c:2436
#3  0x00007ffff745b72a in g_assertion_message_expr (domain=domain@entry=0x7ffff747cfee "GLib", file=file@entry=0x7ffff7481a62 "ghash.c", line=line@entry=377, func=func@entry=0x7ffff7481d10 <__func__.10214> "g_hash_table_lookup_node", expr=expr@entry=0x7ffff7481a48 "hash_table->ref_count > 0") at gtestutils.c:2459
#4  0x00007ffff7423236 in g_hash_table_lookup_node (hash_return=<synthetic pointer>, key=0x7fffbc5110b0, hash_table=0x3b90060) at ghash.c:377
#5  0x00007ffff7423236 in g_hash_table_remove_internal (hash_table=0x3b90060, key=0x7fffbc5110b0, notify=1) at ghash.c:1359
#6  0x00007fffe10fe907 in sock_cb () at /lib64/libostree-1.so.1
#7  0x00007fffe0c02231 in Curl_multi_closed (conn=conn@entry=0x7fffbc5c8800, s=s@entry=49) at ../../lib/multi.c:2477
#8  0x00007fffe0bfed49 in Curl_closesocket (conn=0x7fffbc5c8800, sock=49) at ../../lib/connect.c:1289
[...]

PR #1332, opened 2017-11-07
fetcher/curl: Fix invalid memory access in finalize()

kalev:
Reorder cleanup functions so that curl_multi_cleanup() runs before
self->sockets is destroyed. This avoids an assert and invalid memory
access in sock_cb where self->sockets is dereferenced during
curl_multi_cleanup().

Closes: https://github.com/ostreedev/ostree/issues/1331


cgwalters:
bot, add author to whitelist
and list of awesome people

cgwalters:
@rh-atomic-bot r+ d1d6110

kalev:
Thanks :) Mind if I backport this to Fedora?

cgwalters:
go go gadget backport

PR #1333, opened 2017-11-07
deltas: Don't try to rollsum/bsdiff .xz files

cgwalters:
Fedora switched to 'xz' compress kernel modules, and recently
[RHEL7 did too](https://bugzilla.redhat.com/show_bug.cgi?id=1367496).
This compression defeats bsdiff.

While we have a "rollsum-able" test, we don't have a "bsdiff-able" test as it'd
be very expensive (we'd have to bsdiff, then apply it and compare the result).

Let's do the tactical quick fix here and just not try to delta files ending in
`.xz.`. This avoids us using bsdiff pointlessly for over 4000 files, which is
quite a notable speed increase for generating deltas.

### DIFF
--- a/src/libostree/ostree-repo-static-delta-compilation-analysis.c
+++ b/src/libostree/ostree-repo-static-delta-compilation-analysis.c
@@ -232,8 +232,15 @@ sizename_is_delta_candidate (OstreeDeltaContentSizeNames *sizename)
        * but it's not clear to me that's a major win; we'd still need to
        * maintain a list of compression formats, and this doesn't require
        * allocation.
+       * NB: We explicitly don't have .gz here in case someone might be
+       * using --rsyncable for that.
        */
-      if (g_str_has_suffix (name, ".xz"))
+      const char *dot = g_strrchr (name, '.');
+      if (!dot)
+        continue;
+      const char *extension = dot+1;
+      /* Don't add .gz here, see above */
+      if (g_str_equal (extension, "xz") || g_str_equal (extension, "bz2"))
         return FALSE;
     }
 
@@ -301,7 +308,6 @@ _ostree_delta_compute_similar_objects (OstreeRepo                 *repo,
       const guint64 max_threshold = to_sizenames->size *
         (1.0+similarity_percent_threshold/100.0);
 
-      /* Don't build candidates for the empty object */
       if (!sizename_is_delta_candidate (to_sizenames))
         continue;
 
--- a/src/libostree/ostree-repo-static-delta-compilation-analysis.c
+++ b/src/libostree/ostree-repo-static-delta-compilation-analysis.c
@@ -217,6 +217,30 @@ string_array_nonempty_intersection (GPtrArray    *a,
   return FALSE;
 }
 
+static gboolean
+sizename_is_delta_candidate (OstreeDeltaContentSizeNames *sizename)
+{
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-static-delta-compilation-analysis.c
```diff
@@ -278,20 +302,16 @@ _ostree_delta_compute_similar_objects (OstreeRepo                 *repo,
         (1.0+similarity_percent_threshold/100.0);
 
       /* Don't build candidates for the empty object */
```
jlebon: Nit: repeated comment.

src/libostree/ostree-repo-static-delta-compilation-analysis.c
```diff
@@ -217,6 +217,30 @@ string_array_nonempty_intersection (GPtrArray    *a,
...
+  /* Look for known non-delta-able files (currently just compression like xz) */
+  for (guint i = 0; i < sizename->basenames->len; i++)
+    {
+      const char *name = sizename->basenames->pdata[i];
+      /* We could replace this down the line with g_content_type_guess() or so,
+       * but it's not clear to me that's a major win; we'd still need to
+       * maintain a list of compression formats, and this doesn't require
+       * allocation.
+       */
+      if (g_str_has_suffix (name, ".xz"))
```
jlebon: While we're here, let's also catch e.g. `.gz` and `.bzip2` ?

cgwalters: `.gz` files can be built with `--rsyncable` which is explicitly designed to be friendly to the rsync-style rollsum algorithm we have.  I don't offhand know if many of them are - we do it for the initramfs which is a key target, but it also doesn't end in `.gz` as a filename. 

It looks like there are *vast* number of these files in atomicws mostly due to man pages, CUPS drivers, Emacs Lisp, and console fonts.  Of those, the only one in Atomic Host is console fonts which rev very infrequently.

Hmm.  Dunno.  I think the problem here is it probably doesn't make sense to try hard to delta most of those, but I'm worried about people who might be using `--rsyncable`.

As far as `.bz2`...there aren't many, I have a grand total of 7 hits in atomicws:

```
find /usr/ -name '*.bz2'
/usr/share/doc/dejavu-fonts-common/status.txt.bz2
/usr/share/doc/guile/NEWS.bz2
/usr/share/doc/libgomp/ChangeLog.bz2
/usr/share/doc/libgomp/ChangeLog.graphite.bz2
/usr/share/doc/ncurses/NEWS.bz2
/usr/share/doc/zfs-fuse/contrib/initrd-for-zfs-fuse-on-root.tar.bz2
/usr/share/selinux/packages/container.pp.bz2
```

All in `/usr/share/doc` except the selinux one.   Yeah, guess we might as well ignore that.

jlebon: > Hmm. Dunno. I think the problem here is it probably doesn't make sense to try hard to delta most of those, but I'm worried about people who might be using `--rsyncable`.

Oh right, we don't want to mess that up. Would be interesting to see if we could find a sweet spot by applying a cut-off file size. Anyway, not worth thinking much more about this for now!

jlebon:
@rh-atomic-bot r+ e276103

PR #1336, opened 2017-11-09
bin/summary: Fix --raw option

dbnicholson:
I wanted to inspect a summary file the other day and was saddened to
find it was broken:

$ ostree summary --raw
error: No option specified; use -u to update summary

CC @pwithnall

### DIFF
--- a/src/ostree/ot-builtin-summary.c
+++ b/src/ostree/ot-builtin-summary.c
@@ -210,7 +210,7 @@ ostree_builtin_summary (int argc, char **argv, OstreeCommandInvocation *invocati
             return FALSE;
         }
     }
-  else if (opt_view)
+  else if (opt_view || opt_raw)
     {
       g_autoptr(GBytes) summary_data = NULL;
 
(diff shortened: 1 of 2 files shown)


pwithnall:
This could do with a test case to make sure it doesn’t break again.

dbnicholson:
There was a test case, but it was using `--view --raw`. Changed that back to just `--raw` since that's (IMO) more normal and was the interface that was in place before.

cgwalters:
Hm...I'm not sure I'd say it's "broken"; having `--raw` conceptually be a modifier for `--view` seems intentional, though of course this type of thing probably wants subcommands like `ostree summary update` and `ostree summary view --raw`.

That said I'm fine with this.

@rh-atomic-bot r+ 4bec1bc

dbnicholson:
Right, not "broken", but `ostree summary --raw` was the command that had been in place before. Once I saw that test I thought maybe that was intentional, but IMO it makes more sense as a standalone option.

PR #1337, opened 2017-11-09
More experimental bindings

dbnicholson:
This makes `OstreeRemote` and `OstreeCollectionRef` to be actually usable from bindings.

CC @pwithnall


pwithnall:
@rh-atomic-bot r+ 039215859404ea2963e106a876243fdc16f98b27

PR #1338, opened 2017-11-09
lib/traverse: Port to new style

cgwalters:
Not prep for anything, was just reading this code a bit while
working on rpm-ostree jigdo.


jlebon:
LGTM! I just pushed a minor stylistic fixup to make it look similar to other new-style code of that form.
@rh-atomic-bot r+ 363be5b

Issue #1019, opened 2017-07-19
interrupting/resuming 'rpm-ostree upgrade' multiple times can consume all available disk space


miabbott:
Migrated from projectatomic/rpm-ostree#884

While banging on the interrupt/resume functionality of `rpm-ostree upgrade`, I noticed that I eventually ran out of disk space and was not able to complete the upgrade.

Investigation into what was using all the disk space pointed me in the direction of `/sysroot/ostere/repo/tmp` where it looks like multiple instances of 'fetcher' and 'staging' can be created.

```
# du -h -d 1 /sysroot/ostree/repo/tmp
0       /sysroot/ostree/repo/tmp/cache
27M     /sysroot/ostree/repo/tmp/fetcher-eMI9TC
6.5M    /sysroot/ostree/repo/tmp/staging-36c1069a-9263-4b73-a1a6-38c8744dffde-oqqU1I
30M     /sysroot/ostree/repo/tmp/staging-36c1069a-9263-4b73-a1a6-38c8744dffde-Gx7br4
88M     /sysroot/ostree/repo/tmp/staging-36c1069a-9263-4b73-a1a6-38c8744dffde-ns3s0n
32M     /sysroot/ostree/repo/tmp/fetcher-gfyB14
67M     /sysroot/ostree/repo/tmp/fetcher-R4WjmE
134M    /sysroot/ostree/repo/tmp/staging-36c1069a-9263-4b73-a1a6-38c8744dffde-QEPW3Y
251M    /sysroot/ostree/repo/tmp/staging-36c1069a-9263-4b73-a1a6-38c8744dffde-tP0nLh
20M     /sysroot/ostree/repo/tmp/fetcher-CKfrZN
76M     /sysroot/ostree/repo/tmp/fetcher-cMgwDr
13M     /sysroot/ostree/repo/tmp/fetcher-hrAlhP
573M    /sysroot/ostree/repo/tmp/staging-36c1069a-9263-4b73-a1a6-38c8744dffde-jDOqLz
2.1M    /sysroot/ostree/repo/tmp/fetcher-prjXz4
39M     /sysroot/ostree/repo/tmp/staging-36c1069a-9263-4b73-a1a6-38c8744dffde-KbktZR
29M     /sysroot/ostree/repo/tmp/fetcher-NH92DT
1.4G    /sysroot/ostree/repo/tmp
```

[...]

PR #1343, opened 2017-11-15
Slimmer repo locking

dbnicholson:
This is a slimmed down version of the locking in #1292. It contains just the locking push/pop interfaces and internal locking for transactions and pruning. This should hopefully be easier to review and merge and start testing how the locking actually affects real life use cases.

I still intend to add more of the features in #1292, but will wait on review of this PR. Most of the comments in #1292 have been addressed here. I'll add comments back on that PR where applicable.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -108,6 +108,7 @@ _installed_or_uninstalled_test_scripts = \
 	tests/test-xattrs.sh \
 	tests/test-auto-summary.sh \
 	tests/test-prune.sh \
+	tests/test-concurrency.py \
 	tests/test-refs.sh \
 	tests/test-demo-buildsystem.sh \
 	tests/test-switchroot.sh \
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -37,6 +37,8 @@ G_BEGIN_DECLS
 #define _OSTREE_MAX_OUTSTANDING_FETCHER_REQUESTS 8
 #define _OSTREE_MAX_OUTSTANDING_DELTAPART_REQUESTS 2
 
+#define _OSTREE_DEFAULT_LOCK_TIMEOUT_SECONDS 30
+
 /* In most cases, writing to disk should be much faster than
  * fetching from the network, so we shouldn't actually hit
  * this. But if using pipelining and e.g. pulling over LAN
@@ -157,6 +159,7 @@ struct OstreeRepo {
   guint64 tmp_expiry_seconds;
   gchar *collection_id;
   gboolean add_remotes_config_dir; /* Add new remotes in remotes.d dir */
+  gint lock_timeout_seconds;
 
   OstreeRepo *parent_repo;
 };
@@ -435,6 +438,23 @@ _ostree_repo_get_remote_inherited (OstreeRepo  *self,
                                    GError     **error);
 
 #ifndef OSTREE_ENABLE_EXPERIMENTAL_API
+
+/* All the locking APIs below are duplicated in ostree-repo.h. Remove the ones
+ * here once it's no longer experimental.
+ */
+
+typedef enum {
+  OSTREE_REPO_LOCK_SHARED,
+  OSTREE_REPO_LOCK_EXCLUSIVE
+} OstreeRepoLockType;
+
+gboolean      ostree_repo_lock_push (OstreeRepo          *self,
+                                     OstreeRepoLockType   lock_type,
+                                     GCancellable        *cancellable,
+                                     GError             **error);
+gboolean      ostree_repo_lock_pop (OstreeRepo    *self,
+                                    GCancellable  *cancellable,
+                                    GError       **error);
 
 const gchar * ostree_repo_get_collection_id (OstreeRepo   *self);
 gboolean      ostree_repo_set_collection_id (OstreeRepo   *self,
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -122,6 +122,7 @@ struct OstreeRepo {
 
   GMutex txn_lock;
   OstreeRepoTxn txn;
+  gboolean txn_locked;
 
   GMutex cache_lock;
   guint dirmeta_cache_refcount;
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -456,6 +456,15 @@ gboolean      ostree_repo_lock_pop (OstreeRepo    *self,
                                     GCancellable  *cancellable,
                                     GError       **error);
 
+typedef OstreeRepo OstreeRepoAutoLock;
+
+OstreeRepoAutoLock * ostree_repo_auto_lock_push (OstreeRepo          *self,
+                                                 OstreeRepoLockType   lock_type,
+                                                 GCancellable        *cancellable,
+                                                 GError             **error);
+void          ostree_repo_auto_lock_cleanup (OstreeRepoAutoLock *lock);
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (OstreeRepoAutoLock, ostree_repo_auto_lock_cleanup)
+
 const gchar * ostree_repo_get_collection_id (OstreeRepo   *self);
 gboolean      ostree_repo_set_collection_id (OstreeRepo   *self,
                                              const gchar  *collection_id,
--- a/src/libostree/ostree-repo-prune.c
+++ b/src/libostree/ostree-repo-prune.c
@@ -23,6 +23,7 @@
 
 #include "ostree-core-private.h"
 #include "ostree-repo-private.h"
+#include "ostree-autocleanups.h"
 #include "otutil.h"
 
 typedef struct {
@@ -160,12 +161,20 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
  * Prune static deltas, if COMMIT is specified then delete static delta files only
  * targeting that commit; otherwise any static delta of non existing commits are
  * deleted.
+ *
+ * This function takes an exclusive lock on the @self repository.
  */
 gboolean
 ostree_repo_prune_static_deltas (OstreeRepo *self, const char *commit,
                                  GCancellable      *cancellable,
                                  GError           **error)
 {
+  g_autoptr(OstreeRepoAutoLock) lock =
+    ostree_repo_auto_lock_push (self, OSTREE_REPO_LOCK_EXCLUSIVE, cancellable,
+                                error);
+  if (!lock)
+    return FALSE;
+
   g_autoptr(GPtrArray) deltas = NULL;
   if (!ostree_repo_list_static_delta_names (self, &deltas,
                                             cancellable, error))
@@ -286,6 +295,8 @@ repo_prune_internal (OstreeRepo        *self,
  * Use the %OSTREE_REPO_PRUNE_FLAGS_NO_PRUNE to just determine
  * statistics on objects that would be deleted, without actually
  * deleting them.
+ *
+ * This function takes an exclusive lock on the @self repository.
  */
 gboolean
 ostree_repo_prune (OstreeRepo        *self,
@@ -297,6 +308,12 @@ ostree_repo_prune (OstreeRepo        *self,
                    GCancellable      *cancellable,
                    GError           **error)
 {
+  g_autoptr(OstreeRepoAutoLock) lock =
+    ostree_repo_auto_lock_push (self, OSTREE_REPO_LOCK_EXCLUSIVE, cancellable,
+                                error);
+  if (!lock)
+    return FALSE;
+
   g_autoptr(GHashTable) objects = NULL;
   gboolean refs_only = flags & OSTREE_REPO_PRUNE_FLAGS_REFS_ONLY;
 
@@ -391,6 +408,8 @@ ostree_repo_prune (OstreeRepo        *self,
  *
  * The %OSTREE_REPO_PRUNE_FLAGS_NO_PRUNE flag may be specified to just determine
  * statistics on objects that would be deleted, without actually deleting them.
+ *
+ * This function takes an exclusive lock on the @self repository.
  */
 gboolean
 ostree_repo_prune_from_reachable (OstreeRepo        *self,
@@ -401,6 +420,12 @@ ostree_repo_prune_from_reachable (OstreeRepo        *self,
                                   GCancellable      *cancellable,
                                   GError           **error)
 {
+  g_autoptr(OstreeRepoAutoLock) lock =
+    ostree_repo_auto_lock_push (self, OSTREE_REPO_LOCK_EXCLUSIVE, cancellable,
+                                error);
+  if (!lock)
+    return FALSE;
+
   g_autoptr(GHashTable) objects = NULL;
 
   if (!ostree_repo_list_objects (self, OSTREE_REPO_LIST_OBJECTS_ALL | OSTREE_REPO_LIST_OBJECTS_NO_PARENTS,
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -612,6 +612,66 @@ ostree_repo_lock_pop (OstreeRepo    *self,
     }
 }
 
+/**
+ * ostree_repo_auto_lock_push: (skip)
+ * @self: a #OstreeRepo
+ * @lock_type: the type of lock to acquire
+ * @cancellable: a #GCancellable
+ * @error: a #GError
+ *
+ * Like ostree_repo_lock_push(), but for usage with #OstreeRepoAutoLock.
+ * The intended usage is to declare the #OstreeRepoAutoLock with
+ * g_autoptr() so that ostree_repo_auto_lock_cleanup() is called when it
+ * goes out of scope. This will automatically pop the lock status off
+ * the stack if it was acquired successfully.
+ *
+ * |[<!-- language="C" -->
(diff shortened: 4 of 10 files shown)


src/libostree/ostree-repo-private.h
```diff
@@ -415,6 +419,28 @@ _ostree_repo_get_remote_inherited (OstreeRepo  *self,
 
 #ifndef OSTREE_ENABLE_EXPERIMENTAL_API
 
+typedef enum {
+  OSTREE_REPO_LOCK_SHARED,
+  OSTREE_REPO_LOCK_EXCLUSIVE
+} OstreeRepoLockType;
```
pwithnall: Worthwhile to add a comment mentioning that this is duplicated in `ostree-repo.h`, should be kept in sync, and should be removed when the API graduates from being experimental.

src/libostree/ostree-repo.h
```diff
@@ -107,6 +107,47 @@ OstreeRepo *  ostree_repo_create_at (int             dfd,
 
 #ifdef OSTREE_ENABLE_EXPERIMENTAL_API
 
+/**
+ * OstreeRepoLockType:
+ * The type of repository lock to acquire.
```
pwithnall: Conventionally, this description line goes below the `@` lines. i.e.:
```
/**
 * OstreeRepoLockType:
 * @OSTREE_REPO_…
 * @OSTREE_REPO_…
 *
 * The type of repository lock to acquire.
 *
 * Since: 2017.14
 */
```

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,531 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+    {
+      OstreeRepoLockInfo info;
+      repo_lock_info (lock, &info);
+
+      g_debug ("Free lock: state=%s, depth=%u", info.name, info.len);
+      g_queue_free (lock->stack);
+      if (lock->fd >= 0)
+        {
+          g_debug ("Closing repo lock file");
+          (void) close (lock->fd);
```
pwithnall: Do we want to unlink the lock file at any point, for cleanliness?

Although I can’t think of any way that can be race-free, so probably best avoided.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,531 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+    return NULL;
+  return (OstreeRepoAutoLock *)self;
+}
+
+/**
+ * ostree_repo_auto_lock_cleanup: (skip)
+ * @lock: a #OstreeRepoAutoLock
+ *
+ * A cleanup handler for use with ostree_repo_auto_lock_push(). If @lock
+ * is not %NULL, ostree_repo_lock_pop() will be called on it.
```
pwithnall: Nitpick: Should probably mention that if `ostree_repo_lock_pop()` fails, a critical warning will be emitted.

src/libostree/ostree-repo.c
```diff
@@ -517,6 +1042,10 @@ ostree_repo_finalize (GObject *object)
   g_clear_pointer (&self->remotes, g_hash_table_destroy);
   g_mutex_clear (&self->remotes_lock);
 
+  GHashTable *lock_table = g_private_get (&repo_lock_table);
+  if (lock_table)
+    g_hash_table_remove (lock_table, self);
```
pwithnall: Maybe add:
```
if (g_hash_table_size (lock_table) == 0)
  g_private_replace (&repo_lock_table, NULL);
```
to save a bit of memory if `OstreeRepo` is only used sporadically throughout a process’ lifetime.

src/libostree/ostree-repo-private.h
dbnicholson: Good point. I'll add that comment.

src/libostree/ostree-repo.c
dbnicholson: IMO, you never want to unlink the lock file since it breaks the semantics of file descriptor locking. At the kernel level, the lock can only be managed if all the callers are operating on the same inode. If you unlink it, then another process trying to acquire the lock would then lock a different inode and the lock is useless. I can show you a concrete example we hit in the Endless jenkins scripts with flock if you'd like.

GLnxLockFile does unlink, but it then goes to great lengths when acquiring the lock to make sure that the file hasn't been unlinked and loops until that's not the case. I don't think that complexity outweighs the simple "don't delete the lock file" advice.

src/libostree/ostree-repo.c
dbnicholson: Good call.

src/libostree/ostree-repo.c
dbnicholson: Oh, that's nice. I think it's almost always the case that applications are operating on one repository, so keeping the hash table around all the time always bugged me.

src/libostree/ostree-repo.h
dbnicholson: Good call.

src/libostree/ostree-repo.c
pwithnall: > IMO, you never want to unlink the lock file since it breaks the semantics of file descriptor locking. At the kernel level, the lock can only be managed if all the callers are operating on the same inode. If you unlink it, then another process trying to acquire the lock would then lock a different inode and the lock is useless. I can show you a concrete example we hit in the Endless jenkins scripts with flock if you'd like.

Precisely.

> GLnxLockFile does unlink, but it then goes to great lengths when acquiring the lock to make sure that the file hasn't been unlinked and loops until that's not the case. I don't think that complexity outweighs the simple "don't delete the lock file" advice.

I agree with you.

dbnicholson:
This keeps failing in the ci during `distcheck`. I don't know what's up with that. It passes locally for me.

dbnicholson:
Seems the failures are with libcurl. Here's the relevant part of the travis log:
```
n_committers 4 n_pruners 0
Creating tree tree0
Creating tree tree1
PID 29946 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 29947 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 29948 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 29949 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
error: opendir(staging-b8946530-b5b1-428c-95f7-d830efa8cfd2-FZ5kMg): No such fil
e or directory
process 29948 exited with code 1
1..2
bb3f77a722d57118f3ab3aa3bbc4fa7d8e0b0e403672e37efe11b5fc29bdda92
85c0d517f6fb42b821baf62849b752b9affdde8961fd1b3ea3b9515fcff3128d
01239e103532dea86edef4d7ffe918bbd4a380a67e93f5897091a168a74b9056
ERROR: tests/test-concurrency.py - too few tests run (expected 2, got 0)
ERROR: tests/test-concurrency.py - exited with status 1
```
I guess with libcurl the transaction directory is handled differently?

dbnicholson:
Hmm, no. Even the f26-libsoup builder failed in normal `make check`. I think it's possible that the staging directory handling is generally racy.

dbnicholson:
Finally got an error locally after running `test-concurrency.py` repeatedly. My error was a bit different, though:
```
n_committers 4 n_pruners 0
Creating tree tree0
Creating tree tree1
PID 8848 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 8849 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 8850 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 8851 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
error: No such file or directory
process 8849 exited with code 1
```
Now to debug...

dbnicholson:
@cgwalters I think there are problems in `_ostree_repo_allocate_tmpdir` or `GLnxTmpDir`. I'm seeing a process attempt to reuse a staging directory after another one has removed it. Still debugging...

cgwalters:
Yeah :frowning:.  See https://github.com/ostreedev/ostree/issues/1019 which is related - the problem is right now any txn commit cleans up the whole `tmp/` dir in `cleanup_tmpdir()`.  We should probably just seralize that whole function.

dbnicholson:
Yeah, there are definitely issues with how the tmp directories get used, but I just don't understand where the race is. Even if the staging directories are getting wiped, it should be safe as other processes try to lock them. I spent hours trying to debug this and couldn't quite narrow it down.

One very weird thing. Even on a fresh repository, eventually the exact tempdir `staging-8a1d3c00-143b-4283-b43c-b36518df17d8-SZTteF` would get reused by 2 processes and fail. Like the `SZTteF` is supposedly random, but it repeated every time I ran `prune-concurrency.py` in a loop.

dbnicholson:
I think I got a handle on the worst races in the tmpdir allocation. I think I still see an error when I run my stress test for a long time. A more minor issue is libglnx stomping on errno even when it's ignoring errors.

dbnicholson:
Damn, still failed. I guess I'll keep looking at this.

dbnicholson:
Realized the last commit was leaking. Still looking for remaining races, though.

cgwalters:
Let's peel off those last 4 tmpdir commits to a new PR?

dbnicholson:
> Let's peel off those last 4 tmpdir commits to a new PR?

Alright. Opened #1346.

dbnicholson:
Fixed @pwithnall's comments. This still has the tmpdir fixes, but I'll rebase once #1346 lands.

dbnicholson:
Rebased now with the fixups to @pwithnall's comments still not squashed.

@cgwalters I think this is ready for review. I can keep banging on the tempdir races or remove/disable the concurrency test. All the errors I've ever seen are not from the locking.

pwithnall:
Those fixups address my comments nicely. As far as I’m concerned, this can get merged.

Makefile-tests.am
```diff
@@ -35,6 +35,7 @@ AM_TESTS_ENVIRONMENT += OT_TESTS_DEBUG=1 \
 	LD_LIBRARY_PATH=$$(cd $(top_builddir)/.libs && pwd)$${LD_LIBRARY_PATH:+:$${LD_LIBRARY_PATH}} \
 	PATH=$$(cd $(top_builddir)/tests && pwd):$${PATH} \
 	OSTREE_FEATURES="$(OSTREE_FEATURES)" \
+	PYTHONUNBUFFERED=1 \
```
cgwalters: Oh man that problem...

src/libostree/ostree-repo-prune.c
```diff
@@ -160,12 +161,20 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
  * Prune static deltas, if COMMIT is specified then delete static delta files only
  * targeting that commit; otherwise any static delta of non existing commits are
  * deleted.
+ *
+ * This function takes an exclusive lock on the @self repository.
```
cgwalters: These docs will show up in non-experimental builds but...eh, it's fine.

src/libostree/ostree-repo-prune.c
```diff
@@ -160,12 +161,20 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
...
  */
 gboolean
 ostree_repo_prune_static_deltas (OstreeRepo *self, const char *commit,
                                  GCancellable      *cancellable,
                                  GError           **error)
 {
+  g_autoptr(OstreeRepoAutoLock) lock = NULL;
+  lock = ostree_repo_auto_lock_push (self, OSTREE_REPO_LOCK_EXCLUSIVE,
```
cgwalters: In new code we usually do declare-and-init.

src/libostree/ostree-repo.c
```diff
@@ -156,6 +156,532 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
+      lock_table = g_hash_table_new_full (NULL, NULL, NULL,
+                                          (GDestroyNotify)free_repo_lock);
+      g_private_set (&repo_lock_table, lock_table);
+    }
+
+  OstreeRepoLock *lock = g_hash_table_lookup (lock_table, self);
+  if (lock == NULL)
+    {
+      lock = g_new0 (OstreeRepoLock, 1);
+      lock->stack = g_queue_new ();
```
cgwalters: Very minor but we could also inline the queue and `g_queue_init()` right?

src/libostree/ostree-repo-prune.c
dbnicholson: Well, the lock is still taken when the build is not experimental. Only the APIs aren't exposed on non-experimental builds.

src/libostree/ostree-repo-prune.c
dbnicholson: OK. Previously @pwithnall had preferred the lock to be an out parameter, but I kind of like this way. What do you think?

src/libostree/ostree-repo.c
dbnicholson: Hmm, hadn't thought about that. I'll look at it.

dbnicholson: That worked nicely, thanks.

src/libostree/ostree-repo-prune.c
dbnicholson: Left the interface the way it was and just did it in one line like you say.

dbnicholson:
Added a couple more fixups for your comments.

cgwalters:
One other high level concern; we now have two locks, the sysroot and repo.  Let's document somewhere that the locking order is sysroot, then repo?  AFAICS for the ostree cmdline we always grab the sysroot lock before doing anything, and in rpm-ostree we grab it quite early too.

cgwalters:
We can do some fixups like that to do the docs etc after.   Thanks a lot for working on this!  It looks good to me, but I'd like to play some more with it in master.

Looks like there might be concurrency issues in the tests, we'll see.

@rh-atomic-bot r+ 3210ba3

dbnicholson:
@cgwalters Thanks for looking at this! Indeed, the sysroot lock is separate and they should be documented as separate. I'll keep poking at the concurrency issues. I'm pretty sure it's still somewhere in the tmpdir allocation, but I can no longer reproduce the issue on my laptop. I think I'll need to play with a fast machine.

cgwalters:
I reproduced issues pretty quickly by spinning up a few `while true; do sleep 0.1; done&` busyloops along with `while make check TESTS=tests/test-concurrency.py; do : ; done`.  Looking...

cgwalters:
I'm a bit uncertain about timing out after 30s by default.  Feels like this is probably a case where we may want different behavior if the repo is client vs server side at least?  

It'd maximize compatibility if we didn't lock by default; make it easy to opt-in compatibly with an environment variable too?  I have some worries about doing this for NFS repos at least.

On the client side, the sysroot lock effectively covers the repo already, right?  I guess though an issue is people directly using the `ostree` cmdline on the system repo, but I'm not sure how often that happens.

pwithnall:
> On the client side, the sysroot lock effectively covers the repo already, right? I guess though an issue is people directly using the `ostree` cmdline on the system repo, but I'm not sure how often that happens.

The main reason we need this is for the client side, to add the locking between OSTree (for the OS, used within eos-updater) and flatpak, since on Endless OS they currently share the same repo (`/ostree/repo`). flatpak doesn’t use the sysroot lock at all (and shouldn’t).

dbnicholson:
The single global 30 seconds is not ideal, but I couldn't come up with any better default. Let me if you'd like to see something shorter or longer. In my more fully featured PR, the locking (both enablement and timeout) would be able to be controlled by both admins and applications. I'm not sure what heuristic you could apply to determine that a repo is for a client or server system. I suppose you could apply a different heuristic for a sysroot repo.

That said, I think most people in the real world would never hit these timeouts. They interact with applications that manage access to a specific repo coherently. I.e., rpm-ostree and flatpak already have systems that ensure serialized access to shared repositories. In that case, the repo locking isn't much more than a noop since there's never any contention.

I suppose you could make the locking opt-in, but my experience is that you do want it on by default. We've had a never ending stream of systems that can't update because they have missing objects. Mostly this is self inflicted because we joined the OS and flatpak repos, but it's not as if the issue this is fixing is a corner case. There's nothing that prevents 2 ostree applications from running concurrently and screwing each other. We can make this default off upstream if you prefer, though.

The sysroot lock is great if you're only talking about sysroot applications. Doesn't cover ostree or flatpak generally (i.e., gnome-software vs flatpak CLI) and certainly not the Endless grand unified repo :)

cgwalters:
server side = archive repo right?

dbnicholson:
Yeah, we have all archive repos on the server.

cgwalters:
Split off some prep work here https://github.com/ostreedev/ostree/pull/1352

cgwalters:
I think i'd feel more comfortable with this in the short term if locking is opt in; something like `env OSTREE_EX_REPO_LOCKING=1 ostree...`?  The flatpak/ostree-as-host cases are one thing, but we also need to consider archive repos and also "build cache" repos (e.g. `bare-user` repos like I use for staging ostree-as-host builds).

In practice locking makes sense for all of those...I just want to get some more experience with this before we turn it on by default.

cgwalters:
Rebased this.  I'd like to cut 2017.14 though on Monday, and land this and some other PRs immediately after.

cgwalters:
Lock and load!

@rh-atomic-bot r+ f2570ec

PR #1344, opened 2017-11-15
bin: Port a few commands (diff,remote,static-delta) to new style

cgwalters:
No functional changes, not prep for anything, just keeping up some momentum.


jlebon:
@rh-atomic-bot r+ 5b82295

PR #1345, opened 2017-11-16
build: fix "executible" typo

wjt:
Spotted this while cargo-culting system executable definitions for another project's build system!


jlebon:
Thanks! @rh-atomic-bot r+ db3af4c

PR #1346, opened 2017-11-17
Fix some tmpdir races

dbnicholson:
Running the `test-concurrency.py` script from #1343 shows some races with processes trying to reuse existing tmpdirs. I think there might still be a race lingering, but this makes my stress test run for much longer.


cgwalters:
@rh-atomic-bot r+ c51d93d

PR #1347, opened 2017-11-20
Check commit ref bindings with ostree fsck

pwithnall:
If constructing complex repositories with many refs (in our particular case, for Endless OS), it’s possible to accidentally omit some ref bindings from some commits, or add too many ref bindings to others. `ostree fsck` seems like the right place to catch that kind of error, so add support for checking commit binding metadata there.

The branch checks some things unconditionally (the things which would cause `ostree pull` to error out), and hides other checks behind a new `--verify-back-refs` option which instructs `ostree fsck` to check that each ref listed in a commit’s ref bindings points to that commit (and not to another commit which also lists that ref in its ref bindings).

@dbnicholson might want to take a look at this if he gets time.

### DIFF
--- a/src/ostree/ot-builtin-fsck.c
+++ b/src/ostree/ot-builtin-fsck.c
@@ -189,9 +189,16 @@ ostree_builtin_fsck (int argc, char **argv, OstreeCommandInvocation *invocation,
   gpointer key, value;
   g_hash_table_iter_init (&hash_iter, all_refs);
   while (g_hash_table_iter_next (&hash_iter, &key, &value))
-    if (!fsck_commit_for_ref (repo, value, NULL, key,
-                              &found_corruption, cancellable, error))
-      return FALSE;
+    {
+      const char *refspec = key;
+      const char *checksum = value;
+      g_autofree char *ref_name = NULL;
+      if (!ostree_parse_refspec (refspec, NULL, &ref_name, error))
+        return FALSE;
+      if (!fsck_commit_for_ref (repo, checksum, NULL, ref_name,
+                                &found_corruption, cancellable, error))
+        return FALSE;
+    }
 
 #ifdef OSTREE_ENABLE_EXPERIMENTAL_API
   if (!opt_quiet)
--- a/src/ostree/ot-builtin-fsck.c
+++ b/src/ostree/ot-builtin-fsck.c
@@ -30,6 +30,7 @@
 static gboolean opt_quiet;
 static gboolean opt_delete;
 static gboolean opt_add_tombstones;
+static gboolean opt_verify_back_refs;
 
 /* ATTENTION:
  * Please remember to update the bash-completion script (bash/ostree) and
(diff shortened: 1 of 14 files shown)


pwithnall:
Apparently I forgot to run the entire test suite against this, and some bugs have happened. Oops. I’ll take a look.

pwithnall:
Updated: the code was not stripping the remote off a refspec, so was looking for the refspec rather than the ref name in the commit bindings.

pwithnall:
Ugh, 73d910e82eb9076c3551eb9323d55d0a9c768e22 has made this a bit tricky to rebase. Let’s see what I can do. :-(

pwithnall:
I’ve rebased and pushed. Not sure why one of the tests is failing with
```
./.libs/libostree-1.so: undefined reference to `_ostree_repo_verify_bindings'
```

It all compiles and passes tests fine for me locally (with and without experimental API enabled).

jlebon:
Hmm, likely due to `--without-curl --without-soup`. You probably need to also encase the private table entry with `HAVE_LIBCURL_OR_LIBSOUP`.

pwithnall:
Ah, I didn’t notice the `#ifdef` around the entire file. The verification function should really be defined regardless of whether curl or libsoup are available, since it doesn’t use either of them, and can be used even when not pulling. Potentially, this means it should be in a file other than `ostree-repo-pull.c`; if so, let me know and I’ll move it.

src/ostree/ot-builtin-fsck.c
```diff
@@ -253,6 +255,77 @@ ostree_builtin_fsck (int argc, char **argv, OstreeCommandInvocation *invocation,
...
+                          if (collection_id != NULL)
+                            return glnx_throw (error,
+                                               "Collection–ref (%s, %s) in bindings for commit %s does not exist",
+                                               collection_id, *iter, checksum);
+                          else
+                            return glnx_throw (error,
+                                               "Ref ‘%s’ in bindings for commit %s does not exist",
+                                               *iter, checksum);
+                        }
+                      else if (g_strcmp0 (checksum_for_ref, checksum) != 0)
```
jlebon: Minor: I prefer using `g_str_equal` when possible to signal the reader that the strings at this point *can't* be `NULL`. From that view, using `g_strcmp0` is implying something that doesn't seem true here.

cgwalters:
> Ugh, 73d910e has made this a bit tricky to rebase. Let’s see what I can do. :-(

Ah sorry about that!  I had only been thinking about the locking PR as outstanding.

Overall looks great to me; do you want to do those two tweaks @jlebon suggested?

cgwalters:
@rh-atomic-bot r+ 64587ac

cgwalters:
Some fallout in https://github.com/projectatomic/rpm-ostree/pull/1148#issuecomment-351761632

cgwalters:
https://github.com/ostreedev/ostree/pull/1379

PR #1348, opened 2017-11-26
add back helpful --allow-downgrade err message

dustymabe:


dustymabe:
unfortunately not tested locally since it was simple changing of string. if tests fail i'll investigate

cgwalters:
So this is tricky; the problem here is that libostree is a library, so we can't know if clients use the same option name.  Further, mentioning command line strings isn't useful for GUIs like Cockpit and gnome-software.   Doing this "right" involves creating an error domain and having clients react to that error domain.

That said I'm not sure I care a lot about doing it "right" for this particular case; I suspect most people hitting this will be doing it from the command line.  And further it's kind of an "TLS failure" type thing - in the GUI case you probably want to hide a "continue anyways" option behind lots of warnings.

@rh-atomic-bot r+ b487f20

dustymabe:
i see what you mean about it being tricky. thanks for the explanation.

Issue #878, opened 2017-05-19
add some retry logic for pulling from remotes


dustymabe:
There has been a lot of intermittent network failures lately when pulling from our ostree remotes. While much of this is probably related to the backend fedora infra, i'd like to at least brainstorm how we could make our code more robust to poor networks and/or poor remote servers. 

here are a few examples, but i've seen it myself quite a bit:

- [this screenshot](https://kojipkgs.fedoraproject.org//work/tasks/1921/19621921/screenshot.ppm) from within [this task](https://koji.fedoraproject.org/koji/taskinfo?taskID=19621921)

- [this log](https://kojipkgs.fedoraproject.org//work/tasks/5421/19625421/runroot.log) from within [this task](https://koji.fedoraproject.org/koji/taskinfo?taskID=19625421)

<details>

```
error: Socket I/O timed out

command output:
error: Socket I/O timed out

2017-05-19 16:55:42,082: command returned failure (1)
command returned failure (1)
2017-05-19 16:55:42,082: template command error in /mnt/koji/compose/branched/Fedora-26-20170519.n.0/work/x86_64/Atomic/lorax_templates/atomic-installer/lorax-embed-repo.tmpl:
template command error in /mnt/koji/compose/branched/Fedora-26-20170519.n.0/work/x86_64/Atomic/lorax_templates/atomic-installer/lorax-embed-repo.tmpl:
2017-05-19 16:55:42,083:   runcmd ostree --repo=/var/tmp/lorax.vgsxgpyx/installtree/ostree/repo pull --mirror fedora-atomic fedora/26/x86_64/atomic-host
  runcmd ostree --repo=/var/tmp/lorax.vgsxgpyx/installtree/ostree/repo pull --mirror fedora-atomic fedora/26/x86_64/atomic-host
2017-05-19 16:55:42,086:   subprocess.CalledProcessError: Command '['ostree', '--repo=/var/tmp/lorax.vgsxgpyx/installtree/ostree/repo', 'pull', '--mirror', 'fedora-atomic', 'fedora/26/x86_64/atomic-host']' returned non-zero exit status 1.
  subprocess.CalledProcessError: Command '['ostree', '--repo=/var/tmp/lorax.vgsxgpyx/installtree/ostree/repo', 'pull', '--mirror', 'fedora-atomic', 'fedora/26/x86_64/atomic-host']' returned non-zero exit status 1.
Traceback (most recent call last):
  File "/usr/sbin/lorax", line 273, in <module>
[...]

PR #1349, opened 2017-11-27
fetcher/curl: Stop using CURLOPT_LOW_SPEED_TIME/_LIMIT

cgwalters:
They don't play nicely currently with HTTP2 where we may
have lots of requests queued.

https://github.com/ostreedev/ostree/issues/878#issuecomment-347228854

In practice anyways I think issues here are better solved on a higher level -
e.g. apps today can use an overall timeout on pulls and if they exceed the limit
set the cancellable.

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -788,8 +788,13 @@ initiate_next_curl_request (FetcherRequest *req,
   curl_easy_setopt (req->easy, CURLOPT_PROGRESSFUNCTION, prog_cb);
   curl_easy_setopt (req->easy, CURLOPT_FOLLOWLOCATION, 1L);
   curl_easy_setopt (req->easy, CURLOPT_CONNECTTIMEOUT, 30L);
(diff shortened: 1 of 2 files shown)


alexlarsson:
@rh-atomic-bot r+ e89d37e

Looks good to me.

jlebon:
Let's see if the logs have more details now.
@rh-atomic-bot retry

cgwalters:
Hm, I can't reproduce the failure locally.  Weird.

jlebon:
So, this turned out to be a subtle interaction between libcurl and our docker testing environment. For reasons I don't quite understand, doing a `curl` to `http://example.com/invalid/subpath` will hit the docker-on-Openstack path MTU issue (https://bugzilla.redhat.com/show_bug.cgi?id=1475460). I had previously never seen it triggered on external websites. This does mean we should probably update the hosts to use e.g. `--mtu 1400`, as was originally done in https://github.com/projectatomic/papr/pull/54. I'll update the Ansible playbooks for that. The `LOW_SPEED` opts were allowing us to break out of that hang.

For now, I just pushed a fixup to use a different bad repo URL. ⬆️
@rh-atomic-bot r+ 0eb0484

PR #1350, opened 2017-11-27
ci: Make sure we save gtdr test results on failures

jlebon:
If a test fails, we immediately exit and thus never get a chance to
actually upload the test results. Add a trap so that they always
uploaded, even on failure.


cgwalters:
@rh-atomic-bot r+ 78f7c12

PR #1351, opened 2017-11-28
rofiles-fuse: Fix utime() support

alexlarsson:
We use utimens instead of utime, thus allowing nanosecond timestamps,
and also fixes a bug where we used to passed UTIME_OMIT to tv_nsec
which made the entire operation a no-op.


cgwalters:
Thanks, it's about time we fixed this!

@rh-atomic-bot r+ 93fc0ff

PR #1353, opened 2017-11-28
lib/remote: Add a method to return the URL

joaquimrocha:
When using dynamic remotes (LAN and USB), we cannot use their name with
the common remote related ops (ostree_repo_remote_...) because ostree
doesn't keep this type of remotes in its internal hash table.
Unfortunately this means that we cannot access the URL of those remotes
either (in order to e.g. set the right URL for those remotes in
Flatpak).

Since the URL is actually stored in a key file that belongs to the
OstreeRemote, then we can simply allow users access to it through a
getter.

So this patch adds a method that allows to return the URL directly from
the OstreeRemote without having to go through the OstreeRepo.

The test-repo-finder-config is also updated by this patch to check if
the URL is correct.

### DIFF
--- a/apidoc/ostree-experimental-sections.txt
+++ b/apidoc/ostree-experimental-sections.txt
@@ -19,6 +19,7 @@ OstreeRemote
 ostree_remote_ref
 ostree_remote_unref
 ostree_remote_get_name
+ostree_remote_get_url
 <SUBSECTION Standard>
 ostree_remote_get_type
 </SECTION>
--- a/src/libostree/libostree-experimental.sym
+++ b/src/libostree/libostree-experimental.sym
@@ -93,4 +93,5 @@ global:
 LIBOSTREE_2017.14_EXPERIMENTAL {
 global:
   ostree_remote_get_type;
+  ostree_remote_get_url;
 } LIBOSTREE_2017.13_EXPERIMENTAL;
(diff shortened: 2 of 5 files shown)


pwithnall:
bot, add author to whitelist

cgwalters:
bot, add author to whitelist

cgwalters:
(The jenkins PR tester whitelist is a separate thing...it's a mess)

cgwalters:
Mmm...not sure this one needs to be experimental, right?

cgwalters:
(I mean I have no opposition to it being experimental, but I also don't see a reason why not to mark it for stable)

joaquimrocha:
Hi! I mostly added it there because that's where the ostree_remote_*
methods are. How do we proceed? Should I move the ostree_remote_get_name
the and then change this patch to use stable there?

On Nov 28, 2017 19:34, "Colin Walters" <notifications@github.com> wrote:

(I mean I have no opposition to it being experimental, but I also don't see
a reason why not to mark it for stable)

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ostreedev/ostree/pull/1353#issuecomment-347620465>, or mute
the thread

cgwalters:
Oops yes, I was wrong.

@rh-atomic-bot r+ 76238d1

PR #1354, opened 2017-11-28
lib/commit: Use direct repo writes if fsync is disabled

cgwalters:
For situations where fsync is disabled, there's basically
no reason to do the whole "staging directory" dance.  Just
write directly into the repo.

Today I use `fsync=false` for my build/cache repos.

I briefly considered not allocating a tmpdir at all
in this case, but we actually do want the txn tmpdir
for the non-`O_TMPFILE` case.

Part of https://github.com/ostreedev/ostree/issues/1184


giuseppe:
@rh-atomic-bot r+ 75feeaa

PR #1356, opened 2017-12-01
tests/delta-crosscheck: Disable fsync

cgwalters:
I was running this recently to test the last delta write changes, and this
helps. We should add an option to repo-init to make this easier at some point.


jlebon:
@rh-atomic-bot r+ 30ce650

PR #1358, opened 2017-12-01
lib/repo: Add MT support for transaction_set_ref(), clarify MT rules

cgwalters:
For rpm-ostree I'd like to do importing in parallel with threads; the
code is *almost* ready for that except today it calls
`ostree_repo_transaction_set_ref()`.

Looking at the code, there's really a "transaction" struct here,
not just stats.  Let's lift that struct out, and move the refs
into it under the existing lock.

Clarify the documentation around multithreading for various functions.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -658,19 +658,19 @@ write_content_object (OstreeRepo         *self,
   /* Free space check; only applies during transactions */
   if (self->min_free_space_percent > 0 && self->in_transaction)
     {
-      g_mutex_lock (&self->txn_stats_lock);
-      g_assert_cmpint (self->txn_blocksize, >, 0);
-      const fsblkcnt_t object_blocks = (size / self->txn_blocksize) + 1;
-      if (object_blocks > self->max_txn_blocks)
+      g_mutex_lock (&self->txn_lock);
+      g_assert_cmpint (self->txn.blocksize, >, 0);
+      const fsblkcnt_t object_blocks = (size / self->txn.blocksize) + 1;
+      if (object_blocks > self->txn.max_blocks)
         {
-          g_mutex_unlock (&self->txn_stats_lock);
-          g_autofree char *formatted_required = g_format_size ((guint64)object_blocks * self->txn_blocksize);
+          g_mutex_unlock (&self->txn_lock);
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -1592,29 +1604,39 @@ ostree_repo_transaction_set_refspec (OstreeRepo *self,
...
  *
- * The change will not be written out immediately, but when the transaction
- * is completed with ostree_repo_commit_transaction(). If the transaction
- * is instead aborted with ostree_repo_abort_transaction(), no changes will
- * be made to the repository.
+ * The change will be written when the transaction is completed with
+ * ostree_repo_commit_transaction(); that function takes care of writing all of
+ * the objects (such as the commit referred to by @checksum) before updating the
+ * refs. If the transaction is instead aborted with
+ * ostree_repo_abort_transaction(), no changes to the ref be made to the
```
jlebon: `s/be/will be/`  ?

src/libostree/ostree-repo-commit.c
```diff
@@ -1592,29 +1604,39 @@ ostree_repo_transaction_set_refspec (OstreeRepo *self,
...
+ * repository.
+ *
+ * Note however that currently writing *multiple* refs is not truly atomic; if
+ * the process or system is terminated during
+ * `ostree_repo_commit_transaction()`, it is possible that just some of the refs
```
jlebon: Minor: seems like the convention is to not backtick-escape function names in documentation, right?

src/libostree/ostree-repo-commit.c
```diff
@@ -1573,9 +1584,10 @@ ostree_repo_transaction_set_refspec (OstreeRepo *self,
 {
```
jlebon: Missing multithreading advisory for this one?

src/libostree/ostree-repo-commit.c
cgwalters: gtk-doc should support both, but yeah let's be consistent in not doing it for now.

cgwalters:
This one can wait until after https://github.com/ostreedev/ostree/pull/1359

cgwalters:
OK release done, see fixup :arrow_up: above

jlebon:
@rh-atomic-bot r+ 37d1217

PR #1359, opened 2017-12-04
Release 2017.14

cgwalters:
Time to cut a new release, we've got the libcurl cleanup ordering patch which
several people have hit, along with safe early fixes for tmpdir cleanup. Let's
try to land the locking PR early next cycle.


jlebon:
@rh-atomic-bot r+ 40b562f

PR #1363, opened 2017-12-05
lib/repo-file: Add casts to appease GLib g_object_ref cast PR

cgwalters:
This fixes the build with https://bugzilla.gnome.org/show_bug.cgi?id=790697


jlebon:
@rh-atomic-bot retry

PR #1365, opened 2017-12-05
lib/commit: Refactor file commits to separate subdir from content

cgwalters:
One major thing we can do to speed up local commits is multithreading. In
preparation for that, split up the recursion function so that the subdirectory
case is separate from the content (regfile/symlink) case. Then for non-subdirs,
we can easily peel off worker threads and gather the final checksums and update
the mtree from the main thread.

The diff here looks large but it's pretty straightforward; amazingly this change
compiled the very first time I tried it!


jlebon:
That patch was surprisingly hard to read for the amount of new code, but it looks sane!
@rh-atomic-bot r+ 9bae308

jlebon:
@rh-atomic-bot retry

PR #1366, opened 2017-12-05
lib/repo: Properly list remotes of parent repos

mwleeds:
This commit fixes an infinite loop that happens if you try to list the
remotes of a repo that has a parent repo set. It also adds a unit test
to ensure the right behavior, which is that both the child remotes and
parent remotes are listed.


cgwalters:
Nice, this is ∞ better!  And yeah...we don't have a lot of code exercising parent repos unfortunately 🙁  Anyways will `r+` after our CI comes back.

giuseppe:
bot, retest this please

cgwalters:
@rh-atomic-bot r+ 4815e7b

PR #1367, opened 2017-12-06
Modification to build instructions

knownexus:
I noticed, when attempting to build the latest version of OSTree, that my system would not recognise that it was installed.

This modification seems to resolve that issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,15 +88,35 @@ available online;
 [this code](https://github.com/ostreedev/ostree/blob/master/packaging/Makefile.dist-packaging#L11)
 in OSTree is an example.
 
-Once you have a git clone or recursive archive, building is the
-same as almost every autotools project:
+Once you have a git clone or recursive archive,
+run the following commands to complete the installation process:
 
 ```
-git submodule update --init
-env NOCONFIGURE=1 ./autogen.sh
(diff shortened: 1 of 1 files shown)


cgwalters:
Hmm.  I see that we should have docs for this, but personally I use a "dev" container for most of my work so the `/opt` thing doesn't apply to me.   Can you squash this to one commit at least?

I may also want to move this out of `README.md` into some `README-building.md` guide.

knownexus:
Ok, the commits should be squashed up now

cgwalters:
Hi, sorry about the delay here.  If you don't mind doing a bit more work on this...how about we link to existing well-maintained `HACKING` type files?  I briefly looked and found for example:

 - https://github.com/systemd/systemd/blob/master/HACKING
 - https://github.com/coreutils/coreutils/blob/master/HACKING
 - https://git.gnome.org/browse/glib/tree/HACKING

Obviously not all of the instructions here apply, and there are some ostree-specific things, but there's a *lot* we have in common with other autotools projects, etc.

lucab:
Closing this stale PR. We now have a contributing tutorial at https://ostreedev.github.io/ostree/contributing-tutorial/.

Issue #1362, opened 2017-12-05
ostree pull hangs in low bandwidth environments


jlebon:
Using the libcurl backend on a machine with low bandwidth (< 500 kB/s here), it seems pretty easy to get ostree to hang during a pull operation. I've reproduced this in two separate environments. The issue seems present in git master/latest release/previous release. Disabling pipelining works around the issue.

Once it apparently hangs, `top` shows CPU usage goes to 100%. Using `strace` and `G_MESSAGES_DEBUG=all` shows that it's not actually waiting on a specific resource. In fact, data does come in, but it seems like only 1 of the 8 open sockets are attended to, while the `POLLIN` for the other sockets are completely ignored yet continuously polled.

```
...
poll([{fd=7, events=POLLIN}, {fd=8, events=POLLIN}, {fd=16, events=POLLIN}, {fd=17, events=POLLIN}, {fd=18, events=POLLIN}, {fd=19, events=POLLIN}, {fd=20, events=POLLIN}, {fd=21, events=POLLIN}, {fd=22, events=POLLIN}], 9, 0) = 7 ([{fd=8, revents=POLLIN}, {fd=16, revents=POLLIN}, {fd=17, revents=POLLIN}, {fd=18, revents=POLLIN}, {fd=20, revents=POLLIN}, {fd=21, revents=POLLIN}, {fd=22, revents=POLLIN}])
poll([{fd=7, events=POLLIN}, {fd=8, events=POLLIN}, {fd=16, events=POLLIN}, {fd=17, events=POLLIN}, {fd=18, events=POLLIN}, {fd=19, events=POLLIN}, {fd=20, events=POLLIN}, {fd=21, events=POLLIN}, {fd=22, events=POLLIN}], 9, 0) = 7 ([{fd=8, revents=POLLIN}, {fd=16, revents=POLLIN}, {fd=17, revents=POLLIN}, {fd=18, revents=POLLIN}, {fd=20, revents=POLLIN}, {fd=21, revents=POLLIN}, {fd=22, revents=POLLIN}])
[...]

PR #1368, opened 2017-12-06
pull: Add http2=false remote config option

cgwalters:
This seems to work around
https://github.com/ostreedev/ostree/issues/1362

Though I'm not entirely sure why yet. But at least with this it'll be easier for
people to work around things locally.


jlebon:
@rh-atomic-bot r+ 00af54f

PR #1370, opened 2017-12-08
lib/commit: Fix memleak in bare-user devino hit path

cgwalters:
I noticed this while chasing an entirely different issue:
https://github.com/projectatomic/rpm-ostree/pull/1139


jlebon:
LGTM.

cgwalters:
@rh-atomic-bot r=jlebon 8f62650

PR #1373, opened 2017-12-12
build-sys: Add --disable-http2

cgwalters:
I may punt and use this in Fedora at least for now until we have time to debug
the issues.


jlebon:
@rh-atomic-bot r+ c8ef917

PR #1375, opened 2017-12-13
Locking followup

cgwalters:


src/libostree/ostree-repo.c
```diff
@@ -2742,6 +2749,15 @@ reload_core_config (OstreeRepo          *self,
     self->lock_timeout_seconds = g_ascii_strtoull (lock_timeout_seconds, NULL, 10);
   }
 
+  /* Disable locking by default for now */
+  { gboolean locking;
+    if (!ot_keyfile_get_boolean_with_default (self->config, "core", "locking",
+                                              FALSE, &locking, error))
+      return FALSE;
+    if (!locking)
+      self->lock_timeout_seconds = REPO_LOCK_DISABLED;
```
jlebon: Let's drop the previous block and here do

```c
else
  {
    <parse lock-timeout-secs>
  }
```

?

cgwalters: Reworked, accidentally did `git commit --amend` instead of `fixup` though :arrow_down:

PR #1376, opened 2017-12-14
lib/repo: Add an API to mark a commit as partial

cgwalters:
For the [rpm-ostree jigdo ♲📦](https://github.com/projectatomic/rpm-ostree/issues/1081) work.
We're basically doing "pull" via a non-libostree mechanism, and this
should be fully supported.  As I mentioned earlier we should try to
have `ostree-repo-pull.c` only use public APIs; this gets us closer
to that.


jlebon:
@rh-atomic-bot r+ 326ade5

Issue #1377, opened 2017-12-14
rofiles-fuse: Implement copy up


cgwalters:
Sadly https://sourceware.org/bugzilla/show_bug.cgi?id=22089 is I think going to actually force us to cave here.  See also discussion [about fish](https://bugzilla.redhat.com/show_bug.cgi?id=1415451) as part of the general [Fedora tracker](https://bugzilla.redhat.com/show_bug.cgi?id=1352154).

PR #1378, opened 2017-12-14
lib/core: Add a "break hardlink" API

cgwalters:
This imports the code from rpm-ostree:
https://github.com/projectatomic/rpm-ostree/blob/9ff9f6c997d914cb7d97d6b59d8045ba64a1882c/src/libpriv/rpmostree-util.c#L742

I plan to use this for rofiles-fuse to implement
copyup: https://github.com/ostreedev/ostree/issues/1377

But it's just obviously generally useful for projects using
libostree I think.


cgwalters:
Tests are passing on this one now.

jlebon:
Ahh, that's a nice use of `O_TMPFILE`.
@rh-atomic-bot r+ e4c32ef

PR #1379, opened 2017-12-14
bin/fsck: Make ref binding verification optional

cgwalters:
Today the rpm-ostree test suite uses `refs --create` to save
commits.  I think this is a legitimate use case, and other
people may be doing something similar.

On the other hand, I think we should probably be changing the rpm-ostree test
suite to create "unbound" commits. But let's be maximially compatible here since
we hit a real-world case where something needed to change.

### DIFF
--- a/bash/ostree
+++ b/bash/ostree
@@ -963,6 +963,7 @@ _ostree_fsck() {
         --add-tombstones
         --delete
         --quiet -q
+        --verify-bindings
         --verify-back-refs
     "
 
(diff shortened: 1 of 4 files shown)


cgwalters:
Hm yeah I've been forgetting to do the man pages and bash completion a fair amount.  Ugh...

cgwalters:
Fixup :arrow_up:

pwithnall:
@rh-atomic-bot r+ cd91224

PR #1380, opened 2017-12-14
bin/commit: Support creating "unbound" commits

cgwalters:
We had this basically forced on in the CLI; down the line I'd really like to
make this an API option to commit or so, but given that we found a use case in
the rpm-ostree test suite for "unbound" commits, let's support creating them
from the cmdline.

See: https://github.com/ostreedev/ostree/pull/1379


jlebon:
@rh-atomic-bot r+ 259968d

Issue #1342, opened 2017-11-14
Alias refs don’t work when created pointing to a remote ref


pwithnall:
@dbnicholson pointed this out in a chat today; we haven’t had time to reproduce it properly.

Something along the lines of:
```
ostree refs --alias  --create=new-alias remote1:old-ref
```
doesn’t work.

cgwalters:
Given that remote names are freeform data, I'm not sure about supporting this.

pwithnall:
What do you mean? I was assuming it would either be:
 * supported as a symlink to `refs/remotes/$remote_name/$ref_name`; or
 * explicitly not supported (`ostree refs` returns an error message).

cgwalters:
I'm thinking "explicitly not supported".

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1381

pwithnall:
Should this be mentioned in the documentation somewhere?

PR #1381, opened 2017-12-14
bin/refs: Disallow aliases to remote refs

cgwalters:
It can't really work in general; the client and server would
have to agree on the name of the remote.

Closes: https://github.com/ostreedev/ostree/issues/1342


jlebon:
@rh-atomic-bot r+ f7d8e48

PR #1382, opened 2017-12-14
 rofiles: Add --copyup option

cgwalters:
Sadly https://sourceware.org/bugzilla/show_bug.cgi?id=22089 is I think going to
actually force us to cave here. Even if we got the glibc patch in today, we need
to support the RHEL glibc. See also discussion about fish as part of the general
Fedora tracker.

This is basically needed to unblock rpm-ostree unified core :globe_with_meridians::
projectatomic/rpm-ostree#729

Closes: #1377

### DIFF
--- a/src/rofiles-fuse/main.c
+++ b/src/rofiles-fuse/main.c
@@ -394,14 +394,18 @@ do_open (const char *path, mode_t mode, struct fuse_file_info *finfo)
           if (fd == -1)
             return -errno;
         }
-
-      /* Handle O_TRUNC here only after verifying hardlink state */
-      if (finfo->flags & O_TRUNC)
+      else
         {
-          if (ftruncate (fd, 0) == -1)
+          /* In the non-copyup case we handle O_TRUNC here, after we've verified
+           * the hardlink state above with verify_write_or_copyup().
+           */
+          if (finfo->flags & O_TRUNC)
             {
-              (void) close (fd);
-              return -errno;
+              if (ftruncate (fd, 0) == -1)
+                {
+                  (void) close (fd);
+                  return -errno;
+                }
             }
         }
     }
--- a/src/rofiles-fuse/main.c
+++ b/src/rofiles-fuse/main.c
@@ -35,9 +35,14 @@
 #include <glib.h>
 
 #include "libglnx.h"
+#include "ostree.h"
 
 // Global to store our read-write path
 static int basefd = -1;
+/* Whether or not to automatically "copyup" (in overlayfs terms).
+ * What we're really doing is breaking hardlinks.
+ */
+static gboolean opt_copyup;
 
 static inline const char *
 ENSURE_RELPATH (const char *path)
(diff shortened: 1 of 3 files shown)


cgwalters:
Depends: https://github.com/ostreedev/ostree/pull/1378

src/rofiles-fuse/main.c
```diff
@@ -286,6 +333,9 @@ callback_truncate (const char *path, off_t size)
 static int
 callback_utimens (const char *path, const struct timespec tv[2])
 {
+  /* This one isn't write-verified, we support changing times
+   * even for hardlinked files.
```
jlebon: Can you expand on this? Don't we need to keep bare objects at `OSTREE_TIMESTAMP` to be consistent?

src/rofiles-fuse/main.c
```diff
@@ -324,13 +374,22 @@ do_open (const char *path, mode_t mode, struct fuse_file_info *finfo)
...
         {
           (void) close (fd);
           return r;
         }
 
+      /* In the copyup case, we need to re-open */
+      if (opt_copyup)
+        {
+          (void) close (fd);
+          fd = openat (basefd, path, finfo->flags & ~O_TRUNC, mode);
```
jlebon: Minor: we can just stop filtering out `O_TRUNC` in the copy-up case, right? And only pay the `ftruncate` call in the non-copy-up case.

src/rofiles-fuse/main.c
cgwalters: It's an interesting point...gets a bit into a philosophical question of whether rofiles-fuse is for "anything using hardlinks" or whether it's really specific to ostree.  The "anything using hardlinks" case might be a backup system or non-os-data object store was my thought originally.

For example I could probably have "object semantics" for my `~/Documents` which has a bunch of random PDFs and `.txt` and LibreOffice files etc. that should really never be overwritten in place.  OTOH I've been thinking about storing all of that in git.

So...hmm.  It's not a new issue here and we should think about changing this, but I worry a bit about the backcompat issues.  In practice nothing we use rofiles-fuse for should be touching timestamps for files it didn't create at the time I would guess.

cgwalters:
Consumer: https://github.com/projectatomic/rpm-ostree/pull/1171

src/rofiles-fuse/main.c
dustymabe: > For example I could probably have "object semantics" for my ~/Documents which has a bunch of random PDFs and .txt and LibreOffice files etc. that should really never be overwritten in place. OTOH I've been thinking about storing all of that in git.

I've been using [git-annex](https://git-annex.branchable.com/) for this for some time now. Good news, it's in fedora.

cgwalters: Right, git-annex is likely an excellent example of a possible user of rofiles-fuse semantics, specifically [its direct mode](https://git-annex.branchable.com/direct_mode/).  Though I didn't look closely.

src/rofiles-fuse/main.c
cgwalters: BTW I wish I'd seen this comment while working on this patchset on the plane :airplane:  - it turned out mishandling `O_TRUNC` was the source of bugs.

src/rofiles-fuse/main.c
```diff
@@ -385,7 +385,12 @@ do_open (const char *path, mode_t mode, struct fuse_file_info *finfo)
       if (opt_copyup)
         {
           (void) close (fd);
-          fd = openat (basefd, path, finfo->flags & ~O_TRUNC, mode);
+          /* Note that unlike the initial open, we will pass through
+           * O_TRUNC.  More ideally in this copyup case we'd avoid copying
```
jlebon: Hmm, so then we don't need to `ftruncate` down below again, right? Shouldn't the `if` condition at 399 be changed to

```
if (!opt_copyup && (finfo->flags & O_TRUNC))
```

?

src/rofiles-fuse/main.c
```diff
@@ -385,7 +385,12 @@ do_open (const char *path, mode_t mode, struct fuse_file_info *finfo)
...
+           * the whole file in the first place, but eh.  It's not like we're
+           * high performance anyways.
+           */
+          fd = openat (basefd, path, finfo->flags & ~(O_EXCL|O_CREAT), mode);
```
jlebon: Why are we filtering out `O_EXCL` and `O_CREAT` here? This is something that should've been caught by the first `openat`, right? Or is this just to not have a TOCTOU race on the file existing?

cgwalters: We're internally reopening the file we just created.  So if the user provided `O_EXCL` we need to filter that out.  We might as well drop `O_CREAT` too.

jlebon:
@rh-atomic-bot r+ eec2ff2

PR #1383, opened 2017-12-15
Bump libglnx, use "n items" progress for fsck

cgwalters:
Sooo much nicer.  See also
https://github.com/projectatomic/rpm-ostree/pull/1143

Update submodule: libglnx


jlebon:
That's about 10 times better!
@rh-atomic-bot r+ 74b3348

PR #1384, opened 2017-12-15
build-sys: Use -fno-strict-aliasing by default

cgwalters:
See discussion in https://bugzilla.gnome.org/show_bug.cgi?id=791622

This is what e.g. systemd, the Linux kernel, and lots of other projects do. It's
astonishingly hard to reliably get right; the optimization IMO only really
matters for truly high performance inner loops, but if you're doing
that kind of stuff today you're probably doing it on a GPU anyways.


pwithnall:
@rh-atomic-bot r+ 235ca3de4e9e94119bfd6c782a0bf997bb14185a

cgwalters:
https://github.com/projectatomic/rpm-ostree/pull/1156

PR #1385, opened 2017-12-15
lib/pull: Add more information into journal from pull

cgwalters:
Things like which HTTP version are in use for obvious reasons.
I also want to log the TLS negotation summary.

Extra data that libcurl handliy provides like DNS lookup time
is injected into the journal as key/values separate from the log.


cgwalters:
Marking WIP since I want a bit more time to play with this.  A notable item is that we should also log on failure.

PR #1387, opened 2017-12-20

cgwalters:


jlebon:
@rh-atomic-bot r+ 733ec57

cgwalters:
@rh-atomic-bot retry

jlebon:
I looked into this one a bit earlier. I'm still not entirely sure exactly what's wrong with it, though I did reproduce the issue after upgrading my `rust` package to `rust-1.22.1-1.fc27.x86_64` (and verified that the issue does not reproduce in `rust-1.21.0-1.fc27`. The timeline matches up since 1.22 was just pushed to stable 2 days ago.

cgwalters:
OK, yeah it had to be something like the rust complier changing.  For now I pushed a change to make the context nonblocking :arrow_up:

jlebon:
@rh-atomic-bot r+ 1bbdf78

PR #1390, opened 2018-01-02
tests: Don't assume uid == primary gid

smcv:
Nothing guarantees that each user has a group containing only
themselves. Even if they do, nothing guarantees that its group ID
equals the user ID, particularly if another user earlier in the same
range was created without a corresponding group or vice versa.

---

Applied in Debian since 2017.14.


cgwalters:
Hm, from `test-concurrency.py` in Travis:
> error: opendir(staging-dc9bf54c-5955-4738-9a33-ecc0017fb577-jv0bfY): No such file or directory

May be a new race :running_man:.  Will look at that later.  Thanks for the patch!

@rh-atomic-bot r+ e3ed68b

PR #1391, opened 2018-01-02
build-sys: Link with -ldl for rust build

cgwalters:
I didn't dive into this too much, it looks like something in rust changed that
broke our build. Probably libstd gained a dependency on `-ldl` or so, and that's
handled by cargo? Anyways linking against it isn't going to hurt.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -53,7 +53,6 @@ tests:
 
 context: f26-rust
 inherit: true
-required: false
 container:
     image: registry.fedoraproject.org/fedora:26
 env:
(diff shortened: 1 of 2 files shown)


cgwalters:
https://github.com/ostreedev/ostree/pull/1387#issuecomment-353461297

smcv:
Looks good to me. (Do I count as a reviewer?)

@rh-atomic-bot r+ f55941b

cgwalters:
@rh-atomic-bot delegate=smcv

smcv:
@rh-atomic-bot r+ f55941b

Issue #1392, opened 2018-01-02
Tests fail on some (all?) big-endian CPUs


smcv:
ostree 2017.15 fails its build-time tests on Debian mips and s390x, and the semi-official hppa, powerpc, ppc64 and sparc64 ports: https://buildd.debian.org/status/package.php?p=ostree

I believe those are exactly the architectures that are big-endian and do not disable build-time tests (powerpcspe and m68k are also BE, but they don't run build-time tests, which IMO is deeply misguided but those ports' maintainers get to keep both pieces).

I think this is the failure:

```
++ env LD_PRELOAD=/<<PKGBUILDDIR>>/tests/libreaddir-rand.so ostree --repo=/var/tmp/tap-test.zh9sgb/repo commit --canonical-permissions -b test2 -s 'Metadata string' --add-metadata-string=FOO=BAR --add-metadata-string=KITTENS=CUTE --add-detached-metadata-string=SIGNATURE=HANCOCK '--add-metadata=SOMENUM=uint64 42' --tree=ref=test2
++ cd /var/tmp/tap-test.zh9sgb
++ env LD_PRELOAD=/<<PKGBUILDDIR>>/tests/libreaddir-rand.so ostree --repo=/var/tmp/tap-test.zh9sgb/repo show --print-metadata-key=FOO test2
++ assert_file_has_content test2-meta BAR
++ fpath=test2-meta
++ shift
++ for re in "$@"
++ grep -q -e BAR test2-meta
++ env LD_PRELOAD=/<<PKGBUILDDIR>>/tests/libreaddir-rand.so ostree --repo=/var/tmp/tap-test.zh9sgb/repo show --print-metadata-key=KITTENS test2
++ assert_file_has_content test2-meta CUTE
++ fpath=test2-meta
++ shift
++ for re in "$@"
++ grep -q -e CUTE test2-meta
++ env LD_PRELOAD=/<<PKGBUILDDIR>>/tests/libreaddir-rand.so ostree --repo=/var/tmp/tap-test.zh9sgb/repo show --print-metadata-key=SOMENUM test2
++ assert_file_has_content test2-meta 'uint64 3026418949592973312'
++ fpath=test2-meta
++ shift
++ for re in "$@"
++ grep -q -e 'uint64 3026418949592973312' test2-meta
++ _fatal_print_file test2-meta 'File '\''test2-meta'\'' doesn'\''t match regexp '\''uint64 3026418949592973312'\'''
++ file=test2-meta
[...]

PR #1393, opened 2018-01-03
tests: Assert that byte-order is swapped on LE but not BE CPUs

smcv:
Closes: #1392

---

I'm still not sure I understand why the "obvious" code path through `commit --add-metadata` and `show` byteswaps the metadata - I would have expected that ostree would either define the metadata to be stored in a standardized byte order (in which case machines of the other byte order should byteswap during commit and byteswap back during show), or define the metadata to be stored in the host byte order of the machine that is expected to consume the tree, or store a byte-order indicator. But this does seem to work on both x86_64 (LE) and s390x (BE).

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -130,7 +130,13 @@ experimental_test_scripts = \
 	tests/test-summary-collections.sh \
 	tests/test-pull-collections.sh \
 	$(NULL)
-test_extra_programs = $(NULL)
+test_extra_programs = \
+	tests/get-byte-order \
+	$(NULL)
+
+tests_get_byte_order_SOURCES = tests/get-byte-order.c
+tests_get_byte_order_CFLAGS = $(AM_CFLAGS) $(GLIB_CFLAGS)
+tests_get_byte_order_LDADD = $(GLIB_LIBS)
 
 tests_repo_finder_mount_SOURCES = tests/repo-finder-mount.c
 tests_repo_finder_mount_CFLAGS = $(common_tests_cflags)
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -759,8 +759,20 @@ $OSTREE show --print-metadata-key=FOO test2 > test2-meta
 assert_file_has_content test2-meta "BAR"
 $OSTREE show --print-metadata-key=KITTENS test2 > test2-meta
 assert_file_has_content test2-meta "CUTE"
+
 $OSTREE show --print-metadata-key=SOMENUM test2 > test2-meta
-assert_file_has_content test2-meta "uint64 3026418949592973312"
+case "$("${test_builddir}/get-byte-order")" in
+    (4321)
+        assert_file_has_content test2-meta "uint64 42"
+        ;;
+    (1234)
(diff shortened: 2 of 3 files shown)


cgwalters:
The byte order stuff is a mess; for core ostree we standardized on BE but that doesn't (necessarily) cover user-added metadata.  We also screwed up on static deltas but that's mostly fixed now.

cgwalters:
Nice, thank you very much!  Keeping the test suite running is very much appreciated!  I have a half-done WIP to run the ostree tests in the new [Fedora test standard](https://fedoraproject.org/wiki/Changes/InvokingTestsAnsible) and it will greatly benefit from these fixes.  Thanks again!

@rh-atomic-bot r+ 8e4a15d

smcv:
> for core ostree we standardized on BE but that doesn't (necessarily) cover user-added metadata

Sure, but if I put 42 in, shouldn't I get 42 out, without needing to know or care whether ostree stores it in BE, LE, JSON or hieroglyphs?

cgwalters:
> Sure, but if I put 42 in, shouldn't I get 42 out, without needing to know or care whether ostree stores it in BE, LE, JSON or hieroglyphs?

The problem is it's just one big variant...so we started doing byteswap-if-necessary for the core metadata like the timestamp, but that necessarily swaps the user metadata too.

We probably could teach `show` to byteswap-if-necessary just the core and display user metadata in a separate `g_variant_print()` operation but that would be a bit backwards incompatible.

smcv:
> Keeping the test suite running is very much appreciated

I wish we had more of it, but unfortunately Debian mostly provides build-time testing (`make check`) in a rather artificial environment, so I can't test anything that needs extended attributes, multiple users, root, etc. across all architectures.

The autopkgtest infrastructure runs GNOME-style installed-tests, but I can currently only do that on x86_64 (LXC on official infrastructure and qemu on my laptop) - there used to be arm64 LXC workers but those are currently unavailable.

smcv:
> We probably could teach `show` to byteswap-if-necessary just the core and display user metadata in a separate `g_variant_print()` operation but that would be a bit backwards incompatible.

It certainly feels as though something not right if `ostree commit --add-metadata` and `show` aren't inverses of each other - either both mapping host to network byte order if necessary, or neither doing so. I haven't looked into the underlying C APIs, but I feel as though each C API entry point ought to be able to document whether it expects network or host byte order (with the answer being host byte order 99% of the time). Is this a quirk of the CLI, or is the C API equally problematic?

(I actually think the existence of `g_variant_byteswap()` is a design flaw - I think `g_variant_new_from_data()`, `g_variant_store()` and similar functions should have taken an endianness parameter - but it's far too late for that now.)

Adding new, parallel APIs and deprecating the current ones might be a proportionate response. It's unfortunate that the serialized data structure on disk doesn't carry an indication of how it should be interpreted, but perhaps it isn't too late for rpm-ostree to start writing out metadata in a consistent endianness, with a flag to say it has done so?

cgwalters:
Sorry, meant to follow up earlier.  Yes...new API could make sense.  And a consistent endianness marker.

PR #1396, opened 2018-01-06
rofiles: Fix --copyup when creating a new file

cgwalters:
This tripped up the `docbook-dtds` `%post` in my experiments
with doing rpm-ostree for buildroots.

I cloned and built [xfstests](https://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git)
but haven't yet investigated actually running it.

In the meantime let's do the obvious fix here; we need to distinguish
between "copyup enabled" and "actually did a copyup" in the open path
at least, since if we didn't do a copyup we don't need to re-open.

### DIFF
--- a/src/rofiles-fuse/main.c
+++ b/src/rofiles-fuse/main.c
@@ -247,9 +247,13 @@ gioerror_to_errno (GIOErrorEnum e)
 }
 
 static int
-verify_write_or_copyup (const char *path, const struct stat *stbuf)
+verify_write_or_copyup (const char *path, const struct stat *stbuf,
+                        gboolean *out_did_copyup)
 {
   struct stat stbuf_local;
+
+  if (out_did_copyup)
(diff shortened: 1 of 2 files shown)


jlebon:
> since if we didn't do a copyup we don't need to re-open

The patch makes sense, though can you expand on how this breaks things? This is just an optimization, right? If we didn't actually copy up but `opt_copyup` is `TRUE`, there's no harm in closing and reopening, right?

cgwalters:
>  If we didn't actually copy up but opt_copyup is TRUE, there's no harm in closing and reopening, right?

Right, except (this goes back to a comment you previously made IIRC being prescient) since we're dropping `O_CREAT` things break - since the "second" open is really the first.

jlebon:
@rh-atomic-bot r+ b707a07

PR #1397, opened 2018-01-08
Allow building without libsoup

OYTIS:
Signed-off-by: Anton Gerasimov <anton.gerasimov@openmailbox.org>

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -381,6 +381,14 @@ dist_test_scripts += $(_installed_or_uninstalled_test_scripts)
 test_programs += $(_installed_or_uninstalled_test_programs)
 endif
 
+if !USE_LIBSOUP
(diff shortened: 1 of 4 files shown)


OYTIS:
This is more for discussion than for immediate merger as is. I understand there was probably a good reason to fail in this case, but there are also use cases (like my favourite embedded one) where tests don't matter much and additional dependencies do. Maybe there could be another configure parameter (`--without-soup-please` or something) for such cases.

OYTIS:
That would be great, thank you.

cgwalters:
OK, updated :arrow_up: - look good to you?

OYTIS:
Looks good, thank you.

cgwalters:
@rh-atomic-bot r+ 1383812

PR #1399, opened 2018-01-09
grub2: Exit gracefully if there's no system ostree repository

cgwalters:
Apparently there testing systems that literally install *all*
packages.  Having `ostree-grub2` currently causes grub2 to fail
on a non-ostree managed system.  Let's just gracefully exit
if there's no system repository.

https://bugzilla.redhat.com/show_bug.cgi?id=1532668


jlebon:
@rh-atomic-bot r+ 0484209

PR #1400, opened 2018-01-09
ci: Bump flatpak version (and build to f27)

cgwalters:
Just keeping this updated.


cgwalters:
@alexlarsson Any idea offhand why we wouldn't be finding the icon cache here?

alexlarsson:
There is an optional trigger to rebuild the icon cache, but it needs gtk-update-icon-cache installed. Maybe some dependency changed that caused this to not be installed anymore?

cgwalters:
Got it!  Was missing `hicolor-icon-theme`.  This should be good to go now.

jlebon:
@rh-atomic-bot r+ 3b68820

PR #1401, opened 2018-01-09
Support for booting without initramfs (rebased)

wmanley:
This is #442 rebased and with some additional tests.  We've been using this patch successfully for over a year in combination with #215 and some others.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1700,21 +1700,29 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
   g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", kernel_layout->kernel_namever, NULL);
   ostree_bootconfig_parser_set (bootconfig, "linux", boot_relpath);
 
+  val = ostree_bootconfig_parser_get (bootconfig, "options");
+  g_autoptr(OstreeKernelArgs) kargs = _ostree_kernel_args_from_string (val);
+
   if (kernel_layout->initramfs_namever)
     {
       g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", kernel_layout->initramfs_namever, NULL);
       ostree_bootconfig_parser_set (bootconfig, "initrd", boot_relpath);
     }
-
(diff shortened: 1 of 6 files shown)


vtolstov:
@cgwalters does it possible to update #215 while new grub2 backend not ready? I have already bunch of hosts with /boot on / and i'm always need to prepend /boot inside grub.cfg

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1700,21 +1700,29 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
     {
       g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", kernel_layout->initramfs_namever, NULL);
       ostree_bootconfig_parser_set (bootconfig, "initrd", boot_relpath);
     }
-
-  val = ostree_bootconfig_parser_get (bootconfig, "options");
+  else
+    {
+      g_autofree char *prepare_root_arg = NULL;
+      prepare_root_arg = g_strdup_printf ("init=/ostree/boot.%d/%s/%s/%d/usr/lib/ostree/ostree-prepare-root",
```
cgwalters: This path should use `LIBEXECDIR` I think.

cgwalters:
Did you already rebase #215 then?

src/libostree/ostree-sysroot-deploy.c
wmanley: I don't think so.  See fd1a044d6e75b336ea880530979067926c9b0a09.  I think the idea is that `ostree-prepare-root` needs to be at a fixed location in the rootfs otherwise a different version of ostree with different configure flags couldn't successfully deploy a rootfs with standard configure flags.

wmanley:
> Did you already rebase #215 then?

Yes, I've now created a PR for this too: #1404

src/libostree/ostree-sysroot-deploy.c
cgwalters: Yep, I even agreed with this in https://github.com/ostreedev/ostree/pull/449#issuecomment-238588405 - though for a different reason; I'm not sure how realistic it is to worry about different configure flags.

cgwalters:
@rh-atomic-bot r+ a97c054

PR #1402, opened 2018-01-09
bin/commit: add --keep-metadata option

jlebon:
Clients of libostree such as rpm-ostree make extensive use of the
`ostree commit -b foo --tree=ref=foo` pattern in their tests, e.g. to
simulate an update.

What I'm trying to solve here is that it's often the case that we want
to keep metadata from the previous commit without having to be too
verbose (i.e. reading from the parent, then passing it as an argument).

The new `--keep-metadata` switch makes this really easy. I intend to use
this in the rpm-ostree testsuite to make sure we always carry over the
`source-title` metadata as well as during set up for tests that require
`rpmostree.rpmdb.pkglist` metadata.

I initially implemented this in a small wrapper script that uses the API
directly, though we make use of so many other `ostree commit` functions
that it'd require re-implementing a lot of it.

### DIFF
--- a/bash/ostree
+++ b/bash/ostree
@@ -783,6 +783,7 @@ _ostree_commit() {
     local options_with_args="
         --add-detached-metadata-string
         --add-metadata-string
+        --keep-metadata
         --bind-ref
         --body -m
         --body-file -F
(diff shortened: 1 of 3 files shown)


cgwalters:
I feel like there are probably better patterns for what we're doing in the rpm-ostree testsuite, but - this looks good to me as is generally.

@rh-atomic-bot r+ f6994a2

PR #1403, opened 2018-01-10
Add and use --disable-avahi option

mwleeds:


cgwalters:
Patch seems sane at first glance, but it looks like the CI failure isn't a flake:

> (ostree find-remotes:27104): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed

mwleeds:
@cgwalters I rewrote it to be a `--finders` option instead, at the suggestion of @pwithnall. I'll open that as another PR.

Issue #1405, opened 2018-01-10
pass ostree version in http user agent string


dustymabe:
We have a case where there is a bug in libcurl in old versions of ostree and we are going to work around in the rpm package for newer versions of ostree and also server side in Fedora so that clients can upgrade. What we'd like to be able to do is send clients with the bug one place and clients without the bug to another place, but we have no way to do that because we don't know what version of ostree the client is using.

If we can get that information passed to us in an http header or user agent string then we can implement workarounds in the future if we need them. 

some background here: https://pagure.io/atomic-wg/issue/405

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1406

PR #1406, opened 2018-01-10
lib/fetcher: Add version to USER_AGENT string

cgwalters:
This came up in allowing Fedora infrastructure to work around a libcurl bug with
HTTP2: https://pagure.io/atomic-wg/issue/405

Closes: https://github.com/ostreedev/ostree/issues/1405


jlebon:
@rh-atomic-bot r+ e8bdaae

Issue #1224, opened 2017-09-27
Support faster dir diffing via xattr


cgwalters:
For `rpm-ostree ex livefs` we currently deal with both the commit tree and the "live" tree (and actually the rpm database diff, but that's not related here).  Normally with libostree, a commit tree and a live tree are identical.

But `ex livefs` is all about breaking that barrier - for example, we want the ability to cherry-pick just a subset of changes (perhaps just a new package).  And further, we need to handle the case where things were interrupted.  In the latter case in particular, we *could* try to re-lift the transient state back to a commit but...that feels ugly.

I think the livefs logic would be a lot better if we mostly dealt with checked-out paths.  Then we can handle the interrupted case more easily.  But, it'd be very helpful to have the ability to quickly determine when two directories are identical.

I propose we optionally stick an `ostree.dirtree` xattr on our checked out directories which has its dirtree hash.  At that point we can add a fast non-`GFile` based diff.

jlebon:
Can you flesh this out a bit more on this? Would libostree learn to scrub out (or recalculate?) `ostree.dirtree` prior to linking new files when doing the subpath checkout into `/usr`?

cgwalters:
Yeah, we'd need to scrub the xattr whenever we modify a directory.

PR #1408, opened 2018-01-11
Add ostree checkout overwrite mode "--update"

wmanley:
ostree checkout --update=FROM_COMMIT TO_COMMIT DESTINATION

will assume that `DESTINATION` already contains a checkout of `FROM_COMMIT` and will attempt to do the least amount of work to make it contain `TO_COMMIT`.

The intention is to enhance `ostree admin deploy` to update previous deployments rather than checking out whole new ones.  This should be faster.  Currently on our system with slow eMMC disk it takes 1m40s to do a deploy (~150000 files).  Ideally if only one or two files have changed it should be nearly instantaneous - suitable for interactive development.  This should also reduce eMMC wear.  This can happen in future PRs though.


cgwalters:
I haven't looked at this yet but it seems pretty related to `rpm-ostree livefs` work; see also:
https://github.com/ostreedev/ostree/issues/1224

If this is about development though I find `ostree admin unlock` to be super useful.  I could certainly imagine teaching libostree how to do "live deployment updates" on top of the overlayfs.

cgwalters:
Pretty interesting stuff, thanks a lot for working on this!  The code looks sane at a very quick glance, but I'm going to need a little bit of time to think about it.  In particular I want to try to figure out if this applies to what we have today in `rpm-ostree ex livefs` and also my idea from https://github.com/ostreedev/ostree/issues/1224

But it'd be great if we can pin down the higher level use case again - do you see this as *just* a development thing or do you see it as potentially useful for production hotfixes?

In the [rpm-ostree livefs](https://github.com/projectatomic/rpm-ostree/issues/639) issue one concrete use case I had was being able to hotfix OpenSSH, which is a classic example of a service that is pretty safe to hotfix.

wmanley:
> But it'd be great if we can pin down the higher level use case again - do you see this as just a development thing or do you see it as potentially useful for production hotfixes?

This is just intended as supporting work for #1419.  The motivation for that is faster deploys with large numbers of files on a slow filesystem.  Currently it takes 1m40s to do a deploy and a reboot takes about 10s.  This is too slow - we want to be able to deploy a new version of software to our devices fast enough that our customers don't notice any outage.  In particular we want to add some more containers which increases the number of files.  This will make it even slower.

We can't do a deploy while our customers are using the systems because doing the checkout monopolises the disk.  As a side effect it makes development faster too.  I'll expand on how our build/deploy system works a little later, when I have time...

Note: I'm not live-updating the current deployment with this command.  This isn't a use case I'd considered, but I can see how it could work out well.  I'll expand on this on #1419.

I've made this a separate pull-request because I think it's easier to review that way and easier to test in isolation than bundled with #1419.

wmanley:
I definitely prefer your approach of recording the tree SHAs in xattrs as suggested in #1224.  It's simpler and less error-prone to passing the xattrs in to the `--update`.

In #1419 I introduced the concept of the deploy cache which stored old deploys named by the commit id.  The commit id would be retrieved from the directory name when doing a new deploy.  This would be passed to this `--update` mode when doing a deploy.

This xattrs approach is better for a few reasons:

1. The way the csums are stored is not specific to `ostree admin deploy`, they can also be set when doing a normal checkout.
2. I don't have to special-case `/etc` when updating deploys.  In #1419 I was working around the fact that the csum in of the deploy wouldn't match the real checked-out data because /etc had been merged.  Because these trees would be csumed individually that's no longer an issue - we just unset the root csum xattr before doing etc merge.
3. Updating a live deploy in place becomes possible because we don't need to rename the deployment just because the sha of the contents have changed
4. Updating a live deploy becomes possible because we can cope with non-atomically updated trees - a directory that temporarily doesn't correspond to a particular SHA doesn't have that SHA set on its xattrs.  Theres no risk of a partial update being undetected (assuming the appropriate calls to `sync`.

I'm not saying that live updates would be a good idea, but they'd certainly be possible under this scheme.

I'll have a think about what it would take to implement that.

cgwalters:
As a general request; can you file an issue first for bigger features so we can sketch out the design/use cases a bit before we do the code?    Getting big patches is obviously great, not going to complain 😉!

So...are you thinking it makes sense to do #1224 first?  I can probably take a crack at that sometime soon myself.

wmanley:
> As a general request; can you file an issue first for bigger features so we can sketch out the design/use cases a bit before we do the code? Getting big patches is obviously great, not going to complain 😉!

Yeah, I agree that this is ideally how it would work. In this case time pressures have meant that we don't always have time for this. I spent probably 2-3 days on this.  We're already shipping this and #1419 on our product.  It was a prerequisite for other important customer-facing work (https://stb-tester.com/blog/2018/02/06/stb-tester-v28-released)

As such I'm not precious about changing it to get it merged properly so I don't think we've lost much with this approach in this instance.  I won't be upset if you ask for changes or flat out say no :).

The other element is that I didn't want to propose it until I knew I could do it.  I had a stab at implementing this over a year ago - but failed.  Your commit e6f17b949d8214049faca04d01d82bb28703cc77 made it much easier for me to understand the code, because it removed a layer of abstraction.  So when I had another go and succeeded I raised this PR.  I wouldn't have felt good about coming along and starting a discussion about a new feature that I wanted if I felt like I'd be "demanding" that others implement it for me,

> So...are you thinking it makes sense to do #1224 first? I can probably take a crack at that sometime soon myself.

I think the only thing that would change in this PR if #1224 were implemented is where the initial tree sha would come from.  So changes here would depend only on a small subset of #1224 - specifically recording the SHAs in the xattrs.

How do you see this working? The two questions in my mind are:

1. When does the caller make the guarantee that the directories won't be modified?  For the xattr to actually match the contents of the directory the contents can't change between when the xattr is set to when it is overwritten. I see a few options:

    1. Have an option to the intiial `ostree checkout` like `--record-dirtree-xattrs`. This is a promise to ostree that the caller won't modify the checked-out tree without clearing the xattrs.  And then when you come to do an `ostree checkout --update` ostree just trusts that the xattrs are correct.
    2. Always record dirtree xattrs when checking out and have an option like `--trust-dirtree-xattrs` when updating a checkout.  This flag would signify that the user hasn't modified the checkout since the initial checkout.
    3. Implement both flags
    4. Have a single flag that says both that the xattrs are trustworthy and that xattrs should be recorded.  This could be implied by `ostree checkout --update`
    5. There's no need for an additional flag - `ostree checkout --update` implies this mode.

2. How to handle `sync`/`fsync`?  When updating you'd have to clear the xattr, then update the directory contents, then set the new xattr.  If there's no `fsync` then a power loss could lead to inconsistency between the xattrs and the contents.  At the same time we don't want to add a whole bunch of `fsync` calls as it will be too slow (1810de2b51680dbf35fdbd33d0a8d7e65eadc91f).  I would propose:
    
    1. Recurse through removing the xattrs,
    2. `syncfs`
    3. Update the contents of the directories
    4. `syncfs`
    5. Recurse through adding the new xattrs.

    The increased complexity is probably not unmanageable, but isn't ideal either.  An alternative is to not worry about it and push the complexity to a higher level.  I've already had to handle this case in #1419 (See b68ecc273df52c10e1e2b34ed6fae56c84e1beac) so maybe it's not that big a deal - but having reliable xattrs would allow updates in place.  Maybe these additional guarantees (and the associated complexity) can be pushed into the future.

Thoughts?

wmanley:
I'm going away on holiday so I'm not going to be able to look at this until March.  I'd still be interested in your thoughts in the mean time, I'm just letting you know that I won't be doing anything about them for a little while.

cgwalters:
> We can't do a deploy while our customers are using the systems because doing the checkout monopolises the disk. As a side effect it makes development faster too.

I get the development case for sure.  But I want to understand better how you envision using this for production.  I reread through the thread and your patches, but some important details aren't yet clear to me.

Let's say we have two deployments, current and rollback (the ostree default).  Normally when we do an upgrade, ostree will end up GCing the rollback but only *after* an upgrade has been completed - in the bootloader order we always have 2, but on disk we will transiently have 3 checkouts.  However, we clearly could, on upgrade, immediately "take ownership" of the rollback.  Is something like that what you're thinking?  I see the "deploy cache" in later patches.

The downside of this is that the upgrade is less "transactional" in that the rollback will vanish if we're interrupted.  Today we don't have `ostree rollback` but it's trivial to implement (as it is in rpm-ostree); we could add an API for it, and have that API "resurrect" the rollback.  I'm not too concerned about this though; in practice I feel like the `rpm-ostree deploy <version>` approach is nicer (another thing to drive into ostree CLI).

 Am I right that your approach is about taking control over the rollback?

cgwalters:
Some other notes I have:

The xattr diffing is an *enhancement* but I don't see a need to require it; a valid concern you have is eMMC wear, and (AIUI) the extra reads we do as part of a full diff traverse aren't a problem for that

We should be able to rebase and land some of the preparatory patches now.

> I don't have to special-case /etc when updating deploys.

This is definitely a concern - how we handle the ostree 3-way merge.  Are you using systemd today?  We need to think about how this feature intersects with [staged deployments](https://github.com/ostreedev/ostree/pull/1503) as I'd really like to make that the default in the future.

What filesystem are you using?  Does it have reflinks?  That helps a lot for `/etc`.  Although presumably you can also minimize `/etc` for your use case.

wmanley:
>Let's say we have two deployments, current and rollback (the ostree default). Normally when we do an upgrade, ostree will end up GCing the rollback but only after an upgrade has been completed

Yes, so in #1419 I've modified the GC code to move unused deploys to the cache rather than deleting it:

https://github.com/stb-tester/ostree/blob/09a1e345573caf9c30e70bbf36674f81f88fced6/src/libostree/ostree-sysroot-cleanup.c#L443-L444

This is a lot faster than recursively deleting the deploy as it just involves a `rename` and we can reuse the deploy later without interfering with rollbacks, etc.

> Am I right that your approach is about taking control over the rollback?

No, we take control over the last GCed deploy.

wmanley:
Rebased on master

cgwalters:
Can you split off some of the prep patches here to a separate PR?

PR #1409, opened 2018-01-11
tests/libtest-core: support multiple literal checks

jlebon:
`grep` supports checking multiple fixed strings separated by newlines,
but it's mostly just easier to pass them as separate arguments, so let's
support that. This is now at parity with the similar
`assert_file_has_content`.


jlebon:
Upstreamed from https://github.com/projectatomic/rpm-ostree/pull/1200.

cgwalters:
@rh-atomic-bot r+ 47393b1

PR #1411, opened 2018-01-12
Add support for devicetree files alongside the kernel and initramfs

wmanley:
Much like the (optional) initramfs at `/usr/lib/ostree-boot/initramfs-<SHA256>` or `/usr/lib/modules/$kver/initramfs` you can now optionally include a flattened devicetree (.dtb) file alongside the kernel at `/usr/lib/ostree-boot/devicetree-<SHA256>` or `/usr/lib/modules/$kver/devicetree`.

This is useful for embedded ARM systems which need the devicetree file
loaded by the bootloader for the kernel to discover and initialise
hardware.  See https://en.wikipedia.org/wiki/Device_tree for more
information.

This patch was mostly produced by copy-pasting code for initramfs handling
and renaming `s/initramfs/devicetree/g`.  It's not beautiful, but it is
fairly straightforward.

It may be useful to extend device-tree support in a number ways in the
future.  Device trees dependant on many details of the hardware they
support.  This makes them unlike kernels, which may support many different
hardware variants as long as the instruction-set matches.  This means that
a ostree tree created with a device-tree in this manner will only boot on
a single model of hardware.  This is sufficient for my purposes, but may
not be for others'.

I've tested this on my NVidia Tegra TK1 device which has u-boot running
in syslinux-compatible mode.

Note: if #1404 is merged before this this PR will have to be updated - and vice-versa.

### DIFF
--- a/docs/manual/atomic-upgrades.md
+++ b/docs/manual/atomic-upgrades.md
@@ -95,7 +95,7 @@ collected at any point.
 ## The /ostree/boot directory
 
 However, we want to optimize for the case where the set of
-kernel/initramfs pairs is the same between both the old and new
+kernel/initramfs/devicetree sets is the same between both the old and new
 deployment lists.  This happens when doing an upgrade that does not
 include the kernel; think of a simple translation update.  OSTree
 optimizes for this case because on some systems `/boot` may be on a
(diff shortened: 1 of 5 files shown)


cgwalters:
Looks like the test count needs tweaking:

```
ERROR: tests/test-no-initramfs.sh 8 switching to bootdir with no initramfs layout=/boot # UNPLANNED
ERROR: tests/test-no-initramfs.sh 9 switching from no initramfs to initramfs enabled sysroot layout=/boot # UNPLANNED
ERROR: tests/test-no-initramfs.sh 10 switching from initramfs to no initramfs sysroot with devicetree layout=/boot # UNPLANNED
ERROR: tests/test-no-initramfs.sh - too many tests run (expected 7, got 10)
```

Otherwise looks pretty good to me!

wmanley:
I've done a `git commit --amend && git push --force-with-lease` to force the tests to rerun.  They seemed to have hanged.

cgwalters:
Nice, thanks a lot!

@rh-atomic-bot r+ cf1b8de

PR #1412, opened 2018-01-12
Metadata structure verification (e.g. path traverse) cleanup/strengthening

cgwalters:
I was reading about a recent security issue with both EMC and VMWare:
https://arstechnica.com/information-technology/2018/01/emc-vmware-security-bugs-throw-gasoline-on-cloud-security-fire/
    
It's a classic path traversal problem, and that made me think more about our
handling of this in libostree. Fortunately of course, not being new to this
rodeo, long ago I *did* consider path traversal. Inside the HTTP pull code, we call
`ot_util_filename_validate()`. Also, `fsck` does this too.

The first patch just adds tests for those.
    
So...IMO, we squeak by here and I have no plans to request a CVE; even if there were path traversal practically speaking it mostly only matters for flatpak (maybe buildstream?), or people downloading untrusted ostree content *not* for the bootable OS.  (Someone able to provide you malicious OS content can do whatever they want on the next boot)

That said though even for "OS ostrees" there are use cases for e.g. checking out subtrees as non-root outside of the deployment.

So I went through the code and added a lot of hardening for this; specifically `pull-local --untrusted` and `checkout`.


cgwalters:
This one should be good to go.  Just a Travis flake downloading debs.

jlebon:
Yup, this look good to me!
@rh-atomic-bot r+ d61a342

PR #1414, opened 2018-01-13
find-remotes: Minor fixes to --finders code

mwleeds:
This introduces no functional changes, only cleanups.

### DIFF
--- a/src/ostree/ot-builtin-find-remotes.c
+++ b/src/ostree/ot-builtin-find-remotes.c
@@ -37,8 +37,8 @@ static gboolean opt_pull = FALSE;
 static GOptionEntry options[] =
   {
     { "cache-dir", 0, 0, G_OPTION_ARG_FILENAME, &opt_cache_dir, "Use custom cache dir", NULL },
-    { "finders", 0, 0, G_OPTION_ARG_STRING, &opt_finders, "Use the specified comma separated list of finders (e.g. config,lan,mount)", "FINDERS" },
     { "disable-fsync", 0, 0, G_OPTION_ARG_NONE, &opt_disable_fsync, "Do not invoke fsync()", NULL },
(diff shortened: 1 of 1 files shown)


src/ostree/ot-builtin-find-remotes.c
```diff
@@ -232,14 +232,14 @@ ostree_builtin_find_remotes (int            argc,
   /* Build the array of OstreeRepoFinder instances */
   if (opt_finders != NULL)
     {
-      g_auto(GStrv) finders_strings = NULL;
+      g_autofree const char **finders_strings = NULL;
```
jlebon: This will leak though, right? This will only free the array itself, but not each member, like `g_strfreev`.

src/ostree/ot-builtin-find-remotes.c
```diff
@@ -232,14 +232,14 @@ ostree_builtin_find_remotes (int            argc,
...
 
-      finders_strings = g_strsplit (opt_finders, ",", 0);
+      finders_strings = (const char **)g_strsplit (opt_finders, ",", 0);
       if (!validate_finders_list (finders_strings, context, error))
         return FALSE;
 
-      finders = g_ptr_array_new ();
-      for (char **iter =finders_strings; iter && *iter; iter++)
+      finders = g_ptr_array_new_with_free_func (NULL);
```
jlebon: Any reason for this? Did you mean to use `g_object_unref` as the free function here and get rid of the local `finder_*` vars?

mwleeds: @pwithnall suggested it as a way to make it clear that there's intentionally not a `free_func`

src/ostree/ot-builtin-find-remotes.c
mwleeds: Oops. I guess there's no way to make it `const` then.

jlebon:
@rh-atomic-bot r+ e32ca44

PR #1417, opened 2018-01-15

cgwalters:
I'm still doing release, then versionbump as separate PRs to ensure
the release commit is tested by itself.


pwithnall:
@rh-atomic-bot r+ c22bb3efc658da173d71c05ee51de5f66a70aa71

PR #1418, opened 2018-01-15
ostree admin deploy: Add --no-prune option

wmanley:
If you want cleanup, but don't want to prune the repo.  Pruning can be quite expensive so `ostree admin deploy` can be much faster without pruning.

On the NVidia Tegra TK1 pruning can take 20s, even when there is nothing to delete.

TODO:

- [ ] test it

### DIFF
--- a/src/ostree/ot-admin-builtin-deploy.c
+++ b/src/ostree/ot-admin-builtin-deploy.c
@@ -154,7 +154,7 @@ ot_admin_builtin_deploy (int argc, char **argv, OstreeCommandInvocation *invocat
                                    kargs_strv, &new_deployment, cancellable, error))
     return FALSE;
 
-  OstreeSysrootSimpleWriteDeploymentFlags flags = 0;
+  OstreeSysrootSimpleWriteDeploymentFlags flags = OSTREE_SYSROOT_SIMPLE_WRITE_DEPLOYMENT_FLAGS_NO_CLEAN;
   if (opt_retain)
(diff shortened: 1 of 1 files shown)


cgwalters:
I only quickly glanced at this so far, but I'm a bit uncertain as to the need for the flag and the option?  Can't we just port the cli to use `ostree_sysroot_simple_write_deployment_with_options()` from a5d5333c83c7d0a7ec477243ec602987bf104ce5 ?

wmanley:
> Can't we just port the cli to use ostree_sysroot_simple_write_deployment_with_options() from a5d5333 ?

I had a go at that and we end up duplicating the logic from `ostree_sysroot_simple_write_deployment` that comes up with the list of new deployments.  Instead I have just passed the `NO_CLEAN` flag and done the cleaning ourselves in `ot_admin_builtin_deploy`.  The only theoretical difference is that now the cleaning will happen with a read-write `/boot` rather than a read-only one.

I haven't tested it yet though...

cgwalters:
Going back over this I'm uncertain why we don't just forcibly do `prepare_cleanup()` before starting to write deployments or so.  But that'd require some more tricky investigation.

I'm not worried about the read/write vs ro `/boot` yet, that logic is going to need a lot more work that should end up going along with https://github.com/ostreedev/ostree/issues/1265

So this LGTM.

@rh-atomic-bot r+ 1731f42

PR #1419, opened 2018-01-15
WIP: deploy: Attempt accelerated deploy by updating old deploy

wmanley:
...rather than doing a complete fresh checkout.

Currently on our (embedded) system with slow eMMC disk it takes 1m40s to do
a deploy (~150000 files). Ideally if only one or two files have changed it
should be nearly instantaneous - suitable for interactive development. This
should also reduce eMMC wear.

After applying these patches (and with `--no-prune` #1418) it takes <2s to do a deploy - a 50x-100x speedup.

We store unpacked container images as part of our rootfs.  Checkouts with O(number of changes) rather than O(number of files) is going to be more and more important for us as we add more containers.

On any failure we fall back to doing a full deploy.

This builds on top of #1408.  Although it's not ready I've raised this PR to solicit opinion.

TODO:

- [ ] Merge #1408 and rebase
- [ ] Only take `/usr` from old deploys - rather than removing `/etc`
- [ ] `syncfs` before moving deployment into final directory.
- [ ] Get tests to pass
- [ ] Add expicit new tests for this feature.

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -58,6 +58,7 @@ _installed_or_uninstalled_test_scripts = \
 	tests/test-basic-user.sh \
 	tests/test-basic-user-only.sh \
 	tests/test-basic-root.sh \
+	tests/test-checkout-update.sh \
 	tests/test-pull-subpath.sh \
 	tests/test-archivez.sh \
 	tests/test-remote-add.sh \
--- a/bash/ostree
+++ b/bash/ostree
@@ -704,6 +704,7 @@ _ostree_checkout() {
     local options_with_args="
         --from-file
         --fsync
+        --replace
         --repo
         --subpath
     "
@@ -713,6 +714,10 @@ _ostree_checkout() {
     case "$prev" in
         --from-file)
             __ostree_compreply_all_files
+            return 0
+            ;;
+        --replace)
+            __ostree_compreply_commits
             return 0
             ;;
         --repo|--subpath)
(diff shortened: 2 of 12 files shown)


cgwalters:
> ~150000

Hah, that's quite close to the number of files on this desktop:

```
$ rpm-ostree status
State: idle
Deployments:
● fedora-27:fedora/27/x86_64/workstation
                   Version: 27.66 (2018-02-05 15:55:38)
                BaseCommit: d0100a21ecca78f7931b4278c25528b455f9a0e62eeb4690bb93f46b345dc818
              GPGSignature: Valid signature by 860E19B0AFA800A1751881A6F55E7430F5282EE4
           LayeredPackages: emacs fuse-sshfs git-evtag keepassx krb5-workstation libvirt-client oci-kvm-hook opensc origin-clients pcsc-lite-ccid powerline strace tmux vagrant-libvirt virt-manager xsel ykclient
                            ykpers
$ find /usr | wc -l
146136
$
```

cgwalters:
I just want to note I haven't forgotten about your PRs here; I feel a bit guilty every time I see the PR list as I really want to get them in, but currently the staging bits (as part of auto-updates) have stayed at the top of my priority queue for a while, and these patches also really want a better testing framework than we have today, which is also slowly happening.

As the new ansible/VM framework lands and improves we can more easily test these types of patches.

wmanley:
Rebased on master.

PR #1424, opened 2018-01-17
lib/pull: Port a few functions to new style

cgwalters:
Prep for further work here. This diff is a bit noisy for the delta bits because
the identation was off originally as well.


jlebon:
Nice! @rh-atomic-bot r+ 5340f6e

PR #1425, opened 2018-01-18
lib/pull: Allow specific commits in P2P updates

mwleeds:
Currently users of the find_remotes_async()/pull_from_remotes_async()
functions have no way to specify a commit hash to use instead of the
latest one available. This commit implements an "override-commit-ids"
option analogous to the one used by ostree_repo_pull_with_options().
It's accomplished by returning OstreeRepoFinderResult objects pointing
to the given commit checksum(s) regardless of which ones were available
from the remotes, but in the future this implementation could be
improved to take into account the commits advertised by the remotes.

One effect of this is that flatpak will have the ability to downgrade
apps that use collection IDs
(https://github.com/flatpak/flatpak/issues/1309).

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -4613,7 +4613,9 @@ static void find_remotes_cb (GObject      *obj,
  *
  * The following @options are currently defined:
  *
- *   * `override-commit-ids` (`as`): Array of specific commit IDs to fetch for refs
+ *   * `override-commit-ids` (`as`): Array of specific commit IDs to fetch. The nth
+ *   commit ID applies to the nth ref, so this must be the same length as @refs, if
+ *   provided.
  *
  * @finders must be a non-empty %NULL-terminated array of the #OstreeRepoFinder
  * instances to use, or %NULL to use the system default set of finders, which
@@ -4642,6 +4644,7 @@ ostree_repo_find_remotes_async (OstreeRepo                     *self,
   g_autoptr(OstreeRepoFinder) finder_config = NULL;
   g_autoptr(OstreeRepoFinder) finder_mount = NULL;
   g_autoptr(OstreeRepoFinder) finder_avahi = NULL;
+  g_autofree char **override_commit_ids = NULL;
 
   g_return_if_fail (OSTREE_IS_REPO (self));
   g_return_if_fail (is_valid_collection_ref_array (refs));
@@ -4650,6 +4653,12 @@ ostree_repo_find_remotes_async (OstreeRepo                     *self,
   g_return_if_fail (finders == NULL || is_valid_finder_array (finders));
   g_return_if_fail (progress == NULL || OSTREE_IS_ASYNC_PROGRESS (progress));
   g_return_if_fail (cancellable == NULL || G_IS_CANCELLABLE (cancellable));
+
+  if (options)
+    {
+      (void) g_variant_lookup (options, "override-commit-ids", "^a&s", &override_commit_ids);
+      g_return_if_fail (override_commit_ids == NULL || g_strv_length ((gchar **) refs) == g_strv_length (override_commit_ids));
+    }
 
   /* Set up a task for the whole operation. */
   task = g_task_new (self, cancellable, callback, user_data);
@@ -4895,9 +4904,6 @@ find_remotes_cb (GObject      *obj,
     {
       (void) g_variant_lookup (data->options, "override-commit-ids", "^a&s", &override_commit_ids);
     }
-
-  if (override_commit_ids)
-    g_assert (g_strv_length ((gchar **) refs) == g_strv_length (override_commit_ids));
 
   /* FIXME: In future, we also want to pull static delta superblocks in this
    * phase, so that we have all the metadata we need for accurate size
@@ -5150,7 +5156,8 @@ find_remotes_cb (GObject      *obj,
    * points to for it (unless override-commit-ids was used).
    *
    * @ref_to_latest_commit is indexed by @ref_index, and its values are the
-   * latest checksum for each ref. */
+   * latest checksum for each ref. If override-commit-ids was used,
+   * @ref_to_latest_commit won't be initialized or used.*/
   ref_to_latest_commit = g_new0 (const gchar *, n_refs);
 
   for (i = 0; i < n_refs; i++)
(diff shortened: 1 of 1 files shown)


mwleeds:
This enables https://github.com/flatpak/flatpak/pull/1314

mwleeds:
cc: @pwithnall

src/libostree/ostree-repo-pull.c
```diff
@@ -4658,7 +4658,9 @@ static void find_remotes_cb (GObject      *obj,
  * Pass the results to ostree_repo_pull_from_remotes_async() to pull the given @refs
  * from those remotes.
  *
- * No @options are currently supported.
+ * The following @options are currently defined:
+ *
+ *   * `override-commit-ids` (`as`): Array of specific commit IDs to fetch for refs
```
pwithnall: You should document that this has to be the same length as `refs`, if provided, and that the indexing is the same.

src/libostree/ostree-repo-pull.c
```diff
@@ -4903,6 +4905,7 @@ find_remotes_cb (GObject      *obj,
   g_autoptr(OstreeFetcher) fetcher = NULL;
   g_autofree const gchar **ref_to_latest_commit = NULL;  /* indexed as @refs; (element-type commit-checksum) */
   gsize n_refs;
+  g_autofree char **override_commit_ids = NULL;
```
pwithnall: I think this should be `const`: `g_autofree const char **override_commit_ids = NULL`.

src/libostree/ostree-repo-pull.c
```diff
@@ -4935,13 +4938,13 @@ find_remotes_cb (GObject      *obj,
...
-   * configuration.
-   * Use case: downgrading a flatpak app.
-   */
+  if (data->options)
+    {
+      (void) g_variant_lookup (data->options, "override-commit-ids", "^a&s", &override_commit_ids);
+    }
+
+  if (override_commit_ids)
+    g_assert (g_strv_length ((gchar **) refs) == g_strv_length (override_commit_ids));
```
pwithnall: Hmm, technically it would be more helpful if this precondition check was in the `ostree_repo_find_remotes_async()` function itself, so a failure would be reported to the caller immediately without leaving OSTree in a slightly less well-defined state.

You should probably also phrase it as an implication, i.e. `override_commit_ids != NULL ⇒ lengths are equal`:
```
g_return_if_fail (override_commit_ids == NULL || g_strv_length (…) == g_strv_length (…));
```

src/libostree/ostree-repo-pull.c
```diff
@@ -5204,6 +5207,13 @@ find_remotes_cb (GObject      *obj,
       const CommitMetadata *latest_commit_metadata = NULL;
       g_autofree gchar *latest_commit_timestamp_str = NULL;
 
+      if (override_commit_ids)
+        {
+          g_debug ("%s: Using specified commit ‘%s’ for ref (%s, %s).",
+                   G_STRFUNC, override_commit_ids[i], refs[i]->collection_id, refs[i]->ref_name);
+          continue;
+        }
```
pwithnall: It’s probably worth adding a comment above this explaining that if this path is taken, `ref_to_latest_commit` is no longer valid.

src/libostree/ostree-repo-pull.c
mwleeds: Yeah but I guess we would have to add a `GError` parameter to `find_remotes_async` to give feedback to the caller.

src/libostree/ostree-repo-pull.c
cgwalters: Hmmm...that's not the style we use elsewhere though.  I am going back and forth on whether or not it's "technically" right.

This is (somewhat) personal opinion but I generally don't use `const` in C except for `const char *` since [as Wikipedia says](https://en.wikipedia.org/wiki/Const_(computer_programming)) those are stored in read-only memory often and to me it's really more about "owned" vs "borrowed" in Rust terms but...

Some parts of glib use `const char * const *` which is more technically right but also annoying since other things use `char**` and so in the end one just needs to cast, and really we're buried in legacy here and why doesn't someone invent a time machine and go back and give us Rust 10 years ago...

cgwalters:
LGTM after Philip's comments.

mwleeds:
@pwithnall feedback addressed

pwithnall:
@rh-atomic-bot retry

pwithnall:
bot, retest this please

pwithnall:
It seems I am forever doomed to be unable to remember the syntax for interacting with the bot.

pwithnall:
What the heck, this is one way of rerunning the tests:

@rh-atomic-bot r+ 81bc7501e8434169745d02b50016cfd152ce68c2

pwithnall:
I think these failures are not related to the branch. @cgwalters?

```
+ pkg_install @buildsys-build
+ yum -y install @buildsys-build
Last metadata expiration check: 0:00:50 ago on Wed Jan 24 10:49:23 2018.
Traceback (most recent call last):
  File "/usr/bin/yum", line 58, in <module>
    main.user_main(sys.argv[1:], exit_code=True)
  File "/usr/lib/python3.6/site-packages/dnf/cli/main.py", line 179, in user_main
    errcode = main(args)
  File "/usr/lib/python3.6/site-packages/dnf/cli/main.py", line 64, in main
    return _main(base, args, cli_class, option_parser_class)
  File "/usr/lib/python3.6/site-packages/dnf/cli/main.py", line 99, in _main
    return cli_run(cli, base)
  File "/usr/lib/python3.6/site-packages/dnf/cli/main.py", line 115, in cli_run
    cli.run()
  File "/usr/lib/python3.6/site-packages/dnf/cli/cli.py", line 1014, in run
    return self.command.run()
  File "/usr/lib/python3.6/site-packages/dnf/cli/commands/install.py", line 108, in run
    self.base.read_comps(arch_filter=True)
  File "/usr/lib/python3.6/site-packages/dnf/base.py", line 550, in read_comps
    self._comps._add_from_xml_filename(decompressed)
  File "/usr/lib/python3.6/site-packages/dnf/comps.py", line 328, in _add_from_xml_filename
    ret = comps.fromxml_f(fn)
_libpycomps.ParserError: Fatal parser error
### EXITED WITH CODE 1 AFTER 42s
```

cgwalters:
https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/HNFA7Y32SMHUB7XZ3RGDP4IAJ4ZUPXR4/

@rh-atomic-bot retry

PR #1426, opened 2018-01-18
ci/papr: Update most contexts to f27

cgwalters:
Many of them actually already *were* because they
were inherting.

An exception is flatpak which is being worked on in
https://github.com/ostreedev/ostree/pull/1400


jlebon:
@rh-atomic-bot r+ cd653fe

PR #1427, opened 2018-01-18
bin/delta: Fix compilation with relative subdirs --filename

cgwalters:
Currently we were parsing `opt_filename` twice...I dug through
the history a bit and it looks like it may have been an accident
from refactoring.

What we're fixing here concretely is that using relative subdirectories
like `--filename somesubdir/foo` broke because we were incorrectly
passing the `somesubdir/` again.

Closes: #1423


jlebon:
LGTM, though:

```
src/libostree/ostree-repo-static-delta-compilation.c:1373:19: error: unused variable 'tmp_dfd' [-Werror,-Wunused-variable]
  glnx_autofd int tmp_dfd = -1;
                  ^
1 error generated.
```

cgwalters:
Fixup :arrow_up:

jlebon:
@rh-atomic-bot r+ 09ad144

Issue #1431, opened 2018-01-25
Documentation license question


btreut:
under which license is the documentation?

I tried to populate the information about OSTree in Wikipedia by reusing parts of the introduction, but it got deleted due to Copyright issues. So a more liberal Copyright (at least for parts of the manual) would be very welcome. Currently the entry about OSTree [https://en.wikipedia.org/wiki/OSTree] is a stub and I do not feel to be competent enough to put there more Information, which is not reusing stuff more or less verbally. It was again deleted for being a self-referencing ad, but happily someone else restored it.

cgwalters:
So I saw https://en.wikipedia.org/wiki/Wikipedia:Non-free_content but it's really long; is the TL;DR "license under CC BY-SA or alternatively CC BY-SA and GFDL"?

btreut:
I am not sure, but GNU LGPL is not enough. But I guess GFDL and CC licenses would be OK. 

The guy who deleted my first attempt said: "Thanks for checking on that, Bernhard. Unfortunately GNU licenses aren't compatible with Wikipedia. I'd suggest rewriting the article in your own words. – Joe" but this post is only in Wikipedias history.

cgwalters:
Does anyone have an example of a project that does dual licensing of GFDL and CC BY-SA for their docs?

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1432

btreut:
Thanks a lot.

cgwalters:
@btreut Thank *you* for looking at doing a Wikipedia page!  If there's anything else we can do to help don't hesitate to ask.

PR #1432, opened 2018-01-25
docs: Dual license under CC BY-SA and the GFDL

cgwalters:
Originating issue:  https://github.com/ostreedev/ostree/issues/1431

Mailing list post: https://mail.gnome.org/archives/ostree-list/2018-January/msg00004.html

---

This will allow the text to be used in Wikipedia for example; it
also just makes more sense for documentation than the LGPLv2+.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,3 +108,10 @@ Contributing
 ------------
 
 See [Contributing](CONTRIBUTING.md).
+
+
+Licensing
+-------
+
+The license for the *code* of libostree can be found in [COPYING](COPYING).
+The license for the *documentation* of libostree is: `SPDX-License-Identifier: (CC-BY-SA-3.0 OR GFDL-1.3-or-later)`
--- a/docs/manual/adapting-existing.md
+++ b/docs/manual/adapting-existing.md
@@ -166,3 +166,6 @@ Then to actually deploy this tree for the next boot:
 
 This is essentially what [rpm-ostree](https://github.com/projectatomic/rpm-ostree/)
 does to support its [package layering model](https://rpm-ostree.readthedocs.io/en/latest/manual/administrator-handbook/#package-layering).
+
+###### Licensing for this document:
+`SPDX-License-Identifier: (CC-BY-SA-3.0 OR GFDL-1.3-or-later)`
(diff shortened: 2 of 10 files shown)


cgwalters:
(Don't merge yet - I need to send out a "relicensing ok?" email to people)

cgwalters:
@grimthorpe You are included in this due to https://github.com/ostreedev/ostree/commit/596fc2967eba43394d4e336a39553cac701cc1c5 - is this OK with you?

jlebon:
I do not object. :)

marcusfolkesson:
I think we also should add a SPDX identifier[1] tag to the license text.

E.g.
SPDX-License-Identifier: (CC-BY-SA-3.0 AND GFDL-1.3+)

When on it, we should also add the SPDX-identifier for COPYING file.
Do you want me to send a separate patch for SPDX-identifier on COPYING?

[1] https://spdx.org/sites/cpstandard/files/pages/files/using_spdx_license_list_short_identifiers.pdf

cgwalters:
@marcusfolkesson Makes sense, I'll update this PR for the docs.  As for the source - before you do the PR, just so I know: does that involve touching every source file or just the `COPYING`?  And is there a public example of a project doing that?

cgwalters:
I'm not a SPDX expert but isn't it `OR` not `AND` here?

marcusfolkesson:
@cgwalters
Yes, I think we should replace the license text in every source file with the identifier.
Example on projects that is doing this is: Linux kernel, U-Boot, Barebox, Buildroot and so on.

e.g.
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable.git/tree/drivers/usb/core/driver.c?h=v4.14.15

And yes, it is OR as you suggest, I was a little too fast...

cgwalters:
OK how about this for a footer?

###### Licensing for this document:
`SPDX-License-Identifier: (CC-BY-SA-3.0 OR GFDL-1.3+)`

blixtra:
Kinvolk is fine with the OStree documentation it holds rights to being relicensed CC-BY-SA-3.0 AND GFDL-1.3+

wmanley:
Relicensing is fine by me

maxamillion:
+1 to Relicensing

miabbott:
+1 to Relicensing

cgwalters:
Just for my own future reference, the `man/` directory probably should have been covered here, but we can do that later.

jlebon:
@rh-atomic-bot r+ c5c6ed6

jlebon:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1436, opened 2018-01-26
Migrate to SPDX-License-Identifier

marcusfolkesson:
SPDX License List is a list of (common) open source
licenses that can be referred to by a “short identifier”.
It has several advantages compared to the common "license header texts"
usually found in source files.

Some of the advantages:
* It is precise; there is no ambiguity due to variations in license header
  text
* It is language neutral
* It is easy to machine process
* It is concise
* It is simple and can be used without much cost in interpreted
* environments like java Script, etc.
* An SPDX license identifier is immutable.
* It provides simple guidance for developers who want to make sure the
  license for their code is respected

See http://spdx.org for further reading.


cgwalters:
As best I can tell, Linux at least (mostly?) hasn't hard switched to *only* having SPDX identifiers.  It looks like u-boot has though.  I'm going back and forth on this a little bit.

cgwalters:
I think I'm OK with this though.   Let's just see if anyone else has comments.

cgwalters:
bot, add author to whitelist

giuseppe:
would be using SPDX identifiers have the same legal implications as keeping the header?  I've looked around but could not find anything about using only SPDX

marcusfolkesson:
@cgwalters , no, not yet. We are currently working on it, but the process will take some time. The plan is to have SPDX identifiers in all source files for the kernel. Those files that is already tagged does only have the SPDX tag what I can tell?

See this article in the subject: https://lwn.net/Articles/739183/

@giuseppe . it has the same legal implications what I can tell. I have seen license texts combined with the SPDX-headers in some projects, but I think it is reduntant.

This is a good "intro" to SPDX:
https://spdx.org/sites/cpstandard/files/pages/files/using_spdx_license_list_short_identifiers.pdf

cgwalters:
For example that LWN article you linked to links to an arch/alpha file...but IMO the more useful link is this one:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=6f52b16c5b29b89d92c0e7236f4655dc8491ad70

There we can see it's just adding SPDX to files with *no* license.

I'd feel a lot more comfortable if we weren't near the tip of the spear on this.   I know there's a *few* other examples like u-boot but that's not a big list.  Hence I'm leaning more towards *adding* the SPDX tag rather than replacing the text.

cgwalters:
A lot of times when I'm wondering about best practices, I look at systemd.  And:
https://github.com/systemd/systemd/commit/53e1b683907c2f12330f00feb9630150196f064d

marcusfolkesson:
@cgwalters,  as stated in the commit you refered to: "SPDX license identifiers are a legally binding shorthand, which can be used instead of the full boiler plate text."

In my opinion, SPDX identifiers are preferred because it becomes clear which license is applicable.
The biggest problem I see with license texts is that they can be formulated differently and that it is not as clear if they differ at some point. The only thing to be sure is to read the whole text.
Anyway, ostree is consistent with the LGPL license so it is no real issue, although the license text is not an *exact* copy in all files,

For example:
Most license texts in in ostree has 
"This *program* is free software..." or 
"This *library* is free software...". 
src/libostree/ostree-deployment.c mixes both library and program in the text.

Also, the license text differ how to get the full license text, some files refers to www.gnu.org/licenses and some files want you to write a letter to Free Software Foundation.

A ridiculous thing that does not really matter for sure, but the point is that SPDX is both undoubtedly and consistent.

However, I'm fine with just adding SPDX-tags as well.

cgwalters:
It does indeed seem clearer to me that an SPDX-only future is probably going to happen.  I'm just uncertain about libostree being so near the tip of that.

Basically if someone convinced systemd to make this switch (or a subsystem in the Linux kernel) etc., we can revisit again?

SPDX for the docs seems like a no-brainer because the license situation is more complex, and in the end it's just the docs.  Licensing for the libostree *code* is a lot more important.

> However, I'm fine with just adding SPDX-tags as well.

OK, do you mind reworking this PR?

PR #1437, opened 2018-01-29
lib/core: Expand documentation for ostree_parse_refspec()

pwithnall:
The old documentation had outdated and incomplete annotations, and
didn’t make it very clear that out_remote could legitimately return
NULL.


cgwalters:
Hm, interesting that Travis failed so hard but PAPR is fine.  Anyways LGTM!

@rh-atomic-bot r+ 87bcaa9

Issue #960, opened 2017-06-25
Render mount source as /var


cgwalters:
Moved from: https://github.com/projectatomic/rpm-ostree/issues/849

https://github.com/ostreedev/ostree/pull/859 changed things so that the string `/sysroot/ostree/deploy` shows up in the mount source.  Offhand, fixing this would require something like a `RemovePathPrefix=/sysroot/ostree/deploy/fedora-atomic` tweak in systemd or so.

cgwalters:
I think this is probably should be fixed by https://github.com/ostreedev/ostree/pull/1438

PR #1438, opened 2018-01-29
switchroot: Ensure /sysroot is set to "private" propagation

cgwalters:
Downstream BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1498281

This came up as a problem with `oci-umount` which was trying to ensure some host
mounts like `/var/lib/containers` don't leak into privileged containers.  But
since our `/sysroot` mount wasn't private we also got a copy there.

We should have done this from the very start - it makes `findmnt` way, way less
ugly and is just the obviously right thing to do, will possibly create world
peace etc.

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -244,6 +244,17 @@ main(int argc, char *argv[])
         err (EXIT_FAILURE, "failed to MS_MOVE %s to %s", deploy_path, root_mountpoint);
     }
 
+  /* The /sysroot mount needs to be private to avoid having a mount for e.g. /var/cache
+   * also propagate to /sysroot/ostree/deploy/$stateroot/var/cache
+   *
+   * Now in reality, today this is overridden by systemd: the *actual* way we fix this up
+   * is in ostree-remount.c.  But let's do it here to express the semantics we want
+   * at the very start (perhaps down the line systemd will have compile/runtime option
+   * to say that the initramfs environment did everything right from the start).
+   */
+  if (mount ("none", "sysroot", NULL, MS_PRIVATE, NULL) < 0)
+    err (EXIT_FAILURE, "remounting 'sysroot' private");
+
   if (getpid() == 1)
     {
       execl ("/sbin/init", "/sbin/init", NULL);
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -44,6 +44,15 @@ main(int argc, char *argv[])
   struct stat stbuf;
   int i;
 
+  /* The /sysroot mount needs to be private to avoid having a mount for e.g. /var/cache
+   * also propagate to /sysroot/ostree/deploy/$stateroot/var/cache
+   *
+   * Today systemd remounts / (recursively) as shared, so we're undoing that as early
+   * as possible.  See also a copy of this in ostree-prepare-root.c.
+   */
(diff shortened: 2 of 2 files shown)


cgwalters:
See also https://github.com/ostreedev/ostree/issues/960

rhvgoyal:
Making this private will only prevent future mount propagation. But all the mounts will be visible when container is created?

cgwalters:
We're making the sysroot private *very* early in bootup, long before container systems are started.  That seemed to solve the problem of having two mounts in my testing; make sense?

rhvgoyal:
Still can't understand it. Even if you make it private very early, if a process later does unshare or clone, it will still see this mount point. Just that if it is private, then any further mounts under this mount will not propagate to other mount namespaces. So not sure how making it private made sure that process which are starting later will not see it or will not get copy of this mount upon clone or unshare.

cgwalters:
We could make it unbindable too - this basically seems to be the use case for `MS_UNBINDABLE`. 

But still, you said in:
https://bugzilla.redhat.com/show_bug.cgi?id=1498281#c12
> I think it is a problem where source mount is visible on host in multiple paths

And now the mount isn't visible in multiple paths, right?

rhvgoyal:
MS_UNBINDABLE will not work always. Reason being that libcontainer can change that mount property after doing clone(). IIRC, by default it makes the whole tree "MS_PRIVATE | MS_REC". And that means you will lose the MS_UNBINDABLE property on the mount you don't want to be visible.

rhvgoyal:
I guess we probably need something MS_UNCLONABLE, so that upon clone()/unshare(), certain mount and its children are not cloned in newly created copy of mount namespace. I am not aware of any such thing being available.

cgwalters:
I think what I'm getting at here is, I tested the reproducer from 
https://bugzilla.redhat.com/show_bug.cgi?id=1498281#c0
and this patch fixes it.  Is there another concrete case you're talking about, and if so, can you work up a docker/bwrap/etc. cmdline for it?  Or is this theoretical?

rhvgoyal:
If you just want to go by your test results, feel free to commit it. I am getting more at logical explanation of why you are seeing the results. 

We don't understand yet why your changes fixed the issue. And I think it is important to understand that.

Can you please help me understand how making a mount MS_PRIVATE will make sure it does not leak into container upon clone().

cgwalters:
> We don't understand yet why your changes fixed the issue. And I think it is important to understand that. Can you please help me understand how making a mount MS_PRIVATE will make sure it does not leak into container upon clone().

It's not about not leaking on `clone()` - the fix is to not have duplicate mounts in the first place.  Right?

rhvgoyal:
Yes, if you don't have duplicate mount on host itself that will solve the issue. Ok, so by making it MS_PRIVATE, it will not propagate to other place later I guess and that's how problem has been solved. If yes, that makes sense.

rhvgoyal:
I am assuming that when docker starts, it creates /var/lib/docker/overlay2 mount and it might be propagating under /sysroot/...... and then it leaks into container. So if some mount under /sysroot/.... is
made rprivate, I guess that would mean /var/lib/docker/overlay2 mount will not propagate to /sysroot/...
and that should solve the issue.

cgwalters:
@rh-atomic-bot delegate=rhvgoyal

rhvgoyal:
LGTM

rhvgoyal:
@rh-atomic-bot r+ 4bc681d

PR #1439, opened 2018-01-30
Add SPDX-License-Identifier to source files

marcusfolkesson:
SPDX License List is a list of (common) open source
licenses that can be referred to by a “short identifier”.
It has several advantages compared to the common "license header texts"
usually found in source files.

Some of the advantages:
* It is precise; there is no ambiguity due to variations in license header
  text
* It is language neutral
* It is easy to machine process
* It is concise
* It is simple and can be used without much cost in interpreted
  environments like java Script, etc.
* An SPDX license identifier is immutable.
* It provides simple guidance for developers who want to make sure the
  license for their code is respected

See http://spdx.org for further reading.

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -5,6 +5,8 @@
 <!--
 Copyright 2014 Colin Walters <walters@verbum.org>
 
+SPDX-License-Identifier: LGPL-2.0+
+
 This library is free software; you can redistribute it and/or
 modify it under the terms of the GNU Lesser General Public
 License as published by the Free Software Foundation; either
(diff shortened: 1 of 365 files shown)


man/ostree.repo-config.xml
```diff
@@ -5,6 +5,8 @@
 <!--
 Copyright 2014 Colin Walters <walters@verbum.org>
 
+SPDX-License-Identifier: LGPL-2.0+
```
cgwalters: Hmm, actually...the man pages should probably be under the docs license.  But...let's not try to do that relicensing now.

cgwalters:
@rh-atomic-bot r+ ccc7677

cgwalters:
Thanks!

PR #1441, opened 2018-02-01
lib/checkout: support filtering

jlebon:
This is analogous to the filtering support for the commit API: we allow
library users to skip over checking out specific files. This is useful
in some tricky situations where we *know* that the files to be checked
out will conflict with existing files in subtle ways.

One such example is in rpm-ostree support for multilib. There, we want
to allow checking out a package onto an existing tree, but skipping over
files that are not coloured to our preferred value (e.g. not overwriting
an i686 version of `ldconfig` if we already have the `x86_64` version).
See projectatomic/rpm-ostree#1227 for details.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -39,8 +39,8 @@
 /* Per-checkout call state/caching */
 typedef struct {
   GString *path_buf; /* buffer for real path if filtering enabled */
-  GString *selabel_path_buf; /* buffer for selinux path if labeling enabled */
-  gboolean shared_path_buf; /* we try to use the same buf if we can */
+  GString *selabel_path_buf; /* buffer for selinux path if labeling enabled; this may be
+                                the same buffer as path_buf */
 } CheckoutState;
 
 static void
@@ -48,7 +48,7 @@ checkout_state_clear (CheckoutState *state)
 {
   if (state->path_buf)
     g_string_free (state->path_buf, TRUE);
-  if (state->selabel_path_buf && !state->shared_path_buf)
+  if (state->selabel_path_buf && (state->selabel_path_buf != state->path_buf))
     g_string_free (state->selabel_path_buf, TRUE);
 }
 G_DEFINE_AUTO_CLEANUP_CLEAR_FUNC(CheckoutState, checkout_state_clear)
@@ -784,7 +784,7 @@ push_path_element (OstreeRepoCheckoutAtOptions *options,
 {
   if (state->path_buf)
     push_path_element_once (state->path_buf, name, is_dir);
-  if (state->selabel_path_buf && !state->shared_path_buf)
+  if (state->selabel_path_buf && (state->selabel_path_buf != state->path_buf))
     push_path_element_once (state->selabel_path_buf, name, is_dir);
 }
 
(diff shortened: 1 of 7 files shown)


src/libostree/ostree-repo.h
```diff
@@ -969,7 +997,9 @@ typedef struct {
   OstreeRepoDevInoCache *devino_to_csum_cache;
 
   int unused_ints[6];
-  gpointer unused_ptrs[5];
+  gpointer unused_ptrs[3];
+  OstreeRepoCheckoutFilter filter; /* Since: 2018.2 */
+  gpointer filter_user_data; /* Since: 2018.2 */
```
jlebon: I can transform those two into a single object, similar to `OstreeRepoCommitModifier`, if we want to conserve pointers.

jlebon:
This will be used by https://github.com/projectatomic/rpm-ostree/pull/1227. Marking as WIP for now for tests and initial feedback.

src/libostree/ostree-repo-checkout.c
```diff
@@ -750,6 +766,41 @@ checkout_one_file_at (OstreeRepo                        *repo,
...
+pop_path_element (OstreeRepoCheckoutAtOptions *options,
+                  CheckoutState               *state,
+                  const char                  *name,
+                  gboolean                     is_dir)
+{
+  const size_t n = strlen (name) + (is_dir ? 1 : 0);
+  if (state->path_buf)
+    g_string_truncate (state->path_buf, state->path_buf->len - n);
+  if (state->selabel_path_buf && !state->shared_path_buf)
+    g_string_truncate (state->selabel_path_buf, state->selabel_path_buf->len - n);
```
jlebon: Note that this mostly reverts #936. AFAICT, the algorithm before was good, there was just an off-by-one error to account for the added `/` when truncating after recursing into a subdir.

jlebon:
Added tests and split out some of the prep in #1442. Dropping WIP.

src/libostree/ostree-repo-checkout.c
```diff
@@ -992,18 +1055,28 @@ checkout_tree_at (OstreeRepo                        *self,
                   GError                           **error)
 {
   g_auto(CheckoutState) state = { 0, };
-  // If SELinux labeling is enabled, we need to keep track of the full path string
+
+  if (options->filter)
+    state.path_buf = new_canonicalized_path_buf (options->subpath);
```
jlebon: We can switch this to `/` so that we call back with the subpath prefixed removed already. Hmm, yeah that feels more natural actually.

jlebon:
bot, retest this please

jlebon:
This is ready for review!

src/libostree/ostree-repo-checkout.c
cgwalters: Yeah, makes sense.  Though

src/libostree/ostree-repo.h
cgwalters: Eh, feels unlikely we keep adding a lot more functionality to checkout.  If we do, I think I'd vote we add new high level APIs.

src/libostree/ostree-repo-checkout.c
```diff
@@ -38,13 +38,17 @@
...
+  gboolean shared_path_buf; /* we try to use the same buf if we can */
 } CheckoutState;
 
 static void
 checkout_state_clear (CheckoutState *state)
 {
-  if (state->selabel_path_buf)
+  if (state->path_buf)
+    g_string_free (state->path_buf, TRUE);
+  if (state->selabel_path_buf && !state->shared_path_buf)
```
cgwalters: We could also just check `state->path_buf != state->selabel_path_buf` and lose the boolean, right?

jlebon: True, done in ⬇️!

cgwalters:
@rh-atomic-bot r+ bb40284

jlebon:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #1443, opened 2018-02-02
Introduce new type PAYLOAD_LINK

giuseppe:
It is used to keep track of the payload checksum for files stored in the repository.
   
The goal is that files having the same payload but different xattrs can take advantage of reflinks where supported.

More details here: https://mail.gnome.org/archives/ostree-list/2018-January/msg00012.html

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -30,6 +30,8 @@
 #include <gio/gunixoutputstream.h>
 #include <sys/xattr.h>
 #include <glib/gprintf.h>
+#include <sys/ioctl.h>
+#include <linux/fs.h>
 
 #include "otutil.h"
 #include "ostree.h"
@@ -39,6 +41,12 @@
 #include "ostree-repo-file-enumerator.h"
 #include "ostree-checksum-input-stream.h"
 #include "ostree-varint.h"
+
+/* The standardized version of BTRFS_IOC_CLONE */
+#ifndef FICLONE
+#define FICLONE _IOW(0x94, 9, int)
+#endif
+
 
 /* If fsync is enabled and we're in a txn, we write into a staging dir for
  * commit, but we also allow direct writes into objects/ for e.g. hardlink
@@ -589,6 +597,192 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
   return TRUE;
 }
 
+static gboolean
+_check_support_reflink (OstreeRepo *self, gboolean *supported, GError **error)
+{
+  /* We have not checked yet if the file system supports reflinks, do it here */
+  if (g_atomic_int_get (&self->fs_support_reflink) == 0)
+    {
+      g_auto(GLnxTmpfile) src_tmpf = { 0, };
+      g_auto(GLnxTmpfile) dest_tmpf = { 0, };
+
+      if (!glnx_open_tmpfile_linkable_at (commit_tmp_dfd (self), ".", O_RDWR|O_CLOEXEC,
+                                          &src_tmpf, error))
+        return FALSE;
+      if (!glnx_open_tmpfile_linkable_at (commit_tmp_dfd (self), ".", O_WRONLY|O_CLOEXEC,
+                                          &dest_tmpf, error))
+        return FALSE;
+
+      if (ioctl (dest_tmpf.fd, FICLONE, src_tmpf.fd) == 0)
+        g_atomic_int_set (&self->fs_support_reflink, 1);
+      else if (errno == EOPNOTSUPP)  /* Ignore other kind of errors as they might be temporary failures */
+        g_atomic_int_set (&self->fs_support_reflink, -1);
+    }
+  *supported = g_atomic_int_get (&self->fs_support_reflink) >= 0;
+  return TRUE;
+}
+
+static gboolean
+_create_payload_link (OstreeRepo   *self,
+                      const char   *checksum,
+                      const char   *payload_checksum,
+                      GFileInfo    *file_info,
+                      GCancellable *cancellable,
+                      GError       **error)
+{
+  gboolean reflinks_supported = FALSE;
+  if (!_check_support_reflink (self, &reflinks_supported, error))
+    return FALSE;
+
+  if (!reflinks_supported)
+    return TRUE;
+
+  if (g_file_info_get_file_type (file_info) != G_FILE_TYPE_REGULAR
+      || !G_IN_SET(self->mode, OSTREE_REPO_MODE_BARE, OSTREE_REPO_MODE_BARE_USER, OSTREE_REPO_MODE_BARE_USER_ONLY))
+    return TRUE;
+
+  if (payload_checksum == NULL || g_file_info_get_size (file_info) < self->payload_link_threshold)
+    return TRUE;
+
+  char target_buf[_OSTREE_LOOSE_PATH_MAX + _OSTREE_PAYLOAD_LINK_PREFIX_LEN];
+  strcpy (target_buf, _OSTREE_PAYLOAD_LINK_PREFIX);
+  _ostree_loose_path (target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN, checksum, OSTREE_OBJECT_TYPE_FILE, self->mode);
+
+  if (symlinkat (target_buf, commit_tmp_dfd (self), payload_checksum) < 0)
+    {
+      if (errno != EEXIST)
+        return glnx_throw_errno_prefix (error, "symlinkat");
+    }
+  else
+    {
+      g_auto(OtCleanupUnlinkat) tmp_unlinker = { commit_tmp_dfd (self), g_strdup (payload_checksum) };
+      if (!commit_path_final (self, payload_checksum, OSTREE_OBJECT_TYPE_PAYLOAD_LINK, &tmp_unlinker, cancellable, error))
+        return FALSE;
+    }
+
+  return TRUE;
+}
+
+static gboolean
+_import_payload_link (OstreeRepo   *self,
+                      OstreeRepo   *source,
+                      const char   *checksum,
+                      GCancellable *cancellable,
+                      GError       **error)
+{
+  gboolean reflinks_supported = FALSE;
+  g_autofree char *payload_checksum = NULL;
+  g_autoptr(GInputStream) is = NULL;
+  glnx_unref_object OtChecksumInstream *checksum_payload = NULL;
+  g_autoptr(GFileInfo) file_info = NULL;
+
+  if (!_check_support_reflink (self, &reflinks_supported, error))
+    return FALSE;
+
+  if (!reflinks_supported)
+    return TRUE;
+
+  if (!G_IN_SET(self->mode, OSTREE_REPO_MODE_BARE, OSTREE_REPO_MODE_BARE_USER, OSTREE_REPO_MODE_BARE_USER_ONLY))
+    return TRUE;
+
+  if (!ostree_repo_load_file (source, checksum, &is, &file_info, NULL, cancellable, error))
+    return FALSE;
+
+  if (g_file_info_get_file_type (file_info) != G_FILE_TYPE_REGULAR
+      || g_file_info_get_size (file_info) < self->payload_link_threshold)
+    return TRUE;
+
+  checksum_payload = ot_checksum_instream_new (is, G_CHECKSUM_SHA256);
+
+  guint64 remaining = g_file_info_get_size (file_info);
+  while (remaining)
+    {
+      char buf[8192];
+      gssize ret = g_input_stream_read ((GInputStream *) checksum_payload, buf,
+                                        MIN (sizeof (buf), remaining), cancellable, error);
+      if (ret < 0)
+        return FALSE;
+      remaining -= ret;
+    }
+  payload_checksum = ot_checksum_instream_get_string (checksum_payload);
+
+  return _create_payload_link (self, checksum, payload_checksum, file_info, cancellable, error);
+}
+
+static gboolean
+_try_clone_from_payload_link (OstreeRepo   *self,
+                              const char   *payload_checksum,
+                              GFileInfo    *file_info,
+                              GLnxTmpfile  *tmpf,
+                              GCancellable *cancellable,
+                              GError       **error)
+{
+  gboolean reflinks_supported = FALSE;
+  int dfd_searches[] = { -1, self->objects_dir_fd };
+  if (self->commit_stagedir.initialized)
+    dfd_searches[0] = self->commit_stagedir.fd;
+
+  if (!_check_support_reflink (self, &reflinks_supported, error))
+    return FALSE;
+
+  if (!reflinks_supported)
+    return TRUE;
+
+  for (guint i = 0; i < G_N_ELEMENTS (dfd_searches); i++)
+    {
+      glnx_autofd int fdf = -1;
+      char loose_path_buf[_OSTREE_LOOSE_PATH_MAX];
+      char loose_path_target_buf[_OSTREE_LOOSE_PATH_MAX];
+      char target_buf[_OSTREE_LOOSE_PATH_MAX + _OSTREE_PAYLOAD_LINK_PREFIX_LEN];
+      char target_checksum[OSTREE_SHA256_STRING_LEN+1];
+      int dfd = dfd_searches[i];
+      ssize_t size;
+      if (dfd == -1)
+        continue;
+
+      _ostree_loose_path (loose_path_buf, payload_checksum, OSTREE_OBJECT_TYPE_PAYLOAD_LINK, self->mode);
+
+      size = TEMP_FAILURE_RETRY (readlinkat (dfd, loose_path_buf, target_buf, sizeof (target_buf)));
+      if (size < 0)
+        {
+          if (errno == ENOENT)
+            continue;
+          return glnx_throw_errno_prefix (error, "readlinkat");
+        }
+
+      if (size < OSTREE_SHA256_STRING_LEN + _OSTREE_PAYLOAD_LINK_PREFIX_LEN)
+        return glnx_throw (error, "invalid data size for %s", loose_path_buf);
+
+      sprintf (target_checksum, "%.2s%.62s", target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN, target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN + 3);
+
+      _ostree_loose_path (loose_path_target_buf, target_checksum, OSTREE_OBJECT_TYPE_FILE, self->mode);
+      if (!ot_openat_ignore_enoent (dfd, loose_path_target_buf, &fdf, error))
+        return FALSE;
+
+      if (fdf < 0)
+        {
+          /* If the link is referring to an object that doesn't exist anymore in the repository, just unlink it.  */
+          if (!glnx_unlinkat (dfd, loose_path_buf, 0, error))
+            return FALSE;
+        }
+      else
+        {
+          /* This undoes all of the previous writes; we want to generate reflinked data.   */
+          if (ftruncate (tmpf->fd, 0) < 0)
+            return glnx_throw_errno_prefix (error, "ftruncate");
+
+          if (glnx_regfile_copy_bytes (fdf, tmpf->fd, -1) < 0)
+            return glnx_throw_errno_prefix (error, "regfile copy");
+
+          return TRUE;
+        }
+    }
+  if (self->parent_repo)
+    return _try_clone_from_payload_link (self->parent_repo, payload_checksum, file_info, tmpf, cancellable, error);
+
+  return TRUE;
+}
+
 /* The main driver for writing a content (regfile or symlink) object.
  * There are a variety of tricky cases here; for example, bare-user
  * repos store symlinks as regular files.  Computing checksums
@@ -616,6 +810,8 @@ write_content_object (OstreeRepo         *self,
   GInputStream *file_input; /* Unowned alias */
   g_autoptr(GInputStream) file_input_owned = NULL; /* We need a temporary for bare-user symlinks */
   glnx_unref_object OtChecksumInstream *checksum_input = NULL;
+  glnx_unref_object OtChecksumInstream *checksum_payload_input = NULL;
+  const GFileType object_file_type = g_file_info_get_file_type (file_info);
   if (out_csum)
     {
       /* Previously we checksummed the input verbatim; now
@@ -624,6 +820,7 @@ write_content_object (OstreeRepo         *self,
        * it's not that's not a serious problem because we're still computing a
        * checksum over the data we actually use.
        */
+      gboolean reflinks_supported = FALSE;
       g_autoptr(GBytes) header = _ostree_file_header_new (file_info, xattrs);
       size_t len;
       const guint8 *buf = g_bytes_get_data (header, &len);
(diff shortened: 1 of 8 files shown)


cgwalters:
I wonder if it'd be simpler to add a special xattr that we know about and also filter out.

giuseppe:
how could we use the xattr to lookup for the payload checksum?

cgwalters:
> how could we use the xattr to lookup for the payload checksum?

Yeah, that's an issue.  We could scan all objects, but that'd get slow unless amortized.

So...hmm.  A big picture question here in my mind is the (still unsettled?) degree to which libostree provides low-level APIs for the OCI case, versus doing things at a higher level.

If we provide high level APIs, we're more free to change/optimize the implementation details later.

One other random thought here - how about only doing this for say files over 5MB (or some configurable threshold) instead?  Trading off indexing overhead versus space savings.

Another issue here is I think it needs to be opt-in; otherwise we're taking up extra space for "pure libostree" users who aren't doing containers/rpms/whatever on top.

giuseppe:
yes good point.  I've added a new commit that implements a minimum threshold configuration, the payload link will be created only for files that have a size greater or equal to the threshold.

Are you fine with a default value of 3 MiB?

If you are for disabling this feature by default, we can setup a much bigger value by default, that has the same effect.

---
About OCI, we should probably have a bigger discussion around it.  If we deal with it directly in OSTree, we would end up duplicating a lot of functionalities that are already in `containers/image`.  For a third party project can be painful to interact both with Golang and C but having the OCI bits in `containers/image` has its advantages.  OSTree storage can use/be used with all the other tools that are already using `containers/image`, such as Buildah.

One feature I particularly like is:

https://github.com/containers/image/pull/392

In addition to copy images to the OSTree storage, we can also copy them back to other storages, like to the Docker engine or to a registry.

The OSTree storage part of `containers/image` even if used only for system containers, is a direct mapping from OCI to OSTree, that can be used for any kind of OCI images.  There are no system containers features in the storage part.

giuseppe:
Any feedback on the design?

cgwalters:
> Are you fine with a default value of 3 MiB?

But that's still imposing an unnecessary cost on everyone who is doing "pure ostree" embedded devices, etc.  We could have it be a repo flag like
```
$ cat /ostree/repo/config 
[core]
repo_version=1
mode=bare
contentindex=true
```

or something.  And for Project Atomic systems we'd enable that flag.

But the configuration entrypoint aside, my main concern here is we're basically *doubling* the number of files we create; on this workstation:

```
$ find /ostree/repo/objects/ -name *.file |wc -l
150393
$ find /ostree/repo/objects/ |wc -l
169358
```

And like I said before I'm already not totally happy with how many small metadata files we have.  In fact it'd be interesting to investigate not hardlinking for small objects at all.  Is it really worth it to do hardlinks for e.g. < 1k files?  Less pressure on the filesystem journal to do inode updates?

Another concern is that now pruning objects isn't atomic; it looks like you made the commit process ignore "dangling" payload links, but it should probably delete them so they can be recreated.

cgwalters:
Also big picture, we know this is an issue for SELinux systems, but what about SMACK/AppArmor?  IIRC SMACK uses xattrs but I don't know how aggressive it is with distinct labels like SELinux is.

TBH I don't think it's too worth your time digging into SMACK/AppArmor, but there are definitely ostree users who don't use SELinux, and in that case again we don't need the context indexing, right?

giuseppe:
the number of files >3MiB is a small subset of all the files, this is what I have in the my local repository:
```
$ find -name '*.file' | wc -l
9603
$ find -name '*.payload-link' | wc -l
10
```
but anyway, it still requires the support from the file system to be useful, so I will change the default to `infinite` for now, or do you prefer a different option?

Yes I am not happy as well about the pruning algorithm, do you think we should recalculate everything?  I didn't implement this way as it looks quite expensive (prune will be as expensive as fsck), maybe I can  delete them all and recreate only those for files that are still present in the repository after the pruning.

cgwalters:
Oh wow, you're right...that's actually pretty amazing; on my current workstation the ratio of "3MB+" objects is `211/150393` or `0.1%`.  At a quick glance there's statically linked golang binaries, but also fonts, qemu for some reason, and the kernel/initramfs.

cgwalters:
And on a stock `fedora-atomic:fedora/27/x86_64/atomic-host 27.61 (2018-01-17 15:52:47)` with no packages layered, the ratio is `40/26030`, also at `0.1%`.

cgwalters:
OK so there's a higher level issue here too I just thought of: this code only has an effect for commits created locally (as we do when importing OCI).  When we're pulling objects directly via libostree (`ostree pull`, "rpm-ostree jigdo"), the client won't get this data.

So we'd have to either add this to pull (which in archive fetches would double the number of http requests, probably a non-starter), or add it to some lookaside data (messy).  We can include it in static deltas easily enough though, and I could probably easily enough teach jigdo how to do it too.

cgwalters:
Though hmm...since we're computing checksums at pull time now anyways usually, we could just do two SHA-256 checksums simultaneously.

giuseppe:
I checked "ostree pull" and this code path also creates the `.payload-link`.  I see this like a local optimization that should not go on the wire.  Also since we allow to change the threshold for what file size must be considered, we will need the client to check if the received `.payload-link` must be discarded as it might point to a file bigger than `core.`payload-link-threshold`.

I am still chasing down the failure in the CI, although I've done some changes in the PR:

1. if the symlink is dangling then it is unlinked.
2. I've implemented a lookup in the parent repository
3. At pull time we always check if the link must be created, even for object files that are already present in the repository

giuseppe:
tests pass again ✌️

src/libostree/ostree-repo-prune.c
```diff
@@ -233,6 +260,7 @@ repo_prune_internal (OstreeRepo        *self,
   g_autoptr(GHashTable) reachable_owned = g_hash_table_ref (options->reachable);
   data.reachable = reachable_owned;
 
+
```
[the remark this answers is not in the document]
giuseppe: dropped in the new version

src/libostree/ostree-repo-commit.c
```diff
@@ -633,13 +717,19 @@ write_content_object (OstreeRepo         *self,
         null_input = input = g_memory_input_stream_new_from_data ("", 0, NULL);
       checksum_input = ot_checksum_instream_new_with_start (input, G_CHECKSUM_SHA256,
                                                             buf, len);
-      file_input = (GInputStream*)checksum_input;
+
+      if (xattrs == NULL || self->mode != OSTREE_REPO_MODE_BARE || object_file_type != G_FILE_TYPE_REGULAR)
+        file_input = (GInputStream*)checksum_input;
+      else
+        {
+          checksum_payload_input = ot_checksum_instream_new ((GInputStream*)checksum_input, G_CHECKSUM_SHA256);
```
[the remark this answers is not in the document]
giuseppe: added

giuseppe:
I've added a patch and "pull-local" also creates the `.payload-link` files now

giuseppe:
my only doubt is that I use the checksum for the target in the symlink.  Should the symlink have the target instead like `../../ab/cdef01234....file`?

src/libostree/ostree-repo-commit.c
```diff
@@ -589,6 +589,132 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
...
+      if (dfd == -1)
+        continue;
+
+      _ostree_loose_path (loose_path_buf, actual_payload_checksum, OSTREE_OBJECT_TYPE_PAYLOAD_LINK, self->mode);
+
+      if (TEMP_FAILURE_RETRY (readlinkat (dfd, loose_path_buf, target_checksum, sizeof (target_checksum))) < 0)
+        {
+          if (errno == ENOENT)
+            continue;
+          return glnx_throw_errno (error);
```
cgwalters: `glnx_throw_errno_prefix (error, "readlinkat")`

src/libostree/ostree-repo-commit.c
```diff
@@ -4137,6 +4288,13 @@ _ostree_repo_import_object (OstreeRepo           *self,
         return FALSE;
     }
 
+ exit:
```
cgwalters: Then this one can be put under the `if (has_object)` conditional above and we can avoid a goto?  (Trying to reduce the number of `goto`s here)

tests/pull-test.sh
```diff
@@ -313,6 +313,14 @@ ${CMD_PREFIX} ostree --repo=parentpullrepo rev-parse origin:main > main.txt
 assert_file_has_content main.txt ${rev}
 echo "ok pull specific commit"
 
+repo_init --no-gpg-verify
+${CMD_PREFIX} ostree config --repo=repo set core.payload-link-threshold 0
+${CMD_PREFIX} ostree --repo=repo pull origin main 2>err.txt
```
cgwalters: Let's also add a test that verifies running `sha256sum` against a file directly has the same content as the object's payload link?

src/libostree/ostree-repo-prune.c
```diff
@@ -46,18 +46,43 @@ maybe_prune_loose_object (OtPruneData        *data,
...
+                  if (!ostree_repo_query_object_storage_size (data->repo, OSTREE_OBJECT_TYPE_FILE, target_checksum,
+                                                              &storage_size, cancellable, error))
+                    return FALSE;
+
+                  reachable = storage_size >= data->repo->payload_link_threshold;
+                  if (reachable)
+                    goto exit;
+                }
+            }
           if (objtype == OSTREE_OBJECT_TYPE_COMMIT)
```
cgwalters: nit: This should be `else if` now right?

src/libostree/ostree-repo-prune.c
```diff
@@ -46,18 +46,43 @@ maybe_prune_loose_object (OtPruneData        *data,
...
           guint64 storage_size = 0;
 
+          if (objtype == OSTREE_OBJECT_TYPE_PAYLOAD_LINK)
+            {
+              char loose_path_buf[_OSTREE_LOOSE_PATH_MAX];
+              char target_checksum[OSTREE_SHA256_STRING_LEN+1];
+
+              _ostree_loose_path (loose_path_buf, checksum, OSTREE_OBJECT_TYPE_PAYLOAD_LINK, data->repo->mode);
+              if (readlinkat (data->repo->objects_dir_fd, loose_path_buf, target_checksum, sizeof (target_checksum)) < 0)
+                return glnx_throw_errno (error);
```
cgwalters: `prefix`

src/libostree/ostree-repo-prune.c
```diff
@@ -46,18 +46,43 @@ maybe_prune_loose_object (OtPruneData        *data,
...
+
+              g_autoptr(GVariant) target_key = ostree_object_name_serialize (target_checksum, OSTREE_OBJECT_TYPE_FILE);
+
+              if (g_hash_table_lookup_extended (data->reachable, target_key, NULL, NULL))
+                {
+                  if (!ostree_repo_query_object_storage_size (data->repo, OSTREE_OBJECT_TYPE_FILE, target_checksum,
```
cgwalters: Should also handle `ENOENT` here right?

src/libostree/ostree-repo-commit.c
```diff
@@ -589,6 +589,132 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
...
+
+      if (fdf < 0)
+        {
+          /* If the link is referring to an object that doesn't exist anymore in the repository, just unlink it.  */
+          if (!glnx_unlinkat (dfd, loose_path_buf, 0, error))
+            return FALSE;
+        }
+      else
+        {
+          if (ftruncate (tmpf->fd, 0) < 0)
```
cgwalters: Sort of unfortunate to have done the whole write until now, then toss it but I can't think of a better way.  This could at least use a comment like `/* This undoes all of the previous writes; we want to generate reflinked data */`.

Actually also...if the filesystem doesn't have reflinks, this whole thing isn't actually useful right?  It feels to me like we need to detect this property for the repo and conditionalize use of payload links at all based on that?

cgwalters:
> my only doubt is that I use the checksum for the target in the symlink. Should the symlink have the target instead like ../../ab/cdef01234....file?

Yeah that's probably nicer; we're adding a tiny bit more data on disk and now we need more "parsing" when reading the symlinks, but then the symlinks are actually useful for looking at with `coreutils`.

cgwalters:
In general I personally like splitting up commits in terms of "prep" vs "functionality", whereas here you're adding "part functionality, part functionality, ..." etc.  In particular I personally prefer having tests included with the commit that adds the functionality.

The whole thing really goes together; it doesn't make sense to have payload links written without a way to GC them, etc. right?

src/libostree/ostree-repo-commit.c
```diff
@@ -4189,7 +4228,7 @@ _ostree_repo_import_object (OstreeRepo           *self,
 
       /* If direct import succeeded, we're done! */
       if (direct_was_supported)
-        return TRUE;
+        goto exit;
```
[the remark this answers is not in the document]
giuseppe: yes, I will add the `_import_payload_link()` to `import_one_object_direct()`

src/libostree/ostree-repo-commit.c
giuseppe: by default this option is disabled, so it must be manually enabled before we start looking at the payload

src/libostree/ostree-repo-prune.c
giuseppe: hm.. in that case it won't be in the `data->reachable` hash table?

giuseppe: I've addressed all the other comments, except this one and checking for the reflink support in the file system

giuseppe:
About the check for `reflinks`, what about something like?

```diff
diff --git a/src/libostree/ostree-repo-commit.c b/src/libostree/ostree-repo-commit.c
index 96717669..c10b8661 100644
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -30,6 +30,8 @@
 #include <gio/gunixoutputstream.h>
 #include <sys/xattr.h>
 #include <glib/gprintf.h>
+#include <sys/ioctl.h>
+#include <linux/fs.h>
 
 #include "otutil.h"
 #include "ostree.h"
@@ -40,6 +42,12 @@
 #include "ostree-checksum-input-stream.h"
 #include "ostree-varint.h"
 
+/* The standardized version of BTRFS_IOC_CLONE */
+#ifndef FICLONE
+#define FICLONE _IOW(0x94, 9, int)
+#endif
+
+
 /* If fsync is enabled and we're in a txn, we write into a staging dir for
  * commit, but we also allow direct writes into objects/ for e.g. hardlink
  * imports.
@@ -673,6 +681,28 @@ _try_clone_from_payload_link (OstreeRepo   *self,
   int dfd_searches[] = { -1, self->objects_dir_fd };
   if (self->commit_stagedir.initialized)
     dfd_searches[0] = self->commit_stagedir.fd;
+
+  /* We have not checked yet if the file system supports reflinks, do it here */
+  if (self->payload_fs_support_reflink == 0)
+    {
+      g_auto(GLnxTmpfile) src_tmpf = { 0, };
+      g_auto(GLnxTmpfile) dest_tmpf = { 0, };
+      if (!glnx_open_tmpfile_linkable_at (commit_tmp_dfd (self), ".", O_RDWR|O_CLOEXEC,
+                                          &src_tmpf, error))
+        return FALSE;
+      if (!glnx_open_tmpfile_linkable_at (commit_tmp_dfd (self), ".", O_WRONLY|O_CLOEXEC,
+                                          &dest_tmpf, error))
+        return FALSE;
+
+      if (ioctl (dest_tmpf.fd, FICLONE, src_tmpf.fd) == 0)
+        self->payload_fs_support_reflink = 1;
+      else if (errno == EOPNOTSUPP)
+        self->payload_fs_support_reflink = -1;
+      /* Ignore other kind of errors as they might be temporary failures */
+    }
+  if (self->payload_fs_support_reflink < 0)
+    return TRUE;
+
   for (guint i = 0; i < G_N_ELEMENTS (dfd_searches); i++)
     {
       glnx_autofd int fdf = -1;
@@ -775,7 +805,8 @@ write_content_object (OstreeRepo         *self,
       checksum_input = ot_checksum_instream_new_with_start (input, G_CHECKSUM_SHA256,
                                                             buf, len);
 
-      if (xattrs == NULL || !G_IN_SET(self->mode, OSTREE_REPO_MODE_BARE, OSTREE_REPO_MODE_BARE_USER, OSTREE_REPO_MODE_BARE_USER_ONLY) || object_file_type != G_FILE_TYPE_REGULAR)
+      if (xattrs == NULL || !G_IN_SET(self->mode, OSTREE_REPO_MODE_BARE, OSTREE_REPO_MODE_BARE_USER, OSTREE_REPO_MODE_BARE_USER_ONLY) || object_file_type != G_FILE_TYPE_REGULAR ||
+          self->payload_fs_support_reflink < 0)
         file_input = (GInputStream*)checksum_input;
       else
         {

```

If it is fine, I'd like to implement this as a follow-up patch, as I think it is better implemented in glnx.

Alternatively, what about changing `glnx_regfile_copy_bytes (int fdf, int fdt, off_t max_bytes)` to `glnx_regfile_copy_bytes (int fdf, int fdt, gboolean *was_cloned, off_t max_bytes)` so we can move the check to the first use of `ioctl(.., FICLONE, ...)`?

cgwalters:
> If it is fine, I'd like to implement this as a follow-up patch, as I think it is better implemented in glnx.

Mmm...it's not that much code to just do the standardized `FICLONE` ioctl, it looks pretty sane to me as is without doing new API in libglnx, though I wouldn't have a problem wrapping it there either.

Extending the copy API feels like it involves a lot of churn for other callers just for this special case.

So this approach looks good to me.  The only thing I'd comment on here is that right now it's OK to use an `OstreeRepo` from multiple threads, and at the moment it's possible for them to race on setting the flag.  Today, that race is fairly harmless, but I'd probably add a mutex for this so it helps people remember to be threadsafe in general when operating on the repo.

giuseppe:
hm.. would that be a problem?  In the worse case we will end up having the check happening at the same time in multiple threads.  They will (hopefully!) get the same result after the check.

cgwalters:
It's not a problem today, but as I said:

>  Today, that race is fairly harmless, but I'd probably add a mutex for this so it helps people remember to be threadsafe in general when operating on the repo.

giuseppe:
I've amended the change in the last version here :arrow_up:

src/libostree/ostree-repo-commit.c
```diff
@@ -589,6 +597,169 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
...
+                              GLnxTmpfile  *tmpf,
+                              GCancellable *cancellable,
+                              GError       **error)
+{
+  int dfd_searches[] = { -1, self->objects_dir_fd };
+  if (self->commit_stagedir.initialized)
+    dfd_searches[0] = self->commit_stagedir.fd;
+
+  /* We have not checked yet if the file system supports reflinks, do it here */
+  if (self->payload_fs_support_reflink == 0)
```
cgwalters: Tools like TSAN are going to warn if we read a variable without a mutex, and write with it locked.  It may be easier to use `g_atomic_int_*()`.

src/libostree/ostree-repo-commit.c
```diff
@@ -589,6 +597,169 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
...
+
+static gboolean
+_try_clone_from_payload_link (OstreeRepo   *self,
+                              const char   *payload_checksum,
+                              GFileInfo    *file_info,
+                              GLnxTmpfile  *tmpf,
+                              GCancellable *cancellable,
+                              GError       **error)
+{
+  int dfd_searches[] = { -1, self->objects_dir_fd };
```
cgwalters: Shouldn't we also check the `file_info` here for the same conditions used for `_import_payload_link()`?

src/libostree/ostree-repo-private.h
```diff
@@ -161,6 +164,8 @@ struct OstreeRepo {
   gchar *collection_id;
   gboolean add_remotes_config_dir; /* Add new remotes in remotes.d dir */
   gint lock_timeout_seconds;
+  guint64 payload_link_threshold;
+  int payload_fs_support_reflink; /* The underlying filesystem has support for ioctl (FICLONE..) */
```
cgwalters: Minor bikeshed; how about just `fs_supports_reflink` ?  (It might be useful to know in another circumstance)

src/libostree/ostree-repo-commit.c
giuseppe: yes sure, I'll change it.  Are you fine if I use `g_atomic_int_set ()` instead of the mutex?

cgwalters: Yep that sounds good to me, I should have thought of that first vs the mutex.

src/libostree/ostree-repo-commit.c
giuseppe: I didn't do here as the only caller `write_content_object` of this function already has the check.  The advantage of doing it earlier in `write_content_object` is that we don't need to calculate the payload checksum when we are sure we won't need it.

src/libostree/ostree-repo-private.h
giuseppe: yes, will rename it

src/libostree/ostree-repo-commit.c
cgwalters: It looks to me that `write_content_object()` calls this too.  Right?

giuseppe: yes, `write_content_object()` has the check at line 808:

https://github.com/ostreedev/ostree/pull/1443/files/00482eb2564d103d3d3d220f7061f5e01cf30988..9283d3c128f28cd6e4e856624f7c6f620e390189#diff-7833e28ecd7891cb43d466346f39acbdR810

If we are not going to store a reflink, then the payload checksum is not even calculated

cgwalters: Ah yes, got it, thanks!

giuseppe:
the fixup addresses the last comments

tests/pull-test.sh
```diff
@@ -313,6 +313,24 @@ ${CMD_PREFIX} ostree --repo=parentpullrepo rev-parse origin:main > main.txt
...
+if test `find repo -name '*.payload-link' | wc -l` = 0; then
+    fatal ".payload-link files not found"
```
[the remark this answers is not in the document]
giuseppe: ah, I forgot to check for the reflink support also when creating a file.  I've added a fixup patch to cover this case as well.

How could we cover these tests now?  I needed to add a check as well now, as if reflinks are not supported then we don't get any `.payload-link` file.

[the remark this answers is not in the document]
cgwalters: In today's test framework the best way to do this is to add a `test/installed/test-payload-link.sh`.  To ensure reflink support I'd create a loopback device and `mkfs.xfs -m reflink=1` on it, then make a repo there, etc.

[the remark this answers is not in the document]
giuseppe: thanks for the hint.  I've moved the tests to their own file.  In what check could I expect these tests to run?  All the log files I've checked skip this test as `losetup` fails

giuseppe:
@cgwalters are you fine to merge this?

cgwalters:
The test is being skipped right now though: https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/14c00f4e8527d8b81ffc6d06cca854b3fbd187e8.0.1520269933286777277/artifacts/gdtr-results/libostree_test-payload-link.sh.test.txt

I know our current testing setup is *very* confusing.  I'm working on addressing that (well, some things will be more confusing, others less so) in https://github.com/ostreedev/ostree/pull/1462

Anyways so what you want is to add a case to `tests/installed`, which are basically pure shell scripts that run as root on a FAH VM and not in a container.  I can probably take care of doing this if you prefer.

cgwalters:
The "dev flow" I have for the `tests/installed` stuff is basically to hack in my dev container, then sync my git repo into the vm, and just run the script there.

giuseppe:
> Anyways so what you want is to add a case to tests/installed, which are basically pure shell scripts that run as root on a FAH VM and not in a container. I can probably take care of doing this if you prefer.

I've pushed a new version with the installed test script.  I hope it will pass the CI :-)

cgwalters:
Looks like not, see https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/7a41a77d356fa518b0f3edb8124f8c7a20315bf2.8.1520324472405726051/output.log

giuseppe:
@cgwalters finally managed to get all the tests happy again :-)

cgwalters:
Thanks for all of your work on this!

@rh-atomic-bot r+ 6994f22

alexlarsson:
This regressed flatpak, see https://github.com/ostreedev/ostree/pull/1524 for fix

Issue #1429, opened 2018-01-19
Split source/build fails to build static ostree-prepare-root


akiernan:
If you have split source and build directories, together with static ostree-prepare-root, then the build fails to find the source files.

Not convinced this fix is really "right", but it does work around the problem:

```diff --git a/Makefile-switchroot.am b/Makefile-switchroot.am
index 70aa1c8..b1ac1a6 100644
--- a/Makefile-switchroot.am
+++ b/Makefile-switchroot.am
@@ -43,7 +43,7 @@ if BUILDOPT_USE_STATIC_COMPILER
 ostree_boot_SCRIPTS = ostree-prepare-root

 ostree-prepare-root : $(ostree_prepare_root_SOURCES)
-	$(STATIC_COMPILER) -o $@ -static $(ostree_prepare_root_SOURCES) $(AM_CPPFLAGS) $(AM_CFLAGS) $(DEFAULT_INCLUDES)
+	$(STATIC_COMPILER) -o $@ -static $(srcdir)/src/switchroot/ostree-prepare-root.c $(AM_CPPFLAGS) $(AM_CFLAGS) $(DEFAULT_INCLUDES)
 else
 ostree_boot_PROGRAMS += ostree-prepare-root
 ostree_prepare_root_CFLAGS = $(AM_CFLAGS) -Isrc/switchroot```

cgwalters:
@akiernan I'd `s/$(srcdir)/$(top_srcdir)/` just to match the other custom rules.  Otherwise, this looks right to me; care to make a PR?

cgwalters:
Fixed by https://github.com/ostreedev/ostree/pull/1445

PR #1445, opened 2018-02-05
switchroot: Fix split source/build directory

akiernan:
If you have split source and build directories, then building
static ostree-prepare-root fails to find the source files.

https://github.com/ostreedev/ostree/issues/1429


cgwalters:
Thanks!

@rh-atomic-bot r+ aec8e28

cgwalters:
Looks like flakes.

@rh-atomic-bot retry

PR #1446, opened 2018-02-06
repo: Create uncompressed-object-cache dir dynamically

cgwalters:
Having the `uncompressed-object-cache` directory in `archive` repos by default
is clutter; the functionality should be considered deprecated.

Now we only create the directory if we're doing a checkout with the cache
enabled.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -85,6 +85,17 @@ checkout_object_for_uncompressed_cache (OstreeRepo      *self,
   if (!glnx_fchmod (tmpf.fd, file_mode, error))
     return FALSE;
 
+  if (self->uncompressed_objects_dir_fd == -1)
+    {
+      if (!glnx_shutil_mkdir_p_at (self->repo_dir_fd, "uncompressed-objects-cache", 0755,
+                                   cancellable, error))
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-repo-checkout.c
```diff
@@ -1006,6 +1022,19 @@ checkout_tree_at (OstreeRepo                        *self,
       g_assert (options->force_copy);
     }
 
+  /* Uncompressed archive caches; should be considered deprecated */
```
jlebon: Can you explain why we need this chunk here? We'll try to create & open it on demand in `checkout_object_for_uncompressed_cache` anyway, right? Or is this just to get non-`ENOENT` failures early on? (In that case, let's mention that in the comment?)

cgwalters: We also need to handle the case of reusing an existing cachedir before we try to create objects there.

jlebon:
@rh-atomic-bot r+ 784cf08

cgwalters:
Hm, no obvious error message.

@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

Issue #1315, opened 2017-10-31
Consider "content hash"


cgwalters:
For a [promotion workflow](https://github.com/ostreedev/ostree-releng-scripts/blob/master/do-release-tags) it'd be nice if we added the concept of a "content hash" which was basically just `SHA256(dirtree + dirmeta)` of a commit's rootfs.

Then a promotion system could ensure that *content* hashes remain the same when promoting a commit, and deployment systems could also show this (we'd have an abbreviation like "chash"?).

What I'm a bit unsure about implementing is "pull by content hash"...maybe it'd end up in the summary file?

cgwalters:
What this issue is getting at is that one can't have *both*

 - Chained commit history (git, ostree default)
 - Content promotion

Now...something one could try today would be to forego commit history.  Basically just generate every commit without a parent; then promotion is simply changing the ref.  The downside of this is that I think you'd end up wanting some sort of other database to track "history".

cgwalters:
Bigger picture, in the rpm-ostree world (driven by some usability design from [Cockpit](http://cockpit-project.org/)) we've been focusing on version numbers as the thing that operators see - for a lot of reasons.  One interesting thing here is that most package systems (RPM,dpkg,etc.) as well as OCI don't have built-in history.  In the package system case there's a unique ID (NEVRA in the RPM world), and in the OCI case there's a digest as well as mutable "names" that act a lot like git/ostree branches/refs.  

But in both the pkg and Docker/OCI case, there's no built-in history/connection between independent builds.

abferm:
A problem with dropping history or refs with multiple commits is that you can no longer use the ostree admin tools effectively, because your origin ref will never have any new commits. I may be wrong, but I believe this would affect the ability to generate and use static deltas as well.

A potential solution that just jumped out to me is making commits track content and metadata, while refs/branches track history. In other words I am suggesting removing "parent" from a commit, and making a ref an ordered list of commits. This would allow moving commits between refs without modifying the commit itself

cgwalters:
Static deltas are between arbitrary commits - it doesn't actually require that one is a parent of another in any way.  For example today one can make *reverse* deltas for downgrade scenarios.  Another example is for e.g. Fedora Atomic Host, we make deltas between the last major branch e.g. `fedora/26/x86_64/atomic-host` and `fedora/27/x86_64/atomic-host`, even though those are distinct branches.

[...]

PR #1449, opened 2018-02-08
core: Add API (and standard concept for) content checksum

cgwalters:
There are a few cases for knowing whether a commit has identical
content to another commit.  Some people want to do a "promotion workflow",
where the content of a commit on a tesitng branch is then "promoted"
to a production branch with `ostree commit --tree=ref`.

Another use case I just hit in rpm-ostree deals with
[jigdo](https://github.com/projectatomic/rpm-ostree/issues/1081) where we're
importing RPMs on both the client and server, and will be using the
content checksum, since the client/server cases inject different metadata
into the commit object.

Closes: https://github.com/ostreedev/ostree/issues/1315

### DIFF
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -149,6 +149,7 @@ ostree_validate_structureof_dirtree
 ostree_validate_structureof_dirmeta
 ostree_commit_get_parent
 ostree_commit_get_timestamp
+ostree_commit_get_content_checksum
 ostree_check_version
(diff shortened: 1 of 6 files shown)


dustymabe:
> Closes: #131

is that right?

cgwalters:
> is that right?

Fixed ✓, thanks!

jlebon:
Nice!
@rh-atomic-bot r+ f14ad2a

PR #1450, opened 2018-02-08
lib/pull: Properly remove temporary remotes

mwleeds:
For P2P pulls ostree adds temporary remotes and removes them in
find_remotes_cb(). However, if an OstreeRepoFinderResult gets freed
during the course of that function, the OstreeRemote in the result is
freed but a pointer to it remains in the remotes_to_remove array. This
means that when _ostree_repo_remove_remote() gets called on it at the
end of the function it will fail. In my case the resulting error was
"OSTree-CRITICAL **: _ostree_repo_remove_remote: assertion 'remote->name
!= NULL' failed" but I think it could also seg fault.

This commit adds a reference to the remote so it can be properly removed
when we're finished with it.


mwleeds:
cc: @pwithnall

giuseppe:
LGTM

jlebon:
@rh-atomic-bot r=giuseppe cd2efe2

pwithnall:
Post-hoc: this looks good to me too!

Issue #1451, opened 2018-02-09
Max outstanding fetches assertion in soup


dbnicholson:
I have a job where I ask ostree to pull a lot of commits at once. I'm occasionally hitting an assertion in the soup fetcher:
```
OSTree:ERROR:src/libostree/ostree-fetcher-soup.c:387:start_pending_request: assertion failed (g_hash_table_size (thread_closure->outstanding) < thread_closure->max_outstanding): (24 < 24)
```
Looking at the pull code, I don't understand how this is happening. It seems like it starts queuing requests (rather than immediately starting them) as soon as you get to _OSTREE_MAX_OUTSTANDING_FETCHER_REQUESTS. How is it getting to 3 times that value?

cgwalters:
I think f4d1334e19ce3ab2f8872b1e28da52044f559401 missed deleting the assertions; looking at current master, nothing actually checks against `max_outstanding` in libsoup.

dbnicholson:
For the record, I think the real issue is that the deltas superblock fetches aren't being counted in the total number of fetches - https://github.com/ostreedev/ostree/pull/1453#discussion_r168321706.

PR #1453, opened 2018-02-14
fetcher/soup: Drop outdated max queue size assertion

cgwalters:
Since f4d1334e19ce3ab2f8872b1e28da52044f559401 the primary pull code maintains a
maximum queue. In that commit message I said `Note that I kept an assertion.`.
But I think this is wrong since while it covers a lot of the normal cases, if
one is e.g. trying to fetch a ton of refs, the primary pull code doesn't yet
queue those.  While it'd be nice to queue those, it isn't worth carrying
an extra assertion just in the libsoup side.

Closes: https://github.com/ostreedev/ostree/issues/1451

### DIFF
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -60,7 +60,6 @@ typedef struct {
   int base_tmpdir_dfd;
 
   GVariant *extra_headers;
-  int max_outstanding;
   gboolean transfer_gzip;
 
   /* Our active HTTP requests */
@@ -386,8 +385,6 @@ start_pending_request (ThreadClosure *thread_closure,
   OstreeFetcherPendingURI *pending;
   GCancellable *cancellable;
 
-  g_assert_cmpint (g_hash_table_size (thread_closure->outstanding), <, thread_closure->max_outstanding);
-
   pending = g_task_get_task_data (task);
   cancellable = g_task_get_cancellable (task);
 
@@ -473,7 +470,6 @@ ostree_fetcher_session_thread (gpointer data)
 {
   ThreadClosure *closure = data;
   g_autoptr(GMainContext) mainctx = g_main_context_ref (closure->main_context);
-  gint max_conns;
 
   /* This becomes the GMainContext that SoupSession schedules async
    * callbacks and emits signals from.  Make it the thread-default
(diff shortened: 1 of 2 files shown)


cgwalters:

src/libostree/ostree-fetcher-soup.c
```diff
@@ -386,8 +385,6 @@ start_pending_request (ThreadClosure *thread_closure,
   OstreeFetcherPendingURI *pending;
   GCancellable *cancellable;
 
-  g_assert_cmpint (g_hash_table_size (thread_closure->outstanding), <, thread_closure->max_outstanding);
```
dbnicholson: While I'm generally happy to remove an `assert` that the code is not doing anything to prevent, note that the curl fetcher [has a similarly hardcoded assertion](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-fetcher-curl.c#L870).

I'm also curious why the fetcher is getting to 24 requests when the pull queue code is supposed to stop at 8. Oh, I think I might see it. Most fetches check `fetcher_queue_is_full` before calling `start_fetch` or `start_fetch_deltapart`.

However, all the initial requests are started with `initiate_request`. In that function, if no delta is found then you go into `queue_scan_one_metadata_object`, which eventually checks `fetcher_queue_is_full`. But if a delta is found, then you go to `initiate_delta_request`, which calls `_ostree_fetcher_request_to_membuf` without checking that the fetcher queue is full.

So, I think you can hit this problem if you pull more than 24 deltas. The solution would appear to be that a queue is needed for superblocks.

cgwalters: Thanks, I updated it to drop the assertion from libcurl too.  You're right about the code chain involved here...I think it's fixable, we'd need to maintain integer offsets as iterators over the bits that call `initiate_request()`, but doing too many HTTP requests is better than crashing for now.

dbnicholson: Sure. I think you could refactor a bit with the existing scheme using a `pending_fetch_superblock` set and `start_superblock_fetch` utility function that does the actual `_ostree_fetcher_request_to_membuf` call. Then you'd just check `fetcher_queue_is_full` in `initiate_delta_request` and either add to the set or call `start_superblock_fetch`.

dbnicholson:
@rh-atomic-bot r+ 0953ce0

PR #1454, opened 2018-02-14
Fix static-compiler when CC includes args

akiernan:
Ensure arguments are quoted so that if you include args in CC that
they're handled as a whole.


cgwalters:
@rh-atomic-bot r+ cf3a870

PR #1457, opened 2018-02-19
Make tests compatible with Python 3

smcv:
Because Python 2 reaches EOL in 2020, which is within the projected security-support lifetime of Ubuntu 18.04 and Debian 10, there's an effort within Debian to replace Python 2 with Python 3 as much as possible.

ostree currently uses Python 2 for some tests, but this pull request makes it compatible with Python 3. I believe this would be necessary in Arch Linux, where `/usr/bin/python` is Python 3, if Arch Linux routinely ran the ostree test suite (which it seems they don't).

In Debian we're using an additional patch to switch the Python 2 tests to Python 3, which isn't included in this pull request, but I'm happy to open another one afterwards (or add it to this one) if you're interested: https://salsa.debian.org/debian/ostree/blob/debian/master/debian/patches/debian/Use-Python-3-for-tests.patch

### DIFF
--- a/tests/bootloader-entries-crosscheck.py
+++ b/tests/bootloader-entries-crosscheck.py
@@ -38,8 +38,8 @@ def fatal(msg):
     sys.stderr.write('\n')
     sys.exit(1)
 
-def compare_entries_descending(a, b):
-    return int(b['version']) - int(a['version'])
+def entry_get_version(entry):
+    return int(entry['version'])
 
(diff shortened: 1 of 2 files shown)


cgwalters:
Hmm, we might actually be able to hard switch over to 3.  I think the main thing would be to set up EPEL on our c7 builds.  So please do submit that next patch!

@rh-atomic-bot r+ 7c8f9b0

PR #1462, opened 2018-02-22
ci: Rework installed tests to use Fedora Standard Test interface

cgwalters:
Reusing the way `standard-test-roles` has support for booting
a qcow2 actually gets us to the "VM-in-container" flow.  Plus
Ansible over shell script is sometimes nicer.

https://fedoraproject.org/wiki/CI/Tests#Testing_an_Atomic_Host

This is just an experiment to see if this works; right now we aren't
even injecting the built ostree into the VM, or doing any interesting tests.

One thing that I'm sure we're going to hit is a need to cache the qcow2 in the
CI environment.


jlebon:
Cool! Though we don't have `oci-kvm-hook` right now installed on the CI nodes so no `/dev/kvm` in your container. I'll get it added.

jlebon:
bot, retest this please

cgwalters:
bot, retest this please

jlebon:
bot, retest this please

cgwalters:
>     "msg": "argument must be an int, or have a fileno() method.", 

Damn it Python...

cgwalters:
Using Professional Software Engineering skill `#23` (googling for the error message) I see https://code.vt.edu/nis-ansible-roles/upstream-ansible/commit/af4f0bd0087d5d4067ad11f45dc646230c3ced7f which seems somehow promising...yet I don't see that code in actual ansible git.  Hmm.  Let's try getting more debugging info out of ansible.

jlebon:
Hmm, right I can reproduce this locally:

```
$ sudo docker run --rm -v /var/cache/dnf:/var/cache/dnf:z registry.fedoraproject.org/fedora:27 sh -c 'dnf install -y ansible && ansible -vvv -i 10.8.246.101, all -m shell -a /bin/true'
...
META: ran handlers
Using module file /usr/lib/python2.7/site-packages/ansible/modules/commands/command.py
<10.8.246.101> ESTABLISH CONNECTION FOR USER: None on PORT 22 TO 10.8.246.101
10.8.246.101 | UNREACHABLE! => {
    "changed": false,
    "msg": "argument must be an int, or have a fileno() method.",
    "unreachable": true
}
```

If I pass `-t`, I get:

```
...
paramiko: The authenticity of host '10.8.246.101' can't be established.
The ssh-ed25519 key fingerprint is a1f9a4d65418abceec2988451f974c76.
Are you sure you want to continue connecting (yes/no)?
```

So it seems likely to be related to PAPR not allocating a tty.

jlebon:
Explicitly passing `ANSIBLE_HOST_KEY_CHECKING=False` "works" (as in I get the error I would expect):

```
<10.8.246.101> ESTABLISH CONNECTION FOR USER: None on PORT 22 TO 10.8.246.101
10.8.246.101 | UNREACHABLE! => {
    "changed": false,
    "msg": "No authentication methods available",
    "unreachable": true
}
```

Hmm, I wonder if `-o StrictHostKeyChecking=no` is not actually taking effect. Can you try using `ANSIBLE_HOST_KEY_CHECKING` as well?

cgwalters:
The `paramiko` error was enough of a hint: We were missing `openssh-clients` in the depchain.  When I was testing this interactively, that missing dep was masked since `git-core` pulls it in.

cgwalters:
OK so...let me briefly outline my plans here.

High level goals;

 1) Dedup CI between rpm-ostree and ostree (and potentially other projects)
 2) Increase alignment with https://fedoraproject.org/wiki/CI - specifically things like using Ansible, though in practice we may end up with *both* Ansible + shell scripts?
 3) Move strongly to the VM-in-container approach; the benefits are powerful

In general, let's try to follow the Cockpit path on this.  One of the
first decisions we need to take is how we inject binaries into the VM.
I'd like to go the "build an RPM" path instead of what we do now in rpm-ostree.

Taking this first step, along with the "dedup CI" path means I am inclined to
create a little bit more formal "libpaprci" that is a pure copy-paste lib for now.
Once we find high level patterns that work well, we can consider "baking" them into
PAPR yaml.  (But doing that has the cost that it's harder to run "directly" in
a dev container).

An interesting question here is to what degree it makes sense to have PAPR itself
switch to VM-in-container behind the scenes, particularly for the `cluster` case
with one container and one VM.

Anyways, short term goal is step `#1`: deduping CI between rpm-ostree and ostree and
hence gaining better VM-style testing for ostree.

cgwalters:
@jlebon ^

jlebon:
I'm mostly in agreement with the above, including rolling with a copy-paste lib for now. (Actually, we could have the lib stored in the PAPR container and just scp all environments with a copy of it).

> An interesting question here is to what degree it makes sense to have PAPR itself
switch to VM-in-container behind the scenes, particularly for the cluster case
with one container and one VM.

This is something that PAPR will *have* to learn if we want to migrate to OCP. It also takes us a step closer to integrating the Fedora STI directly in PAPR. (Basically, PAPR takes on the role of resource provisioner, or [testing system](https://fedoraproject.org/wiki/CI/Standard_Test_Interface#Terminology)).

> Move strongly to the VM-in-container approach; the benefits are powerful

One thing to note is that right now all container workloads are performed directly on the executor nodes. And we don't do any form of scheduling other than relying on Jenkins' built-in queue based on executor slots. Before we go all in on the VM-in-container approach, we need better scheduling, which I think would make more sense to tackle until after we move to OCP?

For the time being, I would suggest provisioning a `host` and running containerized tests there. Obviously backwards :upside_down_face:, though at least it unblocks investments into this today that will still be relevant tomorrow. And we still get all the other nice features of the approach (better sharing, reproducibility, portability, etc...).

In the meantime, I'd like to pursue the developments going on in [PACI](https://github.com/projectatomic/projectatomic-ci-infra) to set up the full infra in OCP. Once we're to a point where we can run container workloads as pods there, we can iterate over integrating support for containerized VMs.

cgwalters:
bot, retest this please

cgwalters:
Lifting WIP on this.

cgwalters:
@miabbott would you mind looking at this too?

ci/build-rpm.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/bash
+# Install build dependencies, run unit tests and installed tests.
```
jlebon: Minor: this looks like the wrong description, right?

ci/build-rpm.sh
```diff
@@ -0,0 +1,46 @@
...
+paprcidir=${dn}/libpaprci
+. ${paprcidir}/libbuild.sh
+
+# Auto-provision bootstrap resources if run as root (normally in CI)
+if test "$(id -u)" == 0; then
+    pkg_install_buildroot
+    pkg_install make /usr/bin/rpmbuild git
+fi
+
+# PAPR really should do this
```
jlebon: Oh, good catch! Filed: https://github.com/projectatomic/papr/issues/90.

ci/build-rpm.sh
```diff
@@ -0,0 +1,46 @@
...
+
+# Default libcurl on by default in fedora unless libsoup is enabled
+if test "${OS_ID}" = 'fedora'; then
+    case "${CONFIGOPTS:-}" in
+        *--with-soup*|*--without-curl*) ;;
+        *) CONFIGOPTS="${CONFIGOPTS:-} --with-curl"
+    esac
+fi
+case "${CONFIGOPTS:-}" in
+    *--with-curl*|--with-soup*)
```
jlebon: This looks like it's missing a `*` after the `|`.

ci/libpaprci/libbuild.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/bash
+
+dn=$(cd $(dirname $0) && pwd)
+
+OS_ID=$(. /etc/os-release; echo $ID)
+OS_VERSION_ID=$(. /etc/os-release; echo $VERSION_ID)
+
+pkg_upgrade() {
+    # https://bugzilla.redhat.com/show_bug.cgi?id=1483553
```
jlebon: Optional/not new: we should be able to drop this by now.

ci/libpaprci/libbuild.sh
```diff
@@ -0,0 +1,74 @@
...
+    fi
+}
+
+make() {
+    /usr/bin/make -j $(getconf _NPROCESSORS_ONLN) "$@"
+}
+
+build() {
+    env NOCONFIGURE=1 ./autogen.sh
+    ./configure --prefix=/usr --libdir=/usr/lib64 "$@"
```
jlebon: Optional/not new: ahh, I forgot to upstream the `--sysconfdir=/etc` fix here!

ci/provision-prep.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/bash
+# Prepare the current environment
```
jlebon: It looks like this file isn't used anywhere.

tests/fedora-str/overlay-git.yml
```diff
@@ -0,0 +1,40 @@
...
+    synchronize: src=x86_64/ dest=/root/x86_64/ archive=yes
+  - shell: ostree admin unlock || true
+    changed_when: False
+  # Install the RPMs we already have.  For the test suite we use rpm2cpio
+  # since it depends on libsoup, but we're not using that yet for the sysinstalled tests
+  - shell: >
+      /usr/bin/rpm -Fvh /root/x86_64/*.rpm && \
+      cd / && rpm2cpio /root/x86_64/ostree-tests-2*.rpm | cpio -div
+    changed_when: False
+  - shell: /usr/bin/rpm -Fvh /root/x86_64/*.rpm
```
jlebon: We just did this in the previous task, right?

tests/fedora-str/overlay-git.yml
```diff
@@ -0,0 +1,40 @@
...
+    changed_when: False
+  - command: ostree --version
+    changed_when: False
+    register: ostree_version
+  - set_fact:
+      ostree_version_yaml: "{{ ostree_version.stdout | from_yaml }}"
+  # FIXME: make this work when building RPMs too
```
jlebon: Hmm, how about just checking that the version before and after the RPM install are not the same?

tests/fedora-str/sysinstall-tests.yml
```diff
@@ -0,0 +1,23 @@
...
+  tags:
+  - atomic
+  remote_user: root
+  tasks:
+  - find:
+      paths: /root/tests/installed/
+      patterns: "itest-*.sh"
+    register: installed_tests
+  - name: Run sysinstalled tests
+    shell: /root/tests/installed/run.sh &> /root/installed-tests.log
```
jlebon: This is OK, though things will get tricky when we want to e.g. have tests that reboot/modify the VM. It could be interesting to have a hybrid approach of this and rpm-ostree's `libvm` where tests can seamlessly use the ansible infra for running code remotely, rebooting, pushing/fetching files, etc... Or do you already have an idea in mind on how to approach this?

ci/libpaprci/Makefile.dist-packaging
```diff
@@ -0,0 +1,39 @@
...
+dist-snapshot:
+	if ! test -f $(PKG_VER).tar.xz; then \
+	  $(mypath)/make-git-snapshot.sh "$(topsrcdir)" "$(PKG_VER)" "$(GITREV)" && \
+	  rm -f $(PKG_VER).tar.xz && \
+	  xz $(PKG_VER).tar; \
+  fi
+
+srpm: dist-snapshot
+	if test -f "$(SPEC)"; then \
+	   sed -e "s,^Version:.*,Version: $(GITREV_FOR_PKG)," $(SPEC) > $(DISTGIT_NAME).spec && \
```
jlebon: So, does this not work for some reason? (Judging by the comment in `overlay-git.yaml`).

tests/fedora-str/overlay-git.yml
```diff
@@ -0,0 +1,40 @@
...
+  - atomic
+  remote_user: root
+  tasks:
+  # Install the built RPM to the VM
+  - command: git rev-parse HEAD
+    register: git_rev_head
+    connection: local
+  - name: Copy locally built RPMs
+    synchronize: src=x86_64/ dest=/root/x86_64/ archive=yes
+  - shell: ostree admin unlock || true
```
jlebon: I'm not a big fan of using `admin unlock` for this, given that it's not how it's actually run on client systems. It also limits what tests can do (e.g. testing `admin unlock` itself, or rebooting). Have you considered building our own tree like in rpm-ostree?

Actually, it should work to just `rpm-ostree override replace` the RPMs, reset the ref and change the origin back to use `refspec`, right?

tests/fedora-str/sysinstall-tests.yml
cgwalters: Right, I think the next step is to change some of the installed tests to be playbooks.

ci/libpaprci/Makefile.dist-packaging
cgwalters: It works, I was just trying to proactively prevent any "not testing what was built" type issues like we've had once or twice.  Basically I was thinking the very first test should be checking the binary's built version in some way.  Which we now have.

tests/fedora-str/overlay-git.yml
cgwalters: Mmm.  I use `admin unlock` a lot for testing personally where I really explicitly don't want things to persist.  We'll clearly need `vmcheck` and `override replace` to be used as well though.

(I'd be cool down the line to have `override override replace --transient` or something which would be like more like `replace` + "livefs-on-overlay" but still underneath the ro mount).

cgwalters: Since the previous installed tests code in `fah-prep.sh` (now deleted) also used `admin unlock` i'd vote we keep it for now - and build upon this first step by e.g. copying the a-h-t reboot playbook and using it, etc.

cgwalters:
Fixup for all comments :arrow_up:

jlebon:
@rh-atomic-bot r+ 7407a66

jlebon:
The PR tester failed with:

```
+ ostree --repo=repo remote add origin --set=gpg-verify=false http://127.0.0.1:44687
+ ostree --repo=repo pull --disable-static-deltas origin fedora/27/x86_64/atomic-host
error: [28] Timeout was reached
```

But it worked earlier, so probably a flake?

Anyway, let's try this out! 🎉

PR #1463, opened 2018-02-22
Use Python 3 for tests

smcv:
As requested on #1457, here's the patch I used in Debian to run all tests in a version of Python that is not approaching EOL.

An alternative to this would be to use `$(PYTHON)` as populated by `configure.ac`, but to work for the installed-tests that would require messing around with `.in` files and I'm not sure whether that's really worth the effort.

### DIFF
--- a/ci/travis-install.sh
+++ b/ci/travis-install.sh
@@ -110,7 +110,7 @@ case "$ci_distro" in
             libcurl4-openssl-dev \
             procps \
             zlib1g-dev \
-            python-yaml \
+            python{,3}-yaml \
             ${NULL}
 
         if [ "$ci_in_docker" = yes ]; then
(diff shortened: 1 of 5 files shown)


ci/travis-install.sh
```diff
@@ -110,7 +110,7 @@ case "$ci_distro" in
             libcurl4-openssl-dev \
             procps \
             zlib1g-dev \
-            python-yaml \
+            python{,3}-yaml \
```
smcv: python3-yaml only should be enough on Debian/Ubuntu now that we're using python3 everywhere. python-yaml (which means Python 2) isn't in the Build-Depends any more for the production builds in Debian, and those work fine.

smcv:
I'm no expert on Red Hat derivatives, but your CI changes look entirely reasonable.

cgwalters:
@rh-atomic-bot r+ 88892d9

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

Issue #545, opened 2016-10-24
Handle files in /etc  being deleted/changed during upgrades


cgwalters:
See https://bugzilla.redhat.com/show_bug.cgi?id=1366584 for one example, although I suspect there are others.

The first step in this is going to be handling `ENOENT` when processing modified config files.

What we really should investigate though is doing `/etc` merges only on shutdown or so.

dustymabe:
> We don't change the bootloader or do the /etc merge. Both of those would happen on shutdown, just before reboot.

does that mean we would properly handle the "power loss" case where I've "upgraded" but not yet reboot yet and i suffer a power loss?

cgwalters:
The short answer is "yes".  I'm not sure exactly what your concern is, but the only power loss thing to be concerned about here is that if you get a system crash during the "final processing" phase where we do the /etc merge and bootloader swap, you'd be left in the current system, not the new one.

dustymabe:
ok cool. so if we lost power before reboot we'd simply stay in the state where we are booting into the old deployment on every boot. Would the user just `rpm-ostree upgrade` again to get to the newer deployment (again).

cgwalters:
The way I'd been thinking of implementing things, the update would still be "pending" when the system came back up, so doing a reboot would retry the `/etc` merge and bootloader swap.

cgwalters:
I think we should also try to only do `syncfs()` at this same time too.  This would be even nicer for rpm-ostree livefs style workflows.

The user-visible effect here though would be that it'd be possible for an operation that was applied live could be "undone" after a system crash.

Things get tricky though as we need to think about the flatpak (and other container) use cases.  What degree of expected durability is there?

Perhaps scoping in the change to `syncfs()` would be too hard here and we should do the `/etc` merge first - particularly given that I think in the modern day we basically expect people to be using SSD for the base OS data, so write times and latency imposed by syncing are a lot less of a concern than the days of rotational storage.

cgwalters:
PR: https://github.com/ostreedev/ostree/pull/1503

Issue #1460, opened 2018-02-21
Add support for "pinning" 📌 deployments


cgwalters:
Update 2018-02-23 PR: https://github.com/ostreedev/ostree/pull/1464

---

We should add `ostree admin pin <deploymentid>` which makes the deployment persist until explicitly unpinned.  Example user story: Jane rebases her OS to a new major version `N`, and wants to keep around `N-1` even after a few upgrades for a while so she can easily roll back.

dustymabe:
very similar to https://github.com/projectatomic/rpm-ostree/issues/577. basically the difference is in the heuristic. Both want to keep around old deployments. One asks for them to be garbage collected based on a number, the other asks for them to be garbage collected based on a configurable flag.

PR #1464, opened 2018-02-23
sysroot: Add concept of deployment "pinning" 📌

cgwalters:
See commits.

Also has a prep commit for transient state which I plan to use for "pending" state AKA https://github.com/ostreedev/ostree/issues/545


dustymabe:
any chance this could be extended to support the use case described in https://github.com/ostreedev/ostree/issues/1460#issuecomment-367447536 and https://github.com/projectatomic/rpm-ostree/issues/577 ?

cgwalters:
> any chance this could be extended to support the use case described in #1460 (comment) and projectatomic/rpm-ostree#577 ?

It's not clear what you want there - something like a config file where you can change `2` to `3` or `5` etc?  The semantics of that would seem...strange to me.  It could be implemented on top of this by pinning a number anyways.

dustymabe:
> It's not clear what you want there - something like a config file where you can change 2 to 3 or 5 etc?

yes.

> The semantics of that would seem...strange to me.

To me this is basically the same thing as saying *"keep around X number of snapshots"*. For example on my desktop system I have btrfs snapshots and I essentially keep them all; I have lots of disk so why not? It would be great if I could configure ostree/rpm-ostree to do something similar so I can start moving my systems over to FAW. 

> It could be implemented on top of this by pinning a number anyways.

Not sure what that means exactly. So instead of a deployment the user could pin X where X represents the number of deployments to keep?

cgwalters:
>  For example on my desktop system I have btrfs snapshots and I essentially keep them all; I have lots of disk so why not? It would be great if I could configure ostree/rpm-ostree to do something similar so I can start moving my systems over to FAW.

But do you e.g. snapshot `/home` too?  It's not really the same thing; ostree isn't a backup system.

The way I think of it is; keeping lots of snapshots locally doesn't seem to make sense because you can re-download the from the Internet on-demand.  (And yes, that's an experience we need to consistently enable with rpm-ostree and packages).

But if you still want it, I'm not opposed to adding more policies, but it seems orthogonal to this PR.

dustymabe:
> But do you e.g. snapshot /home too? It's not really the same thing; ostree isn't a backup system.
> 

In my case I do (i actually snapshot the entire filesystem), but I get that ostree isn't going to be exactly what I currently have. I've actually thought about writing a snapper plugin for ostree, like the dnf one, that will call snapper and ask it to make snapshots (btrfs, or thin LV snapshots) automatically on configured filesystems whenever an rpm-ostree operation happens. Haven't really got far on that idea, just a thought so far.

> The way I think of it is; keeping lots of snapshots locally doesn't seem to make sense because you can re-download the from the Internet on-demand. (And yes, that's an experience we need to consistently enable with rpm-ostree and packages).

That is true, but for some reason I lean towards wanting that functionality. if I rebase from f27 to rawhide AND then layer one package my f27 deployment is gone. I'd like it to be a little easier to keep around old deployments than that. The pinning helps, but only works if you proactively think about it. 

> 
> But if you still want it, I'm not opposed to adding more policies, but it seems orthogonal to this PR.

Thank you! I guess it's orthogonal, seemed like a slight variation of the same thing to me.

cgwalters:
> if I rebase from f27 to rawhide AND then layer one package my f27 deployment is gone. I'd like it to be a little easier to keep around old deployments than that. The pinning helps, but only works if you proactively think about it.

I think we'll edit the "rebase instructions" page to include `--pin`.

src/libostree/ostree-sysroot.c
```diff
@@ -1832,3 +1834,43 @@ ostree_sysroot_deployment_unlock (OstreeSysroot     *self,
...
+ * @error: Error
+ *
+ * By default, deployments may be subject to garbage collection. Typical uses of
+ * libostree only retain at most 2 deployments. If @is_pinned is `TRUE`, a
+ * metadata bit will be set causing libostree to avoid automatic GC of the
+ * deployment. However, this is really an "advisory" note; it's still possible
+ * for e.g. older versions of libostree unaware of pinning to GC the deployment.
+ *
+ * This function does nothing and returns successfully if the deployment
+ * is already in the desired pinning state.
```
jlebon: `Since:` ?

tests/test-admin-deploy-2.sh
```diff
@@ -63,3 +63,34 @@ assert_has_dir sysroot/boot/ostree/testos-${bootcsum}
...
+${CMD_PREFIX} ostree admin pin -u 0
+assert_n_pinned 0
+echo "ok pin unpin"
+
+${CMD_PREFIX} ostree admin pin 0
+${CMD_PREFIX} ostree admin pin 1
+assert_n_pinned 2
+os_repository_new_commit
+${CMD_PREFIX} ostree admin upgrade --os=testos
+assert_n_pinned 2
```
jlebon: Maybe let's explicitly also check that the pinned deployments are still there? (I.e. just grep for checksums? Or maybe just checking that the deployment count went up to three.)

src/ostree/ot-admin-builtin-pin.c
```diff
@@ -0,0 +1,82 @@
...
+
+#include "ot-main.h"
+#include "ot-admin-builtins.h"
+#include "ot-admin-functions.h"
+#include "ostree.h"
+#include "otutil.h"
+
+/* ATTENTION:
+ * Please remember to update the bash-completion script (bash/ostree) and
+ * man page (man/ostree-admin-undeploy.xml) when changing the option list.
```
jlebon: Stale comment. Also doesn't look like we added bash completions for this, right?
Let's at least change it to a `TODO`?

src/libostree/ostree-sysroot.c
```diff
@@ -1832,3 +1834,43 @@ ostree_sysroot_deployment_unlock (OstreeSysroot     *self,
...
+ostree_sysroot_deployment_set_pinned (OstreeSysroot     *self,
+                                      OstreeDeployment  *deployment,
+                                      gboolean           is_pinned,
+                                      GError           **error)
+{
+  const gboolean current_pin = ostree_deployment_is_pinned (deployment);
+  if (is_pinned == current_pin)
+    return TRUE;
+
+  g_autoptr(OstreeDeployment) deployment_clone = ostree_deployment_clone (deployment);
```
jlebon: I'm confused here. Why do we need `ostree_deployment_clone`? Can't we just `ostree_deployment_get_origin` and modify that? (Actually, *shouldn't* we do that so that the sysroot deployment array is up to date?)

cgwalters:
Now with another prep commit, and squashed a few fixups :arrow_up:

bash/ostree
```diff
@@ -399,6 +399,37 @@ _ostree_admin_os_init() {
     return 0
 }
 
+_ostree_admin_pin() {
```
jlebon: IIUC, we also need a `pin` entry in `_ostree_admin()`.

bash/ostree
```diff
@@ -399,6 +399,37 @@ _ostree_admin_os_init() {
...
+    local boolean_options="
+        $main_boolean_options
```
jlebon: `--unpin` ?

src/libostree/ostree-sysroot.c
jlebon: Ahh got it, that makes sense.

bash/ostree
cgwalters: I honestly don't grasp bash completion, I don't care too much about it myself at least for `ostree admin`.  Let's not try to polish it for now?

cgwalters: There has to be someone out there who has implemented actually generating this stuff from the cli `--help`, as is it's an enormous maintenance pain doubled by the fact that we need to wrap it in `rpm-ostree` at least too.

jlebon:
Sure, that's fine with me.
@rh-atomic-bot r+ 4374dcf

PR #1468, opened 2018-02-26
bash-completion: Remove `admin` completions

cgwalters:
The `admin` commandline should be considered a demo; I just added
the `pin` command *mostly* so we could use it for unit tests, although
I can imagine other people using it.

But maintaining completions is a lot of overhead right now, let's not
do it for `admin`.

The other command line options that operate on repos we will definitely maintain
since they're used in releng contexts.


bash/ostree
```diff
@@ -180,542 +180,6 @@ _ostree_ostree() {
...
-            ;;
-        *)
-            COMPREPLY=( $( compgen -W "$subcommands" -- "$cur" ) )
-            ;;
-    esac
-
-    return 0
-}
-
-_ostree_cat() {
```
jlebon: Looks like we're nixing `ostree cat` too here. :crying_cat_face:

cgwalters: Hah, excellent use of the sad cat emoji! :smile:   Fixup :arrow_down:

jlebon:
@rh-atomic-bot r+ 8ab48e4

PR #1472, opened 2018-02-28
lib/sysroot: Fix retrieving non-booted pending deployment

jlebon:
If we're booted into a deployment, then any queries for the pending
merge deployment of a non-booted OS will fail due all of them being
considered rollback.

Fix this by filtering by `osname` *before* determining if we've crossed
the booted deployment yet.


jlebon:
I guess this will require an itest since it needs a real system.

jlebon:
(Untested so far).

cgwalters:
This looks good to me, I'd be fine merging as-is without a new test.

jlebon:
SGTM! Dropped WIP label.

cgwalters:
@rh-atomic-bot r+ cf4a8d3

PR #1473, opened 2018-03-01
lib/repo-finder-mount: Update comment about paths

mwleeds:
This updates the gtk-doc comment for OstreeRepoFinderMount to match the
correct flatpak repo path, which was fixed in commit 6db6268df.


cgwalters:
@rh-atomic-bot r+ 4fcc0e3

Issue #1433, opened 2018-01-25
Assertion failure when running `ostree admin upgrade`


mwleeds:
I'm not sure yet if this is reproducible using upstream ostree rather than the one in Endless but I'm making an issue so I don't forget.

Steps to reproduce:
1. Set a collection ID on the `eos` remote in /ostree/repo/config on an Endless 3.3.9 computer.
2. Run `sudo ostree admin upgrade`

Results:
```
endless@endless:/ostree/repo$ sudo ostree admin upgrade

Receiving objects: 99% (7388/7389) 277.3 kB/s 124.8 MB                                                   
(ostree admin upgrade:3092): GLib-CRITICAL **: g_variant_unref: assertion 'value != NULL' failed
1502 metadata, 5887 content objects fetched; 121951 KiB transferred in 456 seconds                       
Copying /etc changes: 15 modified, 0 removed, 20 added
Bootloader updated; bootconfig swap: yes deployment count change: 1
Freed objects: 2.3 MB
```
Oddly ostree seems to continue with the operation after the assertion failure.

cgwalters:
Hmm.  There are a lot of places in the code this could come from unfortunately.  I haven't seen this personally, but I could completely believe there's a bug somewhere.  Offhand though I don't have a good guess.

cgwalters:
Now personally...I tend to run with `G_DEBUG=fatal-warnings` (hm, I need to fix our CI to do this).  Then you'll get a core with a stack trace which is way more useful for debugging.

cgwalters:
Nevermind, the CI already does:

```
$ git grep fatal-warning
Makefile-tests.am:      G_DEBUG=fatal-warnings \
tests/libtest-core.sh:export G_DEBUG=fatal-warnings
```

mwleeds:
@dbnicholson has a patch for this. Turned out the wrong `free_func` was used

PR #1474, opened 2018-03-01
lib/repo-pull: Fix free function for hash table

mwleeds:
The "ref_original_commits" hash table uses string values, not variants,
so fix the free function passed to g_hash_table_new_full (). Since
g_variant_unref isn't NULL safe, this prevents an assertion failure when
a NULL value is inserted.

Dan Nicholson suggested this patch; I'm just submitting it because he's
busy.

Fixes https://github.com/ostreedev/ostree/issues/1433


cgwalters:
Thanks!

@rh-atomic-bot r+ 62ea9ad

PR #1476, opened 2018-03-02
lib/remote: Fix memory leak

mwleeds:


cgwalters:
@rh-atomic-bot r+ 184bb88

Thanks!

PR #1477, opened 2018-03-02
lib/repo-finder-mount: Improve debug message

mwleeds:
This makes it easier to tell which mount is being checked when repos are
found.


cgwalters:
Minor procedural note; our merge bot makes landing small commits in separate PRs a bit more expensive, so it can make sense to do a "small fixups" PR if you plan to do multiple at once.  But not required at all, and at some point we'll optimize the merge bot logic a bit more for this.

@rh-atomic-bot r+ 1f2fe00

mwleeds:
Ah, good point. I'll do that next time

PR #1478, opened 2018-03-03
lib: Fix memory leaks of OstreeRemote

mwleeds:
The _ostree_repo_get_remote() and _ostree_repo_get_remote_inherited()
methods transfer ownership of the returned OstreeRemote to the caller,
so this commit fixes a few call sites that weren't properly freeing it.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -5695,7 +5695,7 @@ ostree_repo_resolve_keyring_for_collection (OstreeRepo    *self,
     }
 
   if (keyring_remote != NULL)
-    return ostree_remote_ref (keyring_remote);
+    return keyring_remote;
   else
     {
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -5695,7 +5695,7 @@ ostree_repo_resolve_keyring_for_collection (OstreeRepo    *self,
     }
 
   if (keyring_remote != NULL)
-    return ostree_remote_ref (keyring_remote);
+    return keyring_remote;
```
cgwalters: Let's use autoptr on the variable still, and then `g_steal_pointer()` here?  We gain consistency and keep the movement towards Rust-like "move semantics".

mwleeds: good point. I pushed another change in the fixup too, so see if that looks good. It's a behavior change but I think in the right direction

cgwalters:
@rh-atomic-bot r+ e42dba3

Thanks!

Issue #1479, opened 2018-03-05
prune: don't ignore `--static-deltas-only` if provided


dustymabe:
If a user provides `--static-deltas-only` then the prune operation should never delete actual commits in the repo. Currently if a user provides `--static-deltas-only` without another argument then it will ignore `--static-deltas-only` and operate on normal commits.

```
[dustymabe@composer ~][STG]$ ./prune-atomic-ostree-repos  --test   
INFO:prune-atomic-ostree-repos:Running ['ostree', 'prune', '--repo', '/mnt/fedora_koji_prod/koji/atomic/27', '--refs-only', '--static-deltas-only', '--no-prune']
INFO:prune-atomic-ostree-repos:####################################                                                                    
INFO:prune-atomic-ostree-repos:                                    
Total objects: 412683                                              
Would delete: 15939 objects, freeing 0 bytes                       

INFO:prune-atomic-ostree-repos:####################################                                                                    
[dustymabe@composer ~][STG]$                                       
[dustymabe@composer ~][STG]$ 
[dustymabe@composer ~][STG]$ ./prune-atomic-ostree-repos  --test
INFO:prune-atomic-ostree-repos:Running ['ostree', 'prune', '--repo', '/mnt/fedora_koji_prod/koji/atomic/27', '--refs-only', '--no-prune']
INFO:prune-atomic-ostree-repos:####################################
INFO:prune-atomic-ostree-repos:
Total objects: 412683
Would delete: 15939 objects, freeing 0 bytes

INFO:prune-atomic-ostree-repos:####################################
```

cgwalters:
Short term PR in https://github.com/ostreedev/ostree/pull/1482

dustymabe:
should this issue be closed now?

cgwalters:
Yep.

PR #1482, opened 2018-03-05
prune: Error if --static-deltas-only without --delete-commit

cgwalters:
The original changes here apparently had the *idea* that `--static-deltas-only`
would be useful in general, but we never implemented that.  The current
situation where it's ignored unless `--delete-commit` is specified is
very misleading and I can easily see it leading to data loss for people.

Let's error out until we have a chance to make it actually useful.

Related: https://github.com/ostreedev/ostree/issues/1479


cgwalters:
@giuseppe You wrote the code here originally; this OK with you for now?

jlebon:
That looks sane to me. Will let @giuseppe have a look as well.

giuseppe:
@cgwalters yes, this looks good to me:

@rh-atomic-bot r+ c38b488

Issue #1480, opened 2018-03-05
prune: incorrect output for `--no-prune`


dustymabe:
`--no-prune` shows us `freeing 0 bytes`. I'd expect this to actually show us a value for how much space we would free.

```
[dustymabe@host repo]$ sudo ostree prune --only-branch f27:fedora/27/ppc64le/atomic-host --depth=8 --no-prune
Total objects: 73077
Would delete: 3415 objects, freeing 0 bytes
```

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1483

PR #1483, opened 2018-03-05
lib/repo: Do account for size with prune --no-prune

cgwalters:
I think this got changed in a refactor.  We definitely want
to total up the amount of space that *would* be freed even
with `--no-prune` AKA `OSTREE_REPO_PRUNE_FLAGS_NO_PRUNE`.

It's actually a bit terrifying this is apparently the first test case for
the `--no-prune` option...

Closes: https://github.com/ostreedev/ostree/issues/1480


jlebon:
@rh-atomic-bot r+ 5b451fd

Issue #1481, opened 2018-03-05
RFE: prune: static delta prune user stories


dustymabe:
This may be already covered in one way or another with existing functionality, if so maybe this is a doc/example RFE more than anything else.

- user story: as an admin of an ostree repo, I only consider the latest HEAD commit in any ref to be relevant with regards to static deltas. I'd like a single command I can run to prune all static deltas but the ones pointing the HEAD of a particular ref (or all refs)

- user story: as an admin of an ostree repo, I want to keep static deltas pointing to commits of `depth=10` (HEAD..HEAD-10). All other static deltas I would like to delete. I'd like a single command I can run to prune all other deltas.

cgwalters:
It seems to me that https://github.com/ostreedev/ostree/issues/1479 is the same as this RFE right?  In other words, if `--static-deltas-only` was honored you could prune them with the same policy as commits right?

dustymabe:
> It seems to me that #1479 is the same as this RFE right? In other words, if --static-deltas-only was honored you could prune them with the same policy as commits right?

Maybe? For some reason for me it is confusing to think about exactly how to combine the options in order to achieve the above mentioned user stories. That's why I said this may be a doc/example RFE more than anything else. We could definitely use some examples in the man pages.

cgwalters:
It'd be `ostree prune --refs-only --static-deltas-only` and `ostree prune --depth=10 --static-deltas-only` right?

dustymabe:
> It'd be `ostree prune --refs-only --static-deltas-only`

So this one says prune all deltas that don't point to HEAD commits?

>  and `ostree prune --refs-only --depth=10 --static-deltas-only` right?

and this one says prune all deltas that point to depth > 10 commits?

That makes sense, but I don't think it's completely obvious with the given docs. For example the `--depth=` argument could be interpreted as commit depth or static delta depth (if that was even a thing). An example would go a long way here, I think.

cgwalters:
> with the given docs.

There *are* no docs for this, `--static-deltas-only` never had any useful semantics.

In general, we treat deltas as a derived artifact whose "lifecycle" is tied to the commit.  Aside from the discussions of "delta only repositories", it doesn't make sense to keep around a delta but not the commit, right?

So all of the prune logic is around the commits.

dustymabe:
> There are no docs for this, --static-deltas-only never had any useful semantics.

[...]

PR #1484, opened 2018-03-05
docs/prune: Document that --static-deltas-only isn't that useful

cgwalters:
This is the documentation followup to: https://github.com/ostreedev/ostree/pull/1482

See also https://github.com/ostreedev/ostree/issues/1481


jlebon:
@rh-atomic-bot r+ 6a46269

Issue #1467, opened 2018-02-26
Unhelpful error messages (specifically when required empty directory is missing)


abferm:
I recently had to restore an ostree repository from a backup. I was greeted with the following error.
```
ostree refs --repo /tmp/repo --list
error: openat: No such file or directory
```
I had to use strace to figure out what was missing. It turns out that it was refs/remotes. Since this is a "server repository" the directory was empty. My backup method did not preserve empty directories, leading to this apparently required directory not existing.

I understand that it may be desired to not automatically create or ignore missing directories like this, however errors should at least tell the user enough information to correct the problem.

cgwalters:
What version of ostree are you using?  I suspect it's old; at one point I spent some time trying to improve our error-prefixing story and addressed a lot of these.  With current git:

```
$ mv repo/refs/remotes{,.orig}
$ ostree --repo=repo2 refs --list
error: opendir(refs/remotes): No such file or directory
$
```

cgwalters:
PR related to this in https://github.com/ostreedev/ostree/pull/1485

PR #1485, opened 2018-03-05
repo/refs: Clean up error prefixing

cgwalters:
Add some "function global" prefixing in line with what we do in
other places now, and drop the "manual filename" prefixing that
is no longer necessary since
https://github.com/GNOME/libglnx/commit/23f7df15006f14ddc3bc2ddee690f7f8604c3ebe

Closes: https://github.com/ostreedev/ostree/issues/1467


jlebon:
@rh-atomic-bot r+ 69d6562

PR #1486, opened 2018-03-05
[WIP] checkout: honor opaque checkouts

giuseppe:


giuseppe:
@nalind does this implementation look good to you?

giuseppe:
Dropping the WIP label

giuseppe:
tests are passing

cgwalters:
A little bit more information - a link to some other code that handles this format e.g. - would be useful in the commit message or as a code comment.

This is a serialization of the
```
A directory is made opaque by setting the xattr "trusted.overlay.opaque"
```

from the kernel `Documentation/filesystems/overlayfs.txt`?

src/libostree/ostree-repo-checkout.c
```diff
@@ -912,6 +914,22 @@ checkout_tree_at_recurse (OstreeRepo                        *self,
...
+      g_autoptr(GVariant) dir_file_contents = g_variant_get_child_value (dirtree, 0);
+      GVariantIter viter;
+      const char *fname;
+      g_autoptr(GVariant) contents_csum_v = NULL;
+      g_variant_iter_init (&viter, dir_file_contents);
+      while (g_variant_iter_loop (&viter, "(&s@ay)", &fname, &contents_csum_v))
+        {
+          is_opaque_whiteout = (g_str_equal (fname, OPAQUE_WHITEOUT_NAME));
+          if (is_opaque_whiteout)
+            break;
```
cgwalters: One thing I don't quite understand is why we're looping over all of the children here.  Doesn't there need to be name matching against the target somehow?   Or hmm...is the serialization that there's *exactly one* child with that name?

giuseppe: I am looking for a child ".wh..wh..opq" and it needs to have exactly that name.  I didn't find a better way for having this search in O(1).  Is there a better way to look for it?

giuseppe:
> A little bit more information - a link to some other code that handles this format e.g. - would be useful in the commit message or as a code comment.

This PR came after the comment here: https://github.com/containers/storage/pull/137#issuecomment-374387421

What I've done is:
```
skopeo copy docker://docker.io/redis@sha256:26c93c5b06eaa323bb1089500f42b0dd158138772348b865e364127f1d554982  ostree:foo@/ostree/repo
ostree ls ociimage/67b22db27e5144c14efdf55d9cef9c3dadffb43593790ec99663fb1b783dbb6b /var/lib/apt/lists/partial
d00755 0 0      0 /var/lib/apt/lists/partial
-00755 0 0      0 /var/lib/apt/lists/partial/.wh..wh..opq
```

what this file does is to rm -rf every entry in the directory where it is present so that `/var/lib/apt/lists/partial` exists but it is empty

I'll amend more information in the commit message

src/libostree/ostree-repo-checkout.c
cgwalters: > I am looking for a child ".wh..wh..opq" and it needs to have exactly that name. 

My question is - should that be the *only* child?

giuseppe: no, there can be other children and they will appear normally in the directory.  The ".wh..wh..opq" file is only to hide anything that is in the lower layers, but the same layer can add more content

cgwalters:
OK, thanks!  I think I understand now.  Just in general though I really prefer if even nearly trivial changes have some sort of rationale.  For example in https://github.com/ostreedev/ostree/commit/c9a9e6c3811002f18de9baa72d32aff5463f19d4

I could have just said "Add bindings section" since it's really obvious *what* I'm doing but even there I added a *why* I'm doing it.  Another example is:
https://github.com/ostreedev/ostree/commit/5183c8f35e5b730dd6d6e9673b05eba355c737b6
Where the *why* is "Making this change since I saw an unprefixed error in an issue.".

This is a far from trivial change and I think deserved a commit message; thanks for fleshing it out a bit!  We should probably still add comments to the code but not blocking on that since the previous code didn't have a lot either (and in some places in libostree we're relatively decent with comments, others not so much).

cgwalters:
And sorry again about the delay here...I think I must have missed notifications, and I sort of use the default github PR viewer as a "review stack" but this one got too buried.  Wish there was a way to change that view to default to "recently changed" and not "initial creation time".

giuseppe:
> And sorry again about the delay here...I think I must have missed notifications, and I sort of use the default github PR viewer as a "review stack" but this one got too buried. Wish there was a way to change that view to default to "recently changed" and not "initial creation time".

no problem :-) I've left it marked as WIP for a long time.

Also, for reference, I've used this Dockerfile to get an opaque checkout:
```
FROM alpine
RUN rm -rf /srv && mkdir -p /srv && touch /srv/a
```

Once it is imported into ostree I see:

```
# ostree ls -R ociimage/aa1c5b4533c0d2c7d18d06c7618a7e4318a71764b9a394202840cc85e39d7630 /
d00755 0 0      0 /
d00755 0 0      0 /etc
-00755 0 0      0 /etc/resolv.conf
d00755 0 0      0 /run
d00755 0 0      0 /run/secrets
d00755 0 0      0 /srv
-00755 0 0      0 /srv/.wh..wh..opq
-00644 0 0      0 /srv/a
```

Issue #1488, opened 2018-03-08
Last signature wins in `ostree summary -u`


jeremy-hiatt:
When the `--gpg-sign=<key>` flag is specified multiple times, only the last signature ends up in the `summary.sig` file. I'm working on a pull request that will fix the problem.

PR #1489, opened 2018-03-08
lib/repo: Fix multi-signature support when generating summary files

jeremy-hiatt:
Ensure that the metadata object is built up with the signatures from all keys
passed to ostree_repo_add_gpg_signature_summary(). Previously only the signature
from the last key would end up in the metadata.

Closes: #1488

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -4811,17 +4811,16 @@ ostree_repo_add_gpg_signature_summary (OstreeRepo     *self,
   /* Note that fd is reused below */
   glnx_close_fd (&fd);
 
-  g_autoptr(GVariant) existing_signatures = NULL;
+  g_autoptr(GVariant) metadata = NULL;
   if (!ot_openat_ignore_enoent (self->repo_dir_fd, "summary.sig", &fd, error))
     return FALSE;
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo.c
```diff
@@ -4830,10 +4829,10 @@ ostree_repo_add_gpg_signature_summary (OstreeRepo     *self,
                       cancellable, error))
         return FALSE;
 
-      new_metadata = _ostree_detached_metadata_append_gpg_sig (existing_signatures, signature_data);
+      metadata = _ostree_detached_metadata_append_gpg_sig (metadata, signature_data);
```
jlebon: One subtle issue here is that we're leaking the previous value of `metadata`. Let's tweak it to something like this instead:

```c
g_autoptr(GVariant) old_metadata = g_steal_pointer (&metadata);
metadata = _ostree_detached_metadata_append_gpg_sig (old_metadata, signature_data);
```

?

jeremy-hiatt:
Hi @jlebon thanks for your feedback. I made the change as you described. Can you please take another look?

jlebon:
Thanks for the patch!
@rh-atomic-bot r+ 5d49262

PR #1490, opened 2018-03-08
configure: add option for libsystemd

Cogitri:
Right now ostree checks for libsystemd and enables support for it if it finds it. This PR changes that behavior by adding an option to enable/disable libsystemd. This is especially useful if one uses a source based distro ( like Gentoo/Exherbo ), where one wants to avoid such automagic detection of dependencies and prefers switches for that instead.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -469,10 +469,26 @@ AC_ARG_WITH(mkinitcpio,
 AM_CONDITIONAL(BUILDOPT_MKINITCPIO, test x$with_mkinitcpio = xyes)
 
 dnl We have separate checks for libsystemd and the unit dir for historical reasons
-PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd], [have_libsystemd=yes], [have_libsystemd=no])
-AM_CONDITIONAL(BUILDOPT_LIBSYSTEMD, test x$have_libsystemd = xyes)
-AM_COND_IF(BUILDOPT_LIBSYSTEMD,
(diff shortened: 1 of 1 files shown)


cgwalters:
This looks sane to me, but would you mind adding the content from your PR submission into the commit text itself?

Cogitri:
Sure, done

cgwalters:
Thanks!

@rh-atomic-bot r+ 7a6a17f

cgwalters:
@rh-atomic-bot r+ 304790c

cgwalters:
BTW one thing you can do here is push `fixup!` commits; would have been easier to see what the bug was.

cgwalters:
3rd time's a charm?  Reviewing automake m4 is soo painful...maybe someday we'll look at Meson.

@rh-atomic-bot r+ f177a06

Cogitri:
Yep, meson is way nicer IMO :)

cgwalters:
Hmm, looks like a flake.

@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

cgwalters:
Hm, something started going seriously wrong in the payload link installed test; it looks like the output is enormous.  May have to log in to S3 and check size and delete some logs, and probably add size sanity checks to PAPR.

jlebon:
Wow, looks like it's about 18M. @giuseppe Can you have a look at this? This looks like there's way too many payload-link files.

jlebon:
@rh-atomic-bot retry

Issue #1491, opened 2018-03-09
Problem cancelling a pull op


joaquimrocha:
PR: https://github.com/ostreedev/ostree/pull/1492

---

I have recently started passing a`GCancellable` to an `ostree_repo_pull_with_options` operation. During some smoke testing, I noticed that I'd get a segfault caused by trying to propagate an error that was NULL, and the reason for this was that the `ostree_repo_prepare_transaction` 
 call (called from `ostree_repo_pull_with_options`) asserts it is not in a pending transaction, early returning FALSE (with no error set) in case it is.

What's happening is that in the first call to `ostree_repo_pull_with_options`, we cancel the op, but since we're using the same `GCancellable` for the `ostree_repo_abort_transaction` (which should reset the transactions to a state where there's no pending transaction), it'll just not reach the point where it resets the flag that `ostree_repo_prepare_transaction` cares for. This means that the crash doesn't happen all the time, but it's still possible to reproduce the crash with a bit of patience.

So a fix for this must ensure that a real transaction abortion happens, but there are a couple of approaches we can follow, all of them with their own quirks:

1) Just abort the transactions in `ostree_repo_prepare_transaction` if it's called with an on-going transaction set: this would correct the problem for any situation, however, it changes the behavior of a public function, and users may be relying on it, so we'd potentially break their code.

2) Always call `ostree_repo_abort_transaction` (it has an early return) before `ostree_repo_prepare_transaction` in the `ostree_repo_pull_with_options`, this ensures that any cancelled abortions won't impede a new pull from happening. Again, this could change the expected behavior of `ostree_repo_pull_with_options` (as users may expect a `FALSE` return if they're still in the middle of a transaction), but since there no error set anyway, this is probably the best approach!

3) Before calling `ostree_repo_abort_transaction` in `ostree_repo_pull_with_options`, verify that the `GCancellable` is not yet cancelled, if it is, call the abort without any cancellable; this won't change the expected behavior of `ostree_repo_pull_with_options`, but since we may be passing no cancellable to it, users will not have the possibility to cancel something that can potentially take long; besides, the original problem can still happen if the unused `GCancelled` is passed and it gets cancelled right before the right flags are reset...

So I would say we go with approach number 2, but talking to @pwithnall , he suggested something like 3) but just passing NULL as the cancellable all the time to the `ostree_repo_abort_transaction`; still, we both agree that maybe @cgwalters could take a look (also, this fix will not prevent the problem from happening in other places).

[...]

PR #1492, opened 2018-03-09
pull: Do not pass cancellable when aborting a transaction

joaquimrocha:
**Please check https://github.com/ostreedev/ostree/issues/1491 first!**

In ostree_repo_pull_with_options, if we pass a cancellable to
ostree_repo_abort_transaction, and that cancellable has been canceled
already, then the abort call will never reach the point where it resets
the needed flags. So the next time an operation is prepared (i.e. a pull
is attempted), it will fail because it believes it already has a pending
transaction.

To fix this we pass the cancellable as NULL. An alternative fix could be
to just check the state of the cancellable before passing it to the
abort operation, but this will not prevent the aforementioned situation
from happening.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -2153,6 +2153,12 @@ ostree_repo_abort_transaction (OstreeRepo     *self,
                                GCancellable   *cancellable,
                                GError        **error)
 {
+  /* Always ignore the cancellable to avoid the chance that, if it gets
+   * canceled, the transaction may not be fully cleaned up.
+   * See https://github.com/ostreedev/ostree/issues/1491 .
+   */
+  cancellable = NULL;
+
   /* Note early return */
   if (!self->in_transaction)
     return TRUE;


cgwalters:
bot, add author to whitelist

cgwalters:
Yep makes sense to me.  But `git grep abort_transaction` shows other callers too.  Mind fixing this up tree-wide?

cgwalters:
Alternatively...we could just do:

```
index 875743b4..9a4d11f4 100644
@@ -2153,6 +2153,12 @@ ostree_repo_abort_transaction (OstreeRepo     *self,
                                GCancellable   *cancellable,
                                GError        **error)
 {
+  /* Always ignore the cancellable; see https://github.com/ostreedev/ostree/issues/1491
+   * Basically a lot of callers will call this on error and we need
+   * to unset the transaction always.
+   */
+  cancellable = NULL;
+
   /* Note early return */
   if (!self->in_transaction)
     return TRUE;
```
?

joaquimrocha:
@cgwalters , I like your suggestion.
We lose the ability to cancel anything in that call that may get stalled but I guess it shouldn't happen that often.
I will change the PR to include that and credit you.

joaquimrocha:
Updated with the mentioned changes.

jlebon:
Thanks for the patch!
@rh-atomic-bot r+ e4ccae2

PR #1493, opened 2018-03-09
tests/str: Rework invocation

cgwalters:
Let's make our `run.sh` generically support any playbook. This is prep for
writing further tests in Ansible. Along with that, rework the Ansible so that
`tests.yml` is a playbook, and then the other bits are just task lists. It's
easier to read.

I also started to add a `use_git_build` variable with the idea that we'll be
able to run these same tests against an upstream image by setting that variable
off.


jlebon:
bot, retest this please

tests/fedora-str/playbook-run.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/bash
+# A thin wrapper for ansible-playbook which has a nice check for
+# TEST_SUBJECTS being set.
 set -xeuo pipefail
 
 dn=$(cd $(dirname $0) && pwd)
```
jlebon:
```
+ exec ansible-playbook --tags=atomic tests/fedora-str/sysinstall-tests.yml
ERROR! the playbook: tests/fedora-str/sysinstall-tests.yml could not be found
### EXITED WITH CODE 1 AFTER 1s
```

Hmm, probably we should just drop the `cd` below this line to make it easier to pass playbooks to this.

cgwalters:
OK, this one is fixed.

jlebon:
@rh-atomic-bot r+ 46957fe

PR #1494, opened 2018-03-12
tests: Don't follow symlinks for payload-link test

cgwalters:
Or things become recursive ∞.  Noticed this start breaking in a
recent PR.


giuseppe:
hm.. still I am not sure how this could be recursive, find should not follow symlinks.  In any case, the amount of logging generated is way too much

giuseppe:
bot, retest this please

cgwalters:
Yeah, I think you're right, it was the sha256 that was a bit slow and had me thinking we were recursing.  I reworked the PR again, let's see?  :arrow_up:

giuseppe:
it seems the tests are still not running, but the patch LGTM

jlebon:
LGTM too. Tests are not being triggered automatically currently. Will fire it off manually.
@rh-atomic-bot r=giuseppe dc03fb1

PR #1496, opened 2018-03-14
lib/fetcher: Allow clients to append to User-Agent

jlebon:
We do already have `http-headers`, which potentially could be used to
allow clients to completely override the field, but it seems like the
more common use case is simply to append.


jlebon:
Leaving WIP for feedback before adding tests.

alexlarsson:
Looks good to me.

cgwalters:
Looks sane to me too.

jlebon:
Dropped WIP and added tests! :arrow_up:

cgwalters:
@rh-atomic-bot r+ a13bb36

PR #1497, opened 2018-03-15
Deployment stage prep

cgwalters:
Prep for work on https://github.com/ostreedev/ostree/issues/545

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -37,13 +37,6 @@
 #include "ostree-bootloader-syslinux.h"
 #include "ostree-bootloader-grub2.h"
 
-static gboolean
-find_booted_deployment (OstreeSysroot       *self,
-                        GPtrArray           *deployments,
-                        OstreeDeployment   **out_deployment,
-                        GCancellable        *cancellable,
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-sysroot.c
```diff
@@ -847,25 +876,26 @@ ostree_sysroot_load_if_changed (OstreeSysroot  *self,
...
     {
       OstreeDeployment *deployment = deployments->pdata[i];
       ostree_deployment_set_index (deployment, i);
     }
 
-  if (!find_booted_deployment (self, deployments, &self->booted_deployment,
-                               cancellable, error))
-    return FALSE;
-  /* Sanity check; note the converse case is fine */
-  if (self->booted_deployment && !self->ostree_booted)
```
jlebon: This is still a valid sanity check though, right? Seems complementary to the one added higher up.

cgwalters: The logic now only sets `booted_deployment` if `self->ostree_booted` is `TRUE` now though right?  I could change it to `g_assert()`?

jlebon:
Ahh right, I see.
@rh-atomic-bot r+ e8a7ef1

PR #1500, opened 2018-03-16
lib/core: Support <remote>: syntax when listing refs

jlebon:
Allow users to pass `<remote>:` to list all refs we have locally
belonging to `<remote>`.


cgwalters:
LGTM, @alexlarsson want to do the `r+`?

alexlarsson:
I agree that it is nicer, with `remote:` and `remote:.`, but old versions of flatpak will still not work even if this is merged. I can update flatpak master and bump the its req to the latest ostree release, but anyone that updates ostree but not flatpak will still run into this issue.

jlebon:
But ostree is broken right now anyway, right? Given that flatpak releases faster than ostree, would it be safe to assume that by the time ostree has this fixed, there's also going to be an updated flatpak as well?

I'm fine adding a special-case for `:.` for backwards compatibility, just want to make sure there's value in carrying it in the first place.

alexlarsson:
I can update flatpak and if you're using the latest flatpak you'll likely pick up the latest ostree too.
However, I guess the issue is if you're using an older flatpak, but for some other reason you get the new ostree.

For instance, Fedora 27 has the stable flatpak (0.10.x), and I can see project atomic needing a new ostree version in F27, which could cause issues here.

jlebon:
OK, added a fixup! :arrow_up: 
It's part of the listing refs path *only* since for all other `ostree_parse_refspec` callers, we do want it to error out on `.`.

jlebon:
Hmm, actually, I think we should make this completely a `list_refs` thing. I don't think most users of the `parse_refspec` API (which is also public) would be ready to handle an empty `ref` portion.

Will get back to this in a bit.

cgwalters:
@rh-atomic-bot r+ 2964194

PR #1501, opened 2018-03-16
tests/str: Migrate some itests to nondestructive/ test group

cgwalters:
My main goal is to migrate the some shell script into Ansible (although that
only sort of started), but also split them into "destructive" and
"nondestructive". The nondestructive ones we can in the future run in parallel
on the same VM.

Since the filenames will be new (hence breaking `git annotate`), I made some
effort to track down the original git commit for the shell script test so it's
easy to know why the test was written later.

This also involved some tweaks so that the `nondestructive.yml` and
`sysinstall-tests.yml` playbooks can share more logic.


tests/fedora-str/nondestructive/bare-user-root.yml
```diff
@@ -0,0 +1,38 @@
...
+    done
+    echo some rootfs data > rootfs/usr/lib/cache.txt
+    # Commit using the host selinux policy
+    ostree --repo=repo commit --selinux-policy / -b rootfs --link-checkout-speedup --tree=dir=rootfs
+    ostree --repo=repo ls rootfs /usr/bin/systemd >ls.txt
+    assert_file_has_content ls.txt '^-007.. 0 0 .*/usr/bin/systemd'
+    ostree --repo=repo ls -X rootfs /usr/lib/dbus-daemon-helper >ls.txt
+    assert_file_has_content ls.txt '^-007.. 0 81 .*security.selinux.*/usr/lib/dbus-daemon-helper'
+    assert_not_file_has_content ls.txt 'user.ostreemeta'
+    echo "ok bare-user link-checkout-speedup with modified xattrs maintains uids"
```
jlebon: Were you planning to "convert" this to ansible? One of my things I hope we avoid is writing lots of shell in YAML. WDYT about just keeping this a regular shell script and just copying it over and executing it?

cgwalters: It is a mixed bag for sure.  I was hoping the end game was Ansible but getting there is going to take a while.

What do you see as the advantage of manual copying?  Making the playbooks smaller?  In this case you're seemingly really arguing for not doing this change at all?

jlebon: I guess I'd like us to use ansible for its strengths while trying to avoid its weaknesses. To me, what makes ansible awesome is its remoting capabilities and the ecosystem around it (e.g. the s-t-r qcow2 provisioning). It's also really good at structuring things using tags and roles.

What I'm not convinced on is whether ansible is good for actually writing the tests themselves. I know a-h-t for example has had good success, though it mostly deals with high-level CLIs.

OTOH if I look at this script here or any of our vmcheck tests, I worry migrating those to pure ansible would make things much harder to maintain and extend. It just feels like trying to make it do something it wasn't designed for. The callback plugin and `allow_duplicates` hacks in a-h-t are good examples of that.

Seeing this shell in YAML reminded me of when PAPR was completely implemented in JJB YAML. At some point, it included some Python in shell in shell in YAML... :/ Of course, shell scripting has its set of issues as well, though it feels like it still comes out ahead, especially if we can still leverage ansible for the parts that are more sticky, like we started to do in rpm-ostree.

I think this is my strawman:

- we use ansible for the s-t-r provisioning
- we use ansible for tagging and organizing tests, e.g.  destructive/nondestructive
- we use ansible to run tests and collect their outputs
- but the tests themselves are still written in their own scripts with whatever composition makes sense there to make life easier, e.g. `libtest.sh`, `libinsttest.sh`, etc...

This doesn't restrict us to tests that can't e.g. survive a reboot. The rpm-ostree dist-git test for example just uses `delegate_to` to run tests directly on the executing machine. That could just be a different category of tests, I suppose?

cgwalters: > but the tests themselves are still written in their own scripts

scripts = shell script?  That seems to cut us off from sharing more code with things that actually use Ansible for controlling the host too (existing openshift-ansible, a-h-t) etc.

One of the tests in this PR (`remotes`) was entirely rewritten into ansible (granted it's tiny).

I guess I'd been thinking down the line we try to recreate the bits from `libtest.sh` for Ansible and reduce our use of shell.

cgwalters: > At some point, it included some Python in shell in shell in YAML...

What I'd been thinking of to avoid that is establishing patterns to better intermix Ansible and shell; something like using Ansible to e.g. parse/filter JSON, then perhaps writing shell-source-able files - or passing via environment?  It's surprising the command module apparently doesn't support passing environment variables.

jlebon: > What I'd been thinking of to avoid that is establishing patterns to better intermix Ansible and shell

Yeah, that sounds good to me! That's similar to what we're doing right now in rpm-ostree. ITSM that's more a case of ansible supporting the shell scripts, rather than the other way around, right? And still leaves the door open for sharing code. E.g. we should be able to easily replace `vm_reboot` with a-h-t's `reboot` role.

In any case, I'm not fundamentally opposed to this change and open to exploring this path if you'd like to try it out.

miabbott: > What I'm not convinced on is whether ansible is good for actually writing the tests themselves. I know a-h-t for example has had good success, though it mostly deals with high-level CLIs.

> OTOH if I look at this script here or any of our vmcheck tests, I worry migrating those to pure ansible would make things much harder to maintain and extend. It just feels like trying to make it do something it wasn't designed for. The callback plugin and allow_duplicates hacks in a-h-t are good examples of that.

I'm in agreement with this sentiment.  I'm regularly saying that we are abusing Ansible to do things it was never intended to do.  Having said that, if it ain't broke, don't fix it...so we are continuing down that path.

> It's surprising the command module apparently doesn't support passing environment variables.

But `shell:` does.

```
- name: echo
  shell:  "echo $VAR"
  environment:
    VAR: "foo"
```

miabbott: To continue this train of thought, I like what @jlebon proposed about how to use Ansible for these tests.  I like the idea of keeping the shell scripts intact and using Ansible to copy them around and fetch logs, etc.

cgwalters: > And still leaves the door open for sharing code. E.g. we should be able to easily replace vm_reboot with a-h-t's reboot role.

Funny enough I actually was playing with that and your clever use of `boot_id` felt hard to replicate exactly there (we need to loop on the tasks).

I think though the "shell calling ansible" path is fairly unique to what's in rpm-ostree today...I had been envisioning that we eventually invert it like what's going on here.
But that "eventually" is going to be a fair ways away.

I think what I'm trying to do basically is elevate Ansible to the point where tests can be written directly in it to start *in addition* to being written in shell.

cgwalters: Ah...the `environment` thing is actually fully generic: http://docs.ansible.com/ansible/latest/playbooks_environment.html

jlebon:
Sorry, forgot to get back to this! Let's just get it in and try it out.
@rh-atomic-bot r+ db603bc

cgwalters:
@rh-atomic-bot r-

cgwalters:
Well since this stalled I'd started reworking it: https://github.com/ostreedev/ostree/pull/1509

And I have other patches which will conflict.

cgwalters:
@rh-atomic-bot force

cgwalters:
Closing in favor of https://github.com/ostreedev/ostree/pull/1509

PR #1502, opened 2018-03-16
rust/bupsplit: minor idiomatic fixes

lucab:
This fixes a few un-idiomatic bits in Rust bupsplit code, getting rid
of some unchecked casts and an assert statement.


cgwalters:
Nice, thanks!

@rh-atomic-bot r+ 5b3cfb8

PR #1503, opened 2018-03-18
WIP: Add concept of "staged" deployment

cgwalters:
Add API to write a deployment state to `/run/ostree/staged-deployment`,
along with a systemd service which runs at shutdown time.

Just compile tested, still WIP.

For: https://github.com/projectatomic/rpm-ostree/issues/40


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2456,61 +2452,259 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
+ * @error: Error
+ *
+ * Check out deployment tree with revision @revision, performing a 3
+ * way merge with @provided_merge_deployment for configuration.
+ *
+ * While this API is not deprecated, you most likely want to use the
+ * ostree_sysroot_stage_tree() API.
+ */
+gboolean
+ostree_sysroot_deploy_tree (OstreeSysroot     *self,
```
jlebon: Maybe let's start by splitting out this `initialize` + `finalize` separation in a prep PR?

cgwalters:
Depends: https://src.fedoraproject.org/rpms/ostree/pull-request/8

cgwalters:
TODO:

 - Clean up the story around kargs
 - Determine whether staged deployment is part of `_get_deployments()`.  Tradeoffs both ways.  If not (as currently) then all tools will need to learn to query for it.
 - tests

cgwalters:
Sigh, needed to move it into the main `/usr/bin/ostree` as that's `install_exec_t` which is what we need to do deployments.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2404,50 +2415,301 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
+  if (!glnx_unlinkat (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED, 0, error))
+    return FALSE;
+
+  if (!sysroot_finalize_deployment (self, self->staged_deployment, NULL, merge_deployment,
+                                    cancellable, error))
+    return FALSE;
+
+  /* TODO: Proxy across flags too? */
+  OstreeSysrootSimpleWriteDeploymentFlags flags = 0;
+  if (!ostree_sysroot_simple_write_deployment (self, ostree_deployment_get_osname (self->staged_deployment),
```
peterbaouoft: Question/minor: noticed in rpm-ostree, seems like we had `rpmostree_syscore_write_deployment` to add some flags for livefs and rollback, do we need to consider a bit here to make it easier to integrate this into rpm-ostree in the future?(if needed :p)

src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -270,6 +270,13 @@ cleanup_old_deployments (OstreeSysroot       *self,
       g_hash_table_replace (active_boot_checksums, bootcsum, bootcsum);
     }
 
+  /* And also the staged deployment, if any */
+  if (self->staged_deployment)
+    {
+      char *deployment_path = ostree_sysroot_get_deployment_dirpath (self, self->staged_deployment);
+      g_hash_table_replace (active_deployment_dirs, deployment_path, deployment_path);
```
peterbaouoft: Question: wanted to confirm, we don't include bootcsum here because the "staged deployment" is not ready for boot yet? ( i.e the merge and boot entries weren't ready)

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2404,50 +2415,301 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
+
+  if (merge_deployment)
+    g_variant_builder_add (builder, "{sv}", "merge-deployment",
+                           serialize_deployment_to_variant (merge_deployment));
+
+  if (override_kernel_argv)
+    g_variant_builder_add (builder, "{sv}", "kargs",
+                           g_variant_new_strv ((const char *const*)override_kernel_argv, -1));
+
+  char *dnbuf = strdupa (_OSTREE_SYSROOT_RUNSTATE_STAGED);
```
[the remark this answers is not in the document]
cgwalters: The basic idea of `/run` is it's guaranteed to go away if the system reboots (it's a `tmpfs`, stored only in memory).  Which I think helps simplify the logic around failure cases.  I feel like we only want to try to process staging data once rather than having it be persistent.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2404,50 +2415,301 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
+  return TRUE;
+}
+
+/* Invoked at shutdown time by ostree-deploy-staged.service */
+gboolean
+_ostree_sysroot_deploy_staged (OstreeSysroot *self,
+                               GCancellable  *cancellable,
+                               GError       **error)
+{
+  /* The service has a ConditionPathExists=/run/ostree/staged-deployment */
```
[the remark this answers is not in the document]
cgwalters: That's not minor, that's a bug!  The service would fail if there was no staged deployments.  Fixed.

src/libostree/ostree-sysroot-deploy.c
cgwalters: I think you're right that integrating this nicely will require pulling down some of that logic.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2388,6 +2398,7 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
     return FALSE;
 
   _ostree_deployment_set_bootcsum (new_deployment, kernel_layout->bootcsum);
+  _ostree_deployment_set_bootconfig_from_kargs (new_deployment, override_kernel_argv);
```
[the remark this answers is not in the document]
cgwalters: Yep, that was all duplicate code.  Fixed, thanks!

src/libostree/ostree-sysroot-cleanup.c
cgwalters: Nope, that was an oversight.  One not revealed because we weren't installing a staged deployment with a new boot checksum.

I have kind of been going back and forth on whether `/boot` should be modified during staging or finalization.  Currently it's during staging.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -808,6 +809,7 @@ static gboolean
 write_origin_file_internal (OstreeSysroot         *sysroot,
                             OstreeSePolicy        *sepolicy,
                             OstreeDeployment      *deployment,
+                            gboolean               if_not_exists,
```
jlebon: I'm confused why we need this. Why not just unconditionally overwrite it in `finalize`? Then even in the stage path, we're sure of using the right label for it post-merge.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2388,66 +2398,308 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
+  return TRUE;
+}
+
+/* Serialize information about a deployment to a variant, used by the staging
+ * code.
+ */
+static GVariant *
+serialize_deployment_to_variant (OstreeDeployment *deployment)
+{
+  g_autoptr(GVariantBuilder) builder = g_variant_builder_new ((GVariantType*)"a{sv}");
```
jlebon: Should we stick with `g_auto` where we can for good form?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2388,66 +2398,308 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
+  g_autofree char *name =
+    g_strdup_printf ("%s.%d", ostree_deployment_get_csum (deployment),
+                     ostree_deployment_get_deployserial (deployment));
```
jlebon: How about keeping those separate to make deconstruction lower down easier?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2388,66 +2398,308 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
+  if (merge_deployment)
+    g_variant_builder_add (builder, "{sv}", "merge-deployment",
+                           serialize_deployment_to_variant (merge_deployment));
+
+  if (override_kernel_argv)
+    g_variant_builder_add (builder, "{sv}", "kargs",
+                           g_variant_new_strv ((const char *const*)override_kernel_argv, -1));
+
+  char *dnbuf = strdupa (_OSTREE_SYSROOT_RUNSTATE_STAGED);
+  const char *parent = dirname (dnbuf);
```
jlebon: Minor: any reason not to collapse these two lines? (i.e. `dirname (strdupa (...))`)

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2388,66 +2398,308 @@ ostree_sysroot_deploy_tree (OstreeSysroot     *self,
...
+                           OstreeDeployment  *merge_deployment,
+                           char             **override_kernel_argv,
+                           OstreeDeployment **out_new_deployment,
+                           GCancellable      *cancellable,
+                           GError           **error)
+{
+  g_autoptr(OstreeDeployment) deployment = NULL;
+  if (!sysroot_initialize_deployment (self, osname, revision, origin, override_kernel_argv,
+                                      &deployment, cancellable, error))
+    return FALSE;
```
jlebon: Hmm, I wonder here if we should do something if there's a deployment already staged. E.g. in rpm-ostree, if we have a `staged` auto-updates policy, we could just continuously be staging newer updates as we see them right? I guess we can put the cleanup there, though it does feel like someone calling `ostree_sysroot_stage_tree` expects it to implicitly delete a previously staged deployment, rather than having them pile up.

src/libostree/ostree-sysroot.c
```diff
@@ -880,17 +939,23 @@ ostree_sysroot_load_if_changed (OstreeSysroot  *self,
         }
     }
 
-  if (self->ostree_booted && self->root_is_sysroot
-      && !self->booted_deployment)
+  const gboolean root_is_ostree_booted =
+    self->ostree_booted && self->root_is_sysroot;
```
jlebon: At this point, we might as well make replace instances of those with a new field.

src/ostree/ot-admin-builtin-deploy.c
```diff
@@ -157,31 +159,41 @@ ot_admin_builtin_deploy (int argc, char **argv, OstreeCommandInvocation *invocat
...
-  if (!ostree_sysroot_deploy_tree (sysroot, opt_osname, revision, origin, merge_deployment,
-                                   kargs_strv, &new_deployment, cancellable, error))
-    return FALSE;
-
-  OstreeSysrootSimpleWriteDeploymentFlags flags = OSTREE_SYSROOT_SIMPLE_WRITE_DEPLOYMENT_FLAGS_NO_CLEAN;
-  if (opt_retain)
-    flags |= OSTREE_SYSROOT_SIMPLE_WRITE_DEPLOYMENT_FLAGS_RETAIN;
-  else
+  if (opt_stage)
     {
```
jlebon: Let's just error out here if any of the `--retain` opts are given?

src/boot/ostree-deploy-staged.service
```diff
@@ -0,0 +1,36 @@
...
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
+
+# For some implementation discussion, see:
+# https://lists.freedesktop.org/archives/systemd-devel/2018-March/040557.html
+[Unit]
+Description=OSTree Deploy Staged
+ConditionPathExists=/run/ostree-booted
```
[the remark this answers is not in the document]
cgwalters: Yeah, can't do that, see https://lists.freedesktop.org/archives/systemd-devel/2018-March/040557.html

src/libostree/ostree-sysroot-deploy.c
cgwalters: Eh, we already need to parse them from the filesystem and had a function for it.

cgwalters:
Pushed a fixup :arrow_up: which addresses a lot of comments (but not all).  In the meantime split out prep patches into https://github.com/ostreedev/ostree/pull/1535

jlebon:
Let's do this!
@rh-atomic-bot r+ 7adfe24

PR #1504, opened 2018-03-19
main, status: Factor out deployment printing into helper

cgwalters:
Prep for staged deployments; they won't be in the primary deployment
list, and we want to print them first.

Also pull in some code from rpm-ostree for the red/bold bits and use
that tree-wide.

Update submodule: libglnx


jlebon:
@rh-atomic-bot r+ ffabf10

PR #1505, opened 2018-03-19
Two deployment stage prep patches

cgwalters:


jlebon:
Nice! @rh-atomic-bot r+ 50356dc

cgwalters:
@rh-atomic-bot retry

PR #1507, opened 2018-03-21
ci/papr: Fix artifacts

cgwalters:
Of course we only notice these things are broken when the CI breaks.

Also add some comments and flip flatpak to `required: true` since it should be
now.


jlebon:
LGTM! Though let's make sure things get artifacted correctly.

jlebon:
@rh-atomic-bot r+ f55545f

PR #1508, opened 2018-03-22
switchroot: Ensure /run/ostree-booted is created even without initramfs

cgwalters:
See https://mail.gnome.org/archives/ostree-list/2018-March/msg00012.html

If ostree-prepare-root is run as pid 1 (i.e we're not using an initramfs), then
anything we write outside the target sysroot (such as `/run/ostree-booted`) will
be lost.

Since `ostree-remount.service` runs fairly early in boot, and is triggered via
`ConditionKernelCommandLine=ostree`, we can just touch the file there in
addition.

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -104,4 +104,19 @@ read_proc_cmdline_ostree (void)
   return ret;
 }
 
+/* This is an API for other projects to determine whether or not the
+ * currently running system is ostree-controlled.
+ */
+static inline void
+touch_run_ostree (void)
+{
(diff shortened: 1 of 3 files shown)


cgwalters:

akiernan:
@rh-atomic-bot r+ 001e843

cgwalters:
Blah, I'm going to have to debug that flake.

@rh-atomic-bot retry

cgwalters:
OK, fix for that flake is in https://github.com/ostreedev/ostree/pull/1509

cgwalters:
@rh-atomic-bot retry

PR #1509, opened 2018-03-22
tests/str: Rework sysinstall-tests to be an entrypoint

cgwalters:
Prep for creating more types of tests.

Move copying of `tests/` into the sysinstall-tests rather than `overlay-git`
as not all test types may need that.

Factored out of https://github.com/ostreedev/ostree/pull/1501


cgwalters:
OK, some more patches here.

tests/installed/itest-payload-link.sh
```diff
@@ -57,29 +56,25 @@ fi
 mkfs.xfs -m reflink=1 ${blkdev}
 
 mount ${blkdev} mnt
+cd mnt
 
-test_tmpdir=$(pwd)/mnt
-cd ${test_tmpdir}
-
+# Test that coreutils will do reflink
```
cgwalters: Also I changed this to just require that reflinks work since that's the point of the test.

cgwalters:
:green_heart:

jlebon:
@rh-atomic-bot r+ 263cb35

jlebon:
(Also, thanks for splitting this up into smaller digestible commits!)

PR #1510, opened 2018-03-22
lib/deltas: Squash some GCC maybe-uninitialized warnings

cgwalters:
These show up in the RPM build, I didn't yet try to figure out why we're not
reproducing them outside of that.


jlebon:
LGTM! Offhand it looks like all of those could have been deduced as safe by a smarter compiler. I wonder if that's why we weren't hitting it (e.g. some clang vs gcc thing maybe).

@rh-atomic-bot r+ 6f7fe14

PR #1512, opened 2018-03-22
 Release 2018.4

cgwalters:
A quick turnaround to include one PR: #1508
"switchroot: Ensure /run/ostree-booted is created even without initramfs"

This fixes ostree when booting without an initramfs.  Thanks to @akiernan for the
bug report and helping review the fix!  I'm working on enhancing
the test suite, which will help in adding some coverage here.

Also for this release I'm going to avoid adding a "stub" symbol section
to the `-released.sym` file; I don't believe it's necessary.


jlebon:
@rh-atomic-bot r+ d2f007c

PR #1514, opened 2018-03-23
Two deployment stage prep patches

cgwalters:
Prep for https://github.com/ostreedev/ostree/pull/1503


jlebon:
Nice! @rh-atomic-bot r+ 81aa7a0

PR #1515, opened 2018-03-23
lib/deploy: Port final bootconfig writing to new style

cgwalters:
The main blocker for doing this before was the `goto out` handling
for remounting `/boot`.  Handle that by factoring out the bits that
require it to a helper function, and do the C/GError equivalent of
"try/finally".

Not prep for anything right now, just decided to do this since I had the file
open.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2241,8 +2241,7 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
                * jump in and open a file there.  See above TODO
                * around doing this in a new mount namespace.
                */
-              int errsv = errno;
-              g_printerr ("warning: Failed to remount /boot read-only: %s\n", strerror (errsv));
+              g_printerr ("warning: Failed to remount /boot read-only: %s\n", strerror (errno));
             }
         }
       if (!success)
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2127,147 +2211,73 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
...
+          if (mount ("/boot", "/boot", NULL, MS_REMOUNT | MS_RDONLY | MS_SILENT, NULL) < 0)
             {
-              g_prefix_error (error, "Bootloader write config: ");
-              goto out;
+              /* Only make this a warning because we don't want to
+               * completely bomb out if some other process happened to
+               * jump in and open a file there.  See above TODO
+               * around doing this in a new mount namespace.
+               */
+              int errsv = errno;
```
jlebon: I know this isn't new, but that line confuses me. Was the intent here to save `errno` *before* the "finally" `mount` and restore it afterwards?

cgwalters: I think it's more that in the past I've been bitten by the error handling functions themselves overwriting `errno`.  Imagine if e.g. you want to do `g_strdup_printf()` to format an error message, and then go on to reference `errno`..  But in this case it's safe to drop.

jlebon:
@rh-atomic-bot r+ cf603f2

PR #1516, opened 2018-03-26
tests/installed: Fix TESTS= being empty

cgwalters:
I broke this in https://github.com/ostreedev/ostree/pull/1509/commits/9b55aaea6f34b7094c44932a3c2e1cf2d54634fd
I'd obviously tested *setting* it locally worked, but I didn't test that
not having it set ran all the tests.

I don't understand why we were doing the `+ ` pattern before; let's
just check if it's empty.

### DIFF
--- a/tests/installed/run.sh
+++ b/tests/installed/run.sh
@@ -4,7 +4,7 @@ set -xeuo pipefail
 
 dn=$(dirname $0)
 for tn in ${dn}/itest-*.sh; do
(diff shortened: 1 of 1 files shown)


cgwalters:
This should fix https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/263cb3565e2f5e2fe1c37747e00a6183d7b5ad48.0.1521737370394136088/artifacts/installed-tests.log

Issue #1233, opened 2017-10-02
Add support for alternative signature systems


jlaako:
I am planning to implement alternative signing/verification system to current gpgme. This would benefit from having a plugin system, or more complete internal API abstracting signature parts.

At the moment gpgme approach has two challenges. One is that it brings in GPLv3 through gnupg dependency which is not optimal for all cases. And another one is performance challenges on large OS rebuilds that can happen in distro development phase in CI systems and developer systems. This is due to way gpgme interfaces with gnupg binary.

My current plan is to utilize PKCS#7 signature and X.509 certificate standards which work well for signature purposes. This also allows flexibility of choice for the used algorithms. GnuTLS has good support for both and the implementation should be fairly straightforward.

At the moment, clean implementation is somewhat challenging because ostree directly relies on gpgme/gpg concepts, instead of abstraction. Some internal abstraction for this part would be nice. One option is plugin interface, but I am assuming one would rarely use more than one signature scheme, so configure-time selection could be more straightforward. Unless there is particular need for multiple independent source repos on one system, in which case plugin system would be good. More schemes than these two could be added in future.

cgwalters:
Hm.  It may be hard to get away from supporting both schemes being compiled in.  I mentioned this somewhere else; but I'd definitely like to make it easier for external users to plug in their own verification.  Right now all logic like the timestamp checks lives inside the libostree pull layer, but if we had a generic API that provided a callback with commit objects we were going to write, then higher level apps could perform any arbitrary signature or security checks they wanted.

Also relatedly, one thing I'd like to do probably is have the signature verification run in a subprocess, and then we can sandbox that subprocess.  (This would of course be ironic in the case of gpg, as libgpgme runs gpg as a subprocess, which in turn forks a daemon, which [can leak](https://github.com/ostreedev/ostree/issues/727#issuecomment-285771058)).

cgwalters:
Right now things like `ostree show` do indeed support showing the commit signature status, and rendering it.  Doing full support for an alternative scheme would indeed involve a lot of work.

But what I'm thinking instead (and I'd like to do with GPG for that matter) is to hook the signature status as an xattr or something off the commit object.  xattr space restrictions may be an issue though; failing that we could look at adding a generic local-only detached metadata that's lifecycle bound to the commit.

cgwalters:
However, all of this aside, you still need to support entirely disabling the GPG dependency, right?

jlaako:
[...]

PR #1518, opened 2018-03-28
Add timestamps to OstreeRepoFinderResult

mwleeds:
cc: @pwithnall

### DIFF
--- a/src/libostree/ostree-repo-finder.h
+++ b/src/libostree/ostree-repo-finder.h
@@ -99,6 +99,8 @@ GPtrArray *ostree_repo_finder_resolve_all_finish (GAsyncResult  *result,
  * @ref_to_checksum: (element-type OstreeCollectionRef utf8): map of collection–ref
  *    pairs to checksums provided by this remote; values may be %NULL to
  *    indicate this remote doesn’t provide that ref
+ * @ref_to_timestamp: (element-type OstreeCollectionRef guint64) (nullable): map of
+ *    collection–ref pairs to timestamps; values may be 0 for various reasons
  * @summary_last_modified: Unix timestamp (seconds since the epoch, UTC) when
  *    the summary file on the remote was last modified, or `0` if unknown
  *
@@ -122,6 +124,15 @@ GPtrArray *ostree_repo_finder_resolve_all_finish (GAsyncResult  *result,
  * should be available locally, so the details for each checksum can be looked
  * up using ostree_repo_load_commit().
  *
+ * @ref_to_timestamp provides timestamps for the set of refs in
+ * @ref_to_checksum. The refs are keys (of type #OstreeCollectionRef) and the
+ * values are guint64 pointers with the timestamp associated with the checksum
+ * provided in @ref_to_checksum. @ref_to_timestamp can be %NULL, and when it's
+ * not, the timestamps are zero when any of the following conditions are met:
+ * (1) the override-commit-ids option was used on
+ * ostree_repo_find_remotes_async (2) there was an error in trying to get the
+ * commit metadata (3) the checksum for this ref is %NULL in @ref_to_checksum.
+ *
  * Since: 2017.8
  */
 typedef struct
@@ -130,6 +141,7 @@ typedef struct
   OstreeRepoFinder *finder;
   gint priority;
   GHashTable *ref_to_checksum;
+  GHashTable *ref_to_timestamp;
   guint64 summary_last_modified;
 
   /*< private >*/
@@ -144,6 +156,7 @@ OstreeRepoFinderResult *ostree_repo_finder_result_new (OstreeRemote     *remote,
                                                        OstreeRepoFinder *finder,
                                                        gint              priority,
                                                        GHashTable       *ref_to_checksum,
+                                                       GHashTable       *ref_to_timestamp,
                                                        guint64           summary_last_modified);
 _OSTREE_PUBLIC
 OstreeRepoFinderResult *ostree_repo_finder_result_dup (OstreeRepoFinderResult *result);
--- a/src/libostree/ostree-repo-finder.h
+++ b/src/libostree/ostree-repo-finder.h
@@ -141,11 +141,11 @@ typedef struct
   OstreeRepoFinder *finder;
   gint priority;
   GHashTable *ref_to_checksum;
-  GHashTable *ref_to_timestamp;
   guint64 summary_last_modified;
+  GHashTable *ref_to_timestamp;
 
   /*< private >*/
-  gpointer padding[4];
+  gpointer padding[3];
 } OstreeRepoFinderResult;
 
 _OSTREE_PUBLIC
--- a/src/libostree/ostree-repo-finder-avahi.c
+++ b/src/libostree/ostree-repo-finder-avahi.c
@@ -844,7 +844,7 @@ ostree_avahi_service_build_repo_finder_result (OstreeAvahiService               
         }
 
       g_ptr_array_add (results, ostree_repo_finder_result_new (remote, OSTREE_REPO_FINDER (finder),
-                                                               priority, supported_ref_to_checksum,
+                                                               priority, supported_ref_to_checksum, NULL,
                                                                GUINT64_FROM_BE (g_variant_get_uint64 (summary_timestamp))));
     }
 }
--- a/src/libostree/ostree-repo-finder-config.c
+++ b/src/libostree/ostree-repo-finder-config.c
@@ -192,7 +192,7 @@ ostree_repo_finder_config_resolve_async (OstreeRepoFinder                  *find
           continue;
         }
 
-      g_ptr_array_add (results, ostree_repo_finder_result_new (remote, finder, priority, supported_ref_to_checksum, 0));
+      g_ptr_array_add (results, ostree_repo_finder_result_new (remote, finder, priority, supported_ref_to_checksum, NULL, 0));
     }
 
   g_ptr_array_sort (results, results_compare_cb);
--- a/src/libostree/ostree-repo-finder-mount.c
+++ b/src/libostree/ostree-repo-finder-mount.c
@@ -545,7 +545,7 @@ G_GNUC_END_IGNORE_DEPRECATIONS
            * the code in ostree_repo_pull_from_remotes_async() will be able to
            * check it just as quickly as we can here; so don’t duplicate the
            * code. */
-          g_ptr_array_add (results, ostree_repo_finder_result_new (remote, finder, priority, supported_ref_to_checksum, 0));
+          g_ptr_array_add (results, ostree_repo_finder_result_new (remote, finder, priority, supported_ref_to_checksum, NULL, 0));
         }
     }
 
--- a/src/libostree/ostree-repo-finder-override.c
+++ b/src/libostree/ostree-repo-finder-override.c
@@ -243,7 +243,7 @@ ostree_repo_finder_override_resolve_async (OstreeRepoFinder                  *fi
   g_hash_table_iter_init (&iter, repo_remote_to_refs);
 
   while (g_hash_table_iter_next (&iter, (gpointer *) &remote, (gpointer *) &supported_ref_to_checksum))
-    g_ptr_array_add (results, ostree_repo_finder_result_new (remote, finder, priority, supported_ref_to_checksum, 0));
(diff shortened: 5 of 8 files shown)


mwleeds:
bot, retest this please

mwleeds:
(The tests pass locally and the error when the CI runs them isn't specific to this PR.)

cgwalters:
> (The tests pass locally and the error when the CI runs them isn't specific to this PR.)

I'm hesitating to agree; it's *just* an experimental API test that failed and this PR only changes experimental API bits.

cgwalters:
FWIW I just pulled your branch and built locally and got the same failure.

cgwalters:
Are you sure you built locally with `--enable-experimental-api`?

mwleeds:
Maybe I confused the timeout error and a seg fault. Maybe it was just the timeout that happened in the other PRs. Was it a seg fault you saw locally?

mwleeds:
Oddly, on my Fedora desktop there wasn't an error but on this Endless laptop there is. I'll look into it.

mwleeds:
I get the same error on the master branch, so I think it's separate, but I can still look into it.

cgwalters:
> Was it a seg fault you saw locally?

Yep.

```
coredumpctl info 1635
           PID: 1635 (lt-test-repo-fi)
           UID: 1000 (walters)
           GID: 1000 (walters)
        Signal: 11 (SEGV)
     Timestamp: Wed 2018-03-28 17:01:50 EDT (2h 42min ago)
  Command Line: /home/walters/src/github/ostreedev/ostree/tests/.libs/lt-test-repo-finder-config -k --tap
    Executable: /srv/walters/src/github/ostreedev/ostree/tests/.libs/lt-test-repo-finder-config
 Control Group: /user.slice/user-0.slice/session-c3.scope
          Unit: session-c3.scope
         Slice: user-0.slice
       Session: c3
     Owner UID: 0 (root)
       Boot ID: b75b277163a34b01870224b19c25c157
    Machine ID: e08ac32f9a5246a58aa97d306c2fbb92
      Hostname: vanadium
       Storage: /var/lib/systemd/coredump/core.lt-test-repo-fi.1000.b75b277163a34b01870224b19c25c157.1635.1522270910000000.lz4
       Message: Process 1635 (lt-test-repo-fi) of user 1000 dumped core.
                
                Stack trace of thread 13794:
                #0  0x00007f3896c10980 OSTREE_IS_REPO_FINDER (/srv/walters/src/github/ostreedev/ostree/.libs/libostree-1.so.1.0.0)
                #1  0x00007f3896c1db44 is_valid_finder_array (/srv/walters/src/github/ostreedev/ostree/.libs/libostree-1.so.1.0.0)
                #2  0x00007f3896c1dfde ostree_repo_find_remotes_async (/srv/walters/src/github/ostreedev/ostree/.libs/libostree-1.so.1.0.0)
                #3  0x0000000000404f04 test_repo_finder_config_find_remotes (lt-test-repo-finder-config)
                #4  0x00007f3894bf92ea g_test_run_suite_internal (libglib-2.0.so.0)
                #5  0x00007f3894bf921b g_test_run_suite_internal (libglib-2.0.so.0)
                #6  0x00007f3894bf94c2 g_test_run_suite (libglib-2.0.so.0)
                #7  0x00007f3894bf94e1 g_test_run (libglib-2.0.so.0)
                #8  0x0000000000405c24 main (lt-test-repo-finder-config)
                #9  0x00007f38945d4f2a __libc_start_main (libc.so.6)
                #10 0x000000000040360a _start (lt-test-repo-finder-config)
```

mwleeds:
I did a `make distclean` and a reboot and now I can't reproduce any errors with `test-repo-finder-config`, and the ones on the CI look like timeouts. Not sure why it says they're seg faults though.

mwleeds:
Oh, interesting. The one you posted looks more relevant than what I have:
[ostree-test-suite.valgrind.log](https://github.com/ostreedev/ostree/files/1858210/ostree-test-suite.valgrind.log)

mwleeds:
Found the problem. Oops

pwithnall:
The two little cleanup commits at the start of the branch look good to me, and you might want to split them into a separate PR so they can get merged ASAP.

I’m not so comfortable with the approach for the main patch, though, for several reasons:
 * Adding an additional hash table with exactly the same keys as `ref_to_checksum` is a waste of memory compared to changing the values of `ref_to_checksum` to be a struct containing checksum and timestamp (for example — but I am not going to suggest doing that).
 * It only exposes timestamps, and not the rest of the commit metadata. We’re going to have to add _another_ hash table if we later find we want to expose more of the commit metadata for clients to make a decision.
 * The repo finding operation already guarantees to return checksums as non-`NULL` only if they’re the latest found across all peers. Is the problem here that you don’t have enough information to make a decision if you’ve also enabled searching for downgrades; or is the problem that if all the peers have an older OSTree commit, you have no way of telling that?
 * OSTree already has API for turning a checksum into commit metadata (`ostree_repo_load_commit()`), so why not use that?

So I have two alternative suggestions:
 1. Use `ostree_repo_load_commit()` to load the commit metadata in your eos-updater/flatpak code, drop the changes in this PR and instead modify `OstreeRepoFinder` to make sure the commit metadata is always downloaded and saved to disk ready for `ostree_repo_load_commit()`. (Grep for `Fill in any gaps in the metadata` in `ostree-repo-pull.c`.)
 2. Or, expose something like the `commit_metadatas` hash table from `ostree-repo-pull.c`, which will map checksums to commit metadata; rather than just exposing something which maps refs to timestamps.

cgwalters:
I haven't looked at the full stack here yet, but one model is indeed to pull the commits with `OSTREE_REPO_PULL_FLAGS_COMMIT_ONLY`.  Then they can be loaded.

We could probably add a generic "callback" mechanism to pulls that passed the ref/commit object pair; I was musing on that as a way to allow [external signatures](https://github.com/ostreedev/ostree/issues/1233).

mwleeds:
> The repo finding operation already guarantees to return checksums as non-NULL only if they’re the latest found across all peers. Is the problem here that you don’t have enough information to make a decision if you’ve also enabled searching for downgrades; or is the problem that if all the peers have an older OSTree commit, you have no way of telling that?

The second one. An offline computer has no way of knowing, just from the OstreeRepoFinderResult data, if the latest commit available from all remotes is an update or a downgrade.

> Use ostree_repo_load_commit() to load the commit metadata in your eos-updater/flatpak code, drop the changes in this PR and instead modify OstreeRepoFinder to make sure the commit metadata is always downloaded and saved to disk ready for ostree_repo_load_commit(). (Grep for Fill in any gaps in the metadata in ostree-repo-pull.c.)

This would mean pulling every remote commit into your local repo, which seems terribly inefficient when all you need is a timestamp, and you're doing the same thing `find_remotes_cb` just did, again.

> Or, expose something like the commit_metadatas hash table from ostree-repo-pull.c, which will map checksums to commit metadata; rather than just exposing something which maps refs to timestamps.

This seems reasonable, although I'm not sure how big the metadata usually is, so I'm not sure how much extra memory we'd be using.

mwleeds:
> changing the values of ref_to_checksum to be a struct containing checksum and timestamp (for example — but I am not going to suggest doing that).

One disadvantage to this is I think you'd have to have some ugly checks in clients-- if ostree version is x, use the data this way, otherwise that way.

pwithnall:
> This would mean pulling every remote commit into your local repo, which seems terribly inefficient when all you need is a timestamp, and you're doing the same thing `find_remotes_cb` just did, again.

All you need *at the moment* is a timestamp. You (or other users of libostree) may need other stuff in future. The number of commits being pulled is likely to be low (in the worst case, it’s going to be the number of peers + the number of remotes + the number of USB keys, and that can only happen if they’re all on different commits).

mwleeds:
Oh I was thinking about looking at the metadata for each commit, but I think you're saying to pull and look at `ostree-metadata`. Are you suggesting that a side effect of calling `find_remotes_async` would be that the `ostree-metadata` for each remote gets pulled into your repo? I don't think that would work because there might be more than one remote providing the same collection ID.

Maybe it would help for you to look at the flatpak side of this? https://github.com/mwleeds/flatpak/commits/lib-list-p2p-updates

pwithnall:
Nope, I’m saying to pull and look at the metadata for each commit. Commits have an additional metadata `a{sv}`.

mwleeds:
Looking again it seems that `find_remotes_cb` doesn't pull the commit metadata from the remote. It just looks at the commit metadata in the summary file (which I'm surprised it's in there) and on disk (if any). Flatpak could do the same thing and look in the summary, but pulling the summary would mean adding the temporary remote to the repo config (pulling the commit metadata has the same problem). Hence why `OstreeRepoFinderResult` providing the right information seems like a good option.

If the use of the unsigned summary file is a concern, we could look at moving some of this information into `ostree-metadata`. I think if a malicious peer crafted a dishonest summary file, they could get you to only pull from them and fail (because `pull_from_remotes_async` doesn't pull refs mapped to NULL), thereby preventing you from updating.

pwithnall:
I’m away for the next week, and I can’t look into this fully now. I think my high-level points from before still stand: make this extensible, and try and reuse existing OSTree API where possible. What the code currently does, and how it currently gets hold of metadata is relevant, but shouldn’t be the main driving force behind the design of the branch.

I haven’t looked at the unsigned summary concern, but it should be investigated in a separate task. I suspect the solution would be to ensure that whatever we pull out of the `summary` file is verified from whatever we eventually get in the signed commit metadata.

src/libostree/ostree-repo-finder.h
```diff
@@ -130,6 +141,7 @@ typedef struct
   OstreeRepoFinder *finder;
   gint priority;
   GHashTable *ref_to_checksum;
+  GHashTable *ref_to_timestamp;
```
mwleeds: I wonder if I need to decrement the padding variable to avoid ABI break. I think so

mwleeds:
I hear what you're saying about extensibility, but I don't think it makes sense to add all the overhead of pulling commit metadata again in the find remotes phase (when it's already happening in the pull phase) just in case someone might have a use for it in the future. As for using existing ostree API, in order to fetch the summary or commit metadata you would have to use `ostree_repo_remote_add` which writes to the config file, potentially leaving it in a bad state of the operation is aborted (ostree can internally just add the remote in memory using `_ostree_repo_add_remote`). That also adds complexity on the client side that I think is unnecessary.

mwleeds:
Another way this could be more extensible is to add a hash table mapping refs to dictionaries, currently with checksums and timestamps but allowing expansion. But I'm not sure that's any more memory efficient than the way it works on this branch, and it means we'd probably want to deprecate `ref_to_checksum` before making the API non-experimental, which means there'd be breakage if old clients are linked against new versions of libostree, and I'm not sure it's worth the cost of changing how the libostree clients currently work.

So I still think the way this branch is currently written is good, but I'll finish writing the flatpak side of it and see if it works.

mwleeds:
(The flatpak side seems to be working).

The code freeze date for the release this feature is planned for is Monday, the same day Philip gets back. So things would be tight if we wait for him, and I'd rather not merge it downstream without merging it upstream. So hopefully we can work this out @cgwalters.

Philip said that having `ref_to_checksum` map refs to structs (with padding) would be another way to do this that might be more memory efficient and extensible, so if you think that's a good idea I can work on it. But it would have to be backwards incompatible, it would require changes to clients, and I'm hesitant to implement something for a hypothetical use case, when what we have now works for the known use cases.

cgwalters:
I had actually forgotten that summary files themselves have an `a{sv}` which maps refs to additional data, the only item of which is timestamps.  This is where I get the [varlink](https://github.com/varlink) people's ideas behind just using JSON.

I'm not too concerned about memory efficiency here yet myself.  

>  But it would have to be backwards incompatible, it would require changes to clients, 

Currently this is all experimental API so we *should* be able to change it, but that said:

It feels like there's a chain here; if we wanted additional data in the future, it's highly likely we'd only want commits that contain upgrades.  Clients could then later do commit-only pulls to get the commit objects.

So I guess I'd lean towards merging as is.

BTW:

> As for using existing ostree API, in order to fetch the summary or commit metadata you would have to use ostree_repo_remote_add which writes to the config file, potentially leaving it in a bad state of the operation is aborted

I'd have no problem adding API to improve that.

cgwalters:
@rh-atomic-bot r+  f61f436

PR #1522, opened 2018-03-29
lib/repo-pull: Improve error message when no summary is found

mwleeds:
In ostree_repo_remote_fetch_summary_with_options(), if no summary is
found on the server and summary verification is enabled, the error
message implies that it's the summary signature that's missing, which is
misleading. This commit adds a more specific error message for the case
of a missing summary, which has the side effect of explicitly checking
for the case that signatures != NULL && summary == NULL after
repo_remote_fetch_summary(), even though that should never happen.

One effect of this is that if you run "flatpak remote-add" with an
incorrect URL you get a more helpful error message, and similarly for
other flatpak operations and other users of ostree.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -5784,6 +5784,13 @@ ostree_repo_remote_fetch_summary_with_options (OstreeRepo    *self,
   if (!ostree_repo_remote_get_gpg_verify_summary (self, name, &gpg_verify_summary, error))
     goto out;
 
+  if (gpg_verify_summary && summary == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+                   "No remote summary found (check that the configured URL is correct)");
+      goto out;
+    }
+
   if (gpg_verify_summary && signatures == NULL)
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -5784,6 +5784,13 @@ ostree_repo_remote_fetch_summary_with_options (OstreeRepo    *self,
   if (!ostree_repo_remote_get_gpg_verify_summary (self, name, &gpg_verify_summary, error))
     goto out;
 
+  if (gpg_verify_summary && summary == NULL)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+                   "No remote summary found (check that the configured URL is correct)");
```
jlebon: How about to be more consistent: "GPG verification enabled, but no summary found (check configured URL in remote config is correct)" ?

I guess technically, both this error and the next one should say "GPG *summary* verification enabled, ..." to be more precise.

mwleeds: Sure

cgwalters:
@rh-atomic-bot r+ 14051ae

PR #1524, opened 2018-03-29
Don't write to parent repo

alexlarsson:
In _try_clone_from_payload_link, don't try to do the clone in the
parent repo, because we don't want to modify that. parent repos are
typically used when you want a shared, immutable base.

For example in flatpak, the parent repo is the system repo which you
don't have write access to, so any modification to it will fail with
EACCES, making it impossible to install via the system helper.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -777,8 +777,6 @@ _try_clone_from_payload_link (OstreeRepo   *self,
           return TRUE;
         }
     }
-  if (self->parent_repo)
-    return _try_clone_from_payload_link (self->parent_repo, payload_checksum, file_info, tmpf, cancellable, error);
 
   return TRUE;
 }


giuseppe:
do you have a log with the error?  If possible, I'd like to fix it in a way that the lookup is still done in the parent repository, but no write should happen there.

alexlarsson:
What fails is the glnx_open_tmpfile_linkable_at()->open(O_TMPFS) gets EACCES in _check_support_reflink() on the parent repo, because its is only writable by root.

But also, even if this would not break, the entire thing seems to depend on any reflinks to work if _check_support_reflink returns true. However if the parent and child repos are on different filesystem you might not be able to reflink between them, even if both are on reflink supporting filesystems.

cgwalters:
@rh-atomic-bot r+ ad11aa1

cgwalters:
Unfortunately while we do run the flatpak unit tests as part of our CI, that doesn't cover "real" usage like this.  And I personally tend to mostly only test ostree git with rpm-ostree.

We can clearly wire up more "real" usage testing in CI, it's one of the things I'm working on now.  Short term I'll add "flatpak sanity checks" to my "before ostree release" checklist.

alexlarsson:
We should probably backport this fix to F28...

cgwalters:
Sure, I can do that, or feel free to do so if you prefer.  I'm not immediately reproducing the problem though; does it only occur when the host fs supports reflinks (are you still on btrfs as `/`)?

cgwalters:
> does it only occur when the host fs supports reflinks (are you still on btrfs as /)?

Answer, yes.

cgwalters:
(Also just leaving this breadcrumb for future self and others: be sure to `pkill -f /usr/libexec/flatpak-system-helper` to be sure you're actually testing the new libostree version you installed)

PR #1525, opened 2018-03-29
commit, payload-reflink: do not write to the parent repo

giuseppe:
reintroduce the feature that was reverted with:

28c7bc6d0e153a0b07bdb82d25473a490765067f

Differently than the original implementation, now we don't attempt any
test for reflinks support on the parent repository, since the test
requires write access to the repository.

Additionally, also check that the two repositories are on the same
device before attempting any reflink.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -678,7 +678,7 @@ _import_payload_link (OstreeRepo   *dest_repo,
 
   /* The two repositories are on different devices */
   if (src_repo->device != dest_repo->device)
-    return FALSE;
+    return TRUE;
 
   if (!_check_support_reflink (dest_repo, &reflinks_supported, error))
     return FALSE;
@@ -729,7 +729,7 @@ _try_clone_from_payload_link (OstreeRepo   *self,
 
   /* The two repositories are on different devices */
   if (self->device != dest_repo->device)
-    return FALSE;
+    return TRUE;
 
   if (!_check_support_reflink (dest_repo, &reflinks_supported, error))
     return FALSE;
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -598,27 +598,30 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
 }
 
 static gboolean
-_check_support_reflink (OstreeRepo *self, gboolean *supported, GError **error)
+_check_support_reflink (OstreeRepo *src, OstreeRepo *dest, gboolean *supported, GError **error)
 {
-  /* We have not checked yet if the file system supports reflinks, do it here */
-  if (g_atomic_int_get (&self->fs_support_reflink) == 0)
+  /* The two repositories are on different devices */
+  if (src->device != dest->device)
+    return FALSE;
+
+  /* We have not checked yet if the destination file system supports reflinks, do it here */
+  if (g_atomic_int_get (&dest->fs_support_reflink) == 0)
     {
-      g_auto(GLnxTmpfile) src_tmpf = { 0, };
+      glnx_autofd int src_fd = -1;
       g_auto(GLnxTmpfile) dest_tmpf = { 0, };
 
-      if (!glnx_open_tmpfile_linkable_at (commit_tmp_dfd (self), ".", O_RDWR|O_CLOEXEC,
-                                          &src_tmpf, error))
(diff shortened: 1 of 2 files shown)


giuseppe:
@cgwalters @alexlarsson would something like this patch work?  I've done some quick tests and it does solve the problem you've seen, I'll keep testing it if the approach makes sense to you

cgwalters:
The code looks sane, I'm not opposed.  But in general today libostree suffers from having very little "parent repo" test suite coverage.  So if we were doing this I'd really like to see the test suite extended to cover it.  We should also only do this *after* spinning a new release to fix flatpak today.

giuseppe:
I've added a test that pulls to a `bare-user` repository from the `bare` repository

giuseppe:
rebased :arrow_up:

src/libostree/ostree-repo-commit.c
```diff
@@ -598,27 +598,30 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
 }
 
 static gboolean
-_check_support_reflink (OstreeRepo *self, gboolean *supported, GError **error)
+_check_support_reflink (OstreeRepo *src, OstreeRepo *dest, gboolean *supported, GError **error)
```
cgwalters: I find it confusing that here the order is `(src, dest)`...

src/libostree/ostree-repo-commit.c
```diff
@@ -664,8 +668,8 @@ _create_payload_link (OstreeRepo   *self,
 }
 
 static gboolean
-_import_payload_link (OstreeRepo   *self,
-                      OstreeRepo   *source,
+_import_payload_link (OstreeRepo   *dest_repo,
```
cgwalters: ...then here it's `(dest, src)`.

src/libostree/ostree-repo-commit.c
```diff
@@ -598,27 +598,30 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
 }
 
 static gboolean
-_check_support_reflink (OstreeRepo *self, gboolean *supported, GError **error)
+_check_support_reflink (OstreeRepo *src, OstreeRepo *dest, gboolean *supported, GError **error)
 {
-  /* We have not checked yet if the file system supports reflinks, do it here */
-  if (g_atomic_int_get (&self->fs_support_reflink) == 0)
+  /* The two repositories are on different devices */
+  if (src->device != dest->device)
```
cgwalters: Since this is the only place where we compare them, maybe it's simplest to move this condition down to the two places that call this with different src,dest?

giuseppe:
added a fixup commit to address the comments :arrow_up:

src/libostree/ostree-repo-commit.c
```diff
@@ -680,7 +676,11 @@ _import_payload_link (OstreeRepo   *dest_repo,
   glnx_unref_object OtChecksumInstream *checksum_payload = NULL;
   g_autoptr(GFileInfo) file_info = NULL;
 
-  if (!_check_support_reflink (src_repo, dest_repo, &reflinks_supported, error))
+  /* The two repositories are on different devices */
+  if (src_repo->device != dest_repo->device)
+    return FALSE;
```
cgwalters: This should be `return TRUE` right?

src/libostree/ostree-repo-commit.c
```diff
@@ -727,7 +727,11 @@ _try_clone_from_payload_link (OstreeRepo   *self,
   if (self->commit_stagedir.initialized)
     dfd_searches[0] = self->commit_stagedir.fd;
 
-  if (!_check_support_reflink (self, dest_repo, &reflinks_supported, error))
+  /* The two repositories are on different devices */
+  if (self->device != dest_repo->device)
+    return FALSE;
```
cgwalters: ditto

src/libostree/ostree-repo-commit.c
giuseppe: yes, correct.  Fixed in a new fixup

cgwalters: OK but that led me to consider "why weren't the tests failing", so I looked at that.  I ended up with the fixup :arrow_down:

cgwalters:
@rh-atomic-bot r+ 95d4b3f

cgwalters:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1526, opened 2018-03-29
lib: Fix a few comments

mwleeds:


cgwalters:
@rh-atomic-bot r+ a64992a

PR #1528, opened 2018-03-31
lib/repo-pull: Rename a variable for clarity

mwleeds:
In libostree, the phrase "commit metadata" has two meanings-- one is the
first dictionary in a commit GVariant that stores metadata such as ref
bindings, and the other is the commit metadata in the summary file,
which stores the commit size, checksum, and timestamp. In
find_remotes_process_refs(), the entire commit GVariant was being
referred to as commit metadata, so this commit changes the variable
name and a comment to make things more consistent.


cgwalters:
Thanks, commits like this are always appreciated!

@rh-atomic-bot r+ 2765719

cgwalters:
@rh-atomic-bot retry

Issue #1523, opened 2018-03-29
Race condition on summary cache


alexlarsson:
Here is the current logic for downloading the summary.

```
repo_remote_fetch_summary () 
{
  summary_sig = download("summary.sig");
  if (same_as_cached(summary_sig))
    return summary_sig, cached_summary;
  summary = download("summary.sig");
  // RACE HERE
  save_in_cache(summary_sig, summary);
  return summary_sig, summary;
}

ostree_repo_remote_fetch_summary_with_options()
{
  summary_sig, summary = repo_remote_fetch_summary ();
  if (!gpg_verify(summary_sig, summary))
    return FAIL("Not gpg verified)";
  return summary;
}
```

As might be obvious by the "RACE HERE" comment, this can race with a server side update to the summary, where you download and cache the new summary.sig but the old summary file. You then get forever into a state where the gpg is not verifying.

Now, this is not a trivial race, because the client side update does write the summary.sig *after* updating the summary. So, typically in a race you would get the old summary.sig and the new summary. Then you would get a gpg error, but next time you would get the new summary.sig and it will solve itself. However, i definately got into the situation where summary.sig was the latest, but summary did not match it. I'm not sure exactly how this happened, but with the multiple levels of caching we have on flathub, i can imagine the order on the download side being reversed.

I think we should either delay the writing of the cache until we have verified the signature, or delete any caches when a gpg verification failed.

ramcq:
> However, i definately got into the situation where summary.sig was the latest, but summary did not match it. I'm not sure exactly how this happened, but with the multiple levels of caching we have on flathub, i can imagine the order on the download side being reversed.

Caching or no, we cannot possibly update two files atomically on the server, so even if we try really hard we'll always see this occasionally/eventually, unless or until we fix #487 or just drop signing summaries entirely now that we have ostree-metadata with signed commits.

alexlarsson:
Yes, but without cacheing the race would be a temporary error, with the caching this is forever stuck until you manually clear the cache.

ramcq:
Sure, I agree with not caching the broken state on the client side. Just saying we can't single out flathub's caching for this - it might make it more common, but could always happen anyway. :)

cgwalters:
> I think we should either delay the writing of the cache until we have verified the signature, or delete any caches when a gpg verification failed.

Yep, the former seems like the first step but I'm fine with both.

PR #1529, opened 2018-04-03
pull: Don't save summary to cache before validating signatures

alexlarsson:
In case of some kind of race or other weirdness we might be getting
non-matching versions of summary.sig and summary, where summary.sig
is the latest version. Currently we're saving them to the cache
directly after downloading them successfully, but they will then fail
to gpg validate. Then on the next run we'll keep using the cached files
even if they are incorrect, until summary.sig changes upstream.

This changes the order so that we verify the signatures before saving
to the cache, thus ensuring that we don't end up in a stuck state.

Fixes https://github.com/ostreedev/ostree/issues/1523

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2913,6 +2913,7 @@ repo_remote_fetch_summary (OstreeRepo    *self,
                            GVariant      *options,
                            GBytes       **out_summary,
                            GBytes       **out_signatures,
+                           gboolean      *out_from_cache,
                            GCancellable  *cancellable,
                            GError       **error)
 {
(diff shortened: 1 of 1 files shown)


cgwalters:
@rh-atomic-bot r+ abcdeab6

cgwalters:
@rh-atomic-bot r+ abdeab6

alexlarsson:
I think CI is broken because 0001-Don-t-write-to-parent-repo.patch was backported to fedora.

cgwalters:
https://github.com/ostreedev/ostree/pull/1530

alexlarsson:
@rh-atomic-bot retest

alexlarsson:
@rh-atomic-bot retry

PR #1530, opened 2018-04-03
ci: Drop patches when building RPM

cgwalters:
Things fell over when downstream added a patch we had already merged
upstream.


jlebon:
@rh-atomic-bot r+ d16f204

PR #1531, opened 2018-04-03
switchroot: Don't log if running as pid1, minor code style cleanups

cgwalters:
If we're running as pid1, avoid printing anything in the normal
success paths as we don't want to affect the physical console by
default; the device may be using a splash screen, etc.

Also cleanup the code a bit to use a single variable
`running_as_pid1`, declare-and-initialize, use the
`bool` type, etc.

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -39,6 +39,7 @@
 #include <fcntl.h>
 #include <stdio.h>
 #include <stdarg.h>
+#include <stdbool.h>
 #include <stdlib.h>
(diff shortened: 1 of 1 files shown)


cgwalters:

jlebon:
This looks sane to me, though I'll give @akiernan a chance to look. CI failures look like Fedora mirror flakes.

cgwalters:
> I consider the information printed quite helpful for debugging.

We do print when things *fail* still.  Have you found the prints useful in the success case?

jlebon:
@rh-atomic-bot r+ c914b06

PR #1533, opened 2018-04-06
fsck: Mark commits with missing or deleted object partial

alexlarsson:
This means we can later use various operations to heal the repository because ostree does not assume all objects are there.
   
This the begining of a fix for https://github.com/ostreedev/ostree/pull/345


cgwalters:
For this I'd initially thought of moving the guts of fsck into the library, much like we did in 5c940987e768523ef1411b65bcaad09fba6befef

Adding new public API for commit parents feels somewhat special case to this...or, hm, maybe someone would want to know "what commits refer to this object" for space reduction purposes.

src/libostree/ostree-repo-traverse.c
```diff
@@ -294,18 +294,142 @@ ostree_repo_traverse_new_reachable (void)
...
+{
+  GVariant *old_parents;
+
+  if (inout_parents == NULL)
+    return;
+
+  old_parents = g_hash_table_lookup (inout_parents, key);
+  if (old_parents == NULL)
+    {
+      /* For the common case of a single pointer we skip using an array to save memory. */
```
cgwalters: One worry I have with this is that there's also going to be a class of objects which are basically in every single commit.  It feels like we may want to distinguish between client-side versus server-side repositories.

In the server side case, if I have say 100 commits of history, and I'm taking the same base component set across my builds (say `bash` binaries) or whatever, then every one of those objects is going to point to all of the parents.

Throwing together some numbers, currently:

```
# ostree ls -R fedora-27:fedora/27/x86_64/workstation | wc -l
129280
```

Say we have server side 100 commits (not unreasonable?) and between them about 75% of the files are in common...that's about:
```
gjs> (((100000*0.75)*100)*8)/(1024*1024)
57.220458984375
```

57 megs of pointer data.  Not a big deal I guess.  I just wanted to work through it.

src/libostree/ostree-repo-traverse.c
```diff
@@ -294,18 +294,142 @@ ostree_repo_traverse_new_reachable (void)
...
+      g_hash_table_insert (inout_parents, g_variant_ref (key), g_variant_ref (parent_key));
+    }
+  else
+    {
+      g_autofree GVariant **new_parents = NULL;
+      gsize len, i;
+
+      len = 0;
```
cgwalters: Very minor but would you mind using declare-and-initialize for new code?

alexlarsson:
Yeah, for historical reasons the API with a separate hashtable is a bit ugly, but it seems like the backtracking could be generally useful even outside fsck.

src/libostree/ostree-repo-traverse.c
alexlarsson: For the record, the flathub repo currently has 2822 refs, and 7530 commits. However, we don't share anywhere near 75% of files, because apps are different, and we don't keep long histories.

There are certain objects that are very common though, like the "default" dirmeta, which will be in essentially every commit... Not going to be a lot of these though.

cgwalters:
Rebased :surfing_man:

cgwalters:
I also added a test to `itest-pull.sh` which is a more "realistic" data set, and verified that re-doing a pull gets us the data back.

One thing I hit on is that this only works for `.file` content objects; deleting a `.dirtree/.dirmeta` will throw an error during traversal.  That's something to fix for later; since file objects are hardlinked (and a lot larger) they're the most likely ones to be corrupted.

@rh-atomic-bot r+ ac9a98f

tests/installed/nondestructive/itest-pull.sh
```diff
@@ -30,6 +30,20 @@ ostree --repo=bare-repo init --mode=bare-user
 ostree --repo=bare-repo remote add origin --set=gpg-verify=false $(cat ${test_tmpdir}/httpd-address)
 log_timestamps ostree --repo=bare-repo pull --disable-static-deltas origin ${host_nonremoteref}
 
+echo "ok pull"
```
alexlarsson: Don't you need to update some number-of-tests count when you add a test?

cgwalters: These tests aren't yet being parsed by something that understands TAP...someday later.

PR #1535, opened 2018-04-11
Staged deploy prep

cgwalters:


jlebon:
Nice! @rh-atomic-bot r+ caff5b9

PR #1513, opened 2018-03-23
tests: Merge installed/ and fedora-str/ directories 

cgwalters:
Let's be opinionated now, and our installed/ test story *is*
Ansible/STR.  Merge `tests/fedora-str` into `tests/installed/`.

Rework the nondestructive tests into a separate playbook run, and parallelize
them for more efficiency.

The destructive tests are also changed to use Ansible more.

Add a higher level `run.sh` entrypoint and update the `README.md`
with some useful tips.

PR #1536, opened 2018-04-11
ci: turn gating back on for most testsuites

jlebon:
Somehow, this slipped through in #1513. We weren't inheriting anymore,
so `branches` defaulted back to just `master`. This means we weren't
gating on most of the containerized builds anymore. Ouch!

It'd make sense to teach PAPR to allow some defaults across all
testsuites, even in the `inherit: false` case. Though it's tempting to
also just change the hardcoded PAPR default to those branches since our
use of Homu + PAPR at this point is pretty ubiquitous, and it doesn't
really hurt for the ones that don't use it.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -28,6 +28,10 @@ artifacts:
 
 # This suite skips the RPMs and does the build+unit tests in a container
 inherit: false
+branches:
+    - master
+    - auto
+    - try
 container:
   image: registry.fedoraproject.org/fedora:27
 context: f27-primary


jlebon:
bot, retest this please

cgwalters:
oops

@rh-atomic-bot r+ 2197de2

jlebon:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #1537, opened 2018-04-11
Don't scan uncompressed_objects_dir if it doesn't exist

owtaylor:
A newly created archive-mode repository won't have a uncompressed-objects-cache
directory, and uncompressed_objects_dir is -1 to flag that. The special meaning of
-1 meaning "cwd" for libglnx means that the current directory was scanned as
if it was an objects directory, producing unexpected results, especially if there
were any two-letter files/subdirs in the current directory.

This fixes a regression from https://github.com/ostreedev/ostree/commit/88d27fb3f1bb4102fddc74e77276011f6ee365ec

I didn't examine the source code to see if there were any other places where uncompressed_objects_dir is used blindly.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1438,7 +1438,8 @@ scan_loose_devino (OstreeRepo                     *self,
         return FALSE;
     }
 
-  if (self->mode == OSTREE_REPO_MODE_ARCHIVE)
+  if (self->mode == OSTREE_REPO_MODE_ARCHIVE &&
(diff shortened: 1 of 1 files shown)


jlebon:
Thanks! @rh-atomic-bot r+ 0afaf73

The only other place which gave me some hesitation was `ostree_repo_checkout_gc`, though it looks like the only way we have uncompressed dirs to clean up is if we initialized the fd to uncompress in the first place.

owtaylor:
Thanks for the review! I agree with your analysis of `ostree_repo_checkout_gc`.

PR #1539, opened 2018-04-13
lib/sysroot: Move staged into deployment list, rework handling

cgwalters:
Followup to: https://github.com/ostreedev/ostree/pull/1503
After starting some more work on on this in rpm-ostree, it is
actually simpler if the staged deployment just shows up in the list.

It's effectively opt-in today; down the line we may make it the default,
but I worry about breaking things that e.g. assume they can mutate
the deployment before rebooting and have `/etc` already merged.

There's not that many things in libostree that iterate over the deployment
list.  The biggest change here is around the
`ostree_sysroot_write_deployments_with_options` API.  I initially
tried hard to support a use case like "push a rollback" while retaining
the staged deployment, but everything gets very messy because that
function truly is operating on the bootloader list.

For now what I settled on is to just discard the staged deployment;
down the line we can enhance things.

Where we then have some new gymnastics is around implementing
the finalization; we need to go to some effort to pull the staged
deployment out of the list and mark it as unstaged, and then pass
it down to `write_deployments()`.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2144,6 +2144,38 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
 {
   g_assert (self->loaded);
 
+  /* It dramatically simplifies a lot of the logic below if we
+   * drop the staged deployment from both the source deployment list,
+   * as well as the target list.  We don't want to write it to the bootloader
+   * now, which is mostly what this function is concerned with.
+   * In the future we though should probably adapt things to keep it.
+   */
+  if (self->staged_deployment)
+    {
+      if (!glnx_unlinkat (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED, 0, error))
+        return FALSE;
+
+      if (!_ostree_sysroot_rmrf_deployment (self, self->staged_deployment, cancellable, error))
+        return FALSE;
+
+      g_assert (self->staged_deployment == self->deployments->pdata[0]);
+      g_ptr_array_remove_index (self->deployments, 0);
+    }
+  /* First new deployment; we'll see if it's staged */
+  OstreeDeployment *first_new =
+    (new_deployments->len > 0 ? new_deployments->pdata[0] : NULL);
+  g_autoptr(GPtrArray) new_deployments_copy = NULL;
+  if (ostree_deployment_is_staged (first_new))
+    {
+      g_assert_cmpint (new_deployments->len, >, 0);
+      new_deployments_copy = g_ptr_array_sized_new (new_deployments->len - 1);
+      for (guint i = 1; i < new_deployments->len; i++)
+        g_ptr_array_add (new_deployments_copy, new_deployments->pdata[i]);
+    }
+  else
+    new_deployments_copy = g_ptr_array_ref (new_deployments);
+  new_deployments = new_deployments_copy;
+
   /* Assign a bootserial to each new deployment.
    */
   assign_bootserials (new_deployments);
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-sysroot.c
```diff
@@ -824,11 +833,16 @@ _ostree_sysroot_reload_staged (OstreeSysroot *self,
...
-          self->staged_deployment =
+          g_autoptr(OstreeDeployment) staged =
             _ostree_sysroot_deserialize_deployment_from_variant (target, error);
-          if (!self->staged_deployment)
+          if (!staged)
             return FALSE;
-          _ostree_deployment_set_bootconfig_from_kargs (self->staged_deployment, kargs);
+
+          _ostree_deployment_set_bootconfig_from_kargs (staged, kargs);
+          if (!load_origin (self, staged, NULL, error))
```
cgwalters: Also this fixes the fact that we weren't loading the origin for staged deployments.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2144,6 +2144,38 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
...
+      if (!glnx_unlinkat (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED, 0, error))
+        return FALSE;
+
+      if (!_ostree_sysroot_rmrf_deployment (self, self->staged_deployment, cancellable, error))
+        return FALSE;
+
+      g_assert (self->staged_deployment == self->deployments->pdata[0]);
+      g_ptr_array_remove_index (self->deployments, 0);
+    }
+  /* First new deployment; we'll see if it's staged */
```
jlebon: Can you expand under what situations this happens? In the finalize path, we've already transformed it into a regular deployment by the time we come here, right?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2144,6 +2144,38 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
...
+   * as well as the target list.  We don't want to write it to the bootloader
+   * now, which is mostly what this function is concerned with.
+   * In the future we though should probably adapt things to keep it.
+   */
+  if (self->staged_deployment)
+    {
+      if (!glnx_unlinkat (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED, 0, error))
+        return FALSE;
+
+      if (!_ostree_sysroot_rmrf_deployment (self, self->staged_deployment, cancellable, error))
```
jlebon: So then here, in the non-finalize path, let's say... `ostree admin undeploy 2`, don't we need to keep the staged deployment? I.e. only remove it from the current deployments, but keep it on disk so it gets added back at the final reload?

cgwalters: The problem with keeping it is we need to do juggling of the deployment arrays; that's what the comment above is about:

```
   /* It dramatically simplifies a lot of the logic below if we
   * drop the staged deployment from both the source deployment list,
   * as well as the target list.  We don't want to write it to the bootloader
   * now, which is mostly what this function is concerned with.
   * In the future we though should probably adapt things to keep it.
   */
```

jlebon: Right I mean, still remove it from the `self->deployments` array for that purpose, but don't actually remove it on disk. Otherwise, IIUC, even something like `undeploy 2` (let's say we're booted into `1`) will delete the staged deployment, right? That seems a bit surprising from a UX standpoint. Anyway, I'm cool with tweaking this in follow-ups!

jlebon:
@rh-atomic-bot r+ 0c8b57c

jlebon:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

jlebon:
OK, I can reproduce this if I upgrade to `gnupg2-2.2.6-1`, which recently went into stable:
https://bodhi.fedoraproject.org/updates/FEDORA-2018-3fc05e009d

jlebon:
https://src.fedoraproject.org/rpms/gpgme/pull-request/3

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

PR #1540, opened 2018-04-17
ci: Pull in patched gpgme packages

cgwalters:
See https://src.fedoraproject.org/rpms/gpgme/pull-request/3


jlebon:
Shoot, you beat me to it. :) This won't work for flatpak though I think, since it uses the `build` function directly, right? I had: https://github.com/jlebon/ostree/commit/5b78d906b8c72bebef3a72c2da7387ca9e42f1e5

cgwalters:
> This won't work for flatpak though I think, since it uses the build function directly, right? I had: jlebon/ostree@5b78d90

Looks like you're right, I force pushed your commit here.

@rh-atomic-bot r+ 5b78d90

jlebon:
@rh-atomic-bot r=cgwalters 7dc759f

PR #1542, opened 2018-04-17
avahi: Don't complain with g_warning if the daemon wasn't running

pwithnall:
This is a normal case when running unit tests in client code
on continuous integration infrastructure. When those tests are
running they will set G_DEBUG=fatal-warnings which will cause
the program to abort if a warning is emitted. Instead, emit
a debug message if the problem was that we couldn't connect to
the daemon.

---

Upstreaming a patch from @smspillaz which we’ve been carrying in Endless OS for a little while.


jlebon:
@rh-atomic-bot r+ e19a35a

PR #1544, opened 2018-04-18
fsck: Only print "marking commit partial" once

cgwalters:
Let's only print if the commit isn't already partial; this
addresses a spam of "marking commit partial" from fsck.


jlebon:
@rh-atomic-bot r+ 391f3ba

jlebon:
@rh-atomic-bot r+ 8dd04a1

jlebon:
@rh-atomic-bot retry

cgwalters:
Moved into https://github.com/ostreedev/ostree/pull/1548

PR #1547, opened 2018-04-19
ci: Temporary libgcrypt workaround for older kernels

jlebon:
I'm trying to get ostree tests to pass in OpenShift as part of our CI
move but I've been seeing lots of failures related to GPG tests. It
finally turned out to be because libgcrypt doesn't behave well on older
kernels that don't have `getrandom()` (the cluster is running on RHEL7).
Thankfully, there's a new build with a fix for this. Pull that in
manually until it gets into stable.

For more information, see:
https://bugzilla.redhat.com/show_bug.cgi?id=1542453


cgwalters:
@rh-atomic-bot r+ f1c216f

PR #1548, opened 2018-04-19
tests/installed: Make reboot task less racy

cgwalters:
This took a whole lot of experimentation.  I hit upon the idea
of doing a `systemctl stop sshd` to avoid the situation where we
might ssh back into the system while it's in the process of shutting
down.

Ultimately the other fix is disabling `ControlMaster`; see
for example: https://github.com/ansible/ansible/issues/17935


jlebon:
@rh-atomic-bot r+ 4cd789f

jlebon:
Ahh OK, that's an instance of https://github.com/ostreedev/ostree/issues/1549.
@rh-atomic-bot r+ bafa31c

cgwalters:
@rh-atomic-bot r+ 9414bea

jlebon:
Whoops! I have this filter that marks all @rh-atomic-bot notifications as read unless it's "Test timed out" or "Test failed". I tweaked it to also leave the "try again" emails unread.

cgwalters:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

jlebon:
I rolled up #1544 into here as well to see if that helps. If not, then I'd suggest we disable `itest-pull.sh` for now and work on it on the side to let pending PRs in.

jlebon:
@rh-atomic-bot r+ 1ba1c38

cgwalters:
>  If not, then I'd suggest we disable itest-pull.sh for now

Yeah I'm on board with that plan if this one bounces...the pain here is rather far exceeding the gain currently.

cgwalters:
Here's what I'm not understanding currently.  Let's take one [that passed](https://github.com/ostreedev/ostree/pull/15430).

In [the pull log](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/972a1921f517f5b2ae0aafed42224161fb4bb6d8.0.1524145048931166589/artifacts/artifacts/installed-nondestructive/logs/itest-pull.sh.log) the test took 9.5 minutes.  Which *should* have exceeded the `240` retries limit that was in place at that time, but actually we got through with `240-183 = 57` retries.  From the Ansible source it looks like `retry_pause = 1`... yet it doesn't seem to be that way locally.  Investigating.

cgwalters:
@rh-atomic-bot r+ 1463077

cgwalters:
@rh-atomic-bot retry

:cry:

jlebon:
@rh-atomic-bot, stop being such a jerk!

cgwalters:
OK, I thought I had it, it seemed fairly robust locally but apparently no dice.  I briefly looked at doing a custom Ansible module, but (again just from initial investigation) it seems we'd really need the engine to understand this; we basically want an atomic combination of`- shell: reboot` + `- wait_for_connection`.

cgwalters:
@rh-atomic-bot r+ 951052c

cgwalters:
Going back to longer timeouts seems to be OK in some manual testing in the CI env.  We'll just live with races temporarily...I'll investigate other approaches.

PR #1551, opened 2018-04-23
Use streaming reads for delta superblock for better memory use with inline parts

alexlarsson:
Instead of maping the entire superblock and referencing any inline parts from that we read the superblock in a streaming fashion from the file descriptor, and skipping any inline delta parts. Instead we record the offsets for said delta parts and stream them as needed from the superblock file.

To test this i generated a flatpak bundle for the Gnome SDK, ending up in a delta file with inlined parts at 288MB. Then i applied this to a bare-user repo with `ostree static-delta apply-offline` and monitored the memory size with top.

In the old code, the maximum size recorded was 556MB vss and 310MB rss.
With the streaming reads this was instead 269MB vss and 44MB rss.

This is mostly interesting for flatpak, as it uses static deltas with inline parts for flatpak bundles.


src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -398,13 +537,33 @@ _ostree_static_delta_part_open (GInputStream   *part_in,
...
+    {
+      if (part_size < 0)
+        {
+          struct stat part_stbuf;
+          if (!glnx_fstat (part_fd, &part_stbuf, error))
+            return FALSE;
+          part_size = part_stbuf.st_size - part_offset;
+        }
+
+      lseek (part_fd, part_offset, SEEK_SET);
```
giuseppe: it should not really happen but do we need to check for errors here?

alexlarsson:
@giuseppe Fixed up your comments

cgwalters:
Hmm.  But we shouldn't be touching those pages, right?    Actually now that I look at this...we should likely be using `MAP_SHARED` to guarantee that.  The unit tests pass at least for me with:
```
$ git diff
diff --git a/src/libotutil/ot-fs-utils.c b/src/libotutil/ot-fs-utils.c
index c4fcd56f..0975cfb6 100644
--- a/src/libotutil/ot-fs-utils.c
+++ b/src/libotutil/ot-fs-utils.c
@@ -177,8 +177,11 @@ ot_fd_readall_or_mmap (int           fd,
       /* The reason we don't use g_mapped_file_new_from_fd() here
        * is it doesn't support passing an offset, which is actually
        * used by the static delta code.
+       *
+       * We also explicitly use MAP_SHARED since the use cases are all
+       * truly read-only.
        */
-      gpointer map = mmap (NULL, len, PROT_READ, MAP_PRIVATE, fd, start);
+      gpointer map = mmap (NULL, len, PROT_READ, MAP_SHARED, fd, start);
       if (map == (void*)-1)
         return glnx_null_throw_errno_prefix (error, "mmap");
 
```

Anyways so in your numbers the (peak?) RSS is way down which is generally good, but on the other hand, if mmap is working as it should then since the pages aren't dirty they should be evicted - we aren't actually "owning" that memory right?

Can you walk this back for me to the real-world problem where it arose?  Was it something like being hit by the OOM killer during a flatpak operation?

cgwalters:
> on low-memory machines (as in < 4 gig). 

Is this low memory or actually 32 bit?  I strongly suspect the latter.   I ran a container locally with `podman run --memory 2G ...` locally, then used this against a 4GB ISO file I had lying around:

```
#!/usr/bin/python3
import os,mmap,sys
f = os.open(sys.argv[1], os.O_RDONLY)
m = mmap.mmap(f, 0, flags=mmap.MAP_SHARED, prot=mmap.PROT_READ)
for x in m:
    pass
```

Watching in `top`, `RES` and `SHR` slowly grew up to about `1.95G`, then stopped.  If other processes were contending for memory, AIUI the kernel is going to be fine discarding clean pages there.

However obviously on 32 bit, we can't map large files, so indeed we need a patch like this.  That said, there is an alternative, which is doing a ["sliding" mmap](https://github.com/gitpython-developers/smmap).

alexlarsson:
I don't see how you could plug in a sliding mmap implementation below GVariant? It expects to have full access to the entire blob you pass to it using regular pointer ops.

cgwalters:
Yeah, a sliding mmap GVariant would be at best a lot of painful work; more painful than this anyways.  Will take a look at this but it's going to be a bit.

src/libotutil/ot-variant-builder.c
```diff
@@ -711,6 +737,300 @@ gvs_calculate_total_size (gsize body_size,
...
+          if (!gvs_fd_read_unaligned_le (fd,
+                                         offset + size - offset_size * (member_info->i + 1),
+                                         offset_size,
+                                         &start, error))
+          return FALSE;
+        }
+      else
+        start = 0;
+
+      start += member_info->a;
```
cgwalters: I think what would help me a lot is if you could annotate some of these functions with the glib functions they came from.  For example, these bits clearly come from `gvs_tuple_get_child()`.  So maybe a comment at the top like:

```
This code derives from:
  - https://gitlab.gnome.org/GNOME/glib/blob/1c177ce0ab36991d33f4fbe0cfe6e003ca121b7e/glib/gvariant-serialiser.c#L851
```

PR #1552, opened 2018-04-23
ci: mark some ci testsuites as required again

jlebon:
Follow up to #1536; we are now running all the testsuites on merges, but
we weren't actually blocking on their success!


cgwalters:
@rh-atomic-bot r+ ec672ad

PR #1553, opened 2018-04-23
 ci: drop BDB1539 workaround 

jlebon:
For reason I don't quite understand, running the PAPR testsuites in OCP
sometimes gets hit with:

```
Running transaction
Failed to obtain the transaction lock (logged in as: root).
```

This is reminiscent of the overlay2 librpm compat issues[1], except that
AFAICT, this OCP instance is running on devicemapper. Yet, this patch
does fix the issue for me. (I initially had it specifically touch
`/var/lib/rpm/Packages`, but nope, that doesn't work).

https://bugzilla.redhat.com/show_bug.cgi?id=1213602

### DIFF
--- a/ci/libpaprci/libbuild.sh
+++ b/ci/libpaprci/libbuild.sh
@@ -6,18 +6,7 @@ OS_ID=$(. /etc/os-release; echo $ID)
 OS_VERSION_ID=$(. /etc/os-release; echo $VERSION_ID)
 
 pkg_upgrade() {
-    # https://bugzilla.redhat.com/show_bug.cgi?id=1483553
-    local ecode=0
-    yum -y distro-sync 2>err.txt || ecode=$?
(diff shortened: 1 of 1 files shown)


jlebon:
Marking as WIP, still investigating a bit more.

cgwalters:
Wouldn't we need `touch /var/lib/rpm/*` ?

jlebon:
> Wouldn't we need `touch /var/lib/rpm/*`

I did initially do that and it didn't seem to make a difference. See e.g. the various test results in: https://github.com/jlebon/ostree/commits/master. Trying to repeat the experiment, though it looks like the CentOS CI OCP instance is having some difficulties right now:

https://bugs.centos.org/view.php?id=14711

cgwalters:
@rh-atomic-bot r+ 4a4e21d

jlebon:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

Issue #1534, opened 2018-04-10
File COPYING is missing but referenced in ReadTheDocs


btreut:
On https://ostree.readthedocs.io/en/latest/#licensing there is a link to COPYING pointing to https://ostree.readthedocs.io/en/latest/COPYING which is missing.

Maybe it should link to https://github.com/ostreedev/ostree/blob/master/COPYING

PR #1554, opened 2018-04-23
README.md: Fix docs link to COPYING

cgwalters:
Because `README.md` also gets rendered into the docs, we need
to link directly to github.

Closes: https://github.com/ostreedev/ostree/issues/1534


jlebon:
@rh-atomic-bot r+ db7150a

PR #1556, opened 2018-04-24
lib/deploy: Also compare deployment csum versions

jlebon:
When comparing deployments to determine whether we need a new
bootversion, we should also check whether the commit "version" metadata
is the same. Otherwise, we may end up with the a bootconfig whose
`title` includes a version that doesn't match the one from the
deployment checksum.

Closes: projectatomic/rpm-ostree#1343

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1896,6 +1896,17 @@ assign_bootserials (GPtrArray   *deployments)
     }
 }
 
+static char*
+get_deployment_nonostree_kargs (OstreeDeployment *deployment)
+{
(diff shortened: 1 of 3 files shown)


cgwalters:
OK, and this is happening for Fedora because of https://pagure.io/pungi/issue/811 or so presumably.

cgwalters:
@rh-atomic-bot r+ 8041347

jlebon:
> OK, and this is happening for Fedora because of https://pagure.io/pungi/issue/811 or so presumably.

Not just for no-op updates, but any update that doesn't have a new kernel or initramfs. E.g. I was able to reproduce this on FAH by upgrading `27.121` to `27.122`, which just has `glusterfs` pkgs as changed.

PR #1558, opened 2018-04-27
Fix staged deployments with no kargs

cgwalters:


jlebon:
LGTM! @rh-atomic-bot r+ 3358698

PR #1559, opened 2018-04-27
deploy: Return staged deployment

cgwalters:
Today rpm-ostree has some code to run a "sanitycheck" on a deployment.
I had initially deleted that when adapting it to use the staging code,
but I realized it should work fine; we just won't see the merged
config, but that's OK.

When I readded that code it started crashing because we didn't
actually return the new deployment object.  We'll gain some coverage
here as I'll land the code to have rpm-ostree use staging, then bump
the rpm-ostree tests here.


jlebon:
@rh-atomic-bot r+ 88fda2d

PR #1560, opened 2018-04-30
man/ostree: Document --repo option a bit more

jlebon:
This new information is already mostly part of `ostree.repo(5)`, though
let's put it in `ostree(1)` as well since that's where the switch is
officially documented.

### DIFF
--- a/man/ostree-init.xml
+++ b/man/ostree-init.xml
@@ -69,13 +69,24 @@ Boston, MA 02111-1307, USA.
         <variablelist>
             <varlistentry>
                 <term><option>--mode</option>="MODE"</term>
-                <listitem><para> Initialize repository in given mode
-                (<literal>bare</literal>, <literal>bare-user</literal>,
-                <literal>archive</literal>). The default is
-                <literal>bare</literal>. Note that for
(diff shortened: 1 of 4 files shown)


dustymabe:
will this hit the man page and the `--help` ?

puiterwijk:
That seems like a lot of magic to me, but thanks for documenting.
Note that `ostree pull-local --help` (for example) also doesn't list any of this. Might be useful to add there too: `--repo=PATH                 Path to OSTree repository (defaults to /sysroot/ostree/repo)`

jlebon:
The `ot-main.c` part of the patch should be inherited into the `--help` output of all the commands that have a `--repo` switch.

The XML part is the man page for `ostree`. It's not in the per-command man pages, in the same way e.g. `git -C` is documented in `git(1)`, but not `git-pull(1)`.

cgwalters:
FWIW the rationale for the current behavior is just that we match what git does with bare repositories.

cgwalters:
@rh-atomic-bot r+ 2b0a947

jlebon:
@rh-atomic-bot retry

PR #1563, opened 2018-04-30
build: Don't distribute Bison-generated parser in dist tarballs

smcv:
We already build it from source, but the generated code is in the dist tarball anyway.


cgwalters:
Free the bison!

@rh-atomic-bot r+ eb5dbf2

cgwalters:
@rh-atomic-bot retry

PR #1566, opened 2018-04-30
Staging prune patches

cgwalters:
OK so it turns out the pruning issues I was having with the staging bits + rpm-ostree really boils down to missing regenerating the baselayer refs in rpm-ostree when doing staging.   (Though the fact that we were pruning the repo at finalization time threw me for a loop a bit unnecessarily)

But I think all of these patches make sense too.


cgwalters:
Split out the simple, more necessary bits into https://github.com/ostreedev/ostree/pull/1567

We can revisit this one after the next release.

cgwalters:
OK, after some more testing, we need to either ship these patches, or write an `ostree/staged` ref.  Although probably we always need the latter.

jlebon:
So now with https://github.com/ostreedev/ostree/pull/1570, we don't need these patches yet for the experimental `StageDeployments=` that just landed in rpm-ostree, right?

cgwalters:
Rebased :surfing_man:

src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -420,25 +420,76 @@ generate_deployment_refs (OstreeSysroot       *self,
...
+   */
+  if (!ostree_repo_add_reachable_refs_for_prune (repo, depth, options->reachable, cancellable, error))
+    return FALSE;
+
+  /* Since ostree was created we've been generating "deployment refs" in
+   * generate_deployment_refs() that look like ostree/0/1 etc. to ensure that
+   * anything doing a direct prune won't delete commits backing deployments.
+   * This bit might allow us to eventually drop that behavior, although we'd
+   * have to be very careful to ensure that all software is updated to use
+   * ostree_sysroot_rune_repo().
```
jlebon: `s/rune/prune`

Actually, do you mean `ostree_sysroot_cleanup_prune_repo`?

src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -501,8 +552,22 @@ _ostree_sysroot_cleanup_internal (OstreeSysroot              *self,
...
+      if (!ostree_sysroot_cleanup_prune_repo (self, &opts, &n_objects_total,
+                                              &n_objects_pruned, &freed_space,
+                                              cancellable, error))
+        return FALSE;
+
+      /* TODO remove printf in library */
+      if (freed_space > 0)
+        {
+          g_autofree char *freed_space_str = g_format_size_full (freed_space, 0);
+          g_print ("Freed objects: %s\n", freed_space_str);
```
jlebon: Wait, but the commit message is implying we're no longer doing prints, no? Or does it actually mean we have the API in place now so that clients can workaround this?

src/libostree/ostree-repo-prune.c
```diff
@@ -301,6 +301,64 @@ repo_prune_internal (OstreeRepo        *self,
...
+ * @cancellable: Cancellable
+ * @error: Error
+ *
+ * Add all commit objects directly reachable via a ref to @reachable.
+ *
+ * Locking: shared
+ * Since: 2018.6
+ */
+gboolean
+ostree_repo_add_reachable_refs_for_prune (OstreeRepo *self,
```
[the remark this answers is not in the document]
cgwalters: I renamed it to `ostree_repo_traverse_reachable_refs` which better fits in with the other traverse bits.

src/libostree/ostree-sysroot-cleanup.c
cgwalters: This is the old method; the new API indeed does not print.

src/libostree/ostree-sysroot-cleanup.c
cgwalters: Yep, rebased :arrow_down:

jlebon:
@rh-atomic-bot r+ 47655d7

PR #1567, opened 2018-05-01
 deploy: Don't prune repo at finalization time by default 

cgwalters:


jlebon:
@rh-atomic-bot r+ be15274

PR #1568, opened 2018-05-01
 lib/deploy: Throw an error if trying to stage when not ostree-booted

cgwalters:


cgwalters:
(I ended up not adding an API from the first commit, `ostree_sysroot_get_booted_deployment()` basically is that API)

jlebon:
@rh-atomic-bot r+ b8c3112

PR #1571, opened 2018-05-02
deploy: Use fdatasync() for new kernel/initramfs by default

cgwalters:
While we do a `syncfs()` plus `FIFREEZE/THAW` for `/boot`, that
only comes during deployment finalization.

The code here today generally assumes that if the file exists
it's been fully written.  So let's do a `fdatasync()` before
we do the `rename()`.

This just came out of looking through the code while working
on deployment staging.  In that scenario there's a much larger
window between when we copy the kernel/initramfs and when we
sync `/boot`.


jlebon:
@rh-atomic-bot retry

PR #1572, opened 2018-05-02
lib: Use `Locking:` term in docs

cgwalters:
This is easier to `git grep` etc. versus ad-hoc English.  Although
we still have some English for the prepare_transaction/commit which
acquire/release in separate phases.


jlebon:
Yes, much better!
@rh-atomic-bot r+ 1b7b2a3

Issue #1569, opened 2018-05-02
commit --tree=ref=foo is slow again


cgwalters:
I noticed that `ostree commit -b testref --tree=ref=fedora-atomic:fedora/27/x86_64/atomic-host` was slow unexpectedly, and looking at strace we're doing full checksums.

The use case here is just "generate a new commit with existing content" which I wanted for testing, but it we also talk about this for [commit promotion](https://ostree.readthedocs.io/en/latest/manual/repository-management/#promoting-content-along-ostree-branches-buildmaster-smoketested).

I'm not sure when this regressed.  Looking at this again, it's a bit hard because we'd have to work out which commit modifier flags actually affect the content versus not.

The more I think about this...we probably want something like:

`ostree commit --from-ref-contents=fedora-atomic:fedora/27/x86_64/atomic-host` and that should explicitly error out if say `--no-xattrs` or `--owner-uid` or whatever are also passed.  Then we can shortcut everything at the toplevel and just generate a new commit object with the same root tree.

cgwalters:
A workaround here today of course is just to use the checkout + commit path which is obviously well tested/optimized since it's a lot more common.

```
# cd /ostree/repo/tmp
# ostree checkout -H fedora-atomic:fedora/27/x86_64/atomic-host t
# ostree commit -b testref -I --consume t
# for ref in fedora-atomic:fedora/27/x86_64/atomic-host testref; do ostree show $ref |grep ContentChecksum; done
ContentChecksum:  3beb31b7ee24ae1a30a5bb409610785ebe687979dbe72b64edef843b30cb196f
ContentChecksum:  3beb31b7ee24ae1a30a5bb409610785ebe687979dbe72b64edef843b30cb196f
#
```

cgwalters:
Offhand I bet this was https://github.com/ostreedev/ostree/commit/0ee9e221beecd2261f75da46afa02d54b1230886

cgwalters:
My recollection was that this worked as of fb1e140dbadb8edae4b869602034c91269fef1df - which was a long time ago now obviously.  I think broadly speaking, we had a lot of trouble juggling the separate cases of "commit from disk, with modifiers" and "commit from tree" (maybe with modifiers).

If we had an API like:

```
_OSTREE_PUBLIC
gboolean      ostree_repo_splice_commit (OstreeRepo                 *self,
       const char                    *checksum,
       const char                    *subpath,
       OstreeMutableTree          *mtree,
[...]

PR #1573, opened 2018-05-04
tests: Tweak staged-deploy test to be faster

cgwalters:
Work around https://github.com/ostreedev/ostree/issues/1569


jlebon:
@rh-atomic-bot r+ 609c15c

PR #1577, opened 2018-05-04
tests: Fix installed tests more 

cgwalters:
OK so I noticed that something was failing and we were missing
`set -xeuo pipefail` in our shells.  That of course revealed
the ansible tests didn't actually work - my only defense
here is spending so much time fighting to get it through CI
and trying something new.

Anyways, to make the staged-deploy tests work we need a task
that actually uses `rpm-ostree override` rather than `usroverlay`.

Let's make this a bit saner and have a clean split between
tests that are "shell-script+usroverlay" and "ansible+override".

### DIFF
--- a/tests/installed/destructive/staged-deploy.yml
+++ b/tests/installed/destructive/staged-deploy.yml
@@ -3,6 +3,7 @@
 
 - name: Write staged-deploy commit
   shell: |
+    set -xeuo pipefail
     cd /ostree/repo/tmp
     # https://github.com/ostreedev/ostree/issues/1569
     ostree checkout -H ${commit} t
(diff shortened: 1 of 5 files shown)


tests/installed/destructive/staged-deploy.yml
```diff
@@ -11,7 +12,7 @@
     orig_mtime=$(stat -c '%.Y' /sysroot/ostree/deploy)
     ostree admin deploy --stage staged-deploy
     new_mtime=$(stat -c '%.Y' /sysroot/ostree/deploy)
-    assert_not_streq "${orig_mtime}" "${new_mtime}"
```
jlebon: It'd be interesting to have a custom ansible module that wraps shell scripts in `set -xeuo pipefail` and imports the `libtest-core.sh` stuff.

PR #1579, opened 2018-05-05
repo: handle GPG_ERR_AMBIGUOUS_NAME in sign_data

muelli:
This should give a more insightful error message if the user provides
a UID which is present on multiple keys.
This happens if you have an old key in your keyring which you are not
actively using any more, e.g. because it is too old.  You still have
your old keys in your keyring, because you want to read old email
encrypted for that key, though.

The gpgme function used by ostree right now complains if a UID is found
on multiple keys:
https://www.gnupg.org/documentation/manuals/gpgme/Listing-Keys.html#index-gpgme_005fget_005fkey
The used API is too simple for that use case.

Note that it would be nicer if ostree picked the only valid signing key out
of the available keys rather than using the simplistic gpgme_get_key
function.  It be nicer, of course, if there was such a gpgme function.


cgwalters:
Thanks for the patch!

@rh-atomic-bot r+ d424bda

PR #1582, opened 2018-05-08
ci: Drop str hotpatch

cgwalters:
The change we want is in the current Fedora repos, and git master
is broken:

> qemu-system-x86_64: -vnc :1: Failed to start VNC server: Failed to bind socket: Address already in use

 https://pagure.io/standard-test-roles/pull-request/186#comment-52440


jlebon:
Thanks for looking into this!
@rh-atomic-bot r+ 5abfaab

PR #1583, opened 2018-05-09
lib/sysroot: Add OSTREE_EX_STAGE_DEPLOYMENTS environment variable 

cgwalters:
I feel like I'm drowning in a pile of experimental-but-almost-stable
features...

Anyways, since we made the feature opt-in in rpm-ostree in
https://github.com/projectatomic/rpm-ostree/pull/1352
let's mirror that a bit here with an environment variable so people
can play with it more easily.

The tests needed some tweaks; specifically we need to reload the
status fact after making changes.  I'm still a bit uncertain
about the Ansible-as-tests.

But we add an upgrade test that uses the new environment variable.

### DIFF
--- a/tests/installed/destructive/staged-deploy.yml
+++ b/tests/installed/destructive/staged-deploy.yml
@@ -27,16 +27,31 @@
 - name: Check that deploy-staged service worked
   shell: |
     set -xeuo pipefail
+    rpm-ostree status
     # Assert that the previous boot had a journal entry for it
     journalctl -b "-1" -u ostree-finalize-staged.service | grep -q -e 'Transaction complete'
     # And there should not be a staged deployment
     test '!' -f /run/ostree/staged-deployment
-- name: Rollback
-  shell: rpm-ostree rollback
-- include_tasks: ../tasks/reboot.yml
-- shell: |
+
+- name: Upgrade with staging
+  shell: |
     set -xeuo pipefail
+    test '!' -f /run/ostree/staged-deployment
+    ostree admin deploy --stage staged-deploy
+    test -f /run/ostree/staged-deployment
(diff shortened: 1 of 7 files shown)


cgwalters:
bot, retest this please

cgwalters:
@rh-atomic-bot try

cgwalters:
OK, this one should be ready for review.  My thought is to land this and then do a release.  Then the next one can be stabilizing the current experimental APIs.

tests/installed/destructive/staged-deploy.yml
```diff
@@ -27,16 +27,31 @@
...
+    env OSTREE_EX_STAGE_DEPLOYMENTS=1 ostree admin upgrade >out.txt
+    test -f /run/ostree/staged-deployment
+    # Debating bouncing back out to Ansible for this
+    firstdeploycommit=$(rpm-ostree status |grep 'Commit:' |head -1|sed -e 's,^ *Commit: *,,')
+    test "${firstdeploycommit}" = "${newcommit}"
+    # Cleanup
     rpm-ostree cleanup -rp
+- import_tasks: ../tasks/query-host.yml
 
 # Ensure we can unstage
```
jlebon: In the following test, we can just skip the `--stage` one before trying to unstage now, right?

cgwalters: We do a `rpm-ostree cleanup -rp` above, so there won't be a staged deployment, right?  The idea is to keep things relatively clean between tests.  Am I understanding you correctly?

jlebon:
Ahh yep, I misread the diff.
@rh-atomic-bot r+ 75cea3e

PR #1584, opened 2018-05-09
ci: Switch libgcrypt URL

cgwalters:
Since the previous one wasn't made into an update, it got GC'd.


jlebon:
@rh-atomic-bot r+ 2197f77

jlebon:
Thanks, missed that easy one!

PR #1587, opened 2018-05-11
ci: Add .papr-ex.yaml

jlebon:
I'd like to start working towards migrating the OSTree CI towards the
new PAPR model running in PACI. In order to take it for a test drive
(pun intended) without yet replacing our current CI, the new PAPR now
knows to look for a `.papr-ex.yaml` file. So as to not conflict with the
current PAPR, no GitHub statuses will be output other than a
`required-ex` label, which plays the same role as `required`. For more
information, see:

https://github.com/projectatomic/papr/commit/85b1d0a91eba6597d01b52b674de38a876f030fe

The `.papr-ex.yaml` version is essentially the same except for:

1. the `FAH27-insttests` runs directly in the scheduled container (notice
   the new `kvm: true` key)
2. the `host:` based contexts (rpm-ostree and flatpak) are missing since
   we don't support provisioning a host just yet

Keeping them in sync could get annoying, though building confidence in
the new approach will allow us to migrate faster.

### DIFF
--- a/.papr-ex.yaml
+++ b/.papr-ex.yaml
@@ -1,0 +1,131 @@
+# https://fedoraproject.org/wiki/CI/Tests
+branches:
+    - master
+    - auto
+    - try
+
+context: FAH27-insttests
+required: true
+
+container:
+  image: registry.fedoraproject.org/fedora:27
+  kvm: true
+
+tests:
+  - ci/fah27-insttests.sh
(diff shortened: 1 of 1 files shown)


cgwalters:
Sounds good to me!

@rh-atomic-bot r+ 703ba1f

PR #1588, opened 2018-05-14
docs: var/local -> var/usrlocal

elboulangero:
It seems to be a typo, please confirm, thanks!

### DIFF
--- a/docs/manual/adapting-existing.md
+++ b/docs/manual/adapting-existing.md
@@ -32,7 +32,7 @@ recommended set:
  - `/opt` → `/var/opt`
  - `/srv` → `/var/srv`
  - `/root` → `/var/roothome`
- - `/usr/local` → `/var/local`
+ - `/usr/local` → `/var/usrlocal`
  - `/mnt` → `/var/mnt`
  - `/tmp` → `/sysroot/tmp`
 


elboulangero:
However `/var/local` already exists, and according to https://www.tldp.org/LDP/sag/html/var-fs.html:

> **/var/local** Variable data for programs that are installed in /usr/local (i.e., programs that have been installed by the system administrator). Note that even locally installed programs should use the other /var directories if they are appropriate, e.g., /var/lock.

So maybe it's not a typo?

jlebon:
> So maybe it's not a typo?

Right, as alluded to in the paragraph above, `/usr` is part of the deployment chroot and so will eventually be garbage collected as you upgrade. The `/usr/local` dir (and other dirs mentioned there) works around this by being a symlink to `/var/usrlocal` so that it persists across upgrades. HTH!

elboulangero:
@jlebon Thanks for your reply.

What I meant is that the documentation suggests the following symlink:

    /usr/local → /var/local

Then just after the `tmpfiles.d` snippet creates directories in `var/usrlocal`. That's a bit of confusing, and it looks like one of them is a typo. Maybe the symlink above should be `/usr/local → /var/usrlocal`.

See https://github.com/ostreedev/ostree/blob/master/docs/manual/adapting-existing.md, sorry but I can't find a way to get a link to the exact line.

cgwalters:
@elboulangero Can you redo the PR with that change?

jlebon:
Ahh right, I see the mixup now. Thanks for bringing this up @elboulangero!

elboulangero:
Thanks for your feedback, I updated the PR

jlebon:
Thanks! @rh-atomic-bot r+ bdba5af

jlebon:
@rh-atomic-bot retry

PR #1590, opened 2018-05-15
u-boot: add support for devicetree

gizero:
Likewise the corresponding support for syslinux introduced by commit
c5112c25e4519835c4cd53f4350c1b2f2a477746, this one enables writing devicetree
filename into the uEnv.txt environment file for u-boot.

Since u-boot does not strictly defines variable names, here 'fdt_file' was
chosen as it appear to be one the most frequently adopted names in u-boot
default environments. Outer boot logic should of course comply with this choice
and use $fdt_file as the device tree file name to pass to boot commands.

This was tested on a custom board booting with u-boot.

### DIFF
--- a/src/libostree/ostree-bootloader-uboot.c
+++ b/src/libostree/ostree-bootloader-uboot.c
@@ -140,6 +140,10 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot     *self,
       if (val)
         g_ptr_array_add (new_lines, g_strdup_printf ("ramdisk_image%s=%s", index_suffix, val));
 
+      val = ostree_bootconfig_parser_get (config, "devicetree");
+      if (val)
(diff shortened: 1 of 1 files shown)


cgwalters:
I guess a big question here is: if u-boot supports syslinux configs, what value is there in keeping the u-boot backend going forward? 

Anyways I'm not a u-boot expert.  Tagging in @wmanley who did the syslinux patch.  Is there anyone else who can give this patch a review?

The high level/code style looks fine.

cgwalters:
bot, add author to whitelist

gizero:
Hi! Gentle ping... Can we have this one merged? I see no objections or complaints. @wmanley Do you have any comment?

wmanley:
Sorry for the delay, this one slipped through.

I don't see anything wrong with the patch.  It's a shame we don't document our expectations WRT various bootloaders, but I guess that it'd get out of sync with the implementation.

I agree that it makes sense to keep the u-boot backend to maximise ostree's portability.

PR #1591, opened 2018-05-17
More useful ostree fsck

alexlarsson:
This adds a --all option to not stop on the first error, and prints out what commits each corrupted object is in.

### DIFF
--- a/src/ostree/ot-builtin-fsck.c
+++ b/src/ostree/ot-builtin-fsck.c
@@ -31,6 +31,7 @@
 
 static gboolean opt_quiet;
 static gboolean opt_delete;
+static gboolean opt_all;
 static gboolean opt_add_tombstones;
 static gboolean opt_verify_bindings;
 static gboolean opt_verify_back_refs;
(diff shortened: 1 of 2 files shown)


src/ostree/ot-builtin-fsck.c
```diff
@@ -94,7 +96,12 @@ fsck_one_object (OstreeRepo            *repo,
               (void) ostree_repo_delete_object (repo, objtype, checksum, cancellable, NULL);
               object_missing = TRUE;
             }
-          else
+          else if (opt_all)
+            {
+              *out_found_corruption = TRUE;
+              g_printerr ("%s\n", temp_error->message);
+            }
+          else if (!opt_all)
```
jlebon: Minor: let's just use `else` here? It's clearly equivalent, but it parses better. :)

src/ostree/ot-builtin-fsck.c
```diff
@@ -31,6 +31,7 @@
```
jlebon: This commit is also updating the libglnx submodule to an older commit.

src/ostree/ot-builtin-fsck.c
```diff
@@ -63,16 +63,31 @@ fsck_one_object (OstreeRepo            *repo,
   if (!ostree_repo_fsck_object (repo, objtype, checksum, cancellable, &temp_error))
```
jlebon: Minor: `s/Inclue/Include/` in the commit message.

cgwalters:
I rebased this, fixed the issues jlebon found.  Only thing I'd like to see on top is a basic test for `fsck --all`.

cgwalters:
I added a test case for `--all`.

@rh-atomic-bot r+ a7dd4a6

jlebon:
@rh-atomic-bot retry

Issue #894, opened 2017-05-30
ostree pull seems to do a lot of dns lookups


alexlarsson:
I've seen a lot of reports of DNS issues during flatpak install. It seems that sometimes one of the dns servers for gnome.org has a hiccup, which tends to abort the entire install.

<details>
<summary>Example user report</summary>
• [ironman:~] bratner $ flatpak install --user http://flatpak.pitivi.org/pitivi.flatpakref
This application depends on runtimes from:
  http://sdk.gnome.org/repo/
Configure this as new remote 'gnome' [y/n]: y
Installing: org.pitivi.Pitivi/x86_64/stable
Required runtime for org.pitivi.Pitivi/x86_64/stable (org.gnome.Platform/x86_64/3.22) is not installed, searching...
Found in remote gnome, do you want to install it? [y/n]: y
Installing: org.gnome.Platform/x86_64/3.22 from gnome
[#=                  ] Downloading: 0 bytes/183.1 MB (0 bytes/s)               
error: While pulling runtime/org.gnome.Platform/x86_64/3.22 from remote gnome: Error resolving 'sdk.gnome.org': Name or service not known
• [ironman:~] bratner $ ping sdk.gnome.org
PING sdk.gnome.org (209.132.180.169) 56(84) bytes of data.
64 bytes from sdk.gnome.org (209.132.180.169): icmp_seq=1 ttl=47 time=235 ms
64 bytes from sdk.gnome.org (209.132.180.169): icmp_seq=2 ttl=47 time=236 ms
^C
--- sdk.gnome.org ping statistics ---
2 packets transmitted, 2 received, 0% packet loss, time 1001ms
rtt min/avg/max/mdev = 235.710/235.972/236.235/0.552 ms
• [ironman:~] bratner $ dig sdk.gnome.org

; <<>> DiG 9.10.3-P4-Ubuntu <<>> sdk.gnome.org
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 1259
;; flags: qr ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;sdk.gnome.org.			IN	A

;; ADDITIONAL SECTION:
sdk.gnome.org.		772	IN	A	209.132.180.169

;; Query time: 1 msec
[...]

PR #1594, opened 2018-05-22
Support retrying downloads if they transiently fail

pwithnall:
Here’s a branch which adds support for retrying each network request in a pull operation if it fails with a transient error (like a socket timeout).

Static delta superblocks are currently not supported because they aren’t queued (that’s a separate issue). If this PR is accepted, I’ll put together another one to add queuing support for them, and remove the FIXME added in this one.

One issue which remains in my testing (using qdisc to increase latency and randomly drop packets) is that DNS queries fail a lot. That’s https://github.com/ostreedev/ostree/issues/894, and isn’t addressed in this PR. In real life, I would expect DNS queries to fail quite a lot less often than HTTP requests (since they’re tiny), but this does warrant further investigation.

I haven’t added tests to the branch because I have no idea how to write a unit test which reliably triggers this error handling path, short of writing a mock `GSocket` implementation and plumbing that in (which would result in a *huge* test harness). Suggestions welcome.

https://github.com/ostreedev/ostree/issues/878

### DIFF
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -1086,8 +1086,8 @@ on_request_sent (GObject        *object,
                   code = G_IO_ERROR_BROKEN_PIPE;
 #else
                   code = G_IO_ERROR_CONNECTION_CLOSED;
-                  break;
 #endif
+                  break;
                 default:
                   code = _ostree_fetcher_http_status_code_to_io_error (msg->status_code);
                   break;
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -1070,18 +1070,12 @@ on_request_sent (GObject        *object,
                 soup_uri_to_string (soup_request_get_uri (pending->request), FALSE);
 
               GIOErrorEnum code;
+
               switch (msg->status_code)
                 {
-                case SOUP_STATUS_NOT_FOUND:
-                case SOUP_STATUS_FORBIDDEN:
-                case SOUP_STATUS_GONE:
-                  code = G_IO_ERROR_NOT_FOUND;
-                  break;
+                /* These statuses are internal to libsoup, and not standard HTTP ones: */
                 case SOUP_STATUS_CANCELLED:
                   code = G_IO_ERROR_CANCELLED;
-                  break;
-                case SOUP_STATUS_REQUEST_TIMEOUT:
-                  code = G_IO_ERROR_TIMED_OUT;
                   break;
                 case SOUP_STATUS_CANT_RESOLVE:
                 case SOUP_STATUS_CANT_CONNECT:
@@ -1092,10 +1086,11 @@ on_request_sent (GObject        *object,
                   code = G_IO_ERROR_BROKEN_PIPE;
 #else
                   code = G_IO_ERROR_CONNECTION_CLOSED;
-#endif
                   break;
+#endif
                 default:
-                  code = G_IO_ERROR_FAILED;
+                  code = _ostree_fetcher_http_status_code_to_io_error (msg->status_code);
+                  break;
                 }
 
               {
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -1072,10 +1072,27 @@ on_request_sent (GObject        *object,
               GIOErrorEnum code;
               switch (msg->status_code)
                 {
-                case 404:
-                case 403:
-                case 410:
+                case SOUP_STATUS_NOT_FOUND:
+                case SOUP_STATUS_FORBIDDEN:
+                case SOUP_STATUS_GONE:
                   code = G_IO_ERROR_NOT_FOUND;
+                  break;
+                case SOUP_STATUS_CANCELLED:
+                  code = G_IO_ERROR_CANCELLED;
+                  break;
+                case SOUP_STATUS_REQUEST_TIMEOUT:
+                  code = G_IO_ERROR_TIMED_OUT;
+                  break;
+                case SOUP_STATUS_CANT_RESOLVE:
+                case SOUP_STATUS_CANT_CONNECT:
+                  code = G_IO_ERROR_HOST_NOT_FOUND;
+                  break;
+                case SOUP_STATUS_IO_ERROR:
+#if !GLIB_CHECK_VERSION(2, 44, 0)
+                  code = G_IO_ERROR_BROKEN_PIPE;
+#else
+                  code = G_IO_ERROR_CONNECTION_CLOSED;
+#endif
                   break;
                 default:
                   code = G_IO_ERROR_FAILED;
--- a/src/libostree/ostree-fetcher-util.c
+++ b/src/libostree/ostree-fetcher-util.c
@@ -218,3 +218,22 @@ _ostree_fetcher_should_retry_request (const GError *error,
 
   return FALSE;
 }
+
+/* Convert a HTTP status code representing an error from libsoup or libcurl to
+ * a #GIOError. This will return %G_IO_ERROR_FAILED if the status code is
+ * unknown or otherwise unhandled. */
+GIOError
+_ostree_fetcher_http_status_code_to_io_error (guint status_code)
+{
+  switch (status_code)
+    {
+    case 403:  /* SOUP_STATUS_FORBIDDEN */
+    case 404:  /* SOUP_STATUS_NOT_FOUND */
+    case 410:  /* SOUP_STATUS_GONE */
+      return G_IO_ERROR_NOT_FOUND;
+    case 408:  /* SOUP_STATUS_REQUEST_TIMEOUT */
+      return G_IO_ERROR_TIMED_OUT;
+    default:
+      return G_IO_ERROR_FAILED;
+    }
+}
--- a/src/libostree/ostree-fetcher-util.c
+++ b/src/libostree/ostree-fetcher-util.c
@@ -52,15 +52,15 @@ fetch_uri_sync_on_complete (GObject        *object,
   data->done = TRUE;
 }
 
-gboolean
-_ostree_fetcher_mirrored_request_to_membuf (OstreeFetcher  *fetcher,
-                                            GPtrArray     *mirrorlist,
-                                            const char     *filename,
-                                            OstreeFetcherRequestFlags flags,
-                                            GBytes         **out_contents,
-                                            guint64        max_size,
-                                            GCancellable   *cancellable,
-                                            GError         **error)
+static gboolean
+_ostree_fetcher_mirrored_request_to_membuf_once  (OstreeFetcher              *fetcher,
+                                                  GPtrArray                  *mirrorlist,
+                                                  const char                 *filename,
+                                                  OstreeFetcherRequestFlags   flags,
+                                                  GBytes                    **out_contents,
+                                                  guint64                     max_size,
+                                                  GCancellable               *cancellable,
+                                                  GError                    **error)
 {
   gboolean ret = FALSE;
   g_autoptr(GMainContext) mainctx = NULL;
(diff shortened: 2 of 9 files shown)


cgwalters:
`ostree-trivial-httpd` already has `--random-500s`; it might actually work to just remove the `for ... seq` loop there in the `test-pull-repeated.sh`?   Though we might need slightly more sophisticated logic like `--random-500s-between=X` and use "number of objects" for `X` or something.

src/libostree/ostree-repo-pull.c
```diff
@@ -2131,6 +2133,24 @@ process_one_static_delta_fallback (OtPullData   *pull_data,
...
 }
 
+static void
+enqueue_one_static_delta_part_request_s (OtPullData           *pull_data,
+                                         FetchStaticDeltaData *fetch_data)
+{
+  if (fetcher_queue_is_full (pull_data))
+    {
+      g_debug ("queuing fetch of static delta %s-%s part %u",
+               fetch_data->from_revision ? fetch_data->from_revision : "empty",
```
jlebon: Minor: you can use the GNU C extension `?:` for this: https://gcc.gnu.org/onlinedocs/gcc/Conditionals.html#Conditionals

pwithnall:
> Cool! How about we split those first 6 commits into a prep PR to start? They're all pretty straight-forward.

Done as #1599.

src/libostree/ostree-repo-pull.c
pwithnall: Done. I’ve also added https://github.com/ostreedev/ostree/pull/1599/commits/b4ea90b89449a498d92b405e8629d3175a2b15cc at the beginning of the branch to convert other conditionals in the file too.

src/libostree/ostree-repo-pull.c
```diff
@@ -502,8 +539,13 @@ idle_worker (gpointer user_data)
   scan_one_metadata_object (pull_data, checksum, scan_data->objtype,
                             scan_data->path, scan_data->recursion_depth,
                             scan_data->requested_ref, pull_data->cancellable, &error);
-  check_outstanding_requests_handle_error (pull_data, &error);
-  scan_object_queue_data_free (scan_data);
+
+  if (should_retry_request (error, scan_data->n_retries_remaining--))
```
jlebon: Why do we need to check for network errors here? `scan_one_metadata_object` either enqueues a request or just scans a local object, right?

src/libostree/ostree-repo-pull.c
```diff
@@ -1286,9 +1336,14 @@ meta_fetch_on_complete (GObject           *object,
...
   pull_data->n_outstanding_metadata_fetches--;
   pull_data->n_fetched_metadata++;
-  check_outstanding_requests_handle_error (pull_data, &local_error);
+
+  if (should_retry_request (local_error, fetch_data->n_retries_remaining--))
+    enqueue_one_object_request_s (pull_data, g_steal_pointer (&fetch_data));
+  else
+    check_outstanding_requests_handle_error (pull_data, &local_error);
+
   if (free_fetch_data)
```
jlebon: Not new here, though I wonder why we need these bools instead of using the `g_steal_pointer` pattern higher up. Anyway, we can look into that later!

src/libostree/ostree-repo-pull.c
```diff
@@ -478,6 +488,33 @@ scan_object_queue_data_free (ScanObjectQueueData *scan_data)
...
+{
+  if (error == NULL || n_retries_remaining == 0)
+    return FALSE;
+
+  /* Return TRUE for transient errors. */
+  if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_TIMED_OUT) ||
+      g_error_matches (error, G_IO_ERROR, G_IO_ERROR_HOST_NOT_FOUND) ||
+      g_error_matches (error, G_IO_ERROR, G_IO_ERROR_HOST_UNREACHABLE) ||
+      g_error_matches (error, G_RESOLVER_ERROR, G_RESOLVER_ERROR_NOT_FOUND) ||
+      g_error_matches (error, G_RESOLVER_ERROR, G_RESOLVER_ERROR_TEMPORARY_FAILURE))
```
jlebon: Wait, I'm confused now how this actually works. I don't see anywhere in either the libsoup or the libcurl fetchers where we translate curl/soup specific transport errors into these. We need to expand handling in those functions to tease out these transient issues, right? E.g. in the libcurl fetcher, I would expect to see something like:

```c
      else if (curlres != CURLE_OK)
        {
           switch (curlres)
             {
             case CURLE_COULDNT_RESOLVE_HOST:
                /* map to G_RESOLVER_ERROR_NOT_FOUND */
                break;
             case CURLE_COULDNT_CONNECT:
                /* map to G_IO_ERROR_HOST_UNREACHABLE */
                break;
             }
        }
```

or something similar...

src/libostree/ostree-repo-pull.c
pwithnall: Hmm, indeed. I blame over-enthusiastic grepping for calls to `check_outstanding_requests_handle_error()`.

src/libostree/ostree-repo-pull.c
pwithnall: I’m not sure it applies in this particular instance, but there are other similar patterns when enqueuing other structs, where `g_steal_pointer (&the_struct)` can’t be used because `the_struct` needs to be referred to again lower down, before it’s eventually potentially freed.

In any case, the code could be restructured to keep a `const` pointer to whatever it needs to refer to later, or something.

src/libostree/ostree-repo-pull.c
pwithnall: In the libsoup case, I guess I’m getting the `GSocket` errors propagated back up through libsoup. I could definitely improve `on_request_sent()` in `ostree-fetcher-soup.c` to propagate some of the higher level errors a bit more successfully though.

I haven’t tested with the libcurl backend.

cgwalters: I'm not sure it's worth trying to maintain a full mapping between the error domains, could just squash the ones we want to retry from curl to `G_IO_ERROR_HOST_UNREACHABLE` or whatever.

pwithnall:
> > `ostree-trivial-httpd` already has `--random-500s`; …
> 
> I’ll check that out tomorrow, thanks.

I’ve updated my branch with some tests based on this, which caught a few more places where network retries were needed.

While making these further changes I did wonder about whether I should be putting the retry support in at a lower level (e.g. in `ostree-fetcher-{soup,curl}.c`). However, there is an advantage to having it implemented in `ostree-fetcher.c`: in future, the queue code could use a timeout from one host as a hint to lower the priority of that host and prioritise trying other mirrors/peers when downloading a particular object. I haven’t implemented that here (and have no immediate plans to), but we have that flexibility in future if needed.

src/libostree/ostree-fetcher-soup.c
```diff
@@ -1072,11 +1072,24 @@ on_request_sent (GObject        *object,
               GIOErrorEnum code;
```
jlebon: One thing that feels odd to me is that the way this is structured, we'll be retrying the whole set of mirrors if the *last one* on the list returns a transient error. So e.g. the first N-1 mirrors could return `NOT_FOUND` and if the Nth mirror times out, we'll be retrying all N mirrors, right? Whereas I think what we want is to retry only the Nth mirror. It seems like fixing this properly would probably require indeed teaching each backend-specific fetcher about retries.

Anyway, I think it's definitely worth getting the current approach in and iterate on that since AFAIK mirrorlists are not widely used yet.

src/libostree/ostree-repo-pull.c
```diff
@@ -195,6 +204,7 @@ typedef struct {
   OstreeObjectType objtype;
   guint recursion_depth; /* NB: not used anymore, though might be nice to print */
   OstreeCollectionRef *requested_ref;  /* (nullable) */
+  guint n_retries_remaining;
```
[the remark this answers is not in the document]
pwithnall: Yes we do, it’s used in `queue_scan_one_metadata_object_c()`.

src/libostree/ostree-fetcher-soup.c
pwithnall: I’d definitely prefer to get the current approach in first. That said, I’m not sure whether I will have time afterwards to work on improvements to mirror handling (I’m working on this to fix a problem with our Endless downloads timing out, and mirrors don’t really impact on that).

However, I’m not entirely convinced about the situation you’re considering. Are we ever expecting the majority of mirrors to return `NOT_FOUND`? Sounds like an unlikely corner case to me. Even so, `OstreeFetcherSoup` maintains state about its mirrors, so I think the problem could be addressed by storing a bit more state there, rather than by rearchitecting things. Specifically, instead of storing a single `mirrorlist_idx` pointing to the best mirror to try, store some flags for each mirror, and track whether each mirror has returned `NOT_FOUND`, or returned a transient error, at all. Prioritise requesting from mirrors which haven’t. (This could also be implemented by storing a priority for each mirror, and demoting it on `NOT_FOUND` or transient errors.)

jlebon: To clarify, I wasn't suggesting you would fix those up. Just something that'd need to be done down the road. As I said, I'm not aware of any big users of mirrorlists yet, so it's not a pressing concern.

> Are we ever expecting the majority of mirrors to return `NOT_FOUND`? Sounds like an unlikely corner case to me.

Heh, sure. The `NOT_FOUND` was an example; anything "non-transient" would work. Anyway, I do agree with you that it could be done without lowering the retry logic down a level.

src/libostree/ostree-repo-pull.c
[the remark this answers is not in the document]
pwithnall: You are absolutely right, sorry. :slightly_frowning_face: 

Fixup pushed.

pwithnall:
I don’t understand why `test-pull-repeated.sh` is failing on Travis. It reliably works fine for me locally. :cry:

jlebon:
I rebased the PR and squashed all the fixups since it makes individual commits harder to review.

src/ostree/ostree-trivial-httpd.c
```diff
@@ -511,6 +524,13 @@ run (int argc, char **argv, GCancellable *cancellable, GError **error)
       goto out;
     }
 
+  if (!(opt_random_408s_percentage >= 0 && opt_random_408s_percentage <= 99))
```
jlebon: Minor: the way I like to write these range type checks usually is:

```c
if (!(0 <= opt_random_408s_percentage && opt_random_408s_percentage <= 99))
```

jlebon: Ahh that's OK, I see you're just following the existing formatting of the 500s one.

src/libostree/ostree-fetcher-util.c
```diff
@@ -144,3 +175,46 @@ _ostree_fetcher_journal_failure (const char *remote_name,
...
+#else
+      g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CONNECTION_CLOSED) ||
+#endif
+      g_error_matches (error, G_RESOLVER_ERROR, G_RESOLVER_ERROR_NOT_FOUND) ||
+      g_error_matches (error, G_RESOLVER_ERROR, G_RESOLVER_ERROR_TEMPORARY_FAILURE))
+    {
+      g_debug ("Should retry request (remaining: %u retries), due to transient error: %s",
+               n_retries_remaining, error->message);
+      return TRUE;
+    }
```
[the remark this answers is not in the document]
pwithnall: That’s already covered by one of the `g_debug()`s at the top of the function, no?
```
    g_debug ("%s: error: %u:%u %s, n_retries_remaining: %u",
             G_STRFUNC, error->domain, error->code, error->message,
             n_retries_remaining);
```

pwithnall:
Ah, the CI test failures are when building with libcurl. I’ll see if I can do anything about that.

pwithnall:
> Ah, the CI test failures are when building with libcurl. I’ll see if I can do anything about that.

Fixed, I think, although some of the CI tests are hitting an internal error.

src/libostree/ostree-fetcher-soup.c
```diff
@@ -1092,10 +1086,11 @@ on_request_sent (GObject        *object,
                   code = G_IO_ERROR_BROKEN_PIPE;
 #else
                   code = G_IO_ERROR_CONNECTION_CLOSED;
-#endif
```
jlebon: Did you mean to move this `#endif`?

jlebon:
bot, retest this please

src/libostree/ostree-fetcher-soup.c
pwithnall: argh, no, the problems of squashing two diffs

jlebon:
```
/tmp/test-tmp-libostree_test-pull-repeated.sh.test-HFQKJZ /tmp/test-tmp-libostree_test-pull-repeated.sh.test-HFQKJZ
7 metadata, 5 content objects fetched; 1 KiB transferred in 0 seconds
Success on iteration 2
```

So here it only needed 2 iterations, which feels pretty low for a test with 40 iterations. But I think it checks out. We're fetching 12 objects, and the chance that one of those returns a 408 5 times in a row is... 4%? Hmm, maybe we should add a `ostree pull --network-retries=N` switch to lower retries to 2 or something. That'd be useful for general use too. The fact that it's greater than 1 at least is comforting.

Anyway, we can easily fix up and tweak these things later.
This looks great overall! Thanks a lot for working on this.

@rh-atomic-bot r+ e283a05

pwithnall:
Thanks! I’ve updated the follow-up #1600 to queue static delta superblock requests.

PR #1596, opened 2018-05-22
Make P2P API public (no longer experimental)

mwleeds:
Currently the API that allows P2P operations (e.g. pulling an ostree ref
from a LAN or USB source) is hidden behind the configure flag
--enable-experimental-api. This commit makes the API public and makes
that flag essentially a no-op (leaving it in place in case we want to
use it again in the future). The P2P API has been tested over the last
several months and proven to work.

This means that since we're no longer using the "experimental" feature
flag, P2P builds of Flatpak will fail when using versions of OSTree from
this commit onwards, until Flatpak is patched in the near future. If you
want to build Flatpak < 0.11.8 with P2P enabled and link against OSTree
2018.6, you'll have to patch Flatpak.  However, since Flatpak won't yet
have a hard dependency on OSTree 2018.6, it needs a new way to determine
if the P2P API in OSTree is available, so this commit adds a "p2p"
feature flag. This way the feature set is more semantically correct than
if we had continued to use the "experimental" feature flag.

In addition to making the P2P API public, this commit makes the P2P unit
tests run by default, changes a few man pages to reflect the changes,
updates the .gitignore, and updates the bash completion to account for
P2P related command line options.


mwleeds:
@cgwalters could you do a release with this commit included in the near future? Sorry to mess up your normal schedule but I'm trying to get Flathub P2P enabled in time for Endless 3.4.1. Thanks!

mwleeds:
Interestingly the CI failure here:
```
ot-builtin-find-remotes.c:(.text+0xbf9): undefined reference to `ostree_repo_find_remotes_async'
```
isn't unique to this branch. You can reproduce it on the master branch using the configure flags used by the CI but with `--enable-experimental-api` instead of `--disable-experimental-api`:
```
./autogen.sh --sysconfdir=/etc --prefix=/usr --libdir=/usr/lib64 --enable-gtk-doc --without-curl --without-soup --disable-gtk-doc --disable-man --disable-rust --without-libarchive --without-selinux --without-smack --without-openssl --without-avahi --without-libmount --disable-rofiles-fuse --enable-experimental-api
```

src/ostree/main.c
```diff
@@ -63,14 +63,12 @@ static OstreeCommand commands[] = {
   { "export", OSTREE_BUILTIN_FLAG_NONE,
     ostree_builtin_export,
     "Stream COMMIT to stdout in tar format" },
-#ifdef OSTREE_ENABLE_EXPERIMENTAL_API
   { "find-remotes", OSTREE_BUILTIN_FLAG_NONE,
     ostree_builtin_find_remotes,
     "Find remotes to serve the given refs" },
   { "create-usb", OSTREE_BUILTIN_FLAG_NONE,
     ostree_builtin_create_usb,
     "Copy the refs to a USB stick" },
```
pwithnall: These two `ostree` subcommands (`find-remotes` and `create-usb`) need to be added to `bash/ostree` to get completion support.

mwleeds: Fixed this

pwithnall:
>  * `ostree_repo_pull_from_remotes_async()` should probably just refer to the main `ostree_repo_pull_with_options()` option set

That’s not entirely trivial to do, since they support different subsets of the options. Each of the options which is supported by both functions has the same semantics in both implementations.

For example, `ostree_repo_pull_from_remotes_async()` will never support the `refs` or `override-commit-ids` options.

Furthermore, an explicit `copy_option()` call has to be added to `ostree_repo_pull_from_remotes_async()` for each option it wants to pass through to `ostree_repo_pull_with_options()`. On that basis, I think it’s best to just explicitly list the options for the two functions separately, but copy the same wording for each equivalent option. Otherwise I can see the list of options in `ostree_repo_pull_with_options()` growing (and its documentation growing), but no `copy_option()` calls being added to `ostree_repo_pull_from_remotes_async()`, and hence its documentation pointing to something which is incorrect.

pwithnall:
@cgwalters: to be clear, are you happy with all of the API here being set in stone for ever more?

cgwalters:
> @cgwalters: to be clear, are you happy with all of the API here being set in stone for ever more?

What would be the alternatives?  Split it into a higher level library?  Doesn't seem worth it.  Something else?

pwithnall:
> What would be the alternatives? Split it into a higher level library? Doesn't seem worth it. Something else?

Tweak it now before making it public. I’m happy the API is the right shape and architecture. There might be naming niggles or minor changes you’d want to make around the function arguments or whatever, which might be annoying once stable.

I was basically just double-checking that you’re happy with this and are not being pressured into getting it in quickly over getting it in correctly.

mwleeds:
For the record I also want to get it right and not just do it quickly. I've looked at most of this API to see if it seems future proof, and I'll look through it again before considering this ready to merge (still need to fix the build failures also of course).

@pwithnall: you said there were discussions that were postponed to this point. Would they be on old PRs or somewhere else?

pwithnall:
> For the record I also want to get it right and not just do it quickly.

I didn’t mean to imply otherwise!

> @pwithnall: you said there were discussions that were postponed to this point. Would they be on old PRs or somewhere else?

I honestly can’t remember. There are a few `FIXME` comments in the experimental code: if any of those affect the API, we should probably look into them. Any discussions which didn’t result in a `FIXME` probably aren’t worth picking up again.

cgwalters:
@mwleeds Can you split the prep commits to a separate PR?

mwleeds:
> @mwleeds Can you split the prep commits to a separate PR?

I'll get the CI happy with this branch before splitting it off; otherwise I'd have no way to know the prep commits have their intended effect.

cgwalters:
Currently the flatpak context is blocking though; we could disable it, or land the prep fixes in flatpak first, then bump the CI context here to pull that newer flatpak version?

mwleeds:
>Currently the flatpak context is blocking though; we could disable it, or land the prep fixes in flatpak first, then bump the CI context here to pull that newer flatpak version?

Sure, we can do the latter

mwleeds:
I ended up using a version check in flatpak, so maybe it makes sense to not add the p2p feature flag

alexlarsson:
This is breaking gnome continuos, because flatpak is now assuming 2018.6  means that the p2p APIs are there, and they are not. I'm gonna revert that change until this is ready...

cgwalters:
Hm, can't the interlock be avoided by building gcontinuous with the experimental APIs?  Or alternatively, changing the API detection to look for a function rather than a version number.  We could also add a `p2p` feature to `ostree --version` that one could use from autoconf or whatever.

mwleeds:
I changed flatpak to use the feature flag: https://github.com/flatpak/flatpak/commit/13f0ee5d5bff4581f18529fd30c43ed50f557230

mwleeds:
So we either need a new 0.10 release of flatpak or we could stop using a tag in `ci/flatpak.sh`. Up to you @alexlarsson?

alexlarsson:
I'd rather not do a LTS release just because of a technicality like this. Can we not point the CI at the flatpak-0.10 branch instead of a tag. That branch changes very little anyway.

mwleeds:
bot, retest this please

mwleeds:
Alright, so I found all the FIXME/TODO/XXX comments in the affected files and looked at each. In my opinion none of them should be considered blockers to merging this PR, and only three seem like they have the potential to cause bugs (see attached).

[ostree_p2p_fixme_comments.csv.txt](https://github.com/ostreedev/ostree/files/2064529/ostree_p2p_fixme_comments.csv.txt)

cgwalters:
Hm, not obvious to me what failed with the installed tests.

mwleeds:
I suppose this is a good opportunity for me to install Atomic so I can debug this.

cgwalters:
Let's just try it and see if it bounces.

@rh-atomic-bot r+ eeda9ea

cgwalters:
Ah.  For excessive OCD, mind squashing the (effective) fixup commit into the main one?

mwleeds:
It's a pretty separate concern than the main commit, no?

cgwalters:
OK, but it's also a requirement of the next commit right?  We don't bisect much today, but let's try to keep things working between commits too.

mwleeds:
Only if you're linking against an old glib

mwleeds:
I changed the order of the commits

cgwalters:
@rh-atomic-bot r+ 22806ee

vtolstov:
can somebody provide more info how to use p2p via lan?

vtolstov:
@mwleeds Thanks!

PR #1598, opened 2018-05-24
bash-completion: Don't add a space after files and directories

mwleeds:
Currently if ostree is completing a file or directory for you it adds a
space to the end, but this is an inconvenience when it's an intermediate
directory in a tree. It's better to let the user add the space after the
final directory, so this commit changes the bash completion to avoid
adding a space when completing files or directories.


jlebon:
Thanks!
@rh-atomic-bot r+ 00a8388

PR #1599, opened 2018-05-25
Preparation for supporting retrying downloads

pwithnall:
Preparatory commits for #1594.


jlebon:
@rh-atomic-bot r+ 70fe54f

PR #1600, opened 2018-05-28
Queue static delta superblocks

pwithnall:
This is a follow-up from #1594 (and is based on it), adding support for queuing downloads of static delta superblocks.


jlebon:
Nice, LGTM!
@rh-atomic-bot r+ bedb796

PR #1601, opened 2018-05-28
ci: Bump to f28

jlebon:
We successfully switched to F28 in rpm-ostree. Time to follow suit here.


cgwalters:
Hm, so this one is just stuck on flatpak now.  Maybe change that one back to f27 for now, let's get everything else in?

jlebon:
Done! :arrow_up:
I also squashed the fixups.

PR #1602, opened 2018-05-29
Cleanup current boot's staging dir if we hit min-free-space-* condition

uajain:
If the pull failed due to lack of free space error, it is
not recommended to hold onto partial caches (in repo/tmp) which
can leave the disk-space full. Low-level free disk space error
reporing is not reliable so we alternatively check for free space
using fstatfs(). Hence, we keep partial caches only if free space is
above a certain threshold. This can be controlled by cache-threshold
config parameter.

### DIFF
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -152,7 +152,7 @@ struct OstreeRepo {
   gid_t target_owner_gid;
   guint min_free_space_percent; /* See the min-free-space-percent config option */
   guint64 min_free_space_mb; /* See the min-free-space-size config option */
-  gboolean emergency_cleanup;
+  gboolean cleanup_stagedir;
 
   guint test_error_flags; /* OstreeRepoTestErrorFlags */
 
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -152,6 +152,7 @@ struct OstreeRepo {
   gid_t target_owner_gid;
   guint min_free_space_percent; /* See the min-free-space-percent config option */
   guint64 min_free_space_mb; /* See the min-free-space-size config option */
+  gboolean emergency_cleanup;
 
   guint test_error_flags; /* OstreeRepoTestErrorFlags */
 
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -522,6 +522,7 @@ _ostree_repo_bare_content_commit (OstreeRepo                 *self,
       const fsblkcnt_t object_blocks = (st_buf.st_size / self->txn.blocksize) + 1;
       if (object_blocks > self->txn.max_blocks)
         {
+          self->emergency_cleanup = TRUE;
           g_mutex_unlock (&self->txn_lock);
           g_autofree char *formatted_required = g_format_size (st_buf.st_size);
           if (self->min_free_space_percent > 0)
@@ -924,6 +925,7 @@ write_content_object (OstreeRepo         *self,
       if (object_blocks > self->txn.max_blocks)
         {
           guint64 bytes_required = (guint64)object_blocks * self->txn.blocksize;
+          self->emergency_cleanup = TRUE;
           g_mutex_unlock (&self->txn_lock);
           g_autofree char *formatted_required = g_format_size (bytes_required);
           if (self->min_free_space_percent > 0)
@@ -1623,6 +1625,7 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
     return FALSE;
 
   self->in_transaction = TRUE;
+  self->emergency_cleanup = FALSE;
   if (self->min_free_space_percent >= 0 || self->min_free_space_mb >= 0)
     {
       struct statvfs stvfsbuf;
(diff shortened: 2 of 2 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -350,6 +351,30 @@ pull_termination_condition (OtPullData          *pull_data)
...
+  int dfd;
+  int free_space;
+  struct statfs st_buf;
+
+  dfd = ostree_repo_get_dfd (repo);
+
+  if (fstatfs (dfd, &st_buf) < 0)
+    return FALSE;
+
+  free_space = st_buf.f_bavail * st_buf.f_bsize >> 30; /* Convert to GB */
```
lucab: Isn't the first multiplication very likely to overflow on large disks? Also, why does this need a *signed* integer?

uajain: Also, I have been suggested to keep a granularity acceptable in `MB` too, so I am thinking here to follow `systemd` convention like `200MB` or `2G`. This will require some parsing of the parameter in the config file.

cgwalters:
Hmm.  Feels like this should relate to `min-free-space-percent` right?  Basically delete caches if we go past that threshold?  (And yes, we'd have to figure out how to work around the fact that flatpak/Endless disable that check, maybe add a new option `min-cache-space-percent` that has the same default value?)

Though to start it seems simpler to just unconditionally delete the caches if we hit `ENOSPC`.

cgwalters:
On the code side I'd look at the implementation of `min-free-space-percent` too.  If we go with the percentage based approach then extracting a helper there would make sense.

src/libostree/ostree-repo-pull.c
```diff
@@ -366,6 +391,18 @@ check_outstanding_requests_handle_error (OtPullData          *pull_data,
...
+
+          if (!dir_check_for_free_space (pull_data->repo))
+            {
+              g_autofree char *msg = g_strdup (error->message);
+
+              g_clear_error (errorp);
+              g_set_error (errorp,
+                           G_IO_ERROR,
+                           G_IO_ERROR_NO_SPACE,
+                           "Not enough disk space : %s", msg) ;
```
cgwalters: It's weird that we're replacing the error - if I get a 404 because I typo'd an URL but see an error message about disk space that'd be very confusing.   Maybe instead suffix the error?

```
g_autofree char *msg = g_steal_pointer (&error->message);
error->message = g_strconat (msg, " (low disk, cache pruned)", NULL);
```
or so?

cgwalters:
> I don't think percentage based evaluation is a good thing here and that might be the case min-free-space-percent is disabled, because on 1TB HDD, it translates to have like ~30GB(3%) of free 

Yes, but an absolute default (in your patch currently, 5GB) works equally badly for small devices (and VMs, even for production ones it's not unusual to have a relatively small rootfs and have a large external volume elsewhere).

> I also want to emphasize that, ENOSPC is not guaranteed by error-propagation. 

I can certainly imagine something else going wrong in some random part of the libcurl or gio libraries or whatever if there's no free space, but I would very much expect to get `ENOSPC` propagated up as `G_IO_ERROR_NO_SPACE` in most cases - notably from the `fallocate()` calls we do before writing object content.  If that's not happening that's definitely a bug to fix first.

Then a next phase would be an active space check like you have here; we could simply check whether the disk is "basically" full (only has `0.1%` space left or something) after a pull op if we had an error that wasn't already `ENOSPC` and in that case also purge the caches.

uajain:
> Yes, but an absolute default (in your patch currently, 5GB) works equally badly for small devices (and VMs, even for production ones it's not unusual to have a relatively small rootfs and have a large external volume elsewhere).

Yes, the default value(Currently 5GB) is debatable but just mentioning here that the `cache-threshold` like `min-free-space-percent` can be configured in `repo/config`.
Just need to tweak it a bit (see https://github.com/ostreedev/ostree/pull/1602#discussion_r191650760) 

I will probably need to look into ENOSPC propagation and get back to you. Afair, Alex also mentioned about unreliability of ENOSPC error codes.

cgwalters:
> Based on Alex's recommendation, currently work-in-progress to extend min-free-space-percent option to accept specific size (1GB, 750MB etc.)

Makes sense to me.  Also one thing that would probably help is to still do space checks with `min_free_space_percent=0`.

uajain:
> Based on Alex's recommendation, currently work-in-progress to extend min-free-space-percent option to accept specific size (1GB, 750MB etc.) 

In https://github.com/ostreedev/ostree/pull/1616 (merged), we now have an `min-free-space-size` config  option to specify exact sizes. The next step would be to clean up staging dir if we hit the min-free-space-* limits. I will update this PR in a while.

src/libostree/ostree-repo-private.h
```diff
@@ -152,6 +152,7 @@ struct OstreeRepo {
   gid_t target_owner_gid;
   guint min_free_space_percent; /* See the min-free-space-percent config option */
   guint64 min_free_space_mb; /* See the min-free-space-size config option */
+  gboolean emergency_cleanup;
```
jlebon: Bikeshed: how about `cleanup_stagedir`? The term "emergency" there is not very precise.

uajain: Addressed the issue in a following fixup! commit. 
Thank you for the review! :+1:

jlebon:
Thanks for the patch!
@rh-atomic-bot r+ eef4fa1

Issue #1613, opened 2018-06-06
Static deltas size should be checked against txn.max_blocks


uajain:
While solving for https://github.com/ostreedev/ostree/pull/1602#issuecomment-394469571 
Currently the size of the delta is compared against the **total** free-space [here]

What I propose is that, it should be compared against max space that can be acquired by the transaction. ( This inherently will honor the `min-free-space-percent` or similar criteria )

`if (delta_required_blocks > pull_data->repo->txn.max_blocks)`

[here]: https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-pull.c#L2505

uajain:
I might be very wrong here  for non-obvious reasons but I just want to run it by experts to understand the situation better. Opened a PR for review, just in case.

uajain:
PR #1614 Merged in master. Closing this issue.

PR #1614, opened 2018-06-06
lib/repo: Enforce min-free-space-* size check for regfile writes in deltas

uajain:
https://github.com/ostreedev/ostree/issues/1613

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -509,6 +509,33 @@ _ostree_repo_bare_content_commit (OstreeRepo                 *self,
 {
   OstreeRealRepoBareContent *real = (OstreeRealRepoBareContent*) barewrite;
   g_assert (real->initialized);
+
+  if ((self->min_free_space_percent > 0 || self->min_free_space_mb > 0) && self->in_transaction)
+    {
+      struct stat st_buf;
+      if (!glnx_fstat (real->tmpf.fd, &st_buf, error))
+        return FALSE;
+
+      g_mutex_lock (&self->txn_lock);
+      g_assert_cmpint (self->txn.blocksize, >, 0);
+
+      const fsblkcnt_t object_blocks = (st_buf.st_size / self->txn.blocksize) + 1;
+      if (object_blocks > self->txn.max_blocks)
+        {
+          g_mutex_unlock (&self->txn_lock);
+          g_autofree char *formatted_required = g_format_size (st_buf.st_size);
+          if (self->min_free_space_percent > 0)
+           return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
+                              self->min_free_space_percent, formatted_required);
+          else
+            return glnx_throw (error, "min-free-space-size %luMB woulid be exceeded, %s more required",
+                               self->min_free_space_mb, formatted_required);
+        }
+      /* This is the main bit that needs mutex protection */
+      self->txn.max_blocks -= object_blocks;
+      g_mutex_unlock (&self->txn_lock);
+    }
+
   ot_checksum_get_hexdigest (&real->checksum, checksum_buf, buflen);
 
   if (real->expected_checksum &&
--- a/tests/installed/nondestructive/itest-pull-space.sh
+++ b/tests/installed/nondestructive/itest-pull-space.sh
@@ -37,9 +37,6 @@ fi
 assert_file_has_content err.txt "min-free-space-size"
 echo "ok min-free-space-size (error)"
 
-umount mnt
-losetup -d ${blkdev}
-rm testblk.img
 
 # min-free-space-size success
 ostree --repo=repo init --mode=bare-user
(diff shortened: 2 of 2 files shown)


jlebon:
bot, add author to whitelist

uajain:
@jlebon As far as I can see, during static delta fetch,  objects are written using `writing_content_objects` (initially) in `ostree-repo-commit.c` and then there is a series of calls to `ostree_repo_bare_content_open/write` APIs.

As `writing_content_objects`  is guarded against min-free-space check and other one is not, I _don't_ quite understand the difference here. Both of functions relate it to writing on disk(I think) but I am so confused. Can you help clearing it up?

src/libostree/ostree-repo-pull.c
```diff
@@ -2508,11 +2508,11 @@ process_one_static_delta (OtPullData                 *pull_data,
    * walk everything twice and keep track of state.
    */
   const guint64 delta_required_blocks = (pull_data->total_deltapart_usize / stvfsbuf.f_bsize);
```
cgwalters: If we do this there's no point in doing the `statvfs()` above, instead of `stvfsbuf.f_bsize` we can use `pull_data->repo->txn.blocksize` like the other code does.

src/libostree/ostree-repo-commit.c
```diff
@@ -493,6 +493,33 @@ _ostree_repo_bare_content_write (OstreeRepo                 *repo,
...
   OstreeRealRepoBareContent *real = (OstreeRealRepoBareContent*) barewrite;
   g_assert (real->initialized);
+
+  if ((repo->min_free_space_percent > 0 || repo->min_free_space_size > 0) && repo->in_transaction)
+    {
+      g_mutex_lock (&repo->txn_lock);
+      g_assert_cmpint (repo->txn.blocksize, >, 0);
+
+      struct stat st_buf;
+      if (!glnx_fstat (real->tmpf.fd, &st_buf, error))
```
uajain: During testing out this patch today, I realized the `st_size` of real->tmpf.fd , is turning out to be 0. Looks like it has to do something with O_TMPFILE. Investigating.

uajain: >Thinking more on this, I'm not sure if this is the best place to check for this since we're always rounding up on each write so we might be greatly overestimating the space a new object takes once we're done with it.

Can you explain this a bit more. I am not sure if I follow. In my observation, when there is a static delta fetch (1-part), there are multiple calls to `_ostree_repo_bare_content_commit` . I thought it would be just at-most one call to `_ostree_repo_bare_content_commit` to commit. Maybe my knowledge of deployment from static-deltas is lacking.

jlebon: Each delta part can have instructions for building multiple files. Check out `_ostree_static_delta_part_execute`. For each content object that gets created, there is a call to `_ostree_repo_bare_content_open`, one or more calls to `_ostree_repo_bare_content_write`, and then a final call to `_ostree_repo_bare_content_commit`. I'm simplifying a bit though (there's different paths for e.g. metadata objects and for committing to archive repos).

uajain:
Shifted the code block in `_ostree_repo_bare_content_commit` where we finally know the final size of the new content object. I have one doubt which I have addressed below

src/libostree/ostree-repo-commit.c
```diff
@@ -509,6 +509,33 @@ _ostree_repo_bare_content_commit (OstreeRepo                 *self,
...
+      if (object_blocks > self->txn.max_blocks)
+        {
+          g_mutex_unlock (&self->txn_lock);
+          g_autofree char *formatted_required = g_format_size (st_buf.st_size);
+          if (self->min_free_space_percent > 0)
+           return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
+                              self->min_free_space_percent, formatted_required);
+          else
+            return glnx_throw (error, "min-free-space-size %luMB would be exceeded, %s more required",
+                               self->min_free_space_size, formatted_required);
```
uajain: These are not particularly helpful error messages, especially the `... more required` part. Currently it denotes the space required to commit for this particular object. There might be subsequent objects(as part of a entire delta) which needs to be committed in subsequent calls.
Can we do better in terms of this error message?

uajain:
Damn, I need to push a fixup! once https://github.com/ostreedev/ostree/pull/1632 is merged. It has a few variable name changes.

src/libostree/ostree-repo-commit.c
```diff
@@ -509,6 +509,33 @@ _ostree_repo_bare_content_commit (OstreeRepo                 *self,
 {
   OstreeRealRepoBareContent *real = (OstreeRealRepoBareContent*) barewrite;
   g_assert (real->initialized);
+
+  if ((self->min_free_space_percent > 0 || self->min_free_space_mb > 0) && self->in_transaction)
+    {
+      g_mutex_lock (&self->txn_lock);
```
jlebon: Ca we can move the locking to after the `fstat`?

uajain: Ah yes, makes sense.

uajain:
Yes, I have been able to trigger it in my testing. Just keep enough free space for delta to download but keep `min-free-space-size` tight enough so that the delta can't be deployed. You'll notice that it is honored and you'll be left with the free disk space exactly as mentioned with `min-free-space-size` config option.

jlebon:
Awesome, thanks a lot for adding a test for this! I just squashed the fixup into the parent commit since it didn't have the correct title (we normally generate these with `git commit --fixup ...`).

@rh-atomic-bot r+ cc7876c

PR #1615, opened 2018-06-06
Don't include libglnx.h in the public header file

mwleeds:


mwleeds:
I think the CI will point out where `#include libglnx.h` needs to be added

mwleeds:
Looks like the `FAH28-insttests` failed under `ex-required`, but probably an unrelated error. I don't think I have permission to get the bot to retry though

mwleeds:
bot, retry please

pwithnall:
bot, retry please

tests/test-include-ostree-h.c
```diff
@@ -0,0 +1,52 @@
...
+#if GLIB_CHECK_VERSION(2, 44, 0)
+#include <ostree.h>
+#endif
+
+static void
+test_include_ostree_h_compiled (void)
+{
+#if !GLIB_CHECK_VERSION(2, 44, 0)
+      g_test_skip ("Skipping test without glib >= 2.44 to avoid depending on"
+                   "libglnx for g_autoptr backports");
```
jlebon: Can you expand on this? Now that you've removed the libglnx include from public headers, this should work even on clients with glib < 2.44, right? Our `GLIB_VERSION_MIN_REQUIRED` is 2.40.

mwleeds: `ostree.h` includes `ostree-autocleanups.h` which uses `G_DEFINE_AUTOPTR_CLEANUP_FUNC` which isn't defined on clients with glib < 2.44. It's different here than for a truly external program because `OSTREE_COMPILATION` is defined; see [here](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-autocleanups.h#L34).

jlebon: Right, but if we want to really test this like an external program would, then we shouldn't have `-DOSTREE_COMPILATION` set, right? Offhand, it seems like adding `-UOSTREE_COMPILATION` should do the trick.

jlebon:
The magic sentence you're looking for is "bot, retest this please". :) Don't worry too much about the `ex-required` one though, it's not used for gating yet.

tests/test-include-ostree-h.c
mwleeds: You're right, that seems like a better solution. I thought it might have side effects but it looks like that symbol is only used in one place.

jlebon:
Thanks! @rh-atomic-bot r+ 8461d5c

PR #1616, opened 2018-06-07
Adjust min-free-space-percent config option to accept specfic size

uajain:


src/libostree/ostree-repo-commit.c
```diff
@@ -1491,6 +1491,59 @@ devino_cache_lookup (OstreeRepo           *self,
...
+  else
+    {
+      gint shifts = 0;
+
+      while (!g_ascii_isalpha (dup_str[i]))
+        i++;
+
+      char unit = dup_str[i];
+      dup_str[i] = '\0';
+      gint size = g_ascii_strtoull (dup_str, NULL, 10);
```
cgwalters: `guint64` again.

src/libostree/ostree-repo-private.h
```diff
@@ -150,7 +150,7 @@ struct OstreeRepo {
   uid_t owner_uid; /* Cache of repo's owner uid */
   uid_t target_owner_uid; /* Ensure files are chowned to this uid/gid */
   gid_t target_owner_gid;
-  guint min_free_space_percent; /* See the min-free-space-percent config option */
+  const char *min_free_space_str; /* See the min-free-space config option */
```
cgwalters: I'm in favor of adding `min-free-space` as an alias for `min-free-space-percent`, but AFAICS this patch isn't doing that yet.

src/libostree/ostree-repo.c
```diff
@@ -2598,6 +2598,19 @@ append_one_remote_config (OstreeRepo      *self,
   return add_remotes_from_keyfile (self, remotedata, path, error);
 }
 
+static gboolean
+min_free_space_str_validate (char *min_free_space_str)
+{
+  GRegex *percent_regex = g_regex_new ("^([0-9]|[1-9][0-9])%$" , 0,0, NULL);
+  GRegex *size_regex = g_regex_new ("^[0-9]+(G|M|T)B$", 0,0, NULL);
```
cgwalters: `g_autoptr()` here - these are leaked currently.  (Also we tend to use `g_once_init...()` for regexps, see elsewhere in the code)

src/libostree/ostree-repo.c
```diff
@@ -2771,18 +2784,22 @@ reload_core_config (OstreeRepo          *self,
       self->zlib_compression_level = OSTREE_ARCHIVE_DEFAULT_COMPRESSION_LEVEL;
   }
 
-  { g_autofree char *min_free_space_percent_str = NULL;
+  { g_autofree char *min_free_space_str = NULL;
+    self->min_free_space_str = NULL;
+
     /* If changing this, be sure to change the man page too */
-    const char *default_min_free_space = "3";
+    const char *default_min_free_space = "3%";
```
cgwalters: Well...but we have a backcompat hazard here as we didn't previously require `%`.  So perhaps it is best to introduce a new config option.

src/libostree/ostree-repo.c
```diff
@@ -2771,18 +2784,22 @@ reload_core_config (OstreeRepo          *self,
       self->zlib_compression_level = OSTREE_ARCHIVE_DEFAULT_COMPRESSION_LEVEL;
   }
 
-  { g_autofree char *min_free_space_percent_str = NULL;
+  { g_autofree char *min_free_space_str = NULL;
+    self->min_free_space_str = NULL;
```
cgwalters: `g_clear_pointer (&self->min_free_space_str, g_free)`

src/libostree/ostree-repo.c
```diff
@@ -2771,18 +2784,22 @@ reload_core_config (OstreeRepo          *self,
...
                                             default_min_free_space,
-                                            &min_free_space_percent_str, error))
+                                            &min_free_space_str, error))
       return FALSE;
 
-    self->min_free_space_percent = g_ascii_strtoull (min_free_space_percent_str, NULL, 10);
-    if (self->min_free_space_percent > 99)
-      return glnx_throw (error, "Invalid min-free-space-percent '%s'", min_free_space_percent_str);
+   if (min_free_space_str_validate (min_free_space_str))
+     self->min_free_space_str = g_strdup (min_free_space_str);
```
cgwalters: `self->min_free_space_str = g_steal_pointer (&min_free_space_str);` and avoid an extra strdup.

src/libostree/ostree-repo.c
```diff
@@ -2598,6 +2598,19 @@ append_one_remote_config (OstreeRepo      *self,
...
 }
 
+static gboolean
+min_free_space_str_validate (char *min_free_space_str)
+{
+  GRegex *percent_regex = g_regex_new ("^([0-9]|[1-9][0-9])%$" , 0,0, NULL);
+  GRegex *size_regex = g_regex_new ("^[0-9]+(G|M|T)B$", 0,0, NULL);
+
+  if (g_regex_match (percent_regex, min_free_space_str, 0, NULL) ||
+      g_regex_match (size_regex, min_free_space_str, 0, NULL))
```
cgwalters: If we're going to use regexps it feels like we might as well use capture groups to get the data out, then `min_free_space_str_parse()` would just need to handle a `(G|M|T)B` suffix.

src/libostree/ostree-repo-commit.c
```diff
@@ -1491,6 +1491,59 @@ devino_cache_lookup (OstreeRepo           *self,
...
+            break;
+          case 'G':
+            shifts = 30;
+            break;
+          case 'T':
+            shifts = 40;
+            break;
+          default:
+                g_warning ("Should not be reached, min_free_space_parse\n");
+	}
```
cgwalters: Indentation seems off in various places like this.

cgwalters:
Also can you add commit messages to your commits?

src/libostree/ostree-repo.c
ramcq: Or interpret a bare number as a percentage for the back compat case?

uajain: @ramcq Won't it give a wrong idea? For e.g.
(i) `min-free-space-percent=3`
(ii) `min-free-space-percent=150MB`
or am I missing something?

I wonder if adding an alias and introducing new config option  are same thing? Do we have an exisitng alias case which I can look-up?

uajain: >If we have two options, an interesting question arises if we want to support setting both simultaneously - I'd be fine trying to do so, I'd also be fine making it an error for now.

I think, making it an error sounds logical to me here.


>I want to be clear here: Is this just an idea of yours or are you planning to actually ship a non-zero min-free-space in EndlessOS?

It's not just an idea but a step in direction of https://github.com/ostreedev/ostree/pull/1602#issuecomment-393289405 . Also, min-free-space (% or specific size) should guard against possibility of hitting `ENOSPC`. So yes, Endless would be very much interesting in this as it provides a bit more fine tuning of the config option.

nedrichards: Endless certainly wants to ship something that does this kind of thing.

uajain: @cgwalters  Quick question: If both `min-free-space` and `min-free-space-percent` are specified in the config file(for some odd reasons), do we error that out, or treat one of them as default (say `min-free-space`) and proceed ahead with discarding the other one ?

uajain:
Updated the PR. I have kept the same `min-free-space-percent` option as default (instead of `min-free-space-size`) for now; But in near future, it makes sense to keep `min-free-space-size` as default. Also, it needs some discussion to decide on the default value.

jlebon:
This looks sane to me. I added a test to `itest-pull-space.sh` as well.
@rh-atomic-bot r+ b7729cd

PR #1619, opened 2018-06-08
Makefile-libostree.am: Drop Makefile dependency

jlebon:
For some reason, this is causing `cargo build` to execute both during a
`make` and a `make install`. This in turn is causing problems on my
system since I'm using `rustup`, so `sudo make install` fails to find
`cargo` in the more restricted `PATH`.

Looking at the librsvg code as an example, I also don't see this
dependency. It seems like any `make` command implicitly depends on
making sure the various Automake makefiles are up to date anyway?


cgwalters:
> It seems like any make command implicitly depends on making sure the various Automake makefiles are up to date anyway?

I think this is more about *after* you changed the `Makefile`, making sure that targets are out of date.

The real problem here is that something is causing the `.a` file to be out of date right?  Hmm.  I'm not immediately reproducing this problem locally.

jlebon:
Yeah, not sure what happened here. I rebuilt from a clean workdir and no longer hit this!

PR #1626, opened 2018-06-14
lib/repo-commit: Delay propagation of errors from abort_transaction()

pwithnall:
If there’s a problem while aborting a transaction, store the error but
don’t report it until the end of the function — do a best effort at
clearing the rest of the transaction state first (since most of it
cannot fail).

If cleanup_tmpdir() fails (which, arguably, should not be a
showstopper), this allows a caller to recover and start a new
transaction in future.


jlebon:
Makes sense to me!
@rh-atomic-bot r+ 39c53e6

PR #1627, opened 2018-06-14
lib/prune: Don't modify dent->d_name in place

mwleeds:
Currently when I run `ostree prune` it hits a seg fault when the
hash_func is used (in this case g_str_hash) from the call stack
_ostree_repo_prune_tmp() -> g_hash_table_contains() ->
g_hash_table_lookup_node(). So the key, in this case dent->d_name, must
be corrupt in some way.

glnx_dirfd_iterator_next_dent() uses readdir() to get the dirent struct.
And according to the man page for readdir(3), "POSIX.1 explicitly notes
that this field should not be used as an lvalue" (in reference to
d_name). So this commit avoids modifying d_name in place and copies it
instead. This seems to avoid the seg fault.

### DIFF
--- a/src/libostree/ostree-repo-prune.c
+++ b/src/libostree/ostree-repo-prune.c
@@ -153,26 +153,29 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
       size_t len;
       gboolean has_sig_suffix = FALSE;
       struct dirent *dent;
+      g_autofree gchar *d_name = NULL;
 
       if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
         return FALSE;
       if (dent == NULL)
         break;
 
-      len = strlen (dent->d_name);
-      if (len > 4 && g_strcmp0 (dent->d_name + len - 4, ".sig") == 0)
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-prune.c
```diff
@@ -153,26 +153,28 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
...
+      g_autofree gchar *d_name = NULL;
 
       if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
         return FALSE;
       if (dent == NULL)
         break;
 
-      len = strlen (dent->d_name);
-      if (len > 4 && g_strcmp0 (dent->d_name + len - 4, ".sig") == 0)
+      d_name = g_strdup (dent->d_name);
```
jlebon: Can you add a comment that we can't reuse `dent->d_name` to increase the odds we don't regress on this?

mwleeds: Done

jlebon:
@rh-atomic-bot r+ 8047f9c

Issue #1628, opened 2018-06-15
ConditionNeedsUpdate needs /usr to have newer modification time than /etc


peterbaouoft:
Hi,
For `/usr` on Atomic host, the modification time is canonicalized to epoch (1970) by ostree, and this makes /usr always older than /etc.

However, to have system-sysusers( or any service with ConditionNeedsUpdate ) to trigger automatically during boot time, /usr needs to be newer than /etc due to ConditionNeedsUpdate=/etc in the service file:

```
[rbao@unused-10-15-17-147 ~]$ cat /usr/lib/systemd/system/systemd-sysusers.service 
[Unit]
Description=Create System Users
Documentation=man:sysusers.d(5) man:systemd-sysusers.service(8)
DefaultDependencies=no
Conflicts=shutdown.target
After=systemd-remount-fs.service
Before=sysinit.target shutdown.target systemd-update-done.service
ConditionNeedsUpdate=/etc

 ................................................
```
Reading a bit from systemd code base, it seems like ConditionNeedsUpdate needs /usr to be strictly newer than /etc (i.e only > and not >=). See [code](https://github.com/systemd/systemd/blob/master/src/shared/condition.c#L417-L481) and [documentation](https://github.com/systemd/systemd/commit/cdf0c7abf8594db37bbb0720f7c18a1765465110).  

Endless seem to already have an implementation to fix this: **Related commit:** https://github.com/endlessm/ostree/commit/9ef61fd1623ffed6580c0309aff8d1b29e671ec7  \o/
Migrated from https://github.com/projectatomic/rpm-ostree/issues/1415

cgwalters:
The idea behind https://github.com/endlessm/ostree/commit/9ef61fd1623ffed6580c0309aff8d1b29e671ec7 seems generally OK to me. I just have minor code style nits, and it'd be good to have a test.

@dbnicholson I vaguely recall this came up before - let's try to get this upstream?

cgwalters:
@peterbaouoft Would you mind testing this patch fixes the problem for rpm-ostree+sysusers?

peterbaouoft:
> @peterbaouoft Would you mind testing this patch fixes the problem for rpm-ostree+sysusers?

No, not at all, will be trying it out =).

peterbaouoft:
Yup, it does fix the problem for `ConditionNeedsUpdate` =). Now systemd-sysusers is triggered at boot time \o/.

dbnicholson:
[...]

PR #1631, opened 2018-06-18
deploy: Delete .updated file from /etc and /var on new deployments

dbnicholson:
Systemd units using ConditionNeedsUpdate run if the mtime of .updated in
the specified directory is newer than /usr. Since /usr has an mtime of
0, there's no way to have an older .updated file. Systemd units
typically specify ConditionNeedsUpdate=/etc or ConditionNeedsUpdate=/var
to support stateless systems like ostree.

Remove the file from the new deployment's /etc and the OS's /var
regardless of where they came from to ensure that these systemd units
run when booting new deployments. This will provide a method to run
services only on upgrade.

Closes: #1628
https://bugzilla.gnome.org/show_bug.cgi?id=752950

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -684,15 +684,10 @@ selinux_relabel_dir (OstreeSysroot                 *sysroot,
 static gboolean
 selinux_relabel_var_if_needed (OstreeSysroot                 *sysroot,
                                OstreeSePolicy                *sepolicy,
-                               OstreeDeployment              *deployment,
+                               int                            os_deploy_dfd,
                                GCancellable                  *cancellable,
                                GError                       **error)
 {
-  const char *osdeploypath = glnx_strjoina ("ostree/deploy/", ostree_deployment_get_osname (deployment));
-  glnx_autofd int os_deploy_dfd = -1;
-  if (!glnx_opendirat (sysroot->sysroot_fd, osdeploypath, TRUE, &os_deploy_dfd, error))
-    return FALSE;
-
   /* This is a bit of a hack; we should change the code at some
    * point in the distant future to only create (and label) /var
    * when doing a deployment.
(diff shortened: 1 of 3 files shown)


cgwalters:
bot, add author to whitelist

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2529,11 +2524,25 @@ sysroot_finalize_deployment (OstreeSysroot     *self,
         return FALSE;
     }
 
+  const char *osdeploypath = glnx_strjoina ("ostree/deploy/", ostree_deployment_get_osname (deployment));
+  glnx_autofd int os_deploy_dfd = -1;
+  if (!glnx_opendirat (self->sysroot_fd, osdeploypath, TRUE, &os_deploy_dfd, error))
+    return FALSE;
+
+  /* Ensure that the new deployment does not have /etc/.updated or
+   * /var/.updated so that systemd ConditionNeedsUpdate=/etc|/var services run
```
cgwalters: Not worth a respin, but just a note: Since `/var` is shared, this applies to all deployments.  It means that if an upgrade is interrupted, on reboot we'll run through `ConditionNeedsUpdate=/var` for the booted deployment too.

In practice, this should be fine; on e.g. Silverblue today that's just `systemd-journal-catalog-update.service`.

cgwalters:
@rh-atomic-bot r+ 19d1884

src/libostree/ostree-sysroot-deploy.c
dbnicholson: By a respin do you just mean for the comment? Because obviously unless the sysroot is changed to have /var per-deployment rather than per-os then nothing can be changed there.

My experience is that all the `ConditionNeedsUpdate` units are using it for optimization - they're relatively slow tasks that you don't want to run on every boot. But nothing enforces that and you do need to make the units relatively idempotent if they're performing actions that you really only want to do once (e.g. use a stamp file or something). Just for the record, we make heavy use of `ConditionNeedsUpdate` units at Endless to perform post-boot migration tasks for things that we can't update via the OS update. There's a pile of semi-horrifying things in https://github.com/endlessm/eos-boot-helper using this mechanism.

cgwalters: > By a respin do you just mean for the comment? 

Yep.  I would have phrased it like "Ensure the new deployment does not have /etc/.updated, and also delete the stateroot's /var/.updated, so that..." or something.

cgwalters:
We're going over the CoreOS/Container Linux patches, and this one is related:

https://github.com/coreos/systemd/commit/5afd5354fb81d20592e00ad9b70a52a68f25b73e

This sounds right but I'd love to have more detail about exactly what unit broke.

Issue #1618, opened 2018-06-08
flatpak tests fail when Avahi is enabled but not running


smcv:
I tried building Flatpak 0.11.8 for Debian experimental against libostree snapshot ecdebeb2, with Avahi support enabled. The installed-tests (in autopkgtest), which do not currently pull in `avahi-daemon`, fail with errors like this:

```
(flatpak install:7036): OSTree-WARNING **: 10:45:27.477: Avahi finder failed abnormally; removing it: Failed to create service browser: Bad state
/usr/lib/flatpak/installed-tests/Flatpak/test-unsigned-summaries.sh: line 81:  7036 Trace/breakpoint trap   ${FLATPAK} ${U} install test-repo org.test.App master
```

I think this is because the fix for #1542 is not (fully?) effective. The Avahi client does this:

```
  client = avahi_client_new (avahi_glib_poll_get (self->poll),
                             AVAHI_CLIENT_NO_FAIL,
                             client_cb, self, &failure);

  if (client == NULL)
    {
      if (failure == AVAHI_ERR_NO_DAEMON)
        g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
                     "Avahi daemon is not running: %s",
                     avahi_strerror (failure));
      else
        g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
                     "Failed to create finder client: %s",
                     avahi_strerror (failure));

      return;
    }
```

but the meaning of `AVAHI_CLIENT_NO_FAIL` is that, instead of failing with `AVAHI_ERR_NO_DAEMON`, the Avahi client will continue to try to connect in the background. Until it has successfully connected, most methods will fail with `AVAHI_ERR_BAD_STATE`.

cc @mwleeds @pwithnall @smspillaz

smcv:
(Endless developers: The packaging is available at https://salsa.debian.org/debian/ostree and https://salsa.debian.org/debian/flatpak if you want to send merge requests with any of your changes.)

mwleeds:
Looking at https://ci.debian.net/packages/f/flatpak/testing/amd64/ it doesn't seem like the tests are actually run. Is there somewhere else to look for this? Or, could you give steps to reproduce it locally?

smcv:
[...]

PR #1639, opened 2018-06-23
avahi: Fail immediately if we can't talk to D-Bus or Avahi

smcv:
We special-case AVAHI_ERR_NO_DAEMON to not cause warnings, but if
we pass AVAHI_CLIENT_NO_FAIL to avahi_client_new, we never actually
see AVAHI_ERR_NO_DAEMON. Instead, we will get AVAHI_ERR_BAD_STATE
when we try to use the client.

Bug: https://github.com/ostreedev/ostree/issues/1618
Signed-off-by: Simon McVittie <smcv@debian.org>

---

This is what I've applied in Debian for now to cope with #1618. It's one of two possible routes for solving that issue. The other would be rather more involved: if the `AvahiClient` can't immediately talk to Avahi, make all methods on the `OstreeRepoFinderAvahi` fail with `G_IO_ERROR_NOT_FOUND` or similar.

### DIFF
--- a/src/libostree/ostree-repo-finder-avahi.c
+++ b/src/libostree/ostree-repo-finder-avahi.c
@@ -1432,8 +1432,7 @@ ostree_repo_finder_avahi_start (OstreeRepoFinderAvahi  *self,
 
   g_assert (self->client == NULL);
 
-  client = avahi_client_new (avahi_glib_poll_get (self->poll),
-                             AVAHI_CLIENT_NO_FAIL,
+  client = avahi_client_new (avahi_glib_poll_get (self->poll), 0,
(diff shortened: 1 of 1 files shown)


cgwalters:
bot, add author to whitelist

cgwalters:
Just one mechanical note, you should say `Closes: #1618` to have github do the auto-close bit.

smcv:
Thanks, I pushed an amended commit message. (The `Bug:` thing is part of the Debian patch metadata convention.)

cgwalters:
Thanks!

@rh-atomic-bot r+ fd4df48

Issue #1640, opened 2018-06-23
test failures on 64-bit big-endian: g_str_has_prefix: assertion 'str != NULL' failed


smcv:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=902209

ostree 2018.6 has test failures on s390x and some unofficial ("ports") architectures, probably involving the newly-non-experimental Avahi support. s390x is a representative example:

```
PASS: tests/test-repo-finder-config 1 /repo-finder-config/init
PASS: tests/test-repo-finder-config 2 /repo-finder-config/no-configs
ERROR: tests/test-repo-finder-config - Bail out! GLib-FATAL-CRITICAL: g_str_has_prefix: assertion 'str != NULL' failed

# OSTree-DEBUG: ostree_repo_finder_resolve_all_async: Resolving refs [(org.example.Collection0, exampleos/x86_64/ref0), (org.example.Collection0, exampleos/x86_64/ref1), (org.example.Collection1, exampleos/x86_64/ref1), (org.example.Collection1, exampleos/x86_64/ref2), (org.example.Collection2, exampleos/x86_64/ref3)] with finders [OstreeRepoFinderConfig]
# OSTree-DEBUG: ostree_repo_finder_config_resolve_async: Checking 21474836480 remotes
# OSTree-DEBUG: Resolved ref (org.example.Collection0, exampleos/x86_64/ref0) to remote ?remote0?.
# OSTree-DEBUG: Resolved ref (org.example.Collection0, exampleos/x86_64/ref1) to remote ?remote0?.
# OSTree-DEBUG: Resolved ref (org.example.Collection0, exampleos/x86_64/ref0) to remote ?remote0-copy?.
# OSTree-DEBUG: Resolved ref (org.example.Collection0, exampleos/x86_64/ref1) to remote ?remote0-copy?.
# OSTree-DEBUG: Resolved ref (org.example.Collection1, exampleos/x86_64/ref2) to remote ?remote1?.
# OSTree-DEBUG: Ignoring remote ?remote1-bad-copy? due to it not advertising any of the requested refs.
# OSTree-DEBUG: Ignoring remote ?remote2? due to no valid collection ID being configured for it: Invalid collection ID (null)
Bail out! GLib-FATAL-CRITICAL: g_str_has_prefix: assertion 'str != NULL' failed
[...]

PR #1641, opened 2018-06-24
OstreeRepoFinderConfig: Fix guint/gsize confusion

smcv:
If a function has a guint "out argument", passing a pointer to a gsize
is not, in general, valid. On an ILP64 platform there is no problem
since guint and gsize are identical, but on an LP64 platform it will
overwrite only the first word of the gsize, leaving the second word
unaffected. On little-endian machines, if the second word is
zero-initialized (as it is here), the result is numerically equal to
the guint, but on big-endian machines the result is around 4 billion
times what it should be, resulting in
ostree_repo_finder_config_resolve_async() reading past the end of
the array and causing undefined behaviour.

In practice this caused assertion failures (and consequently test
failures) on Debian's s390x (z/Architecture), ppc64 (64-bit PowerPC)
and sparc64 (64-bit SPARC) ports.

Closes: #1640

### DIFF
--- a/src/libostree/ostree-repo-finder-config.c
+++ b/src/libostree/ostree-repo-finder-config.c
@@ -96,7 +96,7 @@ ostree_repo_finder_config_resolve_async (OstreeRepoFinder                  *find
   GHashTableIter iter;
   const gchar *remote_name;
   g_auto(GStrv) remotes = NULL;
-  gsize n_remotes = 0;
+  guint n_remotes = 0;
 
   task = g_task_new (finder, cancellable, callback, user_data);
   g_task_set_source_tag (task, ostree_repo_finder_config_resolve_async);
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-finder-config.c
```diff
@@ -106,9 +106,9 @@ ostree_repo_finder_config_resolve_async (OstreeRepoFinder                  *find
 
   /* List all remotes in this #OstreeRepo and see which of their ref lists
    * intersect with @refs. */
-  remotes = ostree_repo_remote_list (parent_repo, (guint *) &n_remotes);
```
cgwalters: Ahh, that cast was a warning sign.

cgwalters:
Thanks a lot for tracking this down!

@rh-atomic-bot r+ 3cfa656

cgwalters:
@rh-atomic-bot retry

src/libostree/ostree-repo-finder-config.c
smcv: They usually are...

cgwalters:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

jlebon:
This is the httpd flake.
@rh-atomic-bot retry

jlebon:
I opened #1644 to try to make investigating this easier.

cgwalters:
@rh-atomic-bot retry

PR #1643, opened 2018-06-25
 ostree repo commit: Speed up composing trees with `--tree=ref`

wmanley:
TODO
====

- [x] Investigate, create test for and fix bug that @wmanley is seeing locally
- [x] Merge #1645 and rebase
- [x] Merge #1655 and rebase

Description
=========

Running `ostree commit --tree=ref=a --tree=dir=b` involves reading the
whole of a into an `OstreeMutableTree` before composing `b` on top.  This
is inefficient if a is a complete rootfs and b is just touching one file.
We process O(size of a + size of b) directories rather than
O(number of touched dirs).

This commit makes `ostree commit` more efficient at composing multiple
directories together.  With `ostree_mutable_tree_fill_empty_from_dirtree`
we create a lazy `OstreeMutableTree` which only reads the underlying
information from disk when needed.  We don't need to read all the
subdirectories just to get the checksum of a tree already checked into the
repo.

This provides great speedups when composing a rootfs out of multiple other
rootfs as we do in our build system.  We compose multiple containers
together with:

    ostree commit --tree=ref=base-rootfs --tree=ref=container1 --tree=ref=container2

and it is much faster now.

As a test I ran

    time ostree --repo=... commit --orphan --tree=ref=big-rootfs --tree=dir=modified_etc

Where modified_etc contained a modified sudoers file under /etc.  I used
`strace` to count syscalls and I seperatly took timing measurements.  To
test with a cold cache I ran

    sync && echo 3 | sudo tee /proc/sys/vm/drop_caches

Results:

|                      | Before | After |
| -------------------- | ------ | ----- |
| Time (cold cache)    |   8.1s | 0.12s |
| Time (warm cache)    |   3.7s | 0.08s |
| `openat` calls       |  53589 |   246 |
| `fgetxattr` calls    |  78916 |     0 |

I'm not sure if this will have some negative interaction with the
`_ostree_repo_commit_modifier_apply` which is short-circuited here.  I
think it was disabled for `--tree=ref=` anyway, but I'm not certain.  All
the tests pass anyway.

I originally implemented this in terms of the `OstreeRepoFile` APIs, but
it was *way* less efficient, opening and reading many files unnecessarily.


src/libostree/ostree-mutable-tree.c
```diff
@@ -299,6 +446,74 @@ ostree_mutable_tree_ensure_parent_dirs (OstreeMutableTree  *self,
...
+set_string(gchar **out, const gchar *in)
+{
+  if (g_strcmp0(*out, in) == 0)
+    return;
+  g_free (*out);
+  *out = g_strdup (in);
+}
+
+/**
+ * ostree_mutable_tree_fill_empty_from_dirtree:
```
wmanley: Possibly a better name would be `ostree_mutable_tree_fast_merge`?  It's more intention revealing.  Or perhaps this should be moved into a `-private.h` header and not be made part of the public API?

jlebon:
bot, add author to whitelist

tests/basic-test.sh
```diff
@@ -450,16 +450,20 @@ echo sudoers2 >tree-C/etc/sudoers
...
 echo sudoers1 >tree-C/etc/sudoers
 
-$OSTREE commit ${COMMIT_ARGS} -b test3-mix1 -s "A" --tree=ref=test3-C1 --tree=ref=test3-D
-$OSTREE commit ${COMMIT_ARGS} -b test3-mix2 -s "A" --tree=dir=tree-C --tree=ref=test3-D
-$OSTREE commit ${COMMIT_ARGS} -b test3-mix3 -s "A" --tree=ref=test3-C1 --tree=dir=tree-D
-$OSTREE commit ${COMMIT_ARGS} -b test3-mix4 -s "A" --tree=dir=tree-C --tree=dir=tree-D
+$OSTREE commit ${COMMIT_ARGS} -b test3-ref-ref -s "A" --tree=ref=test3-C1 --tree=ref=test3-D
+$OSTREE commit ${COMMIT_ARGS} -b test3-dir-ref -s "A" --tree=dir=tree-C --tree=ref=test3-D
+$OSTREE commit ${COMMIT_ARGS} -b test3-ref-dir -s "A" --tree=ref=test3-C1 --tree=dir=tree-D
+$OSTREE commit ${COMMIT_ARGS} -b test3-dir-dir -s "A" --tree=dir=tree-C --tree=dir=tree-D
```
jlebon: Minor/optional: the `-s` switch is optional nowadays.

tests/basic-test.sh
```diff
@@ -450,16 +450,20 @@ echo sudoers2 >tree-C/etc/sudoers
 $OSTREE commit ${COMMIT_ARGS} -b test3-C2 -s "C" --tree=dir=tree-C
```
jlebon: Should this whole hunk be part of the previous commit where we added this test?

src/libostree/ostree-mutable-tree.c
```diff
@@ -47,10 +47,20 @@ struct OstreeMutableTree
 {
   GObject parent_instance;
 
+  /* This is the checksum of the Dirtree object that corresponds to the current
+   * contents of this directory.  contents_checksum can be NULL if the SHA was
+   * never calculated or contents of the mtree has been modified.  Even if
+   * contents_checksum is not NULL it may be out of date. */
```
jlebon: > Even if contents_checksum is not NULL it may be out of date.

Is that true though? We invalidate it before it gets out of date, no?

tests/basic-test.sh
wmanley: You're right, fixed in #1645

tests/basic-test.sh
wmanley: Fixed in #1645

src/libostree/ostree-mutable-tree.c
wmanley: No, see the comment in `ostree_mutable_tree_get_contents_checksum`:

https://github.com/ostreedev/ostree/blob/e120a6b1198aaa785533c76316898f15a804dae1/src/libostree/ostree-mutable-tree.c#L120-L135

We do invalidate it if this mtree is modified, but not if subdirectories of this mtree are modified.

In `state == LAZY` it is always accurate.

wmanley:
I've been doing some more fiddling with this and I think there is a bug in here still.  I'm trying to create a minimal testcase that reproduces the symptom.

cgwalters:
See https://github.com/ostreedev/ostree/issues/1569 - I just did a bit of digging and added a commit link there.

wmanley:
I've figured out what the bug is: `ostree_mutable_tree_get_contents_checksum` expects that one of it's subdirs' `contents_checksum` will be NULL if it's not valid.  This isn't the case for LAZY mtrees, because we're certain of the checksums.

src/libostree/ostree-mutable-tree.c
```diff
@@ -370,3 +589,27 @@ ostree_mutable_tree_new (void)
...
+OstreeMutableTree *
+ostree_mutable_tree_new_from_checksum (OstreeRepo *repo,
+                                       const char *contents_checksum,
+                                       const char *metadata_checksum)
+{
+  OstreeMutableTree* out = (OstreeMutableTree*)g_object_new (OSTREE_TYPE_MUTABLE_TREE, NULL);
+  out->state = MTREE_STATE_LAZY;
+  out->repo = g_object_ref (repo);
+  out->contents_checksum = g_strdup (contents_checksum);
+  out->metadata_checksum = g_strdup (metadata_checksum);
```
wmanley: So I need to invalidate the *parent* checksum here.  Or have a LAZY tree that (publically) has a NULL checksum.

src/libostree/ostree-mutable-tree.c
```diff
@@ -299,6 +446,74 @@ ostree_mutable_tree_ensure_parent_dirs (OstreeMutableTree  *self,
...
+      /* We're not empty - can't convert to a LAZY tree */
+      return FALSE;
+    default:
+      g_return_val_if_reached (FALSE);
+    }
+
+  self->state = MTREE_STATE_LAZY;
+  g_set_object (&self->repo, repo);
+  set_string (&self->contents_checksum, contents_checksum);
+  set_string (&self->metadata_checksum, metadata_checksum);
```
wmanley: So I need to invalidate the *parent* checksum here.  Or have a LAZY tree that (publically) has a NULL checksum.

wmanley: Yep - this was the issue. I've fixed it now.

wmanley:
Ok, tests are passing now, including the new test I wrote for the bug I found in my previous implementation.  The diff is a little more complex now that I've had to add the parent pointer, but I think it all makes sense.

src/libostree/ostree-mutable-tree.c
```diff
@@ -47,10 +47,24 @@ struct OstreeMutableTree
 {
   GObject parent_instance;
 
+  /* The parent directory to this one.  We don't hold a ref because this mtree
+   * is owned by the parent.  We can be certain that any mtree only has one
+   * parent because external users can't set this, it's only set when we create
+   * a child from within this file (see insert_child_mtree). We ensure that the
+   * parent pointer is either valid or NULL because when the parent is destroyed
+   * it sets parent = NULL on all its children (see remove_child_mtree) */
+  OstreeMutableTree *parent;
```
jlebon: This sounds like it might be a good fit for weak refs: https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html#g-object-add-weak-pointer. So then your hash table value free function can just be `g_object_unref`.

src/libostree/ostree-mutable-tree.c
```diff
@@ -90,13 +104,51 @@ ostree_mutable_tree_class_init (OstreeMutableTreeClass *klass)
...
+
+static void
+invalidate_contents_checksum (OstreeMutableTree *self)
+{
+  while (self) {
+    if (!self->contents_checksum)
+      break;
+
+    g_free (self->contents_checksum);
+    self->contents_checksum = NULL;
```
jlebon: These two can be `g_clear_pointer (&self->contents_checksum, g_free)`.

src/libostree/ostree-mutable-tree.c
```diff
@@ -303,9 +343,10 @@ ostree_mutable_tree_ensure_parent_dirs (OstreeMutableTree  *self,
       next = g_hash_table_lookup (subdir->subdirs, name);
       if (!next) 
         {
+          invalidate_contents_checksum (subdir);
```
jlebon: Might be cleaner to just put this call in `insert_child_mtree`?

src/libostree/ostree-mutable-tree.c
```diff
@@ -155,6 +155,10 @@ void
 ostree_mutable_tree_set_metadata_checksum (OstreeMutableTree *self,
                                            const char        *checksum)
 {
+  if (g_strcmp0 (checksum, self->metadata_checksum) == 0)
+    return;
+
+  invalidate_contents_checksum (self->parent);
```
jlebon: Nice fix!

src/libostree/ostree-mutable-tree.c
```diff
@@ -364,6 +478,69 @@ ostree_mutable_tree_ensure_parent_dirs (OstreeMutableTree  *self,
...
+          return FALSE;
+      }
+    case MTREE_STATE_WHOLE:
+      if (g_hash_table_size (self->files) == 0 &&
+          g_hash_table_size (self->subdirs) == 0)
+        break;
+      /* We're not empty - can't convert to a LAZY tree */
+      return FALSE;
+    default:
+      g_return_val_if_reached (FALSE);
```
jlebon: We usually just do `g_assert_not_reached ()` for these.

src/libostree/ostree-mutable-tree.c
jlebon: Yeah, I would agree with that (keeping it private for now).

src/libostree/ostree-mutable-tree.c
wmanley: I did consider this.  I figured it's not worth the overhead both in the source and at runtime.  The difference would be:

* Atomic instructions would be used for the weak refcnt.  It isn't threadsafe to use different `OstreeMutableTree` instances that are part of the same hierarchy from different threads anyway as we have no synchronisation around `contents_checksum`.  We could add more synchronisation here but it hardly seems worth it.
* The (immediate) parents would not be freed until the children had been destroyed: the weak pointer still needs to point to somewhere to check the weak refcnt (assuming it's similar to C++'s `shared_ptr` with `make_shared`)

They're minor points but I think the current implementation is nicer and we avoid a bunch of error handling for codepaths that will never be taken.  I'm not wed to the current implementation so I can change it to use weak_refs if you'd like.

cgwalters: This is just a quick comment, I haven't fully reviewed this patch yet.

Yeah, GLib's weak refs use a global mutex.  I don't know if that really matters here (probably not) - bigger perf wins would probably be having a more efficient string-tree structure.

Having non-owning pointers for trees is pretty standard I believe.

cgwalters: I just looked at Rust's BTree for example, and there the [parent pointer](https://github.com/rust-lang/rust/blob/9f79d2f86af73da483f4fe8e2422c15c79221d32/src/liballoc/btree/node.rs#L71) is a raw pointer.  It kind of has to be in Rust of course, otherwise it'd be cyclic.

wmanley:
I've rebased (and squashed) this since #1655 was merged.

src/libostree/ostree-mutable-tree.c
```diff
@@ -71,7 +84,14 @@ struct OstreeMutableTree
    * and xattrs of this directory.  This can be NULL. */
   char *metadata_checksum;
 
-  /* const char* filename -> const char* checksum */
+  /* ======== Valid for state LAZY: =========== */
+
+  /* The repo so we can look up the checksums. */
+  OstreeRepo *repo;
```
cgwalters: That's going to result in a lot of refs on the repo.  🤔(In Rust we'd just have the mutable tree have a lifetime bounded by that of the repo)

But eh, it's fine for now, if we ever profile and find this is an issue we can perhaps just have the ref be maintained by the toplevel mutable tree, and children chase up to the root.

src/libostree/ostree-mutable-tree.c
```diff
@@ -153,6 +173,83 @@ invalidate_contents_checksum (OstreeMutableTree *self)
...
+  }
+
+  g_clear_object (&self->repo);
+  self->state = MTREE_STATE_WHOLE;
+  return TRUE;
+}
+
+/* _ostree_mutable_tree_make_whole can fail if state == MTREE_STATE_LAZY, but
+ * we have getters that preceed the existence of MTREE_STATE_LAZY which can't
+ * return errors.  So instead this function will fail and print a warning. */
```
cgwalters: Hmm.  One approach I've taken in other places for things like this is to cache the error, changing the object into an "error state".  Have all API functions be "noops", and return the error at the next opportunity.

We might also be able to add a new API to check for and consume that error? 

Not a hard requirement right now though.

src/libostree/ostree-mutable-tree.c
wmanley: > That's going to result in a lot of refs on the repo.

Not too many I think.  There's only one per LAZY `OstreeMutableTree`.  This means that it's only leaf directories that have refs, which initially is only 1 directory (the root).

I agree that it's a shame to churn the repo refcount, but it's unlikely to be worth fixing.

wmanley:
> Can you add some unit tests for this patch too? Or is that too hard to do without lowering the higher level usage too? If that's the case I'd probably prefer having the changes as a single PR.

What did you have in mind?  I added some unit tests for this functionality in 36f7a6d18267814b4d11cb5a89e2ed7993b3d543 which was already merged as part of #1645.

I suppose I could write some tests in C which explicitly call `ostree_mutable_tree_new_from_checksum` and `ostree_mutable_tree_fill_empty_from_dirtree`?

cgwalters:
> What did you have in mind? I added some unit tests for this functionality in 36f7a6d which was already merged as part of #1645.

Let's back up a second here; in this scenario you mention in the commit message:

> ostree commit --tree=ref=base-rootfs --tree=ref=container1 --tree=ref=container2

Are the trees disjoint?  For example `base-rootfs` looks like a traditional system with an empty `/usr/share/containers` directory, then `container1` looks like `/usr/share/containers/container1/usr/*files here*`, `container2` looks like `/usr/share/containers/container2/usr/*files here*`, etc.

If I'm understanding the optimization here, it's really predicated on that.

Now...I'm staring at that test case code and only sort of understanding it.  It looks like it's more testing combinations of `--tree=dir` and `--tree=ref` rather than e.g. disjointness of `--tree=ref`, right?

cgwalters:
(Also I pushed a fixup for the error handling, let me know what you think)

cgwalters:
To rephrase some of the musings from #1569 - in your use case (as well as the promotion case), it feels like we basically want to make it a fatal error if we're not hitting the fast path right?

I completely get that we need to support mixing `--tree=ref` and `--tree=dir` because we initially exposed such as flexible API.  Hmm.  So in the above issue I proposed a `--from-ref-contents` which would be like `--tree=ref`, but you actually need a generalization of that.

How about something like this:

`ostree commit --disjoint-union --ref=base-os:/ --ref=container1:/usr/share/containers/container1 --ref=container2:/usr/share/containers/container2`

The args to `--ref` are `REF:SUBPATH`.

One (IMO) nicer thing about this is that the path prefix is specified at commit time, so one can have the `container1` tree just be rooted at `/` if that makes sense?  Makes it nicer to `ostree ls` them for example.

And further it would be a fatal error if the refs overlapped at all.

wmanley:
> Are the trees disjoint? For example base-rootfs looks like a traditional system with an empty `/usr/share/containers` directory, then container1 looks like `/usr/share/containers/container1/usr/*files here*`, container2 looks like `/usr/share/containers/container2/usr/*files here*`, etc.

That's right, this is what our use-case looks like.

> If I'm understanding the optimization here, it's really predicated on that.

The optimisation is more general than this and not limited to this use-case.  For example: lets say you have a rootfs and want to add dash to it.  Maybe the dash package looks like:

```
/usr/bin/dash
/usr/share/doc/dash/README
/usr/share/man/man1/dash.1.gz
/usr/share/man/man1/sh.1.gz
```

so then we do `ostree commit --tree=ref=rootfs --tree=ref=dash

With the new implementation we only need to read the following dirtrees:

```
rootfs:/
rootfs:/usr
rootfs:/usr/bin
rootfs:/usr/share
rootfs:/usr/share/doc
rootfs:/usr/share/man
rootfs:/usr/share/man/man1
dash:/
dash:/usr
dash:/usr/bin
dash:/usr/share
dash:/usr/share/doc
dash:/usr/share/man
dash:/usr/share/man/man1
```

to produce a complete merged rootfs.  That's 14 dirtree files. This is optimal taking advantage of the fact that ostree is implemented in terms of trees.  

On my laptop I have 48417 directories under `/usr`.  Thats 48417 dirtrees that wouldn't need to be read, including a load under `/usr/share/icons`, `/usr/share/texlive`, `/usr/share/perl`, all of which are irrelevant for adding a dash package to this rootfs.  So this optimisation is more general than just for disjoint refs and I think more generally useful.

    $ find /usr -type d -xdev | wc -l
    48417

There is another orthogonal way in which this implementation is faster too.  I discovered using `strace` that we were reading all the `.file` and `.dirmeta` files of each tree we looked at.  I assume this was caused by the GLib directory iteration code calling `stat` on each returned file. This implementation decodes the `dirtree` `GVariant`s itself directly so is more explicit and can be more careful about not doing more work than necessary.

> Now...I'm staring at that test case code and only sort of understanding it. It looks like it's more testing combinations of --tree=dir and --tree=ref rather than e.g. disjointness of --tree=ref, right?

The intention was to test this code by comparing the results of using `--tree=ref` with an implementation we already have confidence in - the plain `--tree=dir` approach. This implementation will used and tested by those tests.

> To rephrase some of the musings from #1569 - in your use case (as well as the promotion case), it feels like we basically want to make it a fatal error if we're not hitting the fast path right?

I don't think so.  I consider this optimisation more general than that.

> I completely get that we need to support mixing --tree=ref and --tree=dir because we initially exposed such as flexible API. Hmm. So in the above issue I proposed a --from-ref-contents which would be like --tree=ref, but you actually need a generalization of that.
>
>How about something like this:
>
>ostree commit --disjoint-union --ref=base-os:/ --ref=container1:/usr/share/containers/container1 --ref=container2:/usr/share/containers/container2
>
>The args to --ref are REF:SUBPATH.

This is similar to my implementation in #1651, but with a different API.  With #1651 this would look like:

    ostree commit --tree=ref=base-os \
        --tree=prefix=/usr/share/containers/container1 --tree=ref=container1 \
        --tree=prefix=/usr/share/containers/container2 --tree=ref=container2

I had considered the `REF:SUBPATH` API but I thought that would be a more natural API for selecting subtrees of REF rather than specifying where the ref should be "mounted".  This is the API implemented in #1658.

The other issue with the `REF:SUBPATH` API is that you can't also apply it to `--tree=dir=FILENAME:SUBPATH` because `FILENAME` may contain a `:` so you can't disambiguate.

The `--tree=prefix=` is a bit ugly.  I implemented it this way so `GOption` would preserve the order of the prefixes WRT the `--tree` options.

>--disjoint-union would appear in the API as OSTREE_REPO_COMMIT_MODIFIER_FLAGS_REF_DISJOINT_UNION or so? And it would be an error if non-ref sources were provided, any other commit modifiers were specified, etc.
>
> One (IMO) nicer thing about this is that the path prefix is specified at commit time, so one can have the container1 tree just be rooted at / if that makes sense? Makes it nicer to ostree ls them for example.

When building our trees we have a lot of intermediate trees.  So out build process looks like

1. Make container rootfs mounted at /
2. Move container rootfs to /usr/share/containers/container1
3. Merge together different images

With all my branches merged this would look like:

    container1=$(ostree commit --prefix=/usr/share/containers/container1/usr --tree=ref=container1:usr)
    container2=$(ostree commit --prefix=/usr/share/containers/container2/usr --tree=ref=container2:usr)
    final_rootfs=$(ostree commit --tree=ref=$rootfs --tree=ref=$container1 --tree=ref=$container2)

I've considered enhancing our buildsystem to do the merging and moving in one step, but it's not important if ostree is blazing fast at this anyway.

> And further it would be a fatal error if the refs overlapped at all.
>
> --from-ref-contents foo would just be sugar for --disjoint-union --ref=foo:/.
>
> And if you didn't want to change the "schema" for your existing containers refs, that'd just be a matter of using --ref=container1:/

I don't see the advantage of the fatal error as compared to just being as fast as we can be in any given situation.

cgwalters:
>  I assume this was caused by the GLib directory iteration code calling stat on each returned file.

(That's a bug/missed-optimization in `ostree-repo-file-enumerator.c` though it'd be a bit messy to fix).

> I don't see the advantage of the fatal error as compared to just being as fast as we can be in any given situation.

That makes sense to me!  OK, I'm convinced by that.  That said, I think the points I raised about not having the commit modifiers work are still valid; I have a vague worry that someone is relying on e.g. `--tree=ref=blah --owner-uid 42` or whatever.  But I checked flatpak-builder, BuildStream and they're writing filesystems.  So does rpm-ostree.   So...if we break someone's build we can revisit.  

Thanks again for this really nice patch!

@rh-atomic-bot r+ 2d41891

PR #1644, opened 2018-06-25
tests/installed: Wait a bit more for http.server

jlebon:
And also print out the output if it still didn't start up in case there
are error messages hidden in there.

This should hopefully help with diagnosing the flakes we've been seeing
in starting it up.

### DIFF
--- a/tests/installed/libinsttest.sh
+++ b/tests/installed/libinsttest.sh
@@ -58,7 +58,8 @@ run_tmp_webserver() {
     cd -
     child_pid=$!
 
-    for x in $(seq 15); do
+    for x in $(seq 60); do
+        echo "Waiting for web server ($x/60)..." >&2
         # Snapshot the output
         cp ${test_tmpdir}/httpd-output{,.tmp}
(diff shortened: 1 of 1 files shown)


smcv:
See also https://github.com/flatpak/flatpak/pull/1760

cgwalters:
> See also flatpak/flatpak#1760

Thanks for the link, I pushed a fixup! which adapts your same logic, though I only went to 60 seconds for now; if we're failing to start the webserver in a minute something's probably badly wrong. 

Though that said...maybe our tests shouldn't have any timeouts in general and we just rely on higher-level execution frameworks to do it?  Dunno.

cgwalters:
Hm, not sure why that one failed.

@rh-atomic-bot r+ e83e932

jlebon:
@rh-atomic-bot retry

cgwalters:
OK, that one looks like the VM failed to boot.

@rh-atomic-bot retry

cgwalters:
Merged in https://github.com/ostreedev/ostree/pull/1652

PR #1646, opened 2018-06-26
man/ostree.repo-config: Document collection-id

mwleeds:
The collection-id option in the core section was recently made public
but not documented.


cgwalters:
@rh-atomic-bot r+ cdec90b

PR #1647, opened 2018-06-26
lib/repo-commit: Abort a transaction if preparing it fails

pwithnall:
If ostree_repo_prepare_transaction() fails, we should reset the
repository’s state so that the failed call was essentially idempotent.
Do that by calling ostree_repo_abort_transaction() on the failure path.

Typically, the way for preparing a transaction to fail is for its
GCancellable to be triggered, rather than because any of the operations
involved in preparing a transaction are particularly failure prone.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1614,8 +1614,14 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
                                  GCancellable   *cancellable,
                                  GError        **error)
 {
+  g_autoptr(_OstreeRepoAutoTransaction) txn = NULL;
 
   g_return_val_if_fail (self->in_transaction == FALSE, FALSE);
+
+  g_debug ("Preparing transaction in repository %p", self);
+
+  /* Set up to abort the transaction if we return early from this function. */
+  txn = self;
 
   memset (&self->txn.stats, 0, sizeof (OstreeRepoTransactionStats));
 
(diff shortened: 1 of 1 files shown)


pwithnall:
The [test failure](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/02b2f20849d441e4b4ff72c9a79097346407c9ef.1530026445736796672/output.log) seems to be unrelated:
```
TASK [Load status json] ********************************************************
fatal: [localhost]: FAILED! => {"changed": false, "cmd": "rpm-ostree status --json", "msg": "[Errno 2] No such file or directory", "rc": 2}
	to retry, use: --limit @/var/tmp/checkout/tests/installed/nondestructive.retry
```

cgwalters:
> The test failure seems to be unrelated:

Yep, just fixed that.  Hmm, wouldn't it work to just create an `g_autoptr(_OstreeRepoAutoTransaction) txn` here, and set it to NULL at the end of the success path?

pwithnall:
> Hmm, wouldn't it work to just create an g_autoptr(_OstreeRepoAutoTransaction) txn here, and set it to NULL at the end of the success path?

Ah, good suggestion!

cgwalters:
I don't have notifications when one pushes to a PR, so please do add an "Updated" or so comment when you update the PR.  Thanks for the patch!

@rh-atomic-bot r+ ed82835

cgwalters:
@rh-atomic-bot retry

PR #1651, opened 2018-06-27
RFC: ostree commit: Add option `--tree=prefix=PREFIX`

wmanley:
This enhances `ostree commit`'s ability to compose trees.  You can now run

    ostree commit -b rootfs \
        --tree=ref=rootfs \
        --tree=prefix=/usr/lib/machines/container --tree=ref=container

and that will "mount" the container ref under usr/lib/machines/container
in rootfs.  It is much faster than checking the tree out and checking it
back in again.

This is useful for our buildsystem where we ship containers in the rootfs image.

I'm still not sure of the UI.  I'm hoping this PR will spur on discussion.

Alternatives
=========

* Have a single switch that affects all `--tree` options.  The above incantation would become:

        ostree commit -b rootfs \
            --tree=ref=$(ostree commit --orphan --prefix=/usr/lib/machines/container --tree=ref=container) \
            --tree=ref=rootfs

TODO:
=====

- [ ] Don't hard code `dirmeta_0755_0_0`, write it explicitly instead

### DIFF
--- a/man/ostree-commit.xml
+++ b/man/ostree-commit.xml
@@ -99,10 +99,10 @@ Boston, MA 02111-1307, USA.
             </varlistentry>
 
             <varlistentry>
-                <term><option>--tree</option>="dir=PATH" or "tar=TARFILE" or "ref=COMMIT"</term>
+                <term><option>--tree</option>="dir=PATH" or "tar=TARFILE" or "ref=COMMIT" or "prefix=PATH"</term>
 
                 <listitem><para>
-                    Overlay the given argument as a tree.
+                    Overlay the given argument as a tree.  prefix is a little different - --tree=prefix=PATH affects where in the tree the later --tree=ref/tar/dir are merged.
                 </para></listitem>
             </varlistentry>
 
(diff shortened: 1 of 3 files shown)


jlebon:
Hmm, a similar idea I've had was to copy the features of `checkout --union/--union-add/--union-identical` to `commit`. Then you could do e.g.

```
ostree commit --union --tree=ref=rootfs --tree=dir=overlay-rootfs
```

wmanley:
> Hmm, a similar idea I've had was to copy the features of checkout --union/--union-add/--union-identical to commit

On that subject: I started implementing support for overlaying upper directories from overlayfs as part of `ostree commit`:
https://github.com/stb-tester/ostree/commit/34dce7d6337c0689e9e56b714d756764dae9aabd

I stopped when I discovered that you need CAP_SYS_ADMIN to read the xattrs that overlayfs writes.  I'm taking a different approach to make that fast.

jlebon:
Hmm OK, one thing I missed at first is that `--tree=prefix` affects every `--tree` afterwards, not just the next one, until the next `--tree=prefix`. Which is fine I suppose, though it's not clear to me whether the common case would be passing multiple tree at the same prefix, or different ones. In the latter case, it might be cleaner to come up with a syntax that unifies them into a single arg, e.g. something like:

```
ostree commit --tree=ref=foobar --subtree=/usr/lib/mysubtree:ref=baz
```

?

That way they're together so it's easier to tell and harder to get wrong/break during a refactor (and if someone really wants to target a subtree that has `:`, they should just use the API :)).

> Hmm, a similar idea I've had was to copy the features of `checkout --union/--union-add/--union-identical` to `commit`.

Thinking more about this, I think they're mostly orthogonal and could come after if the need arises. Right now, `commit` args have `--union` semantics already; existing files in the mutable tree are overwritten. So my thought was to complement that with a `--union-add`, `--union-identical`, and a e.g. `--no-union` mode, where it errors out if trying to write a file that already exists (the equivalent of the default `checkout` mode).

jlebon:
bot, retest this please

PR #1654, opened 2018-06-27
deploy: Change BootLoaderSpec filenames so they can be used for sorting

martinezjavier:
Currently the BLS snippets are named ostree-$ID-$VARIANT_ID-$index.conf,
but the BLS config files are actually sorted by using the version field
which is the inverse of the index.

In most places, _ostree_sysroot_read_boot_loader_configs() is used to
get the BLS files and this function already returns them sorted by the
version field. The only place where the index trailing number is used is
in the ostree-grub-generator script that lists the BLS files to populate
the grub config file.

But for some bootloaders the BLS filename is the criteria for sorting by
taking the filename as a string version. So on these bootloaders the BLS
entries will be listed in the reverse order.

To avoid that, change the BLS snippets filename to have the version field
instead of the index and also to have the version before deployment name.

Make the filenames to be of the form ostree-$version-$ID-$VARIANT_ID.conf
so the version is before the deployment name.

Signed-off-by: Javier Martinez Canillas <javierm@redhat.com>

### DIFF
--- a/src/boot/grub2/ostree-grub-generator
+++ b/src/boot/grub2/ostree-grub-generator
@@ -66,7 +66,7 @@ populate_menu()
     else
         boot_prefix="${OSTREE_BOOT_PARTITION}"
     fi
-    for config in $(ls -v $entries_path/*.conf); do
+    for config in $(ls -v -r $entries_path/*.conf); do
         read_config ${config}
         menu="${menu}menuentry '${title}' {\n"
         menu="${menu}\t linux ${boot_prefix}${linux} ${options}\n"
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1628,8 +1628,9 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
   g_assert_cmpstr (kernel_layout->bootcsum, ==, bootcsum);
   g_autofree char *bootcsumdir = g_strdup_printf ("ostree/%s-%s", osname, bootcsum);
   g_autofree char *bootconfdir = g_strdup_printf ("loader.%d/entries", new_bootversion);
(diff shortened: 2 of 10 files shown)


martinezjavier:
It seems the test failures are unrelated?

Since I see that in the [required](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/d86570df1ed6f6666af8e09998eb4534780377ae.1530118048012025344/index.html) test suite the [FAH28-insttests](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/d86570df1ed6f6666af8e09998eb4534780377ae.0.1530116631830993808/index.html) failed but this succeeded for [ex-required ](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/d86570df1ed6f6666af8e09998eb4534780377ae.1530117394571930368/index.html).

 And conversely [f28-introspection-tests](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/d86570df1ed6f6666af8e09998eb4534780377ae.1530116660010696448/output.log) failed for ex-required which succeeded in the required test suite.

cgwalters:
That test is flaky sadly...I just don't want to drop it, as at some point I really want to expand our tests to work that way by default.  There's some stuff we're doing in the background to enhance reliability here.
If it fails again in the merge queue we'll look then.

Anyways I thought about this a fair bit, I can't see a way it could break anything - the way we do an atomic swap of all the bootloader configs makes this transition easy, and older versions of libostree will keep using the version field.  The only danger I can think of is if somehow `15_ostree` gets out of sync with the library - but I don't think that will happen in e.g. Anaconda today.

@rh-atomic-bot r+ d86570d

martinezjavier:
@cgwalters agreed and hopefully we should eventually get rid of `15_ostree` too.

PR #1655, opened 2018-06-27
OstreeMutableTree: Add parent pointer and fix bug

wmanley:
This is preparatory work for #1643.  The commits were originally on #1643 and I've added a fixup to address some of the comments from there.  I split this out as it's own PR as requested by @jlebon.

This implements a TODO item from
`ostree_mutable_tree_get_contents_checksum`.  We now no-longer invalidate
the dirtree contents checksum at `get_contents_checksum` time - we
invalidate it when the mtree is modified.  This is implemented by keeping
a pointer to the parent directory in each `OstreeMutableTree`.  This gives
us stronger invariants on `contents_checksum`.

For even stronger guarantees about invariants we could make
`ostree_repo_write_mtree` or similar a member of `OstreeMutableTree` and
remove `ostree_mutable_tree_set_metadata_checksum`.

I've fixed a bug here too.  We now invalidate parent's contents
checksum when our metadata checksum changes, whereas we didn't before.

### DIFF
--- a/src/libostree/ostree-mutable-tree.c
+++ b/src/libostree/ostree-mutable-tree.c
@@ -155,6 +155,10 @@ void
 ostree_mutable_tree_set_metadata_checksum (OstreeMutableTree *self,
                                            const char        *checksum)
 {
+  if (g_strcmp0 (checksum, self->metadata_checksum) == 0)
+    return;
+
+  invalidate_contents_checksum (self->parent);
   g_free (self->metadata_checksum);
   self->metadata_checksum = g_strdup (checksum);
 }
--- a/src/libostree/ostree-mutable-tree.c
+++ b/src/libostree/ostree-mutable-tree.c
@@ -47,10 +47,24 @@ struct OstreeMutableTree
 {
   GObject parent_instance;
 
+  /* The parent directory to this one.  We don't hold a ref because this mtree
+   * is owned by the parent.  We can be certain that any mtree only has one
+   * parent because external users can't set this, it's only set when we create
+   * a child from within this file (see insert_child_mtree). We ensure that the
+   * parent pointer is either valid or NULL because when the parent is destroyed
+   * it sets parent = NULL on all its children (see remove_child_mtree) */
+  OstreeMutableTree *parent;
+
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-mutable-tree.c
```diff
@@ -146,7 +149,7 @@ invalidate_contents_checksum (OstreeMutableTree *self)
       break;
 
     g_free (self->contents_checksum);
```
jlebon: You can drop this line too, that's the beauty of `g_clear_pointer` :)

tests/test-mutable-tree.c
```diff
@@ -39,6 +40,27 @@ test_metadata_checksum (void)
   ostree_mutable_tree_set_metadata_checksum (tree, checksum);
 
   g_assert_cmpstr (checksum, ==, ostree_mutable_tree_get_metadata_checksum (tree));
+
+  /* If a child tree's metadata changes the parent tree's contents needs to be
+   * recalculated */
+  glnx_unref_object OstreeMutableTree *subdir = NULL;
+  g_assert (ostree_mutable_tree_ensure_dir (tree, "subdir", &subdir, &error));
```
jlebon: If we're doing a `g_assert` and we're not gonna check that `error` is still `NULL`, we might as well pass `NULL` here and drop the local variable.

wmanley: Fixed

src/libostree/ostree-mutable-tree.c
wmanley: Fixed

cgwalters:
Awesome, thanks a lot for this!

@rh-atomic-bot r+ a0d0fee

jlebon:
@rh-atomic-bot retry

jlebon:
OK right, this is failing because of #1662 and... not sure what happened with the `FAH28-insttests`, looks like kernel issues? We'll just try again once that PR is merged.

cgwalters:
We can re-queue this one now assuming the other one lands.

@rh-atomic-bot retry

PR #1656, opened 2018-06-28
create-usb: Tweak docs for --destination-repo

mwleeds:
Make it show up in the help output as --destination-repo=DEST so it's
clear that it takes an argument.


jlebon:
Thanks! @rh-atomic-bot r+ e22c730

PR #1657, opened 2018-06-28
tests: Save corefiles back to tests/ directory if one exists

wmanley:
Makes it easier to debug failures from the tests.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -34,6 +34,14 @@ else
 fi
 . ${test_srcdir}/libtest-core.sh
 
+save_core() {
+  if [ -e core ]; then
+    cp core "$test_srcdir/core"
+  fi
+}
+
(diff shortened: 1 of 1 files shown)


cgwalters:
This is fine, though note the coredump semantics depend on the host.  It's common for example to have `systemd-coredump` enabled and that will even grab cores out of containers, etc.

@rh-atomic-bot r+ 45b7e78

jlebon:
One caveat here is that this will be ignored for tests that redefine `trap`s for the same signal:

```
tests/test-repo-finder-mount-integration.sh:58:        trap _mount_cleanup EXIT
tests/test-rofiles-fuse.sh:44:trap cleanup_fuse EXIT
```

Still useful of course for the great majority of other tests in environments without a modified `core_pattern` (which all distros with e.g. `systemd-coredump` will change).

PR #1658, opened 2018-06-28
Enhance command line utils to take a tree/file sha

wmanley:
...in place of a commit sha where reasonable.

The motivation for this is to enhance `ostree commit --tree=ref` to allow composing subtrees.  For example: in our build system we strip away everything not under `/usr` as the last step when building containers.  This PR (in combination with #1651) allows you to do:

    ostree commit --tree=prefix=/usr/lib/machines/container --tree=ref=:container:usr

To do this.  It's fast.

As I was going along I thought it would be useful to have this ability generally in the command line tools. So now many of them accept a treeish instead of a commit.

Previously these tools accepted a COMMIT as:

* REF
* REMOTE:REF
* COMMIT_SHA

And now in addition:

* REMOTE:REF:PATH
* :REF:PATH
* COMMIT_SHA:PATH
* TREE_SHA:PATH

So the idea is that you append `:PATH` to a commit and you can select a tree.  This is the same as the git syntax.  As a ref may already contain a colon if the REMOTE is specified we disambiguate by requiring the remote to be specified, even if it is left empty.


cgwalters:
Offhand the idea is cool, and I love to help push forward the container+OS as single tree concept and make that work better.

It's not a private API if we're adding it to the symbol table :smile:.  We did add a special way for the CLI to call into private library bits, see `ostree-cmdprivate.[ch]`.  That's probably a good approach until we decide to make the API stable?

In just glancing at the code: There are various leftover debug prints, and can you fix the space-between-identifer-and-parent i.e. `g_strcmp0 (` vs  `g_strcmp0(`.  

Thanks again for working on this!  Really looking forward to getting more of your patches in.

wmanley:
> In just glancing at the code: There are various leftover debug prints, and can you fix the space-between-identifer-and-parent i.e. g_strcmp0 ( vs g_strcmp0(.

Sorry, that was sloppy.  I was in a rush to finish it the week before last as I suspected that I wouldn't  have time last week.

I've fixed the formatting issues - at least the ones I spotted.  I had a crack at configuring GNU indent and clang-format to format according to ostree style, but failed on both.  I couldn't get either to align the `*` in function definitions correctly. I think [https://reviews.llvm.org/D27651](D27651) is the clang bug.

I've changed the approach to private API as recommended.  We use `ostree-cmdprivate.h` now.

I've also broken up the main commit into 3 to make it easier to review.  CI passes too.

Please review.

wmanley:
CI isn't passing, but I don't think it's related:

[f28-rust](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/cdfb9e8643eedf4ff8ce958c55a80e4583ad4b3a.1531156574878986240/output.log) says:

> Error: Error downloading packages:
>  Status code: 503 for https://mirrors.fedoraproject.org/metalink?repo=updates-released-f28&arch=x86_64

[FAH28-insttests](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/cdfb9e8643eedf4ff8ce958c55a80e4583ad4b3a.1531156509673376000/output.log) says:

> Could not resolve host: kojipkgs.fedoraproject.org

src/libostree/ostree-repo.c
```diff
@@ -4379,6 +4379,171 @@ ostree_repo_read_commit (OstreeRepo   *self,
...
+      if (!ret_out)
+        {
+          g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
+                       "File \"%s\" not found", path);
+          return FALSE;
+        }
+    }
+
+  if (g_strcmp0 (ostree_repo_file_tree_get_metadata_checksum (ret_out),
+                 "00000000000000000000000000000000") == 0)
```
cgwalters: Hmm...we don't use the all-zero sha256 bit pattern anywhere do we?  The empty tree metadata checksum looks like it's `6e340b9cffb37a989ca544e6bb780a2c78901d3fb33738768511a30617afa01d` (from `/usr/share/empty`) which is itself different from the empty sha256:

```
$ sha256sum </dev/null
e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855  -
$
```

Am I misunderstanding something?  How can this case be hit?

src/libostree/ostree-repo.c
```diff
@@ -4379,6 +4379,171 @@ ostree_repo_read_commit (OstreeRepo   *self,
...
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_INVALID_ARGUMENT,
+                   "No subdirectory specified for tree %s", sha);
+      return FALSE;
+    }
+
+  GFileType type = g_file_query_file_type (
+      (GFile*)ret_out, G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS, cancellable);
```
cgwalters: I think it'd be easier to use `g_file_query_info()` here - that way we just can just throw (return error) if we hit `ENOENT` etc.

src/libostree/ostree-repo.c
```diff
@@ -4379,6 +4379,171 @@ ostree_repo_read_commit (OstreeRepo   *self,
...
+                               &has_object, cancellable, error))
+    return FALSE;
+
+  if (has_object)
+    {
+      g_autoptr(GFile) root_gfile = NULL;
+      if (!ostree_repo_read_commit (self, sha, &root_gfile, NULL, cancellable,
+                                    error))
+        return FALSE;
+      g_autoptr(OstreeRepoFile) root = OSTREE_REPO_FILE (root_gfile);
```
cgwalters: This is relatively minor and it's mostly OK as is, I think I'd prefer though:

```
g_assert (OSTREE_IS_REPO_FILE (root_gfile));
OstreeRepoFile *root = (OstreeRepoFile*) root_gfile;
```

src/libostree/ostree-repo.c
```diff
@@ -4379,6 +4379,171 @@ ostree_repo_read_commit (OstreeRepo   *self,
...
+ * This is a helper for the command line utilities.  It's used to convert a
+ * reference to a tree or a file from the user into an OstreeRepoFile.  It takes
+ * care to provide good error messages.
+ *
+ * @ref can take the form:
+ *
+ *    REMOTE:REF
+ *    :REF
+ *    REF
+ *    COMMIT_SHA
```
cgwalters: This is cool!

src/libostree/ostree-repo.c
```diff
@@ -4379,6 +4379,261 @@ ostree_repo_read_commit (OstreeRepo   *self,
...
+  if (!ostree_repo_has_object (self, OSTREE_OBJECT_TYPE_DIR_TREE, sha,
+                               &has_object, cancellable, error))
+    return FALSE;
+  if (has_object)
+    {
+      /* We're a bit cheeky here - we need to specify a metadata checksum but we
+       * don't have one.  We're getting a subpath here anyway we're not going to
+       * return this invalid file. */
+      g_autoptr(OstreeRepoFile) root =  _ostree_repo_file_new_root (
+          self, sha, "00000000000000000000000000000000");
```
wmanley: This is where I set the "00000000000000000000000000000000" SHA.  I guess the reference above deserves a comment.

src/libostree/ostree-repo.c
wmanley: I set it explicitly below in `resolve_sha_to_tree`.

wmanley:
There's a bug in ostree diff introduced here.  Running:

    ostree --repo=/ostree ls :stbt-node-rootfs/b51ca22876b3b7ff776870bca1e8c1ff92e6bcbb:usr/bin

fails with

    error: Bad ref ":stbt-node-rootfs/0b791b5379ea56d111307f112b2bce0cc5a2d732:usr/bin^": "bin^": No such file or directory

presumably because ostree diff by default adds `^` to the ref to get the previous ref.  I'll need to add a test and fix this.

PR #1660, opened 2018-06-28
sysroot: Reject attempts to pin the staged deployment

cgwalters:
From https://github.com/projectatomic/rpm-ostree/pull/1434#discussion_r198936674

To support it we'd have to actually write it to disk, which...let's
not try that right now.


jlebon:
@rh-atomic-bot r+ 73a72b8

PR #1661, opened 2018-06-29
ci: exclude 'lib' from libsoup configure option

marcusfolkesson:
The option used by configure script is actually
--with-soup/--without-soup.


jlebon:
Good catch, thanks!
@rh-atomic-bot r+ e99087b

PR #1662, opened 2018-06-29
ci/flatpak: Fix to use built ostree version

cgwalters:
Noticed as part of a random failure in this PR:
https://github.com/ostreedev/ostree/pull/1655
that we weren't actually testing the version of ostree
we built in git.  Probably we should be generating RPMs but...later.


jlebon:
I pushed a fixup to drop the `pkg_install_testing`; we're overwriting it anyway, so it doesn't matter if it's from the stable repos.

@rh-atomic-bot r+ c844e41

jlebon:
@rh-atomic-bot r+ 88c89bd

jlebon:
@rh-atomic-bot retry

PR #1665, opened 2018-07-01
boot: Use emergency.target, not emergency.service

akiernan:
Follow systemd units in using emergency.target, not emergency.service
(which is the sole unit, by default, in emergency.target) so we can
easily reconfigure the units which are actived when entering
emergency mode.


cgwalters:
bot, add author to whitelist

cgwalters:
Thanks!

@rh-atomic-bot r+ e9788fd

jlebon:
@rh-atomic-bot retry

Issue #1667, opened 2018-07-03
Separate /var broke in 2018.6


cgwalters:
Regression from https://github.com/ostreedev/ostree/pull/1617

We need an insttest covering this.

cgwalters:
Trivial reproducer setup for a separate `/var` in a testing VM:
```
cp -a /var /sysroot/myvar
echo '/sysroot/myvar /var none bind 0 0' >> /etc/fstab
```

(And if you do e.g. `rpm-ostree override replace` to test, don't forget to `rpm-ostree initramfs --enable` like I did)

cgwalters:
Dropping this systemd unit in `/etc/systemd/system` should work around the issue:

```
$ cat > /etc/systemd/system/workaround-ostree-issue-1667.service << EOF
[Unit]
DefaultDependencies=no
Before=ostree-remount.service

[Service]
ExecStart=/bin/sh -c 'if findmnt -s -n -T /var; then echo Enabling workaround; umount /var; systemctl start --no-block var.mount; fi'

[Install]
WantedBy=multi-user.target
EOF
$ systemctl enable workaround-ostree-issue-1667.service
```

nexfwall:
@cgwalters, thank you. Will try it in the evening, then I come back home.

akiernan:
That workaround is fixing the separate `/var` problem for me. The change in #1668 isn't doing it for me though, but given I'm newly trying to use a separate `/var` I'm not sure I'm seeing the original problem.

jlebon:
@akiernan When testing #1668, make sure to also `rpm-ostree initramfs --enable` since the fix happens in the initramfs.

jlebon:
https://bodhi.fedoraproject.org/updates/ostree-2018.6-4.fc28

akiernan:
@jlebon Thanks - it wasn't that, but I think I've found my problem, just testing it and I'll open a new ticket for it.

nexfwall:
[Workaround](https://github.com/ostreedev/ostree/issues/1667#issuecomment-402308008) isn't worked for me. Because of `/var/home` mount, I think.

```
$ systemctl status workaround-ostree-issue-1667.service
● workaround-ostree-issue-1667.service
   Loaded: loaded (/etc/systemd/system/workaround-ostree-issue-1667.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Wed 2018-07-04 22:09:05 MSK; 3min 14s ago
[...]

PR #1668, opened 2018-07-03
switchroot: Fix regression for separately mounted /var

jlebon:
I made a logical error in #1617 which resulted in the exact *opposite*
behaviour we want when `/var` is a separate mount.

Split this out and lower the number of negations to make it more obvious
that it's correct.

Closes: #1667

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -151,13 +151,16 @@ main(int argc, char *argv[])
   if (chdir (deploy_path) < 0)
     err (EXIT_FAILURE, "failed to chdir to deploy_path");
 
+  /* Default to true, but in the systemd case, default to false because it's handled by
+   * ostree-system-generator. */
   bool mount_var = true;
-  /* In the systemd case, this is handled by ostree-system-generator by default */
-#ifndef HAVE_SYSTEMD_AND_LIBMOUNT
-  /* file in /run can override that behaviour */
-  if (lstat (INITRAMFS_MOUNT_VAR, &stbuf) < 0)
-    mount_var = false;
+#ifdef HAVE_SYSTEMD_AND_LIBMOUNT
+  mount_var = false;
 #endif
+
+  /* file in /run can override the default behaviour so that we definitely mount /var */
+  if (lstat (INITRAMFS_MOUNT_VAR, &stbuf) == 0)
+    mount_var = true;
 
   /* Link to the deployment's /var */
   if (mount_var && mount ("../../var", "var", NULL, MS_MGC_VAL|MS_BIND, NULL) < 0)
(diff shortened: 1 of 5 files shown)


jlebon:
Marking as WIP. I'll sanity check this tomorrow for both the separate `/var` point case and integrated case.

cgwalters:
Test case I was playing with, but I'm officially on vacation today:

```
diff --git a/tests/installed/destructive-ansible.yml b/tests/installed/destructive-ansible.yml
index c72f6b82..3c467f7b 100644
--- a/tests/installed/destructive-ansible.yml
+++ b/tests/installed/destructive-ansible.yml
@@ -12,4 +12,5 @@
     - import_tasks: tasks/install-git.yml
       when: use_git_build
     - import_tasks: tasks/query-host.yml
+    - import_tasks: destructive/var-mount.yml
     - import_tasks: destructive/staged-deploy.yml
diff --git a/tests/installed/destructive/var-mount.yml b/tests/installed/destructive/var-mount.yml
new file mode 100644
index 00000000..7bb0ae85
--- /dev/null
+++ b/tests/installed/destructive/var-mount.yml
@@ -0,0 +1,13 @@
+# https://github.com/ostreedev/ostree/issues/1667
+- name: Set up /var as a mountpoint
+  shell: |
+    set -xeuo pipefail
+    cp -a /var /sysroot/myvar
+    touch /sysroot/myvar/somenewfile
+    echo '/sysroot/myvar /var none bind 0 0' >> /etc/fstab
+- include_tasks: ../tasks/reboot.yml
+- name: Check that deploy-staged service worked
+  shell: |
+    set -xeuo pipefail
+    systemctl status var.mount
+    test -f /var/somenewfile
```

Mind trying this out and verifying it fails without your patch too?

cgwalters:
r=me if this works!

jlebon:
Sigh, spent a while banging my head before I realized I was rsync'ing the wrong ostree bits. :twisted_rightwards_arrows: 

I've verified that:

- this works with no `/var` mountpoint
- this works with `/var` mountpoints
- test case fails without this patch
- test case succeeds with this patch

Thanks for the testcase! One tweak I had to make was:

```diff
diff --git a/tests/installed/tasks/install-git.yml b/tests/installed/tasks/install-git.yml
index 8216afeb..3374cce7 100644
--- a/tests/installed/tasks/install-git.yml
+++ b/tests/installed/tasks/install-git.yml
@@ -8,6 +8,9 @@
   synchronize: src=build/x86_64/ dest=/root/x86_64/ archive=yes
 - name: Install RPMs
   shell: rpm-ostree override replace /root/x86_64/*.rpm
+# Regenerate to make sure new ostree binaries also make it to the initrd
+- name: Regenerate initramfs
+  shell: rpm-ostree initramfs --enable
 - import_tasks: ../tasks/reboot.yml
 - import_tasks: ../tasks/query-host.yml
 - command: ostree --version
```

(I also tweaked the task name in the test).

@rh-atomic-bot r=cgwalters 4370722

PR #1669, opened 2018-07-03
ostree/pull: Add network-retries command line option

### LINKED ISSUE #1659: Add --network-retries=N to ostree pull
PR https://github.com/ostreedev/ostree/pull/1594 added the ability to retry network operations. Expose this new API option on the `ostree pull` command-line with a new switch like `--network-retries=N`.

rfairley:
This exposes a way to specify from the command line the number
of times to retry each download after a network error. If a negative
value is given, then the default number of retries (5) is used. If 0
is given, then errors are returned without retrying.

closes #1659

**Notes**
To-do:
- need to update bash-completion script and man page (will remove the "ME TODO" in the file once done)
- working on adding a test for this

Alternatively instead of initializing to -1, a `G_OPTION_ARG_CALLBACK` function could be used to 1) set a boolean variable to indicate that the `network-retries` argument was given and 2) set the `opt_network_retries` variable, and use the boolean variable to decide whether the option is added to the Variant. However, this may be complicating things.

### DIFF
--- a/tests/test-pull-repeated.sh
+++ b/tests/test-pull-repeated.sh
@@ -23,7 +23,7 @@ set -euo pipefail
 
 . $(dirname $0)/libtest.sh
 
-echo "1..2"
+echo "1..4"
 
 COMMIT_SIGN="--gpg-homedir=${TEST_GPG_KEYHOME} --gpg-sign=${TEST_GPG_KEYID_1}"
 
@@ -47,7 +47,20 @@ ${CMD_PREFIX} ostree --repo=repo rev-parse main
 popd
 echo "ok repeated pull after 500s"
 
-# Now test from a repo which gives error 408 (request timeout) a lot of the time.
+# Sanity check with no network retries and 408s given, pull should fail.
+rm ostree-srv httpd repo -rf
+setup_fake_remote_repo1 "archive" "${COMMIT_SIGN}" --random-408s=99
+
+pushd ${test_tmpdir}
+ostree_repo_init repo --mode=archive
+${CMD_PREFIX} ostree --repo=repo remote add --set=gpg-verify=false origin $(cat httpd-address)/ostree/gnomerepo
+assert_fail ${CMD_PREFIX} ostree --repo=repo pull --mirror origin --network-retries=0 main 2>err.txt
+assert_file_has_content err.txt "\(408.*Request Timeout\)\|\(HTTP 408\)"
+
+popd
+echo "ok no retries after a 408"
+
+# Test pulling a repo which gives error 408 (request timeout) a lot of the time.
 rm ostree-srv httpd repo -rf
 setup_fake_remote_repo1 "archive" "${COMMIT_SIGN}" --random-408s=50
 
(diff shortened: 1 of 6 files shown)


jlebon:
Looks good so far! :+1: 

For tests, check out `test-pull-repeated.sh`. You can look at the original PR #1594 for the test case it added there to get familiar with it.

rfairley:
Thanks, taking a look at this now!

rfairley:
Just added the following:
- tests for `--network-retries` in `test-pull-repeated.sh`
- bash completion and `man` page updates
- moved `assert_fail` bash function from `test_remote_headers.sh` to `libtest.sh` (thinking that function could be used generally, please let me know if this should be kept in the original file)

tests/test-pull-repeated.sh
```diff
@@ -23,9 +23,10 @@ set -euo pipefail
 
 . $(dirname $0)/libtest.sh
 
-echo "1..2"
+echo "1..4"
 
 COMMIT_SIGN="--gpg-homedir=${TEST_GPG_KEYHOME} --gpg-sign=${TEST_GPG_KEYID_1}"
+BIG_NUMBER_RETRIES=8 # Gives error rate <0.5%, with --random-408s=50
```
jlebon: Minor: we don't have to make it a constant upfront. Let's just inline it where it's used? (Though the comment is definitely useful! Let's move that down there too!)

jlebon:
BTW, if you're not familiar with `git rebase -i`, it's definitely worth learning! Check out https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History.

tests/test-pull-repeated.sh
rfairley: Makes sense, making this edit now. Thanks!
One commit plus a prep commit sounds good! Thank you for the link on rebasing, looking over this.

rfairley:
Edits completed! Please let me know if there is anything else.

tests/test-pull-repeated.sh
```diff
@@ -67,3 +80,22 @@ ${CMD_PREFIX} ostree --repo=repo rev-parse main
...
+setup_fake_remote_repo1 "archive" "${COMMIT_SIGN}" --random-408s=50
+
+pushd ${test_tmpdir}
+ostree_repo_init repo --mode=archive
+${CMD_PREFIX} ostree --repo=repo remote add --set=gpg-verify=false origin $(cat httpd-address)/ostree/gnomerepo
+
+# Using 8 network retries gives error rate of <0.5%, when --random-408s=50
+if ${CMD_PREFIX} ostree --repo=repo pull --mirror origin --network-retries=8 main 2>err.txt; then
+   echo "Success with big number of network retries"
+fi
```
jlebon: Minor: usually for tests we don't expect to fail, we just run it directly instead of wrapping it around an if statement. Because we're running with `set -e`, any command that fails will abort the test. See: http://redsymbol.net/articles/unofficial-bash-strict-mode/.

Not worth a respin, just FYI :)

jlebon:
Thanks for working on this!
@rh-atomic-bot r+ d01badc

tests/test-pull-repeated.sh
jlebon: Actually, thinking on this again, wrapping it around an if statement means that we'll silently continue even if it fails. So we do need to change this.

Want to post a follow-up PR once this one is merged? :)

rfairley: Ah I see - previously I tried giving `--network-retries=0` to make it fail, and it did stop and fail, but for the wrong reason: it stopped at `${CMD_PREFIX} ostree --repo=repo rev-parse main` because nothing had been pulled in. Just tried running it directly now, it stopped at the pull command as it should. Will post a follow-up PR.

Thanks for pointing that out! Will read over the link.

PR #1671, opened 2018-07-05
Refactor min-free-space-size error reporting

uajain:
Nitpick fixes .

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -429,6 +429,28 @@ add_size_index_to_metadata (OstreeRepo        *self,
   return g_variant_ref_sink (g_variant_builder_end (builder));
 }
 
+static gboolean
+throw_min_free_space_error (OstreeRepo  *self,
+                            guint64      bytes_required,
+                            GError     **error)
+{
+  const char *err_msg = NULL;
+  g_autofree char *err_msg_owned = NULL;
+
+  if (bytes_required > 0)
+    {
+      g_autofree char *formatted_required = g_format_size (bytes_required);
+      err_msg = err_msg_owned = g_strdup_printf ("would be exceeded, at least %s requested", formatted_required);
+    }
+  else
+    err_msg  = "would be exceeded.";
+
+  if (self->min_free_space_percent > 0)
+    return glnx_throw (error, "min-free-space-percent '%u%%' %s", self->min_free_space_percent, err_msg);
+  else
+    return glnx_throw (error, "min-free-space-size %" G_GUINT64_FORMAT "MB %s", self->min_free_space_mb, err_msg);
+}
+
 typedef struct {
   gboolean initialized;
   GLnxTmpfile tmpf;
@@ -524,13 +546,7 @@ _ostree_repo_bare_content_commit (OstreeRepo                 *self,
         {
           self->cleanup_stagedir = TRUE;
           g_mutex_unlock (&self->txn_lock);
-          g_autofree char *formatted_required = g_format_size (st_buf.st_size);
-          if (self->min_free_space_percent > 0)
-           return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
-                              self->min_free_space_percent, formatted_required);
-          else
-            return glnx_throw (error, "min-free-space-size %luMB woulid be exceeded, %s more required",
-                               self->min_free_space_mb, formatted_required);
+          return throw_min_free_space_error (self, st_buf.st_size, error);
         }
       /* This is the main bit that needs mutex protection */
       self->txn.max_blocks -= object_blocks;
@@ -927,13 +943,7 @@ write_content_object (OstreeRepo         *self,
           guint64 bytes_required = (guint64)object_blocks * self->txn.blocksize;
           self->cleanup_stagedir = TRUE;
           g_mutex_unlock (&self->txn_lock);
-          g_autofree char *formatted_required = g_format_size (bytes_required);
-          if (self->min_free_space_percent > 0)
-            return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s more required",
-                               self->min_free_space_percent, formatted_required);
-          else
-            return glnx_throw (error, "min-free-space-size %" G_GUINT64_FORMAT "MB would be exceeded, %s more required",
-                               self->min_free_space_mb, formatted_required);
+          return throw_min_free_space_error (self, bytes_required, error);
         }
       /* This is the main bit that needs mutex protection */
       self->txn.max_blocks -= object_blocks;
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo.c
```diff
@@ -2693,8 +2693,10 @@ min_free_space_size_validate_and_convert (OstreeRepo    *self,
 
   guint64 min_free_space = g_ascii_strtoull (size_str, NULL, 10);
   if (shifts > 0 && g_bit_nth_lsf (min_free_space, 63 - shifts) != -1)
-    return glnx_throw (error, "Integer overflow detected");
-
+  {
+    g_warning ("min-free-space-size: Integer overflow detected");
```
pwithnall: Why add the `g_warning()`? The error is already being reported to the caller.

uajain: Shouldn't be added? I thought its a benefit to the programmer to notice a warning(and infer exact problem) rather than to see a "Value too high". Maybe I am wrong :thinking:

pwithnall: The problem is already being reported using the `GError`. When this code is called by a UI, the `g_warning()` will (typically) go to the systemd journal, which the user of the UI would probably never see. When this code is called from a console application, that application is probably already doing its own printing out of the `GError` to `stderr`, so you’d end up with the same error being printed twice.

More generally, using `g_warning()` makes it hard to unit test code, since it will typically be escalated into an `abort()` by `G_DEBUG=fatal-warnings`.

uajain: hmm, Ok.
Reworked the PR.

pwithnall:
Looks OK to me. The test failures seem to be unrelated.

pwithnall:
bot, retest this please

src/libostree/ostree-repo-commit.c
```diff
@@ -429,6 +429,20 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+throw_min_free_space_error (OstreeRepo  *self,
+                            guint64      bytes_required,
+                            GError     **error)
+{
+  g_autofree char *formatted_free = g_format_size (bytes_required);
+  if (self->min_free_space_percent > 0)
+    return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s available",
+                       self->min_free_space_percent, formatted_free);
+  else
+    return glnx_throw (error, "min-free-space-size %" G_GUINT64_FORMAT "MB would be exceeded, %s available",
```
cgwalters: I tried this before but the problem is that it's not `%s available`, it's `%s requested` right?  See the 3rd hunk in this patch.  Which...I guess you are implicitly changing here.  So it seems OK, I guess we don't need to report the available space in general.

cgwalters: Or, TL;DR: I think we need `s/available/requested/` here.

uajain: hmm, yeah, Thanks for noticing. We had subtly two cases.

uajain: So, we want to report 3rd hunk as `reserved_blocks * self->txn.blocksize` maybe ? 
(because we there is `requested` in every error reporting)

uajain:
Addressed @cgwalters comments.

src/libostree/ostree-repo-commit.c
```diff
@@ -429,6 +429,20 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
 }
 
+static gboolean
+throw_min_free_space_error (OstreeRepo  *self,
+                            guint64      bytes_required,
+                            GError     **error)
+{
+  g_autofree char *formatted_free = g_format_size (bytes_required);
+  if (self->min_free_space_percent > 0)
+    return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s requested",
```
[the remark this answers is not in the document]
uajain: Right. I can't think of a perfect error message here since these checks run while writing a single file. We don't know the space requested for writing future files so `%s requested` might give a impression that, `%s` is the size we need to complete the whole transaction (which isn't the case).

src/libostree/ostree-repo-commit.c
```diff
@@ -1649,13 +1651,7 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
...
       self->cleanup_stagedir = TRUE;
       g_mutex_unlock (&self->txn_lock);
-      g_autofree char *formatted_free = g_format_size (bytes_required);
-      if (self->min_free_space_percent > 0)
-        return glnx_throw (error, "min-free-space-percent '%u%%' would be exceeded, %s available",
-                           self->min_free_space_percent, formatted_free);
-      else
-        return glnx_throw (error, "min-free-space-size %" G_GUINT64_FORMAT "MB would be exceeded, %s available",
-                           self->min_free_space_mb, formatted_free);
+      return throw_min_free_space_error (self, bytes_required, error);
```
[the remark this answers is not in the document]
uajain: Or
>So, we want to report 3rd hunk as reserved_blocks * self->txn.blocksize maybe ?
(because there is requested in every error reporting)

As per my comment in the last PR review.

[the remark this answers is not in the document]
jlebon: Isn't that just `min-free-space-size` in bytes? (Ignoring the `-percent` case for now). We're dividing by the block size to get min free space in blocks, then re-multiplying it back.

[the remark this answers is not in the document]
uajain: Yeah, we don't have direct access to value in bytes as it's abstracted in `min_free_space_calculate_reserved_blocks`

src/libostree/ostree-repo-commit.c
```diff
@@ -429,6 +429,25 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+                            guint64      bytes_required,
+                            GError     **error)
+{
+  g_autofree char *formatted_required = g_format_size (bytes_required);
+  g_autofree char *err_text = NULL;
+
+  if (bytes_required > 0)
+    err_text = g_strdup_printf ("would be exceeded, at least %s requested", formatted_required);
+  else
+    err_text = g_strdup ("would be exceeded.");
```
jlebon: Cool! Just one final note. There are two ways we could minimize the number of allocations here. First, we can move `formatted_required` inside the `if` branch so we generate it only if we need it. Second, in the `else` case, we don't need to `strdup` since it's a static string. The pattern we usually follow is something like this:

```c
const char *str = NULL;
g_autofree char *str_owned = NULL

if (cond_that_requires_allocation)
  str = str_owned = g_strdup ...
else
  str = "static string"
```

uajain: Thanks for mentioning the pattern. Addressed your comments.

jlebon:
Thanks! I just pushed a small fixup to drop the period at the end. By convention, we don't add periods in error messages.

@rh-atomic-bot r+ fbf8c8a

PR #1673, opened 2018-07-05
tests: Run network retries test for many retries directly

rfairley:
This runs a test involving many retries for the --network-retries
option directly rather than inside a conditional statement, so that
the command does not silently fail and allow the test to continue
running.

Follow-up to #1659.


jlebon:
@rh-atomic-bot r+ da1b6eb

PR #1675, opened 2018-07-07
switchroot: Move late /run/ostree-booted creation to ostree-system-ge…

akiernan:
…nerator

When ostree-prepare-root is pid 1, ostree-prepare-boot defers creation of
/run/ostree-booted, which we were doing in ostree-remount, but that's too
late if we need ostree-system-generator to bind /var. Move the creation
of the /run/ostree-booted marker to ostree-system-generator based on the
existence of the ostree= kernel command line argument (which matches the
condition that ostree-remount used).

### DIFF
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -46,13 +46,8 @@ main(int argc, char *argv[])
   struct stat stbuf;
   int i;
 
-  /* See comments in ostree-prepare-root.c for this.
-   *
-   * This service is triggered via
-   * ConditionKernelCommandLine=ostree
-   * but it's a lot easier for various bits of userspace to check for
-   * a file versus parsing the kernel cmdline.  So let's ensure
-   * the stamp file is created here too.
+  /* When systemd is in use this is normally created via the generator, but
+   * we ensure it's created here as well for redundancy.
    */
   touch_run_ostree ();
 
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -210,8 +210,9 @@ main(int argc, char *argv[])
 
 
   /* We only stamp /run now if we're running in an initramfs, i.e. we're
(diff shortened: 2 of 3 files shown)


cgwalters:
> Something I wasn't clear on, would this need changing to EXIT_SUCCESS if this patch moves forward?

Definitely; see https://bugzilla.redhat.com/show_bug.cgi?id=1331369  While we could have more clearly split out the ostree-as-host portions in the Fedora packaging, people can easily get those bits installed as a dependency of flatpak, etc.  I think the Debian packager was more careful about this, but it still pays to be defensive.

src/switchroot/ostree-remount.c
```diff
@@ -46,16 +46,6 @@ main(int argc, char *argv[])
...
 
-  /* See comments in ostree-prepare-root.c for this.
-   *
-   * This service is triggered via
-   * ConditionKernelCommandLine=ostree
-   * but it's a lot easier for various bits of userspace to check for
-   * a file versus parsing the kernel cmdline.  So let's ensure
-   * the stamp file is created here too.
-   */
-  touch_run_ostree ();
```
cgwalters: I've been trying at least a little bit to support the non-systemd case.  After this change in the "switchroot-as-pid1, !systemd" case nothing will be creating `/run/ostree-booted` right?  It should be harmless to leave this call in to ensure we create the file the same as we did previously.

Maybe just:

```
/* When systemd is in use this is normally created via the generator, but
 * we ensure it's created here as well for redundancy.
 */
  touch_run_ostree ();
```
?

akiernan: I think that makes sense... when I deleted it from there I debated whether creating `/run/ostree-booted` in many places was a sensible approach to cover corner cases. I'm certainly guilty of not thinking about !systemd...

cgwalters:
Looks great, thank you!

@rh-atomic-bot r+ ae36f71

PR #1677, opened 2018-07-09
deploy: Fix overriding kernel args for staged deployments

cgwalters:
This is the inverse of https://github.com/ostreedev/ostree/pull/1558
aka commits cadece6c4f398ca61d21e497bd6e3fbb549f9cf6 and
3358698c86d80821d81443c906621c92672f99fb

Needed to fix `rpm-ostree kargs` test suite with default staging; skipping
a test here for now as eventually what we'll do is turn on the rpm-ostree
suite fully here.


jlebon:
@rh-atomic-bot r+ 74be204

PR #1678, opened 2018-07-12
ostree commit: Add option --use-bare-user-xattrs

wmanley:
For our build-system we have a [patched version of `fakeroot`][1] which
understands the `user.ostreemeta` xattr.  It's intended for use with
`ostree checkout --user-mode --require-hardlinks`.  When programs are run
with `LD_PRELOAD=libfakeroot...` they see the ownership and permissions
that ostree understands, rather than the real ones.

Our build system mostly consists of calls to:

    ostree checkout --user-mode --require-hardlinks $COMMIT root
    LD_PRELOAD=fakeroot <make some changes>
    ostree commit --tree=dir=root --link-checkout-speedup --devino-canonical

in the past this seemed to work fine but at some point I started losing the
setuid bit on `sudo`.  I suspect the previous behaviour was a bug.  This
commit makes the `ostreemeta` preserving behaviour explicit.

`--use-bare-user-xattrs` will currently be ignored when loading a commit
from a tarball.  This can be fixed in the future.

I originally considered implementing this inside `_ostree_repo_commit_modifier_apply` but we don't have access to the file fd to retrieve the xattrs from in there.  I considered using `g_file_info_get_attribute_data (info, "xattr::ostreemeta", ...)` but it GLib has no way of retrieving an attribute that may contain embedded NULs.

This may need a little bit more work (see TODO list below) but I wanted to raise it sooner rather than later to receive feedback.

I'm not super happy with the name

[1]: https://github.com/stb-tester/fakeroot/tree/ostree

**TODO:**

- [ ] Test xattrs in xattrs
- [ ] Decide on and test the precedence of `--canonical-permissions`, `--owner-uid`, `--owner-gid` and `--use-bare-user-xattrs`.  My feeling is that if the xattrs are present they should take precedence.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -3013,6 +3013,60 @@ create_tree_variant_from_hashes (GHashTable            *file_checksums,
   return g_variant_ref_sink (serialized_tree);
 }
 
+static gboolean
+load_bare_user_xattrs (int                       dfd,
+                       const char               *name,
+                       GFileInfo                *file_info,
+                       GFileInfo               **out_modified_info,
+                       GVariant                **out_xattrs,
+                       GCancellable             *cancellable,
+                       GError                  **error)
+{
+  /* I think ENODATA will resolve to G_IO_ERROR_FAILED for now which isn't ideal
+   * but currently glib doesn't know anything about ENODATA: */
+  GIOErrorEnum enodata_gerror_code = g_io_error_from_errno (ENODATA);
+
+  g_autoptr (GError) tmp_error = NULL;
+  g_autoptr(GBytes) bytes = glnx_lgetxattrat (dfd, name, "user.ostreemeta",
+                                              &tmp_error);
+  if (bytes == NULL)
+    {
+      if (g_error_matches (tmp_error, G_IO_ERROR, enodata_gerror_code))
+        goto out_nometa;
+      g_propagate_error (error, g_steal_pointer (&tmp_error));
+      return FALSE;
+    }
+
+  struct stat current_stat;
+  _ostree_gfileinfo_to_stbuf (file_info, &current_stat);
+
+  struct stat new_stat;
+  g_autoptr(GVariant) ret_xattrs = _ostree_filemeta_to_stat (&new_stat, bytes);
+  new_stat.st_mode = (current_stat.st_mode & S_IFMT) | (new_stat.st_mode & 07777);
+
+  if (new_stat.st_uid != current_stat.st_uid ||
+      new_stat.st_gid != current_stat.st_gid ||
+      new_stat.st_mode != current_stat.st_mode)
+    {
+      g_autoptr(GFileInfo) modified_info = g_file_info_dup (file_info);
+
+      g_file_info_set_attribute_uint32 (modified_info, "unix::uid", new_stat.st_uid);
+      g_file_info_set_attribute_uint32 (modified_info, "unix::gid", new_stat.st_gid);
+      g_file_info_set_attribute_uint32 (modified_info, "unix::mode", new_stat.st_mode);
+
+      g_set_object (out_modified_info, modified_info);
+    }
+  else
+    g_set_object (out_modified_info, file_info);
+
+  *out_xattrs = g_steal_pointer (&ret_xattrs);
+  return TRUE;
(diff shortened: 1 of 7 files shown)


cgwalters:
> LD_PRELOAD=fakeroot <make some changes>

Yeah.  For [gnome-continuous](https://git.gnome.org/browse/gnome-continuous) I basically ended up maintaining the set of non-root or setuid things out of band.  What happens with rpm-ostree is pretty similar; the canonical permissions are maintained in the RPM spec `%files` section - it's also out-of-band.

Did you evaluate just how many packages/components/sources have non-root-owned files or setuid files?  I suspect it's similar.

Anyways, let me see if I understand this bug correctly - your patched version of `fakeroot` is writing out its own `user.ostreemeta` xattrs, and the current commit process effectively only uses that if the file is already hardlinked?

Offhand...I feel like we could probably just make this the default - if we detect a `user.ostreemeta` xattr, we adopt it?  The downside of this level of magic would be that it becomes impossible to commit a filesystem tree that happens to contain a bare-user ostree repository.  I'm not sure how much I care about that though - what would be the use case?

wmanley:
> Did you evaluate just how many packages/components/sources have non-root-owned files or setuid files? I suspect it's similar.

I've not looked into this explicitly.  I'm installing stock Ubuntu 14.04 and 16.04 using `multistrap` and then `bwrap`, so I want to avoid modifying the upstream packages.  `/usr` is mostly root:root, but some files have a different gid, perticularly setgid executables like `wall`.  `/etc` has more variety with ssl certificates stored as user `ssl-cert`, and the `shadow`, etc. files as group `shadow`. 

> Anyways, let me see if I understand this bug correctly - your patched version of fakeroot is writing out its own `user.ostreemeta` xattrs, and the current commit process effectively only uses that if the file is already hardlinked?

Exactly

> Offhand...I feel like we could probably just make this the default - if we detect a user.ostreemeta xattr, we adopt it? The downside of this level of magic would be that it becomes impossible to commit a filesystem tree that happens to contain a bare-user ostree repository. I'm not sure how much I care about that though - what would be the use case?

I can't think of any use-case, I implemented it this way because I thought the change of behaviour would be more acceptable if it we're behind an explicit option.  I'd be happy to make it the default.

One open question then would be how to deal with files that don't have `user.ostreemeta` xattrs set.  As I've been using this new code I've found that I want non-xattr files to be interpreted as `root:root 0644`.  `fakeroot` doesn't seem to catch all the file-accesses - it doesn't interpose `creat` for example - but I don't want these files to take on the uid from the build environment - I want them to be root.  So I've made `ostree commit` fall-back to root:root.  I'm not sure of an elegant way of solving this problem without an explicit flag.

I've just pushed two more commits that:

* Make uid:gid default to 0:0
* Make `ostree checkout` also write `ostreemeta` xattrs for directories that it creates.

I'm not sure what you think of these ideas?

wmanley:
I've rebased this and reworked it a bit fixing the xattrs in xattrs case.  It's a bit more general than before because it shares the xattr loading code with `get_final_xattrs`.

I seem to have lost the ability to get the bot to do my bidding.  I guess that's related to my name being commented out in [OWNERS](https://github.com/ostreedev/ostree/blob/master/OWNERS):

>     # Does not currently have affiliation public
>     # - wmanley

@cgwalters: What is affiliation in this context?, and how does one make it public?

/assign @pwithnall

cgwalters:
> @cgwalters: What is affiliation in this context?, and how does one make it public?

https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/publicizing-or-hiding-organization-membership

The bot needs to know you're a member of the organization.  

(Plus, you can think of it like [collecting flair](https://user-images.githubusercontent.com/244096/86394385-80b42180-bc6c-11ea-9243-8b4f83b793cf.png)! )

wmanley:
> The bot needs to know you're a member of the organization.

Great, thanks.  Let's give it a go:

/test sanity

> (Plus, you can think of it like [collecting flair](https://user-images.githubusercontent.com/244096/86394385-80b42180-bc6c-11ea-9243-8b4f83b793cf.png)! )

:smile:

wmanley:
TODO: Rebase and test for conflicts with #2199 and #2198 - check that empty files and files with 0 permissions have xattrs set on checkout.

PR #1679, opened 2018-07-16
build-sys: Link with gpg-error directly

cgwalters:
We use the API, and not linking breaks the build with e.g.
`-fuse-ld=gold` in a Fedora 28 buildroot as gold doesn't do the
"search indirect dependencies" thing.


jlebon:
@rh-atomic-bot r+ 92d8d95

jlebon:
@rh-atomic-bot retry

PR #1681, opened 2018-07-16
Add a change-update-summary config option

mwleeds:


mwleeds:
Fortunately the CI failures look like transient networking errors

man/ostree.repo-config.xml
```diff
@@ -93,6 +93,14 @@ Boston, MA 02111-1307, USA.
         to <literal>false</literal>.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>change-update-summary</varname></term>
+        <listitem><para>Boolean value controlling whether or not to
+        automatically update the summary file after any ref is added,
+        removed, or updated. Defaults to <literal>false</literal>.
```
cgwalters: It's worth mentioning/comparing with `commit-update-summary` from 9e6ac6d822f4dded703a99912f629ea1e9b3bcbc in <https://github.com/ostreedev/ostree/pull/83>.  This is an extension of that.

I understand the argument for doing this; but do also keep in mind my original arguments for not doing it by default - systems may want to ensure that multiple refs are updated atomically.  I suppose though that is (mostly) supported here if one does the ref updates in a single ostree txn via `transaction_set_ref()`.

I am not sure if there was an explicit reason the `commit-update-summary` option was implemented in the `ostree commit` CLI - it seems to me it should be in the library in `ostree_repo_commit_transaction()` right?

src/libostree/ostree-repo.c
```diff
@@ -5385,13 +5385,27 @@ summary_add_ref_entry (OstreeRepo       *self,
...
+ * Locking: exclusive
  */
 gboolean
 ostree_repo_regenerate_summary (OstreeRepo     *self,
                                 GVariant       *additional_metadata,
                                 GCancellable   *cancellable,
                                 GError        **error)
 {
+  /* Take an exclusive lock. This makes sure the commits and deltas don't get
+   * deleted while generating the summary. It also means we can be sure refs
```
cgwalters: Yes, but it doesn't protect against something deleting them immediately afterwards.  I can see this helping in some cases of course, but we'd need a lot more work (probably some sort of client-side retry loop) to ensure consistency from the client's view.

man/ostree.repo-config.xml
mwleeds: >It's worth mentioning/comparing with commit-update-summary from 9e6ac6d in #83. This is an extension of that.

Sure, I added another sentence.

>I understand the argument for doing this; but do also keep in mind my original arguments for not doing it by default - systems may want to ensure that multiple refs are updated atomically. I suppose though that is (mostly) supported here if one does the ref updates in a single ostree txn via transaction_set_ref().

Even with transactions the summary update won't be atomic. It will be updated for each ref update in `ostree_repo_commit_transaction()`. I suppose the alternative would be to teach eos-updater and flatpak to update the summary after an operation but I'm not sure it's worth the complexity that would entail. You'd probably want it to be a configurable option, but why would clients of ostree implement its own option? And perhaps the atomicity won't matter much in practice, since flatpak pulls most dependencies before the app ref, and at least for the USB case GNOME Software shouldn't show an app as able to be put on a USB until it's been fully installed. In the LAN case I think GNOME Software would show the app despite missing dependencies so you could run into errors if you try to install at exactly the wrong time.

>I am not sure if there was an explicit reason the commit-update-summary option was implemented in the ostree commit CLI - it seems to me it should be in the library in ostree_repo_commit_transaction() right?

I had the same question but didn't feel qualified to make that change.

src/libostree/ostree-repo.c
mwleeds: In the case of a ref deletion the summary would only briefly be out of date because the deletion would trigger an update. I'm not sure how you could get around that because I don't know how you could update the two files atomically.

mwleeds:
I think the unit test failures are not related to these changes

mwleeds:
bot, retest this please

man/ostree.repo-config.xml
jlebon: > it seems to me it should be in the library in ostree_repo_commit_transaction() right?

That struck me as odd too. We should open an issue for that.

jlebon:
Thanks! This looks sane to me. I'll let @cgwalters do the final review in case he had other concerns.

man/ostree.repo-config.xml
jlebon: #1689

mwleeds:
I pushed a fixup that changes the implementation to delay summary updates until the end of a transaction, which should be more performant.

src/libostree/ostree-repo-commit.c
```diff
@@ -2201,6 +2201,24 @@ ostree_repo_commit_transaction (OstreeRepo                  *self,
...
+    if (!ot_keyfile_get_boolean_with_default (self->config, "core",
+                                              "change-update-summary", FALSE,
+                                              &update_summary, error))
+      return FALSE;
+
+    if (update_summary && !ostree_repo_regenerate_summary (self,
+                                                           NULL,
+                                                           cancellable,
+                                                           error))
+      return FALSE;
```
jlebon: At this point, it might be worth adding a `update_summary` field in `OstreeRepo`, or at least factor this out to a helper function?

mwleeds: If it were a field initialized when `OstreeRepo` is initialized, there's a possibility its on-disk value is different by the time this function is run. Sure, we could make it a helper function.

mwleeds:
I pushed a fixup to break out a helper function, and another fixup to help if anyone wants to compare the repo mtime with the summary mtime.

src/ostree/ot-builtin-commit.c
```diff
@@ -824,21 +824,32 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
...
+                                                &change_update_summary, error))
         goto out;
 
-      if (update_summary && !ostree_repo_regenerate_summary (repo,
-                                                             NULL,
-                                                             cancellable,
-                                                             error))
-        goto out;
+      /* No need to update it again if we did for each ref change */
+      if (opt_orphan || !change_update_summary)
```
jlebon: Shouldn't this be `if (!opt_orphan && !change_update_summary)`? Or are we trying to retain backwards compatibility with the previous behaviour of updating the summary even for orphan commits? You don't have to make the change in this PR if that's the reasoning. I can't imagine anyone actually relying on that somehow, though I'm cool with making this a separate patch for clarity.

mwleeds: Yes, I was trying to continue doing what the code did before

jlebon:
Thanks!
@rh-atomic-bot r+ 0b3b563

jlebon:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1682, opened 2018-07-17
Add a check for gpg-error via pkg-config

ebassi:
Some downstreams — namely, the Yocto Project — ship gpg-error with a
pkg-config file, and modify gpg-error-config to error out when you try
using it instead of pkg-config.

We can check for gpg-error via pkg-config, and if it's not available,
fall back to gpg-error-config.

Signed-off-by: Emmanuele Bassi <ebassi@gnome.org>


cgwalters:
Thanks!

@rh-atomic-bot r+ f34d613

PR #1686, opened 2018-07-20
ci: Mark insttests as not required

cgwalters:
The reliablity has just not been what we need, and they haven't
yet caught any real bugs.  Until I can carve off some time to
truly make this reliable let's just mark it as not required.
I'd like to gather more statistics on the failure scenarios.


jlebon:
@rh-atomic-bot r+ 686f74c

PR #1691, opened 2018-07-24
lib/repo-commit: Fix min-free-space error message

mwleeds:
Since min_free_space_size_mb is considered before min_free_space_percent
in min_free_space_calculate_reserved_blocks(), it has to be considered
first when generating the error message in order for it to be accurate.


mwleeds:
cc @uajain

jlebon:
Thanks!
@rh-atomic-bot r+ fc2919f

PR #1692, opened 2018-07-26
Minor coverity fixes

jlebon:
See commit messages for details.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2486,7 +2486,10 @@ process_one_static_delta (OtPullData                 *pull_data,
                                                OSTREE_STATIC_DELTA_OPEN_FLAGS_SKIP_CHECKSUM,
                                                NULL, &inline_delta_part,
                                                cancellable, error))
-            return FALSE;
+            {
+              fetch_static_delta_data_free (fetch_data);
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -1819,7 +1819,7 @@ scan_commit_object (OtPullData                 *pull_data,
   if (pull_data->timestamp_check)
     {
       /* We don't support timestamp checking while recursing right now */
-      g_assert (ref);
+      g_assert (ref != NULL);
```
cgwalters: Did you run this through Coverity again or is it a guess?

jlebon: It's a guess. I'm not sure how to request another run. Will look into it.

jlebon: It didn't make a difference, so I dropped that hunk. I left the other two though since I think they're still useful and reworked the commit message. :arrow_down:

jlebon: (To clarify, the first one is definitely a good check to have, while the second one is just Coverity not being smart enough. Though they both did squash warnings).

cgwalters:
@rh-atomic-bot r+ a6014ec

PR #1693, opened 2018-07-30
Deprecate commit-update-summary and rename change-update-summary

jlebon:
Minor follow-ups to #1681.
See individual commit messages.

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -94,7 +94,7 @@ Boston, MA 02111-1307, USA.
       </varlistentry>
 
       <varlistentry>
-        <term><varname>change-update-summary</varname></term>
+        <term><varname>auto-update-summary</varname></term>
         <listitem><para>Boolean value controlling whether or not to
         automatically update the summary file after any ref is added,
         removed, or updated. This covers a superset of the cases covered by
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -87,20 +87,20 @@ Boston, MA 02111-1307, USA.
       </varlistentry>
 
       <varlistentry>
-        <term><varname>commit-update-summary</varname></term>
-        <listitem><para>Boolean value controlling whether or not to
-        automatically update the summary file after a commit.  Defaults
-        to <literal>false</literal>.</para></listitem>
-      </varlistentry>
-
-      <varlistentry>
         <term><varname>auto-update-summary</varname></term>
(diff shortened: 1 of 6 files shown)


jlebon:
I'll rebase this once #1681 gets merged!

/cc @mwleeds

cgwalters:
All looks good to me!  Will let Matthew do review/approve.

@rh-atomic-bot delegate=mwleeds

mwleeds:
Seems pretty reasonable at a high level. I'll look at it more closely later

src/ostree/ot-builtin-commit.c
```diff
@@ -823,33 +822,6 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
...
-                                                "change-update-summary", FALSE,
-                                                &change_update_summary, error))
-        goto out;
-
-      /* No need to update it again if we did for each ref change */
-      if (opt_orphan || !change_update_summary)
-        {
-          gboolean commit_update_summary;
-
-          /* The default for this option is FALSE, even for archive repos,
```
mwleeds: Maybe it would be good to have some explanation of the default value over in `_ostree_repo_maybe_regenerate_summary` since we're losing this explanation. How about: "The default for this option is FALSE, because of the performance cost and because servers may want to atomically update the summary after a set of operations (auto-update-summary only updates atomically within a transaction)."

man/ostree.repo-config.xml
```diff
@@ -87,20 +87,18 @@ Boston, MA 02111-1307, USA.
       </varlistentry>
 
       <varlistentry>
-        <term><varname>commit-update-summary</varname></term>
+        <term><varname>auto-update-summary</varname></term>
```
mwleeds: I like the name `auto-update-summary`, but it also makes it sound like any change that would affect the summary triggers a summary update, which isn't quite true. Maybe we should mention that some operations such as generating a static delta or setting `core.collection-id` in the repo config would make the summary stale but not trigger an update with this option set?

jlebon:
Pushed a fixup for comments! :arrow_up:

mwleeds:
@rh-atomic-bot r+

jlebon:
Weird, it looks like Homu didn't pick up all the fixups.
Let me `rebase --autosquash` locally and retry.

jlebon:
@rh-atomic-bot r=mwleeds 0601308

PR #1694, opened 2018-07-30
OSTree Contributing Tutorial and extending CONTRIBUTING.md  

Bubblemelon:
#### Pull Request Objective

The objective of this PR is to edit and add content the docs related to developing/contributing, so that anyone including new hires, external contributors etc, have a more detailed overview on how to start developing ostree while conforming to ostree's workflow and coding guidelines.  

#### To Do List

- [x] Larger sections of `CONTRIBUTING.md` should be placed in a separate markdown. ( buidling + modifiying + testing sections should stay in this file )

- [x] Have `CONTRIBUTING.md` be brief and concise. 

- [x] Check that all anchors and hyperlinks work.

- [x] Add a new heading  `CONTRIBUTING.md` to give a link the tutorial

- [x] Leave the headings in `CONTRIBUTING.md` the same.

- [x] Have all commits be broken down into 2 commits, one commit for` getting-started.md` and one for `CONTRIBUTING.md`

- [x] Rename `gettting-started.md` to `contributing-tutorial.md` and the title for this doc to `OSTree Contributing Tutorial`

#### Questions

Should a table of contents be placed on top of `CONTRIBUTING.md`? ([readthedocs.io](https://ostree.readthedocs.io/) renders the headers on left hand side panel) -- other docs do not have a table of contents on top

**Jira Card**: https://projects.engineering.redhat.com/browse/COREOS-289

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 Submitting patches
 ------------------
 
-A majority of current maintainers prefer the Github pull request
+A majority of current maintainers prefer the GitHub pull request
 model, and this motivated moving the primary git repository to
 <https://github.com/ostreedev/ostree>.
 
@@ -12,19 +12,19 @@ for more information.  Instead, we use an instance of
 [Homu](https://github.com/servo/homu), currently known as
 `cgwalters-bot`.
 
-As a review proceeds, the preferred method is to push `fixup!`
-commits via `git commit --fixup`.  Homu knows how to use
-`--autosquash` when performing the final merge.  See the
+As a review proceeds, the preferred method is to push `fixup!` commits. Any commits committed with the `--fixup` option will have have the word `fixup!` in its commit title. This is to indicate that this particular commit will be squashed with the commit that was specified in this command, `git commit --fixup <commit ref or hash>`. Homu knows how to use `--autosquash` when performing the final merge.
+
+See the
 [Git documentation](https://git-scm.com/docs/git-rebase) for more
 information.
 
-Alternative methods if you don't like Github (also fully supported):
+Alternative methods if you don't like GitHub (also fully supported):
 
  1. Send mail to <ostree-list@gnome.org>, with the patch attached
  1. Attach them to <https://bugzilla.gnome.org/>
 
 It is likely however once a patch is ready to apply a maintainer
-will push it to a github PR, and merge via Homu.
+will push it to a GitHub PR, and merge via Homu.
 
 Commit message style
 --------------------
@@ -34,6 +34,37 @@ similar to the
 [Linux kernel](https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git).
 
 You may use `Signed-off-by`, but we're not requiring it.
+
+**General Commit Message Guidelines**:
+
+1. Title
+    - Specify the context or category of the changes e.g. `lib` for library changes, `docs` for document changes, `bin/<command-name>` for command changes, etc.
+    - Begin the title with the first letter of the first word capitalized.
+    - Aim for less than 50 characters, otherwise 72 characters max.
+    - Do not end the title with a period.
+    - Use an [imperative tone](https://en.wikipedia.org/wiki/Imperative_mood).
+2. Body
+    - Separate the body with a blank line after the title.
+    - Begin a paragraph with the first letter of the first word capitalized.
+    - Each paragraph should be formatted within 72 characters.
+    - Content should be about what was changed and why this change was made.
+    - If your commit fixes an issue, the commit message should end with `Closes: #<number>`.
+
+Commit Message example:
+
+```bash
+<context>: Less than 50 characters for subject title
+
+A paragraph of the body should be within 72 characters.
+
+This paragraph is also less than 72 characters.
+```
+
+For more information see [How to Write a Git Commit Message](https://chris.beams.io/posts/git-commit/)
+
+**Editing a Committed Message:**
+
+To edit the message from the most recent commit run `git commit --amend`. To change older commits on the branch use `git rebase -i`. For a successful rebase have the branch track `upstream master`. Once the changes have been made and saved, run `git push --force origin <branch-name>`.
 
 Running the test suite
 ----------------------
@@ -136,3 +167,8 @@ Instead do this:
               goto out;
           }
       }
+
+Contributing: Tutorial
+----------------------
+
+For a detailed walk-through on building, modifying, and testing, see this [tutorial on how to start contributing to OSTree](contributing-tutorial.md).
--- a/docs/contributing-tutorial.md
+++ b/docs/contributing-tutorial.md
@@ -1,0 +1,450 @@
+# OSTree Contributing Tutorial
+
+The following guide is about OSTree forking, building, adding a command, testing the command, and submitting the change.
+
+- [Getting Started](#getting-started)
+- [Building OSTree](#building-ostree)
+    - [Install Build Dependencies](#install-build-dependencies)
+    - [OSTree Build Commands](#ostree-build-commands)
+- [Testing a Build](#testing-a-build)
+    - [Testing in a Container](#testing-in-a-container)
+    - [Testing in a Virtual Machine](#testing-in-a-virtual-machine)
+- [Tutorial: Adding a basic builtin command to OSTree](#tutorial-adding-a-basic-builtin-command-to-ostree)
+    - [Modifying OSTree](#modifying-ostree)
+    - [OSTree Tests](#ostree-tests)
+    - [Submitting a Patch](#submitting-a-patch)
+    - [Returning Workflow](#returning-workflow)
+
+---
+
+## Getting Started
+
+Fork https://github.com/ostreedev/ostree, then run the following commands.
+
+```bash
+$ git clone https://github.com/<username>/ostree && cd ostree
+$ git remote add upstream https://github.com/ostreedev/ostree
+$ git checkout master
+$ git fetch upstream && git branch --set-upstream-to=upstream/master master
+```
+Make a branch from master for your patch.
+
+```bash
+$ git checkout -b <name-of-branch>
+$ git branch --set-upstream-to=upstream/master <name-of-branch>
+```
+
+## Building OSTree
+
+### Install Build Dependencies
+
+Execute one of the following group commands as superuser depending on your machine's package manager.
+
+For Fedora:
+
+```bash
+$ dnf install @buildsys-build dnf-plugins-core && \
+dnf builddep ostree
+```
+
+For CentOS:
+
+```bash
+$ yum install yum-utils dnf-plugins-core && \
+yum-builddep ostree
+```
+
+For Debian based distros:
+
+```bash
+$ apt-get update && \
+apt-get install build-essential && \
+apt-get build-dep ostree
+```
+
+[build.sh](../ci/build.sh) will have a list of packages needed to build ostree.
+
+### OSTree Build Commands
+
+These are the basic commands to build OSTree. Depending on the OS that OSTree will be build for, the flags or options for `./autogen.sh` and `./configure` will vary.
+
+See `ostree-build.sh` in this tutorial below for specific commands to building OSTree for Fedora 28 and Fedora 28 Atomic Host.
+
+```bash
+# optional: autogen.sh will run this if necessary
+git submodule update --init
+
+env NOCONFIGURE=1 ./autogen.sh
+
+# run ./configure if makefile does not exist
+./configure
+
+make
+make install DESTDIR=/path/to/install/binary
+```
+
+#### Notes
+
+Running `git submodule update --init` is optional since `autogen.sh` will check to see if one of the submodule files for example from `libglnx/` or from `bsdiff/` exists.
+
+Additionally, `autogen.sh` will check to see if the environment variable `NOCONFIGURE` is set. To run `./configure` manually, run autogen in a modified environment as such, `env NOCONFIGURE=1 ./autogen.sh`.
+
+Otherwise, leave `NOCONFIGURE` empty and `autogen.sh` will run `./configure` as part of the `autogen.sh` command when it executes.
+
+For more information on `--prefix` see [Variables for Installation Directories](https://www.gnu.org/prep/standards/html_node/Directory-Variables.html#Directory-Variables).
+
+`make install` will generate files for `/bin` and `/lib`. If `DESTDIR` is unspecified then OSTree will be installed in the default directory i.e. `/usr/local/bin` and its static libraries in `/usr/local/lib`. Note that the `/usr/local` portion of the path can be changed using the `--prefix` option for `./configure`.
+
+See this [GNU guide on `DESTDIR` Staged Installs](https://www.gnu.org/prep/standards/html_node/DESTDIR.html) for more information.
+
+#### Tip
+
+Make allows parallel execution of recipes. Use `make -j<N>` to speed up the build. `<N>` is typically `$((2 * $(nproc)))` for optimal performance, where `nproc` is the number of processing units (CPU cores) available.
+
+See page 106 of the [GNU Make Manual](https://www.gnu.org/software/make/manual/make.pdf) for more information about the `--jobs` or `-j` option.
+
+## [Testing a Build](#testing-a-build)
+
+It is best practice to build software (definitely including ostree) in a container or virtual machine first.
+
+### Testing in a Container
+
+There are a variety of container engines available and many distributions have pre-packaged versions of e.g. [Podman](https://github.com/projectatomic/libpod) and Docker.
+
+If you choose to use [Docker upstream](https://docs.docker.com/v17.09/engine/installation/linux/docker-ce/fedora/), you may want to follow this [post-installation guide for Docker](https://docs.docker.com/v17.09/engine/installation/linux/linux-postinstall/). This will allow you to run Docker as a non-root user on a Linux based host machine.
+
+You will need to have pushed a remote git branch `$REMOTE_BRANCH` (see `ostree-git.sh below`) in order to pull your changes into a container.
+
+The example below uses Docker to manage containers. Save the contents of this **Dockerfile** somewhere on your machine:
+
+```bash
+# this pulls the fedora 28 image
+FROM registry.fedoraproject.org/fedora:28
+
+# install ostree dependencies
+RUN dnf update -y && \
+    dnf -y install @buildsys-build dnf-plugins-core  && \
+    dnf -y builddep ostree  && \
+    dnf clean all
+
+# clone ostree and update master branch
+COPY ostree-git.sh /
+RUN ../ostree-git.sh
+
+# builds ostree + any additional commands
+COPY ostree-build.sh /
+
+# entry into the container will start at this directory
+WORKDIR /ostree
+
+# run the following as `/bin/sh -c`
+# or enter the container to execute ./ostree-build.sh
+RUN ../ostree-build.sh
+
+```
+
+Save the following bash scripts in the same directory as the Dockerfile. Then change the mode bit of these files so that they are executable, by running `chmod +x ostree-git.sh ostree-build.sh`
+
+```bash
+#!/bin/bash
+
+# ostree-git.sh
+# Clone ostree and update master branch
+
+set -euo pipefail
+
+# Set $USERNAME to your GitHub username here.
+USERNAME=""
+
+# clone your fork of the OSTree repo, this will be in the "/" directory
+git clone https://github.com/$USERNAME/ostree.git
+cd ostree
+
+# Add upstream as remote and update master branch
+git checkout master  
+git remote add upstream https://github.com/ostreedev/ostree.git
+git pull --rebase upstream master
+```
+
+```bash
+#!/bin/bash
+
+# ostree-build.sh
+# Build and test OSTree
+
+set -euo pipefail
+
+# $REMOTE_BRANCH is the name of the remote branch in your
+# repository that contains changes (e.g. my-patch).
+REMOTE_BRANCH=""
+
+# fetch updates from origin
+# origin url should be your forked ostree repository
+git fetch origin
+
+# go to branch with changes
+# if this branch already exists then checkout that branch
+exit_code="$(git checkout --track origin/$REMOTE_BRANCH; echo $?)"
+if [[ "$exit_code" == 1 ]]
+then
+    echo "This branch:" $REMOTE_BRANCH "is not a remote branch."
+    exit
+fi
+
+# make sure branch with changes is up-to-date
+git pull origin $REMOTE_BRANCH
+
+# build OSTree commands for Fedora 28 and Fedora 28 Atomic Host
+./autogen.sh --prefix=/usr --libdir=/usr/lib64 --sysconfdir=/etc
+./configure --prefix=/usr
+make -j$((2 * $(nproc)))
+make install
+
+# any additional commands go here
+```
+
+**Build the container**
+
+Run `docker build` in the same directory of the `Dockerfile` like so:
+
+```bash
+$ docker build -t ostree-fedora-test .
+```
+
+When this build is done, the `-t` option tags the image as `ostree-fedora-test`.  
+
+**Note**: Do not forget the dot **.** at the end of the above command which specifies the location of the Dockerfile.
+
+You will see `ostree-fedora-test` listed when running `docker images`:
+
+```bash
+REPOSITORY                                 TAG                 IMAGE ID            CREATED             SIZE
+ostree-fedora-test                         latest              817c04cc3656        1 day ago          978MB
+```
+
+**Entering the Container**
+
+To **start** the `ostree-fedora-test` container, run:
+
+```bash
+$ docker run -it --rm --entrypoint /bin/sh --name ostree-testing ostree-fedora-test
+```
+
+**Note**:
+
+`--rm` option tells [Docker to automatically clean up the container and remove the file system when the container exits](https://docs.docker.com/engine/reference/run/#clean-up---rm). Otherwise remove it manually by running `docker rm <container name>`.
+
+The state of the container will not be saved when the shell prompt exits. Best practice is modify the Dockerfile to modify the image.
+
+**Testing in a Container Workflow**
+
+1. Edit the changes to OSTree on your local machine.
+2. `git add` to stage the changed files, `git commit` and then `git push origin <local-branch>:<remote-branch>`.
+3. Testing on a _new_ container vs. Testing on an _existing_ container:
+
+    If the `ostree-testing` container was newly built right after your changes have been committed, then the container's build of `ostree` should contain your changes.
+
+    Else: Within the `ostree-testing` container, run `../ostree-build.sh` in the ostree directory. This will pull in changes from your branch and create a new `ostree` build.
+
+4. `make install` will install OSTree in the default location i.e. `/usr/..`in a Fedora 28 container.
+
+5. Test `ostree`.
+
+### Testing in a Virtual Machine
+
+To create a Fedora 28 Atomic Host Vagrant VM, run the following commands:
+
+```bash
+$ mkdir atomic && cd atomic
+$ vagrant init fedora/28-atomic-host && vagrant up
+```
+
+An option is to use `rsync` to transfer `ostree` files to a Vagrant VM.
+
+To find the IP address of a Vagrant VM, run `vagrant ssh-config` in the same directory as the `Vagrantfile`.
+
+**Steps to `rsync` files to test an `ostree` build**:
+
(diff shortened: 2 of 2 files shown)


docs/CONTRIBUTING.md
```diff
@@ -1,11 +1,327 @@
...
+$ git branch --set-upstream-to=upstream/master <name-of-branch>
+```
+
+## [Building OSTree](#building-ostree)
+
+### [Install Build Dependencies](#install-build-dependencies)
+
+```bash
+$ sudo dnf builddep ostree
+$ sudo dnf install @buildsys-build
```
cgwalters: I'd phrase this more like "Use distribution build dependencies (if possible)".  Something like:

> For distributions where ostree is already packaged, it's often easy to just use the package manager.  On Fedora: `sudo dnf install @buildsys-build dnf-plugins-core && dnf builddep ostree`.  On CentOS7, `yum-builddep`.  Debian: `apt builddep ostree`, etc.

I'd imagine there's a generic page about this somewhere.

docs/CONTRIBUTING.md
```diff
@@ -1,11 +1,327 @@
...
+```
+
+### [OSTree Build Commands](#ostree-build-commands)
+
+```bash
+git submodule update --init
+./autogen.sh --prefix=/usr --libdir=/usr/lib64 --sysconfdir=/etc
```
cgwalters: Calling out the submodule init is fine, but note `autogen.sh` will also do it.  It also runs `./configure`, so you should omit that.

docs/CONTRIBUTING.md
```diff
@@ -1,11 +1,327 @@
...
+
+`make install` will generate files for `/etc` and `/usr`
+
+For more information and issues with `./autogen.sh`, `./configure` or `make` see [Build Dependencies](/docs/misc/build-dependencies.md).
+
+## [Testing a Build](#testing-a-build)
+
+**TODO**: This sentense below needs more context and perhaps needs to be rephrased.
+
+Testing OSTree could interfere with the state of your machine. To avoid this, it is preferable to test and build OSTree inside a container or virtual machine.
```
cgwalters: I'd just say:

> It's best practice to build software (definitely including ostree) in a container or virtual machine first.

docs/CONTRIBUTING.md
```diff
@@ -1,11 +1,327 @@
...
+
+### [Testing in a Container](#testing-in-a-container)
+
+Docker is the preferred containerization tool. Follow this guide to [install Docker on your host machine](https://docs.docker.com/v17.09/engine/installation/linux/docker-ce/fedora/). You may need to follow this [post-installation guide for Docker](https://docs.docker.com/v17.09/engine/installation/linux/linux-postinstall/) if you would like to run Docker as a non-root user on a Linux based host machine.
```
cgwalters: This is a pretty strong opinion; I personally use podman now and not Docker.  Fedora also does have a `docker` package included (which isn't the same as upstream).  But as upstream I'd personally help anyone who was using `systemd-nspawn` or whatever too.  Not to mention buliding ostree in flatpaks.

So...how about:

> There are a variety of container engines available; many distributions have pre-packaged versions of e.g. `podman`, `docker`; you can also [use Docker from upstream](https://docs.docker.com/v17.09/engine/installation/linux/docker-ce/).

docs/CONTRIBUTING.md
```diff
@@ -1,11 +1,327 @@
...
+WORKDIR /ostree
+```
+**Build the container**
+
+Run `docker build` in the same directory of the `Dockerfile` like so:
+
+```
+$ docker build -t ostree-fedora-test .
+```
+When this build is done, the `-t` option tags the image as `ostree-fedora-test`.
```
cgwalters: I'm OK with including this but I'd try really hard to avoid doing this myself as it's really, really slow for iteration to re-download the build dependencies each time.  This is why I like the "pet container" pattern more.

docs/CONTRIBUTING.md
```diff
@@ -1,11 +1,327 @@
...
+```
+4. `make install` will install OSTree in the default location of the guest OS.
+
+### [Testing in a Virtual Machine](#testing-in-a-virtual-machine)
+
+See this guide on creating an [Atomic Host Vagrant VM](https://gist.github.com/Bubblemelon/bd9f3cf429d9e04be52f9987baed53c8).
+
+To test a build on the Atomic Host Vagrant VM, simply `scp -r <ostree binary location> vagrant@<ip-address>:~`. You can find the IP address of a Vagrant VM by running `vagrant ssh-config` in the same directory as the `Vagrantfile`.
+
+Run `vagrant ssh` to login to the VM. To override the `Read-only file system` warning, run `ostree admin unlock`. Now replace the original OSTree binary with the new one by running `sudo mv ~/ostree /usr/bin/ostree`.
```
cgwalters: These bits I think are among the most interesting part of the guide; a lot of the earlier instructions are generic for building any autotools software, and some are generic to C/C++/software in general.

This bit is where we're talking about something that's specific to targeting an ostree-managed system.

On that topic, one thing you *absoutely* need to do is sync *all* of the artifacts from the build - most notably including the `/usr/lib64/libostree-1.so.1.0.0` shared library.

There are two paths to doing that: basically `rsync -rlv ~/ostree-install/usr/ /usr/`, or building an RPM.  Both have advantages/disadvantages; I personally tend to use the rsync path more.

(This is where if we want to get really opinionated a `Vagrantfile` probably makes the most sense?)

docs/CONTRIBUTING.md
```diff
@@ -1,11 +1,327 @@
...
+4. Make sure `test-hello-ostree.sh` has executable permissions!
+
+    ```bash
+    chmod +x tests/test-hello-ostree.sh
+    ```
+
+5. Run the test:
+
+    ```
+    make check TESTS="test/test-hello-ostree.sh
```
cgwalters: This all looks good!

miabbott:
There's a lot of good content here, but I agree with the TODO list about keeping `CONTRIBUTING.md` concise.

I'd advocate for moving the bulk of the `Getting Started` through `Tutorial` sections into something like `docs/getting-started.md` and link to that from `CONTRIBUTING.md`.  

I could see enhancing the existing `CONTRIBUTING.md` with the details about the commit message best practices.

docs/CONTRIBUTING.md
```diff
@@ -24,17 +340,64 @@ Alternative methods if you don't like Github (also fully supported):
...
+
+To edit the message from the most recent commit run `git commit --amend`. To change older commits on the branch use `git rebase -i`. For a successful rebase have the branch track `upstream master`. Once the changes have been made and saved, run `git push --force origin <branch-name>`.
+
+## [Returning workflow](#returning-workflow)
+
+When returning to make a new patch, do the following to update your fork with the latest changes to master:
+
+```bash
+git fetch upstream
+git merge upstream/master
```
cgwalters: I'd recommend `git rebase upstream/master` instead.

docs/CONTRIBUTING.md
cgwalters: An even more opinionated path here is for us to just defer to the "Fedora CoreOS SDK".

rfairley:
@miabbott this sounds good to me. I agree on keeping the same sections in `CONTRIBUTING.md`, adding enhancements, and moving the walkthrough-related items (building and tutorial) to a different file.

docs/getting-started.md
```diff
@@ -0,0 +1,395 @@
...
+Make a branch from master for your patch.
+```bash
+$ git checkout -b <name-of-branch>
+$ git branch --set-upstream-to=upstream/master <name-of-branch>
+```
+
+## [Building OSTree](#building-ostree)
+
+### [Install Build Dependencies](#install-build-dependencies)
+
```
miabbott: I'd put a link in this section to the larger `build-dependencies.md` noting there are additional details around the dependencies in that doc.

docs/CONTRIBUTING.md
rfairley: Tried `git rebase upstream/master` and it works better than doing fetch/merge - changing to rebase.

docs/CONTRIBUTING.md
Bubblemelon: Thanks. Made amendments this section.

docs/CONTRIBUTING.md
```diff
@@ -26,15 +26,49 @@ Alternative methods if you don't like Github (also fully supported):
 It is likely however once a patch is ready to apply a maintainer
 will push it to a github PR, and merge via Homu.
 
-Commit message style
---------------------
+## [Commit message style](#commit-message-style)
```
jlebon: We don't have to "linkify" each header like this; GitHub already has a little "link" icon that pops up when you mouse over a header.

docs/CONTRIBUTING.md
```diff
@@ -26,15 +26,49 @@ Alternative methods if you don't like Github (also fully supported):
...
+1. Title
+    - Specify the context or category of the changes .e.g `lib` for library changes, `docs` for document changes, `bin/<command-name>` for command changes, etc.
+    - Begin the title with the first letter of the first word capitalized.
+    - Aim for less than 50 characters, otherwise 72 characters max.
+    - Do not end the title with a period.
+    - Use an [imperative tone](https://en.wikipedia.org/wiki/Imperative_mood).
+2. Body
+    - Separate the body with a blank line after the title.
+    - Begin a paragraph with the first letter of the first word capitalized.
+    - Each paragraph should have be within 72 characters.
```
jlebon: Extra "have" here.

docs/CONTRIBUTING.md
```diff
@@ -26,15 +26,49 @@ Alternative methods if you don't like Github (also fully supported):
...
+    - Content should be about what was changed and why this change was made.
+    - `Closes:` and `Aprroved by:` are autogenerated by `rh-atomic-bot`
```
jlebon: That's more or less an implementation detail. Maybe let's drop this here and in the template below?
One should still do `Closes: #N` if the commit fixes an issue.

docs/misc/build-dependencies.md
```diff
@@ -0,0 +1,199 @@
+# Ostree Build Dependencies
```
jlebon: Do we need this document? One shouldn't encounter this if using `dnf builddep ostree`, right?
Though it might be worth linking to the CI scripts for e.g. test dependencies.

docs/misc/ide-configuration.md
```diff
@@ -0,0 +1,13 @@
+# IDE Configuration
```
jlebon: This feels a bit too off-topic to store here. There are many *many* different potential "IDE" setups out there. It's likely that contributors will already have a preferred way of writing/navigating source code.

docs/CONTRIBUTING.md
Bubblemelon: see the new changes to this section.

docs/CONTRIBUTING.md
Bubblemelon: haha... sure! :laughing: It didn't occur to me that it sounded that way.

docs/getting-started.md
```diff
@@ -0,0 +1,395 @@
...
+$ sudo dnf install @buildsys-build
+```
+
+### [OSTree Build Commands](#ostree-build-commands)
+
+```bash
+git submodule update --init
+./autogen.sh --prefix=/usr --libdir=/usr/lib64 --sysconfdir=/etc
+./configure --prefix=/usr
+make
```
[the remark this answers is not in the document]
rfairley: Thanks, added a note in!

docs/CONTRIBUTING.md
rfairley: Addressed this.

docs/misc/build-dependencies.md
miabbott: I've run into some of the missing deps covered below, but I think it was because I never used `dnf -y install @buildsys-build`;  that seems to land most of the additional deps not covered by `dnf -y build-dep ostree`

docs/misc/ide-configuration.md
Bubblemelon: @jlebon In this case, I will remove this altogether. Do you use an IDE, and if so what it is? and how do you configure all the include paths?

Thanks! :smile:

jlebon: > Do you use an IDE, and if so what it is?

It depends on your definition of "IDE" :)
I use a combination of tmux + vim + universal-ctags + cscope (for cross-project definitions).

> how do you configure all the include paths?

In vim, you can press `gf` to open filenames under your cursor. I also use Command-T and sometimes NERDTree for arbitrary file exploration.

If you mean compiler flags, I just run `make` in a separate terminal. Though you can set up vim to do this with the `:make` command and e.g. navigate from error to error.

docs/CONTRIBUTING.md
Bubblemelon: Okay... noted. This was just in case, that we would make a table of contents on the top of this markdown file

docs/CONTRIBUTING.md
cgwalters: I withdraw this comment; it's fine to demonstrate this, and it can make a lot of sense to have different build containers per project.

docs/CONTRIBUTING.md
rfairley: Addressed this: https://github.com/ostreedev/ostree/pull/1694/commits/51a51264d4c049628435a8afdaa73d5fb536ea34. How does this look?

docs/CONTRIBUTING.md
```diff
@@ -50,21 +50,19 @@ You may use `Signed-off-by`, but we're not requiring it.
 2. Body
     - Separate the body with a blank line after the title.
     - Begin a paragraph with the first letter of the first word capitalized.
-    - Each paragraph should be within 72 characters.
+    - Each paragraph should be formatted within 72 characters.
     - Content should be about what was changed and why this change was made.
-    - `Closes:` and `Approved by:` are autogenerated by `rh-atomic-bot`
+    - If your commit fixes an issue, the commit message should end with `Closes: #<number>`.
+    - Adding `Approved by: ` to end the commit message is optional.
```
Bubblemelon: Isn't approved and closes generated by `rn-atomic bot` ?

docs/getting-started.md
```diff
@@ -219,37 +219,68 @@ $ docker run -it --rm --entrypoint /bin/sh --name ostree-testing ostree-fedora-t
...
+4. Now replace the original OSTree binary with the new one by running `sudo mv ~/ostree /usr/bin/ostree` and as well as the shared object file by running `sudo mv ~/libostree-1.so.1 usr/lib64/libostree-1.so.1`
 
-See this guide on creating an [Atomic Host Vagrant VM](https://gist.github.com/Bubblemelon/bd9f3cf429d9e04be52f9987baed53c8).
+If the `mv` command was successful, running `which ostree` shouldn't return an error. Proceed to test the new OSTree binary, by running the OSTree commands that you've changed.
+
+**Steps to `rsync` files to test an `ostree` build**:
+
+1. Copy your public ssh key from your host machine e.g. `id_rsa.pub` to `/home/vagrant/.ssh/authorized_keys` on the VM.
+
+2. Run `sudo su`, followed by `ssh localhost` then press <kbd>Ctrl</kbd>+<kbd>c</kbd> exit from the decision prompt. This will create the `.ssh` file with the right permissions.
```
Bubblemelon: *to exit

docs/getting-started.md
```diff
@@ -0,0 +1,463 @@
...
+Before a pull request is considered merge ready, your commit messages should fall within the specified guideline. See [Commit message style](CONTRIBUTING.md#commit-message-style).
+
+See [CONTRIBUTING.md](https://github.com/ostreedev/ostree/blob/master/docs/CONTRIBUTING.md#submitting-patches) for information on squashing commits, and alternative options to submit patches.
+
+### Returning Workflow
+
+When returning to make a new patch, do the following to update your fork with the latest changes to master:
+
+```bash
+git fetch upstream
```
miabbott: For this step, I usually do

```
$ git checkout master
$ git pull upstream master
```

Not sure if one is better than the other, though.

docs/CONTRIBUTING.md
```diff
@@ -14,17 +14,21 @@ for more information.  Instead, we use an instance of
 
 As a review proceeds, the preferred method is to push `fixup!`
 commits via `git commit --fixup`.  Homu knows how to use
-`--autosquash` when performing the final merge.  See the
+`--autosquash` when performing the final merge.
+
+Any commits committed with the `--fixup` option will have have the word `fixup!` in its commit title. This is to indicate that this particular commit will be squashed (by Homu when batch merging) with the commit that was specified in the command, `git commit --fixup <commit ref or hash>`.
```
jlebon: Minor: note that "batch merging" refers to something else -- it's when Homu combines multiple PRs into a single test to save time -- so that should just say e.g. "by Homu when merging"?

docs/CONTRIBUTING.md
```diff
@@ -35,6 +39,38 @@ similar to the
...
+    - Aim for less than 50 characters, otherwise 72 characters max.
+    - Do not end the title with a period.
+    - Use an [imperative tone](https://en.wikipedia.org/wiki/Imperative_mood).
+2. Body
+    - Separate the body with a blank line after the title.
+    - Begin a paragraph with the first letter of the first word capitalized.
+    - Each paragraph should be formatted within 72 characters.
+    - Content should be about what was changed and why this change was made.
+    - If your commit fixes an issue, the commit message should end with `Closes: #<number>`.
+    - Adding `Approved by: ` to end the commit message is optional.
```
jlebon: Minor: we can get rid of this line too; the `Approved by:` line is completely transparent to contributors.

docs/getting-started.md
```diff
@@ -0,0 +1,461 @@
...
+
+```bash
+dnf install @buildsys-build dnf-plugins-core && \
+dnf builddep ostree
+```
+
+For CentOS:
+
+```bash
+yum install @buildsys-build dnf-plugins-core && \
```
jlebon: s/dnf-plugins-core/yum-utils/

CentOS unfortunately doesn't have `@buildsys-build` -- see: https://github.com/ostreedev/ostree/blob/521e0ec3ac87b539190f3a6cc310237fbaf39517/ci/libpaprci/libbuild.sh#L39

docs/getting-started.md
```diff
@@ -0,0 +1,461 @@
...
+
+There are a variety of container engines available and many distributions have pre-packaged versions of e.g. [Podman](https://github.com/projectatomic/libpod) and Docker.
+
+Follow this installation guide for [Docker upstream](https://docs.docker.com/v17.09/engine/installation/linux/docker-ce/fedora/) and you may want to follow this [post-installation guide for Docker](https://docs.docker.com/v17.09/engine/installation/linux/linux-postinstall/) if you'd like to run Docker as a non-root user on a Linux based host machine.
+
+The example below is using Docker for virtualization. Save the contents of this **Dockerfile** somewhere on your machine:
+
+```bash
+# this pulls the latest fedora image
+FROM fedora:latest
```
jlebon: Let's use the Fedora registry: `registry.fedoraproject.org/fedora:28`

docs/misc/build-dependencies.md
jlebon: @Bubblemelon or @rfairley can you confirm that this file is no longer necessary if one does `dnf install -y @buildsys && dnf builddep -y ostree` as instructed in the other document?

docs/getting-started.md
Bubblemelon: Thanks for pointing it out.

I supposed CentOS would still need `dnf-plugins-core`, since this will allow [builddep](https://centos.pkgs.org/7/puias-unsupported-x86_64/dnf-plugins-core-0.1.5-3.sdl7.noarch.rpm.html) to run. 

Let me know if I should keep `dnf-plugins-core` as one of the commands to run for CentOS, or if I should specify the individual packages as stated in libbuild.sh and remove the directions to install `dnf-plugins-core` and the `yum-builddep ostree` command.

jlebon: Right, that line would be:

```
yum install yum-utils <list of base pkgs in libbuild.sh>
```

`yum builddep` is provided by `yum-utils` so you don't need `dnf-plugins-core`.

docs/getting-started.md
```diff
@@ -219,37 +219,68 @@ $ docker run -it --rm --entrypoint /bin/sh --name ostree-testing ostree-fedora-t
...
-### [Testing in a Virtual Machine](#testing-in-a-virtual-machine)
+3. To override the `Read-only file system` warning, run `ostree admin unlock`
 
-**To-do**: use rsync or pull on the vm ! 
+4. Now replace the original OSTree binary with the new one by running `sudo mv ~/ostree /usr/bin/ostree` and as well as the shared object file by running `sudo mv ~/libostree-1.so.1 usr/lib64/libostree-1.so.1`
 
-See this guide on creating an [Atomic Host Vagrant VM](https://gist.github.com/Bubblemelon/bd9f3cf429d9e04be52f9987baed53c8).
+If the `mv` command was successful, running `which ostree` shouldn't return an error. Proceed to test the new OSTree binary, by running the OSTree commands that you've changed.
+
+**Steps to `rsync` files to test an `ostree` build**:
```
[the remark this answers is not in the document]
rfairley: Tested the rsync section and it works!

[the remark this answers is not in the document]
rfairley: Will test the scp section.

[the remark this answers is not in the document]
Bubblemelon: okay thank you :smile:

docs/getting-started.md
```diff
@@ -0,0 +1,439 @@
...
+
+## [Testing a Build](#testing-a-build)
+
+It is best practice to build software (definitely including ostree) in a container or virtual machine first.
+
+### Testing in a Container
+
+There are a variety of container engines available and many distributions have pre-packaged versions of e.g. [Podman](https://github.com/projectatomic/libpod) and Docker.
+
+Follow this installation guide for [Docker upstream](https://docs.docker.com/v17.09/engine/installation/linux/docker-ce/fedora/) and you may want to follow this [post-installation guide for Docker](https://docs.docker.com/v17.09/engine/installation/linux/linux-postinstall/) if you'd like to run Docker as a non-root user on a Linux based host machine.
```
cgwalters: Let's prefix this with "If you choose to use Docker upstream, see..."

docs/getting-started.md
```diff
@@ -0,0 +1,439 @@
...
+
+The example below is using Docker for virtualization. Save the contents of this **Dockerfile** somewhere on your machine:
```
cgwalters: s/for virtualization/to manage containers/

docs/getting-started.md
```diff
@@ -0,0 +1,439 @@
...
+2. Run `sudo su`, followed by `ssh localhost` then press <kbd>Ctrl</kbd>+<kbd>c</kbd> to exit from the decision prompt. This will create the `.ssh` file with the right permissions.
+
+3. Using `Vagrant` as the user, run `sudo cp ~/.ssh/authorized_keys /root/.ssh/`. So that user `root` has the the same login credentials.
+
+4. `<ostree-install-dir>` will serve as the local install location for `ostree` and the path to this directory should be _absolute_ when specified in `DESTDIR`.
+
+5. Run the following commands to build OSTree for Atomic Host on your host machine:
+
+        ./autogen.sh
+        ./configure --prefix=/usr
```
cgwalters: First, you *absolutely need* to match `--libdir` which on Fedora is `/usr/lib64` for 64 bit hosts.  It's different for Debian (this is a tricky aspect of all of this...it's a bit sad there's not a portable "configure like the host system" wrapper).  I think meson solves this by having the right defaults.

It's also pretty important to specify `--sysconfdir=/etc`.

It's probably simpler to just recommend `./autogen.sh --prefix=/usr --libdir=/usr/lib64 --sysconfdir=/etc`; it'll be wrong for Debian but oh well.

docs/getting-started.md
rfairley: Thinking about it, the section could include two use cases like so:

If creating a new branch:

```
$ git checkout master
$ git pull upstream master
$ git checkout -b <name-of-patch>
```

If continuing on a branch already created:

```
$ git checkout <name-of-patch>
$ git pull --rebase upstream master
```

(may as well avoid separating the fetch, and no need to `checkout -b` at the end in the case of an already existing feature branch)

@cgwalters suggested rebasing initially. I like the rebase way to update existing branches and avoid interleaving commits - and it saves master from getting interleaved commits if there were local commits made to the forked master. However rebasing is not necessary to update a forked master and branch out - merge works fine then.

Will add in the edit for the two use cases above, if that looks good.

docs/getting-started.md
Bubblemelon: Hmm.. funny, I've built ostree master using those commands and it seems to work fine. Also tested some of ostree's commands. Could you give me an example of what might not work, when built without those flags?

cgwalters: If you change code that ends up in the CLI (`/usr/bin/ostree`) as this guide shows it'll work.  If you change code that ends up in the shared library (e.g. `src/libostree/ostree-repo.c`), your changes won't appear because you'll be creating a new `/usr/lib/libostree-1.so.1.0.0` instead of overwriting the `/usr/lib64/libostree-1.so.1.0.0` .

cgwalters: Aside: You might wonder "what's up with this libdir stuff, why isn't it /usr/lib"?  It's basically that Fedora chose one way many years ago (`lib64`) which is known as "multilib" (solving the 32 and 64 bit parallel install), but Debian took a few more years and implemented a [more general system called multiarch](https://wiki.debian.org/Multiarch/HOWTO).

jlebon:
Let's make that a separate PR once this one is done?

Re. layout -- no strong opinion. I like the way you have it in your demo, though maybe renaming "Getting Started Guide" to "Contributing Tutorial" to make it clear it's about developing ostree, not using it?

rfairley:
@jlebon :+1: sounds good to deal with it separately in its own PR. Agree on changing the name - Getting Started Guide is vague. It'd make sense to also rename this file to `contributing-tutorial.md` and retitle it as `OSTree Contributing Tutorial` - that ties it together IMO.

rfairley:
For now `CONTRIBUTING.md` will link to the markdown page as such: https://github.com/ostreedev/ostree/pull/1694/files#diff-8595297407c24b9ccb87b2a698cca9c2R174. Will change the hyperlink once the file is renamed.

docs/getting-started.md
```diff
@@ -134,7 +136,7 @@ RUN ../ostree-git.sh
 
 # builds ostree + any additional commands
 COPY ostree-test.sh /
-RUN ../ostree-test.sh
+RUN cd ostree && ../ostree-test.sh
```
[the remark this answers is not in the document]
rfairley: After `ostree-git.sh`, the Dockerfile commands still execute from the directory `/`, and the git commands in `ostree-test.sh` expect to be in the `ostree` repo. So a `cd ostree` is needed somewhere before the git commands in `ostree-test.sh`, but I think it should be outside of that script because we want to run the script from the `/ostree` (see [this](https://github.com/ostreedev/ostree/pull/1694/files/b6dc162c239c701beb89377a341370c93db87e1b..8f99b354750e6b8fdffe8d4b4a5c05f6761f5222#diff-a39c9f7aa728d5fa3b973bc6ba49228aR245)).

Just noticed at `/`, `.` and `..` are the same...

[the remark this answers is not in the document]
Bubblemelon: Oh I could just move `WORKDIR /ostree` on top of  `RUN ../ostree-test.sh`. I didn't notice this. Thanks for pointing it out.

docs/getting-started.md
```diff
@@ -0,0 +1,449 @@
...
+
+To **start** the `ostree-fedora-test` container, run:
+
+```bash
+$ docker run -it --rm --entrypoint /bin/sh --name ostree-testing ostree-fedora-test
+```
+
+**Note**:
+
+`--rm` option removes the name of the container i.e. `ostree-testing` whenever the container's shell prompt `#` exits. This is so that the container name `ostree-testing` can be re-used, otherwise remove it manually by running `docker rm <container name>`.
```
miabbott: To be a bit more precise, the `--rm` flag cleans up the container and container file system when the container exits.

https://docs.docker.com/engine/reference/run/#clean-up---rm

Bubblemelon:
I believe that @rfairley checked how the docs renders on readthedocs.io 

So other than the comment that Micah made, all is left is: squashing the the commits, one for CONTRIBUTING.md and one for getting-started.md (or some other name for it - but seems that this titled will be changed in a different PR)

rfairley:
Yes - checked and this should render fine on readthedocs. The actual adding of `contributing-tutorial.md` to the readthedocs pages will be done in a separate PR (per https://github.com/ostreedev/ostree/pull/1694#issuecomment-411876186).

ashcrow:
This is FANTASTIC!

cgwalters:
Thanks a lot for this!  Nice work!

@rh-atomic-bot r+ b24731c

PR #1697, opened 2018-07-31
ostree-remount.service: RemainAfterExit=yes

cgwalters:
This is standard practice for units like this; e.g. it's what
`systemd-remount-fs.service` does.  I think it may be part of
or the whole cause for
https://github.com/projectatomic/rpm-ostree/issues/1471

I haven't reproduced the problem exactly but it seems to me that
if the unit starts and is GC'd, then when systemd goes to execute
a later unit it might end up restarting it.

A noticeable side effect of this is that `systemctl status ostree-remount`
exits with code `0` as expected.

### DIFF
--- a/src/boot/ostree-remount.service
+++ b/src/boot/ostree-remount.service
@@ -31,6 +31,7 @@ Before=systemd-tmpfiles-setup.service
 
 [Service]
 Type=oneshot
+RemainAfterExit=yes
 ExecStart=/usr/lib/ostree/ostree-remount
 StandardInput=null
 StandardOutput=syslog


dustymabe:
this is something I should be able to test on my failing system, right? would I need to make the change and then enable initrfamfs generation in order to get the change to propagate into the initrfd?

cgwalters:
> and then enable initrfamfs generation in order to get the change to propagate into the initrfd?

Don't need to do that, this runs in the main system.  You can just `systemctl edit ostree-remount.service` and add
```
[Service]
RemainAfterExit=yes
```

jlebon:
Good catch! This makes sense to me regardless of its (likely) role in https://github.com/projectatomic/rpm-ostree/issues/1471. (Though cross-posting here too: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=750683 indicates that oneshot services that are depended upon can indeed  get executed multiple times if without `RemainAfterExit=yes`.)

@rh-atomic-bot r+ 6bc5f3d

dustymabe:
> Don't need to do that, this runs in the main system. You can just systemctl edit ostree-remount.service

:+1: 

This seems to have resolved the problem for me. If I add `RemainAfterExit=yes` then I can boot. If I take it away again I can't boot.

PR #1698, opened 2018-08-02
Fix corrupted summary cache

dbnicholson:
If the summary from the cache is used but it doesn't match the signature, fetch it again. While c4c2b5eb helps you from corrupting the summary cache, this helps you get out of it if you happen to get into that situation.

This also adds a couple tests for racing with remote summary updates that I wrote while trying to get to the bottom of this. We were still occasionally seeing this on our image builder, but I believe that's because it uses a cache repo that's used by different versions of ostree depending on what's being built.

This might fix flatpak/flatpak#1450.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -4037,10 +4037,20 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
             if (summary_from_cache)
               {
                 /* The cached summary doesn't match, fetch a new one and verify again */
+                if ((self->test_error_flags & OSTREE_REPO_TEST_ERROR_INVALID_CACHE) > 0)
+                  {
+                    g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                                 "Remote %s cached summary invalid and "
+                                 "OSTREE_REPO_TEST_ERROR_INVALID_CACHE specified",
+                                 pull_data->remote_name);
+                    goto out;
+                  }
+                else
+                  g_debug ("Remote %s cached summary invalid, pulling new version",
+                           pull_data->remote_name);
+
                 summary_from_cache = FALSE;
                 g_clear_pointer (&bytes_summary, (GDestroyNotify)g_bytes_unref);
-                g_debug ("Remote %s cached summary invalid, pulling new version",
-                         pull_data->remote_name);
                 if (!_ostree_fetcher_mirrored_request_to_membuf (pull_data->fetcher,
                                                                  pull_data->meta_mirrorlist,
                                                                  "summary",
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3986,7 +3986,10 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       goto out;
 
     if (bytes_summary)
-      summary_from_cache = TRUE;
+      {
+        g_debug ("Loaded %s summary from cache", remote_name_or_baseurl);
+        summary_from_cache = TRUE;
+      }
 
     if (!pull_data->summary && !bytes_summary)
       {
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -4027,12 +4027,43 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
     if (pull_data->gpg_verify_summary && bytes_summary && bytes_sig)
       {
         g_autoptr(OstreeGpgVerifyResult) result = NULL;
+        g_autoptr(GError) temp_error = NULL;
 
         result = ostree_repo_verify_summary (self, pull_data->remote_name,
                                              bytes_summary, bytes_sig,
-                                             cancellable, error);
-        if (!ostree_gpg_verify_result_require_valid_signature (result, error))
-          goto out;
+                                             cancellable, &temp_error);
+        if (!ostree_gpg_verify_result_require_valid_signature (result, &temp_error))
+          {
+            if (summary_from_cache)
+              {
+                /* The cached summary doesn't match, fetch a new one and verify again */
+                summary_from_cache = FALSE;
+                g_clear_pointer (&bytes_summary, (GDestroyNotify)g_bytes_unref);
+                g_debug ("Remote %s cached summary invalid, pulling new version",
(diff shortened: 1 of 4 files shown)


pwithnall:
This looks pretty awesome to me. I can’t spot any problems with the code, and the tests are excellent.

r+ from me, but I’d like a second opinion since I’ve fallen out of the regular code review loop.

src/libostree/ostree-repo-pull.c
```diff
@@ -4027,12 +4027,44 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
-        if (!ostree_gpg_verify_result_require_valid_signature (result, error))
-          goto out;
+        if (!ostree_gpg_verify_result_require_valid_signature (result, &temp_error))
+          {
+            if (summary_from_cache)
+              {
+                /* The cached summary doesn't match, fetch a new one and verify again */
+                summary_from_cache = FALSE;
+                g_clear_pointer (&bytes_summary, (GDestroyNotify)g_bytes_unref);
+                g_warning ("Remote %s cached summary invalid, pulling new version",
```
cgwalters: See comment in https://github.com/ostreedev/ostree/pull/1685#discussion_r204052444

dbnicholson: I added it because it should really never happen anymore (and the test suite is run with `G_DEBUG=fatal-warnings`). The idea is that if we do start seeing a constant stream of warnings containing this then a new bug has been introduced in ostree that we probably want to fix instead of silently ignoring the summary cache all the time and making it useless.

That said, it is your project so I can drop it or change it to `g_debug`.

pwithnall: I generally think the choice should be between `GError**` and `g_assert_not_reached()`. Since the control flow here is affected by external data, `g_assert_not_reached()` is not a good idea, so this should probably be a `GError**`.

dbnicholson: @pwithnall but in this case it's not an error since it's being recovered from. A returned `GError` would likely never be seen.

I'll just change it to a `g_debug`. Hopefully the test suite will catch if new errors are introduced in the summary cache handling.

cgwalters:
> I'll just change it to a g_debug.

It's also OK by me to add special environment variables just for the test suite - see e.g.
`src/libostree/ostree-repo-commit.c:  if ((self->test_error_flags & OSTREE_REPO_TEST_ERROR_PRE_COMMIT) > 0)`

tests/test-pull-summary-sigs.sh
```diff
@@ -151,4 +151,82 @@ grep static-deltas summary.txt > static-deltas.txt
...
+cmp repo/tmp/cache/summaries/origin ${test_tmpdir}/ostree-srv/gnomerepo/summary.1 >&2
+cmp repo/tmp/cache/summaries/origin.sig ${test_tmpdir}/ostree-srv/gnomerepo/summary.sig.1 >&2
+
+# Simulate a pull race where the client gets the old summary and the new
+# summary signature since it was generated on the server between the
+# requests
+cp ${test_tmpdir}/ostree-srv/gnomerepo/summary.sig{.2,}
+if ${OSTREE} --repo=repo pull origin main; then
+    assert_not_reached "Successful pull with old summary"
+fi
```
jlebon: Minor: would be nice to capture the error message and do an `assert_file_has_content` here to make sure it's the error we expect.

src/libostree/ostree-repo-pull.c
```diff
@@ -4027,12 +4027,44 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
+                g_autoptr(OstreeGpgVerifyResult) retry =
+                  ostree_repo_verify_summary (self, pull_data->remote_name,
+                                              bytes_summary, bytes_sig,
+                                              cancellable, error);
+                if (!ostree_gpg_verify_result_require_valid_signature (retry, error))
+                  goto out;
+              }
+            else
+              {
+                if (temp_error != NULL)
```
jlebon: Hmm, if `temp_error` is `NULL` at this point, then there's something wrong. Otherwise, it means there's a chance we could return `FALSE` with no error set at all.

IOW, let's just drop this and let the `g_propagate_error` precondition check yell at us or alternatively make it a `g_assert`?

tests/test-pull-summary-sigs.sh
dbnicholson: Sure, should have done that. It's equivalent to an earlier test that does check for the appropriate message, but I can repeat it here.

src/libostree/ostree-repo-pull.c
dbnicholson: What happens (I believe, need to go back and check) is that the return from `ostree_repo_verify_summary` isn't checked and just passed into `ostree_gpg_verify_result_require_valid_signature`, which immediately returns `FALSE` if the result is `NULL` without setting an error.

I started looking at either checking the result of `ostree_repo_verify_summary`, but it was making the code a little hairy. It would probably work fine if I passed in `&temp_error` to `ostree_repo_verify_summary`, though, since `ostree_gpg_verify_result_require_valid_signature` will just return immediately without setting the error.

dbnicholson:
Comments addressed. In particular:

* `g_debug` instead of `g_warning` when the invalid cache is detected in the default case
* Pass `&temp_error` to the first `ostree_repo_verify_summary` so that errors there are propagated if the summary isn't cached
* Checked the output from the expected to fail pull with old summary on the remote
* Added an `invalid-cache` test error flag and throw an error when the invalid summary cache is detected to make sure that code is hit

Note that there are some fixups in there to squash and the commit message for "Fetch summary if cached version doesn't match" is now wrong since it talks about emitting a warning. I don't know how the autosquashing goes here exactly, but I'm happy to do the squashing myself and repush after review.

jlebon:
Nice, this looks good to me.

> Note that there are some fixups in there to squash and the commit message for "Fetch summary if cached version doesn't match" is now wrong since it talks about emitting a warning. I don't know how the autosquashing goes here exactly, but I'm happy to do the squashing myself and repush after review.

Yeah, feel free to (squash and) repush with the amended commit messages.

dbnicholson:
Squashed and pushed again. Ready for review.

cgwalters:
Thanks!  Looks great.

@rh-atomic-bot r+ 1429831

PR #1700, opened 2018-08-13
Fix leak in ostree_repo_list_collection_refs

alexlarsson:
We need to have the g_auto(GLnxDirFdIterator) inside the loop, or
we don't correctly clean up when iterating several times.


cgwalters:
@rh-atomic-bot r+ 4d35170

jlebon:
@rh-atomic-bot retry

PR #1701, opened 2018-08-13
ostree_repo_static_delta_generate: Fix leak

alexlarsson:
There is no need to ref the argument of g_variant_builder_add_value


jlebon:
Thanks!
@rh-atomic-bot r+ f29883b

PR #1702, opened 2018-08-14
ostree_repo_pull_from_remotes_async: Fix leak of options

alexlarsson:
copy_option() unnecessarily passed ownership of the value
to g_variant_dict_insert_value, but that already refs, so it was leaked.


cgwalters:
@rh-atomic-bot r+ 5ddb638

Issue #1703, opened 2018-08-14
Build fails with `mkdir: cannot create directory ‘tests’: File exists`


falstaff84:
Using OpenEmbedded the following issue appeared:

```
ERROR: ostree-native-v2017.13-2 do_compile: oe_runmake failed
ERROR: ostree-native-v2017.13-2 do_compile: Function failed: do_compile (log file is located at /workdir/oe/tmp/work/x86_64-linux/ostree-native/v2017.13-2/temp/log.do_compile.114170)
ERROR: Logfile of failure stored in: /workdir/oe/tmp/work/x86_64-linux/ostree-native/v2017.13-2/temp/log.do_compile.114170
...
| ln -fns ../.libs/libreaddir-rand.so tests
| /workdir/oe/tmp/hosttools/mkdir: cannot create directory ‘tests’: File exists
| /workdir/oe/tmp/hosttools/mkdir: cannot create directory ‘tests’: File exists
| Makefile:3746: recipe for target 'tests/.dirstamp' failed
```

It seems to only happen sometimes... On that particular server make run with `-j 32`, so this seems to be some kind of race condition.

PR #1704, opened 2018-08-14
Avoid race condition in case tests directory does not exist

falstaff84:
Make sure the tests directory exists before symlinking files
into it.

Closes: #1703


cgwalters:
@rh-atomic-bot r+ ae12ae6

PR #1706, opened 2018-08-17
man/ostree.repo-config: improve min-free-space-* docs

wjt:
0 is not a valid value (the units are required), and 1MB is taken to
mean 1048576 bytes, not 1000000 bytes (unlike other human-readable sizes
in libostree, which are formatted by g_format_size(), and hence use
power-of-10 units). It's probably a bit late to change this latter
point, but the documentation should mention it.

Define 'MB' etc. more precisely; include the byte counts in the
examples; and improve the formatting of the min-free-space-* paragraphs.

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -129,24 +129,36 @@ Boston, MA 02111-1307, USA.
 
       <varlistentry>
         <term><varname>min-free-space-percent</varname></term>
-        <listitem><para>Integer percentage value (0-99) that specifies a minimum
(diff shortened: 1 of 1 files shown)


wjt:
Attn @mwleeds since you also adjusted these docs in #1687, and @uajain as the author of this feature.

jlebon:
Thanks!
@rh-atomic-bot r=mwleeds 3253ac8

cgwalters:
@rh-atomic-bot retry

PR #1708, opened 2018-08-18
ci: Bump rpm-ostree tag we build for tests

jlebon:
The latest rpm-ostree release no longer requires `python3-devel` so
`dnf builddep` here is no longer pulling it in, subsequently causing issues
when building an older rpm-ostree release. Let's just bump the release
tag so we don't have to also start keeping track of older dependencies.


cgwalters:
@rh-atomic-bot r+ 82caa7f

PR #1709, opened 2018-08-20
lib/repo-pull: Prefer object pull over from-scratch delta if ref exists

pwithnall:
If a ref already exists, we are likely only a few commits behind the
current head of the ref, so it is probably better for bandwidth
consumption to pull the individual objects rather than the from-scratch
delta.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3348,6 +3348,19 @@ initiate_request (OtPullData                 *pull_data,
       return TRUE;
     }
 
+  /* If doing a delta from a ref, look up the from-revision, since we need it
+   * on most paths below. */
+  if (ref != NULL)
+    {
+      g_autofree char *refspec = NULL;
+      if (pull_data->remote_name != NULL)
+        refspec = g_strdup_printf ("%s:%s", pull_data->remote_name, ref->ref_name);
+      if (!ostree_repo_resolve_rev (pull_data->repo,
+                                    refspec ?: ref->ref_name, TRUE,
+                                    &delta_from_revision, error))
+        return FALSE;
+    }
+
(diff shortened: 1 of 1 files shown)


pwithnall:
This is slightly related to #470, in the sense that it tries to address the same issue that client-side chained deltas would, but with a more basic approach.

The idea is that if a ref exists locally, the client is probably only one or two commits out of date on that branch, and since software doesn’t change fast, pulling only the out of date objects has probably a significantly lower bandwidth cost than pulling the entire from-scratch delta.

Apparently Alex has already applied this patch to the Ubuntu flatpak PPA. We’ve been carrying it in Endless for several months, but hadn’t upstreamed it before due to wondering if chained deltas would be a better solution. I still think chained deltas *would* be a better solution, but in the name of not letting the perfect be the enemy of the good, let’s make this incremental improvement first.

cgwalters:
My main concern with this is that it'll break anyone who is doing "delta only" repositories by hand.  Previously, I think it'd entirely avoid object fetches to always have the latest tip as from-scratch, with however many deltas to it one wanted.

That said...while we've talked about delta-only approaches, I am not sure if anyone is actually doing that today.

cgwalters:
@rh-atomic-bot r+ 6cd2c55

ramcq:
@cgwalters Speaking with @alexlarsson about this, we were pondering the "delta-only repo" but until we have a better way to manage/retain them (atm, flatpak only keeps one delta, and ostree will only do a single "hop" - it would be better to keep a round robin until we reach some % of the scratch delta size, and have ostree follow chains) this seemed a more reasonable heuristic.

Relatedly, deltas are going to bloat the summary a lot if we keep multiple per ref - it might make sense to look at adding deltas as commit meta, so they are only fetched if you're actually digging in to a specific ref.

PR #1710, opened 2018-08-20
Minor follow-ups to ostree config work

jlebon:
See individual commit messages for details.

### DIFF
--- a/man/ostree-config.xml
+++ b/man/ostree-config.xml
@@ -51,7 +51,13 @@ Boston, MA 02111-1307, USA.
 
     <refsynopsisdiv>
             <cmdsynopsis>
+                <command>ostree config get</command> <arg choice="req">SECTIONNAME.KEYNAME</arg>
+            </cmdsynopsis>
+            <cmdsynopsis>
                 <command>ostree config get</command> <arg choice="req"> --group=GROUPNAME</arg> <arg choice="req"> KEYNAME</arg>
+            </cmdsynopsis>
+            <cmdsynopsis>
+                <command>ostree config set</command> <arg choice="req">SECTIONNAME.KEYNAME</arg> <arg choice="req">VALUE</arg>
             </cmdsynopsis>
             <cmdsynopsis>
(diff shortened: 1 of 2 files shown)


src/ostree/ot-builtin-config.c
```diff
@@ -88,7 +88,6 @@ ostree_builtin_config (int argc, char **argv, OstreeCommandInvocation *invocatio
 
   if (!strcmp (op, "set"))
     {
-      printf("GROUP NUMBER = %s  %d\n", opt_group, argc);
```
sinnykumari: Damm! Stupid me. Thanks!

man/ostree-config.xml
```diff
@@ -68,7 +74,8 @@ Boston, MA 02111-1307, USA.
 
     <refsect1>
         <title>Example</title>
-        <para><command>$ ostree config get --group=core mode</command></para>
+        <para><command>$ ostree config get core.mode</command></para>
```
sinnykumari: Do we want user to encourage using dots in future? If yes, then +1.

man/ostree-config.xml
```diff
@@ -50,9 +50,15 @@ Boston, MA 02111-1307, USA.
     </refnamediv>
 
     <refsynopsisdiv>
+            <cmdsynopsis>
+                <command>ostree config get</command> <arg choice="req">SECTIONNAME.KEYNAME</arg>
```
sinnykumari: Do we want to encourage users to us dots in future? I thought we will keep existing Dot usecase to maintain backward compatibility and we suggest future usage with using --group option.

jlebon: Hmm yeah, I think it should be fine to support/encourage it. It's really just the `remote "foobar"` case that makes things annoying. In other situations, I'd much rather type `ostree config get core.config` instead of `ostree config get --group=core config`.

src/ostree/ot-builtin-config.c
jlebon: Not at all! This happens to everyone. :)

cgwalters:
@rh-atomic-bot r=sinnykumari 7ae880b

PR #1714, opened 2018-08-29
lib/grub2: Support Debian-style grub.cfg path

fkrull:
In contrast to... at least Fedora, Debian and Debian-derived systems keep their GRUB configuration file on BIOS systems in `/boot/grub/grub.cfg`, rather than `/boot/grub2/grub.cfg`. This patch changes the GRUB detection to look for either of these parts so that GRUB bootloader generation works as expected when doing an `ostree admin deploy` on a Debian system.


cgwalters:
Thanks for the patch!  Offhand, I think this is OK as is, though it might be better as a build-time option?  Do you have the autotools-fu to do that?

fkrull:
A build-time option would also be fine. But adding an option for such a simple check seems overkill to me?

If you prefer the compile option, I can change it. I can probably figure out how to do that with autotools.

cgwalters:
Yeah, not worth redoing it.  I just pushed a quick `fixup!` though that adds a comment as to where each version can be found.  Thanks for the patch!

@rh-atomic-bot r+ d162ea4

PR #1715, opened 2018-08-30
repo: Add an API to get min-free-space-* reserved bytes

uajain:
https://phabricator.endlessm.com/T23694

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1536,23 +1536,28 @@ devino_cache_lookup (OstreeRepo           *self,
   return dev_ino_val->checksum;
 }
 
-static guint64
-min_free_space_calculate_reserved_blocks (OstreeRepo *self, struct statvfs *stvfsbuf)
+static gboolean
+min_free_space_calculate_reserved_blocks (OstreeRepo *self, struct statvfs *stvfsbuf, GError **error)
 {
-  guint64 reserved_blocks = 0;
+  self->reserved_blocks = 0;
 
   if (self->min_free_space_mb > 0)
     {
-      reserved_blocks = (self->min_free_space_mb << 20) / stvfsbuf->f_bsize;
+      if (self->min_free_space_mb > (G_MAXUINT64 >> 20) ||
+          self->txn.blocksize > (1 << 20))
+        return glnx_throw (error, "min-free-space value is greater than the maximum allowed value of %" G_GUINT64_FORMAT " bytes",
+                           G_MAXUINT64 / stvfsbuf->f_bsize);
+
+      self->reserved_blocks = (self->min_free_space_mb << 20) / self->txn.blocksize;
     }
   else if (self->min_free_space_percent > 0)
     {
       /* Convert fragment to blocks to compute the total */
       guint64 total_blocks = (stvfsbuf->f_frsize * stvfsbuf->f_blocks) / stvfsbuf->f_bsize;
-      reserved_blocks = ((double)total_blocks) * (self->min_free_space_percent/100.0);
+      self->reserved_blocks = ((double)total_blocks) * (self->min_free_space_percent/100.0);
     }
 
-  return reserved_blocks;
+  return TRUE;
 }
 
 /**
@@ -1650,11 +1655,16 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
 
   g_mutex_lock (&self->txn_lock);
   self->txn.blocksize = stvfsbuf.f_bsize;
-  guint64 reserved_blocks = min_free_space_calculate_reserved_blocks (self, &stvfsbuf);
+  if (!min_free_space_calculate_reserved_blocks (self, &stvfsbuf, error))
+    {
+      g_mutex_unlock (&self->txn_lock);
+      return FALSE;
+    }
+
   /* Use the appropriate free block count if we're unprivileged */
   guint64 bfree = (getuid () != 0 ? stvfsbuf.f_bavail : stvfsbuf.f_bfree);
-  if (bfree > reserved_blocks)
-    self->txn.max_blocks = bfree - reserved_blocks;
+  if (bfree > self->reserved_blocks)
+    self->txn.max_blocks = bfree - self->reserved_blocks;
   else
     {
       self->cleanup_stagedir = TRUE;
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3293,6 +3293,26 @@ ostree_repo_get_mode (OstreeRepo  *self)
 }
 
 /**
+ * ostree_repo_get_min_free_space:
+ * @self: Repo
+ *
+ * It should be noted that this function should be used only if there
+ * is a transaction active. It is a programmer error to request it
+ * otherwise.
+ *
+ * Returns: Value (in bytes) of min-free-space-* config option
+ * Since: 2018.9
+ */
+guint64
+ostree_repo_get_min_free_space_bytes (OstreeRepo  *self)
+{
+  g_return_val_if_fail (OSTREE_IS_REPO (self), 0);
+  g_return_val_if_fail (self->in_transaction == TRUE, 0);
+
+  return self->reserved_blocks * self->txn.blocksize;
+}
+
+/**
  * ostree_repo_get_parent:
  * @self: Repo
  *
--- a/tests/test-repo.c
+++ b/tests/test-repo.c
@@ -157,24 +157,31 @@ test_repo_get_min_free_space (Fixture *fixture,
 {
   g_autoptr (GKeyFile) config = NULL;
   g_autoptr(GError) error = NULL;
+  typedef struct
+    {
+      const char *val;
+      gboolean should_succeed;
+    } min_free_space_value;
+
   g_autoptr(OstreeRepo) repo = ostree_repo_create_at (fixture->tmpdir.fd, ".",
                                                       OSTREE_REPO_MODE_ARCHIVE,
                                                       NULL,
                                                       NULL, &error);
   g_assert_no_error (error);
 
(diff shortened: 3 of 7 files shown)


src/libostree/ostree-repo.c
```diff
@@ -3292,6 +3292,28 @@ ostree_repo_get_mode (OstreeRepo  *self)
...
+ * Returns: Value(in bytes) of min-free-space-* config option
+ */
+guint64
+ostree_repo_get_min_free_space (OstreeRepo  *self, GError **error)
+{
+  g_return_val_if_fail (OSTREE_IS_REPO (self), 0);
+
+  g_assert (self->txn.blocksize != 0);
+  if (self->reserved_blocks > G_MAXUINT64 / self->txn.blocksize)
+    return glnx_throw (error, "min-free-space(in bytes) value was too high");
```
pwithnall: It seems a bit odd that the code in `ostree_repo_prepare_transaction()` doesn’t check this as well. (I realise it’s because that code operates in terms of blocks rather than bytes.) It would be weird for a transaction to happily go ahead, but then return an error only when you call `ostree_repo_get_min_free_space()`.

src/libostree/ostree-repo.c
```diff
@@ -3292,6 +3292,28 @@ ostree_repo_get_mode (OstreeRepo  *self)
...
+/**
+ * ostree_repo_get_min_free_space:
+ * @self: Repo
+ * @error: error
+ *
+ * Before this function can be used, ostree_repo_prepare_transaction() must have been
+ * called.
+ *
+ * Returns: Value(in bytes) of min-free-space-* config option
+ */
```
pwithnall: The documentation needs a `Since: 2018.x` line (fill in `x` as appropriate).

src/libostree/ostree-repo.c
```diff
@@ -3292,6 +3292,28 @@ ostree_repo_get_mode (OstreeRepo  *self)
   return self->mode;
 }
 
+/**
+ * ostree_repo_get_min_free_space:
+ * @self: Repo
+ * @error: error
+ *
+ * Before this function can be used, ostree_repo_prepare_transaction() must have been
+ * called.
```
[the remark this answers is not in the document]
uajain: Oh yeah, How did I miss this. Thank you.

Thinking on this, we want to only read the value of `min-free-space-*` only when there is an ongoing transaction (`repo->in_transaction`) because that's where the config option is actually computed. If we read outside (e.g. after `ostree_repo_commit_transaction`) it could be inconsistent as the config option might get changed and one will still be reading the old value.

src/libostree/ostree-repo.c
uajain: hmm, this makes sense but I think let's file a separate issue and PR for that?

pwithnall: I suggest it’s handled here, because there is no inconsistency before this PR is merged, and there is an inconsistency afterwards.

src/libostree/ostree-repo-commit.c
```diff
@@ -1650,11 +1650,18 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
 
   g_mutex_lock (&self->txn_lock);
   self->txn.blocksize = stvfsbuf.f_bsize;
-  guint64 reserved_blocks = min_free_space_calculate_reserved_blocks (self, &stvfsbuf);
+  g_assert (self->txn.blocksize != 0);
+  self->reserved_blocks = min_free_space_calculate_reserved_blocks (self, &stvfsbuf);
+  if (self->reserved_blocks > G_MAXUINT64 / self->txn.blocksize)
+    {
+      g_mutex_unlock (&self->txn_lock);
+      return glnx_throw (error, "min-free-space value(in bytes) was too high");
```
pwithnall: It would be good to include the maximum permitted value in this error message. For example:
```
return glnx_throw ("min-free-space value is greater than the maximum allowed value of %" G_GUINT64_FORMAT " bytes", G_MAXUINT64 / self->txn.blocksize);
```

src/libostree/ostree-repo.c
```diff
@@ -3292,6 +3292,35 @@ ostree_repo_get_mode (OstreeRepo  *self)
...
+ * @error: Return location for a #GError
+ *
+ * It should be noted that this function should be used only if there
+ * is a transaction active. It is a programmer error to request it
+ * otherwise.
+ *
+ * If the configuration value of min-free-space-* value is too high(overflow)
+ * then @error is set with status code %G_IO_ERROR_FAILED.
+ *
+ * Returns: Value(in bytes) of min-free-space-* config option
```
pwithnall: Nitpick: Put a space before `(` here and in the paragraph above.

src/libostree/ostree-repo.c
```diff
@@ -3292,6 +3292,35 @@ ostree_repo_get_mode (OstreeRepo  *self)
...
+guint64
+ostree_repo_get_min_free_space_bytes (OstreeRepo  *self, GError **error)
+{
+  g_return_val_if_fail (OSTREE_IS_REPO (self), 0);
+  g_return_val_if_fail (error == NULL || *error == NULL, 0);
+  g_return_val_if_fail (self->in_transaction == TRUE, 0);
+
+  g_assert (self->txn.blocksize != 0);
+  if (self->reserved_blocks > G_MAXUINT64 / self->txn.blocksize)
+    return glnx_throw (error, "min-free-space(in bytes) value was too high");
```
pwithnall: (Make sure this error message is the same as whatever you change the one above to be.)

tests/test-repo.c
```diff
@@ -151,6 +151,57 @@ test_repo_equal (Fixture       *fixture,
...
+  g_autoptr(OstreeRepo) repo = ostree_repo_create_at (fixture->tmpdir.fd, ".",
+                                                       OSTREE_REPO_MODE_ARCHIVE,
+                                                       NULL,
+                                                       NULL, &error);
+  g_assert_no_error (error);
+
+  const char *values_to_test[] = { "500MB",
+                                   "0MB",
+                             //    "17179869185GB", /* Overflow parameter: bytes > G_MAXUINT64 */
+                             //      "17179869183GB", /* bytes approx. under G_MAXUINT64 */ //TODO: Bypass failure
```
pwithnall: Do both of these values error out in `ostree_repo_prepare_transaction()`? If so, that’s probably the best testing we’re going to get, and there’s no point in coming up with contortions to get more coverage.

uajain: Yes, they both error out in `ostree_repo_prepare_transaction` with different appropriate error messages.

> If so, that’s probably the best testing we’re going to get, and there’s no point in coming up with > contortions to get more coverage.

Can you explain this a bit more? Does it mean there is no need for `0MB`, `500MB` test etc.?

pwithnall: As in, don’t try and make the `prepare_transaction()` call artificially succeed so you can test those large values with the `get_min_free_space_bytes()` call.

src/libostree/ostree-repo.c
```diff
@@ -3292,6 +3292,35 @@ ostree_repo_get_mode (OstreeRepo  *self)
...
+ */
+guint64
+ostree_repo_get_min_free_space_bytes (OstreeRepo  *self, GError **error)
+{
+  g_return_val_if_fail (OSTREE_IS_REPO (self), 0);
+  g_return_val_if_fail (error == NULL || *error == NULL, 0);
+  g_return_val_if_fail (self->in_transaction == TRUE, 0);
+
+  g_assert (self->txn.blocksize != 0);
+  if (self->reserved_blocks > G_MAXUINT64 / self->txn.blocksize)
```
uajain: @pwithnall So I was asking about this check. Do we check(again) for overflows here when `min_free_space_calculate_reserved_blocks()` already does it before setting `self->reserved_blocks` ?

tests/test-repo.c
```diff
@@ -151,6 +151,56 @@ test_repo_equal (Fixture       *fixture,
...
+
+  const char *values_to_test[] = { "500MB",
+                                   "0MB",
+                                   "17179869185GB", /* Overflow parameter: bytes > G_MAXUINT64 */
+                                   NULL
+                                 };
+
+  config = ostree_repo_copy_config (repo);
+
+  while (values_to_test[i] != NULL)
```
pwithnall: Nitpick: Use a `for` loop for this. That saves you having to manually increment `i` on every loop path below.

i.e. `for (i = 0; values_to_test[i] != NULL; i++)`

uajain:
@cgwalters @jlebon Can you take a look ? it's LGTM from @pwithnall's side now. 
Thank you.

tests/test-repo.c
```diff
@@ -151,6 +151,50 @@ test_repo_equal (Fixture       *fixture,
...
+  g_autoptr(OstreeRepo) repo = ostree_repo_create_at (fixture->tmpdir.fd, ".",
+                                                      OSTREE_REPO_MODE_ARCHIVE,
+                                                      NULL,
+                                                      NULL, &error);
+  g_assert_no_error (error);
+
+  const char *values_to_test[] = { "500MB",
+                                   "0MB",
+                                   "17179869185GB", /* Overflow parameter: bytes > G_MAXUINT64 */
+                                   NULL
```
cgwalters: It'd be slightly cleaner I think to have this be an array of 2-tuple (struct) `{ const char *val; bool should_succeeed }` and assert that `error` is set if and only if `should_succeed`.

That said, up to you, I don't mind merging as is.

uajain: Thank you for the suggestion.

Pushed a `fixup` commit.

cgwalters:
`test-symbols.sh` wanted leading whitespace in front of symbol names, I pushed a `fixup!`.  Thanks for the patch!

@rh-atomic-bot r+ 9ba887d

cgwalters:
Looks like a flake.

@rh-atomic-bot retry

PR #1716, opened 2018-08-31
lib/fetcher: Fix some memory leaks in curl fetcher

lbonn:


cgwalters:
Thanks!

@rh-atomic-bot r+ c641f27

Issue #1720, opened 2018-09-12
Rework ostree_repo_get_min_free_space_bytes ()


uajain:
In https://github.com/ostreedev/ostree/pull/1715 I added an API to query the `min-free-space-*` config option in a format(`bytes`) which could be useful to application like gnome-software via flatpak wrapper around it. The specific use case for Endless is to place a heuristic around (free-space vs. required space for app-install/update) which can determine(with some probability) if the operation will be successful or not, given the disk-space situation. It also took min-free-space-* into account hence, the genesis for this API.

The min-free-space-* config option is actually parsed into a usable form(`bytes`) during the `prepare_transaction()`. Therefore, while writing this API, I made sure the API is usable only if there is an ongoing transaction so that the value of the option remains consistent and doesn't return inconsistent values outside transaction (which can happen if the config's value is changed in the configuration).

It seems that I have made a mistake that this API can only be read during transaction. The heuristics we place in gnome-software wants to read the `min-free-space-*` option even before the transaction is started by flatpak. Currently it's hard-coded but the eventual goal was to use this API.

Other workarounds suggested is to start and abort a transaction to get the value out with the existing API. 

Please consider this issue as a release blocker. We need collective thinking and how this can rework-ed.

dbnicholson:
Why does the transaction come until play at all? Why not just parse the options once when the repo is opened like 90% of the other repo options? Or just split out the passing to a helper that gets called by the API and when beginning a transaction. Or just call the API when beginning a transaction. As far as I know there's nothing transaction specific about looking at the filesystem stats and the repo options.

ramcq:
Our motivation for adding this was pre-determining whether an operation in GNOME Software will succeed or fail, before we start taking repo locks or doing any disk IO etc. I think opening a transaction ends up writing to disk?

The combination of the volume disk IO/fsyncs around an ostree operation, the fact that the flatpak user -> system pull uses 2x disk space during that operation, and that we're enabling auto updates by defaults, is pretty bad: we're seeing machines which low on disk space getting into loops where they try to auto update, killing interactivity / usability of the machine, fail, then come back and do it again a few hours later. So, your disk is left close to (because Flatpak saves the user repo in /var/tmp), your computer runs really slow, and spits errors randomly about failing updates / low disk. It's pretty cool.

[...]

PR #1722, opened 2018-09-13
RFC: lib/repo: Separate min-free-space-* calculation from transaction code…

uajain:
…path

Earlier, the actual reserved space (in blocks) were calculated inside the
transaction codepath ostree_repo_prepare_transaction(). However, while
reworking on ostree_repo_get_min_free_space_bytes() API, it was realized that
this calculation can be done independently from the transaction's codepaths, hence
enabling the usage for ostree_repo_get_min_free_space_bytes() API irrespective
of whether there is an ongoing transaction or not.

https://github.com/ostreedev/ostree/issues/1720

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2654,6 +2654,37 @@ get_remotes_d_dir (OstreeRepo          *self,
 }
 
 static gboolean
+min_free_space_calculate_reserved_bytes (OstreeRepo *self, guint64 *bytes, GError **error)
+{
+  guint64 reserved_bytes = 0;
+
+  struct statvfs stvfsbuf;
+  if (TEMP_FAILURE_RETRY (fstatvfs (self->repo_dir_fd, &stvfsbuf)) < 0)
+    return glnx_throw_errno_prefix (error, "fstatvfs");
+
+  if (self->min_free_space_mb > 0)
+    {
+      if (self->min_free_space_mb > (G_MAXUINT64 >> 20))
+        return glnx_throw (error, "min-free-space value is greater than the maximum allowed value of %" G_GUINT64_FORMAT " bytes",
+                           (G_MAXUINT64 >> 20));
+
+      reserved_bytes = self->min_free_space_mb << 20;
+    }
+  else if (self->min_free_space_percent > 0)
+    {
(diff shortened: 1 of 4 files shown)


cgwalters:
Hmm.  But the time the repo is opened could be quite distant from the time of a transaction - isn't e.g. gnome-software today a long-running process?

cgwalters:
I would have more leaned towards calculating the space if we're not in a txn when the API is invoked.

uajain:
How about now?

uajain:
@cgwalters @jlebon Any thoughts?

src/libostree/ostree-repo.c
```diff
@@ -2653,6 +2653,37 @@ get_remotes_d_dir (OstreeRepo          *self,
...
 }
 
+static gboolean
+min_free_space_calculate_reserved_bytes (OstreeRepo *self, guint64 *bytes, GError **error)
+{
+  guint64 reserved_bytes = 0;
+
+  struct statvfs stvfsbuf;
+  if (TEMP_FAILURE_RETRY (fstatvfs (self->repo_dir_fd, &stvfsbuf)) < 0)
+    return glnx_throw_errno_prefix (error, "fstatvfs");
```
pwithnall: I would be tempted to set `*bytes = 0` when doing an error return, but it’s not strictly necessary and it’s more a matter of whether that’s part of OSTree policy than anything else. My thinking is that it’s safer to set it to a value than leave it uninitialised and potentially have the caller read it and do something silly (even though they’re not supposed to read it on error).

pwithnall:
bot, retry please

pwithnall:
bot, retest this please

src/libostree/ostree-repo.c
jlebon: Yeah, we don't normally do so this is fine as is to me. Though I wouldn't oppose it either.

pwithnall:
I can’t see any relevant failures in the tests.

pwithnall:
bot, retest this please

pwithnall:
@rh-atomic-bot r+ f77a8eda432a38cbb2535ae40afda94e4ce8e84b

src/libostree/ostree-repo.c
pwithnall: I’ll merge then, unchanged. Thanks for your response!

PR #1724, opened 2018-09-18
OstreeMutableTree: add _remove method

ramcq:
There is no API method to remove a file from a MutableTree besides directly manipulating the GHashTable returned by _get_files. This isn't possible from an introspection binding that transforms the returned GHashTable, and may also leave the tree checksum in an invalid state. ~~Allow NULL to remove a file entry, add GI annotations to show that checksum is nullable.~~ Introduce a new method to allow removing files or subdirectories.


jlebon:
Hmm, there's no way either to remove directories, right? Maybe a dedicated `ostree_mutable_tree_remove` would make more sense to address both of them instead?

I guess we could enhance the semantics of `_replace()` so that a `NULL` checksum can also nuke subdirs, though it wouldn't be at all clear from its name.

ramcq:
Makes sense. I was thinking about how to work around the lack of removal by recursively walking my MutableTree and making a new one without the stuff I wanted to remove, but it seems there's also no API for adding a subdir which is an already existing MutableTree. This whole thing is driving me nuts. Anyway - here is a _remove method, which solves the problem at hand!

pwithnall:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1727, opened 2018-09-20
ci: Disable f28-rpmostree for now

jlebon:
It started failing with:

```
ERROR: tests/check/test-ucontainer.sh - too few tests run (expected 2, got 0)
tap-driver.sh: internal error getting exit status
tap-driver.sh: fatal: I/O or internal error
make[4]: *** [Makefile:4353: tests/check/test-ucontainer.sh.log] Error 1
make[4]: *** Waiting for unfinished jobs....
```

And the artifacts are not being saved for some reason:

```
+ cleanup
+ mv test-suite.log /var/tmp/checkout
mv: cannot stat 'test-suite.log': No such file or directory
+ true
+ mv vmcheck /var/tmp/checkout
mv: cannot stat 'vmcheck': No such file or directory
+ true
```

Let's just disable this for now so that some other pending patches can
go in while we investigate.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -162,7 +162,8 @@ branches:
     - try
 
 context: f28-rpmostree
-required: true
+# XXX: some issues currently failing that need investigating.
+required: false
 
 cluster:
   hosts:


jlebon:
bot, retest this please

cgwalters:
@rh-atomic-bot r+ 508bc36

Issue #1729, opened 2018-09-21
/var/.updated not deleted on deployment


akiernan:
Trying to use systemd's `ConditionNeedsUpdate=/var` it seems as if ostree-sysroot-deploy.c#L2557 isn't looking in the right directory when it unlinks the file, running through `strace`:

`597   unlinkat(12, "var/.updated", 0)   = -1 ENOENT (No such file or directory)`

whereas the file exists:

```
$ ls -l /var/.updated
-rw-r--r-- 1 root root 190 Jan  1  1970 /var/.updated```

akiernan:
Debugging further, it looks like its an issue with separate `/var`

cgwalters:
> Debugging further, it looks like its an issue with separate /var

Ah, right - we need to do the `unlinkat()` from the view of the *booted* deployment - but only if we're booted.  (Correctly handling `ostree admin --sysroot /path/to/os upgrade` is tricky with `/var` mounts, we would need to either mount it ourself, or perhaps require some external process to do so.  I'm honestly though not aware of anyone today doing upgrades like that)

Hmm.  I wonder if would somehow be cleaner to have our systemd generator for `/var` *reverse* things if it detects a `/var` mount by adding a bind mount from the `/var` device/mount to the `/sysroot/.../var`.

cgwalters:
Working on a patch for this.

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1730 - if you wouldn't mind reviewing and testing that'd be great!

PR #1730, opened 2018-09-21
deploy: Fix removing /var/.updated with separate /var mount

cgwalters:
There's some subtlety to this, we don't handle all cases.
But the 99% cases are using `--sysroot deploy` to create an
initial deployment, and then doing upgrades from inside
a booted deployment.

It was only the latter case that didn't work with a separate `/var`.
Fixing all of them would probably require libostree to learn
how to e.g. look at `/etc/fstab` (or worse, systemd mount units?)
and handle the mounting.  I don't think we want to do anything
like that right now, since there are no active drivers for the
use case.

Closes: https://github.com/ostreedev/ostree/issues/1729

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2509,6 +2509,47 @@ sysroot_initialize_deployment (OstreeSysroot     *self,
   return TRUE;
 }
 
+/* Get a directory fd for the /var of @deployment.
+ * Before we supported having /var be a separate mount point,
+ * this was easy. However, as https://github.com/ostreedev/ostree/issues/1729
+ * raised, in the primary case where we're
+ * doing a new deployment for the booted stateroot,
+ * we need to use /var/.  This code doesn't correctly
+ * handle the case of `ostree admin --sysroot upgrade`,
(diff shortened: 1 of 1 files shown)


cgwalters:
Just compile/unit tested, but sadly our test suite today doesn't actively exercise split `/var`.

akiernan:
LGTM

```
root@hvt-341:~# ls -l /var/.updated
-rw-r--r-- 1 root root 190 Jan  1  1970 /var/.updated
root@hvt-341:~# ostree admin deploy master/nano/nano2/nano-image-osa
Copying /etc changes: 4 modified, 0 removed, 15 added
[  164.911751] EXT4-fs (mmcblk0p2): re-mounted. Opts: (null)
warning: Failed to remount /boot read-only: Device or resource busy
Transaction complete; bootconfig swap: yes; deployment count change: 0
Freed objects: 600.9 kB
root@hvt-341:~# ls -l /var/.updated
ls: cannot access '/var/.updated': No such file or directory
```

cgwalters:

akiernan:
@rh-atomic-bot r+ f787f66e9300ba89da7bfdcc96c7098a51c7fc2b

Issue #1731, opened 2018-09-21
Include hostname that failed when propagating CURLE_COULDNT_CONNECT


jlebon:
The error is not super helpful otherwise:

```
$ journalctl -b --no-hostname -p err --no-pager
Sep 21 15:53:46 rpm-ostree[1953]: Txn AutomaticUpdateTrigger on /org/projectatomic/rpmostree1/fedora_workstation failed: [7] Couldn't connect to server
Sep 21 15:54:14 rpm-ostree[1953]: Txn AutomaticUpdateTrigger on /org/projectatomic/rpmostree1/fedora_workstation failed: [7] Couldn't connect to server
```

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1631790

PR #1732, opened 2018-09-21
lib/fetcher-curl: Prefix fatal errors with full URL 

jlebon:
Just include the whole URL that failed if libcurl failed with something
elementary like `CURLE_COULDNT_CONNECT` or `CURLE_COULDNT_RESOLVE_HOST`.

Closes: #1731


cgwalters:
@rh-atomic-bot r+ 5a020d0

PR #1733, opened 2018-09-21
create-usb: Always use archive mode

mwleeds:
Change the create-usb command so that it always creates the destination
repository using the "archive" mode, rather than using archive mode when
xattrs aren't supported and bare-user otherwise. This has a few
advantages:

1. The archive mode works with FAT filesystems, which is what most
USB drives are, and which doesn't support xattrs.

2. At least in some quick testing I did, archive mode is about
twice as performant as bare-user mode, in terms of how long it takes for
the create-usb command to complete.

3. This ensures that a tool can safely change the permissions on
".ostree/repo" and subdirectories after create-usb completes, which is
important for Endless since otherwise you can't use `ostree create-usb`
as root and then `flatpak create-usb` as a non-root user on the same USB
drive (or in other words copy OS updates and apps to the same USB).

### DIFF
--- a/src/ostree/ot-builtin-create-usb.c
+++ b/src/ostree/ot-builtin-create-usb.c
@@ -102,21 +102,17 @@ ostree_builtin_create_usb (int            argc,
 
   /* Open the destination repository on the USB stick or create it if it doesn’t exist.
    * Check it’s below @mount_root_path, and that it’s not the same as the source
-   * repository.
-   *
-   * If the destination file system supports xattrs (for example, ext4), we use
-   * a BARE_USER repository; if it doesn’t (for example, FAT), we use ARCHIVE.
-   * In either case, we want a lossless repository. */
+   * repository. */
   const char *dest_repo_path = (opt_destination_repo != NULL) ? opt_destination_repo : ".ostree/repo";
 
   if (!glnx_shutil_mkdir_p_at (mount_root_dfd, dest_repo_path, 0755, cancellable, error))
     return FALSE;
 
-  OstreeRepoMode mode = OSTREE_REPO_MODE_BARE_USER;
-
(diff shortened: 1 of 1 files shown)


mwleeds:
This does the same as https://github.com/flatpak/flatpak/pull/2124 but for ostree

cgwalters:
For flatpak content, it should be possible to use `BARE_USER_ONLY`.  But that won't work if the USB contains operating system content with xattrs.

>  Is that because archive files are smaller and writing to USB is slow?

If one is syncing from a repo that's already archive, then we don't recompress, just `copy_file_range()` - could be that?

ramcq:
> For flatpak content, it should be possible to use `BARE_USER_ONLY`. But that won't work if the USB contains operating system content with xattrs.
> 
> > Is that because archive files are smaller and writing to USB is slow?
> 
> If one is syncing from a repo that's already archive, then we don't recompress, just `copy_file_range()` - could be that?

I think it's likely to just be CPU compression is cheaper than writing to flash memory. We're exporting from our OS (bare) repo which is what we use for Flatpak too. So we need flat permissions and no xattrs for maximum compatibility, and need to share the stick between the OS updater and app exports - I think that means archive is our only option anyway.

mwleeds:
> If one is syncing from a repo that's already archive, then we don't recompress, just `copy_file_range()` - could be that?

But things are still fsync'd before the command completes right? We're not calling `ostree_repo_set_disable_fsync` anywhere, so that should be the case, but otherwise my performance testing counts for nothing and `gnome-software` is sort of misleading people by showing the copy as finished when `create-usb` finishes.

mwleeds:
oh yeah, `create-usb` has a `--disable-fsync` option

cgwalters:
@rh-atomic-bot r+ 9e82dee

PR #1735, opened 2018-09-24
create-usb: Add a --commit option

mwleeds:
Currently on Endless OS, the OSTree ref for the operating system is
something like os/eos/amd64/eos3, so that's what gets passed to `ostree
create-usb` when copying the OS to a USB drive (for offline updates).
However, when eos-updater checks for updates it pulls the metadata for a
candidate commit and in so doing updates that eos3 ref to point to the
partial commit being examined as a potential update rather than the
deployed commit. This causes `ostree create-usb` to fail with an error
like "No such metadata object
7fb045cb2d1f1f3a81bfc157c6128ff443eb56350315b9536bdb56aee0659863.dirtree".

OSTree creates deployment refs that look like "ostree/1/1/0" to maintain
a pointer to the deployed commit, but create-usb can't use these because
it shows up in the summary as just a ref, not a collection-ref.

So this commit adds a --commit option to the create-usb command, so we
can use the appropriate ref but copy the deployed commit rather than a
(potentially partial) update commit.

### DIFF
--- a/man/ostree-create-usb.xml
+++ b/man/ostree-create-usb.xml
@@ -112,6 +112,15 @@ Boston, MA 02111-1307, USA.
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--commit</option>=COMMIT</term>
+
+                <listitem><para>
+                  Pull COMMIT instead of whatever REF points to. This can only
+                  be used if a single ref is specified.
+                </para></listitem>
+            </varlistentry>
+
         </variablelist>
     </refsect1>
 
--- a/src/ostree/ot-builtin-create-usb.c
+++ b/src/ostree/ot-builtin-create-usb.c
@@ -33,11 +33,13 @@
 
 static gboolean opt_disable_fsync = FALSE;
 static char *opt_destination_repo = NULL;
+static char *opt_commit = NULL;
 
 static GOptionEntry options[] =
   {
     { "disable-fsync", 0, 0, G_OPTION_ARG_NONE, &opt_disable_fsync, "Do not invoke fsync()", NULL },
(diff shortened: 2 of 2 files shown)


mwleeds:
You can always run `create-usb` more than once, so it doesn't seem like much of a limitation to only have `--commit` work with one ref.

cgwalters:
The other option would be for `create-usb` to automatically skip trying to copy partial commits.

cgwalters:
@rh-atomic-bot r+ 7b6fe75

pwithnall:
@rh-atomic-bot retry

mwleeds:
> Would it be worth adding a sanity check that the given commit has a ref binding for the specified ref? Otherwise you could provide two completely unrelated commits/refs and the problem would only be detected when another computer tried to pull from the USB stick and its ref binding checks failed.

This is already the behavior:
```
endless@endless:~$ ostree log eos:ostree-metadata
commit 4ef67468d3a8adef4034acc893998837cd9cf519503660e5d6c21e6e8d60eec4
ContentChecksum:  978b7746df27a18398d9099592a905091a496bf53f9158c1f350d1d410424f66
Date:  2018-09-10 16:32:28 +0000
(no subject)

<< History beyond this commit not fetched >>
endless@endless:~$ sudo ostree create-usb --commit=4ef67468d3a8adef4034acc893998837cd9cf519503660e5d6c21e6e8d60eec4 /run/media/endless/1a9b4cb2-a7ef-4d9b-91a5-6eaf8fdd2bf6/ com.endlessm.Os os/eos/amd64/eos3a

error: Commit 4ef67468d3a8adef4034acc893998837cd9cf519503660e5d6c21e6e8d60eec4: Commit has no requested ref ‘os/eos/amd64/eos3a’ in ref binding metadata (‘ostree-metadata’)
```

Issue #1734, opened 2018-09-24
ostree xattr error when `pull-local` destination is archive repo on filesystem without xattrs


dustymabe:
```
ostree-2018.8.7-e4e6d85ea4e778c61b8f9a0301153dd0ee00b9b6.5bb0359caf4facc535ee3ee8b5ec7c552637c28d.fc28.x86_64
```

- If I pull local from bare-user (local ext2) to archive repo (local ext2) things work.
- If I pull local from archive (local ext2) to archive repo (local/netfs 9pfs/virtfs or smb) things work.
- If I pull local from bare-user (local ext2) to archive repo (local/netfs 9pfs/virtfs or smb) things don't.

I get an error like:

```
bash-4.4# ostree pull-local --repo=/hostsrv/repo/ /srv/repo-build/ fedora/29/x86_64/coreos
error: Importing b7e0a2c196fc139fcf8c2cc2bcc69581ce3ddb6f5d3978de0fced04a5536906f.commit: fsetxattr: Operation not supported
```

mwleeds:
Whatever the explanation for this is, it's probably also the explanation for https://github.com/flatpak/flatpak/issues/2019

cgwalters:
> Whatever the explanation for this is, it's probably also the explanation for flatpak/flatpak#2019

Yeah, probably hitting there because Matthias is using a SELinux-enabled system which has `security.selinux` xattrs on even metadata objects.

PR #1736, opened 2018-09-25
lib/commit: Don't copy xattrs for metadata objects

cgwalters:
Copying the xattrs on metadata objects is wrong in general, we
don't "own" them.  Notably this would fail in the situation of
doing a pull from e.g. a `bare-user` source to a destination
that was on a different mount point (so we couldn't hardlink),
and the source had e.g. a `security.selinux` attribute.

Closes: #1734


jlebon:
@rh-atomic-bot r+ be81d3f

PR #1739, opened 2018-09-28
lib/repo: Validate lock-timeout-secs config

mwleeds:
Now that we're correctly reading negative values for the
"lock-timeout-secs" config key, throw an error if it's not within the
range of values that have semantic meaning. This mirrors the behavior
when other config keys have invalid values.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2833,6 +2833,8 @@ reload_core_config (OstreeRepo          *self,
           return FALSE;
 
         self->lock_timeout_seconds = g_ascii_strtoll (lock_timeout_seconds, NULL, 10);
+        if (self->lock_timeout_seconds < REPO_LOCK_DISABLED);
+          return glnx_throw (error, "Invalid lock-timeout-secs '%s'", lock_timeout_seconds);
       }
   }
 


dbnicholson:
I like this better than relying on an assert since that's a programmer error, not a user error. Actually, there are a bunch of potential errors not caught here that were in my original implementation in https://github.com/ostreedev/ostree/pull/1292/commits/532199858ce283a7e8528a1f2affc7c8069a43fc. See also @pwithnall's review in https://github.com/ostreedev/ostree/pull/1292/commits/532199858ce283a7e8528a1f2affc7c8069a43fc#r147145444.

mwleeds:
That link didn't work for me but I think I found what you're referring to here: https://github.com/ostreedev/ostree/pull/1292#discussion_r147145444

cgwalters:

pwithnall:
Note: You could also potentially use `g_ascii_string_to_[un]signed()` rather than `g_strto[u]ll()`, if you depend on a new enough GLib version (I haven’ŧ checked).

mwleeds:
Oops, didn't intend for y'all to spend time reviewing that last commit. I was pushing it to get it from one computer to another. But yeah, I'll follow-up and fix it.

PR #1740, opened 2018-09-28
boot: Add ostree-finalize-staged.path

jlebon:
Rather than manually starting the `ostree-finalize-staged.service` unit,
we can leverage systemd's path units for this. It fits quite nicely too,
given that we already have a path we drop iif we have a staged
deployment.

---

I started on this thinking we somehow didn't have to change presets. But of course, we do still need a preset for *the path unit itself*. So anyway, just throwing this up here, since I do think overall that using a path unit is nicer, but this is now predicated on getting a preset added in distros (at least for rpm-ostree-based downstreams, we can just do like in #1482, for others... we'd need to mention it clearly in the release notes).

### DIFF
--- a/src/boot/ostree-finalize-staged.path
+++ b/src/boot/ostree-finalize-staged.path
@@ -1,0 +1,28 @@
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library; if not, write to the
+# Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+# Boston, MA 02111-1307, USA.
+
+# For some implementation discussion, see:
+# https://lists.freedesktop.org/archives/systemd-devel/2018-March/040557.html
+[Unit]
+Description=OSTree Monitor Staged Deployment
+Documentation=man:ostree(1)
+
+[Path]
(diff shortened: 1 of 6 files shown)


src/boot/ostree-finalize-staged.path
```diff
@@ -0,0 +1,25 @@
...
+# Boston, MA 02111-1307, USA.
+
+# For some implementation discussion, see:
+# https://lists.freedesktop.org/archives/systemd-devel/2018-March/040557.html
+[Unit]
+Description=OSTree Finalize Staged Deployment
+Documentation=man:ostree(1)
+
+[Path]
+PathExists=/run/ostree/staged-deployment
```
cgwalters: When is this condition evaluated though?  Is it only once at startup?

jlebon: It's monitored using inotify: https://www.freedesktop.org/software/systemd/man/systemd.path.html

It'd be nice if this also *deactivated* the service if we did e.g. `rpm-ostree cleanup -p`, but alas: systemd/systemd#3642

Though at least we'd gain that for free once/if that gets implemented.

cgwalters:
>  but this is now predicated on getting a preset added in distros

Right.  So we can do that now and then land this after the next release?

jlebon:
OK, I opened https://bugzilla.redhat.com/show_bug.cgi?id=1639372 and https://github.com/projectatomic/rpm-ostree/pull/1617.

jlebon:
OK, I've tested this in combination with projectatomic/rpm-ostree#1617. As the commit message mentions, I think we can get this in now for the upcoming release and then in the next release or the one after, drop that `systemctl start` path for good.

jlebon:
Requires: https://src.fedoraproject.org/rpms/ostree/pull-request/11

jlebon:
> error: Installed (but unpackaged) file(s) found:
>    /usr/lib/systemd/system/ostree-finalize-staged.path

Right, this'll be fixed by the dist-git PR, though `FAH28-insttests` is not currently marked as required.

jlebon:
bot, retest this please

cgwalters:
@rh-atomic-bot r+ cdda3a0

cgwalters:
@rh-atomic-bot r+

jlebon:
@rh-atomic-bot retry

Issue #1738, opened 2018-09-28
OstreeAsyncProcess leaks sources


matthiasclasen:
Seen in a valgrind log here: https://github.com/flatpak/flatpak/issues/1700

```
==30958== 79,232 (480 direct, 78,752 indirect) bytes in 5 blocks are definitely lost in loss record 2,790 of 2,806
==30958==    at 0x4C31B25: calloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==30958==    by 0x64B7930: g_malloc0 (gmem.c:129)
==30958==    by 0x64AFBDD: g_source_new (gmain.c:915)
==30958==    by 0x64B3051: g_idle_source_new (gmain.c:5559)
==30958==    by 0x56DDBE8: ensure_callback_locked.part.1 (ostree-async-progress.c:244)
==30958==    by 0x56DE34D: ensure_callback_locked (ostree-async-progress.c:242)
==30958==    by 0x56DE34D: ostree_async_progress_set (ostree-async-progress.c:360)
==30958==    by 0x5700DB8: update_progress (ostree-repo-pull.c:276)
==30958==    by 0x64B2B72: g_timeout_dispatch (gmain.c:4649)
==30958==    by 0x64B20F4: g_main_dispatch (gmain.c:3176)
==30958==    by 0x64B20F4: g_main_context_dispatch (gmain.c:3829)
==30958==    by 0x64B24BF: g_main_context_iterate.isra.26 (gmain.c:3902)
==30958==    by 0x64B254B: g_main_context_iteration (gmain.c:3963)
==30958==    by 0x5706318: ostree_repo_pull_with_options (ostree-repo-pull.c:4121)
```

Looking at the code, ensure_callback_locked creates a new GSource, and sets self->idle_source, and then the idle_invoke_async_progress callback just ... sets the pointer to NULL. How is the source ever going to get destroyed ?

PR #1741, opened 2018-09-28
lib/progress: Fix leak of GSource

cgwalters:
Closes: https://github.com/ostreedev/ostree/issues/1738


jlebon:
@rh-atomic-bot r+ 4a42085

PR #1742, opened 2018-09-28
lib/mutable-tree: Port to new style

jlebon:
Some therapeutic style conversion to finish off the week. Pretty
straightforward overall.


cgwalters:
@rh-atomic-bot r+ a669202

PR #1743, opened 2018-09-28
Add an "ostree config unset" operation

mwleeds:


man/ostree-config.xml
```diff
@@ -62,20 +62,42 @@ Boston, MA 02111-1307, USA.
             <cmdsynopsis>
                 <command>ostree config set</command> <arg choice="req"> --group=GROUPNAME</arg> <arg choice="req"> KEYNAME</arg> <arg choice="req">VALUE</arg>
             </cmdsynopsis>
+            <cmdsynopsis>
+                <command>ostree config unset</command> <arg choice="req">SECTIONNAME.KEYNAME</arg>
+            </cmdsynopsis>
+            <cmdsynopsis>
+                <command>ostree config unset</command> <arg choice="req"> --group=GROUPNAME</arg> <arg choice="req"> KEYNAME</arg>
```
jlebon: Aside: I'm confused now why we use both the `GROUPNAME` and `SECTIONNAME` nouns when they're the same thing, right? I'd vote for harmonizing those to make it the same. Anyway, just writing this down so I don't forget, doesn't have to be in this patch!

tests/test-config.sh
```diff
@@ -53,3 +53,20 @@ assert_file_has_content repo/config "Nightly Flathub"
 assert_file_has_content repo/config "false"
 assert_file_has_content repo/config "http://example.com/ostree/"
 echo "ok config set"
+
+${CMD_PREFIX} ostree config --repo=repo set core.lock-timeout-secs 60
+assert_file_has_content repo/config "lock-timeout-secs=60"
+${CMD_PREFIX} ostree config --repo=repo unset core.lock-timeout-secs
+assert_not_file_has_content repo/config "lock-timeout-secs=60"
```
jlebon: A stronger check here would be that the key doesn't exist at all (i.e. just `lock-timeout-secs=`). And maybe corroborate that with sanity checking that `ostree config get` erroring out?

mwleeds:
I don't really have time to work on this right now but just interesting to note, `flatpak config` has a `--unset` option

man/ostree-config.xml
mwleeds: Added a commit to fix this

tests/test-config.sh
mwleeds: You're right; I made both changes and force-pushed the branch

mwleeds:
This is ready for re-review now, and has a new commit to check for too many arguments. Sorry for the delay

jlebon:
This looks great to me! Will leave it open for a bit to give others time to review.

cgwalters:
@rh-atomic-bot r+ 6b417c3

PR #1745, opened 2018-10-01
bash-completion: Fix --repo autocomplete

mwleeds:
This commit fixes the bash tab completion handling of the "--repo"
argument. Before this commit, the completion only works if "--repo"
comes after the main command. After this commit, you can use "--repo"
directly after "ostree" in the command line, as is natural.


jlebon:
Thanks!
@rh-atomic-bot r+ a5c3177

PR #1746, opened 2018-10-02
man/create-usb: Don't recommend summary updates

mwleeds:
This commit removes the recommendation in the create-usb man page for
the user to update the summary in the source repo before using the
create-usb command. I'm not sure where I got the idea that create-usb
depends on a summary in the source repo. I went back to the first commit
that introduced the create-usb command and even using that a summary
isn't required, so it seems unlikely that this changed recently.

This is good news because the exclusive lock that's taken for summary
updates has been causing problems on Endless (due to other processes
having a lock for the duration of the 30 second acquire time out
period).


cgwalters:
@rh-atomic-bot r+ e1b9df9

PR #1747, opened 2018-10-02
rofiles-fuse: Improve error message for failure to open root

cgwalters:
I was debugging some rpm-ostree work and saw:
`openat: No such file or directory`
and it wasn't immediately obvious it was stderr from `rofiles-fuse`.

Use the `err` API which is better in many ways; in this case
it automatically prefixes with `argv0`.

### DIFF
--- a/src/rofiles-fuse/main.c
+++ b/src/rofiles-fuse/main.c
@@ -25,6 +25,7 @@
 #include <sys/stat.h>
 #include <sys/statvfs.h>
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


jlebon:
@rh-atomic-bot r+ 682ad9b

cgwalters:
@rh-atomic-bot retry

cgwalters:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1750, opened 2018-10-05
lib/sysroot-deploy: Write to journal when finalizing

jlebon:
Write to the journal when starting to finalize a staged deployment.
Combined with the "Transaction completed" message we already emit, this
makes it easy later on to determine whether the operation was successful
by inspecting the journal. This will be used by `rpm-ostree status`.

I also cherry-picked some commits from #1740 here.


cgwalters:
@rh-atomic-bot r+ 8926ff8

PR #1753, opened 2018-10-11
checkout: Support --union-identical and --force-copy{,--zerosized}

cgwalters:
Actually testing the patch to add `--force-copy-zerosized` to
rpm-ostree tripped over the fact that it uses `--union-identical`,
and we just hit an assertion failure with that combination.

Fix this by copying over the logic we have for the hardlink case.


jlebon:
@rh-atomic-bot r+ cc33513

PR #1754, opened 2018-10-12
lib/commit: Don't chown objects to repo target owner

dbnicholson:
The idea is that if the process is running as root, it can change
ownership of newly written files to match the owner of the repo.
Unfortunately, it currently applies in the other direction, too - a
non-root user writing to a root owned repository. If the repo is
writable by the user but owned by root, it can still create files and
directories there, but it can't change ownership of them.

This feature comes from
https://bugzilla.gnome.org/show_bug.cgi?id=738954. As it turns out, this
feature was never completed. It only works on content objects and not
metadata objects, refs, deltas, summaries, etc. Rather than try to fix
all of those, remove the feature until someone has interest in
completing it.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -245,16 +245,7 @@ commit_loose_regfile_object (OstreeRepo        *self,
                              GCancellable      *cancellable,
                              GError           **error)
 {
-  /* We may be writing as root to a non-root-owned repository; if so,
-   * automatically inherit the non-root ownership.
-   */
-  if (self->mode == OSTREE_REPO_MODE_ARCHIVE
-      && self->target_owner_uid != -1)
-    {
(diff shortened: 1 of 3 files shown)


cgwalters:
Yeah...valid.  Ironically I was just thinking about this again today in relation to https://github.com/coreos/coreos-assembler/pull/165 but...you're right it's incomplete and lacks tests and obviously has bugs.

@rh-atomic-bot r+ 7f6a100

dbnicholson:
> Yeah...valid. Ironically I was just thinking about this again today in relation to [coreos/coreos-assembler#165](https://github.com/coreos/coreos-assembler/pull/165) but...you're right it's incomplete and lacks tests and obviously has bugs.

It's definitely a worthy idea in certain scenarios. The alternative is just to add a `getuid() == 0` to the conditional, but I figured since no one had worked on it in 4 years it probably wasn't going to be finished.

PR #1757, opened 2018-10-17
two small error patches

cgwalters:


jlebon:
@rh-atomic-bot r+ 787dbf1

PR #1759, opened 2018-10-18
ostree-prepare-root.service: Run earlier in initrd

jlebon:
Previously, we were preparing the root very late in the boot process;
right before we switch root. The issue with that is that most services
in the initrd that run `After=initrd-root-fs.target` expect that
`/sysroot` already points to the rootfs we'll be pivoting to. Running
this late violates that assumption.

This patch fixes this by making `ostree-prepare-root.service` instead
run right after `sysroot.mount` (the physical sysroot mounted by
systemd) but still before `initrd-root-fs.target` (which is the target
signalling that `/sysroot` is now valid and ready).

This should make it easier to integrate OSTree with other initrd
services such as Ignition.

Related: dustymabe/ignition-dracut#20

### DIFF
--- a/src/boot/dracut/module-setup.sh
+++ b/src/boot/dracut/module-setup.sh
@@ -36,7 +36,7 @@ depends() {
 install() {
     dracut_install /usr/lib/ostree/ostree-prepare-root
     inst_simple "${systemdsystemunitdir}/ostree-prepare-root.service"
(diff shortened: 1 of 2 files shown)


cgwalters:
@rh-atomic-bot r+ bb0f1be

dbnicholson:
We do have a couple initrd services in https://github.com/endlessm/eos-boot-helper/tree/master/dracut that depend on the mount ordering. I haven't looked closely at the ordering yet. @dsd FYI.

PR #1760, opened 2018-10-21
remount: Refactor to helper function instead of loop

cgwalters:
Prep for further work.  It was silly to use a loop on
a static array of two elements.

### DIFF
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -39,13 +39,40 @@
 
 #include "ostree-mount-util.h"
 
+static void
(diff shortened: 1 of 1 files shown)


jlebon:
@rh-atomic-bot r+ cbbddd3

jlebon:
@rh-atomic-bot r+ cbbddd3

jlebon:
Ahh OK, seems like webhooks might be down: https://status.github.com/messages.

jlebon:
I think webhooks are still down. Though let's give it a try.
@rh-atomic-bot retry

PR #1761, opened 2018-10-22
Release 2018.9

uajain:


jlebon:
Yeah, I suppose it's about that time! There's #1760 that seems low risk, and I'd like to get #1740 into this release as well if possible.

mwleeds:
This will have to be redone if https://github.com/ostreedev/ostree/pull/1763 is merged. Sorry about that.

uajain:
> This will have to be redone if #1763 is merged. Sorry about that.

Looks like it's approved but not merged due to some internal error. I'll check again after some time.
@jlebon @cgwalters Do you have anything else(symbol-wise) before I update this release PR?

cgwalters:
> @jlebon @cgwalters Do you have anything else(symbol-wise) before I update this release PR?

Nothing from me.

uajain:
Updated the PR.

cgwalters:
Thanks.  @uajain what time zone are you in?  Often jlebon and I set up an etherpad where we write release notes together.  Both of us are in EST.  Do you want to participate in that too?  If so, would say 2-3pm EST tomorrow work?

uajain:
Hey, @cgwalters I am in India (UTC +5:30) timezone, so I don't think that time suits me. However, I am happy to get involved asynchronously for any matter.

cgwalters:
Hm, some test flakes there.  But let's do this:

@rh-atomic-bot r+ b0eedc9

We can get the release commit merged, and I'll post a link to an etherpad here for us to collaborate on release notes and then take care of pushing the tag.

jlebon:
Fix in #1765.
Actually, maybe I should just inline it here. Will do that.

src/ostree/ot-main.c
```diff
@@ -354,7 +354,7 @@ ostree_option_context_parse (GOptionContext *context,
        * possible to parse */
       g_auto(GStrv) features = g_strsplit (OSTREE_FEATURES, " ", -1);
       g_print ("%s:\n", PACKAGE_NAME);
-      g_print (" Version: %s\n", PACKAGE_VERSION);
+      g_print (" Version: v%s\n", PACKAGE_VERSION);
```
cgwalters: How about making this `g_print (" Version: '%s'\n", PACKAGE_VERSION);` instead?

jlebon: Done! :arrow_down:

cgwalters:
@rh-atomic-bot r+

cgwalters:
Hmm, although the commit is after the release commit right?  Probably need to rebase it to before?

jlebon:
We could do that. Though note whatever out there was relying on this before should just keep working in this release because `2018.9` still parses fine. So it doesn't *have* to be in this release.

cgwalters:
I didn't start on the rest of release yet (relnotes/tag etc)...will look at this tomorrow.

cgwalters:
Set up an etherpad for this release: https://public.etherpad-mozilla.org/p/ostree-2018.9

PR #1762, opened 2018-10-22
README: Add bindings section

cgwalters:
Since rust-libostree now exists too, let's make sure people
know about it.


jlebon:
@rh-atomic-bot r+ 858c9fb

jlebon:
@rh-atomic-bot retry

PR #1765, opened 2018-10-24
ostree/main: Prefix version with "v"

jlebon:
Prefix the `Version` key in the YAML-compatible output of
`ostree --version` with `v` so that it's coerced to a string. The issue
with the previous approach in a nutshell:

```
In [5]: yaml.load("asdf: 2018.10")
Out[5]: {'asdf': 2018.1}
```

It's treating the version number as a floating-point. Now, this is
technically a backwards incompatible change, but given that the previous
approach is inherently broken for our needs, I don't see a way around
breaking it now.


jlebon:
I just inlined this into #1761 since it's needed to get a release out. Let's review and iterate there instead if needed.

PR #1767, opened 2018-10-29
Support mounting /sysroot (and /boot) read-only

cgwalters:
Add a magic file and an API for consumers to tell us it's OK
for us to remount read-write (since it'll just affect our
mount namespace).

### DIFF
--- a/src/libostree/ostree-impl-system-generator.c
+++ b/src/libostree/ostree-impl-system-generator.c
@@ -28,6 +28,7 @@
 #ifdef HAVE_LIBMOUNT
 #include <libmount.h>
 #endif
+#include <sys/statvfs.h>
 #include <stdbool.h>
 #include "otutil.h"
 
@@ -189,8 +190,6 @@ _ostree_impl_system_generator (const char *ostree_cmdline,
                                "[Unit]\n"
                                "Documentation=man:ostree(1)\n"
                                "ConditionKernelCommandLine=!systemd.volatile\n"
-                               /* We need /sysroot mounted writable first */
-                               "After=ostree-remount.service\n"
                                "Before=local-fs.target\n"
                                "\n"
                                "[Mount]\n"
--- a/src/libostree/ostree-sysroot-private.h
+++ b/src/libostree/ostree-sysroot-private.h
@@ -40,6 +40,12 @@ typedef enum {
   OSTREE_SYSROOT_DEBUG_TEST_STAGED_PATH = 1 << 3,
 } OstreeSysrootDebugFlags;
 
+typedef enum {
+      OSTREE_SYSROOT_LOAD_STATE_NONE, /* ostree_sysroot_new() was called */
+      OSTREE_SYSROOT_LOAD_STATE_INIT, /* We've loaded basic sysroot state and have an fd */
+      OSTREE_SYSROOT_LOAD_STATE_LOADED, /* We've loaded all of the deployments */
+} OstreeSysrootLoadState;
+
 /**
  * OstreeSysroot:
  * Internal struct
@@ -51,7 +57,8 @@ struct OstreeSysroot {
   int sysroot_fd;
   GLnxLockFile lock;
 
-  gboolean loaded;
+  OstreeSysrootLoadState loadstate;
+  gboolean mount_namespace_in_use; /* TRUE if caller has told us they used CLONE_NEWNS */
   gboolean root_is_ostree_booted; /* TRUE if sysroot is / and we are booted via ostree */
   /* The device/inode for /, used to detect booted deployment */
   dev_t root_device;
@@ -78,6 +85,10 @@ struct OstreeSysroot {
 #define _OSTREE_SYSROOT_RUNSTATE_STAGED_LOCKED "/run/ostree/staged-deployment-locked"
 #define _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_DIR "/run/ostree/deployment-state/"
 #define _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_DEVELOPMENT "unlocked-development"
+
+gboolean
+_ostree_sysroot_ensure_writable (OstreeSysroot      *self,
+                                 GError            **error);
 
 void
 _ostree_sysroot_emit_journal_msg (OstreeSysroot  *self,
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -227,6 +227,33 @@ ostree_sysroot_new_default (void)
 }
 
 /**
+ * ostree_sysroot_set_mount_namespace_in_use:
+ *
+ * If this function is invoked, then libostree will assume that
+ * a private Linux mount namespace has been created by the process.
+ * The primary use case for this is to have e.g. /sysroot mounted
+ * read-only by default.
+ *
+ * If this function has been called, then when a function which requires
+ * writable access is invoked, libostree will automatically remount as writable
+ * any mount points on which it operates.  This currently is just `/sysroot` and
+ * `/boot`.
+ *
+ * If you invoke this function, it must be before ostree_sysroot_load(); it may
+ * be invoked before or after ostree_sysroot_initialize().
+ *
+ * Since: 2019.7
+ */
+void
+ostree_sysroot_set_mount_namespace_in_use (OstreeSysroot  *self)
+{
+  /* Must be before we're loaded, as otherwise we'd have to close/reopen all our
+     fds, e.g. the repo */
+  g_return_if_fail (self->loadstate < OSTREE_SYSROOT_LOAD_STATE_LOADED);
+  self->mount_namespace_in_use = TRUE;
+}
+
+/**
  * ostree_sysroot_get_path:
  * @self:
  *
@@ -238,6 +265,7 @@ ostree_sysroot_get_path (OstreeSysroot  *self)
   return self->path;
 }
 
+/* Open a directory file descriptor for the sysroot if we haven't yet */
 static gboolean
 ensure_sysroot_fd (OstreeSysroot          *self,
                    GError                **error)
@@ -251,13 +279,51 @@ ensure_sysroot_fd (OstreeSysroot          *self,
   return TRUE;
 }
 
+/* Remount /sysroot read-write if necessary */
+gboolean
+_ostree_sysroot_ensure_writable (OstreeSysroot      *self,
+                                 GError            **error)
+{
+  /* Do nothing if no mount namespace is in use */
+  if (!self->mount_namespace_in_use)
+    return TRUE;
+
+  /* If a mount namespace is in use, ensure we're initialized */
+  if (!ostree_sysroot_initialize (self, error))
+    return FALSE;
+
+  /* If we aren't operating on a booted system, then we don't
+   * do anything with mounts.
+   */
+  if (!self->root_is_ostree_booted)
+    return TRUE;
+
+  /* Check if /sysroot is a read-only mountpoint */
+  struct statvfs stvfsbuf;
+  if (statvfs ("/sysroot", &stvfsbuf) < 0)
+    return glnx_throw_errno_prefix (error, "fstatvfs(/sysroot)");
+  if ((stvfsbuf.f_flag & ST_RDONLY) == 0)
+    return TRUE;
+
+  /* OK, let's remount writable. */
+  if (mount ("/sysroot", "/sysroot", NULL, MS_REMOUNT | MS_RELATIME, "") < 0)
+    return glnx_throw_errno_prefix (error, "Remounting /sysroot read-write");
+
+  /* Reopen our fd */
+  glnx_close_fd (&self->sysroot_fd);
+  if (!ensure_sysroot_fd (self, error))
+    return FALSE;
+
+  return TRUE;
+}
+
 /**
  * ostree_sysroot_get_fd:
  * @self: Sysroot
  *
- * Access a file descriptor that refers to the root directory of this
- * sysroot.  ostree_sysroot_load() must have been invoked prior to
- * calling this function.
+ * Access a file descriptor that refers to the root directory of this sysroot.
+ * ostree_sysroot_initialize() (or ostree_sysroot_load()) must have been invoked
+ * prior to calling this function.
  *
  * Returns: A file descriptor valid for the lifetime of @self
  */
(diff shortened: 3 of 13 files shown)


cgwalters:
WIP for https://github.com/ostreedev/ostree/issues/1265

src/switchroot/ostree-remount.c
```diff
@@ -95,8 +95,11 @@ main(int argc, char *argv[])
       exit (EXIT_SUCCESS);
     }
 
-  do_remount ("/sysroot");
   do_remount ("/var");
 
+  struct stat stbuf;
+  if (lstat ("/sysroot/ostree/.sysroot-ro", &stbuf) < 0)
```
cgwalters: If someone has a better idea than this magic file...I'm all ears.  We could do a kernel argument I guess, but it's not obviously better?

cgwalters:
OK did the trivial rebase :surfing_man: for this, and was testing out this patch to cosa:
```
diff --git a/src/create_disk.sh b/src/create_disk.sh
index 2f9fd779..3f01644c 100755
--- a/src/create_disk.sh
+++ b/src/create_disk.sh
@@ -111,6 +111,9 @@ fi
 
 # we use pure BLS, so don't need grub2-mkconfig
 ostree config --repo rootfs/ostree/repo set sysroot.bootloader none
+# Opt-in to https://github.com/ostreedev/ostree/pull/1767 AKA
+# https://github.com/ostreedev/ostree/issues/1265
+ostree config --repo rootfs/ostree/repo set sysroot.readonly true
 
 if [ "$arch" != "s390x" ]; then
 	# install uefi grub
```

And FCOS goes into emergency mode since `systemd-tmpfiles-setup.service` fails due to `/var` being read-only. I guess the pile of hacks in `ostree-remount.c` isn't working right.

cgwalters:
Toying with
```
diff --git a/src/switchroot/ostree-remount.c b/src/switchroot/ostree-remount.c
index 4b4f5356..f36ecc48 100644
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -120,47 +120,25 @@ main(int argc, char *argv[])
    */
   const bool sysroot_readonly = sysroot_is_configured_ro ();
   if (!sysroot_readonly)
-    do_remount ("/sysroot", !sysroot_readonly);
+    do_remount ("/sysroot", TRUE);
   else
     {
-      /* Since /sysroot is the real physical root, we can't simply remount it
-       * read-only here, as that'd affect e.g. /etc in and also /var if it's not
-       * a separate mount. Instead, we make new read-only bind mount to it,
-       * unmount the original, then move the bind mount to /sysroot.
+      do_remount ("/sysroot", FALSE);
+
+      /* Now, /etc is not normally a bind mount, but remounting the
+       * sysroot above made it read-only since it's on the same filesystem.
+       * Make it a self-bind mount, so we can then mount it read-write.
        */
-      static const char tmp_sysroot[] = "/etc/ostree-sysroot.tmp";
-      static const char sysroot[] = "/sysroot";
-
-      /* This temporary lives in /etc since it needs to be on the same mount. */
-      if (mkdir (tmp_sysroot, 0755) < 0)
-        err (EXIT_FAILURE, "mkdir(%s)", tmp_sysroot);
-      /* Make it a read-only bind mount to /sysroot */
-      if (mount (sysroot, tmp_sysroot, NULL, MS_BIND | MS_PRIVATE, NULL) < 0)
-        err (EXIT_FAILURE, "failed to bind mount %s %s", sysroot, tmp_sysroot);
-      if (mount (tmp_sysroot, tmp_sysroot, NULL, MS_BIND | MS_RDONLY | MS_REMOUNT, NULL) < 0)
-        err (EXIT_FAILURE, "failed to remount ro %s", tmp_sysroot);
-      if (umount (sysroot) < 0)
-        err (EXIT_FAILURE, "while remounting %s read-only: umount", sysroot);
-
-      /* HACK: We can't move a mount that's under a shared namespace. So we
-       * briefly make the sysroot private so that we can move the mount. This
-       * does introduce a race condition where if e.g. another process mounted
-       * something in / it wouldn't be visible in other mount namespaces.  But
-       * we're running quite early, before e.g. any container runtimes should
-       * be starting.
+      if (mount ("/etc", "/etc", NULL, MS_BIND, NULL) < 0)
+        err (EXIT_FAILURE, "failed to make /etc a bind mount");
+      do_remount ("/etc", TRUE);
+      /* If /var was created as as an OSTree default bind mount (instead of being a separate filesystem)
+       * then remounting the root mount read-only also remounted it.
+       * So just like /etc, we need to make it read-write by default.
+       * If it was a separate filesystem, we expect it to be writable anyways,
+       * so it doesn't hurt to remount it if so.
        */
-      if (mount ("none", "/", NULL, MS_PRIVATE, NULL) < 0)
-        err (EXIT_FAILURE, "making / private temporarily");
-      /* Do the move */
-      if (mount (tmp_sysroot, sysroot, NULL, MS_MOVE, NULL) < 0)
-        err (EXIT_FAILURE, "failed to move read-only %s mount", sysroot);
-      /* Make / shared again */
-      if (mount ("none", "/", NULL, MS_SHARED, NULL) < 0)
-        err (EXIT_FAILURE, "making / shared again");
-
-      /* And clean up */
-      if (rmdir (tmp_sysroot) < 0)
-        err (EXIT_FAILURE, "rmdir(%s)", tmp_sysroot);
+      do_remount ("/var", TRUE);
     }
 
   exit (EXIT_SUCCESS);
```

instead.

cgwalters:
OK, the problem was ordering around `var.mount`; fixed that and things seem to work.  Needs more testing around upgrades, and we now also have the cases of:

 - FCOS but not configured this way
 - FAH/Silverblue which don't start out with `rw` on kernel cmdline

cgwalters:
OK, lifting WIP on this!  I still need more testing, but it's ready for review.

I have a working FCOS with this and:

 - https://github.com/coreos/coreos-assembler/pull/736
 - https://github.com/projectatomic/rpm-ostree/pull/1896

Including e.g. package layering, etc.

src/switchroot/ostree-remount.c
```diff
@@ -54,20 +59,33 @@ do_remount (const char *target)
...
+  struct stat stbuf;
+  static const char config_path[] = "/ostree/repo/config";
+  if (stat (config_path, &stbuf) != 0)
+    return false;
+
+  g_autoptr(GKeyFile) keyfile = g_key_file_new ();
+  if (!g_key_file_load_from_file (keyfile, config_path, 0, NULL))
+    return false;
+
+  return g_key_file_get_boolean (keyfile, "sysroot", "readonly", NULL);
```
jlebon: Why do we need to introduce a new key here instead of just handling it based on `ro` being in `/etc/fstab`? IIUC, for FCOS at least, it'll still be `rw` at initrd time to avoid the machine-id mess, and then will be re-mounted `ro` by systemd right before `ostree-remount.service` (though we could still recheck & remount here as we do right now).

Might've missed discussions around this, though keeping the same API as on other traditional systems I think is worth a lot.

src/libostree/ostree-sysroot.c
```diff
@@ -251,13 +279,53 @@ ensure_sysroot_fd (OstreeSysroot          *self,
...
+    return TRUE;
+
+  /* OK, let's remount writable. */
+  if (mount ("/sysroot", "/sysroot", NULL, MS_REMOUNT | MS_RELATIME, "") < 0)
+    return glnx_throw_errno_prefix (error, "Remounting /sysroot read-write");
+
+  /* Reopen our fd */
+  glnx_close_fd (&self->sysroot_fd);
+  if (!ensure_sysroot_fd (self, error))
+    return FALSE;
```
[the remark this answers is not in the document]
cgwalters: Once we remount it read-write, we'll return due to
```
  if ((stvfsbuf.f_flag & ST_RDONLY) == 0)
    return TRUE;
```
right?

src/libostree/ostree-sysroot.c
```diff
@@ -251,13 +279,53 @@ ensure_sysroot_fd (OstreeSysroot          *self,
...
+  /* Check if /sysroot is a read-only mountpoint */
+  struct statvfs stvfsbuf;
+  if (statvfs ("/sysroot", &stvfsbuf) < 0)
+    return glnx_throw_errno_prefix (error, "fstatvfs(/sysroot)");
+  if ((stvfsbuf.f_flag & ST_RDONLY) == 0)
+    return TRUE;
+
+  /* OK, let's remount writable. */
+  if (mount ("/sysroot", "/sysroot", NULL, MS_REMOUNT | MS_RELATIME, "") < 0)
+    return glnx_throw_errno_prefix (error, "Remounting /sysroot read-write");
```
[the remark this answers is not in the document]
cgwalters: We currently do that opportunistically later in deploy.

src/switchroot/ostree-remount.c
cgwalters: As of now, we don't have an `/etc/fstab` in FCOS anymore.

The other thing is...we're trying to create a "middle ground" around "read-only".  Here the physical root *is* writable - it has `/etc` and `/var` by default.  Wouldn't you expect having `ro` in `/etc/fstab` would be truly read-only?

This is really a magical ostree-specific flag I think to opt-in to new behavior for systems builders, not really something admins should see/change directly.

src/ostree/ot-main.c
```diff
@@ -434,10 +435,42 @@ ostree_admin_option_context_parse (GOptionContext *context,
...
+       * namespace and tell the sysroot that we've done so. See the docs for
+       * ostree_sysroot_set_mount_namespace_in_use().
+       *
+       * This is a conservative approch; we could just always
+       * unshare() too.
+       */
+      int sysroot_fd = ostree_sysroot_get_fd (sysroot);
+      g_assert_cmpint (sysroot_fd, !=, -1);
+
+      glnx_autofd int sysroot_subdir_fd = glnx_opendirat_with_errno (sysroot_fd, "sysroot", TRUE);
```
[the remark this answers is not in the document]
cgwalters: This is about handling the "booted" case, where sysroot_fd points to the deployment dir.

It looks like we need to possibly expose the `is_physical` boolean.

src/switchroot/ostree-remount.c
```diff
@@ -54,20 +59,33 @@ do_remount (const char *target)
...
-      if (errno != EINVAL)
-        err (EXIT_FAILURE, "failed to remount %s", target);
-    }
-  else
-    printf ("Remounted: %s\n", target);
+
+  int mnt_flags = MS_REMOUNT | MS_SILENT;
+  if (!writable)
+    mnt_flags |= MS_RDONLY;
+  if (mount (target, target, NULL, mnt_flags, NULL) < 0)
```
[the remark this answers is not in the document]
cgwalters: Good catch, though hmm...I feel like that's probably not necessary and will mask bugs, but...probably better to be conservative?

src/switchroot/ostree-remount.c
jlebon: > The other thing is...we're trying to create a "middle ground" around "read-only". Here the physical root is writable - it has /etc and /var by default. Wouldn't you expect having ro in /etc/fstab would be truly read-only?

OK, I follow. Hmm, I think one confusion here is that "read-only root" can mean a lot of things. Particularly `/etc` could be interpreted as being covered (and I think it'd be really cool to support this too). I wonder if `sysroot.readonly` shouldn't be an enum instead. E.g. `none` (status quo), `base` (everything except `/var` and `/etc`) and `full` (just `/var` is read-write). WDYT? (Not suggesting to implement `full` here, just keeping the option open so we leverage the same knob).

src/libostree/ostree-sysroot.c
```diff
@@ -268,6 +334,21 @@ ostree_sysroot_get_fd (OstreeSysroot *self)
...
 
+/**
+ * ostree_sysroot_is_booted:
+ * @self: Sysroot
+ *
+ * Can only be invoked after `ostree_sysroot_initialize()`.
+ * 
+ * Returns: %TRUE iff the sysroot points to a booted deployment
+ */
+int
```
jlebon: Minor: `gboolean`

src/libostree/ostree-sysroot.c
```diff
@@ -268,6 +334,21 @@ ostree_sysroot_get_fd (OstreeSysroot *self)
   return self->sysroot_fd;
 }
 
+/**
+ * ostree_sysroot_is_booted:
+ * @self: Sysroot
+ *
+ * Can only be invoked after `ostree_sysroot_initialize()`.
+ * 
+ * Returns: %TRUE iff the sysroot points to a booted deployment
```
jlebon: Missing `Since`

src/libostree/ostree-sysroot-private.h
```diff
@@ -40,6 +40,12 @@ typedef enum {
   OSTREE_SYSROOT_DEBUG_TEST_STAGED_PATH = 1 << 3,
 } OstreeSysrootDebugFlags;
 
+typedef enum {
+      OSTREE_SYSROOT_LOAD_STATE_NONE, /* ostree_sysroot_new() was called */
+      OSTREE_SYSROOT_LOAD_STATE_INIT, /* We have a file descriptor */
```
jlebon: How about `We've loaded basic sysroot state and have an fd`?

src/libostree/ostree-sysroot-private.h
```diff
@@ -40,6 +40,12 @@ typedef enum {
...
+      OSTREE_SYSROOT_LOAD_STATE_LOADED, /* We've loaded all of the deploments */
```
jlebon: `deployments`

src/libostree/ostree-impl-system-generator.c
```diff
@@ -163,6 +164,32 @@ _ostree_impl_system_generator (const char *ostree_cmdline,
...
+   * 
+   * However, Fedora CoreOS is an Ignition based system and starts out writable
+   * via rw on the kernel command line.
+   * 
+   * Now, we want to support a read-only /sysroot: https://github.com/ostreedev/ostree/issues/1265
+   * And the way that's currently implemented is in ostree-remount.service.
+   * Ideally systemd would support "Options=bind,rw" to force on a writable bind mount,
+   * but it currently doesn't.  So we handle the ordering here.
+   */
+  if (sysroot_currently_writable)
```
jlebon: I'm still confused by this; doesn't `ostree-remount` always remount `/var` anyway? So can't we unconditionally run `Before=ostree-remount.service`?

cgwalters: OK, you're right we can clean this up a lot.  Done!

cgwalters:
>  I wonder if sysroot.readonly shouldn't be an enum instead. E.g. none (status quo), base (everything except /var and /etc) and full (just /var is read-write).

I feel like they're different knobs.  `/sysroot` is purely ostree territory and this config option in the repo is for OS vendors (e.g. FCOS) to toggle.  But `/etc` is something for the admin to flip, and it's not clear  to me we want admins to be editing the repo config for things like this.

cgwalters:
Maybe the admin API is `touch /etc/.ostree-sealed` or something, and we detect that in `ostree-remount.service`?

cgwalters:
(Somewhat related to this topic though is...I want to change Ignition to do something like write an ostree ref instead.  If it did that, all of the fs-verity and other work would come into play.  And so the "API" for a sealed etc might instead be pointing at an ostree ref or so)

cgwalters:
xref https://github.com/openshift/machine-config-operator/issues/1190 for /etc

jlebon:
OK, this LGTM. Seeing as we have no practical CI covering this: were you able to test this last version successfully in e.g. FCOS?

cgwalters:
Yep, tested with the latest FCOS and [assembler PR](https://github.com/coreos/coreos-assembler/pull/736).

PR #1749, opened 2018-10-04
Don't delete refs having aliases

sinnykumari:
Deleting a ref with aliases makes them dangling. In such
cases, display an error message to the user.

Fixes #1597

Signed-off-by: Sinny Kumari <sinny@redhat.com>

PR #1768, opened 2018-10-30
bin/refs: Disallow creating broken aliases

jlebon:
This is the alias version of #1749. I.e. we want to make sure that one
can't even create an alias which would end up dangling.

See also: https://pagure.io/releng/issue/7891


cgwalters:
@rh-atomic-bot r=sinnykumari 8002073

PR #1769, opened 2018-10-31
bin/pull-local: Add --commit-metadata-only

jlebon:
It might be "local", but e.g. we may be crossing filesystems. So there
are valid use cases for only wanting to pull the commit metadata with
`pull-local`.


jlebon:
Need to add a test for this.

dustymabe:
+1 thanks for this

jlebon:
OK, added test! :arrow_up:

cgwalters:
@rh-atomic-bot r+ 73c2418

PR #1771, opened 2018-10-31
lib/commit: Copy user.ostreemeta only for bare-user

jlebon:
When falling back to copying objects when importing them into a
bare-user repo, we only actually need to transfer over the
`user.ostreemeta` xattr.

This allows the destination repo to be on a separate filesystem that
might not even support `security.selinux`. (I hit this while importing
over virtio-9p).

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -4254,15 +4254,29 @@ import_one_object_direct (OstreeRepo    *dest_repo,
         G_IN_SET (src_repo->mode, OSTREE_REPO_MODE_BARE, OSTREE_REPO_MODE_BARE_USER);
       if (src_is_bare_or_bare_user && !OSTREE_OBJECT_TYPE_IS_META(objtype))
         {
-          g_autoptr(GVariant) xattrs = NULL;
+          if (src_repo == OSTREE_REPO_MODE_BARE)
+            {
+              g_autoptr(GVariant) xattrs = NULL;
+              if (!glnx_fd_get_all_xattrs (src_fd, &xattrs,
+                                           cancellable, error))
+                return FALSE;
+              if (!glnx_fd_set_all_xattrs (tmp_dest.fd, xattrs,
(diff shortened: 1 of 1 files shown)


jlebon:
I don't actually need this anymore, but figured I'd post it up anyway since I had already written it up.

dustymabe:
> I don't actually need this anymore, but figured I'd post it up anyway since I had already written it up.

thanks for doing that especially if it will help others who might hit the same problem

cgwalters:
@rh-atomic-bot r+ 3bcfddd

cgwalters:
Haven't fully verified but I think this is the commit that broke ostree in Anaconda, see https://openqa.fedoraproject.org/tests/345339#step/_do_install_and_reboot/4

I feel like this code isn't fully handling the matrix of possible (src, dest) types, need to think through that more, but there is one obvious bug:

```
$ git diff
index 533cf03a..c95e0b61 100644
@@ -4269,7 +4269,7 @@ import_one_object_direct (OstreeRepo    *dest_repo,
         G_IN_SET (src_repo->mode, OSTREE_REPO_MODE_BARE, OSTREE_REPO_MODE_BARE_USER);
       if (src_is_bare_or_bare_user && !OSTREE_OBJECT_TYPE_IS_META(objtype))
         {
-          if (src_repo == OSTREE_REPO_MODE_BARE)
+          if (src_repo->mode == OSTREE_REPO_MODE_BARE)
             {
               if (!glnx_fd_get_all_xattrs (src_fd, &xattrs,
```

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1803

dustymabe:
> Haven't fully verified but I think this is the commit that broke ostree in Anaconda, see https://openqa.fedoraproject.org/tests/345339#step/_do_install_and_reboot/4
> 
> I feel like this code isn't fully handling the matrix of possible (src, dest) types, need to think through that more, but there is one obvious bug:

thanks for catching this!

PR #1772, opened 2018-11-02
lib/commit: Add devino_cache_hits to txn stats

jlebon:
I found this useful while hacking on rpm-ostree but I think it might be
useful enough to upstream. This stat is really helpful for validating
that a pipeline is hitting the devino cache sweet spot.

### DIFF
--- a/src/libostree/ostree-repo.h
+++ b/src/libostree/ostree-repo.h
@@ -291,8 +291,9 @@ struct _OstreeRepoTransactionStats {
   guint content_objects_total;
   guint content_objects_written;
   guint64 content_bytes_written;
+  guint devino_cache_hits;
 
-  guint64 padding1;
+  guint   padding1;
   guint64 padding2;
   guint64 padding3;
   guint64 padding4;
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-repo.h
```diff
@@ -291,8 +291,9 @@ struct _OstreeRepoTransactionStats {
   guint content_objects_total;
   guint content_objects_written;
   guint64 content_bytes_written;
+  guint devino_cache_hits;
 
-  guint64 padding1;
+  guint   padding1;
```
jlebon: I was going to add a `G_STATIC_ASSERT` for this to make sure `int` is always 32-bit on whatever 32-bit platform we're compiled on. (We're already guaranteed this on 64-bit platforms.) But it turns out glib does this already: https://gitlab.gnome.org/GNOME/glib/blob/460cc723ad6c56f70f24ff8794f669ec522a7942/glib/glib-init.c#L101.

cgwalters:
@rh-atomic-bot r+ 61cffd4

jlebon:
@rh-atomic-bot retry

PR #1773, opened 2018-11-06
lib/repo: Search a list of paths in gpgkeypath for gpg keys

### LINKED ISSUE #773: make gpgkeypath 🗝 support being a list 📓
In f244c702772c69378099685316033d4a6f7b862c we added support for importing a single file, but there are use cases for multiple such as [Fedora GPG keys](https://bugzilla.redhat.com/show_bug.cgi?id=1393545) across major versions.

We probably should try parsing this entry as a single string first, then a list, since `GKeyFile` requires the trailing `;` so it wouldn't be compatible to just make it a list.

rfairley:
This allows specifying gpgpath as a list of
paths that can point to a file or a directory. If a directory path
is given, paths to all regular files in the directory are added
to the remote as gpg ascii keys. If the path is not a directory,
the file is directly added (whether regular file, empty - errors
will be reported later when verifying gpg keys e.g. when pulling).

Adding the gpgkeypath property looks like:

ostree --repo=repo remote add --set=gpgpath="/path/key1.asc,/path/keys.d" R1 https://example.com/some/remote/ostree/repo

Closes #773

**Still left to do**

- [x] Update documentation of gpgkeypath

**Other considerations**

- Checks before adding the keys in `_ostree_gpg_verifier_add_key_ascii_file` and if/how checks will be reported. Right now, entries in the list that are not directories are added automatically (whether regular, or don't exist, ...) (same as before). If the entry is a directory, all regular files in that directory are added, without traversing directories
- Regexp matching of list entries

### DIFF
--- a/src/libostree/ostree-gpg-verifier.c
+++ b/src/libostree/ostree-gpg-verifier.c
@@ -334,7 +334,7 @@ _ostree_gpg_verifier_add_keyfile_path (OstreeGpgVerifier   *self,
 }
 
 /* Add files that exist one level below the directory at @path as ascii
- * key files. If @path exists and cannot be opened as a directory,
+ * key files. If @path cannot be opened as a directory,
  * an error is returned.
  */
 gboolean
--- a/src/libostree/ostree-gpg-verifier.c
+++ b/src/libostree/ostree-gpg-verifier.c
@@ -304,6 +304,70 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
   g_ptr_array_add (self->key_ascii_files, g_strdup (path));
 }
 
+/* Check if the file at @path is a directory, and add ascii key files that
+ * exist one level below the directory. If @path does not lead to a directory,
+ * the file is added.
+ */
+void
+_ostree_gpg_verifier_add_key_ascii_files_check_dir (OstreeGpgVerifier *self,
+                                                    const char        *path,
+                                                    GCancellable      *cancellable,
+                                                    GError           **error)
+{
+  GFile *file;
+  GFileType file_type;
+
+  file = g_file_new_for_path (path);
+  file_type = g_file_query_file_type (file, G_FILE_QUERY_INFO_NONE, cancellable);
+
+  if (file_type == G_FILE_TYPE_DIRECTORY)
+    {
+      GFileEnumerator *direnum;
+      g_autofree char *sep = NULL;
+
+      if (!g_str_has_suffix (path, "/"))
+        sep = g_strdup ("/");
+
+      direnum = g_file_enumerate_children (file, OSTREE_GIO_FAST_QUERYINFO,
+                                          G_FILE_QUERY_INFO_NONE, cancellable, error);
+      if (direnum)
+        {
+          while (TRUE)
+            {
+              GFileInfo *child_info;
+              GFile *child;
+              const char *name;
+              guint32 type;
+
+              if (!g_file_enumerator_iterate (direnum, &child_info, &child,
+                                              NULL, error))
+                break;
+              if (child_info == NULL)
+                break;
+
+              name = g_file_info_get_attribute_byte_string (child_info, "standard::name");
+              type = g_file_info_get_attribute_uint32 (child_info, "standard::type");
+
+              if (type == G_FILE_TYPE_REGULAR)
+                {
+                  g_autofree char *child_path = NULL;
+
+                  child_path = g_strjoin (sep, path, name, NULL);
+
+                  _ostree_gpg_verifier_add_key_ascii_file (self, child_path);
+                }
+            }
+        }
+
+      g_object_unref (direnum);
+      g_object_unref (file);
+    }
+  else
+    {
+      _ostree_gpg_verifier_add_key_ascii_file (self, path);
+    }
+}
+
 gboolean
 _ostree_gpg_verifier_add_keyring_dir (OstreeGpgVerifier   *self,
                                       GFile               *path,
--- a/src/libostree/ostree-gpg-verifier.c
+++ b/src/libostree/ostree-gpg-verifier.c
@@ -310,24 +310,24 @@ _ostree_gpg_verifier_add_keyfile_path (OstreeGpgVerifier   *self,
                                        GCancellable        *cancellable,
                                        GError             **error)
 {
-  GError *temp_error = NULL;
-  if (!_ostree_gpg_verifier_add_keyfile_dir_at(self, AT_FDCWD, path,
-                                               cancellable, &temp_error))
+  g_autoptr(GError) temp_error = NULL;
+  if (!_ostree_gpg_verifier_add_keyfile_dir_at (self, AT_FDCWD, path,
+                                                cancellable, &temp_error))
     {
-      if (temp_error)
-        {
-          /* If failed due to not being a directory, add the file as an ascii key. */
-          if (g_error_matches (temp_error, G_IO_ERROR, G_IO_ERROR_NOT_DIRECTORY))
-            {
-              g_clear_error (&temp_error);
+      g_assert (temp_error);
 
-              _ostree_gpg_verifier_add_key_ascii_file (self, path);
-            }
-          else
-            {
-              g_propagate_error (error, temp_error);
-              return FALSE;
-            }
+      /* If failed due to not being a directory, add the file as an ascii key. */
+      if (g_error_matches (temp_error, G_IO_ERROR, G_IO_ERROR_NOT_DIRECTORY))
+        {
+          g_clear_error (&temp_error);
+
+          _ostree_gpg_verifier_add_key_ascii_file (self, path);
+        }
+      else
+        {
+          g_propagate_error (error, g_steal_pointer (&temp_error));
+
+          return FALSE;
         }
     }
   return TRUE;
@@ -362,7 +362,7 @@ _ostree_gpg_verifier_add_keyfile_dir_at (OstreeGpgVerifier   *self,
         struct dirent *dent;
 
         if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent,
-                                                        cancellable, error))
+                                                         cancellable, error))
           return FALSE;
         if (dent == NULL)
           break;
--- a/src/libostree/ostree-gpg-verifier.c
+++ b/src/libostree/ostree-gpg-verifier.c
@@ -304,7 +304,7 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
   g_ptr_array_add (self->key_ascii_files, g_strdup (path));
 }
 
-void
+gboolean
 _ostree_gpg_verifier_add_keyfile_path (OstreeGpgVerifier   *self,
                                        const char          *path,
                                        GCancellable        *cancellable,
@@ -323,12 +323,14 @@ _ostree_gpg_verifier_add_keyfile_path (OstreeGpgVerifier   *self,
 
               _ostree_gpg_verifier_add_key_ascii_file (self, path);
             }
-        }
-      else
-        {
-          g_propagate_error (error, temp_error);
+          else
+            {
+              g_propagate_error (error, temp_error);
+              return FALSE;
+            }
         }
     }
+  return TRUE;
 }
 
 /* Add files that exist one level below the directory at @path as ascii
@@ -342,40 +344,36 @@ _ostree_gpg_verifier_add_keyfile_dir_at (OstreeGpgVerifier   *self,
                                          GCancellable        *cancellable,
                                          GError             **error)
 {
-  gboolean dir_exist = FALSE;
   g_auto(GLnxDirFdIterator) dfd_iter = { 0, };
 
-  if (!ot_dfd_iter_init_allow_noent (dfd, path,
-                                    &dfd_iter, &dir_exist, error))
+  if (!glnx_dirfd_iterator_init_at (dfd, path, FALSE,
+                                    &dfd_iter, error))
     return FALSE;
 
-  if (dir_exist)
-    {
-      g_autofree char *sep = NULL;
+    g_autofree char *sep = NULL;
 
-      g_debug ("Adding GPG keyfile dir %s to verifier", path);
+    g_debug ("Adding GPG keyfile dir %s to verifier", path);
 
-      if (!g_str_has_suffix (path, "/"))
-        sep = g_strdup ("/");
+    if (!g_str_has_suffix (path, "/"))
+      sep = g_strdup ("/");
 
-      while (TRUE)
-        {
-          struct dirent *dent;
+    while (TRUE)
+      {
+        struct dirent *dent;
 
-          if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent,
-                                                          cancellable, error))
-            return FALSE;
-          if (dent == NULL)
-            break;
+        if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent,
+                                                        cancellable, error))
+          return FALSE;
+        if (dent == NULL)
+          break;
 
-          if (dent->d_type != DT_REG)
-            continue;
+        if (dent->d_type != DT_REG)
+          continue;
 
-          g_autofree char *iter_path = g_strjoin (sep, path, dent->d_name, NULL);
+        g_autofree char *iter_path = g_strjoin (sep, path, dent->d_name, NULL);
 
-          _ostree_gpg_verifier_add_key_ascii_file (self, iter_path);
-        }
-    }
+        _ostree_gpg_verifier_add_key_ascii_file (self, iter_path);
+      }
 
   return TRUE;
 }
--- a/src/libostree/ostree-gpg-verifier.c
+++ b/src/libostree/ostree-gpg-verifier.c
@@ -361,7 +361,6 @@ _ostree_gpg_verifier_add_keyfile_dir_at (OstreeGpgVerifier   *self,
       while (TRUE)
         {
           struct dirent *dent;
-          g_autofree char *iter_path = NULL;
 
           if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent,
                                                           cancellable, error))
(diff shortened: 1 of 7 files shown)


src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,6 +304,70 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
...
+
+                  _ostree_gpg_verifier_add_key_ascii_file (self, child_path);
+                }
+            }
+        }
+
+      g_object_unref (direnum);
+      g_object_unref (file);
+    }
+  else
```
rfairley: I did try a check here for `else if (file_type == G_FILE_TYPE_REGULAR)`. But with this check, and specifying a nonexisting file like in https://github.com/ostreedev/ostree/pull/1773/files#diff-9a33529a3274e91605ccacb39789b065R173, the test reports that there `are keys in the keychain, but none that are valid`.

It should report  `file/directory not found` somewhere, but it'd need to report it before the `pull` command I think

dustymabe:
> ostree --repo=repo remote add --set=gpgpath="/path/key1.asc;/path/keys.d" R1 https://example.com/some/remote/ostree/repo

is that a semi-colon `;` above ? is there a reason to use a semi-colon over a comma?

rfairley:
@dustymabe Initially I was thinking the keyfile where `gpgpath` ends up would need semi-colons so that glibc could parse it (https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html#glib-Key-value-file-parser.description). But the delimiter for glibc to parse keys with multiple values can be changed to any character, so there is no real need to use semi-colons. Commas are also mentioned in that link. I'd be happy to go with commas, they do look nicer than semi-colons as well.

src/libostree/ostree-repo.c
```diff
@@ -5105,7 +5107,26 @@ _ostree_repo_gpg_verify_data_internal (OstreeRepo    *self,
         return NULL;
 
       if (gpgkeypath)
-        _ostree_gpg_verifier_add_key_ascii_file (verifier, gpgkeypath);
+        {
+          if (strchr (gpgkeypath, ','))
```
puiterwijk: Is there any reason you are going for these two different code paths, rather than just always doing the `g_strsplit` and loop, since I would assume that running the split on a string without `,` would result in a vector with one entry?

rfairley: At first I was following from @cgwalters comment on the issue https://github.com/ostreedev/ostree/issues/773#issue-218950903 that a check for the case of a single string should be done first, before interpreting as multiple due to the trailing `;` being required.

Looks like that requirement has changed though since then. In https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-1.1.html#value-types the trailing `;` is now optional, it used to be required in 1.0.

I see the way I've done it though with `split_string`, the trailing separator is not properly handled anyway as an empty string would be added at the end.

`g_key_file_get_string_list` should handle the single case and any trailing `,` properly. I'll add this in, have this always be called and loop. Will also add a test to make sure the trailing `,` is handled.

Thanks for pointing that out!

src/libostree/ostree-gpg-verifier.c
cgwalters: The cleanest API for this I think is to use `ot_dfd_iter_init_allow_noent()`.

src/libotutil/ot-keyfile-utils.c
```diff
@@ -101,6 +101,47 @@ ot_keyfile_get_value_with_default (GKeyFile      *keyfile,
...
+  GError *temp_error = NULL;
+  g_autofree char **ret_value = NULL;
+
+  g_return_val_if_fail (keyfile != NULL, ret);
+  g_return_val_if_fail (section != NULL, ret);
+  g_return_val_if_fail (key != NULL, ret);
+
+  g_key_file_set_list_separator (keyfile, separator);
+
+  ret_value = g_key_file_get_string_list (keyfile, section, key, NULL, &temp_error);
```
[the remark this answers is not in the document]
rfairley: Addressed this

src/libotutil/ot-keyfile-utils.c
```diff
@@ -101,6 +101,47 @@ ot_keyfile_get_value_with_default (GKeyFile      *keyfile,
...
+gboolean
+ot_keyfile_get_string_list_with_default (GKeyFile      *keyfile,
+                                         const char    *section,
+                                         const char    *key,
+                                         char           separator,
+                                         char         **default_value,
+                                         char        ***out_value,
+                                         GError       **error)
+{
+  gboolean ret = FALSE;
```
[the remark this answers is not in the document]
rfairley: Addressed

src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,6 +304,70 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
...
+
+  if (file_type == G_FILE_TYPE_DIRECTORY)
+    {
+      GFileEnumerator *direnum;
+      g_autofree char *sep = NULL;
+
+      if (!g_str_has_suffix (path, "/"))
+        sep = g_strdup ("/");
+
+      direnum = g_file_enumerate_children (file, OSTREE_GIO_FAST_QUERYINFO,
```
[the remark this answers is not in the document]
rfairley: Addressed so this declares-and-initializes in https://github.com/ostreedev/ostree/pull/1773/commits/e75ac9d7c02b4efe95b5054d4622c166f0b0a3bf.

I did also try with the `libglnx` API in https://github.com/ostreedev/ostree/pull/1773/commits/9c7ad70cef213539a3598980366df60dc55ba673.

src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,6 +304,70 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
...
+                                          G_FILE_QUERY_INFO_NONE, cancellable, error);
+      if (direnum)
+        {
+          while (TRUE)
+            {
+              GFileInfo *child_info;
+              GFile *child;
```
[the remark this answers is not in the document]
rfairley: Thanks! Had a look at the examples, and made these `g_autoptr` in https://github.com/ostreedev/ostree/pull/1773/commits/e75ac9d7c02b4efe95b5054d4622c166f0b0a3bf.

src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,6 +304,70 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
   g_ptr_array_add (self->key_ascii_files, g_strdup (path));
 }
 
+/* Check if the file at @path is a directory, and add ascii key files that
+ * exist one level below the directory. If @path does not lead to a directory,
+ * the file is added.
+ */
+void
+_ostree_gpg_verifier_add_key_ascii_files_check_dir (OstreeGpgVerifier *self,
```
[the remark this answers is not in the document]
rfairley: Done - in https://github.com/ostreedev/ostree/pull/1773/commits/e75ac9d7c02b4efe95b5054d4622c166f0b0a3bf. I did end up changing this, to `_ostree_gpg_verifier_add_keyfile_dir_at` in https://github.com/ostreedev/ostree/pull/1773/commits/9c7ad70cef213539a3598980366df60dc55ba673.

rfairley:
Thanks @cgwalters for the review and helpful comments! I pushed a couple of fixup commits.

The first https://github.com/ostreedev/ostree/pull/1773/commits/e75ac9d7c02b4efe95b5054d4622c166f0b0a3bf addresses the changes doing things the first way I had - not using the `libglnx` API.

The second, I used `glnx_dirfd_iterator_init_at` in `ostree-gpg-verifier.c`. With that came a bit of restructuring, but it seems to work well overall. One part where I found I still needed to use `GKeyFile` is https://github.com/ostreedev/ostree/pull/1773/commits/9c7ad70cef213539a3598980366df60dc55ba673#diff-23d4c8887c7dda91613c9451bf3fe7c0R5112 to check if this is a directory. Am just having a look now into `ot_dfd_iter_init_allow_noent()` which you mentioned - sorry looked over this comment before. I think using this would avoid the need to use `GKeyFile`.

rfairley:
Pushed another fixup which cleans up the loop in `ostree-repo.c` from the previous commit and uses `ot_dfd_iter_init_allow_noent()`.

This changes behaviour when coming across a path that does not exist. Previously, the path to a file that did not exist would give a message `No file or directory`. This is because the filepath was added to the verifier if it did not exist.

Now, if the path does not exist, the path does not get added to the verifier. No error message indicates this, so this is silent. A test showing this change in behaviour is this https://github.com/ostreedev/ostree/pull/1773/commits/b13e8421573e7ace8b703c3b87ce71836bb6368a#diff-9a33529a3274e91605ccacb39789b065R220.

A benefit of having `NOENT` be silent is that if multiple `gpgkeypath`s are specified, and at least one of them exists and contains a valid key, the pull operation will not fail due to one of the other paths not existing.

The downside is there is no specific error message given letting the user know a path was imported that doesn't exist.

Either having `NOENT` be silent or enforcing that all the paths exist can be done. Will think about this. Any ideas which way to go, or if there is another way to report `NOENT` would be appreciated!

src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,6 +304,70 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
...
+                break;
+
+              name = g_file_info_get_attribute_byte_string (child_info, "standard::name");
+              type = g_file_info_get_attribute_uint32 (child_info, "standard::type");
+
+              if (type == G_FILE_TYPE_REGULAR)
+                {
+                  g_autofree char *child_path = NULL;
+
+                  child_path = g_strjoin (sep, path, name, NULL);
```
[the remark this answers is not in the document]
rfairley: Addressed this

src/libotutil/ot-keyfile-utils.c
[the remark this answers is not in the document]
rfairley: Addressed, with the decl-and-init

rfairley:
^ Latest commits address the decl-and-inits, and I changed the nonexistent path handling to give `No such file or directory` if any of the paths given in the list don't exist. This behaviour is what `_ostree_gpg_verifier_add_keyring_dir()` follows, I think it makes sense to do the same for `_ostree_gpg_verifier_add_keyfile_path()`.

Sorry for all the flip-flopping in the recent changes, will keep the changes settled now.

rfairley:
Updated the manpage.

There are 2 last questions I have:
- I noticed the documentation in [ostree.repo-config.xml](https://github.com/ostreedev/ostree/blob/master/man/ostree.repo-config.xml#L58) mentions the config file follows the latest Freedesktop standards for Desktop Entry Files. While [GNOME](https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html#glib-Key-value-file-parser.description) mentions commas being acceptable instead of semicolons to separate values, [Freedesktop](https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s04.html) mentions only semicolons. Are we strictly following Freedesktop?
- Currently **any** file in a given directory path is imported. Adding a pattern matcher so that only `*.asc` files are imported would be simple to add in, should this be done? I thought about pattern matching from the paths given in the `gpgkeypath` list, to enable globbing, but it does complicate the logic. Maybe add as an enhancement if requested.

jlebon:
> I noticed the documentation in ostree.repo-config.xml mentions the config file follows the latest Freedesktop standards for Desktop Entry Files. While GNOME mentions commas being acceptable instead of semicolons to separate values, Freedesktop mentions only semicolons. Are we strictly following Freedesktop?

Hmm, it would be nice to support both I think, since traditionally keyfiles use semicolons for lists. So that's an easy gotcha to fall into. What do you think of something like:

- parse as a string first, if there are neither commas or semicolons, just return that
- otherwise, use `g_key_file_set_list_separator` depending on above and parse as a list

?

> Currently any file in a given directory path is imported. Adding a pattern matcher so that only *.asc files are imported would be simple to add in, should this be done?

I think the original reason for this ask was to support just passing `/etc/pki/rpm-gpg`, which only holds keys. So I think doing this later when a use case shows up would be fine.

src/libostree/ostree-repo.c
```diff
@@ -5081,7 +5081,8 @@ _ostree_repo_gpg_verify_data_internal (OstreeRepo    *self,
     }
   else if (remote_name != NULL)
     {
-      g_autofree char *gpgkeypath = NULL;
+      char **gpgkeypath_list = NULL;
```
jlebon: You can use `g_auto(GStrv)` for this and drop the `g_strfreev`s.

src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,12 +304,85 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
...
+_ostree_gpg_verifier_add_keyfile_path (OstreeGpgVerifier   *self,
+                                       const char          *path,
+                                       GCancellable        *cancellable,
+                                       GError             **error)
+{
+  GError *temp_error = NULL;
+  if (!_ostree_gpg_verifier_add_keyfile_dir_at(self, AT_FDCWD, path,
+                                               cancellable, &temp_error))
+    {
+      if (temp_error)
```
jlebon: We don't really need this check right? (Or alternatively, let's turn it into a `g_assert (temp_error);` ?)

src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,12 +304,85 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
   g_ptr_array_add (self->key_ascii_files, g_strdup (path));
 }
 
+gboolean
+_ostree_gpg_verifier_add_keyfile_path (OstreeGpgVerifier   *self,
+                                       const char          *path,
+                                       GCancellable        *cancellable,
+                                       GError             **error)
+{
+  GError *temp_error = NULL;
```
jlebon: The pattern we normally use for this is `g_autoptr(GError)` and then `g_propagate_error (error, g_steal_pointer (&temp_error))`.

src/libostree/ostree-repo.c
rfairley: Thanks! Avoids needing to `g_strfreev` before returning early. Addressed in https://github.com/ostreedev/ostree/pull/1773/commits/1e120d2a41107b9e128e1bed2820eceb080a69fb

src/libostree/ostree-gpg-verifier.c
rfairley: `g_assert` sounds good. Addressed in https://github.com/ostreedev/ostree/pull/1773/commits/1e120d2a41107b9e128e1bed2820eceb080a69fb

src/libostree/ostree-gpg-verifier.c
rfairley: Addressed in https://github.com/ostreedev/ostree/pull/1773/commits/1e120d2a41107b9e128e1bed2820eceb080a69fb

jlebon:
> If both ; and , are present in a string parsed for a single key, then we throw an error? Or perhaps default to separating by the ; in that case, but error might be more informative.

Sure, making it an error SGTM!

rfairley:
⬆️ fixup to add the check for both `;` and `,`, returning an error if both are present.

The function `ot_keyfile_get_string_as_list` deals with this check, and returns a null-terminated array of strings (whether the single-string or list-of-strings case).

src/libostree/ostree-repo.c
```diff
@@ -5100,12 +5099,17 @@ _ostree_repo_gpg_verify_data_internal (OstreeRepo    *self,
...
-                                              &gpgkeypath, error))
+      g_auto(GStrv) gpgkeypath_list = NULL;
+
+      if (!ot_keyfile_get_string_as_list (remote->options, remote->group, "gpgkeypath",
+                                          ";,", &gpgkeypath_list, error))
         return NULL;
 
-      if (gpgkeypath)
-        _ostree_gpg_verifier_add_key_ascii_file (verifier, gpgkeypath);
+      if (gpgkeypath_list)
```
jlebon: Minor: adding explicit blocks here would be nice.

src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,12 +304,85 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
...
+                                         GCancellable        *cancellable,
+                                         GError             **error)
+{
+  g_auto(GLnxDirFdIterator) dfd_iter = { 0, };
+
+  if (!glnx_dirfd_iterator_init_at (dfd, path, FALSE,
+                                    &dfd_iter, error))
+    return FALSE;
+
+    g_autofree char *sep = NULL;
```
jlebon: Minor: looks like from here and below indentation is off.

src/libostree/ostree-gpg-verifier.c
```diff
@@ -304,12 +304,85 @@ _ostree_gpg_verifier_add_key_ascii_file (OstreeGpgVerifier *self,
...
+        if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent,
+                                                         cancellable, error))
+          return FALSE;
+        if (dent == NULL)
+          break;
+
+        if (dent->d_type != DT_REG)
+          continue;
+
+        g_autofree char *iter_path = g_strjoin (sep, path, dent->d_name, NULL);
```
jlebon: You can use `g_build_filename` for this and drop the need for `sep`.

cgwalters: Yeah, though strictly speaking...since we're doing fd-relative here, a cleaner approach would be to open the file here, stashing it an another array, and then later using `gpgme_data_new_from_fd()` just like we do for paths.

Or even more strongly, open every file here and then change the verifier to just iterate over the fds.

(This is an optional followup)

rfairley: Ahh, the utility functions are handy. Added this in.

Storing the fds while reading the paths and later iterating over those to add to the verifier sounds clean, saves doing the intermediate work of storing the path. I added a TODO for this.

src/libostree/ostree-repo.c
rfairley: Addressed!

src/libostree/ostree-gpg-verifier.c
rfairley: Addressed!

jlebon:
This looks good to me. Nice work! :+1: 
Will let @cgwalters have a final lookover.

cgwalters:
Very nice, thanks!

@rh-atomic-bot r+ a802199

cgwalters:
@rh-atomic-bot r+ a802199

rfairley:
Thanks for the review @cgwalters and @jlebon !

Issue #1723, opened 2018-09-14
eliminate file duplication during a local untrusted pull


ramcq:
As mentioned in https://github.com/ostreedev/ostree/issues/1720#issuecomment-421061941, we have a problem in Endless OS using Flatpak apps for some pretty large apps. During an install/upgrade of an app, Flatpak runs the networking code and does the pull as the user (for various good reasons, proxies, auth, etc) into a child repo in /var/tmp, and then uses the priveleged system helper to do a local filesystem pull from the user-owner child repo into the real one.

Because a malicious user could keep FDs open and modify the child repo, this has to be an untrusted pull where every file is copied and verified against the configured GPG key of the system remote. On btrfs this uses reflinks, but everywhere else the consequence of this double pull is that twice the disk space is required during an app install or upgrade.

We've got some 5-6GB apps and some systems which ship with eg 32GB MMCs, so this is a very real problem for us - so we're having to use heuristics to disable automatic upgrades when we are low on disk space - but we'd like to come up with a way to mitigate or eliminate this duplication entirely. As far as I'm aware there are no current kernel APIs to revoke an FD or seal a file.

Some approaches we've thought about:
- We thought about running the networking code as a depriveleged and/or sandboxed system user which can only be accessed by the flatpak system helper - as we have bwrap etc - but this doesn't allow normal access to the user's desktop config (proxies, env vars, etc).
- Implement a "destructive pull" which unlinked each source file as it went. This one seems not too bad although we would probably need some help mapping out exactly how/where to do it. It has the downside that we're not able to entirely eliminate the duplication, merely reduce it to per-file rather than the whole ref. Still, 1GB is a lot better than 6GB.
- "Stealing" the file somehow - mv to a root-owned temp folder, chown to root:root, scanning /proc to look for open fds to the file, then proceeding.
- A variant of the above which used the immutable flag to ensure that after verification, no changes could be made...? The problem is that we wouldn't know when we'd be able to clear that immutable flag, potentially not until a reboot.
- Ask the system helper to create the user child for us, and it does so by creating an unaccessible folder and then bind mounting it so that it's accessible to the user to do the pull. It could then unmount the user bind mount to enforce that no FDs are open, before proceeding to verify and hardlink in-place.

/cc @uajain @wjt @pwithnall @mwleeds

wjt:
> It could then unmount the user bind mount to enforce that no FDs are open, before proceeding to verify and hardlink in-place.

[...]

PR #1776, opened 2018-12-02
Assume same ownership and immutable files in case of untrusted local-pulls

uajain:
In the pull code path, if the caller can guarantee that the files
in the local repo are immutable and checksum-verified, then use the
_OSTREE_REPO_IMPORT_FLAGS_TRUSTED flag to try hardlink or reflink
while importing.

This mainly helps flatpak for enabling a hardlink-able local pull
in the --system case during deploy. If the child repo is owned by root
and is not world-writable, one can pass this to flag solve the double
space requirement problem (i.e. copying each object during import) by
using a hardlink-able(or reflink) pull.

https://github.com/ostreedev/ostree/issues/1723

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -4101,12 +4101,18 @@ import_is_bareuser_only_conversion (OstreeRepo *src_repo,
     && objtype == OSTREE_OBJECT_TYPE_FILE;
 }
 
-/* Returns TRUE if we can potentially just call link() to copy an object. */
+/* Returns TRUE if we can potentially just call link() to copy an object;
+ * if untrusted the repos must be owned by the same uid.
+ */
 static gboolean
 import_via_reflink_is_possible (OstreeRepo *src_repo,
                                 OstreeRepo *dest_repo,
-                                OstreeObjectType objtype)
+                                OstreeObjectType objtype,
+                                gboolean    trusted)
 {
+  /* Untrusted pulls require matching ownership */
+  if (!trusted && (src_repo->owner_uid != dest_repo->owner_uid))
+    return FALSE;
   /* Equal modes are always compatible, and metadata
    * is identical between all modes.
    */
@@ -4166,13 +4172,6 @@ import_one_object_direct (OstreeRepo    *dest_repo,
   GLNX_AUTO_PREFIX_ERROR (errprefix, error);
   char loose_path_buf[_OSTREE_LOOSE_PATH_MAX];
   _ostree_loose_path (loose_path_buf, checksum, objtype, dest_repo->mode);
-
-  if (!import_via_reflink_is_possible (src_repo, dest_repo, objtype))
-    {
-      /* If we can't reflink, nothing to do here */
-      *out_was_supported = FALSE;
-      return TRUE;
-    }
 
   /* hardlinks require the owner to match and to be on the same device */
   const gboolean can_hardlink =
@@ -4339,7 +4338,7 @@ _ostree_repo_import_object (OstreeRepo           *self,
    */
   const gboolean is_bareuseronly_conversion =
     import_is_bareuser_only_conversion (source, self, objtype);
-  gboolean try_direct = trusted;
+  gboolean try_direct = TRUE;
 
   /* If we need to do bareuseronly verification, or we're potentially doing a
    * bareuseronly conversion, let's verify those first so we don't complicate
@@ -4379,11 +4378,20 @@ _ostree_repo_import_object (OstreeRepo           *self,
         }
     }
 
-   /* We try to import via reflink/hardlink. If the remote is explicitly not trusted
-   * (i.e.) their checksums may be incorrect, we skip that.
-   */
-  if (try_direct)
+  /* First, let's see if we can import via reflink/hardlink. */
+  if (try_direct && import_via_reflink_is_possible (source, self, objtype, trusted))
     {
+      /* For local repositories, if the untrusted flag is set, we verify the
+       * checksum first. This assumes then that the files are immutable - the
+       * above check verified that the owner uids match.
+       */
+      if (!trusted)
+        {
+          if (!ostree_repo_fsck_object (source, objtype, checksum,
+                                        cancellable, error))
+            return FALSE;
+        }
+
       gboolean direct_was_supported = FALSE;
       if (!import_one_object_direct (self, source, checksum, objtype,
                                      &direct_was_supported,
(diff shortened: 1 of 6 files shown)


mwleeds:
I think it would be good to link to the flatpak PR in the commit message here so people can refer to that for a more in-depth explanation of why this is needed

src/libostree/ostree-repo-pull.c
```diff
@@ -3929,6 +3929,13 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       if ((flags & OSTREE_REPO_PULL_FLAGS_UNTRUSTED) == 0)
         pull_data->importflags |= _OSTREE_REPO_IMPORT_FLAGS_TRUSTED;
 
+      /* If the caller gaurantees that the files are immutable but still
```
wjt:
```suggestion
      /* If the caller guarantees that the files are immutable but still
```

src/libostree/ostree-repo.h
```diff
@@ -1218,6 +1218,7 @@ gboolean ostree_repo_prune_from_reachable (OstreeRepo             *self,
  * @OSTREE_REPO_PULL_FLAGS_UNTRUSTED: Do verify checksums of local (filesystem-accessible) repositories (defaults on for HTTP)
  * @OSTREE_REPO_PULL_FLAGS_BAREUSERONLY_FILES: Since 2017.7.  Reject writes of content objects with modes outside of 0775.
  * @OSTREE_REPO_PULL_FLAGS_TRUSTED_HTTP: Don't verify checksums of objects HTTP repositories (Since: 2017.12)
+ * @OSTREE_REPO_PULL_FLAGS_IMMUTABLE_FILES: Denotes that the files of local repository are immutable (Since: 2018.10)
```
matthiasclasen: The Since here since dubious, given that we're already at 2018.12

uajain: ah right. What was I thinking :/

mwleeds: No, 2018.10 is right. 2018.9 was the last release.

matthiasclasen: Oh, they aren't date-based? Ignore me

cgwalters:
Offhand, I think we could probably safely just change the semantics of local pulls so that we assume the source files are immutable - particularly when the repos are owned by the same uid (as is the case here right?)

cgwalters:
Something like this?  Compile tested only.

```
index f8f8d070..7ffb4a77 100644
@@ -4101,12 +4101,18 @@ import_is_bareuser_only_conversion (OstreeRepo *src_repo,
     && objtype == OSTREE_OBJECT_TYPE_FILE;
 }
 
-/* Returns TRUE if we can potentially just call link() to copy an object. */
+/* Returns TRUE if we can potentially just call link() to copy an object;
+ * if untrusted the repos must be owned by the same uid.
+ */
 static gboolean
 import_via_reflink_is_possible (OstreeRepo *src_repo,
                                 OstreeRepo *dest_repo,
-                                OstreeObjectType objtype)
+                                OstreeObjectType objtype,
+                                gboolean    trusted)
 {
+  /* Untrusted pulls require matching ownership */
+  if (!trusted && (src_repo->owner_uid != dest_repo->owner_uid))
+    return FALSE;
   /* Equal modes are always compatible, and metadata
    * is identical between all modes.
    */
@@ -4167,13 +4173,6 @@ import_one_object_direct (OstreeRepo    *dest_repo,
   char loose_path_buf[_OSTREE_LOOSE_PATH_MAX];
   _ostree_loose_path (loose_path_buf, checksum, objtype, dest_repo->mode);
 
-  if (!import_via_reflink_is_possible (src_repo, dest_repo, objtype))
-    {
-      /* If we can't reflink, nothing to do here */
-      *out_was_supported = FALSE;
-      return TRUE;
-    }
-
   /* hardlinks require the owner to match and to be on the same device */
   const gboolean can_hardlink =
     src_repo->owner_uid == dest_repo->owner_uid &&
@@ -4379,11 +4378,20 @@ _ostree_repo_import_object (OstreeRepo           *self,
         }
     }
 
-   /* We try to import via reflink/hardlink. If the remote is explicitly not trusted
-   * (i.e.) their checksums may be incorrect, we skip that.
-   */
-  if (try_direct)
+  /* First, let's see if we can import via reflink/hardlink. */
+  if (try_direct && import_via_reflink_is_possible (source, self, objtype, trusted))
     {
+      /* For local repositories, if the untrusted flag is set, we verify the
+       * checksum first. This assumes then that the files are immutable - the
+       * above check verified that the owner uids match.
+       */
+      if (!trusted)
+        {
+          if (!ostree_repo_fsck_object (source, objtype, checksum,
+                                        cancellable, error))
+            return FALSE;
+        }
+
       gboolean direct_was_supported = FALSE;
       if (!import_one_object_direct (self, source, checksum, objtype,
                                      &direct_was_supported,
```

alexlarsson:
@cgwalters I like that idea, however that doesn't seem quite right. In the non-trusted case try_direct is always false, so we'll never hit it.

alexlarsson:
If we change the initialization of try_direct to TRUE it seems ok though. Needs testing first though.

uajain:
`try_direct` just mirrors the `IMPORT_FLAG_TRUSTED`. I think there is a typo in the diff.
It should be `||` instead of `&&` :
```
+  /* First, let's see if we can import via reflink/hardlink. */
+  if (try_direct || import_via_reflink_is_possible (source, self, objtype, trusted))
```

alexlarsson:
@uajain no, that is not right. If try_direct was forced FALSE due to the symlink case above that in the code we should not try it. We just need to initialize it to TRUE.

mwleeds:
Looks like the commit message lost its reference to the flatpak PR

uajain:
@cgwalters I think it's good to go.
I did some testing with my existing flatpak PR and hardlink code-path was hit during local import.
Thanks for the quick patch.

uajain:
Thanks @cgwalters LGTM.

@rh-atomic-bot r+ 5956e24f29d7dbfab996f7677d6e828ff3f85619

cgwalters:
Ah, turns out I basically rewrote `test-pull-untrusted.sh`, which was asserting we didn't hardlink.  Hmm...

OK decided to just nuke that.  @uajain agree? :arrow_up:

uajain:
> Ah, turns out I basically rewrote `test-pull-untrusted.sh`, which was asserting we didn't hardlink. Hmm...
> 
> OK decided to just nuke that. @uajain agree? arrow_up

Perfect!

@rh-atomic-bot r+ https://github.com/ostreedev/ostree/pull/1776/commits/89b5189ded7f05cc32b85971aa570ddaea3e7ae4

PR #1777, opened 2018-12-03
lib/repo: Clarify docs for importing w/ trust

mwleeds:
The way _ostree_repo_import_object() is written, a hardlink copy is only
attempted if the source repo is trusted, so update the docs for
ostree_repo_import_object_from_with_trust() to reflect that.


cgwalters:
@rh-atomic-bot r+ 42b2d24

PR #1779, opened 2018-12-06
lib/repo-commit: Relax min-free-space check in prepare_transaction()

uajain:
We want a case where we can disable the min-free-space check. Initially,
it felt like to add a OSTREE_REPO_PULL_FLAGS_DISABLE_FREE_SPACE_CHECK but
the problem is prepare_transaction() does not have a OstreeRepoPullFlags
parameter which we can parse right here. On top of it, prepare_transaction()
enforces min-free-space check and won't let the transaction proceed if
the check failed.

This is pretty bad in conjunction with "inherit-transaction" as what
Flatpak uses. There is no way to disable this check unless we remove
it altogether from prepare_transaction.

This issue came out to light when flatpak wasn't able to write metadata
after fetching from remote:

[uajain@localhost ~]$ flatpak remote-info flathub org.kde.Platform//5.9
error: min-free-space-size 500MB would be exceeded

Metadata objects helps in housekeeping and restricting them means
restricting crucial UX (like search, new updates) functionalities
in clients like gnome-software. The error banners originated from
these issues are also abrupt and not much helpful to the user. This
is the specific instance of the issue this patches tries to address.

See https://github.com/flatpak/flatpak/issues/2139 for discussion.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1624,9 +1624,18 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
     self->txn.max_blocks = bfree - self->reserved_blocks;
   else
     {
-      self->cleanup_stagedir = TRUE;
-      g_mutex_unlock (&self->txn_lock);
-      return throw_min_free_space_error (self, 0, error);
+      self->txn.max_blocks = 0;
+      /* Don't throw_min_free_space_error here; reason being that
+       * this transaction could be just committing metadata objects
+       * which are relatively small in size and we do not really
+       * want to block them via min-free-space-* value. Metadata
+       * objects helps in housekeeping and hence should be kept
+       * out of the strict min-free-space values.
+       *
+       * The main drivers for writing content objects will always honor
+       * the min-free-space value and throw_min_free_space_error in
+       * case of overstepping the number of reserved blocks.
+       */
     }
   g_mutex_unlock (&self->txn_lock);
 
--- a/tests/installed/nondestructive/itest-pull-space.sh
+++ b/tests/installed/nondestructive/itest-pull-space.sh
@@ -45,6 +45,17 @@ echo 'min-free-space-size=1MB' >> repo/config
 ostree --repo=repo pull-local /ostree/repo ${host_commit}
 echo "ok min-free-space-size (success)"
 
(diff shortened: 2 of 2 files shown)


pwithnall:
The changes look OK to me, though I have not been following the min-free-space work recently. Does your new test fail as expected if the other chunk of the patch is temporarily reverted?

The tests don’t seem too happy, but the failure doesn’t seem directly related to your changes. I don’t really understand this ansible test setup though:
```
+ exec ansible-playbook --tags=atomic -v /srv/code/tests/installed/nondestructive.yml
Using /etc/ansible/ansible.cfg as config file
/usr/share/ansible/inventory/standard-inventory-docker did not meet host_list requirements, check plugin documentation if this is unexpected
/usr/share/ansible/inventory/standard-inventory-local did not meet host_list requirements, check plugin documentation if this is unexpected
/usr/share/ansible/inventory/standard-inventory-qcow2 did not meet host_list requirements, check plugin documentation if this is unexpected
/usr/share/ansible/inventory/standard-inventory-rpm did not meet host_list requirements, check plugin documentation if this is unexpected
ERROR! 'async_commands' is not a valid attribute for a TaskInclude

The error appears to have been in '/srv/code/tests/installed/nondestructive.yml': line 30, column 11, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

    - block:
        - name: Run nondestructive tests
          ^ here

This error can be suppressed as a warning using the "invalid_task_attribute_failed" configuration
```

src/libostree/ostree-repo-commit.c
```diff
@@ -1624,9 +1624,17 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
     self->txn.max_blocks = bfree - self->reserved_blocks;
   else
     {
```
mwleeds: We should set `self->txn.max_blocks` to **some** value in the else block right? Otherwise the code that uses it will be reading uninitialized memory, I think

jlebon:
> The tests don’t seem too happy, but the failure doesn’t seem directly related to your changes.

Nuked in #1783.

This patch looks sane to me, though can you confirm

> Does your new test fail as expected if the other chunk of the patch is temporarily reverted?

jlebon:
@rh-atomic-bot r=mwleeds 7579b63

PR #1782, opened 2018-12-07
bin/pull: Add URL metavar for --url switch

jlebon:
Otherwise it's not obvious that it takes an argument.


cgwalters:
@rh-atomic-bot r+ 1b9aba1

PR #1783, opened 2018-12-08
ci/papr: Drop insttests

cgwalters:
This didn't quite work out, and is now always failing because Ansible
changed.

For now we have some OK coverage via the rpm-ostree suite, let's just
drop this and revisit later.

### DIFF
--- a/.papr.yml
+++ b/.papr.yml
@@ -1,29 +1,3 @@
-# https://fedoraproject.org/wiki/CI/Tests
-branches:
-    - master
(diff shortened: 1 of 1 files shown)


jlebon:
@rh-atomic-bot r+ 41a0d70

Looks like the `f28-rpmostree` suite is also failing right now with:

```
+ fatal 'vmsync failed'
./ci/rpmostree.sh: line 48: fatal: command not found
```

PR #1784, opened 2018-12-10
Fix leak in ostree_content_stream_parse()

alexlarsson:
We need to ref_sink new GVariants for autoptr to work


cgwalters:
@rh-atomic-bot r+ b38afae

Issue #1770, opened 2018-10-31
`ostree admin status` can't check GPG signature unless key is imported first


miabbott:
Using Fedora 29 AH, I observed that `rpm-ostree` will report that a commit is signed with a valid key, but `ostree admin status` does not until the key is imported.

```
# rpm -q ostree rpm-ostree
ostree-2018.8-1.fc29.x86_64
rpm-ostree-2018.8-1.fc29.x86_64

# cat /etc/ostree/remotes.d/fedora-atomic.conf
[remote "fedora-atomic"]
url=https://dl.fedoraproject.org/atomic/repo/
gpg-verify=true
gpgkeypath=/etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-29-primary

# rpm-ostree status
State: idle
AutomaticUpdates: disabled
Deployments:
● ostree://fedora-atomic:fedora/29/x86_64/atomic-host
                   Version: 29.20181030.0 (2018-10-30 15:35:06)
                    Commit: a4cfae321d662d84a7b8c59b59678af5eea0d89c9077affa77331c09341a5278
              GPGSignature: Valid signature by 5A03B4DD8254ECA02FDA1637A20AA56B429476B4

# ostree admin status
* fedora-atomic a4cfae321d662d84a7b8c59b59678af5eea0d89c9077affa77331c09341a5278.0
    Version: 29.20181030.0
    origin refspec: fedora-atomic:fedora/29/x86_64/atomic-host
    GPG: Signature made Tue 30 Oct 2018 03:35:17 PM UTC using RSA key ID A20AA56B429476B4
    GPG: Can't check signature: public key not found

# ostree remote gpg-import fedora-atomic -k /etc/pki/rpm-gpg/RPM-GPG-KEY-fedora-29-primary                                                                                            
Imported 1 GPG key to remote "fedora-atomic"

# ostree admin status
* fedora-atomic a4cfae321d662d84a7b8c59b59678af5eea0d89c9077affa77331c09341a5278.0
    Version: 29.20181030.0
    origin refspec: fedora-atomic:fedora/29/x86_64/atomic-host
    GPG: Signature made Tue 30 Oct 2018 03:35:17 PM UTC using RSA key ID A20AA56B429476B4
    GPG: Good signature from "Fedora 29 <fedora-29@fedoraproject.org>"
```

dustymabe:
[...]

PR #1787, opened 2018-12-14
admin: status - correctly check GPG signature

r4f4:
Closes: #1770

### DIFF
--- a/src/ostree/ot-admin-builtin-status.c
+++ b/src/ostree/ot-admin-builtin-status.c
@@ -141,10 +141,11 @@ deployment_print_status (OstreeSysroot    *sysroot,
       g_autoptr(GString) output_buffer = g_string_sized_new (256);
       /* Print any digital signatures on this commit. */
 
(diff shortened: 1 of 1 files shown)


jlebon:
bot, add author to whitelist

jlebon:
Looks good to me! Just a random note: for the commit message, something more consistent like this would be nice: `admin/status: Correctly check GPG signature`. Would be good too to add a test for this (or just tweak an existing one)? Maybe in `admin-test.sh`?

cgwalters:
> Would be good too to add a test for this (or just tweak an existing one)? Maybe in admin-test.sh?

AFAICS we don't have any admin tests that use gpg, I didn't want to make adding one a prereq for a new contributor, so I said on IRC it'd be OK without.

jlebon:
Yup sure, that's fine. @r4f4 mind just tweaking the commit msg?

r4f4:
@jlebon Done. Thank you for reviewing.

dustymabe:
nice work @r4f4!

PR #1788, opened 2018-12-15
Leaks

matthiasclasen:
Some small leaks that were discovered while valgrinding flatpak.


cgwalters:
@rh-atomic-bot r+ 7493004

cgwalters:
@rh-atomic-bot retry

PR #1790, opened 2018-12-18
grub2: add support for devicetree

ricardosalveti:
Similar as available for u-boot (ce2995e1dc1557c4d97ef5af807eacf3ef4a22d8)
and syslinux (c5112c25e4519835c4cd53f4350c1b2f2a477746), enable parsing
and writing devicetree filename into grub.cfg.

This is required by arm64-based devices running edk2 instead of u-boot
as the main bootloader (e.g. 96boards HiKey and HiKey960).

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>

### DIFF
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -251,10 +251,6 @@ _ostree_bootloader_grub2_generate_config (OstreeSysroot                 *sysroot
       if (devicetree)
         {
           g_string_append (output, "devicetree");
-          if (is_efi)
-            g_string_append (output, GRUB2_EFI_SUFFIX);
-          else
-            g_string_append (output, GRUB2_SUFFIX);
           g_string_append_c (output, ' ');
           g_string_append (output, devicetree);
           g_string_append_c (output, '\n');
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -191,6 +191,7 @@ _ostree_bootloader_grub2_generate_config (OstreeSysroot                 *sysroot
       const char *options;
       const char *kernel;
       const char *initrd;
+      const char *devicetree;
       char *quoted_title = NULL;
       char *uuid = NULL;
       char *quoted_uuid = NULL;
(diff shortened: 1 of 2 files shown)


jlebon:
bot, test pull request

jlebon:
Thanks for the patch!
This looks sane to me, though will let @cgwalters have a look.

src/libostree/ostree-bootloader-grub2.c
```diff
@@ -246,6 +247,19 @@ _ostree_bootloader_grub2_generate_config (OstreeSysroot                 *sysroot
...
         }
 
+      devicetree = ostree_bootconfig_parser_get (config, "devicetree");
+      if (devicetree)
+        {
+          g_string_append (output, "devicetree");
+          if (is_efi)
+            g_string_append (output, GRUB2_EFI_SUFFIX);
+          else
+            g_string_append (output, GRUB2_SUFFIX);
```
cgwalters: I don't think this `if (is_efi)` part is necessary right?  That turns into e.g. `16` or `efi` sometimes for e.g. `linux16` or `linuxefi`, but is it really e.g. `devicetreeefi`?  (Are these devices EFI?)

https://www.gnu.org/software/grub/manual/grub/html_node/devicetree.html

Seems to imply it's always `devicetree`.

cgwalters:
Also can you clarify whether you tested this using the direct shell `ostree-grub-generator` backend or the full `grub2-mkconfig` path?  (It's totally fine to have only tested one path, would just be helpful to know)

src/libostree/ostree-bootloader-grub2.c
ricardosalveti: You are right, it should always be devicetree.

ricardosalveti:
> Also can you clarify whether you tested this using the direct shell ostree-grub-generator backend or the full grub2-mkconfig path? (It's totally fine to have only tested one path, would just be helpful to know)

Tested with ostree-grub-generator only.

src/libostree/ostree-bootloader-grub2.c
ricardosalveti: Let me push a fixup for this.

cgwalters:
Thanks for the patch!

@rh-atomic-bot r+ 7dfe90b

PR #1791, opened 2018-12-20
boot/grub2: Add some more explanation in ostree-grub-generator

jlebon:
Even with the previous docstring, I didn't understand at first the
relationship between the `ostree-grub-generator` script and
`ostree-bootloader-grub2.c`. Throw some more docs to clarify things a
bit.


cgwalters:
@rh-atomic-bot r+ 7ccff41

PR #1796, opened 2019-01-04
WIP: sysroot: Remove immutable bit from /etc/sysconfig/nfs

cgwalters:
Work around for https://src.fedoraproject.org/rpms/nfs-utils/pull-request/7
Currently Fedora-only.


cgwalters:
Don't merge this - I'm just filing it here as a reference.  Will probably close this and just attach it as a patch to the Fedora ostree package.

cgwalters:
Marking WIP so we don't merge, but I did test this patch.

jlebon:
Looks OK to me as a Fedora patch.

cgwalters:
https://koji.fedoraproject.org/koji/taskinfo?taskID=31958795

PR #1797, opened 2019-01-07
admin/status: test GPG signature check

r4f4:
This commit complements PR #1787 

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -109,6 +109,7 @@ _installed_or_uninstalled_test_scripts = \
 	tests/test-admin-pull-deploy-split.sh \
 	tests/test-admin-locking.sh \
 	tests/test-admin-deploy-clean.sh \
+	tests/test-admin-gpg.sh \
 	tests/test-reset-nonlinear.sh \
 	tests/test-oldstyle-partial.sh \
 	tests/test-delta.sh \
--- a/tests/test-admin-gpg.sh
+++ b/tests/test-admin-gpg.sh
@@ -1,0 +1,149 @@
+#!/bin/bash
+#
+# Copyright (C) 2019 Rafael Fonseca <r4f4rfs@gmail.com>
+#
+# SPDX-License-Identifier: LGPL-2.0+
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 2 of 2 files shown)


Makefile-tests.am
```diff
@@ -109,6 +109,7 @@ _installed_or_uninstalled_test_scripts = \
 	tests/test-admin-pull-deploy-split.sh \
 	tests/test-admin-locking.sh \
 	tests/test-admin-deploy-clean.sh \
+    tests/test-admin-gpg.sh \
```
rfairley: Minor: looks like a tab is needed here to indent this line

rfairley:
@rh-atomic-bot retry

rfairley:
bot, retest this please

tests/test-admin-gpg.sh
```diff
@@ -0,0 +1,149 @@
...
+
+# Exports OSTREE_SYSROOT so --sysroot not needed.
+setup_os_repository_signed "archive" "syslinux"
+
+echo "1..2"
+
+${CMD_PREFIX} ostree --repo=sysroot/ostree/repo remote add testos $(cat httpd-address)/ostree/testos-repo
+${CMD_PREFIX} ostree --repo=sysroot/ostree/repo pull-local --gpg-verify=true --remote=testos testos-repo testos/buildmaster/x86_64-runtime
+# This initial deployment gets kicked off with some kernel arguments
+${CMD_PREFIX} ostree admin deploy --karg=root=LABEL=MOO --karg=quiet --os=testos testos:testos/buildmaster/x86_64-runtime
```
cgwalters: To extend this test in the future we should add an upgrade and then test the `ostree admin upgrade` path so we're exercising the GPG pull via that code too (which is mostly the same thing in practice).

tests/test-admin-gpg.sh
```diff
@@ -0,0 +1,149 @@
...
+	      ostree_repo_init testos-repo --mode=${mode}
+    else
+	      ostree_repo_init testos-repo
+    fi
+
+    cd ${test_tmpdir}
+    mkdir osdata
+    cd osdata
+    kver=3.6.0
+    mkdir -p usr/bin ${bootdir} usr/lib/modules/${kver} usr/share usr/etc
```
cgwalters: Future enhancement: Let's avoid duplicating the code for setup (maybe make gpg signatures a flag)?

cgwalters:
@rh-atomic-bot r+ 54ae4d1

tests/test-admin-gpg.sh
rfairley: :+1: Agree - my thought was for this test it'd be okay as `setup_os_repository_signed` is local to this file, but to enhance in future, the one in `libtest.sh` could be updated with a flag.

dustymabe:
thanks @r4f4 !

PR #1798, opened 2019-01-08
pull: If built with --disable-http2, allow enabling via http2=1

cgwalters:
We have a `http2=[0|1]` remote config option; let's have the
`--disable-http2` build option define the default for that.  This way
it's easy to still enable http2 for testing even if
we have it disabled by default.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -141,7 +141,7 @@ AS_HELP_STRING([--disable-http2],
   [Disable use of http2 (default: no)]),,
   [enable_http2=yes])
 AS_IF([test x$enable_http2 != xno ], [
-  AC_DEFINE([BUILDOPT_HTTP2], 1, [Define if we enable http2])
+  AC_DEFINE([BUILDOPT_HTTP2], 1, [Define if we enable http2 by default])
 ], [
   OSTREE_FEATURES="$OSTREE_FEATURES no-http2"
 ])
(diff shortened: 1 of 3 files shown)


jlebon:
How about also:

```diff
diff --git a/configure.ac b/configure.ac
index 8aeec3ab..aad3d3c7 100644
--- a/configure.ac
+++ b/configure.ac
@@ -138,10 +138,10 @@ AM_CONDITIONAL(USE_CURL, test x$with_curl != xno)
 if test x$with_curl = xyes; then OSTREE_FEATURES="$OSTREE_FEATURES libcurl"; fi
 AC_ARG_ENABLE(http2,
 AS_HELP_STRING([--disable-http2],
-  [Disable use of http2 (default: no)]),,
+  [Disable http2 by default (default: no)]),,
   [enable_http2=yes])
 AS_IF([test x$enable_http2 != xno ], [
-  AC_DEFINE([BUILDOPT_HTTP2], 1, [Define if we enable http2])
+  AC_DEFINE([BUILDOPT_HTTP2], 1, [Define if we enable http2 by default])
 ], [
   OSTREE_FEATURES="$OSTREE_FEATURES no-http2"
 ])
```

cgwalters:
So apparently I accidentally pushed the fixup to master :cry: 

https://github.com/ostreedev/ostree/commit/fa35ec38b774b639e0490a12ca22a425b517a6c9

I pushed it here too though with the correct `fixup!` which should merge OK I think.

jlebon:
@rh-atomic-bot r+ a3e8e46

dustymabe:
> So apparently I accidentally pushed the fixup to master cry

:( 

I try to prevent myself from doing things like this by making my git config use `https` for the `upstreamorigin` remote and ssh for my fork (the `origin` remote)

if I `git push` and it asks me for a password, I know i'm pushing to the wrong remote

PR #1804, opened 2019-01-23
tests: Don't skip collection-based summary test

mwleeds:
OSTree's P2P support used to be a compile time option but is now always
enabled. One of the tests was still checking for the old feature flag,
"experimental", which has been renamed to "p2p" and skipping for that
reason. Delete the check so the test always runs.


mwleeds:
bot, retest this please

pwithnall:
@rh-atomic-bot r+ 9ef0f1740589e29f09ba8d4dff7a64e05a0f73de

PR #1805, opened 2019-01-23
Support multiple collection bindings

alexlarsson:
The current `ostree.collection-binding` key only supports a single collection-id. This adds support for a similar key ostree.collections-binding which is instead an array. A pull with a collection-id must match one of the listed ones.

The goal of this is to support subset repositories on e.g. flathub for example one with only free software. The idea is that you have one large repo with the collection id `org.flathub.Stable`, but when you commit the app you also bind the commit to `org.flathub.Stable.free` and create an extra `refs/mirrors/org.flathub.Stable.free/...` ref. We can then from this single repo easily generate split out repositories that only have one of the refs. Such repositories will share (i.e. hardlink) almost objects with the parent repo, except things like the ostree-metadata branch and the summary file. 

In order to test this, the branch also contains a few changes to the ostree command, and with those applied here is an example of how this would be used:

```
# Create main repo with two apps: app1, app2
ostree init --repo=main --mode=archive-z2 --collection-id=org.flathub.Stable
mkdir app
touch app/data
ostree commit --repo=main --branch=app1 --bind-collection-id=org.flathub.Stable.free app
ostree commit --repo=main --branch=app2 --bind-collection-id=org.flathub.Stable.free app

# Put app1 also in org.flathub.Stable.free collection
ostree reset --repo=main --create --collection-id=org.flathub.Stable.free app1 app1

# Pull all refs into "all" repo
ostree --repo=all init --mode=archive-z2 --collection-id=org.flathub.Stable
ostree pull-local --repo=all --collection-id=org.flathub.Stable main

# Pull free refs into free" repo
ostree --repo=free init --mode=archive-z2 --collection-id=org.flathub.Stable.free
ostree pull-local --repo=free --collection-id=org.flathub.Stable.free main
```

This doesn't currently have any tests. I can add them if this seems like a good idea. I would really like to get an ack at least on the metadata key format and name, because then I can start adding the required options to flatpak to use this.


alexlarsson:
Seems like CI failed with:
```
+ fatal 'vmsync failed'
./ci/rpmostree.sh: line 48: fatal: command not found
```

cgwalters:
(Love the idea that flathub clearly separates out proprietary software!)

> ostree init --repo=main --mode=archive-z2

`s/-z2//` nowadays

Anyways I glanced at the new key and code - looks sane to me.

pwithnall:
Sorry for the slow response. I’d be very interested in what @mwleeds has to say too, since I haven’t touched OSTree/P2P stuff for months now and have forgotten everything I thought I knew.

My thoughts:
 * Can you summarise the discussion from IRC the other day, for posterity? In particular, there was some discussion about an alternative plan to generate additional commit objects (one per collection ID) which reused the same dirtree. What were the reasons for not going with that? IIRC it was a pain for static deltas, since the delta superblock refers to commit checksums rather than dirtree checksums. Any other pros/cons?
 * One of the downsides of introducing a new metadata key like `COLLECTIONS_BINDING` is that it ties us in to handling both that *and* the old keys forever (since we can never guarantee that all repositories will update their metadata. In this case, it means all code has to handle:
 ** No binding keys present (original OSTree/flatpak repositories)
 ** Ref binding present but no collection ID binding present (current OSTree/flatpak repositories, but with no collection ID set on the repository)
 ** Ref binding and collection ID present (current OSTree/flatpak repositories, with a collection ID set on the repository) — this is the current ‘state of the art’
 ** Ref binding and collection IDs present (proposed)
 * The question of how this interacts with ref bindings (`OSTREE_COMMIT_META_KEY_REF_BINDING`) is a big one which I think means this needs changes; see below.
 * How do multiple collection IDs interact with GPG keys? See below too.
 * The documentation will need updating in various places to reference this.

The main issue I see is that this is ambiguous when coupled with `OSTREE_COMMIT_META_KEY_REF_BINDING`. If a commit specifies two collection IDs (C1, C2) and two refs (R1, R2) in its bindings, which collection–refs is it actually bound to? Which subset of [(C1, R1), (C1, R2), (C2, R1), (C2, R2)]?

There’s a potential there for confusion, which could potentially lead to security issues where a client could be fooled into thinking a given commit is bound to a collection–ref which the repository author didn’t intend for it to be bound to.

An alternative which avoids this would be to:
 * Add a new `OSTREE_COMMIT_META_KEY_COLLECTION_REFS_BINDING` key which binds the commit to a set of collection–refs. It would have type `a(ss)` and would be added instead of `OSTREE_COMMIT_META_KEY_COLLECTIONS_BINDING`.
 * We would deprecate `OSTREE_COMMIT_META_KEY_COLLECTION_BINDING` in favour of `OSTREE_COMMIT_META_KEY_COLLECTION_REFS_BINDING`.
 * `OSTREE_COMMIT_META_KEY_REF_BINDING` would also be deprecated. In the case where a repository doesn’t have a collection ID set, the set of refs from the set of collection–refs could be used (i.e. just ignoring the collection IDs).

---

Secondly, some thought would need to be given to how this interacts with the GPG keys from different collection IDs. In the use case you care about for flathub, I guess the same GPG key would be used for all the collection IDs? However, if this is to work in general we’d need to support the case where each collection ID has a different GPG key. Would it be OK to have multiple signatures on the commit, one per collection ID? I think commits support that already — it’s more a case of making sure that we verify the correct one based on the collection ID we used to look up the commit.

This is a topic which I have given less thought to, and would appreciate input from others (@cgwalters) on.

> This doesn't currently have any tests. I can add them if this seems like a good idea.

Yes, definitely. I would not want such a core piece of software distribution infrastructure to not have tests against regressions. The collection/ref bindings are a key part of the security of P2P OSTree distribution.

alexlarsson:
About alternate approaches:

The other alternative is to have separate commits for the various other collection ids, the advantage being that there is no need to modify ostree to handle multiple collection ids for a commit. 

However, it just seems weird to have multiple commits for what really *should* be exactly the same thing. That said, the main technical reason for this is about the deltas. 

Deltas are addressed by commit ids, so you need to duplicate all deltas with the commits. This is either very slow and space-intensive, or requires complicated rewriting to regenerate delta headers that can somehow reuse data from the main ones. In the shared commit version we can just symlink the delta subdirectories between all the repos and ignore this problem.

pwithnall:
Would it be advantageous to change deltas to be addressed by dirtree checksums instead of commit checksums, in general? It’s a slightly orthogonal change, but if it were to happen anyway (for example) then this feature would become simpler.

One reason we might want to change static deltas anyway is because various OSTree builders (like the one for EOS, and I believe the one for Atomic OS) ‘promote’ commits from one collection ID/repository to another (for example, once they’ve passed testing) by keeping the dirtree and generating a new commit object. That currently means regenerating the deltas (exactly the same problem as we have here). Would be better to avoid that.

alexlarsson:
I agree with your OSTREE_COMMIT_META_KEY_COLLECTION_REFS_BINDING proposal. I will switch to that.

For GPG signatures. You can sign each commit with however many keys you want. Only the one configured in the remote will be used when verifying it during the pull though, so I think this should be fine.

alexlarsson:
As for changing deltas, its somewhat complicated. At pull time you got the commit id from the summary, and the current commit id from the local repo. This is all you need for the current deltas. If deltas instead where dirtree based then you'd need to first download the new commit object separately to know what deltas you have. 

Furthermore, to find the deltas to use the current code just looks at the list of deltas with the right target commit and sees if we have the corresponding commit object. Except it also looks for .commitpartial state because then it knows it can't trust that commit to be complete. This later would fail if we were using dirtrees to address the deltas. This could be somewhat worked around by limiting which commits you're looking at based on the ref history and then looking at commitpartials for those before looking at their dirtrees, but its much trickier.

Its not completely unthinkable to do this change, but its also not trivial, and its not addressing other issues with deltas such as them bloating the summary file (we should really have the delta list in the commit object detached metadata instead).

pwithnall:
> For GPG signatures. You can sign each commit with however many keys you want. Only the one configured in the remote will be used when verifying it during the pull though, so I think this should be fine.

Cool. I think that’s something we’d want a unit test for, otherwise it could silently break.

> As for changing deltas, its somewhat complicated. …

Probably best to keep it as a separate concern then. Thanks for thinking about it.

alexlarsson:
I believe this is causing the f28-rpmostree ci fail:
```
Jan 25 12:42:04 vmcheck systemd[1]: Starting RPM-OSTree System Management Daemon...
Jan 25 12:42:04 vmcheck rpm-ostree[1671]: /usr/bin/rpm-ostree: error while loading shared libraries: libsolv.so.1: cannot open shared object file: No such file or directory
```

src/ostree/ot-builtin-commit.c
```diff
@@ -447,6 +551,31 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
       goto out;
     }
 
+  if (opt_bind_collection_refs)
+    {
+      int i, out_i;
```
pwithnall: Change to `gsize` and move into the `for` loop statement

src/ostree/ot-builtin-commit.c
```diff
@@ -331,26 +363,92 @@ parse_keyvalue_strings (GVariantBuilder   *builder,
...
 
-static int
-compare_strings (gconstpointer a, gconstpointer b)
-{
-  const char **sa = (const char **)a;
-  const char **sb = (const char **)b;
+          if (!split_collection_ref_pair (opt_bind_collection_refs[i],
+                                          &collection_binding, &ref_binding, NULL))
+            {
+              /* This shouldn't happen because we verified the args early */
```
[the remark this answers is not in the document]
pwithnall: Though where are the args actually verified?

src/ostree/ot-builtin-commit.c
```diff
@@ -447,6 +551,31 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
...
+      opt_bind_refs = g_new0 (char *, g_strv_length (opt_bind_collection_refs) + 1);
+      for (i = 0, out_i = 0; opt_bind_collection_refs[i] != NULL; i++)
+        {
+          g_autofree char *ref_binding = NULL;
+          if (!split_collection_ref_pair (opt_bind_collection_refs[i],
+                                          NULL, &ref_binding, error))
+            goto out;
+
+          if (!g_strv_contains ((const char * const *)opt_bind_refs, ref_binding))
+            opt_bind_refs[out_i++] = g_steal_pointer (&ref_binding);
```
pwithnall: Rather than calilng `g_strv_contains()` here (and getting O(N^2) behaviour, although there are probably never enough bindings for it to matter), perhaps it would be better to sort `opt_binding_refs` once it’s built, and then iterate and pairwise compare for duplicates? That gives the added advantage that the list of bindings is always sorted before being inserted into the commit metadata, which makes later lookups against it easier and stops the argument order to `ostree commit` from affecting the binding order (in a kind of ‘Reproducible Builds’) way.

Maybe best to do this with all the sets of bindings.

src/ostree/ot-builtin-reset.c
```diff
@@ -33,7 +33,12 @@
  * man page (man/ostree-reset.xml) when changing the option list.
  */
 
+static gboolean opt_create;
+static char *opt_collection_id;
+
 static GOptionEntry options[] = {
+  { "create", 'c', 0, G_OPTION_ARG_NONE, &opt_create, "Create the ref if it doesn't exist already", NULL },
```
pwithnall:
```suggestion
  { "create", 'c', 0, G_OPTION_ARG_NONE, &opt_create, "Create the ref if it doesn’t exist already", NULL },
```

I’ll get software using curly quotes and apostrophes if it kills me.

src/ostree/ot-builtin-reset.c
```diff
@@ -33,7 +33,12 @@
...
+  { "collection-id", 0, 0, G_OPTION_ARG_STRING, &opt_collection_id, "Use the collection id for the ref", "COLLECTION-ID" },
```
pwithnall:
```suggestion
  { "collection-id", 0, 0, G_OPTION_ARG_STRING, &opt_collection_id, "Use the collection ID for the ref", "COLLECTION-ID" },
```

src/ostree/ot-builtin-reset.c
```diff
@@ -84,7 +102,13 @@ ostree_builtin_reset (int           argc,
   if (!ostree_repo_prepare_transaction (repo, NULL, cancellable, error))
     goto out;
 
-  ostree_repo_transaction_set_ref (repo, NULL, ref, checksum);
+  if (opt_collection_id)
+    {
+      const OstreeCollectionRef col_ref = { opt_collection_id, (char *)refspec };
```
pwithnall: Nitpick: s/`col_ref`/`collection_ref`/ to keep terminology consistent

src/ostree/ot-builtin-pull-local.c
```diff
@@ -56,6 +57,7 @@ static GOptionEntry options[] = {
   { "gpg-verify", 0, 0, G_OPTION_ARG_NONE, &opt_gpg_verify, "GPG verify commits (must specify --remote)", NULL },
   { "gpg-verify-summary", 0, 0, G_OPTION_ARG_NONE, &opt_gpg_verify_summary, "GPG verify summary (must specify --remote)", NULL },
   { "depth", 0, 0, G_OPTION_ARG_INT, &opt_depth, "Traverse DEPTH parents (-1=infinite) (default: 0)", "DEPTH" },
+  { "collection-id", 0, 0, G_OPTION_ARG_STRING, &opt_collection_id, "Pull from this collection id", "COLLECTION-ID" },
```
pwithnall:
```suggestion
  { "collection-id", 0, 0, G_OPTION_ARG_STRING, &opt_collection_id, "Pull from this collection ID", "COLLECTION-ID" },
```

tests/test-pull-collections.sh
```diff
@@ -74,21 +82,66 @@ ensure_no_collection_ref() {
...
+                echo -n "$SEP"
+                SEP=", "
+                echo -n "'$1'"
+                shift
+                ;;
+        esac
+    done
+}
+
+# Join all coll-ref args after first -- with ", "
```
pwithnall:
```suggestion
# Join all collection–ref args after first -- with ", "
```

src/ostree/ot-builtin-reset.c
```diff
@@ -78,21 +78,29 @@ ostree_builtin_reset (int           argc,
     {
       if (opt_collection_id)
         {
+          const OstreeCollectionRef col_ref = { opt_collection_id, (char *)refspec };
```
pwithnall: The `fixup reset` commit needs to be squashed (just adding this comment so that isn’t forgotten).

src/libostree/ostree-core.h
```diff
@@ -271,10 +271,30 @@ typedef enum {
...
  * as it more strongly binds the commit to the repository and branch.
  *
+ * If you want to bind to multiple collections-ids, use OSTREE_COMMIT_META_KEY_COLLECTION_REFS_BINDING instead.
+ *
  * Since: 2018.6
  */
 #define OSTREE_COMMIT_META_KEY_COLLECTION_BINDING "ostree.collection-binding"
 
+/*
+ * OSTREE_COMMIT_META_KEY_COLLECTION_REFS_BINDING:
```
pwithnall: Repeating this comment here (from elsewhere in my review) so it doesn’t get lost: it would be good to guarantee that the elements in this binding array are ordered alphabetically, to make searching/sorting/reproducibility of commits easier. Probably best to sort by collection ID and then by ref name? (i.e. The same as formatting all the tuples as `$collection_id:$ref_name` and sorting by that.)

pwithnall:
I also left a post-hoc review on the flatpak changes: https://github.com/flatpak/flatpak/pull/2634#pullrequestreview-196990234

pwithnall:
The CI failures all look unrelated to me. The `ex-required` ones are all mirror error 503s from Fedora infrastructure, and the `f28-rpmostree` error seems to be a problem with VMware `vmsync` stuff rather than collection IDs.

src/libostree/ostree-repo-pull.c
```diff
@@ -1530,25 +1530,77 @@ _ostree_repo_verify_bindings (const char  *collection_id,
...
+  g_autoptr(GVariant) collection_refs_binding = NULL;
   g_autofree const char **refs = NULL;
+  gboolean need_ref_binding = FALSE;
+
+  collection_refs_binding = g_variant_lookup_value (metadata, OSTREE_COMMIT_META_KEY_COLLECTION_REFS_BINDING,
+                                                    G_VARIANT_TYPE("a(ss)"));
+
+  if (collection_refs_binding)
+    {
+      if (collection_id != NULL || ref_name != NULL)
```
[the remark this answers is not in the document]
alexlarsson: I don't like that. In that case it might take the else case even though collection_refs_binding is non-NULL. In practice, since that will only happen if collection_id == NULL it will not affect the behaviour, but it makes things less obvious.

src/ostree/ot-builtin-commit.c
```diff
@@ -331,26 +363,92 @@ parse_keyvalue_strings (GVariantBuilder   *builder,
...
+                 gconstpointer ref2p)
+{
+  const OstreeCollectionRef *ref1 = *(const OstreeCollectionRef **)ref1p;
+  const OstreeCollectionRef *ref2 = *(const OstreeCollectionRef **)ref2p;
+  int res;
+
+  res = strcmp (ref1->collection_id, ref2->collection_id);
+  if (res != 0)
+    return res;
+  return strcmp (ref1->ref_name, ref2->ref_name);
```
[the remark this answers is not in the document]
alexlarsson: No, this is meant to be a sorting compare, and ref_equal returns a bool.

src/ostree/ot-builtin-commit.c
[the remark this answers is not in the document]
alexlarsson: Its verified pretty early in ostree_builtin_commit() when we generate the back compat opt_bind_refs().

src/ostree/ot-builtin-commit.c
alexlarsson: That breaks the current "algorithm" of using the first specified ref command line argument as the one to pick for the backwards compat key though.

src/libostree/ostree-core.h
alexlarsson: Guarantee how? This is already done by the changes in commit. Do you mean adding this to the docs?

src/ostree/ot-builtin-commit.c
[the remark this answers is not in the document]
pwithnall: You’re right; I evidently need more coffee.

src/libostree/ostree-repo-pull.c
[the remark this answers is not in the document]
pwithnall: :+1:

src/ostree/ot-builtin-commit.c
pwithnall: Maybe there could be a separate block of code which picks the backwards compat key (and does it as you say)? It would be neat to get reproducible builds and to avoid O(N^2) behaviour.

src/libostree/ostree-core.h
pwithnall: > Do you mean adding this to the docs?

Yup. So it’s an API guarantee which can be relied on by anyone reading this key, and has to be stuck to by anyone writing it.

src/ostree/ot-builtin-commit.c
```diff
@@ -96,6 +97,7 @@ static GOptionEntry options[] = {
   { "orphan", 0, 0, G_OPTION_ARG_NONE, &opt_orphan, "Create a commit without writing a ref", NULL },
   { "no-bindings", 0, 0, G_OPTION_ARG_NONE, &opt_no_bindings, "Do not write any ref bindings", NULL },
   { "bind-ref", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_bind_refs, "Add a ref to ref binding commit metadata", "BRANCH" },
+  { "bind-collection-ref", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_bind_collection_refs, "Add a collection–ref binding to commit metadata", "COLLECTION-ID:BRANCH" },
```
pwithnall: The new command line options also need to be added to `bash/ostree` (sorry, only just remembered about this).

alexlarsson:
I force-pushed a rebase that includes the "fixup reset" commit. Will continue looking at the remaining issues soon.

pwithnall:
@alexlarsson, what’s the status of this? Looks like it almost got in, then the last few review comments didn’t get addressed (or I didn’t notice that they did get addressed). Is this still something you want?

alexlarsson:
Current status is that flatpak already can set these keys, but there is no real use of it in the wild. However, we still want to eventually use it for the free-flathub subset, although that constantly gets on the backburner. It would make a lot of sense for ostree to support this independent of that though, as it seems a natural extension.

PR #1810, opened 2019-02-06
lib/repo-pull: Allow the keyring remote to be overridden

mwleeds:
Currently the P2P code requires you to trust every remote you have
configured to the same extent, because a remote controlled by a
malicious actor can serve updates to refs (such as Flatpak apps)
installed from other remotes.[1] The way this attack would play out is
that the malicious remote would deploy the same collection ID as the
victim remote, and would then be able to serve updates for it.

One possible remedy would be to make it an error to configure remotes
such that two have the same collection ID but differing GPG keys. I
attempted to do that in Flatpak[2] but it proved difficult because it is
valid to configure two remotes with the same collection ID, and they may
then each want to update their keyrings which wouldn't happen
atomically.

Another potential solution I've considered is to add a `trusted-remotes`
option to ostree_repo_find_remotes_async() which would dictate which
keyring to use when pulling each ref. However the
ostree_repo_finder_resolve_async() API would still remain vulnerable,
and changing that would require rewriting a large chunk of libostree's
P2P support.

So this commit represents a third attempt at mitigating this security
hole, namely to have the client specify which remote to use for GPG
verification at pull time. This way the pull will fail if the commits
are signed with anything other than the keys we actually trust to serve
updates.

This is implemented as an option "override-keyring-remotes" for
ostree_repo_pull_from_remotes_async() which dictates the remote to be
used for GPG verification of each OstreeRepoFinderResult. If set, this
value will become the value of the "override-remote-name" option of
ostree_repo_pull_with_options(). Currently "override-remote-name"
affects only local pulls and in that case affects both the name to be
used in the refspec and the remote used for GPG verification. This
commit changes the meaning so that it dictates what remote to use for
GPG verification regardless of whether the pull is local or not. This
seems cleaner than adding a new option and shouldn't be a backwards
compatibility concern because it only makes sense to use a remote in a
refspec name if it can also be used for GPG verification.

This solution is still not perfect because (a) commit metadata will have
already been pulled from the malicious remote during the execution of
ostree_repo_find_remotes_async() which would not be removed on pull
failure and (b) if a global keyring directory is being used rather than
a separate keyring for each remote the pull will still succeed even if
it's not from the intended remote.

[1] https://github.com/flatpak/flatpak/issues/1447
[2] https://github.com/flatpak/flatpak/pull/2601

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1483,7 +1483,7 @@ gpg_verify_unwritten_commit (OtPullData                 *pull_data,
       if (ref != NULL)
         keyring_remote = g_hash_table_lookup (pull_data->ref_keyring_map, ref);
       if (keyring_remote == NULL)
-	keyring_remote = pull_data->remote_name;
+        keyring_remote = pull_data->remote_name;
 
       g_autoptr(GBytes) signed_data = g_variant_get_data_as_bytes (commit);
       g_autoptr(OstreeGpgVerifyResult) result =
@@ -1804,7 +1804,7 @@ scan_commit_object (OtPullData                 *pull_data,
       if (ref != NULL)
         keyring_remote = g_hash_table_lookup (pull_data->ref_keyring_map, ref);
       if (keyring_remote == NULL)
-	keyring_remote = pull_data->remote_name;
+        keyring_remote = pull_data->remote_name;
 
       result = ostree_repo_verify_commit_for_remote (pull_data->repo,
                                                      checksum,
@@ -4412,7 +4412,11 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
         {
           g_autoptr(OstreeCollectionRef) c_r = NULL;
 
+          if (!ostree_validate_collection_id (collection_id, error))
+            goto out;
           if (!ostree_validate_rev (ref_name, error))
+            goto out;
+          if (!ostree_validate_remote_name (keyring_remote_name, error))
             goto out;
 
           c_r = ostree_collection_ref_new (collection_id, ref_name);
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -115,6 +115,7 @@ typedef struct {
   GHashTable       *summary_deltas_checksums;
   GHashTable       *ref_original_commits; /* Maps checksum to commit, used by timestamp checks */
   GHashTable       *gpg_verified_commits; /* Set<checksum> of commits that have been verified */
+  GHashTable       *ref_keyring_map; /* Maps OstreeCollectionRef to keyring remote name */
   GPtrArray        *static_delta_superblocks;
   GHashTable       *expected_commit_sizes; /* Maps commit checksum to known size */
   GHashTable       *commit_to_depth; /* Maps commit checksum maximum depth */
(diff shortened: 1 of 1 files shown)


mwleeds:
Don't merge this until the Flatpak side of it has been implemented

src/libostree/ostree-repo-pull.c
```diff
@@ -1462,23 +1464,32 @@ process_verify_result (OtPullData            *pull_data,
...
+      const char *keyring_remote = NULL;
+
       /* Shouldn't happen, but see comment in process_verify_result() */
       if (g_hash_table_contains (pull_data->gpg_verified_commits, checksum))
         return TRUE;
 
+      if (ref != NULL)
+        keyring_remote = g_hash_table_lookup (pull_data->ref_keyring_map, ref);
+      if (keyring_remote == NULL)
+	keyring_remote = pull_data->remote_name;
```
pwithnall: Indentation looks wrong here.

src/libostree/ostree-repo-pull.c
```diff
@@ -1788,10 +1799,16 @@ scan_commit_object (OtPullData                 *pull_data,
       !g_hash_table_contains (pull_data->gpg_verified_commits, checksum))
     {
       g_autoptr(OstreeGpgVerifyResult) result = NULL;
+      const char *keyring_remote = NULL;
+
+      if (ref != NULL)
+        keyring_remote = g_hash_table_lookup (pull_data->ref_keyring_map, ref);
+      if (keyring_remote == NULL)
+	keyring_remote = pull_data->remote_name;
```
pwithnall: Indentation looks wrong here.

src/libostree/ostree-repo-pull.c
```diff
@@ -4373,6 +4404,26 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
 
+  if (opt_ref_keyring_map_set)
+    {
+      const gchar *collection_id, *ref_name, *keyring_remote_name;
+
+      while (g_variant_iter_loop (ref_keyring_map_iter, "(&s&s&s)", &collection_id, &ref_name, &keyring_remote_name))
+        {
+          g_autoptr(OstreeCollectionRef) c_r = NULL;
+
+          if (!ostree_validate_rev (ref_name, error))
```
pwithnall: Also need to validate the `collection_id` (`ostree_validate_collection_id()`) and `keyring_remote_name` (`ostree_validate_remote_name()`).

pwithnall:
> Don't merge this until the Flatpak side of it has been implemented

Is that [this flatpak PR](https://github.com/flatpak/flatpak/pull/2705)? It’s currently marked as WIP — so presumably this OSTree PR shouldn’t be merged yet? Maybe mark this one as WIP too if that’s the case.

mwleeds:
> LGTM, but you should get a second opinion. Maybe from Alex?

cc @alexlarsson

cgwalters:
> One possible remedy would be to make it an error to configure remotes
such that two have the same collection ID but differing GPG keys.

Right...I think I recall discussion of this in the original P2P code.  I guess a lot of the complexity comes down to supporting key rotation.  If the key can be rotated then it can't be used as a unique ID.

(Perhaps what we should have done is defined "collection ID" = "original GPG key fingerprint" and required any later keys to be signed by the first one or something)

I didn't look in detail at your flatpak PR but the idea is we're basically overriding with a "trusted remote"?  And "trusted remotes" are ones defined statically and not discovered via P2P?

mwleeds:
> I didn't look in detail at your flatpak PR but the idea is we're basically overriding with a "trusted remote"? And "trusted remotes" are ones defined statically and not discovered via P2P?

Yes, the remotes passed to libostree by Flatpak in `ref-keyring-map` are configured remotes which the user has indicated they trust. For a new install the remote is confirmed explicitly and for an update it's the remote the app was originally installed from. Even apps installed from a LAN or USB source use a configured remote as their origin.

cgwalters:
@rh-atomic-bot r+ eee6625

cgwalters:
@rh-atomic-bot r+ f4cd5ac

PR #1812, opened 2019-02-06
lib/repo-pull: Fix docs for n-network-retries option

mwleeds:
Add "Since: 2018.6" and copy the docs to
ostree_repo_pull_from_remotes_async() which also supports it.


jlebon:
Thanks!
@rh-atomic-bot r+ 392a6ce

PR #1813, opened 2019-02-08
lib/repo: Fix Since: annotation for REMOTE_CHANGE_REPLACE

mwleeds:


mwleeds:
I wonder if we should have a magic phrase to avoid wasting CI resources for things like this, like we do for updating submodules

jlebon:
Thanks!
@rh-atomic-bot r+ e59d5ae

> I wonder if we should have a magic phrase to avoid wasting CI resources for things like this

Yeah, we should. The RFE for that is https://github.com/projectatomic/papr/issues/27, though it wouldn't be hard to just embed it in the scripts here directly.

Issue #1774, opened 2018-11-16
ostree admin deploy fails if there are no existing deployments when a bootloader is detected


ajeddeloh:
**Version**
2018.10

**Expected behavior:**
`ostree admin deploy` sets up a deployment and bootloader entries even if it is the first deployment

**Actual behavior:**
`ostree admin deploy` fails if there are no deployments and the grub2 bootloader is detected

**Steps to reproduce:**
``` sh
mkdir sysroot
ostree admin init-fs sysroot
ostree pull-local <ostree repo> --repo sysroot/ostree/repo
ostree admin os-init osname --sysroot sysroot

# assuming only 1 ref
ref=$(ostree refs --repo sysroot/ostree/repo)

# trick ostree into thinking there is a bootloader installed
touch sysroot/boot/grub2/grub.cfg

# fails
ostree admin deploy "$ref" --sysroot sysroot --os osname
```

gives:
```
OSTree:ERROR:src/libostree/ostree-bootloader-grub2.c:356:_ostree_bootloader_grub2_write_config: assertion failed (deployments->len > 0): (0 > 0)
```

It appears it gets stuck in a catch-22 situation writing the bootloader entry where it uses an existing deployment to run grub-mkconfig rather than the pending deployment. Perhaps it should use the pending one in the case there are no completed deployments?

cgwalters:
OK, I reproduced this locally.  One trick one can do here btw on an existing ostree-based system is (assuming that `/var` is on the same filesystem), do:
`cd /ostree/deploy/$stateroot/deploy/var`
and then for the `pull-local`:

```
mkdir sysroot
ostree admin init-fs sysroot
cd sysroot
ostree pull-local --repo=sysroot/repo /sysroot/ostree/repo
```

That way it's just hardlinks and not copies and makes the process nearly free.

I had to do a pull by checksum and not ref though since the deployed system has a remote (`pull-local` should handle this though).

Anyways so what worked for me is to:

```
$ ostree admin deploy 89bfa708d349a5856cc5cd3be441c07e1f96d0be2aa97e2b676f6004e7f6abed --sysroot $(pwd) --os fcos --karg-proc-cmdline
$ touch boot/grub2/grub.cfg
$ ostree admin instutil set-kargs  --sysroot $(pwd) --append=foo=bar
```

Which is what Anaconda is doing today. 

But...yes I think this code is buggy in that it should be looking at the *pending* deployments which are provided for the root to use for `grub2-mkconfig`.

[...]

Issue #1801, opened 2019-01-11
Support for the Raspberry Pi bootloader


fkrull:
I'd like OSTree (i.e. `ostree admin deploy`) to integrate with the Raspberry Pi bootloader like it does with e.g. GRUB. My general idea is that after deploying, ostree would set the kernel and initramfs path as well as kernel options in [/boot/config.txt](https://elinux.org/RPiconfig#Boot). I haven't actually tried any of this since I don't have a bootable image yet, but I feel it should work.

---

The reason I'm opening this issue without actually having figured out if my idea works is because I think it might be better to implement some sort of extension point for bootloader integrations instead. That way, libostree wouldn't need to include support for marginal bootloaders and I wouldn't have to write the config file mangling in C ;)

Some ideas:
* Every binary in a certain directory, let's say `/usr/lib/ostree/bootloaders` participates in the selection process in `_ostree_sysroot_query_bootloader`, by mapping the OstreeBootloaderInterface functions to certain command-line arguments (e.g. `<binary> query` is expected to return a zero exit code if that bootloader is active).
* The path to a bootloader helper is configured in some file, e.g. the repo config as suggested at https://github.com/ostreedev/ostree/issues/1719#issuecomment-441738684. This setting could either override the normal bootloader selection or just add to it.
* Like above, but it's just a fixed path rather than a setting, e.g. if `/usr/lib/ostree/bootloader-helper` exists and is executable. This just seems like a strictly more confusing variant of the previous one though.

---

So yeah. I wanted to at least explain my usecase, but I feel a "plugin" approach would be quite useful here, not least because it saves me from having to work out my details in the context of libostree. Thoughts?

AdrianVovk:
Maybe a hybrid approach could work? There would be many different forks coming from the bootloader helpers. They would need to be queried, paths need to be figured out, etc, and all of that requires forks to the process. What about the bootloaders are defined in C (at least the common ones like GRUB, systemd-boot, uboot, etc) and then the sysadmin or distro packager can add extra bootloader config scripts in `/{etc,usr/lib}/ostree/bootloaders/foo`. There would be a special `stub` bootloader implmented in C that execs `/{etc,usr/lib}/ostree/bootloaders/default` which is a symlink to the correct bootloader.

That way, the most common configurations will remain quick, but the custom stuff will remain extensible

kirbyfan64:
[...]

Issue #1719, opened 2018-09-12
systemd-boot support


kirbyfan64:
From what I can tell, this is basically present (since the loader syntax is identical), except systemd-boot needs the loaders and kernels to be on the ESP (/boot/efi), not where OSTree places them currently (/boot)...

cgwalters:
That sounds likely yes.  I have no opposition to supporting `sd-boot` - the main question is how we detect it (actually the fact that we dynamically detect bootloaders today is a bit of a mess).

kirbyfan64:
Maybe just check for its EFI file? GRUB is a little easily to have automatically installed accidentally forget about it, but systemd-boot is usually manually (and therefore consciously) installed.

AdrianVovk:
I've run into this issue also. I'm trying to develop a distro around OSTree and I'd much rather use systemd-boot than grub for efi

AdrianVovk:
@cgwalters Can the bootloader detection issue be solved with a simple config file? OSTree, when running `ostree admin deploy` can simply look for a file in /usr/etc or whatever and see which bootloader it should use.

Another possibility is to have the installer tell OSTree which bootloader to use. I'd prefer this option, since this gives the installer script the flexibility to decide on the bootloader, not the tree. This way, boot installation/config all happens in one place. Something like: `ostree admin init-boot --sysroot=/mnt --os=foo systemd-boot`

If the config file doesn't exist/command wasn't run, OSTree can (temporarily) fall back to the current behavior

cgwalters:
For people who are building libostree from source (as it sounds like you are?), one thing we could pretty easily do is add a build-time option, like `--with-bootloader=sd-boot` or so.  Would that be sufficient?

My main hesistation on this is where to put the config file, but thinking about it as I'm typing, we could add it as an entry in the repo config?  Something like

```
[ostree-sysroot]
bootloader=<grub|sd-boot>
```

AdrianVovk:
@cgwalters Yes, I am building libostree from source. An option like this would be sufficient. The problem with it is that it will restrict the entire system to a single bootloader. If the user prefers Grub and replaces sd-boot, suddenly OSTree deployments break. You'd need to recompile OSTree to boot with Grub. For minimal distros, though, it would be nice to trim down ostree by restricting bootloader support

[...]

PR #1814, opened 2019-02-08
Add sysroot.bootloader repo config key (fix running bootloader when no deployments exist)

rfairley:
The sysroot.bootloader key is used to configure the bootloader
that OSTree uses when deploying a sysroot. Having this key
allows specifying behavior not to use the default bootloader
backend code, which is preferable when creating a first
deployment from the sysroot (#1774).

As of now, the key can take the values "auto" or "none". If
the key is not given, the value defaults to "auto".

"auto" causes _ostree_sysroot_query_bootloader() to be used
when writing a new deployment, which is the original behavior
that dynamically detects which bootloader to use.

"none" avoids querying the bootloader dynamically. The BLS
config fragments are still written to
sysroot/boot/loader/entries for use by higher-level software.

More values can be supported in future to specify a single
bootloader, different behavior for the bootloader code, or
a list of bootloaders to try.

Resolves: #1774

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -109,22 +109,22 @@ Boston, MA 02111-1307, USA.
         ensure files are on stable storage when performing operations
         such as commits, pulls, and checkouts.  Defaults to
         <literal>true</literal>.</para>
-	<para>
-	  If you disable fsync, OSTree will no longer be robust
-	  against kernel crashes or power loss.
-	</para>
-	<para>
-	  You might choose to disable this for local development
-	  repositories, under the assumption they can be recreated from
-	  source.  Similarly, you could disable for a mirror where you could
-	  re-pull.
-	</para>
-	<para>
-	  For the system repository, you might choose to disable fsync
-	  if you have uninterruptable power supplies and a well tested
-	  kernel.
-	</para>
-	</listitem>
+        <para>
+          If you disable fsync, OSTree will no longer be robust
+          against kernel crashes or power loss.
+        </para>
+        <para>
+          You might choose to disable this for local development
+          repositories, under the assumption they can be recreated from
+          source.  Similarly, you could disable for a mirror where you could
+          re-pull.
+        </para>
+        <para>
+          For the system repository, you might choose to disable fsync
+          if you have uninterruptable power supplies and a well tested
+          kernel.
+        </para>
+        </listitem>
       </varlistentry>
 
       <varlistentry>
@@ -327,6 +327,42 @@ Boston, MA 02111-1307, USA.
       <varlistentry>
         <term><varname>unconfigured-state</varname></term>
         <listitem><para>If set, pulls from this remote will fail with the configured text.  This is intended for OS vendors which have a subscription process to access content.</para></listitem>
+      </varlistentry>
+
+    </variablelist>
+
+  </refsect1>
+
+  <refsect1>
+    <title>[sysroot] Section Options</title>
+
+    <para>
+      Options for the sysroot, which contains the OSTree repository,
+      deployments, and stateroots.  The following entries are defined:
+    </para>
+
+    <variablelist>
+
+      <varlistentry>
+        <term><varname>bootloader</varname></term>
+        <listitem><para>Configure the bootloader that OSTree uses when
+        deploying the sysroot.  This may take the values
+        <literal>bootloader=none</literal> or <literal>bootloader=auto</literal>.
+        Default is <literal>auto</literal>.
+        </para>
+        <para>
+          If <literal>none</literal>, then OSTree will generate only BLS (Boot
+          Loader Specification) fragments in <literal>sysroot/boot/loader/entries/</literal>
+          for the deployment.
+        </para>
+        <para>
+          If <literal>auto</literal>, then in addition to generating BLS
+          fragments, OSTree will dynamically check for the existence of grub2,
+          uboot, and syslinux bootloaders.  If one of the bootloaders is found,
+          then OSTree will generate a config for the bootloader found.  For
+          example, <literal>grub2-mkconfig</literal> is run for the grub2 case.
+        </para>
+        </listitem>
       </varlistentry>
 
     </variablelist>
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3113,6 +3113,32 @@ reload_remote_config (OstreeRepo          *self,
   return TRUE;
 }
 
+static gboolean
+reload_sysroot_config (OstreeRepo          *self,
+                       GCancellable        *cancellable,
+                       GError             **error)
+{
+  { g_autofree char *bootloader = NULL;
+
+    if (!ot_keyfile_get_value_with_default_group_optional (self->config, "sysroot",
+                                                           "bootloader", "auto",
+                                                           &bootloader, error))
+      return FALSE;
+
+    /* TODO: possibly later add support for specifying a generic bootloader
+     * binary "x" in /usr/lib/ostree/bootloaders/x). See:
+     * https://github.com/ostreedev/ostree/issues/1719
+     * https://github.com/ostreedev/ostree/issues/1801
+     */
+    if (!(g_str_equal (bootloader, "auto") || g_str_equal (bootloader, "none")))
+      return glnx_throw (error, "Invalid bootloader configuration: '%s'", bootloader);
+
+    self->bootloader = g_steal_pointer (&bootloader);
+  }
+
+  return TRUE;
+}
+
 /**
  * ostree_repo_reload_config:
  * @self: repo
@@ -3130,6 +3156,8 @@ ostree_repo_reload_config (OstreeRepo          *self,
   if (!reload_core_config (self, cancellable, error))
     return FALSE;
   if (!reload_remote_config (self, cancellable, error))
+    return FALSE;
+  if (!reload_sysroot_config (self, cancellable, error))
     return FALSE;
   return TRUE;
 }
@@ -6063,4 +6091,22 @@ ostree_repo_get_default_repo_finders (OstreeRepo *self)
   g_return_val_if_fail (OSTREE_IS_REPO (self), NULL);
 
   return (const gchar * const *)self->repo_finders;
+}
+
+/**
+ * ostree_repo_get_bootloader:
+ * @self: an #OstreeRepo
+ * 
+ * Get the bootloader configured. See the documentation for the
+ * "sysroot.bootloader" config key.
+ * 
+ * Returns: bootloader configuration for the sysroot
+ * Since: 2019.2
+ */
+const gchar *
+ostree_repo_get_bootloader (OstreeRepo   *self)
+{
+  g_return_val_if_fail (OSTREE_IS_REPO (self), NULL);
+
+  return self->bootloader;
 }
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -5186,8 +5186,12 @@ _ostree_repo_gpg_verify_data_internal (OstreeRepo    *self,
 
       g_auto(GStrv) gpgkeypath_list = NULL;
 
-      if (!ot_keyfile_get_string_as_list (remote->options, remote->group, "gpgkeypath",
-                                          ";,", &gpgkeypath_list, error))
+      if (!ot_keyfile_get_string_list_with_separator_choice (remote->options,
+                                                             remote->group,
+                                                             "gpgkeypath",
+                                                             ";,",
+                                                             &gpgkeypath_list,
+                                                             error))
         return NULL;
 
       if (gpgkeypath_list)
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2310,6 +2310,7 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
   gboolean bootloader_is_atomic = FALSE;
   SyncStats syncstats = { 0, };
   g_autoptr(OstreeBootloader) bootloader = NULL;
+  const char *bootloader_config = NULL;
   if (!requires_new_bootversion)
     {
       if (!create_new_bootlinks (self, self->bootversion,
@@ -2342,8 +2343,22 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
             return glnx_throw_errno_prefix (error, "Remounting /boot read-write");
         }
 
-      if (!_ostree_sysroot_query_bootloader (self, &bootloader, cancellable, error))
-        return FALSE;
+      OstreeRepo *repo = ostree_sysroot_repo (self);
+
+      bootloader_config = ostree_repo_get_bootloader (repo);
+
+      g_debug ("Using bootloader configuration: %s", bootloader_config);
+
+      if (g_str_equal (bootloader_config, "auto"))
+        {
+          if (!_ostree_sysroot_query_bootloader (self, &bootloader, cancellable, error))
+            return FALSE;
+        }
+      else if (g_str_equal (bootloader_config, "none"))
+        {
+          /* No bootloader specified; do not query bootloaders to run. */
+        }
+
       bootloader_is_atomic = bootloader != NULL && _ostree_bootloader_is_atomic (bootloader);
 
       /* Note equivalent of try/finally here */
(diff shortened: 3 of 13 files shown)


cgwalters:
Related: https://github.com/ostreedev/ostree/issues/1719

I need to dive into this more; not entirely sure it's right (but it doesn't seem necessarily wrong either).  There's risk here in changing the grub backend.

I know it's more work but if we can get FCOS systems into having a static grub config (i.e. ostree only writes BLS snippets, use the grub BLS plugin) the whole architecture is much clearer.

rfairley:
Thanks @cgwalters - just pushed a change. `new_deployments` seems to hold the pending deployment as the first entry - so I think this is the data that needs to be passed in order for grub2-mkconfig to use the pending deployment.

> There's risk here in changing the grub backend.

Agreed on the risk - I think now this is a question of "could this change break existing systems that might have been relying on the existing deployment being used rather than the pending".

I'm interested in taking a look at using the BLS plugin in OSTree - I need to read into it more, but now that I've had a look at the bootloader code a bit better, I think I can see what would need to be changed.

rfairley:
As part of this, going to start on an implementation for writing the BLS config, and a way to configure the bootloader (something like https://github.com/ostreedev/ostree/issues/1801#issuecomment-453854856).

One way after brief discussion is to have OSTree always write BLS configs, and ship with a config `bootloader=auto` to use the auto-detecting bootloader functionality currently in place. This can be configured to `bootloader=none` to **not** use the auto-detecting functionality (only BLS configs will be written).

cgwalters:
> One way after brief discussion is to have OSTree always write BLS configs

That's what ostree has done since the very start.  The BLS configs are canonical.  But at the time RHEL7's grub didn't speak BLS (actually no bootloaders did really) so I added the concept of a "bootloader backend" to ostree as a temporary hack.

libostree never *reads* the grub config - it reads the BLS configs it writes as the source of truth.

rfairley:
> That's what ostree has done since the very start

Ahh that's clear to me now, thanks!

```
# ls /boot/loader/entries/ostree-
ostree-1-fedora-atomic.conf  ostree-2-fedora-atomic.conf
```

So what's needed is really having a switch in the bootloader backend to use or not use it (at least for the grub2-mkconfig part), and a way to configure that switch (through `bootloader=`)?

rfairley:
Dropped the edits to the grub2 bootloader code, and started on adding a `sysroot.bootloader` key to the ostree repo config.


WIP, with the following left to do:
- [ ] Errors are given due to the `[sysroot]` group not existing in the default `repo/config` files. Looking for a way to make that field optional.
- [ ] Update the docs.
- [ ] Add a test.

Right now I'm testing with the following commands:

```
mkdir sysroot && ostree admin init-fs sysroot
echo "[sysroot]" >> sysroot/ostree/repo/config
echo "bootloader=none" >> sysroot/ostree/repo/config
ostree pull-local --repo=sysroot/ostree/repo /sysroot/ostree/repo
ostree admin os-init osname --sysroot sysroot
ref=$(ostree refs --repo sysroot/ostree/repo) && mkdir -p sysroot/boot/grub2 && touch sysroot/boot/grub2/grub.cfg
ostree admin deploy "$ref" --sysroot sysroot --os osname -v 2>err.txt
ls sysroot/boot/loader/entries/ostree-1-osname.conf
```

src/libostree/ostree-repo.c
```diff
@@ -3113,6 +3113,53 @@ reload_remote_config (OstreeRepo          *self,
...
+{
+  { g_autofree char *bootloader = NULL;
+
+    if (!ot_keyfile_get_value_with_default (self->config, "sysroot",
+                                            "bootloader", "auto",
+                                            &bootloader, error))
+      {
+        /* XXX: quick hack to make existing config files not fail
+         * on not having the [sysroot] group present. */
+        if (error)
```
jlebon: The pattern for this normally is creating a local_error so you can always check for the exact error, and then `g_propagate_error` if needed.

src/libostree/ostree-repo.c
```diff
@@ -3113,6 +3113,53 @@ reload_remote_config (OstreeRepo          *self,
   return TRUE;
 }
 
+static gboolean
+reload_sysroot_config (OstreeRepo          *self,
+                       GCancellable        *cancellable,
+                       GError             **error)
+{
+  { g_autofree char *bootloader = NULL;
```
jlebon: Why the nested scope here?

src/libostree/ostree-repo.c
```diff
@@ -3113,6 +3113,53 @@ reload_remote_config (OstreeRepo          *self,
...
+    if (bootloader == NULL)
+      return glnx_throw (error, "Invalid empty bootloader configuration");
+
+    /* Can later add support for specifying a single bootloader e.g.
+     * "grub", "sd-boot", "rpi" here too. See:
+     * https://github.com/ostreedev/ostree/issues/1719
+     * https://github.com/ostreedev/ostree/issues/1801
+     */
+    if (strcmp (bootloader, "auto") != 0 &&
+        strcmp (bootloader, "none") != 0)
```
jlebon: `!g_str_equal`

src/libostree/ostree-repo.c
jlebon: Ahh I see, in case we have more configs in `[sysroot]`, I guess?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2310,6 +2311,7 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
   gboolean bootloader_is_atomic = FALSE;
   SyncStats syncstats = { 0, };
   g_autoptr(OstreeBootloader) bootloader = NULL;
+  g_autofree const char *bootloader_config = NULL;
```
jlebon: That should just be `const char*`, no?

rfairley: Ahh, yes, shouldn't be freeing that here.

src/libostree/ostree-repo.c
rfairley: Fixing!

src/libostree/ostree-repo.c
rfairley: Yes - that was it. I was mostly copying the pattern from the `reload_core_config()` and `reload_remtoe_config()` - but also in case we add more to `[sysroot]`.

src/libostree/ostree-repo.c
rfairley: Got it - makes sense.

rfairley:
I'm thinking bd082adaca9a7e95207341af8e7e1e832896f0fc is a clean way to make the `sysroot` group optional by extending the existing API to give a default value if the group is not found.

So far, the `core` group uses this API, aside from a case where we don't handle the error: https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo.c#L2928.

A problem I see with this is not giving the user an error if for some reason `[core]` is removed from the config file. Config files where the key exists but the group `[core]` does not would still have default values applied (wherever `ot_keyfile_get_value_with_default` is used). This shouldn't be a problem if the config is done through `ostree config set` though, and I think the documentation showing `[core]` being needed in the file is sufficient.

rfairley:
Hmm - seeing the tests for keyfile utils https://github.com/ostreedev/ostree/blob/master/tests/test-keyfile-utils.c#L126, probably best to keep the API as it is and instead only make the `[sysroot]` section optional (by checking if the group exists in `reload_sysroot_config` and just returning success if it doesn't exist).

rfairley:
Added tests for the `none` bootloader config option. Ready for another round of review / comments on the approach.

For unit tests for the added keyfile util function `ot_keyfile_get_value_with_default_group_optional()`, the only change from `ot_keyfile_get_value_with_default()` is here: https://github.com/ostreedev/ostree/pull/1814/commits/1f181097da8a1070cc2f6b45e6b6ba0c97dcdab1#diff-11958a5431bcbe882db3033c8de63f5eR190.

Also updated the first comment of this PR to match the commit description of  [`71915df`](/ostreedev/ostree/commit/71915df5e148c4593ba3d44d1f1ebfd16d79c168).

rfairley:
Noticed this is still failing the CI tests on `tests/test-admin-deploy-none.sh`. Will fix these, and add a note in docs for the new option.

rfairley:
Think that should fix the failing test, `${CMD_PREFIX}` was not present on one of the `ostree` commands.

Added some documentation in the ostree repo config manpage - lifting WIP

rfairley:
The reason `test-admin-deploy-none.sh` was failing ([log](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/d37852775ae20ef47ddc821c04d75ddc1c2c1740.0.1550846924962181195/artifacts/gdtr-results/libostree_test-admin-deploy-none.sh.test.txt)) with error  `fsetxattr: Operation not supported` was due to setting xattrs not supported on tmpfs (same as https://github.com/projectatomic/atomic/issues/1247). This occurred during `admin deploy` while copying `/etc/a-new-default-config-file` which had been committed to the tree in `admin-test.sh`. Updated the test to not require copying that file, by setting up a clean repo with `setup_os_repo()` and deploying the tree from there.

Just tidied up the test, a bit - should be passing now.

man/ostree.repo-config.xml
```diff
@@ -333,6 +333,41 @@ Boston, MA 02111-1307, USA.
 
   </refsect1>
 
+  <refsect1>
+    <title>[sysrooot] Section Options</title>
```
cgwalters: `sysroot`

man/ostree.repo-config.xml
```diff
@@ -333,6 +333,41 @@ Boston, MA 02111-1307, USA.
...
+
+    <para>
+      Options for the sysroot, which OSTree deploys the OSTree repository to.  The following entries are defined:
```
cgwalters: How about "which contains the OSTree repository, deployments, and stateroots".

man/ostree.repo-config.xml
```diff
@@ -333,6 +333,41 @@ Boston, MA 02111-1307, USA.
...
+        <literal>bootloader=none</literal> or <literal>bootloader=auto</literal>.
+        Default is <literal>auto</literal>.
+        </para>
+        <para>
+          If <literal>none</literal>, then OSTree will generate only BLS (Boot
+          Loader Specification) fragments in <literal>sysroot/boot/loader/entries/</literal>
+          for the deployment.
+        </para>
+        <para>
+          If <literal>auto</literal>, then OSTree will dynamically check for the
```
cgwalters: "then in addition to generating the BLS fragments, OSTree will..."

src/libostree/ostree-repo.c
```diff
@@ -3113,6 +3113,27 @@ reload_remote_config (OstreeRepo          *self,
...
+{
+  { g_autofree char *bootloader = NULL;
+
+    if (!ot_keyfile_get_value_with_default_group_optional (self->config, "sysroot",
+                                                           "bootloader", "auto",
+                                                           &bootloader, error))
+      return FALSE;
+
+    if (bootloader == NULL)
+      return glnx_throw (error, "Invalid empty bootloader configuration");
```
cgwalters: I don't think this case can happen...did you mean to compare vs the empty string `""`?  Either way it should probably be instead something like:

```
if (!(g_str_equal (bootloader, "auto") || g_str_equal (bootloader, "none"))
```

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2342,8 +2343,36 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
...
+      if (g_str_equal (bootloader_config, "auto"))
+        {
+          if (!_ostree_sysroot_query_bootloader (self, &bootloader, cancellable, error))
+            return FALSE;
+        }
+      else if (g_str_equal (bootloader_config, "none"))
+        {
+          /* No bootloader specified; do not query bootloaders to run. */
+        }
+      else
```
cgwalters: Then we can drop the code here.  Seems better to me to validate it right after parsing above.

man/ostree.repo-config.xml
rfairley: :+1: SGTM, thanks

src/libostree/ostree-repo.c
rfairley: Hmm, for some reason thought it would occur with `bootloader=` in the keyfile, but looks like this just returns an empty string.

Initially I was thinking that `reload_sysroot_config()` should allow any value for `bootloader=$value`, and then have `ostree_sysroot_write_deployments_with_options()` search for something that matches the value (e.g. if we later supported a `/etc/ostree/bootloaders`, `ostree_sysroot_write_deployments_with_options()` would search for `/etc/ostree/bootloaders/my_bootloader` for `bootloader=my_bootloader`). I agree now that validating right at the parsing for this change is better - can always be reworked if `/etc/ostree/bootloaders` or `/usr/lib/ostree/bootloaders` is implemented. Fixing!

src/libostree/ostree-sysroot-deploy.c
rfairley: :+1: Fixing

rfairley:
Working on the review fixes - will try fixing #1827 so that a test

```
if ${CMD_PREFIX} ostree --repo=sysroot/ostree/repo config set sysroot.bootloader some_unsupported_bootloader 2>err.txt; then
    assert_not_reached "unexpectedly succeeded at setting an unsupportd bootloader config"
fi
assert_file_has_content err.txt "Invalid bootloader configuration.*some_unsupported_bootloader"
echo "ok fail on unsupported bootloader config"
```

can be added (right now unexpectedly succeeds).

rfairley:
^ Pushed fixes for the review comments. Since a test case is present for the functionality in this PR, and the failing config test cases are likely to change soon if additional generic bootloader config options are added, I'm thinking it's OK to defer on those tests, and add them in if #1827 is fixed soon.

rfairley:
Sorry for the multiple pushes - the diff from the last 4 pushes (in response to review): https://github.com/ostreedev/ostree/compare/27c643a062844e24e79819c2b24bcb51c37b76cd..1ee554d05a506d18eccbea0d69bf356be15fba9d

src/libostree/ostree-repo.c
```diff
@@ -5186,8 +5186,12 @@ _ostree_repo_gpg_verify_data_internal (OstreeRepo    *self,
```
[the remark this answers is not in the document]
rfairley: Thanks, fixed!

src/libotutil/ot-keyfile-utils.c
```diff
@@ -101,6 +101,44 @@ ot_keyfile_get_value_with_default (GKeyFile      *keyfile,
...
 }
 
+gboolean
+ot_keyfile_get_value_with_default_group_optional (GKeyFile      *keyfile,
+                                                  const char    *section,
+                                                  const char    *value,
+                                                  const char    *default_value,
+                                                  char         **out_value,
+                                                  GError       **error)
+{
```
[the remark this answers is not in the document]
rfairley: :+1: sounds good to me, made this change. I was stuck deciding between the two before (calling out to `ot_keyfile_get_value_with-default` vs re-implementing/copy+paste). Seems better overall to call out and avoid the duplicate code.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2342,8 +2343,24 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
             return glnx_throw_errno_prefix (error, "Remounting /boot read-write");
         }
 
-      if (!_ostree_sysroot_query_bootloader (self, &bootloader, cancellable, error))
-        return FALSE;
+      g_autoptr(OstreeRepo) repo = NULL;
+      if (!ostree_sysroot_get_repo (self, &repo, cancellable, error))
```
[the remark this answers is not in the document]
rfairley: Fixed!

rfairley:
Updated, and fixed the commit messages!
Diff from last review: https://github.com/ostreedev/ostree/compare/1ee554d05a506d18eccbea0d69bf356be15fba9d..2c3cd83a913d395b500b4528f78073838a2d43ab

rfairley:
Also split https://github.com/ostreedev/ostree/pull/1814/commits/5b5093f90318433192f7ff7b1e2c59ed8eb7d534 into its own commit (previously added tests for `sysroot.bootloader` there and did this edit in passing, but later removed the tests).

jlebon:
Awesome, LGTM!
@rh-atomic-bot r+ 2c3cd83

jlebon:
@rh-atomic-bot retry

dustymabe:
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️

AdrianVovk:
:+1:

AdrianVovk:
:+1: Great work!

rfairley:
Thanks :)

PR #1815, opened 2019-02-09
tests: Remove unused includes in libostreetest.c

mwleeds:
Since commit a06bd82cd we no longer use OVERLAYFS_SUPER_MAGIC or
statfs() so remove the includes for linux/magic.h and sys/vfs.h


cgwalters:
@rh-atomic-bot r+ f97af84

PR #1821, opened 2019-02-14
lib/repo-refs: Resolve collection-refs in-memory and in parent repos

mwleeds:
Currently the behavior of ostree_repo_resolve_rev() is that it tries to
resolve a ref to a commit by checking the refs/ directories, but also by
checking for in-memory ref-checksum pairs which are part of an
in-progress transaction and also by checking the parent repo if one
exists. Currently ostree_repo_resolve_collection_ref() only checks the
refs/ directories, so this commit makes its behavior analagous since it
is the analagous API which supports collection-refs.

The impetus for this was that currently Flatpak uses
ostree_repo_resolve_rev() to load a commit after doing a P2P pull in
flatpak_dir_do_resolve_p2p_refs(), but that assumes the ref came from
the same remote that originally provided it, which might not be the case
if more than one remote has the same collection ID configured. And
changing Flatpak to use ostree_repo_resolve_collection_ref() doesn't
work without this patch.

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -537,7 +537,30 @@ ostree_repo_resolve_collection_ref (OstreeRepo                    *self,
                                          cancellable, error))
     return FALSE;
 
-  const char *ret_contents = g_hash_table_lookup (refs, ref);
+  g_autofree char *ret_contents = g_strdup (g_hash_table_lookup (refs, ref));
+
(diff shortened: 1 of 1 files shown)


pwithnall:
The test failure is that `vmsync` thing again, which is unrelated.

pwithnall:
@rh-atomic-bot, r+ 36933d9d987f89199a40ddbfe1adb7d072cf5e05

PR #1822, opened 2019-02-15
lib/repo-pull: Don't omit results that seem out-of-date

mwleeds:
Currently in find_remotes_cb() which is the callback that does the
processing for ostree_repo_find_remotes_async(), if any result doesn't
provide the latest version of any ref it gets dropped from the final set
of results that are returned. This presents a problem if you want to be
robust against a source (configured remote, LAN, or USB) providing
malicious updates, as we do in Flatpak which doesn't trust each remote
to the same degree. Another reason you might want to pull from results
that seem to provide outdated refs is if the pull from the result
providing the latest commit failed for external reasons (say, the USB
drive was removed).

So this commit leaves results in the set returned by
find_remotes_async() even if all their refs are mapped to NULL in the
ref_to_checksum map (meaning they're outdated compared to another
result).

Say I want to pull (com.example.Collection, example-ref) and I have
configured both good-remote and malicious-remote with that collection
ID. Say good-remote advertises commit v1 and remote malicious-remote
advertises commit v2 which seems newer but is signed with the malicious
GPG keyring. With this patch, I can have both remotes in the results
array and when the malicious-remote pull fails for lack of good GPG
signatures (see the "ref-keyring-map" option) a successful pull will
happen from good-remote for the older, legitimate commit.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -5665,10 +5665,11 @@ find_remotes_cb (GObject      *obj,
 
           if (n_latest_refs == 0)
             {
-              g_debug ("%s: Omitting remote ‘%s’ from results as none of its refs are new enough.",
+              /* Here we used to remove the result from the final set, but
+               * let's leave it in since the result providing the latest might
+               * have bad signatures or another failure */
+              g_debug ("%s: Remote ‘%s’ does not have the latest rev of ony ref; leaving in results as a fallback.",
                        G_STRFUNC, result->remote->name);
-              ostree_repo_finder_result_free (g_steal_pointer (&g_ptr_array_index (results, i)));
-              continue;
             }
         }
 


pwithnall:
Would all the code which currently calls `find_remotes_async()` be OK with receiving results like this; or is it effectively an API break because callers are expecting all results to have at least one entry in the `ref_to_checksum` map?

mwleeds:
> Would all the code which currently calls `find_remotes_async()` be OK with receiving results like this; or is it effectively an API break because callers are expecting all results to have at least one entry in the `ref_to_checksum` map?

It's not that we're going to return results with empty `ref_to_checksum` maps. It's that we're going to return results with a `ref_to_checksum` map which might have every ref mapped to `NULL`. While looking into the question of whether any users of the API assume otherwise (that is, whether they assume a result has at least one non-NULL checksum) I found that `ostree_repo_pull_from_remotes_async` does not pull a ref from a result when its checksum is `NULL`, and there's not an easy fix for that. It wouldn't make much sense to merge this change without also changing that behavior, so I'll close this. It turns out it's not strictly necessary for https://github.com/flatpak/flatpak/pull/2705 which I think is because flatpak has some fallback code to use `ostree_repo_pull_with_options` when `ostree_repo_pull_from_remotes_async` fails.

This is also less important than the other patches because it's working toward the goal of being able to pull a legitimate update when a malicious one is being offered, which is less important than the goal of not using the malicious update. Presumably if the attack vector isn't exploitable it won't be used, and you could just remove the bad remote if it's clear what's going on.

PR #1823, opened 2019-02-17
static-delta: Change `show` to display from/to commits

cgwalters:
When writing a delta to a file this may not always be recorded
in the filename, and it's useful data.

This also required teaching `show` to accept a file path.

Note...for some reason `test-deltas.sh` breaks when run from
a tty - we get `SIGTTIN` which implies something is reading from
the tty but it wasn't obvious to me what.

### DIFF
--- a/src/libostree/ostree-repo-static-delta-core.c
+++ b/src/libostree/ostree-repo-static-delta-core.c
@@ -725,16 +725,25 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
                                 GCancellable                  *cancellable,
                                 GError                       **error)
 {
-  g_autofree char *from = NULL;
-  g_autofree char *to = NULL;
-  if (!_ostree_parse_delta_name (delta_id, &from, &to, error))
-    return FALSE;
-
-  g_autofree char *superblock_path = _ostree_get_relative_static_delta_superblock_path (from, to);
-
   glnx_autofd int superblock_fd = -1;
-  if (!glnx_openat_rdonly (self->repo_dir_fd, superblock_path, TRUE, &superblock_fd, error))
-    return FALSE;
+
+  if (strchr (delta_id, '/'))
+    {
+      if (!glnx_openat_rdonly (AT_FDCWD, delta_id, TRUE, &superblock_fd, error))
+        return FALSE;
+    }
+  else
+    {
+      g_autofree char *from = NULL;
+      g_autofree char *to = NULL;
+      if (!_ostree_parse_delta_name (delta_id, &from, &to, error))
+        return FALSE;
+
+      g_autofree char *superblock_path = _ostree_get_relative_static_delta_superblock_path (from, to);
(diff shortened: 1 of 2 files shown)


davisRoman:
Hi Colon,

Thank you very much for this patch. I think this will be a useful feature for ostree.

I was able to pull and compile your branch,delta-dump-filename, however the show command didn't quite work for me. My machine doesn't have the main ostree commit repo because that lives on our build server.  I figured making an empty repo on my machine would suffice as I'm only interested in running the show command to retrieve the start and end commits from the static-delta.

davis@Precision-7920:/tmp$ mkdir ostree-repo
davis@Precision-7920:/tmp$ ostree --repo=ostree-repo init --mode=bare
davis@Precision-7920:/tmp$ ~/ostree/ostree --repo=ostree-repo static-delta show ./static-88cf8a-96bc7b
error: Invalid rev ./static
davis@Precision-7920:/tmp$

Please let me know what you think. 

Thank you,

Davis

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -725,23 +725,52 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+  if (g_variant_n_children (from_commit_v) > 0)
+    {
+      if (!ostree_checksum_bytes_peek_validate (from_commit_v, error))
+        return FALSE;
+      from_commit = ostree_checksum_from_bytes_v (from_commit_v);
+      g_print ("From: %s\n", from_commit);
+    }
+  else
+    {
+      g_print ("From scratch: true\n");
```
jlebon: Minor: maybe `From: <scratch>` here to keep the `From:` format consistent?

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -725,23 +725,52 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
   g_autoptr(GVariant) delta_superblock = NULL;
   if (!ot_variant_read_fd (superblock_fd, 0,
                            (GVariantType*)OSTREE_STATIC_DELTA_SUPERBLOCK_FORMAT,
                            TRUE, &delta_superblock, error))
     return FALSE;
 
   g_print ("Delta: %s\n", delta_id);
+  g_autoptr(GVariant) from_commit_v = NULL;
+  g_variant_get_child (delta_superblock, 2, "@ay", &from_commit_v);
+  g_autofree char *from_commit = NULL;
```
jlebon: Minor: could move this inside the if scope.

cgwalters: Hmm no, it's referenced below right?
`      if (!show_one_part (self, swap_endian, from_commit, to_commit, meta_entries, i,`

jlebon:
@rh-atomic-bot r+ 9ee91ef

PR #1825, opened 2019-02-20
lib/repo-refs: Fix resolving collection-refs

mwleeds:
My last commit "lib/repo-refs: Resolve collection-refs in-memory and in
parent repos" changed ostree_repo_resolve_collection_ref() to check the
in-memory set of refs *after* failing to find the ref on disk but that's
not what we want. We want to use the in-memory set of refs first,
because those are the most up to date commits, and then fall back to the
on-disk repo and finally fall back to checking any parent repo. This
commit makes such a change to the order of operations, which is
consistent with how ostree_repo_resolve_rev() works.

Aside from this change being logical, it also fixes some unit test
failures on an unmerged branch of flatpak:
https://github.com/flatpak/flatpak/pull/2705

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -514,7 +514,7 @@ ostree_repo_resolve_rev_ext (OstreeRepo                    *self,
  * the given @ref cannot be found, a %G_IO_ERROR_NOT_FOUND error will be
  * returned.
  *
- * If you want to check only local refs not remote or mirrored ones, use the
+ * If you want to check only local refs, not remote or mirrored ones, use the
  * flag %OSTREE_REPO_RESOLVE_REV_EXT_LOCAL_ONLY. This is analogous to using
  * ostree_repo_resolve_rev_ext() but for collection-refs.
  *
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -479,6 +479,9 @@ ostree_repo_resolve_rev (OstreeRepo     *self,
  * the parameter @out_rev. Differently from ostree_repo_resolve_rev(),
  * this will not fall back to searching through remote repos if a
  * local ref is specified but not found.
+ *
+ * The flag %OSTREE_REPO_RESOLVE_REV_EXT_LOCAL_ONLY is implied so
+ * using it has no effect.
  */
 gboolean
 ostree_repo_resolve_rev_ext (OstreeRepo                    *self,
@@ -511,7 +514,9 @@ ostree_repo_resolve_rev_ext (OstreeRepo                    *self,
  * the given @ref cannot be found, a %G_IO_ERROR_NOT_FOUND error will be
  * returned.
  *
- * There are currently no @flags which affect the behaviour of this function.
+ * If you want to check only local refs not remote or mirrored ones, use the
+ * flag %OSTREE_REPO_RESOLVE_REV_EXT_LOCAL_ONLY. This is analogous to using
+ * ostree_repo_resolve_rev_ext() but for collection-refs.
  *
  * Returns: %TRUE on success, %FALSE on failure
  * Since: 2018.6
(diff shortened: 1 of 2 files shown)


mwleeds:
bot, retest this please

pwithnall:
bot, retest this please

mwleeds:
I'm not able to reproduce the failure of `test-pull-repeated.sh` locally. I think it's just non-deterministic. According to a comment in that file:
```
# Using 8 network retries gives error rate of <0.5%, when --random-408s=50
```
So I guess we hit that `0.5%` here

mwleeds:
This branch now has two more commits which need review

src/libostree/ostree-repo-refs.c
```diff
@@ -511,7 +514,9 @@ ostree_repo_resolve_rev_ext (OstreeRepo                    *self,
  * the given @ref cannot be found, a %G_IO_ERROR_NOT_FOUND error will be
  * returned.
  *
- * There are currently no @flags which affect the behaviour of this function.
+ * If you want to check only local refs not remote or mirrored ones, use the
```
pwithnall:
```suggestion
 * If you want to check only local refs, not remote or mirrored ones, use the
```

src/libostree/ostree-repo-refs.c
```diff
@@ -535,21 +540,32 @@ ostree_repo_resolve_collection_ref (OstreeRepo                    *self,
 
   /* Check for the ref in the current transaction in case it hasn't been
    * written to disk, to match the behavior of ostree_repo_resolve_rev() */
-  if (self->in_transaction)
+  if (self->in_transaction && self->txn.collection_refs)
```
pwithnall: I don’t think you can access `self->txn.collection_refs` without holding the `txn_lock`, since it could be initialised at any time by another thread calling (for example) `ostree_repo_transaction_set_refspec()` (i.e. `ensure_txn_refs()`).

src/libostree/ostree-repo.h
```diff
@@ -462,9 +462,11 @@ gboolean      ostree_repo_resolve_rev (OstreeRepo  *self,
 /**
  * OstreeRepoResolveRevExtFlags:
  * @OSTREE_REPO_RESOLVE_REV_EXT_NONE: No flags.
+ * @OSTREE_REPO_RESOLVE_REV_EXT_LOCAL_ONLY: Exclude remote and mirrored refs. Since: 2019.2
  */
 typedef enum {
   OSTREE_REPO_RESOLVE_REV_EXT_NONE = 0,
+  OSTREE_REPO_RESOLVE_REV_EXT_LOCAL_ONLY = (1 << 0),
```
pwithnall: Would it make sense to expose two separate flags, `EXCLUDE_MIRRORS` and `EXCLUDE_REMOTES`, like we have for `OstreeRepoListRefsExtFlags`? Might make things a bit more consistent.

mwleeds: I get that the symmetry would be nice, but I don't think it's a distinction we need to make. Because remote and mirrored refs are both remote and both mirrored (in the sense that we serve both over LAN/USB). I'm trying to move Flatpak to use `refs/mirrors/` more, at which point the only distinction will be the on-disk location.

I think the only place the distinction between EXCLUDE_REMOTES and EXCLUDE_MIRRORS is needed for ostree_repo_list_refs() is in the implementation of ostree_repo_traverse_reachable_refs(), because otherwise you'd check the same refs twice.

src/libostree/ostree-repo-refs.c
mwleeds: Good point. In that case I opened https://github.com/ostreedev/ostree/issues/1826 because that rule doesn't seem to be followed consistently

mwleeds:
Addressed the feedback, thanks

src/libostree/ostree-repo.h
pwithnall: Shrug. Maybe a second opinion from @cgwalters would be good.

cgwalters: No strong opinion here either.

mwleeds: Okay, we'll go with `LOCAL_ONLY` then

mwleeds:
This should be merged before the next release to avoid breakage since https://github.com/ostreedev/ostree/pull/1821 was already merged.

mwleeds:
Anyone able to review the fixups and merge this? It's blocking https://github.com/flatpak/flatpak/pull/2705

src/libostree/ostree-repo-refs.c
```diff
@@ -1295,7 +1326,7 @@ ostree_repo_list_collection_refs (OstreeRepo                 *self,
 
   g_string_truncate (base_path, 0);
 
-  for (const char **iter = refs_dirs; iter && *iter; iter++)
+  for (const char **iter = (const char **)refs_dirs->pdata; iter && *iter; iter++)
```
jlebon: Since we have a `GPtrArray` now, it'd be cleaner to not append a final `NULL` and use `refs_dir->len`. But eh, not worth a respin!

jlebon:
LGTM!

@rh-atomic-bot r+ 7a4c5fd

PR #1829, opened 2019-03-01
lib/deploy: Don't include version twice in bootmenu title

jlebon:
In Silverblue right now, the boot menu title looks like this:

    Fedora 29.20190301.0 (Workstation Edition) 29.20190301.0 (ostree)

This is because RPM-OSTree's `mutate-os-release` feature is enabled,
which injects the OSTree version string directly into `VERSION` and
`PRETTY_NAME`. So appending the version string again is a bit redundant.
Let's just do a simple substring check here before adding the version to
the title.


jlebon:
I guess this doesn't catch the case where the OS name really does have a number in it and then the version string happens to match exactly that number. E.g. `TestOS 42` is the actual OS name, and then `TestOS 42` version 42 would just have the title `TestOS 42` instead of `TestOS 42 42`... but why do this to your users?

cgwalters:
@rh-atomic-bot r+ d74edce

cgwalters:
@rh-atomic-bot retry

PR #1835, opened 2019-04-02
lib: Check for http_proxy being the empty string

mwleeds:
On at least one user's computer, g_getenv("http_proxy") returns the
empty string, so check for that and treat it as no proxy rather than
printing a warning.

See https://github.com/flatpak/flatpak/issues/2790


cgwalters:
@rh-atomic-bot r+ dc925be

PR #1838, opened 2019-04-11
WIP: tests/test-config: Fix error regex

rfairley:
Draft PR to see if this works in CI.

Attempts to fix the failure: https://travis-ci.org/ostreedev/ostree/jobs/514465450 (search for `File 'err.txt' doesn't match regexp 'error: Key file` in the raw log)

### DIFF
--- a/tests/test-config.sh
+++ b/tests/test-config.sh
@@ -76,6 +76,8 @@ assert_not_file_has_content repo/config "lock-timeout-secs="
 if ${CMD_PREFIX} ostree config --repo=repo get core.lock-timeout-secs 2>err.txt; then
     assert_not_reached "ostree config get should not work after unsetting a key"
 fi
+cat err.txt
+cat repo/config
 assert_file_has_content err.txt "error: Key file does not have key “lock-timeout-secs” in group “core”"
 
 # Check that it's idempotent


rfairley:
Just made this PR non-draft to try and run it against the Travis CI build. Not sure if this could be why the test fails on the Jessie environment.

rfairley:
Use of `“”` doesn't look like the problem. Pushed another change to help debug.

rfairley:
Seems like the correct error is being given, but slightly different message `error: Key file does not have key 'lock-timeout-secs'` (Ctrl+F for that error in https://api.travis-ci.org/v3/job/518830164/log.txt), whereas we expect `"error: Key file does not have key “lock-timeout-secs” in group “core”"`.

After investigating a bit, seems like the reason for the difference is that https://github.com/GNOME/glib/commit/a12f546b3bd32ae826306c4c7dbd931fe37ec632#diff-bd0f991e03018b0eeac8cd97c5440cfa was applied to the upstream release `2.43.2`, but [jessie has glib2.0 2.42.1-1](https://packages.debian.org/source/jessie/glib2.0) as the upstream. Seems like this would be fixed by updating CI to [stretch (glib2.0 2.50.3-2)](https://packages.debian.org/source/stretch/glib2.0).

rfairley:
Closing this - opening #1839 for the fix

PR #1839, opened 2019-04-12
travis: Update to Debian Stretch

rfairley:
Update to Stretch to use updated version of glib2.0. Fixes CI
failure when parsing error output in tests/test-config.sh. See:

https://github.com/ostreedev/ostree/pull/1838#issuecomment-482186680
https://api.travis-ci.org/v3/job/518830164/log.txt

---

Leaving as WIP, to check if CI tests pass first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ sudo: required
 
 env:
   - ci_distro=ubuntu ci_suite=trusty ci_test=no  # TODO: use libcurl on this
-  - ci_docker=debian:jessie-slim ci_distro=debian ci_suite=jessie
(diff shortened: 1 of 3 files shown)


rfairley:
Just updated this - passing now (though https://travis-ci.org/ostreedev/ostree/jobs/521364798 failed on `tests/test-pull-repeated.sh` - which is known to flake so this should be fine).

Trying with `''` (which Debian Stretch expects) instead of `“”`, `tests/test-config.sh` fails in the Fedora test suites. Since it shouldn't matter the exact character used here, I've just changed it to match any character instead so it can pass on both distros. Not sure if this is a symptom of another problem causing the difference in the type of quotes in the first place.

cgwalters:
@rh-atomic-bot r+ 641c0c4

PR #1841, opened 2019-04-16
lib/deploy: Support locking deployment finalization

jlebon:
Teach `ostree-finalize-staged.service` to check for a file in `/run` to
determine if it should do the finalization. This will be used in
RPM-OSTree, where we want to be able to separate out "preparing updates"
from "making update the default" for more fine-grained control. See:
projectatomic/rpm-ostree#1748

### DIFF
--- a/src/libostree/ostree-fetcher-util.c
+++ b/src/libostree/ostree-fetcher-util.c
@@ -163,18 +163,16 @@ _ostree_fetcher_journal_failure (const char *remote_name,
                                  const char *url,
                                  const char *msg)
 {
-#ifdef HAVE_LIBSYSTEMD
   /* Sanity - we don't want to log this when doing local/file pulls */
   if (!remote_name)
     return;
-  sd_journal_send ("MESSAGE=libostree HTTP error from remote %s for <%s>: %s",
+  ot_journal_send ("MESSAGE=libostree HTTP error from remote %s for <%s>: %s",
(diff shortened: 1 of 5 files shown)


jlebon:
Ahh, keep forgetting to try out the new GitHub draft PR feature.
Anyway, marking as WIP for now. Still testing this!

cgwalters:
Do we expect rpm-ostree will e.g. always write out this file when it starts?

jlebon:
> Do we expect rpm-ostree will e.g. always write out this file when it starts?

Yeah, so basically I'm thinking something like:

- agent calls `UpdateDeployment` with `lock-finalization` option on
    - touch `/run/ostree/staged-deployment-locked`
    - stage deployment
- at a later time, agent calls `UnlockStagedDeployment`, which removes the file

jlebon:
I guess we could always write out the file on boot... Hmm, this ties into https://github.com/projectatomic/rpm-ostree/issues/1747 as well. E.g. if we're driven by an agent, then we could default to always locking finalization? Though I think I'd prefer making it explicit still and just let the "driver" part be about automatic updates and injecting info in `rpm-ostree status` as discussed in that issue.

jlebon:
OK, this one should be good to go!

cgwalters:
@rh-atomic-bot r+ b003fcd

PR #1843, opened 2019-04-23
lib/sysroot: Call `chmod` on overlay dirs when unlocking

jlebon:
Otherwise, we'll be subject to whatever `umask` is currently. Normally,
processes should respect `umask` when creating files and directories,
but specifically for `ostree admin unlock` (or `rpm-ostree usroverlay`),
this poses a problem since e.g. a `/usr` with mode 0700 will break any
daemon that doesn't run as root and needs to read files under `/usr`,
such as polkitd.

This patch just does a `chmod()` after the `mkdir()`. An alternative
would be to do `umask(0000)` after forking into the child process
that'll call `mount()`, but that'd require also moving the `mkdir()`
calls into there, making for a more intrusive patch.

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -1699,6 +1699,22 @@ clone_deployment (OstreeSysroot  *sysroot,
   return TRUE;
 }
 
+/* Do `mkdir()` followed by `chmod()` immediately afterwards to ensure `umask()` isn't
+ * masking permissions where we don't want it to. Thus we avoid calling `umask()`, which
+ * would affect the whole process. */
+static gboolean mkdir_unmasked (int                   dfd,
+                                const char           *path,
+                                int                   mode,
+                                GCancellable         *cancellable,
+                                GError              **error)
+{
+  if (!glnx_shutil_mkdir_p_at (dfd, path, mode, cancellable, error))
+    return FALSE;
+  if (fchmodat (dfd, path, mode, 0) < 0)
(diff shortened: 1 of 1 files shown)


jlebon:
I hit this after trying to `make vmsync` in rpm-ostree targeting Fedora CoreOS. Because of [another issue](https://github.com/coreos/fedora-coreos-config/pull/83), the umask was set to `0077`, which meant that `/usr` was `0700` and caused polkitd to fail to load policy files. At which point, no `rpm-ostree` command could actually run.

src/libostree/ostree-sysroot.c
```diff
@@ -1768,9 +1784,9 @@ ostree_sysroot_deployment_unlock (OstreeSysroot     *self,
          * directly for hotfixes.  The ostree-prepare-root.c helper
          * is also set up to detect and mount these.
          */
-        if (!glnx_shutil_mkdir_p_at (deployment_dfd, ".usr-ovl-upper", 0755, cancellable, error))
+        if (!mkdir_unmasked (deployment_dfd, ".usr-ovl-upper", 0755, cancellable, error))
```
cgwalters: Optional tweak - inherit the mode of the underlying `/usr` rather than hardcoding it.

jlebon: Good idea, done! :arrow_down: 

(Made this a separate commit to be explicit).

jlebon:
(I don't think we technically need to create the workdir with matching permissions, but eh...)

cgwalters:
@rh-atomic-bot r+ 6e7ed8e

PR #1847, opened 2019-05-01
Retry requests on HTTP 503 and 504

wjt:
PR #1594 added logic to retry downloads when they fail due to network errors, or due to HTTP 408 Request Timeout.

In practice, pulling from Flathub frequently fails with 503 Service Unavailable. This is a transient error, so it makes sense to retry when it's encountered.

There is no good code in GIOErrorEnum for this so this patch takes the lazy route of mapping it to G_IO_ERROR_TIMED_OUT, which is already treated as a transient error in _ostree_fetcher_should_retry_request().

While we're here, also map 504 Gateway Timeout to G_IO_ERROR_TIMED_OUT (which does seem correct, if lossy).

----

Aside from a general interest in automatically retrying in the face of transient errors, another reason I am interested in mapping these errors to something other than G_IO_ERROR_FAILED is that we would like to make gnome-software show a more useful error message when pulling an app fails. Right now, users see (an example transcribed from a screenshot):

> Unable to install Blender
> While pulling app/org.blender.Blender/x86_64/stable from remote flathub: Server returned status 503: Service Unavailable

I believe that "Server returned status ..." is untranslated -- it comes from

https://github.com/ostreedev/ostree/blob/3ca1035e989ae2efec3103bd70128ff34aa515aa/src/libostree/ostree-fetcher-soup.c#L1098

Of course, it's not libostree's fault that GNOME Software shows this error message (filtered through libflatpak) verbatim, but while this status code is mapped to G_IO_ERROR_FAILED it's impossible to distinguish from any other errors. I wonder whether it would actually be better to introduce an error domain for HTTP errors (I know libsoup has one but ostree also has a libcurl backend) and propagate the HTTP status code out.

### DIFF
--- a/src/libostree/ostree-fetcher-util.c
+++ b/src/libostree/ostree-fetcher-util.c
@@ -231,6 +231,8 @@ _ostree_fetcher_http_status_code_to_io_error (guint status_code)
     case 410:  /* SOUP_STATUS_GONE */
       return G_IO_ERROR_NOT_FOUND;
     case 408:  /* SOUP_STATUS_REQUEST_TIMEOUT */
+    case 503:  /* SOUP_STATUS_SERVICE_UNAVAILABLE */
+    case 504:  /* SOUP_STATUS_GATEWAY_TIMEOUT */
       return G_IO_ERROR_TIMED_OUT;
     default:
       return G_IO_ERROR_FAILED;
--- a/src/ostree/ostree-trivial-httpd.c
+++ b/src/ostree/ostree-trivial-httpd.c
@@ -47,12 +47,15 @@ static int opt_random_500s_percentage;
 static int opt_random_500s_max = 100;
 static int opt_random_408s_percentage;
 static int opt_random_408s_max = 100;
+static int opt_random_503s_percentage;
+static int opt_random_503s_max = 100;
 static gint opt_port = 0;
 static gchar **opt_expected_cookies;
 static gchar **opt_expected_headers;
 
 static guint emitted_random_500s_count = 0;
 static guint emitted_random_408s_count = 0;
+static guint emitted_random_503s_count = 0;
 
 typedef struct {
   int root_dfd;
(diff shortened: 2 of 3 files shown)


rfairley:
bot, test pull request

cgwalters:
Note there is a [Retry-After](https://tools.ietf.org/html/rfc2616#section-14.37) header - this came up in another context for me recently fetching container image.s  ideally we'd respect that.

I don't have a good sense for the general behavior of HTTP clients when dealing with codes like `503`.  It could clearly also be a situation that could take a day or more to resolve.

jlebon:
Without a `Retry-After`, we could probably do an exponential backoff between each retry so we don't just keep hammering the server (see e.g. `curl`'s `--retry` option). To keep it simple, we could just do this for all codes for which we retry today, rather than just 503s.

cgwalters:
> (see e.g. curl's --retry option)

Do you think it's worth trying to have some of these things be configurable as options in the remote?  I guess we can always do that later.  It'd be nice if some of this actually lived in libcurl rather than its CLI, but I understand why it isn't.

A good example of this actually is I think most user interfaces are going to want to visualize/describe this situation.  There's a big difference between "actively downloading" and "sitting around idly in a retry loop".  In other words we'd probably want to have something in the remote status for this?

jlebon:
@wjt Thoughts on the above feedback? I.e. check for `Retry-After` for when to schedule a retry, and otherwise just default to some delay?

> I wonder whether it would actually be better to introduce an error domain for HTTP errors (I know libsoup has one but ostree also has a libcurl backend) and propagate the HTTP status code out.

Yeah, that makes sense to me. Though I guess we'd technically be breaking compatibility, right? Which is *probably* OK since it's in the error path, though we might want to publicize it and give a heads-up to anyone who could be special-casing `G_IO` errors.

> Do you think it's worth trying to have some of these things be configurable as options in the remote? I guess we can always do that later.

That'd be nice. Just like dnf's `retries` config knob.

wjt:
> @wjt Thoughts on the above feedback? I.e. check for `Retry-After` for when to schedule a retry, and otherwise just default to some delay?

I wasn't aware of this header – I think this makes sense.

> > I wonder whether it would actually be better to introduce an error domain for HTTP errors (I know libsoup has one but ostree also has a libcurl backend) and propagate the HTTP status code out.
> 
> Yeah, that makes sense to me. Though I guess we'd technically be breaking compatibility, right? Which is _probably_ OK since it's in the error path, though we might want to publicize it and give a heads-up to anyone who could be special-casing `G_IO` errors.

I could be wrong, and I agree that it's worth putting this in the release notes and checking in with a few key components (rpm-ostree, eos-updater, flatpak I guess), but [GLib specifically cautions against checking for  `G_IO_ERROR_FAILED`](https://developer.gnome.org/gio/stable/gio-GIOError.html#GIOErrorEnum):

> Note that this domain may be extended in future GLib releases. In general, new error codes either only apply to new APIs, or else replace G_IO_ERROR_FAILED in cases that were not explicitly distinguished before. You should therefore avoid writing code like:
>
> ```c
> if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_FAILED))
>  {
>     // Assume that this is EPRINTERONFIRE
>     ...
>   }
> ```
>
>  but should instead treat all unrecognized error codes the same as G_IO_ERROR_FAILED.

It's true that what we're discussing is also changing the error *domain* for *all* HTTP errors but I think this is probably a safe-enough change. Another option would be to leave the special handling for 403, 404, 410, 408 in place and map other codes to a new domain.

So what we're proposing is:

- Create a new `OSTREE_HTTP_ERROR` domain whose codes are the HTTP error codes
- Either:
  - Remove `_ostree_fetcher_http_status_code_to_io_error()` and replace it with a simple mapping to this domain; or
  - Keep it, but change its `default:` branch to map into `OSTREE_HTTP_ERROR`
- Retry on 503 and 504
- Introduce a delay to retries, defaulting to exponential back-off
- Propagate `Retry-After` far enough out that it can override the exponential back-off
  - I'm not quite sure what the plumbing for these two would look like from a quick read of the code surrounding the calls to `_ostree_fetcher_should_retry_request` though
- Make a detailed note in NEWS
- Verify the new behaviour with major library users

Sound about right? I don't have time to work on this myself at the moment, unfortunately :(

wjt:
I won't be able to work on updating this PR any time soon. My personal motivation is reduced now that Flathub & Endless's CDNs are behaving themselves. :-) 

I'll close this PR. If someone wants to steal the patches, update them as described above & resubmit, go for it!

PR #1849, opened 2019-05-04
docs: fix switchroot source link

TristanCacqueray:
This change updates the switchroot source relative url to an absolute url so
that it works from the readthedocs service. This change also removes
redundant switchroot.sh reference.


cgwalters:
@rh-atomic-bot r+ ccbda10

PR #1851, opened 2019-05-05
prepare-root: remove ignored MS_MGC_VAL flag

TristanCacqueray:
Since Linux 2.4 this is no longer required and the flag is ignored.


cgwalters:
@rh-atomic-bot r+ af4d52b

Issue #1855, opened 2019-05-08
Maybe use mmap for summaries


hughsie:
In GNOME Software (using libflatpak, using libostree) we get reports of large memory usage. On my system there's nearly 10Mb of heap memory allocated by `_ostree_repo_load_cache_summary_if_same_sig()` which seems to use `read()` via `glnx_fd_readall_malloc()` via `glnx_fd_readall_bytes()`.

Is there a reason we have to use heap memory rather than mmap'ing the file? GBytes is smart enough to do `g_mapped_file_new()` -> `g_mapped_file_get_bytes()` without actually copying the data. If this sounds interesting of there's not a good technical reason for the heap allocation I can have a go at a PR for this. Thanks.

cgwalters:
We already have internally `ot_fd_readall_or_mmap()`.

cgwalters:
That said, I think it'd be better for gnome-software to not be a persistent process.

hughsie:
> That said, I think it'd be better for gnome-software to not be a persistent process.

I don't disagree, but we don't have systemd user sessions yet, and we're not at the point we can load packagekitd and flatpak in less than one second. We're getting there, but all the bits are not in place yet.

PR #1856, opened 2019-05-08
lib: mmap the summary to reduce RSS in clients

hughsie:
Closes: #1855


cgwalters:
@rh-atomic-bot r+ cbf1aca

PR #1857, opened 2019-05-09
Minor tweaks

rfairley:
Minor tweaks needed to build when running `make check` from an F30 container.


cgwalters:
@rh-atomic-bot r+ 29b876b

PR #1860, opened 2019-05-17
Update link to ostree-rs

fkrull:
I decided to rename the repo and the crates to be more in line with the gtk-rs naming scheme.


jlebon:
Thanks! @rh-atomic-bot r+ e8ca0c9

PR #1861, opened 2019-05-19
Various fixes to GI metadata

fkrull:
This PR fixes some issue with the GObject-Introspection metadata I ran into when updating the Rust bindings yesterday.

* `ostree_repo_get_min_free_space_bytes` had the wrong name in its doc comment which meant the GI generator didn't find it.
* Several functions were also missing "Since" tags so I added a bunch based on libostree-released.sym. (That wasn't a huge problem, but I noticed and it was bothering me.)
* There were a few functions with inconsistent "nullable" annotations.
* Some functions were missing the "out" annotation on their out parameters. These are always quite easy to spot because the generated code doesn't compile.

(It seems I also ended up nuking some trailing whitespace. Oh well.)

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -485,6 +485,8 @@ _ostree_raw_file_to_archive_stream (GInputStream       *input,
  *
  * Convert from a "bare" file representation into an
  * OSTREE_OBJECT_TYPE_FILE stream suitable for ostree pull.
+ *
+ * Since: 2016.6
  */
 gboolean
 ostree_raw_file_to_archive_z2_stream (GInputStream       *input,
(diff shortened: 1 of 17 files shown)


cgwalters:
Awesome, thanks in particular for the sleuthing :male_detective: on the `Since` annotations!  Hopefully no one is using a libostree remotely that old but it is still useful to see when things came in.

@rh-atomic-bot r+ 39a65de

(While you're here, I briefly looked at using your bindings in rpm-ostree but they need updating to glib-rs `0.8`, didn't look more deeply than that though)

fkrull:
@cgwalters I tried updating to newer glib-rs and gir, but they added a helper type called `ObjectType` and now there's a collision in the generated code (https://github.com/gtk-rs/gir/issues/772). So I've ignored that for now and focused on other aspects.

PR #1863, opened 2019-05-24
ostree-repo-finder: Clarify immutability after construction

pwithnall:
I think this was the intention, but I forgot to document it.


pwithnall:
At that point, those result structs have never been returned to code outside libostree, I think.

jlebon:
This looks sane to me. Will let @mwleeds do the final r+ if his concern has been addressed.
@rh-atomic-bot delegate=mwleeds

mwleeds:
@rh-atomic-bot r+

PR #1870, opened 2019-06-10
bin/refs: Allow overwriting existing ref

dbnicholson:
Currently if you want to update a non-alias ref, you need to first check
if it exists and use either `ostree refs --create` or `ostree reset` as
appropriate. That's unnecessarily complicated and is much less
convenient than the old `write-refs` builtin that simply called
`ostree_repo_set_ref_immediate()` without any checks.

Add a `--force` option to be used with `--create` that does not raise an
error when the destination ref already exists.

### DIFF
--- a/bash/ostree
+++ b/bash/ostree
@@ -952,6 +952,7 @@ _ostree_refs() {
         --collections -c
         --delete
         --list
+        --force
     "
 
     local options_with_args="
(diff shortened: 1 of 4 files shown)


jlebon:
LGTM, thanks!
@rh-atomic-bot r+ e5c7fe1

PR #1872, opened 2019-06-14
Show GPG messages about key expiration

dbnicholson:
Include the key expiration timestamp in the GPG verification result and show it in the description output. While here, also show if the key has been revoked.

I started working on a little more useful testing that involves creating GPG keys on the fly and changing their expiration times as needed, but I didn't get that done. I'm planning to do that for a related feature, but I figured I'd get this work out now instead of sitting on it. Let me know what you think.


dbnicholson:
I pushed 2 more commits to do more detailed testing of key states by generating and manipulating keys on the fly with `gpg`. It works here but as you likely know `gpg` can be pretty fragile.

src/libostree/ostree-gpg-verify-result.c
```diff
@@ -513,6 +513,49 @@ ostree_gpg_verify_result_describe (OstreeGpgVerifyResult *result,
...
 
+static void
+append_expire_info (GString *output_buffer,
+                    const gchar *line_prefix,
+                    const gchar *exp_type,
+                    gint64 exp_timestamp,
+                    gboolean expired)
+{
+  g_autoptr(GDateTime) date_time_utc = NULL;
+  g_autoptr(GDateTime) date_time_local = NULL;
```
cgwalters: General preference in ostree is for declare-and-init, i.e. C99 not C89.

cgwalters:
Not worth a respin just for that, overall looks great!  Thanks a lot!

@rh-atomic-bot r+ d8e2dfc

src/libostree/ostree-gpg-verify-result.c
dbnicholson: Copy and pasting there. I'll fix that for the next thing.

PR #1874, opened 2019-06-17
admin: allow multiple args for pin subcommand

### LINKED ISSUE #1816: Multiple arguments don’t get checked and handled
`ostree-2019.1-3.fc29.x86_64`

When I enter `sudo ostree admin pin 0 1`, deployment 0 gets pinned, but deployment 1 is not respected at all. I think it would be good if all arguments would be checked for correctness and then either an error should be thrown (because just one argument gets handled), or it should handle all arguments, if they get accepted. Don’t know if this also applies for other ostree commands.

The thing is, when I enter a command and it runs through without throwing an error, I assume that everything is done like expected. And as multiple arguments are widespread, I also would assume that this is handled correctly. Sadly, this isn’t the case. Not a big bug of course, but maybe this could save a deployment which instead might get overwritten some time in the future somewhere, and I just like correct and reliable tools, especially if they administer my system. ;)

r4f4:
Fixes #1816

### DIFF
--- a/src/ostree/ot-admin-builtin-pin.c
+++ b/src/ostree/ot-admin-builtin-pin.c
@@ -52,24 +52,39 @@ ot_admin_builtin_pin (int argc, char **argv, OstreeCommandInvocation *invocation
       return FALSE;
     }
 
-  const char *deploy_index_str = argv[1];
-  const int deploy_index = atoi (deploy_index_str);
-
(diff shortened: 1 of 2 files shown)


rfairley:
bot, add author to whitelist

rfairley:
LGTM, thanks!

@rh-atomic-bot r+ 6931106

PR #1876, opened 2019-06-20
lib/util: Drop unneeded ot-tool-util.h includes

rfairley:
With `ot-tool-util.h` made visible in `otutil.h` (in
be2572bf68090a5e277338d2613d3c7d53b0c9e8), drop previous includes
of `ot-tool-util.h` elsewhere.

### DIFF
--- a/src/ostree/ot-admin-builtin-set-origin.c
+++ b/src/ostree/ot-admin-builtin-set-origin.c
@@ -25,7 +25,6 @@
 #include "ot-admin-builtins.h"
 #include "ot-admin-functions.h"
 #include "ostree.h"
-#include "ot-tool-util.h"
 #include "otutil.h"
 
 #include <unistd.h>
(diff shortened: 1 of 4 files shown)


jlebon:
LGTM! @zonggen Want to review this one?
@rh-atomic-bot delegate=zonggen

(BTW, did you notice this by inspection, or did something throw a warning/error at you? If the latter, always good to include that in the commit message!)

cgwalters:
> or did something throw a warning/error at you?

A tool we could include here perhaps?

rfairley:
This was just by inspection :/ I had remembered from encountering the problem of `ot-tool-util.h` missing from `otutil.h` previously (but ended up not needing it) and at the time grepped for `ot-tool-util`.

It'd be good to have a tool that looks for unneeded headers - I could have a look into one.

zonggen:
LGTM too ! Merging

PR #1878, opened 2019-06-25
Alternative signing system

d4s:
Proposed the implementation of alternative signing system (as followup of https://github.com/ostreedev/ostree/issues/1233)

Approach is pretty simple: have a common interface for signing and verification, with implementation of particular sign/verify in separate modules.
- could co-exists with current GPG implementation
- allows to store and use multiple signature types in detached metadata in the same time
- `--with-libsodium` option is needed to build with support of `ed25515` signing engine 

### Current status:
- added the very first version of signing interface
- added "dummy" signing engine
- added "ed25519" signing engine
- added new builtin `ostree sign` (inspired by "ostree gpg-sign") allowing to sign and verify commits
- able to sign/verify commits
- able to sign/verify summary file
- added tests for commits and summary signing/verification
- added tests for pulling verification

not implemented:
- have no support of printing the status on the screen (yet)
- probably not integrated into all places there it should be used (need hints here)

### New configuration keys
For repository and remotes:
- `sign-verify` -- global and per-remote to trigger signature verification of updates
- `verification-key` -- per-remote -- for `ed25519`: base64 encoded public key to use for verification
 - `verification-file` -- per-remote -- for `ed25519`: file with the list of base64 public keys to  use for verification

### Dummy engine
Accept any ASCII string as public/secret key. Used mostly for testing the signing interface itself. Support only single public key for verification.

### Ed25519
Added "well-known" system places for `ed25519` public keys -- expected 1 base64 key per line:
- files
  - `/etc/ostree/trusted.ed25519`
  - `DATADIR + /ostree/trusted.ed25519`
- and directories:
  - `/etc/ostree/trusted.ed25519.d`
  - `DATADIR + /ostree/trusted.ed25519.d`

The same is for revoked keys:
  - `/etc/ostree/revoked.ed25519`
  - `/etc/ostree/revoked.ed25519.d`
  - `DATADIR + /ostree/revoked.ed25519`
  - `DATADIR + /ostree/rvokeded.ed25519.d`

Current logic for verification during the commits/summary file pulling:
- use `verification-key` if it exists in configuration
- use `verification-file` if it exists in configuration
- check public keys from well-known system places

### DIFF
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -709,7 +709,6 @@ ostree_kernel_args_to_string
 <SECTION>
 <FILE>ostree-sign</FILE>
 OstreeSign
-OstreeSignEd25519
 ostree_sign_list_names
 ostree_sign_commit
 ostree_sign_commit_verify
@@ -725,5 +724,4 @@ ostree_sign_set_sk
 ostree_sign_summary
 <SUBSECTION Standard>
 ostree_sign_get_type
-ostree_sign_ed25519_get_type
 </SECTION>
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -723,6 +723,7 @@ ostree_sign_clear_keys
 ostree_sign_load_pk
 ostree_sign_set_pk
 ostree_sign_set_sk
+ostree_sign_summary
 <SUBSECTION Standard>
 ostree_sign_get_type
 ostree_sign_dummy_get_type
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -709,7 +709,6 @@ ostree_kernel_args_to_string
 <SECTION>
 <FILE>ostree-sign</FILE>
 OstreeSign
-OstreeSignDummy
 OstreeSignEd25519
 ostree_sign_list_names
 ostree_sign_commit
@@ -726,6 +725,5 @@ ostree_sign_set_sk
 ostree_sign_summary
 <SUBSECTION Standard>
 ostree_sign_get_type
-ostree_sign_dummy_get_type
 ostree_sign_ed25519_get_type
 </SECTION>
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -719,6 +719,7 @@ ostree_sign_data_verify
 ostree_sign_get_by_name
 ostree_sign_get_name
 ostree_sign_add_pk
+ostree_sign_clear_keys
 ostree_sign_load_pk
 ostree_sign_set_pk
 ostree_sign_set_sk
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -723,7 +723,6 @@ ostree_sign_add_pk
 ostree_sign_load_pk
 ostree_sign_set_pk
 ostree_sign_set_sk
-ostree_sign_ed25519_keypair_generate
 <SUBSECTION Standard>
 ostree_sign_get_type
 ostree_sign_dummy_get_type
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -715,10 +715,9 @@ ostree_sign_list_names
 ostree_sign_commit
 ostree_sign_commit_verify
 ostree_sign_data
+ostree_sign_data_verify
 ostree_sign_get_by_name
 ostree_sign_get_name
-ostree_sign_detached_metadata_append
-ostree_sign_metadata_verify
 ostree_sign_add_pk
 ostree_sign_load_pk
 ostree_sign_set_pk
--- a/man/ostree-commit.xml
+++ b/man/ostree-commit.xml
@@ -251,6 +251,39 @@ Boston, MA 02111-1307, USA.
                     POLICY is a boolean which specifies whether fsync should be used or not.  Default to true.
                 </para></listitem>
             </varlistentry>
+
+            <varlistentry>
+                <term><option>-s, --sign-type</option></term>
+                <listitem><para>
+                    Use particular signature engine. Currently
+                    available <arg choice="plain">ed25519</arg> and <arg choice="plain">dummy</arg>
+                    signature types.
+
+                    The default is <arg choice="plain">ed25519</arg>.
+                </para></listitem>
+
+            </varlistentry>
+            <varlistentry>
+                <term><option>--sign</option>="KEY-ID"</term>
+                <listitem><para>
+                        There <literal>KEY-ID</literal> is:
+                        <variablelist>
+                            <varlistentry>
+                                <term><option>for ed25519:</option></term>
+                                <listitem><para>
+                                        <literal>base64</literal>-encoded secret key for commit signing.
+                                </para></listitem>
+                            </varlistentry>
+
+                            <varlistentry>
+                                <term><option>for dummy:</option></term>
+                                <listitem><para>
+                                            ASCII-string used as secret key.
+                                </para></listitem>
+                            </varlistentry>
+                        </variablelist>
+                </para></listitem>
+            </varlistentry>
         </variablelist>
     </refsect1>
 
--- a/man/ostree-sign.xml
+++ b/man/ostree-sign.xml
@@ -1,0 +1,152 @@
+<?xml version='1.0'?> <!--*-nxml-*-->
+<!DOCTYPE refentry PUBLIC "-//OASIS//DTD DocBook XML V4.2//EN"
+    "http://www.oasis-open.org/docbook/xml/4.2/docbookx.dtd">
+
+<!--
+Copyright 2019 Denis Pynkin <denis.pynkin@collabora.com>
+
+SPDX-License-Identifier: LGPL-2.0+
+
+This library is free software; you can redistribute it and/or
+modify it under the terms of the GNU Lesser General Public
+License as published by the Free Software Foundation; either
+version 2 of the License, or (at your option) any later version.
+
+This library is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with this library; if not, write to the
+Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+Boston, MA 02111-1307, USA.
+-->
+
+<refentry id="ostree">
+
+    <refentryinfo>
+        <title>ostree sign</title>
+        <productname>OSTree</productname>
+
+        <authorgroup>
+            <author>
+                <contrib>Developer</contrib>
+                <firstname>Colin</firstname>
+                <surname>Walters</surname>
+                <email>walters@verbum.org</email>
+            </author>
+        </authorgroup>
+    </refentryinfo>
+
+    <refmeta>
+        <refentrytitle>ostree sign</refentrytitle>
+        <manvolnum>1</manvolnum>
+    </refmeta>
+
+    <refnamediv>
+        <refname>ostree-sign</refname>
+        <refpurpose>Sign a commit</refpurpose>
+    </refnamediv>
+
+    <refsynopsisdiv>
+        <cmdsynopsis>
+            <command>ostree sign</command> <arg choice="opt" rep="repeat">OPTIONS</arg> <arg choice="req">COMMIT</arg> <arg choice="req" rep="repeat">KEY-ID</arg>
+        </cmdsynopsis>
+    </refsynopsisdiv>
+
+    <refsect1>
+        <title>Description</title>
+
+        <para>
+            Add a new signature to a commit.
+
+            Note that currently, this will append a new signature even if
+            the commit is already signed with a given key.
+        </para>
+
+        <para>
+            There are several "well-known" system places for `ed25519` trusted and revoked public keys -- expected single <literal>base64</literal>-encoded key per line.
+        </para>
+
+        <para>Files:
+            <itemizedlist>
+                <listitem><para><filename>/etc/ostree/trusted.ed25519</filename></para></listitem>
+                <listitem><para><filename>/etc/ostree/revoked.ed25519</filename></para></listitem>
+                <listitem><para><filename>/usr/share/ostree/trusted.ed25519</filename></para></listitem>
+                <listitem><para><filename>/usr/share/ostree/revoked.ed25519</filename></para></listitem>
+            </itemizedlist>
+        </para>
+
+        <para>Directories containing files with keys:
+            <itemizedlist>
+                <listitem><para><filename>/etc/ostree/trusted.ed25519.d</filename></para></listitem>
+                <listitem><para><filename>/etc/ostree/revoked.ed25519.d</filename></para></listitem>
+                <listitem><para><filename>/usr/share/ostree/trusted.ed25519.d</filename></para></listitem>
+                <listitem><para><filename>/usr/share/ostree/rvokeded.ed25519.d</filename></para></listitem>
+            </itemizedlist>
+        </para>
+    </refsect1>
+
+    <refsect1>
+        <title>Options</title>
+
+        <variablelist>
+            <varlistentry>
+                <term><option>KEY-ID</option></term>
+                <listitem><para>
+                        <variablelist>
+                            <varlistentry>
+                                <term><option>for ed25519:</option></term>
+                                <listitem><para>
+                                        <literal>base64</literal>-encoded secret (for signing) or public key (for verifying).
+                                </para></listitem>
+                            </varlistentry>
+
+                            <varlistentry>
+                                <term><option>for dummy:</option></term>
+                                <listitem><para>
+                                            ASCII-string used as secret key and public key.
+                                </para></listitem>
+                            </varlistentry>
+                        </variablelist>
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><option>--verify</option></term>
+                <listitem><para>
+                    Verify signatures
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><option>-s, --sign-type</option></term>
+                <listitem><para>
+                    Use particular signature mechanism. Currently
+                    available <arg choice="plain">ed25519</arg> and <arg choice="plain">dummy</arg>
+                    signature types.
+
+                    The default is <arg choice="plain">ed25519</arg>.
+                </para></listitem>
+            </varlistentry>
+           <varlistentry>
+                <term><option>--keys-file</option></term>
+                <listitem><para>
+                    Read key(s) from file <filename>filename</filename>.
+                </para></listitem>
+
+                <listitem><para>
+                    Valid for <literal>ed25519</literal> signature type.
+                    For <literal>ed25519</literal> this file must contain <literal>base64</literal>-encoded
+                    secret key(s) (for signing) or public key(s) (for verifying) per line.
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><option>--keys-dir</option></term>
+                <listitem><para>
+                    Redefine the system path, where to search files and subdirectories with
+                    well-known and revoked keys.
+                </para></listitem>
+            </varlistentry>
+        </variablelist>
+    </refsect1>
+</refentry>
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1505,6 +1505,17 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
 
   if (pull_data->sign_verify)
     {
+      /* Nothing to check if detached metadata is absent */
+      if (detached_metadata == NULL)
+        {
+          g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                               "Can't verify commit without detached metadata");
+          return FALSE;
+        }
+      /* Shouldn't happen, but see comment in process_verify_result() */
+      if (g_hash_table_contains (pull_data->verified_commits, checksum))
+        return TRUE;
+
       gboolean ret = FALSE;
       g_autoptr(GBytes) signed_data = g_variant_get_data_as_bytes (commit);
       /* list all signature types in detached metadata and check if signed by any? */
@@ -1512,13 +1523,14 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
       for (guint i=0; i < g_strv_length (names); i++)
         {
           g_autoptr (OstreeSign) sign = NULL;
-          g_autoptr(GVariant) signatures = NULL;
+          g_autoptr (GVariant) signatures = NULL;
           g_autofree gchar *signature_key = NULL;
           g_autofree GVariantType *signature_format = NULL;
+          g_autofree gchar *pk_ascii = NULL;
 
           if ((sign = ostree_sign_get_by_name (names[i], error)) == NULL)
           {
-              g_error_free (*error);
+              g_clear_error (error);
               continue;
           }
           signature_key = ostree_sign_metadata_key (sign);
@@ -1528,17 +1540,55 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
                                                signature_key,
                                                signature_format);
 
-          /* Set return to true if any sign fit */
           if (!signatures)
             continue;
 
+          /* TODO: load keys for remote here */
+          ostree_repo_get_remote_option (pull_data->repo,
+                                         pull_data->remote_name,
+                                         "verification-key", NULL,
+                                         &pk_ascii, NULL);
+          if (pk_ascii != NULL)
+            {
+              g_autoptr (GVariant) pk = NULL;
+
+              if (!g_strcmp0(ostree_sign_get_name(sign), "dummy"))
+                {
+                  // Just use the string as signature
+                  pk = g_variant_new_string(pk_ascii);
+                }
+              else if (!g_strcmp0(ostree_sign_get_name(sign), "ed25519"))
+                {
+                  gsize key_len = 0;
+                  g_autofree guchar *key = g_base64_decode (pk_ascii, &key_len);
+                  pk = g_variant_new_fixed_array (G_VARIANT_TYPE_BYTE, key, key_len, sizeof(guchar));
+                }
+
+              if (!ostree_sign_set_pk (sign, pk, error))
+                g_clear_error (error);
+            }
+
+          /* Set return to true if any sign fit */
           if (ostree_sign_metadata_verify (sign,
                                            signed_data,
                                            signatures,
                                            error
                                           ))
             ret = TRUE;
+          else
+            g_clear_error (error);
         }
+
+      /* Mark the commit as verified to avoid double verification
+       * see process_verify_result () for rationale */
+      if (ret)
+        {
+        g_hash_table_add (pull_data->verified_commits, g_strdup (checksum));
+        }
+      else
+        g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                             "Can't verify commit");
+
       return ret;
     }
 
@@ -1876,21 +1926,60 @@ scan_commit_object (OtPullData                 *pull_data,
     {
       gboolean ret = FALSE;
       /* list all signature types in detached metadata and check if signed by any? */
-      GStrv names = ostree_sign_list_names();
+      g_auto (GStrv) names = ostree_sign_list_names();
       for (guint i=0; i < g_strv_length (names); i++)
         {
-          g_autoptr (OstreeSign) sign = ostree_sign_get_by_name (names[i], error);
+          g_autoptr (OstreeSign) sign = NULL;
+          g_autofree gchar *pk_ascii = NULL;
 
+          if ((sign = ostree_sign_get_by_name (names[i], error)) == NULL)
+          {
+              g_clear_error (error);
+              continue;
+          }
+          /* TODO: load keys for remote here */
+          ostree_repo_get_remote_option (pull_data->repo,
+                                         pull_data->remote_name,
+                                         "verification-key", NULL,
+                                         &pk_ascii, NULL);
+          if (pk_ascii != NULL)
+            {
+              g_autoptr (GVariant) pk = NULL;
+
+              if (!g_strcmp0(ostree_sign_get_name(sign), "dummy"))
+                {
+                  // Just use the string as signature
+                  pk = g_variant_new_string(pk_ascii);
+                }
+              else if (!g_strcmp0(ostree_sign_get_name(sign), "ed25519"))
+                {
+                  gsize key_len = 0;
+                  g_autofree guchar *key = g_base64_decode (pk_ascii, &key_len);
+                  pk = g_variant_new_fixed_array (G_VARIANT_TYPE_BYTE, key, key_len, sizeof(guchar));
+                }
+
+              if (!ostree_sign_set_pk (sign, pk, error))
+                g_clear_error (error);
+            }
+
+
+          /* Set return to true if any sign fit */
           if (ostree_sign_commit_verify (sign,
                                          pull_data->repo,
                                          checksum,
                                          cancellable,
                                          error))
             ret = TRUE;
+          else
+            g_clear_error (error);
+
         }
-      g_strfreev(names);
-      if (ret == FALSE)
-        return FALSE;
+      if (!ret)
+        {
+          g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                               "Can't verify commit");
+          return FALSE;
+        }
     }
 
   /* If we found a legacy transaction flag, assume we have to scan.
@@ -3857,10 +3946,13 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
                                                         &pull_data->gpg_verify_summary, error))
           goto out;
 #endif /* OSTREE_DISABLE_GPGME */
-
-      /* TODO: read option for remote. */
+      /* Fetch verification settings from remote if it wasn't already
+       * explicitly set in the options. */
       if (!opt_sign_verify_set)
-        opt_sign_verify_set = TRUE;
+        if (!ostree_repo_get_remote_boolean_option (self, pull_data->remote_name,
+                                                    "sign-verify", TRUE,
+                                                    &pull_data->sign_verify, error))
+          goto out;
 
       /* NOTE: If changing this, see the matching implementation in
        * ostree-sysroot-upgrader.c
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1586,7 +1586,7 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
             }
 
           /* Set return to true if any sign fit */
-          if (ostree_sign_metadata_verify (sign,
+          if (ostree_sign_data_verify (sign,
                                            signed_data,
                                            signatures,
                                            &local_error
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -108,6 +108,7 @@ typedef struct {
   gboolean          gpg_verify;
   gboolean          gpg_verify_summary;
   gboolean          sign_verify;
+  gboolean          sign_verify_summary;
   gboolean          require_static_deltas;
   gboolean          disable_static_deltas;
   gboolean          has_tombstone_commits;
@@ -1564,6 +1565,51 @@ _load_public_keys (OtPullData *pull_data,
 }
 
 static gboolean
+_ostree_repo_sign_verify (OtPullData *pull_data,
+                          GBytes *signed_data,
+                          GVariant *metadata)
+{
+  /* list all signature types in detached metadata and check if signed by any? */
+  g_auto (GStrv) names = ostree_sign_list_names();
+  for (char **iter=names; iter && *iter; iter++)
+    {
+      g_autoptr (OstreeSign) sign = NULL;
+      g_autoptr (GVariant) signatures = NULL;
+      const gchar *signature_key = NULL;
+      GVariantType *signature_format = NULL;
+      g_autoptr (GError) local_error = NULL;
+
+      if ((sign = ostree_sign_get_by_name (*iter, &local_error)) == NULL)
+        continue;
+
+      signature_key = ostree_sign_metadata_key (sign);
+      signature_format = (GVariantType *) ostree_sign_metadata_format (sign);
+
+      signatures = g_variant_lookup_value (metadata,
+                                           signature_key,
+                                           signature_format);
+
+      /* If not found signatures for requested signature subsystem */
+      if (!signatures)
+        continue;
+
+      /* Try to load public key(s) according remote's configuration */
+      if (!_load_public_keys (pull_data, sign))
+        continue;
+
+      /* Return true if any signature fit to pre-loaded public keys.
+       * If no keys configured -- then system configuration will be used */
+      if (ostree_sign_data_verify (sign,
+                                   signed_data,
+                                   signatures,
+                                   &local_error))
+        return TRUE;
+    }
+
+  return FALSE;
+}
+
+static gboolean
 ostree_verify_unwritten_commit (OtPullData                 *pull_data,
                                 const char                 *checksum,
                                 GVariant                   *commit,
@@ -1572,21 +1618,24 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
                                 GCancellable               *cancellable,
                                 GError                    **error)
 {
+
+  if (pull_data->gpg_verify || pull_data->sign_verify)
+    /* Shouldn't happen, but see comment in process_verify_result() */
+    if (g_hash_table_contains (pull_data->verified_commits, checksum))
+      return TRUE;
+
+  g_autoptr(GBytes) signed_data = g_variant_get_data_as_bytes (commit);
+
 #ifndef OSTREE_DISABLE_GPGME
   if (pull_data->gpg_verify)
     {
       const char *keyring_remote = NULL;
-
-      /* Shouldn't happen, but see comment in process_verify_result() */
-      if (g_hash_table_contains (pull_data->verified_commits, checksum))
-        return TRUE;
 
       if (ref != NULL)
         keyring_remote = g_hash_table_lookup (pull_data->ref_keyring_map, ref);
       if (keyring_remote == NULL)
         keyring_remote = pull_data->remote_name;
 
-      g_autoptr(GBytes) signed_data = g_variant_get_data_as_bytes (commit);
       g_autoptr(OstreeGpgVerifyResult) result =
         _ostree_repo_gpg_verify_with_metadata (pull_data->repo, signed_data,
                                                detached_metadata,
@@ -1606,59 +1655,17 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
                                "Can't verify commit without detached metadata");
           return FALSE;
         }
-      /* Shouldn't happen, but see comment in process_verify_result() */
-      if (g_hash_table_contains (pull_data->verified_commits, checksum))
-        return TRUE;
 
-      gboolean ret = FALSE;
-      g_autoptr(GBytes) signed_data = g_variant_get_data_as_bytes (commit);
-
-      /* list all signature types in detached metadata and check if signed by any? */
-      g_auto (GStrv) names = ostree_sign_list_names();
-      for (guint i=0; i < g_strv_length (names); i++)
+      if (!_ostree_repo_sign_verify (pull_data, signed_data, detached_metadata))
         {
-          g_autoptr (OstreeSign) sign = NULL;
-          g_autoptr (GVariant) signatures = NULL;
-          const gchar *signature_key = NULL;
-          GVariantType *signature_format = NULL;
-          g_autoptr (GError) local_error = NULL;
-
-          if ((sign = ostree_sign_get_by_name (names[i], &local_error)) == NULL)
-            continue;
-
-          signature_key = ostree_sign_metadata_key (sign);
-          signature_format = (GVariantType *) ostree_sign_metadata_format (sign);
-
-          signatures = g_variant_lookup_value (detached_metadata,
-                                               signature_key,
-                                               signature_format);
-
-          /* If not found signatures for requested signature subsystem */
-          if (!signatures)
-            continue;
-
-          /* Try to load public key(s) according remote's configuration */
-          if (!_load_public_keys (pull_data, sign))
-            continue;
-
-          /* Return true if any signature fit to pre-loaded public keys.
-           * If no keys configured -- then system configuration will be used */
-          if (ostree_sign_data_verify (sign,
-                                       signed_data,
-                                       signatures,
-                                       &local_error))
-            ret = TRUE;
+          g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                               "Can't verify commit");
+          return FALSE;
         }
 
       /* Mark the commit as verified to avoid double verification
        * see process_verify_result () for rationale */
-      if (ret)
-        g_hash_table_add (pull_data->verified_commits, g_strdup (checksum));
-      else
-        g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                             "Can't verify commit");
-
-      return ret;
+      g_hash_table_add (pull_data->verified_commits, g_strdup (checksum));
     }
 
   return TRUE;
@@ -3773,6 +3780,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   gboolean opt_gpg_verify_set = FALSE;
   gboolean opt_gpg_verify_summary_set = FALSE;
   gboolean opt_sign_verify_set = FALSE;
+  gboolean opt_sign_verify_summary_set = FALSE;
   gboolean opt_collection_refs_set = FALSE;
   gboolean opt_n_network_retries_set = FALSE;
   gboolean opt_ref_keyring_map_set = FALSE;
@@ -3809,6 +3817,8 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
         g_variant_lookup (options, "gpg-verify-summary", "b", &pull_data->gpg_verify_summary);
       opt_sign_verify_set =
         g_variant_lookup (options, "sign-verify", "b", &pull_data->sign_verify);
+      opt_sign_verify_summary_set =
+        g_variant_lookup (options, "sign-verify-summary", "b", &pull_data->sign_verify_summary);
       (void) g_variant_lookup (options, "depth", "i", &pull_data->maxdepth);
       (void) g_variant_lookup (options, "disable-static-deltas", "b", &pull_data->disable_static_deltas);
       (void) g_variant_lookup (options, "require-static-deltas", "b", &pull_data->require_static_deltas);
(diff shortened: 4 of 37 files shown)


jlebon:
bot, add author to whitelist

cgwalters:
Only skimmed this so far, I think it is looking reasonable.

Let's split off the ability to disable gpgme as a separate PR first?

Did you consider my suggestion to support a "plugin" mechanism that's just fork/exec of an external process?  Like `/usr/lib/ostree/signing.d/` or so, we iterate over all of them and we have a few well-known return codes like `0=success`, `1=error`, `77=signature type not recognized`, `2=signed but failed to verify` - It's basically what gpgme is doing, just with a lot of additional API for setup, parsing error messages etc. 

The advantage is simplicity and libostree isn't linked against any additional libraries.  The downside is less integration.

But to be clear I'm fine with this approach as well.

I

d4s:
Added the first step in PR https://github.com/ostreedev/ostree/pull/1889 -- ability to build libostree without GPGME.

tests/libtest.sh
```diff
@@ -652,6 +652,16 @@ which_gpg () {
     echo ${gpg}
 }
 
+has_libsodium () {
+    local ret
+    ${CMD_PREFIX} ostree --version > version.txt
+    grep -q -e '- libsodium' version.txt
```
Thread974: You could save a temporary file by simply piping the output of ostree --version to the grep. (and thus remove the rm line)

d4s: yeah. I thought so initially. Unfortunately sometimes pipe fail for some reason, so it return negative value. In tests there function is used twice it leads to incorrect detection of tests which should be run.
I fall into that behavior while reworking `has_gpgme` function above.

src/ostree/ot-builtin-commit.c
```diff
@@ -817,12 +828,64 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
             goto out;
         }
 
-#ifndef OSTREE_DISABLE_GPGME
       if (opt_key_ids)
         {
+          /* Initialize crypto system */
+          if (!opt_sign_name)
+            opt_sign_name = "ed25519";
```
Thread974: This is not valid if !HAVE_LIBSODIUM

src/ostree/ot-builtin-commit.c
```diff
@@ -817,12 +828,64 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
...
+                      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                                   "Invalid KEY '%s'", keyid);
+
+                      goto out;
+                    }
+
+                  secret_key = g_variant_new_fixed_array (G_VARIANT_TYPE_BYTE, key, key_len, sizeof(guchar));
+                }
+#endif
+              if (!ostree_sign_set_sk (sign, secret_key, error))
```
Thread974: secret_key can be NULL here is the parameter was not `dummy` and not `ed22519`

src/ostree/ot-builtin-commit.c
d4s: mmm... `ostree_sign_get_by_name()` will not find it and return NULL, so exit from this part but yes, probably default should be "dummy" in that case, but not sure about the defaults.

src/ostree/ot-builtin-commit.c
d4s: yes, and should de handled by particular module implementation.

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,38 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
     }
 #endif /* OSTREE_DISABLE_GPGME */
 
+  if (pull_data->sign_verify)
+    {
+      gboolean ret = FALSE;
+      g_autoptr(GBytes) signed_data = g_variant_get_data_as_bytes (commit);
+      /* list all signature types in detached metadata and check if signed by any? */
+      GStrv names = ostree_sign_list_names();
```
Thread974: g_auto(GStrv) names;

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,38 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+      for (guint i=0; i < g_strv_length (names); i++)
+        {
+          g_autoptr (OstreeSign) sign = ostree_sign_get_by_name (names[i], error);
```
Thread974: unhandled error?

src/libostree/ostree-sign.c
```diff
@@ -0,0 +1,347 @@
...
+    return FALSE;
+
+  return TRUE;
+}
+
+GStrv ostree_sign_list_names(void)
+{
+  g_debug ("%s enter", __FUNCTION__);
+
+  GType types [] = {
```
Thread974: This array should be global. No need to define it in every func needing the names. It could also contain the name in the form a string to avoid allocating an object just to read the name.

src/ostree/ot-builtin-sign.c
```diff
@@ -0,0 +1,235 @@
...
+
+  for (ii = 0; ii < n_key_ids; ii++)
+    {
+      g_autoptr (GVariant) sk = NULL;
+      g_autoptr (GVariant) pk = NULL;
+      gsize key_len = 0;
+      g_autofree guchar *key = NULL;
+
+
+g_message("Read key: %s", key_ids[ii]);
```
Thread974: spurious g_message? or indent issue

src/ostree/ot-builtin-commit.c
```diff
@@ -817,12 +828,64 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
...
+
+              if (!g_strcmp0(ostree_sign_get_name (sign), "dummy"))
+                {
+                  secret_key = g_variant_new_string (keyid);
+                }
+#if defined(HAVE_LIBSODIUM)
+              else if (!g_strcmp0 (ostree_sign_get_name (sign), "ed25519"))
+                {
+                  key = g_malloc0 (crypto_sign_SECRETKEYBYTES);
+                  if (sodium_hex2bin (key, crypto_sign_SECRETKEYBYTES,
```
Thread974: This code is sodium specific, would it make sense to have it in the `OstreeSignEd25519` instead?

tests/libtest.sh
Thread974: what is ${CMD_PREFIX}  ? `sudo`?

src/libostree/ostree-repo-pull.c
d4s: Shouldn't be an error since we got only supported names from interface. but ok.

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,38 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+            continue;
+
+          if (ostree_sign_metadata_verify (sign,
+                                           signed_data,
+                                           signatures,
+                                           error
+                                          ))
+            ret = TRUE;
+        }
+      g_strfreev(names);
```
Thread974: Should this be removed now?

src/libostree/ostree-repo-pull.c
```diff
@@ -1827,6 +1862,28 @@ scan_commit_object (OtPullData                 *pull_data,
...
+          g_autoptr (OstreeSign) sign = ostree_sign_get_by_name (names[i], error);
+
+          if (ostree_sign_commit_verify (sign,
+                                         pull_data->repo,
+                                         checksum,
+                                         cancellable,
+                                         error))
+            ret = TRUE;
+        }
+      g_strfreev(names);
```
Thread974: Ditto

Makefile-libostree.am
```diff
@@ -260,6 +260,20 @@ libostree_1_la_CFLAGS += $(OT_DEP_SELINUX_CFLAGS)
 libostree_1_la_LIBADD += $(OT_DEP_SELINUX_LIBS)
 endif
 
+libostree_1_la_SOURCES += \
```
Thread974: Is this supposed to be build `--enable-experimental-api=no` ?

Makefile-libostree.am
```diff
@@ -260,6 +260,20 @@ libostree_1_la_CFLAGS += $(OT_DEP_SELINUX_CFLAGS)
...
+	src/libostree/ostree-sign.c \
+	src/libostree/ostree-sign.h \
+	src/libostree/ostree-sign-dummy.c \
+	src/libostree/ostree-sign-dummy.h \
+	src/libostree/ostree-sign-ed25519.c \
```
Thread974: Maybe this (and the .h) should be wrapped by `USE_LIBSODIUM`

Makefile-ostree.am
```diff
@@ -43,6 +43,7 @@ ostree_SOURCES = src/ostree/main.c \
 	src/ostree/ot-builtin-remote.c \
 	src/ostree/ot-builtin-reset.c \
 	src/ostree/ot-builtin-rev-parse.c \
+	src/ostree/ot-builtin-sign.c \
```
Thread974: Does not look very experimental either.

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+          g_autoptr (OstreeSign) sign = NULL;
+          g_autoptr (GVariant) signatures = NULL;
+          g_autofree gchar *signature_key = NULL;
+          g_autofree GVariantType *signature_format = NULL;
+          g_autofree gchar *pk_ascii = NULL;
+          g_autofree gchar *pk_file = NULL;
+
+          if ((sign = ostree_sign_get_by_name (names[i], error)) == NULL)
+          {
+              g_clear_error (error);
```
Thread974: This error will likely never happen, but maybe you could use a local `error` variable instead

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+              continue;
+          }
+          signature_key = ostree_sign_metadata_key (sign);
+          signature_format = (GVariantType *) ostree_sign_metadata_format (sign);
+
+          signatures = g_variant_lookup_value (detached_metadata,
+                                               signature_key,
+                                               signature_format);
+
```
Thread974: nitpick: extra blank line

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+            {
+              g_autoptr (GVariantBuilder) builder = NULL;
+              g_autoptr (GVariant) options = NULL;
+
+              builder = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
+              g_variant_builder_add (builder, "{sv}", "filename", g_variant_new_string (pk_file));
+              options = g_variant_builder_end (builder);
+
+              if (!ostree_sign_load_pk (sign, options, error))
+                g_clear_error (error);
```
Thread974: What happens if you have a `verification-file` but  `ostree_sign_load_pk` fails ?
Maybe accumulate errors?

src/libostree/ostree-sign-dummy.c
```diff
@@ -0,0 +1,187 @@
...
+{
+  g_debug ("%s enter", __FUNCTION__);
+
+  self->data = ostree_sign_dummy_data;
+  self->get_name = ostree_sign_dummy_get_name;
+  self->metadata_key = ostree_sign_dummy_metadata_key;
+  self->metadata_format = ostree_sign_dummy_metadata_format;
+  self->metadata_verify = ostree_sign_dummy_metadata_verify;
+  self->set_sk = ostree_sign_dummy_set_signature;
+  self->set_pk = ostree_sign_dummy_set_signature;
```
Thread974: Are `set_pk` and `set_sk` really both pointing to `ostree_sign_dummy_set_signature` ?

src/libostree/ostree-sign-dummy.c
```diff
@@ -0,0 +1,187 @@
...
+
+  return TRUE;
+}
+
+gchar * ostree_sign_dummy_get_name (OstreeSign *self)
+{
+  g_debug ("%s enter", __FUNCTION__);
+  g_return_val_if_fail (OSTREE_IS_SIGN (self), FALSE);
+
+  g_autofree gchar *name = g_strdup(OSTREE_SIGN_DUMMY_NAME);
```
Thread974: You are leaking `name` since `ostree_sign_get_name` returns a `const char *`.
Just `return OSTREE_SIGN_DUMMY_NAME` is enough here.

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+err:
+  return FALSE;
+}
+
+gchar * ostree_sign_ed25519_get_name (OstreeSign *self)
+{
+  g_debug ("%s enter", __FUNCTION__);
+  g_return_val_if_fail (OSTREE_IS_SIGN (self), FALSE);
+
+  g_autofree gchar *name = g_strdup (OSTREE_SIGN_ED25519_NAME);
```
Thread974: You are leaking `name` since `ostree_sign_get_name` returns a `const char *`.

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+          /* Override key if it is set explicitly */
+          ostree_repo_get_remote_option (pull_data->repo,
+                                         pull_data->remote_name,
+                                         "verification-key", NULL,
+                                         &pk_ascii, NULL);
+          if (pk_ascii != NULL)
+            {
+              g_autoptr (GVariant) pk = NULL;
+
+              if (!g_strcmp0(ostree_sign_get_name(sign), "dummy"))
```
Thread974: Leaking the name here (see ostree_sign_dummy_get_name)
Actually all the calls to `ostree_sign_get_name` leak

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+                {
+                  // Just use the string as signature
+                  pk = g_variant_new_string(pk_ascii);
+                }
+              else if (!g_strcmp0(ostree_sign_get_name(sign), "ed25519"))
```
Thread974: Leaking the name here too

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+  g_autofree gchar *key = g_strdup(OSTREE_SIGN_METADATA_ED25519_KEY);
+  return g_steal_pointer (&key);
+}
+
+gchar * ostree_sign_ed25519_metadata_format (OstreeSign *self)
+{
+  g_debug ("%s enter", __FUNCTION__);
+
+  g_autofree gchar *type = g_strdup (OSTREE_SIGN_METADATA_ED25519_TYPE);
+  return g_steal_pointer (&type);
```
Thread974: This one is freed, but you could as well just do : 
`return OSTREE_SIGN_METADATA_ED25519_TYPE` and return a `const char*`

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+
+  return g_steal_pointer (&name);
+}
+
+gchar * ostree_sign_ed25519_metadata_key (OstreeSign *self)
+{
+  g_debug ("%s enter", __FUNCTION__);
+
+  g_autofree gchar *key = g_strdup(OSTREE_SIGN_METADATA_ED25519_KEY);
+  return g_steal_pointer (&key);
```
Thread974: This one is freed, but you could as well just do : 
`return OSTREE_SIGN_METADATA_ED25519_KEY` and return a `const char*`

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+
+          /* Set return to true if any sign fit */
+          if (ostree_sign_metadata_verify (sign,
+                                           signed_data,
+                                           signatures,
+                                           error
+                                          ))
+            ret = TRUE;
+          else
+            g_clear_error (error);
```
Thread974: Rather than clearing the error, you may consider accumulating the inner errors.
That could give useful informations.

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+                }
+              else if (!g_strcmp0(ostree_sign_get_name(sign), "ed25519"))
+                {
+                  gsize key_len = 0;
+                  g_autofree guchar *key = g_base64_decode (pk_ascii, &key_len);
+                  pk = g_variant_new_fixed_array (G_VARIANT_TYPE_BYTE, key, key_len, sizeof(guchar));
+                }
+
+              if (!ostree_sign_set_pk (sign, pk, error))
+                g_clear_error (error);
```
Thread974: Maybe accumulate errors?

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+                                          ))
+            ret = TRUE;
+          else
+            g_clear_error (error);
+        }
+
+      /* Mark the commit as verified to avoid double verification
+       * see process_verify_result () for rationale */
+      if (ret)
+        {
```
Thread974: nitpick: extra `{`

src/libostree/ostree-repo-pull.c
```diff
@@ -1498,6 +1501,114 @@ ostree_verify_unwritten_commit (OtPullData                 *pull_data,
...
+        g_hash_table_add (pull_data->verified_commits, g_strdup (checksum));
+        }
+      else
+        g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
```
Thread974: Maybe use g_prefix_error here?

src/libostree/ostree-repo-pull.c
```diff
@@ -1827,6 +1938,86 @@ scan_commit_object (OtPullData                 *pull_data,
     }
 #endif /* OSTREE_DISABLE_GPGME */
 
+  if (pull_data->sign_verify &&
+      !g_hash_table_contains (pull_data->verified_commits, checksum))
+    {
+      gboolean ret = FALSE;
+      /* list all signature types in detached metadata and check if signed by any? */
+      g_auto (GStrv) names = ostree_sign_list_names();
+      for (guint i=0; i < g_strv_length (names); i++)
```
Thread974: The strv ends with a NULL, so you can do : 
`for (guint i=0; names[i] != NULL; i++)`
That save some calls to g_strv_length()
There are other places to do that too.

src/libostree/ostree-repo-pull.c
```diff
@@ -1827,6 +1938,86 @@ scan_commit_object (OtPullData                 *pull_data,
...
+                                         "verification-key", NULL,
+                                         &pk_ascii, NULL);
+          if (pk_ascii != NULL)
+            {
+              g_autoptr (GVariant) pk = NULL;
+
+              if (!g_strcmp0(ostree_sign_get_name(sign), "dummy"))
+                {
+                  // Just use the string as signature
+                  pk = g_variant_new_string(pk_ascii);
```
Thread974: This code should be somewhere in dummy implementation.
We have already allocated an object for the current sign.
Just needs a new method for this processing.

src/libostree/ostree-repo-pull.c
```diff
@@ -1827,6 +1938,86 @@ scan_commit_object (OtPullData                 *pull_data,
...
+                }
+              else if (!g_strcmp0(ostree_sign_get_name(sign), "ed25519"))
+                {
+                  gsize key_len = 0;
+                  g_autofree guchar *key = g_base64_decode (pk_ascii, &key_len);
+                  pk = g_variant_new_fixed_array (G_VARIANT_TYPE_BYTE, key, key_len, sizeof(guchar));
```
Thread974: This code should be somewhere in ed25519

src/libostree/ostree-sign-dummy.c
```diff
@@ -0,0 +1,187 @@
...
+
+  return g_steal_pointer (&name);
+}
+
+gchar * ostree_sign_dummy_metadata_key (OstreeSign *self)
+{
+  g_debug ("%s enter", __FUNCTION__);
+
+  g_autofree gchar *key = g_strdup(OSTREE_SIGN_METADATA_DUMMY_KEY);
+  return g_steal_pointer (&key);
```
Thread974: `return OSTREE_SIGN_METADATA_DUMMY_KEY`

src/libostree/ostree-sign-dummy.c
```diff
@@ -0,0 +1,187 @@
...
+}
+
+gchar * ostree_sign_dummy_metadata_format (OstreeSign *self)
+{
+  g_debug ("%s enter", __FUNCTION__);
+
+  g_autofree gchar *type = g_strdup(OSTREE_SIGN_METADATA_DUMMY_TYPE);
```
Thread974: `return OSTREE_SIGN_METADATA_DUMMY_TYPE`

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+
+#undef G_LOG_DOMAIN
+#define G_LOG_DOMAIN "OSTreeSign"
+
+#define OSTREE_SIGN_ED25519_NAME "ed25519"
+
+#define OSTREE_SIGN_METADATA_ED25519_KEY "ostree.sign.ed25519"
+#define OSTREE_SIGN_METADATA_ED25519_TYPE "aay"
+
+#if 0
```
Thread974: Please remove this kind of code if it is no longer in use

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+  g_autofree guchar *sig = NULL;
+#endif
+
+  if ((sign->initialized != TRUE) || (sign->secret_key == NULL))
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                          "Not able to sign: libsodium library isn't initialized properly");
+      goto err;
+    }
+#ifdef HAVE_LIBSODIUM
```
Thread974: Maybe this whole object should not be built if libsodium is not available.
Otherwise the code is cluttered with `#ifdef HAVE_LIBSODIUM`. In the long term, you'd need to build the code twice if you wants to be sure it builds.

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                          "Not able to sign the object");
+      goto err;
+    }
+
+  *signature = g_bytes_new (sig, sig_size);
+  return TRUE;
+#endif /* HAVE_LIBSODIUM */
+err:
```
Thread974: This label is useless

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+                            g_bytes_get_data (data, NULL),
+                            g_bytes_get_size (data),
+                            sign->secret_key))
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                          "Not able to sign the object");
+      goto err;
+    }
+
+  *signature = g_bytes_new (sig, sig_size);
```
Thread974: Consider using [g_bytes_new_take](https://developer.gnome.org/glib/stable/glib-Byte-Arrays.html#g-bytes-new-take) to avoid coyping the signature

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+        }
+    }
+
+  if (ret != TRUE)
+    g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                         "Not able to verify: no valid signatures found");
+#endif /* HAVE_LIBSODIUM */
+
+  return ret;
+err:
```
Thread974: Useless label again?

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+      goto err;
+    }
+
+  *out_secret_key = g_variant_new_fixed_array (G_VARIANT_TYPE_BYTE, sk, crypto_sign_SECRETKEYBYTES, sizeof(guchar));
+  *out_public_key = g_variant_new_fixed_array (G_VARIANT_TYPE_BYTE, pk, crypto_sign_PUBLICKEYBYTES, sizeof(guchar));
+
+  return TRUE;
+#endif /* HAVE_LIBSODIUM */
+
+err:
```
Thread974: Useless label

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+#ifdef HAVE_LIBSODIUM
+  OstreeSignEd25519 *sign = ostree_sign_ed25519_get_instance_private(OSTREE_SIGN_ED25519(self));
+  g_autofree char * hex = NULL;
+  gpointer key = NULL; 
+
+  gsize n_elements = 0;
+  key = (gpointer) g_variant_get_fixed_array (public_key, &n_elements, sizeof(guchar));
+
+  hex = g_malloc0 (crypto_sign_PUBLICKEYBYTES*2 + 1);
+  g_debug ("Read ed25519 public key = %s", sodium_bin2hex (hex, crypto_sign_PUBLICKEYBYTES*2+1, key, n_elements));
```
Thread974: This is a bit weird:  using sodium_bin2hex as part of g_debug.
What happens is your build is in release mode?

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+
+
+static gboolean
+_load_pk_from_stream (OstreeSign *self, GDataInputStream *key_data_in, GError **error)
+{
+  g_return_val_if_fail (key_data_in, FALSE);
+#ifdef HAVE_LIBSODIUM
+  gboolean ret = FALSE;
+
+#if 0
```
Thread974: Don't put that in the pull request

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,588 @@
...
+  g_debug ("Read ed25519 public key = %s", sodium_bin2hex (hex, crypto_sign_PUBLICKEYBYTES*2+1, key, n_elements));
+
+  if (n_elements != crypto_sign_PUBLICKEYBYTES)
+    {
+      g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                           "Incorrect ed25519 public key");
+      goto err;
+    }
+
+  key = g_memdup (key, n_elements);
```
Thread974: This construct is weird too :  key from g_variant_get_fixed_array and key from g_memdup, contains the same data, but they don't have same lifetime, so using the same variable to represent both is error prone imho.

And if the key is already in `sign->public_keys`, it is duplicated but not freed. So the `g_memdup` should be inside the if.

src/libostree/ostree-sign-ed25519.c
Thread974: what is `hex` for? it is not used further in this function???

Makefile-libostree.am
d4s: No, they must exists. Functions from `*ed25519.c` ruturn error "Have no libsodium support" if build without it.

Makefile-libostree.am
d4s: Yes

src/libostree/ostree-repo-pull.c
d4s: good, thanks.

src/libostree/ostree-repo-pull.c
d4s: Not sure about that. What if `verification-file` and `verification-key` set?
With current logic I just return "Verification failed" and it is possible to trace the exact problem with signatures by setting environment variable `G_MESSAGES_DEBUG`.
tbh I prefer to leave it as is.

src/libostree/ostree-repo-pull.c
d4s: see comment above.

src/libostree/ostree-repo-pull.c
d4s: prefer the current way -- looks a bit clearer for me with minimal overhead for.
And `g_strv_length()` is used widely in other places, so do not see why need to use other approach.

src/libostree/ostree-sign-dummy.c
d4s: Yes ;)

src/libostree/ostree-sign-ed25519.c
d4s: At least 2 jumps above.

src/libostree/ostree-sign-ed25519.c
d4s: 2 jumps above -- prefer to exit in that way.

src/libostree/ostree-sign-ed25519.c
d4s: sodium_bin2hex() function converts bytes into a hexadecimal string.

src/libostree/ostree-sign-ed25519.c
d4s: Nice catch!

d4s:
@cgwalters @pwithnall I would be very appreciate for early review of the current state -- just to know if there is any core issues or suggestions.

The support of "summary" file is ongoing.

/assign @pwithnall

pwithnall:
I am not the right person to review this --- I have not been involved in #1233 at all (or OSTree development in the recent past) and don’t have time to work out what’s going on here, sorry.

d4s:
Oh! sorry for annoying -- the bot suggested you as reviewer for some reason.

/cc @cgwalters

man/ostree-commit.xml
```diff
@@ -251,6 +251,39 @@ Boston, MA 02111-1307, USA.
...
             </varlistentry>
+
+            <varlistentry>
+                <term><option>-s, --sign-type</option></term>
+                <listitem><para>
+                    Use particular signature engine. Currently
+                    available <arg choice="plain">ed25519</arg> and <arg choice="plain">dummy</arg>
+                    signature types.
+
+                    The dafault is <arg choice="plain">ed25519</arg>.
```
fdanis-oss: _default_ instead of _dafault_

man/ostree-sign.xml
```diff
@@ -0,0 +1,152 @@
...
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><option>-s, --sign-type</option></term>
+                <listitem><para>
+                    Use particular signature mechanism. Currently
+                    available <arg choice="plain">ed25519</arg> and <arg choice="plain">dummy</arg>
+                    signature types.
+
+                    The dafault is <arg choice="plain">ed25519</arg>.
```
fdanis-oss: _default_ instead of _dafault_

tests/libtest.sh
d4s: nope. defined in lib file for shell.

man/ostree-sign.xml
d4s: thank you!

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,698 @@
...
+{
+  g_debug ("%s enter", __FUNCTION__);
+
+  self->initialized = TRUE;
+  self->secret_key = NULL;
+  self->public_keys = NULL;
+  self->revoked_keys = NULL;
+
+#ifdef HAVE_LIBSODIUM
+  if (sodium_init() < 0)
```
fdanis-oss:
```suggestion
  if (sodium_init() < 0)
```

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,698 @@
...
+              continue;
+            }
+
+          if (crypto_sign_verify_detached ((guchar *) g_variant_get_data (child),
+                                           g_bytes_get_data (data, NULL),
+                                           g_bytes_get_size (data),
+                                           public_key->data) != 0)
+            {
+              /* Incorrect signature! */
+              g_debug("Signature couldn't be verified with key '%s'",
```
fdanis-oss:
```suggestion
              g_debug("Signature couldn't be verified with key '%s'",
```

man/ostree-sign.xml
```diff
@@ -0,0 +1,152 @@
...
+                    Valid for <literal>ed25519</literal> signature type.
+                    For <literal>ed25519</literal> this file must contain <literal>base64</literal>-encoded
+                    secret key(s) (for signing) or public key(s) (for verifying) per line.
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><option>--keys-dir</option></term>
+                <listitem><para>
+                    Redefine the system path, where to search files and subdirectories with
+                    well-known and revoked keys.
```
fdanis-oss:
```suggestion
                    well-known and revoked keys.
```

d4s:
@cgwalters is there any chance for review?
Now I think that the core is complete.

I understand that this PR is pretty big, so I'll be very-very-very appreciate for a quick check if the approach and proposed interface is good enough for the `libostree` as first step.

/assign @cgwalters

src/libostree/ostree-repo-pull.c
cgwalters: The compiler can't know that it will be constant during the loop, so we'll evaluate it each time.  This doesn't matter too much *here* but having apparently `O(N)` code actually be `O(N²)` can be quite surprising if code is moved into an even mildly performance sensitive context.

Best to use an iterator, which we do elsewhere:

```
for (char ** iter = names; iter && *iter; iter++)
{
  const char *name = *iter;
  ...
}
```

src/libostree/ostree-repo-pull.c
```diff
@@ -1829,6 +2002,41 @@ scan_commit_object (OtPullData                 *pull_data,
...
+                                         pull_data->repo,
+                                         checksum,
+                                         cancellable,
+                                         &local_error))
+            ret = TRUE;
+
+        }
+      if (!ret)
+        {
+          g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
```
cgwalters: This is ignoring `local_error`; the message there would likely be very useful right?  We could gather all errors into an array and stringify all of them, or probably simplest just retain the last one.

src/libostree/ostree-repo-pull.c
```diff
@@ -1466,6 +1470,150 @@ process_verify_result (OtPullData            *pull_data,
...
+
+      builder = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
+      g_variant_builder_add (builder, "{sv}", "filename", g_variant_new_string (pk_file));
+      options = g_variant_builder_end (builder);
+
+      if (ostree_sign_load_pk (sign, options, &error))
+        loaded_from_file = TRUE;
+      else
+        {
+          if (error == NULL)
```
cgwalters: This is a programming error if it happens; i.e. we should `g_assert (error)` instead.  The GError convention is that if a function returns a failure value (for the usual booleans, `FALSE`), the `GError **error` parameter *must* be set.  

A similar common pattern is used in e.g. Go - the usual assumption is that any function returning `(somevalue, error)` would have one or the other set.  (There are use cases for returning `nil, nil` in Go, but it's unusual)

Rust of course has a much better story with enums and `Result`.

apidoc/ostree-sections.txt
```diff
@@ -694,3 +694,27 @@ ostree_kernel_args_from_string
...
+ostree_sign_get_name
+ostree_sign_add_pk
+ostree_sign_clear_keys
+ostree_sign_load_pk
+ostree_sign_set_pk
+ostree_sign_set_sk
+ostree_sign_summary
+<SUBSECTION Standard>
+ostree_sign_get_type
+ostree_sign_dummy_get_type
```
cgwalters: Is this one just for our internal testing?  Seems odd to make it public API.

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,698 @@
...
+  self->initialized = TRUE;
+  self->secret_key = NULL;
+  self->public_keys = NULL;
+  self->revoked_keys = NULL;
+
+#ifdef HAVE_LIBSODIUM
+  if (sodium_init() < 0) 
+    {
+      self->initialized = FALSE;
+      g_warning ("libsodium library couldn't be initialized");
```
cgwalters: I tend to avoid `g_warning()` in libraries.  Rather let's do something like:
```
self->initialization_failed = TRUE;
```
And then on the first API call that allows us to return `GError`, we do so.

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,698 @@
...
+
+  g_debug ("%s enter", __FUNCTION__);
+  g_return_val_if_fail (OSTREE_IS_SIGN (self), FALSE);
+  OstreeSignEd25519 *sign = ostree_sign_ed25519_get_instance_private(OSTREE_SIGN_ED25519(self));
+
+#ifdef HAVE_LIBSODIUM
+  guchar *sig = NULL;
+#endif
+
+  if ((sign->initialized != TRUE) || (sign->secret_key == NULL))
```
cgwalters: For example:

```
if (sign->initialization_failed)
  return glnx_throw (error, "Failed to initialize libsodium");
```

cgwalters:
High level, I'm willing to merge this this development cycle - I need a bit more time to work through the details and logic of the signing code, but we don't need to get this perfect before merging.  

I'm happy to help do some post-merge style fixups and such.

One thing I want to understand is - do we see this as something just for "embedded" OSTree users who are making custom OSes and don't want the GPG dep, or do we see this as something that might be used by general purpose public operating systems and app systems - e.g. do we see Flathub adding ed25519 signing in addition to GPG?

Personally I'm uncertain.  There are some advantages of not having the GPG dependency - the way libgpgme forks off the agent helper has been a source of pain, and like most projects using GPG for signing we're not using any of the "web of trust" aspect that makes up so much of GPG.  But there are also some disadvantages; at least for Fedora our hardware signing infrastructure is oriented around GPG, but I don't know how difficult it would be to extend.  Plus general familiarity.

em-:
First and foremost, @cgwalters thank you for the feedback!

I'm replying on behalf of @d4s on where we see this fit in the general scheme of things. 

Our very immediate goal is "embedded", specially so if you define "embedded" as "places where deploying GPL-3 components on the field is problematic" for any reason. :)

That said, we'd be more than happy to see more adoption in general in the long term, as GPG has its on large set of issues not really related to its license. In the end we're doing something akin to what the OpenBSD team already did with their artifacts, so there are compelling reasons to not consider this "embedded-only".

d4s:
> One thing I want to understand is - do we see this as something just for "embedded" OSTree users who are making custom OSes and don't want the GPG dep, or do we see this as something that might be used by general purpose public operating systems and app systems - e.g. do we see Flathub adding ed25519 signing in addition to GPG?

The initial goal was to have signature mechanism for "embedded" users, however I tried to keep sign interface usable for general purposes. Depending on build-time configuration options it is possible to use GPG, ed25519 or both engines in the same time. tbh I'm not so familiar with FlatHub, but I expect it would be not a hard task to add ed25519 signature to it.

Please keep in mind -- current implementation include the common interface and 2 signature engines: ed25519 and simple 'dummy' engine (just an ASCII string as a key). I expect it would be possible to add other signing engines without touching the libostree core.

For commits signing it is possible to use several signature types in the same time, including GPG. To achieve the same for the summary file we need to change at least CLI and probably a bit of internals, so far only single signature type is supported by CLI for summary at the moment.

> Personally I'm uncertain. There are some advantages of not having the GPG dependency - the way libgpgme forks off the agent helper has been a source of pain, and like most projects using GPG for signing we're not using any of the "web of trust" aspect that makes up so much of GPG. But there are also some disadvantages; at least for Fedora our hardware signing infrastructure is oriented around GPG, but I don't know how difficult it would be to extend. Plus general familiarity.

That is why I didn't touch the current GPG implementation -- just add the ability to build without GPG if needed. Probably the current GPG-related code could be dropped sometime if somebody implement the GPG engine for the proposed interface -- but I expect  a lot of additional efforts here. Until that the GPG and new signing interface should co-exist imho.

src/libostree/ostree-repo-pull.c
d4s: Just because the amount of elements _here_ is not expected to be big for years, I thought to leave it as is.
But now carefully re-read that code -- agree, a bit stupid thing I wrote here. Fixed.

apidoc/ostree-sections.txt
d4s: Thank you!

src/libostree/ostree-repo-pull.c
d4s: Reworked error handling -- now the real reason(s) from signature engines are collected instead of just set the common error.

d4s:
@cgwalters I hope all noticed issues has been fixed.
Sorry for the forced update.

d4s:
@cgwalters just a friendly reminder ;)

cgwalters:
OK yep!  Thanks for the ping, I do want to get this in. 
/approve
I think I should be able to review this week.

cgwalters:
@d4s OK I pushed two small commits on top of this so far (please look and let me know if you're OK with them)

cgwalters:
I know the `glnx_throw` etc. isn't quite idomatic glib for other projects, but I think libostree has heavily benefited from `__attribute__((cleanup))` (I can't imagine writing C without it anymore) - so that's the reason for the current style.

src/libostree/ostree-repo-pull.c
```diff
@@ -1466,6 +1470,167 @@ process_verify_result (OtPullData            *pull_data,
...
+
+      builder = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
+      g_variant_builder_add (builder, "{sv}", "filename", g_variant_new_string (pk_file));
+      options = g_variant_builder_end (builder);
+
+      if (ostree_sign_load_pk (sign, options, error))
+        loaded_from_file = TRUE;
+      else
+          g_debug("Unable to load public keys for '%s' from file '%s': %s",
+                  ostree_sign_get_name(sign), pk_file, (*error)->message);
```
cgwalters: Using `GError` correctly requires strict discipline.  This bit is wrong because it should be supported to pass `NULL` for the error to mean "ignore errors".  And then we'd deref a `NULL` pointer here.

This bit should use a `local_error` too as the bit below does.

But...I also don't understand at a higher level why we aren't just erroring out here if loading the key fails?

src/libostree/ostree-repo-pull.c
```diff
@@ -1466,6 +1470,167 @@ process_verify_result (OtPullData            *pull_data,
...
+      if (_load_public_keys (sign, repo, remote_name, &local_error))
+        {
+          /* Return true if any signature fit to pre-loaded public keys.
+           * If no keys configured -- then system configuration will be used */
+          if (ostree_sign_data_verify (sign,
+                                       signed_data,
+                                       signatures,
+                                       &local_error))
+            {
+              g_clear_error (error);
```
cgwalters: Similarly here I think it's best if we hoist the `local_error` out of the loop here; call it something like `last_error`, and the loop shouldn't be touching `error`.

At the end of the loop we just propagate the error if one is set, or return the "no supported signed type".

src/libostree/ostree-repo-pull.c
```diff
@@ -4168,6 +4425,70 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
+                                                                     "summary",
+                                                                     OSTREE_FETCHER_REQUEST_OPTIONAL_CONTENT,
+                                                                     pull_data->n_network_retries,
+                                                                     &bytes_summary,
+                                                                     OSTREE_MAX_METADATA_SIZE,
+                                                                     cancellable, error))
+                      goto out;
+
+                    if (_ostree_repo_sign_verify (pull_data->repo, pull_data->remote_name, bytes_summary, signatures, error))
+                      ret = TRUE;
```
cgwalters: This should look more like:
```
if (!sign_verify (...))
  {
     glnx_prefix_error (error, "Verifying summary")
     goto out;
  }
```

or so, since there's no `temp_error` involved, right?

src/libostree/ostree-sign-dummy.c
```diff
@@ -0,0 +1,205 @@
...
+static void
+_ostree_sign_dummy_class_init (OstreeSignDummyClass *self)
+{
+  g_debug ("%s enter", __FUNCTION__);
+}
+
+static void
+_ostree_sign_dummy_init (OstreeSignDummy *self)
+{
+  g_debug ("%s enter", __FUNCTION__);
```
cgwalters: Let's remove all of these?  We're not using them elsewhere in the code.

src/libostree/ostree-sign-dummy.c
```diff
@@ -0,0 +1,205 @@
...
+  self->pk_ascii = NULL;
+}
+
+gboolean ostree_sign_dummy_set_sk (OstreeSign *self, GVariant *key, GError **error)
+{
+  g_debug ("%s enter", __FUNCTION__);
+
+  OstreeSignDummy *sign =  _ostree_sign_dummy_get_instance_private(OSTREE_SIGN_DUMMY(self));
+
+  if (sign->sk_ascii != NULL)
```
cgwalters: This bit is unnecessary, `g_free()` is happy to ignore `NULL`.

src/libostree/ostree-sign-dummy.c
```diff
@@ -0,0 +1,205 @@
...
+
+  for (gsize i = 0; i < g_variant_n_children(signatures); i++)
+    {
+      g_autoptr (GVariant) child = g_variant_get_child_value (signatures, i);
+      g_autoptr (GBytes) signature = g_variant_get_data_as_bytes(child);
+
+      gsize sign_size = 0;
+      g_bytes_get_data (signature, &sign_size);
+      g_autofree gchar *sign_ascii = g_strndup(g_bytes_get_data (signature, NULL), sign_size);
+      g_debug("Read signature %d: %s", (gint)i, sign_ascii);
```
cgwalters: This isn't assured to be `NUL` terminated, can't pass it to `%s` right?

src/libostree/ostree-sign-dummy.c
```diff
@@ -0,0 +1,205 @@
...
+      g_debug("Stored signature %d: %s", (gint)i, sign->pk_ascii);
+
+      if (!g_strcmp0(sign_ascii, sign->pk_ascii))
+          ret = TRUE;
```
cgwalters: Why not just `return TRUE` here?

src/libostree/ostree-sign-ed25519.c
cgwalters: We do have CI that builds in multiple configurations at least.

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,714 @@
...
+    return glnx_throw (error, "Unknown ed25519 revoked key type");
+
+  OstreeSignEd25519 *sign = _ostree_sign_ed25519_get_instance_private(OSTREE_SIGN_ED25519(self));
+
+  const gchar *rk_ascii = g_variant_get_string (revoked_key, NULL);
+  gsize n_elements = 0;
+  gpointer key = g_base64_decode (rk_ascii, &n_elements);
+
+  g_autofree char * hex = g_malloc0 (crypto_sign_PUBLICKEYBYTES*2 + 1);
+  g_debug ("Read ed25519 revoked key = %s", sodium_bin2hex (hex, crypto_sign_PUBLICKEYBYTES*2+1, key, n_elements));
```
cgwalters: `sodium_bin2hex` calls `malloc()` right?  So we need to do `g_autofree char *hex = sodium_bin2hex(...); g_debug ("%s", hex)`.

src/libostree/ostree-sign-ed25519.c
```diff
@@ -0,0 +1,714 @@
...
+  /* Use simple file format with just a list of base64 public keys per line */
+  while (TRUE)
+    {
+      gsize len = 0;
+      g_autofree char *line = g_data_input_stream_read_line (key_data_in, &len, NULL, error);
+      g_autoptr (GVariant) pk = NULL;
+      gboolean added = FALSE;
+
+      if (*error != NULL)
+        goto err;
```
cgwalters: I'm finding it hard to get over the error handling style - I know e.g. the Linux kernel makes heavy use of `goto` but again in all of our new code we try to avoid it.

Actually this case needs to be:

```
if (!line)
  return FALSE;
```

Because it's supported again to pass `NULL` for `GError`.

src/libostree/ostree-sign-ed25519.c
cgwalters: (Nevermind about this one, read the docs)

d4s:
First of all -- thank you for review, I understand it's not a trivial task to review XXXL requests.

> @d4s OK I pushed two small commits on top of this so far (please look and let me know if you're OK with them)

I am absolutely ok with them.

Do I need to rework the rest of the PR to follow the style? I mean the usage of `return` instead of `goto ...` and `glnx_*` for error handling

src/libostree/ostree-repo-pull.c
d4s: Thanks! The idea of that behavior is to give a chance for the key added for the remote with `verification-key` option as ascii string, please see the check below of `pk_ascii`.

cgwalters:
> Do I need to rework the rest of the PR to follow the style?

I'm happy to pitch in and do some of it too!  It will also help me get familiar with the code.  I'll do a bit of that today.

src/libostree/ostree-sign-dummy.c
d4s: actually no.
this is from `g_strndup()` description:
>  * Duplicates the first @ n bytes of a string, returning a newly-allocated
>  * buffer @ n + 1 bytes long which will always be nul-terminated.

src/libostree/ostree-sign-ed25519.c
d4s: correct, I think it should be the separate MR, since the current one is too huge already

src/libostree/ostree-sign-dummy.c
d4s: Thanks for removing ;-)

src/libostree/ostree-sign-dummy.c
d4s: my bad. Indeed, with a new style such kind of optimizations are more visible. Thanks!

src/libostree/ostree-repo-pull.c
d4s: Thank you. Indeed

d4s:
@cgwalters sorry for the delay, hope I address all your comments in a proper way.
Added a couple of minor fixes in addition.

d4s:
@cgwalters just a friendly reminder :)
Hope all your comments are fixed and suggestions applied.

d4s:
hmmm... @cgwalters, just a reminder about this PR ;)

cgwalters:
/test sanity

OK at this point I think we can commit to getting this in the next release, there will likely be followup work around this, but we can do it in master.

/lgtm

cgwalters:
That's weird, I don't know what's going on with the `do-not-merge/invalid-owners-file` label.

cgwalters:
@d4s I submitted a followup PR here https://github.com/ostreedev/ostree/pull/2058 - mind reviewing?

Issue #959, opened 2017-06-25
ability to repair a repo


untoreh:
would it be possible something like
`ostree pull --repair remote branch`
that fschecks the repository and pulls in corrupted objects?

cgwalters:
See https://github.com/ostreedev/ostree/pull/345

cgwalters:
A user hit a case where the current and previous commits were missing objects; `ostree fsck` naturally marked them both partial.  Then doing e.g. `ostree admin upgrade` will pull the missing objects for the *current* commit but not previous.  The problem is the prune bombs out looking for missing objects from the previous commit.

We do need to make it ergonomic to pull missing objects for all commits.  Further, I think when we're going to delete a commit anyways we can just ignore missing objects.

v217:
I changed a file in the readonly `/usr` directory (e.g. `/usr/bin/gnome-session`).
Then I ran `ostree fsck`.
And `ostree fsck` correctly indicated me that the filesystem has been tampered with.
Is there any way to repair the system without reinstalling fedora silverblue from scratch.
(Rollback did not work and neither the link to https://github.com/ostreedev/ostree/pull/345 was helpful)

v217:
`rpm-ostree reset` did not work.

The best advice, I could find was in this blog post: https://blog.duti.me/blog/breaking-silverblue/
 

```
sudo ostree pull fedora:fedora/39/x86_64/testing/silverblue
sudo ostree checkout fedora:fedora/39/x86_64/testing/silverblue
sudo ostree pull fedora:fedora/39/x86_64/silverblue
sudo ostree checkout fedora:fedora/39/x86_64/silverblue
rpm-ostree deploy  <latest commit hashsum>
```

Probably there is an easier way to check, if files have been modified and in this case download the unmodified files, and deploy them? Any feedback is very welcome! Thanks

v217:
I have not tried that yet, because I could not find a lot of documentation:
¿`ostree pull --subpath`  to just download the files where bit flips happened?

Anyway, I think this is useful documentation, so people not too familiar with ostree and how it mimics git, can repair their deployed filesystem in the exceptional case of bit flips.

PR #1880, opened 2019-06-26
fsck: Throw and error and return non-zero for non-verified commits

jwessel:
After the corruption has been fixed with "ostree fsck -a --delete", a
second run of the "ostree fsck" command will print X partial commits
not verified and exit with a zero.

The zero exit code makes it hard to detect if a repair operation needs
to be run, it is easier to check an exit code than try to parse the
text that is emitted from the ostree fsck.

Signed-off-by: Jason Wessel <jason.wessel@windriver.com>

### DIFF
--- a/src/ostree/ot-builtin-fsck.c
+++ b/src/ostree/ot-builtin-fsck.c
@@ -444,7 +444,7 @@ ostree_builtin_fsck (int argc, char **argv, OstreeCommandInvocation *invocation,
     }
   else if (n_partial > 0)
     {
-      g_print ("%u partial commits not verified\n", n_partial);
+      return glnx_throw (error, "%u partial commits not verified", n_partial);
     }
 
   if (found_corruption)


cgwalters:
bot, add author to whitelist

cgwalters:
I see where you're coming from; the current behavior is confusing.  However I don't think we can change it as it's really not an error to have partial commits.

For example, it's convenient to `ostree pull --commit-metadata-only --depth=50` or so to fetch the commit history.

I'd need to know a little bit more about how you're using `fsck` to have a different recommendation (please try to include a high level rationale in commit messages!) but my guess is this is for something like a "repair system" UI in a device.  I suspect the calling code instead should check for partial commits for refs that it cares about (e.g. whether the booted ref is partial) and also treat that as an error condition to be fixed.

We could also bake this in as `fsck --error-on-partial-refs` or so.  I think flatpak may have similar logic?

See also https://github.com/ostreedev/ostree/issues/959

cgwalters:
To restate, the core issue is that it's valid to have partial commits for reasons other than fsck pruned bad objects, and libostree doesn't have a way to distinguish.

Another option perhaps is to write e.g. `fsck-partial` into the statefile `state/<hash>.commitpartial` which would mean "partial, and expected to exist but was pruned by fsck" and `fsck` would continue to error out until the commit was re-pulled.  Right now the partial stamp file is empty, so it'd be fully compatible to write a rationale into it.

jwessel:
> Another option perhaps is to write e.g. `fsck-partial` into the statefile `state/<hash>.commitpartial` which would mean "partial, and expected to exist but was pruned by fsck" and `fsck` would continue to error out until the commit was re-pulled. Right now the partial stamp file is empty, so it'd be fully compatible to write a rationale into it.

I believe you described the answer to my question.  My question was if there was a way to have the fsck --delete (when it deletes something) leave the tree in a state where the next fsck will also exit with zero.   This is mainly so that a fix/pull process can execute at a later point.  A command line option for the fsck would be fine for the partial commits, but it seems that there are numerous reasons you might have partial commits.  Perhaps we even need both?

jwessel:
This pull req is dropped in favor of https://github.com/ostreedev/ostree/pull/1887

PR #1884, opened 2019-07-08
tests/sizes.js: Fix byte array unpacking

dbnicholson:
Recent GJS changed how byte arrays are unpacked with some assumptions
that they are likely strings. Manually use get_child_value() and
get_byte() to ensure the correct value is parsed when checking the
`ostree.sizes` metadata.

The upstream test is currently passing fine with GJS 1.56.2, but at
Endless we (unfortunately) have a downstream change that adds the object
type as an additional byte in the array. This is parsed incorrectly by
`deep_unpack()`. We can carry this patch downstream, but this change
makes the test more robust regardless.

One more commit is included that I noticed when inspecting the test log.

### DIFF
--- a/tests/test-sizes.js
+++ b/tests/test-sizes.js
@@ -64,10 +64,10 @@ assertEquals(nSizes, 2);
 let expectedUncompressedSizes = [12, 18];
 let foundExpectedUncompressedSizes = 0;
 for (let i = 0; i < nSizes; i++) {
-    let sizeEntry = sizes.get_child_value(i).deep_unpack();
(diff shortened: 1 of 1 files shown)


mwleeds:
bot, retest this please

cgwalters:
@rh-atomic-bot r+ 7974eba

dbnicholson:
Just for reference, @ptomato tracked this down to a spidermonkey bug in https://gitlab.gnome.org/GNOME/gjs/issues/260.

PR #1886, opened 2019-07-10
Validate config keys

r4f4:
WIP on the second patch because I want to confirm whether this is the desired approach.


r4f4:
To properly test this, #1898 is needed.

r4f4:
Dropping WIP since with #1898 the tests pass.

r4f4:
Updated to make use of #1899 in the tests.

r4f4:
> Ouch I'm sorry for never reviewing this one!! My bad 😦 Thanks for the work here. I'm trying to catch up on reviews here. Hope things are going well with you!

No problem. I also forgot about it and never drew attention to it.
All good, I hope you're well too :)

PR #1887, opened 2019-07-10
RFC - Implementation of partialcommit reason

jwessel:
This is an implementation as well as a test case for the partialcommit reason in order to allow the fsck operation to continue to exit with non-zero status after an "ostree fsck --delete" operation has been run.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1874,6 +1874,56 @@ ensure_txn_refs (OstreeRepo *self)
 }
 
 /**
+ * ostree_repo_mark_commit_partial_reason:
+ * @self: Repo
+ * @checksum: Commit SHA-256
+ * @is_partial: Whether or not this commit is partial
+ * @in_state: Reason bitmask for partial commit
+ * @error: Error
+ *
+ * Commits in "partial" state do not have all their child objects written. This
+ * occurs in various situations, such as during a pull, but also if a "subpath"
+ * pull is used, as well as "commit only" pulls.
+ *
+ * This function is used by ostree_repo_pull_with_options(); you
+ * should use this if you are implementing a different type of transport.
+ *
+ * Since: 2019.3
+ */
+gboolean
+ostree_repo_mark_commit_partial_reason (OstreeRepo     *self,
+                                        const char     *checksum,
+                                        gboolean        is_partial,
+                                        OstreeRepoCommitState in_state,
+                                        GError        **error)
+{
+  g_autofree char *commitpartial_path = _ostree_get_commitpartial_path (checksum);
+  if (is_partial)
+    {
+      glnx_autofd int fd = openat (self->repo_dir_fd, commitpartial_path,
+                                   O_EXCL | O_CREAT | O_WRONLY | O_CLOEXEC | O_NOCTTY, 0644);
+      if (fd == -1)
+        {
+          if (errno != EEXIST)
+            return glnx_throw_errno_prefix (error, "open(%s)", commitpartial_path);
+        }
+      else
+        {
+          if (in_state & OSTREE_REPO_COMMIT_STATE_FSCK_PARTIAL)
+            if (glnx_loop_write (fd, "f", 1) < 0)
+              return glnx_throw_errno_prefix (error, "write(%s)", commitpartial_path);
+        }
+    }
+  else
+    {
+      if (!ot_ensure_unlinked_at (self->repo_dir_fd, commitpartial_path, 0))
+        return FALSE;
+    }
+
+  return TRUE;
+}
+
+/**
  * ostree_repo_mark_commit_partial:
  * @self: Repo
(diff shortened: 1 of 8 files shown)


jwessel:
Looks to me like the test fails due to setfxattr() not being available on some of the fedora systems.  I can simply modify the test.

src/libostree/ostree-repo-commit.c
```diff
@@ -1915,6 +1923,31 @@ ostree_repo_mark_commit_partial (OstreeRepo     *self,
...
+ *
+ * Since: 2017.15
+ */
+gboolean
+ostree_repo_mark_commit_partial (OstreeRepo     *self,
+                                 const char     *checksum,
+                                 gboolean        is_partial,
+                                 GError        **error)
+{
+  return ostree_repo_mark_commit_partial_reason (self, checksum, is_partial, 0, error);
```
cgwalters: I'd prefer `OSTREE_REPO_COMMIT_STATE_NORMAL` over `0`.

src/libostree/ostree-repo-commit.c
```diff
@@ -1915,6 +1923,31 @@ ostree_repo_mark_commit_partial (OstreeRepo     *self,
...
+ * @self: Repo
+ * @checksum: Commit SHA-256
+ * @is_partial: Whether or not this commit is partial
+ * @error: Error
+ *
+ * Commits in "partial" state do not have all their child objects written. This
+ * occurs in various situations, such as during a pull, but also if a "subpath"
+ * pull is used, as well as "commit only" pulls.
+ *
+ * This function is used by ostree_repo_pull_with_options(); you
```
cgwalters: Looks like bad copy/paste?

src/libostree/ostree-repo.c
```diff
@@ -3737,7 +3737,20 @@ load_metadata_internal (OstreeRepo       *self,
...
-            *out_state |= OSTREE_REPO_COMMIT_STATE_PARTIAL;
+            {
+              glnx_autofd int fd = openat (self->repo_dir_fd, commitpartial_path, O_RDONLY, 0644);
+
+              *out_state |= OSTREE_REPO_COMMIT_STATE_PARTIAL;
+              if (fd != -1)
+                {
+                  char reason;
+                  if (read (fd, &reason, 1) == 1)
+                  {
```
cgwalters: Indentation is off here.

src/libostree/ostree-repo.c
```diff
@@ -3737,7 +3737,20 @@ load_metadata_internal (OstreeRepo       *self,
           if (!glnx_fstatat_allow_noent (self->repo_dir_fd, commitpartial_path, NULL, 0, error))
             return FALSE;
           if (errno == 0)
-            *out_state |= OSTREE_REPO_COMMIT_STATE_PARTIAL;
+            {
+              glnx_autofd int fd = openat (self->repo_dir_fd, commitpartial_path, O_RDONLY, 0644);
```
cgwalters: Missing `O_CLOEXEC`.  I'd use `ot_openat_ignore_enoent()`.

src/ostree/ot-builtin-fsck.c
```diff
@@ -410,7 +411,11 @@ ostree_builtin_fsck (int argc, char **argv, OstreeCommandInvocation *invocation,
             }
 
           if (commitstate & OSTREE_REPO_COMMIT_STATE_PARTIAL)
-            n_partial++;
+	    {
```
cgwalters: Indentation looks off; we only use spaces in ostree code (part of GNU style).

src/ostree/ot-builtin-fsck.c
```diff
@@ -127,7 +127,7 @@ fsck_one_object (OstreeRepo            *repo,
                   if ((state & OSTREE_REPO_COMMIT_STATE_PARTIAL) == 0)
                     {
                       g_printerr ("Marking commit as partial: %s\n", parent_commit);
-                      if (!ostree_repo_mark_commit_partial (repo, parent_commit, TRUE, error))
+                      if (!ostree_repo_mark_commit_partial_reason (repo, parent_commit, TRUE, OSTREE_REPO_COMMIT_STATE_FSCK_PARTIAL, error))
```
cgwalters: This is fine, but I just noticed that there's a race condition here - if `fsck` is interrupted between deleting the first object and marking the commit partial, we'd have a partial commit without the flag.

We should probably mark all commits partial *before* deletion.

(And it looks like we're marking the commit partial each time, which is a bit ugly, but not a new bug)

cgwalters:
Do you think you'll have a chance to work on this soon?  If not I or someone else could probably pick up this PR and put the few finishing touches on it while retaining you as author if that's OK with you!

jwessel:
I can make the cleanups and submit a new version next week.

jwessel:
This is replaced by https://github.com/ostreedev/ostree/pull/1910

PR #1889, opened 2019-07-18
Add conditional use of GPG sign&verify

d4s:
Followup of PR https://github.com/ostreedev/ostree/pull/1878#issuecomment-506987229

This patchset allows to build libostree without GPGME support.
The new option `--without-gpgme` has been added for `configure` to disable GPGME support. Build rules and code adopted for optional support of GPGME signing and verification.


configure.ac
```diff
@@ -210,25 +210,31 @@ m4_ifdef([GOBJECT_INTROSPECTION_CHECK], [
...
-    ],[ have_gpgme=no ])
-])
-AS_IF([ test x$have_gpgme = xno ], [
-   AC_MSG_ERROR([Need GPGME_PTHREAD version $LIBGPGME_DEPENDENCY or later])
-])
-OSTREE_FEATURES="$OSTREE_FEATURES gpgme"
-PKG_CHECK_MODULES(OT_DEP_GPG_ERROR, [gpg-error], [], [
+AC_ARG_WITH(gpgme,
+	    AS_HELP_STRING([--with-gpgme], [Use gpgme @<:@default=yes@:>@]),
+	    [], [with_gpgme=$with_gpgme_default])
```
cgwalters: Shouldn't this be
`with_gpgme=yes`

src/libostree/ostree-repo-pull.c
```diff
@@ -3592,10 +3606,12 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       (void) g_variant_lookup (options, "subdir", "&s", &dir_to_pull);
       (void) g_variant_lookup (options, "subdirs", "^a&s", &dirs_to_pull);
       (void) g_variant_lookup (options, "override-remote-name", "s", &pull_data->remote_refspec_name);
+#if defined(HAVE_GPGME)
       opt_gpg_verify_set =
         g_variant_lookup (options, "gpg-verify", "b", &pull_data->gpg_verify);
```
cgwalters: Maybe we should explicitly error out if this is set to `TRUE` but it's disabled at build time?  Dunno.

cgwalters:
Overall, this looks really good, thanks a lot for doing this!  I'm going to leave this open for a bit, and will try to get some time to do a local build and play with it.  But I'm inclined to merge this soon unless anyone else has comments.

configure.ac
cgwalters: (Nothing seems to define `with_gpgme_default`)

d4s:
> Overall, this looks really good, thanks a lot for doing this! I'm going to leave this open for a bit, and will try to get some time to do a local build and play with it. But I'm inclined to merge this soon unless anyone else has comments.

Thanks. Unfortunately only now I discovered that flatpack build is failed because it relies to gpgme. Looks like I need to reverse the check logic from `#ifdef HAVE_GPGME` to something like `#ifndef DISABLE_GPGME`.

src/libostree/ostree-autocleanups.h
```diff
@@ -50,7 +50,9 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (OstreeRepoDevInoCache, ostree_repo_devino_cache_u
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (OstreeAsyncProgress, g_object_unref)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (OstreeBootconfigParser, g_object_unref)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (OstreeDeployment, g_object_unref)
+#if defined(HAVE_GPGME)
```
cgwalters: Right, we can't have `#ifdef` in public headers.

Makefile-libostree-defines.am
```diff
@@ -48,6 +47,12 @@ libostree_public_headers = \
 	src/libostree/ostree-kernel-args.h \
 	$(NULL)
 
+if USE_GPGME
+libostree_public_headers += \
+	src/libostree/ostree-gpg-verify-result.h \
```
cgwalters: I think we probably need to keep installing this, and write a `ostree-gpg-verify-result-dummy.c` or so instead that's compiled if GPG is disabled?

Basically aim to keep the same public API, just make GPG operations an error.  We should also make it easy for apps to detect if libostree doesn't support GPG, maybe return `G_IO_ERROR_NOT_IMPLEMENTED` or so.

d4s: Ok. Will add that.

d4s: Should we support the "classic" CLI interface as well? i.e. command 'gpg_sign' and gpg-related options for 'commit', 'pull' etc.?

cgwalters: I'm uncertain.  Offhand...it seems nice if we can avoid having a different API/ABI depending on whether gpg is on or off.  But, I don't think maintaining the same ABI is a hard requirement either.

In practice, I expect most general purpose distributions to continue to enable GPG, and people who want to turn it off will do so globally.  However, it does highlight that dependent software like flatpak will also need to learn how to deal with this too.

Probably, we need to *generate* data in `ostree.h` with our build-time feature flags.  We can't use `#ifdef HAVE_GPG` there, that's what's breaking the CI. 
What we should do instead at build time generate `ostree-build-options.h` or so, that does `#define OSTREE_BUILDOPT_GPG` if it's enabled, and then install that header.

That way inside our own public headers we can use it, and so can dependent software.

cgwalters: The alternative is to maintain the same public API, and just return errors.

src/libostree/ostree-repo-pull.c
d4s: added the appropriate check for that.

Makefile-libostree-defines.am
d4s: Changes:
- `src/libostree/ostree-gpg-verify-result.h` is installed
- added `ostree-gpg-verify-result-dummy.c` with dummy public functions
- some unneeded "ifdef"s removed since I do not remove public API anymore
- rename define from "HAVE_GPGME" to "OSTREE_DISABLE_GPGME"

After checking of CLI utilities I decide to keep the removing of util and gpg-related options. At least we already have the same behavior for "curl", so...

believe the generation of `ostree.h` is bit out of scope of current PR.
However I think it would be enough to rename already generated `config.h` into `ostree-build-options.h` and include it to installables.

d4s: @cgwalters any thought about updated version?

cgwalters: OK related to the detection, I did https://github.com/ostreedev/ostree/pull/1890

cgwalters:
Looks like the test suite is still running the GPG tests, so we'd need to do the `ostree --version | grep gpgme` dance but...hm, why did just Travis fail?  

Other than that I think this is probably good to go.

Makefile-libostree-defines.am
d4s: hm.... does it means I need to return back logic "#ifdef HAVE_GPGME" and hope downstream will check if feature is enabled?

d4s:
> Looks like the test suite is still running the GPG tests, 

yes, some tests use signing as a part of testing process. Did not change the tests themselves yet since later I have to change them to adopt for working with proposed signing/verification mechanism.

> so we'd need to do the `ostree --version | grep gpgme` dance but...hm, why did just Travis fail?

at least it fails in the same way as couple of previous MRs.
btw I have the same 2 not passed tests on local Debian Buster as Travis have.

Makefile-libostree-defines.am
cgwalters: The main thing we *needed* to fix is having `#ifdef HAVE_GPGME` in our public headers.  It just can't work.  And I believe your current PR has addressed that.

The next problem is though; apps like flatpak probably want to detect whether or not libostree supports gpg.   The PR I did isn't quite ergonomic for C code to do build-time checking; can't do `strstr()` in the C preprocessor.  I think we'd have to explode the feature flags into `#defines`.  But it was easy to do #1890 and it does allow runtime dispatch at least.

d4s: > The main thing we _needed_ to fix is having `#ifdef HAVE_GPGME` in our public headers. It just can't work. And I believe your current PR has addressed that.

but it is already fixed. In last available PR I did not touch public headers.
Sorry for force pushing to the source branch -- but otherwise changes would be even worse to read since I have to rework `ifdef` logic to `ifndef`

> The next problem is though; apps like flatpak probably want to detect whether or not libostree supports gpg. 

yeah, that's why I did that reversion to `#ifndef OSTREE_DISABLE_GPGME`  -- now all apps just build and dummy GPG stuff will throw an error in runtime if there is no GPGME support but it is requested by app.
So the fixed version shouldn't break public API.

d4s:
@cgwalters now all lights are green, thanks to #1892.

cgwalters:
Awesome, thanks so much for working on this!  There are some other things I'd like to tweak but nothing blocking, better to get this in and iterate from there.  Thanks again!

@rh-atomic-bot r+ df5c6d3

PR #1891, opened 2019-07-26
lib/pull: Allow downgrade protection with overrides

jlebon:
There's a valid use case for enabling the timestamp downgrade check
while still also using override commits.

We'll make use of this in Fedora CoreOS, where the agent specifies the
exact commit to upgrade to, while still enforcing that it be newer.


cgwalters:
Nice!

@rh-atomic-bot r+ e23feb3

jlebon:
@rh-atomic-bot retry

jlebon:
@rh-atomic-bot retry

PR #1892, opened 2019-07-26
Fix CI tests

dbnicholson:
Some of the CI tests are always failing, which makes it hard to know if your PR is adding new failures.


dbnicholson:
I did what I could but I don't believe I can fix the `f29-rpmostree` failure. As far as I can tell that's fixed by https://github.com/projectatomic/rpm-ostree/commit/45b6186be0826889d805ba55bee3be75b042ed35, but that's only in rpm-ostree 2019.3 and the f29 test is running rpm-ostree 2019.1. I don't know how to make it run something newer.

Anyways, this is ready for review.

tests/test-gpg-signed-commit.sh
```diff
@@ -233,37 +233,52 @@ echo "ok verified with key2 primary expired"
...
-assert_file_has_content test2-show '^Found 2 signatures'
-assert_not_file_has_content test2-show 'public key not found'
-assert_file_has_content test2-show "${key1_id}"
-assert_file_has_content test2-show "${key2_sub_id}"
-assert_file_has_content test2-show 'Good signature from "Test Key 1 <>"'
-assert_file_has_content test2-show 'BAD signature from "Test Key 2 <>"'
-assert_file_has_content test2-show 'Key expired'
-assert_file_has_content test2-show "Primary key ID ${key2_id}"
-assert_file_has_content test2-show 'Primary key expired'
+if ! ${GPG} --homedir=${tmpgpg_home} --quick-set-expire ${key2_fpr} seconds=1 ${key2_sub_fpr}; then
```
jlebon: Hmm, could we do something like this instead here:

```
if ! (${GPG} --quick-set-expire || :) |& grep -q SUBKEY-FPRS; then
...
```

?

dbnicholson: Sure, I think that could be a little better. I usually shy away from scraping tool output since it will just break things when upstream changes them, but I guess that's better than papering over other issues. I did look at the usage output for my current gpg and the one on debian stretch and it is as you say.

gpg-2.2.12:
```
$ gpg --quick-set-expire 
usage: gpg [options] --quick-set-exipre FINGERPRINT EXPIRE [SUBKEY-FPRS]
```
gpg-2.1.18:
```
# gpg --quick-set-expire
usage: gpg [options] --quick-set-exipre FINGERPRINT EXPIRE
```

dbnicholson:
> Just one comment, otherwise looks good to me!
> 
> > I did what I could but I don't believe I can fix the `f29-rpmostree` failure.
> 
> It _should_ work to just bump `RPMOSTREE_TAG` to `v2019.3`, but feel free to leave that one alone for now.

I'll give it a shot.

dbnicholson:
All tests passing now. I forget, does homu take care of squashing my fixups or should I rewrite the history myself?

cgwalters:
Homu will squash.  Thanks so much for doing this!  I owe you :beers: 

@rh-atomic-bot + d4baf60

tests/test-gpg-signed-commit.sh
dbnicholson: For some reason `(${GPG} --quick-set-expire || :) |&` didn't actually send the output to `grep` when I ran it through the test suite (it did work in my interactive shell). Instead I added an explicit `2>&1` in the `${GPG}` command.

dbnicholson:
> Homu will squash. Thanks so much for doing this! I owe you 

Well, one of them I introduced myself, so :stuck_out_tongue:

dbnicholson:
@cgwalters it doesn't look like the merge was initiated.

jlebon:
All green, beautiful!

> @cgwalters it doesn't look like the merge was initiated.

Ahh yup, we're missing an `r`.
@rh-atomic-bot r+ d4baf60

PR #1894, opened 2019-08-06
admin/init-fs: Add a --modern switch

cgwalters:
This skips creating the default stuff in the physical sysroot.
I don't recall why I did that to be honest; it originated with
the first commit of this file.  It might not have ever been
necessary.

In any case, it's not necessary now with Fedora CoreOS, so
prune it and let's have a clean `/`.

Keep the old behavior by default though to avoid breaking anyone.

### DIFF
--- a/src/ostree/ot-admin-builtin-init-fs.c
+++ b/src/ostree/ot-admin-builtin-init-fs.c
@@ -30,7 +30,10 @@
 
 #include <glib/gi18n.h>
 
+static gboolean opt_modern;
+
 static GOptionEntry options[] = {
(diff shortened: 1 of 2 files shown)


cgwalters:
@rh-atomic-bot r=ajeddeloh 02c390c

cgwalters:
@rh-atomic-bot r=ajeddeloh 7ab13fb

cgwalters:
OK, added the missing `echo ok`

cgwalters:
@rh-atomic-bot delegate=ajeddeloh

ajeddeloh:
@rh-atomic-bot r+ d1dc848

PR #1898, opened 2019-08-09
config/set: Don't fail when setting negative value

### LINKED ISSUE #1893: `ostree config set` fails when setting negative values
If you try to set a negative value for a config option (e.g `lock-timeout-secs` with value `-1` means `LOCK_DISABLED`) you'll get instead:
```
./ostree config --repo=/tmp/test-repo-6Zwkyk/ set core.lock-timeout-secs -1
error: Unknown option -1
```
So `-1` is being considered a new command line option instead of a value for the just entered one.

r4f4:
Some config options have special meaning when the value is negative. For
example, `core.lock-timeout-secs` when set to -1 means timeout disabled.
Currently, `g_option_context_parse` fails when parsing negative values:
```
ostree config set core.lock-timeout-secs -1
error: Unknown option -1
```
We need to tell it to ignore unknown options instead of erroring out and
let the `ostree_builtin_config` code handle the value properly.

Fixes #1893

### DIFF
--- a/src/ostree/ot-builtin-config.c
+++ b/src/ostree/ot-builtin-config.c
@@ -74,6 +74,8 @@ ostree_builtin_config (int argc, char **argv, OstreeCommandInvocation *invocatio
   int correct_argc;
 
   context = g_option_context_new ("(get KEY|set KEY VALUE|unset KEY)");
+  /* Needed so we don't fail when passing negative integers as VALUE */
+  g_option_context_set_ignore_unknown_options (context, TRUE);
 
   if (!ostree_option_context_parse (context, options, &argc, &argv, invocation, &repo, cancellable, error))
     return FALSE;
--- a/src/ostree/ot-builtin-config.c
+++ b/src/ostree/ot-builtin-config.c
@@ -71,7 +71,7 @@ ostree_builtin_config (int argc, char **argv, OstreeCommandInvocation *invocatio
   g_autofree char *section = NULL;
   g_autofree char *key = NULL;
   g_autoptr(GKeyFile) config = NULL;
-  int correct_argc;
+  int i, correct_argc;
 
   context = g_option_context_new ("(get KEY|set KEY VALUE|unset KEY)");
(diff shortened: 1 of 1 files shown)


rfairley:
Doing this makes sense to me. Just one minor typo in the commit title (`Do'nt`). CI test failed on `pull-repeated` - which is known to flake.

r4f4:
I fixed the typo.
Also I added a check for unknown options as a separate commit in case you like the approach and want to use it.

jlebon:
Using `g_option_context_set_ignore_unknown_options` is OK, though it's unfortunate to be inconsistent with the rest of the CLI. Hmm, WDYT about instead adding a `--value` switch? Then it'd be e.g.:

```
ostree config set core.lock-timeout-secs --value=-1
```

r4f4:
Should the switch be used always or only when setting negative values?

jlebon:
Basically, all these invocations would be equivalent:

```
$ ostree config set GROUP.KEY VAL
$ ostree config set --group=GROUP KEY VAL
$ ostree config set GROUP.KEY --value=VAL
$ ostree config set --group=GROUP KEY --value=VAL
```

We should probably support `--key` too for completeness, but meh... could leave that out for now. I don't expect us to have a key that starts with `-` anytime soon. :)

r4f4:
Ok. You'd still get an error from `g_option_context_parse` when ignoring unknown is off though when using
```
$ ostree config set group.key value
$ ostree config set --group=group key value
```
if value is negative: "Unknown option: -X"

jlebon:
WDYT about doing #1899 instead?

rfairley:
> WDYT about doing #1899 instead?

With #1899 I'm in favor of keeping the CLI conistent and requiring `--` or a `--value` key to be specified. Adding a `--value=` key would be helpful here (myself I looked for a `--value` first before trying `--`). I think that would provide enough alternative options for adding values starting with `-`, such that we don't need to tell GLib to ignore the unknown options here anymore.

r4f4:
Closing this since #1899 has been merged.

PR #1899, opened 2019-08-13
bin: Better handle --

jlebon:
We would stop passing through `--` and args after it to the underlying
command in `ostree_run`. This made it impossible to use `--` to tell the
parser that following args starting with `-` really are positional.

AFAICT, that logic for `--` here came from a time when we parse options
manually in a big loop, in which case breaking out made sense (see
97558276e4f855442337be01abc8f90cf0dd1810).

There's an extra step here, which is that glib by default leaves the
`--` in the list of args, so we need to take care to remove it from the
list after parsing.

Closes: #1898

### DIFF
--- a/src/ostree/ot-main.c
+++ b/src/ostree/ot-main.c
@@ -152,11 +152,6 @@ ostree_run (int    argc,
             }
         }
 
-      else if (g_str_equal (argv[in], "--"))
-        {
-          break;
-        }
-
       argv[out] = argv[in];
     }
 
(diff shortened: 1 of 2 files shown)


r4f4:
LGTM

rfairley:
LGTM!

@rh-atomic-bot r+ e80fe53f9f3f854282e0efb157af7f9cf7388137

rfairley:
@rh-atomic-bot retry

rfairley:
@rh-atomic-bot retry

Looks like infra flakes :snowflake:

rfairley:
@rh-atomic-bot retry

PR #1901, opened 2019-08-21
lib/deploy: handle FIFREEZE ENOSYS failure

dsd:
When running under qemu, unimplemented ioctls such as FIFREEZE
return ENOSYS, and this causes the deployment to fail.

Catch this and handle it like EOPNOTSUPP.

I'm not sure if qemu's behaviour is fully correct here (or if it should
return EOPNOTSUPP) but it's trivial to handle regardless.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1374,7 +1374,7 @@ fsfreeze_thaw_cycle (OstreeSysroot *self,
            * the filesystem is already frozen (EBUSY).
            * OK, let's just do a syncfs.
            */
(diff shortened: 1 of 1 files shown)


cgwalters:
@rh-atomic-bot r+ 9e19feb

I don't understand though; how is qemu responding to system calls?  qemu is just executing the Linux kernel and it's the *filesystem* handling that syscall right?

dsd:
ioctls often take parameters that vary on different architectures. For example int, struct.
So qemu needs to be taught which parameters are given or returned from each ioctl & how to convert them to/from the host arch.

cgwalters:
Oh, I see, this is cross-architecture qemu emulation, not KVM.  OK.

Issue #1907, opened 2019-08-30
ostree-2019.3 spams terminal with gpg-connect-agent messages for every flatpak operation


kparal:
With `ostree-2019.3-3.fc30.x86_64` I see this:
```
$ flatpak install peek
Looking for matches…
gpg-connect-agent: no running gpg-agent - starting '/usr/bin/gpg-agent'
gpg-connect-agent: waiting for the agent to come up ... (5s)
gpg-connect-agent: connection to agent established
Found similar ref(s) for ‘peek’ in remote ‘flathub’ (system).
Use this remote? [Y/n]: y
Found ref ‘app/com.uploadedlobster.peek/x86_64/stable’ in remote ‘flathub’ (system).
Use this ref? [Y/n]: y
gpg-connect-agent: no running gpg-agent - starting '/usr/bin/gpg-agent'
gpg-connect-agent: waiting for the agent to come up ... (5s)
gpg-connect-agent: connection to agent established
Required runtime for com.uploadedlobster.peek/x86_64/stable (runtime/org.gnome.Platform/x86_64/3.32) found in remote flathub
Do you want to install it? [Y/n]: n
```

With `ostree-2019.2-1.fc30.x86_64` I see this:
```
$ flatpak install peek
Looking for matches…
Found similar ref(s) for ‘peek’ in remote ‘flathub’ (system).
Use this remote? [Y/n]: y
Found ref ‘app/com.uploadedlobster.peek/x86_64/stable’ in remote ‘flathub’ (system).
Use this ref? [Y/n]: y
Required runtime for com.uploadedlobster.peek/x86_64/stable (runtime/org.gnome.Platform/x86_64/3.32) found in remote flathub
Do you want to install it? [Y/n]: n
```

Compare for yourself which one looks more readable. The gpg-agent messages are printed out **very** frequently, for a single `flatpak update` command that has "nothing to do", I saw about 20 lines of gpg-agent messages (depends on what's happening in the background, usually it's just 3-6 lines of messages).

This is happening on Fedora 30, but Arch users see it as well:
https://bbs.archlinux.org/viewtopic.php?id=248745

This happens regardless of `gpg-agent` actually running or not running in the background.

cgwalters:
Likely fallout from https://github.com/ostreedev/ostree/pull/1799

cgwalters:
@dbnicholson

dbnicholson:
[...]

PR #1908, opened 2019-08-30
lib/gpg: Only show gpg-connect-agent stderr on failures

dbnicholson:
When listing GPG keys, the temporary GPG homedir will be constructed by
simply copying the remote's trusted keys to the pubring.gpg file. In
that case, no GPG operations spawning gpg-agent will be run. When
gpg-connect-agent is run to cleanup the homedir, it will helpfully print
on stderr that it's starting gpg-agent like so:

gpg-connect-agent: no running gpg-agent - starting '/usr/bin/gpg-agent'
gpg-connect-agent: waiting for the agent to come up ... (5s)
gpg-connect-agent: connection to agent established

Send gpg-connect-agent's stderr to a pipe and only send it to the
application's stderr if an error was encountered.

Fixes: #1907

### DIFF
--- a/src/libotutil/ot-gpg-utils.c
+++ b/src/libotutil/ot-gpg-utils.c
@@ -25,6 +25,7 @@
 
 #include <stdlib.h>
 
+#include <gio/gunixoutputstream.h>
 #include "libglnx.h"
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Nice, thanks!

@rh-atomic-bot r+ 745d7fd

Issue #1859, opened 2019-05-16
Fix kargs ordering to be "value" aware


jlebon:
Right now we just store the order of kargs by their keys. Which means we flush them out one key at a time (possibly multiple times for each associated value). But in fact, we need to be more flexible than this. See e.g. https://www.ibm.com/support/knowledgecenter/en/linuxonibm/com.ibm.linux.z.lgdd/lgdd_r_largepage_setup.html:

```
hugepagesz=1M hugepages=8 hugepagesz=2G hugepages=2
```

Trying with rpm-ostree (which mostly uses a private copy of the kargs code in libostree):

```
[root@localhost ~]# rpm-ostree kargs
console=tty0 console=ttyS0,115200n8 rootflags=defaults,prjquota rw $ignition_firstboot root=UUID=6ba949a3-e232-42cd-8411-4eeb45e25c05 ostree=/ostree/boot.1/fedora-coreos/05798738ee6c0b344460758df21199601b74a1ef26af3d3c1f79ec5e39157d7e/0 ignition.platform.id=qemu
[root@localhost ~]# rpm-ostree kargs --append hugepagesz=1M --append hugepages=8
...
[root@localhost ~]# rpm-ostree kargs --append hugepagesz=2G --append hugepages=2
...
[root@localhost ~]# rpm-ostree kargs
console=tty0 console=ttyS0,115200n8 rootflags=defaults,prjquota rw $ignition_firstboot root=UUID=6ba949a3-e232-42cd-8411-4eeb45e25c05 ostree=/ostree/boot.1/fedora-coreos/05798738ee6c0b344460758df21199601b74a1ef26af3d3c1f79ec5e39157d7e/0 ignition.platform.id=qemu hugepagesz=1M hugepagesz=2G hugepages=8 hugepages=2
```

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/1909

PR #1909, opened 2019-08-30
lib/kernel-args: Store kernel args as key/value entries

rfairley:
Define an `OstreeKernelArgsEntry` structure, which holds
both the key and the value. The kargs order array stores
entries for each key/value pair, instead of just the keys.
The hash table is used to locate entries, by storing
entries in a pointer array for each key. This preserves
ordering information about each key/value pair when
appending/replacing/deleting kargs, and maintains the
same public interface.

Fixes: #1859

---

Proposing this as an approach for #1859 . `tests/test-kargs.c` worked locally. Letting `tests/test-admin-deploy-karg.sh` run in CI as a seemingly unrelated error occurred in my local container `error: Cleaning deployments: ioctl(EXT2_IOC_GETFLAGS): Bad file descriptor`. Not yet manually verified when installed with `rpm-ostree` - leaving WIP for now.

### DIFF
--- a/src/libostree/ostree-kernel-args.c
+++ b/src/libostree/ostree-kernel-args.c
@@ -30,6 +30,111 @@ struct _OstreeKernelArgs {
   GHashTable *table;
 };
 
+struct _OstreeKernelArgsEntry {
+  char *key;
+  char *value;
+};
+
+char *
+_ostree_kernel_args_entry_get_key (const OstreeKernelArgsEntry *e)
+{
+  return e->key;
+}
+
+char *
+_ostree_kernel_args_entry_get_value (const OstreeKernelArgsEntry *e)
+{
+  return e->value;
+}
+
+void
+_ostree_kernel_args_entry_set_key (OstreeKernelArgsEntry *e,
+                                   char  *key)
+{
+  e->key = key;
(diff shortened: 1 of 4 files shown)


cgwalters:
/approve (in Prow terms)

I wonder whether we really need the complexity of implementing an ordered map versus just using an array - it's not like people have hundreds of kargs.  But I'm also fine keeping it as you've done!  Code looks good offhand.

rfairley:
Apologies for not getting back to this sooner. Now rebased, and tested with with `rpm-ostree-2019.6-1.fc31` and `ostree-2019.5.13.gaadc4db0-2.fc31` (RPM built at this commit) present in FCOS. The expected order `hugepagesz=1M hugepages=8 hugepagesz=2G hugepages=2` now shows in `/proc/cmdline` using the commands in #1859. Also tried a `cosa build` with `ostree-2019.5.13.gaadc4db0-2.fc31` installed in the `coreos-assembler` image; `/proc/cmdline` in FCOS when run looks to be formed correctly.

rfairley:
Seeing `ERROR: tests/test-pull-repeated.sh` in libsoup tests - known flake.

PR #1910, opened 2019-08-30
Implementation of partialcommit reason

jwessel:
This a cleaned up version of the rfc fsck patch. It contains the implementation of the partial commit reason as well as a test case for the partialcommit reason.  The intent is to allow the fsck operation to continue to exit with non-zero status after an "ostree fsck --delete" operation has been run, which has removed objects.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3768,10 +3768,19 @@ load_metadata_internal (OstreeRepo       *self,
           g_autofree char *commitpartial_path = _ostree_get_commitpartial_path (sha256);
           *out_state = 0;
 
-          if (!glnx_fstatat_allow_noent (self->repo_dir_fd, commitpartial_path, NULL, 0, error))
+          glnx_autofd int fd = -1;
+          if (!ot_openat_ignore_enoent (self->repo_dir_fd, commitpartial_path, &fd, error))
             return FALSE;
-          if (errno == 0)
-            *out_state |= OSTREE_REPO_COMMIT_STATE_PARTIAL;
+          if (fd != -1)
+            {
+              *out_state |= OSTREE_REPO_COMMIT_STATE_PARTIAL;
+               char reason;
+               if (read (fd, &reason, 1) == 1)
+                 {
+                   if (reason == 'f')
+                     *out_state |= OSTREE_REPO_COMMIT_STATE_FSCK_PARTIAL;
+                 }
+            }
         }
     }
   else if (self->parent_repo)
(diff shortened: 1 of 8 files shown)


src/libostree/ostree-repo.c
```diff
@@ -3771,7 +3771,20 @@ load_metadata_internal (OstreeRepo       *self,
           if (!glnx_fstatat_allow_noent (self->repo_dir_fd, commitpartial_path, NULL, 0, error))
             return FALSE;
           if (errno == 0)
-            *out_state |= OSTREE_REPO_COMMIT_STATE_PARTIAL;
+            {
+              glnx_autofd int fd = -1;
+              ot_openat_ignore_enoent (self->repo_dir_fd, commitpartial_path, &fd, error);
```
cgwalters: Still need to do:
```
if (!ot_openat_ignore_enoent (...))
  return FALSE;
```

The idea is that we still want to surface *other* I/O errors like `EIO`, `EPERM` etc.

src/ostree/ot-builtin-fsck.c
```diff
@@ -410,7 +411,11 @@ ostree_builtin_fsck (int argc, char **argv, OstreeCommandInvocation *invocation,
             }
 
           if (commitstate & OSTREE_REPO_COMMIT_STATE_PARTIAL)
-            n_partial++;
+            {
+              n_partial++;
+              if (commitstate & OSTREE_REPO_COMMIT_STATE_FSCK_PARTIAL)
+n_fsck_partial++;
```
cgwalters: Indentation looks off here, we just use spaces.

src/ostree/ot-builtin-fsck.c
```diff
@@ -450,5 +455,8 @@ ostree_builtin_fsck (int argc, char **argv, OstreeCommandInvocation *invocation,
   if (found_corruption)
     return glnx_throw (error, "Repository corruption encountered");
 
+  if (n_fsck_partial > 0)
+    return glnx_throw (error, "%u fsck deleted partial commits not verified\n", n_partial);
```
cgwalters: No newlines in single line error messages.

src/libostree/ostree-repo.c
cgwalters: Actually we could just stop doing the `stat()` call above too; no reason to do `stat()` as an existence check + `open()` when one can just do `open()` and check for `ENOENT`.

cgwalters:
bot, add author to whitelist

cgwalters:
I'd like to get this in, so I applied the suggested changes above and force-pushed to your branch.  Hope that's OK with you!

jwessel:
Thanks for all your help!

cgwalters:
To clarify I only updated your branch, this PR still needs to merge via CI.

@rh-atomic-bot r+ a2be082

PR #1912, opened 2019-09-03
Gate ostree-trivial-httpd on BUILDOPT_TRIVIAL_HTTPD

akiernan:
When building without --enable-trivial-httpd-cmdline, don't build or install
the ostree-trivial-httpd binary.

### DIFF
--- a/Makefile-ostree.am
+++ b/Makefile-ostree.am
@@ -143,12 +143,13 @@ ostree_SOURCES += src/ostree/ot-builtin-pull.c
 endif
 
 if USE_LIBSOUP
-# Eventually once we stop things from using this, we should support disabling this
+if BUILDOPT_TRIVIAL_HTTPD
 ostree_SOURCES += src/ostree/ot-builtin-trivial-httpd.c
 pkglibexec_PROGRAMS += ostree-trivial-httpd
 ostree_trivial_httpd_SOURCES = src/ostree/ostree-trivial-httpd.c
 ostree_trivial_httpd_CFLAGS = $(ostree_bin_shared_cflags) $(OT_INTERNAL_SOUP_CFLAGS)
 ostree_trivial_httpd_LDADD = $(ostree_bin_shared_ldadd) $(OT_INTERNAL_SOUP_LIBS)
+endif
 
 if !USE_CURL
 # This is necessary for the cookie jar bits
--- a/configure.ac
+++ b/configure.ac
@@ -195,6 +195,9 @@ AC_ARG_ENABLE(trivial-httpd-cmdline,
   [Continue to support "ostree trivial-httpd" [default=no]])],,
   enable_trivial_httpd_cmdline=no)
 AM_CONDITIONAL(BUILDOPT_TRIVIAL_HTTPD, test x$enable_trivial_httpd_cmdline = xyes)
+AS_IF([test x$with_soup = xno && test x$enable_trivial_httpd_cmdline = xyes], [
+  AC_MSG_ERROR([trivial-httpd enabled, but libsoup is not; libsoup is needed for trivial-httpd])
+])
 AM_COND_IF(BUILDOPT_TRIVIAL_HTTPD,
   [AC_DEFINE([BUILDOPT_ENABLE_TRIVIAL_HTTPD_CMDLINE], 1, [Define if we are enabling ostree trivial-httpd entrypoint])]
 )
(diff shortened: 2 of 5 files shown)


akiernan:
I did wonder when I did this if it was the way it was because of the tests... I guess this means it is.

cgwalters:
bot, add author to whitelist

Makefile-ostree.am
```diff
@@ -143,12 +143,14 @@ ostree_SOURCES += src/ostree/ot-builtin-pull.c
 endif
 
 if USE_LIBSOUP
+if BUILDOPT_TRIVIAL_HTTPD
 # Eventually once we stop things from using this, we should support disabling this
```
akiernan: Presumably this comment should be removed along with this changeset?

akiernan:
bot, retest this please

akiernan:
I'm a bit confused on this... the f29 failures look totally unrelated to me - is this good to merge, or is there something I'm missing?

configure.ac
```diff
@@ -192,6 +192,9 @@ AC_ARG_ENABLE(trivial-httpd-cmdline,
   [Continue to support "ostree trivial-httpd" [default=no]])],,
   enable_trivial_httpd_cmdline=no)
 AM_CONDITIONAL(BUILDOPT_TRIVIAL_HTTPD, test x$enable_trivial_httpd_cmdline = xyes)
+AS_IF([test x$with_soup = xno && test x$enable_trivial_httpd_cmdline = xyes], [
+  AC_MSG_WARN([trivial-httpd enabled, but libsoup is not; libsoup is needed for trivial-httpd])
```
cgwalters: I think this should probably be fatal; it's too easy to miss warnings in the autoconf output spew.

Makefile-ostree.am
cgwalters: Yep

akiernan:
> If it's that easy, sure; but did you test make check etc?

That was where I went wrong initially... so I added a initial commit which pushes `--enable-trivial-httpd-cmdline` into all the test paths; maybe that's cheating...

Right now the combinations which get covered by Travis for `make check` pass.

> better to refactor the tests so that we also use a basic webserver like Python's or whatever, and only require trivial-httpd for the tests that use its tweaks like --random-500s=50.

That's quite a lot more than I was really planning to do. I wonder if anyone's really using `trivial-httpd`, or if it's just being used for the tests?

cgwalters:
> That was where I went wrong initially... so I added a initial commit which pushes --enable-trivial-httpd-cmdline into all the test paths; maybe that's cheating...
> Right now the combinations which get covered by Travis for make check pass.

Hmm, OK.  Will have to look to see how that's working.

> That's quite a lot more than I was really planning to do. I wonder if anyone's really using trivial-httpd, or if it's just being used for the tests?

Yeah I definitely don't expect you to dive into that.  I hope no one is using it besides the tests.

I think it might not actually be too hard though to split out the tests which use the trivial-httpd features.

akiernan:
> the f29 failures look totally unrelated to me

This was just the f29 failing in setup, it's working now and I can see real failures...

akiernan:
@rh-atomic-bot r+ 2629328b39ae117876f4854409c2231220a30b9c

dbnicholson:
Sorry, but this is broken. `BUILDOPT_TRIVIAL_HTTPD` says whether you want `ostree trivial-httpd` to be supported, not whether you want the separate `ostree-trivial-httpd` libexec program. So, this commit says that both `ostree trivial-httpd` should be supported _and_ `ostree-trivial-httpd` should be installed. That's what was happening before, but now the test suite is completely if you don't build with `--enable-trivial-httpd-cmdline`.

What are you trying to achieve exactly? Why is building ostree-trivial-httpd a problem?

akiernan:
> What are you trying to achieve exactly?

Avoiding the libexec helper being installed in all cases. When I started looking at it, I guess I'd not realised that it really was just there as part of the test suite, not something most people will actually want.

Maybe what I really want is just delete it after `make install`

dbnicholson:
Debian puts it in a separate ostree-tests package since it's used by the installed tests.

If you want to not install it at all (which would break the installed tests, but probably you don't care), then you could add `ostree-trivial-httpd` to `check_PROGRAMS` so it's only built for the in-tree tests. That still wouldn't match the semantics of the current configure option, though. You'd probably want to add a second one to not install the libexec program.

akiernan:
I guess I waded into this assuming it was a bug... seems like that's not true and it's expected behaviour.

Reverting it seems like the best approach and taking the debian route of packaging it separately.

akiernan:
https://github.com/ostreedev/ostree/pull/1950

PR #1914, opened 2019-09-05
Bootloader no boot partition

palvarez89:
This is #1404 rebased

Currently tests are not passing.


palvarez89:
Hm... So I tried to reproduce the error (in my development environment within docker) and it worked.

I've started looking at the implementation, and I'm not sure this will work when using the `--sysroot` option.

My current use-case: I'm constructing a rootfs that later will be used to create a disk image. I don't think knowing the current mount points will help for this case.

cgwalters:
Closing since https://github.com/ostreedev/ostree/pull/2149 merged.

PR #1915, opened 2019-09-05
lib/gpg: Don't kill gpg-agent on newer gnupg

dbnicholson:
GnuPG 2.1.17 contains a bug fix so that `gpg-agent` is killed when the
entire GPG home directory is deleted[1]. If the host's GnuPG is new
enough, then we don't need to bother calling `gpg-connect-agent` to kill
the agent since it will be cleaned up on its own.

Get the GnuPG version from the GPGME OpenPGP engine info and parse it to
see if it matches this criteria.

1. https://dev.gnupg.org/T2756

### DIFF
--- a/src/libotutil/ot-gpg-utils.c
+++ b/src/libotutil/ot-gpg-utils.c
@@ -495,7 +495,7 @@ ot_gpgme_kill_agent (const char *homedir)
   /* If gnupg is at least 2.1.17, gpg-agent will exit when the homedir
    * is deleted.
    */
-  guint gnupg_major, gnupg_minor, gnupg_patch;
+  guint gnupg_major = 0, gnupg_minor = 0, gnupg_patch = 0;
   if (get_gnupg_version (&gnupg_major, &gnupg_minor, &gnupg_patch))
     {
       if ((gnupg_major > 2) ||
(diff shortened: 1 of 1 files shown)


src/libotutil/ot-gpg-utils.c
```diff
@@ -439,11 +439,75 @@ ot_gpgme_new_ctx (const char *homedir,
...
+
 void
 ot_gpgme_kill_agent (const char *homedir)
 {
   g_return_if_fail (homedir != NULL);
 
+  /* If gnupg is at least 2.1.17, gpg-agent will exit when the homedir
+   * is deleted.
+   */
+  guint gnupg_major, gnupg_minor, gnupg_patch;
```
cgwalters: The code as written is fine, but I tend to have a preference to initialize variables since it's very easy for code refactoring to change assumptions about that; IOW
` guint gnupg_major = 0, gnupg_minor = 0, gnupg_patch = 0`.

dbnicholson: Oh, definitely. I'm surprised the compiler didn't warn about use of uninitialized values.

cgwalters:
@rh-atomic-bot r+ b9e841f

PR #1916, opened 2019-09-06
ci/prow: add helper dockerfiles

vrutkovs:
Add helper dockerfiles for Prow builds. The base for all tests is `Dockerfile.buildroot` which only contains builddeps. Other dockerfiles build  ostree with various config options

### DIFF
--- a/ci/prow/Dockerfile
+++ b/ci/prow/Dockerfile
@@ -1,4 +1,0 @@
-FROM registry.fedoraproject.org/fedora:30
-WORKDIR /src
-COPY . .
-RUN ./ci/build.sh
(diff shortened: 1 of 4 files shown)


cgwalters:
Please, let's use https://github.com/coreos/coreos-assembler/pull/730 ?

cgwalters:
We should be merging that today.

vrutkovs:
Does ostree need coreos-assembler as a base too? I'd rather use a minimal base image - and use coreos-assembler:buildroot for rpm-ostree specifically

cgwalters:
I want ostree's tests to use cosa (and rpm-ostree), yes.

vrutkovs:
Sure, done

cgwalters:
Sorry about the delay here...I think we ended up doing this in a different way, by having a shared fcos buildroot container image built elsewhere e.g.

Issue #1913, opened 2019-09-05
ostree causing issues with multiple KDE users


tlrdrb:
https://bugs.kde.org/show_bug.cgi?id=411421 my issue
https://bugs.kde.org/show_bug.cgi?id=411415 another user
https://bugs.archlinux.org/task/63614 arch linux bug report

all three examples show strange behavior when using latest ostree version. downgrading to 2019.2 solves each of our issues.

what I just realized is that as soon as latte-dock is started and is run in debug mode, a message about gpg agent is displayed, and after that, the system starts displaying this behavior. Coincidentally, there was a commit 20 days ago about gpg agent, so it may be related.

tlrdrb:
steps to reproduce: 

- set up a simple arch linux or latest manjaro + KDE environment
- make sure ostree is latest version
- install latte-dock

- run latte dock in debug mode latte-dock -d
- press windows/super/meta key for a period longer than 1 second
- observe complete freeze.

- downgrade ostree to 2019.2
- observe none of this behavior.

found more mentions of this:

https://forum.manjaro.org/t/testing-update-2019-08-28-kernels-nvidia-kde-dev-mesa-python-haskell/100660/25

cgwalters:
Um, wow.  It sounds somehow this is a duplicate of https://github.com/ostreedev/ostree/issues/1907 but I don't have any idea offhand how...

Oh, hmm, I guess if latte-dock is querying libflatpak it might be spamming stderr?

tlrdrb:
No idea about that, but these messages are displayed:

gpg-connect-agent: no running gpg-agent - starting '/usr/bin/gpg-agent'
gpg-connect-agent: waiting for the agent to come up ... (5s)
gpg-connect-agent: connection to agent established
[debug 15:47:20.553553] -  -------- ||| Here we switch virtual desktop
[debug 15:47:20.553553] - 0 .  "DP-2"  -  Plasma::Types::TopEdge
[debug 15:47:20.553553] - 1 .  "DP-2"  -  Plasma::Types::BottomEdge
[debug 15:47:20.553553] -  -------- sorted -----
[debug 15:47:20.553553] - 0 .  false  -  "DP-2"  -  Plasma::Types::BottomEdge
[debug 15:47:20.553553] - 1 .  false  -  "DP-2"  -  Plasma::Types::TopEdge
[...]

PR #1917, opened 2019-09-06
lib/gpg: Use g_spawn_sync to kill gpg-agent

dbnicholson:
For reasons I don't understand, GSubprocess doesn't play nice with KDE's
plasmashell. I assume this has something to do with the GSubprocess
using the glib worker thread while plasmashell uses the glib main
loop. Instead, just use g_spawn_sync to fork and wait in the current
thread.

Fixes: #1913

### DIFF
--- a/src/libotutil/ot-gpg-utils.c
+++ b/src/libotutil/ot-gpg-utils.c
@@ -509,31 +509,32 @@ ot_gpgme_kill_agent (const char *homedir)
     }
 
   /* Run gpg-connect-agent killagent /bye */
-  g_autoptr(GError) local_error = NULL;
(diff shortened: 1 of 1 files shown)


cgwalters:
> For reasons I don't understand, GSubprocess doesn't play nice with KDE's
plasmashell. I assume this has something to do with the GSubprocess
using the glib worker thread while plasmashell uses the glib main
loop. 

&lt;excited child in front of class raises hand&gt;  know I know!

It's because of `SIGCHLD` - the it's related to the GLib worker thread indeed.  Basically `g_spawn_sync` doesn't register a handler, but GSubprocess is more generic.  Theoretically perhaps it could be changed to only add a watch if the wait was async.

cgwalters:
Thanks a lot for tracking this down!

@rh-atomic-bot r+ 24015c1

PR #1918, opened 2019-09-09
fsck: Fix version in docs, tweak error text

cgwalters:
I think this error message is clearer.


jlebon:
@rh-atomic-bot r+ 5a6b2e5

PR #1919, opened 2019-09-13
prepare-root: remember to remove /sysroot.tmp

bgilbert:
Without this, rerunning `ostree-prepare-root` will fail in `mkdir()` because `/sysroot.tmp` already exists, which complicates debugging from the dracut emergency shell.


cgwalters:
@rh-atomic-bot r+ 225dd93

PR #1923, opened 2019-09-17
lib/gpg-verify: Fix out arg annotation

jlebon:
Was trying to use the GI bindings to verify a commit signature, and hit
this.

### DIFF
--- a/src/libostree/ostree-gpg-verify-result-dummy.c
+++ b/src/libostree/ostree-gpg-verify-result-dummy.c
@@ -202,7 +202,7 @@ ostree_gpg_verify_result_get_all (OstreeGpgVerifyResult *result,
  * ostree_gpg_verify_result_describe:
  * @result: an #OstreeGpgVerifyResult
  * @signature_index: which signature to describe
- * @output_buffer: a #GString to hold the description
+ * @output_buffer: (out): a #GString to hold the description
  * @line_prefix: (allow-none): optional line prefix string
  * @flags: flags to adjust the description format
  *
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-gpg-verify-result-dummy.c
```diff
@@ -202,7 +202,7 @@ ostree_gpg_verify_result_get_all (OstreeGpgVerifyResult *result,
  * ostree_gpg_verify_result_describe:
  * @result: an #OstreeGpgVerifyResult
  * @signature_index: which signature to describe
- * @output_buffer: a #GString to hold the description
+ * @output_buffer: (out): a #GString to hold the description
```
cgwalters: I don't think this is right; in Rust terms it's like a `&mut String`, in Java `StringBuilder` - it's not a return value from the function which is what `(out)` is for.

jlebon: OK yup, I should've read more carefully. What confused me is that with this patch, it Just Worked:

```
In [6]: r = repo.verify_commit_for_remote("0d56adacb518350906bc11b4978e7d45823a3b5439c24bb2b539b11a33bdd1cd", "tmpremote")

In [8]: s = r.describe(0, None, 0)

In [9]: s.str
Out[9]: 'Signature made Thu Sep 19 10:56:51 2019 using RSA key ID 936F29E9519CE313\nGood signature from "Jonathan Lebon <jlebon@redhat.com>"\n'
```

But yeah, *without* this patch, it does work to just allocate a `GString` and pass it in:

```
In [17]: r = repo.verify_commit_for_remote("0d56adacb518350906bc11b4978e7d45823a3b5439c24bb2b539b11a33bdd1cd", "tmpremote")

In [23]: s = GLib.string_new()

In [24]: r.describe(0, s, None, 0)

In [26]: s.str
Out[26]: 'Signature made Thu Sep 19 10:56:51 2019 using RSA key ID 936F29E9519CE313\nGood signature from "Jonathan Lebon <jlebon@redhat.com>"\n'
```

PR #1927, opened 2019-09-25
Release 2019.4

jlebon:
Tiny release. Just want to get out the important bugfixes instead of
backporting patches (notably the gpg-agent stuff and
`ostree-finalize-staged.service` ordering).


cgwalters:
@rh-atomic-bot r+ e72b11d

PR #1930, opened 2019-10-11
ci: Honor ARTIFACTS environment variable

cgwalters:
This is set by the OpenShift Prow pod-utils:
https://github.com/openshift/test-infra/blob/master/prow/pod-utilities.md

Prep for having OSTree use that.


jlebon:
@rh-atomic-bot r+ 5186b8d

PR #1931, opened 2019-10-14
libotutil: Port keyfile-utils.c to new style

cgwalters:
I'm seeing a rather inscrutable/convoluted error from clang-analyzer
when trying to rebase libostree:

```
Error: CLANG_WARNING: [#def2]
libostree-2019.2/src/libotutil/ot-keyfile-utils.c:197:3: warning: Potential leak of memory pointed to by 'value_list'
```

After staring at it for a bit, I *think* the problem is that since
GLib isn't part of this compilation unit, it doesn't know about
the "GLib error convention" such that `g_key_file_get_value()` always
returns `NULL` if an error is set.  If it didn't adhere to that
convention we'd leak the value.

I added a `g_assert()` in there.

That said in the process of tracing through this I found it
way easier to read as "new style" code, so this also ports the
code.


cgwalters:
(Also while looking at this I only realized after reading the report for the 3rd time it's from clang-analyzer, a FOSS project, not Coverity...so it's *way* easier to run manually, and we could also do so in CI)

cgwalters:
`test-pull-repeated.sh` flakes

jlebon:
> libostree-2019.2/src/libotutil/ot-keyfile-utils.c:197:3: warning: Potential leak of memory pointed to by 'value_list'

Hmm, that's in `ot_keyfile_get_string_list_with_separator_choice`, no?

Which... looking closer at it does look like we have a leak. I think we need:

```diff
diff --git a/src/libotutil/ot-keyfile-utils.c b/src/libotutil/ot-keyfile-utils.c
index 9d5903ce..8e6da535 100644
--- a/src/libotutil/ot-keyfile-utils.c
+++ b/src/libotutil/ot-keyfile-utils.c
@@ -154,7 +154,7 @@ ot_keyfile_get_string_list_with_separator_choice (GKeyFile      *keyfile,
   guint sep_count = 0;
   gchar sep = '\0';
   g_autofree char  *value_str = NULL;
-  g_autofree char **value_list = NULL;
+  g_auto(GStrv) value_list = NULL;

   g_return_val_if_fail (keyfile != NULL, FALSE);
   g_return_val_if_fail (section != NULL, FALSE);
@@ -215,9 +215,8 @@ ot_keyfile_get_string_list_with_default (GKeyFile      *keyfile,

   g_key_file_set_list_separator (keyfile, separator);

-  g_autofree char **ret_value = g_key_file_get_string_list (keyfile, section,
-                                                            key, NULL, &temp_error);
-
+  g_auto(GStrv) ret_value = g_key_file_get_string_list (keyfile, section,
+                                                        key, NULL, &temp_error);
   if (temp_error)
     {
       if (g_error_matches (temp_error, G_KEY_FILE_ERROR,
```

right?

cgwalters:
Yeah, I think you're right - want to submit a PR with just that for clarity, and I'll rebase this on top?

Note looking at this, I think we need one other change:
```
@@ -211,7 +211,7 @@ ot_keyfile_get_string_list_with_default (GKeyFile      *keyfile,
                            G_KEY_FILE_ERROR_KEY_NOT_FOUND))
         {
           g_clear_error (&temp_error);
-          ret_value = default_value;
+          ret_value = g_strdupv (default_value);
         }
       else
         {
```

cgwalters:
(Obligatory "this bug, patch, review and conversation would have never happened if this code was Rust" comment)

jlebon:
> Note looking at this, I think we need one other change:

Ahh yes, indeed. Added in https://github.com/ostreedev/ostree/pull/1939!

cgwalters:
Hm, looks like a ci-operator race.
/retest

cgwalters:
rpm-md repo failures never get old!
/retest

jlebon:
Hmm, not sure what's going on there:

```
 + ./configure --sysconfdir=/etc --prefix=/usr --libdir=/usr/lib64 --enable-gtk-doc --with-curl --enable-installed-tests=exclusive
ls: cannot access '.': Operation not permitted
configure: error: working directory cannot be determined
error: build error: running '/bin/bash -c set -o errexit; umask 0002; env SKIP_INSTALLDEPS=1 ./ci/build.sh' failed with exit code 2 
```

PR #1932, opened 2019-10-14
libostree: Add an assert to pacify clang-analyzer

cgwalters:
Got this error when trying to rebase libostree in RHEL:

```
Error: CLANG_WARNING: [#def1]
libostree-2019.2/src/libostree/ostree-repo-checkout.c:375:21: warning: Access to field 'disable_xattrs' results in a dereference of a null pointer (loaded from variable 'repo')
```

I think what's happening is it sees us effectively testing
`if (repo == NULL)` via the `while (current_repo)`.  Let's
tell it we're sure it's non-null right after the loop.


cgwalters:
Another `test-pull-repeated.sh` flake.

PR #1937, opened 2019-10-14
WIP: Add a zipl bootloader backend

cgwalters:
zipl is a bit special in that it parses the BLS config files
directly *but* we need to run the command to update the "boot block".

Hence, we're not generating a separate config file like the other
backends.  Instead, extend the bootloader interface with a `post_bls_sync`
method that is run in the same place we swap the `boot/loader` symlink.

We write a "stamp file" in `/boot` that says we need to run this command.
This opens the door to making things eventually-consistent/reconcilable
by later adding a systemd unit to run `zipl` if we're interrupted via
a systemd unit - I think we should eventually take this approach
everywhere rather than requiring `/boot/loader` to be a symlink.


cgwalters:
(Only compile tested!)

Prashanth684:
> (Only compile tested!)

i'd like to test these changes locally as part of creating an fcos/rhcos image on s390x. any idea how i can do that ?

cgwalters:
> i'd like to test these changes locally as part of creating an fcos/rhcos image on s390x. any idea how i can do that ?

See https://github.com/coreos/coreos-assembler/blob/master/README-devel.md#using-overrides

vtolstov:
can you share link to this kind of bootloader - zipl ?

cgwalters:
> can you share link to this kind of bootloader - zipl ?

See https://github.com/ibm-s390-tools/s390-tools/ for https://en.wikipedia.org/wiki/Linux_on_z_Systems

tuan-hoang1:
Given we make `*out_is_active = FALSE;` and specifically `set sysroot.bootloader zipl` in `create_disk.sh`, I think we still need this, right ?
```
diff --git a/src/libostree/ostree-repo.c b/src/libostree/ostree-repo.c
index b654aff2..94116290 100644
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3185,7 +3185,7 @@ reload_sysroot_config (OstreeRepo          *self,
      * https://github.com/ostreedev/ostree/issues/1719
      * https://github.com/ostreedev/ostree/issues/1801
      */
-    if (!(g_str_equal (bootloader, "auto") || g_str_equal (bootloader, "none")))
+    if (!(g_str_equal (bootloader, "auto") || g_str_equal (bootloader, "none") || g_str_equal (bootloader, "zipl")))
 
     g_free (self->bootloader);
```

cgwalters:
>     if (!(g_str_equal (bootloader, "auto") || g_str_equal (bootloader, "none") || g_str_equal (bootloader, "zipl")))

Oh right.  Thanks, I merged that bit and added the other bootloaders too for completeness.  This bit is ugly...need to rewrite the bootloader stuff, but let's do the easy change for now.

tuan-hoang1:
To summarize:
Colin's 542f550 commit : add a "zipl" check in libostree/ostree-repo.c
Colin's 5646da4 commit : add `_ostree_bootloader_zipl_new()` to create bootloader object for zipl case.
My last 2 commits : add a comment to make it clearer for why we do `_ostree_bootloader_zipl_new()` instead inside `libostree/ostree-sysroot-deploy.c`. Also update git commit message with some more details.

Colin @cgwalters, thanks a lot for quick turnaround ! This works perfectly well. I think we are good to go now.

tuan-hoang1:
Looks like there was some permission issue in ostree source tree ? I pulled your remote tree and make a branch on my local repo, maybe that's an issue ? Let me clone your tree then ...

cgwalters:
Repushed to retrigger CI.

src/libostree/ostree-repo.c
```diff
@@ -3184,8 +3184,13 @@ reload_sysroot_config (OstreeRepo          *self,
      * binary "x" in /usr/lib/ostree/bootloaders/x). See:
      * https://github.com/ostreedev/ostree/issues/1719
      * https://github.com/ostreedev/ostree/issues/1801
+     * Also, dedup these strings with the bootloader implementations
      */
-    if (!(g_str_equal (bootloader, "auto") || g_str_equal (bootloader, "none")))
+    if (!(g_str_equal (bootloader, "auto") || g_str_equal (bootloader, "none")
+          || g_str_equal (bootloader, "grub2")
+          || g_str_equal (bootloader, "syslinux")
+          || g_str_equal (bootloader, "U-Boot")
```
jlebon: If we allow these other ones too now, we should also handle them in `ostree_sysroot_write_deployments_with_options`, no?

cgwalters: Probably...it just grows into a bigger PR that has the potential to break other things and I'd like to fast track zipl support into the next release and circle back and clean up the bootloader bits after that.  Sound OK?

cgwalters: Per IRC chat, removed the non-zipl changes here.

PR #1938, opened 2019-10-14
OWNERS: New file

cgwalters:
I tried to balance reflecting the reality of who works on libostree
today with keeping some of the existing committers - particularly
committers from multiple organizations.

Part of switching libostree over to OpenShift Prow.


cgwalters:
Merging this manually to bootstrap us to OpenShift Prow!

PR #1939, opened 2019-10-15
src/libotutil: Fix strv memory leak

jlebon:
We were only freeing the array and not the members.

Caught by `clang-analyzer` in:
https://github.com/ostreedev/ostree/pull/1931


rfairley:
/lgtm

Thanks for catching this!

PR #1942, opened 2019-10-16
ci: Skip all yum operations if SKIP_INSTALLDEPS is set

cgwalters:
This is used by our OpenShift Prow job; we use the
cosa buildroot container:

https://github.com/coreos/coreos-assembler/pull/730

And using `yum` at all means we can flake on fetching rpm metadata.


cgwalters:
More travis flakes

PR #1943, opened 2019-10-16
More scan build

cgwalters:


cgwalters:
Repodata flakes
/override f29-flatpak

PR #1946, opened 2019-10-16
ostree-metadata commit API

dbnicholson:
There are 2 pieces of this PR:

* Add an `ostree_repo_regenerate_metadata` API that rebuilds the summary __and__ the `ostree-metadata` commit and signs them both. Currently the `ostree-metadata` commit is only created by the `ostree summary` CLI. This adds a way for callers to reliably generate it.

* ~~Add the local ref map to the `ostree-metadata` commit. This spawned out of discussion in https://lists.freedesktop.org/archives/flatpak/2019-October/001784.html and https://github.com/flatpak/flatpak/pull/3167#pullrequestreview-302032248. Essentially, it seems that since you can't trust the ref map in the peer's summary file, it should be included in the mirrored `ostree-metadata` commit like the additional summary metadata so that all you need from the peer's summary is a list of what mirrored refs it has.~~

~~I can split out the 2nd part, but I figured I'd put them both here for discussion.~~


alexlarsson:
I don't really understand the local ref map. Is this only about the flatpak xa.cache issue, or do you actually expect ostree to use this? If so, how? 

In a situation where you're pulling N refs from M different p2p remotes you'll end up pulling the ostree-metadata ref from *one* of those M different remotes. And the rest of the N refs from some "random" subset depending on what refs each node has in their respective summary file. If the ostree-metadata you end up with is more recen" than some of the other peers, do you expect to fail pulling those refs? That seems like a bad idea to me.

dbnicholson:
> I don't really understand the local ref map. Is this only about the flatpak xa.cache issue, or do you actually expect ostree to use this? If so, how?
> 
> In a situation where you're pulling N refs from M different p2p remotes you'll end up pulling the ostree-metadata ref from _one_ of those M different remotes. And the rest of the N refs from some "random" subset depending on what refs each node has in their respective summary file. If the ostree-metadata you end up with is more recent than some of the other peers, do you expect to fail pulling those refs? That seems like a bad idea to me.

The way I see it, you pull the ostree-metadata ref _and_ the summary file from _all_ of the p2p peers. That's how you reliably determine what subset of the remote each of them have. As I said in https://github.com/flatpak/flatpak/pull/3167#discussion_r335685468, since you can't verify the peer's summary file, the only source of verification it provides is the mirrored ostree-metadata file from the time when it mirrored from the remote. Having a copy of the remote's ref map in the ostree-metadata file allows you to verify that the commits in the peer are actually from the remote without fetching the commit objects from the peer.

However, I'm not the P2P expert.

alexlarsson:
> The way I see it, you pull the ostree-metadata ref _and_ the summary file from _all_ of the p2p peers. That's how you reliably determine what subset of the remote each of them have. As I said in [flatpak/flatpak#3167 (comment)](https://github.com/flatpak/flatpak/pull/3167#discussion_r335685468), since you can't verify the peer's summary file, the only source of verification it provides is the mirrored ostree-metadata file from the time when it mirrored from the remote. Having a copy of the remote's ref map in the ostree-metadata file allows you to verify that the commits in the peer are actually from the remote without fetching the commit objects from the peer.
> 
> However, I'm not the P2P expert.

Well, that is not how any ostree code is written today. The unique identified for the ref (i.e. the ref name + the collection id) can with the current repo layout only be stored once (as `refs/mirrors/$collection_id/$ref`), and it is pulled like any other ref, ie.  from the highest priority available peer. Now, in *theory* one could pull it from all, but that is not how any code today works.

But, the summary isn't really necessary to verify the the ref, because the commit itself is also signed, and it includes a timestamp so that we can know if the update is a downgrade or not. So, I don't see the point of verifying it with the original summary.

The reason that flapak needs this information is because it stores info related to the refs inside the ostree-metadata itself (rather than in the commit itself), and it then needs to know what commit this data is about. As such, having the commit map in the summary could help (although the form this has uses much more space than the solution i did for flatpak only, as it duplicates every ref in the table, as well as using asciified commit ids instead of raw and including all refs rather than the required ones only.

dbnicholson:
Given that the last commit does not have good agreement, I'll break that out into a separate PR and leave this just for the new API to handle updating the summary and ostree-metadata commit in one shot.

dbnicholson:
I rebased this on master now and dropped the "ref-map in ostree-metadata" commit. But I did add on an additional change I originally did in #490 to make the summary and summary.sig in a tmpdir before renaming. Although this doesn't solve a race (as #490 was trying to do), it does mean that you won't publish a summary without a matched summary.sig file.

src/libostree/ostree-repo.c
```diff
@@ -5944,9 +5920,82 @@ ostree_repo_regenerate_summary (OstreeRepo     *self,
...
+ * ostree_repo_regenerate_summary() and %OSTREE_SUMMARY_GVARIANT_FORMAT for
+ * additional details on its contents.
+ *
+ * Locking: exclusive
+ *
+ * Since: 2019.7
+ */
+gboolean
+ostree_repo_regenerate_metadata (OstreeRepo     *self,
+                                 GVariant       *additional_metadata,
```
jlebon: Maybe `summary_additional_metadata` to make it clear this goes in the summary file?

src/libostree/ostree-repo.c
```diff
@@ -5944,9 +5920,82 @@ ostree_repo_regenerate_summary (OstreeRepo     *self,
...
+                                 const gchar   **key_ids,
+                                 const gchar    *homedir,
+                                 GCancellable   *cancellable,
+                                 GError        **error)
```
jlebon: Whenever we add new extensible APIs, I lean towards adding a `GVariant *options` arg to help future proof it a bit more. WDYT?

src/libostree/ostree-repo.c
```diff
@@ -6009,22 +6009,58 @@ regenerate_metadata (OstreeRepo     *self,
...
+
+          return glnx_throw_errno_prefix (error, "%s",
+                                          "Unable to rename summary signature file");
+        }
+    }
+  else
+    {
+      g_debug ("Deleting existing unmatched summary.sig file");
+      if (!ot_ensure_unlinked_at (self->repo_dir_fd, "summary.sig", error))
+        return glnx_prefix_error (error, "Unable to delete summary signature file: ");
```
jlebon: A small improvement is hoisting this out of the if and up to before doing the `summary` rename. That way, if we're able to rename the `summary` but not the signature, we don't have to hope that we can remove the old one.

dbnicholson:
I decided to dust off this PR this week. It's mostly the same as before (I'll respond to the original comments above). The main difference is that I changed the API to take a `GVariant *options` where the GPG keys and/or sign keys can be specified. I don't think non-GPG signing existed when I first started this.

The other thing I did was make the `ostree-metadata` commit generation an explicit internal API. Back when flatpak was in the `ostree-metadata` business, @pwithnall suggested that there should be public API in ostree to create the commit. Several years later, flatpak has dropped it and I don't think there's any intention of bringing it back. I don't know if there are any other users of that commit that would benefit from a public API now.

src/libostree/ostree-repo.c
dbnicholson: I think since the metadata goes in both the summary and the `ostree-metadata` commit then that's not entirely true. But I don't mind if it changes name.

src/libostree/ostree-repo.c
dbnicholson: Excellent suggestion. I did that and dropped the explicit GPG parameters. Instead, you specify them in `options` with or without non-GPG signing keys. There are other things I think could be added in the future. For example, flatpak has an option to only list a subset of refs in the summary. I think that was mostly done because flathub's summary file was getting enormous.

src/libostree/ostree-repo.c
dbnicholson: But if you remove the old summary before writing the new one, then you're left with no summary signature if there are errors. I think it's preferable to leave it the old signature in place until you know the new summary has been written.

I suppose if you've specified that you're going to create the new summary without a signature, then you might as well remove the old one, though. Like:

```
if (!signing)
  remove_old_signature()
rename_new_summary()
if (signing)
  if (!rename_new_signature())
    remove_old_signature()
```

Is that what you meant? I think it makes sense either way, but I can see the advantage of doing it that way.

cgwalters (commented):
Only gave this a really superficial skim so far...having to context switch in a lot here :smile: 

Very superficially, seems sane.

src/libostree/ostree-repo.c
```diff
@@ -3525,16 +3525,16 @@ reload_sysroot_config (OstreeRepo          *self,
...
   {
     return glnx_throw (error, "Invalid bootloader configuration: '%s'", bootloader);
   }
   /* Parse bls-append-except-default string list. */
   g_auto(GStrv) read_values = NULL;
-  if (!ot_keyfile_get_string_list_with_default (self->config, "sysroot", "bls-append-except-default", 
+  if (!ot_keyfile_get_string_list_with_default (self->config, "sysroot", "bls-append-except-default",
                                                 ';', NULL, &read_values, error))
       return glnx_throw(error, "Unable to parse bls-append-except-default");
-
```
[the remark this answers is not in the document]
dbnicholson: Good point. I only noticed after the fact that my editor is now following the `.editorconfig` setting to do that.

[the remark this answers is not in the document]
dbnicholson: It seemed weird to only strip the trailing whitespace in the files I was touching, so I went all the way and added a prep commit that does all the C files. I can split that out to a separate PR if you prefer.

cgwalters (approved):
Nice!

PR #1953, opened 2019-10-23
async-progress: Plug memory leak while destroying GSource

uajain:
See https://gitlab.gnome.org/GNOME/glib/commit/71973c722

PR #1954, opened 2019-10-23
lib/pull: Avoid calling destroy on unref'ed GSource

jlebon:
We're creating the timer source and then passing ownership to the
context, but because we didn't free the pointer, we would still call
`g_source_destroy` in the exit path. We'd do this right after doing
`unref` on the context too, which would have already destroyed and
unref'ed the source.

Drop that and just restrict the scope of that variable down to make
things more obvious.

Just noticed this after reviewing #1953.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3573,7 +3573,6 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   g_autofree char **refs_to_fetch = NULL;
   g_autoptr(GVariantIter) collection_refs_iter = NULL;
   g_autofree char **override_commit_ids = NULL;
-  GSource *update_timeout = NULL;
   gboolean opt_gpg_verify_set = FALSE;
   gboolean opt_gpg_verify_summary_set = FALSE;
   gboolean opt_collection_refs_set = FALSE;
@@ -4505,6 +4504,8 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
 
   if (pull_data->progress)
     {
+      g_autoptr(GSource) update_timeout = NULL;
+
       /* Setup a custom frequency if set */
       if (update_frequency > 0)
         update_timeout = g_timeout_source_new (pull_data->dry_run ? 0 : update_frequency);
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -4505,6 +4504,8 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
 
   if (pull_data->progress)
     {
+      g_autoptr(GSource) update_timeout = NULL;
```
cgwalters: Wait, we're relying on that source staying active at least as long as we're in the main loop, right?

jlebon: Doesn't it take ownership of the source? I'm looking at https://gitlab.gnome.org/GNOME/glib/blob/master/glib/gmain.c#L1131.

jlebon: I guess another way to fix this is:

```diff
diff --git a/src/libostree/ostree-repo-pull.c b/src/libostree/ostree-repo-pull.c
index bcb8b420..429c76e5 100644
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3573,7 +3573,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   g_autofree char **refs_to_fetch = NULL;
   g_autoptr(GVariantIter) collection_refs_iter = NULL;
   g_autofree char **override_commit_ids = NULL;
-  GSource *update_timeout = NULL;
+  g_autoptr(GSource) update_timeout = NULL;
   gboolean opt_gpg_verify_set = FALSE;
   gboolean opt_gpg_verify_summary_set = FALSE;
   gboolean opt_collection_refs_set = FALSE;
@@ -4514,7 +4514,6 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       g_source_set_priority (update_timeout, G_PRIORITY_HIGH);
       g_source_set_callback (update_timeout, update_progress, pull_data, NULL);
       g_source_attach (update_timeout, pull_data->main_context);
-      g_source_unref (update_timeout);
     }

   /* Now await work completion */
```

so we actually keep a ref in the out path. Though AFAICT, we should be able to let the context take care of it for us.

cgwalters: I always found sources being "destroyed" confusing relative to their unref state.  I *think* it's required (or just best practice) to also destroy the source?  See e.g. https://gitlab.gnome.org/GNOME/glib/blob/d515a1e85a2d9c81c37024fb1b47dfeed5f50eb6/glib/gmain.c#L2100
But maybe that code only kicks in if something else went wrong?

cgwalters:
OK I think this is right indeed.  The maincontext does destroy any attached sources.
/lgtm

PR #1955, opened 2019-10-24
Revert grub2 exit, add new grub2: Honor /boot/.grub2-bls-enabled 

cgwalters:
In this case this is reverted in Fedora; I don't like carrying downstream patches.  Let's revert it here, and I added a different approach.

### DIFF
--- a/src/boot/grub2/grub2-15_ostree
+++ b/src/boot/grub2/grub2-15_ostree
@@ -26,13 +26,6 @@ if ! test -d /ostree/repo; then
     exit 0
 fi
 
-# Gracefully exit if the grub2 configuration has BLS enabled,
-# since there is no need to create menu entries in that case.
-. /etc/default/grub
-if test ${GRUB_ENABLE_BLSCFG} = "true"; then
-    exit 0
-fi
-
 # Make sure we're in the right environment
 if ! test -n "${GRUB_DEVICE}"; then
     echo "This script must be run as a child of grub2-mkconfig" 1>&2


martinezjavier:
@cgwalters I just had a trivial comment, but other than that I agree with your changes.

cgwalters:
> @cgwalters I just had a trivial comment, but other than that I agree with your changes.

What's the comment?  I don't see it here?

src/boot/grub2/grub2-15_ostree
```diff
@@ -26,10 +26,12 @@ if ! test -d /ostree/repo; then
...
-# Gracefully exit if the grub2 configuration has BLS enabled,
-# since there is no need to create menu entries in that case.
-. /etc/default/grub
-if test ${GRUB_ENABLE_BLSCFG} = "true"; then
+# This is a special file which can be opt-in created to
+# say that the installed grub2 speaks BLS.  See
+# commit 985a14100295c99d0c6d712bfbee0ec52a3a1601
+# https://discussion.fedoraproject.org/t/boot-entries-gone-after-upgrade/8026/3
+# https://github.com/ostreedev/ostree/pull/1929
+if test -f /boot/.grub2-bls-enabled; then
```
jlebon: Hmm, in the idea I had in https://github.com/ostreedev/ostree/pull/1929#issuecomment-535618871, I suggested checking for *both* `GRUB_ENABLE_BLSCFG` and a stamp file that that means "this bootloader supports blscfg" (so probably should be named e.g. `.grub2-bls-supported`).

With this approach, if someone turns off `GRUB_ENABLE_BLSCFG` again, we go back to no entries. Not that folks would normally want to do that, but seems worth handling it given the consequences. (One case that comes to mind for example is someone rolling back to a previous deployment that doesn't enable `GRUB_ENABLE_BLSCFG`.)

martinezjavier:
> What's the comment? I don't see it here?

Sorry, it's now. I noticed that never submitted my comment.

src/boot/grub2/grub2-15_ostree
martinezjavier: > With this approach, if someone turns off `GRUB_ENABLE_BLSCFG` again, we go back to no entries. Not that folks would normally want to do that, but seems worth handling it given the consequences. (One case that comes to mind for example is someone rolling back to a previous deployment that doesn't enable `GRUB_ENABLE_BLSCFG`.)

@jlebon I thought that @cgwalters idea here was to not allow the user to choose but instead let the installer / OS to specify whether the installed bootloader has BLS  support or not and just use it if that's the case.

That way you won't use a config option that's specific to Fedora GRUB but instead a protocol that's defined by OSTree that bootloaders can use to specify that are able to parse BLS files.
But you are right too that maybe users should be allowed to disable the option.

Probably this PR should be split in two, so the revert can be merged to drop the downstream patch while the second commit is discussed.

cgwalters:
Updated the PR to just revert

cgwalters:
bot, retest this please

PR #1956, opened 2019-10-24
tests/export: Guard with check for libarchive

akiernan:
If we are built without libarchive support, this test fails:

  error: This version of ostree is not compiled with libarchive support
  ...
  ERROR: tests/test-export.sh - too few tests run (expected 5, got 0)
  ERROR: tests/test-export.sh - exited with status 1


jlebon:
bot, retest this please

PR #1957, opened 2019-10-24
Upstream Endless sizes metadata changes

dbnicholson:
At Endless we've been using the `ostree.sizes` metadata for years in [eos-updater](https://github.com/endlessm/eos-updater/) to tell users the download size of the installation. I realize that static deltas provide the same information, but deltas are not always available. However, since the sizes data is opt-in, others don't need to use it.

Unfortunately, we have some downstream changes that we'd like to get upstream.

1. We include all object types instead of just files so that the metadata accurately represents all the objects in a commit.
2. We added an additional byte to the `ostree.sizes` `ay` entries that represents the object type. This is used when reading out the metadata to see if the repo already has that object. That object is then removed from the download stats.
3. We added an API that reads the metadata and summarizes. It sums up the sizes from all the entries, checking if each object is already in the local repo. This requires 2.

I think 1 and 3 are not controversial, but 2 might be. If any readers of `ostree.sizes` are validating the length of each entry, they may error on the additional byte. However, since the entry array already ends in 2 variable length integers, the length of the array is not fixed. Also, I checked through a github code search and I'm pretty sure the only users of this data are the ostree test suite, eos-updater, and very old versions of xdg-app (pre-flatpak).

So, I think it's likely that these changes will not cause any issues, but I have no proof of that. Anyways, let me know what you think.


dbnicholson:
CC @pwithnall who maintains eos-updater.

src/libostree/ostree-repo-commit.c
```diff
@@ -408,18 +453,176 @@ add_size_index_to_metadata (OstreeRepo        *self,
...
+ * @error: a #GError
+ *
+ * Reads the size data for the commit stored in the %ostree.sizes key in
+ * the commit metadata. If this data is not available, %FALSE is
+ * returned and @error is set to %G_IO_ERROR_NOT_FOUND.
+ *
+ * Returns: %TRUE on success, %FALSE on failure
+ *
+ * Since: 2019.5
+ */
```
dbnicholson: This is the API as we have it in Endless, but I feel like it's not quite right. What this function actually returns is a summary of the `ostree.sizes` data, so something like `ostree_repo_get_commit_size_summary()` might make more sense. Also, the 6 out parameters seems excessive. Should this be a struct?

This API only returns a summary, but there's nothing that gets you the entries. Parsing the entries manually is tricky because of the variable length integers, and `_ostree_read_varuint64()` is not exposed publicly. I think an `ostree_repo_read_commit_sizes()` returning an array of `OstreeContentSizeCacheEntry` (possibly renamed) would also be nice so that the program could like through all the entries instead of just getting the summary.

cgwalters:
Haven't looked at the code yet but
/approve
the idea.
/hold
since this needs some deeper review.
/meow
just because

src/libostree/ostree-repo-commit.c
cgwalters: I'd name this something more like `ostree_repo_get_missing_object_sizes()` or so.

Having a struct instead of the out parameters makes sense to me.

I wonder whether it'd be more ergonomic though for this to be part of `OstreeAsyncProgress` - during a pull if e.g. `OSTREE_REPO_PULL_FLAG_CALCULATE_MISSING_SIZES` is passed, we provide all of this data as part of the progress?

All callers who want this would have to basically do a pull to get the commit, then call this API, then determine if they wanted to resume the pull, right?

dbnicholson: > I'd name this something more like `ostree_repo_get_missing_object_sizes()` or so.

I can't think of any really great names, although the `missing` is a bit of a misnomer because it returns both object size information as well as the number of objects in both total and local counts. That's why I suggested `get_commit_size_summary`, but this is almost certainly going to be used get the size of the objects you don't have. Maybe `get_commit_object_summary` since it's not just sizes. Feel strongly one way or the other?

> Having a struct instead of the out parameters makes sense to me.

:+1:

> I wonder whether it'd be more ergonomic though for this to be part of `OstreeAsyncProgress` - during a pull if e.g. `OSTREE_REPO_PULL_FLAG_CALCULATE_MISSING_SIZES` is passed, we provide all of this data as part of the progress?
> 
> All callers who want this would have to basically do a pull to get the commit, then call this API, then determine if they wanted to resume the pull, right?

That's interesting. The way it's used in our updater now is that we always do a metadata only pull and then use the API to get statistics about the full pull. That data is then exported over DBus to callers that want to present it like the About page in control center and possibly in GNOME software.

@pwithnall any thoughts about what this API should look like?  What's in the PR is what eos-updater is using now.

dbnicholson: How about this. Instead of figuring out the best way to use this summarizing API right now, I just add an API that parses the entries and returns them in an array? Then the caller can do what they want since the summarizing part is simple once you can read the entries but application specific. I think that the ability to parse the checksums, variable length integers and object type are needed regardless.

Over in eos-updater we can inline the calculation it needs (it ignores the number of object out params now anyways) after calling the new entries parsing API to get an array of entries.

pwithnall: Sorry I’m late to the party. From my point of view, either is fine. Adding an API to OSTree to just return the metadata seems like a reasonable compromise if people can’t agree on semantics for things otherwise. From the eos-updater point of view, the important things are:
 * Not relying on custom downstream libostree API
 * Not ending up with a different, forked implementation of functionality which exists upstream (getting commit sizes)

So going with a metadata-only API would be fine as long as we didn’t then deviate on how to store and calculate the summaries from it.

dbnicholson: Currently eos-updater doesn't store anything about the summaries. It just passes the size totals out as DBus properties. And the size totals are simple to calculate - just loop over each of the entries, add the sizes for that entry to the "full" totals, check if the local repo has the object in the entry, add the sizes to the "needed" totals if the object doesn't exist locally.

To me the important part is that we get the changes to the commit metadata upstream:

* Include all objects in the sizes array
* Include the object type in each size entry so you can reliably determine if you have the object corresponding to the entry or not

Once the metadata has this information and it can be parsed easily, what to do with it is really application specific.

dbnicholson:
> Overall code looks good to me. Bigger picture...if we could start over this would all probably be redesigned, because having the commit object contain a list of all objects makes having the tree structure (dirtree objects) somewhat pointless - it probably would have just been easier to have a single big metadata object. But oh well.

I've thought about this several times, and it does not have its downsides. The sizes metadata for our OS commits are big, and our commit objects are on the order of several MB. To the point that when I want to look at the commit metadata in python I first use GVariant to remove the array before letting pygobject deserialize the metadata variant into native types. I've considered that maybe you could put this object manifest somewhere else, but sooner or later you have to fetch it. So, the tree structure does have its benefits.

The corollary is that static deltas pretty much fit the bill here - a packed format that only contains the objects that aren't in the commit you're deriving the target from. The downsides being that you have a combinatorial explosion of deltas to create and that the delta might have some objects that you already have from a different commit that's not the source commit.

src/libostree/ostree-repo-commit.c
cgwalters: > That's why I suggested get_commit_size_summary

I am having a mental hurdle using `summary` for that since it's not related to the summary file.  `get_commit_size_info()`?

cgwalters:
I have a lingering concern that someone else out there is using the sizes metadata...hm, probably any code parsing the previous data would fatally error with the new data?  But I guess since Endless has been shipping with this we can't easily e.g. rename the metadata?  I hesitate to say it but maybe this should be a build time option?

That said I'm also willing to just merge it and hope that anyone else out there using the size metadata would be able to adapt, if they exist.

src/libostree/ostree-repo-commit.c
```diff
@@ -408,18 +453,176 @@ add_size_index_to_metadata (OstreeRepo        *self,
             g_hash_table_lookup (self->object_sizes, e_checksum);
           _ostree_write_varuint64 (buffer, e_size->archived);
           _ostree_write_varuint64 (buffer, e_size->unpacked);
+          g_string_append_c (buffer, (gchar) e_size->objtype);
```
cgwalters: The core incompatible change being this.

dbnicholson: Indeed. However, as I stated in a couple other comments, I really doubt people are using it. Primarily because there's no exported way to read the variable length integers. You'd have to duplicate the code from ostree as you can see in this old xdg-app commit that removes its usage - https://github.com/alexlarsson/xdg-app/commit/be71ceb31f715074d10d4c4e37e39ba72f0c7549.

I did a github code search on `ostree.sizes`, which unfortunately is sorta useless because it ignores the `.`. However, I trawled through [many pages of it](https://github.com/search?q=ostree.sizes&type=Code) and the only actual references were copies of ostree itself (including ours which expands on upstream) and xdg-app (as noted above).

Furthermore, because of the variable length integers, no code could have been asserting a fixed length. What could be done is asserting that there were no more bytes in the array after reading the 2nd integer. I didn't see that anywhere, though.

So, I think it is understandable to want to use a different key. In fact, that's how I was first approaching this. However, after reviewing the facts above, I came to the conclusion that it was highly unlikely that this would break anything besides the ostree testsuite.

cgwalters: OK yep, let's get this in then!

dbnicholson:
I've updated this now with a lower level API providing an array of the parsed `ostree.sizes` entries rather than the higher level summary API that was there before. You can see it being used in the last commit. I think that's simple enough personally. One thing is that I didn't know the best place to put it. Since it was operating on purely commit objects, I put it in core.

Unfortunately, I get a bunch of test failures locally after rebasing on master. In particular, 4d17cd917fe4fb67d2f9e4881b6693ac7289f423 seems to be running tests on my system in Catalan (ca_AD.utf8).

dbnicholson:
> Unfortunately, I get a bunch of test failures locally after rebasing on master. In particular, [4d17cd9](https://github.com/ostreedev/ostree/commit/4d17cd917fe4fb67d2f9e4881b6693ac7289f423) seems to be running tests on my system in Catalan (ca_AD.utf8).

#1991 seems to fix that.

src/libostree/ostree-repo-commit.c
dbnicholson: I dropped this API in the most recent changes. Instead, there's a lower level API that just parses the metadata and returns an array representing it. The `show` builtin has an example of summing the metadata as this API was doing before. It's pretty trivial and what I plan to do in eos-updater (except that it doesn't care about all the sums, so parts of it can be left out).

jlebon:
bot, retest this please

dbnicholson:
Thanks guys!

PR #1959, opened 2019-10-25
Initial fs-verity support

cgwalters:
Using [fs-verity](https://www.kernel.org/doc/html/latest/filesystems/fsverity.html) is natural for OSTree because it's file-based,
as opposed to block based (like [dm-verity](https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/verity.html)).  This only covers
files - not symlinks or directories.  And we clearly need to
have integrity for the deployment directories at least.

More background on fs-verity:

- https://lwn.net/Articles/763729/
- https://lwn.net/Articles/752614/
    
So making this truly secure would need a lot more work.  Nevertheless,
I think it's time to start experimenting with it.  Among other things,
it does *finally* add an API that makes files immutable, which will
help against some accidental damage.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -32,6 +32,10 @@
 #include <glib/gprintf.h>
 #include <sys/ioctl.h>
 #include <linux/fs.h>
+#include <ext2fs/ext2_fs.h>
+#ifdef HAVE_LINUX_FSVERITY_H
+#include <linux/fsverity.h>
+#endif
 
 #include "otutil.h"
 #include "ostree.h"
@@ -168,6 +172,113 @@ ot_security_smack_reset_fd (int fd)
 #endif
 }
 
+/* Wrapper around the fsverity ioctl, compressing the result to
+ * "success, unsupported or error".  This is used for /boot where
+ * we enable verity if supported.
+ * */
+gboolean
+_ostree_tmpf_fsverity_core (GLnxTmpfile *tmpf,
+                            _OstreeFeatureSupport fsverity_requested,
+                            gboolean    *supported,
+                            GError     **error)
+{
+  /* Set this by default to simplify the code below */
+  if (supported)
+    *supported = FALSE;
+
+  if (fsverity_requested == _OSTREE_FEATURE_NO)
+    return TRUE;
+
+#ifdef HAVE_LINUX_FSVERITY_H
+  GLNX_AUTO_PREFIX_ERROR ("fsverity", error);
+
+  /* fs-verity requires a read-only file descriptor */
+  if (!glnx_tmpfile_reopen_rdonly (tmpf, error))
+    return FALSE;
+
+  struct fsverity_enable_arg arg = { 0, };
+  arg.version = 1;
+  arg.hash_algorithm =  FS_VERITY_HASH_ALG_SHA256;  /* TODO configurable? */
+  arg.block_size = 4096; /* FIXME query */
+  arg.salt_size = 0; /* TODO store salt in ostree repo config */
+  arg.salt_ptr = 0;
+  arg.sig_size = 0; /* We don't currently expect use of in-kernel signature verification */
+  arg.sig_ptr = 0;
+
+  if (ioctl (tmpf->fd, FS_IOC_ENABLE_VERITY, &arg) < 0)
+    {
+      switch (errno)
+        {
+          case ENOTTY:
+          case EOPNOTSUPP:
+            return TRUE;
+          default:
+            return glnx_throw_errno_prefix (error, "ioctl(FS_IOC_ENABLE_VERITY)");
+        }
+    }
+  
+  if (supported)
+    *supported = TRUE;
+#endif
+  return TRUE;
+}
+
+/* Enable verity on a file, respecting the "wanted" and "supported" states.
+ * The main idea here is to optimize out pointlessly calling the ioctl()
+ * over and over in cases where it's not supported for the repo's filesystem,
+ * as well as to support "opportunistic" use (requested and if filesystem supports).
+ * */
+gboolean
+_ostree_tmpf_fsverity (OstreeRepo  *self,
+                       GLnxTmpfile *tmpf,
+                       GError    **error)
+{
+#ifdef HAVE_LINUX_FSVERITY_H
+  g_mutex_lock (&self->txn_lock);
+  _OstreeFeatureSupport fsverity_wanted = self->fs_verity_wanted;
+  _OstreeFeatureSupport fsverity_supported = self->fs_verity_supported;
+  g_mutex_unlock (&self->txn_lock);
+
+  switch (fsverity_wanted)
+    {
+      case _OSTREE_FEATURE_YES:
+        {
+          if (fsverity_supported == _OSTREE_FEATURE_NO)
+            return glnx_throw (error, "fsverity required but filesystem does not support it");
+        }
+        break;
+      case _OSTREE_FEATURE_MAYBE:
+        break;
+      case _OSTREE_FEATURE_NO:
+        return TRUE;
+    }
+
+  gboolean supported = FALSE;
+  if (!_ostree_tmpf_fsverity_core (tmpf, fsverity_wanted, &supported, error))
+    return FALSE;
+
+  if (!supported)
+    {
+      if (G_UNLIKELY (fsverity_wanted == _OSTREE_FEATURE_YES))
+        return glnx_throw (error, "fsverity required but filesystem does not support it");
+
+      /* If we got here, we must be trying "opportunistic" use of fs-verity */
+      g_assert_cmpint (fsverity_wanted, ==, _OSTREE_FEATURE_MAYBE);
+      g_mutex_lock (&self->txn_lock);
+      self->fs_verity_supported = _OSTREE_FEATURE_NO;
+      g_mutex_unlock (&self->txn_lock);
+      return TRUE;
+    }
+  
+  g_mutex_lock (&self->txn_lock);
+  self->fs_verity_supported = _OSTREE_FEATURE_YES;
+  g_mutex_unlock (&self->txn_lock);
+#else
+  g_assert_cmpint (self->fs_verity_wanted, !=, _OSTREE_FEATURE_YES);
+#endif
+  return TRUE;
+}
+
 /* Given an O_TMPFILE regular file, link it into place. */
 gboolean
 _ostree_repo_commit_tmpf_final (OstreeRepo        *self,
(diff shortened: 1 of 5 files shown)


cgwalters:
I think Android is basically going to continue using dm-verity for the base OS, and use fs-verity just for apps which come as single .zip files.

So they don't have the problem of verifying full filesystem trees.

I was thinking about this some more, and a rather gross hack might be to store all of our directories/symlinks as effectively tarballs in the real root, then on boot we unpack them into a tmpfs, and use overlayfs with the tmpfs and the underlying fs-verity files.

But clearly what we really want is an extension to fs-verity that lets us "seal" a directory and set of symlinks in it too.

cgwalters:
Continuing my half-baked ideation...

Maybe we can use a dual-partition dm-verity to hold the directories/symlinks, and an overlayfs pointing to the underlying real rootfs with fs-verity files?  

But...what's bugging me here is that if an attacker gains `CAP_SYS_ADMIN` and the ability to write directly to the underlying block device, then on reboot, even though the *files* are fs-verity protected, all of the rest of the disk (starting from the root directory) isn't.  And it's known that [mounting malicious filesystems is unsafe](https://lwn.net/Articles/755593/).  I wonder if the fs-verity developers (or potential Android/ChromeOS use of it) have considered this.

cgwalters:
Also, one thing we can and should do is use fs-verity for `/boot` - this PR needs to do the same thing for `/boot` that it does for `/objects` (at least when `/boot` is a separate FS).  For Fedora CoreOS we could probably turn on `ext4+verity` for `/boot` by default even if we don't make the switch for `/`.  That could at least help us prove out some of the secure/trusted boot bits combining with fs-verity even if we aren't doing it for the rest of userspace yet.

cgwalters:
OK a number of tweaks, biggest is that we now also do fsverity for files in `/boot`.

Also now:
Depends: https://gitlab.gnome.org/GNOME/libglnx/merge_requests/11

cgwalters:
I think the remaining files to cover are the config file (circular, need to write it then verity it), and the origin file.

cgwalters:
Side note: I just verified (heh) that fs-verity is perfectly happy to run without any privileges.  This means that ostree's fsverity support would work just fine for flatpak user installs.  It'd work for random config files like `~/.bashrc`.  

I'm excited about the potential of fs-verity!

cgwalters:
Lifted the WIP from this - it definitely works and does something useful today: makes files immutable.  I particularly want to do this for FCOS for `/boot` to start since we don't have the readonly bind mount over `/boot`.

The question is; is that worth merging as is?  I'm uncertain.  I think it's quite likely we'll need a "sign this file" callback.  And...looking at that, we have the big decision whether to import [the existing C code](https://git.kernel.org/pub/scm/linux/kernel/git/ebiggers/fsverity-utils.git/tree/cmd_sign.c) or to execute it as a subprocess (or, try to get a library made out of it upstream)  (Or, git submodule it).

And, none of the signing will really be worth much until we implement a full plan per above.

And, actually need support for shipping the signatures out of band too.  That's going to be hard to do elegantly while retaining backwards compatibility.  Bigger picture, if one can *assume* verity, then we don't need the "basic OSTree sha256" anymore...which argues for thinking of this more like a fully separate "mode" for OSTree with its own repository format...

cgwalters:
I updated this to disable even the "opportunistic" use of fs-verity (previously it'd only happen if you happened to `mkfs.ext4 -O verity` which isn't the default so extremely unlikely to do accidentally), but at least now it's totally off unless one sets the repo config flag to require it.

(That said also, we need to care for the case that `/boot` is of a different filesystem type)

jlebon:
OK, so confirm here, without the "measure" step, this is essentially a stronger version of `chattr +i`, right? And this patch also acts as a stepping stone towards when we do figure out integrity checking?

cgwalters:
> OK, so confirm here, without the "measure" step, this is essentially a stronger version of chattr +i, right? 

Stronger in some ways, but also weaker in others; e.g. it *does* allow `unlink()` and `link()` which we really want.  It also does allow `chmod u+s` and `chown foo:bar` which we don't...

> And this patch also acts as a stepping stone towards when we do figure out integrity checking?

I think a bottom line is this is useful as a starting point.

src/libostree/ostree-repo-commit.c
```diff
@@ -168,6 +172,74 @@ ot_security_smack_reset_fd (int fd)
...
+                /* If we got here, we must be trying "opportunistic" use of fs-verity */
+                g_assert_cmpint (fsverity_enabled, ==, _OSTREE_FEATURE_MAYBE);
+                g_mutex_lock (&self->txn_lock);
+                self->fs_verity = _OSTREE_FEATURE_NO;
+                g_mutex_unlock (&self->txn_lock);
+                return TRUE;
+              }
+            else
+              {
+                return glnx_throw_errno_prefix (error, "(config required) ioctl(FS_IOC_ENABLE_VERITY)");
```
jlebon: Or let's just throw the same error as we would've above? `fsverity required but filesystem does not support it`

jlebon: Minor: could reorg this to return upfront and de-indent

src/libostree/ostree-repo.c
```diff
@@ -3033,6 +3035,10 @@ reload_core_config (OstreeRepo          *self,
       }
   }
 
+  if (!ot_keyfile_get_boolean_with_default (self->config, "fsverity", "required",
```
jlebon: Maybe `ex-fsverify` to mark it experimental for now?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -110,32 +112,59 @@ install_into_boot (OstreeSePolicy *sepolicy,
...
-           * non-SELinux xattrs for the kernel anyways aside from perhaps
-           * IMA but that's its own story.
-           */
-          g_auto(OstreeSepolicyFsCreatecon) fscreatecon = { 0, };
-          const char *boot_path = glnx_strjoina ("/boot/", glnx_basename (dest_subpath));
-          if (!_ostree_sepolicy_preparefscreatecon (&fscreatecon, sepolicy,
-                                                    boot_path, S_IFREG | 0644,
-                                                    error))
-            return FALSE;
-          return glnx_file_copy_at (src_dfd, src_subpath, NULL, dest_dfd, dest_subpath,
```
jlebon: Hmm, tempting to just fold the fsverity bit into libglnx and just make it another flag.

cgwalters: The "hardlink or copy"?  Yeah...but we also have the bits around SELinux here.

src/libostree/ostree-repo.c
cgwalters: Yeah, makes sense.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -113,32 +115,59 @@ install_into_boot (OstreeSePolicy *sepolicy,
...
+  if (fchown (tmp_dest.fd, src_stbuf.st_uid, src_stbuf.st_gid) != 0)
+    return glnx_throw_errno_prefix (error, "fchown");
+
+  if (fchmod (tmp_dest.fd, src_stbuf.st_mode & 07777) != 0)
+    return glnx_throw_errno_prefix (error, "fchmod");
+
+  if (fdatasync (tmp_dest.fd) < 0)
+    return glnx_throw_errno_prefix (error, "fdatasync");
+
+  if (!_ostree_tmpf_fsverity_core (&tmp_dest, NULL, error))
```
jlebon: Shouldn't this be `_ostree_tmpf_fsverity` so it respects the settings?

src/libostree/ostree-sysroot-deploy.c
jlebon: I mean folding `_ostree_tmpf_fsverity_core` into libglnx, and exposing it via e.g. `GLNX_FILE_COPY_FSVERITY`. We can look at that later though!

src/libostree/ostree-sysroot-deploy.c
cgwalters: Yeah...the problem is that `/boot` may be a different filesystem type.   The repo is trying to cache the supported state for the repo.  Messy...

I tweaked this so we don't enable verity for `/boot` if we're not using it for `/` - but currently if it's `required` by `/` we downgrade to "opportunistic" for `/boot`.

cgwalters:
To expand on this:

> I was thinking about this some more, and a rather gross hack might be to store all of our directories/symlinks as effectively tarballs in the real root (that are fs-verity protected), then on boot we unpack them into a tmpfs, and use overlayfs with the tmpfs and the underlying fs-verity files.

It may help to only separate files with nontrivial size (maybe > 1k?).  Also with this it may be simplest to have one big "verity pack file" for the initial deployment, then create small "delta pack files" for updates.  Then at some point we do a "repack"?

Also related to signatures and such, in this model we'd need to verify the signature on the verity pack files as well as every "loose file", since we can't rely on overlayfs to do that for us.

OR we try to go with the in-kernel verity signature support.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -113,32 +115,67 @@ install_into_boot (OstreeSePolicy *sepolicy,
...
+
+  /* Today we don't have a config flag to *require* verity on /boot,
+   * and at least for Fedora CoreOS we're not likely to do fsverity on
+   * /boot soon due to wanting to support mounting it from old Linux
+   * kernels.  So change "required" to "maybe".
+   */
+  _OstreeFeatureSupport boot_verity = _OSTREE_FEATURE_NO;
+  if (repo->fs_verity_wanted != _OSTREE_FEATURE_NO)
+    boot_verity = _OSTREE_FEATURE_MAYBE;
+  if (!_ostree_tmpf_fsverity_core (&tmp_dest, boot_verity, NULL, error))
```
jlebon: Won't this cause one `ioctl` each time even if it's not supported? Hmm, how about we add an optional `fsverity_wanted` parameter to `_ostree_tmpf_fsverity` that overrides the one from `OstreeRepo`? Then we could use it here too and get the automatic no-op.

Ohh, but right the kernel might support it, but not the filesystem on `/boot`. Hmm, so probably instead we should have a `boot_fsverity` in the `OstreeSysroot` object? But ehhh, this is good as is too. The number of files we install in `/boot` pales in comparison to `/` so it's a small optimization we could do later.

cgwalters: Yeah, an extra syscall or two for this doesn't matter much, but I try to keep the total small since I sometimes just resort to `strace -f` - we have a lot more objects than kernels.

PR #1960, opened 2019-10-28
Keep last version of deployment

r4f4:
Putting this up for initial review while I try to work on adding tests.
This PR is implementing the solution proposed in https://github.com/coreos/rpm-ostree/issues/1905.


jlebon:
bot, add author to whitelist

src/libostree/ostree-sysroot.c
```diff
@@ -1645,22 +1653,52 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
...
+          metadata = g_variant_get_child_value (variant, 0);
+          g_variant_lookup (metadata, OSTREE_COMMIT_META_KEY_VERSION, "s", &deployment_version);
+        }
+
+      if ((is_booted || is_merge) && !last_version)
+        last_version = g_steal_pointer (&deployment_version);
+
+      const gboolean is_previous_version =
+          (osname_matches && last_version && passed_crossover &&
+           (g_strcmp0 (deployment_version, last_version) < 0));
```
jlebon: Should be `!= 0` here, no?

src/libostree/ostree-sysroot.c
```diff
@@ -1645,22 +1653,52 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
...
           || !osname_matches
           || (retain_pending && !passed_crossover)
           || (is_booted || is_merge)
-          || (retain_rollback && passed_crossover))
+          || (retain_rollback && passed_crossover)
+          || (retain_previous && is_previous_version && !n_previous))
         g_ptr_array_add (new_deployments, g_object_ref (deployment));
 
+      /* Just keep one previous version */
+      n_previous += (guint8)is_previous_version;
```
jlebon: Hmm, how about just a `retained_previous_version` boolean flag instead to detect that we've already done it?

src/ostree/ot-admin-builtin-deploy.c
```diff
@@ -174,7 +174,7 @@ ot_admin_builtin_deploy (int argc, char **argv, OstreeCommandInvocation *invocat
       if (opt_not_as_default)
         return glnx_throw (error, "--stage cannot currently be combined with --not-as-default");
       if (!ostree_sysroot_stage_tree (sysroot, opt_osname, revision, origin, merge_deployment,
-                                      kargs_strv, &new_deployment, cancellable, error))
+                                      kargs_strv, &new_deployment, TRUE, cancellable, error))
```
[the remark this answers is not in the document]
r4f4: Ok. Done.

src/libostree/ostree-sysroot.c
```diff
@@ -1645,22 +1653,52 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
...
+      const gchar *csum = ostree_deployment_get_csum (deployment);
+      g_autoptr(GVariant) variant = NULL;
+      g_autoptr(GVariant) metadata = NULL;
+
+      g_autofree char *deployment_version = NULL;
+
+      OstreeRepo *repo = ostree_sysroot_repo (sysroot);
+      /* XXX Copying ot_admin_checksum_version() + bits from
+       *     ot-admin-builtin-status.c.  Maybe this should be
+       *     public API in libostree? */
```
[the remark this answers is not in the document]
r4f4: Sure

src/libostree/ostree-sysroot.c
```diff
@@ -1645,22 +1653,52 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
...
+      if (ostree_repo_load_variant (repo, OSTREE_OBJECT_TYPE_COMMIT, csum,
+                                    &variant, NULL))
+        {
+          metadata = g_variant_get_child_value (variant, 0);
+          g_variant_lookup (metadata, OSTREE_COMMIT_META_KEY_VERSION, "s", &deployment_version);
```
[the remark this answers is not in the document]
r4f4: Ok

src/libostree/ostree-sysroot.c
r4f4: Ok

src/libostree/ostree-sysroot.c
r4f4: Yes, the deployments are already ordered.

src/libostree/ostree-sysroot.c
```diff
@@ -1645,22 +1678,41 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
...
+      if (is_booted || is_merge)
+        {
+          g_autofree char *deployment_version =
+            ostree_deployment_get_version (sysroot, deployment);
+
+          if (!last_version)
+            last_version = g_steal_pointer (&deployment_version);
+
+          is_previous_version = (osname_matches && passed_crossover && last_version &&
+                                 (g_strcmp0 (deployment_version, last_version) != 0));
```
jlebon: Hmm, we should be checking this on every iteration, not just when it's the booted or merge deployment, no? Otherwise, it'll never trigger.

src/ostree/ot-admin-builtin-deploy.c
```diff
@@ -173,8 +175,11 @@ ot_admin_builtin_deploy (int argc, char **argv, OstreeCommandInvocation *invocat
         return glnx_throw (error, "--stage cannot currently be combined with --retain arguments");
       if (opt_not_as_default)
         return glnx_throw (error, "--stage cannot currently be combined with --not-as-default");
+      OstreeSysrootSimpleWriteDeploymentFlags flags = 0;
```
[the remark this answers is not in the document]
r4f4: Ok, I'll try to do it in a separate PR.

src/libostree/ostree-sysroot.c
```diff
@@ -1553,6 +1553,31 @@ ostree_sysroot_init_osname (OstreeSysroot       *self,
   return TRUE;
 }
 
+static gchar *
+ostree_deployment_get_version (OstreeSysroot    *sysroot,
```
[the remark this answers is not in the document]
r4f4: Added a new commit for this.

src/libostree/ostree-sysroot.c
r4f4: Yes, what I want is to stop checking versions once I've already retained a previous version.

cgwalters:
Hmm.  Not opposed but this feels a bit like it's crossing into logic that may be better purely in rpm-ostree?   On the other hand maybe any future `apt/ebuild-ostree` tools would want this too.

r4f4:
Rebased and squashed commits.

r4f4:
> Hmm. Not opposed but this feels a bit like it's crossing into logic that may be better purely in rpm-ostree? On the other hand maybe any future `apt/ebuild-ostree` tools would want this too.

@jlebon any comments?

tests/installed/destructive/staged-deploy.yml
```diff
@@ -113,5 +113,22 @@
...
     commit: "{{ rpmostree_status['deployments'][0]['checksum'] }}"
 
+- name: Previous version is retained during staged
+  shell: |
+    set -xeuo pipefail
+    ostree --repo=/ostree/repo rfs --create nonstaged-deploy "${commit}"
+    ostree admin deploy nonstaged-deploy
+    ostree admin status > status.txt
+    grep -vqFe '(staged)' status.txt
+    ostree admin deploy --stage --retain-previous staged-deploy
```
jlebon: `--retain-previous-version`

Though... note there is no CI currently testing this. But we might make use of these tests (even if we lift them out of the playbooks) once we rework CI here. For this PR, we'll have to make sure we test it manually for now (and at least we'll get some CI coverage in the rpm-ostree testsuite once it makes use of it).

Here are some steps on how to do that using cosa: https://github.com/coreos/coreos-assembler/blob/master/README-devel.md#using-overrides

src/libostree/ostree-deployment.c
```diff
@@ -63,6 +63,40 @@ ostree_deployment_get_bootserial (OstreeDeployment *self)
...
 
+/*
+ * ostree_deployment_get_version:
+ * @self: Deployment
+ *
+ * Returns: (transfer full): The deployment's version
+ */
+char *
+ostree_deployment_get_version (OstreeDeployment  *self,
+                               OstreeRepo        *repo,
```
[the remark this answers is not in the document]
jlebon: Actually, can we punt on adding this new API for now? Let's just put it in `ostree-deployment-private.h` and prefix with `_`. That way at least we still get the deduping.

src/libostree/ostree-deployment.c
```diff
@@ -63,6 +63,40 @@ ostree_deployment_get_bootserial (OstreeDeployment *self)
...
+                               GError           **error)
+{
+  gchar *version = NULL;
+
+  if (!repo)
```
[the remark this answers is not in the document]
r4f4: Done.

src/libostree/ostree-deployment.c
```diff
@@ -63,6 +63,40 @@ ostree_deployment_get_bootserial (OstreeDeployment *self)
...
+    return NULL;
+
+  /* Try extracting a version for this deployment. */
+  const gchar *csum = ostree_deployment_get_csum (self);
+  g_autoptr(GVariant) variant = NULL;
+  g_autoptr(GVariant) metadata = NULL;
+
+  /* XXX Copying ot_admin_checksum_version() + bits from
+   *     ot-admin-builtin-status.c.  Maybe this should be
+   *     public API in libostree? */
```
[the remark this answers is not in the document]
r4f4: Done

src/libostree/ostree-deployment.c
```diff
@@ -63,6 +63,40 @@ ostree_deployment_get_bootserial (OstreeDeployment *self)
...
+{
+  gchar *version = NULL;
+
+  if (!repo)
+    return NULL;
+
+  /* Try extracting a version for this deployment. */
+  const gchar *csum = ostree_deployment_get_csum (self);
+  g_autoptr(GVariant) variant = NULL;
+  g_autoptr(GVariant) metadata = NULL;
```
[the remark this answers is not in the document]
r4f4: Done

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2814,6 +2814,7 @@ ostree_sysroot_stage_tree (OstreeSysroot     *self,
                            OstreeDeployment  *merge_deployment,
                            char             **override_kernel_argv,
                            OstreeDeployment **out_new_deployment,
+                           OstreeSysrootSimpleWriteDeploymentFlags flags,
```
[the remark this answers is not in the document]
r4f4: Done

src/libostree/ostree-deployment.h
```diff
@@ -72,6 +72,8 @@ _OSTREE_PUBLIC
 OstreeBootconfigParser *ostree_deployment_get_bootconfig (OstreeDeployment *self);
 _OSTREE_PUBLIC
 GKeyFile *ostree_deployment_get_origin (OstreeDeployment *self);
+_OSTREE_PUBLIC
+char *ostree_deployment_get_version (OstreeDeployment *self, OstreeRepo *repo, GError **error);
```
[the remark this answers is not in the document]
r4f4: It's a private function now, so not needed anymore.

src/libostree/ostree-deployment.c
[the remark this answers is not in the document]
r4f4: Done

tests/installed/destructive/staged-deploy.yml
r4f4: Fixed that and changed the test to look more like what I was doing when testing it manually.

src/libostree/ostree-sysroot.c
```diff
@@ -1586,6 +1586,9 @@ ostree_sysroot_init_osname (OstreeSysroot       *self,
  * specified, then no cleanup will be performed after adding the
  * deployment. Make sure to call ostree_sysroot_cleanup() sometime
  * later, instead.
+ *
+ * If %OSTREE_SYSROOT_SIMPLE_WRITE_DEPLOYMENT_FLAGS_RETAIN_PREVIOUS_VERSION is
+ * specified, then the previous version will not be garbage collected.
```
jlebon: OK, so testing this now, I don't think this is quite what we want. The case I'm trying to capture is this:

Imagine we have two deployments:
- `31.0` (booted)
- `30.0`

What I want to prevent here is `30.0` getting blown away if we `rpm-ostree install foobar`. But e.g. if we `rpm-ostree upgrade` and get `31.1`, then regular GC rules apply. But the logic below doesn't GC it.

Essentially, I think we just want `RETAIN_PREVIOUS_VERSION` to mean "keep at least 2 versions around". So I think we need to tweak the logic so that we just do this: get the version from `new_deployment`, and make sure that the first deployment that has a different version is retained. If it would've been retained anyway by existing rules, then the condition is also satisfied.

Does that make sense?

jlebon:
I'm using this rpm-ostree patch to test this:

```diff
diff --git a/src/daemon/rpmostree-sysroot-upgrader.c b/src/daemon/rpmostree-sysroot-upgrader.c
index e3f5acef..d16ea710 100644
--- a/src/daemon/rpmostree-sysroot-upgrader.c
+++ b/src/daemon/rpmostree-sysroot-upgrader.c
@@ -1380,11 +1380,12 @@ rpmostree_sysroot_upgrader_deploy (RpmOstreeSysrootUpgrader *self,

       g_auto(RpmOstreeProgress) task = { 0, };
       rpmostree_output_task_begin (&task, "Staging deployment");
-      if (!ostree_sysroot_stage_tree (self->sysroot, self->osname,
+      if (!ostree_sysroot_stage_tree_with_flags (self->sysroot, self->osname,
                                       target_revision, origin,
                                       self->cfg_merge_deployment,
                                       self->kargs_strv,
                                       &new_deployment,
+                                      OSTREE_SYSROOT_SIMPLE_WRITE_DEPLOYMENT_FLAGS_RETAIN_PREVIOUS_VERSION,
                                       cancellable, error))
         return FALSE;
     }
```

So here's how you would test this using cosa (assuming your cosa workdir is at `/srv/fcos`):

```
ostree-dir$ make install DESTDIR=/src/fcos/overrides/rootfs
rpm-ostree-dir$ make install DESTDIR=/src/fcos/overrides/rootfs
/srv/fcos$ cosa build
/srv/fcos$ cosa run
```

Inside the VM, you can do something like this:

```
$ ostree commit -b tmp --tree=ref=<checksum-of-booted-deployment> --add-metadata-string version=my-version-1
$ rpm-ostree rebase :tmp
<reboot>
$ ostree commit -b tmp --tree=ref=tmp --add-metadata-string version=my-version-2
$ rpm-ostree upgrade
<reboot>
$ rpm-ostree install ltrace
```

Once you're up and running, you can use `make vmsync` in rpm-ostree to hack on top of this without having to rebuild FCOS each time.

src/libostree/ostree-sysroot.c
r4f4: Yes. With the updated version, my tests seem to reflect the intended behaviour you described here.

src/libostree/ostree-deployment.c
```diff
@@ -63,6 +63,33 @@ ostree_deployment_get_bootserial (OstreeDeployment *self)
...
+  if (self->version != NULL)
+    return self->version;
+
+  /* Try extracting a version for this deployment. */
+  const gchar *csum = ostree_deployment_get_csum (self);
+
+  gchar *version = NULL;
+  g_autoptr(GVariant) variant = NULL;
+  if (ostree_repo_load_variant (repo, OSTREE_OBJECT_TYPE_COMMIT, csum,
+                                &variant, error))
```
jlebon: This should be e.g.

```c
if (!ostree_repo_load_variant (...))
  return FALSE
<continue here>
```

src/libostree/ostree-sysroot.c
```diff
@@ -1738,9 +1743,15 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
   if (!booted_deployment && !merge_deployment && (retain_pending || retain_rollback))
     retain = TRUE;
 
+  /* tracks current versioned deployment */
+  OstreeRepo *repo = ostree_sysroot_repo (sysroot);
+  const gchar *last_version =
```
jlebon: `new_version`?

src/libostree/ostree-sysroot.c
```diff
@@ -1758,6 +1769,13 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
        * deployments, fall back on merge deployment */
       const gboolean passed_crossover = booted_deployment ? !before_booted : !before_merge;
 
+      gboolean is_previous_version = FALSE;
+      if (passed_crossover && osname_matches && !retained_previous_version)
+        {
+          const gchar *version = _ostree_deployment_get_version (deployment, repo, error);
+          is_previous_version = version && (!last_version || (g_strcmp0 (version, last_version) < 0));
```
jlebon: Hmm, should be `!= 0` right?

jlebon: Actually, I think this can just be simplified to:

```
is_previous_version = (g_strcmp0 (version, last_version) != 0);
```

src/libostree/ostree-sysroot.c
```diff
@@ -1738,9 +1743,15 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
   if (!booted_deployment && !merge_deployment && (retain_pending || retain_rollback))
     retain = TRUE;
 
+  /* tracks current versioned deployment */
+  OstreeRepo *repo = ostree_sysroot_repo (sysroot);
+  const gchar *last_version =
+    _ostree_deployment_get_version (new_deployment, repo, error);
```
jlebon: I think here, we should also check if the `booted_deployment` or `merge_deployment` have different versions from `new_version`. We never prune those, so it would fulfill the criteria. So then, we'd set `retained_previous_version = TRUE` here.

If they *aren't* different (or there is no booted/merge), then the logic below is good.

src/libostree/ostree-sysroot.c
```diff
@@ -1738,6 +1743,26 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
...
+
+  gboolean retained_previous_version = FALSE;
+  if (booted_deployment)
+    {
+      const gchar *booted_version =
+        _ostree_deployment_get_version (booted_deployment, repo, error);
+      retained_previous_version = (g_strcmp0 (booted_version, new_version) != 0);
+    }
+
+  if (!retained_previous_version && merge_deployment)
```
jlebon: Could also check here that the booted and merge are not the same, e.g. `&& !ostree_deployment_equal (merge_deployment, booted_deployment)`. But not a big deal either since we cache the version now (though actually, this will still do work in the `NULL` case -- we should probably have a `version_is_cached` boolean field too in `OstreeDeployment` or something, but we can do that as a follow-up!).

jlebon:
@r4f4 Could you rebase and squash all the fixups? Will give it a final review and test!

r4f4:
Rebased and squashed!

r4f4:
What do I have to do for this test to pass? Just add it to `apidoc/ostree-sections.txt`?
```
Verifying all public symbols are documented:
--- expected-documented.txt	2020-02-04 21:05:06.052151036 +0000
+++ found-documented.txt	2020-02-04 21:05:06.076151036 +0000
@@ -392,7 +392,6 @@
 ostree_sysroot_set_mount_namespace_in_use
 ostree_sysroot_simple_write_deployment
 ostree_sysroot_stage_tree
-ostree_sysroot_stage_tree_with_flags
```

src/libostree/ostree-sysroot.c
r4f4: Done

r4f4:
Are you guys still interested on this or should I close the PR?

jlebon:
Yes, I'm still interested in this! I need to set aside some time to test it out.

r4f4:
Ok. I'll try to rebase and fix the conflicts.

src/ostree/ot-admin-builtin-deploy.c
```diff
@@ -195,6 +197,7 @@ ot_admin_builtin_deploy (int argc, char **argv, OstreeCommandInvocation *invocat
   OstreeSysrootDeployTreeOpts opts = {
     .override_kernel_argv = kargs_strv,
     .overlay_initrds = overlay_initrd_chksums ? (char**)overlay_initrd_chksums->pdata : NULL,
+    .flags = opt_retain_previous ? OSTREE_SYSROOT_SIMPLE_WRITE_DEPLOYMENT_FLAGS_RETAIN_PREVIOUS_VERSION : 0,
```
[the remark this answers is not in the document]
jlebon: We also need to use the new `_with_options` version of the API in this case otherwise it'll get lost.

jlebon (commented):
OK cool, this works well for me in both the staging and non-staging paths (after adding the missing bit in `ot-admin-builtin-deploy.c`).

Here's how you can pretty easily test this locally in a VM:

```
[root@qemu0 ~]# rpm-ostree status
State: idle
Deployments:
* ostree://fedora:fedora/x86_64/coreos/next
                   Version: 34.20210503.1.1 (2021-05-12T22:57:51Z)
                    Commit: c229af6df100607f73216ecda14dfc3a71e4c7094f4a7cdd260dc97a5c650a3e
              GPGSignature: Valid signature by 8C5BA6990BDB26E19F2A1A801161AE6945719A39
                  Unlocked: development

[root@qemu0 ~]# ostree commit -b previous-version --add-metadata-string=version=34.20210503.1.0 --tree=ref=c229af6df100607f73216ecda14dfc
3a71e4c7094f4a7cdd260dc97a5c650a3e
b6437a1dae3345aa73908bee2fc4606f56e81a338cb4a9082c51042572294534

[root@qemu0 ~]# ostree admin deploy previous-version --not-as-default
Copying /etc changes: 8 modified, 0 removed, 30 added
Bootloader updated; bootconfig swap: yes; bootversion: boot.0.1, deployment count change: 1

[root@qemu0 ~]# ostree commit -b same-version --add-metadata-string=version=34.20210503.1.1 --tree=ref=c229af6df100607f73216ecda14dfc3a71
e4c7094f4a7cdd260dc97a5c650a3e
3478fbc53e070b3d57c86893417329385cd7bc746670d8d2933b93e51b595b02

[root@qemu0 ~]# ostree admin deploy same-version --retain-previous-version
Copying /etc changes: 8 modified, 0 removed, 30 added
Bootloader updated; bootconfig swap: yes; bootversion: boot.1.1, deployment count change: 1
Freed objects: 190 bytes

[root@qemu0 ~]# rpm-ostree status
State: idle
Deployments:
  ostree://same-version
                   Version: 34.20210503.1.1 (2021-05-14T20:11:40Z)
                    Commit: 3478fbc53e070b3d57c86893417329385cd7bc746670d8d2933b93e51b595b02

* ostree://fedora:fedora/x86_64/coreos/next
                   Version: 34.20210503.1.1 (2021-05-12T22:57:51Z)
                    Commit: c229af6df100607f73216ecda14dfc3a71e4c7094f4a7cdd260dc97a5c650a3e
              GPGSignature: Valid signature by 8C5BA6990BDB26E19F2A1A801161AE6945719A39
                  Unlocked: development

  ostree://previous-version
                   Version: 34.20210503.1.0 (2021-05-14T20:11:00Z)
                    Commit: b6437a1dae3345aa73908bee2fc4606f56e81a338cb4a9082c51042572294534
```

Then you can also test the staging path by nuking the pending deployment (`rpm-ostree cleanup -p`) and re-deploying, but this time with `--stage` followed by `ostree admin finalize-staged`.

In practice at the rpm-ostree level I think we should just omit the flag in the `rpm-ostree rebase` case, which is like what we're doing here since we're changing branches, but this works to test the change here. (Or maybe even only enable this in the `NO_PULL_BASE` case to start.)

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1923,24 +1923,8 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
-       *     public API in libostree? */
-      if (ostree_repo_load_variant (repo, OSTREE_OBJECT_TYPE_COMMIT, csum,
-                                    &variant, NULL))
-        {
-          metadata = g_variant_get_child_value (variant, 0);
-          g_variant_lookup (metadata, OSTREE_COMMIT_META_KEY_VERSION, "s", &deployment_version);
-        }
-    }
+  const gchar *deployment_version =
+      _ostree_deployment_get_version (deployment, repo, error);
```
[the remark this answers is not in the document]
r4f4: Sorry, I don't think I follow. The previous `if (repo) {` was replaced by `g_return_val_if_fail (repo != NULL, NULL);`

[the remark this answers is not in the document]
jlebon: `g_return_val_if_failed` has different semantics. It's more of a [precondition](https://en.wikipedia.org/wiki/Design_by_contract) which we always expect to be true.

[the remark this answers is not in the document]
jlebon: Yup, exactly!

r4f4:
@jlebon I think I've addressed all the points and the test is now passing `[2021-05-20T17:04:53.378Z] --- PASS: ext.ostree.destructive.staged-deploy.sh (151.61s)`

src/libostree/ostree-deployment.c
```diff
@@ -87,6 +87,32 @@ ostree_deployment_get_bootserial (OstreeDeployment *self)
   return self->bootserial;
 }
 
+char *
```
jlebon, line 90: Should return a `const char*` to signal we own the string.

src/libostree/ostree-deployment.c
```diff
@@ -87,6 +87,32 @@ ostree_deployment_get_bootserial (OstreeDeployment *self)
...
+
+  g_autoptr(GVariant) variant = NULL;
+  if (!ostree_repo_load_variant (repo, OSTREE_OBJECT_TYPE_COMMIT, csum,
+                                 &variant, error))
+    return NULL;
+
+  g_autoptr(GVariant) metadata = g_variant_get_child_value (variant, 0);
+  g_variant_lookup (metadata, OSTREE_COMMIT_META_KEY_VERSION, "s", &self->version);
+
+  self->version_is_cached = TRUE;
```
jlebon, line 111: The pattern I like to use is:

```
if (!is_cached) {
  // cache it
  is_cached = TRUE;
}
return cached_value;
```

Though this is fine too of course!

src/libostree/ostree-sysroot.h
```diff
@@ -196,9 +196,13 @@ gboolean ostree_sysroot_stage_overlay_initrd (OstreeSysroot  *self,
...
                                               GError        **error);
 
+/**
+ * finalization_flags: only used in the staging path
+ */
 typedef struct {
   gboolean unused_bools[8];
-  int unused_ints[8];
+  int unused_ints[7];
+  int finalization_flags;
```
jlebon, line 205: Let's invert the order here?

src/libostree/ostree-deployment.c
```diff
@@ -87,6 +87,32 @@ ostree_deployment_get_bootserial (OstreeDeployment *self)
   return self->bootserial;
 }
 
+char *
+_ostree_deployment_get_version (OstreeDeployment  *self,
+                                OstreeRepo        *repo,
+                                GError           **error)
```
jlebon, line 93: Error-handling for this function is funky. `NULL` could mean an error occurred or there is no version. Since this is opportunistic anyway and all callers need to gracefully deal with missing version strings, maybe simpler to just drop the `error` arg entirely instead of having them now check if `error` was set.

src/libostree/ostree-sysroot.c
```diff
@@ -1884,6 +1889,27 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
...
+    }
+
+  if (!retained_previous_version && merge_deployment &&
+      !ostree_deployment_equal (merge_deployment, booted_deployment))
+    {
+      const gchar *merge_version =
+        _ostree_deployment_get_version (merge_deployment, repo, error);
+      retained_previous_version = (g_strcmp0 (merge_version, new_version) != 0);
+    }
+
```
jlebon, line 1912: Some comments for these would be helpful.

westurner:
- coreos/rpm-ostree#577

PR #1961, opened 2019-10-28
lib/pull: Tweak update_timeout logic again

jlebon:
I was hitting `SIGSEGV` when running `cosa build` and narrowed it down
to #1954. What's happening here is that because we're using the default
context, when we unref it in the out path, it may not actually destroy
the `GSource` if it (the context) is still ref'ed elsewhere. So then,
we'd still get events from it if subsequent operations iterated the
context.

This patch is mostly a revert of #1954, except that we still keep a ref
on the `GSource`. That way it is always safe to destroy it afterwards.
(And I've also added a comment to explain this better.)

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3573,6 +3573,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   g_autofree char **refs_to_fetch = NULL;
   g_autoptr(GVariantIter) collection_refs_iter = NULL;
   g_autofree char **override_commit_ids = NULL;
+  g_autoptr(GSource) update_timeout = NULL;
   gboolean opt_gpg_verify_set = FALSE;
   gboolean opt_gpg_verify_summary_set = FALSE;
   gboolean opt_collection_refs_set = FALSE;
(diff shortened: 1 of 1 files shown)


jlebon:
Sorry for the churn on this.

The net diff here wrt pre-#1954 is:

```diff
diff --git a/src/libostree/ostree-repo-pull.c b/src/libostree/ostree-repo-pull.c
index bcb8b420..586b34fe 100644
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3573,7 +3573,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
   g_autofree char **refs_to_fetch = NULL;
   g_autoptr(GVariantIter) collection_refs_iter = NULL;
   g_autofree char **override_commit_ids = NULL;
-  GSource *update_timeout = NULL;
+  g_autoptr(GSource) update_timeout = NULL;
   gboolean opt_gpg_verify_set = FALSE;
   gboolean opt_gpg_verify_summary_set = FALSE;
   gboolean opt_collection_refs_set = FALSE;
@@ -3675,6 +3675,10 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
     pull_data->async_error = &pull_data->cached_async_error;
   else
     pull_data->async_error = NULL;
+
+  /* Note we're using the thread default (or global) context here, so it may outlive the
+   * OtPullData object if there's another ref on it. Thus, always detach/destroy sources
+   * local to the `ostree_repo_pull*` operation rather than trying to transfer ownership. */
   pull_data->main_context = g_main_context_ref_thread_default ();
   pull_data->flags = flags;

@@ -4514,7 +4518,6 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       g_source_set_priority (update_timeout, G_PRIORITY_HIGH);
       g_source_set_callback (update_timeout, update_progress, pull_data, NULL);
       g_source_attach (update_timeout, pull_data->main_context);
-      g_source_unref (update_timeout);
     }

   /* Now await work completion */
```

which makes it more obvious what's going on.

cgwalters:
Ah yes, right.  I had a vague notion that destroy was there for a reason...and now we have a comment so that will help in the future!

/lgtm

PR #1962, opened 2019-10-29
lib/deploy: Also install HMAC file into /boot

jlebon:
To allow for FIPS mode, we need to also install the HMAC file from
`/usr/lib/modules` to `/boot` alongside the kernel image where the
`fips` dracut module will find it. For details, see:

https://github.com/coreos/fedora-coreos-tracker/issues/302

Note I didn't include the file in the boot checksum since it's itself a
checksum of the kernel, so we don't really gain much here other than
potentially causing an unnecessary bootcsum bump.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -878,6 +878,8 @@ typedef struct {
   int   boot_dfd;
   char *kernel_srcpath;
   char *kernel_namever;
+  char *kernel_hmac_srcpath;
+  char *kernel_hmac_namever;
   char *initramfs_srcpath;
   char *initramfs_namever;
   char *devicetree_srcpath;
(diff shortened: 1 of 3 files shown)


cgwalters:
I think this is more an rpm-ostree bug:

```
$ cat /usr/lib/modules/5.2.16-200.fc30.x86_64/.vmlinuz.hmac
654f49c15416a589cd6892e6900385e4c1fcc2e34213b177972085d1ef0cb0e355f4c57eed973f6c44a2279c367b95f90f3a14aeda5fb9e71cdf03c405a7c89f  /boot/vmlinuz-5.2.16-200.fc30.x86_64
```

is also wrong.  Or to say it more precisely, how we *built* it is wrong; ostree shouldn't be working around build problems.

Hm, maybe it'd work to just make the file purely relative?

jlebon:
> Hm, maybe it'd work to just make the file purely relative?

Yeah, seems like the cleanest solution. Done in https://github.com/coreos/rpm-ostree/pull/1934 now!

PR #1963, opened 2019-10-29
Release 2019.5

cgwalters:
This is mostly;

 - `scan-build` fixes
 - zipl backend

Plus some attempted changes that got reverted.  But I'd like to get the zipl stuff out officially.


jlebon:
So I think https://github.com/ostreedev/ostree/pull/1962 is in a good position now. WDYT about pulling it in?

cgwalters:
> So I think #1962 is in a good position now. WDYT about pulling it in?

Yeah, makes sense.

darkmuggle:
👍 Thank you most kindly for working on this!

PR #1966, opened 2019-11-02
Fix tests with musl

akiernan:
When running under musl rather than glibc, the tests fail because of minor differences. There's three commits here
* the first is straightforward (handle EPIPE with some `|| true` additions)
* the second is possibly more contentious (default to C.UTF-8 as the locale command doesn't exist)
* third looks like a problem somewhere between coreutils and musl and this is simply a workaround for someone elses problem - I could see you not wanting this one...


cgwalters:
The latter does indeed seem like a workaround for a musl problem.  I'd be a little concerned if anything else on your target system is relying on this `cp`... 

But, if it's *just* that...OK.

/approve
/lgtm

Issue #1951, opened 2019-10-21
Rework BLS fragment writing to do "two phase commit"


cgwalters:
Currently we require `/boot/loader` be a symbolic link, so that we can transactionally replace all of the entries.    This causes various problems because it's an OSTree-specific invention.  

Another approach would be to use "journaling", something like this:

- First, write `/boot/loader/.ostree-txn` which would be a single file containing the new list of OSTree-specific

Completion:

- If `/boot/loader/.ostree-txn` exists, add all bootloader entries referenced by it, then delete all unreferenced ones
- `unlink(/boot/loader.ostree-txn)`

Now if we're interrupted between these steps, we can add a systemd unit which does `ConditionPathExists=/boot/loader/.ostree-txn` and does the work on bootup.  If we also detect that this would have changed the default boot entry, then we reboot.

AdrianVovk:
Cool. This'll solve the sd-boot problem.

Would it be possible for OSTree to also be able to operate directly on /efi (systemd's preferred mount point for the ESP)

cgwalters:
Also just recording some other thoughts here.  In some cases, we end up writing kernel+initramfs to FAT.  And in general there's been longstanding issues with FS journaling versus bootloaders; see https://github.com/ostreedev/ostree/pull/1049

I think we should create a higher level "protocol" between things writing data in `/boot` and the ESP and the bootloaders.  For example, we could have separate "checksum files" for the BLS fragments, kernel/initramfs etc.  Something like this:

```
/boot/loader/entries/foo.conf
/boot/loader/entries/foo.conf.sha256
/boot/vmlinuz-1
/boot/vmlinuz-1.sha256
/boot/vmlinuz-2
/boot/vmlinuz-2.sha256
/boot/initramfs-1
/boot/initramfs-1.sha256
````

etc.  And the bootloader would validate these before trying to boot a particular entry.  Or it'd at least verify the BLS fragment checksum since that should be basically free.

Things writing kernels/initramfs to `/boot` should try to ensure there's at least one valid bootable entry, and the filesystem is fully sync'd before deleting things.

cgwalters:
xref https://marc.info/?l=linux-fsdevel&m=157168785821373&w=2
(I replied in that thread also linking to this issue)

bam80:
What is the status of this?

damianatorrpm:
This would also fix systemd-boot usage in Silverblue which uses /boot/loader/entries/ not /boot/loader.0/
[...]

PR #1967, opened 2019-11-03
Use rename of directories instead of symbolic links in boot partition.

valentindavid:
This implements  @AdrianVovk idea from https://github.com/ostreedev/ostree/issues/1719#issuecomment-449895351 to solve issue #1719.

This uses `renameat2` to do atomic swap of the loader directory in the
boot partition. It fallsback to non-atomic rename. This stays atomic
on filesystems supporting links but also provide a non-atomic behavior
when filesystem does not provide any atomic alternative.

This is working with SystemD boot on EFI using boot loader
specifications.

There is still the issue of losing `/loader/loader.conf` with SystemD
boot. Maybe we should think about copying other files from previous loader directories.

### DIFF
--- a/src/libostree/ostree-sysroot-cleanup.c
+++ b/src/libostree/ostree-sysroot-cleanup.c
@@ -216,6 +216,15 @@ cleanup_other_bootversions (OstreeSysroot       *self,
   const int cleanup_subbootversion = self->subbootversion == 0 ? 1 : 0;
   /* Reusable buffer for path */
   g_autoptr(GString) buf = g_string_new ("");
+  struct stat stbuf;
+
+  if ((glnx_fstatat_allow_noent (self->sysroot_fd, "boot/loader", &stbuf, AT_SYMLINK_NOFOLLOW, error))
+      && (!S_ISLNK (stbuf.st_mode)))
+    {
+      g_string_truncate (buf, 0); g_string_append_printf (buf, "boot/loader.%d", self->bootversion);
+      if (!glnx_shutil_rm_rf_at (self->sysroot_fd, buf->str, cancellable, error))
+        return FALSE;
+    }
 
   /* These directories are for the other major version */
   g_string_truncate (buf, 0); g_string_append_printf (buf, "boot/loader.%d", cleanup_bootversion);
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2026,18 +2026,18 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
   return TRUE;
 }
 
-/* We generate the symlink on disk, then potentially do a syncfs() to ensure
+/* We generate the directory on disk, then potentially do a syncfs() to ensure
  * that it (and everything else we wrote) has hit disk. Only after that do we
  * rename it into place.
  */
 static gboolean
-prepare_new_bootloader_link (OstreeSysroot  *sysroot,
-                             int             current_bootversion,
-                             int             new_bootversion,
-                             GCancellable   *cancellable,
-                             GError        **error)
+prepare_new_bootloader_dir (OstreeSysroot  *sysroot,
+                            int             current_bootversion,
+                            int             new_bootversion,
+                            GCancellable   *cancellable,
+                            GError        **error)
 {
-  GLNX_AUTO_PREFIX_ERROR ("Preparing final bootloader swap", error);
+  GLNX_AUTO_PREFIX_ERROR ("Preparing new bootloader directory", error);
   g_assert ((current_bootversion == 0 && new_bootversion == 1) ||
             (current_bootversion == 1 && new_bootversion == 0));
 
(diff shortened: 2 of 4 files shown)


AdrianVovk:
The challenge with this is that the ESP is vfat, and therefore rename will always be non-atomic. It's still better than what I'm doing now (manually copying files into /efi)

https://github.com/ostreedev/ostree/issues/1951
^ here's a proposed solution that I think could solve this same problem without needing the renameat2 call.

bam80:
Is this PR still proposed or abandoned?

valentindavid:
> Is this PR still proposed or abandoned?

I can rebase and take care of it a bit. But since there was no interaction from maintainers, then I am not sure whether they  are considering it.

bam80:
@valentindavid have you been able to pay some attention to this issue again? The changes are highly anticipated here, in one form or another. Thanks.

jjardon:
@mwleeds @rfairley any chance to take a look to this? We currently need this to make GNOME images work, see https://gitlab.gnome.org/GNOME/gnome-build-meta/-/commit/79fb62e0d243a21ab58dc1dda439c23db5d474ab

jjardon:
@cgwalters Hey! Any chance to take a look to this?

bam80:
Friendly ping.

AdrianVovk:
@valentindavid This PR conflicts with upstream again. GNOME OS uses an ancient version of OSTree so they have yet to run into this issue, but on my OS I just updated to the latest version and I'm about to be downgrading again because this patch is broken.

@cgwalters Is there any chance for this patch to be reviewed or considered at all? Without this patch, I have to follow every single `ostree admin deploy` anywhere in the process of building, generating ISOs, installing the OS, and updating the OS with a bunch of ugly commands that mess with the contents of the ESP. With this patch, it just works. OSTree operates on BLS files, but has yet to support the bootloader that was actually built for the standard :shrug:

damianatorrpm:
[updated.txt](https://github.com/ostreedev/ostree/files/5433897/updated.txt)
Rebased patch file.

kowalski7cc:
Any news on this?

runiq:
Hi, any updates here, by any chance? :)

jjardon:
In case is useful, this patch has been updated in GNOME OS to work with libostree-2021.2: https://gitlab.gnome.org/GNOME/gnome-build-meta/-/blob/41.rc/files/ostree/no-boot-symlink.patch

AdrianVovk:
carbonOS has been using this patch for a while. It works flawlessly so far! Hopefully this will get merged at some point...

@jjardon that same patch works for 2021.3 also

cgwalters:
Hi, sorry about the delay.  

I am OK in principle with the idea and code.  But I'd like to have some design to close the atomicity hole longer term.

If you (or someone else) doesn't mind rebasing we can try to get this merged.

src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -216,6 +216,15 @@ cleanup_other_bootversions (OstreeSysroot       *self,
...
   g_autoptr(GString) buf = g_string_new ("");
+  struct stat stbuf;
+
+  if ((glnx_fstatat_allow_noent (self->sysroot_fd, "boot/loader", &stbuf, AT_SYMLINK_NOFOLLOW, error))
+      && (!S_ISLNK (stbuf.st_mode)))
+    {
+      g_string_truncate (buf, 0); g_string_append_printf (buf, "boot/loader.%d", self->bootversion);
+      if (!glnx_shutil_rm_rf_at (self->sysroot_fd, buf->str, cancellable, error))
+        return FALSE;
+    }
```
dbnicholson, line 227: Is this logic even right? In `swap_bootloader`, `loader.<newversion>` is exchanged with `loader`. That means that the current bootloader directory is now named `loader.<newversion>`. What's even at `loader.<currentversion>`? This is also in conflict with the code below that cleans up `loader.<currentversion>`. So, after this change, there's no more `loader.*` bootloader directories left. Is that correct?

dbnicholson (changes requested):
I didn't finish thinking through all of this, but I wouldn't want to merge this without a lot more comments and tests. Is the symlink fallback path tested at all?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1872,38 +1872,89 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+static gboolean
+renameat2_exchange (int         olddirfd,
+                    const char *oldpath,
+                    int         newdirfd,
+                    const char *newpath,
+                    gboolean   *is_atomic,
+                    GError    **error)
+{
+  if (renameat2(olddirfd, oldpath, newdirfd, newpath, RENAME_EXCHANGE) == 0)
+    return TRUE;
```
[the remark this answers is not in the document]
valentindavid: We start within `ostree_sysroot_write_deployments_with_options` with:
```
    bootloader_is_atomic = bootloader != NULL && _ostree_bootloader_is_atomic (bootloader);
```

Then we only mark `bootloader_is_atomic` as `FALSE` if we do not manage to do a `renameat2`. But if we manage, we want to still keep `FALSE` if that was the value given by `_ostree_bootloader_is_atomic`.

src/libostree/ostree-sysroot-cleanup.c
valentindavid: Right. Only `loader` directory is left after deployment is finished successfully. We could just use `loader.tmp` and `loader`. But if I remember correctly, we need to support older deployments that are still using symlinks. For that reason we still to need to keep this 0/1 version names. Also maintaining a patch for long term I cannot really afford changing too much the logic.

valentindavid: Reading the code, I do not think the support for older deployment is working. I will try to add tests for that.

src/libostree/ostree-sysroot-deploy.c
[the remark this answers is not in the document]
dbnicholson: OK, I missed that part.

dbnicholson:
I'm still not totally sure what the best approach here is, but I think the fundamental change from replacing `loader` to exchanging `loader` is likely to have subtle effects that need more detailed analysis.

Since the replacing process needs to be supported in case `RENAME_EXCHANGE` is not available or there's no existing `loader`, I think the best approach would be to make it always work as if it's replacing. An approach I think would work is:

1. Use the existing symlink replacement, which involves creating a temporary `loader.tmp` symlink and the renaming it to `loader`. This is the simple tried and true approach.
2. If symlinking fails because of `EPERM`, then recursively copy the new loader directory to `loader.tmp` and then exchange it with `loader` using `RENAME_EXCHANGE` + the renaming fallbacks you already coded. When exchanging has occurred, recursively delete the exchanged `loader.tmp` directory.

martinezjavier:
> Since the replacing process needs to be supported in case `RENAME_EXCHANGE` is not available

@dbnicholson  FYI, I worked on a patch series to add `RENAME_EXCHANGE` support to vfat, latest revision is: https://lkml.org/lkml/2022/6/1/1049

dbnicholson:
> > Since the replacing process needs to be supported in case `RENAME_EXCHANGE` is not available
> 
> @dbnicholson FYI, I worked on a patch series to add `RENAME_EXCHANGE` support to vfat, latest revision is: https://lkml.org/lkml/2022/6/1/1049

That's great. Of course, you'd still need to support the fallback since it will be some time until that's commonly available (presuming it gets merged).

I've come around a bit on this idea in the sense that I think it's reasonable to assume `RENAME_EXCHANGE` as the primary mechanism since there will always be filesystems like vfat that can't do symlinks but presumably they can all gain support for `RENAME_EXCHANGE`. However, I still feel like any major changes in this area require exhaustive testing.

ricardosalveti:
I'm currently playing with https://github.com/ricardosalveti/ostree/commit/731b73f3c2fabb4b8b0ac1d101bdc70015e206a9, which is based on the original patch from this PR which keeps the current behavior for new systems and systems that are already deployed, but also supports loader as a directory (relying on RENAME_EXCHANGE).

That way I can create new deployments (using systemd-boot) with the extra kernel patches without breaking my current deployments are are ok with links (using grub / u-boot).

Still validating with my yocto integration, can post the final set after I'm done testing it.

martinezjavier:
FYI the patch-set got merged by a subsystem maintainer and are heading towards the Linux v5.20 release.

GrabbenD:
Is GNOME OS's patch still the best workaround until this is finished?
https://gitlab.gnome.org/GNOME/gnome-build-meta/-/blob/45.beta/files/ostree/no-boot-symlink.patch

GrabbenD:
I ended up working around this issue by making `/` a ext4 partition and then I mounted ESP on `/boot/esp` which makes `/boot` ext4 thus allowing symbolic links

Here's the commands:
https://github.com/M1cha/archlinux-ostree/issues/1#issuecomment-1700747666

bam80:
update?

dbnicholson:
I've thought about this several times as I would really like to have this in Endless to support our systemd-boot systems. What always makes me anxious is trying to figure out how to handle compatibility with the vast majority of our systems that have symlink based deployments. There are 2 main issues I'm concerned with:

* The semantics of the deployment are different if `loader` is a symlink to a `loader.N` directory versus `loader` itself being a directory. What do the swapped `loader.N` directories mean in pure directory mode? I think the answer is that they don't really match the semantics of the symlink deployment. It's essentially a different algorithm even though it's seemingly only a small change from the symlink approach.
* What do you do with systems that already have symlink based deployments? Potentially you could write a migration to the directory scheme, but that means you can't roll back to a system with a libostree that doesn't understand the directory scheme.

So, to me this requires a couple additional pieces of implementation and policy.

* The directory deployment scheme is treated separately from the symlink deployment scheme. Your system can be in one scheme or the other, but the semantics are different and shouldn't be mixed.
* If your system is already using a symlink based deployment, it should stay that way. You can opt in to a migration to the directory scheme, but this means you may potentially lose roll back targets. At Endless we'd do that at some major version check point and probably have to delete some old deployments to ensure users didn't try to roll back to a system where they'd be screwed. Or maybe we'd never migrate anyone because the downsides are too great.
* If you're on a new deployment (i.e., no existing `/boot/loader` or `/boot/ostree`), then the directory scheme is preferred.

bam80:
Status?

champtar:
closing in favor of https://github.com/ostreedev/ostree/pull/3359

PR #1968, opened 2019-11-04
libostree: Add ostree_async_progress_copy_state()

ptomato:
This allows copying the state from one OstreeAsyncProgress object to
another, atomically, without invoking the callback. This is needed in
libflatpak, in order to chain OstreeAsyncProgress objects so that you
can still receive progress updates when iterating a different
GMainContext than the one that the OstreeAsyncProgress object was
created under.

### DIFF
--- a/src/libostree/libostree-devel.sym
+++ b/src/libostree/libostree-devel.sym
@@ -19,6 +19,7 @@
 
 /* Add new symbols here.  Release commits should copy this section into -released.sym. */
 LIBOSTREE_2019.6 {
+  ostree_async_progress_copy_state;
 } LIBOSTREE_2019.4;
 
 /* Stub section for the stable release *after* this development one; don't
--- a/src/libostree/libostree-devel.sym
+++ b/src/libostree/libostree-devel.sym
@@ -18,7 +18,7 @@
 ***/
 
 /* Add new symbols here.  Release commits should copy this section into -released.sym. */
-LIBOSTREE_2019.5 {
+LIBOSTREE_2019.6 {
 } LIBOSTREE_2019.4;
 
 /* Stub section for the stable release *after* this development one; don't
--- a/src/libostree/ostree-async-progress.c
+++ b/src/libostree/ostree-async-progress.c
@@ -425,6 +425,40 @@ ostree_async_progress_set_uint64 (OstreeAsyncProgress       *self,
 }
 
 /**
+ * ostree_async_progress_copy_state:
+ * @self: An #OstreeAsyncProgress to copy from
+ * @dest: An #OstreeAsyncProgress to copy to
+ *
+ * Atomically copies all the state from @self to @dest, without invoking the
+ * callback.
+ * This is used for proxying progress objects across different #GMainContexts.
+ *
+ * Since: 2019.6
+ */
+void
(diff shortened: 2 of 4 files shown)


jlebon:
bot, add author to whitelist

src/libostree/libostree-devel.sym
```diff
@@ -19,6 +19,7 @@
 
 /* Add new symbols here.  Release commits should copy this section into -released.sym. */
 LIBOSTREE_2019.5 {
```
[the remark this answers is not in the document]
ptomato: I did that in a separate commit, I'm not familiar with the conventions that you have for bumping the version so I hope that's OK.

ptomato:
Thanks for the review!

> Can you also link to the corresponding flatpak PR in the commit message for reference?

Done.

> Hmm, don't we need to do `ensure_callback_locked` on the destination `OstreeAsyncProgress` so its callback gets fired?

I documented the function as not invoking the callback, so that was intentional. Of course, it may not be the right decision :smile:

My reasoning (which I'm not sure I'm explaining well enough, it may become clearer after reading the linked flatpak PR) was that libflatpak uses this API to ensure that an `OstreeAsyncProgress`'s callback is called even when the main context it associates with is not being iterated. It does this by creating a "new" progress object in the new main context and copying the state from the "old" object, then when the new object's callback is fired, copy the state back to the old object and manually call the old object's callback. So I think if we also had `ostree_async_progress_copy_state()` also invoke the callback, we'd get a spurious callback invocation when copying from old to new, and it wouldn't invoke anyway when copying from new to old because the old object's main context isn't being iterated.

ptomato:
@jlebon Mind taking another look? I've addressed the comments, and the test failure seems to be transient due to disk space.

src/libostree/ostree-async-progress.c
```diff
@@ -424,6 +424,42 @@ ostree_async_progress_set_uint64 (OstreeAsyncProgress       *self,
...
+  g_return_if_fail (OSTREE_IS_ASYNC_PROGRESS (dest));
+
+  g_mutex_lock (&self->lock);
+
+  if (self->dead)
+    goto out;
+
+  GLNX_HASH_TABLE_FOREACH_KV (self->values, void *, key, GVariant *, value)
+    {
+      if (value)
```
jlebon: Is this on purpose? Seems like if the value for a key is `NULL` in the original table, we'd want to copy that state too, no? Maybe just:

```c
if (value)
  g_variant_ref (value);
g_hash_table_replace (dest->values, key, value); 
```

?

ptomato: This was on purpose since `ostree_async_progress_get()` will crash if the value for a requested key is `NULL`, but I don't think it makes much difference either way. I guess if you had an invalid progress object before, it may as well be invalid after the copy, otherwise it could mask bugs, so maybe this is a bit better. I'll test the flatpak fix with this change, and if no surprises, update this PR.

ptomato: OK, updated.

ptomato:
bot, retest this please

mwleeds:
/retest
(looks like it's f29-rpmostree that's flaky this time)

PR #1969, opened 2019-11-06
Makefile: declare ostree_boot_SCRIPTS and append values

ricardosalveti:
ostree_boot_SCRIPTS was being set on both Makefile-boot.am and
Makefile-switchroot.am, causing the first one to be replaced by the
other at the final Makefile, so declare as empty and append on both
places instead.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>


cgwalters:
Thanks!

/approve
/lgtm

PR #1970, opened 2019-11-07
lib/keyfile: Treat "group not found" the same as "key not found"

cgwalters:
Prep for fsverity, where I want to create a new group
`[fsverity]` in the keyfile that has default values.  We should
treat the absence of a group the same as absence of a key
in these "with defaults" APIs.


cgwalters:
OK, fixed the tests for this one.

PR #1971, opened 2019-11-08
build: create tests directory for split builds

akiernan:
When `--disable-dependency-tracking` is in effect with separate build
directory, the tests directory isn't created as a result of the
dependency generation, which leads to a build race for the tests
directory being created and failures:
```
  Making all in .
  make[2]: Entering directory 'TOPDIR/build/tmp/work/riscv64-yoe-linux-musl/ostree/2019.5-r0/build'
  (echo '[Test]' > tests/test-local-pull-depth.sh.test.tmp; \
  echo 'Type=session' >> tests/test-local-pull-depth.sh.test.tmp; \
  echo 'Exec=env G_TEST_SRCDIR=/usr/libexec/installed-tests/libostree G_TEST_BUILDDIR=/usr/libexec/installed-tests/libostree /usr/libexec/installed-tests/libostree/test-local-pull-depth.sh' >> tests/test-local-pull-depth.sh.test.tmp; \
  mv tests/test-local-pull-depth.sh.test.tmp tests/test-local-pull-depth.sh.test)
  /bin/sh: tests/test-local-pull-depth.sh.test.tmp: No such file or directory
  /bin/sh: line 1: tests/test-local-pull-depth.sh.test.tmp: No such file or directory
  /bin/sh: line 2: tests/test-local-pull-depth.sh.test.tmp: No such file or directory
  mv: cannot stat 'tests/test-local-pull-depth.sh.test.tmp': No such file or directory
  make[2]: *** [Makefile:9282: tests/test-local-pull-depth.sh.test] Error 1
```

### DIFF
--- a/buildutil/glib-tap.mk
+++ b/buildutil/glib-tap.mk
@@ -116,6 +116,7 @@ installed_testcases = $(test_programs) $(installed_test_programs) \
 installed_test_meta_DATA = $(installed_testcases:=.test)
 
 %.test: %$(EXEEXT) Makefile
+	@$(MKDIR_P) $(dir $@)
 	$(AM_V_GEN) (echo '[Test]' > $@.tmp; \
 	echo 'Type=session' >> $@.tmp; \
 	echo 'Exec=env G_TEST_SRCDIR=$(installed_testdir) G_TEST_BUILDDIR=$(installed_testdir) $(installed_testdir)/$(notdir $<)' >> $@.tmp; \


cgwalters:
Ah interesting.  Is there much speedup from `--disable-dependency-tracking`?
/approve
/lgtm

cgwalters:
Need to track down that repo finder flake.
/override f29-introspection-tests

akiernan:
> Ah interesting. Is there much speedup from `--disable-dependency-tracking`?

Looks like its of the order of 0.5s for OSTree on my laptop.

But it's the default for OpenEmbedded where you're never rebuilding inside a working directory and usually building large numbers of packages - in that world 0.5s is likely worth it.

PR #1976, opened 2019-11-21
tests/repo-finder: Run realpath() on /tmp

cgwalters:
This fixes running this test case inside
https://github.com/cgwalters/coretoolbox

### DIFF
--- a/tests/test-repo-finder-mount.c
+++ b/tests/test-repo-finder-mount.c
@@ -62,7 +62,9 @@ setup (Fixture       *fixture,
     g_clear_error (&error);
   g_assert_no_error (error);
 
-  fixture->working_dir = g_file_new_for_path (fixture->tmpdir.path);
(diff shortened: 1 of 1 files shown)


jlebon:
Hmm, where was the test failing? Are we not handling symlinks correctly somewhere else?

cgwalters:
> Hmm, where was the test failing? 

The test serializes file paths to strings, and something in the GLib stack was effectively doing `realpath`, so we need them to match.

jlebon:
bot, retest this please

jlebon:
```
tests/test-repo-finder-mount.c:448:          g_error ("Unknown result ‘%s’ with keyring ‘%s’.",
maint.mk: found error message ending in period
```

jlebon:
/lgtm

Known flakes:
/override f29-primary

And this one I think we need to adjust the test deps or just bump to a newer tag. Clearly not related to this patch though.
/override f29-rpmostree

PR #1980, opened 2019-12-01
README.md: Tweak text about BuildStream

ssssam:
BuildStream no longer uses libostree internally (see
https://gitlab.com/BuildStream/buildstream/issues/387). It still
has first class support for interoperating with libostree repos.


cgwalters:
Mind re-pushing this to retrigger Travis to get past the flake?

ssssam:
Done, but i'm not sure if it's helped ..

PR #1984, opened 2019-12-12
lib/repo: Create repo directories as 0775

jlebon:
For repo structure directories like `objects`, `refs`, etc... we should
be more permissive and let the system's `umask` narrow down the
permission bits as wanted.

This came up in a context where we want to be able to have read/write
access on an OSTree repo on NFS from two separate OpenShift apps by
using supplemental groups[1] so we don't require SCCs for running as the
same UID (supplemental groups are part of the default restricted SCC).

[1] https://docs.openshift.com/container-platform/3.11/install_config/persistent_storage/persistent_storage_nfs.html#nfs-supplemental-groups

### DIFF
--- a/src/libostree/ostree-core-private.h
+++ b/src/libostree/ostree-core-private.h
@@ -30,6 +30,12 @@ G_BEGIN_DECLS
 /* It's what gzip does, 9 is too slow */
 #define OSTREE_ARCHIVE_DEFAULT_COMPRESSION_LEVEL (6)
 
+/* Note the permissive group bits. We want to be liberal here and let individual machines
+ * narrow permissions as needed via umask. This is important in setups where group ownership
+ * can matter for repo management (like OpenShift). */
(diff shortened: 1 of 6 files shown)


cgwalters:
/approve
Optional: Hoist some of those to a shared constant.

jlebon:
I've updated this now to keep the default mode bits centralized!
Though I still need to test that this solves the OCP + NFS + supplemental groups workflow.

jlebon:
OK, tested this in stage OpenShift and seems to work fine, so:
/hold cancel

jlebon:
I didn't touch the sysroot-only paths in this patch, only the paths which are involved in managing an OStree archive repo on the origin server.

cgwalters:
/lgtm
Seems sane, though I wonder if in the future we need to make this configurable or so?  (Or, hm, maybe default ACLs would work?)

Issue #1977, opened 2019-11-22
Inline signature for static-delta superblock


fdanis-oss:
If we use static-delta with all parts inlined into the single file, the `superblock` that precedes everything else is not signed. 
While the commits contained in the single static-delta file are signed so we can check them and operate on trusted data, the `superblock` isn't signed in any way, so it end up operating on untrusted data to:
1. actually find where the trusted data is, and
2. check whether the update is fit for the current device by looking at the collection id stored in the metadata

An idea to sign the `superblock` is to generate a signature of all its data (serialized), except the signature, then add the signature to the `superblock` itself as an additional metadata field.

Verifying the `superblock` would then involve the following steps:
1. fetch data from the `superblock`
2. serialize the fetched data except the signature to reconstruct a `pseudo-superblock`
3. check the computed serialization of the `pseud-superblock` against the original signature

The GVariant parsing code would still operate on untrusted data, but at least we can verify that the extracted contents are trusted.

Signing the whole bundle would avoid operating on untrusted data but would also require to copy the whole bundle away from external storage to protect from malicious storage devices trying to play TOCTOU tricks on us (for instance, an update shipped on a bundle on a usb key that malicously changes the contents after the first read used to verify the contents).

Do you have any suggestion about what would be the best way to address this issue?

Thanks!

Issue #707, opened 2017-02-23
TLS pinning and mirrors


cgwalters:
Currently, if one configures `tls-ca-path` for a remote, we apply that to *all* HTTP requests.  That's good if one is using a remote with a simple `url`, i.e. fetching metadata and content from the same host.

However, I would like to support fetching *metadata* from one server via TLS pinning, and content from S3/Google Cloud Storage/Cloudfront/Akamai etc. 

I feel relatively confident that we could change the meaning of this remote config:

```
[remote "exampleos"]
url=https://exampleos.com/repo
tls-ca-path=/etc/pki/tls/exampleos-ca.pem
contenturl=https://s3.amazonaws.com/exampleos/repo
```
to only apply `tls-ca-path` to `url`, not `contenturl`.

However - we should support potentially pinning the mirrors as well.  I can imagine some people want this - it *does* give you stronger protection against DoS attacks and some people may want confidentiality and not just integrity.  (For example, your OS requires a subscription)

So perhaps we also have `tls-ca-path-content`  And `tls-client-{cert,key}-path-content`.

cgwalters:
Another option is something like:

```
[remote "exampleos"]
url=https://exampleos.com/repo
tls-ca-path=/etc/pki/tls/exampleos-ca.pem
contentremote=exampleos-content

[exampleos-content]
url=https://s3.amazonaws.com/exampleos/repo
```

Then it's easy to apply every option we already have like `tls-ca-path` to the content, without doubling the options.  The code side of this would be easier I think - we'd just reinitialize the fetcher with another config.

PR #1985, opened 2019-12-13
Static-delta's superblock signature support

fdanis-oss:
While the commits contained in the single `static-delta` file are signed so we can check them and operate on trusted data, the `superblock` isn't signed in any way, so it end up operating on untrusted data to:
1. actually find where the trusted data is, and
2. check whether the update is fit for the current device by looking at the collection id stored in the metadata

These commits generate a signature of all `superblock` data (serialized), except inline commits and the signature, then add the signature to the `superblock` itself as an additional metadata field.

New APIs are added to check if a `superblock` is signed, and to verify its signature.

This is part of the _Add support for alternative signature systems_ #1233 and is based on top of _Alternative signing system_ #1878

Fixes: #1977


fdanis-oss:
Hi @cgwalters 
These commits are based on #1878, so only the last 6 commits are relevant to add static-delta's superblock signature
/assign @cgwalters

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -873,3 +873,201 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+              signatures_index++;
+            }
+        }
+      g_variant_unref (item);
+    }
+  if (signatures)
+    signatures[signatures_index] = NULL;
+
+  if (out_value && ret)
+    *out_value = signatures;
```
d4s, line 967: if `ret == FALSE` it is needed to free `signatures`

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -873,3 +873,201 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+ * to %NULL.
+ *
+ * Returns: TRUE if static data superblock has signatures, FALSE otherwise.
+ *
+ * Since: 2019.6
+ */
+gboolean
+ostree_repo_static_delta_is_signed (OstreeRepo       *self,
+                                    const char       *delta_id,
+                                    char           ***out_value,
```
d4s, line 897: for discussion: probably it is better to use GPtrArray instead of GStrv here?
In that case you can avoid playing with memory reallocation below

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -873,3 +873,201 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+  if (!ot_variant_read_fd (superblock_fd, 0,
+                           (GVariantType*)OSTREE_STATIC_DELTA_SUPERBLOCK_FORMAT,
+                           TRUE, &delta_superblock, error))
+    return FALSE;
+
+  /* Check if there are signatures in the superblock for signature engine */
+  const gchar *signature_key = ostree_sign_metadata_key(sign);
+  GVariantType *signature_format = (GVariantType *) ostree_sign_metadata_format(sign);
+  delta_meta = g_variant_get_child_value (delta_superblock, 0);
+  if (delta_meta == NULL)
```
d4s, line 1025: return glnx_throw ()

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -873,3 +873,201 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+}
+
+/**
+ * ostree_repo_static_delta_verify_signature:
+ * @self: Repo
+ * @delta_id: delta path
+ * @sign: Signature engine used to check superblock
+ * @error: Error
+ *
+ * Verify the superblock of static-data, except inline commits,
```
d4s, line 979: Verify pseudo-superblock constructed from original superblock from delta file.
Does not include inlined commits or signatures themselves into signed data.

fdanis-oss:
Hi @cgwalters , can I kindly ask you a review of those commits, please?

cgwalters:
Right, this came up before...I'd have to really dig to find it, but I think where we ended up was to say to use signed summary files.  The summary file contains a checksum of the deltas (though I'd want to double check all this).

There is also ensuring transport integrity - signing is a great "backstop", but I've been pushing "pinned TLS", see e.g. https://github.com/ostreedev/ostree/issues/707 and ... https://pagure.io/fedora-infrastructure/issue/5372 (which links to this thread https://mail.gnome.org/archives/ostree-list/2016-September/msg00015.html )

Is that not sufficient for you?  Anything involving "inline" signatures quickly gets into exactly the problem you're talking about and we'd need to carefully review the code.

d4s:
@cgwalters this is for other use case.
We can prepare the self-containg upgrade bundle as a single file, for instance:
`ostree static-delta generate --from=... --to=...  --inline  --min-fallback-size=1024 --filename update.bundle`

In this case the user need to have only a single file `update.bundle` for offline updates  -- just need to copy it to any mass-storage device with any supported filesystem.
The commit inside of that bundle is signed and is verified with signapi, but here we want to verify the meta-information of delta file itself which is used during extracting data. And need to do that  _before_ extracting the contents of the file and checking the commit to prevent some attack vectors caused by header modification and/or re-creation of the update bundle.

cgwalters:
I understand you guys are probably shipping this already, which complicates things (sorry about taking so long to get to this) - please correct me if that's not the case though!

I agree with the problem statement, but I think a much simpler solution would be to wrap the existing static delta format, i.e. we have as a GVariant layout `a{sv}ay` where

- a{sv}: signatures
- ay existing delta variant

In other words all we've invented is a dead simple *concatenation* of the two.  The building and verification hence becomes trivial, in contrast to what this code seems to be doing today.  A downside is that existing ostree can't parse those static deltas, but that doesn't seem particularly problematic (if you want them verified, the code to verify them needs to be deployed).

fdanis-oss:
@cgwalters I understand this is a really simpler way to do it, but I think the ability to read signed static deltas with legacy ostree version is really better as it simplifies the transition period. With these patches, we can generate only signed deltas, which can update to the new ostree version.

cgwalters:
Split out the prep commit to https://github.com/ostreedev/ostree/pull/2144

cgwalters:
To clarify, is this only about "offline" updates?  Are you using signed summaries today or not for online updates?

cgwalters:
One risk with this approach is that we're relying on GVariant to give us a bit-for-bit exact re-serialization of the data that was signed originally.  More concretely in a scenario where the endpoint device has e.g. glib-2.58, glib-2.62 changes the serialization of `a{sv}` which is used in deltas.

See also https://gitlab.gnome.org/GNOME/glib/commit/d54e10632a575963790d50661ae75692bb159b37
It does seem unlikely that GLib changes anything here, but last time I talked with Ryan he wanted to reserve the right to do so.

fdanis-oss:
Currently, we only use it for "offline" updates

fdanis-oss:
@cgwalters Thanks for explaining the risk with GVariant re-serialization.

I think there should be 2 GVariant layouts:
- for a non signed static delta we should keep the current `OSTREE_STATIC_DELTA_SUPERBLOCK_FORMAT` GVariant layout to avoid breakage of current and other ostree usage like flatpak.
- for signed static delta, the `a{sv}ay`

Are you OK with this?

If I understand correctly, the signature should be computed from the `ay` part, i.e. the `OSTREE_STATIC_DELTA_SUPERBLOCK_FORMAT` serialized, right?  
I'm not sure to understand how to generate it from the ot_variant_builder*() calls in ostree_repo_static_delta_generate()

fdanis-oss:
@cgwalters I uploaded a new version which concatenates the signature to the static delta.

This adds 2 new commits to add signature support to `static-delta apply-offline` and `static-delta show` commands.

This also request Glib2 >= 2.60 to prevent a crash due to an alignment problem when reading the delta variant, see
https://gitlab.gnome.org/GNOME/glib/-/issues/1626

cgwalters:
> @cgwalters I uploaded a new version which concatenates the signature to the static delta.

Thanks!  Will look at this.

> This also request Glib2 >= 2.60 to prevent a crash due to an alignment problem when reading the delta variant, see
https://gitlab.gnome.org/GNOME/glib/-/issues/1626

Ah.

> To avoid this dependency do you think adding a 64-bit integer padding value between signatures and delta variant (i.e. a{sv}tay) could be acceptable and do the trick?

It feels slightly unfortunate to carry forward forever an unused 8 bytes to work around a GLib but that exists today, but OTOH...no one is going to ever care about those 8 bytes given the relative sizes of everything else :smile:

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -873,3 +906,154 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+ * Verify if the static-delta has signature(s).
+ * If the option is not set, or if an error is returned, @out_value will be set
+ * to %NULL.
+ *
+ * Returns: TRUE if static data has signatures, FALSE otherwise.
+ *
+ * Since: 2020.1
+ */
+gboolean
+ostree_repo_static_delta_is_signed (OstreeRepo       *self,
```
cgwalters, line 927: One property of GVariant is it will never fail - you can take any arbitrary bytestream and interpret it as GVariant of any given type and it will always return *something*.

What's going on here is pretty weird because we're basically scraping the first N bytes for any strings matching `ostree.sign.`.  This seems potentially fragile - what if a non-signed delta *happened* to contain those bytes somehow (e.g. we're updating ostree itself and the target binary embedded in the delta payload contains it).

Upon reflection...it might be cleaner to define a magic header for these signed deltas.  For example we could require the first bytes be `OSTSGNDT` (short for "ostree signed delta") and then the data after *that* is GVariant.  The reason for 8 bytes is alignment so we can `mmap()`.  WDYT?

(Also since I've now given bad advice on this PR I am happy to run with the ball a bit and update things to do that)

cgwalters: Or alternatively, we could require signed deltas to have a filename suffix like `.ostsd` or so.   This would seem OK for the use case of "copy delta to USB stick", but I guess is more problematic for storing them in a repository.

fdanis-oss: I pushed a new version with the magic header.
I also changed the order of signatures and delta variant, i.e. `taya{sv}` where:
- t: magic header `OSTSGNDT`
- ay: existing delta variant
- a{sv}: signatures

With this order change GLib takes care of alignment and it removes the dependency on Glib2 >= 2.60

cgwalters:
Are we intentionally not verifying the delta signature as part of applying it, and instead users/systems using ostree would need to invoke `ostree_repo_static_delta_verify_signature()` or its wrapper command before applying a delta?

cgwalters:
I know you've waited a long time on this but I'd also like to try to cut an OSTree release so we keep the train going...any objections to doing one without this?

fdanis-oss:
> Are we intentionally not verifying the delta signature as part of applying it, and instead users/systems using ostree would need to invoke `ostree_repo_static_delta_verify_signature()` or its wrapper command before applying a delta?

Yes, it was intentional
Do you prefer it to have it checked during apply process?

> I know you've waited a long time on this but I'd also like to try to cut an OSTree release so we keep the train going...any objections to doing one without this?

It will be great to have it merged but I don't want to block a release for it

cgwalters:
> Do you prefer it to have it checked during apply process?

What about introducing new `gpg-verify-deltas` and `sign-verify-deltas` options?  They'd have to default to off because everyone is either relying on summary signatures today or doing out-of-band verification.

Basically the situation here seems non-symmetric with how verification for commits and summaries work, and I worry that this "manual verify required" step will surprise people.

Having `gpg-verify` default to `true` in the original ostree was intentional to make people turn it off if they really didn't want it.  Now here we can't default on due to backcompat, but I think the "explicit" aspect here is important, and also having it owned by the repository config.

cgwalters:
Also regarding this:

> Currently, we only use it for "offline" updates

Is this like storing it on a USB stick or so?  We have a `create-usb` subcommand today that could be a good fit for this; basically try to be a bit more opinionated about where the files are stored and create more convention for it.

fdanis-oss:
@cgwalters 
> > Do you prefer it to have it checked during apply process?
> 
> What about introducing new `gpg-verify-deltas` and `sign-verify-deltas` options? They'd have to default to off because everyone is either relying on summary signatures today or doing out-of-band verification. But then you and others who want signed deltas can turn it on.
> 
> Basically the situation here seems non-symmetric with how verification for commits and summaries work, and I worry that this "manual verify required" step will surprise people.
> 
> Having `gpg-verify` default to `true` in the original ostree was intentional to make people turn it off if they really didn't want it. Now here we can't default on due to backcompat, but I think the "explicit" aspect here is important, and also having it owned by the repository config.

I'm not sure to understand it correctly.
I should add a new option set by `ostree config --repo=repo set core.sign-verify-deltas true` to automatically check signature of the delta file.
This verification should be done in `ostree_repo_static_delta_execute_offline`. It will need new parameters for the signature type and keys.

Am I right?

cgwalters:
> I should add a new option set by ostree config --repo=repo set core.sign-verify-deltas true to automatically check signature of the delta file.

Right.

> This verification should be done in ostree_repo_static_delta_execute_offline. It will need new parameters for the signature type and keys.

We'd expect that to be the same signature/keys as for verifying commits, right?  Which already lives in the repository configuration.

(BTW, feel free to push back against this if you disagree.  A lot of pull request reviews are a bit of a discussion/negotiation - I'm not saying we *must* fix this before merging necessarily, but if we're saying merge as is I'd at least like to have it be clear in the docs probably that verification must be done out of band)

d4s:
> > This verification should be done in ostree_repo_static_delta_execute_offline. It will need new parameters for the signature type and keys.
> 
> We'd expect that to be the same signature/keys as for verifying commits, right? Which already lives in the repository configuration.
> 
> (BTW, feel free to push back against this if you disagree. A lot of pull request reviews are a bit of a discussion/negotiation - I'm not saying we _must_ fix this before merging necessarily, but if we're saying merge as is I'd at least like to have it be clear in the docs probably that verification must be done out of band)

in addition to configured options in repo config -- I have a strong opinion what we have to support any available signing subsystem with a random keys in CLI for signing/verification. This is useful in build automation and allows to check deltas prepared for different targets, i.e. signed with keys not aimed for the current system.

fdanis-oss:
> (BTW, feel free to push back against this if you disagree. A lot of pull request reviews are a bit of a discussion/negotiation - I'm not saying we _must_ fix this before merging necessarily, but if we're saying merge as is I'd at least like to have it be clear in the docs probably that verification must be done out of band)

This exactly how I understand it. Up to now I agree with your request :).
Btw, is there an IRC channel or something similar on which we can discuss ostree implementation?

fdanis-oss:
> > I should add a new option set by ostree config --repo=repo set core.sign-verify-deltas true to automatically check signature of the delta file.
> 
> Right.
> 
> > This verification should be done in ostree_repo_static_delta_execute_offline. It will need new parameters for the signature type and keys.
> 
> We'd expect that to be the same signature/keys as for verifying commits, right? Which already lives in the repository configuration.

Yes, I thought to use the `_signapi_init_for_remote` and `_sign_verify_for_remote` calls to do this.
But `_signapi_init_for_remote` needs a `remote_name` and I don't know how to set it as, afaiu, a static delta doesn't have it.
Do you have any idea how I can do it?

So I sent a new version checking the delta signature in `execute_offline`, adding the signature parameters to `static-delta apply-offline` CLI and with the following changes to previous commits:
- `lib/deltas: Add signature check API for static-delta superblock`
  remove public API `ostree_repo_static_delta_is_signed`
  moving `ostree_repo_static_delta_is_signed` code to static function and call it from `ostree_repo_static_delta_verify_signature` before trying to verify signature
- `bin/static-delta: Add command to verify delta signature`
  remove call to `ostree_repo_static_delta_is_signed` as it is done in `ostree_repo_static_delta_verify_signature`
- Update tests as `static-delta verify` call with unsigned delta now return failure

cgwalters:
> Btw, is there an IRC channel or something similar on which we can discuss ostree implementation?

There is `#ostree` on freenode - my time zone is EST.  I was also thinking about setting up periodic "ostree contributor meetings" via some video system.

fdanis-oss:
Finally found a correct way to load keys from well-known directories, as for the signed commit verification.

src/ostree/ot-builtin-static-delta.c
```diff
@@ -438,13 +446,63 @@ ot_static_delta_builtin_apply_offline (int argc, char **argv, OstreeCommandInvoc
...
+  opt_sign_name = opt_sign_name ?: OSTREE_SIGN_NAME_ED25519;
+#endif
+
+  if (opt_sign_name)
+    {
+      sign = ostree_sign_get_by_name (opt_sign_name, error);
+      if (!sign)
+        {
+          g_print("Sign-type not supported\n");
+          return FALSE;
```
cgwalters, line 460: In the "GError model" if `FALSE` is returned then the `error` must be set.

You want just:
`return glnx_throw (error, "Signing type %s is not supported", opt_sign_name);`

src/ostree/ot-builtin-static-delta.c
```diff
@@ -438,13 +446,63 @@ ot_static_delta_builtin_apply_offline (int argc, char **argv, OstreeCommandInvoc
       return FALSE;
     }
 
+#if defined(HAVE_LIBSODIUM)
+  /* Initialize crypto system */
+  opt_sign_name = opt_sign_name ?: OSTREE_SIGN_NAME_ED25519;
```
cgwalters, line 451: Hmm...we don't have this default elsewhere I don't think.  Is it really necessary?  In other code the default comes from the repository configuration or the CLI and that's it.

src/ostree/ot-builtin-static-delta.c
```diff
@@ -438,13 +446,63 @@ ot_static_delta_builtin_apply_offline (int argc, char **argv, OstreeCommandInvoc
...
+          g_autoptr (GVariant) pk = g_variant_new_string(key_ids[i]);
+          if (!ostree_sign_add_pk(sign, pk, error))
+            return FALSE;
+        }
+      if ((n_key_ids == 0) || opt_keysfilename)
+        {
+          g_autoptr (GVariantBuilder) builder = NULL;
+          g_autoptr (GVariant) options = NULL;
+
+          builder = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
```
cgwalters, line 476: Not a blocker but: our code style is "declare-and-initialize" as one unit (C99+) like:
`g_autoptr (GVariantBuilder) builder =  g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));`

src/ostree/ot-builtin-static-delta.c
```diff
@@ -438,13 +446,63 @@ ot_static_delta_builtin_apply_offline (int argc, char **argv, OstreeCommandInvoc
...
+          if (opt_keysdir)
+            g_variant_builder_add (builder, "{sv}", "basedir", g_variant_new_string (opt_keysdir));
+          /* The last chance for verification source -- system files */
+          if (opt_keysfilename)
+            g_variant_builder_add (builder, "{sv}", "filename", g_variant_new_string (opt_keysfilename));
+          options = g_variant_builder_end (builder);
+
+          if (!ostree_sign_load_pk (sign, options, error))
+            {
+              /* If it fails to load system default public keys, consider there no signature engine */
```
cgwalters, line 487: I really don't like silently masking errors.  This is one of the reasons for the refactoring done after the initial signapi landed, see e.g. https://github.com/ostreedev/ostree/pull/2105

Basically if the system is explicitly configured to use e.g. `ed25519`, then if we somehow fail to load that key, that should be a fatal error.

cgwalters (changes requested):

src/ostree/ot-builtin-static-delta.c
fdanis-oss: This is initialized this way in [ostree_builtin_commit()](https://github.com/ostreedev/ostree/blob/master/src/ostree/ot-builtin-commit.c#L879), [ostree_builtin_sign()](https://github.com/ostreedev/ostree/blob/master/src/ostree/ot-builtin-sign.c#L108) and [ostree_builtin_summary()](https://github.com/ostreedev/ostree/blob/master/src/ostree/ot-builtin-summary.c#L106).

Iiuc, the default crypto system is only defined in configuration file for remote repositories.
When we try to apply a static delta offline, there's no reference to a remote repository.

cgwalters:
OK I think we can get this in an address some things as followups.  Thanks for your patience on this!
/approve
/lgtm

cgwalters (approved):

cgwalters:
Followup in https://github.com/ostreedev/ostree/pull/2203

Issue #1983, opened 2019-12-11
Can't get test-switchroot to pass


akiernan:
I've been trying to get `test-switchroot` to pass, but I'm struggling to get it to. I've got `ostree-prepare-root` built statically, but my problems seem to start before then:

```
    libostree 2019.7 (devel)
    features: libsoup gpgme libarchive selinux libmount devel p2p
    ===============


    introspection:                                yes
    Rust (internal oxidation):                    no
    rofiles-fuse:                                 yes
    HTTP backend:                                 libsoup
    "ostree trivial-httpd":                       no
    SELinux:                                      yes
    cryptographic checksums:                      glib
    systemd:                                      yes
    libmount:                                     yes
    libarchive (parse tar files directly):        yes
    static deltas:                                yes (always enabled now)
    O_TMPFILE:                                    yes
    wrpseudo-compat:                              no
    man pages (xsltproc):                         yes
    api docs (gtk-doc):                           no
    installed tests:                              
    gjs-based tests:                              yes
    dracut:                                       no
    mkinitcpio:                                   no
    Static compiler for ostree-prepare-root:      gcc
    Experimental API                              no
    grub2-mkconfig path:                          grub-mkconfig
```

and then running just the switchroot test `sudo make check TESTS=tests/test-switchroot.sh` it ends up failing pretty much immediately with the check for `+ grep -qx /sysroot/var/this_is_ostree_var files`.

I've been fiddling with it for a few hours, without obviously making any progress (mostly getting more and more confused as to what the heck is going on with the tmpfs mounts, which don't seem to work anything like I'd expect).

Is there something obvious I'm missing to make it pass/an example of it running somewhere I can compare against?

akiernan:
It's systemd and libmount, which in combination trigger https://github.com/ostreedev/ostree/blob/master/src/switchroot/ostree-prepare-root.c#L176 so `/var` mounting is deferred and the associated `/var` tests fail.

PR #1987, opened 2019-12-29
Skip /var test if running with systemd and libmount

akiernan:
If running with systemd and libmount then /var mounting is deferred for systemd. Skip the relevant tests in this case as it will always fail.

Fixes: #1983

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -559,6 +559,18 @@ skip_without_user_xattrs () {
     fi
 }
 
+_have_systemd_and_libmount=''
+have_systemd_and_libmount() {
+    if test "${_have_systemd_and_libmount}" = ''; then
+        if [ $(ostree --version | grep -c -e '- systemd' -e '- libmount') -eq 2 ]; then
+            _have_systemd_and_libmount=yes
+        else
+            _have_systemd_and_libmount=no
+        fi
+    fi
+    test ${_have_systemd_and_libmount} = yes
+}
+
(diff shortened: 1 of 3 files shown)


tests/test-switchroot.sh
```diff
@@ -56,7 +56,7 @@ find_in_env() {
 		"$1" "$tmpdir"
 		enter_fs "$tmpdir"
 		ostree-prepare-root /sysroot
-		find /
+		find / \( -path /proc -o -path /sysroot/proc \) -prune -o -print
```
cgwalters: OK with this, but is it really just `/proc`?  Wouldn't e.g. `/sys` and other filesystems be there too?  Oh, I see the script just mounts that.

akiernan: I should've added a comment in there - because processes from outside are visible, recursing down into those proc directory fails when process die.

tests/libtest.sh
```diff
@@ -559,6 +559,18 @@ skip_without_user_xattrs () {
     fi
 }
 
+_have_systemd_and_libmount=''
+have_systemd_and_libmount() {
+    if test "${_have_systemd_and_libmount}" = ''; then
+        if [ $(ostree --version | grep -c -e '- systemd' -e '- libmount') -eq 2 ]; then
```
jlebon: Note `ostree --version` is actually valid parsable YAML. I think we make use of that fact in other places already. But heh, I like this.

PR #1988, opened 2019-12-29
test-switchroot.sh: Find ostree-prepare-root in installed tests

akiernan:
When running with installed tests, ostree-prepare-root (probably) exists in /usr/lib. Add heuristics to look for it based on the directory we're running from.


tests/test-switchroot.sh
```diff
@@ -2,6 +2,16 @@
 
 this_script="${BASH_SOURCE:-$(readlink -f "$0")}"
 
+OSTREE_PREPARE_ROOT=$(dirname "${this_script}")/../ostree-prepare-root
```
akiernan: Looking at this again, given that this file is sourced, I'm really not sure I've got this check in the right place?

PR #1991, opened 2020-01-21
tests/core: Really pick C.UTF-8 locale

dbnicholson:
The case-ignoring regex `^(C|en_US)` will match any locale that starts
with `c`. On my system this is `ca_AD.utf8`, which breaks the test
suite. Instead, use a single regex that includes the joining `.` rather
than 2 separate regexes. This also changes `head` to use the `-n`
option, which has been preferred for at least 10 years in the coreutils
version and is supported by busybox as well.

### DIFF
--- a/tests/libtest-core.sh
+++ b/tests/libtest-core.sh
@@ -41,7 +41,7 @@ assert_not_reached () {
 # If we can't find the locale command assume we have support for C.UTF-8
 # (e.g. musl based systems)
 if type -p locale >/dev/null; then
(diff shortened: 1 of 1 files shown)


dbnicholson:
The [f29-rpmostree](https://s3.amazonaws.com/aos-ci/ghprb/ostreedev/ostree/5135a1e58ade2bfafc8c1fda359540eafd72531e.7.1579628445419709544/index.html) error looks entirely unrelated to these changes. Can this be merged?

PR #1992, opened 2020-01-21
Fix with_... vs have_... in autoconf

akiernan:
There's a couple of confused with_... vs have_... in configure.ac which leads to surprising behaviour when disabling the relevant features.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -504,11 +504,11 @@ AS_IF([ test x$with_libsystemd != xno ], [
     AS_IF([ test x$have_libsystemd = xyes], [
         AC_DEFINE([HAVE_LIBSYSTEMD], 1, [Define if we have libsystemd.pc])
         PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd])
-        with_libsystemd=yes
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -504,11 +504,11 @@ AS_IF([ test x$with_libsystemd != xno ], [
     AS_IF([ test x$have_libsystemd = xyes], [
         AC_DEFINE([HAVE_LIBSYSTEMD], 1, [Define if we have libsystemd.pc])
         PKG_CHECK_MODULES([LIBSYSTEMD], [libsystemd])
-        with_libsystemd=yes
+        have_libsystemd=yes
     ], [
-        with_libsystemd=no
```
jlebon: Hmm, I'm not sure if this is correct. This is resolving `with_libsystemd=maybe` to either `yes` or `no`, right? And we only set it to `yes` if `have_libsystemd` is `yes`.

akiernan: I agree... I'll go back to the issue I bumped into and have another look at it.

akiernan:
Splitting into two PRs

PR #1993, opened 2020-01-21
ci/rpmostree: Bump to 2019.4

dbnicholson:
The vmcheck tests in 2019.3 fail because of an SSH control socket issue
on overlayfs. This is fixed in 2019.4[1]. That has some other changes
such as using Python 3 in tests. The package dependencies have been
synced from the rpm-ostree CI for that.

Unfortunately, this is no longer a totally representative test of f29
since it has 2019.3 in updates. But that's the price you pay for
exercising someone else's CI from your own CI.

1. https://github.com/coreos/rpm-ostree/commit/c89f81c1385ef095616b0e7001926572a20057b2

### DIFF
--- a/ci/rpmostree.sh
+++ b/ci/rpmostree.sh
@@ -6,7 +6,7 @@ set -xeuo pipefail
 # Frozen to a tag for now to help predictability; it's
 # also useful to test building *older* versions since
 # that must work.
-RPMOSTREE_TAG=v2019.3
+RPMOSTREE_TAG=v2019.4
 
 dn=$(dirname $0)
 . ${dn}/libpaprci/libbuild.sh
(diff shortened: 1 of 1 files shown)


dbnicholson:
I'm not sure this is really all that's needed, but I need to open a PR to exercise the CI...

dbnicholson:
This fixed it. I believe an alternative fix would be to stay with 2019.3 but then just cherry pick coreos/rpm-ostree@c89f81c after cloning.

cgwalters:
Thanks so much for looking at this!  We're still in a transition period with CI - rpm-ostree's CI was just reworked, and hopefully we'll soon unify and extend them using the coreos-assembler work.

/approve
/lgtm

Issue #1921, opened 2019-09-15
ostree on live PXE systems


bgilbert:
When Fedora CoreOS is booted as a live PXE system, ostree would ideally function in read-only mode, where it refuses to deploy updates (since that isn't a meaningful operation on a live system) but still provides read-only status info.

rpm-ostreed currently refuses to start on such systems because ostree is unhappy:

```
Error setting up sysroot: Unexpected state: /run/ostree-booted found and in / sysroot but not in a booted deployment
```

Live PXE systems don't need anything in `/boot`, since the kernel and initramfs are provided externally.  Fedora CoreOS could generate a fake BLS entry if needed, though note that `/proc/cmdline` doesn't contain an `ostree` argument so we'd need to deal with that also (#1920).

What's the best way to proceed here?

cgwalters:
We could "standardize" `/run/ostree-live` here?

cgwalters:
https://github.com/ostreedev/ostree/pull/1995 is a WIP PR for this.

jlebon:
This oddly works now in a locally built FCOS live ISO and PXE artifacts:

```
[root@localhost ~]# rpm-ostree status
State: idle
Deployments:
● ostree://fedora:fedora/x86_64/coreos/testing-devel
                   Version: 31.20200526.dev.0 (2020-05-26T19:18:19Z)
                    Commit: f1bbe310d7ec3ac03e1dba2c819f43daf3c7f6c88c000b20892d7d4a25881f2e
              GPGSignature: (unsigned)
[root@localhost ~]# cat /proc/cmdline
BOOT_IMAGE=fedora-coreos-31.20200526.dev.0-live-kernel-x86_64 ip=dhcp rd.neednet=1 initrd=fedora-coreos-31.20200526.dev.0-live-initramfs.x86_64.img console=tty0 console=ttyS0 ignition.config.url=http://192.168.122.1:8000/config-fcos.ign ignition.firstboot ignition.platform.id=metal BOOTIF=01-52-54-00-44-96-c9
```

jlebon:
Repurposing this card now to just generalize it to having ostree be "live system-aware".

cgwalters:
Moving this comment over https://github.com/coreos/fedora-coreos-config/pull/3009#issuecomment-2237350414

> I think this is basically another instance of ostreedev/ostree#1921 (which I've just retitled); so then ostree-prepare-root would detect that this is a live environment and just not even try to set up composefs there.

[...]

PR #1995, opened 2020-01-22
WIP: ostree live support

cgwalters:
WIP for https://github.com/ostreedev/ostree/issues/1921


cgwalters:
OK, fixed this up and got it to work.  No CI here yet, but I did test this manually with an override and `cosa buildextend-live`.

I've got some thoughts/code in motion for support for testing on "live" systems - it would be most ergonomic with https://github.com/coreos/ignition/issues/935

jlebon:
I think this needs a rebase too on top of fixed CI.

src/libostree/ostree-sysroot.h
```diff
@@ -32,6 +32,16 @@ G_BEGIN_DECLS
 #define OSTREE_IS_SYSROOT(obj) \
   (G_TYPE_CHECK_INSTANCE_TYPE ((obj), OSTREE_TYPE_SYSROOT))
 
+/**
+ * OSTREE_SYSROOT_LIVE_PATH:
+ *
+ * The existence of this file declares that the system
+ * is booted "live" (fully in RAM).
+ *
+ * Since: 2020.1
```
jlebon: 2020.4

src/libostree/ostree-sysroot.h
```diff
@@ -32,6 +32,16 @@ G_BEGIN_DECLS
 #define OSTREE_IS_SYSROOT(obj) \
   (G_TYPE_CHECK_INSTANCE_TYPE ((obj), OSTREE_TYPE_SYSROOT))
 
+/**
+ * OSTREE_SYSROOT_LIVE_PATH:
+ *
+ * The existence of this file declares that the system
+ * is booted "live" (fully in RAM).
```
jlebon: I think one important thing worth noting here is that unlike the seemingly similar `/run/ostree-booted`, this isn't written by OSTree itself. I.e. it's an agreed-upon API between OSTree and OS vendors. Maybe something like:

> It is written by external code which drive OSTree in live contexts (e.g. live ISOs or live PXE boots).

?

src/libostree/ostree-sysroot.c
```diff
@@ -284,6 +284,9 @@ gboolean
 _ostree_sysroot_ensure_writable (OstreeSysroot      *self,
                                  GError            **error)
 {
+  if (self->root_is_ostree_live)
+    return glnx_throw (error, "sysroot does not support persistent changes");
```
jlebon: Minor: how about prefixing with `found /run/ostree-live`?

jlebon:
Hmm OK, so just thinking about this some more and looking at the FCOS side; do we want to think more deeply about what the ideal integration would look like between the OS and libostree?

For example, right now FCOS has to play hacks with the cmdline to fool `ostree-prepare-root`: https://github.com/coreos/fedora-coreos-config/blob/f49ef1237d6b75f39410147001d14f594daf79cf/overlay.d/05core/usr/lib/dracut/modules.d/20live/ostree-cmdline.sh. Is this how we want things to work long-term?

I'm wondering if we should instead have a `ostree=live` karg or something that signals `ostree-prepare-root` to (1) automatically detect the only deployment in `/sysroot` and prepare root on that, and (2) write the `/run/ostree-live` stamp file.

/cc @bgilbert

cgwalters:
> I'm wondering if we should instead have a ostree=live karg or something that signals ostree-prepare-root to (1) automatically detect the only deployment in /sysroot and prepare root on that, and (2) write the /run/ostree-live stamp file.

Yeah, that seems reasonable.

I guess there's a much bigger topic around whether it's worth doing the "deployment root" for live systems at all.  It might not be really hard to rework things so that libostree supports just not switching root in this case and operating in read-only mode.   But OTOH, FCOS already shipped a live system built this way.

cgwalters:
OTOH live systems don't support in-place updates, so there aren't any backcompat constraints really; we are free to change the API between live systems and libostree.

cgwalters:
OK in the meantime I split off the prep patch in https://github.com/ostreedev/ostree/pull/2049

cgwalters:
Gah the rabbit hole here is deep.  So today, the FCOS live bits create an overlayfs for `/etc`.  But I just realized that since FCOS also has the [readonly sysroot](https://github.com/ostreedev/ostree/pull/1767) bits on, `ostree-remount.service` would ordinarily try to set up the bind mount for `/etc` except it doesn't do that only because `ConditionKernelCmdLine=ostree` isn't met.

PR #1996, opened 2020-01-25
build: fix systemd feature advertisement

akiernan:
17db0f15a798 ("configure: add option for libsystemd") exposed
--without-libsystemd to allow systemd to be disabled even if the systemd
pkgconfig script was present, introducing a new variable
with_libsystemd; there are now three, almost identical variables:

- with_libsystemd [yes, no, maybe] - controlled by --without-libsystemd,
  resolved into yes/no by the initial checks
- have_libsystemd [yes, no, <undefined>] - only set if with_libsystemd
  is yes/maybe, otherwise undefined
- with_systemd [yes, <undefined>] - yes if have_systemd is yes,
  otherwise undefined

with_systemd is the earliest variable and was previously set by a set of
checks for dracut and mkinitcpio. These checks were changed for a
systemd check in 9e2763106be0 ("lib: Use sd_journal directly
(optionally)"). This commit also introduced BUILDOPT_LIBSYSTEMD, which
will always match BUILDOPT_SYSTEMD.

Fix the confusion by removing with_systemd which will always be yes when
with_libsystemd=yes, or undefined if with_libsystemd=no. We can ignore
the with_libsystemd=maybe case because it will always be resolved into
yes/no before with_systemd is set.

And replace all uses of BUILDOPT_SYSTEMD with BUILDOPT_LIBSYSTEMD, since
they again always match.

This fixes both the advertised features and the summary output when
systemd is disabled by using with_libsystemd which is always defined.

Fixes: 5c62a7e4d0a5 ("build: Expose systemd in OSTREE_FEATURES")
Fixes: 17db0f15a798 ("configure: add option for libsystemd")
Supersedes: #1992

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -214,7 +214,7 @@ libostree_1_la_CFLAGS += $(OT_DEP_AVAHI_CFLAGS)
 libostree_1_la_LIBADD += $(OT_DEP_AVAHI_LIBS)
 endif
 
-if BUILDOPT_LIBSYSTEMD
+if BUILDOPT_SYSTEMD
 libostree_1_la_CFLAGS += $(LIBSYSTEMD_CFLAGS)
 libostree_1_la_LIBADD += $(LIBSYSTEMD_LIBS)
 endif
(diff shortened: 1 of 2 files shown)


cgwalters:
Awesome, thanks so much for cleaning this up!
/approve
/lgtm

PR #1998, opened 2020-02-03
Add .cci.jenkinsfile

cgwalters:
See https://github.com/jlebon/coreos-ci

This is just a start to test, cut down from the rpm-ostree version.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,0 +1,36 @@
+// See https://github.com/jlebon/coreos-ci
+// This was originally copied from the rpm-ostree one
+@Library('github.com/coreos/coreos-ci-lib@master') _
+
+COSA_IMAGE = 'quay.io/coreos-assembler/coreos-assembler:latest'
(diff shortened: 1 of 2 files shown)


cgwalters:
OK...debating a bit whether to merge this as is or to try to just push forward and drop PAPR testing entirely.

cgwalters:
Leaning a bit towards merging as is, so if someone could drop a `/lgtm` that'd be great!

.cci.jenkinsfile
```diff
@@ -0,0 +1,36 @@
+// See https://github.com/jlebon/coreos-ci
+// This was originally copied from the rpm-ostree one
+@Library('github.com/coreos/coreos-ci-lib@master') _
+
+COSA_IMAGE = 'quay.io/coreos-assembler/coreos-assembler:latest'
+COSA_BUILDROOT_IMAGE = 'registry.svc.ci.openshift.org/coreos/cosa-buildroot:latest'
+
+stage("Build") {
+parallel rpms: {
```
jlebon: Maybe just `build`?

cgwalters: Heh I saw that and was going to fix it but didn't want to wait for a whole other cycle of CI runs...

PR #1999, opened 2020-02-03
ci: Replace PAPR with CoreOS CI 

cgwalters:
Move the alternative builds into the Jenkinsfile.

Update it to do a FCOS build + kola run.

We drop the flatpak/rpm-ostree runs for now; the former
will needs some work to do the automatic virt bits.  The
latter I think we can circle back to when we e.g. figure
out how to include rpm-ostree's tests in kola runs.


cgwalters:
/override papr-stub

Yes, a pass! :tada:

.cci.jenkinsfile
```diff
@@ -34,3 +62,51 @@ codestyle: {
...
+  }
+},
+buildopts: {
+  coreos.pod(image: COSA_BUILDROOT_IMAGE, runAsUser: 0) {
+      checkout scm
+      sh """
+        set -xeuo pipefail
+        git submodule update --init
+
+        git clean -dfx && git submodule foreach git clean -dfx
```
jlebon: Hmm, we don't need this, right? Each pod should get a fresh workspace.

.papr.yml
```diff
@@ -1,145 +1,9 @@
-# This suite skips the RPMs and does the build+unit tests in a container
+# Stub until we drop PAPR entirely
```
jlebon: Mehh... let's just nuke it if we're not gonna leave any job in it? Once we drop PAPR here, I think it's just bubblewrap really that needs to migrate off before we can kill it entirely!

.cci.jenkinsfile
```diff
@@ -5,8 +5,8 @@
 COSA_IMAGE = 'quay.io/coreos-assembler/coreos-assembler:latest'
 COSA_BUILDROOT_IMAGE = 'registry.svc.ci.openshift.org/coreos/cosa-buildroot:latest'
 
-stage("Build") {
-parallel rpms: {
+stage("Build and sanitychecks") {
```
jlebon: Bikeshed/optional: maybe just `Build` ? And then have the parallel branches be e.g. `normal`, `minimal`, and `codestyle`?

.cci.jenkinsfile
```diff
@@ -15,8 +15,36 @@ parallel rpms: {
         git fetch origin --tags
         git submodule update --init
 
+        env SKIP_INSTALLDEPS=1 ./ci/build.sh
+      """
+      try {
+        sh """
+        make check
```
jlebon: I'd put this in a separate step, e.g. `Check` or something.
When tests fail, having more distinct and descriptive stages makes it easier to narrow down what went wrong I find. But yeah, totally good to just tweak this later too!

.cci.jenkinsfile
cgwalters: Yeah, but I'm running multiple builds here.  I could split it into a separate pod but...some worry about the cost of repeated source cloning.

.papr.yml
cgwalters: I worried that papr would barf if there wasn't a `papr.yaml` until we removed the webhooks entirely?

jlebon: It should be able to handle it gracefully.

.cci.jenkinsfile
jlebon: I don't follow; this branch allocates its own pod already, no?

cgwalters: You're right that the first clean is unnecessary, but I was trying to be defensive when the section is copy-pasted.  However, I switched this now to use a `git worktree` which is actually nicer anyways.

jlebon: Ahh gotcha, I somehow missed the copy/paste pattern you were going for. But yeah, `git worktree` is cleaner.

Issue #1900, opened 2019-08-20
Multiple device trees for OSTree deploying


starnight:
Thanks to William's effort.  OSTree supports a device tree file for a single model of hardware since commit 720e2ec9bc40 "Add support for devicetree files alongside the kernel and initramfs".

However, there are lots of embedded systems and single board computers in the world.  Each of them has a specified device tree.

According to current mechanism, if we have "n" models of hardware, then we have "n" device trees and have to build "n" ostrees for installation and updating.  This will be a big release management problem for common distributions.

Actually, this issue is already recognized by William of the existing functionality:
https://github.com/ostreedev/ostree/pull/1411

"It may be useful to extend device-tree support in a number ways in the future. Device trees depend on many details of the hardware they support. This makes them unlike kernels, which may support many different hardware variants as long as the instruction-set matches. This means that a ostree tree created with a device-tree in this manner will only boot on a single model of hardware. This is sufficient for my purposes, but may not be for others'."

This problem is particularly relevant now as Linux & Mesa mainline support for ARM boards is so much better than before, e.g. RPi, Rockchip, Amlogic.  And as the list the names of all the different RPi3 device trees so that it's clear that the existing ostree implementation doesn't really work even if you only want to make a "Raspberry Pi 3" image.

The device tree source list of Raspberry Pi 3 in Linux mainline kernel:
arch/arm64/boot/dts/broadcom/bcm2837-rpi-3-a-plus.dts
arch/arm64/boot/dts/broadcom/bcm2837-rpi-3-b.dts
arch/arm64/boot/dts/broadcom/bcm2837-rpi-3-b-plus.dts
arch/arm64/boot/dts/broadcom/bcm2837-rpi-cm3-io3.dts

Besides, the device tree is closely linked to the kernel version, and should be updated whenever the kernel is updated.  For example, Arch Linux ARM and Fedora aarch64 install the entire tree of built device trees directly alongside the kernel & initramfs, and then rely on the boot loader (u-boot) to select the right device tree
at boot time.

What is the best way forward to solving this limitation?

Perhaps the entire device tree folder hierarchy should be deployed

agners:
We do have the same problem and currently carry a rather hakish patch which copies all device tree's in the given source dir. However, checksum is not accounted for currently.

I was thinking about this a bit more, and I think the best way to implement this would be to only support the new deployment model (in `/usr/lib/modules/$kver`), this will simplify the code as we do not have to filter by checksum when searching for device trees.

[...]

PR #2001, opened 2020-02-04
deploy: support devicetree directory

agners:
Add support for a devicetree directory at /usr/lib/modules/$kver/dtb/.
In ARM world a general purpose distribution often suppports multiple
boards with a single operating system. However, OSTree currently only
supports a single device tree, which does not allow to use the same
OSTree on different ARM machines. In this scenario typically the boot
loader selects the effective device tree.

This adds device tree directory support for the new boot artefact
location under /usr/lib/modules. If the file `devicetree` does not
exist, then the folder dtb will be checked. All devicetrees are hashed
into the deployment hash. This makes sure that even a single devicetree
change leads to a new deployment and hence can be rolled back.

The loader configuration has a new key "devicetreepath" which contains
the path where devicetrees are stored. This is also written to the
U-Boot variable "fdt_path". The boot loader is expected to use this path
to load a particular machines device tree from.

Closes: #1900
Signed-off-by: Stefan Agner <stefan.agner@toradex.com>

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -217,6 +217,79 @@ dirfd_copy_attributes_and_xattrs (int            src_parent_dfd,
   return TRUE;
 }
 
+static gint
+str_sort_cb (gconstpointer name_ptr_a, gconstpointer name_ptr_b)
+{
+  const gchar *name_a = *((const gchar **) name_ptr_a);
+  const gchar *name_b = *((const gchar **) name_ptr_b);
+
+  return g_strcmp0 (name_a, name_b);
+}
+
+static gboolean
+checksum_dir_recurse (int          dfd,
+                  const char      *path,
+                  OtChecksum      *checksum,
+                  GCancellable    *cancellable,
+                  GError         **error)
+{
+  g_auto(GLnxDirFdIterator) dfditer = { 0, };
+  g_autoptr (GPtrArray) d_entries = g_ptr_array_new_with_free_func (g_free);
+
+  if (!glnx_dirfd_iterator_init_at (dfd, path, TRUE, &dfditer, error))
+    return FALSE;
+
+  while (TRUE)
+    {
+      struct dirent *dent;
+
+      if (!glnx_dirfd_iterator_next_dent (&dfditer, &dent, cancellable, error))
+        return FALSE;
+
+      if (dent == NULL)
+        break;
+
+      g_ptr_array_add (d_entries, g_strdup (dent->d_name));
+    }
+
+  /* File systems do not guarantee dir entry order, make sure this is
+   * reproducable
+   */
+  g_ptr_array_sort(d_entries, str_sort_cb);
+
+  for (gint i=0; i < d_entries->len; i++)
+    {
+      const gchar *d_name = (gchar *)g_ptr_array_index (d_entries, i);
+      struct stat stbuf;
+
+      if (!glnx_fstatat (dfditer.fd, d_name, &stbuf,
+                         AT_SYMLINK_NOFOLLOW, error))
+        return FALSE;
+
+      if (S_ISDIR (stbuf.st_mode))
+        {
+          if (!checksum_dir_recurse(dfditer.fd, d_name, checksum, cancellable, error))
+            return FALSE;
+        }
+      else
+        {
+          int fd;
+
+          if (!ot_openat_ignore_enoent (dfditer.fd, d_name, &fd, error))
+            return FALSE;
+          if (fd != -1)
+            {
+              g_autoptr(GInputStream) in = g_unix_input_stream_new (fd, FALSE);
+              if (!ot_gio_splice_update_checksum (NULL, in, checksum, cancellable, error))
+                return FALSE;
+            }
+        }
+
+    }
+
+  return TRUE;
+}
+
 static gboolean
 copy_dir_recurse (int              src_parent_dfd,
                   int              dest_parent_dfd,
@@ -1065,6 +1138,9 @@ get_kernel_from_tree_usrlib_modules (int                  deployment_dfd,
   g_clear_object (&in);
   glnx_close_fd (&fd);
 
+  /* Check for /usr/lib/modules/$kver/devicetree first, if it does not
+   * exist check for /usr/lib/modules/$kver/dtb/ directory.
+   */
   if (!ot_openat_ignore_enoent (ret_layout->boot_dfd, "devicetree", &fd, error))
     return FALSE;
   if (fd != -1)
@@ -1074,6 +1150,23 @@ get_kernel_from_tree_usrlib_modules (int                  deployment_dfd,
       in = g_unix_input_stream_new (fd, FALSE);
       if (!ot_gio_splice_update_checksum (NULL, in, &checksum, cancellable, error))
         return FALSE;
+    }
+  else
+    {
+      struct stat stbuf;
+      /* Check for dtb directory */
+      if (!glnx_fstatat_allow_noent (ret_layout->boot_dfd, "dtb", &stbuf, 0, error))
+        return FALSE;
+
+      if (errno == 0 && S_ISDIR (stbuf.st_mode))
+        {
+          /* devicetree_namever set to NULL indicates a complete directory */
+          ret_layout->devicetree_srcpath = g_strdup ("dtb");
+          ret_layout->devicetree_namever = NULL;
+
+          if (!checksum_dir_recurse(ret_layout->boot_dfd, "dtb", &checksum, cancellable, error))
+            return FALSE;
+        }
     }
 
   g_clear_object (&in);
@@ -1730,15 +1823,24 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
 
   if (kernel_layout->devicetree_srcpath)
     {
-      g_assert (kernel_layout->devicetree_namever);
-      if (!glnx_fstatat_allow_noent (bootcsum_dfd, kernel_layout->devicetree_namever, &stbuf, 0, error))
-        return FALSE;
-      if (errno == ENOENT)
+      /* If devicetree_namever is set a single device tree is deployed */
+      if (kernel_layout->devicetree_namever)
         {
-          if (!install_into_boot (repo, sepolicy, kernel_layout->boot_dfd, kernel_layout->devicetree_srcpath,
-                                  bootcsum_dfd, kernel_layout->devicetree_namever,
-                                  sysroot->debug_flags,
-                                  cancellable, error))
+          if (!glnx_fstatat_allow_noent (bootcsum_dfd, kernel_layout->devicetree_namever, &stbuf, 0, error))
+            return FALSE;
+          if (errno == ENOENT)
+            {
+              if (!install_into_boot (repo, sepolicy, kernel_layout->boot_dfd, kernel_layout->devicetree_srcpath,
+                                      bootcsum_dfd, kernel_layout->devicetree_namever,
+                                      sysroot->debug_flags,
+                                      cancellable, error))
+                return FALSE;
+            }
+        }
+      else
+        {
+          if (!copy_dir_recurse(kernel_layout->boot_dfd, bootcsum_dfd, kernel_layout->devicetree_srcpath,
+                                sysroot->debug_flags, cancellable, error))
             return FALSE;
         }
     }
@@ -1849,6 +1951,15 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
     {
       g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", kernel_layout->devicetree_namever, NULL);
       ostree_bootconfig_parser_set (bootconfig, "devicetree", boot_relpath);
+    }
+  else if (kernel_layout->devicetree_srcpath)
+    {
+      /* If devicetree_srcpath is set but devicetree_namever is NULL, then we
+       * want to point to a whole directory of device trees.
+       * See: https://github.com/ostreedev/ostree/issues/1900
+       */
+      g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", kernel_layout->devicetree_srcpath, NULL);
+      ostree_bootconfig_parser_set (bootconfig, "fdtdir", boot_relpath);
     }
 
   /* Note this is parsed in ostree-impl-system-generator.c */
--- a/src/libostree/ostree-bootloader-uboot.c
+++ b/src/libostree/ostree-bootloader-uboot.c
@@ -144,6 +144,10 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot     *self,
       if (val)
         g_ptr_array_add (new_lines, g_strdup_printf ("fdt_file%s=%s", index_suffix, val));
 
+      val = ostree_bootconfig_parser_get (config, "fdtdir");
+      if (val)
+        g_ptr_array_add (new_lines, g_strdup_printf ("fdtdir%s=%s", index_suffix, val));
+
       val = ostree_bootconfig_parser_get (config, "options");
       if (val)
         {
--- a/tests/test-admin-deploy-uboot.sh
+++ b/tests/test-admin-deploy-uboot.sh
@@ -25,9 +25,11 @@ set -euo pipefail
 . $(dirname $0)/libtest.sh
 
 # Exports OSTREE_SYSROOT so --sysroot not needed.
-setup_os_repository "archive" "uboot"
+kver="3.6.0"
+modulesdir="usr/lib/modules/${kver}"
+setup_os_repository "archive" "uboot" ${modulesdir}
 
-extra_admin_tests=1
+extra_admin_tests=2
 
 . $(dirname $0)/admin-test.sh
 
@@ -52,3 +54,27 @@ assert_file_has_content sysroot/boot/uEnv.txt "kernel_image2="
 assert_file_has_content sysroot/boot/uEnv.txt "kernel_image3="
 
 echo "ok merging uEnv.txt files"
+
+cd ${test_tmpdir}
+os_repository_new_commit "uboot test" "test with device tree directory"
+
+devicetree_path=osdata/${modulesdir}/dtb/asoc-board.dtb
+devicetree_overlay_path=osdata/${modulesdir}/dtb/overlays/overlay.dtbo
+
+mkdir -p osdata/${modulesdir}/dtb
+echo "a device tree" > ${devicetree_path}
+mkdir -p osdata/${modulesdir}/dtb/overlays
+echo "a device tree overlay" > ${devicetree_overlay_path}
+
+bootcsum=$(
+  (echo "new: a kernel uboot test" && echo "new: an initramfs uboot test" &&
+    cat ${devicetree_path} ${devicetree_overlay_path} ) |
+  sha256sum | cut -f 1 -d ' ')
+
+${CMD_PREFIX} ostree --repo=testos-repo commit --tree=dir=osdata/ -b testos/buildmaster/x86_64-runtime
+${CMD_PREFIX} ostree admin upgrade --os=testos
+assert_file_has_content sysroot/boot/uEnv.txt "fdtdir="
+assert_file_has_content sysroot/boot/ostree/testos-${bootcsum}/dtb/asoc-board.dtb 'a device tree'
+assert_file_has_content sysroot/boot/ostree/testos-${bootcsum}/dtb/overlays/overlay.dtbo 'a device tree overlay'
+
+echo "ok deploying fdtdir"


cgwalters:
/approve
/ok-to-test
This seems sane to me; I don't have expertise in ARM and DT bits personally.  Would love to get a review from someone who does.

agners:
@starnight mind having a look at this (since you created the issue)?

@gizero or @guyshapiro since you contributed to the U-Boot support previously, maybe you can have a look as well?

akiernan:
Give me a few days and I'll test it here. Though our primary use case is DTBs embedded in FIT images, but at least I can check it doesn't break anything.

starnight:
@agners Thanks for your PR.  We are going to find a way to test it on our distribution.

agners:
@d4s since you are supporting ARM devices too, you might be interested in that too?

akiernan:
Just pulled this into a test build, works for me. My only query would be the trailing slash on `fdt_path` which we don't have on other paths which are written to uEnv.txt

agners:
@akiernan seems reasonable, dropped the trailing slash.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1730,16 +1766,46 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+
+              if (!glnx_fstatat_allow_noent (bootcsum_dfd, dent->d_name, &stbuf, 0, error))
+                return FALSE;
+              if (errno == ENOENT)
+                {
+                  if (!install_into_boot (repo, sepolicy, dfditer.fd, dent->d_name,
+                                    bootcsum_dfd, dent->d_name,
+                                    sysroot->debug_flags,
+                                    cancellable, error))
+                    return FALSE;
```
jlebon: Just making sure I understand here; this is installing all the subdirs found in `dtb/` directly under `/ostree/$osname-$bootcsum` right? Wouldn't it be cleaner to keep them together in `/ostree/$osname-$bootcsum/dtb` instead?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1730,16 +1766,46 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+                                      bootcsum_dfd, kernel_layout->devicetree_namever,
+                                      sysroot->debug_flags,
+                                      cancellable, error))
+                return FALSE;
+            }
+        }
+      else
+        {
+          g_auto(GLnxDirFdIterator) dfditer = { 0, };
+          if (!glnx_dirfd_iterator_init_at (kernel_layout->boot_dfd, "dtb/", FALSE, &dfditer, error))
```
jlebon: Can we make this `devicetree_srcpath` instead of hardcoding `dtb/`?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1850,6 +1916,11 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
       g_autofree char * boot_relpath = g_strconcat ("/", bootcsumdir, "/", kernel_layout->devicetree_namever, NULL);
       ostree_bootconfig_parser_set (bootconfig, "devicetree", boot_relpath);
     }
+  else if (kernel_layout->devicetree_srcpath)
+    {
```
jlebon: Maybe another comment here too like:

```c
# If devicetree_srcpath is set but devicetree_namever is NULL, then we want to point to a whole directory of device trees.
# See: https://github.com/ostreedev/ostree/issues/1900
```

src/libostree/ostree-sysroot-deploy.c
agners: Yes. I remember I was thinking about creating a directory there, but don't remember why I dropped it. I guess it was due to the additional complexity. But if that is preferred, we can of course add it.

src/libostree/ostree-sysroot-deploy.c
agners: Agreed, should be the same except the missing slash, but that shouldn't really matter.

starnight:
Hi @agners,

Thanks for your effort!  I have backported the commit to our [OSTree branch](https://github.com/endlessm/ostree) and did some experiment for the `dtb` folder.

Here is the original `uEnv.txt`:
```
$ cat /boot/uEnv.txt
kernel_image=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/vmlinuz-5.4.0-7-generic
ramdisk_image=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/initramfs-5.4.0-7-generic.img
fdt_file=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/devicetree-5.4.0-7-generic
bootargs=rw splash plymouth.ignore-serial-consoles quiet loglevel=0 ostree=/ostree/boot.1/eos/f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/0
```

I try to make ostree produce `dtb` folder by removing the devicetree file.
Here is the steps and result:
```
$ mkdir fdtpath
$ cd fdtpath/
$ sudo ostree admin status
* eos 826ebcc82551e81b0b5b0d06ea8cbaea3fb756d632a2bfec142765cd10b1532b.0
    Version: 3.8.0~200323-220027
    origin refspec: eos:os/eos/arm64/master
    GPG: Signature made Tue 24 Mar 2020 06:54:03 AM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Sun 01 Jul 2029 12:18:11 AM CST
    GPG: Signature made Tue 24 Mar 2020 06:54:06 AM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Wed 10 Jun 2026 01:47:38 AM CST
$ sudo ostree checkout 826ebcc82551e81b0b5b0d06ea8cbaea3fb756d632a2bfec142765cd10b1532b
$ cd 826ebcc82551e81b0b5b0d06ea8cbaea3fb756d632a2bfec142765cd10b1532b/
$ sudo rm lib/modules/5.4.0-7-generic/devicetree 
$ sudo ostree commit -b test --tree=dir=`pwd`
960c8b88ed4471fc5cdc8d97a469506dfff0a89eaddd4b91315ab78804c14daf
$ sudo ostree admin deploy test
Copying /etc changes: 17 modified, 0 removed, 16 added
Transaction complete; bootconfig swap: yes; deployment count change: 1
Freed objects: 13.9 MB
$ cat /boot/uEnv.txt 
kernel_image=/ostree/eos-54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6/vmlinuz-5.4.0-7-generic
ramdisk_image=/ostree/eos-54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6/initramfs-5.4.0-7-generic.img
bootargs=rw splash plymouth.ignore-serial-consoles quiet loglevel=0 ostree=/ostree/boot.0/eos/54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6/0
kernel_image2=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/vmlinuz-5.4.0-7-generic
ramdisk_image2=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/initramfs-5.4.0-7-generic.img
fdt_file2=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/devicetree-5.4.0-7-generic
bootargs2=rw splash plymouth.ignore-serial-consoles quiet loglevel=0 ostree=/ostree/boot.0/eos/f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/0
$ ls /boot/ostree/eos-54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6/
initramfs-5.4.0-7-generic.img  vmlinuz-5.4.0-7-generic
$ ls /boot/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/
devicetree-5.4.0-7-generic  initramfs-5.4.0-7-generic.img  vmlinuz-5.4.0-7-generic
$ sudo ostree admin status
  eos 960c8b88ed4471fc5cdc8d97a469506dfff0a89eaddd4b91315ab78804c14daf.0 (pending)
    origin refspec: test
* eos 826ebcc82551e81b0b5b0d06ea8cbaea3fb756d632a2bfec142765cd10b1532b.0
    Version: 3.8.0~200323-220027
    origin refspec: eos:os/eos/arm64/master
    GPG: Signature made Tue 24 Mar 2020 06:54:03 AM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Sun 01 Jul 2029 12:18:11 AM CST
    GPG: Signature made Tue 24 Mar 2020 06:54:06 AM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Wed 10 Jun 2026 01:47:38 AM CST
```

* Original ostree folder for boot at `/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db` locates at INDEX **1** in `uEnv.txt` after new commit `test` is deployed.
* New ostree folder for boot at `/ostree/eos-54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6` locates at INDEX **0** in `uEnv.txt` after new commit `test` is deployed.
* Because new ostree is at INDEX **0**, next boot will use it, instead of the original one.
* New ostree has only kernel and ramdisk. Original ostree has kernel, ramdisk and devicetree FIT image. But both of them do not have the `dtb` folder. `dtb` should be produced by the backported commit and is expected appears in the new ostree for boot. This will makes next boot failed, due to no device FIT image, nor `dtb` folder. And it does boot failed.

I am not sure no `dtb` in the new ostree is caused by older OSTree which we are using, or I missed something else.  Any idea for the experiment will be appreciated.

agners:
@starnight not sure if I understand correctly, but the `dtb` directory must be put in place by the build system. While OSTree could create a empty folder, it would not really help, somebody needs to deploy the device tree here.

agners:
Another thing we might consider is copying directories recursively. This would be helpful if device tree overlays are stored in a sub directory (e.g. RPi makes use of overlays, and often they are in a directory "overlays")

starnight:
Thanks to @agners' reply.  Yes!  I made the `dtb` directory wrong.

To have the same reference base, I prepared the upstream's OSTree with commit dd2a305 ("deploy: support devicetree directory").

I re-do new ostree and deploy it.
```
$ mkdir fdtpath
$ cd fdtpath/
$ ostree admin status
* eos 826ebcc82551e81b0b5b0d06ea8cbaea3fb756d632a2bfec142765cd10b1532b.0
    Version: 3.8.0~200323-220027
    origin refspec: eos:os/eos/arm64/master
    GPG: Signature made Tue 24 Mar 2020 06:54:03 AM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Sun 01 Jul 2029 12:18:11 AM CST
    GPG: Signature made Tue 24 Mar 2020 06:54:06 AM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Wed 10 Jun 2026 01:47:38 AM CST
$ sudo ostree checkout 826ebcc82551e81b0b5b0d06ea8cbaea3fb756d632a2bfec142765cd10b1532b
$ cd 826ebcc82551e81b0b5b0d06ea8cbaea3fb756d632a2bfec142765cd10b1532b/
$ sudo rm usr/lib/modules/5.4.0-7-generic/devicetree
$ sudo cp -r usr/lib/firmware/5.4.0-7-generic/device-tree usr/lib/modules/5.4.0-7-generic/dtb
$ sudo ostree commit -b test --tree=dir=`pwd`
$ ls usr/lib/modules/5.4.0-7-generic/
config  initramfs.img  kernel         modules.alias.bin  modules.builtin.bin  modules.dep.bin  modules.order    modules.symbols      System.map  vmlinuz
dtb     initrd         modules.alias  modules.builtin    modules.dep          modules.devname  modules.softdep  modules.symbols.bin  vdso
$ ls usr/lib/modules/5.4.0-7-generic/dtb/
actions  allwinner  altera  amd  amlogic  apm  arm  bitmain  broadcom  cavium  freescale  hisilicon  intel  lg  marvell  mediatek  qcom  realtek  renesas  sprd  synaptics  ti  xilinx
$ sudo ostree admin deploy test
Copying /etc changes: 22 modified, 1 removed, 89 added
error: Installing kernel: regfile copy: Is a directory
```
It shows `egfile copy: Is a directory` error.
According to @agners' [comment](https://github.com/ostreedev/ostree/pull/2001#issuecomment-604442430), I try to have only some device trees under `usr/lib/modules/5.4.0-7-generic/dtb` directly:
```
$ ls usr/lib/modules/5.4.0-7-generic/dtb/
bcm2711-rpi-4-b.dtb  bcm2837-rpi-3-b.dtb  bcm2837-rpi-3-b-plus.dtb  bcm2837-rpi-cm3-io3.dtb
```
Then, commit and deploy device tree again.
We have new `uEnvt.txt` with **fdt_path now**:
```
$ cat /boot/uEnv.txt 
kernel_image=/ostree/eos-54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6/vmlinuz-5.4.0-7-generic
ramdisk_image=/ostree/eos-54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6/initramfs-5.4.0-7-generic.img
fdt_path=/ostree/eos-54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6
bootargs=rw splash plymouth.ignore-serial-consoles quiet loglevel=0 ostree=/ostree/boot.0/eos/54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6/0
kernel_image2=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/vmlinuz-5.4.0-7-generic
ramdisk_image2=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/initramfs-5.4.0-7-generic.img
fdt_file2=/ostree/eos-f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/devicetree-5.4.0-7-generic
bootargs2=rw splash plymouth.ignore-serial-consoles quiet loglevel=0 ostree=/ostree/boot.0/eos/f0bc8343b896de2ebca20fc1eb4dfa123fef6ed8865b19dce4086bc6551e72db/0
```
Also check how are the device tree binaries deployed:
```
$ ls /boot/ostree/eos-54b04738f4585cd29442c0454da1cafc66e9ec003c0a343891d072f8cd1db5d6
bcm2711-rpi-4-b.dtb  bcm2837-rpi-3-b.dtb  bcm2837-rpi-3-b-plus.dtb  bcm2837-rpi-cm3-io3.dtb  initramfs-5.4.0-7-generic.img  vmlinuz-5.4.0-7-generic
```
Now we understand the designed idea:
1. A devicetree directory must locate at `usr/lib/modules/$kver/dtb/` (mentioned in commit message)
2. The device tree folder should have only device tree binaries, no recursive folders.
3. The device tree binaries in the device tree folder are not aggregated in a single subfolder under the boot ostree path. They are located at the same level of kernel and ramdisk files. (@jlebon raised this [question](https://github.com/ostreedev/ostree/pull/2001#discussion_r390588257) before)

starnight:
According to the final test result in previous [comment](https://github.com/ostreedev/ostree/pull/2001#issuecomment-605803184), I think that will be great if subfolders can work recursively in `dtb` folder.  Because, that is the way device tree binaries get installed by the official Linux kernel, and that structure is installed by Fedora, Ubuntu, etc, and also the RPi firmware gives us a path to use involving a subdirectory.

agners:
Sorry for the delay on this topic. I started to think about a different direction altogether.

Instead of deploying all of them, and deferring the decision which device tree to use to the boot loader, I think the OS (user) should be in charge here. So when OSTree knows which device tree to use, then we do not have to deploy multiple.

Basically, what I was thinking about:

```
ostree admin deploy --device-tree=thatdevicetree.dtb
```

This then would deploy only that device tree to /boot, and create a `devicetree` entry in the loader conf (and for U-Boot, we would set the `fdtfile` environment variable). This also fits the BLS, as there is no `devicetree_path`, but there is `devicetree` (see https://systemd.io/BOOT_LOADER_SPECIFICATION/). I would use the current Fedora storage location to search for device trees `/usr/lib/modules/$kver/dtb/`.

Furthermore, it would also extend to overlays nicely:

```
ostree admin deploy --device-tree=thatdevicetree.dtb --dt-overlay="overlay1.dtb overlay2.dtb
ostree admin deploy --device-tree=thatdevicetree.dtb --dt-overlay-append=overlay1.dtb
```

This would deploy the right overlays to `/boot` and set `devicetree-overlay` loader conf. So far I think Fedora does not support any overlays, so the source directory for this has not been defined really. I would use `/usr/lib/modules/$kver/dtb/overlays/`.

The downside of this approach is that a OS image will come with a default device tree deployed, hence it will only work for that particular machine. We could also consider a hybird of the two approaches: Deploy all of them if `--device-tree` is not set, and let the boot loader decide in this case, but if `--device-tree` is set we deploy only the one.

@starnight @jlebon @cgwalters thoughts?

akiernan:
There's two use cases come to mind here. Either you're deploying a general purpose OS, targeting many platforms, which you deploy everything for, or you're deploying a single purpose OS, targeting a single hardware platform.

I think it's the second world you're thinking about here, but even there, you don't always want to have to have just one DTB deployed - for a BBB you might want to apply overlays at boot time to support the hardware, rather than build every possible combination of device tree to support the available capes.

In the world I'm running in, we have two base hardware models (which have their own DTB) and then a small set of overlays that we apply based on various criteria to determine what we actually boot with. For example we have a tile on the board which has both traditional serial and SPI interfaces and depending on the exact use case, we'll pick one or the other at boot time.

Whilst I'm sure a single DTB deployment is useful, I do wonder if you should simply include that in your deployment ref, rather than making it a deployment time option.

agners:
@akiernan I think your first observation still applies, and is quite accurate. It really is a trade off, a question of who does which choices.

The model I proposed would allow to reuse the same OSTree, but might need different OSTree deployments (and hence images) for different devices. I see that this depends on the exact use case if that is an advantage over what we have today. But it is not a step back as with that OSTree would at least support multiple device tree in a single OSTree commit.

> though I do still wonder if you include what you want in your ref rather than doing it at deploy.
That is what I try to avoid, since I want to push that one update with support for all my devices (from that perspective the OS gets more general purpose).

My current case is that overlays are manually selected by the user. We control that by setting a U-Boot variable. However, that has the downside that this cannot be rolled back. Using a `ostree admin deploy` command would make this rollback capable...

I've a question to your use case:
> In the world I'm running in, we have two base hardware models (which have their own DTB) and then a small set of overlays that we apply based on various criteria to determine what we actually boot with.
Who decides which overlays should get applied? Is that the boot loader?

So in your case, my Initial proposal having a fdt_path which points to the location where all the device trees (and overlays) are stored would be better since you can let the boot loader pick the right device tree as well as the overlays?

I see arguments for both cases, and I think one option does not necessary exclude the other option.

akiernan:
>> In the world I'm running in, we have two base hardware models (which have their own DTB) and then a small set of overlays that we apply based on various criteria to determine what we actually boot with.

> Who decides which overlays should get applied? Is that the boot loader?

We ship `/usr/lib/ostree-boot/uEnv.txt` which overrides `findfdt` to set it, though our model is many images, where each image is specific to a use case. So whilst it's in the bootloader, it's picked based on the deployed image.

> So in your case, my Initial proposal having a fdt_path which points to the location where all the device trees (and overlays) are stored would be better since you can let the boot loader pick the right device tree as well as the overlays?

We ship both our base and overlay DTBs as part of a (signed) FIT image, so we're avoiding the need to resolve OSTree filesystem paths in the bootloader; if we weren't doing that, I think `fdt_path` would work. 

> I see arguments for both cases, and I think one option does not necessary exclude the other option.

Definitely true!

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1730,16 +1766,46 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+
+          while (TRUE)
+            {
+              struct dirent *dent;
+
+              if (!glnx_dirfd_iterator_next_dent (&dfditer, &dent, cancellable, error))
+                  return FALSE;
+              if (dent == NULL)
+                break;
+
```
agners, line 1797: There is a check missing for directories here, which leads to the following error if the source is a directory:

```
error: Installing kernel: linkat(overlays): Operation not permitted
```

Alternatively, and probably better, we go the recursive appraoch.

agners:
I continued my initial approach and just pushed a new version with the following notable changes:

- preserve the `dtb` directory in the target directory
- checksum and deploy sub directories found in `dtb`
- changed U-Boot and BLS variable to `fdtdir` (U-Boot uses this in `extlinux.conf` which follows largely BLS otherwise, see https://github.com/u-boot/u-boot/blob/master/doc/README.distro#L71)
- dropped trailing `/` in the `fdtdir` variable
- make sure files are check-summed in predictable order to prevent hash mismatches
- add unit test

agners:
> We ship both our base and overlay DTBs as part of a (signed) FIT image, so we're avoiding the need to resolve OSTree filesystem paths in the bootloader; if we weren't doing that, I think fdt_path would work.

@akiernan how do you ship the FIT image? As vmlinuz in the OSTree? And initramfs is in that package too, I assume? I guess that works since initramfs is optional...

akiernan:
>> We ship both our base and overlay DTBs as part of a (signed) FIT image, so we're avoiding the need to resolve OSTree filesystem paths in the bootloader; if we weren't doing that, I think fdt_path would work.

> @akiernan how do you ship the FIT image? As vmlinuz in the OSTree? And initramfs is in that package too, I assume? I guess that works since initramfs is optional...

As vmlinuz in OSTree. We don't need initramfs as our target is known, so we boot with `ostree-prepare-root` as `init`.

agners:
Anybody came around to test this updated version? I think it should address all comments. If there are still question/concerns, let me know.

akiernan:
Only thing I've spotted so far is the commit message needs a `s/fdt_patch/fdtdir/`

akiernan:
I've just realised `fdtpath` isn't done per-deployment though in `uEnv.txt`, shouldn't I have `fdtpath2` for the old deployment?

agners:
@akiernan yeah I was once wondering too why it did not behave as expected, when I suddenly realized that I was using an unpatched binary still.

Replaced `fdt_path` with `fdtdir` in the commit message and rebased the code.

agners:
@jlebon, @cgwalters I think this is ready to be merged.

cgwalters:
Honestly I don't know much about DT, and can't quite spare the cycles to dive in myself right now.

Maybe worth asking on the mailing list for any other reviewers?

I skimmed the code and it looks sane.  If the community feels the code is ready, then I'm OK to merge.

agners:
It seems that @akiernan is fine with the latest version.

@starnight, can you maybe have a look at the new version?

akiernan:
> It seems that @akiernan is fine with the latest version.
> 

Yep, functionally it seems fine to me.

nullr0ute:
I can take a look at some point if review is needed from a DT PoV. For Fedora IoT up to now we're requiring EBBR which means the firmware provides the DT which means it doesn't need to be located on the filesystem.

agners:
@nullr0ute thanks! Yeah, unfortunately, EBBR/firmware provided dt is still not the reality for lots of devices. For those cases, shipping multiple device trees to avoid having to build multiple OS images is the best thing we can do.

starnight:
Thanks to @agners' new implementation.  I tested branch `multi-device-tree` at commit 5f08649 ("deploy: support devicetree directory").  Here is the test I have done and result:
* Remove the `devicetree`
* Add `broadcom` and `amlogic` dtbs to `usr/lib/modules/5.6.0-7-generic/dtb` for multiple dtb folders test
```
$ sudo ostree admin status
* eos 1432c89ac02b1186ecbc0c3b6f1af2677aa09d4812cb97c4bd0fa46721342227.0
    Version: 3.9.0~200621-230039
    origin refspec: eos:os/eos/arm64/master
    GPG: Signature made Mon 22 Jun 2020 07:56:54 AM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Sun 01 Jul 2029 12:18:11 AM CST
    GPG: Signature made Mon 22 Jun 2020 07:56:57 AM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Wed 10 Jun 2026 01:47:38 AM CST
$ sudo ostree checkout 1432c89ac02b1186ecbc0c3b6f1af2677aa09d4812cb97c4bd0fa46721342227
$ cd 1432c89ac02b1186ecbc0c3b6f1af2677aa09d4812cb97c4bd0fa46721342227
$ sudo rm lib/modules/5.6.0-7-generic/devicetree
$ sudo mkdir usr/lib/modules/5.6.0-7-generic/dtb
$ sudo cp -r usr/lib/firmware/5.6.0-7-generic/device-tree/{broadcom,amlogic} usr/lib/modules/5.6.0-7-generic/dtb/
$ sudo ostree commit -b test --tree=dir=`pwd`
813b69a3a4aaa446816e6f8dca08c8552042c63e8ba8e0a1e9cc4032eaf4fb89
$ sudo ostree admin deploy test
Copying /etc changes: 23 modified, 1 removed, 97 added
Transaction complete; bootconfig swap: yes; deployment count change: 1
Freed objects: 4.0 MB
$ sudo ostree admin status
  eos 813b69a3a4aaa446816e6f8dca08c8552042c63e8ba8e0a1e9cc4032eaf4fb89.0 (pending)
    origin refspec: test
* eos 1432c89ac02b1186ecbc0c3b6f1af2677aa09d4812cb97c4bd0fa46721342227.0
    Version: 3.9.0~200621-230039
    origin refspec: eos:os/eos/arm64/master
    GPG: Signature made Mon 22 Jun 2020 07:56:54 AM CST using RSA key ID 9E08D8DABA02FC46
    GPG: Good signature from "EOS OSTree Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Sun 01 Jul 2029 12:18:11 AM CST
    GPG: Signature made Mon 22 Jun 2020 07:56:57 AM CST using RSA key ID FCF17B17F1F8E157
    GPG: Good signature from "EOS Flatpak Signing Key 1 <maintainers@endlessm.com>"
    GPG: Key expires Wed 10 Jun 2026 01:47:38 AM CST
$ cat /boot/uEnv.txt 
kernel_image=/ostree/eos-8a4f394fb3b8d81f37e1fdb62376335955fc807dc021c9af6c0d59fa96a21dae/vmlinuz-5.6.0-7-generic
ramdisk_image=/ostree/eos-8a4f394fb3b8d81f37e1fdb62376335955fc807dc021c9af6c0d59fa96a21dae/initramfs-5.6.0-7-generic.img
fdtdir=/ostree/eos-8a4f394fb3b8d81f37e1fdb62376335955fc807dc021c9af6c0d59fa96a21dae/dtb
bootargs=rw splash plymouth.ignore-serial-consoles quiet loglevel=0 ostree=/ostree/boot.0/eos/8a4f394fb3b8d81f37e1fdb62376335955fc807dc021c9af6c0d59fa96a21dae/0
kernel_image2=/ostree/eos-79e3923eeb8cdee5043c24fe571d224a794209622776f08a9a58bf08c7496ebb/vmlinuz-5.6.0-7-generic
ramdisk_image2=/ostree/eos-79e3923eeb8cdee5043c24fe571d224a794209622776f08a9a58bf08c7496ebb/initramfs-5.6.0-7-generic.img
fdt_file2=/ostree/eos-79e3923eeb8cdee5043c24fe571d224a794209622776f08a9a58bf08c7496ebb/devicetree-5.6.0-7-generic
bootargs2=rw splash plymouth.ignore-serial-consoles quiet loglevel=0 ostree=/ostree/boot.0/eos/79e3923eeb8cdee5043c24fe571d224a794209622776f08a9a58bf08c7496ebb/0
$ ls /boot/ostree/eos-8a4f394fb3b8d81f37e1fdb62376335955fc807dc021c9af6c0d59fa96a21dae/
dtb  initramfs-5.6.0-7-generic.img  vmlinuz-5.6.0-7-generic
$ ls /boot/ostree/eos-8a4f394fb3b8d81f37e1fdb62376335955fc807dc021c9af6c0d59fa96a21dae/dtb/*
/boot/ostree/eos-8a4f394fb3b8d81f37e1fdb62376335955fc807dc021c9af6c0d59fa96a21dae/dtb/amlogic:
meson-a1-ad401.dtb                meson-gxbb-kii-pro.dtb         meson-gxbb-wetek-hub.dtb          meson-gxl-s905w-tx3-mini.dtb        meson-gxm-q201.dtb
meson-axg-s400.dtb                meson-gxbb-nanopi-k2.dtb       meson-gxbb-wetek-play2.dtb        meson-gxl-s905x-hwacom-amazetv.dtb  meson-gxm-rbox-pro.dtb
meson-g12a-sei510.dtb             meson-gxbb-nexbox-a95x.dtb     meson-gxl-s805x-libretech-ac.dtb  meson-gxl-s905x-khadas-vim.dtb      meson-gxm-s912-libretech-pc.dtb
meson-g12a-u200.dtb               meson-gxbb-odroidc2.dtb        meson-gxl-s805x-p241.dtb          meson-gxl-s905x-libretech-cc.dtb    meson-gxm-vega-s96.dtb
meson-g12a-x96-max.dtb            meson-gxbb-p200.dtb            meson-gxl-s905d-libretech-pc.dtb  meson-gxl-s905x-nexbox-a95x.dtb     meson-sm1-khadas-vim3l.dtb
meson-g12b-a311d-khadas-vim3.dtb  meson-gxbb-p201.dtb            meson-gxl-s905d-p230.dtb          meson-gxl-s905x-p212.dtb            meson-sm1-sei610.dtb
meson-g12b-odroid-n2.dtb          meson-gxbb-vega-s95-meta.dtb   meson-gxl-s905d-p231.dtb          meson-gxm-khadas-vim2.dtb
meson-g12b-s922x-khadas-vim3.dtb  meson-gxbb-vega-s95-pro.dtb    meson-gxl-s905d-phicomm-n1.dtb    meson-gxm-nexbox-a1.dtb
meson-g12b-ugoos-am6.dtb          meson-gxbb-vega-s95-telos.dtb  meson-gxl-s905w-p281.dtb          meson-gxm-q200.dtb

/boot/ostree/eos-8a4f394fb3b8d81f37e1fdb62376335955fc807dc021c9af6c0d59fa96a21dae/dtb/broadcom:
bcm2711-rpi-4-b.dtb  bcm2837-rpi-3-a-plus.dtb  bcm2837-rpi-3-b.dtb  bcm2837-rpi-3-b-plus.dtb  bcm2837-rpi-cm3-io3.dtb  northstar2  stingray
```
The DTBs are deployed by following the Linux installation.  It looks good!

cgwalters:
Possible fallout in https://github.com/ostreedev/ostree/issues/2154

PR #2003, opened 2020-02-10
docs: Fix 'package layering' rpm-ostree link

crobinso:
The referenced fragment name has changed in the published rpm-ostree docs.
Fix it.

Also, it seems ostree's readthedocs content is out of date, b35f337dc7c
is from May 2019 but the published content doesn't contain that fix


cgwalters:
/approve
/lgtm
Yeah the docs build isn't automated at the moment :cry:

PR #2009, opened 2020-02-14
deploy: Avoid trying to change immutable state unnecessarily

cgwalters:
For some reason I haven't fully debugged (probably a recent
kernel change), in the case where the immutable bit isn't set,
trying to call `EXT2_IOC_SETFLAGS` without it set returns `EINVAL`.

Let's avoid calling the `ioctl()` if we don't have anything to do.

This fixes a slew of `make check` failures here in my toolbox
environment.

(kernel is `5.5.0-0.rc6.git0.1.fc32.x86_64` with `xfs`)

### DIFF
--- a/src/libostree/ostree-linuxfsutil.c
+++ b/src/libostree/ostree-linuxfsutil.c
@@ -68,6 +68,10 @@ _ostree_linuxfs_fd_alter_immutable_flag (int            fd,
     }
   else
     {
+      gboolean prev_immutable_state = (flags & EXT2_IMMUTABLE_FL) > 0;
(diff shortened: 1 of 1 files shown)


cgwalters:
I investigated this a bit more and it does seem to have something to do with containerization (thankfully, otherwise things would blow up if there was a kernel change here that affected ostree outside of container).

OK and looking at the code... this might have been introduced by https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f991492ed11055934f1b35615cb1b435325939bf

PR #2012, opened 2020-02-18
Update ostree-pull.xml with info about pulled refs location and access

clime:
Adds information into description section of `ostree pull` command about pulled refs - where they are stored and how they can be accessed in OSTree commands.

### DIFF
--- a/man/ostree-pull.xml
+++ b/man/ostree-pull.xml
@@ -143,6 +143,14 @@ Boston, MA 02111-1307, USA.
     <refsect1>
         <title>Description</title>
 
+      <para>
+	  Without --mirror, this command will create new refs
+	  under <literal>remotes/REMOTE/</literal> directory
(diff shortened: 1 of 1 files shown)


cgwalters:
/lgtm cancel
Looks like the build is failing with an XML syntax error; check the Travis logs

clime:
It didn't like things like `<REMOTE>` or `<BRANCH>` so I removed the those angle brackets. I hope it will be clear from the context that `REMOTE` and `BRANCH` are variables.

clime:
By the way should i squash the commits or is it ok like this (in case the checks succeed)?

clime:
ping

cgwalters:
> By the way should i squash the commits or is it ok like this (in case the checks succeed)?

Yes please squash - we prefer a cleaner git history.

clime:
squashed.

PR #2015, opened 2020-02-21
Release 2020.2

cgwalters:
"Brown paper bag" release that actually sets the
`is_release_build=yes` flag and also fixes the
`Since:` on a few new functions.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -7,10 +7,10 @@ dnl Seed the release notes with `git-shortlog-with-prs <previous-release>..`. Th
 dnl `git-evtag` to create the tag and push it. Finally, create a GitHub release and attach
 dnl the tarball from `make dist`.
 m4_define([year_version], [2020])
-m4_define([release_version], [2])
+m4_define([release_version], [3])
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -2,12 +2,12 @@ AC_PREREQ([2.63])
...
+dnl is_release_build=yes
 dnl Seed the release notes with `git-shortlog-with-prs <previous-release>..`. Then use
 dnl `git-evtag` to create the tag and push it. Finally, create a GitHub release and attach
 dnl the tarball from `make dist`.
 m4_define([year_version], [2020])
-m4_define([release_version], [2])
+m4_define([release_version], [3])
 m4_define([package_version], [year_version.release_version])
 AC_INIT([libostree], [package_version], [walters@verbum.org])
 is_release_build=no
```
akiernan: ITYM to flip this one...

akiernan: Ignore me... I'm not reading the right commit :|

cgwalters: No worries, it is confusing!  Thanks for double checking.

PR #2018, opened 2020-03-02
ci: migrate to new coreos-ci project

jlebon:
Use the new custom steps. I think we could simplify things further by
using `fcosBuild` but let's start with this for now.


jlebon:
Requires: https://github.com/coreos/coreos-ci/pull/9

jlebon:
OK, this is ready to go!

PR #2025, opened 2020-03-11
tests: Rework tests/installed → tests/kola

cgwalters:
Previously we made an effort to use the [Fedora Standard Test Interface](https://docs.fedoraproject.org/en-US/ci/standard-test-interface/).
This effort was not very successful; the primary thing that
it really died on is Ansible just didn't support rebooting
very well.  I think that's since gotten better, but even
then, Ansible wasn't the best thing for a test framework
for us anyways.

In the meantime Fedora CoreOS happened emphasizing Ignition
and not "post-hoc reconciliation" models like Ansible over
ssh.

And, [coreos-assembler](https://github.com/coreos/coreos-assembler) happened too.

Furthermore, we really need to test OSTree's interaction
with Ignition as we've invented several special things there.

Then most recently, I've been working on having
cosa/kola support running externally defined tests:
https://github.com/coreos/coreos-assembler/pull/1215

There's a lot of things to clean up after this but at least this
works for me:

```
$ cd /srv/fcos
$ cosa kola run -- --parallel 4 --output-dir tmp/kola -E ~/src/github/ostreedev/ostree/ 'ext.ostree.*'
```

NOTE: This *does not* drop ostree binaries into the target.  See:
https://github.com/coreos/coreos-assembler/pull/1252#issuecomment-600623315

This drops our dependency on Python in the installed tests, and
also fixes a few bugs that came up.

I disabled the `itest-bare-user-root.sh` one because it's
entangled with the shell script infrastructure for the unit tests.


cgwalters:
OK, a major update here, see the commit message for details.  All tests passing here.  Out with the Ansible, in with the kola!

cgwalters:
Updated to `s/deps/data`.  I guess we could wait a bit till the cosa container is rebuilt and try running this in our CI...but OTOH I wouldn't want to block if that fails necessarily because even having this in master is super useful for me because at least anyone running locally can run tests more easily.

cgwalters:
Interesting, so that Travis failure is:

```
ERROR: tests/test-concurrency.py
================================
n_committers 4 n_pruners 0
Creating tree tree0
Creating tree tree1
PID 30315 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 30316 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 30317 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 30318 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 30316 exited OK
PID 30318 exited OK
PID 30317 exited OK
PID 30315 exited OK
n_committers 6 n_pruners 3
Creating tree tree0
Creating tree tree1
Creating tree tree2
PID 30321 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 30322 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 30323 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 30324 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 30325 ostree --repo=repo commit --fsync=0 -b tree2 --tree=dir=tree2
PID 30326 ostree --repo=repo commit --fsync=0 -b tree2 --tree=dir=tree2
error: opendir(staging-89ed2f5d-09e9-4dc1-8360-9d7011bc2edd-anBXJW): No such file or directory
PID 30327: ostree --repo=repo prune --refs-only
PID 30328: ostree --repo=repo prune --refs-only
PID 30329: ostree --repo=repo prune --refs-only
process 30322 exited with code 1
PID 30326 exited OK
PID 30324 exited OK
PID 30321 exited OK
PID 30323 exited OK
PID 30325 exited OK
PID 30328 exited OK
PID 30327 exited OK
PID 30329 exited OK
A child process exited abnormally
1..2
014ccd06e67eacefc6a965c01a3fe3aa953cbac93cd17ab49b7cc35325b9e8b9
67eddf38725595bf7a23c692101b6edded083c8fbc5b099ce7be5afa6f56f90c
00867bcf06c15f34b3b1ce4d62eeb13747ec470c1b52bf74f5834475fa8875d9
0fbf3241a95691c836497e5c1ae0422d67fe9e71ba8fade043c64c58970a5b0d
ok no concurrent prunes
PASS: tests/test-concurrency.py 1 no concurrent prunes
d62b8c33421f8dadd6844646cb46e7f7bf8e9dfe79db01c6829ac282fa8d3a3a
83798fcef5f1893b7848103e039021de22eaeb8615cb05e9bfa3b4b16352605a
10461655e44945cd0c923c6451de43c5cf0ad0ef11f62f5c24aabb7e9a6a1485
7a566b367f6a250f2e1433e42ad4f84120b2d6ce87856f9f3f7bddacc690689f
Total objects: 72
No unreachable objects
fd51a8b23108e97ec7d57dea2a13b9ada796e0998b1acd364b94e900aaf4d80b
Total objects: 73
No unreachable objects
Total objects: 73
No unreachable objects
ERROR: tests/test-concurrency.py - too few tests run (expected 2, got 1)
ERROR: tests/test-concurrency.py - exited with status 1
```

Worth investigating at some point.

tests/kola/destructive/var-mount.sh
```diff
@@ -0,0 +1,23 @@
...
+# https://github.com/ostreedev/ostree/issues/1667
+set -xeuo pipefail
+
+. ${KOLA_EXT_DATA}/libinsttest.sh
+
+n=$(nth_boot)
+case "${n}" in
+  1) 
+    require_writable_sysroot
+    # Hack this off for now
```
jlebon: Interesting to think about possibly using an Ignition config for this now as a test.

tests/kola/libinsttest.sh
```diff
@@ -0,0 +1,91 @@
...
+    fi
+}
+
+nth_boot() {
+    journalctl --list-boots | wc -l
+}
+
+kola_reboot() {
+    kill -TERM $$
+    sleep 1h
```
jlebon: Minor: hmm, 1h seems like overkill... could probably use 5m or something? If a mostly vanilla FCOS node takes 59m to reboot, then that raises some flags in either the overlayed libostree or the kola test harness or somewhere in FCOS.

For comparison, the upgrade tests in kola use 2m (we could probably bump that up a bit to be safe, but we haven't had a need yet).

cgwalters: I *think* actually this can never happen because bash will `waitpid()` for `kill` to exit but the signal to the parent will have already been delivered synchronously.  So OK, changed to `2m`.

tests/kola/destructive/var-mount.sh
cgwalters: Oh yes, more tests to be written and reworked now...just trying to port what exists.  This particular test though is more trying to test the "anaconda path" and not all the special stuff that happens w/ignition+ostree.

PR #2028, opened 2020-03-13
ci: Test kola --upgrades

cgwalters:
This one is very relevant for ostree.


jlebon:
/lgtm

See also https://github.com/coreos/coreos-ci-lib/pull/14.

PR #2029, opened 2020-03-13
ci: use `fcosKola` for running kola tests

jlebon:
We were using `--no-test-exit-error` for upgrade tests but weren't
actually checking for test failures after.

Instead of running kola directly, just use the `fcosKola` custom step
which automatically takes care of e.g. running tests in parallel and
archiving results.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -77,20 +77,7 @@ parallel fcos: {
         coreos-assembler build
       """)
     }
-    stage("FCOS basic") {
-      timeout(time: 30, unit: 'MINUTES') {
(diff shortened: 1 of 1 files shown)


cgwalters:
Whee now this is blocked too...

cgwalters:
A lot of lessons learned from this...among them that I clearly should have at least done a `kola run` with the original ro-sysroot stuff and ensured that that passed.

Now that we have this CI set up, plus the [run-ext-bin](https://github.com/ostreedev/ostree/pull/2025) work will get us in a better place.  It's tempting to just force merge things here and push through rather than trying to hack it...

PR #2037, opened 2020-03-19
lib/fetcher-util: retry download on G_IO_ERROR_PARTIAL_INPUT

mwleeds:
Add G_IO_ERROR_PARTIAL_INPUT to the list of error codes caused by
transient networking errors which lead us to retry the request. When
attempting to install the spotify flatpak you often get the error
message "Connection terminated unexpectedly" and the download of the deb
file fails. In this case, libsoup is setting G_IO_ERROR_PARTIAL_INPUT
and sometimes a subsequent download attempt is successful, so we should
treat it as transient.

Ideally we would behave as wget does in this case and retry the download
picking up where we left off in the file rather than starting over, but
that would require changes to libsoup I think.

Sadly this patch does not fix the flatpak installation of spotify in the
face of such errors, because flatpak doesn't use libostree to download
extra data, but presumably it's possible we could encounter such an
error pulling from an ostree repo, so the patch is still correct.

### DIFF
--- a/src/libostree/ostree-fetcher-util.c
+++ b/src/libostree/ostree-fetcher-util.c
@@ -202,6 +202,7 @@ _ostree_fetcher_should_retry_request (const GError *error,
   if (g_error_matches (error, G_IO_ERROR, G_IO_ERROR_TIMED_OUT) ||
       g_error_matches (error, G_IO_ERROR, G_IO_ERROR_HOST_NOT_FOUND) ||
       g_error_matches (error, G_IO_ERROR, G_IO_ERROR_HOST_UNREACHABLE) ||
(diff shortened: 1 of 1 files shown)


cgwalters:
/approve
/lgtm

At one point I tried to have ostree support partial fetches, but it significantly complicates the state machine.

mwleeds:
> ERROR: tests/test-concurrency.py - too few tests run (expected 2, got 1)

Do we know if that one is flaky?

jlebon:
It's likely not a flake, but it's not new either.

jlebon:
I opened https://github.com/ostreedev/ostree/issues/2038 to track that one.

PR #2039, opened 2020-03-21
main/commit: Rework control flow to use --tree=X path

cgwalters:
Rework the simple cases of "commit ." and "commit argv[1]" to
generate the more general "--tree=X --tree=Y" path, so that we
only have one primary control flow here.

Prep for a future patch around loading SELinux policy from
the first argument.


jlebon:
OK yup. This is much easier to review with whitespace changes ignored.
/lgtm

cgwalters:
/hold
needs CI fixes from https://github.com/ostreedev/ostree/pull/2040

cgwalters:
test-pull-repeated flake

PR #2041, opened 2020-03-23
 repo/commit: Add support for --selinux-policy-from-base

cgwalters:
The [dev-overlay](https://github.com/coreos/coreos-assembler/blob/332c6ab3b91778d904224c3c960d9cc4739d60bd/src/cmd-dev-overlay)
script shipped in coreos-assembler mostly exists to deal
with the nontrivial logic around SELinux policy.  Let's make
the use case of "commit some binaries overlaying a base tree, using
the base's selinux policy" just require a magical
`--selinux-policy-from-base` argument to `ostree commit`.

A new C API was added to implement this in the case of `--tree=ref`;
when the base directory is already checked out, we can just reuse
the existing logic that `--selinux-policy` was using.

Requires: https://github.com/ostreedev/ostree/pull/2040
Requires: #2039


cgwalters:
test-pull-repeated flake

src/libostree/libostree-devel.sym
```diff
@@ -20,7 +20,7 @@
 /* Add new symbols here.  Release commits should copy this section into -released.sym. */
 LIBOSTREE_2020.2 {
 global:
-  someostree_symbol_deleteme;
+  ostree_repo_commit_modifier_sepolicy_from;
```
jlebon: Bikeshed: how about `ostree_repo_commit_modifier_set_sepolicy_from` to match the existing `ostree_repo_commit_modifier_set_sepolicy`?

Hmm, and maybe also we just spell it out entirely it's from what, i.e. `from_commit` ?

src/ostree/ot-builtin-commit.c
```diff
@@ -637,12 +633,33 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
       g_clear_object (&object_to_commit);
       if (strcmp (tree_type, "dir") == 0)
         {
+          if (first && opt_selinux_policy_from_base)
+            {
+              opt_selinux_policy = g_strdup (tree);
+              opt_selinux_policy_from_base = FALSE;
+            }
```
jlebon: Heh, clever.

src/libostree/ostree-repo-commit.c
```diff
@@ -4279,6 +4281,48 @@ ostree_repo_commit_modifier_set_sepolicy (OstreeRepoCommitModifier
...
+{
+  GLNX_AUTO_PREFIX_ERROR ("setting sepolicy from commit", error);
+  g_autofree char *commit = NULL;
+  g_autoptr(GFile) root = NULL;
+  if (!ostree_repo_read_commit (repo, rev, &root, &commit, cancellable, error))
+    return FALSE;
+  const char policypath[] = "usr/etc/selinux";
+  g_autoptr(GFile) policyroot = g_file_get_child (root, policypath);
+  if (!g_file_query_exists (policyroot, NULL))
+    return TRUE;  /* No policy, nothing to do */
```
jlebon: Hmm, I'm wondering if this should be a hard error instead. It feels like if you call this API, you expect an actual `OstreeSePolicy` to get set on the modifier by the end. For example, making it a `G_IO_ERROR_NOT_FOUND` allows callers to handle it and specify a fallback policy via `set_sepolicy`.

Could also make it a e.g. `allow_noent` parameter.

src/libostree/libostree-devel.sym
cgwalters: Pretty long identifier but...eh, sure.

src/libostree/ostree-repo-commit.c
cgwalters: I don't want all build scripts to bifurcate between "support selinux" versus not.  This way e.g. the [dev-overlay script](https://github.com/coreos/coreos-assembler/blob/332c6ab3b91778d904224c3c960d9cc4739d60bd/src/cmd-dev-overlay) wouldn't need to check whether a policy exists...which is what this is doing anyways.

This way we basically just Do The Right Thing - the only reason this isn't on by default is a conservatism around behavioral changes.

Now yes we'd end up with silently unlabeled files if for some reason the policy isn't there - but if the policy isn't there *other* things are going to break too...and unlabeled files will quickly be noticed if the system is enforcing, etc.

PR #2051, opened 2020-04-01
lib: fix typo in function docs

fkrull:
Kind of a mild typo, but it caused the Rust bindings generator to produce broken code because the "Since" tag didn't work.


cgwalters:
Hi @fkrull - thanks for the patch!  Did you see [this PR](https://github.com/ostreedev/ostree/pull/2048) btw?
/approve
/ok-to-test
/lgtm

Issue #2053, opened 2020-04-03
Delta pulls broken when repo has commitpartial


alexlarsson:
If we're pulling a commit using static deltas, and for some reason we have a commitpartial for the pulled commit, then when we save the commit from the static delta superblock and scan it we will trigger the scanning of the root dirtree and dirmeta in scan_commit_object() due to the is_partial check. 

This is not going to *break* anything, but it will mean we will download everything twice, both as delta parts and separate objects.

I think we need to disable the dirmeta/dirtree scanning if the commit came from the delta.

alexlarsson:
For various reasons this was triggered in flatpak causing https://github.com/flatpak/flatpak/pull/3524

jlebon:
Threw something up in https://github.com/ostreedev/ostree/pull/2054.

PR #2054, opened 2020-04-03
lib/pull: Don't scan commit objects we fetch via deltas

jlebon:
When we're fetching a commit via static delta, we already take care of
fetching the full commit, so there's no need to also scan it using the
regular object workflow.

Closes: #2053

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -47,6 +47,7 @@ cosaPod(runAsUser: 0, memory: "9Gi", cpu: "4") {
     checkout scm
     unstash 'build'
     shwrap("""
+      chown -R -h builder: .
       # Move the bits into the cosa pod (but only if major versions match)
       buildroot_id=\$(cat installed/buildroot-id)
       osver=\$(. /usr/lib/os-release && echo \$VERSION_ID)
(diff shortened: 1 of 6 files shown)


jlebon:
Not tested/needs tests.

cgwalters:
Ouch...pretty embarrassing bug.  This looks good to me; we could probably test coverage of this by temporarily doing a `chmod a-r repo-srv/objects` or so in a test that does a static delta pull.

alexlarsson:
I'm not sure about this one. It completely skips over scan_commit_object() in the delta case. But will that not break the fetching of parent commits (if you use --depth=...)? Unless I misunderstand you need to skip only the scanning of root dirtree/metadata.

alexlarsson:
Alternatively we could scan the parent when handling the commit in the superblock.

cgwalters:
> But will that not break the fetching of parent commits (if you use --depth=...)? Unless I misunderstand you need to skip only the scanning of root dirtree/metadata.

Yep, I think you're right.

> Alternatively we could scan the parent when handling the commit in the superblock.

It's a bit tricky right now because it's actually the "write object" paths which verify the checksum, and we want to do that before we parse it.  Wouldn't be hard of course to just verify the commit checksum explicitly in the delta code before parsing it, but it'd be another special case.

I think I like your other idea of only skipping the scan for the metadata/content when fetched from delta, would just require passing another bit down into `ScanObjectQueueData`.

jlebon:
Revived this now! I tried the "pass through another parameter" approach but the number of function signatures impacted quickly grew because we also have to have it survive the detached metadata fetching first since we don't scan commits until they're verified. I ended up just adding another hash table to `OtPullData`. It's less efficient but more legible.

cgwalters (approved):
Nice!

PR #2058, opened 2020-04-05
pull: Update key loading function to match error style

cgwalters:
This code wasn't written with idiomatic GError usage; it's not standard
to construct an error up front and continually append to its
message.  The exit from a function is usually `return TRUE`,
with error conditions before that.

Updating it to match style reveals what I think is a bug;
we were silently ignoring failure to parse key files.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1488,14 +1488,10 @@ _load_public_keys (OstreeSign *sign,
                    const gchar *remote_name,
                    GError **error)
 {
-
   g_autofree gchar *pk_ascii = NULL;
   g_autofree gchar *pk_file = NULL;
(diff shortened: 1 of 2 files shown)


cgwalters:
> @d4s: changing LGTM is restricted to collaborators

Hm, right.  Are you willing to sign up to do some ongoing code review in general?  I am happy to invite you to the organization.

d4s:
> > @d4s: changing LGTM is restricted to collaborators
> Hm, right. Are you willing to sign up to do some ongoing code review in general? I am happy to invite you to the organization.

Thank you! I'll be happy to became a part of the team.
Especially keeping in mind that there are a lot of follow up work after integration of #1878 ;)

cgwalters:
@d4s want to retry a `/lgtm` here?

PR #2059, opened 2020-04-06
commit: Add --base argument

cgwalters:
I was trying to followup the `--selinux-policy-from-base` work
to add a `cosa build --fast=overlay` for coreos-assembler,
but hit on the fact that using e.g. `--owner-uid` disables
commit optimizations.

A while ago, https://github.com/ostreedev/ostree/pull/1643 landed
which optimized this for the case where no modifications are provided.
But, we really need the SELinux policy bits, and it's super convenient
to run `ostree commit` as non-root.

It's fairly surprising actually that it's taken us so long to
iterate on a good interface for this "commit changes on top of a base"
model.  In practice, many nontrivial cases really end up needing
to do a (hardlink) checkout, and that case is optimized.

But for this coreos-assembler work I want to directly overlay onto
a commit object another commit object.

That previous PR above added exactly the API we need, so let's
expose it in the CLI.

What you can see happening in the test is that we provide
`--owner-uid 42`, but that only applies to directories/files
that were added in the commit.

And now that I look at this, I think what we really want here
is to avoid changing directories that exist in the base, but
eh; in practice the main use here is for `--owner-uid 0` while
committing as non-root; and that works fine with this since
the baseline uid will be zero as well.


src/ostree/ot-builtin-commit.c
```diff
@@ -600,7 +602,32 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
...
+      g_autoptr(GFile) root = NULL;
+      if (!ostree_repo_read_commit (repo, opt_base, &root, &base_commit, cancellable, error))
+        goto out;
+      OstreeRepoFile *rootf = (OstreeRepoFile*) root;
+
+      mtree = ostree_mutable_tree_new_from_checksum (repo,
+                                                     ostree_repo_file_tree_get_contents_checksum (rootf),
+                                                     ostree_repo_file_tree_get_metadata_checksum (rootf));
+
+      if (opt_selinux_policy_from_base)
```
jlebon, line 617: Let's also reflect that in the help string for `selinux-policy-from-base`. E.g. something like:

> Set SELinux labels based on --base or otherwise the first --tree argument

?

tests/archive-test.sh
```diff
@@ -68,3 +68,11 @@ echo "ok cat-file"
 cd ${test_tmpdir}
 $OSTREE fsck
 echo "ok fsck"
+
+mkdir -p test-overlays
+date > test-overlays/overlaid-file
+$OSTREE commit ${COMMIT_ARGS} -b test-base --base test2 --owner-uid 42 --owner-gid 42 test-overlays/
+$OSTREE ls -R test-base > ls.txt
+cat ls.txt
+assert_streq "$(grep '42.*42' ls.txt | wc -l)" 2
```
jlebon, line 77: Could also check here that `ls.txt` has more than just 2 lines to make sure we did overlay on top of a base.

PR #2063, opened 2020-04-12
pull: Cleanup signature verification functions

cgwalters:
Rework this to use standard error handling - we fail fast if
we can't load the public keys for example.  Preserve
the semantics of accepting the first valid signature.  The
first signature error is captured, the others are currently
compressed into a `(and %d more)` prefix.

And now that I look at this more closely there's a lot of
duplication between the two code paths in pull.c for verifying;
will dedup this next.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1470,11 +1470,23 @@ process_gpg_verify_result (OtPullData            *pull_data,
 }
 #endif /* OSTREE_DISABLE_GPGME */
 
+static gboolean
+get_signapi_remote_option (OstreeRepo *repo,
+                           OstreeSign *sign,
+                           const char *remote_name,
+                           const char *keysuffix,
+                           char      **out_value,
+                           GError    **error)
+{
+  g_autofree char *key = g_strdup_printf ("verification-%s-%s", ostree_sign_get_name (sign), keysuffix);
+  return ostree_repo_get_remote_option (repo, remote_name, key, NULL, out_value, error);
+}
+
 /* _load_public_keys:
  *
  * Load public keys according remote's configuration:
- * inlined key passed via config option `verification-key` or
- * file name with public keys via `verification-file` option.
+ * inlined key passed via config option `verification-<signapi>-key` or
+ * file name with public keys via `verification-<signapi>-file` option.
  *
  * If both options are set then load all all public keys
  * both from file and inlined in config.
@@ -1493,15 +1505,10 @@ _load_public_keys (OstreeSign *sign,
   gboolean loaded_from_file = TRUE;
   gboolean loaded_inlined = TRUE;
 
-  ostree_repo_get_remote_option (repo,
-                                 remote_name,
-                                 "verification-file", NULL,
-                                 &pk_file, NULL);
-
-  ostree_repo_get_remote_option (repo,
-                                 remote_name,
-                                 "verification-key", NULL,
-                                 &pk_ascii, NULL);
+  if (!get_signapi_remote_option (repo, sign, remote_name, "file", &pk_file, error))
+    return FALSE;
+  if (!get_signapi_remote_option (repo, sign, remote_name, "key", &pk_ascii, error))
+    return FALSE;
 
   /* return TRUE if there is no configuration for remote */
   if ((pk_file == NULL) &&(pk_ascii == NULL))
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -1598,24 +1597,33 @@ _ostree_repo_sign_verify (OstreeRepo *repo,
         continue;
 
       /* Try to load public key(s) according remote's configuration */
-      if (_load_public_keys (sign, repo, remote_name, &local_error))
+      if (!_load_public_keys (sign, repo, remote_name, error))
+        return FALSE;
```
d4s, line 1601: The same idea with multiple signature types here.

src/libostree/ostree-repo-pull.c
```diff
@@ -2001,39 +2009,46 @@ scan_commit_object (OtPullData                 *pull_data,
...
       g_auto (GStrv) names = ostree_sign_list_names();
-      for (char **iter=names; !ret && iter && *iter; iter++)
+      for (char **iter=names; iter && *iter; iter++)
         {
           g_autoptr (OstreeSign) sign = NULL;
-          g_autoptr (GError) local_error = NULL;
 
-          if ((sign = ostree_sign_get_by_name (*iter, &local_error)) == NULL)
-            continue;
+          if ((sign = ostree_sign_get_by_name (*iter, error)) == NULL)
```
d4s, line 2022: and here

src/libostree/ostree-repo-pull.c
```diff
@@ -1571,20 +1571,19 @@ _ostree_repo_sign_verify (OstreeRepo *repo,
...
       g_autoptr (OstreeSign) sign = NULL;
       g_autoptr (GVariant) signatures = NULL;
       const gchar *signature_key = NULL;
       GVariantType *signature_format = NULL;
-      g_autoptr (GError) local_error = NULL;
 
-      if ((sign = ostree_sign_get_by_name (*iter, &local_error)) == NULL)
-        continue;
+      if ((sign = ostree_sign_get_by_name (*iter, error)) == NULL)
+        return FALSE;
```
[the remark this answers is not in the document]
cgwalters: Oh wow yes, the test suite relies on having each one try to parse the key type and ignore failures.

How about we change things so that instead it's e.g.:

```
verification-key=ed25519:<blah>
```

src/libostree/ostree-repo-pull.c
cgwalters: OK, though I struggle a bit with the fact that as of right now, we have exactly one signing implementation behind this that people should use (they shouldn't use `dummy` of course, in fact I'd like to change it so it requires something like `env OSTREE_TESTSUITE_FLAGS=enable-dummy-sigs ostree pull...` e.g.).

src/libostree/ostree-repo-pull.c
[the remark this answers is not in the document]
d4s: hm... good idea for adding prefix for `verification-key`, I think that `verification-file` should to use it as well. But that is not related to the current case -- for instance `ed25519` may work without those settings if public signatures are placed inside the "well-known" directories.

Let me to describe why I using the current approach:
I expect that `commitmeta` may contain several signature types with several signatures for each type. Currently we query the sign module for which signature types we are supporting and try to check if there are any of them in detached metadata -- one by one. I didn't wanted to propose separate API call for checking the engine by signature name from metadata.

So, for example in case if the commit has signed with only valid `ed25519` -- we should be able to check it. But with proposed changes it is leading that only `dummy` engine is queried and stop the check for the rest of engines since the detached `commitmeta` file doesn't contains signature of `dummy` type.
As a result it leads to  the failure of `tests/test-signed-pull.sh` for all `ed25519` test cases.

The similar problem we have if there is a `dummy` signature configuration is failed for some reason while loading public keys -- there will be no chance for other signature types to check if the commit is signed with correct signature.

The current behavior is designed to be able to work with multiple signature types -- and allows to sign the commit with several signatures and several signature types in the same time. I expect it will be useful for switching between signature types as soon as we will have some additional engines besides the `dummy` and `ed25519`. To be honest the `dummy` module has been added mostly for prototyping and checking the multiple signature types work.

src/libostree/ostree-repo-pull.c
d4s: hm... I wanted to add ability to switch off the `dummy` engine in a build time.
Please let me know which approach is more preferable?

cgwalters:
OK I reworked this - @d4s mind taking another look?

PR #2064, opened 2020-04-14
Only enable "dummy" signature type with opt-in env variable

cgwalters:
I don't want to even have to think about people using
this in production.


cgwalters:
@d4s thanks for the review, though as a general rule here in this repo we are using a "merge bot", which takes care of merging PRs only when CI is passing.  A bit more info in https://github.com/coreos/fedora-coreos-tracker/issues/263

So usually it's just enough to drop `/lgtm` and let the bot do it.

In this case...the coreosci is failing and I don't understand why yet.  Need to debug that.

d4s:
@cgwalters got it. Sorry for that mistake.
just in case -- I did local tests for this PR.

PR #2068, opened 2020-04-15
lib/pull: Two cosmetic internal function renames

cgwalters:
I'm mainly doing this to sanity check the CI state right now.

However, I also want to more cleanly/clearly distinguish
the "sign" code from the "gpg" code.

Rename one function to include `gpg`.

For the other...I think what it's really doing is using the remote
config, so change it to include `remote` in its name.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1440,7 +1440,7 @@ static_deltapart_fetch_on_complete (GObject           *object,
 
 #ifndef OSTREE_DISABLE_GPGME
 static gboolean
-process_verify_result (OtPullData            *pull_data,
+process_gpg_verify_result (OtPullData            *pull_data,
                        const char            *checksum,
                        OstreeGpgVerifyResult *result,
                        GError               **error)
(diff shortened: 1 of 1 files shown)


cgwalters:
OK yeah, something fell over in coreosci I think...but I simply cannot see what's wrong from looking at the Jenkins UI.  I just logged into the cluster and it looks like the relevant pods have already been GC'd.

cgwalters:
Ahh wait I see, looking at the legacy UI it's clear: https://jenkins-coreos-ci.apps.ci.centos.org/job/github-ci/job/ostreedev/job/ostree/view/change-requests/job/PR-2068/1/console

`groovy.lang.MissingPropertyException: No such property: cosaDir for class: org.codehaus.groovy.runtime.GStringImpl`

So this is likely fallout from groovy.lang.MissingPropertyException: No such property: cosaDir for class: org.codehaus.groovy.runtime.GStringImpl

jlebon:
Yup, sorry about that! Patch in https://github.com/ostreedev/ostree/pull/2069.

PR #2070, opened 2020-04-15
travis: Add some libsodium coverage

cgwalters:
As far as I can tell we're not gating on this right now.  From
a quick glance, it looks like Debian stable has `libsodium-dev`
but only Ubuntu eoan does which we're not testing right now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,9 @@ dist: trusty
 sudo: required
 
 env:
-  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch
-  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_configopts="--with-curl"
+  # debian has libsodium-dev, ubuntu doesn't in core at least
+  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_configopts="--with-libsodium" ci_pkgs="libsodium-dev"
+  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_configopts="--with-curl --with-libsodium --without-gpgme" ci_pkgs="libsodium-dev"
   - ci_docker=ubuntu:xenial ci_distro=ubuntu ci_suite=xenial
   - ci_docker=ubuntu:bionic ci_distro=ubuntu ci_suite=bionic
 
--- a/ci/travis-Dockerfile.in
+++ b/ci/travis-Dockerfile.in
@@ -2,7 +2,7 @@ FROM @ci_docker@
 ENV container docker
 
 ADD ci/travis-install.sh /travis-install.sh
-RUN ci_suite="@ci_suite@" ci_distro="@ci_distro@" ci_in_docker=yes /travis-install.sh
+RUN ci_suite="@ci_suite@" ci_distro="@ci_distro@" ci_pkgs="@ci_pkgs@" ci_in_docker=yes /travis-install.sh
 
 ADD . /home/user/ostree
 RUN chown -R user:user /home/user/ostree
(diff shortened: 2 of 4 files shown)


cgwalters:
Also...I really should have brought this up before (maybe I did and can't find it) but...remind me again why we're using libsodium instead of openssl for this?  Nothing else depends on libsodium in Fedora CoreOS today - we have curl configured for openssl and ostree already can use openssl for sha256.

It looks like it [has an interface we want](https://www.openssl.org/docs/manmaster/man3/EVP_DigestSignInit.html) right?

d4s:
@cgwalters the initial discussion is here: https://github.com/ostreedev/ostree/issues/1233
The problem is in licensing. The openssl is ok for Fedora but not for other distros, especially aimed for embedded.

.travis.yml
```diff
@@ -4,7 +4,7 @@ sudo: required
 
 env:
   - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch
-  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_configopts="--with-curl"
+  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_configopts="--with-curl --with-libsodium" ci_pkgs="libsodium-dev"
```
d4s, line 7: May I ask for additional test in different configuration with disabled GPG (`--without-gpgme`)?
So we need to test 3 build environments:
- GPG-only
- GPG + libsodium
- libsodium-only

This is that I do for checking local changes -- I discover some issues only in one of those configuration.

cgwalters: Well, our existing CI is gpg-only for the most part because we're not pulling in libsodium-dev.  But yep, changed this.

.travis.yml
```diff
@@ -3,8 +3,9 @@ dist: trusty
 sudo: required
 
 env:
-  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch
-  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_configopts="--with-curl"
+  # debian has libsodium-dev, ubuntu doesn't in core at least
+  - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_pkgs="libsodium-dev"
```
d4s, line 7: we have to explicitly provide `--with-libsodium` for both new tests.
Or change the autodetection to true in  `configure.ac`. For now we don't use the `libsodium` by default even if the package exists in the system:
```bash
$ dpkg -l libsodium-dev
ri  libsodium-dev:amd64 1.0.17-1     amd64        Network communication, cryptography and signaturing library - headers
```
```bash
$ ./configure  | grep sodium
    libsodium (ed25519 signatures):               no
```

d4s: @cgwalters I'm not sure if you plan to enable the `libsodium` support by default, so not sure about the exact change.

And additional thing -- it seems to me that `ci_suite=stretch` should be `ci_suite=buster`, but again I'm unsure if it is the idea or issue. Sorry about that.

cgwalters: Right, I had that originally and then got confused and thought it was enabled by default.  Fixed.

d4s:
@cgwalters I'll fix the build on debian/ubuntu and push the change to MR

d4s:
@cgwalters I've submit MR with needed changes to your branch

cgwalters:
> @cgwalters I've submited PR with needed changes to your branch

Thanks!  The "submit PR to PR" is actually a fairly cool pattern I hadn't seen used before.  We've often used the fact that Github normally allows repository committers to push to PRs instead (though a downside of that is the PR author really needs to know that their branch will be updated/force-pushed).

PR #2072, opened 2020-04-16
Use `sign-ed25519` for the feature name

cgwalters:
`libsodium` is an implementation detail.  In particular, I'd like
to consider using OpenSSL for ed25519 (if libsodium isn't configured
and openssl is).

So switch the name of the exposed feature and adjust the tests.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -253,7 +253,7 @@ AS_IF([test x$with_libsodium != xno], [
     AS_IF([ test x$have_libsodium = xno ], [
        AC_MSG_ERROR([Need LIBSODIUM version $LIBSODIUM_DEPENDENCY or later])
     ])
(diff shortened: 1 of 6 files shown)


jlebon:
Seems reasonable to me! Will let @d4s do final review and approval.
/approve

cgwalters:
> Am I understood correctly: you want to implement the same functionality for ed25519 with openssl API and switch between them with HAVE_LIBSODIUM and HAVE_OPENSSL macroses, right?

Yep.

> In that case I'd propose to rename --with-libsodium option to --with-ed25519-libsodium, so later it would be possible to add --with-ed25519 or --with-ed25519-openssl which will be mutually exclusive with libsodium-based implementation.

Yeah, sounds reasonable to me; but let's do it separately from this PR?

/test sanity

PR #2073, opened 2020-04-17
lib: Move internal binding verification API to repo.c

cgwalters:
`ostree-repo-pull.c` is rather monstrous; I plan to split it
up a bit.  There's actually already a `pull-private.h` but
that's just for the binding verification API.  I think that one
isn't really pull specific.  Let's move it into the "catchall"
`repo.c`.

PR #2074, opened 2020-04-17
tests: Remove (spurious) key generation to stdout

cgwalters:
Nothing is using this right now, I don't quite understand
the intention of this part of the test.

And doing it to stdout will cause random test failures because
the key might happen to start with `ok` which will make TAP
think it's a passed test, and it's an error to have unplanned tests.


cgwalters:
@d4s this happened in https://github.com/ostreedev/ostree/pull/2073#issuecomment-615240484

d4s:
argh! thanks for finding that.
my fault -- there should be created a lot of files in well known files with fake signatures.
Let me to really fix the test.

PR #2075, opened 2020-04-17
tests/signed-commit: fix the test of well-known places

d4s:
Commit e474033e removed the redirection of incorrect public keys
aimed to generate a lot of files without correct public signature.

Fix the test by returning back the creation of files containing
incorrect public keys for ed25519.

Signed-off-by: Denis Pynkin <denis.pynkin@collabora.com>


cgwalters:
OpenShift CI is overloaded, but the other two are fine, so
/override ci/prow/sanity

Issue #2038, opened 2020-03-20
test-concurrency.py sometimes fails


jlebon:
See e.g. https://github.com/ostreedev/ostree/pull/2025#issuecomment-600917509:

```
ERROR: tests/test-concurrency.py
================================
n_committers 4 n_pruners 0
Creating tree tree0
Creating tree tree1
PID 30315 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 30316 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 30317 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 30318 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 30316 exited OK
PID 30318 exited OK
PID 30317 exited OK
PID 30315 exited OK
n_committers 6 n_pruners 3
Creating tree tree0
Creating tree tree1
Creating tree tree2
PID 30321 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 30322 ostree --repo=repo commit --fsync=0 -b tree0 --tree=dir=tree0
PID 30323 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 30324 ostree --repo=repo commit --fsync=0 -b tree1 --tree=dir=tree1
PID 30325 ostree --repo=repo commit --fsync=0 -b tree2 --tree=dir=tree2
PID 30326 ostree --repo=repo commit --fsync=0 -b tree2 --tree=dir=tree2
error: opendir(staging-89ed2f5d-09e9-4dc1-8360-9d7011bc2edd-anBXJW): No such file or directory
PID 30327: ostree --repo=repo prune --refs-only
PID 30328: ostree --repo=repo prune --refs-only
PID 30329: ostree --repo=repo prune --refs-only
process 30322 exited with code 1
PID 30326 exited OK
PID 30324 exited OK
PID 30321 exited OK
PID 30323 exited OK
PID 30325 exited OK
PID 30328 exited OK
PID 30327 exited OK
PID 30329 exited OK
A child process exited abnormally
1..2
014ccd06e67eacefc6a965c01a3fe3aa953cbac93cd17ab49b7cc35325b9e8b9
67eddf38725595bf7a23c692101b6edded083c8fbc5b099ce7be5afa6f56f90c
[...]

PR #2077, opened 2020-04-17
lib/commit: Check that dirent is a directory before cleaning 

jlebon:
I've only noticed this by inspection. But I think it's possible for
`cleanup_txn_dir` to get called with the `staging-...-lock` file since
it matches the prefix.

Make the checking here stronger by verifying that it's a directory. If
it's not a directory (lockfile), then follow the default pruning expiry
logic so that we still cleanup stray lockfiles eventually.


cgwalters:
Ooh...this could be the source of https://github.com/ostreedev/ostree/issues/2038
/approve
/lgtm

Issue #1793, opened 2018-12-26
Factory reset with OSTree


AdrianVovk:
Hello!

What would be a safe way to "factory reset" an OSTree deployment?

I'm unsure what folders I can delete from `/ostree` to clear out `/etc` and `/var` without messing anything up.

It would be nice if there was an `ostree admin reset` command that would wipe out these folders safely

Thank you! Happy holidays!

cgwalters:
You generally shouldn't change anything under `/ostree` manually.  Factory reset is mostly:

```
rm -rf /var/*
rsync -rlv --delete /usr/etc /etc
```

However, that will also e.g. remove `/etc/fstab` which some setups may not want.

We also don't currently have a model for "factory" kernel arguments, and hence no way to reset those.

But the high level intent is that almost all state is under `/etc` and `/var`.

AdrianVovk:
That is something I can do, but not from outside a deployment. If I wanted to do this from an installer, for example, it would be nice to be able to do something like `ostree admin reset --sysroot=/mnt` instead of somehow mounting /etc (?) and /var and then clearing them out...

Kernel arguments can be specified somehow from /usr or even be left blank like they are on the default deployment. They aren't an issue; anything important should be run with `ostree deploy` by the distro and anything the user added should disappear. I haven't looked into the ostree configuration dirs, but could one specify kargs in the repo config?

akiernan:
My factory restore is a secure erase of (a separate) `/var` and a rewrite of the rootfs from an image inside the bootloader. But being able to do a deploy without the three-way merge of `/etc` would be enough for me - getting a cleaned up deployment after someone's fiddled with `/etc` would be a useful developer workflow. 

Then again my use case is for thousands of completely identical devices, modulo a very small amount of state which I store outside of the standard filesystem.

AdrianVovk:
@cgwalters I tested the commands you gave, and they have the right idea

I've had most success running this set of commands:
```
$ systemctl isolate emergency
[Type root password]
$ rm -rf /etc/{,.}*
$ rm -rf /var/{,.}*
$ rm -rf /home/{,.}*
$ rsync -aHX /usr/etc/* /etc
$ systemctl reboot
```

I have a goal of getting the system to boot with an empty /etc (which it cannot do at the moment). I'm waiting for systemd/systemd#11287 for this

It would still be nice to be able to wipe these directories from an external mount, like from recovery media. How would I do that?

Thank you

wmanley:
[...]

PR #2081, opened 2020-04-21
deploy: Add --no-merge

cgwalters:
All of the underlying libostree APIs have supported passing `NULL`
for a merge deployment for...a long time.  But we never plumbed
it up into the CLI.

Add a `--no-merge` option to aid people who want to do a "factory reset":
https://github.com/ostreedev/ostree/issues/1793

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -74,6 +74,7 @@ parallel fcos: {
         mkdir -p overrides/rootfs
         mv insttree/* overrides/rootfs/
         rmdir insttree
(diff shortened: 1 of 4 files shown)


cgwalters:
Argh, state shared across the admin base test into uboot.  Let's just make it a separate test.

cgwalters:
OK I just squashed in the same fix as https://github.com/coreos/rpm-ostree/pull/2067 here.

jlebon:
Nice!
/lgtm

PR #2083, opened 2020-04-24
Doc updates

agners:
Small docs updates which I find were missing/unclear while trying to understand the OSTree repository types/format better.


cgwalters:
Awesome, thanks!
/approve
/ok-to-test
/lgtm

PR #2084, opened 2020-04-24
signapi: expose metadata format and key

d4s:
Explicitly expose functions for querying the metadata format
and key name used by OstreeSign object:
 - ostree_sign_metadata_format
 - ostree_sign_metadata_key

This allows to use the same metadata format and key name
by 3-rd party applications using signapi.

I thought there is no need to use them externally but we have at least one case then it helps to avoid duplication of name and format in code.

### DIFF
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -719,6 +719,8 @@ ostree_sign_get_name
 ostree_sign_add_pk
 ostree_sign_clear_keys
 ostree_sign_load_pk
+ostree_sign_metadata_format
+ostree_sign_metadata_key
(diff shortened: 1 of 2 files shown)


jlebon:
(I've now unhooked @rh-atomic-bot from this repo.)

Patch itself seems reasonable to me.
/approve

cgwalters:
/lgtm
but two things.  First, everything with `_OSTREE_PUBLIC` should be listed in symbols, or should not have `_OSTREE_PUBLIC`.  (So this is fixing a prior bug)
Second, I would like to consider reworking this API considerably but that can come after this.

Issue #1243, opened 2017-10-03
OstreeRepoListRefsExtFlags GIR changed


dbnicholson:
The GIR for OstreeRepoListRefsExtFlags changed from `enumeration` to `bitfield`. Here's what I have from a 2016.15 system:
```
    <enumeration name="RepoListRefsExtFlags"
                 c:type="OstreeRepoListRefsExtFlags">
      <member name="repo_list_refs_ext_none"
              value="0"
              c:identifier="OSTREE_REPO_LIST_REFS_EXT_NONE">
        <doc xml:space="preserve">No flags.</doc>
      </member>
    </enumeration>
```
And from a 2017.11 system:
```
    <bitfield name="RepoListRefsExtFlags" c:type="OstreeRepoListRefsExtFlags">
      <member name="none"
              value="0"
              c:identifier="OSTREE_REPO_LIST_REFS_EXT_NONE">
        <doc xml:space="preserve">No flags.</doc>
      </member>
      <member name="aliases"
              value="1"
              c:identifier="OSTREE_REPO_LIST_REFS_EXT_ALIASES">
        <doc xml:space="preserve">Only list aliases.  Since: 2017.10</doc>
      </member>
      <member name="exclude_remotes"
              value="2"
              c:identifier="OSTREE_REPO_LIST_REFS_EXT_EXCLUDE_REMOTES">
        <doc xml:space="preserve">Exclude remote refs.  Since: 2017.11</doc>
      </member>
    </bitfield>
```
These were built on different systems, so perhaps this is a change in `g-ir-scanner`. However, maybe this change was due to 7ed881b.

cgwalters:
Hm yes...fun.  It was indeed that commit.  The values are ABI compatible from a C perspective, but not IABI obviously.  The value was *named* `Flags` but was missing one of the magic bits like `/* flags */` or just using left shifts to convince gtk-doc/g-ir-scanner that it was flags.

cgwalters:
Did you have a static analysis tool that noticed this IABI change, or did a script actually break?

dbnicholson:
I was just testing one of our python scripts that uses `OSTree.RepoListRefsExtFlags.REPO_LIST_REFS_EXT_NONE` and it broke when I ran it against a newer ostree.

cgwalters:
[...]

PR #2089, opened 2020-04-30
lib: Coerce flags enums to GIR bitfields

dbnicholson:
The GI scanner decides if an `enum` is really a `bitfield` if it finds
any values that have left shifts. With an `enumeration`, the
introspecting language may error or convert to a different type if the
user tries to combine values. Change all Flags `enum`s to use
left-shifted values so that they're represented as `bitfield`s in the
GIR.

The primary bug here is that you can't combine `REFS_ONLY` and
`NO_PRUNE` when calling `OSTree.Repo.prune()` from an introspected
language.

This is an IABI break since the typelib will change from `enumeration`
to `bitfield`. `OstreeRepoImportFlags` is internal but the change is
included here to prepare for a subsequent name that would require bit
shifting to operate correctly as a flag.


cgwalters:
Ah yes.    OK so the core question is

> This is an IABI break since the typelib will change from enumeration to bitfield. 

IABI breaks are especially painful for dynamic languages - and compounding the problem, code searches for dynamic languages are harder.

I did a quick check at Debian code search:

 - [RepoImportFlags](https://codesearch.debian.net/search?q=RepoImportFlags&literal=1)
 - [RepoPruneFlags](https://codesearch.debian.net/search?q=RepoPruneFlags&literal=1)

No hits outside of ostree/ostree-go.  

But I know of stuff not packaged in Debian, for example here's [a hit in ostree-releng-scripts](https://github.com/ostreedev/ostree-releng-scripts/search?q=RepoPruneFlags&unscoped_q=RepoPruneFlags).  

I think that's fine in Python to change right?

> The primary bug here is that you can't combine REFS_ONLY and NO_PRUNE when calling OSTree.Repo.prune() from an introspected language.

And presumably you want to write such code?  Where is it, just for future reference?

dbnicholson:
> But I know of stuff not packaged in Debian, for example here's [a hit in ostree-releng-scripts](https://github.com/ostreedev/ostree-releng-scripts/search?q=RepoPruneFlags&unscoped_q=RepoPruneFlags).

Github doesn't do too bad of a job [here](https://github.com/search?q=RepoPruneFlags&type=Code). Relevant ones:

https://github.com/ostreedev/ostree-releng-scripts/blob/d7d3984914389849bffec689b243512304dd1aec/reset-if-staging
https://github.com/lirios/ostree-upload/blob/d0a722c6a87fe30a4ad7ae9d16c314833d61016d/src/pusher.rs
https://github.com/refi64/flatpod
https://github.com/projectatomic/atomic/blob/4cf54d1e7a77dccfc8e4d52d6bf8d89c68e47705/Atomic/syscontainers.py

A few more are in some endlessm private repos that you wouldn't see.

> I think that's fine in Python to change right?

In pygobject all the uses there would be fine, although I swear it used to use a different format for enumeration vs bitfield. In #1243 I had been using `OSTree.RepoListRefsExtFlags.REPO_LIST_REFS_EXT_NONE` and that changed to `OSTree.RepoListRefsExtFlags.NONE` when it got switched to a bitfield. But today I'm not seeing that kind of notation with pygobject 3.32 I have on hand. Anyways, none of the searched examples are using that funky notation. Actually, now that I look closer I think it was really a gobject-instrospection change since it specifies `member_name` in the GIR. Presumably pygobject was just following what it said there.

The other examples are rust and JS, and I do know what happens with those, but the usage looks fine. Now that I think on it again, I don't think this will cause any breakage since you could only set a single value before and now you can combine them. If you went the other way - bitfield to enumeration, you could definitely have uses out in the wild that would break.

> > The primary bug here is that you can't combine REFS_ONLY and NO_PRUNE when calling OSTree.Repo.prune() from an introspected language.
> 
> And presumably you want to write such code? Where is it, just for future reference?

I just wanted to have a dry run mode for a pruning script I was writing like `ostree prune --no-prune`, so I was going to combine the flags. Unfortunately, most of our ostree management tooling is in a private repo. I actually would like to get them out now that we've moved our repo configuration somewhere else. What's left in there has some crazy Endless specific stuff but nothing sensitive.

cgwalters:
> Github doesn't do too bad of a job here.

Nice, I hadn't tried a global search like that recently.  Some projects there I hadn't heard of, neat.

/approve

> The other examples are rust and JS, and I do know what happens with those, but the usage looks fine. 

Rust is a statically typed language and almost all static languages differentiate enums and flags.  I can't think of one that doesn't.

For example, it is [currently an enum](https://fkrull.gitlab.io/ostree-rs/ostree/enum.RepoPruneFlags.html) which implies exclusivity.

So this will be an IABI break for them...but on balance I think we should do it.  If this was e.g. a core part of a GTK+ API the calculus might be different.

/lgtm

PR #2091, opened 2020-05-06
commit: Add --mode-ro-executables

cgwalters:
I think we should encourage removing the writable bits from
executables.  This has happened to me:
https://thomask.sdf.org/blog/2019/11/09/take-care-editing-bash-scripts.html

And not having the writable bit may help prevent hardlink
corruption with OSTree in some cases.

We can't do this by default, but add a convenient CLI flag
for it.

### DIFF
--- a/src/ostree/ot-builtin-commit.c
+++ b/src/ostree/ot-builtin-commit.c
@@ -56,6 +56,7 @@ static gboolean opt_no_xattrs;
 static char *opt_selinux_policy;
 static gboolean opt_selinux_policy_from_base;
 static gboolean opt_canonical_permissions;
+static gboolean opt_ro_executables;
 static gboolean opt_consume;
 static gboolean opt_devino_canonical;
 static char *opt_base;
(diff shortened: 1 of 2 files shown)


src/ostree/ot-builtin-commit.c
```diff
@@ -111,6 +112,7 @@ static GOptionEntry options[] = {
   { "owner-uid", 0, 0, G_OPTION_ARG_INT, &opt_owner_uid, "Set file ownership user id", "UID" },
   { "owner-gid", 0, 0, G_OPTION_ARG_INT, &opt_owner_gid, "Set file ownership group id", "GID" },
   { "canonical-permissions", 0, 0, G_OPTION_ARG_NONE, &opt_canonical_permissions, "Canonicalize permissions in the same way bare-user does for hardlinked files", NULL },
+  { "mode-w-xor-x", 0, 0, G_OPTION_ARG_NONE, &opt_mode_w_xor_x, "Ensure executable files are not writable", NULL },
```
jlebon, line 115: Hmm, that's a bit cryptic (also I think it's technically a nand, not a xor :) ). Maybe something like `--ro-executables`?

tests/basic-test.sh
```diff
@@ -541,6 +541,24 @@ fi
...
+cd ${test_tmpdir}
+rm test2-checkout -rf
+$OSTREE checkout test2 test2-checkout
+cd test2-checkout
+touch user-wx
+chmod 0755 user-wx
+touch group-wx
+chmod 0575 group-wx
+touch both-wx
+chmod 0775 both-wx
```
jlebon, line 553: Maybe a sanity-check that a writable but not executable file doesn't get its write perms stripped?

src/ostree/ot-builtin-commit.c
cgwalters: "xor" is about the "rule" and not the operation - https://en.wikipedia.org/wiki/W%5EX

jlebon: Ahh gotcha, didn't realize that.

Issue #2080, opened 2020-04-20
signing: explicit expected signature types


cgwalters:
First...I find terminology tricky here because when we say "signing" it could e.g. apply to GPG too.  I've been starting to use the term "signapi" to distinguish the two.  What do people think about that?  

A lot of the signapi APIs¹ mirror the way GPG works.  And that would seem to make sense...except, they also have some important differences.  

I think the core thing that really bothers me right now is that we have a `sign_verify` boolean (matching GPG).  And the logic here is "accept any valid signature from a signapi backend".  We end up using the list of signatures in the commit we pulled as a way to determine which backends to use.

It would make me way more comfortable instead if `sign-verify` was a list of strings, so the client side instead explicitly says which signatures it expects to see.

```
sign-verify=ed25519,somethingelse
```

This does currently happen *implicitly* right now because the repo has to now say `verification-ed25519-key`, but doing the above would make the intent very clear.

For example, we would immediately hard error out if we have `sign-verify=ed25519` and there isn't an `ed25519` signature.

(It'd actually be really nice if we could make "gpg" be a backend behind the signapi...I didn't think about that much yet though)

/cc @d4s 

¹ Hm well that redundancy is an example of how tricky naming things is...

d4s:
Initially I used GPG as example, having in mind that somebody and sometime may want to switch GPG to signapi as well ;)

The explicit list of needed engines looks reasonable. But I want to keep 2 cases as well:
- using the implicit behavior, i.e. "use any available correctly configured engine"
- deny the signapi usage at all (current default)

If it can be managed by `true` and `false` strings for `sign-verify` option -- it would be wonderful, since in that case we will keep backward compatibility with the initial behavior (yes, I know that this is the new feature for libostree but it is a half-year old feature for Apertis distro ;) ) But it is acceptable any other string values for those 2  cases.

Regarding `verification-<engine>-key` and `verification-<engine>-file` variables -- they may exist, but even if they don't -- the engine could be used if it support other configuration approach.
For example the current `ed25519` implementation works well without having those variables in config file. As a fallback variant it tries to load any valid public signature from files placed in well-known places, for instance in files from `/usr/share/ostree/trusted.ed25519.d/` or `/etc/ostree/trusted.ed25519.d/` directories. The similar behavior is for revoked signatures.

[...]

PR #2092, opened 2020-05-07
remote-add: Add --sign-verify=KEYTYPE=[inline|file]:PUBKEYREF 

cgwalters:
Per https://github.com/ostreedev/ostree/issues/2080#issuecomment-623614483

A huge benefit of ed25519 (and ECC in general) is that keys are very
short - short enough that it's completely reasonable to inline
them into a command line argument.

And I think that's a good model; it makes the keys very visible.

For example, someone could easily copy-paste a commandline
argument from a webpage (secured via TLS) that says to run
`ostree remote add --verify-ed25519=KEY`.

### DIFF
--- a/src/ostree/ot-remote-builtin-add.c
+++ b/src/ostree/ot-remote-builtin-add.c
@@ -32,6 +32,7 @@ static gboolean opt_no_sign_verify;
 static gboolean opt_if_not_exists;
 static gboolean opt_force;
 static char *opt_gpg_import;
+static char **opt_sign_verify;
 static char *opt_contenturl;
 static char *opt_collection_id;
 static char *opt_sysroot;
@@ -46,6 +47,7 @@ static GOptionEntry option_entries[] = {
   { "set", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_set, "Set config option KEY=VALUE for remote", "KEY=VALUE" },
   { "no-gpg-verify", 0, 0, G_OPTION_ARG_NONE, &opt_no_gpg_verify, "Disable GPG verification", NULL },
   { "no-sign-verify", 0, 0, G_OPTION_ARG_NONE, &opt_no_sign_verify, "Disable signature verification", NULL },
+  { "sign-verify", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_sign_verify, "Verify signatures using KEYTYPE=inline:PUBKEY or KEYTYPE=file:/path/to/key", "KEYTYPE=[inline|file]:PUBKEY" },
   { "if-not-exists", 0, 0, G_OPTION_ARG_NONE, &opt_if_not_exists, "Do nothing if the provided remote exists", NULL },
   { "force", 0, 0, G_OPTION_ARG_NONE, &opt_force, "Replace the provided remote if it exists", NULL },
   { "gpg-import", 0, 0, G_OPTION_ARG_FILENAME, &opt_gpg_import, "Import GPG key from FILE", "FILE" },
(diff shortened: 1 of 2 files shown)


src/ostree/ot-remote-builtin-add.c
```diff
@@ -46,6 +47,7 @@ static GOptionEntry option_entries[] = {
   { "set", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_set, "Set config option KEY=VALUE for remote", "KEY=VALUE" },
   { "no-gpg-verify", 0, 0, G_OPTION_ARG_NONE, &opt_no_gpg_verify, "Disable GPG verification", NULL },
   { "no-sign-verify", 0, 0, G_OPTION_ARG_NONE, &opt_no_sign_verify, "Disable signature verification", NULL },
+  { "verify-ed25519", 0, 0, G_OPTION_ARG_STRING, &opt_verify_ed25519, "Verify signatures ed25519 public key PUBKEY", "PUBKEY" },
```
jlebon, line 50:
```suggestion
  { "verify-ed25519", 0, 0, G_OPTION_ARG_STRING, &opt_verify_ed25519, "Verify signatures using ed25519 public key string", "PUBKEY" },
```

?

cgwalters: I wish github offered "squash suggestion into commit, add Co-authored-by"

d4s:
@cgwalters agree, it is a good idea to add remote with a key.
However I don't think `--verify-ed25519` option is a good name for usage.
It is too `ed25519`-centric, while the signapi for now is able to operate with several signing engines.
Please check options in [src/ostree/ot-builtin-sign.c](https://github.com/ostreedev/ostree/blob/master/src/ostree/ot-builtin-sign.c) as instance -- if we use option `--sign-type` for engine selection, we may pass-through the key, file with keys or basedir as additional option. And in this case it works for **any** engine we have, so do not need to provide engine-specific things in `ostree` builtins.

cgwalters:
> However I don't think --verify-ed25519 option is a good name for usage. It is too ed25519-centric, while the signapi for now is able to operate with several signing engines.

Yes, I split this out of a more generic patch to support `--signapi-verify=TYPE=KEY`...hm, OK I'll do that instead.

cgwalters:
> Yes, I split this out of a more generic patch to support --signapi-verify=TYPE=KEY...hm, OK I'll do that instead.

Oh right I remember why I switched now - we don't know that new key types will be ergonomic to add to the commandline inline.  So it'd have to be more like:

`remote add --sign-verify=<keytype>=inline:<key>`
`remote add --sign-verify=<keytype>=file:/path/to/file`
or something?

cgwalters:
OK reworked this - @d4s what do you think?

tests/test-signed-pull.sh
```diff
@@ -98,13 +98,30 @@ ${CMD_PREFIX} ostree --repo=repo config set 'remote "origin"'.verification-dummy
 ${CMD_PREFIX} ostree --repo=repo config unset 'remote "origin"'.verification-dummy-file
 test_signed_pull "dummy" ""
 
+repo_init
+if ${CMD_PREFIX} ostree --repo=repo remote add other --sign-verify=trustme=inline:ok http://localhost 2>err.txt; then
+    assert_not_reached "remote add with invalid keytype succeeded"
+fi
+assert_file_has_content err.txt 'Requested signature type is not implemented'
+if ${CMD_PREFIX} ostree --repo=repo remote add other --sign-verify=ed25519 http://localhost 2>err.txt; then
```
d4s, line 106: 1. Please move all `ed25519` related tests below -- after `has_sign_ed25519` check -- in other case it throws an error if there is no ed25519 support: `error: Requested signature type is not implemented`
2. And as far as you there -- please add the test for `dummy` engine with inlined key -- it should to work everywhere.

cgwalters: OK, done.

PR #2093, opened 2020-05-08
ostree-prepare-root: Requires=sysroot.mount

cgwalters:
With just `After=` we'll still try to run in the scenario
where `sysroot.mount` fails because the rootfs didn't appear.
And this will end up spewing an error which can confuse people
into thinking something is wrong at the ostree level.

This has come up numerous times w/{Fedora,RHEL} CoreOS, most
recently while looking at
https://bugzilla.redhat.com/show_bug.cgi?id=1803130


vtolstov:
Nice addition, I'm more then two time hit this issue

PR #2094, opened 2020-05-09
Changing link to file about contributing, in readme.

zpiotr:
You don't must to go to docs and then to file.  Now you can just click on the link in the readme file. It's a very small change but I think it's now easier to go to contribution file than before.


cgwalters:
Hi @zpiotr and thanks for your contribution!  Just one small request, can you adjust your commit message to match the style of the rest of the logs?  See `git log` here.

For example:

```
README.md: Fix link to CONTRIBUTING.md

We should link to the target and not the symlink.
```

zpiotr:
Okey, name is changed.

cgwalters:
/lgtm cancel
Wait we have 3 commits now

cgwalters:
OK I think I see, you are making these commits from the Github web interface or so, which always makes a new commit or something?

Can you scroll to the bottom of this PR and check the "Allow edits from maintainers" box?

zpiotr:
Yep, I using web interface because I think it's a too small commit to use git for this.  Yes, web interface always make new commits, and "Allow edits from maintainers"  is "on". I have a question, can you tell me what can I do for you? I programming in python, C, and golang. I'm new in contributing and I don't know what can I do.

PR #2095, opened 2020-05-10
signing: Change API to create instances directly

cgwalters:
This cleans up the verification code; it was weird how
we'd get the list of known names and then try to create
an instance from it (and throw an error if that failed, which
couldn't happen).

### DIFF
--- a/src/libostree/ostree-repo-pull-verify.c
+++ b/src/libostree/ostree-repo-pull-verify.c
@@ -142,6 +142,9 @@ _signapi_load_public_keys (OstreeSign *sign,
   return TRUE;
 }
 
+/* Iterate over all known signing types, and check if the commit is signed
+ * by at least one.
+ */
 gboolean
 _sign_verify_for_remote (OstreeRepo *repo,
                           const gchar *remote_name,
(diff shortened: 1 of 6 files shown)


src/libostree/ostree-repo-pull-verify.c
```diff
@@ -150,31 +150,19 @@ _sign_verify_for_remote (OstreeRepo *repo,
...
+  for (guint i = 0; i < signers->len; i++)
     {
-      g_autoptr (OstreeSign) sign = NULL;
-      g_autoptr (GVariant) signatures = NULL;
-      const gchar *signature_key = NULL;
-      GVariantType *signature_format = NULL;
-
-      if ((sign = ostree_sign_get_by_name (*iter, NULL)) == NULL)
-        {
-          n_unknown_signatures++;
```
cgwalters, line 168: Also I think previously I said something like we were parsing the list of signatures from the commit object - in doing this commit I realized I was wrong.  

But we have a deeper problem which is that because we've mixed in "signapi signatures" with other metadata, we can't do anything like say "this commit was signed, but with an engine we don't know about".

Oh hmm, I guess we could actually check the metadata to see if there's any `ostree.sign.*` keys that we don't know about.  So yeah this is fine, nevermind!

PR #2098, opened 2020-05-13
finalize: Add RequiresMountsFor=/boot too

cgwalters:
In https://bugzilla.redhat.com/show_bug.cgi?id=1827712
some OpenShift CI is seeing `/boot` being unmounted before
`ostree-finalize-staged.service` runs or completes.

We finally tracked this down to a bug elsewhere, but
I think we should add this because it clearly shows
our requirements.

### DIFF
--- a/src/boot/ostree-finalize-staged.service
+++ b/src/boot/ostree-finalize-staged.service
@@ -23,7 +23,7 @@ Documentation=man:ostree(1)
 ConditionPathExists=/run/ostree-booted
 DefaultDependencies=no
 
-RequiresMountsFor=/sysroot
+RequiresMountsFor=/sysroot /boot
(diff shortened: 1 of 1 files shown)


vtolstov:
if i have mounted boot but in readonly mode? does requires mounts for check also that path is writable ?

cgwalters:
> if i have mounted boot but in readonly mode? does requires mounts for check also that path is writable ?

This doesn't affect readonly state.  See https://github.com/ostreedev/ostree/pull/1767 for that.

vtolstov:
@cgwalters thanks!

cgwalters:
xref https://ask.fedoraproject.org/t/silverblue-missing-boot-option/7330/5

PR #2099, opened 2020-05-14
lib/pull: Add `timestamp-check-from-rev`

jlebon:
The way `timestamp-check` works might be too restrictive in some
situations. Essentially, we need to support the case where users want to
pull an older commit than the current tip, but while still guaranteeing
that it is newer than some even older commit.

This will be used in Fedora CoreOS. For more information see:
https://github.com/coreos/rpm-ostree/pull/2094
https://github.com/coreos/fedora-coreos-tracker/issues/481

### DIFF
--- a/src/libostree/ostree-repo-pull-private.h
+++ b/src/libostree/ostree-repo-pull-private.h
@@ -114,6 +114,7 @@ typedef struct {
   guint             n_imported_content;
 
   gboolean          timestamp_check; /* Verify commit timestamps */
+  char             *timestamp_check_from_rev;
   int               maxdepth;
   guint64           max_metadata_size;
   guint64           start_time;
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1593,6 +1593,7 @@ scan_commit_object (OtPullData                 *pull_data,
                                      commit, error))
     return glnx_prefix_error (error, "Commit %s", checksum);
 
+  guint64 new_ts = ostree_commit_get_timestamp (commit);
   if (pull_data->timestamp_check)
     {
       /* We don't support timestamp checking while recursing right now */
@@ -1611,10 +1612,21 @@ scan_commit_object (OtPullData                 *pull_data,
             return glnx_prefix_error (error, "Reading %s for timestamp-check", ref->ref_name);
 
           guint64 orig_ts = ostree_commit_get_timestamp (orig_commit);
-          guint64 new_ts = ostree_commit_get_timestamp (commit);
           if (!_ostree_compare_timestamps (orig_rev, orig_ts, checksum, new_ts, error))
             return FALSE;
         }
+    }
+  if (pull_data->timestamp_check_from_rev)
+    {
+      g_autoptr(GVariant) commit = NULL;
+      if (!ostree_repo_load_commit (pull_data->repo, pull_data->timestamp_check_from_rev,
+                                    &commit, NULL, error))
+        return glnx_prefix_error (error, "Reading %s for timestamp-check-from-rev",
+                                  pull_data->timestamp_check_from_rev);
+
+      guint64 ts = ostree_commit_get_timestamp (commit);
+      if (!_ostree_compare_timestamps (pull_data->timestamp_check_from_rev, ts, checksum, new_ts, error))
+        return FALSE;
     }
 
   /* If we found a legacy transaction flag, assume all commits are partial */
@@ -3270,6 +3282,7 @@ initiate_request (OtPullData                 *pull_data,
  *   * require-static-deltas (b): Require static deltas
  *   * override-commit-ids (as): Array of specific commit IDs to fetch for refs
  *   * timestamp-check (b): Verify commit timestamps are newer than current (when pulling via ref); Since: 2017.11
+ *   * timestamp-check-from-rev (s): Verify that all fetched commit timestamps are newer than timestamp of given rev; Since: 2020.4
  *   * metadata-size-restriction (t): Restrict metadata objects to a maximum number of bytes; 0 to disable.  Since: 2018.9
  *   * dry-run (b): Only print information on what will be downloaded (requires static deltas)
  *   * override-url (s): Fetch objects from this URL if remote specifies no metalink in options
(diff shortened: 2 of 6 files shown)


cgwalters:
This seems fine but...I don't see any downsides to changing libostree to do what rpm-ostree is doing by default instead.

jlebon:
> This seems fine but...I don't see any downsides to changing libostree to do what rpm-ostree is doing by default instead.

Hmm, can you expand? The pull code has no concept of deployments, right? How would it know what the base revision is?

cgwalters:
> Hmm, can you expand? The pull code has no concept of deployments, right? How would it know what the base revision is?

I think `OstreeSysrootUpgrader` would pass this in internally down to the pull code.  But...I guess we forked that for rpm-ostree.

dustymabe:
> @cgwalters 
> This seems fine but...I don't see any downsides to changing libostree to do what rpm-ostree is doing by default instead.

I think I agree (if technically possible). The description of the behavior in https://github.com/coreos/rpm-ostree/pull/2094#issue-418155181 is what I thought downgrade protection meant universally:

- *For rpm-ostree, it means not deploying any commit which is older than what the current deployment is on.*

> @jlebon
> Hmm, can you expand? The pull code has no concept of deployments, right? How would it know what the base revision is?

If that's true then do we *need* downgrade protection here (in addition to rpm-ostree)?

cgwalters:
> If that's true then do we need downgrade protection here (in addition to rpm-ostree)?

I want libostree to work on its own too - with this we aren't helping `ostree admin upgrade`.  To be clear, fixing `OstreeSysrootUpgrader` isn't a blocker from my PoV, but let's keep in mind that it should probably be fixed for this too.

cgwalters:
aside: I am pretty happy with how the "layers" in libostree ended up, but man there are some fun "layer-crossing" problems (this always happens).

dustymabe:
> > If that's true then do we need downgrade protection here (in addition to rpm-ostree)?
> 
> I want libostree to work on its own too - with this we aren't helping `ostree admin upgrade`. To be clear, fixing `OstreeSysrootUpgrader` isn't a blocker from my PoV, but let's keep in mind that it should probably be fixed for this too.

Yep I was forgetting about libostree without rpm-ostree.

jlebon:
> I want libostree to work on its own too - with this we aren't helping ostree admin upgrade. To be clear, fixing OstreeSysrootUpgrader isn't a blocker from my PoV, but let's keep in mind that it should probably be fixed for this too.

Making the OSTree sysroot upgrader use this flag too WFM. I misunderstood your original comment as suggesting an alternative way to solve this.

jlebon:
Updated sysroot upgrader! :arrow_up:

jlebon:
All green!

cgwalters:
/lgtm
Though I'd also like to also change `OstreeSysrootUpgrader` to use the deployment at some point.

jlebon:
>  Though I'd also like to also change `OstreeSysrootUpgrader` to use the deployment at some point.

Isn't that what the last commit here is doing?

cgwalters:
Yep, you're right, thanks!

cgwalters:
Hum interesting, [this PR changing unrelated code failed here](https://jenkins-coreos-ci.apps.ci.centos.org/job/github-ci/job/ostreedev/job/ostree/job/PR-2101/1/artifact/test-suite.log).

I haven't debugged why yet; at first I thought it might be we were making commits in under a second, but that still wouldn't appear as a downgrade.  I wonder if it could be something like clock skew on the builder machine.

cgwalters:
```
PASS: tests/test-admin-deploy-none.sh 21 no duplicate version strings in title
+++ env LD_PRELOAD=/home/jenkins/agent/workspace/ci_ostreedev_ostree_PR-2101-WEW6OONKDMYJBVECT3A3VL423YJK6UHLCQP2P777HY3QTQYSGOKA/tests/libreaddir-rand.so ostree rev-parse --repo=sysroot/ostree/repo testos/buildmaster/x86_64-runtime
++ head_rev=8f8e8eae4021f28891ef763dd8f1a3a9ec2662b55f78010ddb3891f4b9352ada
+++ env LD_PRELOAD=/home/jenkins/agent/workspace/ci_ostreedev_ostree_PR-2101-WEW6OONKDMYJBVECT3A3VL423YJK6UHLCQP2P777HY3QTQYSGOKA/tests/libreaddir-rand.so ostree rev-parse --repo=sysroot/ostree/repo 'testos/buildmaster/x86_64-runtime^^^^'
++ prev_rev=84ec817e489bf8a1e3db3c7f7232086319a892fca603a076b8efbe04ffe6333e
++ assert_not_streq 8f8e8eae4021f28891ef763dd8f1a3a9ec2662b55f78010ddb3891f4b9352ada 84ec817e489bf8a1e3db3c7f7232086319a892fca603a076b8efbe04ffe6333e
++ test 8f8e8eae4021f28891ef763dd8f1a3a9ec2662b55f78010ddb3891f4b9352ada = 84ec817e489bf8a1e3db3c7f7232086319a892fca603a076b8efbe04ffe6333e
++ env LD_PRELOAD=/home/jenkins/agent/workspace/ci_ostreedev_ostree_PR-2101-WEW6OONKDMYJBVECT3A3VL423YJK6UHLCQP2P777HY3QTQYSGOKA/tests/libreaddir-rand.so ostree admin upgrade --os=testos --override-commit=84ec817e489bf8a1e3db3c7f7232086319a892fca603a076b8efbe04ffe6333e
4 metadata, 0 content objects imported
Copying /etc changes: 0 modified, 0 removed, 0 added
Bootloader updated; bootconfig swap: yes; deployment count change: 0
++ fatal 'downgraded without --allow-downgrade?'
++ echo downgraded without '--allow-downgrade?'
downgraded without --allow-downgrade?
++ exit 1
+ run_exit_cmds
+ for expr in "${libtest_exit_cmds[@]}"
+ eval save_core
++ save_core
++ '[' -e core ']'
+ for expr in "${libtest_exit_cmds[@]}"
+ eval libtest_cleanup_gpg
++ libtest_cleanup_gpg
++ local gpg_homedir=/var/tmp/tap-test.vahbPW/gpghome
++ gpg-connect-agent --homedir /var/tmp/tap-test.vahbPW/gpghome killagent /bye
gpg-connect-agent: no running gpg-agent - starting '/usr/bin/gpg-agent'
gpg-connect-agent: waiting for the agent to come up ... (5s)
gpg-connect-agent: connection to agent established
Freed objects: 824 bytes
OK closing connection
ERROR: tests/test-admin-deploy-none.sh - too few tests run (expected 29, got 21)
ERROR: tests/test-admin-deploy-none.sh - exited with status 1
```

jlebon:
> Hum interesting, this PR changing unrelated code failed here.

Ahhh I think I know what this is. In the first diff hunk in that file (https://github.com/ostreedev/ostree/pull/2099/files#diff-5092e9dd89d398d2a7b46be8e0dbe787R307), I added a sanity-check that `ostree admin upgrade` with an override to an older commit failed without `--allow-downgrade`, but the test isn't strictly ensuring that the two commits are at least one second apart. Will fix!

jlebon:
https://github.com/ostreedev/ostree/pull/2106

PR #2100, opened 2020-05-14
ci: Install kola tests

cgwalters:
This builds on
https://github.com/coreos/coreos-assembler/pull/1441
to install our tests rather than running them from the source
directory.  This model will more cleanly allow us to ship
our tests along with a test container or elsewhere, separate
from the source directory.

Also prep for https://github.com/ostreedev/ostree/pull/2048

### DIFF
--- a/tests/kola/Makefile
+++ b/tests/kola/Makefile
@@ -1,0 +1,13 @@
+DESTDIR ?=
+
+TESTDIRS := $(shell find -mindepth 1 -maxdepth 1 -type d)
+LIBSCRIPTS := $(shell ls *.sh)
+
+KOLA_TESTDIR = $(DESTDIR)/usr/lib/coreos-assembler/tests/kola/ostree/
+
+all:
+	for x in $(LIBSCRIPTS); do bash -n "$${x}"; done
+
+install:
+	install -D -m 0644 -t $(KOLA_TESTDIR) $(LIBSCRIPTS)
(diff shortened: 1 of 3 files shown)


tests/kola/Makefile
```diff
@@ -0,0 +1,13 @@
+DESTDIR ?=
+
+TESTDIRS := destructive nondestructive
```
jlebon, line 3: Could just use e.g. `find -type d` here, no?

cgwalters:
To expand on this, there's a few directions we could go with this.

One I am leaning towards a bit is where we have a `quay.io/coreos/tests` container that does:
```
FROM quay.io/coreos-assembler:latest
RUN git clone https://github.com/ostreedev/ostree && cd ostree && make -C tests/kola install
RUN git clone https://github.com/coreos/ignition && cd ignition && make -C tests/kola install
```
etc.

There's already an `ostree-tests` RPM...but using it would be a bit awkward anyways because for cosa we need to support src != target.  Plus I'd like to support updating the tests *without* updating the content.

dustymabe:
I don't think I have an opinion here. Thanks for continuing to keep testing at a high priority.

tests/kola/Makefile
cgwalters: Mmm...yeah but we want to exclude `.`...I had to spend a bit thinking about a clean way to do that, but done.

PR #2101, opened 2020-05-17
pull: Only have API to disable signapi for local pulls

cgwalters:
There's a lot of historical baggage associated with GPG verification
and `ostree pull` versus `ostree pull-local`.  In particular nowadays,
if you use a `file://` remote things are transparently optimized
to e.g. use reflinks if available.

So for anyone who doesn't trust the "remote" repository, you should
really go through through the regular
`ostree remote add --sign-verify=X file://`
path for example.

Having a mechanism to say "turn on signapi verification" *without*
providing keys goes back into the "global state" debate I brought
up in https://github.com/ostreedev/ostree/issues/2080

It's just much cleaner architecturally if there is exactly one
path to find keys: from a remote config.

So here in contrast to the GPG code, for `pull-local` we explictily
disable signapi validation, and the `ostree_repo_pull()` API just
surfaces flags to disable it, not enable it.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3277,6 +3277,8 @@ initiate_request (OtPullData                 *pull_data,
  *   * override-remote-name (s): If local, add this remote to refspec
  *   * gpg-verify (b): GPG verify commits
  *   * gpg-verify-summary (b): GPG verify summary
+ *   * disable-sign-verify (b): Disable signapi verification of commits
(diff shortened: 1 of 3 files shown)


cgwalters:
Note we're still supporting the "global" keys found for ed25519/dummy for the main `ostree pull` path, this just discourages it more and also avoids carrying forward the GPG mistakes unnecessarily (I don't think anyone needs "pull local but verify" per above).

PR #2102, opened 2020-05-17
ci: Test for clock skew

cgwalters:
I saw `tests/test-admin-deploy.none.sh` fail in one CI run, and
I want to check if it was because of clock skew, so fail
fast if we detect that.

xref https://github.com/ostreedev/ostree/pull/2099#issuecomment-629805375

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -488,6 +488,11 @@ EOF
     cd ${oldpwd} 
 }
 
+timestamp_of_commit()
+{
+  date --date="$(ostree --repo=$1 show $2 | grep -Ee '^Date: ' | sed -e 's,^Date: *,,')" '+%s'
+}
+
 os_repository_new_commit ()
 {
     boot_checksum_iteration=${1:-0}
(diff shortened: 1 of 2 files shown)


cgwalters:
Also saw the same test fail in Travis in https://travis-ci.org/github/ostreedev/ostree/jobs/688423079

tests/libtest.sh
```diff
@@ -529,6 +534,13 @@ os_repository_new_commit ()
     echo "content iteration ${content_iteration}" > usr/bin/content-iteration
 
     ${CMD_PREFIX} ostree --repo=${test_tmpdir}/testos-repo commit  --add-metadata-string "version=${version}" -b $branch -s "Build"
+    if ${CMD_PREFIX} ostree --repo=${test_tmpdir}/testos-repo rev-parse ${branch} 2>/dev/null; then
+        prevdate=$(timestamp_of_commit ${test_tmpdir}/testos-repo "${branch}"^)
+        newdate=$(timestamp_of_commit ${test_tmpdir}/testos-repo "${branch}")
+        if [ $((${prevdate} > ${newdate})) = 1 ]; then
```
dustymabe, line 540: why not just `if [ $((${prevdate} > ${newdate})) ];` here? according to http://mywiki.wooledge.org/ArithmeticExpression we shouldn't need the `=1` comparison.

dustymabe: sorry my comment was late.. it was mostly for my understanding to see if we needed it or not.

PR #2104, opened 2020-05-18
remount: Also run Before=systemd-hwdb-update.service

cgwalters:
I saw a race with the readonly sysroot with RHEL CoreOS because
we hadn't mounted `/etc` writable by the time `systemd-hwdb-update.service`
ran.

### DIFF
--- a/src/boot/ostree-remount.service
+++ b/src/boot/ostree-remount.service
@@ -26,9 +26,10 @@ Conflicts=umount.target
 After=-.mount var.mount
 After=systemd-remount-fs.service
 # But we run *before* most other core bootup services that need write access to /etc and /var
+# In the future we should try to somehow replace/augment systemd-remount-fs.service
 Before=local-fs.target umount.target
 Before=systemd-random-seed.service plymouth-read-write.service systemd-journal-flush.service
-Before=systemd-tmpfiles-setup.service
+Before=systemd-tmpfiles-setup.service systemd-hwdb-update.service
 
 [Service]
 Type=oneshot


cgwalters:
One thing I'm not remembering here is why we don't just do all of `ostree-remount` in the initramfs.   Or at least, if the rootfs is already writable, there's no reason to wait.

cgwalters:
Oh right, it's because `ostree-prepare-root.c` doesn't link to glib today.  It's not at all a problem for those of us shipping an initramfs to do so, but is more of a problem for the no-initramfs statically-linked crowd.

Hmm.  I guess we could use a kernel argument, or honestly just don't use `GKeyFile` and just read the config line by line.

akiernan:
> It's not at all a problem for those of us shipping an initramfs to do so, but is more of a problem for the no-initramfs statically-linked crowd.

/me raises hand!

I guess my only concern with a kernel argument is ensuring it gets in there on upgrade, if folk are hardcoding their boot parameters.

jlebon:
Hmm, interesting, this is causing Kola failures. E.g.:

```
--- FAIL: ostree.remote (150.79s)
        harness.go:779: Cluster failed starting machines: machine "3bd322f2-eaff-4351-8594-c8020baa67f3" failed basic checks: some systemd units failed:
systemd-udev-settle.service loaded failed failed udev Wait for Complete Device Initialization
```

cgwalters:
Obsoleted by https://github.com/ostreedev/ostree/pull/2113

PR #2106, opened 2020-05-20
tests/admin-test: Ensure that commits are 1s apart

jlebon:
Otherwise the new check we added there to verify that upgrading without
`--allow-downgrade` fails itself fails.

See: https://github.com/ostreedev/ostree/pull/2099#issuecomment-629805840

### DIFF
--- a/tests/admin-test.sh
+++ b/tests/admin-test.sh
@@ -300,6 +300,8 @@ echo "ok no duplicate version strings in title"
 
 # Test upgrade with and without --override-commit
 # See https://github.com/GNOME/ostree/pull/147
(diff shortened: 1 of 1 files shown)


cgwalters:
OK I had missed that you added a new test that raced.  I kept thinking this was somehow related to the code.  

However, we still have a mysterious failure there (and why is Travis reproducing this so much more reliably?  I still suspect some sort of clock skew).  

Filed https://github.com/ostreedev/ostree/pull/2107

cgwalters:
OK this clearly can't break anything and I want to get both of these in so we can debug more

PR #2108, opened 2020-05-20
switchroot/remount: Neuter sysroot.readonly for now

jlebon:
We're hitting issues with the read-only remounts racing with various
services coming up. Let's neuter it for now until we rework how it
works.

See: https://github.com/coreos/fedora-coreos-tracker/issues/488


jlebon:
Known outstanding Travis CI failure.

PR #2109, opened 2020-05-20
admin-test: Add set -x

cgwalters:
Still seeing this fail in CI and this should help debug.


cgwalters:
OK wow I think I understand now; Travis apparently handles stdout/stderr separately, and because stdout is buffered when we hit an error it's *way* up there in the logs after the rest of the buffered stdout.  Viewing the raw log is much more sensible.

jlebon:
OK, I think I *finally* got to the bottom of this here: https://github.com/ostreedev/ostree/pull/2110.

PR #2110, opened 2020-05-21
tests/admin-test: Fix --allow-downgrade check

jlebon:
We were doing a check to verify that `ostree admin upgrade` wouldn't
accept a downgrade without `--allow-downgrade`. However, there's no
guarantee that the commit it's upgrading from is older than HEAD^ (what
we're upgrading to). Specifically, if the test runs fast enough, the
timestamps could be equal, since the lowest resolution is seconds.

Rework the test so that we first upgrade to HEAD, which we're sure is at
least 1 second apart from HEAD^, and *then* check that downgrade
protection is enforced.

We also can't use `rev-parse testos/buildmaster/x86_64-runtime` as a way
to know what commit the host is sitting on since the ref might've gone
ahead. Instead, just use `ostree admin status | head -n1`. (I played
with using the `ostree/I/J/K` refs, but those depend on what the
boot/subbootversion is and can easily change if we change previous
tests).

### DIFF
--- a/tests/admin-test.sh
+++ b/tests/admin-test.sh
@@ -302,23 +302,31 @@ echo "ok no duplicate version strings in title"
 # See https://github.com/GNOME/ostree/pull/147
 sleep 1
 os_repository_new_commit
-${CMD_PREFIX} ostree pull --repo=sysroot/ostree/repo --commit-metadata-only --depth=-1 testos:testos/buildmaster/x86_64-runtime
+# upgrade to the latest
+${CMD_PREFIX} ostree admin upgrade --os=testos
 head_rev=$(${CMD_PREFIX} ostree rev-parse --repo=sysroot/ostree/repo testos/buildmaster/x86_64-runtime)
-prev_rev=$(${CMD_PREFIX} ostree rev-parse --repo=sysroot/ostree/repo testos/buildmaster/x86_64-runtime^^^^)
+prev_rev=$(${CMD_PREFIX} ostree rev-parse --repo=sysroot/ostree/repo testos/buildmaster/x86_64-runtime^)
 assert_not_streq ${head_rev} ${prev_rev}
-# check that we can't "upgrade" to an older commit without --allow-downgrade
+# Don't use `ostree admin status | head -n 1` directly here because `head`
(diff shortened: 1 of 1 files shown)


jlebon:
>(I played with using the ostree/I/J/K refs, but those depend on what the boot/subbootversion is and can easily change if we change previous tests).

Let me know if I'm missing a quick and easy way to do this.

jlebon:
/hold cancel

OK, there is something really weird in Travis CI going on here. But I think it should work now.

From test comment:

```
# For reasons I cannot explain, one of the Debian Travis tests doesn't like it
# when using `ostree admin status | head -n 1 > status.txt` directly. So split
# it out to two separate steps. XXX: should investigate this further.
```

We should probably dig into this because it seems like it's consistently happening only in the Travis CI branch configured with `ci_configopts="--with-curl --with-ed25519-libsodium --without-gpgme"`, whereas `ci_configopts="--with-ed25519-libsodium"` works fine. So something subtle going on here (code in `ostree admin status` not expecting a pipe in that code path maybe?). Anyway, ran out of cycles on this for now.

cgwalters:
It might have something to do with the ambient `SIGPIPE` handling.  Hmm...yes a quick ~professional software engineer investigation~ google search for "travis sigpipe" turns this up: https://github.com/ldc-developers/ldc/issues/2351#issue-261386046

jlebon:
Ahh yup, I think you're right, good catch! So basically ostree's writes to the pipe buffer is racing with `head -n1` processing it and exiting. If it exits before ostree is done writing to the buffer, ostree gets a `SIGPIPE`, and because we have `set -euo pipefail`, ostree's exit code propagates up and causes us to exit.

Anyway, for now I've just clarified the comment and just added a linkback to this PR for details.

cgwalters:
Something in the stack resets `SIGPIPE` to `SIG_IGN` I think.  And we should probably not fail if `g_print()` fails.  Anyways, messy topic.  Thanks a ton for chasing this down!
/approve
/lgtm

cgwalters:
Or to quote René Magritte: Ceci n'est pas une SIGPIPE

PR #2113, opened 2020-05-24
Move ro /sysroot bind mount of /etc into initramfs

cgwalters:
We recently disabled the read-only /sysroot handling:
https://github.com/ostreedev/ostree/pull/2108/commits/e35b82fb891daee823fcce421ae8f1442b630ea2

The core problem was that a lot of services run early in the
real root and want write access to things like `/var` and `/etc`.

In trying to do remounts while the system is running we introduce
too many race conditions.

Instead, just make the `/etc` bind mount in the initramfs right
after we set up the main root.  This is much more natural really,
and avoids all race conditions since nothing is running in the
sysroot yet.

The main awkward part is that since we're not linking
`ostree-prepare-root` to GLib (yet) we have a hacky parser
for the config file.  But, this is going to be fine I think.

In order to avoid parsing the config twice, pass state from
`ostree-prepare-root` to `ostree-remount` via a file in `/run`.

### DIFF
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -81,24 +81,6 @@ do_remount (const char *target,
   printf ("Remounted %s: %s\n", writable ? "rw" : "ro", target);
 }
 
-static bool
-sysroot_is_configured_ro (void)
-{
-  struct stat stbuf;
-  static const char config_path[] = "/ostree/repo/config";
-  if (stat (config_path, &stbuf) != 0)
-    return false;
-
-  g_autoptr(GKeyFile) keyfile = g_key_file_new ();
-  if (!g_key_file_load_from_file (keyfile, config_path, 0, NULL))
-    return false;
-
-  if (g_key_file_get_boolean (keyfile, "sysroot", "readonly", NULL))
-    puts ("Ignoring sysroot.readonly config; see https://github.com/coreos/fedora-coreos-tracker/issues/488.");
-
-  return false;
-}
-
 int
 main(int argc, char *argv[])
 {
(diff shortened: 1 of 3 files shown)


src/switchroot/ostree-remount.c
```diff
@@ -124,25 +106,10 @@ main(int argc, char *argv[])
...
+  if (unlink (_OSTREE_SYSROOT_READONLY_STAMP) == 0)
     {
-      /* Now, /etc is not normally a bind mount, but remounting the
-       * sysroot above made it read-only since it's on the same filesystem.
-       * Make it a self-bind mount, so we can then mount it read-write.
-       */
-      if (mount ("/etc", "/etc", NULL, MS_BIND, NULL) < 0)
-        err (EXIT_FAILURE, "failed to make /etc a bind mount");
-      do_remount ("/etc", true);
+      do_remount ("/sysroot", false);
```
jlebon, line 112: Couldn't we ensure that the sysroot is mounted read-only from `ostree-prepare-root`? That way there's no gap either in the real root as for when `sysroot.readonly` is enforced.

cgwalters: That'd make a lot of sense, but I'd prefer that as a followup cleanup because it affects a whole lot more of the code paths, mostly around handling of `/var`.  And thing like the `/run/ostree/initramfs-mount-var` path versus the fstab generator etc.

cgwalters:
I should note I tested this on FCOS (and so is CI) - we get a correctly `ro` sysroot again.

And I also tested on a FSB32 system by appending the binary to the current initramfs (and verified we have the new journal message):
`May 26 08:32:39 localhost ostree-prepare-root[682]: sysroot configured read-only: 0, currently writable: 0`

src/switchroot/ostree-remount.c
jlebon: Right yeah. OK filed https://github.com/ostreedev/ostree/issues/2115 as a follow up.

cgwalters:
Followup in https://github.com/ostreedev/ostree/pull/2160

PR #2116, opened 2020-05-27
tests/kola: Move to tests/kolainst

cgwalters:
Follow the precedent set in https://github.com/coreos/rpm-ostree/pull/2106
and rename the directory, to more clearly move away from the
"uninstalled" test model.  Prep for Rust-based tests.


jlebon:
```
make[3]: *** No rule to make target 'tests/libtest-core.sh'.  Stop.
```

Hmm, not sure what's going on here.

cgwalters:
Needed to fix the symlink; the unit tests share this bit.

PR #2121, opened 2020-06-03
core: Add OSTREE_COMMIT_META_KEY_ARCH

cgwalters:
Add a standard key for this.  We actually had a case in OpenShift
builds recently where a ppc64le image was pushed over an `x86_64`
one and this started failing at runtime with a not immediately
obvious error.

I'll probably end up changing rpm-ostree at least to use
the RPM architecture for this key and fail if it doesn't match
the booted value.

Possibly that should live in ostree but it would involve adding
architecture schema here, which gets into a big mess.  Let's
just standardize the key.

### DIFF
--- a/src/libostree/ostree-core.h
+++ b/src/libostree/ostree-core.h
@@ -212,6 +212,19 @@ typedef enum {
  * Since: 2014.9
  */
 #define OSTREE_COMMIT_META_KEY_VERSION "version"
+
+/**
+ * OSTREE_COMMIT_META_KEY_ARCHITECTURE:
+ *
+ * GVariant type `s`.  Intended to describe the CPU architecture.  This is a freeform string, and some distributions
+ * which have existing package managers might want to match that schema.  If you
+ * don't have a prior schema, it's recommended to use `uname -m` by default (i.e. the Linux kernel schema).  In the future
+ * ostree might include a builtin function to compare architectures.
+ *
+ * Since: 2020.4
+ */
+#define OSTREE_COMMIT_META_KEY_ARCHITECTURE "ostree.architecture"
+
 /**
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-core.h
```diff
@@ -212,6 +212,19 @@ typedef enum {
...
+ * OSTREE_COMMIT_META_KEY_ARCH:
+ *
+ * GVariant type `s`. Architecture; this is a freeform string, and some distributions
+ * which have existing package managers might want to match that schema.  It's recommended
+ * to use e.g. `uname -m` by default (i.e. the Linux kernel schema).  In the future
+ * ostree might include a builtin function to compare architectures.
+ *
+ * Since: 2020.4
+ */
+#define OSTREE_COMMIT_META_KEY_ARCH "ostree.arch"
```
jlebon, line 226: Small bikeshed to just spell this out as `OSTREE_COMMIT_META_KEY_ARCHITECTURE "ostree.architecture"`, though good as is too.

cgwalters: > doesn't that theoretically break going from a 32-bit to 64-bit kernel? 

Right, so far we haven't hit that problem for any system we care about that I know.  e.g. no one sane runs `i686` kernels anymore, and for ARM in Fedora it's just `aarch64`.  But I guess if we did do this in rpm-ostree we'd probably want to use the libdnf architecture mapping for compatible types to do it "properly".

cgwalters: > I'm not sure if 32-bit kernels would still show x86_64 on a 64-bit capable CPU even if it was compiled for 32-bit. 

No, 32 bit kernels report as `i686`.

cgwalters:
https://github.com/coreos/fedora-coreos-config/pull/457

PR #2126, opened 2020-06-08
Improve checkout man page

agners:
This merge request adds all missing options to the checkout man page. There is actually one mistake which really confused me (`-u` for `--user-mode`) which I kept in a separate commit.

### DIFF
--- a/man/ostree-checkout.xml
+++ b/man/ostree-checkout.xml
@@ -109,6 +109,14 @@ Boston, MA 02111-1307, USA.
             </varlistentry>
 
             <varlistentry>
+                <term><option>--whiteouts</option></term>
+
+                <listitem><para>
+                    Process whiteout files (Docker style).
+                </para></listitem>
+            </varlistentry>
+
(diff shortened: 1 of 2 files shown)


man/ostree-checkout.xml
```diff
@@ -131,6 +139,68 @@ Boston, MA 02111-1307, USA.
...
+                    Suppress mode bits outside of 0775 for directories (suid,
+                    world writable, etc.).
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>--skip-list</option>="FILE"</term>
+
+                <listitem><para>
+                    File containing list of files to skip.
```
jlebon, line 192: Maybe

```suggestion
                    Skip checking out the absolute file paths listed in FILE, one per line.
```

?

agners: Sounds resonable, updated!

jlebon:
Thanks!

/lgtm

PR #2127, opened 2020-06-08
tests/inst: Add destructive test framework

cgwalters:
This adds infrastructure to the Rust test suite for destructive
tests, and adds a new `transactionality` test which runs
rpm-ostree in a loop (along with `ostree-finalize-staged`) and
repeatedly kills them.

The main goal here is to flush out any "logic errors".  I plan
to further extend this to reboots and then force poweroffs.


cgwalters:
OK a lot more work here; we're testing that we reliably survive forced poweroffs.  Still TODO:

- Re-merge the "kill -9" code and alternate between multiple "interrupt strategies" of "none, kill -9, reboot -ff" etc.
- Test that we *successfully fail* if we e.g. omit `fsync()`

cgwalters:
OK lifting WIP on this - fault injection would be another level, and I need to solve some other problems before doing that like being able to pull containers/binaries/packages from the "host" cosa container at least in qemu so we don't rely on internet access (as I really want to be able to run this test in a loop and not have it randomly flake for internet reasons).

tests/inst/src/destructive.rs
```diff
@@ -0,0 +1,580 @@
...
+            ",
+            testref = TESTREF,
+            booted_commit = booted_commit
+        )
+        .context("Failed pre-upgrade cleanup")?;
+        let res: Result<bool> = rt.block_on(async move { run_upgrade_or_timeout(sleeptime).await });
+        let res = res.context("Failed during upgrade")?;
+        if res {
+            println!(
+                "Failed to interrupt upgrade, attempt {}/{}",
```
jlebon, line 436: We could lower the timeout by e.g. 10% in this case to increase the odds for the next time (maybe after we get to 5 retries or something).

tests/inst/src/destructive.rs
```diff
@@ -0,0 +1,580 @@
...
+
+/// The data passed across reboots by serializing
+/// into the AUTOPKGTEST_REBOOT_MARK
+#[derive(Serialize, Deserialize, Debug, Default)]
+#[serde(rename_all = "kebab-case")]
+struct RebootMark {
+    // forced will be true if the reboot was `-ff`
+    reboot_strategy: Option<InterruptStrategy>,
+    iter: u32,
+    before: u32,
```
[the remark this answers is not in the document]
cgwalters: Yeah it's just informational.

tests/inst/src/destructive.rs
```diff
@@ -0,0 +1,580 @@
...
+        )?;
+
+        if firstrun {
+            // Also disable some services (like zincati) because we don't want automatic updates
+            // in our reboots, and it currently fails to start.  The less
+            // we have in each reboot, the faster reboots are.
+            sh_execute!("systemctl disable --now zincati fedora-coreos-pinger")?;
+            // And prepare for updates
+            sh_execute!("rpm-ostree cleanup -pr")?;
+            generate_update(&commit)?;
```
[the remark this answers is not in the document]
cgwalters: Yeah, but then the rest of the code needs to handle the special case of having no "previous" commit for the upgrade target ref.

cgwalters:
OK ended up doing more fixes and tweaks here; I noticed that the results weren't including the "no-interrupt" case, and fixing/handling that required another tricky special case.

I also noticed there were fewer "completed" results for live interrupts than I expected, and that turned out to be me forgetting it needs to be `systemctl stop ostree-finalize-staged` and not `start`.

tests/inst/src/destructive.rs
cgwalters: I thought about doing things like this - an issue I've seen is that when I start up 4 instances of this test in parallel, there's heavy CPU usage in the VMs where they're all doing the initial setup, and that inflates the timing for the test upgrade.

The better fix I think would be something like an `env OSTREE_PAUSE_POINT=pre-deploy,post-deploy,pre-cleanup` where the harness can control when the process continues.  (Or maybe implement this with scripting gdb or so)

tests/inst/src/destructive.rs
```diff
@@ -0,0 +1,630 @@
...
+        Ok(())
+    })?;
+    Ok(())
+}
+
+#[itest(destructive = true)]
+fn other() -> Result<()> {
+    testinit()?;
+    Ok(())
+}
```
jlebon, line 630: Is this test leftover from testing or is it meaningful? Otherwise, let's save the VM. :)

cgwalters: Thanks, removed :recycle:

PR #2128, opened 2020-06-11
tests: Add a pre-signed-pull.sh test

cgwalters:
I'm thinking about adding an implementation of ed25519 signatures
with OpenSSL (so we can ship the feature with Fedora CoreOS
without requiring an additional library) and in preparation for
that it's essential that we validate that libsodium-generated
signatures and OpenSSL-generated signatures are compatible.

I don't know if they are yet actually, but the goal of this
new test is to add a pre-generated repository with a signed
commit generated by libsodium.

This will catch if e.g. there's ever a change in libsodium,
or if existing libsodium implementation versions (e.g. the
one in Debian) might differ from what we ship here.


cgwalters:
This test isn't about hardcoding the key - it's about hardcoding the *signature* including the `.commitmeta`, and really the full repository which we treat as a black box and aren't generating at test time.

Having other tests use the same key feels like it's of pretty small value but I'm not opposed to splitting it out either.

d4s:
@cgwalters yes, I'm fully understood the purpose.
As I said -- I'm not sure for 100% if the key split is very useful -- just a little improvement. Do I need to wait the change or approve as is? I'm happy with both

cgwalters:
I think let's get this in if that's OK with you!

PR #2129, opened 2020-06-16
sign/ed25519: Output failed signatures in error message

cgwalters:
To aid debuggability, when we find a commit that isn't signed
by our expected key, output a specific error message with the
key.

(And then add code to switch to just printing the count beyond 3
 because the test suite injects 100 keys and hopefully no one
 ever actually does that)

### DIFF
--- a/src/libostree/ostree-sign-ed25519.c
+++ b/src/libostree/ostree-sign-ed25519.c
@@ -202,6 +202,9 @@ gboolean ostree_sign_ed25519_data_verify (OstreeSign *self,
 
   g_debug ("verify: data hash = 0x%x", g_bytes_hash(data));
 
+  g_autoptr(GString) invalid_signatures = NULL;
+  guint n_invalid_signatures = 0;
+
   for (gsize i = 0; i < g_variant_n_children(signatures); i++)
     {
       g_autoptr (GVariant) child = g_variant_get_child_value (signatures, i);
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-sign-ed25519.c
```diff
@@ -242,7 +250,17 @@ gboolean ostree_sign_ed25519_data_verify (OstreeSign *self,
...
 
-  return glnx_throw (error, "no valid ed25519 signatures found");
+  if (invalid_signatures)
+    {
+      g_assert_cmpuint (n_invalid_signatures, >, 0);
+      /* The test suite has a key ring with 100 keys.  This seems insane, let's
+       * cap a reasonable error message at 3.
+       */
+      if (n_invalid_signatures > 3)
+        return glnx_throw (error, "ed25519: Failed to validate, tried %u keys", n_invalid_signatures);
```
jlebon, line 260: How about

```suggestion
        return glnx_throw (error, "ed25519: Signature couldn't be verified; tried %u keys", n_invalid_signatures);
```

?

To make it more consistent with the `n<=3` case.

cgwalters: I merged that and tweaked things further so we are less redundant in the multiple key case.

d4s:
/lgtm
fixed one test

cgwalters:
> fixed one test

:+1: thanks!

d4s:
Hm... did I broke travis?

PR #2131, opened 2020-06-17
signapi: Change API to also return a success message

cgwalters:
This is the dual of https://github.com/ostreedev/ostree/pull/2129/commits/1f3c8c5b3de978f6e069c24938967f823cce7ee8
where we output more detail when signapi fails to validate.

Extend the API to return a string for success, which we output
to stdout.

This will help the test suite *and* end users validate that the expected
thing is happening.

In order to make this cleaner, split the "verified commit" set
in the pull code into GPG and signapi verified sets, and have
the signapi verified set contain the verification string.

We're not doing anything with the verification string in the
pull code *yet* but I plan to add something like
`ostree pull --verbose` which would finally print this.


cgwalters:
@d4s any thoughts on this?

d4s:
sorry, somehow missed that PR
/approve
/lgtm

cgwalters:
I queued a retest (unfortunately manual at the moment)

PR #2133, opened 2020-06-17
ci: Import latest ci-commitmessage-submodules from rpm-ostree

jlebon:
Especially for https://github.com/coreos/rpm-ostree/pull/2079.


cgwalters:
Ah, some more drift:
> ci/ci-commitmessage-submodules.sh: line 5: ci/libbuild.sh: No such file or directory

PR #2134, opened 2020-06-17
Fix abort on verify

d4s:
Small fixes for case if file with public keys doesn't contain correctly formatted keys.
Allow to return the correct error instead of aborting.


src/libostree/ostree-sign-ed25519.c
```diff
@@ -546,6 +546,8 @@ _load_pk_from_file (OstreeSign *self,
         return glnx_throw (error, 
                            "signature: ed25519: no valid keys in file '%s'",
                            filename);
+      else
+        return FALSE;
```
cgwalters, line 550: Well...there's a larger rework that should take place here.  But yes this is right for now.

PR #2132, opened 2020-06-17
libostree-devel.sym: Remove nonexistent stub symbol

cgwalters:
This should have been removed when we added symbols to this list.

PR #2135, opened 2020-06-17
tests: Check that example symbol isn't released

mwleeds:
For the motivation for this see #2132.


mwleeds:
Had to fix some silly mistakes. Sorry, I'll test locally next time. I did check locally that the new library function `assert_file_has_content_once` works as advertised.

cgwalters:
Thank you!
/approve
/lgtm

PR #2137, opened 2020-06-18
sysroot: Remove unimplemented ostree_sysroot_lock_with_mount_namespace

cgwalters:
This came in with 5af403be0cc64df50ad21cef05f3268ead256d6d but
was never implemented.

I noticed this now because the Rust ostree bindings generate a
wrapper for it which the linker tries to use.


cgwalters:
```
walters@toolbox ~/s/g/o/ostree (master)> rg ostree_sysroot_lock_with_mount
src/libostree/ostree-sysroot.h
105:gboolean ostree_sysroot_lock_with_mount_namespace (OstreeSysroot  *self, GError **error);
walters@toolbox ~/s/g/o/ostree (master)> 
```

PR #2143, opened 2020-06-30
lib/repo: Handle EACCES for POSIX locking

jlebon:
If `glnx_make_lock_file` falls back to `flock`, on NFS this uses POSIX
locks (`F_SETLK`). As such, we need to be able to handle `EACCES` as
well as `EAGAIN` (see `fnctl(2)`).

I think this is what coreos-ostree-importer has been hitting, which runs
on RHEL7 in the Fedora infra and does locking over an NFS share where
multiple apps could concurrently pull things into the repo.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1933,7 +1933,7 @@ rename_pending_loose_objects (OstreeRepo        *self,
   return TRUE;
 }
 
-/* Try to lock a transaction stage directory created by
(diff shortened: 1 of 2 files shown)


jlebon:
@dustymabe FYI, this *should* fix the importer error that happens when it runs at the same time as the rawhide sync job like:

```
error: tmpdir cleanup: Cleaning up txn dir staging-cad29946-95e8-42f8-8ae0-8920cc82f8e4-fU2hNK: Permission denied
```

PR #2144, opened 2020-07-03
lib/deltas: convert ostree_repo_static_delta_generate to new style

cgwalters:
The "new style" code generally avoids `goto err` because it conflicts
with `__attribute__((cleanup))`.  This fixes a compiler warning.

Signed-off-by: Frédéric Danis <frederic.danis@collabora.com>


cgwalters:
Split from https://github.com/ostreedev/ostree/pull/1985/

PR #2145, opened 2020-07-08
sysroot: Support customising /boot on root for syslinux and u-boot

wmanley:
This introduces a new config value "sysroot.boot_path_on_disk":

> Tell ostree where the bootloader will be looking for the kernel.  This is
> the location of `/boot` relative to the root of the boot partition.
>
> This allows explicitly telling ostree that the boot partition is or is
> not separate from the root partition.  If `/boot` is on root set this to
> `/boot`.  If it's on seperate partition this should be set to `/`.
>
> This must be either an absolute path (starting with `/`) or `auto`.
> Defaults to `auto`.
>
> This setting currently only respected by the u-boot and syslinux
> bootloaders.

This commit is based on an original by Colin Walters (#215), but takes a
different tack.  #215 attempts to determine automatically whether `/boot`
is on root or not, but there are always situations where this isn't
possible so I'm adding a config option here so the user can make it
explicit.

Future commits can add the automatic detection in, but this means that we
don't need to be completely general with the auto-detection.

See also some background here: https://github.com/ostreedev/ostree/pull/1404#issuecomment-654169968

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -372,6 +372,33 @@ Boston, MA 02111-1307, USA.
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>boot_path_on_disk</varname></term>
+        <listitem>
+          <para>
+            Tell ostree where the bootloader will be looking for the kernel.
+            This is the location of <literal>/boot</literal> relative to the
+            root of the boot partition.
+          </para>
+          <para>
+            This allows explicitly telling ostree that the boot partition is or
+            is not seperate from the root partition.  If
+            <literal>/boot</literal> is on root set this to
+            <literal>/boot</literal>.  If it's on seperate partition this
+            should be set to <literal>/</literal>.
+          </para>
+          <para>
+            This must be either an absolute path (starting with
+            <literal>/</literal>) or <literal>auto</literal>.  Defaults to
+            <literal>auto</literal>.
+          </para>
+          <para>
+            This setting currently only respected by the u-boot and syslinux
+            bootloaders.
+          </para>
+        </listitem>
+      </varlistentry>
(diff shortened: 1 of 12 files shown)


wmanley:
Actually, nevermind this I've got a better solution...

cgwalters:
I know this is a lot more code but I do like the explicit-ness of this.
/approve

man/ostree.repo-config.xml
```diff
@@ -372,6 +372,33 @@ Boston, MA 02111-1307, USA.
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>boot_path_on_disk</varname></term>
```
cgwalters, line 376: Maybe just `boot_path`?

src/libostree/ostree-sysroot.c
```diff
@@ -2140,3 +2140,40 @@ ostree_sysroot_deployment_set_pinned (OstreeSysroot     *self,
...
+  if (g_str_equal (boot_path_on_disk, "auto"))
+    {
+      /* TODO: Be smarter about working this out automatically */
+      g_free (boot_path_on_disk);
+      boot_path_on_disk = g_strdup ("/");
+    }
+
+  if (!boot_path_on_disk || boot_path_on_disk[0] != '/')
+    {
+      g_error_new (G_IO_ERROR, G_IO_ERROR_FAILED, "sysroot.boot_path_on_disk "
```
cgwalters, line 2173: I'd do:
```
return glnx_null_throw ("sysroot.boot_path is not absolute: %s", boot_path);
```

cgwalters:
Closing this since https://github.com/ostreedev/ostree/pull/2149 merged.
Thanks for all the work on this!

PR #2147, opened 2020-07-09
pull: Add OSTREE_PULL_MAX_OUTSTANDING_WRITES env var

cgwalters:
We need to expose more control over how fast we write to disk
in order to allow OS vendors to more finely tune latency vs
throughput for OS updates.

See https://github.com/openshift/machine-config-operator/issues/1897

This is a crude hack I'm using for testing; need to also
expose this via API, document it etc.


jlebon:
Any preliminary results on how effective this is? Feels like the better thing to throttle is size of outstanding objects instead of number of objects, though that's a bigger patch. If it unblocks https://github.com/openshift/machine-config-operator/issues/1897 though, WFM! (Could probably make it a repo config knob instead/in addition?).

cgwalters:
Moved this to https://github.com/ostreedev/ostree/pull/2152

Issue #2125, opened 2020-06-05
Support for deploying sysroot tree without writing deployment in /boot


mildred:
In some cases, it can be desirable to have a sysroot managed by ostree, but never boot to it, such as chroots or systemd-machined systems running on top of a host.

For those cases, it would be great to be able to `ostree admin deploy` without actually configuring the boot loader.

It happens in :

https://github.com/ostreedev/ostree/blob/8996ead2ea7422903ee4d4a4b4024c44c866a7fe/src/ostree/ot-admin-builtin-deploy.c#L204

The call to `ostree_sysroot_simple_write_deployment` should not happen in those cases.

The workaround I found for the moment is to use `--retain --not-as-default` as to avoid changing the host system boot entries. But not having the guest system register in the bootloader would be even better.

agners:
Hm, so this would need to be a per deployment option. Currently there is `sysroot.bootloader`, so this is a sysroot wide option. I guess `sysroot.bootloader` could be treated as the default, which can be overwritten by a deployment option.

cgwalters:
You can just use `ostree checkout` right?

agners:
@cgwalters I guess `/etc` handling would still be missing for this use-case.

mildred:
`--retain --not-as-default` would be sufficient if only rpm-ostree allowed to pass those options when updating a deployment. Also, doesn't ostree read files in /boot/loader/entries to know which deployments are active ? In that case, not configuring the bootloader is not really an option.

cgwalters:
> For those cases, it would be great to be able to ostree admin deploy without actually configuring the boot loader.

... 

> Also, doesn't ostree read files in /boot/loader/entries to know which deployments are active ? In that case, not configuring the bootloader is not really an option.

You are asking for conflicting things. 

Basically, the `ostree admin` portion of ostree (aka `OstreeSysroot`) does have an opinionated model for how to manage bootable filesystem trees and the bootloader config.

You can totally use ostree for non-bootable trees (containers etc.) but in that scenario your higher level code owns generating and garbage collecting them, not ostree.

(For example, you really want something like locking around not pruning a deployment root when a container is using it...which should definitely not live in ostree itself but in your container tooling)

PR #2149, opened 2020-07-15
sysroot: Support /boot on root or as seperate filesystem for syslinux and u-boot

### LINKED ISSUE #1452: support /boot in the root filesystem; i.e. no separate /boot
Whenever someone does manual partitioning and they don't add a separate `/boot` partition. They end up getting an error in anaconda: see [this BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1497094).

I looked and I couldn't find a specific issue that addressed this but there is https://github.com/ostreedev/ostree/issues/876 which seems somewhat related but mostly focusing on atomicity of the filesystem I think.

Can we support `/boot` in the root filesystem? Of course, for efi systems there will always be the vFAT `/boot/efi`.

Note: This may be an issue that needs code work in Anaconda instead of in ostree.

wmanley:
We use a similar trick to having a `sysroot -> .` symlink on the real root
here to support both /boot on root as well as on a separate filesystem.  No
matter how it's mounted `/boot/xyz` will always refer to the file you'd
expect.

This is nicer than my previous attempts at this because there's no
configuration nor auto-detection required.

Fixes #1452

Replaces #1404, #1914 and #2145

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1998,6 +1998,12 @@ prepare_new_bootloader_link (OstreeSysroot  *sysroot,
   g_assert ((current_bootversion == 0 && new_bootversion == 1) ||
             (current_bootversion == 1 && new_bootversion == 0));
 
+  /* This allows us to support both /boot on a seperate filesystem to / as well
+   * as on the same filesystem. */
+  if (TEMP_FAILURE_RETRY (symlinkat (".", sysroot->sysroot_fd, "boot/boot")) < 0)
+    if (errno != EEXIST)
+      return glnx_throw_errno_prefix (error, "symlinkat");
+
   g_autofree char *new_target = g_strdup_printf ("loader.%d", new_bootversion);
 
   /* We shouldn't actually need to replace but it's easier to reuse
--- a/src/libostree/ostree-bootloader-syslinux.c
+++ b/src/libostree/ostree-bootloader-syslinux.c
@@ -89,15 +89,15 @@ append_config_from_loader_entries (OstreeBootloaderSyslinux  *self,
       val = ostree_bootconfig_parser_get (config, "linux");
       if (!val)
         return glnx_throw (error, "No \"linux\" key in bootloader config");
-      g_ptr_array_add (new_lines, g_strdup_printf ("\tKERNEL %s", val));
+      g_ptr_array_add (new_lines, g_strdup_printf ("\tKERNEL /boot%s", val));
 
       val = ostree_bootconfig_parser_get (config, "initrd");
       if (val)
-        g_ptr_array_add (new_lines, g_strdup_printf ("\tINITRD %s", val));
+        g_ptr_array_add (new_lines, g_strdup_printf ("\tINITRD /boot%s", val));
 
       val = ostree_bootconfig_parser_get (config, "devicetree");
       if (val)
-        g_ptr_array_add (new_lines, g_strdup_printf ("\tDEVICETREE %s", val));
+        g_ptr_array_add (new_lines, g_strdup_printf ("\tDEVICETREE /boot%s", val));
 
       val = ostree_bootconfig_parser_get (config, "options");
       if (val)
@@ -150,10 +150,13 @@ _ostree_bootloader_syslinux_write_config (OstreeBootloader  *bootloader,
           if (kernel_arg == NULL)
             return glnx_throw (error, "No KERNEL argument found after LABEL");
 
-          /* If this is a non-ostree kernel, just emit the lines
-           * we saw.
+          /* If this is a non-ostree kernel, just emit the lines we saw.
+           *
+           * We check for /ostree (without /boot prefix) as well to support
+           * upgrading ostree from <v2020.4.
            */
-          if (!g_str_has_prefix (kernel_arg, "/ostree/"))
+          if (!g_str_has_prefix (kernel_arg, "/ostree/") &&
+              !g_str_has_prefix (kernel_arg, "/boot/ostree/"))
             {
               for (guint i = 0; i < tmp_lines->len; i++)
                 {
--- a/src/libostree/ostree-bootloader-uboot.c
+++ b/src/libostree/ostree-bootloader-uboot.c
@@ -134,19 +134,19 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot     *self,
                        "No \"linux\" key in bootloader config");
           return FALSE;
         }
-      g_ptr_array_add (new_lines, g_strdup_printf ("kernel_image%s=%s", index_suffix, val));
+      g_ptr_array_add (new_lines, g_strdup_printf ("kernel_image%s=/boot%s", index_suffix, val));
 
       val = ostree_bootconfig_parser_get (config, "initrd");
       if (val)
-        g_ptr_array_add (new_lines, g_strdup_printf ("ramdisk_image%s=%s", index_suffix, val));
+        g_ptr_array_add (new_lines, g_strdup_printf ("ramdisk_image%s=/boot%s", index_suffix, val));
 
       val = ostree_bootconfig_parser_get (config, "devicetree");
       if (val)
-        g_ptr_array_add (new_lines, g_strdup_printf ("fdt_file%s=%s", index_suffix, val));
+        g_ptr_array_add (new_lines, g_strdup_printf ("fdt_file%s=/boot%s", index_suffix, val));
 
       val = ostree_bootconfig_parser_get (config, "fdtdir");
       if (val)
-        g_ptr_array_add (new_lines, g_strdup_printf ("fdtdir%s=%s", index_suffix, val));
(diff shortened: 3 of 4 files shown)


wmanley:
jenkins ci is failing with the same issue that is currently plaguing master:

> fork: retry: Resource temporarily unavailable

but I consider this ready for review. If this goes in I'll be glad to close the PRs with the other attempts :).

I've only tested this on real systems with /boot on root and in syslinux mode.  Our systems have u-boot, but they read the syslinux config.

vtolstov:
"\yeah, i'm really need this too as i have /boot on root

wmanley:
> jenkins ci is failing with the same issue that is currently plaguing master:

CI is passing now, presumably thanks to #2151.  Thanks @jlebon and @cgwalters.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1998,6 +1998,12 @@ prepare_new_bootloader_link (OstreeSysroot  *sysroot,
   g_assert ((current_bootversion == 0 && new_bootversion == 1) ||
             (current_bootversion == 1 && new_bootversion == 0));
 
+  /* This allows us to support both /boot on a seperate filesystem to / as well
+   * as on the same filesystem. */
+  if (TEMP_FAILURE_RETRY (symlinkat (".", sysroot->sysroot_fd, "boot/boot")) < 0)
```
cgwalters, line 2003: Right, the simplicity of this is clearly tempting.  I think what I'm pausing on here though is that so far libostree doesn't really create files except in a few places.

But...I just want to revisit again, were we overthinking this previously around detecting `/boot` being a mount point?  Can't we just `statvfs(/)` and `statvfs(/boot)` and check the device IDs?

Passing down a boolean for that to the bootloader generation shouldn't be too hard.

cgwalters:
Thanks a lot for working on this BTW!

cgwalters:
To be clear I'm leaning towards https://github.com/ostreedev/ostree/pull/2145 but also happy to continue discussion.

wmanley:
> what I'm pausing on here though is that so far libostree doesn't really create files except in a few places.

AIUI `/boot` is one of those places - ostree creates and manages the `/boot/loader` symlink. Ideally I think it would be nice to do this in `ostree admin init-fs` instead, but for backwards compatibility creating it during deploy is necessary.

Could you expand on the reasoning for this policy? I'm having difficulty thinking of downsides.

An alternative would be to check for the `/boot/boot -> .` symlink and only prepend `/boot` if it's detected.  This would have the perverse effect that for users where boot is on root the symlink must exist but won't be used - but it would preserve backwards compatibility and mean that ostree wouldn't have to create that symlink at deploy time.

> But...I just want to revisit again, were we overthinking this previously around detecting `/boot` being a mount point? Can't we just `statvfs(/)` and `statvfs(/boot)` and check the device IDs?

I think there's a general problem that the system that is running `ostree admin deploy` isn't necessarily the same as the system that will be booting the deployed image, i.e when using the `--sysroot` option.  This has come up recently for me as I've been deploying to SD and then booting on another system, but I could imagine other scenarios where `ostree admin deploy --sysroot` is used as part of a larger build process.  The point being that you can't necessarily rely on probing the current filesystem setup to determine how it will be used/perceived by the bootloader.

I ran into this when working on #1404, which prompted the creation of #2145 - which, as you say, makes things explicit.  There's more discussion here: https://github.com/ostreedev/ostree/pull/1404#issuecomment-654169968

> To be clear I'm leaning towards #2145 but also happy to continue discussion.

I have a preference for this one rather than #2145 because:

* There's no configuration needed
* It defers to the bootloader at boot time, rather than relying on correctly specifying how the bootloader was configured ahead of time.  This is the central point that makes auto-detection so treacherous. We'd be using mount information of the currently booted system as a proxy for information on how the bootloader will behave when the system (or image) is booted.
* The implementation is super simple - #2145 was a pain to write to get concatenation of paths correct without double `//` or missing ones.
* Booting with ostree already depends on the bootloaders knowing how to resolve symlinks, so there's no additional requirements here.
* It's consistent with how we recommend that `/ostree -> sysroot/ostree`, etc., albeit `/boot` is a shared location, while these other symlinks
* It doesn't require further work in terms of improving auto-detection of how `/boot` is set up.  My experience of trying to get the auto-detection "right" in #1404 was that it was really quite difficult to write tests for.

I think there are some commits in #2145 that would be worth salvaging, like the `sysroot.bootloader` config option improvements.  But I can't see what benefits the `sysroot.boot_path` config option really provides over this PR, but I'd like to hear more about the file creation issue.

cgwalters:
@martinezjavier do you have an opinion on this one?

martinezjavier:
> @martinezjavier do you have an opinion on this one?

I think that I'm leaning towards this PR instead of #2125, not only because is simpler as you mentioned which is appealing but also because this issue is also present for GRUB.

GRUB expects the kernel and initrd images to be loaded from a path relative to the root of the partition where these are located. For example the `menuentry` commands (when not using the BLS module) would have something like the following:

```
linux /ostree/fedora-13eb3067f1bbc3ed7128faf435a87a5cbad451f99c2696380b7c7b65c57d8f03/vmlinuz-5.6.0-0.rc5.git1.1.fc33.x86_64
initrd /ostree/fedora-13eb3067f1bbc3ed7128faf435a87a5cbad451f99c2696380b7c7b65c57d8f03/initramfs-5.6.0-0.rc5.git1.1.fc33.x86_64.img
```

because the kernel and initrd images are in the `/ostree` directory that is in the root of the partition mounted as `/boot`. But if `/boot` is not a mount point, these should be something like:

```
linux /boot/ostree/fedora-13eb3067f1bbc3ed7128faf435a87a5cbad451f99c2696380b7c7b65c57d8f03/vmlinuz-5.6.0-0.rc5.git1.1.fc33.x86_64
initrd /boot/ostree/fedora-13eb3067f1bbc3ed7128faf435a87a5cbad451f99c2696380b7c7b65c57d8f03/initramfs-5.6.0-0.rc5.git1.1.fc33.x86_64.img
```

The GRUB `blscfg` module just translates the `linux` and `initrd` fields from the BLS snippets to GRUB `linux` and `initrd` commands, so the current BLS snippets generated by OSTree only work if `/boot` is a mount point.

By having the `/boot/boot -> .` symlink, the BLS snippets could always have a `/boot` prefix in the `linux` and `initrd` fields. If there is a boot partition, the symlink will just be followed by the bootloaders to the root of the filesystem in the boot partition.

So my suggestion would be to not make this something specific to syslinux and u-boot but instead add the prefix to the BLS snippets. That way `src/libostree/ostree-bootloader-{syslinux,uboot}.c` won't need any changes and could just use the value of the fields in the BLS snippets which already will have the path with the prefix.

I remember there was a similar issue with the `zipl` tool, that always expected the paths in the BLS snippets to have a `/boot` prefix. And the tool had to be changed to support the BLS generated by OSTree (see commit ibm-s390-tools/s390-tools@d71628326d8).

As @wmanley mentioned, OSTree already relies heavily on symlinks. Adding another one to cover this case doesn't seem that bad to me.

cgwalters:
I think I'd be a lot happier with this if this symlink wasn't something we invented in ostree, but e.g. something upstream bootloader maintainers recommended to create as well.  I have no idea how to drive that consensus though.

cgwalters:
OK restarted the CI
/lgtm

bam80:
So will it work for grub also?

wmanley:
> So will it work for grub also?

I'm sure it could be made to work for grub, but I haven't changed anything WRT grub here.  I'd expect the equivalent change for grub to be straightforward.

src/libostree/ostree-sysroot-deploy.c
cmurf: RE: Btrfs

statfs() f_fsid will be different per subvolume on the same file system. The first half (8 nibbles?) of the ID reported by `stat -f` seems to be consistent regardless of subvolume, but I'm not certain if the 2nd half of the ID can "roll over" once subvol ID's get high enough.

inode 256 is reserved for Btrfs subvolumes, i.e. if the file system is btrfs and the inode is 256, then it is a subvolume (or snapshot).

`mountpoint -d` returns the same value for each mountpoint that's the same Btrfs file system, even when backed by different subvolumes. This value looks like MAG:MIN but it doesn't match the MAG:MIN if the actual physical device partition.

Maybe more challenging is that a subvolume could be named anything and exist anywhere on a Btrfs file system, and yet mounted at /boot. This is relevant because this path to subvolume is what's needed for the bootloader to find it. I see BTRFS_IOC_INO_LOOKUP and BTRFS_IOC_TREE_SEARCH ioctl's used by `btrfs subvolume list` command.

cgwalters:
So...hmm, I think what we didn't end up doing here is this part:

> but instead add the prefix to the BLS snippets

?

martinezjavier:
> So...hmm, I think what we didn't end up doing here is this part:
> 
> > but instead add the prefix to the BLS snippets
> 
> ?

Yeah, it seems that was not done.

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/2705

PR #2150, opened 2020-07-16
pull: Also append bytes written

cgwalters:
This is very useful information that we get from the transaction
stats.  Append it to the final display if we're not inheriting
the transaction.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -3358,6 +3358,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
    * value, otherwise NULL. Used for logging.
    */
   const char *the_ref_to_fetch = NULL;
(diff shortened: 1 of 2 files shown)


cgwalters:
Split out from my work on https://github.com/openshift/machine-config-operator/issues/1897
example:
```
+ env OSTREE_PULL_MAX_OUTSTANDING_WRITES=1 chrt --idle 0 ionice -c 3 ostree --repo=repo pull-local srv-repo/repo bcfae65a2ab0b4ab5afcd1c9f1cce30b300e5408b04fc8bdec51a143ab593d40
+ fio --rw=write --ioengine=sync --fdatasync=1 --directory=etcdtest --size=30m --bs=2300 --name=etcd
7602 metadata, 30681 content objects imported; 2.2 GB content written       
```

PR #2151, opened 2020-07-16
ci: Constrain parallel build jobs

jlebon:
The default `_NPROCESSORS_ONLN` heuristic we have isn't cgroups aware.
So it thinks it has e.g. 40 CPUs when running in a k8s pod. This can
then blow through our allocated resource limits.

Declare some modest amount of RAM and CPU resources and override `make`
parallelism.

This matches what rpm-ostree now does in
https://github.com/coreos/rpm-ostree/pull/2155.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -2,7 +2,8 @@
 
 stage("Build") {
 parallel normal: {
-  cosaPod(buildroot: true, runAsUser: 0) {
+  def n = 5
+  cosaPod(buildroot: true, runAsUser: 0, memory: "2Gi", cpu: "${n}") {
       checkout scm
       stage("Core build") {
         shwrap("""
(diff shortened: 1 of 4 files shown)


jlebon:
```
+ ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   2348   524 ?        Ss   18:50   0:00 /usr/bin/sleep infinity
root          14  0.0  0.0      0     0 ?        Z    18:50   0:00 [sh] <defunct>
root          28  0.0  0.0      0     0 ?        Z    18:50   0:00 [sh] <defunct>
```

Let's try https://github.com/coreos/coreos-ci-lib/pull/33 ?

cgwalters:
Man the regressions from my PRs just don't stop :cry: 
I pushed https://github.com/ostreedev/ostree/pull/2151/commits/6a5f97c14522f46f202f6cf9449c6e9fe0f4d15b
Software is hard...fix one thing, break at least 1.5 other things...

jlebon:
Woohoo all green!
/lgtm

jlebon:
Oh right. Wanna do the honours? :)

PR #2152, opened 2020-07-16
pull: Add --per-object-fsync

cgwalters:
This is the opposite of
https://github.com/ostreedev/ostree/issues/1184

Motivated by OpenShift seeing etcd performance issues during
OS updates: https://github.com/openshift/machine-config-operator/issues/1897

Basically, if we switch to invoking `fsync()` as we go, it makes
ostree performance worse (in my tests, 31s to write 2G versus 7s if we
delay sync) but it avoids *huge* outliers in `fsync()` time for etcd.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -51,15 +51,33 @@
 #define FICLONE _IOW(0x94, 9, int)
 #endif
 
-
-/* If fsync is enabled and we're in a txn, we write into a staging dir for
- * commit, but we also allow direct writes into objects/ for e.g. hardlink
- * imports.
+/* Understanding ostree's fsync strategy
+ * 
+ * A long time ago, ostree used to invoke fsync() on each object,
+ * then move it into the objects directory.  However, it turned
+ * out to be a *lot* faster to write the objects into a separate "staging"
+ * directory (letting the filesystem handle writeback how it likes)
+ * and then only walk over each of the files, fsync(), then rename()
+ * into place.  See also https://lwn.net/Articles/789024/
+ *
+ * (We also support a "disable fsync entirely" mode, where you don't
+ * care about integrity; e.g. test suites using disposable VMs).
+ *
+ * This "delayed fsync" pattern though is much worse for other concurrent processes
+ * like databases because it forces a lot to go through the filesystem
+ * journal at once once we do the sync.  So now we support a `per_object_fsync`
+ * option that again invokes `fsync()` directly.  This also notably
+ * provides "backpressure", ensuring we aren't queuing up a huge amount
+ * of I/O at once.
  */
+
+/* The directory where we place content */
 static int
 commit_dest_dfd (OstreeRepo *self)
 {
-  if (self->in_transaction && !self->disable_fsync)
+  if (self->per_object_fsync)
+    return self->objects_dir_fd;
+  else if (self->in_transaction && !self->disable_fsync)
     return self->commit_stagedir.fd;
   else
     return self->objects_dir_fd;
@@ -420,7 +438,7 @@ commit_loose_regfile_object (OstreeRepo        *self,
   /* Ensure that in case of a power cut, these files have the data we
    * want.   See http://lwn.net/Articles/322823/
    */
-  if (!self->in_transaction && !self->disable_fsync)
+  if (!self->disable_fsync && self->per_object_fsync)
     {
       if (fsync (tmpf->fd) == -1)
         return glnx_throw_errno_prefix (error, "fsync");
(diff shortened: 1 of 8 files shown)


cgwalters:
For lots and lots more info, see: https://hackmd.io/WeqiDWMAQP2sNtuPRul9QA

(I will probably copy/paste that doc into here for posterity too when it's finalized)

cgwalters:
I am also considering adding a repo option for this to force it on by default.  

But the next step here is to change e.g. rpm-ostree to set this by default when it's pulling from a local repo for updates.

src/libostree/ostree-repo-private.h
```diff
@@ -38,13 +38,17 @@ G_BEGIN_DECLS
...
+ * (via GTask) all writing to disk.  Eventually we may
+ * use io_uring which handles backpressure correctly.
+ * Also, in "immediate fsync" mode, this helps provide
+ * much more backpressure, helping our I/O patterns
+ * be nicer for any concurrent processes, such as etcd
+ * or other databases.
+ * https://github.com/openshift/machine-config-operator/issues/1897
  * */
-#define _OSTREE_MAX_OUTSTANDING_WRITE_REQUESTS 16
+#define _OSTREE_MAX_OUTSTANDING_WRITE_REQUESTS 3
```
jlebon, line 51: It shouldn't though would be good to do a sanity-check that over-the-network pull performance isn't affected by this.

src/libostree/ostree-repo-commit.c
```diff
@@ -2377,6 +2409,9 @@ ostree_repo_commit_transaction (OstreeRepo                  *self,
   if (!rename_pending_loose_objects (self, cancellable, error))
     return FALSE;
 
+  if (!fsync_object_dirs (self, cancellable, error))
+    return FALSE;
```
jlebon, line 2413: Is there a purpose to the refactor into a separate function? Seems like `fsync` behaviour in this path hasn't really changed here, right?

src/libostree/ostree-repo-commit.c
```diff
@@ -2377,6 +2409,9 @@ ostree_repo_commit_transaction (OstreeRepo                  *self,
   if (!rename_pending_loose_objects (self, cancellable, error))
```
jlebon, line 2409: Can't we skip this function entirely in the case where we're writing directly into the object dir?

cgwalters: Probably, but per above we'd need to make more of the transaction/staging dir conditional.

src/libostree/ostree-repo-commit.c
cgwalters: In `per-object-fsync` mode we aren't using the staging dir; we still allocate one because making that conditional would be a notably bigger patch.  So we need to sync the actual target `objects/` directory.  And in any case I think we can get away with doing it this way rather than syncing the staging dir in both cases.

src/libostree/ostree-repo-private.h
cgwalters: With plain `kola http-server` over localhost I get 155MB/s and changing this slows it down from 9s to 15s...but that doesn't really matter a lot IMO.

I briefly investigated trying out `tc` but ended up doing `strace -f -e write --inject=write:delay_enter=5ms kola http-server 2>/dev/null` and there's no appreciable difference with that (~21MB/s, still a quite fast connection speed obviously).

cgwalters:
flake in test-concurrency.py still, I thought we'd fixed that :cry: 

cgwalters:
It's also interesting to compare the effect of `--fsync-incremental` on "fast ethernet" HTTP pulls.  It looks like it about doubles time from 35 to 68s with the same data, with the same "avoids huge latency spike" effect.

The more I think about this though I think ultimately what we want is to better control the total bandwidth of writes.  Which will [come with cgroups v2](https://andrestc.com/post/cgroups-io/) but we can't rely on that yet.

src/libostree/ostree-repo-commit.c
```diff
@@ -1835,6 +1853,50 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
...
+
+  if (self->disable_fsync)
+    return TRUE;  /* No fsync?  Nothing to do then. */
+
+  if (!glnx_dirfd_iterator_init_at (self->objects_dir_fd, ".", FALSE, &dfd_iter, error))
+    return FALSE;
+  while (TRUE)
+    {
+      struct dirent *dent;
+      g_auto(GLnxDirFdIterator) child_dfd_iter = { 0, };
```
jlebon, line 1873: Unused variable.

src/libostree/ostree-repo-commit.c
```diff
@@ -1835,6 +1853,50 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
...
+        continue;
+      /* All object directories only have two character entries */
+      if (strlen (dent->d_name) != 2)
+        continue;
+
+      glnx_autofd int target_dir_fd = -1;
+      if (!glnx_opendirat (self->objects_dir_fd, dent->d_name, FALSE,
+                           &target_dir_fd, error))
+        return FALSE;
+      if (fsync (target_dir_fd) == -1)
```
jlebon, line 1889: To confirm, in the `self->per_object_fsync` path, all the file objects at least should already have been `fsync`'ed during `commit_loose_regfile_object`, right? So ideally, this shouldn't actually be syncing much at this stage.

Maybe a comment about that?

src/libostree/ostree-repo-pull.c
```diff
@@ -3286,6 +3286,7 @@ initiate_request (OtPullData                 *pull_data,
  *   * disable-sign-verify (b): Disable signapi verification of commits
  *   * disable-sign-verify-summary (b): Disable signapi verification of the summary
  *   * depth (i): How far in the history to traverse; default is 0, -1 means infinite
+ *   * fsync-incremental (b): Perform disk writes more slowly, avoiding a single large I/O sync
```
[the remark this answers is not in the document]
cgwalters: Yeah, will do.

jlebon:
Hmm, Travis CI looks like a flake. Restarted it.
/lgtm

cgwalters:
It looks like for some reason the GPG agent socket is being written into the tempdir.  Need to chase that down but

PR #2155, opened 2020-07-24
lib/deploy: Add support for overlay initrds 

jlebon:
In FCOS and RHCOS, the need to configure software in the initramfs has
come up multiple times. Sometimes, using kernel arguments suffices.
Other times, it really must be a configuration file. Rebuilding the
initramfs on the client-side however is a costly operation. Not only
does it add complexity to the update workflow, it also erodes a lot of
the value obtained from using the baked "blessed" initramfs from the
tree itself.

One elegant way to address this is to allow specifying multiple
initramfses. This is supported by most bootloaders (notably GRUB) and
results in each initrd being overlayed on top of each other.

This patch allows libostree clients to leverage this so that they can
avoid regenerating the initramfs entirely. libostree itself is agnostic
as to what kind and how much data overlay initrds contain. It's up to
the clients to enforce such boundaries.

To implement this, we add a new option to the deploy APIs called
`overlay_initrds`: this is an array of filepaths to CPIO archives. We
copy these files into `/boot` and add them to the BLS as another
`initrd` entry.


cgwalters:
There's no use case I can think of for a "pristine" ostree shipping multiple initramfses.  So what if instead

- We just taught the bootconfig parser to correctly preserve multiple initramfs fields (done in this code)
- Downstream, projects like rpm-ostree could create e.g. `/boot/rpm-ostree/custom-initramfs-foo.img` and generate a new deployment using it.

IOW we think of this more like kernel arguments (which don't affect the commit checksum).

jlebon:
OK, updated this now based on feedback and https://github.com/coreos/rpm-ostree/pull/2170#issuecomment-673564367. We no longer look for the initramfs in the tree; it's instead passed via the new deploy APIs, checksummed, copied into `/boot`, and added into the BLS entry. The end result is that this is now an even more lightweight operation for clients!

jlebon:
Split out the trivial/prep patches into https://github.com/ostreedev/ostree/pull/2172.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2669,6 +2722,63 @@ _ostree_deployment_set_bootconfig_from_kargs (OstreeDeployment *deployment,
...
+/* Checksums overlay initrds of a deployment and updates the OstreeDeployment object. This
+ * is needed so that ostree_sysroot_write_deployments_with_options() can determine whether a
+ * bootswap is necessary when comparing deployments. */
+gboolean
+_ostree_checksum_overlay_initrds (OstreeSysroot    *self,
+                                  OstreeDeployment *deployment,
+                                  GCancellable     *cancellable,
+                                  GError          **error)
+{
+  /* On new deployments; overlay initrds are not yet written to /boot nor part of the
```
cgwalters, line 2734: Ownership of the overlay initrds is confusing to me.  For a libostree API user trying to add a new overlay, they own the files as long as the deployment is...staged?  Would a best practice be to use a temporary file?  (And why wouldn't we support e.g. an `O_TMPFILE`?)

What if we had a clearly distinct API like
`char *ostree_sysroot_install_overlay_initrd (OstreeSysroot *sysroot, int fd)` that would be called before creating a deployment using it?  Then ownership would be clear - create an `O_TMPFILE` with your initrd, call that API and get a checksum for it that you can pass down into the set of overlay initrds when adding another deployment.

If you call `install_overlay()` and don't call `deploy()/stage()` then they'd be GC'd, as one would expect.

jlebon: > What if we had a clearly distinct API like
`char *ostree_sysroot_install_overlay_initrd (OstreeSysroot *sysroot, int fd)` that would be called before creating a deployment using it? Then ownership would be clear - create an `O_TMPFILE` with your initrd, call that API and get a checksum for it that you can pass down into the set of overlay initrds when adding another deployment.

Hmm yeah, that sounds cleaner. Will give that a try!

jlebon: Did this now!

jlebon:
Added an installed tests and marked ready for review!

Requires: #2204

jlebon:
Rebased this now that #2204 is merged!

src/libostree/ostree-sysroot.c
```diff
@@ -815,6 +815,24 @@ list_deployments_process_one_boot_entry (OstreeSysroot               *self,
     return FALSE;
 
   ostree_deployment_set_bootconfig (deployment, config);
+  char **overlay_initrds = ostree_bootconfig_parser_get_overlay_initrds (config);
+  g_autoptr(GPtrArray) initrds_chksums = NULL;
+  for (char **it = overlay_initrds; it && *it; it++)
+    {
+      const char *basename = glnx_basename (*it);
+      if (strlen (basename) != (_OSTREE_SHA256_STRING_LEN + strlen (".img")))
```
cgwalters, line 823: It'd be nice to allow people to have a "short name" for their overlay initrd to describe its purpose without needing to unpack it to look.

Basically we append the checksum to the name they provide?

cgwalters (approved):
One comment, otherwise I think this is looking good!

src/libostree/ostree-sysroot.c
cgwalters: IOW it's `/boot/ostree/initramfs-overlays/kdump-41af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3.img` and not just `/boot/ostree/initramfs-overlays/41af286dc0b172ed2f1ca934fd2278de4a1192302ffa07087cea2682e7d372e3.img`

cgwalters: If you prefer we can do this as a followup too.

jlebon: Are you suggesting expanding the overlay initrd staging API as e.g.:

```diff
diff --git a/src/libostree/ostree-sysroot.h b/src/libostree/ostree-sysroot.h
index 3a3b6a77..ef5a353c 100644
--- a/src/libostree/ostree-sysroot.h
+++ b/src/libostree/ostree-sysroot.h
@@ -189,6 +189,7 @@ gboolean ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
 _OSTREE_PUBLIC
 gboolean ostree_sysroot_stage_overlay_initrd (OstreeSysroot  *self,
                                               int             fd,
+                                              const char      short_id,
                                               char          **out_checksum,
                                               GCancellable   *cancellable,
                                               GError        **error);
```

?

Hmm, or maybe we could change the semantics of `OstreeSysrootDeployTreeOpts.overlay_initrds` to have the checksum be optionally prefixed by an ID?

:thinking: Honestly, while the code here does support multiple overlays, I think in practice it's not something that will be used often. E.g. in the rpm-ostree use case, there is just a single "initramfs-etc" initrd, and my hope is that it's the one way through which users interact with this feature.

jlebon: I think I'd vote for leaving the API simple and we can do the second approach mentioned above if it comes up? (Because the first approach would obviously be an API break.)

cgwalters: OK, fine by me!

cgwalters (approved):
/lgtm

PR #2156, opened 2020-07-30
app: Fix various CLI metavariable names

jlebon:
- Use `REV` instead of `REF` in places where we meant it.
- Fix `commit --parent` actually taking a commit checksum and not a ref.
- Fix `ostree admin switch` using `REF` instead of `REFSPEC`.

### DIFF
--- a/src/ostree/ot-builtin-commit.c
+++ b/src/ostree/ot-builtin-commit.c
@@ -94,7 +94,7 @@ parse_fsync_cb (const char  *option_name,
  */
 
 static GOptionEntry options[] = {
-  { "parent", 0, 0, G_OPTION_ARG_STRING, &opt_parent, "Parent ref, or \"none\"", "REF" },
+  { "parent", 0, 0, G_OPTION_ARG_STRING, &opt_parent, "Parent commit checksum, or \"none\"", "COMMIT" },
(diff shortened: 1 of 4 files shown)


src/ostree/ot-builtin-commit.c
```diff
@@ -94,7 +94,7 @@ parse_fsync_cb (const char  *option_name,
  */
 
 static GOptionEntry options[] = {
-  { "parent", 0, 0, G_OPTION_ARG_STRING, &opt_parent, "Parent ref, or \"none\"", "REF" },
+  { "parent", 0, 0, G_OPTION_ARG_STRING, &opt_parent, "Parent checksum, or \"none\"", "CHECKSUM" },
```
agners, line 97: Thinking a bit more about it, can we use `COMMIT` here? Then we can use the `CHECKSUM` literal for arbitrary object types, e.g. as processed by `ostree show`...

cgwalters: Yep agree

jlebon: Fixed!

cgwalters:
We have CI fixes pending, this doesn't break them.
/override  continuous-integration/travis-ci/pr
/lgtm

cgwalters:
/override continuous-integration/jenkins/pr-merge continuous-integration/travis-ci/pr

PR #2157, opened 2020-07-31
Improve man pages

agners:
This mainly adds a basic glossary to define some literals commonly used. Writing them down helps to understand the man pages better and picking the right term/literal when adding features. There are probably some cases where the "wrong" literal has been used, but once we have a clear definition of the literals, it will become easier to harmonize their usage.


cgwalters:
Thank you so much for the patch!
/approve
/lgtm

PR #2158, opened 2020-07-31
ci: test FCOS PXE and ISO install

bgilbert:
Make sure we don't break the FCOS live image.  PXE is probably sufficient, but also test the ISO image for good measure.


cgwalters:
Followup inhttps://github.com/ostreedev/ostree/pull/2161

PR #2159, opened 2020-07-31
Show commit checksum of parent, if present

agners:
This is useful for ostree log on client side where often not the
full history of a branch is available. It is also helpful for
ostree show to show if a particular commit has a parent.


agners:
We could also add a flag to suppress `Parent` where not useful, but the way `ostree log` traverses parents does not make it straight forward. Still could be done, if preferred...

PR #2160, opened 2020-08-01
remount: Still remount /sysroot writable if not configured ro

cgwalters:
Regression from https://github.com/ostreedev/ostree/pull/2113/commits/35642259175973617da937f3cab6ce5f13c95077
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1862568

We still need to remount writable if it's not configured on;
because it may need OS adjustments it needs to be opt-in.


jlebon:
Makes sense to me!
/lgtm

PR #2161, opened 2020-08-01
ci: Fix ISO testing

cgwalters:
Regression from
https://github.com/ostreedev/ostree/pull/2158/commits/5d7f897908dbf7f471ddfdbd6c29a84ac6bc0bda

I'm not sure how (or if) this passed before, the job logs have
been GC'd.

This is a bit confusing but basically right now ostree/rpm-ostree's
CI jobs don't use `/srv/fcos` - it might make sense to port
these to `fcosBuild` but that needs investigation.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -79,7 +79,7 @@ parallel fcos: {
         coreos-assembler build
         coreos-assembler buildextend-metal
         coreos-assembler buildextend-metal4k
-        coreos-assembler buildextend-live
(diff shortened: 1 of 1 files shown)


bgilbert:
Whoops.  Thanks for the fix.

While you're there, `buildextend-live` could take `--fast` now.

cgwalters:
Thanks, taking that as a lgtm.
since that will be fixed by https://github.com/ostreedev/ostree/pull/2162

PR #2162, opened 2020-08-01
tests/repo-finder: Explicitly commit empty dir

cgwalters:
We were committing the whole tempdir, which seems to fail
in Travis because the GPG agent Unix domain socket ends up there too,
and ostree refuses to commit sockets.


cgwalters:
Fixed by other PR

PR #2163, opened 2020-08-05
 ci: Barf on unset umask

cgwalters:
Since it's just not a sane thing to do and will cause various
failures in our test suite.


cgwalters:
```
umask: 0000
```

Ughhh.  Evil.

cgwalters:
https://github.com/coreos/coreos-ci-lib/pull/34

Issue #2165, opened 2020-08-07
Avoid need to download remote config when pulling


pwithnall:
The code around [lines 3748–3768 of `ostree-repo-pull.c`](https://github.com/ostreedev/ostree/blob/master/src/libostree/ostree-repo-pull.c#L3748-L3768) requires downloading the `config` file from a remote to check two options:
 - `core.mode`, which it just does to abort if the mode isn’t `OSTREE_REPO_MODE_ARCHIVE`
 - `core.tombstone-commits` (a boolean)

Checking these two options means that `config` gets pulled by every client on every pull they do. On high latency or lossy connections, this adds noticeable time to a pull. It all seems a bit unnecessary.

Would it be possible to copy those options into the `summary` file, so that they can be checked using the copy of that which the client has already downloaded (or loaded from a cache)? Or is there some design reason for downloading `config` which I’m missing?

cgwalters:
Totally agree, this is just a legacy thing we should get rid of indeed.

pwithnall:
Fix available in #2167

PR #2166, opened 2020-08-07
pull: Add summary-{,sig-}bytes options to ostree_repo_pull()

pwithnall:
These allow the `summary` and `summary.sig` files to be cached at a
higher layer (for example, flatpak) between related pull operations (for
example, within a single flatpak transaction). This avoids
re-downloading `summary.sig` multiple times throughout a transaction,
which increases the transaction’s latency and introduces the possibility
for inconsistency between parts of the transaction if the server changes
its `summary` file part-way through.

In particular, this should speed up flatpak transactions on machines
with high latency network connections, where network round trips have a
high impact on the latency of an overall operation.


pwithnall:
See also #2165

PR #2167, opened 2020-08-10
Avoid need to download remote config when pulling

### LINKED ISSUE #2165: Avoid need to download remote config when pulling

pwithnall:
Fixes #2165. See that and the commit messages for details.

This doesn’t need additional changes in flatpak for it to be used there too. It does require repositories to regenerate their `summary` files before the non-fallback code path is taken in clients, though. That’ll happen soon enough naturally.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -2001,6 +2001,8 @@ start_fetch (OtPullData *pull_data,
                                       is_meta ? meta_fetch_on_complete : content_fetch_on_complete, fetch);
 }
 
+/* Deprecated: code should load options from the `summary` file rather than
+ * downloading the remote’s `config` file, to save on network round trips. */
 static gboolean
 load_remote_repo_config (OtPullData    *pull_data,
                          GKeyFile     **out_keyfile,
@@ -3757,30 +3759,6 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
       if (!ostree_repo_open (pull_data->remote_repo_local, cancellable, error))
         goto out;
     }
-  else
-    {
-      if (!load_remote_repo_config (pull_data, &remote_config, cancellable, error))
-        goto out;
-
-      if (!ot_keyfile_get_value_with_default (remote_config, "core", "mode", "bare",
-                                              &remote_mode_str, error))
-        goto out;
-
-      if (!ostree_repo_mode_from_string (remote_mode_str, &pull_data->remote_mode, error))
-        goto out;
-
-      if (!ot_keyfile_get_boolean_with_default (remote_config, "core", "tombstone-commits", FALSE,
-                                                &pull_data->has_tombstone_commits, error))
-        goto out;
-
-      if (pull_data->remote_mode != OSTREE_REPO_MODE_ARCHIVE)
-        {
-          g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
-                       "Can't pull from archives with mode \"%s\"",
-                       remote_mode_str);
-          goto out;
-        }
-    }
   }
 
   /* Change some option defaults if we're actually pulling from a local
@@ -3854,6 +3832,8 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
     g_autoptr(GVariant) deltas = NULL;
     g_autoptr(GVariant) additional_metadata = NULL;
     gboolean summary_from_cache = FALSE;
+    gboolean remote_mode_loaded = FALSE;
+    gboolean tombstone_commits = FALSE;
 
     if (summary_sig_bytes_v)
       {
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -4172,6 +4152,42 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
+                                                      &pull_data->has_tombstone_commits, error))
+              goto out;
+
+            if (pull_data->remote_mode != OSTREE_REPO_MODE_ARCHIVE)
+              {
+                g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                             "Can't pull from archives with mode \"%s\"",
+                             remote_mode_str);
+                goto out;
+              }
```
jlebon, line 4189: This check should be done in both the new and old paths, no?

src/libostree/ostree-repo-pull.c
```diff
@@ -4172,6 +4152,42 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
...
           }
+
+        if (g_variant_lookup (additional_metadata, OSTREE_SUMMARY_MODE, "s", &remote_mode_str) &&
+            g_variant_lookup (additional_metadata, OSTREE_SUMMARY_TOMBSTONE_COMMITS, "b", &tombstone_commits))
+          {
+            if (!ostree_repo_mode_from_string (remote_mode_str, &pull_data->remote_mode, error))
+              goto out;
+            pull_data->has_tombstone_commits = tombstone_commits;
+          }
+        else if (pull_data->remote_repo_local == NULL)
```
jlebon, line 4163: Hmm, don't we need to fallback also if the remote doesn't have a summary file? I.e. I think this needs to be moved out of the `if (pull_data->summary)` if-statement and executed if both it's a remote repo and we didn't already fetch the info we need from the optional summary file.

src/libostree/ostree-repo.c
```diff
@@ -5823,6 +5823,24 @@ ostree_repo_regenerate_summary (OstreeRepo     *self,
                                  g_variant_new_uint64 (GUINT64_TO_BE (g_get_real_time () / G_USEC_PER_SEC)));
   }
 
+  {
+    g_autofree char *remote_mode_str = NULL;
```
cgwalters, line 5827: We can land this hunk to simply add the data to the summary now as a prep patch.

pwithnall:
Sorry for the delay. I’ve updated as per the review comments

src/libostree/ostree-repo-pull.c
pwithnall: Actually, no. Turns out it should only be done on the `pull_data->remote_repo_local == NULL` path (the old behaviour) or a lot of the unit tests fail. I don’t know why this is. `load_remote_repo_config()` was previously only called on that path.

jlebon: Ahh yes, you're right.

> Turns out it should only be done on the pull_data->remote_repo_local == NULL path (the old behaviour) or a lot of the unit tests fail. I don’t know why this is.

Probably because we support `pull-local` from repos that aren't archive repos.

jlebon:
Looks like `test-summary-view.sh` needs to be adjusted:

```
# ([('main', (uint64 12538021362599460864, [byte 0xd9, 0xc5, 0x1e, 0x2a, 0xe5, 0xfe, 0x81, 0x9a, 0x94, 0x0d, 0xba, 0xef, 0x1f, 0xfc, 0xac, 0x98, 0xfa, 0x18, 0x88, 0x2c, 0x6c, 0x74, 0x37, 0x8a, 0x32, 0x40, 0x64, 0xc9, 0xf4, 0x42, 0xbe, 0xc8], {'ostree.commit.timestamp': <uint64 1601051836>})), ('other', (10808639105689190400, [0xbe, 0x51, 0xcd, 0xd8, 0xa1, 0x45, 0x91, 0x87, 0xd4, 0x3a, 0x64, 0x99, 0xdc, 0x45, 0x90, 0x87, 0x79, 0x39, 0x98, 0x15, 0xc4, 0xe6, 0xdf, 0x22, 0xc4, 0x6d, 0x93, 0x9c, 0x98, 0xcf, 0x83, 0xe2], {'ostree.commit.timestamp': <uint64 1601051836>}))], {'ostree.summary.mode': <'archive-z2'>, 'ostree.summary.last-modified': <uint64 1601051836>, 'ostree.summary.tombstone-commits': <false>})
+ fatal 'File '\''raw-summary.txt'\'' doesn'\''t match fixed string list '\''{'\''ostree.summary.last-modified'\'': <uint64'\'''
+ echo File ''\''raw-summary.txt'\''' 'doesn'\''t' match fixed string list ''\''{'\''ostree.summary.last-modified'\'':' '<uint64'\'''
File 'raw-summary.txt' doesn't match fixed string list '{'ostree.summary.last-modified': <uint64'
```

LGTM otherwise!
/approve

pwithnall:
Adjusted that test. Not sure what the continuous-integration/jenkins/pr-merge failure is about. The Jenkins page shows only green ticks.

cgwalters (approved):
/lgtm

PR #2170, opened 2020-08-13
dracut: Create reproducible images

jprvita:
Without reproducible images, a rebuild of the initrd will create a
different image file (due to things like creation time of the files in
the cpio archive) even if the actual contents in it are exactly the
same, adding an unnecessary download during updates.

Adding 'reproducible=yes' avoids this and creates the same image files
for the same content.

### DIFF
--- a/src/boot/dracut/ostree.conf
+++ b/src/boot/dracut/ostree.conf
@@ -16,3 +16,4 @@
 # Boston, MA 02111-1307, USA.
 
 add_dracutmodules+=" ostree systemd "
+reproducible=yes


jprvita:
We are shipping this change on Endless OS downstream.

jlebon:
Wouldn't it be cleaner to have whatever runs `dracut` ensure it passes `--reproducible`? This is what rpm-ostree does (and actually, Fedora/RHEL doesn't ship this config at all nowadays; see https://bugzilla.redhat.com/show_bug.cgi?id=1331369).

Realistically, this is probably fine and I agree that most (all?) users of ostree + dracut would want this. Though it feels like this is at the wrong level of the stack; given that `--reproducible` affects the whole initramfs ISTM this belongs in the layer that binds ostree and dracut together.

Anyway, weak
/approve

Will let @cgwalters have a look.

cgwalters:
Agree with @jlebon - this is a distro decision, not ostree's decision (though it clearly benefits ostree-based systems a lot).

I also think basically everyone should be using `--with-dracut=yesbutnoconf` nowadays as Fedora is - the config file we install is equivalent to just having `--add=ostree` on the dracut cmdline, which is utterly trivial to do (and is also what rpm-ostree does).

But that said, this doesn't hurt so
/lgtm

PR #2172, opened 2020-08-17
Miscellaneous patches split out of #2155

jlebon:
See individual commit messages.


cgwalters:
Nice!
I had to triple check the karg one, but I see the flow now.
/lgtm

PR #2181, opened 2020-08-25
tests/inst: Port to new sh-inline repo

cgwalters:
I cleaned up my fork of commandspec (see git log) and am
planning to publish to crates.  Port to the new API in prep
for that.


cgwalters:
See https://github.com/cgwalters/rust-sh-inline/commits/master

PR #2182, opened 2020-08-25
Fix mkinitcpio with newer systemd versions

mbilker:
`ostree-prepare-root.service` may race with systemd failing to determine if `/sysroot` is valid because of `/etc/os-release` not being available yet.

Tested this under Arch Linux on multiple systems with different some having SSDs and some having HDDs.

Related: #1759

### DIFF
--- a/src/boot/mkinitcpio/ostree
+++ b/src/boot/mkinitcpio/ostree
@@ -5,6 +5,6 @@ build() {
     add_binary /usr/lib/ostree/ostree-remount
 
     add_file /usr/lib/systemd/system/ostree-prepare-root.service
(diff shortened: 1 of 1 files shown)


mbilker:
I should note that I assume the original mkinitcpio install script worked before but broke sometime in the last 7 years since it was contributed to ostree.

cgwalters:
Thanks for the patch!  LGTM, it's the same as what we have in the dracut glue today.
/approve
/ok-to-test
/lgtm

cgwalters:
Ah right CI needs https://github.com/coreos/rpm-ostree/pull/2206 to pass
/override  continuous-integration/jenkins/pr-merge

PR #2183, opened 2020-08-26
tests/inst: Update to published sh-inline crate

cgwalters:
And I made a few more API tweaks, such as supporting `Path`
objects directly and also not needing e.g. `commit = commit`, see

- https://github.com/cgwalters/rust-sh-inline/commit/cfa7c71126f23545a7d4782cad650eab60e74204
- https://github.com/cgwalters/rust-sh-inline/commit/679bce4cc7ce65641e0c9bd33654510575583de8

### DIFF
--- a/tests/inst/Cargo.toml
+++ b/tests/inst/Cargo.toml
@@ -14,8 +14,7 @@ structopt = "0.3"
 serde = "1.0.111"
 serde_derive = "1.0.111"
 serde_json = "1.0"
-# To be published on crates.io soon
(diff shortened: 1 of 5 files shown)


jlebon:
Nice! Next natural step is full string interpolation. :)

/lgtm

cgwalters:
> Nice! Next natural step is full string interpolation. :)


https://github.com/cgwalters/rust-sh-inline/issues/1

cgwalters:
/override ci/prow/sanity
Prow is having issues, this code is only run in coreosci right now anyways.

PR #2188, opened 2020-09-01
Add indexes for deltas outside of the summary

alexlarsson:
As mentioned in my [recent mail](https://mail.gnome.org/archives/ostree-list/2020-August/msg00018.html) on repo scalability listing all the deltas in the summary is quite wasteful. For example, for flathub it currently makes the summary file 1.1 MB larger.

Typically you only need the list of deltas when pulling a commit, and all you need then is the list of deltas going to that particular commit. So, this PR adds a separate delta-index/ directory with one file per target commit, containing the subset of deltas affecting it. These indexes are updated when the summary is updated and the in-summary delta index would normally be updated (at similar cost). Note that this had to be put outside of the old deltas/ directory dues to a bug (fixed in the PR) that breaks older ostree files if there are regular files in the wrong place in this directory.

The patch adds a `core.no-deltas-in-summary` option that allows you to drop the deltas from the summary. However, as that would break older versions looking for the deltas there it is off by default. This makes this change look not very useful, however I have plans of adding a new form of summary with other efficiency improvements, and since those forms will not be read by older versions of ostree they can default to not contain delta indexes.


alexlarsson:
I rebased this on top of 2020.6, and furthermore, including the cleanups from #2193, so lets land that first.

alexlarsson:
New rebase on top of the landed cleanup.

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -982,3 +982,170 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+      g_autofree char *index_path = _ostree_get_relative_static_delta_index_path (to);
+
+      if (froms == NULL)
+        {
+          /* No index to this checksum seen, delete if it exists */
+
+          g_debug ("Removing delta index for %s", to);
+          if (!glnx_shutil_rm_rf_at (repo->repo_dir_fd, index_path,
+                                     cancellable, error))
+            return FALSE;
```
jlebon, line 1092: Since it's a single file, how about just `unlinkat()` instead of pulling out the big guns? E.g.

```suggestion
          if (!ot_ensure_unlinked_at (repo->repo_dir_fd, index_path, error))
            return FALSE;
```

man/ostree.repo-config.xml
```diff
@@ -249,6 +249,20 @@ Boston, MA 02111-1307, USA.
         costly).
         </para></listitem>
       </varlistentry>
+
+      <varlistentry>
+        <term><varname>no-deltas-in-summary</varname></term>
+        <listitem><para>Boolean value controlling whether or not OSTree puts
+        an index of available deltas in the summary file. Defaults to false.
```
jlebon, line 256: Maybe

```suggestion
        <listitem><para>Boolean value controlling whether OSTree should skip
        putting an index of available deltas in the summary file. Defaults to false.
```

?

That way the "Defaults to false." reads better.

man/ostree.repo-config.xml
```diff
@@ -249,6 +249,20 @@ Boston, MA 02111-1307, USA.
...
+        </para>
+        <para>
+        Since 2020.7 OSTree creates and uses delta indexes outside the summary file,
```
jlebon, line 259:
```suggestion
        Since 2020.7 OSTree can use delta indexes outside the summary file,
```

src/libostree/ostree-repo-pull-private.h
```diff
@@ -74,7 +74,8 @@ typedef struct {
   GBytes           *summary_data;
   GBytes           *summary_data_sig;
   GVariant         *summary;
-  GHashTable       *summary_deltas_checksums;
+  GHashTable       *summary_deltas_checksums; /* Filled from summary and delta indexes */
```
jlebon, line 77: IIUC, this should be

```suggestion
  GHashTable       *summary_deltas_checksums; /* Filled from summary or delta indexes */
```

 right?

src/libostree/ostree-repo-pull.c
```diff
@@ -2623,6 +2661,132 @@ validate_variant_is_csum (GVariant       *csum,
...
+          g_variant_get_child (ref, 1, "v", &csum_v);
+
+          if (!validate_variant_is_csum (csum_v, error))
+            goto out;
+
+          guchar *csum_data = g_malloc (OSTREE_SHA256_DIGEST_LEN);
+          memcpy (csum_data, ostree_checksum_bytes_peek (csum_v), 32);
+          g_hash_table_insert (pull_data->summary_deltas_checksums,
+                               g_strdup (delta),
+                               csum_data);
```
jlebon, line 2712: Minor: could factor out all this for sharing with the in-summary version.

tests/test-delta.sh
```diff
@@ -249,6 +260,41 @@ ${CMD_PREFIX} ostree --repo=repo2 ls ${samerev} >/dev/null
 
 echo 'ok pull empty delta part'
 
+rm -rf repo/delta-indexes
+${CMD_PREFIX} ostree --repo=repo summary -u
+${CMD_PREFIX} ostree summary -v --raw --repo=repo > summary.txt
+assert_file_has_content summary.txt "ostree\.static\-deltas"
```
jlebon, line 266: Should also sanity-check here (and similarly below) that the directory is empty.

tests/test-delta.sh
```diff
@@ -249,6 +260,41 @@ ${CMD_PREFIX} ostree --repo=repo2 ls ${samerev} >/dev/null
...
+
+${CMD_PREFIX} ostree --repo=repo config set core.no-deltas-in-summary true
+${CMD_PREFIX} ostree --repo=repo summary -u
+
+${CMD_PREFIX} ostree summary -v --raw --repo=repo > summary.txt
+assert_not_file_has_content summary.txt "ostree\.static\-deltas"
```
jlebon, line 272: Should also sanity-check here (and similarly below) that the directory isn't empty.

jlebon (commented):
Approach looks sane to me overall. Thanks for splitting out the commits nicely!
Mostly minor comments. Coding style needs updating.

cgwalters (commented):
OK I replied on list; overall I understand the need but I think we should do a bit more design.  Maybe a realtime meeting on this?  We could create a shared hackmd.io doc and go through some of the positives/negatives of different approaches.

src/libostree/ostree-repo-pull-private.h
alexlarsson: Well, the code does handle the case where they come from both. I mean, we don't really *expect* that to happen, but it could (if mixing version in some weird way maybe), and we handle it.

tests/test-delta.sh
alexlarsson: Its not though, we always generate the delta indexes, even if they are *also* in the summary. This is because eventually we want to have some new version of the summary not have the deltas but the old one have it, and then new clients will use the delta indexes and old clients the deltast in the old summary.

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -982,3 +982,170 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+    {
+      const char *delta_name = g_ptr_array_index (all_deltas, i);
+      g_autofree char *from = NULL;
+      g_autofree char *to = NULL;
+      GPtrArray *deltas_to_commit = NULL;
+
+      if (!_ostree_parse_delta_name (delta_name, &from, &to, error))
+        return FALSE;
+
+      if (opt_to_commit != NULL && strcmp (to, opt_to_commit) != 0)
```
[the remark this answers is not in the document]
alexlarsson: No, this is not right. If opt_to_commit is NULL (i.e. produce indexes for all commits) we don't want to do this check, because all "to" are different to NULL and will be skipped.

alexlarsson:
@cgwalters Sure, having a real life meeting would be good. I'm almost ready with a new version of the summaries though, might as well finish that so everyone could have a cursory look at it before the meeting.

tests/test-delta.sh
jlebon: Ahh right OK, forgot about that.

src/libostree/ostree-repo-static-delta-core.c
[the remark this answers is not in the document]
jlebon: Right OK. Hmm, in that case I'd prefer

```suggestion
      if (opt_to_commit != NULL && !g_str_equal (to, opt_to_commit))
```

but consider that optional.

dbnicholson:
@alexlarsson have you checked what the size difference is for the flathub summary file without the delta index? My hunch was that `xa.cache` would be the biggest chunk, but having a look at one of our summary files it does appear that the delta listing is pretty big.

dbnicholson:
> @alexlarsson have you checked what the size difference is for the flathub summary file without the delta index? My hunch was that `xa.cache` would be the biggest chunk, but having a look at one of our summary files it does appear that the delta listing is pretty big.

Nevermind, you said 1.1 MB in the cover letter.

dbnicholson:
> The patch adds a `core.no-deltas-in-summary` option that allows you to drop the deltas from the summary. However, as that would break older versions looking for the deltas there it is off by default. This makes this change look not very useful, however I have plans of adding a new form of summary with other efficiency improvements, and since those forms will not be read by older versions of ostree they can default to not contain delta indexes.

For sure it would be pretty hard to turn that on for quite some time, although it wouldn't break old clients, only force them to do object fetches. I might say that this repo option isn't worth the trouble if you have a summary format in mind that only new clients will fetch.

Overall I think this approach is nice. I'm interested to see how it plays with a different summary format, though. The only other thing that came to mind is that I don't believe the index file is updated when a delta is created or pruned, only when the summary is updated. I think that's OK and probably not worth the trouble, but it could be nice to move the reindexing work out of the summary path.

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -895,3 +982,168 @@ _ostree_repo_static_delta_dump (OstreeRepo                    *self,
...
+
+  return g_bytes_equal (existing_data, data);
+}
+
+gboolean
+_ostree_repo_static_delta_reindex (OstreeRepo                 *repo,
+                                   const char                 *opt_to_commit,
+                                   GCancellable               *cancellable,
+                                   GError                    **error)
+{
```
dbnicholson, line 1023: Even though the summary path does take an exclusive lock, I think you should at least take a shared lock here to ensure you're not racing with another process pruning deltas. An exclusive lock would (I believe) also ensure that another process isn't creating deltas while indexing. With the CLI you added, this can also be called outside of the summary generation path.

alexlarsson:
> > The patch adds a `core.no-deltas-in-summary` option that allows you to drop the deltas from the summary. However, as that would break older versions looking for the deltas there it is off by default. This makes this change look not very useful, however I have plans of adding a new form of summary with other efficiency improvements, and since those forms will not be read by older versions of ostree they can default to not contain delta indexes.
> 
> For sure it would be pretty hard to turn that on for quite some time, although it wouldn't break old clients, only force them to do object fetches. I might say that this repo option isn't worth the trouble if you have a summary format in mind that only new clients will fetch.
> 
> Overall I think this approach is nice. I'm interested to see how it plays with a different summary format, though. The only other thing that came to mind is that I don't believe the index file is updated when a delta is created or pruned, only when the summary is updated. I think that's OK and probably not worth the trouble, but it could be nice to move the reindexing work out of the summary path.

Yeah, I'm not sure how useful the option is to use this for older repos. It was mainly a way to test this work while it was done. The later work enables this feature automatically for indexed summaries, while keeping it for the backwards-compat summaries. So, only apps that declare they support new-format summaries will get them.

I don't think we want to reindex all the time a ref changes. Its a O(n-refs-in-repo) operation similar to summary update, so if you add multiple refs in an operation (like adding a bew flathub build) you will get n^2 behaviour. So, you want to update this only when you know you are in a steady state, just like the summary.

alexlarsson:
Pushed a pure rebase. Will add a few more commits targeting the "minimal ostree changes" approach to summary alternatives.

src/libostree/ostree-repo-static-delta-core.c
alexlarsson: I think a shared lock makes sense, to avoid a parallel prune removing commits that the generated indexes will refer too. However, I think an exclusive lock is overkill (and unnecessarily deadlock risky), because at most a parallel call that was generated before the reindex is finished will be missing. 

If you want to guarantee that a newly generated delta is index you have to manually ensure reindex is called after the delta is fully created. An exclusive lock on the indexing will not help you with that, all it will do is delay the start of delta creation if a reindex is running.

alexlarsson:
I added some commits, and now the pull code doesn't use the summary at all if it doesn't need it, which is the case for non-mirror pulls that specify commit ids for all the refs (assuming the remote has deltas outside of the summary).

This is good in general, as we don't need to download a potentially large file and not use it. But it also means that callers of ostree are very flexible in terms of when summary files are needed. i.e. the caller are free to communicate the commit ids out of band in a more efficient way than the summary file without ostree itself causing an inefficient download of the summary file.

src/ostree/ot-builtin-static-delta.c
```diff
@@ -226,7 +226,7 @@ ot_static_delta_builtin_reindex (int argc, char **argv, OstreeCommandInvocation
   if (!ostree_option_context_parse (context, reindex_options, &argc, &argv, invocation, &repo, cancellable, error))
     return FALSE;
 
-  if (!ostree_cmd__private__ ()->ostree_static_delta_reindex (repo, opt_to_rev, cancellable, error))
+  if (!ostree_repo_static_delta_reindex (repo, 0, opt_to_rev, cancellable, error))
```
dbnicholson, line 229: Obviously if this is public API (which to me it should be), then I think you should reorder the commits so the `cmd_private__` API isn't ever added.

alexlarsson: Yeah, I didn't want to make the new changes to large to review though.

alexlarsson: New version is reordered

alexlarsson:
Reordered per @dbnicholson comment and fixed a bug in when deltas are required. Looking at using this from flatpak atm.

cgwalters:
An overall concern I have with this is basically: at what point can e.g. flathub pull the trigger and stop adding deltas to the summary?  It's not clear to me how many flatpak clients are on long-lifecycle distributions but I bet it's probably at least 20-30%?  Do we have that data?

Basically, we merge this and it might be...how many months/years for that?

> I added some commits, and now the pull code doesn't use the summary at all if it doesn't need it, which is the case for non-mirror pulls that specify commit ids for all the refs (assuming the remote has deltas outside of the summary).

This bit though is an interesting topic.  To rephrase this, it sounds like if flathub deploys the updated ostree quickly (easy) then new clients can ignore the large summary file in most normal cases of applying updates?  Which helps address the above concern a lot.

alexlarsson:
> An overall concern I have with this is basically: at what point can e.g. flathub pull the trigger and stop adding deltas to the summary? It's not clear to me how many flatpak clients are on long-lifecycle distributions but I bet it's probably at least 20-30%? Do we have that data?
> 
> Basically, we merge this and it might be...how many months/years for that?

The plan is to keep them there indefinitely, or until we have basically no users. However, the creation of the summary would be completely managed by flatpak, so we can limit which deltas we put in the summary. For example, if we keep only the x86-64 deltas in the summary then it would not become increasingly bloated as we add arches, keeping old 86-64 clients working.

Then in *parallel* to the summary file we would write a different summary format without the delta info in. These would be only used by new flatpak version, and use a more efficient format and allow delta updates. Flatpak will use these to pre-resolve any pulls to the individual commit so the ostree code would not need to pull the (compat) summary.

> > I added some commits, and now the pull code doesn't use the summary at all if it doesn't need it, which is the case for non-mirror pulls that specify commit ids for all the refs (assuming the remote has deltas outside of the summary).
> 
> This bit though is an interesting topic. To rephrase this, it sounds like if flathub deploys the updated ostree quickly (easy) then new clients can ignore the large summary file in most normal cases of applying updates? Which helps address the above concern a lot.

Yes, that is the plan.

jlebon:
> Then in parallel to the summary file we would write a different summary format without the delta info in. These would be only used by new flatpak version, and use a more efficient format and allow delta updates. Flatpak will use these to pre-resolve any pulls to the individual commit so the ostree code would not need to pull the (compat) summary.

Question about this: why wouldn't we want this semantic in libostree itself? Specifically: if we have delta indexes, why wouldn't we also teach libostree pull-via-ref to just fetch the ref content and check for deltas the new way? (And thus avoid the summary entirely). That should simplify the flatpak case too since it no longer needs to store a separate summary file.

And then, once we have *that*, we don't really need to change the summary format at all; we can just leave all the info in there.

To rephrase: let's make sure that the summary is never in the hot path for pull operations and have for sole purpose discoverability (and for that, you do want all the static deltas listed).

alexlarsson:
> > Then in parallel to the summary file we would write a different summary format without the delta info in. These would be only used by new flatpak version, and use a more efficient format and allow delta updates. Flatpak will use these to pre-resolve any pulls to the individual commit so the ostree code would not need to pull the (compat) summary.
> 
> Question about this: why wouldn't we want this semantic in libostree itself? Specifically: if we have delta indexes, why wouldn't we also teach libostree pull-via-ref to just fetch the ref content and check for deltas the new way? (And thus avoid the summary entirely). That should simplify the flatpak case too since it no longer needs to store a separate summary file.

I'm not sure that is always better. It depends on the details whether this is an optimization or not, because you have to do a network roundtrip per pulled ref file. So, if you are pulling a single ref, its clearly an optimization. But if you are pulling 10 refs then it depends on the size of the summary, for a smaller repo it may be better to use the summary in this case.

For flatpak it doesn't really change anything as we need a bunch of metadata available locally to resolve dependencies, etc. 

Also, take the most common operation "flatpak update", it would have to iterate over all installed apps and fetch the ref file for that, even if most are likely unchanged. This is much better done locally with a single summary file, especially if we have an efficient way to update the cached copy (as the incremental changes in it are bound to be small).

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -1205,3 +1205,168 @@ ostree_repo_static_delta_verify_signature (OstreeRepo       *self,
...
+        {
+          g_auto(GVariantDict) index_builder = OT_VARIANT_BUILDER_INITIALIZER;
+          g_auto(GVariantDict) deltas_builder = OT_VARIANT_BUILDER_INITIALIZER;
+          g_autoptr(GVariant) index_variant = NULL;
+          g_autoptr(GBytes) index = NULL;
+
+          /* We sort on from here so that the index file is reproducible */
+          g_ptr_array_sort (froms, (GCompareFunc)g_strcmp0);
+
+          g_variant_dict_init (&deltas_builder, NULL);
```
wmanley, line 1364: This has GVariant type `a{say}`.  An alternative would be `a(ayayay)`.  This way from and to are stored as separate elements, so we don't need to do our own serialisation (`$from-$to`) and $from and $to wouldn't need to be encoded as hex, halving the size.

cgwalters:
I rebased :surfer: this in prep for merging.

src/libostree/ostree-repo-static-delta-core.c
alexlarsson: It is using the same encoding as the old summary file is. It is indeed not very efficient, and if we were to still try to keep all the delta files in the summary we should make it more efficient. However, the new format just keeps a few deltas in each file, so in practice you will not download a lot of data. So, changing this is not as important, especially as it would require different code in the ostree pull codepath to handle the new format.

cgwalters:
Where did we land on signatures for this?

Do we believe signatures are unnecessary because we already check the signature on the commit object for deltas and validate the objects?

alexlarsson:
I added a ostree_repo_gpg_sign_data() because I want to sign my summaries. Flatpak uses this to allow passing updated remote metadata changes via summary properties. For example adding a new GPG key or redirecting to a new URL. And, we use the gpg to be able to forward this metadata update to the system helper where we can't rely on the TLS.

alexlarsson:
@cgwalters we do actually record the digest of the summary in the summary, not only the name.

We do sign the commit from the delta, so I guess the worry would be that an attacker would feed the client a delta that the delta handling code mismanages to cause some exploit in the delta apply code? 

Honestly, I'm not to worried about this. Delta generation is suuuper complex, but applying them is pretty basic, and we end up verifying the result.

cgwalters:
Yes...but my recollection here is that we went to signing the summary partly to ensure that deltas were directly covered by signatures because of concerns like this.

Hmm.  Also since the last release ostree learned how to have deltas include inline signatures.  Perhaps we should support making that required for split delta indices?

Something like `gpg-verify-deltas` (defaulting to false for backcompat)?

Would it work for e.g. flathub to GPG sign deltas this way?

> Delta generation is suuuper complex, but applying them is pretty basic, and we end up verifying the result.

Yeah, it's also C code though.  Ideally eventually ostree learns to fork off the delta processing as a low-privilege helper that can just pass back objects as `O_TMPFILE` file descriptors or so.

cgwalters:
(Feel free to push back again and say "I'm fine shipping delta superblocks without gpg as part of moving to split delta indices")

Hmm, or maybe we error out only if `gpg-verify-summary=true`?

alexlarsson:
Signing deltas seem like something that might be interesting, but I'm really not overly worried about it, and it does make creating the deltas a bit more complex. Especially for flathub where we create deltas on separate machines, so we would have to patch in the signatures when we import them.

I'd rather skip this for now.

cgwalters:
Rebased this to retrigger CI
/lgtm

cgwalters:
```
error: Full sync: During fsfreeze-thaw: write(watchdog syncfs complete): Broken pipe
```

Hum that's...interesting.  `EPIPE` here apparently means the reading end was closed.  But that error is being lost to the void.

cgwalters:
OK looks like this needs a rebase since https://github.com/ostreedev/ostree/pull/2205 merged first.

alexlarsson:
Rebased on master. However, some of the stuff here kinda negates the work in #2205, in particular that we now download the config file in the no-need-for-summary case. @pwithnall If you're still interested in this work we should look at a client-side cache for that too. (And given that config changes seldom we can probably only refresh the case once a day or so.)

alexlarsson:
As an aside, while reviewing this I noticed that we now have the possibility of caching the summary even if we don't have a summary.sig (due to the etags), but we don't currently do this. Not sure how important that is, but it is another possible optimization.

pwithnall:
> @pwithnall If you're still interested in this work we should look at a client-side cache for that too. (And given that config changes seldom we can probably only refresh the case once a day or so.)

I am still interested, but don’t have time to work on it at the moment. Happy to review anything you put together if you CC me in.

It would be good to avoid downloading the `config` file regularly. Round trips are bad. Although at least it’s a lot smaller than the `summary`.

> As an aside, while reviewing this I noticed that we now have the possibility of caching the summary even if we don't have a summary.sig (due to the etags), but we don't currently do this. Not sure how important that is, but it is another possible optimization.

Certainly possible. How often would you expect to have a `summary` without `summary.sig` though? Are you thinking that could happen when something goes wrong on the client side, or are you thinking about repos with summary signing turned off?

alexlarsson:
> Certainly possible. How often would you expect to have a `summary` without `summary.sig` though? Are you thinking that could happen when something goes wrong on the client side, or are you thinking about repos with summary signing turned off?

I don't think it will happen currently as all existing users rely on GPG signatures. However, there has been some arguments that for most users just using TLS/https instead of GPG is probablly pretty ok security-wise. However this is not great in practice currently because the current summary cache:ing is summary.sig based. 

However, I don't personally have any use for this. Flatpak is GPG for the forseable future, because we need the signatures for forwarding trust to the system helper.

Issue #2194, opened 2020-09-12
Add -Werror=shadow to CFLAGS


mwleeds:
I don't have time for it at the moment but perhaps ostree wants to use `-Werror=shadow` as Flatpak is in https://github.com/flatpak/flatpak/pull/3853

There's a note in the `configure.ac` about sharing compiler flags between the two projects.

PR #2195, opened 2020-09-12
Avoid shadowing local variables

mwleeds:
This should help with code readability.

Fixes https://github.com/ostreedev/ostree/issues/2194

### DIFF
--- a/src/ostree/ot-builtin-gpg-sign.c
+++ b/src/ostree/ot-builtin-gpg-sign.c
@@ -171,9 +171,8 @@ delete_signatures (OstreeRepo *repo,
 
       while (!g_queue_is_empty (&signatures))
         {
-          GVariant *child = g_queue_pop_head (&signatures);
-          g_variant_builder_add_value (&signature_builder, child);
-          g_variant_unref (child);
+          g_autoptr(GVariant) sigchild = g_queue_pop_head (&signatures);
+          g_variant_builder_add_value (&signature_builder, sigchild);
         }
 
       g_variant_dict_insert_value (&metadata_dict,
--- a/src/ostree/ot-remote-builtin-add.c
+++ b/src/ostree/ot-remote-builtin-add.c
@@ -104,7 +104,6 @@ ot_remote_builtin_add (int argc, char **argv, OstreeCommandInvocation *invocatio
   g_autoptr(GString) sign_verify = NULL;
   const char *remote_name;
   const char *remote_url;
-  char **iter;
   g_autoptr(GVariantBuilder) optbuilder = NULL;
   g_autoptr(GVariant) options = NULL;
   gboolean ret = FALSE;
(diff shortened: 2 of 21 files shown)


mwleeds:
(Fixed a few shadows in the tests)

src/ostree/ot-builtin-gpg-sign.c
```diff
@@ -171,7 +171,7 @@ delete_signatures (OstreeRepo *repo,
 
       while (!g_queue_is_empty (&signatures))
         {
-          GVariant *child = g_queue_pop_head (&signatures);
+          child = g_queue_pop_head (&signatures);
           g_variant_builder_add_value (&signature_builder, child);
           g_variant_unref (child);
```
cgwalters, line 176: Let's also rename this one rather than increasing the scope.  We could also use `g_autoptr()` here i.e.:
```suggestion
          g_autoptr(GVariant) sigchild = g_queue_pop_head (&signatures);
          g_variant_builder_add_value (&signature_builder, sigchild);
```

src/ostree/ot-remote-builtin-add.c
```diff
@@ -191,7 +191,7 @@ ot_remote_builtin_add (int argc, char **argv, OstreeCommandInvocation *invocatio
                             g_variant_new_variant (g_variant_new_boolean (FALSE)));
     }
 
-  for (char **iter = opt_sign_verify; iter && *iter; iter++)
+  for (iter = opt_sign_verify; iter && *iter; iter++)
```
cgwalters, line 194: Let's instead move the *other* `iter` to be scoped into the other loop.

cgwalters (changes requested):
Cool, thanks for doing this!  Mostly LGTM but in a few cases let's reduce scope rather than increase.

src/libostree/ostree-repo-commit.c
```diff
@@ -4007,7 +4007,8 @@ write_dfd_iter_to_mtree_internal (OstreeRepo                  *self,
       if (!glnx_fstatat (src_dfd_iter->fd, dent->d_name, &stbuf, AT_SYMLINK_NOFOLLOW, error))
         return FALSE;
 
-      g_autoptr(GFileInfo) child_info = _ostree_stbuf_to_gfileinfo (&stbuf);
```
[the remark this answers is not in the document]
mwleeds: Yeah that seems better

src/ostree/ot-builtin-gpg-sign.c
mwleeds: Sure

cgwalters:
Not sure what went wrong with CI there.  Tried restarting it.

jlebon:
There was a CI blip this morning. Should be better now!
/approve

cgwalters (approved):
/lgtm

PR #2197, opened 2020-09-15
checkout: Always fall back to copy on EMLINK

cgwalters:
See https://github.com/coreos/rpm-ostree/issues/2231#issuecomment-692789811

It's debatable if we should even be hardlinking the empty file
at all.  (It's also *very* questionable that Python requires
so many empty files but that's another problem)

Let's just always fall back to copying in this case.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -466,10 +466,18 @@ checkout_file_hardlink (OstreeRepo                          *self,
     }
   else if (linkat (srcfd, loose_path, destination_dfd, destination_name, 0) == 0)
     ret_result = HARDLINK_RESULT_LINKED;
-  else if (!options->no_copy_fallback && (errno == EMLINK || errno == EXDEV || errno == EPERM))
+  else if (!options->no_copy_fallback && (errno == EXDEV || errno == EPERM))
     {
-      /* EMLINK, EXDEV and EPERM shouldn't be fatal; we just can't do the
+      /* In most cases, an ostree user will want hardlinks and will request
+       * it by setting the no_copy_fallback flag.  For the cases where
+       * that's not set, treat failure to hardlink as not-fatal; we just can't do the
        * optimization of hardlinking instead of copying.
+       */
+    }
+  else if (errno == EMLINK)
+    {
+      /* Cases like Python __init__.py being empty can cause us to hit EMLINK.
+       * In this case copying is preferable to not working at all.
        */
     }
   else if (allow_noent && errno == ENOENT)


jlebon:
WDYT about gating this behaviour on the target file having size 0? Maybe it's the right thing to do in the nonzero case but it's less obvious to me than the trivial size 0 case, so I'm thinking we don't do that until a use case actually shows up.

cgwalters:
> WDYT about gating this behaviour on the target file having size 0? 

Right, we've already been there: 673cacd633f9d6b653cdea530657d3e780a41bbd

I guess we could make it unconditional.  The benefits of hardlinking zero sized files seems very marginal to me in general.

jlebon:
> I guess we could make it unconditional.

That sounds reasonable to me. I mean, at this point the file object is really just about holding the permissions and ownership information. We're not really saving any space otherwise (other than the inode space, but presumably that's not something we're making worse vs non-OSTree systems which always copy).

cgwalters:
https://github.com/ostreedev/ostree/pull/2198

But, I don't think it obsoletes this one necessarily.  Particularly if we have reflinks then we should be perfectly happy to use that.  (Or, even use reflinks instead but that's a larger thread)

jlebon:
My concern with this is that it's very "uncharacteristic" of OSTree. An assumption OSTree allows us to make is that checkouts are cheap because it's just hardlinks. If we introduce some way to silently copy files even with `no_copy_fallback`, it breaks that assumption. If we get `EMLINK`, then to be conservative we should assume that there might be a lot more filepaths coming backed by that same file object. If we fallback to copying for all those and it's a nontrivial file size, that could be a lot of unexpected disk writes (and in the worst case, result in `ENOSPC`).

Another way to say this: in the `no_copy_fallback` case, I think it's better to expose users to `EMLINK` than to possibly surprise them by writing X bytes of data to the disk. It likely implies something wrong with their OSTree composes rather than something they actually intend.

wmanley:
> If we get `EMLINK`, then to be conservative we should assume that there might be a lot more filepaths coming backed by that same file object. If we fallback to copying for all those and it's a nontrivial file size, that could be a lot of unexpected disk writes (and in the worst case, result in `ENOSPC`).

One option would be: if you get `EMLINK` then make a copy into `$repo/duplicates/$sha.0.file` and then link from there.  That way we won't get that same cliff-edge behaviour when we go over the maximum number of links.

`ostree gc` would have to be enhanced to remove anything under `$repo/duplicates` that has `st_nlink == 1`, but apart from that there shouldn't be too much additional complexity introduced in other parts of ostree.

jlebon:
> One option would be: if you get `EMLINK` then make a copy into `$repo/duplicates/$sha.0.file` and then link from there. That way we won't get that same cliff-edge behaviour when we go over the maximum number of links.

Yeah, that would work. Though I'm still dubious on whether we really need this at all. The zero-size case is kind of a special one because it's easy to picture multiple packages all shipping zero-sized files for $reasons and those all converging on the same SHA256. It's harder to imagine for > 0. So I still lean more on just not changing anything here until a valid use case shows up.

cgwalters:
Yeah let's revisit this if we hit a non-zero-sized case.

PR #2198, opened 2020-09-16
checkout: Don't hardlink zero sized files

cgwalters:
Alternative to https://github.com/ostreedev/ostree/pull/2197

Python's (usually) zero-sized `__init__.py` files can provoke
us hitting the hardlink limits on some filesystems (`EMLINK`).
At least one Fedora rpm-ostree user hit this.

The benefits of hardlinking here are quite marginal; lots
of hardlinks can behave suboptimally in particular filesystems
like BTRFS too.

This builds on prior code which made this an option, introduced
in https://github.com/ostreedev/ostree/commit/673cacd633f9d6b653cdea530657d3e780a41bbd
Now we just do it uncondtionally.

Also this provoked a different bug in a very obscure user mode checkout
case; when the "real" permissions were different from the "physical"
permissions, we would still hardlink.  Fix the test case for this.

### DIFF
--- a/man/ostree-checkout.xml
+++ b/man/ostree-checkout.xml
@@ -163,7 +163,7 @@ Boston, MA 02111-1307, USA.
                 <option>-z</option></term>
 
                 <listitem><para>
-                    Do not hardlink zero-sized files.
+                    This option does nothing; the functionality is now always on by default.
                 </para></listitem>
             </varlistentry>
 
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -637,8 +637,12 @@ checkout_one_file_at (OstreeRepo                        *repo,
 
       need_copy = FALSE;
     }
-  else if ((options->force_copy_zerosized && is_reg_zerosized) || override_user_unreadable)
+  else if (is_reg_zerosized || override_user_unreadable)
     {
+      /* In https://github.com/ostreedev/ostree/commit/673cacd633f9d6b653cdea530657d3e780a41bbd we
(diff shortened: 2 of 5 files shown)


cgwalters:
OK no we need https://github.com/ostreedev/ostree/pull/2199 first.

cgwalters:
Rebased :surfer: on top of https://github.com/ostreedev/ostree/pull/2199

jlebon (commented):
Let's also mark it as deprecated in the header, and `HIDDEN` in the `checkout` CLI? LGTM otherwise!

cgwalters:
> Let's also mark it as deprecated in the header, 

I don't think we have a way to deprecate bitfield entries, they actually right now also aren't documented individually sadly.  I'm not super worried about marking it as deprecated because we're going to support it forever.

> and HIDDEN in the checkout CLI?

Done.

jlebon:
> I don't think we have a way to deprecate bitfield entries, they actually right now also aren't documented individually sadly.

I meant a `/* Deprecated */` comment the same way a few other deprecated options are marked.

/lgtm

jlebon:
I think `test-libarchive.sh` needs to be adjusted.

```
+ '[' 612737938 == 612737937 ']'
+ fatal 'Files '\''usr/bin/foo0'\'' and '\''usr/bin/bar0'\'' are not hardlinked'
+ echo Files ''\''usr/bin/foo0'\''' and ''\''usr/bin/bar0'\''' are not hardlinked
Files 'usr/bin/foo0' and 'usr/bin/bar0' are not hardlinked
```

cgwalters:
> I think test-libarchive.sh needs to be adjusted.

Thanks, fixed!

PR #2199, opened 2020-09-16
checkout: Ensure copies of unreadable usermode checkouts are readable

cgwalters:
The extreme special case of "zero mode" files like `/etc/shadow`
comes up again.  What we want is for "user mode" checkouts to
override it to make the file readable; otherwise when operating
as non-root without `CAP_DAC_OVERRIDE` it becomes very difficult
to work with.

Previously, we were hardlinking these files, but then it intersects
with *another* special case around zero sized files, which is
*also* true for `/etc/shadow`.

Trying to avoid hardlinking there unveiled this bug - when
we go to do a copy checkout, we need to override the mode.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -613,9 +613,12 @@ checkout_one_file_at (OstreeRepo                        *repo,
     }
 
   const gboolean is_symlink = (g_file_info_get_file_type (source_info) == G_FILE_TYPE_SYMBOLIC_LINK);
+  const guint32 source_mode = g_file_info_get_attribute_uint32 (source_info, "unix::mode");
+  const gboolean is_unreadable = (!is_symlink && (source_mode & S_IRUSR) == 0);
   const gboolean is_whiteout = (!is_symlink && options->process_whiteouts &&
                                 g_str_has_prefix (destination_name, WHITEOUT_PREFIX));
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-repo-checkout.c
```diff
@@ -613,9 +613,12 @@ checkout_one_file_at (OstreeRepo                        *repo,
     }
 
   const gboolean is_symlink = (g_file_info_get_file_type (source_info) == G_FILE_TYPE_SYMBOLIC_LINK);
+  const guint32 source_mode = g_file_info_get_attribute_uint32 (source_info, "unix::mode");
+  const gboolean is_unreadable = (!is_symlink && (source_mode & ~S_IFMT) == 0);
```
jlebon, line 617: Should we be more strict here and also mask out setuid/setgid bits? E.g.

```suggestion
  const gboolean is_unreadable = (!is_symlink && (source_mode & (S_IRWXU | S_IRWXG | S_IRWXO)) == 0);
```

jlebon (approved):
One comment, LGTM otherwise!

src/libostree/ostree-repo-checkout.c
cgwalters: Or, how about just `& S_IRUSR` ?

jlebon: SGTM

cgwalters: Updated with that.

PR #2200, opened 2020-09-22
ot-fs-utils.c: try $TMPDIR if /var/tmp fails

9999years:
In `ot_map_anonymous_tmpfile_from_content` we call `libglnx`'s [`glnx_open_anonymous_tmpfile`][glnx_open_anonymous_tmpfile] function, which attempts to create an fd in `/var/tmp` regardless of the value of `$TMPDIR`.

This is _usually_ okay, but can fail in some contexts, such as in the [NixOS][1] build environment, which doesn't have `/var` mapped at all.

To avoid failing in this case, if `glnx_open_anonymous_tmpfile` fails, we retrieve the value of `$TMPDIR` manually and try calling `glnx_open_anonymous_tmpfile_full` directly (which is equivalent to `glnx_open_anonymous_tmpfile` save for allowing us to choose the directory the fd is created in).

[1]: https://nixos.org/
[glnx_open_anonymous_tmpfile]: https://gitlab.gnome.org/GNOME/libglnx/-/blob/7e3a19958e6af94b764c772a58adf47e6cc4b678/glnx-fdio.c#L305-318

### DIFF
--- a/src/libotutil/ot-fs-utils.c
+++ b/src/libotutil/ot-fs-utils.c
@@ -206,7 +206,30 @@ ot_map_anonymous_tmpfile_from_content (GInputStream *instream,
 {
   g_auto(GLnxTmpfile) tmpf = { 0, };
   if (!glnx_open_anonymous_tmpfile (O_RDWR | O_CLOEXEC, &tmpf, error))
-    return NULL;
+    {
+      /* glnx_open_anonymous_tmpfile hardcodes /var/tmp as a base directory.
+       * If it doesn't exist, such as in a NixOS build or other constrained
+       * systems, glnx_open_anonymous_tmpfile will fail.
+       *
+       * Therefore, we try again using the directory in $TMPDIR if possible.
+       */
+      if ((*error)->code == G_IO_ERROR_NOT_FOUND)
+        {
+          GError *tmp_error;
+          tmp_error = NULL;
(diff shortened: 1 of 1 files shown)


cgwalters:
Hi, thank you for the patch!

This seems like something better fixed in libglnx though, right?  The git for that is here: https://gitlab.gnome.org/GNOME/libglnx
Not opposed to a workaround here though similar to this.

9999years:
Opened [libglnx#20](https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/20).

cgwalters:
Closing this since we have https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/20 merged and I believe that fully covers this.  (if that's not correct please reopen).  Thanks again!

9999years:
Mind cutting a new release so we can downstream this?

cgwalters:
> Mind cutting a new release so we can downstream this?

Somewhat soon, there's a few things that need a bit more work before that I think.

That said: https://github.com/ostreedev/ostree/pull/2208 bumps the submodule, and feel free to cherry pick that libglnx patch downstream if you like.  To get it to apply what I've done in the past is use e.g. `git format-patch -1` on the libglnx commit, then manually edit the diff and change it to `a/libglnx/foo.c` instead of just `a/foo.c`, then it can be applied at the toplevel.

Issue #2154, opened 2020-07-24
v2020.4 triggers assertion failure when upgrading with deployments with device trees


akiernan:
New in v2020.4 compared to v2020.3:

```
root@gvt-441:~# ostree admin deploy 331814d256100fe9dd08b1c75f59b2cb3880877713d9a78af5a81b1a1c1acdaf
[   69.732284] EXT4-fs (mmcblk0p2): re-mounted. Opts:
[   69.737492] ext4 filesystem being remounted at /sysroot supports timestamps until 2038 (0x7fffffff)
Copying /etc changes: 2 modified, 3 removed, 10 added
[   88.382227] EXT4-fs (mmcblk0p2): re-mounted. Opts: (null)
[   88.390336] ext4 filesystem being remounted at /boot supports timestamps until 2038 (0x7fffffff)
**
OSTree:ERROR:../git/src/libostree/ostree-sysroot-deploy.c:1775:install_deployment_kernel: assertion failed (kernel_layout->bootcsum == bootcsum): ("4f3375c34a0e378b04bdfc77d82d75d6f726c4cf065e048f22e9ba7df0617163" == "f71de0cbc10f1dd60423c02c280317742d329a2fc0eba5eeab2282291bde1935")
Bail out! OSTree:ERROR:../git/src/libostree/ostree-sysroot-deploy.c:1775:install_deployment_kernel: assertion failed (kernel_layout->bootcsum == bootcsum): ("4f3375c34a0e378b04bdfc77d82d75d6f726c4cf065e048f22e9ba7df0617163" == "f71de0cbc10f1dd60423c02c280317742d329a2fc0eba5eeab2282291bde1935")
```

Can't immediately see a commit which is causing this, so starting to bisect it.

cgwalters:
Hmm possibly https://github.com/ostreedev/ostree/pull/2001 ?

akiernan:
> Hmm possibly #2001 ?

That was my first thought, but I definitely tested that one at the time (though I probably just brought that commit in rather than the whole tree).

akiernan:
> > Hmm possibly #2001 ?
> 
> That was my first thought, but I definitely tested that one at the time (though I probably just brought that commit in rather than the whole tree).

Clearly my testing was rubbish, because that's the commit that breaks it.

jlebon:
To clarify, this only fails when you're using device trees, right? Or are you seeing failures even when deploying an OSTree without? That would be much worse.

[...]

PR #2202, opened 2020-09-24
 deploy: Remove deployment bootcsum assertion

cgwalters:
When support for devicetree was added, it created a problem
because old and new ostree versions would compute different
checksums for the "boot data".  The scenario here is:

- Have system with ostree < 2020.4
- Reboot into system with ostree 2020.5
- Try to perform an operation that would retain
  that previous booted deployment (common)

Currently ostree iterates over all the deployments
that will be retained and calls `install_deployment_kernel()`,
even for the booted one (which is a bit silly), but
just to verify that all boot data for the targeted
deployments are installed.

This then re-computes the checksum and we'd trip this
assertion.

In practice though, we don't strictly require them to match;
the only thing that will happen if they don't is that we'll
end up with another copy of the kernel/initramfs - and
that only temporarily until the previous deployment
gets GC'd.

Longer term, I think what we really want to do anyways
is probably closer to like a little ostree repo for `/boot`
so that we can e.g. still hardlink kernels there even if
the initramfs changes, or hardlink both kernel/initramfs
if just the devicetree changes, etc.

Closes: https://github.com/ostreedev/ostree/issues/2154

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1028,7 +1028,8 @@ _ostree_kernel_layout_new (void)
 
 /* See get_kernel_from_tree() below */
 static gboolean
-get_kernel_from_tree_usrlib_modules (int                  deployment_dfd,
+get_kernel_from_tree_usrlib_modules (OstreeSysroot       *sysroot,
+                                     int                  deployment_dfd,
                                      OstreeKernelLayout **out_layout,
                                      GCancellable        *cancellable,
                                      GError             **error)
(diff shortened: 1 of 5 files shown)


cgwalters:
OK reworked and marked as ready for review!

cgwalters:
> Currently ostree iterates over all the deployments that will be retained and calls install_deployment_kernel(), even for the booted one (which is a bit silly), but just to verify that all boot data for the targeted deployments are installed.

I only realized in looking at this how this logic is exacerbating the problem - if ostree had been doing the more obvious thing and only calling `install_deployment_kernel()` for deployments that actually need writing to disk, we wouldn't see this.

I don't think we're tracking right now the "needs writing to disk" but it probably wouldn't be hard.

Anyways, that's a separate optimization.

tests/test-osupdate-dtb.sh
```diff
@@ -0,0 +1,55 @@
...
+${CMD_PREFIX} ostree --repo=testos-repo commit --tree=dir=osdata/ -b testos/buildmaster/x86_64-runtime
+${CMD_PREFIX} ostree --repo=sysroot/ostree/repo pull-local --remote=testos testos-repo testos/buildmaster/x86_64-runtime
+${CMD_PREFIX} ostree --repo=sysroot/ostree/repo remote add --set=gpg-verify=false testos file://$(pwd)/testos-repo testos/buildmaster/x86_64-runtime
+${CMD_PREFIX} env OSTREE_SYSROOT_DEBUG=${OSTREE_SYSROOT_DEBUG},no-dtb ostree admin deploy --os=testos testos:testos/buildmaster/x86_64-runtime
+assert_has_file sysroot/boot/ostree/testos-${bootcsum}/vmlinuz-3.6.0
+assert_not_has_file sysroot/boot/ostree/testos-${bootcsum}/dtb/asoc-board.dtb 'a device tree'
+${CMD_PREFIX} ostree --repo=testos-repo commit --tree=dir=osdata/ -b testos/buildmaster/x86_64-runtime
+env OSTREE_SYSROOT_DEBUG=${OSTREE_SYSROOT_DEBUG},no-dtb ${CMD_PREFIX} ostree admin upgrade --os=testos
+${CMD_PREFIX} ostree --repo=testos-repo commit --tree=dir=osdata/ -b testos/buildmaster/x86_64-runtime
+${CMD_PREFIX} ostree admin upgrade --os=testos
```
jlebon, line 53: Let's sanity-check here that the new bootcsum dir has a `dtb/` dir?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1137,37 +1138,41 @@ get_kernel_from_tree_usrlib_modules (int                  deployment_dfd,
...
-  if (fd != -1)
-    {
-      ret_layout->devicetree_srcpath = g_strdup ("devicetree");
-      ret_layout->devicetree_namever = g_strdup_printf ("devicetree-%s", kver);
-      in = g_unix_input_stream_new (fd, FALSE);
-      if (!ot_gio_splice_update_checksum (NULL, in, &checksum, cancellable, error))
-        return FALSE;
-    }
-  else
+  /* Testing aid for https://github.com/ostreedev/ostree/issues/2154# */
```
jlebon, line 1141: Minor/optional: extraneous `#` at the end of link.

jlebon (approved):
Some minor comments, LGTM otherwise!

tests/test-osupdate-dtb.sh
cgwalters: I did that but also added a test that there still wasn't `dtb/` in the *previous* bootcsum but that failed...precisely because of the interaction of https://github.com/ostreedev/ostree/pull/2202#issuecomment-698586616 

I think ultimately it shouldn't be a problem to start installing `dtb/` for previous deployments - that bug is independent of this, but it'd be good to fix too.

cgwalters:
Flake on apt repodata

PR #2205, opened 2020-09-29
Add support for ETag and Last-Modified headers for summary and summary.sig

pwithnall:
See the commit messages. This adds support (in the soup and curl backends) for the `Last-Modified`/`ETag`/`If-None-Match`/`If-Modified-Since` HTTP headers and hooks them up for requests for `summary` and `summary.sig` so they don’t have to be re-downloaded if unchanged from last time. In particular, this will shave 590B off each response for `summary.sig` in the cached (common) case.

This PR is related to #2167 in that it’s a network efficiency improvement, but can be reviewed/merged independently.


pwithnall:
One possible improvement to think about after this (if it gets merged) is that the separate HTTP requests to get `summary.sig` and then (if it’s changed) `summary` could be combined: previously `summary.sig` was essentially acting as the `ETag` for `summary`, but once HTTP ETags are supported properly, then perhaps that could be rethought. If a client has an ETag for its cached `summary`, it could request `summary` from the server immediately (without requesting `summary.sig` first) and rely on the ETag it sends to make sure it only downloads the full `summary` file when it’s actually changed.

That would mean a shift to using inline signatures to protect `summary`, though (otherwise you still need to download `summary.sig` separately). I can’t remember all of the tradeoffs to know whether that’s helpful, unhelpful, or indifferent.

jlebon (commented):
Overall idea makes sense to me. Though worth noting there was a suggestion from @alexlarsson recently to have deltas for summary files as well which I think can affect the discussion here. E.g. if we were to support summary deltas, then presumably they'd be in some form which makes querying for "new summary wrt what I have locally" cheap, which intersects with what's going on here.

Though support for cache semantics makes sense on its own too and is compatible with existing OSTree webservers today.

src/libostree/ostree-fetcher-curl.c
```diff
@@ -575,6 +590,179 @@ write_cb (void *ptr, size_t size, size_t nmemb, void *data)
...
+ *    <day-name>, <day> <month> <year> <hour>:<minute>:<second> GMT
+ *
+ * Note that this only accepts the full-year and GMT formats specified by
+ * RFC 1123. It doesn’t accept RFC 850 or asctime formats.
+ *
+ * Example:
+ *    Wed, 21 Oct 2015 07:28:00 GMT
+ */
+static GDateTime *
+parse_rfc2616_date_time (const char *buf,
```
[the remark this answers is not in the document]
pwithnall: There’s `soup_date_new_from_string()`, but it’s in libsoup and this code is in the libcurl backend, so I assume libsoup is not an option there. Have I misunderstood?

src/libostree/ostree-repo-pull.c
```diff
@@ -2700,6 +2701,66 @@ _ostree_repo_verify_summary (OstreeRepo   *self,
...
+    return;
+
+  if (!glnx_openat_rdonly (self->cache_dir_fd, file, TRUE, &fd, NULL))
+    return;
+
+  if (out_etag != NULL)
+    {
+      g_autoptr(GBytes) etag_bytes = NULL;
+
+      etag_bytes = glnx_fgetxattr_bytes (fd, "user.etag", NULL);
```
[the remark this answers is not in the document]
pwithnall: So declarations not necessarily at the start of code blocks? I thought there were problems with doing that with autocleanup, where the autocleanup would happen but the initialisation might not (if not reached by the time the block is exited)?

src/libostree/ostree-repo-pull.c
```diff
@@ -2700,6 +2701,66 @@ _ostree_repo_verify_summary (OstreeRepo   *self,
...
+            }
+
+          /* Nul-terminate and return */
+          if (buf_len > 0)
+            *out_etag = g_strndup ((const char *) buf, buf_len);
+          else
+            *out_etag = NULL;
+        }
+      else
+        *out_etag = NULL;
```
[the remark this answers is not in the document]
pwithnall: Could do, although it would likely get missed if this function was ever expanded to do error handling. I can change it if you want.

src/libostree/ostree-repo-pull-private.h
```diff
@@ -72,7 +72,11 @@ typedef struct {
   gboolean          has_tombstone_commits;
 
   GBytes           *summary_data;
+  char             *summary_etag;
+  guint64           summary_last_modified;  /* seconds since the epoch */
```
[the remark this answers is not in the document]
pwithnall: I think `ETag` support is actually more important than `Last-Modified` support because it doesn’t have race conditions around updates/checks in the same second, and doesn’t have issues with clock synchronisation. Both require some level of support on the server (typically, aiui, servers are configured to serve everything as if it’s not cacheable, as that’s the safest option). I would expect that all web servers would support both headers, though. They’ve been around a long time.

If xattr support isn’t available then ETags won’t work, but should degrade nicely, so that shouldn’t cause a problem.

src/libostree/ostree-fetcher-curl.c
```diff
@@ -335,7 +341,16 @@ check_multi_info (OstreeFetcher *fetcher)
...
           curl_easy_getinfo (easy, CURLINFO_RESPONSE_CODE, &response);
-          if (!is_file && !(response >= 200 && response < 300))
+
+          if (!is_file && response == 304)
+            {
+              /* Version on the server is unchanged from the version we have
+               * cached locally; report this as an out-argument, a zero-length
+               * response buffer, and no error. */
+              req->out_not_modified = TRUE;
+            }
```
jlebon, line 351: Optional: should we check here that at least one of `if_none_match` or `if_modified_since` was provided? That way, if somehow a HTTP server returns 304 even though we didn't provide either of these, we still error out as before.

jlebon (commented):
Would be nice to have a basic test for this where we check that we avoid pulling the summary. Something like just doing another `ostree pull` and sanity-checking that the inode for the summary didn't change? (Since we always create a new file when caching.) Hmm, though on the `ostree-trivial-httpd` side, briefly looking at `SoupServer`, it seems like we'd have to implement that logic ourselves like [this test here](https://gitlab.gnome.org/GNOME/libsoup/-/blob/4596b4ec4c14a0e473e3fa0c3f41a0107d846cc8/tests/cache-test.c#L55-68).

src/libostree/ostree-fetcher-curl.c
```diff
@@ -575,6 +590,179 @@ write_cb (void *ptr, size_t size, size_t nmemb, void *data)
...
+            guint       n_digits,
+            guint       min,
+            guint       max,
+            guint      *out)
+{
+  guint64 number;
+  const char *end_ptr = NULL;
+  gint saved_errno = 0;
+
+  g_return_val_if_fail (n_digits == 2 || n_digits == 4, FALSE);
```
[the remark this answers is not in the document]
pwithnall: Whoops, good catch. It was used before I refactored the function a bit, and then I forgot to plumb it back in. Will update.

src/libostree/ostree-fetcher-curl.c
pwithnall: Good suggestion, will tweak things to do that (and in the libsoup implementation too).

src/libostree/ostree-repo-pull.c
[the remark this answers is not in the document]
pwithnall: OK. Tweaked a couple of declarations to move them to where they’re used, but most declarations in the PR are assigned by pointer so can’t be changed.

pwithnall:
> Would be nice to have a basic test for this where we check that we avoid pulling the summary

Working on that now; will hopefully have something ready tomorrow.

pwithnall:
Updated. I’ve fixed a couple of bugs and added a unit test (although see its commit message: it doesn’t actually fail with the ETag/Last-Modified changes reverted). All my changes are as fixups for ease of review, and can be squashed later.

src/libostree/ostree-fetcher-curl.c
```diff
@@ -575,6 +591,178 @@ write_cb (void *ptr, size_t size, size_t nmemb, void *data)
...
+  GTask *task = G_TASK (user_data);
+  FetcherRequest *req;
+
+  req = g_task_get_task_data (task);
+
+  if (real_size > strlen ("ETag: ") &&
+      strncasecmp (buffer, "ETag: ", strlen ("ETag: ")) == 0)
+    {
+      g_clear_pointer (&req->out_etag, g_free);
+      req->out_etag = g_strstrip (g_strdup (buffer + strlen ("ETag: ")));
```
cgwalters, line 743: `g_strstrip()` is dangerous here because if the buffer has *leading* whitespace we lose track of the real malloc pointer.  I think you just want `g_strchomp()`?

src/libostree/ostree-fetcher-curl.c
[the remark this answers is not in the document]
cgwalters: Another big thing to keep in mind is that when updating host systems at least ostree usually runs as unrestricted root (I'd eventually like to fix this) and things like the string parsing we're doing here are an attack surface and need careful auditing.

To be clear did you write this code from scratch or is it inspired by anything else?

Let's at least add some unit tests for this? 

We've already copied in some libsoup code for URIs, maybe we could copy in the libsoup code?

src/libostree/ostree-repo-pull.c
```diff
@@ -2772,11 +2834,38 @@ _ostree_repo_load_cache_summary_if_same_sig (OstreeRepo        *self,
...
+static void
+store_file_cache_properties (int         dir_fd,
+                             const char *filename,
+                             const char *etag,
+                             guint64     last_modified)
+{
+  glnx_autofd int fd = -1;
+  struct timeval time_vals[] =
+    {
+      { .tv_sec = last_modified, .tv_usec = 0 },  /* access */
```
cgwalters, line 2846: Why reset the access time versus `UTIME_OMIT`?

src/libostree/ostree-fetcher-soup.c
```diff
@@ -431,6 +438,24 @@ create_pending_soup_request (OstreeFetcherPendingURI  *pending,
...
+    }
+
+  if (SOUP_IS_REQUEST_HTTP (pending->request) && pending->if_modified_since > 0)
+    {
+      glnx_unref_object SoupMessage *msg = soup_request_http_get_message ((SoupRequestHTTP*) pending->request);
+      g_autofree char *mod_date = NULL;
+      g_autoptr(GDateTime) date_time = NULL;
+
+      date_time = g_date_time_new_from_unix_utc (pending->if_modified_since);
+      mod_date = g_date_time_format (date_time, "%a, %d %b %Y %H:%M:%S %Z");
```
cgwalters, line 456: This kind of stuff I think would look more readable in declare-and-initialize style.  (Not a blocker though)

src/libostree/ostree-fetcher-curl.c
```diff
@@ -575,6 +591,178 @@ write_cb (void *ptr, size_t size, size_t nmemb, void *data)
...
+
+  return g_date_time_new_utc (year_int, month_name_index + 1, day_int,
+                              hour_int, minute_int, second_int);
+}
+
+/* CURLOPT_HEADERFUNCTION */
+static size_t
+response_header_cb (const char *buffer, size_t size, size_t n_items, void *user_data)
+{
+  const size_t real_size = size * n_items;
```
cgwalters, line 733: Does libcurl protect us against overflow?  Hmm, [the docs](https://curl.haxx.se/libcurl/c/CURLOPT_HEADERFUNCTION.html) say that "size is always one" so I guess so.

src/libostree/ostree-fetcher-curl.c
```diff
@@ -575,6 +591,178 @@ write_cb (void *ptr, size_t size, size_t nmemb, void *data)
...
+  req = g_task_get_task_data (task);
+
+  if (real_size > strlen ("ETag: ") &&
+      strncasecmp (buffer, "ETag: ", strlen ("ETag: ")) == 0)
+    {
+      g_clear_pointer (&req->out_etag, g_free);
+      req->out_etag = g_strstrip (g_strdup (buffer + strlen ("ETag: ")));
+    }
+  else if (real_size > strlen ("Last-Modified: ") &&
+           strncasecmp (buffer, "Last-Modified: ", strlen ("Last-Modified: ")) == 0)
```
cgwalters, line 746: I think this would read more nicely with a `static char last_modified[] = "Last-Modified: "` somewhere above, would also protect against tyops.  Same for etag.

src/libostree/ostree-fetcher-curl.c
```diff
@@ -575,6 +591,178 @@ write_cb (void *ptr, size_t size, size_t nmemb, void *data)
...
+
+      /* Check termination */
+      if (lm_len < 2 ||
+          lm_buf[lm_len - 2] != '\r' ||
+          lm_buf[lm_len - 1] != '\n')
+        dt = NULL;
+      else
+        dt = parse_rfc2616_date_time (lm_buf, lm_len - 2);
+
+      req->out_last_modified = (dt != NULL) ? g_date_time_to_unix (dt) : 0;
```
cgwalters, line 760: Do we need an `if (req->out_last_modified) g_date_time_unref (req->out_last_modified);` here?

src/libostree/ostree-fetcher-curl.c
```diff
@@ -575,6 +591,178 @@ write_cb (void *ptr, size_t size, size_t nmemb, void *data)
...
+      req->out_etag = g_strstrip (g_strdup (buffer + strlen ("ETag: ")));
+    }
+  else if (real_size > strlen ("Last-Modified: ") &&
+           strncasecmp (buffer, "Last-Modified: ", strlen ("Last-Modified: ")) == 0)
+    {
+      g_autoptr(GDateTime) dt = NULL;
+      const char *lm_buf = buffer + strlen ("Last-Modified: ");
+      size_t lm_len = real_size - strlen ("Last-Modified: ");
+
+      /* Check termination */
```
cgwalters, line 752: Why do we do this in the last-modified case but not etag?

src/libostree/ostree-fetcher-curl.c
pwithnall: Are you sure? `g_strstrip(x)` is just `g_strchug(g_strchomp(x))`, and `g_strchug()` doesn’t increment the input pointer, [it uses `memmove()`](https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/gstrfuncs.c#L2284) to shift the string contents down over any leading whitespace. The pointer is unchanged.

src/libostree/ostree-repo-pull.c
pwithnall: Because I didn’t know about `UTIME_OMIT`. Fixed, thanks

src/libostree/ostree-fetcher-soup.c
pwithnall: Fixed

src/libostree/ostree-fetcher-curl.c
pwithnall: True, fixed

src/libostree/ostree-fetcher-curl.c
pwithnall: Yeah, it’s a bit of a weird API. I’ll add an assertion to be sure

src/libostree/ostree-fetcher-curl.c
pwithnall: No, `req->out_last_modified` is a `guint64`

src/libostree/ostree-fetcher-curl.c
pwithnall: Simplicity of the code; stripping the termination is handled in the ETag case by `g_strstrip()`, and it seemed overkill to check that the last two bytes actually were `\r\n`, given that the set of allowed characters in an ETag is basically everything, and that all we do with it is to store it in an xattr and then regurgitate it. My thinking was that as long as we’re not actually parsing the ETag (and correctly handle it as an arbitrary byte array) it doesn’t need validating, just storing and regurgitating.

I can tighten up the parsing on the ETag if you want, though; I think the HTTP spec is not 100% liberal with it.

src/libostree/ostree-fetcher-curl.c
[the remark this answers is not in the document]
pwithnall: > To be clear did you write this code from scratch or is it inspired by anything else?

It’s written from scratch.

> Let's at least add some unit tests for this?

I’ll have a go.

> We've already copied in some libsoup code for URIs, maybe we could copy in the libsoup code?

Maybe, but it was [rewritten recently](https://gitlab.gnome.org/GNOME/libsoup/-/commit/4596b4ec4c14a0e473e3fa0c3f41a0107d846cc8), so is in basically the same position as this code (except more general and has support for multiple RFC date formats tied together in the same code). The older version of the code was tied to `SoupDate` and also supported multiple date formats in one code path.

I think having a minimal in-house implementation makes sense, given that we only need to support the `Last-Modified` header, which only uses one specific HTTP date format (RFC 2616). If we were supporting more headers with other formats then maybe things would be different.

src/libostree/ostree-fetcher-curl.c
jlebon: Looking at https://stackoverflow.com/a/50180173, it seems like it's possible to have arbitrary amounts of whitespace before the value as well. And that'll throw off the `len != 29` check in the parser. So maybe simpler to just `g_strstrip (g_strdup (...))` here too.

src/libostree/ostree-fetcher-curl.c
```diff
@@ -721,8 +785,32 @@ initiate_next_curl_request (FetcherRequest *req,
...
+      req->req_headers = curl_slist_append (req->req_headers, if_none_match);
+    }
+
+  if (req->if_modified_since > 0)
+    {
+      g_autofree char *mod_date = NULL;
+      g_autoptr(GDateTime) date_time = NULL;
+
+      date_time = g_date_time_new_from_unix_utc (req->if_modified_since);
+      mod_date = g_date_time_format (date_time, "If-Modified-Since: %a, %d %b %Y %H:%M:%S %Z");
```
jlebon, line 802: Could use new style here. ;)

src/libostree/ostree-fetcher-curl.c
```diff
@@ -335,7 +342,17 @@ check_multi_info (OstreeFetcher *fetcher)
...
+          if (!is_file && response == 304 &&
+              (req->if_none_match != NULL || req->if_modified_since > 0))
+            {
+              /* Version on the server is unchanged from the version we have
+               * cached locally; report this as an out-argument, a zero-length
+               * response buffer, and no error. */
+              req->out_not_modified = TRUE;
+            }
+
+          if (!is_file && !(response >= 200 && response < 300) && response != 304)
```
jlebon, line 355: Minor: I think it'd be nice to match the libsoup version of this and use `else if` here. I like that it makes it really obvious that if the first branch is taken, it's not an error condition.

tests/test-rfc2616-dates.c
```diff
@@ -0,0 +1,123 @@
...
+test_ostree_parse_rfc2616_date_time (void)
+{
+#if GLIB_CHECK_VERSION(2, 62, 0)
+G_GNUC_BEGIN_IGNORE_DEPRECATIONS
+  const struct
+    {
+      const char *rfc2616;
+      const char *expected_iso8601;  /* (nullable) if parsing is expected to fail */
+    }
+  tests[] =
```
jlebon, line 41: Nice!

jlebon (approved):
This looks great overall! Really just minor things.

Could you squash the commits? We don't use the Homu bot here anymore which means that there is no autosquashing.

src/libostree/ostree-date-utils.c
```diff
@@ -0,0 +1,166 @@
...
+  for (month_name_index = 0; month_name_index < G_N_ELEMENTS (month_names); month_name_index++)
+    {
+      if (strncmp (month_names[month_name_index], month_name, 3) == 0)
+        break;
+    }
+  if (month_name_index >= G_N_ELEMENTS (month_names))
+    return NULL;
+  if (*(month_name + 3) != ' ')
+    return NULL;
+  if (!parse_uint (year, 4, 0, 9999, &year_int))
```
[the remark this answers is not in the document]
pwithnall: I like to aim higher than [Busted](https://www.youtube.com/watch?v=Tu7HoGZaspo) in all that I do.

src/libostree/ostree-fetcher-curl.c
```diff
@@ -575,6 +592,50 @@ write_cb (void *ptr, size_t size, size_t nmemb, void *data)
...
+static size_t
+response_header_cb (const char *buffer, size_t size, size_t n_items, void *user_data)
+{
+  const size_t real_size = size * n_items;
+  GTask *task = G_TASK (user_data);
+  FetcherRequest *req;
+
+  /* libcurl says that @size is always 1, but let’s check
+   * See https://curl.haxx.se/libcurl/c/CURLOPT_HEADERFUNCTION.html */
+  g_assert (size == 1);
```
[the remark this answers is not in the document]
pwithnall: No, `g_assert()` is appropriate here. The `g_assert_*()` functions are intended for use in unit tests, and [have all sorts of test-specific behaviour](https://gitlab.gnome.org/GNOME/glib/-/blob/master/glib/gtestutils.c#L2885). `g_assert()` does not, and is correctly compiled out if OSTree were to be built with `G_DISABLE_ASSERT`. (I don’t advise doing that, though, it makes debugging real-world bugs a pain.)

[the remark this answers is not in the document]
jlebon: Hmm, we use `g_assert_*()` macros a lot in this codebase (and [rpm-ostree](https://github.com/coreos/rpm-ostree) too).

> The `g_assert_*()` functions are intended for use in unit tests, and have all sorts of test-specific behaviour. `g_assert()` does not, and is correctly compiled out if OSTree were to be built with `G_DISABLE_ASSERT`.

Ahh OK, so the primary difference you're pointing out here is that `g_assert()` is elided with `G_DISABLE_ASSERT` while the other ones are not. Which is true, though OTOH... I don't think anyone is going to come out and tell us that these pesky assertions are blocking them from the undefined behaviour they're looking for :).

Anyway, fine as is too of course. Just wanted to mention they're used extensively elsewhere too.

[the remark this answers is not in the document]
pwithnall: > Just wanted to mention they're used extensively elsewhere too.

Technically OSTree is in the wrong there, but in practical terms if this hasn’t caused problems already then it’s unlikely to :)

src/libostree/ostree-fetcher-curl.c
pwithnall: Done

src/libostree/ostree-fetcher-curl.c
pwithnall: I tried that and it breaks fetches if done in libcurl. Looking more closely, the structure of the code doesn’t match: in libsoup, the `if` block is entirely for error handling, but in libcurl, the first `if` block is for error handling and the subsequent `else if` and `else` blocks are all for success handling. We want to take one of those `else if`/`else` blocks even if `req->out_not_modified` has been set.

So I haven’t made this change.

pwithnall:
All review comments addressed (or commented on) and all the fixups squashed.

src/libostree/ostree-fetcher-curl.c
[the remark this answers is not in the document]
cgwalters: Thanks so much for the added tests!

cgwalters (approved):
/lgtm

cgwalters:
Looks like Travis is stuck, hmm.  Not sure how to retrigger that one without another force push.  Let's give it a bit and see if it comes back, otherwise we can override.

PR #2210, opened 2020-10-06
Add Packit integration

jlebon:
This is a basic `.packit.yaml` integration file which will allow us have
continuous builds of OSTree in cosa and upstream CI. If things go well,
we'll likely deploy this in other build tools like rpm-ostree.

Prompted by wanting to get #2155 out to unblock
https://github.com/coreos/rpm-ostree/pull/2170.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -1,0 +1,11 @@
+# build into f32-coreos-continuous on every commit to master
+jobs:
+    - job: production_build
+      trigger: commit
+      metadata:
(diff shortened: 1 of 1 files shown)


jlebon:
So if I read the Packit docs correctly, I *think* that's all we actually need.

cgwalters:
Sure, happy to try this though as I said I have a long term concern about "yet another system".
/lgtm

PR #2211, opened 2020-10-07
lib/deploy: Don't leak fd when checksumming dtbs

jlebon:
Likely the root of https://bugzilla.redhat.com/show_bug.cgi?id=1886149.


jlebon:
Hmm, or maybe it's clearer to pass ownership to the `GInputStream`. OK, did that now! :arrow_up:

cgwalters (approved):
/lgtm

PR #2212, opened 2020-10-07
ci: Make Packit ignore downstream patches

jlebon:
We don't really carry "Fedora-only" patches in dist-git. So we want to
nuke all the patches which exist there.

Follow-up to #2210.


cgwalters (approved):
It's rather surprising this isn't a first class operation...
/lgtm

jlebon:
This can't break tests.

PR #2215, opened 2020-10-13
deploy: Remove (transfer none) from fd arg

cgwalters:
GI complains.  And in general one needs to assume that file
descriptors aren't stolen.


jlebon:
Ahh, thanks!
/lgtm

Issue #479, opened 2016-08-31
specify additional cmdline to kernel when compose tree


vtolstov:
How can i add some things to kernel cmdline when compose tree?
Or how i control cmdline when boot ostree system?

cgwalters:
`ostree admin deploy --karg=`.  Which is used by e.g. the [Fedora/CentOS Anaconda installer](https://github.com/rhinstaller/anaconda/blob/1a4d17b066d9a0050d04556f03815df12a1209c3/pyanaconda/packaging/rpmostreepayload.py#L321)

vtolstov:
Thanks, i miss it with using rpm-ostree =(.
This is really hard use rpm-ostree without ostree admin =).
But not ostree admin upgrade works fine, because i'm deploy ref via ostree admin deploy and it not complain about unknown refspec

vtolstov:
Last qeustion - i'm add custom cmdline with treefile (dracut additional cmdline and specify kernel cmdline to dracut). Why after composing tree i don't have this options in dracut config and grub entries does not contains this line?

cgwalters:
We don't currently support that, but we could.

cgwalters:
Basically for now though, use the kickstart https://github.com/rhinstaller/pykickstart/blob/master/docs/kickstart-docs.rst#bootloader verb.

paulvt:
Another thing to note is that `--karg` allows one to update or set a kernel boot argument, but not to remove one.
For know, I use `--karg=foo=` to at least clear it, but the relevant part of the kernel does complain it cannot handle the value `''`.

dustymabe:
Marking this as medium difficulty, although I suspect it might lean towards hard. @cgwalters WDYT?

cgwalters:
> Marking this as medium difficulty, although I suspect it might lean towards hard. @cgwalters WDYT?

As a general rule I think for ostree/rpm-ostree, the difficulty level is weighted to a much greater degree by the person's knowledge (e.g. bootloaders, C or Rust, whether it requires learning the test framework etc.) compared to other projects.  I understand the goal of labeling the difficulty level of issues, but I'm not sure it's worth trying to figure out the exact levels here.  If someone wants to dive in they should feel free to ask questions on an issue I'd say.

dustymabe:
[...]

PR #1836, opened 2019-04-08
Add support for default kargs

rfairley:
WIP - will update shortly once commits are tidied up and a test for the default kargs files is added.

Resolves: #479

PR #2217, opened 2020-10-14
Support kargs.d directories for default kargs (rebased)

agners:
This is a rebased version of #1836. The rebase mainly required trivial conflict fixes. There was one larger conflict with https://github.com/ostreedev/ostree/commits/52a6224606e4a057b806225d93a2d0239a004723 which required a bit more in depth changes to preserve/adhere to the idea of that change. All unit tests run by `make check` are passing.

We would like a method to set and update kernel arguments via vanilla OSTree updates. Besides this approach we were also considering using OSTree metadata. But it seems this approach with files in-tree is more flexible. Feedback welcome!

Add the following config directories for setting default kernel
arguments:

/etc/ostree/kargs.d (host config)
/usr/lib/ostree-boot/kargs.d (base config)

These directories contain files whose contents consist of a karg
snippet, which is a collection kernel parameter keys and values.
Example of a snippet:

```
KEY=FOO KEYFLAG ANOTHERKEY=VALUE1 ANOTHERKEY=VALUE2
```

Snippets are read in alphanumeric order of the karg snippet filename
when generating the final kernel options from the host and base config
directories. Ordering may be specified by prefixing a number, e.g.
`/etc/ostree/kargs.d/4000_localhost`.

The bootconfig key `ostree-kargs-generated-from-config` indicates
whether the kargs were generated by ostree from the kargs.d directories
(`true`), or copied from the previous deployment (`false`). Editing the
kargs through the command line (e.g. `ostree admin deploy --karg=`) will
set the flag to `false`, and kargs will be copied from the previous
deployment for subsequent deployments.

Also add support for `ostree admin instutil set-kargs` so that installer
programs using this command (such as Anaconda) remain managed by the
kargs.d directories from the first deployment.

Closes: #479

[rebased to master]
Signed-off-by: Stefan Agner <stefan.agner@toradex.com>

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2709,6 +2709,18 @@ allocate_deployserial (OstreeSysroot           *self,
   return TRUE;
 }
 
+static void
+bootconfig_set_kargs (OstreeBootconfigParser *bootconfig,
+                      const char             *opts,
+                      gboolean                generated)
+{
+  ostree_bootconfig_parser_set (bootconfig, "options", opts);
+  ostree_bootconfig_parser_set (bootconfig,
+                                "ostree-kargs-generated-from-config",
+                                generated ? "true" : "false");
+}
+
+
 void
 _ostree_deployment_set_bootconfig_from_kargs (OstreeDeployment *deployment,
                                               char            **override_kernel_argv)
@@ -2727,6 +2739,7 @@ _ostree_deployment_set_bootconfig_from_kargs (OstreeDeployment *deployment,
       g_autoptr(OstreeKernelArgs) kargs = ostree_kernel_args_new ();
       ostree_kernel_args_append_argv (kargs, override_kernel_argv);
       g_autofree char *new_options = ostree_kernel_args_to_string (kargs);
+      bootconfig_set_kargs(bootconfig, new_options, FALSE);
       ostree_bootconfig_parser_set (bootconfig, "options", new_options);
     }
 }
@@ -2824,9 +2837,241 @@ get_var_dfd (OstreeSysroot      *self,
   return glnx_opendirat (base_dfd, base_path, TRUE, ret_fd, error);
 }
 
+/* Get a GFile* referring to the commit object at ref for the file at path. */
+static GFile*
+get_file_from_repo (OstreeRepo    *repo,
+                    const char    *ref,
+                    const char    *path,
+                    GCancellable  *cancellable,
+                    GError       **error)
+{
+  g_autoptr(GFile) root = NULL;
+  if (!ostree_repo_read_commit (repo, ref, &root,
+                                NULL, cancellable, error))
+    return NULL;
+  return g_file_resolve_relative_path (root, path);
+}
+
+static gboolean
+add_kargs_from_filesystem (GTree         *kargs_configs,
+                           int            deployment_dfd,
+                           const char    *dirpath,
+                           GCancellable  *cancellable,
+                           GError       **error)
+{
+  int kargs_dfd = -1;
+  if (!ot_openat_ignore_enoent (deployment_dfd, dirpath,
+      &kargs_dfd, error))
+    return FALSE;
+  if (kargs_dfd > 0)
+    {
+      g_auto(GLnxDirFdIterator) dfd_iter = { 0, };
+      if (!glnx_dirfd_iterator_init_at (kargs_dfd, ".", TRUE, &dfd_iter, error))
+        return FALSE;
+      while (TRUE)
+        {
+          struct dirent *dent = NULL;
+          if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent,
+                                                           cancellable, error))
+            return FALSE;
+          if (dent == NULL)
+            break;
+
+          if (!(dent->d_type == DT_REG || dent->d_type == DT_LNK))
+            continue;
+
+          const char *karg_name = dent->d_name;
+          if (g_tree_lookup (kargs_configs, karg_name))
+            continue;
+
+          g_autofree char *karg_contents = NULL;
+          if (!(karg_contents = glnx_file_get_contents_utf8_at (kargs_dfd,
+                                                                karg_name, NULL,
+                                                                cancellable,
+                                                                error)))
+            return FALSE;
+
+          g_tree_insert (kargs_configs, g_strdup (karg_name),
+                         g_steal_pointer (&karg_contents));
+        }
+    }
+
+  return TRUE;
+}
+
+static gboolean
+add_kargs_from_commit (OstreeSysroot *sysroot,
+                       GTree         *kargs_configs,
+                       int            deployment_dfd,
+                       const char    *dirpath,
+                       const char    *revision,
+                       GCancellable  *cancellable,
+                       GError       **error)
+{
+  struct stat stbuf;
+  if (!glnx_fstatat_allow_noent (deployment_dfd, dirpath, &stbuf, 0, error))
+    return FALSE;
+  const gboolean dir_exists = (errno == 0);
+
+  if (dir_exists)
+    {
+      g_autoptr(GFile) karg_file = get_file_from_repo (ostree_sysroot_repo (sysroot),
+                                                       revision,
+                                                       dirpath,
+                                                       cancellable,
+                                                       error);
+      g_autoptr(GFileEnumerator) dir_enum = NULL;
+      g_autoptr(GFile) child = NULL;
+      g_autoptr(GFileInfo) child_info = NULL;
+      g_autoptr(GError) temp_error = NULL;
+      dir_enum = g_file_enumerate_children (karg_file,
+                                            OSTREE_GIO_FAST_QUERYINFO,
+                                            G_FILE_QUERY_INFO_NOFOLLOW_SYMLINKS,
+                                            cancellable, error);
+      if (!dir_enum)
+        return FALSE;
+
+      while ((child_info = g_file_enumerator_next_file (dir_enum, NULL,
+                                                        &temp_error)) != NULL)
+        {
+          g_clear_object (&child);
+          const char *karg_name = g_file_info_get_name (child_info);
+          child = g_file_get_child (karg_file, karg_name);
+
+          g_autofree char *karg_contents = NULL;
+          if (!g_file_load_contents (child, cancellable, &karg_contents, NULL,
+                                     NULL, &temp_error))
+            {
+              g_propagate_error (error, g_steal_pointer (&temp_error));
+              return FALSE;
+            }
+          if (karg_contents && !g_tree_lookup (kargs_configs, karg_name))
+            g_tree_insert (kargs_configs, g_strdup (karg_name),
+                           g_steal_pointer (&karg_contents));
+          g_clear_object (&child_info);
+        }
+      if (temp_error)
+        {
+          g_propagate_error (error, g_steal_pointer (&temp_error));
+          return FALSE;
+        }
+    }
+
+  return TRUE;
+}
+
+static gboolean
+collect_karg (gpointer key,
+              gpointer value,
+              gpointer kargs_data)
+{
+  const char *karg_contents = value;
+  OstreeKernelArgs *kargs = kargs_data;
+
+  ostree_kernel_args_parse_append (kargs, karg_contents);
+
+  return FALSE;
+}
+
+static gboolean
+sysroot_regenerate_kargs (OstreeSysroot  *self,
+                          const char     *revision,
+                          int             deployment_dfd,
+                          char          **out_kargs,
+                          GCancellable   *cancellable,
+                          GError        **error)
+{
+  g_autoptr(GTree) kargs_configs = g_tree_new_full ((GCompareDataFunc)g_ascii_strcasecmp,
+                                                    NULL, g_free, g_free);
+
+  if (!add_kargs_from_filesystem (kargs_configs, deployment_dfd,
+                                  _OSTREE_SYSROOT_KARGS_HOST, cancellable,
+                                  error))
+    return FALSE;
+
+  if (revision)
+    {
+      if (!add_kargs_from_commit (self, kargs_configs, deployment_dfd,
+                                  _OSTREE_SYSROOT_KARGS_BASE, revision,
+                                  cancellable, error))
+        return FALSE;
+    }
+  else
+    {
+      if (!add_kargs_from_filesystem (kargs_configs, deployment_dfd,
+                                      _OSTREE_SYSROOT_KARGS_BASE,
+                                      cancellable, error))
+        return FALSE;
+    }
+
+  g_autoptr(OstreeKernelArgs) kargs = ostree_kernel_args_new ();
+  g_tree_foreach (kargs_configs, (GTraverseFunc)collect_karg, kargs);
+  g_autofree char *kargs_contents = ostree_kernel_args_to_string (kargs);
+
+  ot_transfer_out_value (out_kargs, &kargs_contents);
+  return TRUE;
+}
+
+static gboolean
+sysroot_finalize_kargs (OstreeSysroot     *self,
+                        OstreeDeployment  *merge_deployment,
+                        OstreeDeployment  *deployment,
+                        int                deployment_dfd,
+                        const char        *revision,
+                        GCancellable      *cancellable,
+                        GError           **error)
+{
+  OstreeBootconfigParser *bootconfig = ostree_deployment_get_bootconfig (deployment);
+
+  /* If we got an override in this deployment, nothing to do here */
+  if (ostree_bootconfig_parser_get (bootconfig, "options") != NULL)
+    return TRUE;
+
+  /* If we got a merge deployment, check if we should copy the kargs from the
+   * merge deployment.
+   */
+  gboolean kargs_generated_from_config = FALSE;
+  if (merge_deployment)
+    {
+      OstreeBootconfigParser *merge_bootconfig = ostree_deployment_get_bootconfig (merge_deployment);
+      if (merge_bootconfig)
+        {
+          kargs_generated_from_config = g_strcmp0 (ostree_bootconfig_parser_get (merge_bootconfig,
+                                                                                 "ostree-kargs-generated-from-config"),
+                                                   "true") == 0;
+          if (!kargs_generated_from_config)
+            {
+              /* Copy kargs from the merge deployment. */
+              const char *merge_opts = ostree_bootconfig_parser_get (merge_bootconfig,
+                                                                     "options");
+              bootconfig_set_kargs (bootconfig, merge_opts, FALSE);
+            }
+        }
+    }
+
+  /* If we received indication to regenerate kargs from the deployment config,
+   * regenerate the kargs. If there is no merge deployment (e.g. this is the
+   * first deployment), then default to regenerating kargs from this deployment.
+   */
+  if (kargs_generated_from_config || !merge_deployment)
+    {
+      g_autofree char *opts = NULL;
+      if (!sysroot_regenerate_kargs (self, revision, deployment_dfd, &opts,
+                                     cancellable, error))
+        return FALSE;
+      bootconfig_set_kargs (ostree_deployment_get_bootconfig (deployment),
+                            opts, TRUE);
+    }
+
+  return TRUE;
+}
+
+
+
 static gboolean
 sysroot_finalize_deployment (OstreeSysroot     *self,
                              OstreeDeployment  *deployment,
+                             const char        *revision,
                              OstreeDeployment  *merge_deployment,
                              GCancellable      *cancellable,
                              GError           **error)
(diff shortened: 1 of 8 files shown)


wmanley:
This is really neat.  Currently we enable in-tree kargs by having a file `/usr/lib/ostree-boot/kargs.env` that looks like:

```sh
KARGS="console=/dev/ttyS0,115200 ..."
```

with a shell script to do the deploy like this:

```sh
eval "$(ostree cat "$commit_sha" /usr/lib/ostree-boot/kargs.env || true)"
KARGS_ARG="--karg-none $(/usr/bin/printf "--karg-append=%s " $KARGS)"
ostree admin deploy ... $KARGS_ARG "$commit_sha"
```

but an ostree built-in solution would be a lot cleaner.

agners:
Removed a TODO and simplified `tests/libtest.sh` changes a bit.

Also successfully tested this changes in our OE/meta-updater based environment (with changes to meta-updater, see https://github.com/agners/meta-updater/tree/dunfell-kargs.d).

jlebon:
One concern I have about this PR which I feel like we should flesh out a bit first is how updating from older OSTree will work. At least in FCOS/RHCOS, kargs are currently all untracked. I think we'll need some sort of "adoption" scheme where untracked kargs get translated to `/etc/ostree/kargs.d` drop-ins? The challenge there is that there isn't really an easy way to tell what's OEM and what's user-added. Probably safest to assume everything untracked is "user-added", though that might lead to duplicate kargs once composes do start shipping kargs in `/usr/lib/ostree/kargs.d`.

kelvinfan001:
+1 to using an "adoption scheme" where the merge deployment's kargs are translated to drop-ins in `kargs.d`. We can then optimize commands like `ostree admin deploy --karg=` to drop files into `kargs.d`. This way, we no longer have two potential sources of truth and we no longer need the `ostree-kargs-generated-from-config` marker. 
Regarding potential duplicates once composes start shipping kargs in `/usr/lib/ostree/kargs.d`, could we make it so that during the adopting stage, where OSTree looks at the current BLS entry and generates files to drop into `/etc/ostree/kargs.d`, it also checks whether such kargs already exist in `/usr/lib/ostree/kargs.d` and skips those?
Let me know if I am overlooking anything, though!

cgwalters:
I haven't done a deep dive on this yet but the more I think about this what I'd propose is:

- Support `/usr/lib/ostree/kargs.d` (allow OS vendors to change kargs naturally)
- Don't support `/etc/ostree/kargs.d` - instead user config is directly added to the BLS file

IOW we're keeping the idea that the BLS file is a standard that admins and tools can edit directly and not inventing something ostree specific.

cgwalters:
We're having some live discussion on this, other topics:

- In CoreOS at least we need to support platform-specific kargs (`ignition.platform.id`) so there is also a concept of kargs that need to live outside of either ostree or user configuration - for us they're part of the disk image.
- Ordering matters for things like `hugepages=` - we should carefully define how ordering works.  And a specific example of that where this is relevant is the `console=` argument which is messy because it may also cross the "ostree default" and "user configurable" worlds.

I think combining these we are going to be sure we handle the cases of both removing default kargs as well as appending.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2824,34 +2837,250 @@ get_var_dfd (OstreeSysroot      *self,
   return glnx_opendirat (base_dfd, base_path, TRUE, ret_fd, error);
 }
 
+/* Get a GFile* referring to the commit object at ref for the file at path. */
```
kelvinfan001, line 2840: For my understanding, does this mean "Get a GFile* that points to the file at path in the content for ref"?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2824,34 +2837,250 @@ get_var_dfd (OstreeSysroot      *self,
...
+                       GError       **error)
+{
+  struct stat stbuf;
+  if (!glnx_fstatat_allow_noent (deployment_dfd, dirpath, &stbuf, 0, error))
+    return FALSE;
+  const gboolean dir_exists = (errno == 0);
+
+  if (dir_exists)
+    {
+      g_autoptr(GFile) karg_file = get_file_from_repo (ostree_sysroot_repo (sysroot),
```
kelvinfan001, line 2918: Would it be clearer to name `karg_file` something like `karg_dir` to clearly distinguish it from the children files in `usr/lib/ostree-boot/kargs.d` that contain the kargs?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2824,34 +2837,250 @@ get_var_dfd (OstreeSysroot      *self,
...
+          g_autofree char *karg_contents = NULL;
+          if (!g_file_load_contents (child, cancellable, &karg_contents, NULL,
+                                     NULL, &temp_error))
+            {
+              g_propagate_error (error, g_steal_pointer (&temp_error));
+              return FALSE;
+            }
+          if (karg_contents && !g_tree_lookup (kargs_configs, karg_name))
+            g_tree_insert (kargs_configs, g_strdup (karg_name),
+                           g_steal_pointer (&karg_contents));
```
kelvinfan001, line 2950: The way this currently works, if there is a host configured karg in `/etc/ostree/kargs.d`, and the vendor adds the *same* karg at some point in the future during an update but with a different name for the karg file, then there will be a duplicates in the BLS entry. 
Additionally, it is possible that there are multiple host configured kargs all in the _same_ karg file, and when the vendor adds a subset of these kargs at some point during an update, we would have duplicates.

To deal with these two potential problems, perhaps we can make it so that each karg file must only have one karg and the karg KEY must be in the name of the karg file. However, then we would need to have some extra logic to ignore the number preceding the karg file or add the ordering info to the karg file.

kelvinfan001:
Re: https://github.com/ostreedev/ostree/pull/2217#issuecomment-718050080, if we don't support `/etc/ostree/kargs/d`, how would the user remove OEM kargs across updates? Also, if we make users directly edit the BLS entry for configuration, then user configuration would be overridden if the base kargs in `/usr` remove and re-add (or add and remove) the same karg that a user manually added (or removed) in the BLS entry.

cgwalters:
> Re: #2217 (comment), if we don't support /etc/ostree/kargs.d and make users directly edit the BLS entry for configuration, then user configuration would be overridden if the vendor removes and re-adds (or adds and removes) the same karg that a user manually added (or removed) in the BLS entry.

Yeah agree, and thanks for bringing this up.  I think having configuration in one place (the BLS file) just fundamentally conflicts with supporting all of the general corner cases of the systemd style "split configuration" model.

My initial thought here is we just say that overriding vendor kargs isn't supported.  I suspect most people who want this don't even have per-system configuration at all - they are making custom ostree builds that they control and just want a convenient mechanism to add/remove kargs at build time without needing to wrap `ostree admin upgrade` with a script (per above).  So those cases don't care at all about this problem.

But if we do want to support this, we could have `/etc/ostree/kargs.d` *only* support removals of vendor args perhaps?

Or the other approach is to take this problem upstream to the BLS spec; something like
```
options-removed=foo bar
options=console=ttyS0 nosmt
```

And `options-removed` would suppress `/usr/lib/ostree/kargs.d` trying to inject `foo`.

cgwalters:
One angle to look at this from then is: Would we use vendor kargs in CoreOS?  I'm not sure.

I guess one use case I can think of is for RHCOS we added [random.trust_cpu=on](https://github.com/openshift/os/blob/e392f3912741b973b0826bb9b988c7a847ad0255/image.yaml#L11) while we waited for a kernel fix to be backported.  After that's done we can safely remove the karg - so that's a perfect use case for vendor kargs.  And in this case, there is absolutely no sane reason someone would want to turn that argument *off*.

kelvinfan001:
Could we use a third directory `/usr/etc/ostree/kargs.d` for the purpose of storing the "everything in between" default configuration kargs that can't belong in `/usr/lib/ostree-boot/kargs.d`? For example, for FCOS, kargs added by Ignition could be dropped in this directory. This would essentially be a directory that marks the kargs that it contains as explicitly not user configured. Alternatively, perhaps we could put this "marker" in the karg file itself. This way. when a "factory reset" command (`ostree admin deploy --no-merge`) is run, we just need to remove everything in the `/etc` directory (or remove all kargs in `/etc` that have not been marked as "default config").

However, for existing nodes, the "adoption stage" might be a little tricky since it will be difficult for OSTree to distinguish between what is "default" and what is user configured in the existing BLS entry.

jlebon:
Since this is still being discussed, let's
/hold
it.

I've been thinking about this, and I have an idea which I *think* would resolve some of the problems mentioned earlier. The core issue is that by looking at `/proc/cmdline`, it's impossible to tell what kargs came from where and so it's impossible to manage them correctly. Specifically, for OSTree what we really want to know is whether a karg is "managed" or not.

One way we can easily clear this up is by introducing a separator karg like `ostree.kargs` which sits between managed kargs and "everything else". So then roughly the algorithm during updates to generate the new `options` BLS line would be:
1. write kargs from `/usr/lib/ostree/kargs.d` and `/etc/ostree/kargs.d` (with the latter possibly overriding the former)
2. write `ostree.kargs` karg
3. pass through all the kargs after `ostree.kargs` from the source (in the booted sysroot case, that's `/proc/cmdline`)

So then, it becomes safe to upgrade to such an OSTree version because all the preexisting kargs are essentially inherited as "unmanaged".

The idea here is that the source of truth for the *managed* part of the BLS lives in `/usr/lib/ostree/kargs.d` and `/etc/ostree/kargs.d` and users have to work with that. But we still also support unmanaged kargs and don't have to force the world to stop editing the BLS config and moving to the new way (and e.g. `rpm-ostree kargs --append` would still work fine).

Now, for CoreOS at least, it's likely that we wouldn't make the kargs in [`image.yaml`](https://github.com/coreos/fedora-coreos-config/blob/f39fc8d0cb795534f0ad4d473dd8cf99f0ff1579/image.yaml#L5-L10) into vendor kargs because users might have already removed them (IOW, we can't safely make managed what was unmanaged). At least, we'd have to publicize a transition period. We could still use vendor kargs for *new* kargs though, and opening the door to "user kargs" is still fine (and that's what we're really interested in for https://github.com/coreos/fedora-coreos-tracker/issues/318). And of course, other OSes with more locked down configs would still find vendor kargs useful as mentioned above.

jlebon:
(Had this comment ready earlier, but I split it out of https://github.com/ostreedev/ostree/pull/2217#issuecomment-720759328 for clarity.)

> But in both cases, a "factory reset" doesn't seem to be easily doable unless we can reliably distinguish between user per-system configuration, vendor configuration, and everything in between (e.g. Ignition-related or platform-specific kargs in FCOS).

I think the concept of a "factory reset" in practice is going to have to live at a higher-level than OSTree because there are different grades possible (some discussions about that in https://github.com/coreos/fedora-coreos-tracker/issues/399). And as you mentioned, a karg which looks like a "user karg" to OSTree might actually be a "integration/distro karg" (like the Ignition platform ID one). And similarly for files, you may not actually want to nuke all the files in `/etc` which aren't in the commit (e.g. `/etc/fstab` for OSes which still use that). It wouldn't be useful I think to try to establish a classification scheme for these that lives in OSTree.

kelvinfan001:
I also like the idea of "managed" (by OSTree) vs "unmanaged" kargs and using `ostree.kargs` as the separator. It'll essentially make `/etc/ostree/kargs.d` a directory for removing/modifying "managed" kargs _only_. Though one thing that might be a concern is ordering which could matter in some cases.

> don't have to force the world to stop editing the BLS config

Yeah it seems unideal to have an "OSTree-only" way of doing things. But say a user wants to add a host karg, then under this design they would append it in the BLS config directly (not through `/etc/ostree/kargs.d`). But say we then ship the same karg as a vendor karg, then we'd have a duplicate. @lucab mentioned that duplicates could potentially be "valid" and have different meanings (instead of the last one overriding the earlier ones).

> I think the concept of a "factory reset" in practice is going to have to live at a higher-level than OSTree

That makes sense. So OSTree would only "factory reset" the kargs that are managed by OSTree. IOW, we're not going to try to factory reset to whatever kargs Ignition might have set at first boot (if the user modified those kargs).

> we can't safely make managed what was unmanaged

Might be unnecessarily complicating things, but wouldn't we be able to reasonably accurately infer which "managed" kargs (or which kargs we are intending to manage) were modified? E.g. for FCOS, if `systemd.unified_cgroup_hierarchy=0` were missing, then we can be pretty sure that the user would want it removed through `/etc/ostree/kargs.d`.

cgwalters:
> One way we can easily clear this up is by introducing a separator karg like ostree.kargs which sits between managed kargs and "everything else".

Hmm.  It'd work today, but it feels much less clean than adding a proper separate BLS key `options.vendor=` or so.  Also I worry a bit some admin would be like "what's this" and remove it (though this is clearly a "doctor it hurts when I...") situation.

This whole discussion has shades of https://github.com/coreos/rpm-ostree/pull/2170 too - it's quite similar in that we're trying to add a file in `/etc` that is really just a proxy for content in `/boot` that needs synchronization.

I still feel my bottom line on this is: Let's just support the `/usr` variant since that's what the people who submitted this PR want.
Or hmm you know I bet we could make this even stronger: If there is `/usr/lib/ostree/kargs.d`, then we default to using *just* that on upgrades instead of `/proc/cmdline`.

For FCOS/Silverblue style systems we wouldn't use `/usr/lib/ostree/kargs.d` because we need local customization, but we could then take all of these merge issues to a secondary phase.

cgwalters:
Let me verify this: @wmanley do you have any per-device kargs or would it be sufficient for you to have the canonical kargs live in the ostree commit per above?

ricardosalveti:
For iot-based devices (without local customizations), supporting `/usr/lib/ostree/kargs.d` would be quite useful as the same group of devices (based on the same hardware platform) would all have the same kargs.

AdrianVovk:
Doesn't the per-device kernel live in the ostree commit anyway?

cgwalters:
> Doesn't the per-device kernel live in the ostree commit anyway?

Yes...are you saying that e.g. if one is building a custom kernel one can already bake in command line arguments via e.g.
```

config CMDLINE
	string "Default kernel command string"
	default ""
	help
	  On some architectures (EBSA110 and CATS), there is currently no way
	  for the boot loader to pass arguments to the kernel. For these
	  architectures, you should supply some command-line options at build
	  time by entering them here. As a minimum, you should specify the
	  memory size and the root device (e.g., mem=64M root=/dev/nfs).
```
?

cgwalters:
Having `/etc` support would clean up the [MCO](https://github.com/openshift/machine-config-operator/) kargs support for sure; if it can change to writing files in `/etc/ostree/kargs.d` then all the "merge logic" lives in ostree and not the MCO.  I imagine there are other cases like this.

wmanley:
> Let me verify this: @wmanley do you have any per-device kargs or would it be sufficient for you to have the canonical kargs live in the ostree commit per above?

We don't have any per-device kargs because we have a different ostree commit for each device type, so kargs stored under `/usr` would work fine for us.

sinnykumari:
> Having `/etc` support would clean up the [MCO](https://github.com/openshift/machine-config-operator/) kargs support for sure; if it can change to writing files in `/etc/ostree/kargs.d` then all the "merge logic" lives in ostree and not the MCO. I imagine there are other cases like this.

It feels to me that allowing kargs to write as file in /etc/ostree/kargs.d brings additional complexity for MCO. Users can directly drop a file here using MacineConfig and whenever next time ostree kargs gets called kargs will be updated. While in the current approach, MCO allows adding additional kargs only through specifying value in kernelArguments in a MachineConifg.

cgwalters:
> It feels to me that allowing kargs to write as file in /etc/ostree/kargs.d brings additional complexity for MCO. Users can directly drop a file here using MacineConfig and whenever next time ostree kargs gets called kargs will be updated. While in the current approach, MCO allows adding additional kargs only through specifying value in kernelArguments in a MachineConifg.

Right; this would be a new way to write kernel arguments via the MCO (include them in Ignition), however I was more thinking about this as a new *backend* for the current MCO kernel arguments logic.  So it would eventually simplify the code.

That said...to make this whole thread even more twisty, this intersects with https://github.com/openshift/machine-config-operator/issues/1190 and its underlying issue https://github.com/ostreedev/ostree/issues/2220 - fixing that would really be a prerequisite for managing kargs via `/etc` I'd say.

kelvinfan001:
So right now, I see two main proposals for how we can tackle this PR:

**Option 1: Managed vs Unmanaged Kargs**
Summary:
Have a separation between managed and unamanaged kargs using an ostree.kargs karg as the separator sitting in between them. The source of truth of managed kargs lives in /usr/lib/ostree/kargs.d and /etc/ostree/kargs.d. Unmanaged kargs are still supported, where users would directly edit the BLS entry and OSTree won't be involved.
Advantages:
- Provides a standard way to change command line options through OSTree
- Gives OSTree more control over the merging of kargs throughout updates (host configuration will persist and be reconciled with vendor kargs across updates)
Disadvantages:
- Introduces need for a separator karg
- Potential confusion and unexpected behaviour if users use managed kargs and edit the BLS entry directly at the same time (e.g. duplicate kargs, karg ordering issues)
- Possibly need some "adoption" logic (but would also work OK without)

**Option 2: Just support /usr/lib/ostree/kargs.d**
Summary:
If /usr/lib/ostree/kargs.d exists, use that as the single source of truth for all kargs.
Advantages:
- Simpler, less likely to break anything
- Already has what most embedded-type users want
Disadvantages:
- Solves fewer problems (e.g. there would still not be a standard way to change command line options through OSTree)

AdrianVovk:
@cgwalters I meant that having per-device kernel args shouldn't stop you from reading from /usr as the canonical location because the per-device kernel is in /usr anyways

sinnykumari:
> Right; this would be a new way to write kernel arguments via the MCO (include them in Ignition), however I was more thinking about this as a new _backend_ for the current MCO kernel arguments logic. IOW rather than the MCO running `rpm-ostree kargs` it would just write `/etc/ostree/kargs.d/machineconfig`. So it would eventually simplify the code.

yes, that is my understanding as well. 
My main concern is what if user learns about this feature and uses both way to apply kargs:
1. Specifying karnelArguments field in MachinConfig
2. Specifying ignition config  to write a file into /etc/ostree/kargs.d/custom-kargs with some kargs

It is not difficult to handle both cases but creates multiple source of truth to look for all users added kargs. Debugging kargs related issues can get trickier compared to today as we would need to look for any added file into /etc/ostree/kargs.d/

travier:
According to me, both options currently have too many disadvantages:

> **Option 1: Managed vs Unmanaged Kargs**

- I think this option requires an adoption strategy, otherwise we need a transition period and manual admin action which is unwanted from my point of view.
- Having managed kargs come first may cause ordering issues.
- Would `/etc` only mask managed kargs or also override them. In both cases this seems confusing from a user point of view.

> **Option 2: Just support `/usr/lib/ostree/kargs.d`**

- If `/usr` owns all the kargs then we can not have per instance kargs so it's unusable for FCOS.

----

I think we need to stick to a single source of truth. Either it's the BLS config or the configuration in the real root (`/etc` & `/usr`).

AdrianVovk:
So if there's both /usr/lib/ostree/kargs.d, and /etc/ostree/kargs.d:

- On pre-existing systems, on first boot with new ostree the options from the BLS are dumped into /etc, and then a distro-specific tool is invoked that can process that file and decide what to keep and what to strip out/is default in /usr
- either the BLS is wiped on every deploy or the same logic used today with --kargs can be repurposed (as if the config files just add --kargs to the command line)
- ostree loads from both /usr and /etc (/etc masks files in /usr if the name is the same) and then orders the files lexicographically. That's the order the keys appear in (so an admin can easily order keys before/after/I'm between the system-provided keys using the standard ##_whatever.conf pattern).

Alternatively, maybe this logic can be opt-in (or out if rpm-ostree can handle disabling this functionality on upgrade via the repo config) or maybe only enabled for newly-created repos/deployments?

kelvinfan001:
After some discussion with the CoreOS team, we've decided that for the non embedded use case (e.g. FCOS), attempting to create a new OSTree way of shipping vendor kargs while supporting host configured kargs simultaneously would create many problems. For this PR, I believe it should be sufficient to just support `/usr/ostree/kargs.d`, since that would satisfy the embedded use case. IOW, for non embedded distros using OSTree, we would want to keep the current way of doing things; we would simply check for the existence of `/usr/ostree/kargs.d`, and if it does not exist, the BLS would be the source of truth. This would actually simplify the implementation a bit compared to its current state. @agners what do you think? Would you be interested in making those adjustments?

agners:
@kelvinfan001 thanks for the update and sorry for the late reply. I no longer work in the project. Some feedback still from that use case's perspective:

> For this PR, I believe it should be sufficient to just support /usr/ostree/kargs.d, since that would satisfy the embedded use case.

That might work out yes. The initial idea was to separate distro provided kargs (e.g. OS supplier) and project provided kargs (e.g. embedded product team). The way I was thinking to implement it is that the product team uses `/etc` (its default in `/usr/etc`), and the distro provider provides `/usr` (just like any other distro). But the use case could likely also be satisfied by defining file names in `/usr/lib/ostree/kargs.d`. In any case, using files inside the OSTree seems like a much better approach than deployment specified BLS defaults, which then carry on on every deployment. For the embedded use case it would be fine if the whole feature is opt-in IMHO.

I don't plan to work on this personally. If this approach is fine, I think whoever is working on it should open a new pull request with such an implementation.

PR #2218, opened 2020-10-15
travis: Add a 32 bit build

cgwalters:
In the past we've had 32 bit bugs that were caught by
the compiler, let's add this to Travis.


cgwalters:
Motivated by looking at https://bugzilla.redhat.com/show_bug.cgi?id=1888436

jlebon:
/lgtm
though Travis looks unhappy.

jlebon:
Ahh, missing a
/hold cancel

PR #2219, opened 2020-10-15
ostree-prepare-root: print st_dev and st_ino as 64-bit ints

jlebon:
This matches what systemd does and should work fine on all platforms.

Possibly resolves: bugzilla.redhat.com/show_bug.cgi?id=1888436


jlebon:
Wait no, I don't think this is right actually.
/hold

jlebon:
OK digging into the type of `st_ino` is confusing here because it also depends on whether [LFS](https://users.suse.com/~aj/linux_lfs.html) is enabled. Let's just match what systemd does and always cast to `uint64_t`. :arrow_up: 

/hold cancel

cgwalters:
Ah right we're passing a 64 bit value in args but systemd is only shifting 32.

It's unfortunate this isn't caught by the compiler...probably gcc isn't smart enough to understand the systemd journaling API?

cgwalters (approved):
/lgtm
/hold
for manual testing

cgwalters:
/hold cancel
This fix makes sense even if there are other bugs.

Issue #2223, opened 2020-10-26
Using OSTree with Raspberry Pi (Raspberry Pi OS)


goldyfruit:
Hi,

I'm trying to use OSTree with my Raspberry Pi to deploy atomic system updates but so far I'm facing an issue during the `ostree admin deploy`.

Every time I'm trying to deploy an update I'm getting this error:
```
error: Failed to find kernel in /usr/lib/modules, /usr/lib/ostree-boot or /boot
```

I'm using `debootstrap` to build my image and then this image is commit into my repo.

I would like to know if it's possible to use OSTree with Raspberry Pi OS? I guess I'm missing something...
Does anyone ever tried to use OSTree with Raspberry Pi OS?

ricardosalveti:
AFAIK the raspberry pi OS maintains the kernel at the first partition (vfat), which gets loaded right after the boot firmware, so wonder if this error is simply because the kernel is not installed at the rootfs as well.

Check for the kernel package, see if it is deploying it somewhere in the rootfs.

The other issue I see with rpi os is the lack of a standard bootloader (e.g. u-boot), so the standard ostree boot path might not work right.

If you want to use ostree but without the rpi os, recommend you to check https://github.com/advancedtelematic/meta-updater or https://github.com/foundriesio/meta-lmp (https://docs.foundries.io/latest/reference-manual/linux/linux.html).

jlebon:
Fedora IoT also targets the Raspberry Pi: https://docs.fedoraproject.org/en-US/iot/reference-platforms/.

goldyfruit:
@ricardosalveti The `rapsberry-kernel` deploys modules into `/lib/modules` _(as for all Debian like distributions)_ and kernels into `/boot`.
```
/boot/kernel.img
/boot/kernel7.img
/boot/kernel7l.img
/boot/kernel8.img
```

@jlebon I would to stay on Debian like (so far).

Thanks guys :+1:

ricardosalveti:
> @ricardosalveti The `rapsberry-kernel` package deploys modules into `/lib/modules` _(as for all Debian like distributions)_ and kernels into `/boot`.
> 
> ```
> /boot/kernel.img
> /boot/kernel7.img
> /boot/kernel7l.img
> /boot/kernel8.img
> ```

So this is probably the issue, as I believe ostree might be looking for a more standard naming for the kernel.

From src/libostree/ostree-sysroot-deploy.c:

```
[...]

PR #2224, opened 2020-10-26
Bootloader probing and construction refactoring

wmanley:
These are some refactorings I made a little while ago in preparation for adding a new bootloader class, but the new bootloader wasn't necessary after all.  I think the refactorings still have value though.

I'm raising this now so it can be used in the `OstreeBootloaderRpi` implementation.  See discussion on #2223.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -6323,7 +6323,7 @@ ostree_repo_get_default_repo_finders (OstreeRepo *self)
  * Get the bootloader configured. See the documentation for the
  * "sysroot.bootloader" config key.
  *
- * Returns: bootloader configuration for the sysroot
+ * Returns: (transfer none): bootloader configuration for the sysroot
  * Since: 2019.2
  */
 const gchar *
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -1048,7 +1048,6 @@ ostree_repo_finalize (GObject *object)
   g_mutex_clear (&self->txn_lock);
   g_free (self->collection_id);
   g_strfreev (self->repo_finders);
-  g_free (self->bootloader);
 
   g_clear_pointer (&self->remotes, g_hash_table_destroy);
   g_mutex_clear (&self->remotes_lock);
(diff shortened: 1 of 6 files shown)


src/libostree/ostree-repo.c
```diff
@@ -3186,28 +3185,28 @@ reload_sysroot_config (OstreeRepo          *self,
...
-     * binary "x" in /usr/lib/ostree/bootloaders/x). See:
-     * https://github.com/ostreedev/ostree/issues/1719
-     * https://github.com/ostreedev/ostree/issues/1801
-     * Also, dedup these strings with the bootloader implementations
-     */
-    if (!(g_str_equal (bootloader, "auto") || g_str_equal (bootloader, "none")
-          || g_str_equal (bootloader, "zipl")))
-      return glnx_throw (error, "Invalid bootloader configuration: '%s'", bootloader);
+  if (!ot_keyfile_get_value_with_default_group_optional (self->config, "sysroot",
+                                                         "bootloader", "auto",
```
lucab, line 3191: I guess the `"auto"` here can be un-hardcoded in favor of the `CFG_SYSROOT_BOOTLOADER_OPTS_STR[]` entry?

src/libostree/ostree-sysroot.c
```diff
@@ -1337,32 +1364,38 @@ _ostree_sysroot_query_bootloader (OstreeSysroot     *sysroot,
...
+        CFG_SYSROOT_BOOTLOADER_OPT_GRUB2,
+        CFG_SYSROOT_BOOTLOADER_OPT_UBOOT,
+      };
+      for (int i = 0; i < G_N_ELEMENTS (probe); i++)
+        {
+          g_autoptr(OstreeBootloader) bl = _ostree_sysroot_new_bootloader_by_type (
+              sysroot, probe[i]);
+          gboolean is_active = FALSE;
+          if (!_ostree_bootloader_query (bl, &is_active, cancellable, error))
+            return FALSE;
```
lucab, line 1387: I think that here on error you may want to have keep-going logic, in order to continue and query bootloaders further in the array which may be fine, maybe?

src/libostree/ostree-sysroot.c
```diff
@@ -1324,10 +1325,36 @@ ostree_sysroot_repo (OstreeSysroot *self)
   return self->repo;
 }
 
+static OstreeBootloader*
+_ostree_sysroot_new_bootloader_by_type (
+    OstreeSysroot                 *sysroot,
+    OstreeCfgSysrootBootloaderOpt  bl_type)
+{
```
lucab, line 1332: It is my understanding that this function (judging by the calling logic) is not meant to be reached with an `AUTO` bootloader setting, despite using the `OstreeCfgSysrootBootloaderOpt` type which allows that.
If so, it would be good having an explicit assertion (outside of the `default` catch-all switch case) to call out that special case. It would also help casual readers landing here.

src/libostree/ostree-sysroot.c
```diff
@@ -1324,10 +1325,36 @@ ostree_sysroot_repo (OstreeSysroot *self)
...
+    default:
+      g_assert_not_reached ();
+    }
+}
+
 /**
  * ostree_sysroot_query_bootloader:
  * @sysroot: Sysroot
  * @out_bootloader: (out) (transfer full) (allow-none): Return location for bootloader, may be %NULL
+ * @out_bootloader_config: (out) (transfer none) (allow-none): Return location for value of ostree repo config variable sysroot.bootloader, may be %NULL
```
lucab, line 1357: This doesn't seem to exist?

wmanley: Thanks, removed.

src/libostree/ostree-sysroot.c
wmanley: I've added a `case` and a comment.

src/libostree/ostree-repo.c
wmanley: I've tried this and I think `CFG_SYSROOT_BOOTLOADER_OPTS_STR[CFG_SYSROOT_BOOTLOADER_OPT_AUTO]` is a rather awkward way to say "auto". It might make sense to add `ot_keyfile_get_enum_value_with_default` in the future and use it for "repo.mode" too, but I don't think it's worth it right now.

src/libostree/ostree-sysroot.c
wmanley: I've done it this way so as not to change current behaviour.

wmanley:
@lucab: Thanks for the review.

src/libostree/ostree-sysroot.c
lucab: Ah, fair point. It still looks a bit weird to me, but no need to touch it in this PR.

src/libostree/ostree-repo.c
lucab: Ack, understandable, I wasn't sure whether it was a simple leftover, I'm fine to keep it as is.

lucab (approved):
/lgtm

wmanley:
Hmm, not sure why @lucab's lgtm was rejected.  You're in the list of reviewers and your a member of the ostreedev organisation. @cgwalters: any idea?

lucab (approved):
/lgtm

lucab:
My organization membership was not public, thus the bot could not detect it. Updated my orga/profile settings, it seems to be fine now.

wmanley:
> My organization membership was not public, thus the bot could not detect it. Updated my orga/profile settings, it seems to be fine now.

Thanks :)

wmanley:
Oh bother: The fixup!s got merged too.  I was under the impression that the bot took care of rebasing.  Maybe that was the old bot.  I will be more careful in the future.

PR #2225, opened 2020-10-27
lib/fetch-curl: Unref timeout source

jlebon:
The timeout timer should always be one-shot, so let's just always
destroy it in the callback. The main context has its own ref on it, so
it won't be freed behind its back.

This *should* fix a leak that was brought up in
https://bugzilla.redhat.com/show_bug.cgi?id=1891761.

Reported-by: Milan Crha <mcrha@redhat.com>


cgwalters (approved):
/lgtm

PR #2228, opened 2020-11-02
Drop use of `volatile`

jlebon:
As detailed in
https://gitlab.gnome.org/GNOME/glib/-/issues/600#note_877282, volatile
isn't actually needed in these contexts because the atomic operations
already give us strong enough guarantees. In GCC 11, this triggers a
diagnostic due to the volatile qualifier getting dropped anyway.

There is a WIP to do the same in glib:
https://gitlab.gnome.org/GNOME/glib/-/merge_requests/1719

This obsoletes this downstream patch:
https://src.fedoraproject.org/rpms/ostree/c/b8c5a6fb

### DIFF
--- a/src/libostree/ostree-diff.h
+++ b/src/libostree/ostree-diff.h
@@ -42,7 +42,7 @@ typedef enum {
 typedef struct _OstreeDiffItem OstreeDiffItem;
 struct _OstreeDiffItem
 {
-  volatile gint refcount;
+  gint refcount;  /* atomic */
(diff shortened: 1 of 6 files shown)


lucab (commented):
:+1:

pwithnall:
:ok_hand:

pwithnall (approved):
LGTM

cgwalters (approved):
/lgtm

PR #2229, opened 2020-11-02
lib/deltas: Annotate from checksum as nullable

dbnicholson:
Without this you can't create a scratch delta from GI. While here,
switch the deprecated allow-none annotations to nullable.


lucab:
/approve
/hold

Change looks good to me, but I don't know what's the contract around changes in public introspection signatures.

dbnicholson:
> Change looks good to me, but I don't know what's the contract around changes in public introspection signatures.

I think in this case it should be entirely safe since it's just removing an artificial restriction that the C API doesn't impose. You can certainly call `ostree_repo_static_delta_generate` with a `NULL` `from` argument since that's the intended semantics to create a from scratch delta. But I'm fine to wait for someone else to weigh in.

jlebon:
Yeah, this seems safe to me! Thanks!
/lgtm

PR #2230, opened 2020-11-03
pull: Don't save into cache passed in GByte summaries

alexlarsson:
The cache shouldn't be affected by the user passing in some other
summary as it may not be the "official one".

I ran into this in flatpak where the passed summary was correct, but
the re-saving of the cache updated the mtime of the cached file which
led to later http If-Modified-Since calls failing to update.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -4545,7 +4545,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
             }
         }
 
-      if (!summary_from_cache && bytes_summary && bytes_sig)
+      if (!summary_from_cache && bytes_summary && bytes_sig && summary_sig_bytes_v == NULL)
         {
           if (!pull_data->remote_repo_local &&
               !_ostree_repo_cache_summary (self,


src/libostree/ostree-repo-pull.c
```diff
@@ -4545,7 +4545,7 @@ ostree_repo_pull_with_options (OstreeRepo             *self,
             }
         }
 
-      if (!summary_from_cache && bytes_summary && bytes_sig)
+      if (!summary_from_cache && bytes_summary && bytes_sig && summary_sig_bytes_v == NULL)
```
lucab, line 4548: What about turning this into a `gboolean summary_sig_from_cache` and setting it to `TRUE` in the same place where the bytes are fetched from the GVariant? That would be symmetrical to `summary_from_cache` which already undergoes similar logic, and would make clearer what's going on (yes, I'm almost got lost reading keeping track of the logic in this functions).

alexlarsson: `summary_sig_bytes_v != NULL` isn't really about the cache though. Its about if the user passed in a custom summary. 

So, neither the summary or the summary.sig came from the cache, yet we should not write anything to the cache because they are not from the canonical location.

alexlarsson: But, yes, this function is way to complex, and the naming of summary_sig_bytes_v is perhaps not optimal.

cgwalters (approved):
/lgtm

PR #2233, opened 2020-11-06
sysroot: Fix up some GI nullable annotations

cgwalters:
Hit `ostree_sysroot_repo()` shouldn't be nullable while using
the ostree Rust bindings.


cgwalters:
Travis networking flake

PR #2236, opened 2020-11-10
ci/travis: move to newer base distro

lucab:
This removes the old pinned distro (Ubuntu Trusty 14.04) from Travis,
moving to the newer default distro (Ubuntu Xenial 16.04).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: c
-dist: trusty
 sudo: required
 
 env:


lucab:
I'm not sure whether this is going to work without further tweaks to the scripts (at which point it may be better to directly jump to github actions), but it's worth a try as otherwise all PRs are red right now.

jlebon:
Nice, thanks for the fix!

/lgtm

lucab:

We only care about Travis being green on this one.

PR #2239, opened 2020-11-16
ci: run ci-release-build.sh on GitHub

lucab:
This adds a GitHub action in order to run ci-release-build.sh on
release PRs (detected via the `kind/release` label).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,27 @@
+---
+name: Release sanity
+
+on:
+  pull_request:
+    branches: [master]
+    types: [labeled]
+
+jobs:
+  ci-release-build:
+    name: "Sanity check release commits"
+    if: ${{ github.event.label.name == 'kind/release' }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+          fetch-depth: '0'
+      - name: Checkout (HEAD)
+        run: git checkout HEAD
+      - name: Check release sanity (HEAD)
+        run: ci/ci-release-build.sh
+      - name: Checkout (HEAD^)
+        run: git checkout HEAD^
+      - name: Check release sanity (HEAD^)
+        run: ci/ci-release-build.sh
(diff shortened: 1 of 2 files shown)


lucab:
This is how it looks on my fork (on a non-release branch): https://github.com/lucab/ostree/runs/1405983142?check_suite_focus=true

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+
+on:
+  pull_request:
+    branches: [master]
+    types: [labeled]
+
+jobs:
+  ci-release-build:
+    name: "Sanity check release commits"
+    if: ${{ github.event.label.name == 'kind/release' }}
```
jlebon, line 12: Is this a label a human would add? Is it possible to just make it queue it off of the commit title being "Release ..." instead?

jlebon (approved):
Cool idea! Just one question, but LGTM as is too.

.github/workflows/release.yml
lucab: Yes, it would need to be manually added. I'll check to see if the latter is possible.

cgwalters: Hum...configuring which CI to run via labels sounds really nice as a general pattern.  For example it'd be nice to label things "ci/trivial" to skip the expensive tests.  But we'd need to plumb that through various systems to make it work.

ci/ci-release-build.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
cgwalters, line 1: Why out of curiosity?

lucab: https://nixos.wiki/wiki/NixOS#Comparison_with_traditional_Linux_Distributions

Not really relevant here, I can drop that if preferred. I seldomly use `nix` myself, but it's more like a good habit.

cgwalters (approved):
/lgtm

.github/workflows/release.yml
lucab: @jlebon I've scouted GH docs a bit, and I think I found all the primitives I need to put that together. I'll try gluing them together in a separate repo and report back if it works.

jlebon: Having it be a manual step WFM as long as we document it somewhere (hmm, I think we should have something like https://github.com/coreos/rpm-ostree/blob/master/docs/RELEASE.md for ostree?)

PR #2240, opened 2020-11-16
workflows/release: pattern-match on PR title

lucab:
This adds an additional condition in order to sanity check all
PRs starting with `Release` (case-insensitive).

Ref: https://github.com/ostreedev/ostree/pull/2239#discussion_r524422740

/cc @jlebon


jlebon:
Cool!
/lgtm

PR #2241, opened 2020-11-17
Release 2020.8

lucab:
This release mostly contains scalability improvements and bugfixes.

Caching-related HTTP headers are now supported on summaries and signatures, so that they do not have to be re-downloaded if not changed in the meanwhile.

Summaries and delta have been reworked to allow more fine-grained fetching.
It is now possible to store deltas in detached metadata outside of summary files, so that only relevant ones can be pulled when downloading a particular commit.
In particular, deltas can now be stored in a separate directory indexed by target commit, thus grouping the subset of deltas affecting it. These indexes are updated when the summary is updated and the in-summary delta index would normally be updated.

Related to the above, a new core option has been added to drop the deltas from the summary. However, as that would break older versions looking for the deltas there, it is off by default. 

Finally, this fixes several bugs related to atomic variables, HTTP timeouts, and 32-bit architectures.

---

```
Alexander Larsson (16):
      deltas: Add _ostree_get_relative_static_delta_index_path()
      deltas: Add ostree_repo_list_static_delta_indexes() function
      deltas: Update delta indexes when updating summary
      deltas: Add and document no-deltas-in-summary config option
      deltas: Make ostree_repo_static_delta_reindex() public
      deltas: Add CLI ops to list and reindex delta-indexes
      deltas: Use delta indexes when pulling
      deltas: Add tests for delta indexes
      deltas: Take a shared repo lock while reindexing deltas
      deltas: Set `indexed-deltas` key in the config and summary
      pull: Only download summary if we need it for the pull operation
      tests: Add a testcase to ensure we're not using the summary if we don't need it
      Add ostree_repo_gpg_sign_data()
      ostree pull: Add more g_debug spew around fetching deltas
      ostree_repo_gpg_sign_data: Fix API doc argument name
      pull: Don't save into cache passed in GByte summaries

Colin Walters (7):
      Post-release version bump
      deploy: Remove (transfer none) from fd arg
      travis: Add a 32 bit build
      sysroot: Fix up some GI nullable annotations
      bin/checkout: Port some to new style
      deployment: Add a bunch of docs and fix annotations
      deployment: Ensure query_deployments_for returns nullable values

Dan Nicholson (1):
      lib/deltas: Annotate from checksum as nullable

Felix Krull (1):
      lib: fix GI parameter tags

Jonathan Lebon (4):
      ostree-prepare-root: print st_dev and st_ino as 64-bit ints
      lib/fetcher-curl: Use G_SOURCE_REMOVE instead of FALSE
      lib/fetch-curl: Unref timeout source
      Drop use of `volatile`

Kelvin Fan (1):
      docs: Fix various typos

Luca BRUNO (5):
      ci/travis: move to newer base distro
      ci: run ci-release-build.sh on GitHub
      workflows/release: pattern-match on PR title

Philip Withnall (5):
      libostree: Add support for ETag and Last-Modified headers
      lib/pull: Hook up HTTP caching headers for summary and summary.sig
      tests: Add simple test for summary file caching
      ostree/trivial-httpd: Add Last-Modified/ETag support
      tests: Split RFC 2616 date parsing code out and add tests

William Manley (8):
      ostree_repo_get_bootloader: Document transfer none
      Refactor: Centralise choosing the appropriate bootloader
      Refactor: sysroot.bootloader: Store enum value rather than string
      Add support for explicitly requesting any specific bootloader type
      Refactor `ostree_sysroot_query_bootloader`
      Tests: Refactor bootloader-entries-crosscheck
```

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -184,9 +184,12 @@ libostree_1_la_SOURCES += \
 endif # USE_GPGME
 
 symbol_files = $(top_srcdir)/src/libostree/libostree-released.sym
-if BUILDOPT_IS_DEVEL_BUILD
-symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
-endif
+
+## Uncomment this include when adding new development symbols.
+#if BUILDOPT_IS_DEVEL_BUILD
+#symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
+#endif
+
 # http://blog.jgc.org/2007/06/escaping-comma-and-space-in-gnu-make.html
 wl_versionscript_arg = -Wl,--version-script=
 EXTRA_DIST += \
--- a/configure.ac
+++ b/configure.ac
@@ -7,10 +7,10 @@ dnl Seed the release notes with `git-shortlog-with-prs <previous-release>..`. Th
 dnl `git-evtag` to create the tag and push it. Finally, create a GitHub release and attach
 dnl the tarball from `make dist`.
 m4_define([year_version], [2020])
-m4_define([release_version], [8])
+m4_define([release_version], [9])
 m4_define([package_version], [year_version.release_version])
 AC_INIT([libostree], [package_version], [walters@verbum.org])
-is_release_build=yes
+is_release_build=no
 AC_CONFIG_HEADER([config.h])
 AC_CONFIG_MACRO_DIR([buildutil])
 AC_CONFIG_AUX_DIR([build-aux])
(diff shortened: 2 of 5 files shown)


lucab:
/hold

I will hold-cancel once the release-notes have been finalized too.

alexlarsson:
So far it looks good to me.

jlebon:
PR and notes LGTM. Looks like the release sanity check GitHub action isn't working (both checks say `OK: commit is not a release`, which seems wrong). We don't have to block on figuring that out though.

Will let @cgwalters take a look.

/approve

lucab:
@jlebon indeed, the action is partly wrongly because it acts on the resulting merge, not directly on the PR-branch. Post-release I'll need to update it this way: https://github.com/actions/checkout#checkout-pull-request-head-commit-instead-of-merge-commit

cgwalters (approved):
/lgtm

lucab:
Tagged: https://github.com/ostreedev/ostree/releases/tag/v2020.8

Issue #2245, opened 2020-11-22
test-pull-summary-sigs.sh fails when run on a tmpfs


smcv:
Versions: libostree 2020.8 or git master (commit 024b169a tested); Debian unstable rolling release as of 2020-11-22.

libostree 2020.8 failed to build with test failures on most of Debian's official autobuilders (all except mips*el). I believe this is correlated with the tests being run with `/var/tmp` on a tmpfs: my understanding is that for architectures where enough RAM or RAM+swap is available, Debian is moving towards doing production builds in a chroot hosted on a tmpfs, in order to make dependency installation and build happen as fast as possible.

I was able to reproduce this by building libostree on a virtual machine with small loopback filesystems mounted on `/media/btrfs`, `/media/ext4` and `/media/xfs`, and a tmpfs on `/media/tmpfs`, then running commands like:

```
make check TESTS=tests/test-pull-summary-sigs.sh TEST_TMPDIR=/media/xfs
```

The test repeatably passed (5/5 consecutive runs) on btrfs, ext4 and xfs, but repeatably failed (5/5 consecutive runs) on tmpfs.

[test-suite.log](https://github.com/ostreedev/ostree/files/5579509/test-suite.log)

smcv:
It appears to be possible to fix this by using touch(1) to ensure that the mtime of the `summary` and `summary.sig` changes every time they are overwritten.

PR #2246, opened 2020-11-22
test-pull-summary-sigs: Set timestamps to serve expected files

smcv:
If this is not done, the test can fail when the temporary directory is
a tmpfs: for example this happens during build-time testing with /var/tmp
on tmpfs or TEST_TMPDIR pointing to a tmpfs, or installed-tests with
gnome-desktop-testing-runner allocating the test directory on a tmpfs.

In particular, many of Debian's official autobuilders now do the entire
build and test procedure in a chroot hosted on a tmpfs, to improve build
performance and prevent fsync overhead.

In this situation, it appears that overwriting summary.sig with a copy
of summary.sig.2 is not sufficient for the web server to tell the
libostree client that it needs to be re-downloaded. I'm not completely
sure why, because tmpfs does appear to have sub-second-resolution
timestamps, but forcing a distinct mtime is certainly enough to
resolve it.

Resolves: https://github.com/ostreedev/ostree/issues/2245  
Bug-Debian: https://bugs.debian.org/975418  

### DIFF
--- a/tests/test-pull-summary-sigs.sh
+++ b/tests/test-pull-summary-sigs.sh
@@ -175,6 +175,8 @@ cd ${test_tmpdir}
 # Reset to the old valid summary and pull to cache it
 cp ${test_tmpdir}/ostree-srv/gnomerepo/summary{.1,}
 cp ${test_tmpdir}/ostree-srv/gnomerepo/summary.sig{.1,}
+touch -t 200101010101 ${test_tmpdir}/ostree-srv/gnomerepo/summary
+touch -t 200101010101 ${test_tmpdir}/ostree-srv/gnomerepo/summary.sig
 repo_reinit
 ${OSTREE} --repo=repo pull origin main
 assert_has_file repo/tmp/cache/summaries/origin
@@ -186,6 +188,7 @@ cmp repo/tmp/cache/summaries/origin.sig ${test_tmpdir}/ostree-srv/gnomerepo/summ
 # summary signature since it was generated on the server between the
 # requests
 cp ${test_tmpdir}/ostree-srv/gnomerepo/summary.sig{.2,}
+touch -t 200202020202 ${test_tmpdir}/ostree-srv/gnomerepo/summary.sig
 if ${OSTREE} --repo=repo pull origin main 2>err.txt; then
     assert_not_reached "Successful pull with old summary"
 fi
@@ -197,6 +200,7 @@ cmp repo/tmp/cache/summaries/origin.sig ${test_tmpdir}/ostree-srv/gnomerepo/summ
 
 # Publish correct summary and check that subsequent pull succeeds
 cp ${test_tmpdir}/ostree-srv/gnomerepo/summary{.2,}
+touch -t 200202020202 ${test_tmpdir}/ostree-srv/gnomerepo/summary
 ${OSTREE} --repo=repo pull origin main
 assert_has_file repo/tmp/cache/summaries/origin
 assert_has_file repo/tmp/cache/summaries/origin.sig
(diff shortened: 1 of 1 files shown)


dbnicholson:
I wonder if this has more to do with the web server part because the code is relying on the HTTP `Last-Modified` header, which only has second resolution. So, the issue, I believe, is that the builders are going too fast in your case. Another way to fix would be to throw in a `sleep 1` before the copies.

smcv:
> I wonder if this has more to do with the web server part because the code is relying on the HTTP Last-Modified header, which only has second resolution

Quite possibly it's that, and the difference between tmpfs and the on-disk filesystems is literally just that tmpfs is so fast.

(However, my test VM itself had its disk image stored on tmpfs, so it should have been nearly as fast.)

> Another way to fix would be to throw in a sleep 1 before the copies.

I'd prefer to force the timestamps as I've done here, without artificially slowing down the test, because slower tests make the implement/test/debug/fix cycle slower.

dbnicholson:
> I'd prefer to force the timestamps as I've done here, without artificially slowing down the test, because slower tests make the implement/test/debug/fix cycle slower.

That's fine, but it would be nice if you could update the commit message if that's the case. You could experiment by using the previous file as a reference so that the modification time is unchanged. Like `touch -r summary.1 summary.2` early in the process to force them to have the same modification times.

jlebon:
See also discussions about this recently in https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/21.

Hmm, I think another way to solve this is to use `cp -a` when copying summary files so that timestamps are preserved, right? (Well, I guess we'd still need at least one `touch` to make `summary.1` older.)

wmanley:
I'm a bit confused by this.  It seems that both the soup and the curl fetchers set both the `If-None-Match` and `If-Modified-Since`. And ostree-trivial-httpd prefers checking the Etag to see if things have changed - so I'm not sure why the mtime should matter.

smcv:
> I think another way to solve this is to use cp -a when copying summary files so that timestamps are preserved, right? (Well, I guess we'd still need at least one touch to make summary.1 older.)

Right, we could force the mtime of `summary.1` and `summary.sig.1` to differ from that of `summary.2` and `summary.sig.2` at a granularity of at least 1 second, and then use `cp -a` or `cp -p` instead of `cp` to make sure that's carried over to `summary`. The diffstat would be fairly similar.

> It seems that both the soup and the curl fetchers set both the If-None-Match and If-Modified-Since. And ostree-trivial-httpd prefers checking the Etag to see if things have changed - so I'm not sure why the mtime should matter.

Sorry, your guess is as good as mine for this. Maybe it *shouldn't* matter - and I'm surprised we get such consistently different results for tmpfs and the other three filesystems I tried - but the result of my tests is that it *does* matter.

On one hand, it's unsatisfying to merge something like this without being able to say exactly why it's a solution; but on the other hand, the different versions of the summary having mtimes that differ by at least a second seems like a more realistic simulation of real-world uses of libostree, and we can't gate acceptance into distros on tests passing if the tests don't reliably pass.

smcv:
> both the soup and the curl fetchers

FWIW, we're using the libsoup fetcher in Debian at the moment, although we could transition to libcurl if the libostree maintainers strongly prefer that one.

smcv:
Vaguely related, I have wondered whether the summary and the signature should be a single file in order to bypass cache-coherency issues completely: see <https://github.com/flatpak/flatpak/issues/3983> (sorry, at the time I opened that issue I thought that was a piece of design done in Flatpak).

cgwalters (approved):
We can do this for now and revisit any of the improved suggestions in followups.
/lgtm
Thanks for the PR!

PR #2253, opened 2020-12-17
pull: Allow disabling commit binding verification

dbnicholson:
In some cases such as backups or mirroring you may want to pull commits
from one repo to another even if there commits that have incorrect
bindings. Fixing the commits in the source repository to have correct
bindings may not be feasible, so provide a pull option to disable
verification.

For Endless we have several repositories that predate collection IDs and
ref bindings. Later these repositories gained collection IDs to support
the features they provide and ref bindings as the ostree tooling was
upgraded. These repositories contain released commits that were valid to
the clients they were targeting at the time. Correcting the bindings is
not really an option as it would mean invalidating the repository
history.

### DIFF
--- a/bash/ostree
+++ b/bash/ostree
@@ -849,6 +849,7 @@ _ostree_pull_local() {
         --gpg-verify-summary
         --require-static-deltas
         --untrusted
+        --disable-verify-bindings
     "
 
     local options_with_args="
@@ -904,6 +905,7 @@ _ostree_pull() {
         --untrusted
         --bareuseronly-files
         --dry-run
+        --disable-verify-bindings
     "
 
     local options_with_args="
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1620,15 +1620,17 @@ scan_commit_object (OtPullData                 *pull_data,
   if (!ostree_repo_load_commit (pull_data->repo, checksum, &commit, &commitstate, error))
     return FALSE;
 
-  /* If ref is non-NULL then the commit we fetched was requested through the
-   * branch, otherwise we requested a commit checksum without specifying a branch.
-   */
-  g_autofree char *remote_collection_id = NULL;
-  remote_collection_id = get_remote_repo_collection_id (pull_data);
-  if (!_ostree_repo_verify_bindings (remote_collection_id,
-                                     (ref != NULL) ? ref->ref_name : NULL,
-                                     commit, error))
-    return glnx_prefix_error (error, "Commit %s", checksum);
+  if (!pull_data->disable_verify_bindings) {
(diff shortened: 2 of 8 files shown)


bash/ostree
```diff
@@ -849,6 +849,7 @@ _ostree_pull_local() {
         --gpg-verify-summary
         --require-static-deltas
         --untrusted
+        --disable-verify-bindings
```
lucab, line 852: Just wondering, did you consider making this a `--verify-bindings=<true|false>`, defaulting to `true`?
I think that would be slightly more aligned with `ostree fsck` current option, although I'm not use if the usecase is the same.

dbnicholson: I hadn't considered that since it didn't match any of the other options there. It's exactly the same use case as fsck. Our repos also fail that 😁

In the case of fsck it's off by default. For pulling it's on by default. I don't believe the fsck option does that kind of handling, but maybe I missed it.

lucab: My personal taste would be to turn this PR so that at the end of the day you can do `ostree pull --verify-bindings=false`, but I'd be happy to hear somebody else's opinion on the UX here.

jlebon: AFAIK, `--flag=<true|false>` isn't really easy with glib. There's no built-in support for it, so you'd have to e.g. implement your own callback or parse the string yourself. @dbnicholson is right that it doesn't match any other options we have right now. Probably better to just stick with the same paradigm.

cgwalters: Yep, I agree that is a better UX, but improving the CLI in this aspect should probably be a distinct effort.

src/libostree/ostree-repo-pull.c
```diff
@@ -1620,15 +1620,17 @@ scan_commit_object (OtPullData                 *pull_data,
...
-  /* If ref is non-NULL then the commit we fetched was requested through the
-   * branch, otherwise we requested a commit checksum without specifying a branch.
-   */
-  g_autofree char *remote_collection_id = NULL;
-  remote_collection_id = get_remote_repo_collection_id (pull_data);
-  if (!_ostree_repo_verify_bindings (remote_collection_id,
-                                     (ref != NULL) ? ref->ref_name : NULL,
-                                     commit, error))
-    return glnx_prefix_error (error, "Commit %s", checksum);
+  if (!pull_data->disable_verify_bindings) {
```
cgwalters, line 1623: (Though I just noticed this doesn't match the GNU coding style but eh, we can fix in a followup)

dbnicholson: Oops, sorry. I actually like the GNU style, but was in a different place before running into this issue.

dbnicholson: See #2254

Issue #2255, opened 2021-01-06
OSTree extlinux.conf on Nvidia Jetson Xavier


halverneus:
My goal is to create an image that can be flashed to an Nvidia Jetson Xavier device that is based on OSTree. I use the script, below, to create the rootfs. According to the [Nvidia Documentation](https://docs.nvidia.com/jetson/l4t/index.html#page/Tegra%2520Linux%2520Driver%2520Package%2520Development%2520Guide%2Fbootflow_jetson_xavier.html%23wwconnect_header) the device requires the kernel, devicetree and initrd to be referenced by <rootfs>/boot/extlinux/extlinux.conf. After running the below script, this file does not exist.

My questions are:
- [ ] Does OSTree currently support using extlinux.conf as the boot configuration file?
- [ ] If not, can I get a pointer into the OSTree code where I might be able to implement this change (perhaps with minor guidance)?
- [ ] If so, what am I not doing right or misunderstanding about creating the rootfs?

```bash
#!/bin/bash

set -euo pipefail # Exits script on any error.

# To run this script a few base packages are required. Run the following command
# to install this base packages:
#
# apt install -y build-essential debootstrap qemu-user-static ostree
#
# Requires file Jetson_Linux_R32.4.4_aarch64.tbz2 to be in the same directory as
# this setup.sh script.

HOMEDIR="/home/$USER"
echo "Will be using SSH key from the following home directory: $HOMEDIR"

# Start fresh and remove any existing builds. Existing root filesystems may have
# the immutable flag set, so the flag needs to be cleared before the directory
# can be removed.
sudo chattr -i out/Linux_for_Tegra/rootfs/ostree/deploy/mine/deploy/* || /bin/true
sudo rm -rf out || /bin/true
mkdir out
cd out

# Extract the Jetson tools
tar -xf ../Jetson_Linux_R32.4.4_aarch64.tbz2
cd Linux_for_Tegra

# Build the filesystem.
sudo qemu-debootstrap --arch=arm64 bionic rootfs

sudo /bin/bash -c "echo \"deb http://ports.ubuntu.com/ubuntu-ports bionic main universe\" > rootfs/etc/apt/sources.list"

[...]

PR #2256, opened 2021-01-07
Bootloaders: Support for U-boot in extlinux.conf compatibility mode

wmanley:
Supports the bootloader that comes with the NVidia Tegra TK-1 and NX.

See also: #2255

TODO:

- [ ] Post-rebase testing

### DIFF
--- a/src/libostree/ostree-bootloader-syslinux.c
+++ b/src/libostree/ostree-bootloader-syslinux.c
@@ -26,6 +26,7 @@
 #include <string.h>
 
 static const char syslinux_config_path[] = "boot/syslinux/syslinux.cfg";
+static const char extlinux_config_path[] = "boot/extlinux/extlinux.conf";
 
 struct _OstreeBootloaderSyslinux
 {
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-bootloader-syslinux.c
```diff
@@ -124,7 +133,19 @@ _ostree_bootloader_syslinux_write_config (OstreeBootloader  *bootloader,
     glnx_file_get_contents_utf8_at (self->sysroot->sysroot_fd, syslinux_config_path, NULL,
                                     cancellable, error);
   if (!config_contents)
-    return FALSE;
+    {
+      if (errno != ENOENT)
```
cgwalters, line 137: Having `errno` be set after a function returns a `GError` is...somewhat undefined behavior.  It'd be better to do:
```
  if (g_error_matches (temp_error, G_IO_ERROR, G_IO_ERROR_NOT_FOUND))
```
which is a pattern you can find all over the codebase (xref https://docs.rs/openat-ext/0.1.9/openat_ext/trait.OpenatDirExt.html#tymethod.open_file_optional )

src/libostree/ostree-bootloader-syslinux.c
```diff
@@ -84,12 +92,13 @@ append_config_from_loader_entries (OstreeBootloaderSyslinux  *self,
       if (regenerate_default && i == 0)
         g_ptr_array_add (new_lines, g_strdup_printf ("DEFAULT %s", val));
 
-      g_ptr_array_add (new_lines, g_strdup_printf ("LABEL %s", val));
+      g_ptr_array_add (new_lines, g_strdup_printf ("\nLABEL %s", val));
```
cgwalters, line 95: The way we're adding newlines into what is intended to be an "array of lines" feels...slightly wrong though it's clearly going to work.  But WDYT about:
```
      g_ptr_array_add (new_lines, g_strdup (""));
      g_ptr_array_add (new_lines, g_strdup_printf ("LABEL %s", val));
```
instead?

cgwalters (commented):
Thanks a ton for working on this!  Seems reasonable to me offhand.

We clearly need to beef up our CI around bootloaders...hm, it might not be too bad for me to hack up something that switches the fcos image to use extlinux.

PR #2261, opened 2021-01-09
sysroot: Handle ro /boot but rw /sysroot

cgwalters:
The recent change in https://github.com/coreos/fedora-coreos-config/pull/659
broke some of our tests that do `mount -o remount,rw /sysroot` but
leave `/boot` read-only.

We had code for having `/boot` read-only before `/sysroot` but
in practice we had a file descriptor for `/sysroot` that we opened
before the remount that would happen later on.

Clean things up here so that in the library, we also remount
`/boot` at the same time we remount `/sysroot` if either are readonly.

Second, adapt the client side code to check for either being
readonly to enable the mount namespace.  Now honestly we could
almost certainly just set this unconditionally - the original
client code here is just being excessively conservative.  But
I'd like to make that cleanup independently of this fix.


lucab (approved):

cgwalters:
flakes

PR #2263, opened 2021-01-12
repo: Make ostree_repo_create_at take nullable options

cgwalters:
Hit this when trying to use the Rust bindings.


lucab (approved):
/lgtm

PR #2267, opened 2021-01-12
Pull depth fixes

dbnicholson:
Here are a couple fixes I found when trying to do a `pull-local --depth=-1`. My use case is that I'm trying to merge 2 repos together locally but the repos have truncated histories due to pruning.

### DIFF
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1113,6 +1113,18 @@ on_metadata_written (GObject           *object,
   check_outstanding_requests_handle_error (pull_data, &local_error);
 }
 
+static gboolean
+is_parent_commit (OtPullData *pull_data,
+                  const char *checksum)
+{
+  /* FIXME: Only parent commits are added to the commit_to_depth table,
+   * so if the checksum isn't in the table then a new commit chain is
+   * being started. However, if the desired commit was a parent in a
+   * previously followed chain, then this will be wrong.
+   */
+  return g_hash_table_contains (pull_data->commit_to_depth, checksum);
+}
+
 static void
 meta_fetch_on_complete (GObject           *object,
                         GAsyncResult      *result,
@@ -1158,7 +1170,8 @@ meta_fetch_on_complete (GObject           *object,
            * We may be pulling from a partial repository that ends in
            * a dangling parent reference. */
           else if (objtype == OSTREE_OBJECT_TYPE_COMMIT &&
-                   pull_data->maxdepth != 0)
+                   pull_data->maxdepth != 0 &&
+                   is_parent_commit (pull_data, checksum))
             {
               g_clear_error (&local_error);
               /* If the remote repo supports tombstone commits, check if the commit was intentionally
@@ -1542,8 +1555,6 @@ scan_commit_object (OtPullData                 *pull_data,
   else
     {
       depth = pull_data->maxdepth;
-      g_hash_table_insert (pull_data->commit_to_depth, g_strdup (checksum),
-                           GINT_TO_POINTER (depth));
     }
 
 #ifndef OSTREE_DISABLE_GPGME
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-repo-pull.c
```diff
@@ -1820,10 +1820,44 @@ scan_one_metadata_object (OtPullData                 *pull_data,
...
         }
 
+      g_autoptr(GError) local_error = NULL;
       if (!_ostree_repo_import_object (pull_data->repo, pull_data->remote_repo_local,
                                        objtype, checksum, pull_data->importflags,
-                                       cancellable, error))
-        return FALSE;
+                                       cancellable, &local_error))
+        {
+          /* When traversing parents, do not fail on a missing commit.
```
jlebon, line 1828: Tempting to refactor this somehow to share with the remote case.

At least let's add a comment somewhere around here to say that this matches the remote logic in `meta_fetch_on_complete`.

dbnicholson: I gave that a little thought but couldn't come up with the right abstraction. `handle_missing_commit (OtPullData *, FetchObject Data *, const char*)`? Then it would queue the tombstone fetch or import it based on `remote_repo_local`.

tests/test-pull-depth.sh
```diff
@@ -63,3 +63,40 @@ find repo/state -name '*.commitpartial' | wc -l > commitpartialcount
...
+find repo/objects -name '*.commit' | wc -l > commitcount
+assert_file_has_content commitcount "^1$"
+find repo/state -name '*.commitpartial' | wc -l > commitpartialcount
+assert_file_has_content commitpartialcount "^0$"
+
+${CMD_PREFIX} ostree --repo=repo pull --depth=-1 origin main
+find repo/objects -name '*.commit' | wc -l > commitcount
+assert_file_has_content commitcount "^1$"
+find repo/state -name '*.commitpartial' | wc -l > commitpartialcount
+assert_file_has_content commitpartialcount "^0$"
```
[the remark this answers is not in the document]
dbnicholson: Sure, makes sense. I had been following the strategy from the first test case of just building on further depth.

tests/test-pull-depth.sh
```diff
@@ -63,3 +63,40 @@ find repo/state -name '*.commitpartial' | wc -l > commitpartialcount
...
+
+echo "ok pull depth missing parent"
+
+# Check that it errors if the ref head commit is missing.
+cd ${test_tmpdir}
+rm -f ostree-srv/gnomerepo/objects/*/*.commit
+rm -rf repo
+mkdir repo
+ostree_repo_init repo
+${CMD_PREFIX} ostree --repo=repo remote add --set=gpg-verify=false origin $(cat httpd-address)/ostree/gnomerepo
```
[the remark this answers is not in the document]
dbnicholson: No, I think deleting the commit and ref should be good enough.

src/libostree/ostree-repo-pull.c
dbnicholson: Added something on the last commit. I tried to figure out how to encapsulate the condition as well, but it I couldn't come up with any ideas about handling the slightly different outer logic between the 2 cases. Let me know if you see a better plan.

dbnicholson:
Ready for review again. I can squash the history if this looks good.

jlebon:
> Added something on the last commit. I tried to figure out how to encapsulate the condition as well, but it I couldn't come up with any ideas about handling the slightly different outer logic between the 2 cases. Let me know if you see a better plan.

Yeah agreed, there's no clean way to do this. I think I personally prefer what you had before, so let's drop that commit?

jlebon:
Also, feel free to squash in your commits. Our merge bot doesn't autosquash anymore.

dbnicholson:
> > Added something on the last commit. I tried to figure out how to encapsulate the condition as well, but it I couldn't come up with any ideas about handling the slightly different outer logic between the 2 cases. Let me know if you see a better plan.
> 
> Yeah agreed, there's no clean way to do this. I think I personally prefer what you had before, so let's drop that commit?

Sure, I'll just leave in the cross referencing comments.

jlebon (approved):
Thanks a lot!
/lgtm

PR #2269, opened 2021-01-17
WIP: Use fsverity library, require keys and add signatures too

cgwalters:
At the time we added fsverity code to ostree, fsverity was
just a CLI tool; since then it has
gained a C shared library which wraps all the signature
bits and the enablement `ioctl()` conveniently.

This makes it much easier for us to support signatures, so
do so.  Note that at this time, because ostree doesn't define
a mechanism to transport fsverity signatures across repositories,
this is mostly only useful for locally-generated signatures.

The idea however is this is a starting point from which we can
build more support, including signature transport, remote keys,
etc.

In order to simplify things, drop support for "opportunistic"
use of fsverity.  In practice we expect people using this
to set it up fully, or not at all.  The "read only files"
aspect *is* useful, but complicated the code too much relative
to its benefit.

Also drop support for using fsverity for `/boot` for now; in
practice most things there are read by the bootloader,
which doesn't know about fsverity.  Instead those should
be covered by e.g. Secure Boot.  This ensures that
we only have one high level API `_ostree_tmpf_fsverity()`
that is invoked from the core commit path.

xref https://lwn.net/Articles/842002
xref https://github.com/ostreedev/ostree/pull/1959
xref https://github.com/coreos/rpm-ostree/issues/1883


cgwalters:
This requires libfsverity from git master as we use https://git.kernel.org/pub/scm/linux/kernel/git/ebiggers/fsverity-utils.git/commit/?id=f76d01b8ce8ce13538bac89afa8acfea9e2bdd57

cgwalters:
Initial user story here (see https://lwn.net/ml/fedora-devel/94055225-94c3-4d6b-a21a-08f5d4632437@www.fastmail.com/ ) is basically:

"As a Fedora CoreOS user I can provide my own encryption keys via Ignition, and use it to enable fsverity on the ostree repository on firstboot, and have files from OS updates also signed using that key"

Next steps after this are:

- Add an API to sign all existing objects (bootstrapping)
- Documentation
- Support fetching a key from kernel keyring (libfsverity enhancement?)

Longer term:

- Also add support for automatically using fsverity on `/etc` too?
- Design a way to ship fsverity signatures (theoretically could pretend they're an xattr on the wire...ugh) - or instead go for a more serious rework of ostree as brainstormed in https://github.com/ostreedev/ostree/pull/1959
- Add support to xfs for fsverity

cgwalters:
https://lore.kernel.org/linux-fscrypt/2369a655-d7e8-467c-8003-860d35ef1026@www.fastmail.com/T/#u

PR #2272, opened 2021-01-28
commit: Optionally translate ima xattr

puiterwijk:
The security.ima extended attribute can only be written by root, but
there is the user.ima attribute that can be set by e.g. ima-evm-utils.
Allowing translation from user.ima to security.ima means that the tree
can be prepared with user.ima attributes, and after translation, stored
as security.ima in the tree.

This would result in a deployed security.ima attribute without needing
to run the tree compose as root.

Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -49,7 +49,7 @@
 #endif
 
 /* Understanding ostree's fsync strategy
- * 
+ *
  * A long time ago, ostree used to invoke fsync() on each object,
  * then move it into the objects directory.  However, it turned
  * out to be a *lot* faster to write the objects into a separate "staging"
@@ -3319,6 +3319,34 @@ get_final_xattrs (OstreeRepo                       *self,
         }
     }
 
+  if (modifier && (modifier->flags & OSTREE_REPO_COMMIT_MODIFIER_FLAGS_IMA_TRANSLATE) > 0)
+    {
+      /* If there is no xattrs to start with, there's nothing to translate */
+      if (ret_xattrs)
+        {
+          GVariantBuilder builder;
+          g_variant_builder_init (&builder, G_VARIANT_TYPE ("a(ayay)"));
+          guint n = g_variant_n_children (ret_xattrs);
+          for (guint i = 0; i < n; i++)
+            {
+              const char *name = NULL;
+              g_autoptr(GVariant) value = NULL;
+
+              g_variant_get_child (ret_xattrs, i, "(^&ay@ay)",
+                                  &name, &value);
+
+              if (strcmp (name, "user.ima") == 0)
+                {
+                  name = "security.ima";
+                }
+              g_variant_builder_add (&builder, "(@ay@ay)",
+                                    g_variant_new_bytestring (name),
+                                    value);
+            }
+          ret_xattrs = g_variant_ref_sink (g_variant_builder_end (&builder));
+        }
+    }
+
   if (original_xattrs && ret_xattrs && g_variant_equal (original_xattrs, ret_xattrs))
     modified = FALSE;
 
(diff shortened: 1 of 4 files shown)


jlebon:
> This would result in a deployed security.ima attribute without needing
to run the tree compose as root.

Note you can also do this also via the xattr callback. For example this is how rpm-ostree sets file caps when importing RPMs: https://github.com/coreos/rpm-ostree/blob/e88a736e55eeb91ced5619c4ee9119d4cf705068/src/libpriv/rpmostree-importer.cxx#L818-L833.

Though if `user.ima` is an established convention, it might make sense indeed to have support for it in libostree.

cgwalters:
> Though if user.ima is an established convention, it might make sense indeed to have support for it in libostree.

Yeah I am OK with this, but I'd also like to see high level support in this in rpm-ostree i.e. declarative in the manifest something like:

```
ima:
  key: /path/to/rsa.priv
```
(and support for passing the key in the kernel keyring or via file descriptor too)

It looks like this would require today forking off `evmctl` for each unsigned file; a bit like the state of fsverity before it gained a shared library, but eh.

Anyways WDYT about just generalizing this so that e.g. *all* `user.security` xattrs are translated to `security.`?  Or maybe it's:
`ostree commit --xattr-map=user.ima=security.ima`?  That wouldn't require a new commit modifier flag (i.e. new ostree API) and could simply be done in the CLI commit modifier path, and easily replicated in rpm-ostree.

puiterwijk:
@cgwalters There's actually a libimaevm, which has `sign_hash`. I have a draft patch for ostree to support that, but then with arguments to `ostree commit`.

ashcrow:
The travis tests failures look to show a lot of infra flakes (can't reach a gnome repo).

For jenkins CI I see:
```
--- FAIL: ext.ostree.destructive-rs.transactionality (134.80s)

        harness.go:709: kolet failed: : kolet run-test-unit failed: Error: Unit kola-runext.service exited with code 1

2021-01-29T21:41:39Z cli: Unit kola-runext.service exited with code 1: Process exited with status 1
```

I do see the above jenkins failure in a few other PRs such as:
- https://github.com/ostreedev/ostree/pull/2273
- https://github.com/ostreedev/ostree/pull/2256
- https://github.com/ostreedev/ostree/pull/2187

which may indicate this is a known failing test and not directly related to this change.

@cgwalters @lucab @travier @kelvinfan001 would one of you mind taking a look and verifying?

ashcrow:
Is there a way to force a retest?

cgwalters:
For Jenkins it's logging in and clicking retry; done now.
Travis requires a force push AFAIK; I hope to drop Travis soon probably in favor of using Prow, or maybe GH actions.

But let's reach a conclusion on the high level interface; I think I'd lean most towards
> ostree commit --xattr-map=user.ima=security.ima
and we could also just hardcode doing that in rpm-ostree by default probably.  (ostree is intended to be a flexible shared library, opinionated where necessary; rpm-ostree is more opinionated)

lucab:
CI failure seems to be a race/regression in rpm-ostree, tracked at https://github.com/coreos/rpm-ostree/issues/2531.

For the short term, I managed to get a :heavy_check_mark: by adding some retries around the racing logic.

ashcrow:
**Bump** for continued conversation :smiley:

puiterwijk:
Thanks, @ashcrow for the ping. Given that @cgwalters wants the generic translation, I'm now instead adding this to rpm-ostree as something that's read from the treefile instead.
I'll open that a PR against rpm-ostree later today.

puiterwijk:
I opened the rpm-ostree PR (as WIP since I still need to finish tests) at https://github.com/coreos/rpm-ostree/pull/2563

PR #2273, opened 2021-01-31
README.md: Also link apt2ostree

cgwalters:
Since the topic of Debian+ostree-for-host comes up fairly often.


cgwalters:
will debug that soon.

PR #2275, opened 2021-02-02
CI: run as non-root and add sanitizer build+check

cgwalters:
ci: Don't install deps if running as non-root

This way we run in Prow too.

---

ci: Add new build-check-sanitized.sh

All C/C++ projects should use the sanitizers (and static analysis)
in their CI.  We had this but lost it in one of our CI shuffles;
let's readd it.


jlebon (approved):
/lgtm

cgwalters:
This can't affect Jenkins CI

PR #2276, opened 2021-02-03
tests/ext/destructive: enhance test logic

lucab:
This enhances external-tests logic, ensuring that destructive tests
have some retries and context to pinpoint failures, and that failed-state services
are reset between iterations.

### DIFF
--- a/tests/inst/src/rpmostree.rs
+++ b/tests/inst/src/rpmostree.rs
@@ -1,7 +1,6 @@
-use anyhow::Result;
+use anyhow::{Context, Result};
 use serde_derive::Deserialize;
-use serde_json;
-use std::process::{Command, Stdio};
+use std::process::Command;
 
 #[derive(Deserialize)]
 #[serde(rename_all = "kebab-case")]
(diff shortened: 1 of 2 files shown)


tests/inst/src/rpmostree.rs
```diff
@@ -25,9 +24,29 @@ pub(crate) struct Deployment {
 }
 
 pub(crate) fn query_status() -> Result<Status> {
-    let cmd = Command::new("rpm-ostree")
-        .args(&["status", "--json"])
-        .stdout(Stdio::piped())
-        .spawn()?;
-    Ok(serde_json::from_reader(cmd.stdout.unwrap())?)
```
cgwalters, line 32: Ugh and we were missing a check of the exit status here...such an evil trap in the Rust command API.
(Which you have correctly fixed)
I have this bit https://github.com/cgwalters/coretoolbox/blob/29a862b62587e1b462d577a45037e8c395e654ab/src/cmdrunext.rs#L8 we could cargo cult around or...perhaps crateify.  I thought https://docs.rs/subprocess/0.2.6/subprocess/ had an API for this but I guess not?

tests/inst/src/rpmostree.rs
```diff
@@ -25,9 +24,29 @@ pub(crate) struct Deployment {
...
+        if res.status.success() || retries >= 10 {
+            break res;
+        }
+        std::thread::sleep(pause);
+    };
+
+    if !cmd_res.status.success() {
+        anyhow::bail!(
+            "running 'rpm-ostree status' failed: {}",
+            String::from_utf8_lossy(&cmd_res.stderr)
```
cgwalters, line 47: As a general rule...I think it's better to just redirect child process stderr to our stderr rather than capturing it, because the output can be arbitrarily large.

cgwalters (commented):
/lgtm

PR #2277, opened 2021-02-03
Add --enable-sanitizers, fix `make check` with it

cgwalters:
It's cleaner if this is an build option rather than being
kludged into the CI layer.

Notably we can't use `LD_PRELOAD` anymore with ASAN, so update
our tests to check for `ASAN_OPTIONS`.


cgwalters:
Flakes

PR #2279, opened 2021-02-11
docs: Describe using scratch/empty deltas for initial fetches

cgwalters:
Came up with a user hitting ratelimiting from S3.


jlebon (approved):
/lgtm

PR #2282, opened 2021-02-19
refs: Make ostree_repo_resolve_rev{,_ext}() use (nullable)

cgwalters:
We have an `allow_noent` boolean that controls this, but
were missing the `(nullable)` annotation, so the Rust bindings
panic when the ref doesn't exist instead of being `Option<GString>`.

### DIFF
--- a/src/libostree/ostree-repo-refs.c
+++ b/src/libostree/ostree-repo-refs.c
@@ -449,7 +449,7 @@ _ostree_repo_resolve_rev_internal (OstreeRepo     *self,
  * @self: Repo
  * @refspec: A refspec
  * @allow_noent: Do not throw an error if refspec does not exist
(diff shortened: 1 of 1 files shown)


dbnicholson:
Interesting. I use this on python all the time. I guess it doesn't care about the `nullable` return. I can't remember if I actually have privileges to do this or not, but

/lgtm

cgwalters:
> I use this on python all the time. I guess it doesn't care about the nullable return.

Yeah; a regret I have with G-I is not focusing more on the needs of static languages at the time.  Obviously then the main goal was dynamic languages because at the time they had the "memory safety+prototyping/development speed" aspects whereas C is "performance + bindability".

When I invent a time machine...well there's a lot of things to fix, but taking Rust back to the 1990s or at least early 2000s will be up there on the list :wink: 

>  I can't remember if I actually have privileges to do this or not, but

You do and you should!

PR #2285, opened 2021-02-26
deploy: Add subbootversion to journal

cgwalters:
To help debug an issue we've seen where `/boot` isn't
in sync with the `/ostree/boot` dir, let's log to the journal
what we're doing.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1675,6 +1675,7 @@ static gboolean
 swap_bootlinks (OstreeSysroot *self,
                 int            bootversion,
                 GPtrArray     *new_deployments,
+                char         **out_subbootdir,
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2326,6 +2330,7 @@ write_deployments_bootswap (OstreeSysroot     *self,
                         cancellable, error))
     return FALSE;
 
+  *out_subbootdir = g_steal_pointer (&new_subbootdir);
```
dbnicholson, line 2333: Should you check that `out_subbootdir` isn't `NULL` like in `swap_bootlinks`? It seems like you're updating all the callers of both helpers to pass a valid pointer, but it seems safer in case it's reused.

cgwalters:
Yeah, that's a good best practice.  Fixed.

Issue #2283, opened 2021-02-22
OSTree boot links can get out of sync


jlebon:
We've had one instance recently in RHCOS where a system failed to boot with:

```
systemd[1]: Starting OSTree Prepare OS/...
ostree-prepare-root[887]: ostree-prepare-root: Couldn't find specified OSTree root '/sysroot//ostree/boot.1/rhcos/a77c6a70e3037f40c6dc3634d31e63f797d4a5cf9059f5927141ca53690d1942/0': No such file or directory
systemd[1]: ostree-prepare-root.service: Main process exited, code=exited, status=1/FAILURE
systemd[1]: ostree-prepare-root.service: Failed with result 'exit-code'.
systemd[1]: Failed to start OSTree Prepare OS/.
systemd[1]: ostree-prepare-root.service: Triggering OnFailure= dependencies.
```

And digging deeper it seemed to be due to the bootversion being out of sync between the bootloader and the sysroot:

```
$ ls -la / sysroot/ostree/boot.*/rhcos/*
/sysroot/ostree/boot.0.1/rhcos/57fdb4d93971e78e2ba1dfb53ac646d8372e016bb97dedf2cdadd44b355ec7e9:
total 0
drwxr-xr-x. 2 root root  15 Jan 21 09:35 .
drwxr-xr-x. 4 root root 150 Jan 21 09:35 ..
lrwxrwxrwx. 1 root root  95 Jan 21 09:35 0 -> ../../../deploy/rhcos/deploy/046de410ba240083996855d797207e09374bbfdf65e5be34baa1e1f1ab49918a.0

/sysroot/ostree/boot.0.1/rhcos/a77c6a70e3037f40c6dc3634d31e63f797d4a5cf9059f5927141ca53690d1942:
total 0
drwxr-xr-x. 2 root root  15 Jan 21 09:35 .
drwxr-xr-x. 4 root root 150 Jan 21 09:35 ..
lrwxrwxrwx. 1 root root  95 Jan 21 09:35 0 -> ../../../deploy/rhcos/deploy/ba247419bcd04867a38aa336701cf54cf62d36114bd125e64198567ec14f262a.0

/sysroot/ostree/boot.0/rhcos/57fdb4d93971e78e2ba1dfb53ac646d8372e016bb97dedf2cdadd44b355ec7e9:
total 0
drwxr-xr-x. 2 root root  15 Jan 21 09:35 .
drwxr-xr-x. 4 root root 150 Jan 21 09:35 ..
[...]

PR #2287, opened 2021-03-02
lib/sysroot: Add comments and debug statements around sysroot parsing

jlebon:
Was looking at this code more closely today to investigate issues
related to bootlink mismatches (#2283).


lucab (approved):
/lgtm

PR #2288, opened 2021-03-04
docs: Add Torizon to related projects and OS

leograba:
The Torizon platform, includin the TorizonCore OS, the TorizonCore
Builder Tool and the Torizon OTA, use OSTree as a base for update the
host OS, while the user focus on application development using Docker.

Add TorizonCore to the list of Operating systems and distributions using
OSTree.

Add Torizon and its components to the list of related projects.


cgwalters (approved):
Thanks!
/lgtm

PR #2296, opened 2021-03-11
Add an API+CLI to inject metadata for bootable OSTree commits

cgwalters:
I was doing some rpm-ostree work and I wanted to compare two
OSTree commits to see if the kernel has changed.  I think
this should be a lot more natural.

Add `ostree commit --bootable` which calls into a new generic
library API `ostree_commit_metadata_for_bootable()` that
discovers the kernel version and injects it as an `ostree.linux`
metadata key.  And for extra clarity, add an `ostree.bootable`
key.

It's interesting because the "core" OSTree layer is all about
generic files, but this is adding special APIs around bootable
OSTree commits (as opposed to e.g. flatpak as well as
things like rpm-ostree's pkgcache refs).

Eventually, I'd like to ensure everyone is using this and
hard require this metadata key for the `ostree admin deploy`
flow - mainly to prevent accidents.


cgwalters:
OK yeah we need to drop Travis, it's been dead for a while due to the Docker rate limiting.  I may look at GH actions for that.

src/libostree/ostree-repo-os.h
```diff
@@ -0,0 +1,45 @@
...
+/** 
+ * OSTREE_METADATA_KEY_BOOTABLE:
+ *
+ * GVariant type `b`: Set if this commit is intended to be bootable
+ */
+#define OSTREE_METADATA_KEY_BOOTABLE "ostree.bootable"
+/** 
+ * OSTREE_METADATA_KEY_LINUX:
+ *
+ * GVariant type `s`: Contains the Linux kernel version
```
lucab, line 37: s/version/release/ perhaps? (matching `uname` nomenclature)

src/libostree/ostree-repo-os.c
```diff
@@ -0,0 +1,74 @@
...
+      if (!g_file_enumerator_iterate (dir_enum, &child_info, &child_path,
+                                      cancellable, error))
+        return FALSE;
+      if (child_info == NULL)
+        break;
+      if (g_file_info_get_file_type (child_info) != G_FILE_TYPE_DIRECTORY)
+        continue;
+    
+      g_autoptr(GFile) kernel_path = g_file_resolve_relative_path (child_path, "vmlinuz");
+      if (!g_file_query_exists (kernel_path, NULL))
```
lucab, line 65: Is there something elsewhere in ostree that guarantees that there is exactly a single `/usr/lib/modules/*/vmlinuz` entry in the content of a commit?

src/libostree/ostree-repo-os.c
```diff
@@ -0,0 +1,74 @@
...
+ * @root: Root filesystem to be committed
+ * @dict: Dictionary to update
+ *
+ * Update provided @dict with standard metadata for bootable OSTree commits.
+ */
+_OSTREE_PUBLIC
+gboolean
+ostree_commit_metadata_for_bootable (GFile *root, GVariantDict *dict, GCancellable *cancellable, GError **error)
+{
+  g_autoptr(GFile) modules = g_file_resolve_relative_path (root, "usr/lib/modules");
```
lucab, line 44: This looks like pretty much the same logic of https://github.com/ostreedev/ostree/blob/v2020.8/src/libostree/ostree-sysroot-deploy.c#L1054

cgwalters: Yeah...hard to share logic because using the libostree API abstraction means the content to be committed may not be realized as physical files, and we need to support that here.  (For example, this API allows reading from an `archive` repository directly, and while it's usually better to have a checkout, it can be very convenient to do some operations on an archive repo directly)

Whereas the sysroot logic always operates on a `bare` repo with real files.

That logic is also trying to handle the legacy cases, which this one is explicitly not supporting.

jlebon (approved):
Missing manpage and bash completion (which yeah, would be nice if that was automatic). Otherwise, LGTM!

src/libostree/ostree-repo-os.c
cgwalters: Hmm it indeed looks like the deploy code isn't enforcing this, it's just taking the first one.  IMO that's a bug though, we should be stricter.  Let's do that here.

cgwalters:
> Missing manpage and bash completion (which yeah, would be nice if that was automatic).

Yeah, I was reading https://github.com/clap-rs/clap/issues/552 recently.

jlebon (approved):
LGTM

cgwalters:
> In general, is ostree supposed to handle multiple (inactive) kernel images in a commit?

No; we should only have one.  OSTree is all about a mechanism to transactionally swap between versioned bootable filesystem  trees.  We don't have code to install or switch to alternative kernels inside a single ostree commit.  Instead, OS builders can ship multiple OSTree commits, and switch between them client or server side.  These ostree commits will share storage via hardlinks.

This is very relevant to OpenShift where we want to expose the ability to switch between the mainline `kernel` and `kernel-rt`.  Today the MCO uses rpm-ostree client side overrides but down the line I'd like to have it be a pre-built ostree commit.

cgwalters:
Or another way to say this is - handling multiple kernel versions is no different than handling multiple versions of any other (userspace) component of the ostree payload, e.g. systemd - those are distinct commits which one can generate client or server side; it wouldn't make much sense to offer a way to switch systemd versions inside a single commit.

lucab:
@cgwalters thanks for the explanation, it clarified all my doubts!

Issue #1009, opened 2017-07-17
grub2-mkconfig when invoked directly generates bad config on EFI


cmurf:
I'm still running into this with Fedora-Workstation-ostree-x86_64-Rawhide-20170714.n.0.iso
atomic-1.18.1-4.fc27.x86_64
ostree-2017.8-2.fc27.x86_64
rpm-ostree-2017.7-1.fc27.x86_64

[chris@frawhide ~]$ sudo ostree admin instutil grub2-generate -v 0
OT: Deployment 6dc795302af832537ad225050a00441726350937f4f9b218b01f5e885c0e1619.0 unlocked=0
**
OSTree:ERROR:src/libostree/ostree-bootloader-grub2.c:167:_ostree_bootloader_grub2_generate_config: assertion failed: (grub2_boot_device_id != NULL)
Aborted
[chris@frawhide ~]$ sudo ostree admin instutil grub2-generate -v 1
OT: Deployment 6dc795302af832537ad225050a00441726350937f4f9b218b01f5e885c0e1619.0 unlocked=0
**
OSTree:ERROR:src/libostree/ostree-bootloader-grub2.c:167:_ostree_bootloader_grub2_generate_config: assertion failed: (grub2_boot_device_id != NULL)
Aborted
[chris@frawhide ~]$

Recalling 
https://bugzilla.redhat.com/show_bug.cgi?id=1293986
https://github.com/ostreedev/ostree/pull/165

This command isn't really supposed to be user servicable, but on Workstation there needs to be a way to regenerate the grub.cfg. When I run grub2-mkconfig it does not pick up items in /etc/default/grub and it also uses linux16 and initrd16 commands, instead of linuxefi/initrdefi as appropriate for this UEFI installation.

cgwalters:
See also https://github.com/projectatomic/rpm-ostree/issues/594

cmurf:
https://github.com/projectatomic/rpm-ostree/issues/594 says on EFI do:
`grub2-mkconfig > /boot/efi/EFI/fedora/grub.cfg.new && mv /boot/efi/EFI/fedora/grub.cfg{.new,}`

But this creates a grub.cfg with linux16/initrd16 commands which is incorrect. I don't know why the original grub.cfg came out correct during install, but post-install it results in invalid grub commands.

cgwalters:
Yep, confirmed that bug.  Hmm.  Man this code sucks :cry:

cgwalters:
[...]

PR #2298, opened 2021-03-15
Add configure option for unsuffixed GRUB2 commands linux/initrd

KloudJack:
GRUB starting with version 2.02 permits the use of the standard `linux` and `initrd` commands for both EFI boot in *-efi installations, as well as 32-bit BIOS boot in i386-pc installations.

This makes the use of the -16 and -efi suffixes for BIOS and EFI boot obsolete in modern distros (The efi suffix is a RHEL/Fedora-ism anyway that doesn't apply to distros not based on these).

Additionally, booting the kernel in real-mode on x86 BIOS systems with the linux16/initrd16 commands should be an exception and not the default.

All modern distros, including RHEL and Fedora versions starting with RHEL 8 and Fedora 28, include or are based on a grub >2.02 release, so I think moving forward ostree should drop these suffixes for generated grub configurations.

Closes: #1009

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -548,6 +548,14 @@ AM_COND_IF(BUILDOPT_SYSTEMD_AND_LIBMOUNT,
   AC_DEFINE([BUILDOPT_LIBSYSTEMD_AND_LIBMOUNT], 1, [Define if systemd and libmount]))
 if test x$with_libsystemd = xyes; then OSTREE_FEATURES="$OSTREE_FEATURES systemd"; fi
 
+AC_ARG_WITH(modern-grub,
+	    AS_HELP_STRING([--with-modern-grub],
+                           [Omit grub linux and initrd suffixes for EFI/BIOS booting on GRUB >2.02 (default: no)]),,
+              [with_modern_grub=no])
+AS_IF([ test x$with_modern_grub = xyes], [
+  AC_DEFINE([WITH_MODERN_GRUB], 1, [Define if we have a GRUB version newer than 2.02])
+])
+
 AC_ARG_WITH(builtin-grub2-mkconfig,
             AS_HELP_STRING([--with-builtin-grub2-mkconfig],
                            [Use a builtin minimal grub2-mkconfig to generate a GRUB2 configuration file (default: no)]),,
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -28,25 +28,26 @@
 
 #include <string.h>
 
-/* I only did some cursory research here, but it appears
- * that we only want to use "linux16" for x86 platforms.
- * At least, I got a report that "linux16" is definitely wrong
- * for ppc64.  See
- * http://pkgs.fedoraproject.org/cgit/rpms/grub2.git/tree/0036-Use-linux16-when-appropriate-880840.patch?h=f25
- * https://bugzilla.redhat.com/show_bug.cgi?id=1108296
- * among others.
+/* Maintain backwards compatibility with legacy GRUB
+ * installations that might rely on the -16 suffix
+ * for real-mode booting.
+ * Allow us to override this at build time if we're
+ * using a modern GRUB installation.
  */
-#if defined(__i386__) || defined(__x86_64__)
+#if !defined(WITH_MODERN_GRUB) && ( defined(__i386__) || defined(__x86_64__) )
 #define GRUB2_SUFFIX "16"
 #else
 #define GRUB2_SUFFIX ""
 #endif
-/* https://github.com/projectatomic/rpm-ostree-toolbox/issues/102#issuecomment-316483554
- * https://github.com/rhboot/grubby/blob/34b1436ccbd56eab8024314cab48f2fc880eef08/grubby.c#L63
- *
- * This is true at least on Fedora/Red Hat Enterprise Linux for aarch64.
+/* Maintain backwards compatibility with legacy GRUB
+ * installations that might rely on the -efi suffix
+ * for RHEL/fedora-based distros.
+ * Allow us to override this at build time if we're
+ * using a modern GRUB installation that makes the
+ * suffix-less linux/initrd commands synonymous for
+ * both EFI and BIOS booting.
  */
-#if defined(__aarch64__)
+#if defined(WITH_MODERN_GRUB) || defined(__aarch64__)
 #define GRUB2_EFI_SUFFIX ""
 #else
 #define GRUB2_EFI_SUFFIX "efi"


dbnicholson:
Since ostree has no control over what grub is installed on the target system, I don't think this should be unilaterally done. We would likely revert this at Endless as we have systems that predate this standardization. I would think this should be a build time option or ideally it would probe grub at runtime to see if it's needed or not. I don't know if the latter is actually possible, though.

KloudJack:
@dbnicholson I see the Endless OS project uses a GRUB version forked on an older Fedora GRUB fork's branch (24-26)? Those Fedora-specific branches all seems to use grub-2.02-beta3 as a basis which itself doesn't include x86 EFI support yet, so I suppose this would also break current EOS releases as they still rely on the `linuxefi` command for EFI boot.

As I understand it, libostree aims to be distribution-agnostic. So personally it feels a little strange to _exclusively_ support a distribution-specific fork of a project as universal as GRUB in the *nix world. Though I do understand the reasoning behind the initial use of the RHEL-fork (to provide EFI boot functionality that was not yet available in the main project).

However, GRUB 2.02 was released over 4 years ago and offers this functionality in mainline since; I feel there has to be a demarcation on functionality support for out-of-date software, let alone outdated _distro-specific_ support code (though I'm not the one to decide on that).

Ostree may not have control over the grub version deployed, but it should set version requirements for dependencies or support code like the ostree-bootloader-* components, not just for the sake of the sanity of project maintainers, but also developers integrating libostree.

If you integrate libostree, GRUB, and any modern, existing or self-built, non-RHEL distro, then you'll find that deployments for EFI are broken, as linuxefi isn't a recognized command for mainline GRUB builds. Personally I think it's unfair to persist on functionality supporting a subset of users exclusively, though I'm willing to find a compromise that makes everyone happy.

> I would think this should be a build time option or ideally it would probe grub at runtime to see if it's needed or not. I don't know if the latter is actually possible, though. 

I suppose we could test for the presence of the `linuxefi.mod` grub module at runtime, and use that as a decision factor for supplementing the efi suffix to the `linux` and `initrd` commands. That feels like a bit of a hack though, and I'm not sure we could always rely on the modules being in the same place on boot.

I have no idea of the requirements that EOS poses for its bootloader, but perhaps the GRUB version the project uses should at least be rebased on a modern RHEL/Fedora GRUB fork's branch just for future-proofing and minimizing future dev efforts there. :+1:

KloudJack:
@dbnicholson Would you be happy if we just remove the linux**16** and initrd**16** suffix for now, or does Endless require real-mode BIOS boot for x86*?

As the decision per ostree-bootloader-grub2 for suffixing **efi** is based only on arch == arm64, proper EFI detection is broken for any other arch on GRUB anyway. Due to this the suffix-less boot on x86* and other archs would be catchall and work for EFI and BIOS systems on modern mainline GRUB. For systems that require the **efi** suffix **and** where EFI detection is broken, it can still be overridden with the `_OSTREE_GRUB2_IS_EFI=1` environment var. So behaviour should be unchanged for distros like EOS.

I think proper detection of EFI/BIOS and usage of either (linux/initrd)**efi** commands should rely on finding the active grub module directory (e.g. i386-pc vs x86_64-efi, etc.) and detecting certain modules in it, but that's another PR. Also I'm not sure how much the ostree maintainers want to continue supporting the bootloader-specific integration stuff when BLS seems to be the way forward (the RH devs seem to agree as the rhboot-GRUB tree has BLS support while mainline does not yet).

dbnicholson:
Actually, I forgot that we don't even ship the grub parts and rely entirely on the BLS handling in grub. So, I think from an Endless perspective this won't matter at all.

My point, though, is that while you can make requirements on other parts of userspace in ostree, the actual bootloader, like the hardware itself, is entirely out of ostree's control. And grub in particular has been heavily patched by every vendor for years since the mainline release process basically didn't exist. Like you say, the were all based on a 2.02 beta + a pile of patches, so it's pretty difficult to say exactly what features everybody has.

Anyways, since it will have no effect on our products, I have no objections. But this is falling into a trap that many of our engineers have fallen into many times - this will run live on systems trying to update, not just on brand new systems running today's software. If you are trying to support systems updating via ostree, then you need to consider that the environment the deployment code will run in is likely a much older environment than what you have right now. And if they can't make a new deployment, then they're essentially stuck forever. You're essentially writing a distro upgrade tool here.

jlebon:
> But this is falling into a trap that many of our engineers have fallen into many times - this will run live on systems trying to update, not just on brand new systems running today's software. If you are trying to support systems updating via ostree, then you need to consider that the environment the deployment code will run in is likely a much older environment than what you have right now. And if they can't make a new deployment, then they're essentially stuck forever.

Yeah, I concur. And the bootloader specifically might have never been updated at all since installation (in FCOS/RHCOS at least, we're trying to address this with [bootupd](https://github.com/coreos/bootupd/)).

It's probably safer to just leave this code alone until we can eventually nuke it entirely in favour of pure BLS (which as you've said is the way forward). FCOS and RHCOS already use `bootloader=none` at least, but we still need to drive it into other currently Anaconda+rpm-ostree-based Fedora spins.

cgwalters:
For OSes/distros that know they have the new grub, we could add a build-time option to do this.  Effectively `--with-modern-grub` and `#ifdef WITH_MODERN_GRUB` etc.

jlebon:
What would be ideal (and this came up before too and I'm sure will come up again) is if GRUB wrote down its version in some distro-independent machine-readable format (e.g. JSON) in a file in `/boot` or something.

cgwalters:
Heh, bootupd does write a JSON file with that today.  I don't think ostree should require that of course, but it'd be nice to support linking them together.

Or grub could inject some feature flags into its binary that we could parse too?   Though that gets into finding the grub location.

So I think build option is going to be easiest for now.

cgwalters:
@KloudJack I can take on the autotools glue for this if you want.

KloudJack:
> But this is falling into a trap that many of our engineers have fallen into many times - this will run live on systems trying to update, not just on brand new systems running today's software. If you are trying to support systems updating via ostree, then you need to consider that the environment the deployment code will run in is likely a much older environment than what you have right now. And if they can't make a new deployment, then they're essentially stuck forever.

Very good point. Worst-case systems will have last run grub-install half a decade ago at install, even if they've always upgraded to the latest stable release of their distro.

> For OSes/distros that know they have the new grub, we could add a build-time option to do this. Effectively `--with-modern-grub` and `#ifdef WITH_MODERN_GRUB` etc.

> So I think build option is going to be easiest for now.

Agreed, that's probably the most backwards-compatible solution and easiest to maintain until the bootloader-specific code gets dropped in favor of BLS (whenever that might be).

> I can take on the autotools glue for this if you want.

No worries, I'll update this PR's commit using `WITH_MODERN_GRUB` for the SUFFIX preprocessor conditionals in ostree-bootloader-grub2, and add the configure flag to configure.ac.

configure.ac
```diff
@@ -548,6 +548,14 @@ AM_COND_IF(BUILDOPT_SYSTEMD_AND_LIBMOUNT,
   AC_DEFINE([BUILDOPT_LIBSYSTEMD_AND_LIBMOUNT], 1, [Define if systemd and libmount]))
 if test x$with_libsystemd = xyes; then OSTREE_FEATURES="$OSTREE_FEATURES systemd"; fi
 
+AC_ARG_WITH(modern-grub,
```
jlebon, line 551: Hmm, I wonder if this should just be e.g. `--with-grub-2.02` instead. (Not sure Autoconf supports periods in these switches.) Today's modern GRUB is tomorrow's ancient GRUB. :)

Or maybe we should be specific about the feature this is enabling, which might be safer given that each distro carries so many patches. E.g. `--without-grub-efi-16-suffixes`?

cgwalters: There's also the argument to flip the default now - rip off the band-aid and require people who need it (Endless and probably Fedora derivatives for now to be safe) to explicitly pass `--with-grub-efi-16` or.

cgwalters:
This all said @KloudJack I want to be sure we're on the same page - from my PoV both paths are actually legacy - the strongly preferred option is where grub natively reads `/boot/loader/entries`.  That was the idea from the start - it gets ostree out of the business of having bootloader backends at all.

I think the patches for grub+BLS aren't upstream yet but if you pull them, then you can set the bootloader backend to "none".  Among other massive benefits, this avoids `os-prober` being involved.

configure.ac
cgwalters: Though I'm conflicted on this myself...if we all agree the real end goal is everyone has a grub that does BLS, then maybe it's not worth the disruption from this.

> Today's modern GRUB is tomorrow's ancient GRUB. :)

True in general but the idea here is we won't care about the grub version in the future.

So...I know this is all very bike-sheddable but I am circling in on just accepting `--with-modern-grub` as is.

cgwalters (approved):

PR #2302, opened 2021-03-17
tests/inst: Fix lots of `cargo clippy` warnings

cgwalters:
Prep for doing this in CI.

PR #2303, opened 2021-03-17
ci: Add a Github Action for Rust for tests/inst

cgwalters:
Let's dip our toes in GH Actions (specifically aiming to migrate
away from Travis).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,27 @@
+---
+name: Rust
+on:
+  push:
+    branches: [master]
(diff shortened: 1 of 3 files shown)


lucab:
Great, thanks! Assuming https://github.com/ostreedev/ostree/pull/2302 goes in quickly, we can maybe directly turn on clippy here too.

cgwalters:
Ahhh right, the plan to do clippy with this is going to fall over on our C dependencies.

kelvinfan001 (approved):
LGTM

PR #2306, opened 2021-03-18
Drop minimal rust/ library

cgwalters:
This was my first experiment with using Rust in this way; I gained
a lot of knowledge from it.  But, we don't really gain
anything from the code as it is today - while it is "bit fiddling"
code, the C code is well tested.

We have a lot of compile-time options, and trimming them will be
helpful.

We've also gotten pushback on hard requiring Rust client side.

Instead, what I'd like to do is hopefully soon create an `ostree-system`
crate that uses the existing `ostree` library and can contain
code drained from the rpm-ostree Rust and used by other projects perhaps.

So the goal here is really more Rust, but we need to focus our
efforts on where it's most valuable.


lucab:
A bit sad to see this go, but I think it's the right way.

CI is red because `ostree/tests/inst` lacks the lockfile, and some dependency versions drifted into a broken state.

lucab (approved):

jlebon (approved):
Makes sense to me.

Plus, we can always go back to oxidizing this codebase in earnest in the future when the ecosystem is more ready for it.

cgwalters:
> CI is red because ostree/tests/inst lacks the lockfile, and some dependency versions drifted into a broken state.

Yeah, we should add a lockfile.  Will do that as a followup, but for now I opened a PR to fix the upstream problem and added a commit to use it here.

cgwalters:
To elaborate on this a bit: I think the ostree core C code is mostly OK - there's always of course things that in hindsight need redesign, or more testing, etc.  But the value in rewriting more of the core in Rust I think is less valuable than the new direction of writing new high level functionality in Rust (possibly with low level helpers in the C library).  Plus writing tests in Rust, etc.  

That said https://github.com/ostreedev/ostree/pull/2343 just landed too which ensures Rust has a foothold in the core, and beyond tests we can use it for parts of the build system too perhaps; and maybe down the line reconsider and introduce a hard dependency on Rust here too.  

Probably an example of the biggest thing to rewrite would be the pull code - doing HTTP in a memory safe language (ideally out of process) etc.  But it's also the pull code that would be most likely to introduce regressions if rewritten, so here we are :smile:

Issue #2307, opened 2021-03-19
Release 2021.1


cgwalters:
# Release process

The release process follows the usual PR-and-review flow, allowing an external reviewer to have a final check before publishing a release.

## Requirements

This guide requires:

 * a web browser (and network connectivity)
 * `git`
 * GPG setup and personal key for signing
 * [git-evtag](https://github.com/cgwalters/git-evtag/)
 * write access to the git repository
 * upload access to this project on GitHub

## Release checklist

- Prepare local environment:
  - [ ]  `git remote get-url --push origin`
  - [ ]  validate that the output above points to `git@github.com:ostreedev/ostree.git`
  - [ ] `git checkout master && git pull`
  - [ ] `git clean -fd`
  - [ ] `RELEASE_VER=yyyy.n` (matching `package_version` in `configure.ac`)
  - [ ] `git checkout -b release-${RELEASE_VER}`

- Prepare the release commits:
  - [x] `sed -i -e 's/^is_release_build=no/is_release_build=yes/' configure.ac`
  - [x] move the new-symbols stanza (if any) from `src/libostree/libostree-devel.sym` to `src/libostree/libostree-released.sym`
  - [x] comment the `src/libostree/libostree-devel.sym` include in `Makefile-libostree.am`
  - [x] update `tests/test-symbols.sh` with the new digest from `sha256sum src/libostree/libostree-released.sym`
  - [x] `git commit -a -m "Release ${RELEASE_VER}"`
  - [x] `RELEASE_COMMIT=$(git rev-parse HEAD)`
  - [x] `./autogen.sh && make dist`
  - [x] update `year_version` and `release_version` in `configure.ac` for the next development cycle
  - [x] `sed -i -e 's/^is_release_build=yes/is_release_build=no/' configure.ac`
  - [x] `git commit -a -m 'configure: post-release version bump'`

- Open a PR to create the release:
  - [x] `git push -u origin release-${RELEASE_VER}`
  - [x] open a web browser and create a PR for the branch above, titled `Release ${RELEASE_VER}`
  - [x] make sure the resulting PR contains two commits
  - [ ] in the PR body, write a short summary of relevant changes since last release (using `git shortlog` too)

- [ ] get the PR reviewed, approved and merged

- Publish the tag:
  - [ ] `git fetch origin && git checkout ${RELEASE_COMMIT}`
  - [ ] `git-evtag sign v${RELEASE_VER}`
[...]

PR #2308, opened 2021-03-19
Release 2021.1

cgwalters:
Release 2021.1

---

configure: post-release version bump

---

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -188,9 +188,9 @@ endif # USE_GPGME
 symbol_files = $(top_srcdir)/src/libostree/libostree-released.sym
 
 # Uncomment this include when adding new development symbols.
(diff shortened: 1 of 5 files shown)


cgwalters:
Following https://github.com/ostreedev/ostree/issues/2307

cgwalters:
(Will make a hackmd.io doc for release notes on Monday)

jlebon (approved):
Hmm, GitHub Action for release commit sanity check is assuming it has a `config.h` even though it's not building the tree.

CoreOS CI was failing on the openat issue. I restarted it now that https://github.com/ostreedev/ostree/pull/2306 was merged.

cgwalters:
> Hmm, GitHub Action for release commit sanity check is assuming it has a config.h even though it's not building the tree.

Yeah, I think that was ported to GH actions from the old CI but without testing it.

PR #2311, opened 2021-03-25
switchroot: fix reading /proc/cmdline with multiple lines

nikita-dubrovskii:
/proc/cmdline may contian multiple lines ending with '\n',
but only the first line would be taken, so we may end up with:
```
ostree-prepare-root: No OSTree target; expected ostree=/ostree/boot.N/...
```
To prevent this we have to read entire file and use both '\n' and ' ' as separators.

This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1941302

Signed-off-by: Nikita Dubrovskii <nikita@linux.ibm.com>

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -35,6 +35,10 @@
 #define INITRAMFS_MOUNT_VAR "/run/ostree/initramfs-mount-var"
 #define _OSTREE_SYSROOT_READONLY_STAMP "/run/ostree-sysroot-ro.stamp"
 
+// This limit depends on the architecture and is between 256 and 4096 characters.
+// It is defined in the file ./include/asm/setup.h as COMMAND_LINE_SIZE.
+#define COMMAND_LINE_SIZE 4096
+
 static inline int
 path_is_on_readonly_fs (const char *path)
 {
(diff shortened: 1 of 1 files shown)


jlebon:
> /proc/cmdline may contian multiple lines ending with '\n'

OK, that's interesting. I didn't realize that was supported. But indeed, systemd for example supports this: https://github.com/systemd/systemd/pull/11983.

/ok-to-test

src/switchroot/ostree-mount-util.h
```diff
@@ -50,58 +54,33 @@ static inline char *
...
-  if (cmdline[len-1] == '\n')
-    cmdline[len-1] = '\0';
-out:
-  if (f)
-    fclose (f);
-  return cmdline;
+  char buf[COMMAND_LINE_SIZE] = {'\0'};
+  size_t len = fread(&buf, 1, sizeof (buf), f);
+  fclose(f);
+  return strndup(buf, len);
```
jlebon, line 63: Hmm, can we make this be something like

```suggestion
  return strndup (buf, MIN(len, sizeof (buf)));
```

instead to be safe?

src/switchroot/ostree-mount-util.h
```diff
@@ -50,58 +54,33 @@ static inline char *
...
+  char *cmdline = read_proc_cmdline ();
   char *ret = NULL;
 
-  cmdline = read_proc_cmdline ();
   if (!cmdline)
     err (EXIT_FAILURE, "failed to read /proc/cmdline");
 
-  iter = cmdline;
-  while (iter != NULL)
+  const char* next = strtok (cmdline, " \n");
```
jlebon, line 75: Should be we use `strtok_r` instead to be safe? This code might end up getting used in a context where it matters in the future.

src/switchroot/ostree-mount-util.h
nikita-dubrovskii: why should we? `fread` couldn't return more than `sizeof(buf)`

src/switchroot/ostree-mount-util.h
nikita-dubrovskii: i was thinking about it, but this function is only called within `switchroot`'s  `main` functions (which are pretty straightforward and simple), and `pthreads` are not used there. 
is it worth doing this code more complex right now?

jlebon:
So actually, thinking more on this and trying it out locally, there's an obvious issue with this: OSTree itself can't produce valid configs with newlines, because it can't be correctly rendered into a BLS entry (AFAICT, the spec has no way to escape newlines or special characters).

We could support *reading* newlines, but it's of limited utility if it's actually impossible to carry forward that kernel cmdline forward in new deployments. Looking at the RHBZ you linked to, it seems like they were able to fix their config without issues. So overall, I don't think this is worth doing until there's a more compelling use case.

jlebon:
The way I figured this out is I did:

```
$ rpm-ostree kargs --append "foo
bar"
```

and while OSTree finalization was happy to create a broken BLS config, it of course wasn't parsed correctly by GRUB's `blscfg` because newlines separate key value pairs.

Tentatively closing; feel free to reopen if you'd like to discuss this more!

cgwalters:
https://www.freedesktop.org/wiki/Specifications/BootLoaderSpec/ is very light on details here, but clearly `options` could e.g. support escapes like `\n`.  I don't have a strong opinion if that's a good idea or not.  If we chose to support it it'd likely lead to a mess of trying to get the various bootloaders to support it.

PR #2314, opened 2021-03-26
cci: Update for buildroot changes

cgwalters:
See latest in https://github.com/coreos/fedora-coreos-tracker/blob/master/docs/ci-and-builds.md

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -3,7 +3,7 @@
 stage("Build") {
 parallel normal: {
   def n = 5
-  cosaPod(buildroot: true, runAsUser: 0, memory: "2Gi", cpu: "${n}") {
+  buildPod(runAsUser:0, memory: "2Gi", cpu: "${n}") {
       checkout scm
       stage("Core build") {
         shwrap("""
@@ -21,19 +21,25 @@ parallel normal: {
             make syntax-check
           """)
         } finally {
+            shwrap("cat test-suite.log || true")
             archiveArtifacts allowEmptyArchive: true, artifacts: 'test-suite.log'
         }
       }
-      shwrap("""
-        make install DESTDIR=\$(pwd)/insttree/
-        tar -c -C insttree/ -zvf insttree.tar.gz .
-      """)
-      stash includes: 'insttree.tar.gz', name: 'build'
+      stage("Build installed tests") {
+        shwrap("make -C tests/kolainst")
+      }
+      stage("Generate artifacts") {
+        shwrap("""
+          make install DESTDIR=\$(pwd)/installed/rootfs
+          make -C tests/kolainst install DESTDIR=\$(pwd)/installed/tests
+        """)
+      }
+      stash includes: "installed/", name: 'build'
   }
 },
 // A minimal build, helps test our build options
 minimal: {
-  cosaPod(buildroot: true, runAsUser: 0) {
+  buildPod(runAsUser:0) {
       checkout scm
       shwrap("""
         git submodule update --init
@@ -48,7 +54,7 @@ minimal: {
   }
 },
 codestyle: {
-  cosaPod {
+  buildPod(runAsUser:0) {
       checkout scm
       shwrap("""
         # Jenkins by default only fetches the branch it's testing. Explicitly fetch master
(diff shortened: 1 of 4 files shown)


jlebon:
Hmm, bunch of failures and for some reason it's not archiving `test-suite.log`. Try inlining a `cat test-suite.log` maybe?

cgwalters:
OK yeah I should have immediately pulled the buildroot image and run the same commands locally; being able to trivially do so is a huge point of all of this.

The failure was then obvious: https://github.com/coreos/fedora-coreos-config/pull/911

cgwalters:
Well, that's just exciting:

```
       Message: Process 570832 (mv) of user 100999 dumped core.
                
                Stack trace of thread 57473:
                #0  0x00007f242b809442 __read (libc.so.6 + 0xf2442)
                #1  0x00007f242b79a122 _IO_file_underflow@@GLIBC_2.2.5 (libc.so.6 + 0x83122)
                #2  0x00007f242b78e1b8 __getdelim (libc.so.6 + 0x771b8)
                #3  0x000055a8af2c52ad yesno (/usr/bin/mv + 0x92ad)
                #4  0x000055a8af2c99b6 copy_internal (/usr/bin/mv + 0xd9b6)
                #5  0x000055a8af2cd133 do_move (/usr/bin/mv + 0x11133)
                #6  0x000055a8af2c09be main (/usr/bin/mv + 0x49be)
                #7  0x00007f242b73f1e2 __libc_start_main (libc.so.6 + 0x281e2)
                #8  0x000055a8af2c0c0e _start (/usr/bin/mv + 0x4c0e)
```

cgwalters:
OK one thing going on here is I think ye olde umask bug - we're suddenly triggering `mv` to warn about replacing existing files;
```
++ mv ostree-srv/gnomerepo/objects/6d/8ddc0347c6bdf6494972713cee589ad760449b3495f2957f60f58c6900f8e3.commitmeta.bak ostree-srv/gnomerepo/objects/6d/8ddc0347c6bdf6494972713cee589ad760449b3495f2957f60f58c6900f8e3.commitmeta
mv: replace 'ostree-srv/gnomerepo/objects/6d/8ddc0347c6bdf6494972713cee589ad760449b3495f2957f60f58c6900f8e3.commitmeta', overriding mode 0000 (---------)? +++ run_exit_cmds
```

I may try to just hack in a umask again here and *not* dive into figuring out where umask is really coming from...

cgwalters:
OK I think I understand one major problem - we're now running the build and tests in the workspace instead of the `/srv` emptydir.  Or at least I'm pretty sure in Jenkins we were doing builds in `/srv` before.  At least locally with podman I definitely have ostree issues interacting with `fuse-overlayfs` (it's also slooooow).

jlebon:
> OK I think I understand one major problem - we're now running the build and tests in the workspace instead of the `/srv` emptydir. Or at least I'm pretty sure in Jenkins we were doing builds in `/srv` before. At least locally with podman I definitely have ostree issues interacting with `fuse-overlayfs` (it's also slooooow).

cosa builds are done in `/srv`, but there shouldn't be any changes to component builds (it was built in the workspace before too).

jlebon:
> [2021-03-26T22:54:14.751Z] gawk: cmd. line:392: (FILENAME=- FNR=2) fatal: cannot open pipe `cat >&3' for output: Cannot allocate memory

Hmm, that's odd. There's no difference between `cosaPod` and `buildPod` wrt memory requests, but I guess... let's try bumping the memory request to e.g. 3Gi?

jlebon:
Ohh one major difference is that `cosaPod` uses `dumb-init` but not `buildPod`. Hmm, yeah we should definitely do that:
- https://github.com/coreos/fedora-coreos-config/pull/916
- https://github.com/coreos/coreos-ci-lib/pull/69

jlebon:
(Restarted CI here now that we have `dumb-init`.)

cgwalters:
OK yeah that's working *much* better.  I see a missing `python3-pyyaml` dep; testing a fix and validating we're not missing anything else.

jlebon:
Hmm, not clear what went wrong here. It looks like `read line` failed somehow. Maybe try that `make -C tests/kolainst install DESTDIR=...` invocation locally in the buildroot container?

cgwalters:
Hooray!  CI is passing here.

PR #2315, opened 2021-03-26
.lgtm.yml: Add package deps

cgwalters:
I was recently looking at systemd's CI and came across this.
AFAICS it's a proprietary analysis but it looks useful.

We still have `clang-analyzer` for example for FOSS analysis.


jlebon:
Hmm, is this missing some more configuration to have it analyze the C bits?

.lgtm.yml
```diff
@@ -0,0 +1,40 @@
+# See https://lgtm.com/
+extraction:
+  cpp:
+    prepare:
+      packages:
+      - "autoconf-archive"
```
cgwalters, line 6: I cargo culted these from `travis-install.sh` - obviously annoying to duplicate, at least since this one is declarative I can see about parsing it for our Debian/Ubuntu builds.

cgwalters:
I think it may not actually take effect until it lands in master.

Other references I used for this:

- https://github.com/tpm2-software/tpm2-tools/blob/master/.lgtm.yml
- https://github.com/systemd/systemd/blob/main/.lgtm.yml

So let's see?

cgwalters:
(I just enabled required reviews on this repo btw)

jlebon (approved):

PR #2316, opened 2021-03-26
Two lgtm.com analysis test commits

cgwalters:
manual-tests: Fix unused variables

lgtm.com harshly makes our JS grade "D" due to these unused
variables.

---

repo: Refactor ostree_repo_remote_list() to decl-and-init style

Mostly doing this to tickle lgtm.com's detection of C/C++ changes.

---

### DIFF
--- a/manual-tests/upgrade-loop.js
+++ b/manual-tests/upgrade-loop.js
@@ -24,9 +24,6 @@
 // could watch the output, and assert that the system is in a
 // consistent state if this script is killed and restarted. randomly.
 
-const GLib = imports.gi.GLib;
-const Gio = imports.gi.Gio;
-
 const OSTree = imports.gi.OSTree;
 
 let sysroot = OSTree.Sysroot.new_default();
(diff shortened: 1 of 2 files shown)


cgwalters:
OK I ran through https://lgtm.com/help/lgtm/adding-language-to-project and I think it's working now, let's see.

src/libostree/ostree-repo.c
```diff
@@ -1928,17 +1927,15 @@ ostree_repo_remote_list (OstreeRepo *self,
...
     {
-      GList *list, *link;
       guint ii = 0;
 
       remotes = g_new (char *, n_remotes + 1);
 
-      list = g_hash_table_get_keys (remotes_ht);
-      list = g_list_sort (list, (GCompareFunc) strcmp);
+      GList *list = g_hash_table_get_keys (remotes_ht);
+      GList *link = g_list_sort (list, (GCompareFunc) strcmp);
```
lucab, line 1938: I think this changes the actual logic. The for-loop below seems to rely on `list`, which should be sorted at that point, while `link` is set up directly in the init step.

cgwalters: Thanks, I just dropped that commit; it was mostly to test LGTM.com's C/C++ handling which as it turns out is less useful than I thought.

PR #2319, opened 2021-03-30
repo: Fix load_variant_if_exists to return a nullable value

cgwalters:
Another introspection fix I hit when trying to use this in Rust.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3686,6 +3686,10 @@ load_metadata_internal (OstreeRepo       *self,
   g_return_val_if_fail (OSTREE_OBJECT_TYPE_IS_META (objtype), FALSE);
   g_return_val_if_fail (objtype == OSTREE_OBJECT_TYPE_COMMIT || out_state == NULL, FALSE);
 
+  /* Ensure this is set to NULL if we didn't find the object */
+  if (out_variant)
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo.c
```diff
@@ -4452,12 +4452,13 @@ ostree_repo_query_object_storage_size (OstreeRepo           *self,
...
  * @objtype: Object type
  * @sha256: ASCII checksum
- * @out_variant: (out) (transfer full): Metadata
+ * @out_variant: (out) (nullable) (transfer full): Metadata
  * @error: Error
  *
  * Attempt to load the metadata object @sha256 of type @objtype if it
  * exists, storing the result in @out_variant.  If it doesn't exist,
- * %NULL is returned.
+ * @out_variant will be set to %NULL and the function will still
```
jlebon, line 4460: Hmm, AFAICT `*out_variant` isn't explicitly set to `NULL` in the `ENOENT` case.

cgwalters: Good catch!  In practice for callers they usually need to zero-initialize anyways for reliability, but indeed let's be explicit here.

jlebon:
Oh right, this still needs https://github.com/ostreedev/ostree/pull/2314.

PR #2323, opened 2021-04-07
build-sys: Remove --enable-experimental-api

cgwalters:
It was added for the collections bits, but we made that stable.
It's now just cruft and we're very unlikely to reuse the infrastructure
again.

Motivated by a unit test failure when running from a tarball:
https://github.com/ostreedev/ostree/issues/2313

### DIFF
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -107,6 +107,7 @@ OSTREE_SUMMARY_GVARIANT_STRING
 OSTREE_SUMMARY_GVARIANT_FORMAT
 ostree_metadata_variant_type
 ostree_validate_checksum_string
+ostree_validate_collection_id
 ostree_checksum_to_bytes
 ostree_checksum_to_bytes_v
 ostree_checksum_from_bytes
@@ -304,6 +305,7 @@ ostree_repo_is_system
 ostree_repo_is_writable
 ostree_repo_create_at
 ostree_repo_create
+ostree_repo_get_collection_id
 ostree_repo_get_bootloader
 ostree_repo_get_path
 ostree_repo_get_mode
@@ -319,6 +321,7 @@ ostree_repo_remote_delete
 OstreeRepoRemoteChange
 ostree_repo_remote_change
 ostree_repo_remote_list
+ostree_repo_remote_list_collection_refs
 ostree_repo_remote_get_url
 ostree_repo_remote_get_gpg_verify
 ostree_repo_remote_get_gpg_verify_summary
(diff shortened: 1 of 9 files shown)


apidoc/ostree-sections.txt
```diff
@@ -618,17 +618,6 @@ ostree_remote_get_url
...
-<SECTION>
-<FILE>ostree-repo-experimental</FILE>
-ostree_repo_find_remotes_async
-ostree_repo_find_remotes_finish
-ostree_repo_pull_from_remotes_async
-ostree_repo_pull_from_remotes_finish
-ostree_repo_resolve_keyring_for_collection
-OSTREE_REPO_METADATA_REF
-OSTREE_META_KEY_DEPLOY_COLLECTION_ID
-</SECTION>
```
dbnicholson: I think this is and the hunk below are unintended. They are no longer experimental after 8fbf19c9f54ba34ae7ebbc540b08566b8dd18848 and I think this was just not clearly renaming the section when shifting it from `apidoc/ostree-experimental-sections.txt` in that commit. Probably these should be moved into the regular `ostree-repo` section.

apidoc/ostree-sections.txt
```diff
@@ -683,18 +691,6 @@ ostree_repo_finder_override_add_uri
 ostree_repo_finder_override_get_type
 </SECTION>
 
-<SECTION>
-<FILE>ostree-misc-experimental</FILE>
-ostree_repo_get_collection_id
-ostree_repo_set_collection_id
-ostree_validate_collection_id
```
dbnicholson, line 690: I think this should be in the `ostree-core` section and the others should be in the `ostree-repo` section. They're normal APIs and the doc section just didn't get updated in 8fbf19c9f54ba34ae7ebbc540b08566b8dd18848.

apidoc/.gitignore
```diff
@@ -49,5 +49,4 @@
 /tmpl/*.bak
 /tmpl/ostree-unused.sgml
 /xml
-/version.xml
```
dbnicholson, line 52: Were these changes intentional? I can never tell WTF is going on with this file, but usually this happens for me when I don't configure with `--enable-gtk-doc`.

dbnicholson (changes requested):

cgwalters:
OK, I think I got the gtk-doc adjustments right.

dbnicholson (approved):
I didn't look super closely at all the changes, but in principle I agree and the CI seems happy.

cgwalters:
OK, I just changed two things here:

- PR review approval is now required for merge
- Auto-merge is now available

So if you're comfortable mind clicking the latter?

dbnicholson:
I approved before, but now I will merge. Is it expected that one of the kola tests has been running for 1.5 hours or is it just hung?

cgwalters:
Ah sorry, I meant to do auto-merge on CI green, not to force merge.  

>  Is it expected that one of the kola tests has been running for 1.5 hours or is it just hung?

Some of the tests are expensive and the cluster can get loaded, and we have a lot of tests but...it looks like the main culprit is a recent flake

```
[2021-04-07T21:33:34.547Z] --- FAIL: coreos.boot-mirror.luks (5027.68s)

[2021-04-07T21:33:34.547Z]         boot-mirror.go:391: machine "c396dc6d-e7aa-4273-aa3b-458def908ff6" failed to start: ssh journalctl failed: ssh: handshake failed: read tcp 127.0.0.1:34292->127.0.0.1:43897: read: connection reset by peer
```

PR #2325, opened 2021-04-07
API doc gitignore cleanup

dbnicholson:
I think this should make it such that I never accidentally include diffs to apidoc/.gitignore in my commits again.


cgwalters (approved):
Thanks!

PR #2331, opened 2021-04-08
travis: Bump Ubuntu versions

cgwalters:
xenial is EOL soon, let's bump to the latest LTS and the latest.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ env:
   # debian has libsodium-dev, ubuntu doesn't in core at least
   - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_configopts="--with-ed25519-libsodium" ci_pkgs="libsodium-dev"
   - ci_docker=debian:buster-slim ci_distro=debian ci_suite=stretch ci_configopts="--with-curl --with-ed25519-libsodium --without-gpgme" ci_pkgs="libsodium-dev"
-  - ci_docker=ubuntu:xenial ci_distro=ubuntu ci_suite=xenial
-  - ci_docker=i386/ubuntu:bionic ci_distro=ubuntu ci_suite=bionic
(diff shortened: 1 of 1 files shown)


dbnicholson:
Hmm, perhaps I was a bit hasty. The focal build blocked waiting to be told what timezone to use at https://travis-ci.org/github/ostreedev/ostree/jobs/766353610#L1948:
```
Setting up tzdata (2021a-0ubuntu0.20.04) ...
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
Configuring tzdata
------------------
Please select the geographic area in which you live. Subsequent configuration
questions will narrow this down by presenting a list of cities, representing
the time zones in which they are located.
  1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
  2. America     5. Arctic     8. Europe    11. SystemV
  3. Antarctica  6. Asia       9. Indian    12. US
Geographic area: 
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```
That's lame. I really don't think we should have to resort to adding a debconf entry for this.

dbnicholson:
Apparently if `/etc/timezone` already exists, then you skip all the debconf stuff. I think this is an issue because the newer ubuntu containers don't include tzdata. Anyways, should be easy enough to fix and just put `Etc/UTC` in `/etc/timezone` like debian does on their images and ubuntu used to do.

cgwalters:
Ahhh I see, I need to make some more of these CI contexts blocking; that will ensure the button is "merge when CI passes".

dbnicholson:
I think #2332 should fix this in a more general way.

PR #2332, opened 2021-04-08
travis: Don't ask any debconf questions when installing packages

dbnicholson:
Currently the Ubuntu builds are stuck waiting for an answer on what
timezone to use. That could be fixed, but generally the way to do these
types of installs is to set the DEBIAN_FRONTEND environment variable to
`noninteractive` so that debconf just chooses a default. This is what
debootstrap does, for instance. I tested installing tzdata on a local
focal container this way and it just chooses `Etc/UTC` as the timezone.


cgwalters (dismissed):

dbnicholson:
The i386/ubuntu:focal build failed because they dropped i386 - https://wiki.ubuntu.com/FocalFossa/ReleaseNotes#Upgrades_on_i386. I think we should just remove it and get on with life.

cgwalters (approved):

cgwalters:
Ah yes.  Though we are going to need a 32 bit build, because a lot of people are using ostree on armhf (32 bit).  Will look at that at some point...

dbnicholson:
> Ah yes. Though we are going to need a 32 bit build, because a lot of people are using ostree on armhf (32 bit). Will look at that at some point...

Good point. I'll just add an `i386/debian:buster-slim` build then.

PR #2333, opened 2021-04-08
Add back a travis 32 bit build

dbnicholson:
To try to catch issues that would appear on more frequently deployed 32 bit armv7.


cgwalters:
Thanks for doing this!

PR #2335, opened 2021-04-09
repo: Add ostree_repo_write_regfile

cgwalters:
This API is push rather than pull, which makes it much more
suitable to use cases like parsing a tar file from external
code.

Now, we have a large mess in this area internally because
the original file writing code was pull based, but static
deltas hit the same problem of wanting a push API, so I added
this special `OstreeRepoBareContent` just for writing regular
files from a push API.

Eventually...I'd like to deprecate the pull based API,
and rework things so that for regular files the push API
is the default, and then `write_content_object()` would
be split up into archive/bare cases.

In this world the `ostree_repo_write_content()` API would
then need to hackily bridge pull to push and it'd be
less efficient.

Anyways for now due to this bifurcation, this API only
works on non-archive repositories, but that's fine for
now because that's what I want for the `ostree-ext-container`
bits.

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -66,6 +66,8 @@ libostree_1_la_SOURCES = \
 	src/libostree/ostree-checksum-input-stream.h \
 	src/libostree/ostree-chain-input-stream.c \
 	src/libostree/ostree-chain-input-stream.h \
+	src/libostree/ostree-content-writer.c \
+	src/libostree/ostree-content-writer.h \
 	src/libostree/ostree-lzma-common.c \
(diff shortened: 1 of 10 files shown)


cgwalters:
Hmm this one isn't showing up in the Rust binding generation, trying to figure out why.

cgwalters:
OK it does work, I just had to learn how to use gtk-rs a bit better; the list of objects to expose is manual.

Working patch in https://github.com/ostreedev/ostree-rs-ext/pull/8 which depends https://github.com/ostreedev/ostree-rs/pull/2

jlebon (approved):
Nice!

PR #2340, opened 2021-04-14
Add an API to verify a commit signature explicitly

cgwalters:
We have a bunch of APIs to do GPG verification of a commit,
but that doesn't generalize to signapi.  Further, they
require the caller to check the signature status explicitly
which seems like a trap.

This much higher level API works with both GPG and signapi.
The intention is to use this in things that are doing "external
pulls" like the ostree-ext tar import support.  There we will
get the commitmeta from the tarball and we want to verify it
at the same time we import the commit.


cgwalters:
I think we probably also want a high level API that does the validation and writes the commit object + partial state as one unit.

src/libostree/ostree-repo-pull-verify.c
```diff
@@ -346,6 +347,111 @@ _process_gpg_verify_result (OtPullData            *pull_data,
...
+                                          GError       **error)
+{
+  g_assert (self);
+  g_assert (remote_name);
+  g_assert (commit_data);
+
+  gboolean gpg = !(flags & OSTREE_REPO_VERIFY_FLAGS_NO_GPG);
+  gboolean signapi = !(flags & OSTREE_REPO_VERIFY_FLAGS_NO_SIGNAPI);
+  // Must ask for at least one type of verification
+  g_assert (gpg || signapi);
```
lucab, line 389: I think this would be nicer as a `glnx_throw`.

src/libostree/ostree-repo-pull-verify.c
```diff
@@ -346,6 +347,111 @@ _process_gpg_verify_result (OtPullData            *pull_data,
...
+  if (commit_metadata == NULL)
+    return glnx_throw (error, "Can't verify commit without detached metadata");
+  if (!validate_metadata_size ("Commit metadata", commit_metadata, error))
+    return FALSE;
+  g_autoptr(GVariant) commit_metadata_v = g_variant_new_from_bytes (G_VARIANT_TYPE_VARDICT, commit_metadata, FALSE);
+
+  g_autoptr(GString) results_buf = g_string_new ("");
+  gboolean verified = FALSE;
+
+  if (!ostree_repo_remote_get_gpg_verify (self, remote_name,
```
lucab, line 403: Reading this function from here on, I'm left a bit confused.
There are operations to setup GPG done even with `NO_GPG` flags, and same goes for signapi counterpart.
I would have instead expected two major separate blocks, gated by the relevant flag, and stopping as soon as one of the requested verification method fails. Any reason for approaching it in this intermixed way instead?

src/libostree/ostree-repo-pull-verify.c
```diff
@@ -346,6 +347,111 @@ _process_gpg_verify_result (OtPullData            *pull_data,
...
+    return FALSE;
+
+  if (!(gpg || signapi_verifiers))
+    return glnx_throw (error, "Cannot verify commit for remote %s; GPG verification disabled, and no signapi verifiers configured", remote_name);
+
+#ifndef OSTREE_DISABLE_GPGME
+  if (gpg)
+    {
+      g_autoptr(OstreeGpgVerifyResult) result =
+        _ostree_repo_gpg_verify_with_metadata (self, commit_data,
```
lucab, line 419: I think this may be missing a NULL check.

src/ostree/ot-admin-builtin-status.c
```diff
@@ -31,7 +31,10 @@
 
 #include <glib/gi18n.h>
 
+static gboolean opt_verify;
+
 static GOptionEntry options[] = {
+  { "verify", 'V', 0, G_OPTION_ARG_NONE, &opt_verify, "Print the commit verification status", NULL },
```
lucab, line 37: I'm not familiar with signed setups, but this does not seem to offer a way to choose between GPG and signapi verification. Are they usually either both enabled or both disabled? Wouldn this be better suited as a verb with flags for different methods?

src/libostree/ostree-repo-pull-verify.c
cgwalters: Hmm; it's a calling-program error, invoking the API in an invalid way.  In Rust this would be a `panic!()` e.g.

But eh, changed.

src/libostree/ostree-repo-pull-verify.c
cgwalters: Yes!  Hooray for code review.

src/ostree/ot-admin-builtin-status.c
cgwalters: I think using both is supported, but not expected (except perhaps as a transition period).  But if both are enabled in the remote config, then I'd generally expect us to do both.

I can't think of a need for a CLI consumer to explicitly just check one.

src/libostree/ostree-repo-pull-verify.c
cgwalters: The intent here is to throw an error upfront if no signature types are enabled in the repo.  That is a separate error (IMO) from the case where the program caller disables one or both. 

I think having them in each block would look cleaner...but OTOH with the current flow I like this "double check" we have with the `verified = TRUE;` flag.

lucab (approved):

Issue #2286, opened 2021-02-26
Make repo locking API public


dbnicholson:
Back in #1292 and temporarily in it's slimmer cousin #1555 the repo locking APIs were public. I think it would be useful to have the API public, particularly for cases when you're going to do repo maintenance and want to ensure no one else is modifying the repo. On our Endless ostree server I've wanted this many times for doing unusual operations on repos to the point that I basically copied the whole locking scheme into python and our tooling relies on that to work.

Now we have a use case similar to https://raw.githubusercontent.com/endlessm/eos-meta/master/eos-tech-support/eos-fix-ostree-repo where we want to make dangerous changes on client repos. In particular, we want to undo the dubious decision we made many years ago to have flatpak use `/ostree/repo` as its repo. In order to split these up reliably we really want to make sure none of our various ostree programs (gnome-software and eos-updater primarily) can alter the repo while the surgery is happening.

Anyways, I think the locking has proven to be robust, although I do share your concern about deadlocking when there are multiple threads in play. I don't know what to do about it besides tell apps to be careful with exclusive locks, though.

What do you think?

cgwalters:
Yeah, I think exposing the APIs makes sense.

Issue #2337, opened 2021-04-12
Wrong locking during more complex prunes


alexlarsson:
While doing some research into pruning I noticed that in ot-builtin-prune.c, if we take the more complex path that uses `ostree_repo_prune_from_reachable()` the exclusive repo lock is only held inside `ostree_repo_prune_from_reachable()`, so we compute the reachable set outside the lock. 

If this where to race with a commit operation that created a new commit which we didn't traverse then it would be pruned away.

cgwalters:
Yeah, I think you're right.  We clearly need a test that tries to reproduce this race.  Also, fixing this would require fixing https://github.com/ostreedev/ostree/issues/2286

jlebon:
Oh hey, this is fixed by https://github.com/ostreedev/ostree/pull/2808!

PR #2341, opened 2021-04-15
repo: Make locking APIs public

cgwalters:
Doing anything even somewhat sophisticated requires this;
turns out our own `ostree prune` CLI wants this, e.g.
https://github.com/ostreedev/ostree/issues/2337

Closes: https://github.com/ostreedev/ostree/issues/2286

### DIFF
--- a/apidoc/ostree-sections.txt
+++ b/apidoc/ostree-sections.txt
@@ -319,6 +319,10 @@ ostree_repo_get_min_free_space_bytes
 ostree_repo_get_config
 ostree_repo_get_dfd
 ostree_repo_get_default_repo_finders
+ostree_repo_lock_pop
+ostree_repo_lock_push
+ostree_repo_auto_lock_push
+ostree_repo_auto_lock_cleanup
 ostree_repo_hash
 ostree_repo_equal
 ostree_repo_copy_config
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -444,7 +444,7 @@ pop_repo_lock (OstreeRepo  *self,
   return TRUE;
 }
 
-/*
+/**
  * ostree_repo_lock_push:
  * @self: a #OstreeRepo
  * @lock_type: the type of lock to acquire
@@ -470,9 +470,10 @@ pop_repo_lock (OstreeRepo  *self,
  * %TRUE is returned.
  *
  * Returns: %TRUE on success, otherwise %FALSE with @error set
+ * Since: 2021.2
  */
 gboolean
-_ostree_repo_lock_push (OstreeRepo          *self,
+ostree_repo_lock_push (OstreeRepo          *self,
                         OstreeRepoLockType   lock_type,
                         GCancellable        *cancellable,
                         GError             **error)
@@ -538,8 +539,8 @@ _ostree_repo_lock_push (OstreeRepo          *self,
     }
 }
 
-/*
- * _ostree_repo_lock_pop:
+/**
+ * ostree_repo_lock_pop:
  * @self: a #OstreeRepo
  * @cancellable: a #GCancellable
  * @error: a #GError
@@ -560,11 +561,12 @@ _ostree_repo_lock_push (OstreeRepo          *self,
  * %TRUE is returned.
  *
  * Returns: %TRUE on success, otherwise %FALSE with @error set
+ * Since: 2021.2
  */
 gboolean
-_ostree_repo_lock_pop (OstreeRepo    *self,
-                       GCancellable  *cancellable,
-                       GError       **error)
+ostree_repo_lock_pop (OstreeRepo    *self,
+                      GCancellable  *cancellable,
+                      GError       **error)
 {
   g_return_val_if_fail (self != NULL, FALSE);
   g_return_val_if_fail (OSTREE_IS_REPO (self), FALSE);
@@ -628,7 +630,7 @@ _ostree_repo_lock_pop (OstreeRepo    *self,
 }
 
 /*
- * _ostree_repo_auto_lock_push: (skip)
+ * ostree_repo_auto_lock_push: (skip)
  * @self: a #OstreeRepo
  * @lock_type: the type of lock to acquire
  * @cancellable: a #GCancellable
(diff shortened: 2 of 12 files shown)


tests/test-core.js
```diff
@@ -134,4 +134,12 @@ w.write(inline_content.slice(10), null)
 let actual_checksum = w.finish(null)
 assertEquals(actual_checksum, networks_checksum)
 
+// Basic locking API sanity test
+repo.lock_push(OSTree.RepoLockType.SHARED, null);
+repo.lock_push(OSTree.RepoLockType.SHARED, null);
+repo.lock_pop(null);
+repo.lock_pop(null);
+repo.lock_push(OSTree.RepoLockType.EXCLUSIVE, null);
+repo.lock_pop(null);
```
dbnicholson, line 143: Since this wasn't public, I ended up writing basically the same locking implementation in python for our (sadly internal) tooling. I ended up writing a bunch of tests for it and I'd say a major improvement is to open another `OstreeRepo` so you have 2 different threads competing for the lock (rather than trying to write a multi-threaded test from a single `OstreeRepo`. So:
```
let repo2 = OSTree.Repo.new(repoPath);
repo2.open(null);
repo.lock_push(...)
repo2.lock_push(...)
```

Another thing I'd strongly suggest is setting the repo lock timeout to `0` so that any issues here don't just block the test suite for 30 seconds:
```
config = repo.copy_config()
config.set_integer("core", "lock-timeout-secs", 0)
repo.write_config(config)
```
I'd like to be able to change the locking configuration at runtime without using the repo configuration, but that can be handled later.

Finally, testing the actual error states would be great. I think this is how it would go in JS:
```
repo.lock_push(OSTree.RepoLockType.EXCLUSIVE, null);
assertThrows("Locking repo", {
  repo2.lock_push(OSTree.RepoLockType.EXCLUSIVE, null);
})
```

dbnicholson (changes requested):
If you want to wait, that's fine, but I think `OstreeRepoAutoLock` APIs should be public, too. It makes it vastly simpler to use from C since you don't have to litter the code with a bunch of `ostree_repo_lock_pop` calls in the error paths.

I had some suggestions for more testing, but I can also follow up with a copy of the tests I wrote for our python implementation of this.

cgwalters:
> If you want to wait, that's fine, but I think OstreeRepoAutoLock APIs should be public, too.

Yeah, fair.  OK I think particularly given https://github.com/ostreedev/ostree/issues/2344 too we may want a bit more time to think about the locking API.  I'm going to defer this to the next release.

apidoc/ostree-sections.txt
```diff
@@ -319,6 +319,10 @@ ostree_repo_get_min_free_space_bytes
 ostree_repo_get_config
 ostree_repo_get_dfd
 ostree_repo_get_default_repo_finders
+ostree_repo_lock_pop
+ostree_repo_lock_push
+ostree_repo_auto_lock_push
```
dbnicholson, line 324: I think `OstreeRepoAutoLock` should be here or the type won't be documented.

src/libostree/ostree-repo.c
```diff
@@ -470,9 +470,10 @@ pop_repo_lock (OstreeRepo  *self,
...
  *
  * Returns: %TRUE on success, otherwise %FALSE with @error set
+ * Since: 2021.2
  */
 gboolean
-_ostree_repo_lock_push (OstreeRepo          *self,
+ostree_repo_lock_push (OstreeRepo          *self,
                         OstreeRepoLockType   lock_type,
                         GCancellable        *cancellable,
                         GError             **error)
```
dbnicholson, line 479: Indentation is off now.

src/libostree/ostree-repo.h
```diff
@@ -1498,6 +1498,48 @@ gboolean ostree_repo_regenerate_summary (OstreeRepo     *self,
                                          GCancellable   *cancellable,
                                          GError        **error);
 
+
+/**
+ * OstreeRepoLockType:
+ * @OSTREE_REPO_LOCK_SHARED: A "read only" lock; multiple readers are allowed.
+ * @OSTREE_REPO_LOCK_EXCLUSIVE: A writable lock at most one writer can be active, and zero readers.
```
pwithnall, line 1505:
```suggestion
 * @OSTREE_REPO_LOCK_EXCLUSIVE: A writable lock: at most one writer can be active, and zero readers.
```

src/libostree/ostree-repo.h
```diff
@@ -1498,6 +1498,48 @@ gboolean ostree_repo_regenerate_summary (OstreeRepo     *self,
...
+                                     GCancellable        *cancellable,
+                                     GError             **error);
+_OSTREE_PUBLIC
+gboolean      ostree_repo_lock_pop (OstreeRepo    *self,
+                                    GCancellable  *cancellable,
+                                    GError       **error);
+
+/* C convenience API only */
+#ifndef __GI_SCANNER__
+typedef OstreeRepo OstreeRepoAutoLock;
```
pwithnall, line 1528: I suspect this will need a gtk-doc comment to keep gtk-doc happy. Probably something like:
```
/**
 * OstreeRepoAutoLock:
 *
 * Opaque type. See ostree_repo_auto_lock_push() for details.
 *
 * Since: 2021.2
 */
```

tests/test-repo.c
```diff
@@ -249,6 +249,29 @@ test_write_regfile_api (Fixture *fixture,
   g_assert_cmpstr (checksum, ==, "23a2e97d21d960ac7a4e39a8721b1baff7b213e00e5e5641334f50506012fcff");
 }
 
+/* Just a sanity check of the C autolocking API */
+static void
+test_repo_autolock (Fixture *fixture,
+                        gconstpointer test_data)
```
pwithnall, line 255: Nitpick: Indentation is off here.

Also ‘sanity’ in the comment above could be considered a bit ableist. How about “Basic check of the C autolocking API”?

cgwalters:
Let's obsolete this in favor of https://github.com/ostreedev/ostree/pull/2348 ?  They really need to land together, and we can collaborate on fixups in that PR.

dbnicholson:
Sure. I think I already addressed the comments there.

PR #2342, opened 2021-04-15
tests/inst: Make nondestructive tests runnable as unit tests

cgwalters:
Ideally in the future we change more of our unit tests to
support running installed; we've tried this in the past with
https://wiki.gnome.org/Initiatives/GnomeGoals/InstalledTests

I'd like to pick that back up again.  This takes a step
towards that by having our Rust tests.

To make this even easier, add a `tests/run-installed`
which runs the installed tests (uninstalled, confusingly
but conveniently for now).


cgwalters:
I folded this into https://github.com/ostreedev/ostree/pull/2343 since they're closely related.

PR #2343, opened 2021-04-15
build-sys: Add toplevel workspace Cargo.toml

cgwalters:
build-sys: Add toplevel workspace Cargo.toml

rust-analyzer is happier with this because it understands
the project structure out of the box.

We aren't actually again adding a dependency on Rust/cargo in the core,
this is only used to make `cargo build` work out of the box to build
the Rust test code.

---

tests/inst: Make nondestructive tests runnable as unit tests

Ideally in the future we change more of our unit tests to
support running installed; we've tried this in the past with
https://wiki.gnome.org/Initiatives/GnomeGoals/InstalledTests

I'd like to pick that back up again.  This takes a step
towards that by having our Rust tests.

To make this even easier, add a `tests/run-installed`
which runs the installed tests (uninstalled, confusingly
but conveniently for now).

---


lucab (approved):

cgwalters:
Ah heh right, needed to fix the tests install path.  So this is actually even better now I think.  I also folded in https://github.com/ostreedev/ostree/pull/2342

cgwalters:
Marking as draft to hold until after 2021.2

cgwalters:
2021.2 is out, lifting draft.

Issue #2344, opened 2021-04-15
Improve multi-threaded repo locking


dbnicholson:
Currently the repo locking is entirely thread local, which is very safe but makes it difficult to use in a multi-threaded application. The canonical issue would be if you took an exclusive lock and then spawned some threads with tried to take a shared lock. This would deadlock because the threads could never acquire the lock even though the application holds an exclusive lock on the repo.

This isn't theoretical. Back when I was working on #1292 I wanted to make `ostree_repo_prune_object` take an exclusive lock. However, it's actually used deep within the pull code to try to cleanup a tombstone commit (IIRC). The main thread acquires a shared repo lock when it starts a transaction and then the pull code spawns a bunch of threads. When a pull thread hits this path it deadlocks because it actually has no connection to the main thread lock.

If we're to make the repo locking public, it's pretty likely that someone would take an exclusive lock and then run the pull code. I know I've done that many times in our tooling when doing some maintenance task.

dbnicholson:
I'm sure there's probably a data structure that's more appropriate for this, but this is the idea I came up with.

The actual lock file descriptor is stored in `OstreeRepo` with a mutex guarding it. When threads want to change the lock state, they acquire the mutex then change the flock. The reason I didn't do this initially was because of the recursive nature of the locking. For a single thread, it's easy to have a single linear stack that represents the locking recursion.

For multiple threads that's not possible since they can act in any order. The way I thought of to handle this is to maintain a stack per thread in thread local storage like now so that when an individual thread pops a lock state it knows what type of lock state it's dropping. In the shared mutex guarded code, there would just be 2 counters for how many shared and exclusive states or on the stack. This maintains the latching aspect of it. I.e., the lock stays exclusive until all exclusive states have been popped. When all shared and exclusive states have been popped, the lock is unlocked.

cgwalters:
OK from [your original commit](https://github.com/ostreedev/ostree/commit/4e78ddd2da4c0a2ced9527ef20a30f8e3b8567b8)

> The lock also attempts to be thread safe by storing the lock state in
thread local storage with GPrivate. This means that each thread will
have an independent lock for each repository it opens. There are some
drawbacks to that, but it seemed impossible to manage the lock state
coherently in the face of multithreaded access.

But...the patch Alexander had here originally https://bugzilla.gnome.org/show_bug.cgi?id=759442#c2 was *way* simpler and also by benefit of being file based (like the sysroot lock) also works across processes, which is I think what we actually want.

[...]

PR #2348, opened 2021-04-16
Improve multi-threaded locking

dbnicholson:
As described in #2344, the current repo locking is multithread-safe but doesn't actually allow for coordination between threads using the same `OstreeRepo`. This refactors the locking so it's maintained per-`OstreeRepo` while retaining the thread safety and recursive features.

I'm going to add some more tests, but I wanted to get some review on the actual changes. This is on top of #2341 because I wanted to run the tests added there (and likely build on them).

Fixes: #2344


dbnicholson:
I think this is ready now. I think I covered all the locking cases in the last commit but would be happy to add more.

src/libostree/ostree-repo.c
```diff
@@ -405,26 +412,40 @@ pop_repo_lock (OstreeRepo  *self,
...
+  g_autoptr(GMutexLocker) locker = g_mutex_locker_new (&self->lock.mutex);
+  g_return_val_if_fail (self->lock.fd != -1, FALSE);
 
   OstreeRepoLockInfo info;
-  repo_lock_info (lock, &info);
+  repo_lock_info (self, &info);
   g_return_val_if_fail (info.len > 0, FALSE);
 
   g_debug ("Pop lock: state=%s, depth=%u", info.name, info.len);
+  int state_to_drop = GPOINTER_TO_INT (stack->pdata[stack->len - 1]);
```
lucab, line 426: I think here there is a missing check to prevent `stack->len - 1` from underflowing, which I think can happen in case of unbalanced `unlock()` calls from a consuming thread that previously performed at least one `lock()` call.

src/libostree/ostree-repo.c
```diff
@@ -347,49 +347,56 @@ push_repo_lock (OstreeRepo          *self,
...
   OstreeRepoLockInfo info;
-  repo_lock_info (lock, &info);
+  repo_lock_info (self, &info);
   g_debug ("Push lock: state=%s, depth=%u", info.name, info.len);
 
   if (info.state == LOCK_EX)
     {
       g_debug ("Repo already locked exclusively, extending stack");
-      g_queue_push_head (&lock->stack, GINT_TO_POINTER (LOCK_EX));
+      self->lock.exclusive++;
```
lucab, line 382: I think this can overflow (wrap back to 0), breaking lock correctness.

src/libostree/ostree-repo.c
```diff
@@ -347,49 +347,56 @@ push_repo_lock (OstreeRepo          *self,
...
       g_debug ("Locking repo %s", next_state_name);
-      if (!do_repo_lock (lock->fd, flags))
+      if (!do_repo_lock (self->lock.fd, flags))
         return glnx_throw_errno_prefix (error, "Locking repo %s failed",
                                         next_state_name);
 
-      g_queue_push_head (&lock->stack, GINT_TO_POINTER (next_state));
+      if (next_state == LOCK_EX)
+        self->lock.exclusive++;
+      else
```
lucab, line 397: I think both of these can overflow (wrap back to 0), breaking lock correctness.

src/libostree/ostree-repo.c
```diff
@@ -210,62 +221,51 @@ free_repo_lock_table (gpointer data)
...
-
 typedef struct {
   guint len;
   int state;
   const char *name;
 } OstreeRepoLockInfo;
 
 static void
-repo_lock_info (OstreeRepoLock *lock, OstreeRepoLockInfo *out_info)
+repo_lock_info (OstreeRepo *self, OstreeRepoLockInfo *out_info)
```
lucab, line 243: This seems to internally assume that `self->lock` details are unchanged throughout body execution, i.e. that the function can be called if and only if run the caller already holds the related mutex.
If that is indeed the intended design, it may be nice to let the caller be explicitly aware of that, e.g. by requiring a mutex locker/guard as an input argument.

src/libostree/ostree-repo.c
```diff
@@ -1060,6 +1090,11 @@ ostree_repo_finalize (GObject *object)
         g_private_replace (&repo_lock_table, NULL);
     }
 
+  g_mutex_lock (&self->lock.mutex);
```
lucab, line 1093: It looks like the rest of this function does not try lock the other mutexes, which I guess could be useful in case of a poisoned mutex due to a thread panic.
Is racing between a lock and a finalize operation (e.g. across different threads) a concern for the library?

src/libostree/ostree-repo.c
dbnicholson: Good call, I think that is something that could happen.

src/libostree/ostree-repo.c
dbnicholson: I'm inclined to think that it should fail if you're going to overflow since I can't imagine any real application intentionally recursing the lock that far. What do you think?

src/libostree/ostree-repo.c
dbnicholson: Right.

src/libostree/ostree-repo.c
dbnicholson: It's just a helper function to not have a bunch of adhoc conversions of lock state to reporting info, but I guess since `GMutexLocker` is already in use it could be used to ensure the mutex is taken.

src/libostree/ostree-repo.c
dbnicholson: I don't think it is a concern. I was thinking about what would happen if you tried to finalize while another thread is still actively using the same object, but I think you'd have bigger problems if that was the case. I think I'll just drop this lock/unlock and assume that anyone using threads will clean them up before finalizing or deal with the consequences.

Makefile.am
```diff
@@ -31,7 +31,7 @@ AM_CPPFLAGS += -DDATADIR='"$(datadir)"' -DLIBEXECDIR='"$(libexecdir)"' \
 	-DOSTREE_COMPILATION \
 	-DG_LOG_DOMAIN=\"OSTree\" \
   -DOSTREE_GITREV='"$(OSTREE_GITREV)"' \
-	-DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_40 '-DGLIB_VERSION_MAX_ALLOWED=G_ENCODE_VERSION(2,50)' \
+	-DGLIB_VERSION_MIN_REQUIRED=GLIB_VERSION_2_44 '-DGLIB_VERSION_MAX_ALLOWED=G_ENCODE_VERSION(2,50)' \
```
cgwalters, line 34: Potentially worth refactoring this as a prep commit.  It looks like Debian stable has 2.58 and oldstable has 2.50, so we can almost certainly go to at least 2.50.

And I bet there's other cruft we could drop then, like our backported autocleanup macros.

src/libostree/ostree-repo.c
cgwalters: Yeah, we could add `g_assert_cmpint (self->lock.exclusive, <, G_MAXINT);` but it seems unlikely.
Ah, TIL GCC [has builtins for this](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html#Integer-Overflow-Builtins) similar to what Rust exposes via e.g. [https://doc.rust-lang.org/std/primitive.i32.html#method.overflowing_add](overflowing_add).

src/libostree/ostree-repo.c
```diff
@@ -405,26 +412,40 @@ pop_repo_lock (OstreeRepo  *self,
...
   g_debug ("Pop lock: state=%s, depth=%u", info.name, info.len);
+  int state_to_drop = GPOINTER_TO_INT (stack->pdata[stack->len - 1]);
   if (info.len > 1)
     {
-      int next_state = GPOINTER_TO_INT (g_queue_peek_nth (&lock->stack, 1));
+      int next_state;
+
+      if (state_to_drop == LOCK_EX)
+        {
+          g_return_val_if_fail (self->lock.exclusive > 0, FALSE);
```
cgwalters, line 433: In ostree code I tend to prefer assertions over `g_return_val_if_fail` because we're maintaining the root filesystem and crashing is better than corruption.  
i.e.

```suggestion
          g_assert_cmpint (self->lock.exclusive, >, 0);
```

(Personally I'd say even for desktop apps this is better but that battle feels kind of lost)

src/libostree/ostree-repo.c
```diff
@@ -435,16 +456,20 @@ pop_repo_lock (OstreeRepo  *self,
...
+      if (!do_repo_unlock (self->lock.fd, flags))
         return glnx_throw_errno_prefix (error, "Unlocking repo failed");
     }
 
-  g_queue_pop_head (&lock->stack);
+  if (state_to_drop == LOCK_EX)
+    self->lock.exclusive--;
+  else
+    self->lock.shared--;
+  g_ptr_array_remove_index (stack, stack->len - 1);
```
cgwalters, line 467: Hmm wait, now that we have the two counters, do we actually need to store the stack?  Since the valid state transitions can only be:

shared (++) -> exclusive (++)

i.e. we can never have any shared *after* exclusive, so why maintain the stack at all versus just the two counters?

cgwalters (approved):

Makefile.am
dbnicholson: Sure, do you want to go for 2.50?

src/libostree/ostree-repo.c
dbnicholson: I'm fine with that. I had been using `g_return*` rather than `g_assert*` under the idea that there's less chance they would be disabled, but I guess it's possible either `G_DISABLE_ASSERT` or `G_DISABLE_CHECKS` could be defined.

src/libostree/ostree-repo.c
dbnicholson: I briefly came to the same conclusion, but I think it's necessary to know what state to drop when there are multiple threads. In fact, what's here has a bug and I'm going to beef up the multithreaded test to check this.

Here's the scenario if I can get this ascii timeline to work out:
```
Thread 1	Thread 2	Flock state
========	========	===========
push exclusive			exclusive
		push shared	exclusive
pop				shared
```
With the current implementation, the flock would remain exclusive after the pop because the counters would say `exclusive == 2` and `shared == 0` before. In order to correctly return from exclusive to shared when thread 1 pops, you need to correctly track how many exclusive and shared locks are held instead of just incrementing `exclusive` every time once the lock is exclusive. Then it would be `exclusive == 1` and `shared == 1`, but you still wouldn't know which state to drop unless it was stored with the thread that is doing the dropping.

dbnicholson: I just pushed 3f78672 which tests this scenario and will now fail.

dbnicholson: That failed as expected:
```
[2021-04-27T21:19:26.049Z] ERROR: tests/test-repo - Bail out! OSTree:ERROR:tests/test-repo.c:459:test_repo_lock_multi_thread: assertion failed (error == NULL): Locking repo shared failed: Resource temporarily unavailable (g-io-error-quark, 27)
```

dbnicholson: And now with c7de18a `tests/test-repo` passes. Unless you still think that's incorrect, I'm going to clean up the series.

src/libostree/ostree-repo.c
dbnicholson: glib even has wrappers for the GCC builtins in https://developer.gnome.org/glib/stable/glib-Bounds-checked-integer-arithmetic.html. These were added in 2.48. However, after looking at adding the assertions, I'm pretty sure I'd do the check and increment separately. The check would be before changing the file lock state, and the increment would be after the file lock call succeeded.

Makefile.am
cgwalters: SGTM!

dbnicholson: Hmm, bumping to 2.50 results in a pile of deprecation warnings for `GSimpleAsyncResult` usage, which was deprecated in 2.46 in favor of `GTask`. See https://developer.gnome.org/gio/stable/GSimpleAsyncResult.html. I'm going to leave this at 2.44 as I don't really want to go refactor all of that as part of this work.

src/libostree/ostree-repo.c
cgwalters: > In order to correctly return from exclusive to shared when thread 1 pops, you need to correctly track how many exclusive and shared locks are held instead of just incrementing exclusive every time once the lock is exclusive. Then it would be exclusive == 1 and shared == 1, but you still wouldn't know which state to drop unless it was stored with the thread that is doing the dropping.

I need to think about this a bit more.  So what we're trying to support is a case where higher level code grabs an exclusive lock, but then some portion of that operation wants a shared lock, right?  And the idea here is that everything that has access to the same `OstreeRepo` instance is part of the same process.  

That part makes sense to me.

But your state diagram has *threads* - it should be separate processes and `OstreeRepo*` instances right?

dbnicholson: It would be great to only think about processes and just have a mutex that guards the lock state. Unfortunately, once you're running threads the order that the lock is taken and released is no longer linear. 2 worker threads aren't going to coordinate to release their locks in the same order they were acquired. And since the lock has multiple states, you need to know what each particular thread is trying to release. If it was just a single lock/unlock state like a mutex, then you could just have a single counter and when it gets to 0 you unlock.

I honestly wouldn't have given multithreaded code using the same `OstreeRepo` much thought except that the pull code already spawns a bunch of threads. So even in ostree you could trigger this behavior just by taking an exclusive lock somewhere and then running some pull operation _if_ a lock was taken by any of the pull threads.

What actually tipped me off to this was that I wanted to make `ostree_repo_prune_object` take an exclusive lock. In the vast majority of cases that works fine - `ostree_repo_prune` takes an exclusive lock and the lock increments each time `ostree_repo_prune_object` is called. Except that when you try to write a commit object it also tries to prune a potentially existing old tombstone commit with `ostree_repo_prune_object`. That immediately caused pull to deadlock since the pull thread's lock (with what's in ostree right now) is disconnected from the main thread. OK, let's make them coordinate by keeping the lock in the shared `OstreeRepo`. Oh wait, that's not so simple to track correctly.

An alternative I just thought of that would require less state tracking in ostree would be to have the caller specify the state to drop in the `pop()` call. Then you could have 2 counters in the `OstreeRepo` and just decrement what the caller said without the per-thread stack keeping track of what the thread's last push call was. I don't really know what that API would look like.

cgwalters: But all that locking is internal to ostree today; we can just add an `_ostree_repo_prune_object_unlocked()` that assumes the caller has locked, and use that in the pull code right?

dbnicholson: I don't know what happened to the comment I made the other day.

Yes, in that case you could add some unlocked variant that's only available internally. That case actually violates the semantics since the transaction only has a shared lock and the expectation is that you have an exclusive lock when deleting. Regardless, I think in general once you allow the locking to be made externally you will sooner or later run into a case where someone is taking a lock externally in a way that the ostree internals haven't arranged for. To me the only real fix is to make the locking completely safe to use regardless of how the application has decided to use processes and threads.

In reviewing why the locking implementation is the way it is, it occurred to me that all the complexity is due to not taking the lock type in the `ostree_repo_lock_pop` call. The reason it doesn't is to make the API friendly - you push the lock type you want an pop will just take it off when you say to. However, doing that requires the semi-complex stack of lock states so that you know which one to remove.

However, if you drop that requirement and make the caller specify the lock type to pop, then the tracking becomes very simple again - 2 counters and as long as the number of exclusive locks is greater than 0 than the lock is exclusive. I think that makes the whole thing much simpler and implemented it in https://github.com/dbnicholson/ostree/tree/mt-locking-alt. Let me know if that looks better to you and I can move it here.

src/libostree/ostree-repo.c
```diff
@@ -374,31 +386,47 @@ push_repo_lock (OstreeRepo          *self,
...
+  if (info.state == LOCK_EX || info.state == next_state)
+    {
+      g_debug ("Repo already locked %s, extending stack", info.name);
+    }
+  else
+    {
+      const char *next_state_name = lock_state_name (next_state);
 
       g_debug ("Locking repo %s", next_state_name);
       if (!do_repo_lock (lock->fd, flags))
```
pwithnall: Maybe worth an assertion somewhere round here that checks a lock is never ‘upgraded’ from exclusive to shared.

src/libostree/ostree-repo.c
```diff
@@ -542,6 +593,7 @@ ostree_repo_lock_push (OstreeRepo          *self,
 /**
  * ostree_repo_lock_pop:
  * @self: a #OstreeRepo
+ * @lock_type: the type of lock to release
```
pwithnall, line 558: Might be worth pointing out that it’s a programmer error to release the wrong type of lock, compared to what your program acquired in the first place.

src/libostree/ostree-repo.c
```diff
@@ -660,7 +718,11 @@ ostree_repo_auto_lock_push (OstreeRepo          *self,
 {
   if (!ostree_repo_lock_push (self, lock_type, cancellable, error))
     return NULL;
-  return (OstreeRepoAutoLock *)self;
+
+  OstreeRepoAutoLock *auto_lock = g_new0 (OstreeRepoAutoLock, 1);
```
pwithnall, line 684: Could the need to heap-allocate be avoided by making the `OstreeRepoAutoLock` struct public (but opaque), and having callers stack-allocate it?

Or, if that’s too intrusive, you could use `g_alloca()` here and require that `OstreeRepoAutoLock` is always used with `g_autoptr()` in a single scope.

src/libostree/ostree-repo.c
```diff
@@ -172,52 +172,43 @@ G_DEFINE_TYPE (OstreeRepo, ostree_repo, G_TYPE_OBJECT)
...
  * The intended usage is to take a shared lock when writing objects or
  * reading objects in critical sections. Exclusive locks are taken when
  * deleting objects.
  *
- * To allow fine grained locking within libostree, the lock is
- * maintained as a stack. The core APIs then push or pop from the stack.
- * When pushing or popping a lock state identical to the existing or
- * next state, the stack is simply updated. Only when upgrading or
+ * To allow fine grained locking within libostree, the lock is treated
+ * as a stack. The core APIs then push or pop from the stack. When
```
pwithnall, line 204: Stack no longer exists.

src/libostree/ostree-repo-private.h
```diff
@@ -104,6 +104,13 @@ typedef struct {
   fsblkcnt_t max_blocks;
 } OstreeRepoTxn;
 
+typedef struct {
+  GMutex mutex;
```
pwithnall, line 108: Might be worth adding a comment to say that every other member of `OstreeRepoLock` should be accessed only with `mutex` held

pwithnall (commented):
LGTM. A few minor comments. I haven’t looked at the test additions in `test-repo.c`.

src/libostree/ostree-repo.c
```diff
@@ -214,7 +214,8 @@ static GPrivate repo_lock_table = G_PRIVATE_INIT (free_repo_lock_table);
 
 typedef struct {
   int fd;
-  GQueue stack;
+  guint shared; /* Number of shared locks */
```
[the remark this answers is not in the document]
dbnicholson: Good point.

src/libostree/ostree-repo.c
```diff
@@ -374,31 +386,47 @@ push_repo_lock (OstreeRepo          *self,
...
+      next_state = LOCK_SH;
+      counter = &(lock->shared);
+    }
+
+  /* Check for overflow */
+  g_assert_cmpuint (*counter, <, G_MAXUINT);
+
+  if (info.state == LOCK_EX || info.state == next_state)
+    {
+      g_debug ("Repo already locked %s, extending stack", info.name);
```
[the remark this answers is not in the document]
dbnicholson: Yeah, I need to go back and clean some of that up.

src/libostree/ostree-repo.c
```diff
@@ -374,31 +386,47 @@ push_repo_lock (OstreeRepo          *self,
...
   g_debug ("Push lock: state=%s, depth=%u", info.name, info.len);
 
-  if (info.state == LOCK_EX)
+  int next_state;
+  guint *counter;
+  if (lock_type == OSTREE_REPO_LOCK_EXCLUSIVE)
     {
-      g_debug ("Repo already locked exclusively, extending stack");
-      g_queue_push_head (&lock->stack, GINT_TO_POINTER (LOCK_EX));
+      next_state = LOCK_EX;
```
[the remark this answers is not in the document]
dbnicholson: They're actually not identical because `flags` can contain `LOCK_NB`. I suppose I could use separate constants rather than reusing the flock operation flags. I kinda wanted to add `OSTREE_REPO_LOCK_UNLOCKED` to `OstreeRepoLockType` and use those constants for the state tracking, but then I don't think that's really appropriate in the API since it's not something you could pass to the any of the locking functions.

src/libostree/ostree-repo.c
dbnicholson: :+1:

src/libostree/ostree-repo.c
dbnicholson: :+1:

src/libostree/ostree-repo.c
dbnicholson: I did want to stack allocate but couldn't figure out a way that would be as nice to use as `g_autoptr()`. Currently you can do:
```
g_autoptr(OstreeRepoAutoLock) lock = ostree_repo_auto_lock_push(repo, ...);
if (!lock)
  return FALSE;
```
With the stack allocation, I think it would be:
```
g_auto(OstreeRepoAutoLock) lock = { 0, };
if (!ostree_repo_auto_lock_push(repo, &lock, ...))
  return FALSE;
```
I guess that's not so bad. You could probably provide an initializer macro, too. I could go either way there.

Although I can't think of a reason why you'd want to use it outside of a single scope, requiring it seems a bit intrusive.

src/libostree/ostree-repo.c
dbnicholson: Right, all this documentation needs to be cleaned up.

src/libostree/ostree-repo.c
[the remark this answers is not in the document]
pwithnall: `flags` only contains `LOCK_NB` in `pop_repo_lock()`. This comment is about `push_repo_lock()`.

src/libostree/ostree-repo.c
pwithnall: > I guess that's not so bad. You could probably provide an initializer macro, too. I could go either way there.

Yeah, or you could change `ostree_repo_auto_lock_push()` to return an `OstreeRepoAutoLock` structure by value. Should be fine, since it’s only two pointers.

dbnicholson: That's how I was thinking about it initially, but then how do you detect errors? Particularly if the structure is opaque. Unlike `autoptr` where you can use `NULL`, here you'd have to check if one of the members was in an error state since you'd actually have to return a structure. Even if it's not opaque it would be kind of ugly to use it that way.

pwithnall: You could check the `error` out argument. But true, it is quite idiosyncratic. :shrug:

src/libostree/ostree-repo.c
dbnicholson: I stopped using the stack, but the same principle applied to the counters, so I added a check for that.

src/libostree/ostree-repo.c
dbnicholson: I changed all these checks to use `g_assert*` or `g_error`.

src/libostree/ostree-repo.c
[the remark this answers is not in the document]
dbnicholson: It does (it has to be passed through to the actual open fd/flock call):
```
  int flags = (lock_type == OSTREE_REPO_LOCK_EXCLUSIVE) ? LOCK_EX : LOCK_SH;
  if (!blocking)
    flags |= LOCK_NB;
```
But you're right that the assignments could be more localized so they're relationship is clearer.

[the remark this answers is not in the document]
dbnicholson: I put the assignment to `next_state` in this block so it's clearer how it's derived and how it's different from `flags`.

src/libostree/ostree-repo.c
dbnicholson: I put this in the description text.

dbnicholson:
I pushed some fixups to address @pwithnall's review. I left the API `ostree_repo_auto_lock_push` the way it is allocating an `OstreeRepoAutoLock` structure (but switched to `g_slice_*`). I'm preparing a patch to switch to stack allocated if that's preferred.

src/libostree/ostree-repo.c
dbnicholson: I implemented a stack allocated version in https://github.com/dbnicholson/ostree/commit/bd0fb98597ec29a839f67cb75416107368b9cb48. I'm pull that in here if that's the preferred API.

src/libostree/ostree-repo.c
cgwalters: > Yes, in that case you could add some unlocked variant that's only available internally. That case actually violates the semantics since the transaction only has a shared lock and the expectation is that you have an exclusive lock when deleting.

Hmm.  The tombstone object is very special though - clients just test whether or not it exists.  Here a separate client with a shared lock querying the commit (that doesn't exist ) running simultaneously with another client writing that commit is racy, but in a safe way.  We're not talking about *losing* data here, conceptually it's deleting a file as part of the act of *creating* data.

My mental model here is still thinking of shared = read, exclusive = write.

dbnicholson: > My mental model here is still thinking of shared = read, exclusive = write.

It took me a while to not think of them that way, but I've also touched this code by far the most. Maybe we should rename the lock names before making them public? I think I only named them that way initially since those were the flock names. The semantics it's really trying to encode are to not delete things that others might be using. I think you could name those `ACCESS` and `DELETE`. Probably there are better names, though.

dbnicholson:
I've already said a bunch here and probably need to get out of the way to let you chew on this, but just one last thought.

I think if the API is changed so that pop requires the type (the 2nd commit), then we can decide later to change the semantics from "all threads lock independently" to "all threads sharing an OstreeRepo lock cooperatively" without breaking the API. Basically, I'm pushing for the latter because I feel like you'd expect the locking to cooperate wherever you share the OstreeRepo structure, but I can see cases where you want the threads to act independently even if they're accessing the same structure.

src/libostree/ostree-repo.c
cgwalters: There's lots of terms of art in https://en.wikipedia.org/wiki/Isolation_(database_systems)

I think shared/exclusive is better than read/write - it's just easier to think of read/write in many cases.  That said for example in Rust there's a sentiment that `&/&mut` are misnamed and it should have been `&/&excl` or so (short for "exclusive"), because `&` in some cases still allows [interior mutability](https://doc.rust-lang.org/reference/interior-mutability.html)).

dbnicholson:
@cgwalters any thoughts?

src/libostree/ostree-repo.c
```diff
@@ -678,7 +678,7 @@ ostree_repo_auto_lock_push (OstreeRepo          *self,
   if (!ostree_repo_lock_push (self, lock_type, cancellable, error))
     return NULL;
 
-  OstreeRepoAutoLock *auto_lock = g_new0 (OstreeRepoAutoLock, 1);
+  OstreeRepoAutoLock *auto_lock = g_slice_new (OstreeRepoAutoLock);
```
cgwalters, line 681: Out of curiosity, does this still matter today?  I thought modern malloc had subsumed the optimizations here.

cgwalters (approved):
Let's commit to getting this in for the next release.  We can do any further fixes as followups.

cgwalters:
Can you rebase on latest `main` to pull in the CI fixes?

src/libostree/ostree-repo.c
cgwalters: So we got into this discussion around the complexity of the state tracking, which was simplified by having the caller track which lock they acquired.  That's better, but we were also having a discussion around the *semantics* of locking that I'm not sure is resolved.  You said you wanted to change `ostree_repo_delete_object()` to acquire an exclusive lock, which makes some sense.

I do still think the right fix is either:

 - For removing *tombstone* objects to not require an exclusive lock
 - Change the pull code to call an `_unlocked()` variant

Or honestly, it's going to be an ABI hazard to change the locking semantics of any functions at all, and that may apply in this case.

Actually `_delete_object()` is also inherently very special and low level - you can easily just remove a file object referenced by a commit, etc.  So we could instead add `ostree_repo_delete_object_locked()` or so...but under what scenarios should that be used?

Were you actually planning to use this for something?

src/libostree/ostree-repo.c
dbnicholson: I honestly have no idea. It seemed like a good idea, but maybe this is based on 2012 knowledge. I also wrote a stack allocated version in https://github.com/dbnicholson/ostree/commit/bd0fb98597ec29a839f67cb75416107368b9cb48 as discussed in one of the comments above.

src/libostree/ostree-repo.c
dbnicholson: The semantics and hazards is exactly why I started working on this again. In the case of `_delete_object()`, there are several ways to make it work. In addition to the ways you mentioned above, I also have a local branch that adds a `_delete_object_if_exists()` which simply does a `_have_object()` before calling into the real `_delete_object()` to avoid taking the lock unless you really need to. It's still racy, but for the case we're describing here it's not really different. Also just documenting that `_delete_object()` is low level and you take an exclusive lock before using it would be fine, too.

Anyways, I wasn't trying to focus on this specific case. I don't have plans to use `_delete_object()` directly. I was trying to use it as a concrete example of the general issue I'm worried about. The semantics of the repo locking without this PR are that threads operate on separate lock file descriptors. If you have multiple threads operating on the same `OstreeRepo`, then sooner or later you're likely to have a deadlocked application when someone adds an innocuous lock call in ostree's internals.

I find this to not be a theoretical issue since ostree itself becomes multithreaded in pull, one of the most common operations someone would do on an ostree repo. So, I believe this hazard is not just a corner case of someone trying to run a multithreaded ostree program, but actually pretty likely to occur. All that has to happen is someone adds a locking call somewhere in an API that pull uses. My focus on `_delete_object()` was that it's a realistic (or even desired) candidate.

With this PR the hazard reduces a lot and it actually gives the application (or even libostree itself) options. Threads will share the lock file descriptor and coordinate on locking. This is the same as if you wrote a multithreaded `flock()` program and opened the file descriptor before spawning threads. If you want the threads to act independently (e.g. a pull in one thread and a prune in another) you just open separate `OstreeRepo`s in each thread. I think this is a lot better and probably just needs some clearer documentation of the semantics.

dbnicholson:
> Can you rebase on latest `main` to pull in the CI fixes?

With or without squashing?

dbnicholson:
> > Can you rebase on latest `main` to pull in the CI fixes?
> 
> With or without squashing?

I went without squashing for now.

cgwalters:
We aren't using a merge bot that understands `fixup!` right now unfortunately, so please squash those.

dbnicholson:
> We aren't using a merge bot that understands `fixup!` right now unfortunately, so please squash those.

Done. Some of the fixups were conflicting, so I don't think a merge bot would have been able to resolve it anyways. I usually try to wait on squashing because some people want to only see the differences from what they've reviewed before.

src/libostree/ostree-repo.c
pwithnall: > I thought modern malloc had subsumed the optimizations here.

That is correct. The performance gap between the two is a lot narrower now, and gslice is basically unmaintained. We are (still) [looking at dropping it](https://gitlab.gnome.org/GNOME/glib/-/issues/1079). It’s fine to keep using it, but it might secretly become a call to `g_malloc()` in future.

dbnicholson: Good to know! I'll file a followup to change back `g_new`.

pwithnall: Not really worth the churn, imo. They both work fine; I prefer `g_malloc()`+friends for new code, but am not going around changing old code from one to the other.

cgwalters: Any way you slice it, OK by me.

PR #2350, opened 2021-04-18
libtest: On failure, make it clearer what has happened

smcv:
If we fail as a result of `set -x`, It's often not completely obvious
which command failed or how. Use a trap on ERR to show the command that
failed, and its exit status.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -46,6 +46,14 @@ run_exit_cmds() {
 }
 trap run_exit_cmds EXIT
 
+report_err () {
+  local exit_status="$?"
+  { { local BASH_XTRACEFD=3; } 2> /dev/null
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Nice, that's some veritable bash sorcery!  

Tried this out with

```
diff --git a/tests/basic-test.sh b/tests/basic-test.sh
index 75333f4d..c80829aa 100644
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -42,6 +42,8 @@ else
     fi
 fi
 
+false is true you know
+
 # This should be dynamic now
 assert_not_has_dir repo/uncompressed-objects-cache
 ```

and yep, I see in `test-suite.log`:

```
Unexpected nonzero exit status 1 while running: false is true you know
```

Issue #2351, opened 2021-04-18
test-remote-gpg-import.sh intermittently failing: Can't check signature: public key not found


smcv:
Seen intermittently on Debian 11 autobuilders with 2021.2 and 2021.1. #2313 mentions a similar-looking failure on Arch Linux.

<details>
<summary>A log, generated on Debian armhf (ARMv7 EABI hardfloat) using 2021.2 patched with #2350</summary>

```
ERROR: tests/test-remote-gpg-import.sh
======================================

# testlabel.txt: security.selinux: No such attribute
++ BUILT_WITH_ASAN=
++ test -n ''
++ test -n ''
++ test -z ''
++ test -z ''
++ CMD_PREFIX='env LD_PRELOAD=/<<PKGBUILDDIR>>/tests/libreaddir-rand.so'
++ test -n /<<PKGBUILDDIR>>
++ OSTREE_HTTPD=/<<PKGBUILDDIR>>/ostree-trivial-httpd
++ _have_user_xattrs=
++ _have_systemd_and_libmount=
++ _have_strace_fault_injection=
++ libtest_exit_cmds+=(libtest_cleanup_gpg)
++ ED25519PUBLIC=
++ ED25519SEED=
++ ED25519SECRET=
+ unset OSTREE_GPG_HOME
+ setup_fake_remote_repo1 archive
+ mode=archive
+ shift
+ commit_opts=
+ '[' 0 -eq 0 ']'
++ pwd
+ oldpwd=/var/tmp/tap-test.5Ur8lk
+ mkdir ostree-srv
+ cd ostree-srv
+ mkdir gnomerepo
+ ostree_repo_init gnomerepo --mode=archive
+ repo=gnomerepo
+ shift
+ env LD_PRELOAD=/<<PKGBUILDDIR>>/tests/libreaddir-rand.so ostree --repo=gnomerepo init --mode=archive
+ test -n 1
+ echo -e 'disable-xattrs=true\n'
+ mkdir gnomerepo-files
+ cd gnomerepo-files
+ echo first
+ mkdir baz
+ echo moo
+ echo alien
+ env LD_PRELOAD=/<<PKGBUILDDIR>>/tests/libreaddir-rand.so ostree --repo=/var/tmp/tap-test.5Ur8lk/ostree-srv/gnomerepo commit --add-metadata-string version=3.0 -b main -s 'A remote commit' -m 'Some Commit body'
+ mkdir baz/deeper
[...]

PR #2352, opened 2021-04-19
tests: Test without a cache directory by default

dbnicholson:
Several tests generate summaries and then expect to use the generated
summary immediately. However, this can cause intermittent test failures
when they inadvertantly get a cached summary file. This typically
happens when the test is run on a filesystem that doesn't support user
extended attributes. In that case, the caching code can only use the
last modified time, which only has 1 second granularity. If tests don't
carefully manage the summary modification times or the repo cache then
they are likely subject to races in some test environments.

This introduces an environment variable `OSTREE_SKIP_CACHE` that
prevents the repo from using a cache directory. This is enabled by
default in tests and disabled for tests that are a explicitly trying to
test the caching behavior.

Fixes: #2313
Fixes: #2351

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3291,7 +3291,7 @@ ostree_repo_open (OstreeRepo    *self,
   if (!glnx_opendirat (self->repo_dir_fd, "tmp", TRUE, &self->tmp_dir_fd, error))
     return FALSE;
 
-  if (self->writable)
+  if (self->writable && getenv ("OSTREE_SKIP_CACHE") == NULL)
     {
       if (!glnx_shutil_mkdir_p_at (self->tmp_dir_fd, _OSTREE_CACHE_DIR, DEFAULT_DIRECTORY_MODE, cancellable, error))
         return FALSE;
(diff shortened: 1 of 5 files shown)


dbnicholson:
This is a pretty large hammer and any suggestions on the variable name are welcome.

cgwalters:
Wouldn't it be simpler to add a `sleep 1` after the places where we regenerate the summary?

dbnicholson:
That's what I suggested in another one off these issues (and have done myself), but there are 2 drawbacks.

1. It slows down the tests artificially.
2. It requires the person writing the test to understand when the caching is implicitly going to have an effect on the test.

In the linked issues the test is just running `pull` and it's not clear why the summary caching would have an effect unless you're familiar with this part of the code. And the test worked fine until the summary caching changes were made somewhat recently.

I'm fine if this isn't the way to do it. My other suggestion was `rm -rf repo/tmp/cache/summaries` in the few needed places.

cgwalters (approved):
> That's what I suggested in another one off these issues 

Right sorry, as I get older I am getting more forgetful =/

OK yeah, this is fine.  Thanks for this patch!

dbnicholson:
Heh, I was about to change course and sprinkle in the necessary handing in the tests.

dbnicholson:
> Right sorry, as I get older I am getting more forgetful =/

No problem, I didn't mean to imply that you should be able to remember every comment here! More like, "I did think of that previously and here's why I think this is better".

PR #2353, opened 2021-04-27
pull: Cleanly error out on unknown schemes

cgwalters:
Previous to this we'd trip an assertion `abort()` deep in the curl code if e.g.
a user did `ostree remote add foo htttp://...` etc.

Motivated by considering supporting "external remotes" where code outside
ostree does a pull, but we want to reuse the signing verification infrastructure.

### DIFF
--- a/src/libostree/ostree-fetcher-uri.c
+++ b/src/libostree/ostree-fetcher-uri.c
@@ -117,3 +117,21 @@ _ostree_fetcher_uri_to_string (OstreeFetcherURI *uri)
 {
   return soup_uri_to_string ((SoupURI*)uri, FALSE);
 }
+
+
+/* Only accept http, https, and file; particularly curl has a ton of other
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-fetcher-util.h
```diff
@@ -33,6 +33,9 @@ G_BEGIN_DECLS
  */
 #define OSTREE_FETCHER_USERAGENT_STRING (PACKAGE_NAME "/" PACKAGE_VERSION)
 
+gboolean
+_ostree_fetcher_validate_mirrorlist (GPtrArray *mirrorlist, GError **error);
```
jlebon, line 37: Missing a `git add` for this bit?

cgwalters: Ah that's cruft, I realized it wasn't needed since we do (phew) check the returned mirrorlist URIs.

jlebon (approved):

PR #2355, opened 2021-04-30
docs: Add more details about 3-way merge

jlebon:
This came up a few times so let's go into more details in the docs.


dbnicholson (approved):
This is great! I've also had to explain this many times and am glad to see the details laid out. The "propagated forever" part is such a crucial detail that is not apparent when you start dipping your toes into the ostree waters.

Issue #2362, opened 2021-05-14
Publish API docs


dbnicholson:
Forgive me if they're already published or this has already been discussed, but it would be really nice if the API docs were published somewhere. The version at https://developer.gnome.org/ostree/stable/ is from 2014.4. Yikes.

It would appear that ostreedev.github.io/api would be usable alongside the regular ostree docs. I probably don't have the privileges to get this done, but I'd be happy to work on it.

jlebon:
Agreed, that'd be great! Hmm, we could have a GitHub Action which builds the docs and pushes them to some git repo. It'd be pretty messy to push it to the main branch of this repo itself though, so it might have to be a different website from https://ostreedev.github.io/ostree/ ?

Maybe simplest to just have a separate `ostreedev/ostree-api-doc` repo with its own GitHub Pages which we link to from https://ostreedev.github.io/ostree/.

dbnicholson:
How does the GitHub pages integration work right now? I haven't used it a ton, but at Endless I recall there was a separate gh-pages branch and our CI would build the docs and push them to that branch. I don't see any of that here and presumably jekyll builds the stuff in `docs` before it's published. Oh, I think I see from https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll that github pages automatically generates them.

In line with using a separate branch, another option (not sure how complex this would be) would be to have a simple top level index with links to the current docs and the API docs as separate endpoints like https://ostreedev.github.io/ostree/manual and https://ostreedev.github.io/ostree/reference. I don't quite know if `manual` is the best way to describe the existing docs. In that scheme an ambitious person could even generate the man pages and add them as a separate endpoint. The downside there is that I think you'd have to do the jekyll build in the CI instead of letting github pages do it automatically.

I think a separate repo would also work, but I'm guessing that would require a bit more effort to pass in a github token to allow pushing to a separate repository.

dbnicholson:
#2363 does this but pushes to the `gh-pages` branch in this repo as seems to be the convention. It's not super pretty but it seems to work.

dbnicholson:
Unfortunately, the [build](https://github.com/ostreedev/ostree/runs/2668612246) failed:

```
remote: Permission to ostreedev/ostree.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/ostreedev/ostree.git/': The requested URL returned error: 403
```

[...]

PR #2363, opened 2021-05-21
Publish libostree API docs

dbnicholson:
This adds a GitHub workflow to build the Jekyll docs with Jekyll Actions and push it to the `gh-pages` rather than use GitHub's fully automated flow. The reason to do this is to build the gtk-doc API docs and then include them into the Jekyll static site.

I'm no jekyll or github pages expert, but it seems to work correctly as you can see it at the following pages:

https://dbnicholson.github.io/ostree/
https://dbnicholson.github.io/ostree/#api-reference
https://dbnicholson.github.io/ostree/reference/

In order to put this in place the ostree repo Pages settings would need to be changed to use `gh-pages` as the branch with the root directory.

Fixes: #2362

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,62 @@
+---
+name: Docs
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  docs:
+    name: Build documentation
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      # This is taken from ci/travis-install.sh but should probably be
+      # refactored.
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y \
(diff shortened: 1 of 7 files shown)


cgwalters (approved):
Thank you so much for doing this!

cgwalters:
Ah and fun, looks like our CI testing fell over with newer systemd, at least one output string changed to `Deactivated successfully.`.

cgwalters:
Ahh and the `main` rename broke Prow, ok will look at that too.

AdrianVovk:
I'm not sure what exactly is wrong but I cannot access the API docs. I just get error 404. Happens on multiple devices, one of which just visited that page for the first time so there's no possibility of a caching issue on my end

dbnicholson:
> I'm not sure what exactly is wrong but I cannot access the API docs. I just get error 404.

Someone needs to turn it on :old_key: https://github.com/ostreedev/ostree/issues/2362#issuecomment-855250964

PR #2364, opened 2021-05-25
ci: Fix staged-delay to work with newer systemd

cgwalters:
Yeah, we should stop parsing the text; I need to dig at that
at some point.


cgwalters:
Briefly looked at using structured logging, but it's only slightly less gross.  systemd doesn't log any metadata about the unit with these messages.  (I think it would make sense though to have at least some way to track historical unit state more precisely)

dbnicholson (approved):

cgwalters:
ratelimits, still need to move off Travis

cgwalters:
Needs fixing separately
/override ci/prow/sanity

PR #2366, opened 2021-05-26
docs: Change IRC channel to libera.chat from freenode

pwithnall:
Freenode appears to be imploding.

Signed-off-by: Philip Withnall <pwithnall@endlessos.org>

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ make install DESTDIR=/path/to/dest
 ## Contact and discussion forums
 
 OSTree has a [mailing list](https://mail.gnome.org/archives/ostree-list/) and
-there is also an `#ostree` channel on FreeNode.  However, asynchronous+logged
+there is also an `#ostree` channel on [Libera.Chat](ircs://irc.libera.chat/ostree).  However, asynchronous+logged
 communication is preferred for nontrivial questions.
 
 ## Contributing


pwithnall:
Marking this as a draft as it needs a bit of discussion (and for the IRC channel to actually be moved).

cgwalters (approved):

README.md
```diff
@@ -147,7 +147,7 @@ make install DESTDIR=/path/to/dest
 ## Contact and discussion forums
 
 OSTree has a [mailing list](https://mail.gnome.org/archives/ostree-list/) and
-there is also an `#ostree` channel on FreeNode.  However, asynchronous+logged
+there is also an `#ostree` channel on irc.libera.chat.  However, asynchronous+logged
```
jlebon, line 150: Minor: I think the stylized name is "Libera.Chat".

jlebon (approved):
Let's go ahead with this? Fedora has also already moved over.

alexlarsson:
For the record, flatpak (and related projects like freedesktop-sdk) decided that this was a good time to just drop irc. We're now 100% on matrix, without any bridging to irc (which just limit matrix).

 We still have a (owned by us) channel on libera.chat, but only to point people to the right place.

README.md
pwithnall: Pushed an updated version which turns this into an actual link (I hope I’ve got the link syntax right)

cgwalters:
I need to get on Matrix I guess...but anyways, right now we can just merge this and if we want to make it a redirect to Matrix later then that's doable.

ramcq:
> I need to get on Matrix I guess...but anyways, right now we can just merge this and if we want to make it a redirect to Matrix later then that's doable.

If by redirect, you mean gateway, that experience is strictly worse for people on both sides of the gateway, compared to having a native Matrix chat. Given my life involves being connected to both IRC and Matrix for the time being, I don't mind either way but "both" is a worse answer all round IMO. What would be better is choosing, and merging the PR and articulating where the party is at now. :)

travier:
With the rapid Freenode demise, I'd vote in favor of merging this one to reflect the current status and moving the discussion about how to do Matrix with or without bridging to an issue.

jlebon:
Docs only.
/override ci/prow/sanity

jlebon:
OK, we can keep the Matrix chat going in https://github.com/ostreedev/ostree/issues/2378.

Issue #2359, opened 2021-05-12
Regression in tests/test-gpg-signed-commit.sh with GnuPG 2.3


cbiedl:
Greetings,

rebuilding ostree using the recently released GnuPG 2.3 version caused failures in tests/test-gpg-signed-commit.sh:

    ERROR: tests/test-gpg-signed-commit.sh - too few tests run (expected 9, got 4)
    ERROR: tests/test-gpg-signed-commit.sh - exited with status 1

It seems - I might be wrong - GnuPG changed the output slightly, and ostree can not handle that. Could you please check?

This first showed up in the Debian autopkgtest of ostree (URL might not be stable): https://ci.debian.net/data/autopkgtest/unstable/i386/o/ostree/12266734/log.gz

Test was done using the versions in Debian unstable (2020.8-2) and experimental (2021.2-2), and according to git, no changes have been made to that test since then.

Failing test is tests/test-gpg-signed-commit.sh:240:

    assert_not_file_has_content test2-show 'Key expired'

The content of test2-show in my tests were

Using GnuPG 2.2.27 (pass):

    commit 8e0d719f8d3efb5e6e13cb02be7738a48730f8f27a9029dd9e7905f658467aef
    Parent:  a873c6d18c6e6d9da6bef54c5a93ccb62e14e6236a69ba5c37fed5146b784bf5
    ContentChecksum:  836bef3087db7055f34240586910fdb8c65d5d8b7f0587d6d2b9e83ba81933bd
    Date:  2021-05-12 11:27:29 +0000

        A GPG signed commit

        Signed commit body

    Found 2 signatures:

      Signature made Wed May 12 13:27:30 2021 using RSA key ID B3B25CA8394773D9
      Good signature from "Test Key 1 <>"

      Signature made Wed May 12 13:27:30 2021 using RSA key ID 276132F419CFFB80
      BAD signature from "Test Key 2 <>"
      Primary key ID BBC53D20DC671DE2
      Primary key expired Wed May 12 13:27:33 2021

Using GnuPG 2.3.1 (fail):

    commit b45b54b1a1a2ce0af29c53534c978b3ba617d99706c3e23bef42ddd31875a23e
    Parent:  0338445b23b4278306bad18f79c76d69964b58e0ddde84c650234298ddfc65b8
    ContentChecksum:  836bef3087db7055f34240586910fdb8c65d5d8b7f0587d6d2b9e83ba81933bd
    Date:  2021-05-12 11:16:39 +0000

[...]

PR #2367, opened 2021-05-28
tests/gpg: Don't assert subkey expiration when only primary expired

dbnicholson:
In gnupg 2.3.0[1], if a primary key is expired and a subkey does not
have an expiration or its expiration is older than the primary key, the
subkey's expiration will be reported as the primary's. Previously a
subkey without an expiration would not report one that regardless of the
primary key's expiration.

This caused a regression in a test setting an expiration on a primary
key. The test was checking that the subkey was not expired by asserting
that there was no `Key expired` line in the signature verification
output. With gnupg 2.3.0+, it will show as expired, causing the test to
fail.

Remove the assertion since it's not consistent across gnupg versions. In
practice we don't care whether the subkey is considered expired or not
as long as the signature verification fails when the primary key is
expired.

1. https://dev.gnupg.org/T3343

Fixes: #2359


jlebon (approved):
Thanks for the nice commit message. LGTM!

jlebon:
Hitting pull limits in Travis.

PR #2370, opened 2021-06-07
repo: Use g_new for OstreeRepoAutoLock

dbnicholson:
GSlice is effectively deprecated and has little to no advantage over
using the system allocator on Linux.

This was discussed in https://github.com/ostreedev/ostree/pull/2348#discussion_r645864806. I started swapping out all the GSlice usage tree-wide, but I wasn't totally sure if that was wanted. If you prefer, I can finish that off.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -678,7 +678,7 @@ ostree_repo_auto_lock_push (OstreeRepo          *self,
   if (!ostree_repo_lock_push (self, lock_type, cancellable, error))
     return NULL;
 
-  OstreeRepoAutoLock *auto_lock = g_slice_new (OstreeRepoAutoLock);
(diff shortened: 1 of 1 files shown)


dbnicholson:
> This was discussed in [#2348 (comment)](https://github.com/ostreedev/ostree/pull/2348#discussion_r645864806). I started swapping out all the GSlice usage tree-wide, but I wasn't totally sure if that was wanted. If you prefer, I can finish that off.

However, for full disclosure I looked at the `GSlice` usage in https://github.com/ostreedev/ostree/blob/main/src/libotutil/ot-variant-builder.c and it seemed like it was fairly intimately connected to the code in question. The other usages in libostree were straightforward allocations.

cgwalters (approved):

dbnicholson:
Travis/Dockerhub rate limiting failure... This is a simple change, so merging rather than waiting for the planets to align.

PR #2373, opened 2021-06-09
deploy: Warn if we find content in the deployment's /var

cgwalters:
This will be ignored, so let's make it very clear
people are doing something wrong.  Motivated by a bug
in a build pipeline that injected `/var/lib/rpm` into an ostree
commit which ended up crashing rpm-ostree because it was an empty db
which it wasn't expecting.

It *also* turns out rpm-ostree is incorrectly dumping content in the
deployment `/var` today, which is another bug.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2632,6 +2632,39 @@ _ostree_deployment_set_bootconfig_from_kargs (OstreeDeployment *deployment,
     }
 }
 
+// Perform some basic static analysis and emit warnings for things
+// that are likely to fail later.  This function only returns
+// a hard error if something unexpected (e.g. I/O error) occurs.
+static gboolean
+lint_deployment_fs (OstreeSysroot     *self,
+                    OstreeDeployment  *deployment,
+                    int                deployment_dfd,
+                    GCancellable      *cancellable,
+                    GError           **error)
+{
+  g_auto(GLnxDirFdIterator) dfd_iter = { 0, };
+  glnx_autofd int dest_dfd = -1;
+  gboolean exists;
+
+  if (!ot_dfd_iter_init_allow_noent (deployment_dfd, "var", &dfd_iter, &exists, error))
+    return FALSE;
+  while (exists)
+    {
+      struct dirent *dent;
+
+      if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
+        return FALSE;
+      if (dent == NULL)
+        break;
+
+      fprintf (stderr, "note: Deploying commit %s which contains content in /var/%s that will be ignored.\n",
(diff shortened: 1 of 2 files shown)


cgwalters:
ratelimits

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2632,6 +2632,40 @@ _ostree_deployment_set_bootconfig_from_kargs (OstreeDeployment *deployment,
...
+  glnx_autofd int dest_dfd = -1;
+  gboolean exists;
+
+  if (!ot_dfd_iter_init_allow_noent (deployment_dfd, "var", &dfd_iter, &exists, error))
+    return FALSE;
+  while (exists)
+    {
+
+      struct dirent *dent;
+
```
jlebon, line 2655: Looks roomy there.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2632,6 +2632,40 @@ _ostree_deployment_set_bootconfig_from_kargs (OstreeDeployment *deployment,
...
+      if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
+        return FALSE;
+      if (dent == NULL)
+        break;
+
+      ot_journal_print (LOG_WARNING, "Deploying commit %s which contains content in /var/%s that will be ignored.",
```
jlebon, line 2661: Hmm, wouldn't `g_warning` be a better fit here? E.g. in cosa for example this will run in supermin and we don't collect journal logs there. Or in Anaconda, it'd be harder to miss if it shows up inline with the rest of the Anaconda output/logs.

jlebon (approved):
LGTM overall. Note CoreOS CI is currently down so we're missing some coverage on this.

src/libostree/ostree-sysroot-deploy.c
cgwalters: I think logging to the journal as a general rule has more elegance.  But...yes you're right, we have issues where we're not always run under systemd.

I used `fprintf (stderr, ...)` and not `g_warning()` because we use `G_DEBUG=fatal-warnings` in some cases.

It's funny: I changed the code to do that but didn't change the test which grepped the journal...and the test still passed, because of course the kola exttests are logging to the journal.

dbnicholson: Could `ot_journal_print` just write to stderr if the journal isn't enabled?

cgwalters: Ultimately...a big (huge, really) benefit of systemd over SysV is that actually just calling `printf/fprintf(stderr)` Just Works too.  So I think this is fine.  We could add such infrastructure, but it's not clear in this case it'd really be better.

jlebon (approved):

Issue #2371, opened 2021-06-07
Use GitHub Actions instead of Travis CI


dbnicholson:
Many of the PRs here fail because the travis builds get rate limited. I don't see this project moving from travis-ci.org to travis-ci.com, so we should get off of Travis CI. While there are likely other CI systems, I think GitHub Actions does everything we're currently doing on Travis (build ostree and run `make check`) and is free.

I started poking at this the other day but figured I better check that others are on board first. What do you think?

wmanley:
FWIW I've moved some projects to github actions and have been really pleased with how fast builds happen.

cgwalters:
Oh definitely.  I was trying this in https://github.com/ostreedev/ostree/pull/2305 but haven't gotten back to it yet.  If anyone plans to work on this please self-assign this issue, feel free to reuse that code or force push the PR.

dbnicholson:
Ah, I knew I'd seen something you'd done fly by. That's a good start, but I think I'm going to try to use the docker setup so that we can get more coverage than just whatever Ubuntu VM GitHub supplies.

Anyone else can feel free to take over, but I will probably try to do it later this week.

PR #2379, opened 2021-06-18
Replace Travis CI with GitHub Actions

dbnicholson:
This replaces the Travis CI infrastructure with GitHub Actions. It seems to work pretty well and there are just a few minor differences from the Travis setup. Since travis-ci.org stopped running builds 3 days ago, now seems as good a time as any to make the switch fully to GitHub Actions.

Fixes: #2371


.github/workflows/tests.yml
```diff
@@ -1,27 +1,106 @@
...
+              apt-get update
+              apt-get install -y git libc6:amd64 libstdc++6:amd64
+
+          - name: Ubuntu Focal (20.04)
+            image: ubuntu:focal
+            pre-checkout-setup: |
+              apt-get update
+              apt-get install -y git
+
+          - name: Ubuntu Groovy (20.10)
```
smcv, line 75: groovy is a short-term-support release and has already been superseded by hirsute (21.04), so it might make more sense to go for hirsute - or Debian bullseye, which is roughly contemporary.

.github/workflows/tests.yml
```diff
@@ -1,27 +1,106 @@
...
+    #
+    # extra-packages: Packages to install in addition to those in
+    #   `ci/gh-install.sh`. This can be used to support features from
+    #   additional `configure` options.
+    #
+    # configure-options: Options to pass to `configure`.
+    strategy:
+      matrix:
+        include:
+          - name: Debian Buster (10) with sign-ed25519
```
smcv, line 35: Maybe use Debian bullseye (11), which is currently in freeze and will hopefully be released on 2021-07-31, for one of the Debian builds?

.github/workflows/tests.yml
dbnicholson: Yeah, this was just a straight copy of the travis configuration. I agree we could target different releases here.

I actually think it would be better to use release phase names. I.e., stable and testing (and maybe even unstable) for Debian. The docker repo has those tags. Similarly, Ubuntu has latest pointing to the current LTS (focal), rolling pointing to the current short term release (hirsute) and devel pointing to the unreleased codename (impish).

We could certainly do all of that here, but I kind of wanted to just convert the setup as is.

smcv: > I actually think it would be better to use release phase names. I.e., stable and testing (and maybe even unstable) for Debian

There isn't much point in doing both testing and unstable; they're meant to always be close, except perhaps during a freeze or a big transition. I'd recommend testing over unstable if you just want "something newish", because unstable sometimes has transient brokenness (uninstallability etc.) which the migration process from unstable to testing is designed to filter out.

cgwalters (approved):
Awesome!  LGTM to merge as is, we can tweak things like which releases are tested as followups too from my PoV.  (But also good to make the tweaks smcv is recommending!)

dbnicholson:
> Awesome! LGTM to merge as is, we can tweak things like which releases are tested as followups too from my PoV. (But also good to make the tweaks smcv is recommending!)

I'll do a follow up and CC smcv. I might add Arch and/or Alpine if I can figure them out, but I think that's getting pretty tangential to this work.

PR #2380, opened 2021-06-18
Don't fail build when systemd unit path not defined

dbnicholson:
In configure the systemd unit path is optional, but in the code it's
assumed to be defined. Add an `#ifdef` that throws an error when it's
not defined like the handling of `HAVE_LIBMOUNT` below it.


dbnicholson:
I ran a build locally with `--with-systemdsystemunitdir=no`, checked that `SYSTEM_DATA_UNIT_PATH` wasn't defined in `config.h` and the build succeeded.

cgwalters (approved):

cgwalters:
Thank you!

PR #2382, opened 2021-06-21
Update GitHub Tests workflow configurations

dbnicholson:
This changes the existing Debian and Ubuntu configurations to use the release stage tags instead of the codename tags. A Debian testing build is added while the Ubuntu groovy build is changed from groovy to hirsute via the `rolling` tag. An Alpine Linux build is added, too.

CC @smcv

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,8 +35,13 @@ jobs:
 
       matrix:
         include:
-          - name: Debian Buster (10) with sign-ed25519
-            image: debian:buster-slim
+          # Debian builds. Currently stable and testing are tested.
+          # Other options would be stable-backports, oldstable,
+          # oldstable-backports and unstable.
+          #
+          # https://hub.docker.com/_/debian
+          - name: Debian Stable with sign-ed25519
+            image: debian:stable-slim
             pre-checkout-setup: |
               apt-get update
               apt-get install -y git
@@ -45,8 +50,8 @@ jobs:
             configure-options: >-
               --with-ed25519-libsodium
 
-          - name: Debian Buster (10) with curl, sign-ed25519 and no gpgme
-            image: debian:buster-slim
+          - name: Debian Stable with curl, sign-ed25519 and no gpgme
+            image: debian:stable-slim
             pre-checkout-setup: |
               apt-get update
               apt-get install -y git
@@ -59,8 +64,8 @@ jobs:
 
           # A 32 bit build to act as a proxy for frequently deployed 32
           # bit armv7
-          - name: Debian Buster (10) 32 bit
-            image: i386/debian:buster-slim
+          - name: Debian Stable 32 bit
+            image: i386/debian:stable-slim
             # This is pretty nasty. The checkout action uses an x86_64
             # node binary in the container, so we need to provide an
             # x86_64 ld.so and libstdc++.
@@ -69,14 +74,26 @@ jobs:
               apt-get update
               apt-get install -y git libc6:amd64 libstdc++6:amd64
 
-          - name: Ubuntu Focal (20.04)
-            image: ubuntu:focal
+          - name: Debian Testing
+            image: debian:testing-slim
             pre-checkout-setup: |
               apt-get update
               apt-get install -y git
 
-          - name: Ubuntu Groovy (20.10)
-            image: ubuntu:groovy
+          # Ubuntu builds. Unfortunately, when the latest release is
+          # also the latest LTS, latest and rolling are the same. Other
+          # options would be to test the previous LTS by name or to test
+          # the devel tag, which is the unreleased version.
+          #
+          # https://hub.docker.com/_/ubuntu
+          - name: Ubuntu Latest LTS
+            image: ubuntu:latest
+            pre-checkout-setup: |
+              apt-get update
+              apt-get install -y git
+
+          - name: Ubuntu Latest Release
+            image: ubuntu:rolling
             pre-checkout-setup: |
               apt-get update
               apt-get install -y git
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,6 +30,9 @@ jobs:
     #
     # configure-options: Options to pass to `configure`.
     strategy:
+      # Let other configurations continue if one fails.
+      fail-fast: false
+
       matrix:
         include:
           - name: Debian Buster (10) with sign-ed25519
--- a/ci/gh-install.sh
+++ b/ci/gh-install.sh
@@ -26,8 +26,6 @@
 set -euo pipefail
 set -x
 
-NULL=
-
 # Get the OS release info
 . /etc/os-release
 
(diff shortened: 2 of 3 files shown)


dbnicholson:
This follows on from some of the discussion in #2379.

.github/workflows/tests.yml
```diff
@@ -19,21 +19,30 @@ jobs:
...
+      # Let other configurations continue if one fails.
+      fail-fast: false
+
       matrix:
         include:
-          - name: Debian Buster (10) with sign-ed25519
-            image: debian:buster-slim
+          # Debian builds. Currently stable and testing are tested.
+          # Other options would be stable-backports, oldstable,
+          # oldstable-backports and unstable.
```
smcv, line 41: I don't know how the Docker images are set up, but as general information: -backports is an overlay (like an Ubuntu PPA or Fedora COPR) rather than a fully independent version, and packages are not installed from -backports by default, even there's a -backports apt source configured in `/etc/apt/sources.list` or `/etc/apt/sources.list.d/`. If you want to install packages from -backports you have to use something like `apt-get -t stable-backports install libfoo1`.

.github/workflows/tests.yml
```diff
@@ -66,18 +75,38 @@ jobs:
               apt-get update
               apt-get install -y git libc6:amd64 libstdc++6:amd64
 
-          - name: Ubuntu Focal (20.04)
-            image: ubuntu:focal
+          - name: Debian Testing
```
smcv, line 78: If you want to avoid inflating the number of parallel builds, you could combine this with one of the build variants above.

smcv (commented):
This all looks reasonable to me, FWIW. I can't comment on Alpine.

dbnicholson:
Looks like the alpine tests are going to pass and I believe alpine upstream doesn't run the test suite, so I'm going to split that out to a separate branch to maybe be merged later.

smcv:
For the Debian builds, if you want to run `hexdump` you'll need `bsdextrautils` (for versions with util-linux >= 2.35.2-3, i.e. Debian 11  'bullseye', Ubuntu 20.10 'groovy' or later) or `bsdmainutils` (for older versions).

From experience, the best way is usually to have the default be the new behaviour, and use the old behaviour for specific suites, because there's a finite number of old suites (that we care about) and an infinite number of new suites. Pseudocode:

```
case "$suite" in
    (buster|focal|bionic) # also stretch, xenial, trusty if we still care about those
        use bsdmainutils
        ;;
    (*)
        use bsdextrautils
        ;;
```

smcv:
> For the Debian builds, if you want to run `hexdump` you'll need `bsdextrautils` (for versions with util-linux >= 2.35.2-3, i.e. Debian 11 'bullseye', Ubuntu 20.10 'groovy' or later) or `bsdmainutils` (for older versions)

This was pulled in by the dependency chain `debhelper -> man-db -> bsdextrautils` previously.

dbnicholson:
> > For the Debian builds, if you want to run `hexdump` you'll need `bsdextrautils` (for versions with util-linux >= 2.35.2-3, i.e. Debian 11 'bullseye', Ubuntu 20.10 'groovy' or later) or `bsdmainutils` (for older versions)
> 
> This was pulled in by the dependency chain `debhelper -> man-db -> bsdextrautils` previously.

Thanks. I also saw `dh-autoreconf -> quilt -> bsdextrautils | bsdmainutils`. I wish I could just pass that optional syntax to apt on the command line, but I don't think it supports that. For now I just added `bsdmainutils` since it has a dependency on `bsdextrautils`, but I'll see about adding a suite conditional. I tried to make that part a little simpler by just sourcing `/etc/os-release`, but it appears that only the stable debian releases actually fill in the details there. I guess an empty/unset `VERSION_ID` could be assumed to be testing/unstable.

smcv:
`VERSION_CODENAME` is usually the way to distinguish between suites (the codenames don't collide between Debian and Ubuntu and are what's in the apt repository); and, yes, an empty/unset `VERSION_ID` and/or `VERSION_CODENAME` implies testing/unstable (which isn't really a version, it's newer than all versions).

As of right now, we're in pre-release freeze for Debian 11 'bullseye', so `VERSION_ID` and `VERSION_CODENAME` *are* set even in testing/unstable, but don't count on that - it's usually not true.

smcv:
> For now I just added bsdmainutils since it has a dependency on bsdextrautils, but I'll see about adding a suite conditional

I can do this bit afterwards, if you want. bsdmainutils is a transitional package, which will exist in Debian 11 and then probably disappear.

dbnicholson:
> As of right now, we're in pre-release freeze for Debian 11 'bullseye', so `VERSION_ID` and `VERSION_CODENAME` _are_ set even in testing/unstable, but don't count on that - it's usually not true.

Kinda weird:
```
$ docker run --rm debian:testing-slim cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux bullseye/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
$ docker run --rm debian:unstable-slim cat /etc/os-release
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```
So, unstable-slim has `VERSION_CODENAME=bullseye` but testing-slim has no `VERSION_CODENAME`. I don't know how `/etc/os-release` is constructed in Debian. I know we have a weird process at Endless to manage it. Either way, I agree that `VERSION_CODENAME` not set should be assumed to be testing or unstable, which should be good enough for build dependency purposes.

Anyways, if I'm going to have a suite conditional in there, maybe I should just go back to the straight `Build-Depends` list. The reason I switched initially (which I forgot to put in the commit message) was that `dh-systemd` no longer exists in testing.

.github/workflows/tests.yml
dbnicholson: Good point. Taking a closer look at `stable-backports`, it just has the stable source in `/etc/apt/sources.list.d` with no other special configuration. So, it would just use the default apt preferences that downgrades the priority since the `Release` file has `NotAutomatic: yes`.

I wasn't planning to add builds for those, so I could just remove the comments here to just mention `oldstable` and `unstable`.

.github/workflows/tests.yml
dbnicholson: Good point. I don't feel strongly here one way or the other.

dbnicholson:
I think this is ready for full review. @smcv thanks for the comments so far. Let me know if the conditional package handling makes sense. The jenkins failure is unrelated and I couldn't remember if there was a way to retrigger that test.

smcv:
> So, unstable-slim has `VERSION_CODENAME=bullseye` but testing-slim has no `VERSION_CODENAME`

Looks like `testing-slim` hasn't been updated for a while, and so didn't pick up the new `base-files` yet. Every package in testing has to get there via a few days in unstable without serious regressions[1] and base-files doesn't get a special exception to that rule.

Debian testing does have `VERSION_CODENAME=bullseye`, but the Docker images can lag behind the state of the apt archive - although I'm surprised they're that far behind (it would have to be more than a month).

[1] not strictly true, there's a shortcut route that can be used in emergencies, but the release team tries very hard not to use it, because packages that go via that shortcut get basically no user testing before it's too late

ci/gh-install.sh
```diff
@@ -37,42 +35,99 @@ case "$ID" in
...
+        #
+        # Ubuntu package data:
+        # https://packages.ubuntu.com/source/impish/ostree
+        #
         # TODO: fetch this list from the Debian packaging git repository?
+
+        # First construct a list of Build-Depends common to all
+        # versions. This includes build-essential, which is assumed to
+        # be installed on all Debian builders. We also add gjs to allow
+        # the JS tests to run even though gjs is explicitly disable in
```
smcv, line 51: Oh, thanks for the reminder, I should put that back after bullseye gets released.

ci/gh-install.sh
```diff
@@ -37,42 +35,99 @@ case "$ID" in
...
+            systemd
+        )
+
+        # Distro specific packages. Matching is on VERSION_CODENAME from
+        # /etc/os-release. Debian testing and unstable may not have this
+        # set, so assume an empty or unset value represents those.
+
+        # hexdump was previously provided by bsdmainutils but is now in
+        # bsdextrautils.
+        case "${VERSION_CODENAME:-}" in
```
smcv, line 120: Yes, this conditional installation makes sense to me.

smcv (commented):
I'll take another look when I'm more caffeinated but this seems fine so far.

smcv (approved):

ci/gh-install.sh
cgwalters: Yep, we'll stay on target there.

cgwalters (approved):

jlebon (approved):
Cool, thanks a lot for all these CI improvements!

jlebon:
CI is all green, so merging! (We should be able to make a subset of those blocking and get the auto-merge feature back soon: https://github.com/openshift/release/pull/19571).

Issue #2385, opened 2021-06-22
ostree_repo_write_commit: respect SOURCE_DATE_EPOCH


lucab:
There are currently (at least) two ways to perform a commit and record its timestamp:
 *  `ostree_repo_write_commit_with_time`: this takes an explicit `time` parameter as the commit timestamp
 * `ostree_repo_write_commit`: this internally retrieves the "current" timestamp and uses that as the commit timestamp

For the latter case specifically, it would be nice to allow consumers to transparently set the definition of "current" in a deterministic way, and without having to plumb it as a configurable option through interfaces/layers.

Folks involved in the reproducible-builds effort already come up with standard way of doing this via the `SOURCE_DATE_EPOCH` env variable, see https://reproducible-builds.org/docs/source-date-epoch/.
It would good for libostree `ostree_repo_write_commit` to be aware of this env variable in order to override the logic which injects the current timestamp.

PR #2386, opened 2021-06-22
lib/commit: respect SOURCE_DATE_EPOCH for commit timestamp

lucab:
This tweaks `ostree_repo_write_commit` so that it checks for the
envinroment variable `SOURCE_DATE_EPOCH` as a way to override
the current time, which is used as the commit timestamp.

Ref: https://reproducible-builds.org/docs/source-date-epoch/
Ref: https://reproducible-builds.org/specs/source-date-epoch/
Closes: https://github.com/ostreedev/ostree/issues/2385

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -69,6 +69,7 @@ _installed_or_uninstalled_test_scripts = \
 	tests/test-remote-add.sh \
 	tests/test-remote-headers.sh \
 	tests/test-commit-sign.sh \
+	tests/test-commit-timestamp.sh \
 	tests/test-export.sh \
 	tests/test-help.sh \
 	tests/test-libarchive.sh \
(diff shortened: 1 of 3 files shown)


dbnicholson (changes requested):
Looks great, but can you document it in the gtk-doc?

lucab:
@dbnicholson yup, amended!

cgwalters (approved):
Nice!  This makes total sense though I would like to know where (if?) you plan to use this.  For example do you see us changing coreos-assembler to set this?  

IOW in what context did this come up?

lucab:
> Luca suggested to use this env variable which works fine for us too

I don't strictly see them as exclusive. I think this was generally good to have for other consumers we don't control or possibly know about, as proper FLOSS citizens. For the rpm-ostree/osbuild case specifically, we can totally have a dedicated `--timestamp` option if we want to expose it in a more polished way (i.e. a knob that users are encourage to use directly).

cgwalters:
Makes sense!  My goal here really was just to ensure we have "backreferences" in the commit messages/comments in the code from known likely users so that if e.g. 2 years from now someone comes by and needs to consider changing something about this they can look at what the consuming code is doing to see if it may break, etc.

PR #2394, opened 2021-07-09
ci/release-build: evaluate package_version from m4 definition

lucab:
This fixes the ci-release-build.sh script to directly source
and evaluate 'package_version' from its m4 definition, without
requiring a fully configured source tree.

### DIFF
--- a/ci/ci-release-build.sh
+++ b/ci/ci-release-build.sh
@@ -10,14 +10,26 @@ set -euo pipefail
 HEAD=${PAPR_COMMIT:-HEAD}
 
 git log --format=%B -n 1 $HEAD > log.txt
-trap "rm -f log.txt" EXIT
+trap "rm -f version.m4 log.txt" EXIT
 
 if grep -q ^is_release_build=yes configure.ac; then
     echo "*** is_release_build is set to yes ***"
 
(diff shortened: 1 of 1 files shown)


cgwalters:
$kingdom for equivalent of `cargo metadata` right?

ci/ci-release-build.sh
```diff
@@ -10,14 +10,26 @@ set -euo pipefail
...
     echo "*** is_release_build is set to yes ***"
 
-    V=$(grep -Po '^#define PACKAGE_VERSION "\K[0-9]+\.[0-9]+(?=")' config.h)
+    # assemble a short m4 macro file to evaluate 'package_version'
+    cat > version.m4 <<EOF
+m4_divert(-1)
+m4_changequote([, ])
+EOF
+
+    grep m4_define configure.ac | grep _version >> version.m4
```
dbnicholson, line 24: Another possibility here is to move the `m4_define` calls out of configure.ac and then `m4_include` it. Then it would be a little more straightforward to get it here. And you could even store this boilerplate in a file next to it.

lucab: Good point, thanks! I doubt anybody loves digging into autoconf nowadays, so I'd rather keep the m4 side simple and hide the ugly things in the CI script.

lucab:
This worked as expected on https://github.com/ostreedev/ostree/pull/2396.

PR #2401, opened 2021-07-15
Remote GPG key info

dbnicholson:
This is split out from #2260. It adds an API, `ostree_repo_remote_get_gpg_keys`, and CLI (`remote list-gpg-keys`) for getting information about GPG keys associated with a remote (or the global keys when no remote is specified).

That's useful on its own, but the latter part of the PR adds an implementation to generate [OpenPGP Web Key Directory URLs](https://datatracker.ietf.org/doc/html/draft-koch-openpgp-webkey-service#section-3.1) so that you can get a well defined place to download an updated key. In #2260 I was trying to go further and have ostree API to go fetch those and import them, but I think with the appropriate information available it would be easy to write a tool that does the fetching and importing itself. That part isn't strictly necessary for anything, but I found it convenient to have the implementation here so it could be usable by any OSTree consumer that relies on GPG verification.

There are a couple parts that I'm not really sure about. First, the key listing is basically a GVariant version of [gpgme_key_t](https://www.gnupg.org/documentation/manuals/gpgme/Key-objects.html) and friends. I'd kinda prefer to work with a real data structure, but I was unsure about committing to a type there.

Second, Web Key Directory (WKD) URLs use an obscure base32 encoding referred to as [zbase32](https://en.wikipedia.org/wiki/Base32#z-base-32) that's supposed to be more usable for humans. I find the use dubious as what it's encoding is a SHA1 checksum, so the encoding is always 32 characters and I know I could never handle that in my head. Anyways, that's what's in the spec and I copied in the author's reference C implementation. If that's troublesome, I could also code up a naive implementation. It's just base32 with an alternate alphabet and this isn't in any fast path.


src/libostree/ostree-gpg-verifier.c
```diff
@@ -185,6 +185,91 @@ _ostree_gpg_verifier_import_keys (OstreeGpgVerifier  *self,
...
+
+  if (out_keys != NULL)
+    *out_keys = g_steal_pointer (&keys);
+
+  ret = TRUE;
+
+ out:
+  if (tmp_dir != NULL) {
+    ot_gpgme_kill_agent (tmp_dir);
+    (void) glnx_shutil_rm_rf_at (AT_FDCWD, tmp_dir, NULL, NULL);
```
cgwalters, line 267: We can probably make these autocleanups too in the future.

src/libostree/ostree-repo.c
```diff
@@ -2353,6 +2353,130 @@ ostree_repo_remote_gpg_import (OstreeRepo         *self,
...
+      g_auto(GVariantBuilder) subkeys_builder = OT_VARIANT_BUILDER_INITIALIZER;
+      g_variant_builder_init (&subkeys_builder, G_VARIANT_TYPE ("a(a{sv})"));
+      g_auto(GVariantBuilder) uids_builder = OT_VARIANT_BUILDER_INITIALIZER;
+      g_variant_builder_init (&uids_builder, G_VARIANT_TYPE ("a(a{sv})"));
+      for (gpgme_subkey_t subkey = key->subkeys; subkey != NULL;
+           subkey = subkey->next)
+        {
+          g_auto(GVariantDict) subkey_dict = OT_VARIANT_BUILDER_INITIALIZER;
+          g_variant_dict_init (&subkey_dict, NULL);
+          g_variant_dict_insert_value (&subkey_dict, "fingerprint",
```
cgwalters, line 2424: I'd love to figure out a good way to have reliable documentation for these keys.  It's a downside of the gvariant approach.

cgwalters (commented):
Only skimmed this so far but I think it makes sense.

Mind splitting out the prep patches as a separate PR?

src/libostree/ostree-gpg-verifier.c
dbnicholson: I think so. The gpgme code could use some sprucing up.

src/libostree/ostree-repo.c
```diff
@@ -2353,6 +2353,144 @@ ostree_repo_remote_gpg_import (OstreeRepo         *self,
...
+                                             g_variant_dict_end (&metadata_dict));
+      g_ptr_array_add (keys, g_variant_ref_sink (key_variant));
+    }
+
+  if (out_keys)
+    *out_keys = g_steal_pointer (&keys);
+
+  return TRUE;
+#else /* OSTREE_DISABLE_GPGME */
+  return glnx_throw (error, "GPG feature is disabled in a build time");
```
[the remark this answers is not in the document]
dbnicholson: Right. I'm pretty sure I copied this from another optional GPG feature, but there's no reason to copy the slightly wrong grammar.

src/libostree/ostree-repo.h
```diff
@@ -1425,6 +1425,48 @@ gboolean      ostree_repo_remote_get_gpg_verify_summary (OstreeRepo  *self,
...
+/**
+ * OSTREE_GPG_KEY_GVARIANT_FORMAT:
+ *
+ * - aa{sv} - Array of subkeys. Each a{sv} dictionary represents a
+ *   subkey. The primary key is the first subkey. The following keys are
+ *   currently recognized:
+ *   - key: `fingerprint`, value: `s`, key fingerprint hexadecimal string
+ *   - key: `created`, value: `x`, key creation timestamp (seconds since
+ *     the Unix epoch in UTC, big-endian)
+ *   - key: `expires`, value: `x`, key expiration timestamp (seconds since
```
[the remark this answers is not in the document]
dbnicholson: Possibly. This is copying the gpgme approach where `0` is no expiration (as noted a little later in the comment). I could special case `0` to allow for a `mx`. According to the [OpenPGP RFC](https://datatracker.ietf.org/doc/html/rfc4880#section-5.2.3.6) either missing or `0` are interpreted as not expiring. I think treating `0` as never expires is a little simpler and consistent with gpgme.

src/libotutil/zbase32.c
```diff
@@ -0,0 +1,141 @@
...
+		return b2a_l_extra_Duffy(os, lengthinbits);
+}
+
+static zstr b2a(const czstr os)
+{
+	return b2a_l(os, os.len*8);
+}
+
+char *
+zbase32_encode(const unsigned char *data, size_t length)
```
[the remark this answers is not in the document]
dbnicholson: Yeah, that makes sense. I think you could also add an assert that length is less than `SIZE_MAX / 8`. I don't see how the rest of the code could overflow that condition.

man/ostree-remote.xml
```diff
@@ -65,6 +65,9 @@ Boston, MA 02111-1307, USA.
             <cmdsynopsis>
                 <command>ostree remote gpg-import</command> <arg choice="opt" rep="repeat">OPTIONS</arg> <arg choice="req">NAME</arg> <arg choice="opt" rep="repeat">KEY-ID</arg>
             </cmdsynopsis>
+            <cmdsynopsis>
+                <command>ostree remote list-gpg-keys</command> <arg choice="req">NAME</arg>
```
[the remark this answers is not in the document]
dbnicholson: Good catch. I like `gpg-list-keys`.

src/libostree/ostree-repo.c
dbnicholson: Agreed. You can do decently in the `FORMAT` documentation, but it could definitely be better.

dbnicholson: I could also just turn this into a few boxed structs if that would be better. I actually really prefer working with actual data structures rather than variants, but I chose a variant to allow more flexibility for future changes.

dbnicholson:
> Mind splitting out the prep patches as a separate PR?

For the prep do you mean everything before `lib/repo: Add ostree_repo_remote_get_gpg_keys()`?

cgwalters (approved):
OK, let's merge this - we can do further commits as followups.

dbnicholson:
> OK, let's merge this - we can do further commits as followups.

Are you planning a release? Most of the outstanding stuff is pretty minimal, but @lucab's suggestion of `ostree remote gpg-list-keys` instead of `ostree remote list-gpg-keys` should probably be done before a release goes out.

PR #2406, opened 2021-07-26
fix: Avoid wild pointers

liujianqiang-niu:
Pointer command is dangerous if there is no assignment

Log: Avoid wild pointers

### DIFF
--- a/src/ostree/ot-main.c
+++ b/src/ostree/ot-main.c
@@ -148,7 +148,6 @@ ostree_run (int    argc,
             OstreeCommand *commands,
             GError **res_error)
 {
-  OstreeCommand *command;
   GError *error = NULL;
   GCancellable *cancellable = NULL;
 #ifndef BUILDOPT_TSAN
@@ -187,7 +186,7 @@ ostree_run (int    argc,
 
   argc = out;
 
-  command = commands;
+  OstreeCommand *command = commands;
   while (command->name)
     {
       if (g_strcmp0 (command_name, command->name) == 0)


lucab:
Thanks for the PR. Do you maybe have more context to add for the rationale of this patch?
Initializing this variable is likely a good idea, but from my reading of the code I think it is not currently possible to access it while not yet initialized.
Additionally, it looks like it could directly move the `command = commands` assignment upfront, instead of NULL-setting it.

liujianqiang-niu:
> Thanks for the PR. Do you maybe have more context to add for the rationale of this patch?
> Initializing this variable is likely a good idea, but from my reading of the code I think it is not currently possible to access it while not yet initialized.
> Additionally, it looks like it could directly move the `command = commands` assignment upfront, instead of NULL-setting it.

Thank you for your comments.

Yes, it's safe at the moment. We are just used to setting the pointer to null when initializing it.

So as to avoid misoperation when adding patches or adding functions in the future.

Of course, here we can directly set command = commands

I think both ways are OK. Do you think it needs to be modified into your way? Initialize command = commands

wmanley:
It might be better to remove `command` entirely. AFAICS it's just an alias for `commands`

liujianqiang-niu:
> It might be better to remove `command` entirely. AFAICS it's just an alias for `commands`

Thank you for your comments.

Subsequent procedures have pointer operation, ostree_ option_ context_ new_ with_ The commands function cannot get the initial address.

I don't think it's a good idea.

dbnicholson:
The other way to do this would be the modern ostree development process and just declare `OstreeCommand` where it's used:
```
diff --git a/src/ostree/ot-main.c b/src/ostree/ot-main.c
index d153dcec..bbaba8b0 100644
--- a/src/ostree/ot-main.c
+++ b/src/ostree/ot-main.c
@@ -148,7 +148,6 @@ ostree_run (int    argc,
             OstreeCommand *commands,
             GError **res_error)
 {
-  OstreeCommand *command;
   GError *error = NULL;
   GCancellable *cancellable = NULL;
 #ifndef BUILDOPT_TSAN
@@ -187,7 +186,7 @@ ostree_run (int    argc,
 
   argc = out;
 
-  command = commands;
+  OstreeCommand *command = commands;
   while (command->name)
     {
       if (g_strcmp0 (command_name, command->name) == 0)
```
Then there's no chance of using it uninitialized.

cgwalters:
Thank you for your contribution!

> Then there's no chance of using it uninitialized.

Agree!  @liujianqiang-niu mind reworking per the above advice?  Thank you!

liujianqiang-niu:
> Thank you for your contribution!
> 
> > Then there's no chance of using it uninitialized.
> 
> Agree! @liujianqiang-niu mind reworking per the above advice? Thank you!

I agree.

I will modify it.

lucab (approved):
LGTM, thanks!

Issue #2405, opened 2021-07-25
Tests fail with SOURCE_DATE_EPOCH set


smcv:
Debian builds official packages with [SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/) set to the date in the Debian changelog, to get a reproducible result from the same source code.

Unfortunately, since 2021.3 this seems to be making several build-time unit tests fail with this message:

```
downgraded without --allow-downgrade?
```

I think this is probably because the tests build more than one ostree commit, expecting the second one to have a strictly higher mtime, but the new `SOURCE_DATE_EPOCH` support gives them both the same mtime.

At the moment I'm working around this with `unset SOURCE_DATE_EPOCH` before running the tests in the packaging, but ideally the affected tests would unset this themselves.

dbnicholson:
I agree with the workaround. Tests that want to check that behavior can set it explicitly. I think there's only one test that does handle it and it already does manage the variable. @lucab what do you think?

lucab:
Yes, I'd agree it would be good to fix the tests upstream (i.e. here).
@smcv do you maybe have the logs from the failing test somewhere? I couldn't find the 2021.3 packages/runs on debian yet. Is the failure showing up through a plain `make check`, or is it going through some other path?

It would be good to see what is the situation across the whole testsuite. If it is only `upgrade with and without override-commit` having issues, we can surgically fix the epochs there. If it is more widespread, `libtest.sh` may carry the tweak. Or perhaps both.

smcv:
> do you maybe have the logs from the failing test run somewhere?

From a previous attempt: [ostree_2021.3-1_amd64_20210725t181136.build.gz](https://github.com/ostreedev/ostree/files/6877501/ostree_2021.3-1_amd64_20210725t181136.build.gz)

> I couldn't find the 2021.3 packages/runs on debian infra yet

I didn't upload the version that was failing tests, because it was failing tests :-)

The finalized version of 2021.3-1 with the workaround is only in `experimental` so far, because Debian is in freeze for the Debian 11 release.

> Is the failure showing up through a plain `make check`, or is it going through some other path?

https://salsa.debian.org/debian/ostree/-/blob/debian/master/debian/test.sh wraps `make check`, and currently has the workaround.

> If it is more widespread

[...]

PR #2407, opened 2021-07-26
tests: Unset SOURCE_DATE_EPOCH

smcv:
Some distributions set this during build in order to have reproducible
builds from the same source code: for example, Debian uses the date
from debian/changelog.

However, some of our tests assume that `ostree commit` will result in
a commit with the current date/time, and SOURCE_DATE_EPOCH breaks that
assumption. Unset it for our build-time tests.

Resolves: https://github.com/ostreedev/ostree/issues/2405

### DIFF
--- a/ci/gh-build.sh
+++ b/ci/gh-build.sh
@@ -44,6 +44,11 @@ ${make}
 
 # Run the tests both using check and distcheck.
 ${make} check
+
+# Some tests historically failed when package builds set this.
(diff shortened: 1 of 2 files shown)


smcv:
I'm trying this with a Debian package build, but I think it'll work.

smcv:
A modified Debian package build, reverting my packaging-level workaround for #2405 and adding this patch instead, was successful.

lucab (approved):

PR #2411, opened 2021-08-17
builtins/commit: check for conflicting permissions options

lucab:
This explicitly checks for commit command options asking for both
non-zero UID/GID and canonical permissions at the same time,
which are incompatible.

### DIFF
--- a/src/ostree/ot-builtin-commit.c
+++ b/src/ostree/ot-builtin-commit.c
@@ -555,6 +555,23 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
       detached_metadata = g_variant_ref_sink (g_variant_builder_end (builder));
     }
 
+  /* Check for conflicting options */
+  if (opt_canonical_permissions && opt_owner_uid > 0)
+    {
+      glnx_throw (error, "Cannot specify both --canonical-permissions and non-zero --owner-uid");
+      goto out;
+    }
+  if (opt_canonical_permissions && opt_owner_gid > 0)
+    {
+      glnx_throw (error, "Cannot specify both --canonical-permissions and non-zero --owner-gid");
+      goto out;
+    }
+  if (opt_selinux_policy && opt_selinux_policy_from_base)
+    {
+      glnx_throw (error, "Cannot specify both --selinux-policy and --selinux-policy-from-base");
+      goto out;
+    }
+
   if (opt_no_xattrs)
     flags |= OSTREE_REPO_COMMIT_MODIFIER_FLAGS_SKIP_XATTRS;
   if (opt_consume)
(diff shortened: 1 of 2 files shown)


tests/basic-test.sh
```diff
@@ -437,7 +437,7 @@ echo "ok user checkout"
 $OSTREE commit ${COMMIT_ARGS} -b test2 -s "Another commit" --tree=ref=test2
 echo "ok commit from ref"
 
-$OSTREE commit ${COMMIT_ARGS} -b test2 -s "Another commit with modifier" --tree=ref=test2 --owner-uid=`id -u`
+$OSTREE commit ${COMMIT_ARGS} -b test2 -s "Another commit with modifier" --tree=ref=test2 --owner-uid=0
```
cgwalters, line 440: I think we should instead stop running this test in `bare-user-only` mode, but continue to run it for other repo modes.  And while we're here, use e.g. `ostree ls` to verify that we got the expected uid in the commit.

src/ostree/ot-builtin-commit.c
```diff
@@ -441,33 +441,32 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
   context = g_option_context_new ("[PATH]");
 
   if (!ostree_option_context_parse (context, options, &argc, &argv, invocation, &repo, cancellable, error))
-    goto out;
+    return glnx_prefix_error (error, "Parsing options");
```
cgwalters, line 444: I think we should avoid mixiong `goto out` and `return` style in a single function because it's then easy to miss the manual cleanups when refactoring code.

In this case I think we can push this forward into fully using autocleanups with `_ostree_repo_auto_transaction_start (repo, cancellable, error);`?  

Oh I see, that's private API.  Blah.  Probably time to make it public.

tests/basic-test.sh
lucab: Reading through the whole testsuite, my understanding is that this basic test is only here to reach the modifier logic (and check that it doesn't blatantly segfaults, I think). To that extent, tweaking the numeric value here should play the same role.

There are already other tests doing the check you are looking for, e.g. https://github.com/ostreedev/ostree/blob/38c14b374567f736533f7f515595208fe359819f/tests/test-basic-root.sh#L40-L42

src/ostree/ot-builtin-commit.c
lucab: Oh sorry, I didn't consider that, only noticed a large chunks of not-strictly-needed gotos and tweaked them.

I've kept split commits, so can I drop that part from here. Or I can try to break+nest the logic here into multiple layers with separate cleanups (to keep the new error prefixes). I'd rather not delve into transaction changes at this point.

cgwalters: Ah yep cool; I think I'd say let's be conservative and just do the second commit here, and split the first one (err prefixing) to a separate PR and investigate making the auto txn bits public.

lucab:
Dropped one commit from here, it only contains the new check now.

PR #2412, opened 2021-08-17
lib/commit: autofix permissions for bare-user-only

lucab:
This tweaks commit logic to detect bare-user-only repositories and
canonicalize permissions automatically.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -3286,22 +3286,35 @@ _ostree_repo_commit_modifier_apply (OstreeRepo               *self,
                                     GFileInfo                *file_info,
                                     GFileInfo               **out_modified_info)
 {
+  gboolean canonicalize_perms = FALSE;
+  gboolean has_filter = FALSE;
   OstreeRepoCommitFilterResult result = OSTREE_REPO_COMMIT_FILTER_ALLOW;
   GFileInfo *modified_info;
 
-  if (modifier == NULL ||
-      (modifier->filter == NULL &&
-       (modifier->flags & OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS) == 0))
+  /* Auto-detect bare-user-only repo, force canonical permissions. */
+  if (self->mode == OSTREE_REPO_MODE_BARE_USER_ONLY)
+    canonicalize_perms = TRUE;
+
+  if (modifier != NULL)
+    {
+      if ((modifier->flags & OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS) != 0)
+        canonicalize_perms = TRUE;
+      if (modifier->filter != NULL)
+        has_filter = TRUE;
+    }
+
+  if (!(canonicalize_perms || has_filter))
     {
       *out_modified_info = g_object_ref (file_info);
-      return OSTREE_REPO_COMMIT_FILTER_ALLOW;
+      return OSTREE_REPO_COMMIT_FILTER_ALLOW; /* Note: early return (no actions needed) */
     }
 
   modified_info = g_file_info_dup (file_info);
-  if (modifier->filter)
+
+  if (has_filter)
     result = modifier->filter (self, path, modified_info, modifier->user_data);
 
-  if ((modifier->flags & OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS) != 0)
+  if (canonicalize_perms)
     {
       guint mode = g_file_info_get_attribute_uint32 (modified_info, "unix::mode");
       switch (g_file_info_get_file_type (file_info))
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -3286,22 +3286,33 @@ _ostree_repo_commit_modifier_apply (OstreeRepo               *self,
...
+  gboolean canonicalize_perms = FALSE;
+  gboolean has_filter = FALSE;
   OstreeRepoCommitFilterResult result = OSTREE_REPO_COMMIT_FILTER_ALLOW;
   GFileInfo *modified_info;
 
-  if (modifier == NULL ||
-      (modifier->filter == NULL &&
-       (modifier->flags & OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS) == 0))
+  /* Auto-detect bare-user-only repo, force canonical permissions. */
+  // XXX(lucab): This seems to break `checkout --union-identical`.
```
lucab, line 3295: Still need to investigate what's going on here. From what I've observed so far, the is-identical logic is failing due to a checksum mismatch on symlinks.

dbnicholson:
Obviously it would be good to add a test that committing to a bare-user-only repo without `--canonical-permissions` results in canonical permissions begin used. I think it would also be useful to document in `OstreeRepoCommitModifierFlags` that `OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS` is implied when the target repository is in `bare-user-only` mode.

Also, are you planning to handle `OSTREE_REPO_COMMIT_MODIFIER_FLAGS_SKIP_XATTRS` in this PR?

lucab:
@dbnicholson I'd like to scope this to the uid/gid/mode part only, and tackle xattrs separately. Thanks for your other suggestions, I'll incorporate them.

Right now though this is a WIP as I'm blocked tracking down the non-identical glitch I'm seeing, which for some reason does not happen in CI (possibly because it runs as root in a container).

src/libostree/ostree-repo-commit.c
lucab: Found the problem, fixed by https://github.com/ostreedev/ostree/pull/2415.

lucab:
Rebased, this is now ready to land.

src/libostree/ostree-repo.h
```diff
@@ -678,10 +678,14 @@ typedef OstreeRepoCommitFilterResult (*OstreeRepoCommitFilter) (OstreeRepo    *r
  * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_NONE: No special flags
  * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_SKIP_XATTRS: Do not process extended attributes
  * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_GENERATE_SIZES: Generate size information.
- * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS: Canonicalize permissions for bare-user-only mode.
+ * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS: Canonicalize permissions, for bare-user-only mode.
```
jlebon, line 681: Should we just simplify this to

```suggestion
 * @OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS: Canonicalize permissions
```

?

Since the only reason you would pass this explicitly now is if you want the behaviour on non-bare-user-only repos.

tests/test-basic-user-only.sh
```diff
@@ -112,3 +112,15 @@ $OSTREE checkout --force-copy perms out
 $OSTREE checkout ${CHECKOUT_H_ARGS} --union-identical perms out
 $OSTREE fsck
 echo "ok checkout checksum with canonical perms"
+
+cd ${test_tmpdir}
+rm repo -rf
+ostree_repo_init repo init --mode=bare-user-only
+rm files -rf && mkdir files
+echo afile > files/afile
+$OSTREE commit --no-xattrs -b perms files
```
jlebon, line 121: We should be able to drop `--canonical-permissions` from `COMMIT_ARGS` and use `COMMIT_ARGS` here too, no?

src/libostree/ostree-repo-commit.c
```diff
@@ -3286,22 +3286,32 @@ _ostree_repo_commit_modifier_apply (OstreeRepo               *self,
...
+  gboolean canonicalize_perms = FALSE;
+  gboolean has_filter = FALSE;
   OstreeRepoCommitFilterResult result = OSTREE_REPO_COMMIT_FILTER_ALLOW;
   GFileInfo *modified_info;
 
-  if (modifier == NULL ||
-      (modifier->filter == NULL &&
-       (modifier->flags & OSTREE_REPO_COMMIT_MODIFIER_FLAGS_CANONICAL_PERMISSIONS) == 0))
+  /* Auto-detect bare-user-only repo, force canonical permissions. */
+  canonicalize_perms |= (self->mode == OSTREE_REPO_MODE_BARE_USER_ONLY);
```
jlebon, line 3295: Minor: this is OK, though to be consistent with the rest of the codebase might be worth avoiding bitwise ops for these pseudo-booleans.

src/libostree/ostree-repo-commit.c
```diff
@@ -3286,22 +3286,32 @@ _ostree_repo_commit_modifier_apply (OstreeRepo               *self,
...
       *out_modified_info = g_object_ref (file_info);
-      return OSTREE_REPO_COMMIT_FILTER_ALLOW;
+      return OSTREE_REPO_COMMIT_FILTER_ALLOW; /* Note: early return (no actions needed) */
     }
 
   modified_info = g_file_info_dup (file_info);
-  if (modifier->filter)
+
+  if (has_filter)
     result = modifier->filter (self, path, modified_info, modifier->user_data);
```
jlebon, line 3312: Aside: hmm, we should be able to just return early here if the filter returned `OSTREE_REPO_COMMIT_FILTER_SKIP`. Need to double-check, maybe there's a subtle reason we still need to modify the `GFileInfo`.

tests/test-basic-user-only.sh
lucab: Yes! And... it uncovered another bug :disappointed:
It looks like file adoption (`commit --consume`) is also broken for bare-user-only.

src/libostree/ostree-repo-commit.c
lucab: How would you prefer to write this? `if (mode is user-only) { canonicalize_perms = TRUE }`?

jlebon: Yup, exactly! With a newline and less braces. :)

lucab:
Rebased once more.

cgwalters (approved):

Issue #2410, opened 2021-08-06
fsck fails after ostree commit for bare-user-only repo


cxrvh:
Files committed with `ostree commit` into a bare-user-only repo [0] are found corrupted when checked with `ostree fsck` although the files are fine (SHA256 sum is valid after checkout):
```
Validating refs...
Validating refs in collections...
Enumerating objects...
Verifying content integrity of 1 commit objects...
fsck objects (1/4) [===          ]  25%
error: In commits d2d11564c36e4ce7a27063029e79e3623458354a06ac2a1a1f8d6c4e2dae6bbf:
fsck content object 840ce1349a9e6ab32c02cbe26d5f60c8c46847e1e60af545a0f0bdab571290de:
Corrupted file object; checksum expected='840ce1349a9e6ab32c02cbe26d5f60c8c46847e1e60af545a0f0bdab571290de'
actual='9ae371068638ba594f6874af81ccda2e2b5a61d87f7141acbb994355b0e410b7'
```
### How to reproduce
```
mkdir commit
echo "Value" > commit/value

# This fails
ostree init --mode bare-user-only --repo=bare-user-only
ostree --repo=bare-user-only commit -s 'Commit test' -b 'test' commit
ostree --repo=bare-user-only fsck

# This works
ostree init --mode bare-user --repo=bare-user
ostree --repo=bare-user commit -s 'Commit test' -b 'test' commit
ostree --repo=bare-user fsck
```

`ostree fsck` validates Flatpak repos (bare-user-only according to the config) without issues.
On Arch Linux (without SELinux), the `ostree checksum` on the object files returns the expected checksum (just as in Flatpak repos):
```
ostree checksum commit/value
840ce1349a9e6ab32c02cbe26d5f60c8c46847e1e60af545a0f0bdab571290de

ostree checksum bare-user-only/objects/84/0ce1349a9e6ab32c02cbe26d5f60c8c46847e1e60af545a0f0bdab571290de
840ce1349a9e6ab32c02cbe26d5f60c8c46847e1e60af545a0f0bdab571290de
```

### System
The behavior is reproducible on Fedora 33 (2021.2, SELinux enforcing) & Arch Linux (2021.3), with and without `--no-xattrs`:
```
ostree --version
libostree:
 Version: '2021.3'
 Git: v2021.3
 Features:
  - libsoup
  - gpgme
  - sign-ed25519
  - ex-fsverity
  - libarchive
  - openssl
  - avahi
  - libmount
[...]

PR #2415, opened 2021-08-19
lib/checkout: use canonical permissions in bare-user-only mode

lucab:
This adds a new `OSTREE_CHECKSUM_FLAGS_CANONICAL_PERMISSIONS`
checksumming flag, which is needed in bare-user-only mode
to ignore local IDs.
It also automatically enables canonical permissions for checkouts in
bare-user-only mode.

### DIFF
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -31,7 +31,7 @@ if is_bare_user_only_repo repo; then
     # In bare-user-only repos we can only represent files with uid/gid 0, no
     # xattrs and canonical permissions, so we need to commit them as such, or
     # we end up with repos that don't pass fsck
-    COMMIT_ARGS="--canonical-permissions"
+    COMMIT_ARGS="--canonical-permissions --no-xattrs"
     DIFF_ARGS="--owner-uid=0 --owner-gid=0 --no-xattrs"
     # Also, since we can't check out uid=0 files we need to check out in user mode
     CHECKOUT_U_ARG="-U"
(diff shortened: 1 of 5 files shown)


lucab:
This locally passed a non-root `make check` run. I think the CI here does not have a job covering that.

tests/test-basic-user-only.sh
```diff
@@ -99,3 +100,14 @@ if ! skip_one_without_user_xattrs; then
     $OSTREE fsck -q
     echo "ok hardlink pull from bare-user"
 fi
+
+cd ${test_tmpdir}
+rm repo -rf
+ostree_repo_init repo init --mode=bare-user-only
+rm files -rf && mkdir files
+echo afile > files/afile
+$OSTREE --repo=repo commit ${COMMIT_ARGS} -b perms files
```
cgwalters, line 109: Shouldn't we be running `ostree fsck` here?

lucab: Ack, I've added it.
Although this PR is not the fix for https://github.com/ostreedev/ostree/issues/2410.
`ostree fsck` only happens to work here because `${COMMIT_ARGS}` values let this test land on a happy path.

tests/basic-test.sh
```diff
@@ -31,7 +31,7 @@ if is_bare_user_only_repo repo; then
     # In bare-user-only repos we can only represent files with uid/gid 0, no
     # xattrs and canonical permissions, so we need to commit them as such, or
     # we end up with repos that don't pass fsck
-    COMMIT_ARGS="--canonical-permissions"
+    COMMIT_ARGS="--canonical-permissions --no-xattrs"
```
jlebon, line 34: I guess this is the bit we should be able to get rid of once we fix the commit side of this.

jlebon (approved):
Makes sense to me!

tests/basic-test.sh
lucab: Yes, it is the same workaround we suggested at https://github.com/ostreedev/ostree/issues/2410#issuecomment-896144099.

Issue #2357, opened 2021-05-06
ostree on Beaglebone Black


jallwine:
I'm looking into what it would take to use ostree to manage full system atomic updates on Beaglebone boards. The official Beaglebone images for their boards are generated using scripts from this repository: https://github.com/beagleboard/image-builder 
They create a root filesystem using debootstrap and apt-get and create a /boot directory where the kernel, initrd, device trees and device tree overlays are placed. A /boot/uEnv.txt file is created with a number of environment variables that specify which device tree to load and which overlays to apply. I'm trying to figure out how to adapt those scripts to make the rootfs and boot process work with ostree. Does anyone have any insights into what needs to happen to make that work?

wmanley:
Step one would be to commit your rootfs into an ostree.  You will have to adapt it so ostree will deploy it correctly

* This covers much of it: https://ostree.readthedocs.io/en/stable/manual/adapting-existing/
* There's a bit here about where you put your kernel and initrd: https://ostree.readthedocs.io/en/stable/manual/deployment/#the-system-boot
* And a bit here about where to put device-trees: https://github.com/ostreedev/ostree/pull/2001

jallwine:
Thanks @wmanley! When it says:

> This configuration file will include a special ostree= kernel argument that allows the initramfs to find (and chroot() into) the specified deployment.

Do I need to ensure my initrd does that?

wmanley:
I don't know about initrd as I don't use it.  See #1401.

mike-sul:
This meta layer https://github.com/advancedtelematic/meta-updater is a good example of OSTree application for full system atomic updates, and this is an example of its utilization in the scope of an end-to-end product https://github.com/foundriesio/lmp-manifest.

dbnicholson:
@jallwine for the uEnv.txt, I was sure we had managed the environment variables at Endless, but when I looked closer there wasn't any of that. Apparently we've always just embedded the correct details into u-boot or something like that since we've controlled the hardware.

However, since we initially started using ostree, what you can do is populate `/usr/lib/ostree-boot/uEnv.txt` in the rootfs and ostree will merge this together with the arguments it needs to specify the `kernel_image`, `ramdisk_image` and `bootargs`. See https://github.com/ostreedev/ostree/blob/master/tests/test-admin-deploy-uboot.sh#L42 for an example in the test suite.

jallwine:
[...]

PR #2416, opened 2021-08-20
Run /usr/lib/ostree-boot/setup.sh if it exists during u-boot bootloader setup

jallwine:
Standard Beaglebone images have a custom version of U-Boot installed that looks for specific variables in the uEnv.txt file as well as specific files for loading various device tree overlays. In order to properly set up the /boot/loader.# directory at `ostree admin deploy` time, I added a check for the existence of /usr/lib/ostree-boot/setup.sh in the deployment and run it so it can set up the necessary symlinks and modifications to uEnv.txt (this is much easier than implementing it explicitly in C). The script is passed the sysroot path, the deployment path, and the /boot/loader.# path so that it can properly set everything up. If the script fails to run (not executable, for example) or exits with an error status, the deployment will fail.

See https://github.com/ostreedev/ostree/issues/2357 for more info.

### DIFF
--- a/src/libostree/ostree-bootloader-uboot.c
+++ b/src/libostree/ostree-bootloader-uboot.c
@@ -65,6 +65,79 @@ _ostree_bootloader_uboot_get_name (OstreeBootloader *bootloader)
   return "U-Boot";
 }
 
+/* Run system's setup.sh script, if it exists in $deployment/usr/lib/ostree-boot/ */
+static gboolean
+run_system_setup (OstreeBootloaderUboot   *self,
+                    const int bootversion,
+                    GCancellable            *cancellable,
+                    GError                 **error)
+{
+  // code to find bootargs is duplicated from create_config_from_boot_loader_entries
+  // is there a better way to know the deployment path?
+  g_autoptr(GPtrArray) boot_loader_configs = NULL;
+  OstreeBootconfigParser *config;
+  const char *val;
+  const char *bootargs = NULL;
+
+  if (!_ostree_sysroot_read_boot_loader_configs (self->sysroot, bootversion, &boot_loader_configs,
+                                                 cancellable, error)) {
+    return FALSE;
+  }
+
+  for (int i = 0; i < boot_loader_configs->len; i++) {
+    config = boot_loader_configs->pdata[i];
+
+    val = ostree_bootconfig_parser_get (config, "options");
+    if (val) {
+      bootargs = val;
+      break;
+    }
+  }
+
+  if(bootargs == NULL) {
+    return FALSE;
+  }
+
+  g_autoptr(OstreeKernelArgs) kargs = NULL;
+  const char *setup_path = NULL;
+  const char *setup_path_relative = NULL;
+  const char *ostree_arg = NULL;
+  GFile *sysroot_file = ostree_sysroot_get_path(self->sysroot);
+  g_autofree char* sysroot_path = g_file_get_path(sysroot_file);
+
+  kargs = ostree_kernel_args_from_string (bootargs);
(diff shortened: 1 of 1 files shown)


jallwine:
Thanks for taking a look @dbnicholson! I agree we should discuss further. I'm not sure if I'm doing it the best way, but it seems to work. I went this route due to how Beaglebone's custom version of U-Boot works. A lot of the custom U-Boot code is here: https://github.com/beagleboard/u-boot/blob/v2019.04-bbb.io-am335x/include/configs/ti_armv7_common.h The key pieces are under the uname_boot variable when it detects a uname_r variable set in the uEnv.txt file.

1) My `setup.sh` script copies the deployment's uEnv.txt file over to the proper /boot/loader.#/uEnv.txt. It overwrites what the ostree uboot configuration generated because I'm only concerned with the variables used by Beaglebone's custom version of U-Boot (this is why I was originally leaning towards creating a new bootloader configuration). All I need to change in my uEnv.txt file is add the ostree=<deployment path> to the cmdline argument. This is handled by these two lines:

```
cp ${SYSROOT}${DEPLOY}/boot/uEnv.txt uEnv.txt
sed -i "/^cmdline=/ s,$, ostree=${DEPLOY}," uEnv.txt
```

The rest of the lines are creating symlinks so that Beaglebone's custom U-Boot can find specific paths at boot. In most Beaglebone uEnv.txt files, there's a uname_r variable that should be set to the kernel version of the system. uname_r is used in a number of paths in U-Boot. These paths include places /lib/firmware and /boot. When preparing my rootfs, I set the uname_r variable to always be the value `current` rather than a specific kernel version and then create the proper symlinks to the actual kernel verison (i.e. symlink vmlinuz-current to vmlinuz-4.19.94-ti-r43), initramfs (initrd.img-4.19.94-ti-r43 -> initrd.img-current). There's also a dtbs/<kernel version> folder that contains a number of device tree overlays that can be pulled in at boot by setting specific variables (disable_uboot_overlay_emmc, disable_uboot_overlay_video, disable_uboot_overlay_audio, disable_uboot_overlay_wireless, and disable_uboot_overlay_adc). Again, I symlink that directory to dtbs/current. What's problematic is the real sysroot of the filesystem needs to have symlinks from files in /lib/firmware to the /lib/firmware in the deployment as well as symlinks from files in /boot to the deployment. So, on the real /sysroot I also created symlinks from /lib to /boot/loader/lib and /boot/vmlinuz-current to /boot/loader/vmlinuz-current, /boot/initrd.img-current to /boot/loader/initrd.img-current, etc. With it set up this way, `ostree admin deploy` needs to create the symlinks from /boot/loader.#/vmlinuz-current to the deployment's /boot/vmlinuz-current (and the initrd, dtbs folder, etc.). My `setup.sh` script creates those /boot/loader.# links.

2) I pass in the sysroot path, the deployment path and the /boot/loader.# path as the first three arguments to setup.sh. That way the script has the ability to create links however it needs to whether `ostree admin deploy` is run from a build process or on a client.

Any thoughts you have on the best way forward are appreciated!

jallwine:
Oh, and there’s one other thing that’s been in the back of my mind, but I haven’t implemented yet in the setup.sh. The U-Boot build does get updated occasionally and needs to be written to the first portion of the Beaglebone’s eMMC (I’ve run into complications with device tree overlays. There’s a script that needs to be run on a client after an ostree admin deploy that does that, so it seems appropriate to put it in setup.sh.

jowa2021:
Hello.
I am not a master like most of you here, but this seems dangerous and deceitful. Maybe I misinterpreted everything I have read here, so, can someone please help me understand more clearly ?
Would it be wrong for me to summarize this by saying that this potentially could be used to trick another individual into thinking  they were looking at one thing, but its actually another ??

PR #2419, opened 2021-08-23
bin/remote: Rename list-gpg-keys to gpg-list-keys

dbnicholson:
As pointed out in the original review, `gpg-list-keys` fits better
alongside the existing `gpg-import`.

Changes were done with:

```
git grep -l list-gpg-keys | xargs sed -i 's/list-gpg-keys/gpg-list-keys/'
for src in $(git ls-files '*list-gpg-keys*'); do
  dst=${src/list-gpg-keys/gpg-list-keys}
  git mv "$src" "$dst"
done
```


dbnicholson:
This is a followup to #2401.

lucab (approved):

Issue #2413, opened 2021-08-18
api: make transactions autocleanups public


lucab:
From the code review at https://github.com/ostreedev/ostree/pull/2411#discussion_r690416686:

>  I think we can push this forward into fully using autocleanups with
> `_ostree_repo_auto_transaction_start (repo, cancellable, error);`?
> 
> Oh I see, that's private API. Blah. Probably time to make it public.

cgwalters:
Side note: I think this one needs to be manually bound in Rust, unless we make it a boxed type.

lucab:
Regarding the switch to a boxed type, https://github.com/ostreedev/ostree-rs/pull/33/files sketches an alternative/different implementation (which I think I'm liking more than status quo) along the following lines:
 * a dedicated internal `repo` reference, nullable
 * explicit commit/abort methods which do consume the guard
 * auto-cleanup logic which only triggers an abort call if the guard as not been consumed (via explicit commit/abort) before

lucab:
I made a pass on this and ported back some of the enhancements from the Rust side, PR at https://github.com/ostreedev/ostree/pull/2454.

I also looked into moving this guard to a gobject boxed type. I admit my shallow knowledge in this area, so please correct me if here below I'm making any wrong statements.

I think that in this specific case `OstreeRepoAutoTransaction` cannot be expressed as a boxed type due to semantic mismatches. In particular, it is my understanding that a boxed type is primarily meant for type-values that can be (trivially) copied, and thus requires a mandatory non-nullable `GBoxedCopyFunc`.
In our case instead `OstreeRepoAutoTransaction` is a guard which enforces scope/ownership, so I think we won't actually expose a way to duplicate an existing guard.

lucab:
From an out-of-band chat, it would still be preferable to have auto-generated bindings for this.
Going back to a boxed type for this, we can add internal reference counting to `OstreeRepoAutoTransaction` so that copy/free functions can operate on light references, while still keeping a single guard underneath.

PR #2421, opened 2021-08-25
lib: improve transactions auto-cleanup logic

lucab:
This fixes some aspects of OstreeRepoAutoTransaction and re-aligns
it with the logic in flatpak. Specifically:
 * link to the underlying repo through refcounting
 * bridge internal errors to warning messages
 * verify the input pointer type

This is a preparation step before exposing this logic as a public API.

Refs:
 * https://github.com/ostreedev/ostree/issues/2413
 * https://github.com/flatpak/flatpak/pull/3282/commits/3e0a8f296ac9c7ad246c4eaa38ae2e472a0c1529


src/libostree/ostree-repo-private.h
```diff
@@ -234,9 +234,16 @@ typedef OstreeRepo _OstreeRepoAutoTransaction;
 static inline void
 _ostree_repo_auto_transaction_cleanup (void *p)
 {
+  if (p == NULL || !OSTREE_IS_REPO(p))
```
cgwalters, line 237: I think these are two different things; if the `OSTREE_IS_REPO(p)` check fails then there must be e.g. memory corruption involved.

How about:

```
if (p == NULL)
  return;
g_return_if_fail (OSTREE_IS_REPO (p));
```

Now in theory, API entrypoints like `ostree_repo_abort_transaction()` could also be doing the dynamic type check so we could drop it here, but I haven't bothered with that in most of our code so far.

cgwalters (approved):

src/libostree/ostree-repo-private.h
lucab: Thanks for the feedback, it looks indeed better, I'll update that way.
For completeness, I added this check because if this function becomes public API there is an higher chance of users mistakenly calling this with a wrong parameter, so some basic sanity check seems worth.

lucab:
I think the CI hit https://github.com/ostreedev/ostree/issues/2038.

PR #2428, opened 2021-08-30
tests: fix bare mode unprivileged 'make check'

lucab:
This tweaks the testsuite in order to fix unprivileged `make check`,
specifically around bare mode.
There is one case of a broken `ostree fsck` which I previously introduced
in a testcase without realizing it wasn't actually passing; I've left it commented
as a FIXME for now as I didn't get to the bottom of the issue.
After this, unprivileged `make check` should be green again.

Closes: https://github.com/ostreedev/ostree/pull/2423

### DIFF
--- a/tests/basic-test.sh
+++ b/tests/basic-test.sh
@@ -455,11 +455,18 @@ $OSTREE commit ${COMMIT_ARGS} --skip-if-unchanged -b trees/test2 -s 'should not 
 $OSTREE ls -R -C test2
 new_rev=$($OSTREE rev-parse test2)
 assert_streq "${old_rev}" "${new_rev}"
+$OSTREE fsck
 echo "ok commit --skip-if-unchanged"
 
 cd ${test_tmpdir}/checkout-test2-4
+# Unfortunately later tests depend on this right now, so commit anyways
 $OSTREE commit ${COMMIT_ARGS} -b test2 -s "no xattrs" --no-xattrs
-echo "ok commit with no xattrs"
+if have_selinux_relabel; then
+    echo "ok # SKIP we get an injected security.selinux xattr regardless, so we can't do this"
+else
+    $OSTREE fsck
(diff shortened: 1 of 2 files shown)


lucab:
@cgwalters this should make your unprivileged flow green again, assuming that I managed to reproduce the same setup locally in a faithful way. It would be helpful if you could run a `make check` on this locally to confirm.

tests/basic-test.sh
```diff
@@ -793,7 +793,8 @@ rm files -rf && mkdir files
 mkdir files/worldwritable-dir
 chmod a+w files/worldwritable-dir
 $OSTREE commit ${COMMIT_ARGS} -b content-with-dir-world-writable --tree=dir=files
-$OSTREE fsck
+# FIXME(lucab): this seems to fail in unprivileged bare mode.
+#$OSTREE fsck
```
cgwalters, line 797: Let's conditionalize this on `if is_bare_user_only_repo` like below?

cgwalters (changes requested):

tests/basic-test.sh
lucab: I thing this does work in bare-user-only mode. That's the reason why I did not see this failing when I added it initially.
I can make this conditional on `have_selinux_relabel` and keep the FIXME note, if you prefer.

cgwalters: Hmm you're saying the fsck only fails if selinux/xattrs?   Sure, we can make that conditional but...oh I see, there's also e.g.
`echo "ok # SKIP checksums won't match up in bare-user-only"` above here.

I think we should fix things so that `fsck` *does* work in that case too.  But it's fine to make it conditional too for now and then we circle back and fix it and that other conditional too.

lucab: Ack, I made this conditional and filed https://github.com/ostreedev/ostree/issues/2432 for later.

Issue #2389, opened 2021-06-30
Non-idempotent result of ostree admin upgrade


RomanValov:
If there are no upgrades available run of `ostree admin upgrade` does nothing.
Subsequent invocations of `ostree admin upgrade` also does nothing:

```
[root@ostree ~]# ostree admin upgrade
Receiving metadata objects: 1/(estimating) 98 bytes/s 196 bytes                 
No update available.
[root@ostree ~]# ostree admin upgrade
2 metadata, 0 content objects fetched; 788 B transferred in 3 seconds; 0 bytes content written
No update available.
[root@ostree ~]# ostree admin upgrade
2 metadata, 0 content objects fetched; 788 B transferred in 2 seconds; 0 bytes content written
No update available.
[root@ostree ~]# ostree admin upgrade
2 metadata, 0 content objects fetched; 788 B transferred in 2 seconds; 0 bytes content written
No update available.
```

If there are existing upgrades they're downloaded and deployed.
Subsequent invocations of `ostree admin upgrade` updates pending deployment with new deployment number. Thus `upgrade` command is acting like `deploy` command:

```
[root@ostree ~]# ostree admin upgrade
134 metadata, 72 content objects fetched; 76509 KiB transferred in 21 seconds; 90.9 MB content written
Copying /etc changes: 6 modified, 0 removed, 13 added
Bootloader updated; bootconfig swap: yes; bootversion: boot.0.1, deployment count change: 1
[root@ostree ~]# ostree admin upgrade
2 metadata, 0 content objects fetched; 788 B transferred in 3 seconds; 0 bytes content written
Copying /etc changes: 6 modified, 0 removed, 13 added
Transaction complete; bootconfig swap: no; bootversion: boot.0.0, deployment count change: 0
[root@ostree ~]# ostree admin upgrade
2 metadata, 0 content objects fetched; 788 B transferred in 3 seconds; 0 bytes content written
Copying /etc changes: 6 modified, 0 removed, 13 added
Transaction complete; bootconfig swap: no; bootversion: boot.0.1, deployment count change: 0
[root@ostree ~]# ostree admin upgrade
2 metadata, 0 content objects fetched; 788 B transferred in 2 seconds; 0 bytes content written
Copying /etc changes: 6 modified, 0 removed, 13 added
Transaction complete; bootconfig swap: no; bootversion: boot.0.0, deployment count change: 0
[root@ostree ~]# ls -l /ostree/deploy/fedora-iot/deploy/
total 16
[...]

PR #2430, opened 2021-09-02
upgrade: Stabilize deployment staging

cgwalters:
We're waaay overdue for this, it's been the default
in rpm-ostree for years, and solves several important bugs
around not capturing `/etc` while things are running.

Also, `ostree admin upgrade --stage` (should) become idempotent.

Closes: https://github.com/ostreedev/ostree/issues/2389

### DIFF
--- a/src/libostree/ostree-sysroot-upgrader.h
+++ b/src/libostree/ostree-sysroot-upgrader.h
@@ -35,12 +35,15 @@ G_BEGIN_DECLS
  * OstreeSysrootUpgraderFlags:
  * @OSTREE_SYSROOT_UPGRADER_FLAGS_NONE: No options
  * @OSTREE_SYSROOT_UPGRADER_FLAGS_IGNORE_UNCONFIGURED: Do not error if the origin has an unconfigured-state key
+ * @OSTREE_SYSROOT_UPGRADER_FLAGS_STAGE: Enable "staging" (finalization at shutdown); recommended
+ *    (Since: 2021.4)
  *
  * Flags controlling operation of an #OstreeSysrootUpgrader.
  */
 typedef enum {
   OSTREE_SYSROOT_UPGRADER_FLAGS_NONE = (1 << 0),
   OSTREE_SYSROOT_UPGRADER_FLAGS_IGNORE_UNCONFIGURED = (1 << 1),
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-sysroot-upgrader.h
```diff
@@ -35,12 +35,14 @@ G_BEGIN_DECLS
  * OstreeSysrootUpgraderFlags:
  * @OSTREE_SYSROOT_UPGRADER_FLAGS_NONE: No options
  * @OSTREE_SYSROOT_UPGRADER_FLAGS_IGNORE_UNCONFIGURED: Do not error if the origin has an unconfigured-state key
+ * @OSTREE_SYSROOT_UPGRADER_FLAGS_STAGE: Enable "staging" (finalization at shutdown); recommended
```
lucab, line 38: Looking at other flag-enums, I think we usually add a `Since:` annotation when introducing new variants.

lucab, line 38: Looking at other flag-enums, I think we usually add a `Since:` annotation when introducing new variants.

tests/kolainst/destructive/staged-deploy.sh
```diff
@@ -70,7 +67,7 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
     ostree checkout -H "${origcommit}" t
     ostree commit --no-bindings --parent="${origcommit}" -b staged-deploy -I --consume t
     newcommit=$(ostree rev-parse staged-deploy)
-    env OSTREE_EX_STAGE_DEPLOYMENTS=1 ostree admin upgrade >out.txt
+    ostree admin upgrade --stage >out.txt
```
lucab, line 70: It looks like this may be wrongly testing the old/installed libostree instead of the new logic from the PR. The `ostree` binary though seems fresh. The failed run in CI shows this:
```
+ ostree admin upgrade --stage

ostree[1077]: value "((OstreeSysrootUpgraderFlags) 4)" of type 'OstreeSysrootUpgraderFlags' is invalid or out of range for property 'flags' of type 'OstreeSysrootUpgraderFlags'
```

cgwalters: Thanks for debugging that.  It was because we aren't using `glib-mkenums` for this one and so need to define it twice.

lucab (approved):

PR #2434, opened 2021-09-07
Add support for "custom remotes"

cgwalters:
This will be helpful for the "ostree native container" work in
https://github.com/ostreedev/ostree-rs-ext/

Basically in order to reuse GPG/signapi verification, we need
to support adding a remote, even though it can't be used via
`ostree pull`.  (At least, not until we merge ostree-rs-ext into ostree, but
 even then I think the principle stands)

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -355,6 +355,14 @@ Boston, MA 02111-1307, USA.
         <listitem><para>If set, pulls from this remote will fail with the configured text.  This is intended for OS vendors which have a subscription process to access content.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>custom-backend</varname></term>
+        <listitem><para>If set, pulls from this remote via libostree will fail with an error that mentions the value.
+                        It is recommended to make this a software identifier token (e.g. "examplecorp-fetcher"), not freeform text ("ExampleCorp Fetcher").
+                        This is intended to be used by higher level software that wants to fetch ostree commits via some other mechanism, while still reusing the core libostree infrastructure around e.g. signatures.
+                        </para></listitem>
+      </varlistentry>
+
     </variablelist>
 
   </refsect1>
--- a/src/ostree/ot-remote-builtin-add.c
+++ b/src/ostree/ot-remote-builtin-add.c
@@ -35,6 +35,7 @@ static char *opt_gpg_import;
 static char **opt_sign_verify;
 static char *opt_contenturl;
 static char *opt_collection_id;
+static char *opt_custom_backend;
 static char *opt_sysroot;
 static char *opt_repo;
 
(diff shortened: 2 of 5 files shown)


src/ostree/ot-remote-builtin-add.c
```diff
@@ -51,6 +52,7 @@ static GOptionEntry option_entries[] = {
   { "if-not-exists", 0, 0, G_OPTION_ARG_NONE, &opt_if_not_exists, "Do nothing if the provided remote exists", NULL },
   { "force", 0, 0, G_OPTION_ARG_NONE, &opt_force, "Replace the provided remote if it exists", NULL },
   { "gpg-import", 0, 0, G_OPTION_ARG_FILENAME, &opt_gpg_import, "Import GPG key from FILE", "FILE" },
+  { "custom-backend", 0, 0, G_OPTION_ARG_STRING, &opt_custom_backend, "This remote has content not fetched via libostree", "NAME" },
```
lucab, line 55: Is `--contenturl` here below compatible with this?
In other word, is a custom-backed handler meant to handle a `(Option<Url>, Option<ContentUrl>)` for a remote?

tests/pull-test.sh
```diff
@@ -598,6 +598,15 @@ assert_file_has_content err.txt "ONE BILLION DOLLARS"
 
 echo "ok unconfigured"
 
+cd ${test_tmpdir}
+${CMD_PREFIX} ostree --repo=repo remote add --set=gpg-verify=false --custom-backend=ostree-rs-ext fromcontainer
+if ${CMD_PREFIX} ostree --repo=repo pull fromcontainer main 2>err.txt; then
```
lucab, line 603: It may be good to cover the behavior of summary/list/show-url too.

man/ostree.repo-config.xml
```diff
@@ -355,6 +355,11 @@ Boston, MA 02111-1307, USA.
         <listitem><para>If set, pulls from this remote will fail with the configured text.  This is intended for OS vendors which have a subscription process to access content.</para></listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>custom-backend</varname></term>
+        <listitem><para>If set, pulls from this remote via libostree will fail with an error.  This is intended to be used by higher level software that wants to fetch ostree commits via some other mechanism, while still reusing the core libostree infrastructure.</para></listitem>
```
lucab, line 360: This is worded like the description of a boolean field (maybe it was in a initial code sketch?).
However my understanding is that the value of this option is really a freeform kind/tag which can be used to dispatch/gate further operations to relevant handlers (including decoding the optional URL too).

cgwalters: > (maybe it was in a initial code sketch?).

It was, well spotted.   I tweaked the description.

src/ostree/ot-remote-builtin-add.c
cgwalters: It's allowed to set `url` in addition; in fact I think I'm going to rework ostree-rs-ext to require that for container fetches.

Updated the test suite.

For `--contenturl`...I don't see a reason to disallow it explicitly.  Maybe it makes sense for the backend.  If it doesn't, don't set it.

lucab (approved):

PR #2435, opened 2021-09-08
repo-pull: legacy_transaction_resuming flag ignored

RBuddel:
Hello ostreedev maintainers

When trying to do a pull form a remote with the --require-static-delta flag set after a previous attempt was aborted it would always indicate that everything was already pulled. But doing an deploy of the commit afterwards would throw an error that files are missing.
Investigating the problem i found out that the legacy_transaction_resuming flag was not considered in the delta-path, so it would always mark an legacy_transaction as done/not partial, even if files were missing. Using the commitstate_is_partial function will correctly assess which deltaparts are missing and redownload the missing parts.

With regards
RBuddel


lucab:
Thanks for the patch, LGTM!

lucab (approved):

PR #2440, opened 2021-09-10
two small unit test patches

cgwalters:
tests/pull-test: Avoid duplicating test numbers

We do this in other places; avoids touching two numbers when
adding tests.  Let computers do the addition.

---

tests: Add new TAP APIs

Having to touch a global test counter when adding tests is
a recipe for conflicts between PRs.

The TAP protocol allows *ending* with the expected number of
tests, so the best way to do this is to have an explicit
API like our `tap_ok` which bumps a counter, then end with `tap_end`.

I ported one test as a demo.

---

### DIFF
--- a/tests/kolainst/libtest-core.sh
+++ b/tests/kolainst/libtest-core.sh
@@ -35,6 +35,17 @@ assert_not_reached () {
     fatal "$@"
 }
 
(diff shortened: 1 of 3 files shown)


jlebon (approved):
Cool!

PR #2447, opened 2021-09-28
repo: Add an API to init `OstreeSePolicy` from commit directly

cgwalters:
This is part of `OstreeCommitModifier`, but I'm not using
that in some of the ostree-ext Rust code.

It just makes more sense as a direct policy API, where it should
have been in the first place.  There's already support for
setting a policy object on a commit modifier, so that's all the
old API needs to do now.

### DIFF
--- a/src/libostree/ostree-sepolicy.c
+++ b/src/libostree/ostree-sepolicy.c
@@ -29,6 +29,7 @@
 #include "otutil.h"
 
 #include "ostree-sepolicy.h"
+#include "ostree-repo.h"
 #include "ostree-sepolicy-private.h"
 #include "ostree-bootloader-uboot.h"
 #include "ostree-bootloader-syslinux.h"
@@ -47,6 +48,7 @@ struct OstreeSePolicy {
   int rootfs_dfd;
   int rootfs_dfd_owned;
   GFile *path;
+  GLnxTmpDir tmpdir;
 
 #ifdef HAVE_SELINUX
   GFile *selinux_policy_root;
@@ -76,6 +78,8 @@ static void
 ostree_sepolicy_finalize (GObject *object)
 {
   OstreeSePolicy *self = OSTREE_SEPOLICY (object);
+
+  (void) glnx_tmpdir_delete (&self->tmpdir, NULL, NULL);
 
   g_clear_object (&self->path);
   if (self->rootfs_dfd_owned != -1)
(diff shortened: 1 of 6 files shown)


src/libostree/ostree-sepolicy.c
```diff
@@ -266,6 +270,50 @@ get_policy_checksum (char        **out_csum,
...
+    {
+       OstreeRepoCheckoutAtOptions coopts = {0,};
+       coopts.mode = OSTREE_REPO_CHECKOUT_MODE_USER;
+       coopts.subpath = glnx_strjoina ("/", policypath);
+     
+       if (!ostree_repo_checkout_at (repo, &coopts, tmpdir.fd, policypath, commit, cancellable, error))
+         return glnx_prefix_error (error, "policy checkout"), NULL;
+    }
+
+  return ostree_sepolicy_new_at (tmpdir.fd, cancellable, error);
```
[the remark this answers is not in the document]
cgwalters: Yeah I think you're right, I think I missed the transfer of tmpdir ownership.

src/libostree/ostree-sepolicy.c
```diff
@@ -266,6 +270,56 @@ get_policy_checksum (char        **out_csum,
 
 #endif
 
+/**
+ * ostree_sepolicy_new_from_commit:
+ * @repo: The repo
+ * @commit: SHA-256 commit checksum
```
jlebon, line 276: Not sure if this affects bindings, but this is missing `cancellable`.

src/libostree/ostree-sepolicy.c
```diff
@@ -266,6 +270,56 @@ get_policy_checksum (char        **out_csum,
...
+ *
+ * Returns: (transfer full): A new policy
+ */
+OstreeSePolicy*
+ostree_sepolicy_new_from_commit (OstreeRepo  *repo,
+                                 const char  *commit,
+                                 GCancellable *cancellable,
+                                 GError     **error)
+{
+  GLNX_AUTO_PREFIX_ERROR ("setting sepolicy from commit", error);
```
jlebon, line 292: `strjoina` hack to include commit hash?

jlebon (approved):

src/libostree/ostree-sepolicy.c
cgwalters: I've been trying to be more careful about adding checksums into error messages, because we now have several codepaths that log the checksum twice just inside ostree, and if the caller does it too the error message can become hard to read.

cgwalters:
Hooray for useful CI coverage, I had missed that we supported passing refs and not just commits into this API.
  (I should clearly have actually tested this locally too)
Anyways, should be fixed now.

lucab:
Re-surfacing a comment I made somewhere else, I think `ostree_sepolicy_get_path()` may become less than useful after this. Maybe a good candidate for deprecation?

src/libostree/ostree-sepolicy.c
```diff
@@ -266,6 +270,58 @@ get_policy_checksum (char        **out_csum,
...
+    return FALSE;
+
+  if (g_file_query_exists (policyroot, NULL))
+    {
+       OstreeRepoCheckoutAtOptions coopts = {0,};
+       coopts.mode = OSTREE_REPO_CHECKOUT_MODE_USER;
+       coopts.subpath = glnx_strjoina ("/", policypath);
+     
+       if (!ostree_repo_checkout_at (repo, &coopts, tmpdir.fd, policypath, commit, cancellable, error))
+         return glnx_prefix_error_null (error, "policy checkout"), NULL;
```
lucab, line 314: Stray trailing `NULL`, I think.

lucab (approved):

cgwalters:
> Re-surfacing a comment I made somewhere else, I think ostree_sepolicy_get_path() may become less than useful after this. 
> Maybe a good candidate for deprecation?

Sure, done

src/libostree/ostree-sepolicy.c
```diff
@@ -266,6 +270,58 @@ get_policy_checksum (char        **out_csum,
...
+OstreeSePolicy*
+ostree_sepolicy_new_from_commit (OstreeRepo  *repo,
+                                 const char  *rev,
+                                 GCancellable *cancellable,
+                                 GError     **error)
+{
+  GLNX_AUTO_PREFIX_ERROR ("setting sepolicy from commit", error);
+  g_autoptr(GFile) root = NULL;
+  g_autofree char *commit = NULL;
+  if (!ostree_repo_read_commit (repo, commit, &root, &commit, cancellable, error))
```
jlebon, line 296:
```suggestion
  if (!ostree_repo_read_commit (repo, rev, &root, &commit, cancellable, error))
```

jlebon (changes requested):

src/libostree/ostree-sepolicy.c
cgwalters: THanks!

jlebon (approved):

PR #2448, opened 2021-09-30
bin/commit: Fix --tree=tar with --selinux-policy

cgwalters:
The logic for `--selinux-policy` ended up in the `--tree=dir`
path, but there's no reason for that.  Fix the imported
labeling with `--tree=tar`.  Prep for use with containers.

### DIFF
--- a/src/ostree/ot-builtin-commit.c
+++ b/src/ostree/ot-builtin-commit.c
@@ -602,6 +602,17 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
       filter_data.skip_list = skip_list;
       modifier = ostree_repo_commit_modifier_new (flags, commit_filter,
                                                   &filter_data, NULL);
+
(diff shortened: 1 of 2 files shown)


src/ostree/ot-builtin-commit.c
```diff
@@ -686,6 +686,18 @@ ostree_builtin_commit (int argc, char **argv, OstreeCommandInvocation *invocatio
       tree_type = g_strndup (tree, eq - tree);
       tree = eq + 1;
 
+      if (first && opt_selinux_policy)
```
lucab, line 689: I think this could even go outside of the for-tree-iter body, the `first` condition is effectively constraining it to only run once.

lucab (approved):

src/ostree/ot-builtin-commit.c
cgwalters: Hmm true but it was a bit more subtle than that.  Updated :new:

PR #2449, opened 2021-09-30
lib: Add an API to construct a MutableTree from a commit

cgwalters:
lib: Add an API to construct a `MutableTree` from a commit

This is nicer than having the caller parse the commit
object, or indirect via the `OstreeRepoFile*` object of the root.

Will be used in ostree-rs-ext around tar parsing.

---


jlebon (approved):

cgwalters:
Amazing that I made effectively the same mistake in two separate PRs.

jlebon (approved):

PR #2452, opened 2021-10-01
libarchive: Always ensure we have a label

cgwalters:
The way SELinux works, there isn't a concept of "something without
a label".  There *is* the `unlabeled_t` type that the kernel makes
up when it finds no label on a mounted filesystem.  But that's still
a label, it's just a kernel fallback.

When we go to import a tar archive, in most cases it won't
have SELinux labels, and even if it does, we want to overwrite it.

Now, the concept of a "whiteout" file arose relatively recently
as a special file type that's used in the container ecosystem alongside
overlayfs.

For some reason, presumably due to its special nature, at least the
Fedora SELinux policy doesn't define a label for it.

However, when it lands on disk, the kernel will still assign it
a default label.

And that's the problem - the computed checksum won't match, which
will appear as corruption.

One doesn't see this problem when e.g. extracting a tarball
temporarily to disk, and then committing it because the act of extracting
to disk will assign a label.

This problem was hit when importing container layers for ostree-container
work in ostree-rs-ext.

### DIFF
--- a/src/libostree/ostree-sepolicy.c
+++ b/src/libostree/ostree-sepolicy.c
@@ -599,6 +599,25 @@ ostree_sepolicy_get_label (OstreeSePolicy    *self,
   return TRUE;
 }
 
+// If policy doesn't specify a label, try a fallback.
+gboolean
+_ostree_sepolicy_require_label (OstreeSePolicy *policy, const char *relpath,
+                                guint32 unix_mode, char **out_label, 
+                                GCancellable *cancellable, GError **error)
+{
+  char *label = NULL;
+  if (!ostree_sepolicy_get_label (policy, relpath, unix_mode, &label, cancellable, error))
+    return FALSE;
+  if (!label)
+    {
+      if (!ostree_sepolicy_get_label (policy, "/usr/share/some-generic-thing", unix_mode, &label, cancellable, error))
+        return FALSE;
+    }
+  *out_label = label;
+  return TRUE;
+}
+
+
 /**
  * ostree_sepolicy_restorecon:
  * @self: Self
--- a/src/libostree/ostree-repo-libarchive.c
+++ b/src/libostree/ostree-repo-libarchive.c
@@ -27,6 +27,7 @@
 #include "ostree.h"
 #include "ostree-core-private.h"
 #include "ostree-repo-private.h"
+#include "ostree-sepolicy-private.h"
 
 #ifdef HAVE_LIBARCHIVE
 #include <archive.h>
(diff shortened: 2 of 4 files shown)


travier:
Reading on whiteout files, I found (https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt) which suggests that whiteout are 0:0 char devices. Still looking into the `.wh` format.

src/libostree/ostree-sepolicy.c
```diff
@@ -599,6 +599,25 @@ ostree_sepolicy_get_label (OstreeSePolicy    *self,
...
+_ostree_sepolicy_require_label (OstreeSePolicy *policy, const char *relpath,
+                                guint32 unix_mode, char **out_label, 
+                                GCancellable *cancellable, GError **error)
+{
+  char *label = NULL;
+  if (!ostree_sepolicy_get_label (policy, relpath, unix_mode, &label, cancellable, error))
+    return FALSE;
+  if (!label)
+    {
+      if (!ostree_sepolicy_get_label (policy, "/usr/share/some-generic-thing", unix_mode, &label, cancellable, error))
```
jlebon, line 613: I think a more appropriate thing to do here would be to keep going up from `relpath` and stopping until we have a parent dir which has a label. This more closely matches the default heuristic used by SELinux.

travier:
http://aufs.sourceforge.net/aufs.html > More reading for the `.wh.....` format for filenames.

src/libostree/ostree-sepolicy.c
travier, line 613: Should we call this `/usr/share/i-will-never-exist`?

travier (approved):
Minor nit / question to make sure I understand the logic here but LGTM.

cgwalters:
After a big of digging, the cause of this is likely
https://github.com/fedora-selinux/selinux-policy/blob/1715509773b3387a0c74423c05d53eb401d9b470/policy/modules/kernel/files.fc#L321

And in practice, it is quite simply broken to have anything `/var/tmp` in ostree at all.   And more generally it looks like effectively all use of `<<none>>` in the policy is for things that shouldn't be in ostree at all.

So, on one hand the answer could be "don't do that", and that's basically what https://github.com/ostreedev/ostree-rs-ext/pull/112 is doing.

Perhaps we should instead try to drive more opinions on this stuff into the ostree core.

src/libostree/ostree-sepolicy.c
cgwalters: I'm not sure.  In the case of this `/var/tmp` file, yes.   On the other hand, if someone is trying to add `/foobar`, I don't think we want it to be `root_t`.

jlebon: I think it should be safe to trust the policy. E.g. `/foobar` would get caught by https://github.com/fedora-selinux/selinux-policy/blob/1715509773b3387a0c74423c05d53eb401d9b470/policy/modules/kernel/files.fc#L5.

cgwalters:
We can address this one in a different way.

Issue #2446, opened 2021-09-28
consider ignoring non-regular,non-symlink files in `/etc`


cgwalters:
https://bugzilla.redhat.com/show_bug.cgi?id=1945274 is an issue where a privileged kubernetes daemonset is writing a socket into `/etc`.  This makes ostree upgrades barf.

Now, they should clearly move it to `/run`.

However, one option is for us to just ignore it instead of erroring out.  Some brief investigation shows that e.g. `git add somesocket` is a silent no-op, which is an argument in favor of ignoring it.

PR #2453, opened 2021-10-01
deploy: Ignore sockets, fifos in /etc during merge

cgwalters:
https://bugzilla.redhat.com/show_bug.cgi?id=1945274 is an issue where a privileged
kubernetes daemonset is writing a socket into `/etc`.  This makes ostree upgrades barf.

Now, they should clearly move it to `/run`.  However, one option is for us to
just ignore it instead of erroring out.  Some brief investigation shows that
e.g. `git add somesocket` is a silent no-op, which is an argument in favor of ignoring it.

Closes: https://github.com/ostreedev/ostree/issues/2446

### DIFF
--- a/tests/test-admin-deploy-etcmerge-cornercases.sh
+++ b/tests/test-admin-deploy-etcmerge-cornercases.sh
@@ -51,6 +51,9 @@ chmod 700 ${etc}/a/long/dir/forking
 # Symlink to nonexistent path, to ensure we aren't walking symlinks
 ln -s no-such-file ${etc}/a/link-to-no-such-file
 
+# fifo which should be ignored
+mkfifo "${etc}/fifo-to-ignore"
+
 # Remove a directory
 rm ${etc}/testdirectory -rf
 
@@ -65,6 +68,10 @@ newetc=${newroot}/etc
 
 assert_file_has_content ${newroot}/usr/etc/NetworkManager/nm.conf "a default daemon file"
 assert_file_has_content ${newetc}/NetworkManager/nm.conf "a modified config file"
+
+if test -e "${newetc}"/fifo-to-ignore; then
+  fatal "Should not have copied fifo!"
+fi
 
 assert_file_has_mode() {
   stat -c '%a' $1 > mode.txt
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -488,9 +488,7 @@ copy_modified_config_file (int                 orig_etc_fd,
     }
   else
     {
-      return glnx_throw (error,
-                         "Unsupported non-regular/non-symlink file in /etc '%s'",
-                         path);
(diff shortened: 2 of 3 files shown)


jlebon:
Feels like we should just ignore everything that's not a file/symlink/dir to be more consistent, no?

cgwalters:
Yeah, fair.  Done.

cgwalters:
Heh, the fact that `apt install netcat` on debian barfs and says "there are both netcat-openbsd and netcat-traditional, you need to pick one" is just a very Debian thing to do.
Like I'm sure there are 2% of users who care but...

tests/test-admin-deploy-etcmerge-cornercases.sh
```diff
@@ -51,9 +51,21 @@ chmod 700 ${etc}/a/long/dir/forking
...
 
+# Socket, which should be ignored
+(cd "$etc" && setpriv --pdeathsig SIGTERM -- netcat -lU sock-to-ignore)&
+for x in $(seq 10); do
+  if test -S "${etc}/sock-to-ignore"; then
+    break
+  fi
+  sleep 0.5
+done
+trap 'kill -9 $(jobs -p) || true' EXIT ERR
```
jlebon, line 62: This will stomp over any existing traps for those signals. I think we have `libtest_exit_cmds` for this.

tests/test-admin-deploy-etcmerge-cornercases.sh
```diff
@@ -66,6 +78,12 @@ newetc=${newroot}/etc
 assert_file_has_content ${newroot}/usr/etc/NetworkManager/nm.conf "a default daemon file"
 assert_file_has_content ${newetc}/NetworkManager/nm.conf "a modified config file"
 
+if test -e "${newetc}"/sock-to-ignore; then
+  fatal "Should not have copied socket!"
+fi
+kill $(jobs -p)
```
jlebon, line 84: Maybe let's just capture the PID of the netcat when you spawn it (`$!`) and use that here and in the trap?

jlebon: Or, maybe it's simpler to just test with a FIFO instead for this test?

jlebon (approved):
Some comments, but LGTM overall!

tests/test-admin-deploy-etcmerge-cornercases.sh
cgwalters: Yeah, done - we have socket coverage with the other test.

tests/test-admin-deploy-etcmerge-cornercases.sh
cgwalters: Bash use of `trap` is full of traps

jlebon (approved):

yummypeng:
It seems that copy_dir_recurse() needs this logic too.

PR #2454, opened 2021-10-05
repo/private: allow committing/aborting through a transaction guard

lucab:
This enhances the auto-transaction logic, augmenting the scope of a
transaction guard.
It allows committing or aborting a transaction through its guard.
It also supports tracking the completion status of a transaction
guard, avoiding double commits/aborts, while retaining the auto-cleanup
logic.

Ref: https://github.com/ostreedev/ostree/issues/2413#issuecomment-930327836

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1672,14 +1672,18 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
                                  GCancellable   *cancellable,
                                  GError        **error)
 {
+  g_assert (self != NULL);
+
   guint64 reserved_bytes = 0;
 
   g_return_val_if_fail (self->in_transaction == FALSE, FALSE);
 
   g_debug ("Preparing transaction in repository %p", self);
 
-  /* Set up to abort the transaction if we return early from this function. */
-  g_autoptr(_OstreeRepoAutoTransaction) txn = self;
+  /* Set up to abort the transaction if we return early from this function.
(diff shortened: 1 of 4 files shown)


src/libostree/ostree-repo.c
```diff
@@ -711,6 +711,118 @@ ostree_repo_auto_lock_cleanup (OstreeRepoAutoLock *auto_lock)
...
+ */
+gboolean
+_ostree_repo_auto_transaction_abort (OstreeRepoAutoTransaction  *txn,
+                                     GCancellable               *cancellable,
+                                     GError                    **error)
+{
+  g_assert (txn != NULL);
+
+  if (txn->repo == NULL) {
+    return glnx_throw (error, "transaction already completed");
```
lucab, line 760: I was unsure about what to do in this case (and on commit too), but I eventually settled for what seemed to me the less-surprising approach.
The Rust logic does not have this problem as the guard itself is consumed on commit/abort.

src/libostree/ostree-repo-commit.c
```diff
@@ -1736,7 +1740,7 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
     return FALSE;
 
   /* Success: do not abort the transaction when returning. */
-  txn = NULL; (void) txn;
+  txn->repo = NULL; (void) txn;
```
cgwalters, line 1743: Minor nit, I think the `(void) txn;` is now unnecessary because we're performing an active assignment from the PoV of the compiler.

cgwalters (approved):

PR #2476, opened 2021-11-02
lib: Fix a bad call to g_file_get_child

valentindavid:
In Glib, since commit 3a6e8bc8876e149c36b6b14c6a25a718edb581ed,
`g_file_get_child` does not accept absolute path as paramater anymore.

The broken assertion was encountered during `ostree admin deploy`
command for the checkout of subpath `etc`.

Example of error log:
```
(ostree admin deploy:1640): GLib-GIO-CRITICAL **: 03:42:00.570: g_file_get_child: assertion '!g_path_is_absolute (name)' failed

(ostree admin deploy:1640): GLib-GIO-CRITICAL **: 03:42:00.570: g_file_query_info: assertion 'G_IS_FILE (file)' failed
**
OSTree:ERROR:src/ostree/ot-main.c:232:ostree_run: assertion failed: (success || error)
Bail out! OSTree:ERROR:src/ostree/ot-main.c:232:ostree_run: assertion failed: (success || error)
```

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -1389,7 +1389,7 @@ ostree_repo_checkout_at (OstreeRepo                        *self,
   g_autoptr(GFile) target_dir = NULL;
 
   if (strcmp (options->subpath, "/") != 0)
-    target_dir = g_file_get_child (commit_root, options->subpath);
+    target_dir = g_file_resolve_relative_path (commit_root, options->subpath);
   else
     target_dir = g_object_ref (commit_root);
   g_autoptr(GFileInfo) target_info =


src/libostree/ostree-repo-checkout.c
```diff
@@ -1388,10 +1388,18 @@ ostree_repo_checkout_at (OstreeRepo                        *self,
...
-    target_dir = g_file_get_child (commit_root, options->subpath);
-  else
+  if (strcmp (options->subpath, "/") != 0) {
+    if (options->subpath[0] == '/') {
+      g_autoptr(GFile) abs_root_dir = g_file_new_for_path ("/");
+      g_autoptr(GFile) subpath = g_file_new_for_path (options->subpath);
+      g_autofree char* rel_subpath = g_file_get_relative_path (abs_root_dir, subpath);
+      target_dir = g_file_get_child (commit_root, rel_subpath);
+    } else {
+      target_dir = g_file_get_child (commit_root, options->subpath);
```
mcatanzaro, line 1398: I don't think this is right. The second parameter to g_file_get_child() should be a basename, i.e. it should surely not contain the path separator /.

valentindavid: So I suppose we need to use `g_file_resolve_relative_path`.

valentindavid: @mcatanzaro I have changed the code. Could you tell me if I am now using correctly glib api?

mcatanzaro: Well... the new version doesn't make sense either. g_path_skip_root() assumes you are starting from an absolute path, then g_file_resolve_relative_path() converts it back to an absolute path. So that doesn't make sense, right?

valentindavid: It is not the same absolute path. We compose for example `/root/path` with `/sub/path` and we expect to get `/root/path/sub/path`.
However, I have noticed that `g_path_skip_root` returns NULL on relative path and I believe we need to accept `sub/path` as subpath to behave like `/sub/path`.

mcatanzaro: > It is not the same absolute path. We compose for example /root/path with /sub/path and we expect to get /root/path/sub/path.

Ack!

valentindavid: I have looked more. The commit root path is a ostree specific implementation of GFile. And it is considered as the root. So passing an absolute or relative path as parameter will resolve to the same. So we should just switch `g_file_get_child` to `g_file_resolve_relative_path` without any other changes.

cgwalters:
I think that's a flake, not seeing the real error though.
/override Ubuntu Latest Release
Missing `capsh` looks like, will investigate

cgwalters (approved):

PR #2479, opened 2021-11-09
ci: Require `libcap2-bin` for `capsh`

cgwalters:
This was previously pulled in indirectly, but it looks like we need
to require it explicitly in newer Ubuntu.

### DIFF
--- a/ci/gh-install.sh
+++ b/ci/gh-install.sh
@@ -84,6 +84,7 @@ case "$ID" in
             libsoup2.4-dev
             libsystemd-dev
             libtool
+            libcap2-bin
             procps
             python3
             python3-yaml


lucab (approved):

cgwalters:
OK that's one problem, but the real one seems to be:
>  ERROR: tests/test-sysroot-c - missing test plan

Which seems like something going wrong in the glib TAP integration talking to automake?  `ubuntu:latest` has `2.68.4-1ubuntu1`.  Which actually matches Fedora 34 too...and both have automake 1.16.  Digging more...

cgwalters:
Going to ship this since it fixes at least one bug, will try to figure out the TAP failure separately.

PR #2481, opened 2021-11-12
lib: Avoid dereferencing NULL error values

refi64:
Otherwise, this will segfault when callers don't need any exact errors.

Signed-off-by: Ryan Gonzalez <ryan.gonzalez@collabora.com>

### DIFF
--- a/src/libostree/ostree-sign-ed25519.c
+++ b/src/libostree/ostree-sign-ed25519.c
@@ -487,12 +487,16 @@ _load_pk_from_stream (OstreeSign *self,
   while (TRUE)
     {
       gsize len = 0;
-      g_autofree char *line = g_data_input_stream_read_line (key_data_in, &len, NULL, error);
       g_autoptr (GVariant) pk = NULL;
       gboolean added = FALSE;
+      g_autoptr(GError) local_error = NULL;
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-sign-ed25519.c
```diff
@@ -487,12 +487,16 @@ _load_pk_from_stream (OstreeSign *self,
...
       gsize len = 0;
-      g_autofree char *line = g_data_input_stream_read_line (key_data_in, &len, NULL, error);
+      g_autoptr(GError) local_error = NULL;
+      g_autofree char *line = g_data_input_stream_read_line (key_data_in, &len, NULL, &local_error);
       g_autoptr (GVariant) pk = NULL;
       gboolean added = FALSE;
 
-      if (*error != NULL)
-        return FALSE;
+      if (local_error != NULL)
```
lucab, line 495: Minor: the ordering of this feels slightly weird, I'd suggest to reorder variable declarations so that this error-check comes right after the corresponding `g_data_input_stream_read_line` call.

lucab (commented):
Good catch! Fixes look sane to me, I just have a minor style comment.

src/libostree/ostree-sign-ed25519.c
refi64: Done!

lucab (approved):

Issue #2485, opened 2021-11-19
/run/ostree-booted present in initramfs after ostree-prepare-root?


kjbracey-arm:
Some of my initramfs scripting that was previously working has stopped working, and I'm not totally sure where. But I'd like to understand the intent here.

We've included a full ostree installation in an initramfs to perform various integrity checks, and those run after the `ostree-prepare-root /rootfs`, but before moving the mounts. (So that they can locate the current deployment from `ostree admin`, but we don't want to inadvertently start using anything from the rootfs yet).

`ostree-prepare-root` is creating `/run/ostree-booted`.

This then means that following commands including `ostree fsck` and `ostree diff` fail, because they see that we're "ostree booted", try to mess with the non-existent `/sysroot`, and we get `error: Remounting /sysroot read-write: No such file or directory`.

`ostree admin --sysroot=/rootfs --print-current-dir` works, because the `--sysroot` option suppresses the "ostree booted" flag, but you can't do that for the non-admin commands.

Shouldn't `ostree-prepare-root` be creating `/rootfs/run/ostree-booted`, if anything? Surely we don't want that created in the current initramfs layout prior to remounting? If it is being set for a good reason, maybe `ostree` commands need to be more selective about when to remount `/sysroot`, or more tolerant of failure?

This scripting was working - I'm guessing an ostree update changed something in the way `/run/ostree-booted` was created. This is now at 2021.1, and I believe it was previously at 2020.7. Could maybe have been some other environment change.

Going to attempt to work around it by (temporarily?) `rm`ing `/run/ostree-booted` from the initramfs.

jlebon:
Yeah, really we shouldn't be trying to remount any repo at all here. Fix in https://github.com/ostreedev/ostree/pull/2486. I've manually verified that this fixes running `ostree fsck` from the initramfs.

kjbracey-arm:
Thanks! My `rm /run/ostree-booted` from the initramfs seems to be a functioning workaround for 2021.1 - nothing else seems to care about that prior to the remount.

Although maybe I should have been more careful - it seems that in my environment the initramfs `/run` **does** actually move to become the sysroot `/run`, but then there's also a separate thing in sysroot init scripts to recreate the `/run/ostree-booted` anyway, so my deletion doesn't seem to break anything...

PR #2486, opened 2021-11-19
app: Only remount /sysroot if needed

jlebon:
We should only try to remount `/sysroot` if we're actually handling the
sysroot repo and the repo isn't writable. We already have public APIs to
check each of those, so let's use them.

Closes: #2485

### DIFF
--- a/src/ostree/ot-main.c
+++ b/src/ostree/ot-main.c
@@ -238,7 +238,7 @@ ostree_run (int    argc,
   return 0;
 }
 
-/* Process a --repo arg; used below, and for the remote builtins */
+/* Process a --repo arg. */
 static OstreeRepo *
 parse_repo_option (GOptionContext *context,
                    const char     *repo_path,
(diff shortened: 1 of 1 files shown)


kjbracey-arm:
I can confirm this patch fixes my script, applied to 2021.1.

cgwalters (approved):

lucab:
I think this resulted in a change of behavior for a very specific CLI usecase.
rpm-ostree CI does some hackish manipulation via `ostree --repo /ostree/foo/bar`, which apparently was relying on the previous logic here to remount sysroot as RW.
I'm tying to amend the CI hacks in https://github.com/coreos/rpm-ostree/pull/3246, but I guess it would be nice to special case `--repo /ostree/...` and `--repo /sysroot/...` so that ostree automatically remounts `/sysroot`.
Or maybe there is a smarter way to do the detection via `vstatvfs()`?

jlebon:
> I'm tying to amend the CI hacks in [coreos/rpm-ostree#3246](https://github.com/coreos/rpm-ostree/pull/3246), but I guess it would be nice to special case `--repo /ostree/...` and `--repo /sysroot/...` so that ostree automatically remounts `/sysroot`. Or maybe there is a smarter way to do the detection via `vstatvfs()`?

Hmm, maybe. I think this is somewhat specific to the rpm-ostree CI which likes to keep test repos in `/ostree/repo/` to ensure hardlinks with the sysroot repo. Maybe simplest is to have a hidden stamp file in `/run` somewhere and we just always remount rw in that case in this code? That way in the rpm-ostree CI, we'd just need a `touch /run/...` upfront.

Though if the only fallout is https://github.com/coreos/rpm-ostree/pull/3246, then I'd vote for not doing anything at all for now. :)

lucab:
Yes I haven't seen any other fallout, at least yet.
I'm not super happy with the proliferation of magic stamp files as reliable interfaces for consumers.
Thinking a bit more about my previous copmment, the statfs approach would probably be to broad.
The rpm-ostree got easily fixed already.
Overall, I agree on:
 * keeping things as is going forward
 * if we notice further breakage for similar cases, just special-casing the /ostree path-prefixes

PR #2490, opened 2021-11-25
lib: misc static analysis fixes

lucab:
This fixes a few warnings from coverity, none of which really
interesting.

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -770,8 +770,8 @@ initiate_next_curl_request (FetcherRequest *req,
     curl_easy_setopt (req->easy, CURLOPT_URL, uri);
   }
 
-  curl_easy_setopt (req->easy, CURLOPT_USERAGENT,
-                    self->custom_user_agent ?: OSTREE_FETCHER_USERAGENT_STRING);
+  (void) curl_easy_setopt (req->easy, CURLOPT_USERAGENT,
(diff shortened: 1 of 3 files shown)


jlebon (approved):

travier (approved):

jlebon:
```
[2021-11-25T15:54:03.891Z] --- FAIL: ext.ostree.destructive-rs.transactionality (607.71s)
[2021-11-25T15:54:03.891Z]         harness.go:413: TIMEOUT[10m0s
```

Ahh the recent timeout work strikes again.

If I'm following how this is all wired up, I think we can do something like

```diff
diff --git a/tests/kolainst/install-wrappers.sh b/tests/kolainst/install-wrappers.sh
index 4fbb05c6..befff487 100755
--- a/tests/kolainst/install-wrappers.sh
+++ b/tests/kolainst/install-wrappers.sh
@@ -10,6 +10,7 @@ while read line; do
   cat >${testdir}/${line} << EOF
 #!/bin/bash
 set -xeuo pipefail
+# kola: { "timeoutMin": 15 }
 dn=\$(dirname $0)
 exec \${KOLA_EXT_DATA}/ostree-test run-destructive ${line}
 EOF
```

?

Not ideal of course since it indiscriminately applies to all Rust tests, but meh. I guess a more proper approach would be to have each Rust test define its timeout and add a `get-timeout` command to `insttestmain.rs` that `install-wrappers.sh` can use?

lucab:
(I realized I went to far and one of the fixes was not technically correct, so I dropped that. The new push did not hit any timeout)

PR #2494, opened 2021-12-06
Update FSF license notices to use URL instead of address

jmarrero:
This PR should cleanup the errors related to `incorrect-fsf-address`
https://fedoraproject.org/wiki/Common_Rpmlint_issues#incorrect-fsf-address

as seen on:
https://fedora.softwarefactory-project.io/zuul/build/626111229f584791960b28198e5e2d4b

However this seems like a huge diff for this... if this does not provide any value I can just close it. I am confused with the error from the build but then the documentation says that `In all cases, upstream should be informed about this. This is the only requirement with respect to this error.` So this might not be a big deal to update the package.

Also keeping as draft until I can test if this change removes the zuul error by local testing(hopefully).


lucab:
Yes, this is bound to be a large diff but anyway it should be done at some point.

By the way, FSF now recommends pointing to https://www.gnu.org/licenses/ instead of using the office address, see https://www.gnu.org/licenses/gpl-howto.en.html.

cgwalters:
Thanks for the patch!  Agreed on using the URL.  That way a few years from now landlord raising the Boston rent for the FSF doesn't indirectly lead down the line to them moving again and then thousands more pull requests :wink:

jmarrero:
Found this nice one liner to test locally
```
grep -r -E '(675\s+Mass\s+Ave|59\s+Temple\s+Place|Franklin\s+Steet|02139|02111-1307)'
```
in https://github.com/rpm-software-management/rpmlint/issues/85

and updated the text to use the URL.

src/libostree/libostree-released.sym
```diff
@@ -14,9 +14,7 @@
   Lesser General Public License for more details.
  
   You should have received a copy of the GNU Lesser General Public
-  License along with this library; if not, write to the
```
lucab, line 17: This changes the overall digest of the file, thus I think you'll have to update `tests/test-symbols.sh` too, see
https://github.com/ostreedev/ostree/blob/a07b8d6c0076d424555c37b6fa462ce279b831c0/tests/test-symbols.sh#L58-L63

jmarrero: Thank you @lucab, fixed in latest push.

PR #2496, opened 2021-12-07
lib/repo: fix problematic invariant checks

lucab:
This tweaks a few invariant checks, turning them into safer assertions instead.
For more details on each, see individual commits.
One of them came up through casual code reading, so I did also check the rest of the file while at it.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -1599,7 +1599,8 @@ _ostree_repo_update_mtime (OstreeRepo        *self,
 GKeyFile *
 ostree_repo_get_config (OstreeRepo *self)
 {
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo.c
```diff
@@ -3814,7 +3818,8 @@ ostree_repo_equal (OstreeRepo *a,
 OstreeRepoMode
 ostree_repo_get_mode (OstreeRepo  *self)
 {
-  g_return_val_if_fail (self->inited, FALSE);
```
lucab, line 3817: I'm somehow surprised no static analyzer flagged this "implicit boolean to OstreeRepoMode coercion".

cgwalters (approved):

Issue #2480, opened 2021-11-09
Support CLI integration of ostree-rs-ext


cgwalters:
Basically what I think we should support is something like:
`/usr/lib/ostree/ostree-ext`
which if detected, `/usr/bin/ostree` automatically gains support for e.g. `ostree ex-container` mapping to `ostree-ext container`.

(Right now for rpm-ostree we are vendoring the library, so most likely we'd make it so `/usr/lib/ostree/ostree-ext` was a hardlink to `/usr/bin/rpm-ostree`, and then auto-detect via `argv[0]` that we should behave as `ostree-ext`)

In general this path leads to a more seamless UX.

lucab:
I agree this would be nice to have. Bunch of questions to push this forward:
 * How do you feel about dedicating a binary name or CLI verb to this extension mechanism (`ostree-extra` or `ostree extra`)? That way I think we can directly map a subverb to an extension binary, without clashes.
 * What's the expected arrangement for `rpm-ostree` and `ostree-rs-ext` binaries? Given that they do provide basically the same verbs, did you intend them to be mutually exclusive for installation?

cgwalters:
> How do you feel about dedicating a binary name or CLI verb to this extension mechanism (ostree-extra or ostree extra)? That way I think we can directly map a subverb to an extension binary, without clashes.

I like the git (and cargo) approach of "if git/cargo $foo is not recognized, look for `git-$foo` instead".

That said, it's not clear to me that ostree needs this level of external extensibility, versus hardcoding support for ostree-ext.  

> What's the expected arrangement for rpm-ostree and ostree-rs-ext binaries? Given that they do provide basically the same verbs, did you intend them to be mutually exclusive for installation?

We should *definitely* support people using ostree-ext independently of rpm-ostree.  But...the thing is, dealing with Rust and vendoring and crate dependencies etc., it seems by far simpler to me to have rpm-ostree vendor ostree-ext instead of shipping it independently, at least in Fedora.

This is what I was getting at with the above:

> (Right now for rpm-ostree we are vendoring the library, so most likely we'd make it so /usr/lib/ostree/ostree-ext was a hardlink to /usr/bin/rpm-ostree, and then auto-detect via argv[0] that we should behave as ostree-ext)

lucab:
Yes, I'd be happy with that lookup logic too (assuming it's sanely feasible).
[...]

PR #2500, opened 2021-12-20
main: add support for CLI extensions via external binaries

lucab:
This adds some logic to detect and dispatch unknown subcommands to
extensions available in `$PATH`. Additional commands can be
implemented by adding relevant `ostree-$verb` binaries to the system.

As an example, if a `/usr/bin/ostree-extcommand` extension is provided,
the execution of `ostree extcommand --help` will be dispatched to that
as `ostree-extcommand extcommand --help`.

Closes: https://github.com/ostreedev/ostree/issues/2480

### DIFF
--- a/tests/test-cli-extensions.sh
+++ b/tests/test-cli-extensions.sh
@@ -1,0 +1,27 @@
+#!/usr/bin/env bash
+#
+# Copyright (C) 2021 Red Hat Inc.
+# SPDX-License-Identifier: LGPL-2.0+
+
+set -euo pipefail
+
+. $(dirname $0)/libtest.sh
+
+echo '1..2'
+
+mkdir -p ./localbin
+export PATH="./localbin/:${PATH}"
+ln -s /usr/bin/env ./localbin/ostree-env
+${CMD_PREFIX} ostree env --help >out.txt
+assert_file_has_content out.txt "with an empty environment"
+rm -rf -- localbin
+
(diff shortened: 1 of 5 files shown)


tests/test-cli-extensions.sh
```diff
@@ -0,0 +1,27 @@
...
+
+. $(dirname $0)/libtest.sh
+
+echo '1..2'
+
+mkdir -p ./localbin
+export PATH="./localbin/:${PATH}"
+ln -s /usr/bin/env ./localbin/ostree-env
+${CMD_PREFIX} ostree env --help >out.txt
+assert_file_has_content out.txt "with an empty environment"
```
cgwalters, line 16: How about e.g.:
```suggestion
${CMD_PREFIX} env sometestenv=foo ostree env >out.txt
assert_file_has_content out.txt '^sometestenv=foo`
```

or so?

cgwalters (approved):
Cool, this is nice and straightforward.

cgwalters:
Eh rather than re-roll CI for a nit, just going to merge; we can do the above as a followup.

tests/test-cli-extensions.sh
lucab: I also realized there were some leftovers in the environment. https://github.com/ostreedev/ostree/pull/2501 takes care of all that.

PR #2501, opened 2021-12-21
tests/cli-extensions: tweak test logic

lucab:
This updates the test logic for CLI extensions, actually checking
for functional output from the subcommand.
It also cleans up some environmental leftover.

Ref: https://github.com/ostreedev/ostree/pull/2500


jmarrero (approved):
lgtm

cgwalters (approved):

PR #2502, opened 2021-12-21
two minor clang-analyzer fixes

cgwalters:
soup-uri: Fix clang-analyzer warning

Fixes `Argument with 'nonnull' attribute passed null`.

(In upstream libsoup this code is gone, it uses `GUri` from glib,
 which we probably could now too, but one thing at a time)

---

tests: Fix clang-analyzer not seeing through `g_error()`

Basically due to the glib structured logging rework we lost the
`noreturn` attribute on `g_error()`.
This is fixed in glib as of https://gitlab.gnome.org/GNOME/glib/-/commit/f97ff20adf4eb7b952dd83e2c13046fe9e282f50

But we might as well just throw an error here.

---


jmarrero (approved):
lgtm

dbnicholson (approved):
Looks good!

PR #2504, opened 2022-01-03
bsdiff: bump submodule, pick up fix for CVE-2014-9862

lucab:
This updates the bsdiff submodule to latest upstream revision, in
order to pick up additional bound checks for CVE-2014-9862.

Ref:
 * https://github.com/mendsley/bsdiff/pull/31
 * https://www.x41-dsec.de/lab/advisories/x41-2020-006-bspatch/


lucab:
We should really have @dependabot taking care of timely submodule updates here. I'll take care of adding that in a separate PR.

cgwalters (approved):
Thanks for finding this!

Hopefully those crash reports are from not-intentionally corrupted deltas :frowning: 

Also, clearly we and/or bsdiff upstream should be investing in fuzzing.

lucab:
https://github.com/ostreedev/ostree/pull/2506 adds the dependabot config.

Issue #2510, opened 2022-01-06
Cleanup from staged deployments


dbnicholson:
I finally got around to changing our updater to use staged deployments and one thing we lose is pruning of the rollback deployment. Since the ref isn't removed until the new deployment is finalized, the objects are still on disk until some later process prunes. Our updater runs a full cleanup after staging, so the old rollback deployment would get pruned when a new update comes in. However, that may not happen for a long time and it effectively means that you always have 3 deployments on disk.

We could solve this downstream in Endless, but fixing this is something that any user of ostree staged deployments could benefit from. My idea is to have a sysroot autocleanup mode that only runs the cleanup if a known file exists and then delete it when the cleanup completes.

For example, `/sysroot/.cleanup` is written out when the deployment is finalized. Add an API (or a cmdprivate) that runs `ostree_sysroot_cleanup` only when `/sysroot/.cleanup` exists and deletes it when done. Call this from `ostree admin cleanup --auto`. Add a systemd unit like `ostree-sysroot-auto-cleanup.service` with `ConditionPathExists=/sysroot/.cleanup` and `ExecStart=/usr/bin/ostree admin cleanup --auto`.

WDYT?

cgwalters:
Hmm.  I am not opposed to this, but so far the vision for ostree had been more of as a library.  That said this has come up a few times but it would be really nice to try to have more shared daemon code.  I've had this offhand thought that we could try to start that daemon code in Rust in ostree-rs-ext?

OTOH, we could also ship some services like this as a build-time (off by default?) option?

The problem I see here is if we suddenly start shipping more on-by-default systemd services we could be interfering with user code.

dbnicholson:
Right, there are definitely some competing interests here.

* Any updater daemon may want handle this kind of cleanup themselves and then the systemd service is going to get in the way.

* The logic to encode this state is really best to live in the staging finalization. When we were discussing this for eos-updater, @wjt suggested the stamp file has to be in `/etc` so you don't get into a situation where eos-updater created the stamp file, you have an unclean shutdown, and then the cleanup mechanism triggers on the next boot even though you didn't actually write out the deployment yet. Having it in ostree after the call to write out the deployment means it could only exist when a staged deployment has been finalized but not pruned.

* I had trouble figuring out where to actually put this in eos-updater. Startup? A random idle time? Having an independent service that handles this during boot seemed nice even though it might lock the repo and block eos-updater.

[...]

PR #2511, opened 2022-01-07
Sysroot auto cleanup

dbnicholson:
This is my proof of concept to initiate cleanup on the next boot of a staged deployment. In #2510 we discussed whether it was a good idea to add another systemd unit when the cleanup may be handled better by a management daemon like `rpm-ostree`.

I think this is a nice general purpose way to address the issue, but I definitely concede it's not the only way to handle it. I think even without the systemd unit the rest of the commits are useful. It would be possible to stash some state indicating cleanup from outside ostree, but really ostree is by far best positioned since it handles writing out the staged deployment via `ostree admin finalize-staged`.

Fixes: #2510


src/boot/ostree-auto-cleanup.service
```diff
@@ -0,0 +1,43 @@
...
+After=sysinit.target basic.target
+Conflicts=shutdown.target
+Before=shutdown.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStop=/usr/bin/ostree admin cleanup --auto
+ProtectSystem=strict
+ReadWritePaths=/sysroot /boot
```
cgwalters, line 36: Interesting, I like this.  But don't we also have to touch `/var` and/or `/etc` for the `.updated` files?  IIRC I hit that last time trying to restrict things.

src/libostree/libostree-devel.sym
```diff
@@ -20,12 +20,17 @@
    - uncomment the include in Makefile-libostree.am
 */
 
+LIBOSTREE_2022.2 {
+global:
+  ostree_sysroot_auto_cleanup;
```
cgwalters, line 25: New public API feels overkill for this if the expectation is that it will be invoked by a separate systemd unit.  Related below:

src/libostree/ostree-sysroot-cleanup.c
```diff
@@ -555,6 +555,48 @@ ostree_sysroot_cleanup (OstreeSysroot       *self,
...
+ * Since: 2022.2
+ */
+gboolean
+ostree_sysroot_auto_cleanup (OstreeSysroot       *self,
+                             gboolean            *out_cleaned,
+                             GCancellable        *cancellable,
+                             GError             **error)
+{
+  struct stat stbuf;
+  if (!glnx_fstatat_allow_noent (self->sysroot_fd, _OSTREE_SYSROOT_AUTOCLEANUP, &stbuf, AT_SYMLINK_NOFOLLOW, error))
```
cgwalters, line 580: Since the systemd unit is already doing this, do we need a new API at all versus simply `ExecStart=ostree admin cleanup`?

src/boot/ostree-auto-cleanup.service
dbnicholson: No, that happens when finalizing a deployment, not when cleaning up. See the comments in `ostree-finalize-staged.service`. I tested this in a downstream way and it worked fine.

Over in endlessm/eos-updater#298 @wjt pointed out that having `/sysroot` writable means essentially the whole system is writable. One thing that could be nicer would be to limit this whole exercise to just pruning since finalizing also does `ostree_sysroot_prepare_cleanup`. Then you could limit the writable paths to just `/sysroot/repo`.

src/libostree/libostree-devel.sym
dbnicholson: I guess this was with an eye toward having the systemd unit disabled and letting the management daemon handle it itself. But I did do this in endlessm/eos-updater#298 without needing the API. I do think it's cleaner to have the semantics contained within libostree, though.

src/libostree/ostree-sysroot-cleanup.c
dbnicholson: Maybe not. I think you could do this all in the CLI or via systemd unit settings. It felt cleaner to encapsulate it in libostree, but it's not necessary. I could go either way.

src/boot/ostree-auto-cleanup.service
```diff
@@ -0,0 +1,43 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library. If not, see <https://www.gnu.org/licenses/>.
+
+[Unit]
+Description=OSTree Automatic Sysroot Cleanup
+Documentation=man:ostree-admin-cleanup(1)
+ConditionPathExists=/sysroot/.cleanup
```
wmanley, line 19: FWIW, we do something similar, but with the condition inverted:

    # This is deleted by deploy.sh:
    ConditionPathExists=!/sysroot/.ostree-cleaned

This made migration easier. If you're upgrading ostree with the deploy then the old version of ostree won't know to create this file, so when booting into the new deployment cleaning won't run.

dbnicholson: That is a nice property. However, doesn't that mean you're running a prune on every boot, even if you haven't made a new deployment?

wmanley: No, we delete .ostree-cleaned only just before ostree admin deploy.

PR #2512, opened 2022-01-10
libotutil: avoid leaking builder memory on error

lucab:
This swaps the order of a couple of input sanity checks, in order
to fix a minor memory leak due to an early-return on the error
path.
Memory for the result is now allocated only after input has been
sanity-checked.
It fixes a static analysis warning highlighted by Coverity.


jmarrero (approved):
lgtm

PR #2513, opened 2022-01-10
ostree: check g_setenv return value

lucab:
This adds proper return-value checks on g_setenv calls.
It fixes a static analysis warning highlighted by Coverity.

### DIFF
--- a/src/libostree/ostree-sepolicy.c
+++ b/src/libostree/ostree-sepolicy.c
@@ -422,7 +422,10 @@ initable_init (GInitable     *initable,
     {
       const char *policy_rootpath = gs_file_get_path_cached (policy_root);
 
-      g_setenv ("LIBSELINUX_DISABLE_PCRE_PRECOMPILED", "1", FALSE);
+      /* TODO(lucab): get rid of this setenv(), it may be unsafe in a multi-thread context. */
+      if (!g_setenv ("LIBSELINUX_DISABLE_PCRE_PRECOMPILED", "1", FALSE))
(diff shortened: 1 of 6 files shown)


src/libostree/ostree-sepolicy.c
```diff
@@ -422,7 +422,10 @@ initable_init (GInitable     *initable,
     {
       const char *policy_rootpath = gs_file_get_path_cached (policy_root);
 
-      g_setenv ("LIBSELINUX_DISABLE_PCRE_PRECOMPILED", "1", FALSE);
+      /* TODO(lucab): get rid of this setenv(), it may be unsafe in a multi-thread context. */
+      if (!g_setenv ("LIBSELINUX_DISABLE_PCRE_PRECOMPILED", "1", FALSE))
```
lucab, line 426: @cgwalters do you maybe know/remember what is the effect of this?
It looks like it was first introduced in [this commit](https://github.com/ostreedev/ostree/commit/3337334be5c7abbb79796994523e445db53df838#diff-9c8abb06ee43e47264c447cf374c8d43b2ba9688758f8dbe8282aacda21aee8eR224), but ours is the only occurrence across all of [GitHub](https://github.com/search?q=LIBSELINUX_DISABLE_PCRE_PRECOMPILED&type=code) and [Debian](https://codesearch.debian.net/search?q=LIBSELINUX_DISABLE_PCRE_PRECOMPILED) sources.

lucab: Thanks for the hints, I will go and dig more into libselinux history.
It seems feasible to drop, but I will confirm and put that into a separate PR.

PR #2515, opened 2022-01-11
main: Also support CLI extensions in `/usr/libexec/libostree/ext`

cgwalters:
In fixing https://github.com/coreos/rpm-ostree/pull/3323
I felt that it was a bit ugly we're installing `/usr/bin/ostree-container`.

It's kind of an implementation detail.  We want users to use
`ostree container`.

Let's support values outside of $PATH too.

For example, this also ensures that TAB completion for `ost` expands
to `ostree ` with a space.

### DIFF
--- a/src/ostree/ot-main.c
+++ b/src/ostree/ot-main.c
@@ -41,6 +41,10 @@ static gboolean opt_verbose;
 static gboolean opt_version;
 static gboolean opt_print_current_dir;
 
+// TODO: make this public?  But no one sane wants to use our C headers
+// to find where to put files.  Maybe we can make it printed by the CLI?
+#define _OSTREE_EXT_DIR PKGLIBEXECDIR "/ext"
+
 static GOptionEntry global_entries[] = {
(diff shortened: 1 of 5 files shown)


jmarrero (approved):
lgtm

src/ostree/ot-main.c
```diff
@@ -41,6 +41,10 @@ static gboolean opt_verbose;
 static gboolean opt_version;
 static gboolean opt_print_current_dir;
 
+// TODO: make this public?  But no one sane wants to use our C headers
+// to find where to put files.  Maybe we can make it printed by the CLI?
+#define _OSTREE_EXT_DIR PKGLIBEXECDIR "/ext"
```
dbnicholson, line 46: `/usr/share/pkgconfig/ostree.pc` like `/usr/share/pkgconfig/systemd.pc`? Not a blocker, but I think it provides a natural place to provide info about the CLI. Another alternative is to put it in the `--version` output like `Features`.

cgwalters: Good suggestion!  Done, thanks!

PR #2520, opened 2022-01-18
builtin/commit: reject empty metadata keys/values

lucab:
This adds two more checks to the metadata parsing logic in order
to reject empty metadata keys/values upfront.


dbnicholson:
An empty value is entirely valid, but rejecting empty keys would be good. However, wouldn't the validation be better in libostree rather than in the CLI?

dbnicholson:
It looks like the rpm-ostree case is particularly about the `version` key. There it makes sense to reject empty values, but the semantics of keys in general is unknown. It would be entirely reasonable for `my-custom-key` to have a meaningful empty value.

dbnicholson:
> I'm not sure if by "There" you meant doing it in rpm-ostree, or just speaking about the `version` key. Since the `version` metadata key is also meaningful to OSTree, I'm assuming you meant the latter. Or do you see a use case for empty version strings?

Yeah, I meant `version` by "there". I don't see allowing `version` to be empty being useful and I think it would be reasonable to add a special case in the metadata verification at commit time that it's value (and maybe other known `OSTREE_COMMIT_META_KEY_*` values) is not empty. You just can't generally say that empty string values are invalid for arbitrary metadata keys, IMO.

tests/test-basic-user-only.sh
```diff
@@ -54,6 +54,16 @@ fi
 assert_file_has_content err.txt "Content object.*invalid mode.*with bits 040.*"
 echo "ok failed to commit suid"
 
+cd ${test_tmpdir}
+rm repo-input -rf
+ostree_repo_init repo-input init --mode=archive
+rm files -rf && mkdir files
+if $CMD_PREFIX ostree --repo=repo-input commit -b willfail --tree=dir=files --add-metadata-string='=FOO'; then
+    assert_not_reached "committed an empty metadata key"
+fi
```
jlebon, line 63: Minor/optional: I like to also verify the error message is what we expect even if it makes the test slightly more implementation-dependent. This way you're sure that you're testing the checks you think you are and won't unknowingly regress if it turns out you weren't.

tests/test-basic-user-only.sh
```diff
@@ -54,6 +54,16 @@ fi
...
+$CMD_PREFIX ostree --repo=repo-input commit -b willfail --tree=dir=files --add-metadata-string='FOO='
```
jlebon, line 64: Minor: committing this to `willfail` might be slightly misleading.

jlebon (approved):
> > I'm not sure if by "There" you meant doing it in rpm-ostree, or just speaking about the `version` key. Since the `version` metadata key is also meaningful to OSTree, I'm assuming you meant the latter. Or do you see a use case for empty version strings?
> 
> Yeah, I meant `version` by "there". I don't see allowing `version` to be empty being useful and I think it would be reasonable to add a special case in the metadata verification at commit time that it's value (and maybe other known `OSTREE_COMMIT_META_KEY_*` values) is not empty. You just can't generally say that empty string values are invalid for arbitrary metadata keys, IMO.

Right OK, yup. I agree with that!

@lucab Do you want to try to tackle that part in this PR or should we leave it as a follow-up?

cgwalters:
For the record I agree with Dan here - I think if we could go back in time, it would make sense to require non-empty values.  But I think we're also all agreed we cannot start enforcing that constraint now.

Empty metadata keys are almost certainly a mistake.

There is a subtle but important issue in this - in an ideal world, we would reject this mostly at *build* time but not client side.  IOW, let's say in a scenario where an OS builder was (due to a bug) writing an empty metadata value, it'd be nice to still support pulling and deploying older commits until they can have the build side fixed.

We don't today though really distinguish in the code between "build" vs "client" side.  Doing so would be interesting.

lucab:
Thanks all for the feedback! Yes I agree I was too aggressive/optimistic.
There are legit cases for empty metadata string values, and the validation logic is shared and entangled across writing/pulling/fscking flows. That makes it very unwieldy to retro-add this kind of validation (e.g. on `version`) without invasive rework and a good chance of breaking compatibility.

@jlebon addressed the minor remarks, I won't add more stuff around here. From my side, it can be merged.

cgwalters:
Are you planning a followup which rejects an empty value for the `version` key?

lucab:
@cgwalters if you feel that it would be ok (compatibility-wise) to have that check in this same common validation logic, yes I can do that. I'm personally not fully sure anymore. I don't want to increase code complexity around this, and I'm not in a good position to judge whether "existing repos/commits with an empty `version`" can be considered broken.
From your last comment, are you perhaps in favor of leaving the generic case as is, but validating `version` only specially?

cgwalters:
I'm weakly in favor of validating `version` specifically.  But it's not the most critical thing going on right now either, so we can just leave this as is too.

dbnicholson:
> There is a subtle but important issue in this - in an ideal world, we would reject this mostly at _build_ time but not client side. IOW, let's say in a scenario where an OS builder was (due to a bug) writing an empty metadata value, it'd be nice to still support pulling and deploying older commits until they can have the build side fixed.
> 
> We don't today though really distinguish in the code between "build" vs "client" side. Doing so would be interesting.

Can't you distinguish the build side by putting the validation in `ostree_repo_write_commit_with_time`? I know there are some occasions when you would generate a new commit on the client side, but the vast majority of commit generation is going to be on the builder.

PR #2521, opened 2022-01-20
deploy: Also log to journal if we time out global sync()

cgwalters:
We do implicitly have this data because we log timings via structured
metadata in a later journal entry, but it's quite common to lose
the structured metadata because a lot of tooling just grabs the default
syslog-compatible text from `journalctl`.

Let's be louder when we hit this case as a general rule too; I think
most people shipping ostree systems want to see if it's happening.


jmarrero (approved):
:shipit:

PR #2523, opened 2022-01-21
builtin: use GCancellable and GError everywhere

lucab:
This reworks `ostree ls` top-level logic so that cancellation
tokens and error details are plumbed through all codepaths.
It also gets rid of all previous goto jumps.


jmarrero (approved):
lgtm

PR #2524, opened 2022-01-21
lib/deploy: When deleting staged deployment, delete any lock

jlebon:
Otherwise, any future staged deployment will also automatically be
locked even if not requested. Likely we should fold the locking into the
primary `staged-deployment` serialized GVariant instead.


jlebon:
This is related tohttps://discussion.fedoraproject.org/t/34925, where a user was unable to manually override Zincati, even if they did `rpm-ostree cleanup -p` first and then upgraded.

cgwalters (approved):

jlebon:
Added a test for this, but had to also fix symlinks in the external tests to avoid symlink loops and breaking kola. I have no idea why CI here isn't hitting that issue.

cgwalters:
Unit tests are broken by the symlink change.

lucab (approved):

PR #2526, opened 2022-01-24
lib/repo: open file only if required

lucab:
This tightens up the logic for opening a file while inspecting its
xattrs. The only codepath fetching xattrs from a FD is the one
handling 'bare' mode.
It also rearranges the else-assert flow, mostly for future-proofing.


cgwalters (approved):

cgwalters:
CI failure looks like a flake?  Not seeing real errors in `test-suite.log`.

lucab:
I also didn't manage to locate any real error in the logs. Re-triggered, it passed.

PR #2528, opened 2022-01-27
tests: Stop using inventory crate

cgwalters:
tests: Stop using inventory crate

I was reading this thread
https://internals.rust-lang.org/t/from-life-before-main-to-common-life-in-main/16006/30
and that reminded me about this code, which it turns out actually
doesn't compile with my default local cargo config:
```
$ cat ~/.cargo/config
[target.x86_64-unknown-linux-gnu]
rustflags = ["-Ctarget-cpu=native", "-C", "link-arg=-fuse-ld=lld"]

[profile.release]
incremental = true
$ cargo b
...
error: linking with `cc` failed: exit status: 1
  |
  = note: "cc" "-m64" "/var/srv/walters/src/github/ostreedev/ostree/target/debug/deps/ostree_test-4ca8e730f9dc6ffc.10325uqlhkyr5uol.rcgu.o" "/var/srv/walte"
  = note: ld.lld: error: undefined symbol: __start_linkme_NONDESTRUCTIVE_TESTS
          >>> referenced by 22nn09lfsklfqvyy
          >>>               /var/srv/walters/src/github/ostreedev/ostree/target/debug/deps/ostree_test-4ca8e730f9dc6ffc.22nn09lfsklfqvyy.rcgu.o:(ostree_tes)

```

For now let's just go back to having a static list of functions.
We don't have *too* many of those.

---

Update to nix 0.23

Part of general crate updates.

---

Update to ostree-ext 0.6

Part of general crate updates.

---

Update to rand 0.8

Part of general crate updates.

---

Update to sh-inline 0.2

Syncing up with the latest.

---


cgwalters:
OK, rebased :surfer: and lifting draft on this!

jlebon (approved):

PR #2529, opened 2022-01-28
libotutil: Avoid infinite recursion during error unwinding

smcv:
When we clean up from an error, for example copy_file_range() failing
while we generate a static delta (perhaps caused by
https://gitlab.gnome.org/GNOME/libglnx/-/issues/3 or by a
genuine write error), we might free a variant builder that has a
non-null parent. Previously, this caused infinite recursion and a stack
overflow, repeatedly freeing the same object, but Luca Bruno suggested
that the intention here appears to have been to free the parent object.

Partially resolves https://github.com/ostreedev/ostree/issues/2525
(the other bug reported in that issue needs to be resolved by updating
libglnx to a version where libglnx#3 has been fixed).

cc @alexlarsson @lucab @oleid 

---

This resolves half of #2525 (#2530 is the other half).

### DIFF
--- a/src/libotutil/ot-variant-builder.c
+++ b/src/libotutil/ot-variant-builder.c
@@ -830,7 +830,7 @@ static void
 ot_variant_builder_info_free (OtVariantBuilderInfo *info)
 {
   if (info->parent)
-    ot_variant_builder_info_free (info);
+    ot_variant_builder_info_free (info->parent);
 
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Thanks!

cgwalters:

I looked at the logs and the test looked fine.  Not sure what happened there.

Issue #2499, opened 2021-12-15
support ostree-ext tar mode


cgwalters:
Today https://github.com/ostreedev/ostree-rs-ext/ implements effectively a new repository mode.  The main thing is that we split off xattrs into a separate content-addressed file; partly because I was uncertain about xattrs making it through tar transformations. 

We should support all ostree operations here, both reads and commits.  This will involve some code duplication with ostree-rs-ext unfortunately due to the Rust/C problem.  Or...*in theory* we could try to expose a hidden but stable C API from ostree-rs-ext as a C shared library...but, eek.

lucab:
Perhaps without going into the uncharted territory of linking the C and Rust codebases together, for the moment it could be enough to investigate a "metadata read-only" mode purely in C here.
As an intermediate step, that would allow basic handling/introspection through libostree and most of the metadata command to work on such repo (ls, log, show, fsck, etc). Writing and checkout would still live on the Rust extension only, at least for the moment.

cgwalters:
Having it be read-only would definitely be a useful start.

PR #2532, opened 2022-01-31
lib/core: introduce 'bare-split-xattrs' mode

lucab:
This adds read-only support for a new `bare-split-xattrs` mode.

It should contain all the necessary logic to inspect and check a
repository, but without support for committing/pulling.
This is primarily meant to unlock further progress outside of the
C side of libostree.

Remaining work is to implement the read-write flow, which can be
done separately. The idea is to hopefully prototype/implement
that in in Rust first, and later re-consider what to do. 
Here we put a `OSTREE_EXP_WRITE_BARE_SPLIT_XATTRS`
env gate in place.

Ref: https://github.com/ostreedev/ostree/issues/2499


src/libostree/ostree-repo.c
```diff
@@ -4171,6 +4176,41 @@ repo_load_file_archive (OstreeRepo *self,
...
+static GVariant *
+_ostree_repo_read_detached_xattrs (OstreeRepo   *self,
+                                   const char   *checksum,
+                                   GCancellable *cancellable,
+                                   GError      **error)
+{
+  g_assert (self != NULL);
+
+  g_autofree char *xattr_path = NULL;
+  xattr_path = g_strdup_printf ("objects/%c%c/%s.file.xattrs",
```
cgwalters, line 4188: Two minor nits; why not do declare-and-initialize i.e.:
`g_autofree char *xattr_path = g_strdup_printf (...)`?

But second, we could also consider a stack allocation here.  At one point I was trying to reduce the number of mallocs we do in potentially hot paths of reading lots of files.  I don't have a strong opinion on that though.  We can come back to it later.

cgwalters (commented):
Looking good so far!  Thanks so much for picking this up!

src/libostree/ostree-repo.c
lucab: Agreed. I do plan to eventually fold this back into the existing `_ostree_loose_path()` logic, which does not involve the heap.

However I first wanted to ask whether the double `.file.xattrs` extension here happened on purpose or by mistake?
It is the only object-type with a double suffix, and I can foresee that may cause future headaches.

It gets suffixed [here in ostree-rs-ext](https://github.com/ostreedev/ostree-rs-ext/blob/d1466efcb949e1359d377e8d3fca11f3913d61e4/lib/src/tar/export.rs#L306), which made me wonder whether it was actually intentional.
I'd be inclined to turn that into a single `.file-xattrs` extension (or similar) instead, as part of the v1 format.

cgwalters: Single extension makes sense to me.

cgwalters: We can do that change in format version 1

lucab:
I think I can checkpoint the PR at this point. It should contain all the necessary pieces for read-only access to a `bare-split-xattrs` repo, with basic docs and tests.

If we merge this, we should be able to continue prototyping writing logic on the Rust side.
If at some point we want to implement the writing logic here in C, there is an experimental env flag in place to iteratively do that.

src/libostree/ostree-core.c
```diff
@@ -1228,6 +1229,10 @@ ostree_object_type_to_string (OstreeObjectType objtype)
       return "commitmeta";
     case OSTREE_OBJECT_TYPE_PAYLOAD_LINK:
       return "payload-link";
+    case OSTREE_OBJECT_TYPE_FILE_XATTRS:
+      return "file-xattrs-link";
```
jlebon, line 1233:
```suggestion
      return "file-xattrs";
```

tests/test-basic-bare-split-xattrs.sh
```diff
@@ -0,0 +1,47 @@
...
+fi
+${OSTREE} fsck --all
+echo "ok commit not implemented"
+rm -rf -- repo files
+
+cd ${test_tmpdir}
+mkdir -p "${test_tmpdir}/files"
+touch files/foo
+${OSTREE} init --mode "${mode}"
+OSTREE_EXP_WRITE_BARE_SPLIT_XATTRS=true ${OSTREE} commit --orphan -m "experimental" files
```
jlebon, line 42: Until we have full write support, would it make sense to add a tiny xz-compressed bare-split-xattrs repo test fixture to test read support with?

src/libostree/ostree-core.c
lucab: Ouch!

jlebon (approved):
LGTM!

tests/test-basic-bare-split-xattrs.sh
cgwalters: I think I'd vote for using `quay.io/coreos-assembler/fcos:stable` as a fixture, injecting our binary into that and running tests there.

cgwalters (approved):

cgwalters:
Looks like CI fixes are needed:
```
[](https://jenkins-coreos-ci.apps.ocp.ci.centos.org/blue/organizations/jenkins/github-ci%2Fostreedev%2Fostree/detail/PR-2532/13/pipeline#step-93-log-308)[](https://jenkins-coreos-ci.apps.ocp.ci.centos.org/blue/organizations/jenkins/github-ci%2Fostreedev%2Fostree/detail/PR-2532/13/pipeline#step-93-log-309)[2022-03-02T16:49:18.798Z] ERROR: tests/test-basic-bare-split-xattrs.sh - missing test plan

[2022-03-02T16:49:18.798Z] ERROR: tests/test-basic-bare-split-xattrs.sh - exited with status 1```

tests/test-basic-bare-split-xattrs.sh
lucab: Yes it would be good to test against the actual FCOS content in the long run.
We need to switch the pipeline to emit format v1 artifacts first though.

PR #2533, opened 2022-02-01
lib/commit: clean up assertions

lucab:
This aligns all the assertion in the module. In particular, it gets
rid of all `g_return_val_if_fail` instances which may fail without
properly setting GError to the caller.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -912,8 +912,9 @@ write_content_object (OstreeRepo         *self,
                       GCancellable       *cancellable,
                       GError            **error)
 {
+  g_assert (expected_checksum != NULL || out_csum != NULL);
+
   GLNX_AUTO_PREFIX_ERROR ("Writing content object", error);
-  g_return_val_if_fail (expected_checksum || out_csum, FALSE);
 
   if (g_cancellable_set_error_if_cancelled (cancellable, error))
     return FALSE;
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-commit.c
```diff
@@ -912,8 +912,11 @@ write_content_object (OstreeRepo         *self,
                       GCancellable       *cancellable,
                       GError            **error)
 {
+  g_assert (self != NULL);
+  g_assert (OSTREE_IS_REPO (self));
```
jlebon, line 916: Feels like we should keep these checks only at API entrypoints and otherwise assume it is correct. WDYT?

src/libostree/ostree-repo-commit.c
```diff
@@ -2160,9 +2177,13 @@ ostree_repo_transaction_set_collection_ref (OstreeRepo                *self,
...
-  g_return_if_fail (OSTREE_IS_REPO (self));
-  g_return_if_fail (self->in_transaction == TRUE);
-  g_return_if_fail (ref != NULL);
+  g_assert (self != NULL);
+  g_assert (OSTREE_IS_REPO (self));
+  g_assert (self->in_transaction == TRUE);
+  g_assert (ref != NULL);
+
+  // TODO(lucab): introduce a method with error-returning in order to deprecate
+  // this one, because it can silently fail.
```
jlebon, line 2186: Is there a reason you didn't want to migrate it to `g_assert`?

jlebon (approved):
Cool! Some minor comments, but LGTM overall.

src/libostree/ostree-repo-commit.c
lucab: Yes, I can drop a few of those on the internal-only functions then.

src/libostree/ostree-repo-commit.c
lucab: This one below is not a programming error, but it triggers depending on the given `checksum` input.
So this should be a normal runtime error and signal back the validation failure, but the current signature cannot accommodate that.

Under the current API I don't see a clear winning move out of this problematic situation, so I opted for leaving it untouched.

src/libostree/ostree-repo-commit.c
lucab: Done.

jlebon (approved):

PR #2534, opened 2022-02-03
Add COPR integration Makefile

jlebon:
I'd like to enable auto-builds of this repo to
[copr.fedorainfracloud.org/coprs/g/CoreOS/continuous](https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/) so it could
eventually feed into
https://github.com/coreos/fedora-coreos-tracker/issues/910.

### DIFF
--- a/.copr/Makefile
+++ b/.copr/Makefile
@@ -1,0 +1,8 @@
+srpm:
+	dnf install -y git
+	ci/make-git-snapshot.sh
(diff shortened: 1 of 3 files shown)


cgwalters (approved):

jlebon:
Some test timeouts; we recently bumped those, so should get better. This can't affect CI here anyway.

PR #2535, opened 2022-02-04
lib/repo: Add commit version metadata to summary metadata

dbnicholson:
The commit metadata `version` key is well established but getting it for
a remote commit is cumbersome since the commit object needs to be
fetched and loaded. Including it in the summary additional metadata
allows a much more convenient view of what each of the remote refs
represents.

I was working on our release process at Endless and thought it would be so much easier to query the current OS version. I think the `version` metadata is well established enough to warrant inclusion in the summary.


cgwalters (approved):

dbnicholson:
Me thinks this is the same CI timeout issue that's been happening lately. Certainly would be unaffected by these changes.

Issue #2482, opened 2021-11-15
Add `ostree prune --commit-only`


jlebon:
On large OSTree repos, it's expensive to compute reachability. If we want to prune things one bit at a time to be explicit by invoking `ostree prune` multiple times, this can quickly add up.

This is a proposal to add a `ostree prune --commit-only` option where it would only delete the selected commit objects and do nothing else. Then we could have multiple cheap `ostree prune --commit-only` calls and only a single expensive `ostree prune` invocation which calculates full reachability once to shake off all orphan objects.

See also: https://github.com/coreos/fedora-coreos-releng-automation/pull/79#discussion_r735774851

lucab:
For clarity, how will this new flag relate to the existing `ostree prune --delete-commit <COMMIT>`?

dustymabe:
I think `--delete-commit` is when you know what commit you want to delete. This RFE is requesting that when we run a prune that spans many commits (i.e. delete all commits in the history of `$ref` that are older than `X` days) that it just deletes the commit and leaves the content objects. Then a later prune operation can sweep the repo and delete all objects that aren't reachable.

Issue #2338, opened 2021-04-14
Prune with depth != -1 can delete reachable commits


alexlarsson:
Consider a repo with 4 commits (1-4) and 2 branches a pointing at 1 and b pointing at 2. Like so:
```
 a -> 1 -> 2 -> 3 -> 4
       b --^
```
When pruning to depth 2, ostree_repo_pull() first traverses from a == 1 with depth 2, so it adds 1-3 to reachable. Then it traverses from b with depth 2, and ideally it should add 2-4 to reachable.

However, what really happens is that `ostree_repo_traverse_commit_union_with_parents()` terminates the loop in the first iteration (at b == 2) because 2 is already in `inout_reachable`, so 4 is never reached to be added to the reachable set.

If maxdepth is not set then we're ok because if some commit is in reachable, then everything reachable from it is already in the set.

So, either we need to track the depth we reached from each commit in the reachable set, or we don't short circuit the iteration when depth != -1.

cgwalters:
Good catch.  Did you find this via code inspection or did you actually hit it?

Avoiding short-circuit seems simplest.

alexlarsson:
I was re-implementing prune in flatpak for performance reasons, and saw it while reading the code.

cgwalters:
There are definitely a bunch of optimizations we could make for prune.  The problem domain is quite analogous to language runtime garbage collection.

Caching reachable objects makes sense.  Sharding seems doable too (e.g. have multiple repos split by commit hash prefix or ref name).  Another technique could be to each time we find an object is reachable, we bump its mtime or so.  Then we can avoid traversing e.g. dirtree objects that have been bumped recently in a "fast prune" operation.

alexlarsson:
Here is the output running on a clone of the main repo:

```
Pruning old commits
Finding reachable objects (depth=10)
Elapsed time: 80576.1 sec
Listing all objects
Elapsed time: 514.5 sec
Computing unreachable objects
Elapsed time: 3902.6 sec
Total objects: 16866210
Deleted 5208581 objects, 1.2 TB freed
```
Note: It didn't actually delete the objects it was just a dry-run.

alexlarsson:
And yeah, its in archive mode.

wmanley:
> Enumerating all reachable objects in the flathub repo took 22h. I made a version of prune where it stored all the reachable objects from each commit in a `.commitmeta2` file (like the other commit meta, but clients don't download it). This took the enumeration down to 39 minutes.

[...]

Issue #2021, opened 2020-03-09
Idea for more efficient prune operation


alexlarsson:
Pruning large repos is quite an expensive operation because it requires mark-and-sweeping over the entire reachable set of objects, which is quite a lot of i/o. We need to do this because we have no tracking of the reverse-mapping from object to what commits it is part of. Tracking that would be hard, because it is very dynamic.

However, there are some things that are static that we could easily track which would make this more efficient. For example, we could easily cache the set of all objects a commit references, which would make the mark phase require a lot less i/o. In fact, if this was a separate cache file all we would need to do is:
1) Read refs files
2) Read commit files finding parents (to required depth)
3) Read cache file
4) Do full travese for each commit not in the cache
5) List all objects
6) Delete unreached objects
7) Update cache

In the common case where there is very few new commits (compared to the size of the repo) we basically completely remove i/o to non-commit objects. Additionally, we could generate the traversals for the cache ahead of time to make the actual locking of the repo for the prune shorter.

alexlarsson:
Flatpak now does this and various other optimization: https://github.com/flatpak/flatpak/pull/4231
We really need this for the flathub repos which are huuuge.

PR #2536, opened 2022-02-08
src/ostree: Add --commit-only option to ostree prune

saqibali-2k:
Recently we have noticed exceedingly long execution times
for multiple invocations of `ostree prune`. This is a result of
calculating full reachability on each invocation.

The --commit-only flag provides an alternative strategy. It will only
traverse and delete commit objects to avoid the more expensive
reachability calculations. This allows us to chain multiple `--commit-only`
commands cheaply, and then follow with a more expensive `ostree prune`
invocation at the end to clean up orphaned meta and content objects.

closes: https://github.com/ostreedev/ostree/issues/2482


src/ostree/ot-builtin-prune.c
```diff
@@ -343,5 +377,8 @@ ostree_builtin_prune (int argc, char **argv, OstreeCommandInvocation *invocation
     g_print ("Deleted %u objects, %s freed\n",
              n_objects_pruned, formatted_freed_size);
 
+  if (opt_commit_only) {
+    g_print("NOTE: this only includes commit objects\n");
+  }
```
cgwalters, line 382: Mmmm...a bit of a bikeshed but I don't think we need this print.  The caller explicitly asked for this via `--commit-only`, it's a bit like ordering a sandwich and then the person giving it to you says "this is a sandwich" :wink:

src/libostree/ostree-repo-traverse.c
```diff
@@ -660,6 +680,33 @@ ostree_repo_traverse_commit_union (OstreeRepo      *repo,
...
+ * @maxdepth: Traverse this many parent commits, -1 for unlimited
+ * @inout_reachable: Set of reachable objects
+ * @cancellable: Cancellable
+ * @error: Error
+ *
+ * Update the set @inout_reachable containing all COMMIT objects reachable
+ * from @commit_checksum, traversing @maxdepth parent commits.
+ */
+gboolean
+ostree_repo_traverse_commit_only_union (OstreeRepo      *repo,
```
cgwalters, line 696: I think what we should do instead is have
`ostree_repo_traverse_commit_with_flags` that accepts `OstreeRepoCommitTraverseFlags` as a parameter too.

Another way to say this is just to make `ostree_repo_traverse_commit_internal()` the public API.

To restate another way, this is I think fixing a historical bug that this API didn't take flags; it's only the explicit external iteration API (`ostree_repo_commit_traverse_iter_init_commit`) that takes flags right now.

src/ostree/ot-builtin-prune.c
```diff
@@ -103,9 +105,19 @@ traverse_keep_younger_than (OstreeRepo *repo, const char *checksum,
   /* This is the first commit in our loop, which has a ref pointing to it. We
    * don't want to auto-prune it.
    */
-  if (!ostree_repo_traverse_commit_union (repo, checksum, 0, reachable,
-                                          cancellable, error))
-    return FALSE;
+  if (opt_commit_only)
+    {
+      if (!ostree_repo_traverse_commit_only_union (repo, checksum, 0, reachable,
```
cgwalters, line 110: Having a new "with flags" be the public API simplifies this, because then we can just compute the desired flags, and call the single public API.

cgwalters (commented):
Really nice work on this!  :rocket: 

Some minor nits, nothing major.

It'd be nice to have some new tests in e.g. `tests/test-prune.sh`.

src/libostree/ostree-repo-prune.c
```diff
@@ -125,7 +129,11 @@ maybe_prune_loose_object (OtPruneData           *data,
       else
         data->n_reachable_content++;
     }
-
+    if (commit_only && (objtype != OSTREE_OBJECT_TYPE_COMMIT))
+    {
+      g_debug ("Keeping object (not commit) %s.%s", checksum,
+               ostree_object_type_to_string (objtype));
+    }
```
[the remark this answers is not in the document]
saqibali-2k: I was trying to keep the expected logging consistent. The previous behaviour was that `maybe_prune_loose_object` logs when it doesn't delete an object (ie. the object is reachable). Even though an orphaned object that is not a commit is not reachable, we still skip pruning it. That was my thinking but I'm OK with taking it out.

src/libostree/ostree-repo-prune.c
```diff
@@ -334,9 +332,18 @@ ostree_repo_traverse_reachable_refs (OstreeRepo *self,
   GLNX_HASH_TABLE_FOREACH_V (all_refs, const char*, checksum)
     {
       g_debug ("Finding objects to keep for commit %s", checksum);
-      if (!ostree_repo_traverse_commit_union (self, checksum, depth, reachable,
-                                              cancellable, error))
+      if (commit_only)
+       {
```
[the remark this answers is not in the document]
saqibali-2k: > GNU coding style which we follow has two space indent, this looks like just one.

Ah shoot, I realised I had done this and I went back to fix it but it looks like I still missed a couple :disappointed: 

>(IMO though the real fix is to have programming languages come with an opinionated out of the box formatter, like Rust and Go do, but...that ship long, long ago sailed for C smile )

Ye that would be awesome. The amount of time gofmt has saved me...

src/ostree/ot-builtin-prune.c
saqibali-2k: Personally, I felt a little confused when the `Total objects: ` value printed out didn't match the total number of objects in the repo. However, I realised that this only includes the number of objects traversed or considered for pruning. I thought that message might clear things up, but maybe it doesn't achieve what I was going for.

Maybe a compromise would be to change `Total objects:` to `Total objects traversed:`?

src/libostree/ostree-repo-traverse.c
saqibali-2k: I agree, that sounds like a good idea to me. When I was thinking about this, I was worried about doubling the functionality of `ostree_repo_traverse_commit_union` and potentially making that function obsolete. Is that a concern here?

saqibali-2k:
> Really nice work on this! rocket
> 
> Some minor nits, nothing major.
> 
> It'd be nice to have some new tests in e.g. `tests/test-prune.sh`.

Thanks for the review! I'll be sure to add those tests :+1:

src/libostree/ostree-repo-prune.c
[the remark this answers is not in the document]
cgwalters: OK, fine by me to keep too.

src/libostree/ostree-repo-traverse.c
cgwalters: We are indeed making that function obsolete, but that's fine!  I don't think we need to mark it as explicitly deprecated or anything.

See also https://lwn.net/Articles/585415/

src/ostree/ot-builtin-prune.c
saqibali-2k: Thats looks great!

saqibali-2k:
Using this comment as a "do not merge" label since I don't have permissions. Will update this comment when I want to remove the "label".

src/libostree/ostree-repo.h
```diff
@@ -1180,6 +1191,16 @@ gboolean ostree_repo_traverse_commit_union_with_parents (OstreeRepo         *rep
                                                          GCancellable       *cancellable,
                                                          GError            **error);
 
+_OSTREE_PUBLIC
+gboolean ostree_repo_traverse_commit_with_flags (OstreeRepo                     *repo,
```
cgwalters, line 1195: New API also needs to be listed in `libostree-devel.sym`.

I'm not sure how this patch can work without that actually.

src/libostree/ostree-repo-prune.c
```diff
@@ -299,26 +307,19 @@ repo_prune_internal (OstreeRepo        *self,
...
- * Since: 2018.6
- */
-gboolean
-ostree_repo_traverse_reachable_refs (OstreeRepo *self,
-                                     guint       depth,
-                                     GHashTable *reachable,
-                                     GCancellable *cancellable,
-                                     GError      **error)
+static gboolean
+ostree_repo_traverse_reachable_refs_with_options (OstreeRepo                    *self,
```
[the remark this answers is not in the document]
cgwalters: Also while you're here, mind dropping the `ostree_` from the function prefix?  In new code I've been trying to reserve that for actual public functions.  Further a common pattern is to add `_internal` or `_impl` to the function.

So this could be just e.g. `traverse_reachable_internal()` or perhaps just `traverse_impl` or so.

src/libostree/ostree-repo-prune.c
```diff
@@ -299,26 +307,19 @@ repo_prune_internal (OstreeRepo        *self,
...
+                                                  guint                          depth,
+                                                  GHashTable                    *reachable,
+                                                  GCancellable                  *cancellable,
+                                                  GError                       **error)
 {
+  gboolean commit_only = flags & OSTREE_REPO_COMMIT_TRAVERSE_FLAG_COMMIT_ONLY;
+  OstreeRepoCommitTraverseFlags traverse_flags = OSTREE_REPO_COMMIT_TRAVERSE_FLAG_NONE;
+  if (commit_only)
+    traverse_flags |= OSTREE_REPO_COMMIT_TRAVERSE_FLAG_COMMIT_ONLY;
+
```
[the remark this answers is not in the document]
saqibali-2k: > Or is this trying to filter out other flags (which don't exist yet?)

Nope, just a mistake on my part.

src/libostree/ostree-repo-prune.c
```diff
@@ -409,9 +437,15 @@ ostree_repo_prune (OstreeRepo        *self,
    * the deletion.
    */
 
+  OstreeRepoCommitTraverseFlags traverse_flags = 0;
```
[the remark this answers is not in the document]
saqibali-2k: Initialising to the flag will make the code a little bit more clear so I'll do that :+1:

cgwalters (approved):
Thanks for this, great work!

src/libostree/libostree-released.sym
```diff
@@ -229,6 +229,7 @@ global:
         ostree_repo_transaction_stats_get_type;
         ostree_repo_traverse_commit;
         ostree_repo_traverse_commit_union;
+        ostree_repo_traverse_commit_with_flags;
```
cgwalters, line 232: Wait actually, we don't want to make this change. What should happen is the `-devel`....ahh, I see you fell into a trap of being the first person to add a symbol in this cycle.

Can you try adding this patch?

```
index 02ae9c6a..c03360e6 100644
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -170,10 +170,10 @@ endif # USE_GPGME
 
 symbol_files = $(top_srcdir)/src/libostree/libostree-released.sym
 
-# Uncomment this include when adding new development symbols.
-# if BUILDOPT_IS_DEVEL_BUILD
-# symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
-# endif
+Uncomment this include when adding new development symbols.
+if BUILDOPT_IS_DEVEL_BUILD
+symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
+endif
 
 # http://blog.jgc.org/2007/06/escaping-comma-and-space-in-gnu-make.html
 wl_versionscript_arg = -Wl,--version-script=
```

saqibali-2k: Can do!

cgwalters (approved):

jlebon:
Nice work on this @saqibali-2k!

wmanley:
See also: #2021 and there's some discussion about faster pruning on #2338.

PR #2537, opened 2022-02-09
lib/gpg-verify-result: Add missing floating annotation

jlebon:
I think I'm hitting issues due to this while using the Rust bindings:
https://github.com/coreos/rpm-ostree/pull/3406#issuecomment-1033084956

The bindings for those APIs use `from_glib_full` which says:

> Because ownership can only be transferred if something is already
> referenced, this is unsuitable for floating references.

### DIFF
--- a/src/libostree/ostree-gpg-verify-result-dummy.c
+++ b/src/libostree/ostree-gpg-verify-result-dummy.c
@@ -143,7 +143,7 @@ ostree_gpg_verify_result_lookup (OstreeGpgVerifyResult *result,
  * an invalid @signature_index.  Use ostree_gpg_verify_result_count_all() to
  * find the number of signatures in @result.
  *
- * Returns: a new, floating, #GVariant tuple
+ * Returns: (transfer floating): a new, floating, #GVariant tuple
  **/
 GVariant *
 ostree_gpg_verify_result_get (OstreeGpgVerifyResult *result,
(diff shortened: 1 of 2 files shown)


jlebon:
And I guess the dummy versions should match this too for consistency. :arrow_up: 

Still trying this out to see if it fixes my issue.

cgwalters:
Ohhh yes :cry: floating references.  Looks sane to me offhand

jlebon:
Yup, this unblocks me! (Hitting more issues now, but not related to this.)

cgwalters (approved):

cgwalters:
Looks like a flake, but still need to dig into that.

cgwalters:
> 2022-02-09T19:41:13.3576058Z error: opendir(staging-39746821-bc63-4215-838d-209817e18391-jYggLj): No such file or directory

In `test-concurrency.py` again.  That's likely a real bug, but also a flake.

cgwalters:
I had to dig this up and recall, per the docs:

> floating: alias for none, can be used for floating objects.

So I think the way this works is generally because the value will be used at least by bindings, that use will consume the reference.

Hmm...so...previous to this, we were basically double-free'ing?  I want to understand this more.

Anyways https://github.com/ostreedev/ostree-rs/pull/48

PR #2538, opened 2022-02-11
core: Mark `ostree_create_directory_metadata` as `(not nullable)`

cgwalters:
So I can drop an unnecessary use of `unwrap()` in Rust.


dbnicholson (approved):
Not sure what happened to LGTM, but this looks fine.

PR #2539, opened 2022-02-11
lib/bootloader: use ot_journal_print() instead of sd-journal

chergert:
This needs to use the helper so that USE_LIBSYSTEMD still works as
expected.

### DIFF
--- a/src/libostree/ostree-bootloader-zipl.c
+++ b/src/libostree/ostree-bootloader-zipl.c
@@ -21,7 +21,6 @@
 #include "ostree-bootloader-zipl.h"
 #include "ostree-deployment-private.h"
 #include "otutil.h"
-#include <systemd/sd-journal.h>
 #include <string.h>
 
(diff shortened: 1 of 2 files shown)


cgwalters (approved):

cgwalters:
Looks like we need an `#include <syslog.h>` - hmm, this seems to be indirectly pulled in in `ostree-sysroot-deploy.c` but I don't see where offhand (except indirectly via
```
#ifdef HAVE_LIBSYSTEMD
#include <systemd/sd-journal.h>
#endif
```
)

chergert:
Is it worth adding an `#else` there to `#include <syslog.h>`?

cgwalters:
> Is it worth adding an #else there to #include <syslog.h>?

SGTM (But I'd like to know how `ostree-sysroot-deploy.c` is compiling without that).  Or another way to say it is probably the better fix is to do these includes in `otutil.h` so they apply to all users of `ot_journal_send()` automatically.

cgwalters:
Unrelated flake
/ok-to-test

PR #2542, opened 2022-02-15
lib/tar: Add some error prefixing

cgwalters:
We're trying to debug a problem with a tar stream with hardlinks,
and I think this will be helpful.

### DIFF
--- a/src/libostree/ostree-repo-libarchive.c
+++ b/src/libostree/ostree-repo-libarchive.c
@@ -364,6 +364,7 @@ aic_ensure_parent_dir (OstreeRepoArchiveImportContext *ctx,
                        GCancellable        *cancellable,
                        GError             **error)
 {
+  GLNX_AUTO_PREFIX_ERROR ("ostree-tar: Failed to create parent", error);
   /* Who should own the parent dir? Since it's not in the archive, it's up to
    * us. Here, we use the heuristic of simply creating it as the same user as
    * the owner of the archive entry for which we're creating the dir. This is OK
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-repo-libarchive.c
```diff
@@ -631,6 +635,7 @@ aic_handle_file (OstreeRepoArchiveImportContext *ctx,
                  GCancellable       *cancellable,
                  GError            **error)
 {
+  GLNX_AUTO_PREFIX_ERROR ("ostree-tar: Failed to import file", error);
```
jlebon, line 638:
```suggestion
  GLNX_AUTO_PREFIX_ERROR ("ostree-tar: Failed to handle file", error);
```

So that it's different from the prefix for `aic_import_file` (and more closely matches the function name).

Would be good to include the path in there too, though ideally just in a few places so it's not repeated all the way up.

src/libostree/ostree-repo-libarchive.c
```diff
@@ -526,6 +528,7 @@ aic_handle_dir (OstreeRepoArchiveImportContext *ctx,
                 GCancellable       *cancellable,
                 GError            **error)
 {
+  GLNX_AUTO_PREFIX_ERROR ("ostree-tar: Failed to import directory", error);
```
jlebon, line 531: And then

```suggestion
  GLNX_AUTO_PREFIX_ERROR ("ostree-tar: Failed to handle directory", error);
```

for consistency?

cgwalters: Yep thanks, I handled it!

jlebon (approved):

cgwalters:
More flakes :cry: 
Will debug those at some point...

Issue #2543, opened 2022-02-16
Finalizing staged deployments broken on /boot automount


dbnicholson:
Recently we changed our updater to use staged deployments in endlessm/eos-updater#298. That worked fine on systems where `/boot` is a persistent mount point, but it fails on systems that use systemd-boot where `/boot` is the automounted EFI system partition. There are 2 problems with this:

1. If the `/boot` automount expires, the `ostree-finalize-staged.service` unit runs immediately since it has `RequiresMountsFor=/boot`. With nothing keeping the automount from expiring, this can happen at any point prior to shutdown and ruin the feature. This actually deadlocks in systemd, but it would be bad even without the automounting bugs.
2. If `RequiresMountsFor=/boot` is removed and instead just `After=boot.mount` is used, then the service is only triggered on shutdown but the ordering remains. However, if the automount has expired, systemd will ignore the request to remount it since the automount is scheduled to be stopped.

See systemd/systemd#22528 for details. Maybe the solution here is staged deployments are not supported on `/boot` automounts, but I wanted to open for discussion.

dbnicholson:
The dirty idea I had was to change `ostree admin finalize-staged` so it opens `/boot` (and `/sysroot`) when starting and then blocks waiting for `SIGTERM` via `sigwait`. This would effectively turn `/boot` into a persistent mount if an deployment was scheduled for finalizing since the automount would never expire.

cgwalters:
I think that approach isn't dirty at all - it makes sense to me.  The code is already heavily oriented towards using directory file descriptors, so we already have a natural mechanism to hold open the mounts.

cgwalters:
And that would actually change us from using `ExecStop=` to `ExecStart=` which is definitely more natural.

dbnicholson:
Alright, I'll put something together. One question I had is, what if someone unstages the deployment? Should it watch `/run/ostree/staged-deployment` for deletion and stop itself without finalizing? I think finalizing is pretty much idempotent, so probably it could be allowed to run even if someone unstages.

However, you would need to change the builtin to not initially lock the sysroot since that would prevent doing anything else with it until the unit was stopped. So, I think you'd want to block on the signal, receive the signal, lock the sysroot, load it again (so the state of the deployments is up to date), and then finalize.

cgwalters:
[...]

PR #2544, opened 2022-02-16
finalize-staged: Ensure /boot and /sysroot automounts don't expire

dbnicholson:
If `/boot` or `/sysroot` are automounts, then the unit will be stopped
as soon as the automounts expire. That's would defeat the purpose of
using systemd to delay finalizing the deployment until shutdown. This is
not uncommon as `systemd-gpt-auto-generator` will create an automount
unit for `/boot` when it's the EFI System Partition and there's no fstab
entry.

Instead of relying on systemd to run the command via `ExecStop` at the
appropriate time, have `finalize-staged` open `/boot` and `/sysroot` and
then block on `SIGTERM`. Having the directories open will prevent the
automounts from expiring, and then we presume that systemd will send
`SIGTERM` when it's time for the service to stop. Finalizing the
deployment still happens when the service is stopped. The difference is
that the process is already running.

In order to keep from blocking legitimate sysroot activity prior to
shutdown, the sysroot lock is only taken after the signal has been
received. Similarly, the sysroot is reloaded to ensure the state of the
deployments is current.

Fixes: #2543

### DIFF
--- a/src/ostree/ot-admin-builtin-finalize-staged.c
+++ b/src/ostree/ot-admin-builtin-finalize-staged.c
@@ -1,5 +1,6 @@
 /*
  * Copyright (C) 2018 Red Hat, Inc.
+ * Copyright (C) 2022 Endless OS Foundation LLC
  *
  * SPDX-License-Identifier: LGPL-2.0+
  *
@@ -19,8 +20,9 @@
 
 #include "config.h"
 
-#include "config.h"
-
+#include <glib-unix.h>
+#include <sched.h>
+#include <signal.h>
 #include <stdlib.h>
 
 #include "ot-main.h"
@@ -32,9 +34,22 @@
 #include "ostree-cmd-private.h"
 #include "ostree.h"
 
+static gboolean opt_hold;
+
 static GOptionEntry options[] = {
+  { "hold", 0, 0, G_OPTION_ARG_NONE, &opt_hold, "Hold /boot open during finalization", NULL },
   { NULL }
 };
+
+static gboolean
+sigterm_cb (gpointer user_data)
+{
+  gboolean *running = user_data;
+  g_print ("Received SIGTERM, exiting\n");
+  *running = FALSE;
+  g_main_context_wakeup (NULL);
+  return G_SOURCE_REMOVE;
+}
 
 /* Called by ostree-finalize-staged.service, and in turn
  * invokes a cmdprivate function inside the shared library.
@@ -50,13 +65,56 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
 
   g_autoptr(GOptionContext) context = g_option_context_new ("");
   g_autoptr(OstreeSysroot) sysroot = NULL;
+
+  /* First parse the args without loading the sysroot to see what options are
+   * set. */
   if (!ostree_admin_option_context_parse (context, options, &argc, &argv,
-                                          OSTREE_ADMIN_BUILTIN_FLAG_SUPERUSER,
+                                          OSTREE_ADMIN_BUILTIN_FLAG_NO_LOAD,
                                           invocation, &sysroot, cancellable, error))
     return FALSE;
 
-  if (!ostree_cmd__private__()->ostree_finalize_staged (sysroot, cancellable, error))
-    return FALSE;
+  if (opt_hold)
+    {
+      /* Load the sysroot unlocked so that a separate namespace isn't
+       * created. */
+      if (!ostree_admin_sysroot_load (sysroot,
+                                      OSTREE_ADMIN_BUILTIN_FLAG_SUPERUSER | OSTREE_ADMIN_BUILTIN_FLAG_UNLOCKED,
+                                      cancellable, error))
+        return FALSE;
+
+      /* In case it's an automount, open /boot so that the automount doesn't
+       * expire until before this process exits. If it did expire and got
+       * unmounted, the service would be stopped and the deployment would be
+       * finalized earlier than expected.
+       */
+      int sysroot_fd = ostree_sysroot_get_fd (sysroot);
+      glnx_autofd int boot_fd = -1;
+      g_debug ("Opening /boot directory");
+      if (!glnx_opendirat (sysroot_fd, "boot", TRUE, &boot_fd, error))
+        return FALSE;
+
+      /* We want to keep /boot open until the deployment is finalized during
+       * system shutdown, so block on SIGTERM under the assumption that it will
+       * be received when systemd stops the unit.
+       */
+      gboolean running = TRUE;
+      g_unix_signal_add (SIGTERM, sigterm_cb, &running);
+      g_print ("Waiting for SIGTERM\n");
+      while (running)
+        g_main_context_iteration (NULL, TRUE);
+    }
+  else
+    {
+      /* Load the sysroot with the normal flags and actually finalize the
+       * deployment. */
+      if (!ostree_admin_sysroot_load (sysroot,
+                                      OSTREE_ADMIN_BUILTIN_FLAG_SUPERUSER,
+                                      cancellable, error))
+        return FALSE;
+
+      if (!ostree_cmd__private__()->ostree_finalize_staged (sysroot, cancellable, error))
+        return FALSE;
+    }
 
   return TRUE;
 }
--- a/Makefile-boot.am
+++ b/Makefile-boot.am
@@ -41,6 +41,7 @@ systemdsystemunit_DATA = src/boot/ostree-prepare-root.service \
 	src/boot/ostree-boot-complete.service \
 	src/boot/ostree-finalize-staged.service \
 	src/boot/ostree-finalize-staged.path \
+	src/boot/ostree-finalize-staged-hold.service \
 	$(NULL)
 systemdtmpfilesdir = $(prefix)/lib/tmpfiles.d
 dist_systemdtmpfiles_DATA = src/boot/ostree-tmpfiles.conf
@@ -70,6 +71,7 @@ EXTRA_DIST += src/boot/dracut/module-setup.sh \
 	src/boot/ostree-finalize-staged.path \
 	src/boot/ostree-remount.service \
 	src/boot/ostree-finalize-staged.service \
+	src/boot/ostree-finalize-staged-hold.service \
 	src/boot/grub2/grub2-15_ostree \
 	src/boot/grub2/ostree-grub-generator \
 	$(NULL)
--- a/src/boot/ostree-finalize-staged-hold.service
+++ b/src/boot/ostree-finalize-staged-hold.service
@@ -1,0 +1,35 @@
+# Copyright (C) 2018 Red Hat, Inc.
+# Copyright (C) 2022 Endless OS Foundation LLC
+#
+# This library is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2 of the License, or (at your option) any later version.
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library. If not, see <https://www.gnu.org/licenses/>.
+
+# See https://github.com/ostreedev/ostree/pull/2543 for background.
+[Unit]
+Description=Hold /boot Open for OSTree Finalize Staged Deployment
+Documentation=man:ostree(1)
+ConditionPathExists=/run/ostree-booted
+DefaultDependencies=no
+
+RequiresMountsFor=/sysroot /boot
+After=local-fs.target
+Before=basic.target final.target
+
+[Service]
+Type=exec
+
+# This is explicitly run in the root namespace to ensure an automounted
+# /boot doesn't time out since autofs doesn't handle mount namespaces.
+#
+# https://bugzilla.redhat.com/show_bug.cgi?id=2056090
+ExecStart=+/usr/bin/ostree admin finalize-staged --hold
--- a/src/boot/ostree-finalize-staged.service
+++ b/src/boot/ostree-finalize-staged.service
@@ -29,6 +29,11 @@ Before=basic.target final.target
 After=systemd-journal-flush.service
 Conflicts=final.target
 
+# Start the hold unit and ensure it stays active throughout this
+# service.
+Wants=ostree-finalize-staged-hold.service
+After=ostree-finalize-staged-hold.service
+
 [Service]
 Type=oneshot
 RemainAfterExit=yes
--- a/src/ostree/ot-main.c
+++ b/src/ostree/ot-main.c
@@ -579,41 +579,11 @@ on_sysroot_journal_msg (OstreeSysroot *sysroot,
 }
 
 gboolean
-ostree_admin_option_context_parse (GOptionContext *context,
-                                   const GOptionEntry *main_entries,
-                                   int *argc,
-                                   char ***argv,
-                                   OstreeAdminBuiltinFlags flags,
-                                   OstreeCommandInvocation *invocation,
-                                   OstreeSysroot **out_sysroot,
-                                   GCancellable *cancellable,
-                                   GError **error)
+ostree_admin_sysroot_load (OstreeSysroot            *sysroot,
+                           OstreeAdminBuiltinFlags   flags,
+                           GCancellable             *cancellable,
+                           GError                  **error)
 {
-  /* Entries are listed in --help output in the order added.  We add the
-   * main entries ourselves so that we can add the --sysroot entry first. */
-
-  g_option_context_add_main_entries (context, global_admin_entries, NULL);
-
-  if (!ostree_option_context_parse (context, main_entries, argc, argv,
-                                    invocation, NULL, cancellable, error))
-    return FALSE;
-
-  if (!opt_print_current_dir && (flags & OSTREE_ADMIN_BUILTIN_FLAG_NO_SYSROOT))
-    {
-      g_assert_null (out_sysroot);
-      /* Early return if no sysroot is requested */
-      return TRUE;
-    }
-
-  g_autoptr(GFile) sysroot_path = NULL;
-  if (opt_sysroot != NULL)
-    sysroot_path = g_file_new_for_path (opt_sysroot);
-
-  g_autoptr(OstreeSysroot) sysroot = ostree_sysroot_new (sysroot_path);
-  if (!ostree_sysroot_initialize (sysroot, error))
-    return FALSE;
-  g_signal_connect (sysroot, "journal-msg", G_CALLBACK (on_sysroot_journal_msg), NULL);
-
   if ((flags & OSTREE_ADMIN_BUILTIN_FLAG_UNLOCKED) == 0)
     {
       /* If we're requested to lock the sysroot, first check if we're operating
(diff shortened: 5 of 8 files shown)


dbnicholson:
This past the test suite for me locally but it's otherwise totally untested.

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -51,10 +52,58 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+  sigemptyset (&set);
+  sigaddset (&set, SIGTERM);
+  ret = sigprocmask (SIG_BLOCK, &set, NULL);
+  if (ret != 0)
+    {
+      errno = ret;
+      return glnx_throw_errno_prefix (error, "sigprogmask");
+    }
+  g_debug ("Waiting for SIGTERM");
+  ret = sigwait (&set, &sig);
```
cgwalters, line 94: I'm definitely OK with this as is, but we might consider using e.g. `g_unix_signal_add` and blocking in an event loop instead.  That would allow us to generalize in the future to e.g. also using inotify to watch staged deployments.

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -51,10 +52,58 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+  if (ret != 0)
+    {
+      errno = ret;
+      return glnx_throw_errno_prefix (error, "sigwait");
+    }
+
```
cgwalters, line 100: Might as well add a log entry here?  Maybe just:
`g_print ("Caught SIGTERM, finalizing.\n")`?

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -51,10 +52,58 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+   * gets unmounted, the service will be stopped.
+   */
+  int sysroot_fd = ostree_sysroot_get_fd (sysroot);
+  glnx_autofd int boot_fd = -1;
+  g_autoptr(OstreeDeployment) deployment = NULL;
+  g_autofree char *deployment_path = NULL;
+  glnx_autofd int deployment_fd = -1;
+  g_debug ("Opening /boot directory");
+  if (!glnx_opendirat (sysroot_fd, "boot", TRUE, &boot_fd, error))
+    return FALSE;
```
cgwalters, line 70: OK as is, but perhaps note that `_ostree_sysroot_ensure_boot_fd()` exists; we'd just have to expose it via API (perhaps private) to use it.

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -51,10 +52,58 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+  if (ret != 0)
+    {
+      errno = ret;
+      return glnx_throw_errno_prefix (error, "sigprogmask");
+    }
+  g_debug ("Waiting for SIGTERM");
+  ret = sigwait (&set, &sig);
+  if (ret != 0)
+    {
+      errno = ret;
```
cgwalters, line 97: Oh wow, another set of APIs that don't set errno.  Did you just know this offhand?

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -51,10 +52,58 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
     return FALSE;
 
+  /* In case they're automounts, open /boot and the booted deployment on
+   * /sysroot so that the automounts don't expire. If they do expire and either
+   * gets unmounted, the service will be stopped.
+   */
+  int sysroot_fd = ostree_sysroot_get_fd (sysroot);
+  glnx_autofd int boot_fd = -1;
+  g_autoptr(OstreeDeployment) deployment = NULL;
+  g_autofree char *deployment_path = NULL;
```
cgwalters, line 66: In new code I tend to prefer declare-and-initialize style (here on line 74).

cgwalters:
```
--- FAIL: fcos.upgrade.basic/upgrade-from-current (25.94s)
[2022-02-16T23:42:10.925Z]             basic.go:328: expected reboot into version 35.20220216.dev.0.kola, but got version 35.20220216.dev.0
```

Hmm, that looks likely to be related to this.

cgwalters:
```
Feb 16 23:41:46.048773 unknown[1309]: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
Feb 16 23:41:46.050395 ostree[1309]: error: Read-only file system
Feb 16 23:41:46.050743 systemd[1]: Stopping OSTree Finalize Staged Deployment...
Feb 16 23:41:46.052224 systemd[1]: dbus-broker.service: Deactivated successfully.
Feb 16 23:41:46.054235 systemd[1]: ostree-finalize-staged.service: Main process exited, code=exited, status=1/FAILURE
Feb 16 23:41:46.054465 systemd[1]: ostree-finalize-staged.service: Failed with result 'exit-code'.
Feb 16 23:41:46.054840 systemd[1]: Stopped OSTree Finalize Staged Deployment.
```

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -51,10 +52,58 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+  glnx_autofd int deployment_fd = -1;
+  g_debug ("Opening /boot directory");
+  if (!glnx_opendirat (sysroot_fd, "boot", TRUE, &boot_fd, error))
+    return FALSE;
+  deployment = ostree_sysroot_require_booted_deployment (sysroot, error);
+  if (deployment == NULL)
+    return FALSE;
+  deployment_path = ostree_sysroot_get_deployment_dirpath (sysroot, deployment);
+  g_debug ("Opening booted deployment /%s", deployment_path);
```
cgwalters, line 75: And wait, why do we need this one?  How would automounts be involved for it?

src/ostree/ot-admin-builtin-finalize-staged.c
dbnicholson: Good call. I had been thinking the same while pondering whether to watch `/run/ostree/staged-deployment`, so might as well just do it now.

src/ostree/ot-admin-builtin-finalize-staged.c
dbnicholson: I did it for completeness, but you're right that I don't think it's needed. ostree is the one that sets up the `/sysroot` mount, so we can be sure it won't be an automount. I'll drop that.

src/ostree/ot-admin-builtin-finalize-staged.c
dbnicholson: Fair. I ended up dropping the deployment handling here. I have noticed that in some scenarios gcc (at least on my system) seems to think you're using variables uninitialized when you don't explicitly initialize auto variables to `NULL`. I guess it probably is because of some code reordering. Anyways, that's tangential.

src/ostree/ot-admin-builtin-finalize-staged.c
dbnicholson: :+1:

dbnicholson:
> ```
> Feb 16 23:41:46.048773 unknown[1309]: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
> Feb 16 23:41:46.050395 ostree[1309]: error: Read-only file system
> Feb 16 23:41:46.050743 systemd[1]: Stopping OSTree Finalize Staged Deployment...
> Feb 16 23:41:46.052224 systemd[1]: dbus-broker.service: Deactivated successfully.
> Feb 16 23:41:46.054235 systemd[1]: ostree-finalize-staged.service: Main process exited, code=exited, status=1/FAILURE
> Feb 16 23:41:46.054465 systemd[1]: ostree-finalize-staged.service: Failed with result 'exit-code'.
> Feb 16 23:41:46.054840 systemd[1]: Stopped OSTree Finalize Staged Deployment.
> ```

This is because I changed to to use `OSTREE_ADMIN_BUILTIN_FLAG_UNLOCKED` so that initially the sysroot is not locked. However, I missed the part about setting up the mount namespace so that `/boot` and `/sysroot` can be remounted read-write. I wanted to factor out the code in `ostree_admin_option_context_parse`, but I got a little confused about the different use cases and left a `FIXME`. If you see a good way to handle that, I'll give it a shot.

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -51,10 +62,46 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+    g_main_context_iteration (NULL, TRUE);
+
+  /* Since the sysroot was initially started unlocked, setup a mount namespace
+   * to allow /boot and /sysroot to be writable, lock the sysroot, and reload
+   * it to ensure the state of the deployments is current.
+   *
+   * FIXME: This overlaps with the mount namespace handling in
+   * ostree_admin_option_context_parse. That should be factored out.
+   */
+  if (unshare (CLONE_NEWNS) < 0)
```
cgwalters, line 97: Ohhh, right. I think it would be cleaner is to have systemd do this for us via e.g.
`MountFlags=slave`
in the unit.  See e.g.
https://github.com/coreos/rpm-ostree/blob/baa2eba2d77647204be2baff2b385e16220e0fac/src/daemon/rpm-ostreed.service.in#L11
(In rpm-ostree we always do everything via the daemon; unfortunately that's not true of ostree today, but in *this* case we know we're running in systemd and can use its features)

(Although I guess this direction conflicts with people who are trying to do ostree without systemd, but...I think the onus is going to be increasingly on them to at least parse a subset of unit files)

But, OK as is too - i.e. we can merge this and do followups.

cgwalters (approved):
Looks generally good to me.  Did you test this locally now?

At least our CI will cover this pretty well, but we just need to check any failures there versus the random flakes.

dbnicholson:
I think the LGTM issue should be fixed by #2545.

src/ostree/ot-admin-builtin-finalize-staged.c
dbnicholson: Just to make sure I'm following this right, because the unit has `ProtectHome` and `ReadOnlyPaths`, then systemd is already going to setup a mount namespace and there's no reason to create another one. Is that right? I sorta feel like you'd want to have code to check whether the process was in its own namespace or not.

Another thing I thought of while looking at this section. The unit talks about how only `ProtectHome` can be used, but then the sysroot goes and remounts `/boot` and `/sysroot` readwrite. Furthermore, it talks about how it needs to remove `/var/.updated`, but isn't that within the stateroot `/var`? Oh, I see that there's some handling for `/var` as a separate mount. Anways, this all looks like it could be handles with:

```
ProtectSystem=strict
ReadWritePaths=/var
```

with the mount handling. Maybe another day, though.

dbnicholson:
> Looks generally good to me. Did you test this locally now?
> 
> At least our CI will cover this pretty well, but we just need to check any failures there versus the random flakes.

Yeah, thanks for the kola tests. I kinda wish I could get that environment going locally.

Let me try to take this for a spin in my problematic VM with the `/boot` automount and make sure it actually does what I think it will.

dbnicholson:
No changes in that last force push. Just wanted to rebase on the LGTM fix.

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -51,10 +62,46 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+    return glnx_throw_errno_prefix (error, "setting up mount namespace: unshare(CLONE_NEWNS)");
+  ostree_sysroot_set_mount_namespace_in_use (sysroot);
```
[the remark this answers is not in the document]
cgwalters: If it helps we can move this entire thing into a `ostree_cmd__private__ ()->ostree_finalize_staged` that can use all the internal APIs.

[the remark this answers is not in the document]
cgwalters: A good reference for this is `ostree-system-generator.c` - the binary just calls into a private API from the library.

src/ostree/ot-admin-builtin-finalize-staged.c
jlebon: Personally I've definitely called `ostree admin finalize-staged` manually when debugging stuff. I could fake out the `INVOCATION_ID` I guess, but maybe more elegant to do:
- if in systemd unit, don't unshare
- otherwise unshare or die trying

?

src/ostree/ot-admin-builtin-finalize-staged.c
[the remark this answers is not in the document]
dbnicholson: Maybe, but the complication is in loading the sysroot appropriately, which currently always happens in the builtins (and ostree-system-generator doesn't do). I think there are advantages to moving the whole thing internal, but the downside is repeating all of the setup code in `src/ostree`.

dbnicholson:
Fun times. At least in my VM, when systemd sets up a mount namespace (due to either of `ProtectHome` or `ReadOnlyPaths`), then autofs doesn't consider the open `/boot` FD and expires the mount. If I remove both of `ProtectHome` and `ReadOnlyPaths`, then it works fine. I also found https://bugzilla.redhat.com/show_bug.cgi?id=1569146 and https://www.spinics.net/lists/autofs/msg01694.html.

cgwalters:
> Once the initial process in the root namespace exits, then there won't be anything keeping /boot from expiring since the subsequent command is in a separate namespace.

In my proposal, both processes stay active until they are both complete.

jlebon:
Can't a service unit only have a single top-level process though? `systemd.service(5)` says:

> Unless Type= is oneshot, exactly one command must be given. When Type=oneshot is used, zero or more commands may be specified. Commands may be specified by providing multiple command lines in the same directive, or alternatively, this

And even in the `oneshot` case, the commands are executed sequentially.

> I think the only way to do this is to handle the mount namespace in process as I was doing on my last attempt while ensuring it's run in the root namespace. Then the initial /boot FD stays open in the root namespace while the rest happens in the new namespace. You'd basically have to recreate everything systemd is doing, but I don't think it would be that complex and we'd have full control over exactly what should be mounted. For example, /var in the stateroot vs /var as separate filesystem. Well, after reading the systemd namespace setup it's a little more complex than I was thinking.

Hmm, I think another simpler approach would be having the fd holding happen in a separate unit without a mount namespace which runs `Before=ostree-finalize-staged.service`.

cgwalters:
> Hmm, I think another simpler approach would be having the fd holding happen in a separate unit without a mount namespace which runs Before=ostree-finalize-staged.service.

Yeah, also SGTM.

AdrianVovk:
carbonOS is using a boot automount, but I don't necessarily see a reason to not just set `TimeoutIdleSec=0` to disable auto-unmount on `/boot`. Of course, it would be _nice_ if OSTree could just handle auto-unmounting `/boot`, but I don't think it's a super high priority either.

Perhaps it should be documented somewhere that staged deployments are incompatible with /boot timing out (maybe in the API doc for creating a staged deployment?) until this fix lands?

dbnicholson:
I moved away from this because we were going to EOL the product that this affects, but now that decision has been reversed. My plan here is:

* Refactor `ostree_admin_option_context_parse` so that the setup can be composed elsewhere.
* Add the `finalize-staged` `--hold-boot` (or maybe `--wait`?) option to skip all the setup, open `/boot` and block until `SIGTERM` (mostly already in the PR). Make the `finalize-staged` non-`hold-boot` option do the rest of the normal sysroot setup and the actual finalizatoin.
* Add an `ostree-finalize-staged-hold-boot.service` unit that explicitly runs in the root namespace with `ExecStart=+/usr/bin/ostree admin finalize-staged --hold-boot`. Add `Wants`/`After=ostree-finalize-staged-hold-boot.service` to `ostree-finalize-staged.service`.

Seem reasonable? I think another way would be to just add a separate `finalize-staged-hold-boot` command instead of implementing it as an option.

cgwalters:
Two systemd units as jlebon originally suggested with two distinct CLI verbs seems conceptually cleanest to me, it's just a bit more verbose to type out.

You're right that there are a ton of options that will cause a mount namespace to be created, but we can just basically not use any unit options in the `ostree-finalize-staged-hold-boot.service`.

dbnicholson:
> You're right that there are a ton of options that will cause a mount namespace to be created, but we can just basically not use any unit options in the `ostree-finalize-staged-hold-boot.service`.

I'm going with no options in the unit and the explicit `ExecStart=+` notation you mentioned earlier to ensure this is in the root namespace.

I'd like to add a test in `tests/kolainst/destructive/` since it seems like I should be tweak the system to have `/boot` automounted. However, I'd really like to do that locally to speed up iterations instead of cycling through PR checks. Any suggestions on how to do that?

cgwalters:
> However, I'd really like to do that locally to speed up iterations instead of cycling through PR checks. Any suggestions on how to do that?

A cool thing IMO about https://github.com/coreos/coreos-assembler is that the entire build and test tooling is one (large) container image you can run anywhere you like.  Though it also requires doing a build in the corresponding userspace (e.g. f36).  But it's basically using https://coreos.github.io/coreos-assembler/working/#using-overrides to drop in the new ostree binaries, plus `kola run ext.ostree.destructive`.  (Actually also need a `makesudoinstall -C tests/kolainst` to install the tests in the cosa container userspace)

dbnicholson:
> A cool thing IMO about https://github.com/coreos/coreos-assembler is that the entire build and test tooling is one (large) container image you can run anywhere you like. Though it also requires doing a build in the corresponding userspace (e.g. f36). But it's basically using https://coreos.github.io/coreos-assembler/working/#using-overrides to drop in the new ostree binaries, plus `kola run ext.ostree.destructive`. (Actually also need a `makesudoinstall -C tests/kolainst` to install the tests in the cosa container userspace)

That thing is really cool. I'm sure I wasn't driving it very well, but that was **way** faster than how I would have normally done something like that. I wish we had something like that for Endless.

dbnicholson:
I think this is ready to review again. I ended up not touching the `ostree_admin_option_context_parse` and instead just calling it twice with different flags depending on the context. That seems to have worked.

The added kola test was extremely helpful to make sure this was actually doing what I thought it would do.

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -50,13 +65,57 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
 
   g_autoptr(GOptionContext) context = g_option_context_new ("");
   g_autoptr(OstreeSysroot) sysroot = NULL;
+
+  /* First parse the args without loading the sysroot to see what options are
```
cgwalters, line 69: I'm OK with this but...I think we could also bypass the whole thing and just do

```
bool opt_hold = argc > 2 && strcmp (argv[1], "--hold") == 0;
```
or so?  Or use an environment variable.

Dunno.  I don't feel really strongly.  My main concern is that someone later comes by and makes some sort of change that isn't ready for `ostree_admin_option_context_parse` being invoked twice.

cgwalters (approved):
Looks good to me as is, we can address anything else via followup commits too.

src/ostree/ot-admin-builtin-finalize-staged.c
dbnicholson: Eh, I'm personally not a fan of adding adhoc option handing. What if the first option is `--verbose`or something else?

I think I can go back to the refactoring route and split out the sysroot opening.

cgwalters: I'm OK with the code as is too.

Though...hmm, wouldn't it work to use plain old `g_option_context_parse()` at least?  That would avoid any issues with invoking `ostree_admin_option_context_parse` twice.

dbnicholson: I think it would except that you'd have to duplicate the handling of the `--sysroot` option at least. I think the refactoring I added to initialize but not load the sysroot should work well. See the new prep commit I added.

src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -50,13 +65,56 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+
+      /* We want to keep /boot open until the deployment is finalized during
+       * system shutdown, so block on SIGTERM under the assumption that it will
+       * be received when systemd stops the unit.
+       */
+      gboolean running = TRUE;
+      g_unix_signal_add (SIGTERM, sigterm_cb, &running);
+      g_print ("Waiting for SIGTERM\n");
+      while (running)
+        g_main_context_iteration (NULL, TRUE);
```
cgwalters, line 104: I'm fine with this as is, but in general I think it's even cleaner where possible to *not* install a `SIGTERM` handler and let the kernel simply kill the process.  That way there's one fewer context switch at shutdown time.

IOW, we could just do:

```
while (true)
  g_main_context_iteration (NULL, TRUE);
```

cgwalters (approved):
Nice, thanks so much for doing this!

src/ostree/ot-admin-builtin-finalize-staged.c
dbnicholson: Very true. I guess I like the `SIGTERM` handler so that the service exits gracefully instead of showing as failed, although in this case you'd probably never look at the status of the service.

cgwalters: What would show as failed?

```
[root@cosa-devsh ~]# cat /etc/systemd/system/testunit.service
[Service]
ExecStart=sleep infinity
[root@cosa-devsh ~]# systemctl start testunit
[root@cosa-devsh ~]# systemctl status testunit
● testunit.service
     Loaded: loaded (/etc/systemd/system/testunit.service; static)
     Active: active (running) since Tue 2022-08-30 20:21:23 UTC; 1s ago
   Main PID: 1792 (sleep)
      Tasks: 1 (limit: 1042)
     Memory: 284.0K
        CPU: 1ms
     CGroup: /system.slice/testunit.service
             └─ 1792 sleep infinity

Aug 30 20:21:23 cosa-devsh systemd[1]: Started testunit.service.
[root@cosa-devsh ~]# systemctl stop testunit
[root@cosa-devsh ~]# systemctl status testunit
○ testunit.service
     Loaded: loaded (/etc/systemd/system/testunit.service; static)
     Active: inactive (dead)

Aug 30 20:21:27 cosa-devsh systemd[1]: Stopping testunit.service...
Aug 30 20:21:27 cosa-devsh systemd[1]: testunit.service: Deactivated successfully.
Aug 30 20:21:27 cosa-devsh systemd[1]: Stopped testunit.service.
```

cgwalters: PR in https://github.com/ostreedev/ostree/pull/2704

dbnicholson: I guess I was thinking that systemd would see the non-0 exit status and interpret it as failed. But it makes sense since it's the parent and can process the status appropriately.

PR #2546, opened 2022-02-17
build-sys: Drop `-Werror=aggregate-return`

cgwalters:
This is failing for me as of recently but only when I build
without optimization.  I don't think we've ever written any
code that returned a large structure by value.

Let's just drop this one.


dbnicholson (approved):
Works for me.

PR #2547, opened 2022-02-18
[RFC][WIP] port to libsoup3

q66:
Hello,

this is an initial port of libostree to use libsoup3. The reason for this is:

1) GNOME and webkit are migrating to libsoup3, and other upstreams should do so as well as the new generation is significantly cleaned up, brings http/2 support and other things
2) libsoup2 and libsoup3 cannot exist in the same process together
3) libsoup2 is in maintenance mode and everything is expected to migrate over time

This pull request does several things:

1) Upgrades glib requirement to at least 2.68 (though it does not yet actually fix up all the warnings that result from this - it only fixes up the bare minimum to get it to build)
2) changes all usage of SoupURI to GUri (which is needed for libsoup3) - this is why the above is needed
3) actually ports the fetcher as well as the server
4) removes the soupuri copypasta from the build (though does not yet remove the actual files)

For now this is rough, as I wanted to get comments first. Notably, I was wondering wthether to attempt some kind of dual-libsoup compatibility. I have already attempted this, but it's fairly complicated and it would take a lot of effort to coerce the code to work both ways (with a lot of stuff being wildly different between the generations). Additionally, I wanted to find out how the maintainers feel about bumping up the dependency requirements (glib), as that is a significant thing to consider. If this is okay, it would probably be a good idea to start by updating the glib requirement in the codebase on its own first.
 
It is expected that this rework is buggy for now. I wanted to get this out so that people are aware of it, and so that somebody else doesn't start working on it and waste effort.

In any case, I'm looking forward to this getting a review.


q66:
The reason the dependency is bumped is the migration to `GUri`; `SoupURI` is gone in libsoup3, and the copy-pasted `SoupURI` inside the codebase appears to be pretty much a workaround. `GUri` as it is requires at least 2.66, while 2.68 is necessary for `G_URI_FLAGS_SCHEME_NORMALIZE`.

So the options are basically:

1) just switch to `GUri` as it is right now; that means dropping a bunch of code even for libcurl, and is likely better going forward, as long as you can deal with the newer version
2) keep the copypasted `SoupURI`; enable the copypasted `SoupURI` even for libsoup backend, add internal code for conversion to `GUri` where `libsoup` needs it; change the build system to only need newer glib when compiling libsoup (because regardless of how things done, newer libsoup will pull in newer glib requirement because it uses `GUri` in its public API), and somehow deal with the glib version incompatibilities (which will likely be its own can of worms because it'd need ifdefs or wrappers around the codebase)

cgwalters:
> just switch to GUri as it is right now; that means dropping a bunch of code even for libcurl, and is likely better going forward, as long as you can deal with the newer version

Yeah, SGTM.  *But* I don't want to speak for everyone - this definitely has implications for the flatpak ecosystem.  There's a somewhat related thread over here: https://github.com/flatpak/flatpak/issues/2241#issuecomment-1044434100

So summoning @smcv

mwleeds:
I think the related Flatpak PR is https://github.com/flatpak/flatpak/pull/4582 unless I'm missing something.

smcv:
GLib 2.68 is going to be available in Ubuntu 22.04 (coming soon) and Debian 12, but a dependency on GLib 2.68 would rule out backporting libostree to Ubuntu 20.04 (the current LTS right now) and Debian 11.

If it's possible to limit the dependency to 2.66, then that would open up the option of backporting to Debian 11, which would be nice.

At the moment, all backport and PPA builds of Flatpak use libostree 2020.8, which is the version shipped in Debian 11, and that has been sufficient so far (although there are some annoying bugs for which I might try to backport patches at some point).

However, if Flatpak wants to start depending on new features of libostree 2022+, then a GUri dependency would make that awkward.

The other side of this is that Flatpak itself uses libsoup directly. If Flatpak and libostree are ported to libsoup 3, distros that compile libostree against libsoup will need to be able to switch both of them from libsoup 2 to libsoup 3 as a flag-day, because libsoup doesn't have versioned symbols and so mixing the two libraries will crash.

At the moment, the .deb world is still building libostree against libsoup, because it is not clear to me what the advantages and disadvantages of libsoup vs libcurl are, so I stuck with the status quo instead of making arbitrary changes; and the Ubuntu and PPA packages are derived from my Debian packages, so they inherit that policy. If there are reasons why I should switch to libcurl, that's entirely possible to do.

dbnicholson:
At Endless we use libsoup in ostree and I don't think we'd switch over to libcurl without a lot of testing. Even though I don't have any reasons to doubt libcurl, download failures == you can't upgrade failures. The ostree/flatpak libsoup flag day does not sound fun, though.

q66:
I added back support for libsoup2; it can be enabled with `--with-soup2`. The higher GLib requirement is still in place (I will try to get it down to 2.66 though) as `GUri` is still used in `ostree-fetcher-uri` so that the copypasta can be dropped and so that the work becomes easier for soup3 support. The soup2 support is written against relatively modern soup2 API (i.e. all the deprecated conditional stuff is gone, I set the baseline to 2.70.0 which matches Ubuntu 20.04) and to share as much as possible with the soup3 support, it does not use `SoupRequest` and the likes.

This should address things for distributions that will not be switching to soup3 yet.

q66:
changed minimum glib version to 2.66; updated the GUri code to only use the 2.68 scheme normalization when new enough and otherwise it normalizes manually

dbnicholson:
Can you split out the GUri change? It seems useful on its own and a necessary prerequisite for the rest of the work.

After that, is there any reason soup3 can't be a separate fetch backend instead of replacing the soup2 one? The hard work of abstracting the fetch backend has already been done, so I don't see a good reason not to leave the soup2 one alone. That would allow ostree and flatpak to migrate in a much cleaner way and let consumers using the soup2 backend continue to do so. At some point in the future the soup2 backend could be dropped.

q66:
well, at this point soup2 and soup3 share a lot of code; if you prefer that soup2 code remains mostly untouched and soup3 code is made separate, I can do that too

i was planning to split the GUri changes all along, so sure

dbnicholson:
I personally think you should leave the soup2 code entirely alone. That ensures there are no regressions introduced and removing the backend later is simpler. At that point any duplication just goes away and there are no leftover abstraction layers from soup2/3 code sharing. Maybe other ostree developers feel differently, though.

q66:
yeah, sounds good; let's do that

cgwalters:
Dan's suggestion SGTM!  It will definitely be easier for distributors to avoid a chained flag day that way.

(Although, have the recursive dependencies of libflatpak been analyzed?  e.g. does gnome-software also use libsoup directly?  I'd be surprised if it didn't.)

q66:
I've previously compiled a more or less complete list of all software that uses libsoup (https://gitlab.gnome.org/GNOME/libsoup/-/issues/218); we're gradually helping upstreams port things over

i'm also working on a linux distribution that currently uses libsoup3 only, so that gets me a testbed to verify things

q66:
ok, first part of the work is at https://github.com/ostreedev/ostree/pull/2551

q66:
this is now rebased on the GUri work; the soup3 backend is added as a separate thing

i adjusted the CI to use soup2, unfortunately it's not available in Debian yet - I should figure out a way to get it tested (as there are probably bugs right now)

the httpd is ported to support both versions of libsoup at once

configure.ac
```diff
@@ -152,10 +152,18 @@ AS_IF([test x$enable_http2 != xno ], [
...
 AC_ARG_WITH(soup,
 	    AS_HELP_STRING([--with-soup], [Use libsoup @<:@default=yes@:>@]),
 	    [], [with_soup=$with_soup_default])
+AC_ARG_WITH(soup2,
+	    AS_HELP_STRING([--with-soup2], [Use libsoup2 @<:@default=no@:>@]),
+	    [], [with_soup2=no])
+AS_IF([test x$with_soup2 != xno], [
+    SOUP_DEPENDENCY=$SOUP2_DEPENDENCY
+    with_soup=yes
+])
```
dbnicholson, line 166: I think `--with-soup` should continue to be soup2 and the default. The soup3 backend should be a new option `--with-soup3`. I don't think it makes sense to switch to a brand new backend for the default. The soup2 and curl backends have seen years of testing in the field and I don't think "because soup3 is available" is a good reason to switch that right away. The goal to eradicate soup2 from a system has merit, but having a tested and working fetch backend is likely a much stronger goal for most people. As I stated before, if you're using ostree for a bootable OS, you don't get updates if the fetch backend has bugs. That's the worst kind of bug.

When the soup3 backend gets some testing in the field, we can flip around the options, IMO.

q66: ok, that's fair, we can do that

q66:
sure, that's okay, the CI matrix has debian testing already, so we can either switch that one to soup3 or add a new one for testing + soup3, i'll leave that decision to the maintainers here

dbnicholson:
IMO, let's start with testing + explicit `--with-soup3`. I'm less familiar the FCOS parts of the CI, but that might be workable, too.

q66:
I changed stuff so that `--with-soup3` is used to enable libsoup3; I suggest that we wait for Debian testing to pick up libsoup3, then add the CI and fix up any remaining issues (we'll have plenty of time through the gnome 43 development cycle to get this in)

q66:
hm, not really sure what is causing those tests failures...

dbnicholson:
@q66 Are you planning on picking this up again? Need help with it?

q66:
the PR should be pretty much ready, there were some CI fails but I wasn't sure how to deal with them (probably would need somebody else to take a look)

q66:
rebased

mattst88:
Is this waiting on anything other than a maintainer to check the CI and/or merge it?

q66:
as far as i can tell, the soup session is created in the thread in the new soup3 fetcher, and all api usage is already async? or do you see something that i have missed?

q66:
in any case, the draft-ness is just something i missed, i didn't realize it was what was preventing you from reviewing it; that said, feel free to push your fixes in this branch, and let me know if you need me to do anything else

dbnicholson:
> as far as i can tell, the soup session is created in the thread in the new soup3 fetcher, and all api usage is already async? or do you see something that i have missed?

I see this in some of the test logs:

```
# libsoup:ERROR:../libsoup/soup-session.c:1458:message_completed: assertion failed: (item->context == soup_thread_default_context ())
# Bail out! libsoup:ERROR:../libsoup/soup-session.c:1458:message_completed: assertion failed: (item->context == soup_thread_default_context ())
```

I think if that can be fixed and the test suite passes, then I think this can go in as a first cut. However, I think it would be better to not use threads, right? Then you avoid any of the soup3 thread gotchas and you get the HTTP/2 benefits.

Can I push to your branch? I guess I could make a PR on top of it with my changes.

Makefile-libostree.am
```diff
@@ -226,7 +226,11 @@ libostree_1_la_CFLAGS += $(OT_DEP_CURL_CFLAGS)
 libostree_1_la_LIBADD += $(OT_DEP_CURL_LIBS)
 else
 if USE_LIBSOUP
+if USE_LIBSOUP3
+libostree_1_la_SOURCES += src/libostree/ostree-fetcher-soup3.c
+else
 libostree_1_la_SOURCES += src/libostree/ostree-fetcher-soup.c
+endif
```
dbnicholson, line 233: I think like in the fetcher code itself, libsoup (i.e. soup2) and libsoup3 should be totally separate if possible.

Makefile-libostree.am
```diff
@@ -267,7 +271,7 @@ OSTree_1_0_gir_INCLUDES = Gio-2.0
 OSTree_1_0_gir_CFLAGS = $(libostree_1_la_CFLAGS)
 OSTree_1_0_gir_LIBS = libostree-1.la
 OSTree_1_0_gir_SCANNERFLAGS = --warn-all --identifier-prefix=Ostree --symbol-prefix=ostree $(GI_SCANNERFLAGS)
-OSTree_1_0_gir_FILES = $(libostreeinclude_HEADERS) $(filter-out %-private.h %/ostree-soup-uri.h,$(libostree_1_la_SOURCES))
+OSTree_1_0_gir_FILES = $(libostreeinclude_HEADERS) $(filter-out %-private.h,$(libostree_1_la_SOURCES))
```
dbnicholson, line 274: I think if you rebase this will fall out.

configure.ac
```diff
@@ -153,9 +153,17 @@ dnl When bumping the libsoup-2.4 dependency, remember to bump
...
+SOUP3_DEPENDENCY="libsoup-3.0 >= 3.0.0"
 AC_ARG_WITH(soup,
 	    AS_HELP_STRING([--with-soup], [Use libsoup @<:@default=yes@:>@]),
 	    [], [with_soup=$with_soup_default])
+AC_ARG_WITH(soup3,
+	    AS_HELP_STRING([--with-soup3], [Use libsoup3 @<:@default=no@:>@]),
+	    [], [with_soup3=no])
+AS_IF([test x$with_soup3 != xno], [
+    SOUP_DEPENDENCY=$SOUP3_DEPENDENCY
+    with_soup=yes
```
dbnicholson, line 165: I'd prefer if the soup3 handling was distinct from the soup2 handling rather than being an option if possible. In other words, I don't think enabling soup3 should set `with_soup=yes` and I think `OSTREE_FEATURES` should say `libsoup3`.

dbnicholson (changes requested):
Some nitpicks on the build side. I'd prefer duplication as separate options rather than having soup2 vs soup3 be treated the same just like how the fetch backends are separated. That will likely require some duplication of build settings, but I'd rather have that then having them intertwined. ostree-trivial-httpd makes this more awkward, but I'm going to treat that as a special case.

q66:
alright, i will rebase later, if you can't push into my PR (i know gitlab allows it but github might not), you could just post a git patch and i can apply it myself

the log indeed seems to indicate wrong usage, I'll see what i can do (unfortunately i have much less time to work on this than before)

dbnicholson:
> alright, i will rebase later, if you can't push into my PR (i know gitlab allows it but github might not), you could just post a git patch and i can apply it myself

What do you know, I could push. Feel free to squash those 3 patches into yours or polish the commit messages if you prefer. I also ran into dd98a2a4cfea85e8a666f3a2b7e23240fe0a59d6, which would go away on a rebase.

dbnicholson:
Well, it wasn't going to run the tests without a rebase, so I did it and force-pushed back to your branch. Locally I only have 2 errors left:

```
ERROR: tests/test-pull-large-metadata.sh
========================================
...
# **
# libsoup:ERROR:../libsoup/soup-session.c:1458:message_completed: assertion failed: (item->context == soup_thread_default_context ())
# Bail out! libsoup:ERROR:../libsoup/soup-session.c:1458:message_completed: assertion failed: (item->context == soup_thread_default_context ())
```

I'm not sure what's going on there.

```
ERROR: tests/test-remote-cookies.sh
===================================
...
+ ostree --repo=repo pull origin main
error: Server returned status 403: Forbidden
...
+ ostree --repo=repo pull origin main
error: Could not connect to 127.0.0.1: Connection refused
```

The connection refused happens after trivial-httpd dies. Unfortunately, trivial-httpd's stderr gets closed when it's daemonized, so it's hard to tell what's going on.

dbnicholson:
I fixed all the soup2 regressions, but there are still piles of soup3 test errors to comb through. Part of it has to do with dconf that I'm guessing soup3 pulls in.

dbnicholson:
With a couple environment variables stolen from libsoup's test suite, it's down to one failing test:

```
ERROR: tests/test-pull-large-metadata.sh
========================================
...
# libsoup:ERROR:../libsoup/soup-session.c:1458:message_completed: assertion failed: (item->context == soup_thread_default_context ())
```

Makefile-libostree.am
dbnicholson: I redid this part so they're a bit more separated at the expense of a little duplication.

configure.ac
dbnicholson: I redid this part so that the soup3 option is distinct just like the curl option.

dbnicholson:
I squashed my fixups into @q66's patch. I then decided to give the full single thread async approach. I think it turned out nicely and (IMO) is simpler. And it got rid of the last test failure.

@cgwalters or @jlebon either of you care to review?

dbnicholson:
> Looks sane to me, thanks! I do still prefer declare-and-initialize for new code, but it's fine.

I was fixing those as I noticed them. Let me take one more sweep to get them all.

dbnicholson:
Thanks for getting this going @q66!

PR #2551, opened 2022-02-21
lib: bump glib requirement to 2.66 and port to GUri

q66:
This removes the old SoupURI copypasta from previous generation of libsoup and opens up a path for a simple libsoup3 port.

~~There are a coupe files that extensively use the deprecated `g_simple_async_result` APIs. For now, I guarded them with:~~ This is now fixed, everything using `g_simple_async_result` has been ported to use `GTask`.

Prepares for https://github.com/ostreedev/ostree/pull/2547


q66:
converted remaining deprecated stuff to `GTask`

src/libostree/ostree-repo-static-delta-processing.c
```diff
@@ -333,28 +335,25 @@ _ostree_static_delta_part_execute_async (OstreeRepo      *repo,
...
 
-  asyncdata->result = g_simple_async_result_new ((GObject*) repo,
-                                                 callback, user_data,
-                                                 _ostree_static_delta_part_execute_async);
-
-  g_simple_async_result_set_op_res_gpointer (asyncdata->result, asyncdata,
-                                             static_delta_part_execute_async_data_free);
-  g_simple_async_result_run_in_thread (asyncdata->result, static_delta_part_execute_thread, G_PRIORITY_DEFAULT, cancellable);
-  g_object_unref (asyncdata->result);
+  task = g_task_new (G_OBJECT (repo), cancellable, callback, user_data);
```
cgwalters, line 338: I think this can now just be `g_autoptr(GTask) task = g_task_new (...)` and then we can drop the manual `g_object_unref()` below.

cgwalters (approved):
Nice!  Thank you so much for doing this!

cgwalters:
> error: Server returned status 401: Unauthorized

CI failure looks related - possibly the `G_URI_HIDE_PASSWORD` bits?

q66:
yes, i gotta investigate the CI failure...

q66:
it would seem the issue is due to quirky/badly documented behavior of GUri, I changed things to be safe. Other than that I still have a few cleanups to do here, and I noticed I accidentally removed a warning flag from configure, so I will finish these up, but other than that this should be good.

cgwalters:
OK, looks like a few tests are failing on too-old glib.  I can probably help dig into that.

q66:
which ones? i don't see any failing tests, other than those where configure fails because of glib being < 2.66

q66:
that said, it seems curl still needs some conversion to modern glib idioms, as the -Werror build is failing for that

q66:
ok, i pushed all that i wanted

cgwalters (approved):
This looks great to me, thank you so much!

I'll take a look at adjusting our CI for the new requirement.

PR #2557, opened 2022-03-04
apidoc: add missing page includes

lucab:
This fixes some missing sections in API reference, adding all the
relevant includes.


cgwalters (approved):

dbnicholson:
Nice! Whenever I build locally I see the pile of warnings and really want them to be fixed.

lucab:
There are still several warnings in the docs generation, related to other things that I didn't investigate.

But I was mostly bothered by not having some functions in https://ostreedev.github.io/ostree/reference/reference.html#api-index-full, and this should fix that.

PR #2560, opened 2022-03-07
test-prune: Read to the end of cut(1) output

smcv:
If we use head(1) to take only the first two lines, then cut(1) and
earlier pipeline entries are killed by SIGPIPE (if they have not already
terminated), and that's flagged as an error under `set -o pipefail`.
Use an equivalent sed command to take exactly the second line, but
without SIGPIPE.

---

The fact that I need to know this does not say great things about the Unix pipeline philosophy.

### DIFF
--- a/tests/test-prune.sh
+++ b/tests/test-prune.sh
@@ -350,7 +350,7 @@ tap_ok --commit-only and --delete-commit
 # Test --delete-commit when it creates orphaned commits
 reinitialize_commit_only_test_repo
 # get the current HEAD's parent on dev branch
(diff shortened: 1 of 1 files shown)


dbnicholson:
These types of stupid details are why I rarely use `pipefail` in my own scripts even though it's clearly the thing to do.

cgwalters (approved):
Nice, thanks!

> These types of stupid details are why I rarely use pipefail in my own scripts even though it's clearly the thing to do.

The alternative of ignoring errors from earlier commands is  way worse though.

PR #2563, opened 2022-03-10
tmpfiles: Create `/run/ostree`

cgwalters:
This is referenced by https://github.com/ostreedev/ostree-rs-ext/blob/9645cee4f29786ba51ae9d62a52eeef9230146fd/lib/src/globals.rs#L16
specifically used for the (container image) pull secret in
`/run/ostree/auth.json`.

Let's pre-create the directory so users don't have to.

Motivated by https://github.com/openshift/machine-config-operator/pull/3007#discussion_r824172564

### DIFF
--- a/tests/inst/src/sysroot.rs
+++ b/tests/inst/src/sysroot.rs
@@ -1,5 +1,8 @@
 //! Tests that mostly use the API and access the booted sysroot read-only.
 
+use std::os::unix::prelude::PermissionsExt;
+use std::path::Path;
+
 use anyhow::Result;
 use ostree_ext::prelude::*;
 use ostree_ext::{gio, ostree};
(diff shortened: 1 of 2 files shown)


jlebon (approved):

cgwalters:
As best I can tell, the FCOS CI Jenkins is not scheduling pods into the cluster.  The [underlying cluster](https://console-openshift-console.apps.ocp.fedoraproject.org/) seems OK.

The two people who would know the most about this (dustymabe and jlebon) are on PTO today.

Hmm, but [this job](https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/job/coreos-assembler/job/PR-2750/5/console) seems to at least have executed a build, but failed to push to the registry.  There's a variety of "coreos-assembler" pods that are running `sleep 24h` from...something?

Looking at [the system log](https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/log/all) I don't see anything obvious.  It does look like the default "openshift" jenkins-kubernetes cloud is at least present.

tests/inst/src/sysroot.rs
```diff
@@ -45,3 +48,13 @@ fn test_immutable_bit() -> Result<()> {
...
     Ok(())
 }
+
+#[itest]
+fn test_tmpfiles() -> Result<()> {
+    if skip_non_ostree_host() {
+        return Ok(());
+    }
+    let metadata = Path::new("/run/ostree").metadata()?;
+    assert_eq!(metadata.permissions().mode(), 0o755);
```
lucab, line 58: This failed in CI with:
```
kola-runext-ostree-test[1578]: thread '<unnamed>' panicked at 'assertion failed: `(left == right)`
kola-runext-ostree-test[1578]:   left: `16877`,
kola-runext-ostree-test[1578]:  right: `493`', tests/inst/src/sysroot.rs:58:5
```

cgwalters: Always nice to get those CI failures that let you know CI is working.

PR #2565, opened 2022-03-11
Add .clang-format file + tree-wide: Run `clang-format`

cgwalters:
lib: Fix one include

We need all the ostree bits here.

---

tree-wide: Run clang-format

This is a one-time tree wide reformatting to ensure consistency
going forward.

---

ci: Validate clang-format

---


cgwalters:
(This needs some work to compile, leaving as draft for discussion anyways since we should get outstanding patches in first)

jlebon:
:+1: from me. This is disruptive, but worth it long-term.

(Also, for anyone not familiar, check out `git log -S` if you don't use it yet. It'll often get you to the commit you're interested in faster than iteratively using `git annotate`.)

jlebon:
:+1: from me. This is disruptive, but worth it long-term.

(Also, for anyone not familiar, check out `git log -S` if you don't use it yet. It'll often get you to the commit you're interested in faster than iteratively using `git annotate`.)

jlebon:
If we do this, we probably also want to do the same thing as in https://github.com/coreos/rpm-ostree/pull/3604.

dbnicholson:
This is fine by me. I'd rather fix it all in one mega commit than hundreds of little fixes scattered in PRs.

dbnicholson:
https://github.com/marketplace/actions/clang-format-check seems like a drop in.

jlebon:
Looks like the impact to `git blame` can be minimized: https://mobile.twitter.com/pamelafox/status/1556779165247893506.

cgwalters:
OK, lifting draft on this.

cgwalters:
Ah exciting, looks like clang-format has different behavior between clang-15 and clang-16 (fedora 37 and 38 respectively).

dbnicholson:
I'm all in favor of this although I haven't looked at the details. The one thing I ask is that if the policy gets tweaked for any reason that it also gets captured in the `.editorconfig` file if possible.

ci/codestyle.sh
```diff
@@ -15,12 +15,11 @@ done
...
 
-# Will uncomment this once we reformat
-#if command -v clang-format; then
-#    echo -n "checking clang-format... "
-#    git ls-files '**.c' '**.cxx' '**.h' '**.hpp' | xargs clang-format --Werror --dry-run
-#    echo "ok"
-#fi
+if command -v clang-format; then
+    echo -n "checking clang-format... "
+    git ls-files '**.c' '**.cxx' '**.h' '**.hpp' | xargs clang-format --Werror --dry-run
```
jlebon, line 20: Having to copy-paste this invocation and modify it to apply `clang-format` the first time before it goes in your shell history isn't ideal. Let's add a `clang-format` make target for this? Like in https://github.com/coreos/rpm-ostree/commit/14a8b7c92235da56acb977e9ba5029006c3acb0b.

jlebon (approved):
We could also add a `.git-blame-ignore-revs` [like in rpm-ostree](https://github.com/coreos/rpm-ostree/commit/aea8432d71b15bfd12a611ea439cfe03f2397a36) to mitigate the impact of this change as part of this PR (this repo uses merge commits, so we already know what commit hash to ignore). Fine to add separately too.

ci/codestyle.sh
cgwalters: > let's add a clang-format make target for this? Like in https://github.com/coreos/rpm-ostree/commit/14a8b7c92235da56acb977e9ba5029006c3acb0b.

That already landed in https://github.com/ostreedev/ostree/blob/main/Makefile.am#L136

cgwalters:
Right exciting, and the Debian stable version is too old.  Added some logic that tries to skip in that case...

cgwalters:
> The one thing I ask is that if the policy gets tweaked for any reason that it also gets captured in the .editorconfig file if possible.

As I understand it, our `.editorconfig` is providing the baseline.  We could also try to keep things like [max line length](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length) in sync, but...I think we basically now expect people to use `make clang-format` when submitting patches in the general case.

dbnicholson:
> > The one thing I ask is that if the policy gets tweaked for any reason that it also gets captured in the .editorconfig file if possible.
> 
> As I understand it, our `.editorconfig` is providing the baseline. We could also try to keep things like [max line length](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#max_line_length) in sync, but...I think we basically now expect people to use `make clang-format` when submitting patches in the general case.

Sure, but the entire point of configuring the editor is so that it does it for you. Might as well make them consistent so you don't forget the extra step or have the CI bounce you back. It would be frustrating to have your editor fighting with the CI over how the code is supposed to look.

dbnicholson:
> Sure, but the entire point of configuring the editor is so that it does it for you. Might as well make them consistent so you don't forget the extra step or have the CI bounce you back. It would be frustrating to have your editor fighting with the CI over how the code is supposed to look.

Sorry, forgot one more part of this. EditorConfig seems to be less expressive than what you can do with `clang-format`. That's fine. All I'm asking is that they not be inconsistent.

.clang-format
```diff
@@ -2,3 +2,4 @@ Language: Cpp
 BasedOnStyle: GNU
 ColumnLimit: 100
 ForEachMacros: ['GLNX_HASH_TABLE_FOREACH', 'GLNX_HASH_TABLE_FOREACH_V', 'GLNX_HASH_TABLE_FOREACH_KV', 'GLNX_HASH_TABLE_FOREACH_IT']
+AlignEscapedNewlines: DontAlign
```
jlebon, line 5: Is this something we should add to the rpm-ostree one too?

jlebon (approved):

cgwalters:
> Sure, but the entire point of configuring the editor is so that it does it for you. Might as well make them consistent so you don't forget the extra step or have the CI bounce you back. It would be frustrating to have your editor fighting with the CI over how the code is supposed to look.

Right, editorconfig is a subset.  In the current setup (AFAIK) the editor will not fight with clang-format.  However, it's quite likely there are things that clang-format will want to fix that aren't handled by editorconfig (line length is an obvious one, but I'm sure there are more).  That's kind of inherent in all of this.

cgwalters:
OK I gave this some more spot checks all around; I didn't see anything that looked definitely wrong.  But, let's give this a day or so before we merge.

PR #2566, opened 2022-03-14
ci: Disable Ubuntu LTS

cgwalters:
It doesn't have a new enough glib.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -16,6 +16,7 @@ jobs:
   docs:
     name: Build documentation
     runs-on: ubuntu-latest
+    container: registry.ci.openshift.org/coreos/fcos-buildroot:testing-devel
     steps:
       - name: Checkout repository
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


q66:
should be able to re-enable it soon (as a new LTS will come in a month)

also, I believe some other things in the CI also rely on too old systems

.github/workflows/docs.yml
```diff
@@ -16,46 +16,15 @@ jobs:
...
-            libgpgme11-dev \
-            liblzma-dev \
-            libmount-dev \
-            libselinux1-dev \
-            libsoup2.4-dev \
-            libcurl4-openssl-dev \
-            procps \
-            zlib1g-dev \
-            python3-yaml
+        run: ./ci/installdeps.sh
```
jlebon, line 27: We should probably swallow those up into the buildroot image too. And similarly for rpm-ostree's `ci/install-test-deps.sh`.

jlebon (approved):

dbnicholson:
Wow, just missed the boat. Presumably, 22.04 will be released in a month, in which case it would be come both `latest` and `rolling`. Possibly we could switch this to `latest` and `devel` where the latter is unreleased and would never coincide with `latest`.

PR #2569, opened 2022-03-18
deploy: Try to rebuild policy in new deployment if needed

WOnder93:
Whenever the user has SELinux enabled and has any local
modules/modifications installed, it is necessary to rebuild the policy
in the final deployment, otherwise ostree will leave the binary policy
files unchanged from last deployment as it detects difference against
the base content (in rpm-ostree case this is the RPM content).

To avoid the situation where the policy binaries go stale once any local
customization of the policy is made, try to rebuild the policy as part
of sysroot_finalize_deployment(). Use the special
--rebuild-if-modules-changed switch, which detects if the input module
files have changed relative to last time the policy was built and skips
the most time-consuming part of the rebuild process if modules are
unchanged (thus making this a relatively cheap operation if the user
hasn't made any modifications to the shipped policy).

As suggested by Jonathan Lebon, this uses bubblewrap (via
g_spawn_sync()) to perform the rebuild inside the deployment's
filesystem tree, which also means that ostree will have a runtime
dependency on bubblewrap.

Partially addresses: https://github.com/coreos/fedora-coreos-tracker/issues/701

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2926,9 +2926,12 @@ sysroot_finalize_selinux_policy (int deployment_dfd, GError **error)
                           SEMODULE_HELP_ARGC, &exit_status, &stdout, error))
     return FALSE;
   if (!g_spawn_check_exit_status (exit_status, error))
-    return FALSE;
+    return glnx_prefix_error (error, "failed to run semodule");
   if (!strstr(stdout, "--rebuild-if-modules-changed"))
-    return TRUE;
+    {
+      ot_journal_print (LOG_INFO, "semodule does not have --rebuild-if-modules-changed");
+      return TRUE;
+    }
 
   static const gchar * const SEMODULE_REBUILD_ARGV[] = {
     "semodule", "-N", "--rebuild-if-modules-changed"
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2830,6 +2830,118 @@ get_var_dfd (OstreeSysroot      *self,
   return glnx_opendirat (base_dfd, base_path, TRUE, ret_fd, error);
 }
 
+#ifdef HAVE_SELINUX
+static void
+child_setup_fchdir (gpointer data)
+{
+  int fd = (int) (uintptr_t) data;
+  int rc __attribute__((unused));
+
+  rc = fchdir (fd);
+}
+
+/*
+ * Derived from rpm-ostree's rust/src/bwrap.rs
+ */
+static gboolean
+run_in_deployment (int deployment_dfd,
+                   const gchar * const *child_argv,
+                   gsize child_argc,
+                   gint *exit_status,
+                   gchar **stdout,
+                   GError **error)
+{
+  static const gchar * const COMMON_ARGV[] = {
+    "/usr/bin/bwrap",
+    "--dev", "/dev", "--proc", "/proc", "--dir", "/run", "--dir", "/tmp",
+    "--chdir", "/",
+    "--die-with-parent",
+    "--unshare-pid",
+    "--unshare-uts",
+    "--unshare-ipc",
+    "--unshare-cgroup-try",
+    "--ro-bind", "/sys/block",    "/sys/block",
+    "--ro-bind", "/sys/bus",      "/sys/bus",
+    "--ro-bind", "/sys/class",    "/sys/class",
+    "--ro-bind", "/sys/dev",      "/sys/dev",
+    "--ro-bind", "/sys/devices",  "/sys/devices",
+    "--bind", "usr", "/usr",
+    "--bind", "etc", "/etc",
+    "--bind", "var", "/var",
+    "--symlink", "/usr/lib",      "/lib",
+    "--symlink", "/usr/lib32",    "/lib32",
+    "--symlink", "/usr/lib64",    "/lib64",
+    "--symlink", "/usr/bin",      "/bin",
+    "--symlink", "/usr/sbin",     "/sbin",
+  };
+  static const gsize COMMON_ARGC = sizeof (COMMON_ARGV) / sizeof (*COMMON_ARGV);
+
+  gsize i;
+  GPtrArray *args = g_ptr_array_sized_new (COMMON_ARGC + child_argc + 1);
+  g_autofree gchar **args_raw = NULL;
+
+  for (i = 0; i < COMMON_ARGC; i++)
+    g_ptr_array_add (args, (gchar *) COMMON_ARGV[i]);
+
+  for (i = 0; i < child_argc; i++)
+    g_ptr_array_add (args, (gchar *) child_argv[i]);
+
+  g_ptr_array_add (args, NULL);
+
+  args_raw = (gchar **) g_ptr_array_free (args, FALSE);
+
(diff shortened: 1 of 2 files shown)


WOnder93:
BTW, I did test this in a dummy Fedora CoreOS VM where I reproduced the issue (basically by installing a dummy SELinux module and then doing `rpm-ostree rebase ...` that involved a `selinux-policy` package upgrade) and I was able to verify that the SELinux policy was up to date afterwards with this patch.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2866,6 +2929,17 @@ sysroot_finalize_deployment (OstreeSysroot     *self,
...
+    "semodule", "-N", "--rebuild-if-modules-changed"
+  };
+  gint exit_status;
+  if (!run_in_deployment(deployment_dfd, SEMODULE_CMD,
+                         sizeof(SEMODULE_CMD) / sizeof(*SEMODULE_CMD),
+                         &exit_status, error))
+    return FALSE;
+  if (!g_spawn_check_exit_status(exit_status, NULL))
+    g_message ("Failed to refresh SELinux policy - the policy contents may be inconsistent");
+
```
jlebon, line 2942: I think we need to gate all this on `HAVE_SELINUX`.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2866,6 +2929,17 @@ sysroot_finalize_deployment (OstreeSysroot     *self,
         return FALSE;
     }
 
+  static const gchar * const SEMODULE_CMD[] = {
+    "semodule", "-N", "--rebuild-if-modules-changed"
+  };
+  gint exit_status;
+  if (!run_in_deployment(deployment_dfd, SEMODULE_CMD,
```
jlebon, line 2936: Convention in this repo is

```suggestion
  if (!run_in_deployment (deployment_dfd, SEMODULE_CMD,
```

for function calls.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2866,6 +2929,17 @@ sysroot_finalize_deployment (OstreeSysroot     *self,
...
+                         sizeof(SEMODULE_CMD) / sizeof(*SEMODULE_CMD),
+                         &exit_status, error))
+    return FALSE;
+  if (!g_spawn_check_exit_status(exit_status, NULL))
+    g_message ("Failed to refresh SELinux policy - the policy contents may be inconsistent");
```
jlebon, line 2941: I guess you made this non-fatal to handle ratcheting it in? Let's figure out instead how to auto-detect it so we can tighten this. Two obvious options:
- Feature detection at runtime: e.g. search for switch in output of `semodule --help`? This would be best I think but can get quite verbose.
- Compile-time version check: add autoconf goop to detect if `libselinux` is high enough? Hmm, it looks like on Fedora `semodule` is part of `policycoreutils`. But they're all versioned together, right?

jlebon (commented):
Awesome, thanks a lot for working on this!

It looks like this is updating the libglnx submodule unintentionally.

src/libostree/ostree-sysroot-deploy.c
WOnder93: The second option won't work, because we depend on a given `semodule` version *inside* the deployment, which may not match the runtime where the `ostree` is running (may be an older/newer/completely different deployment).

The first option I had thought off, it just didn't seem worth it. But since you request it, I'll try to do it this way.

src/libostree/ostree-sysroot-deploy.c
WOnder93: I guess you're right.. While technically the code doesn't depend on libselinux, if ostree isn't built with `HAVE_LIBSELINUX`, then it can't relabel files in the deployment and thus attempting any selinux-policy management would be pointless anyway...

WOnder93: Wait, sorry, actually it's `HAVE_SELINUX`. not `HAVE_LIBSELINUX`, and it means simply "SELinux support", not "use libselinux", so it's even more obvious that this should be under `HAVE_SELINUX` :)

cgwalters (commented):
Thank you so much for working on this!

cgwalters:
Tangentially related but I want to mention; on the coreos side we've been pulling things in a whole new direction with https://github.com/coreos/enhancements/blob/main/os/coreos-layering.md

(There the policy rebuild would happen in a container and be applied centrally, not happen per host/machine)

src/libostree/ostree-sysroot-deploy.c
WOnder93: I already went with scanning the --help output, which turned out to be not that verbose nor ugly. Please have a look and let me know if it works for you.

It's not entirely clear to me how it would work with the env variable, but the idea doesn't sound very appealing to me... Yet, it's not up to me so I'll submit to whatever you guys prefer :)

cgwalters:
Hmm, the transactionality test failed...but passed on rerun.  Unfortunately it seems we don't save logs from the failing case, grr.

Offhand...it seems like this should be transactional.  But let me see if I can replicate any failures locally.

cgwalters:
Our merge process doesn't squash `fixup!` so I went ahead and did that and also rebased to pick up the fixes to the test suite, ran locally and seems to work.

Added a commit on top for more logging.  Now that I look, we don't have the updated `semodule` in CI here.  I'm looking at trying to use FCOS `next-devel` i.e. F36 but running into another issue...

In the meantime though I think we can merge this.

cgwalters (approved):

WOnder93:
> Our merge process doesn't squash `fixup!` so I went ahead and did that and also rebased to pick up the fixes to the test suite, ran locally and seems to work.

Then maybe [CONTRIBUTING.md](https://github.com/ostreedev/ostree/blob/main/docs/CONTRIBUTING.md#submitting-patches) needs updating? ;)

> In the meantime though I think we can merge this.

:tada:

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2830,6 +2830,69 @@ get_var_dfd (OstreeSysroot      *self,
...
+    "--ro-bind", "/sys/bus",      "/sys/bus",
+    "--ro-bind", "/sys/class",    "/sys/class",
+    "--ro-bind", "/sys/dev",      "/sys/dev",
+    "--ro-bind", "/sys/devices",  "/sys/devices",
+    "--bind", "usr", "/usr",
+    "--bind", "etc", "/etc",
+    "--bind", "var", "/var",
+    "--symlink", "/usr/lib",      "/lib",
+    "--symlink", "/usr/lib32",    "/lib32",
+    "--symlink", "/usr/lib64",    "/lib64",
```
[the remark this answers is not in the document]
dbnicholson: Coming in to this late, but why not do an `--ro-bind` of `.` to `/` followed by the writable binds for `/usr`, `/etc` and `/var`? Presumably the usr-merge symlinks are already setup properly in the deployment.

cgwalters:
Just to xref, https://github.com/openshift/os/pull/962/commits/e9741465f35ac9e38352d4fc35176f4d8c4fe7fa landed which is a worse version of this - in particular the policy is only rebuilt early during the boot, at which point we may already be subject to problems with drift.  This PR rebuilds *before* booting, which avoids those issues.

Issue #2427, opened 2021-08-27
Proposal: Merge github.com/ostreedev/ostree and github.com/ostreedev/ostree-rs{,-ext} into this repo


cgwalters:
This is a followup to https://mail.gnome.org/archives/ostree-list/2021-April/msg00000.html

So far I think the model has been pretty successful (at least for us in rpm-ostree land).

But...I'm already hitting metal complexity + ergonomic issues in dealing with 3 git repos.  For example, 

# Proposed: Create a `2021-lts` git branch

Take the current state of this repo, and we will create a "long term support" branch.  Anyone who still cannot take a hard dependency on Rust in 2022 can use this branch.

Or, if you are mostly satisfied with the existing feature set and want stability, you can follow this branch.  If we execute on this proposal, I'd plan to use this for RHEL8.

# Proposed: Fold ostree-rs and ostree-rs-ext into main 

In the short term, the `main` branch still will support the existing C library and binaries as is.

However, in the medium term (e.g. somewhere early in 2022?) we actively take a hard dependency on Rust (and `cargo` etc) to build and run.

Benefits:

- Being able to version + release a single thing
- Single test suite infrastructure (right now bits are duplicated across multiple repos)
- We can write new key functionality in Rust

# What about non-Rust consumers?

A key question here is whether `libostree-1.so` depends on Rust.  If we do, it means that projects like rpm-ostree will end up linking in Rust code twice, which is going to lead to bloat and problems.

In this model, we would probably need to do soname bumps corresponding to the Rust crate semver bumps.  

One possibility here is that the shared library calls back into Rust via weak symbols.  So the Rust code gets compiled in the consumer binary (e.g. rpm-ostree).

AdrianVovk:
Would the gobject-introspection api stick around or continue to get extended with new features? I'm not interested in porting os-updated to rust...

cgwalters:
Thanks for the feedback! 

> Would the gobject-introspection api stick around or continue to get extended with new features? I'm not interested in porting os-updated to rust...

Yeah, I was trying to touch on this in the last proposal but did so indirectly.

I think a pivotal question indeed is "do we break the libostree API or not".  I had been leaning towards not breaking the API, and this feedback strengthens that.  (Also if we *did* break the C API this would be a parallel-installable library, but let's keep that out of scope for now)

jlebon:
[...]

PR #2575, opened 2022-03-31
Merge ostree-rs into this repository

cgwalters:
Starting the ball rolling on https://github.com/ostreedev/ostree/issues/2427

Implemented by following https://stackoverflow.com/questions/1425892/how-do-you-merge-two-git-repositories - specifically `git filter-repo --to-subdirectory-filter rust`.  So if you have a different opinion on what the subdirectory should be named (`rust/bindings`? in prep for merging ostree-rs-ext too?) now's the time to say!

I pushed one small commit on top which fixes up the build system, but that's it so far.


jmarrero:
Would ostree-rs-ext live under the same `rust` dir? If so `rust/lib/src`? and `rust/bindings/src`? or both merged under `rust/src`?

cgwalters:
> Would ostree-rs-ext live under the same rust dir? If so rust/lib/src? and rust/bindings/src? or both merged under rust/src?

Yeah, we should figure this out now.  Hmm...how about `rust-bindings/` for ostree-rs and `rust-ext` for ostree-ext?

(The thing is this topic really opens the question of the C library hard depending on Rust code, and that adds another wrinkle to this)  I guess that could just be `rust`?

cgwalters:
OK a few things here.  First a [rhel8 branch exists](https://github.com/ostreedev/ostree/tree/rhel8) that...well, we could rename it `2022-branch` or something.  ("stable" is IMO something that becomes meaningless over time, what happens with two stable branches etc.)

> It'd be good I think to hold off on that until we sort out if/when we could have libostree hard-require Rust.

Yeah.  I think I'd like to say that we (the upstream/git main) can, and anyone who can't follow can use the rhel8/2022-branch for now.  We (upstream) will still maintain the 2022 branch for stability and security fixes.

It's not like we're going to go on a wholesale rewrite anytime soon, so even if we make Rust a hard requirement it's very likely that most things will be easily backportable.

cgwalters:
OK, I've reworked this to be `rust-bindings/` and fixed `cargo fmt` incorrectly trying to reformat the `sys/` bits.

As of now, this does *not* make ostree hard depend on Rust.  New features may start to depend on Rust.

rust-bindings/.github/workflows/rust.yml
```diff
@@ -0,0 +1,68 @@
+name: Rust
```
jlebon, line 1: Don't we need to lift this up to `.github/workflows`?

jlebon (commented):
This looks sane to me overall. I was expecting to see at least some CI added that builds the bindings.

cgwalters:
>  I was expecting to see at least some CI added that builds the bindings.

Yeah, fair.  Done!

jlebon (approved):

cgwalters:
OK, test run had a new clearly unrelated flake: https://github.com/coreos/fedora-coreos-tracker/issues/1192

There's going to be a lot of followup to this, but let's get this ball rolling!

PR #2577, opened 2022-04-09
Publish man pages

dbnicholson:
I'm sure this could be polished more (the systemd setup is a solid reference), but this publishes the manpages in the hosted ostree docs. I'm a little unsure the build system integration is the best. Right now it requires a separate configure option and is part of a normal automake target, but I also have a much simpler version that just adds the custom `manhtml` target.

### DIFF
--- a/man/ostree.xml
+++ b/man/ostree.xml
@@ -151,7 +151,7 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <term><citerefentry><refentrytitle>ostree-admin-cleanup</refentrytitle><manvolnum>1</manvolnum></citerefentry></term>
                 
                 <listitem><para>
-                    &nbsp;Delete untagged
+                    Delete untagged
                     deployments and repository objects.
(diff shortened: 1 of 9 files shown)


dbnicholson:
This is up at https://dbnicholson.github.io/ostree/man/ if you want a preview. As you can see, it's pretty simplistic, but I think it's good enough to start with.

man/ostree.xml
```diff
@@ -151,7 +151,7 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <term><citerefentry><refentrytitle>ostree-admin-cleanup</refentrytitle><manvolnum>1</manvolnum></citerefentry></term>
                 
                 <listitem><para>
-                    &nbsp;Delete untagged
+                    Delete untagged
```
cgwalters, line 154: Agree, seems like this is likely to be a non-breaking change.

cgwalters (approved):
Awesome!! Thanks so much!

PR #2580, opened 2022-04-19
s390x: generate sd-boot at its own partition

nikita-dubrovskii:
SE-qcow2 image comes with extra partition for `sd-boot`.  
Required by: https://github.com/coreos/coreos-assembler/pull/2822

### DIFF
--- a/src/libostree/ostree-bootloader-zipl.c
+++ b/src/libostree/ostree-bootloader-zipl.c
@@ -21,12 +21,17 @@
 #include "ostree-bootloader-zipl.h"
 #include "ostree-deployment-private.h"
 #include "otutil.h"
+#include <sys/mount.h>
+#include <sys/stat.h>
 #include <string.h>
 
-#define SECURE_EXECUTION_BOOT_IMAGE     "/boot/sd-boot"
+#define SECURE_EXECUTION_PARTITION      "/dev/disk/by-label/se"
+#define SECURE_EXECUTION_MOUNTPOINT     "/sysroot/se"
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-bootloader-zipl.c
```diff
@@ -67,6 +72,25 @@ _ostree_bootloader_zipl_get_name (OstreeBootloader *bootloader)
   return "zipl";
 }
 
+static gboolean
+_ostree_secure_execution_mount(GError **error)
+{
+  const char *device = realpath (SECURE_EXECUTION_PARTITION, NULL);
+  if (device == NULL)
+    return glnx_throw_errno_prefix(error, "s390x SE: resolving %s", SECURE_EXECUTION_PARTITION);
+  if (mount (device, SECURE_EXECUTION_MOUNTPOINT, "ext4", 0, NULL) < 0)
```
cgwalters, line 81: I think this is OK as is, but one thing we could consider here is reworking the zipl invocation to be a systemd unit we activate.  In that case, we can start using all the systemd features, like device dependencies, handling mount namespaces etc.

cgwalters (approved):

src/libostree/ostree-bootloader-zipl.c
nikita-dubrovskii: I'm thinking about moving all SecExe related stuff into zipl itself. Maybe a wrapper within `s390-tools`. 
Wrapper (without any bash scripts) exists now as part of proposed [PR](https://github.com/coreos/coreos-installer/pull/851) for coreos-installer

cgwalters (approved):

PR #2583, opened 2022-04-20
fetcher/curl: Consistently check return value `curl_easy_setopt`

cgwalters:
Static analyzers don't like when we only check it sometimes.

And we definitely want to know if any of these are failing.


dbnicholson (approved):
Kinda seems like you could make a wrapper since it's always asserting the same value, but meh.

cgwalters:
CI failure is unrelated

PR #2584, opened 2022-04-21
ci: Mask zincati for synthetic update

cgwalters:
Our CI started falling over because coreos-assembler checks
for units stuck activating as of https://github.com/coreos/coreos-assembler/pull/2810

Really need to centralize the code for this and fix the root
problem, but...not today.

xref https://github.com/coreos/coreos-assembler/pull/2814

### DIFF
--- a/tests/kolainst/destructive/staged-deploy.sh
+++ b/tests/kolainst/destructive/staged-deploy.sh
@@ -40,6 +40,9 @@ EOF
     cd /ostree/repo/tmp
     # https://github.com/ostreedev/ostree/issues/1569
     ostree checkout -H ${commit} t
+    # xref https://github.com/coreos/coreos-assembler/pull/2814
+    systemctl mask --now zincati
+    # Create a synthetic commit for upgrade
     ostree commit --no-bindings --parent="${commit}" -b staged-deploy -I --consume t
     newcommit=$(ostree rev-parse staged-deploy)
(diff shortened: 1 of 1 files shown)


dbnicholson (approved):

dbnicholson:
Looking at the reference, it would be nice to teach `ostree commit` how to retain metadata from a commit. I've done that in python many times and `flatpak build-commit-from` does that too. There are generated keys to skip and it wouldn't always work since the metadata might be too context dependent, but for simple cases it could be nice.

lucab:
Indeed I think that's a quite common usecase. Do we prefer to specify as input an explicit keys to retain, or a set of keys to ignore, or some kind of pattern/regexp (in whichever direction)? Is there a similar user-based metadata filtering mechanism already somewhere in ostree, in order to keep some kind of UX consistency?

dbnicholson:
Also, thanks for fixing the CI! There's little more annoying that seeing your PR failed, digging around to find out what the error was and seeing something that probably wasn't (but maybe not?) caused by you.

jlebon:
There is `--keep-metadata` for this (see https://github.com/ostreedev/ostree/pull/1402). But I think another thing we should probably fix is to have Zincati not infinitely get restarted.

One idea is to exit with a specific error code if it detects this (and other non-recoverable situations) and use `RestartPreventExitStatus=`.

PR #2588, opened 2022-04-21
docs: Add new IMA document

cgwalters:
Now that the fixed code for `ima-sign` landed in
https://github.com/ostreedev/ostree-rs-ext/pull/283

### DIFF
--- a/docs/ima.md
+++ b/docs/ima.md
@@ -25,14 +25,17 @@ files would (depending on policy) not be readable or executable.
 ## IMA signatures and OSTree checksum
 
 Mechanically, IMA signatures appear as a `security.ima` extended attribute
-on the file.   This is a signed digest of just the file content.
+on the file.   This is a signed digest of just the file content (and not
+any metadata)
 
-OSTree has first-class support for extended attributes;
-they are included in the object digest along with other key file attributes
-such as uid, gid and mode.
+OSTree's checksums in contrast include not just the file content, but also 
+metadata such as uid, gid and mode and extended attributes;
 
 Together, this means that adding an IMA signature to a file in the OSTree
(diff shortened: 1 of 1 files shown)


cgwalters:
:thinking: Hmm, I could imagine actually an ostree mode or format that actually *replaces* the ostree checksum with the (portable) IMA EVM checksum or so.  Would be quite nontrivial though.

docs/ima.md
```diff
@@ -0,0 +1,111 @@
...
+## Further references
+
+- https://sourceforge.net/p/linux-ima/wiki/Home/
+- https://wiki.gentoo.org/wiki/Integrity_Measurement_Architecture
+- https://fedoraproject.org/wiki/Changes/Signed_RPM_Contents
+- https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/managing_monitoring_and_updating_the_kernel/enhancing-security-with-the-kernel-integrity-subsystem_managing-monitoring-and-updating-the-kernel
+
+
+###### Licensing for this document:
+`SPDX-License-Identifier: (CC-BY-SA-3.0 OR GFDL-1.3-or-later)`
```
jlebon, line 111: If you just want it in the markdown but unrendered, you can include it at the top in HTML quotes like this: `<!--- SPDX-License-Identifier: (CC-BY-SA-3.0 OR GFDL-1.3-or-later) --->` .

jlebon (approved):
Cool, this is very informative!

docs/ima.md
cgwalters: This is in other documents, but I agree a comment is better.

jlebon (approved):

Issue #1576, opened 2018-05-04
display failure to finalize staged deployments


cgwalters:
Copying [this comment](https://github.com/projectatomic/rpm-ostree/pull/1321#issuecomment-385062253) from rpm-ostree discussion:

>  We'll need to think about how to display failures in status on reboot (back into the same deployment). Maybe ostree-finalize-staged could drop a marker file in e.g.` /ostree/deploy/$osname/deploy/$csum.$i.unfinalized`? And then we can print out something like:
> `Check logs using ``journalctl -b -1 -u ostree-finalize-staged.service``.`

PR #2589, opened 2022-04-23
Add an `ostree-boot-complete.service` to propagate staging failures

cgwalters:
Quite a while ago we added staged deployments, which solved
a bunch of issues around the `/etc` merge.  However...a persistent
problem since then is that any failures in that process that
happened in the *previous* boot are not very visible.

We ship custom code in `rpm-ostree status` to query the previous
journal.  But that has a few problems - one is that on systems
that have been up a while, that failure message may even get
rotated out.  And second, some systems may not even have a persistent
journal at all.

A general thing we do in e.g. Fedora CoreOS testing is to check
for systemd unit failures.  We do that both in our automated tests,
and we even ship code that displays them on ssh logins.  And beyond
that obviously a lot of other projects do the same; it's easy via
`systemctl --failed`.

So to make failures more visible, change our `ostree-finalize-staged.service`
to have an internal wrapper around the process that "catches" any
errors, and copies the error message into a file in `/boot/ostree`.

Then, a new `ostree-boot-complete.service` looks for this file on
startup and re-emits the error message, and fails.

It also deletes the file.  The rationale is to avoid *continually*
warning.  For example we need to handle the case when an upgrade
process creates a new staged deployment.  Now, we could change the
ostree core code to delete the warning file when that happens instead,
but this is trying to be a conservative change.

This should make failures here much more visible as is.

Closes: https://github.com/ostreedev/ostree/issues/1576


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3478,6 +3478,65 @@ _ostree_sysroot_finalize_staged (OstreeSysroot *self,
...
+    return FALSE;
+
+  glnx_autofd int failure_fd = -1;
+  if (!ot_openat_ignore_enoent (self->boot_fd, _OSTREE_FINALIZE_STAGED_FAILURE_PATH, &failure_fd, error))
+    return FALSE;
+  // If we didn't find a failure log, then there's nothing to do right now.
+  // (Actually this unit shouldn't even be invoked, but we may do more in the future)
+  if (failure_fd == -1)
+    return TRUE;
+  g_autofree char *failure_data = glnx_fd_readall_utf8 (failure_fd, NULL, cancellable, NULL);
```
jlebon, line 3532: Shouldn't this be

```suggestion
  g_autofree char *failure_data = glnx_fd_readall_utf8 (failure_fd, NULL, cancellable, error);
```

?

jlebon (approved):
Some minor comments, but looks sane to me overall.

I think a follow-up to this is to have the `rpm-ostree status` code use this instead so it's more reliable. Let's file something for that?

lucab:
A general concern with this trend of relying on `/boot` for more primitives is that it is generally a small partition and in many cases may be on a non-journaled filesystem. Though this usage seems small enough and hopefully won't cause troubles, so LGTM.

src/libostree/ostree-sysroot-deploy.c
cgwalters: Good catch!

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3478,6 +3478,65 @@ _ostree_sysroot_finalize_staged (OstreeSysroot *self,
...
+/* Invoked at shutdown time by ostree-finalize-staged.service */
+gboolean
+_ostree_sysroot_finalize_staged (OstreeSysroot *self,
+                                 GCancellable  *cancellable,
+                                 GError       **error)
+{
+  g_autoptr(GError) local_error = NULL;
+  if (!_ostree_sysroot_ensure_boot_fd (self, error))
+    return FALSE;
+  if (!_ostree_sysroot_finalize_staged_inner (self, cancellable, &local_error))
```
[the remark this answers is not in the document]
cgwalters: I cleaned this bit up

cgwalters:
> A general concern with this trend of relying on /boot for more primitives is that it is generally a small partition and in many cases may be on a non-journaled filesystem. Though this usage seems small enough and hopefully won't cause troubles, so LGTM.

I guess we could have this somewhere in `/sysroot` too.  But my rationale for `/boot` is this problem only happens when we need to change that.  And yeah, it is going to be much smaller than kernels and initramfsen.

jlebon (approved):

Issue #2592, opened 2022-04-30
OstreeRepoAutoTransaction used with uninitialized refcount in ostree_repo_prepare_transaction()


smcv:
While trying to port Flatpak to FUSE 3 (https://github.com/flatpak/flatpak/pull/4326), I needed to upgrade the version of libostree used to build it in CI from 2020.8 to a version that uses FUSE 3, and tried upgrading to 2022.2. The build that is using FUSE 3 happens to be the one that also runs the tests under valgrind.

Having done so, several tests fail like this:

```
+ env G_SLICE=always-malloc valgrind -q --leak-check=no --error-exitcode=1 --gen-suppressions=all --num-callers=30 --suppressions=/home/runner/work/flatpak/flatpak/_build/../tests/flatpak.supp --suppressions=/home/runner/work/flatpak/flatpak/_build/../tests/glib.supp flatpak build-update-repo --gpg-homedir=/tmp/test-flatpak-5K7eYm/gpghome --gpg-sign=7B0961FD repos/test
--68438-- WARNING: unhandled amd64-linux syscall: 315
--68438-- You may be able to write your own handler.
--68438-- Read the file README_MISSING_SYSCALL_OR_IOCTL.
--68438-- Nevertheless we consider this a bug.  Please report
--68438-- it at http://valgrind.org/support/bug_reports.html.
Updating appstream branch
==68438== Conditional jump or move depends on uninitialised value(s)
==68438==    at 0x49F19E5: _ostree_repo_auto_transaction_unref (ostree-repo.c:832)
==68438==    by 0x4A068B7: glib_autoptr_clear_OstreeRepoAutoTransaction (ostree-repo-private.h:555)
==68438==    by 0x4A068B7: glib_autoptr_cleanup_OstreeRepoAutoTransaction (ostree-repo-private.h:555)
==68438==    by 0x4A068B7: ostree_repo_prepare_transaction (ostree-repo-commit.c:1692)
==68438==    by 0x1FBE61: flatpak_repo_generate_appstream (in /home/runner/work/flatpak/flatpak/_build/flatpak)
==68438==    by 0x159FDC: flatpak_builtin_build_update_repo (in /home/runner/work/flatpak/flatpak/_build/flatpak)
==68438==    by 0x137CC4: main (in /home/runner/work/flatpak/flatpak/_build/flatpak)
==68438== 
{
   <insert_a_suppression_name_here>
   Memcheck:Cond
[...]

PR #2593, opened 2022-04-30
Fix abort-on-error behaviour of transactions

smcv:
* repo: Factor out _ostree_repo_auto_transaction_new()
    
    This will allow the direct allocation in
    ostree_repo_prepare_transaction() to be replaced with a call to this
    function, avoiding breaking encapsulation.

* repo: Correctly initialize refcount of temporary transaction
    
    Previously, the reference count was left uninitialized as a result of
    bypassing the constructor, and the intended abort-on-error usually
    wouldn't have happened.
    
    Fixes: 8a9737a "repo/private: move OstreeRepoAutoTransaction to a boxed type"
    Resolves: https://github.com/ostreedev/ostree/issues/2592

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1689,10 +1689,10 @@ ostree_repo_prepare_transaction (OstreeRepo     *self,
   g_debug ("Preparing transaction in repository %p", self);
 
   /* Set up to abort the transaction if we return early from this function.
-   * This needs to be manually built here due to a circular dependency. */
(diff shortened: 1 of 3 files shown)


smcv:
I think this is ready for review, but I'm marking it as a draft while I see whether it fixes the CI for flatpak/flatpak#4326.

smcv:
I *think* this is right, but I won't know whether it fixed my failing CI for at least another hour (running the Flatpak test-suite under valgrind is *slow*). https://github.com/flatpak/flatpak/runs/6240039656?check_suite_focus=true

smcv:
The Flatpak test-suite has finally passed under valgrind, and yes, a backport of these commits to 2022.2 does resolve the test failure.

dbnicholson (approved):

PR #2601, opened 2022-05-09
Add APIs to get xattrs from disk

cgwalters:
I'm aiming to do some more work on the Rust side around `fsck`
like functionality, and this is a useful primitive.  There isn't
a great Rust crate for xattrs, and I think it's better to share this
code.


src/libostree/libostree-devel.sym
```diff
@@ -20,7 +20,11 @@
    - uncomment the include in Makefile-libostree.am
 */
 
-
+LIBOSTREE_2022.4 {
+global:
+  ostree_fd_get_all_xattrs;
+  ostree_get_all_xattrs_at;
```
lucab, line 26: Naming related, could we try to reduce the distance between these two via a common prefix like `ostree_xattrs_get_all_{at,from_fs}` or `ostree_fsmeta_*`?

src/libostree/ostree-core.c
```diff
@@ -836,6 +836,48 @@ gboolean ostree_break_hardlink (int               dfd,
...
+  GVariant *ret = NULL;
+  if (!glnx_fd_get_all_xattrs (fd, &ret, cancellable, error))
+    return NULL;
+  return ret;
+}
+
+/**
+ * ostree_get_all_xattrs_at:
+ * @fd: File descriptor
+ * @cancellable: Cancellable
```
lucab, line 862: Minor: `@path` is missing right above this.

src/libostree/ostree-core.c
```diff
@@ -836,6 +836,48 @@ gboolean ostree_break_hardlink (int               dfd,
...
+ * @error: Error
+ *
+ * Retrieve all extended attributes in a canonical (sorted) order from
+ * the given path, relative to the provided directory file descriptor.
+ * Crucically, this must be used currently to retrieve extended attributes
```
lucab, line 867: Typo: `s/Crucically/Crucially/`

If I'm reading this correctly, I think this can be reworded in `If path is a symbolic link, it does not get dereferenced`. Although I wonder if we should future-proof this with an additional `int flags` argument.

src/libostree/libostree-devel.sym
cgwalters: Good suggestion, fixed that!

src/libostree/ostree-core.c
cgwalters: Addressed that except the `int flags` - not opposed to that or anything, but OTOH we're not a kernel either.  Adding new C API is much cheaper than new system calls.  I can't think what we'd ever add there that we wouldn't *also* want in the fd path?

lucab: In the ends it's a just a matter of design decisions.
Keeping this simpler without flags and adding new APIs if needed it's fine too for me :+1: .

cgwalters:
```

0s
Run ./autogen.sh --enable-gtk-doc --enable-man --enable-man-html
fatal: unsafe repository ('/__w/ostree/ostree' is owned by someone else)
To add an exception for this directory, call:

	git config --global --add safe.directory /__w/ostree/ostree
```
Exciting unrelated CI failure, will look

wmanley:
> I'm aiming to do some more work on the Rust side around `fsck` like functionality, and this is a useful primitive. There isn't a great Rust crate for xattrs, and I think it's better to share this code.

In ostree-oxide I used the [xattr](https://crates.io/crates/xattr/0.2.2) crate.

Regarding `fsck` like functionality I've played around with a parallel implementation of `ostree-prune` using `rayon`: https://github.com/ostreedev/ostree-oxide/blob/master/ostree-repo/src/prune.rs

Issue #2591, opened 2022-04-25
Changes needed in OSTree for grub password support 


saqibali-2k:
We want to land [GRUB password support](https://github.com/coreos/fedora-coreos-tracker/issues/134) in FCOS. That is, we want to: 
1. Gate users from accessing GRUB shell without a password 
2. Prevent them from booting into non-default ostree deployments without a password

The first part has been handled by sourcing in a `user.cfg` into `grub.cfg` (https://github.com/coreos/coreos-assembler/pull/2809) and adding sugar to butane which will create `user.cfg` in `/boot/grub2/` (https://github.com/coreos/butane/pull/339). 

The second part can be achieved via modifications to OSTree. A new key can be added to `ostree/repo/config` which will allow us to append to BLS configs in `/boot/loader/entries`. Specifically, we want to add the line `grub_users ""` (see [BootLoaderSpecByDefault](url)) to non-default entries. This key may be specified by default in [coreos-assembler](https://github.com/coreos/coreos-assembler/blob/72e915618e33dfabe420be19e681893f4e578a6e/src/create_disk.sh#L424-L427).

PR #2605, opened 2022-05-11
ostree-systroot-deploy: parse bls-append-except-default key

saqibali-2k:
We want to parse a new "bls-append-except-default" key from ostree config. The
key-value pairs specified by this key will be added to the generated
BLS fragments of non-default deployments. They must follow the format
"key1,value1;key2,value2" and so on.

This change will allow us to land GRUB password support in FCOS.

closes: https://github.com/ostreedev/ostree/issues/2591

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -404,6 +404,17 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>bls-append-except-default</varname></term>
+        <listitem><para>A semicolon seperated string list of key-value pairs. For example:
+        <literal>bls-append-except-default=key1=value1;key2=value2</literal>. These key-value 
+        pairs will be injected into the generated BLS fragments of the non-default deployments.
+        In other words, the BLS fragment of the default deployment will be unaffected by
+        <literal>bls-append-except-default</literal>.
+        </para>
+        </listitem>
+      </varlistentry>
+
     </variablelist>
 
   </refsect1>
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2082,6 +2082,50 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
   g_autofree char *options_key = ostree_kernel_args_to_string (kargs);
   ostree_bootconfig_parser_set (bootconfig, "options", options_key);
 
+  g_autoptr(GError) local_error = NULL;
+  GKeyFile *config = ostree_repo_get_config (repo);
+  gchar **read_values = g_key_file_get_string_list (config, "sysroot", "bls-append-except-default", NULL, &local_error);
+  /* We can ignore not found errors */
+  if (!read_values)
+    {
+      gboolean not_found = g_error_matches (local_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND) || \
+                           g_error_matches (local_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_GROUP_NOT_FOUND);
+      if (not_found)
+        {
+          g_clear_error (&local_error);
+        }
+      else
+        {
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return FALSE;
+        }
+    }
+
+  /* Only append to this BLS config if:
+   * - this is not the default deployment
+   */
+   /* If deployment was prepended, it is the new default */
+  gboolean is_new_default = (ostree_deployment_get_index (deployment) == 0);
+  gboolean allow_append = !is_new_default;
+  if (allow_append)
+    {
+      /* get all key value pairs in bls-append */
+      for (char **iter = read_values; iter && *iter; iter++)
+        {
+          const char *key_value = *iter;
+          const char *sep = strchr (key_value, '=');
+          if (sep == NULL)
+            {
+              glnx_throw (error, "bls-append-except-default key must be of the form \"key1=value1;key2=value2...\"");
+              return FALSE;
+            }
+          g_autofree char *key = g_strndup (key_value, sep - key_value);
+          g_autofree char *value = g_strdup (sep + 1);
+          ostree_bootconfig_parser_set (bootconfig, key, value);
+        }
+
+    }
+
   glnx_autofd int bootconf_dfd = -1;
   if (!glnx_opendirat (sysroot->boot_fd, bootconfdir, TRUE, &bootconf_dfd, error))
     return FALSE;


cgwalters:
Sorry that no one responded to the issue earlier.  This is a bit abstracted and while I understand the concrete use case of bootloader locking, I don't quite understand how `grub_users,` works.

Oh, I think I see - it's a comma separated list of options so this maps in the end to a BLS fragment:

`grub_users=` or so?

And then the GRUB BLS parser ends up synthesizing...OK I see the docs https://www.gnu.org/software/grub/manual/grub/grub.html#Security

OK.  If this is all correct I think it'd be useful to spell this out a bit more explicitly, link to that page etc.

Now onto the UX; I think it'd be a bit cleaner if `bls-append-except-default` was a *separate list* and not a boolean.  That way there's only a single key to set for this and not two, and the feature becomes more flexible.

Are there any use cases for `bls-append` *without* `bls-append-ignore-default`?  

Hmm...actually, and wouldn't this potentially break other bootloaders to inject something GRUB specific there?  Will others like zipl know to ignore this key?  I guess we can partially get out of that by only injecting this config into the repo config on architectures that use grub, but...kind of unclean.  I guess I'd strawman something like a prefix `x-grub-grub_users` or so, then other bootloaders know to ignore keys prefixed with `x-` that don't match them (e.g. `x-zipl` would only apply to zipl, etc.)

saqibali-2k:
Thanks for the reply Colin.

> Are there any use cases for `bls-append` _without_ `bls-append-ignore-default`?

Not at the moment. I decided to add `bls-append-ignore-default` key since most people would expect `bls-append` to add to every BLS fragment and not just the ones that correspond to non-default deployments.

>Now onto the UX; I think it'd be a bit cleaner if bls-append-except-default was a separate list and not a boolean. That way there's only a single key to set for this and not two, and the feature becomes more flexible.

That definitely will introduce more flexibility. I can switch to that approach.

> Hmm...actually, and wouldn't this potentially break other bootloaders to inject something GRUB specific there? Will others like zipl know to ignore this key? I guess we can partially get out of that by only injecting this config into the repo config on architectures that use grub, but...kind of unclean.

I thought this was not an issue since, as you mentioned, we don't inject GRUB specific details indiscriminately. I'm actually a little confused on why this would be considered unclean. In my head, a user incorrectly inserting GRUB details when their system is using zipl is similar to a user setting `sysroot.bootloader=foobar` - both are user errors. Thanks in advance on any corrections in my thinking.

cgwalters:
> I thought this was not an issue since, as you mentioned, we don't inject GRUB specific details indiscriminately. I'm actually a little confused on why this would be considered unclean. In my head, a user incorrectly inserting GRUB details when their system is using zipl is similar to a user setting sysroot.bootloader=foobar - both are user errors. Thanks in advance on any corrections in my thinking.

From the CoreOS model we have pre-generated disk images, it isn't the *user* that wrote `sysroot.bootloader`, nor the user that actually typed `ostree config sysroot.bls_append = "grub_users"`.

OK I took this thread to https://github.com/coreos/coreos-assembler/pull/2854#discussion_r871753848

cgwalters:
> That definitely will introduce more flexibility. I can switch to that approach.

Related to this, I'm a little uncertain about adding a config option for which we don't know of a valid use case.  I see the rationale in the *symmetry* of having an option that does it unconditionally, don't get me wrong.  

So my proposal is basically just `bls-append-except-default` as its own standalone list to start.  But, if you prefer to *also* add `bls-append`, I am also OK with that.

man/ostree.repo-config.xml
```diff
@@ -404,6 +404,17 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>bls-append-except-default</varname></term>
+        <listitem><para>A semicolon seperated list of key-value pairs. For example:
+        <literal>bls-append=key1,value1;key2,value2</literal>. These key-value pairs
```
jlebon, line 410: This needs to be updated.

jlebon, line 410: `GKeyFile`s support string lists natively. It might be simpler to document the key as a string list instead (which actually uses the same `;`-separated format documented here) and in the code we use [`get_string_list`](https://docs.gtk.org/glib/method.KeyFile.get_string_list.html) instead of parsing the string ourselves.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2082,6 +2082,47 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
 
+  g_autoptr(GError) local_error = NULL;
+  GKeyFile *config = ostree_repo_get_config (repo);
+  g_autofree char *read_value = g_key_file_get_string (config, "sysroot", "bls-append-except-default", &local_error);
+  /* We can ignore not found errors */
+  gboolean not_found = g_error_matches (local_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND) || \
+                       g_error_matches (local_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_GROUP_NOT_FOUND);
+  if (not_found) 
+    {
+      g_clear_error(&local_error);
```
jlebon, line 2093:
```suggestion
      g_clear_error (&local_error);
```

And a few places lower down.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2082,6 +2082,47 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+    {
+      g_propagate_error (error, g_steal_pointer (&local_error));
+      return FALSE;
+    }
+
+  /* Only append to this BLS config if:
+   * - this is not the default deployment
+   */
+   /* If deployment was prepended, it is the new default */
+  gboolean is_new_default = ostree_deployment_get_index(deployment) == 0;
```
jlebon, line 2105: Personally like wrapping these in parentheses for clarity. E.g.

```suggestion
  gboolean is_new_default = (ostree_deployment_get_index (deployment) == 0);
```

cgwalters:
CI failure is unrelated to this PR; it's fallout from our switching fcos to 36 and having a new gcc which has new exciting `-fanalyzer` failures which offhand look like false positives, but needs digging.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2082,6 +2082,44 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
   g_autofree char *options_key = ostree_kernel_args_to_string (kargs);
   ostree_bootconfig_parser_set (bootconfig, "options", options_key);
 
+  g_autoptr(GError) local_error = NULL;
+  GKeyFile *config = ostree_repo_get_config (repo);
+  gchar **read_values = g_key_file_get_string_list (config, "sysroot", "bls-append-except-default", NULL, &local_error);
+  /* We can ignore not found errors */
+  gboolean not_found = g_error_matches (local_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND) || \
```
cgwalters, line 2089: This is fine as is (I looked and it seems `g_error_matches` checks for `error == NULL`, but it's more idiomatic to check `if (read_values == NULL)` first.

IOW (sorry whitespace damage, typing in GH comments):

```
if (!read_values)
  {
 gboolean not_found = g_error_matches (local_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_KEY_NOT_FOUND) || \
                       g_error_matches (local_error, G_KEY_FILE_ERROR, G_KEY_FILE_ERROR_GROUP_NOT_FOUND);
  if (not_found)
    {
      g_clear_error (&local_error);
    }
  else
   {
      g_propagate_error (error, g_steal_pointer (&local_error));
      return FALSE;
   }
  }
```

man/ostree.repo-config.xml
```diff
@@ -404,6 +404,17 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>bls-append-except-default</varname></term>
+        <listitem><para>A semicolon seperated list of key-value pairs. For example:
+        <literal>bls-append-except-default=key1,value1;key2,value2</literal>. These key-value
```
cgwalters, line 410: Any reason to use `,` over e.g. `=` which would seem more obvious?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2082,6 +2082,44 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+    {
+      /* get all key value pairs in bls-append */
+      for (char **iter = read_values; iter && *iter; iter++)
+        {
+          const char *key_value = *iter;
+          const char *comma = strchr (key_value, ',');
+          if (comma == NULL)
+            glnx_throw (error, "bls-append key must be of the form \"key1,value1;key2,value2...\"");
+          const char *key = g_strndup (key_value, comma - key_value);
+          const char *value = g_strdup (comma + 1);
```
cgwalters, line 2117: These need `g_autofree`.

man/ostree.repo-config.xml
saqibali-2k: I'm personally used to thinking of key-value pairs as two item tuples (comma-separated and ordered pair). Which is why my go to was to use `,`. However, I'm not bound by it. I'm OK with switching to `=`

saqibali-2k:
Before this PR lands, I will make another PR to amend https://github.com/coreos/coreos-assembler/pull/2854. The use of `,` will be changed to `=`.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2082,6 +2082,47 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+  if (allow_append)
+    {
+      /* get all key value pairs in bls-append */
+      for (char **iter = read_values; iter && *iter; iter++)
+        {
+          const char *key_value = *iter;
+          const char *sep = strchr (key_value, '=');
+          if (sep == NULL)
+            glnx_throw (error, "bls-append key must be of the form \"key1=value1;key2=value2...\"");
+          g_autofree const char *key = g_strndup (key_value, sep - key_value);
```
cgwalters, line 2119: I think we need to drop the `const` here, since `free()` will warn if passed a `const` pointer.   Are we not getting compiler warnings here?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2082,6 +2082,47 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
+  gboolean allow_append = !is_new_default;
+  if (allow_append)
+    {
+      /* get all key value pairs in bls-append */
+      for (char **iter = read_values; iter && *iter; iter++)
+        {
+          const char *key_value = *iter;
+          const char *sep = strchr (key_value, '=');
+          if (sep == NULL)
+            glnx_throw (error, "bls-append key must be of the form \"key1=value1;key2=value2...\"");
```
cgwalters, line 2118: Missing a `return` here.

cgwalters (commented):
Looking close, though...CI is failing and I suspect the failures are legitimate.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2082,6 +2082,50 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
   g_autofree char *options_key = ostree_kernel_args_to_string (kargs);
   ostree_bootconfig_parser_set (bootconfig, "options", options_key);
 
+  g_autoptr(GError) local_error = NULL;
+  GKeyFile *config = ostree_repo_get_config (repo);
+  gchar **read_values = g_key_file_get_string_list (config, "sysroot", "bls-append-except-default", NULL, &local_error);
```
cgwalters, line 2087: Also while we're here...I think we should move all parsing this value into `reload_sysroot_config()` so that any failures in parsing the value are surfaced when the repo config is loaded, and not when we're going to install the kernel.

(This would have made the CI failure a lot more obvious, and also I think the unit tests would fail in the same way)

saqibali-2k:
> Digging into the kola logs:
> bls-append-except-default key must be of the form "key1=value1;key2=value2..."

Interesting how only that test is getting that issue...I would expect other tests to have the same issue.

cgwalters:
> OK, CI seems to be good now :)

Hmm, did you push any new code?  Do you think it was somehow a flake? That seems...unlikely to me.

src/libostree/ostree-sysroot-deploy.c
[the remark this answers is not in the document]
cgwalters: This one needs to be addressed too right?

cgwalters (approved):

cgwalters:
OK I'm fine to land as is but would be nice to address the two bits above as followup.  Thanks for the patch!

src/libostree/ostree-sysroot-deploy.c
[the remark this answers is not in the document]
travier: Addressed in https://github.com/ostreedev/ostree/pull/2666

travier: Addressed in https://github.com/ostreedev/ostree/pull/2666

PR #2608, opened 2022-05-15
ci: Drop gcc `-fanalyzer` by default

cgwalters:
sysroot: Quiet one gcc `-fanalyzer` warning

I don't think this can happen in practice, but add an assertion
to verify it.

---

ci: Drop gcc `-fanalyzer` by default

There's a pile of new warnings in GCC 12 from F36, I dug into
them but most of them seem inscrutable - I am not seeing the bugs.

For now to unblock further work, stop using this flag.

---

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -1267,7 +1267,7 @@ ostree_sysroot_get_booted_deployment (OstreeSysroot       *self)
 OstreeDeployment *
 ostree_sysroot_require_booted_deployment (OstreeSysroot *self, GError **error)
 {
-  g_return_val_if_fail (self->loadstate == OSTREE_SYSROOT_LOAD_STATE_LOADED, NULL);
+  g_assert (self->loadstate == OSTREE_SYSROOT_LOAD_STATE_LOADED);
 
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-sysroot.c
```diff
@@ -1896,6 +1896,7 @@ ostree_sysroot_simple_write_deployment (OstreeSysroot      *sysroot,
   /* tracks when we come across the booted deployment */
   gboolean before_booted = TRUE;
   gboolean before_merge = TRUE;
+  g_assert (deployments);
```
lucab, line 1899: I think in this case the compiler is complaining about the `NULL` that can be returned by the `g_return_val_if_fail()` here:
 https://github.com/ostreedev/ostree/blob/9cdd98125063734ad0c016d51f51a0c8e724645a/src/libostree/ostree-sysroot.c#L1303
Let's directly change that into an assertion maybe?

cgwalters: Ahh, nicely spotted!  I changed that and a few others.

cgwalters:
OK, looks like just a flake in one of the GH actions suites; this should be good to go!

PR #2610, opened 2022-05-17
ci: Move codestyle and min build to GH actions

cgwalters:
Part of general momentum to leave heavy lifting (e.g. VM tests)
to Jenkins and Prow and use the cheaper/faster GH actions for
plain builds.


cgwalters:
OK, this one is working now!

Issue #2586, opened 2022-04-21
Teach `--karg-delete` option to `ostree admin deploy`


jpds:
Feature request for a `ostree admin deploy --karg-delete=`. I recently had to remove a boot option from a deployment and the only workaround I had was to manually edit GRUB and then use `--karg-proc-cmdline`.

champtar:
This was implemented a while ago

PR #2612, opened 2022-05-19
Teach --karg-delete option to ostree admin deploy

HuijingHei:
See https://github.com/ostreedev/ostree/issues/2586

### DIFF
--- a/src/ostree/ot-admin-builtin-deploy.c
+++ b/src/ostree/ot-admin-builtin-deploy.c
@@ -41,6 +41,7 @@ static gboolean opt_no_prune;
 static gboolean opt_no_merge;
 static char **opt_kernel_argv;
 static char **opt_kernel_argv_append;
+static char *opt_kernel_argv_delete;
 static gboolean opt_kernel_proc_cmdline;
 static char *opt_osname;
 static char *opt_origin_path;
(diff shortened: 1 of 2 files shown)


travier (approved):

src/ostree/ot-admin-builtin-deploy.c
```diff
@@ -90,6 +92,18 @@ ot_admin_builtin_deploy (int argc, char **argv, OstreeCommandInvocation *invocat
     return FALSE;
   }
 
+  if (opt_kernel_arg_none && opt_kernel_argv_delete)
+  {
```
cgwalters, line 96: GNU style has two spaces for indent here.  But I think we'll eventually do https://github.com/ostreedev/ostree/pull/2565 anyways; feel free to fix in a later patch though!

cgwalters (approved):
Thanks, nice and simple!

bauen1:
Is it intentional, that `--karg-delete` can only be specified **once** ?

That is quite unlike `--karg-append` and makes it significantly less useful, additionally no warning (or rather error) is issued when it is specified multiple times.

cgwalters:
I see no reason not to support multiple invocations of `--karg-delete` indeed.

PR #2614, opened 2022-05-23
s390x: ensure SecureExecution is enabled before sd-boot generation

nikita-dubrovskii:
By design `SecureExecution` is enabled during CoreOS build. So if systems runs in non-SE mode, skip `sd-boot` generation. 

https://github.com/coreos/fedora-coreos-tracker/issues/1205

### DIFF
--- a/src/libostree/ostree-bootloader-zipl.c
+++ b/src/libostree/ostree-bootloader-zipl.c
@@ -25,6 +25,7 @@
 #include <sys/stat.h>
 #include <string.h>
 
+#define SECURE_EXECUTION_SYSFS_FLAG     "/sys/firmware/uv/prot_virt_guest"
 #define SECURE_EXECUTION_PARTITION      "/dev/disk/by-label/se"
 #define SECURE_EXECUTION_MOUNTPOINT     "/sysroot/se"
 #define SECURE_EXECUTION_BOOT_IMAGE     SECURE_EXECUTION_MOUNTPOINT "/sd-boot"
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -109,6 +110,14 @@ _ostree_bootloader_zipl_write_config (OstreeBootloader  *bootloader,
   return TRUE;
 }
 
+static gboolean _ostree_secure_execution_is_enabled (GCancellable *cancellable) {
+  gsize len = 0;
+  g_autofree char *data = glnx_file_get_contents_utf8_at (-1, SECURE_EXECUTION_SYSFS_FLAG, &len, cancellable, NULL);
+  if ( !data )
```
cgwalters, line 116: Can be fixed in a followup, but GNU C style has no spaces here, i.e. `if (!data)`

dustymabe:
Looks like he addressed the code review comments. 👍

dustymabe (approved):

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -109,6 +110,14 @@ _ostree_bootloader_zipl_write_config (OstreeBootloader  *bootloader,
   return TRUE;
 }
 
+static gboolean _ostree_secure_execution_is_enabled (GCancellable *cancellable) {
```
jlebon, line 113: I'm surprised at this signature (no `GError`). We need to make really sure here that we discern between "Secure Execution is disabled" (OK) and "couldn't read Secure Execution enablement" (hard error).

Is `/sys/firmware/uv/prot_virt_guest` always expected to exist on s390x machines, even if Secure Execution is disabled?

nikita-dubrovskii: no GError, becase `/sys/firmware/uv/prot_virt_guest` exists only on machines with enabled SecureExecution. so if `openat` fails, error isn't propagated to the caller

jlebon: Then ideally, to be strict we should handle `ENOENT` gracefully here but otherwise error out.

nikita-dubrovskii: https://github.com/ostreedev/ostree/pull/2615

Issue #2594, opened 2022-05-04
Duplicate "--os=OSNAME..." line in output of "ostree admin config-diff --help".


junjieyuan:
```text
[junjie@xps ~]$ ostree admin config-diff --help
用法：
  ostree admin config-diff [选项…]

Diff current /etc configuration versus default

帮助选项：
  -h, --help         显示帮助选项

应用程序选项：
  --os=OSNAME        Use a different operating system root than the current one
  --sysroot=PATH     Create a new OSTree sysroot at PATH
  --os=OSNAME        Use a different operating system root than the current one
  -v, --verbose      Print debug information during command processing
  --version          Print version information and exit

```

And ostree have two ```lf``` at end of output, is it by design?

lucab:
Thanks for the report! Good catch, this is definitely a bug in `ostree admin config-diff`.
I didn't check further, but I think this duplication may be happening because of this spurious call:
https://github.com/ostreedev/ostree/blob/7fffc2e91e47f868aa309fc9c7296970f5eb887e/src/ostree/ot-admin-builtin-diff.c#L42

PR #2616, opened 2022-05-25
Remove dup "--os=OSNAME" line in output of "ostree admin config-diff --help"

HuijingHei:
- Small fix for https://github.com/ostreedev/ostree/issues/2594
- Fix GNU style of two spaces for indent in https://github.com/ostreedev/ostree/pull/2612


HuijingHei:
https://github.com/ostreedev/ostree/issues/2594#issuecomment-1117574159
What @lucab said is totally correct, test locally with the patch, the issue is fixed

PR #2621, opened 2022-05-31
ci: Drop libsoup build in jenkins

cgwalters:
ci: Drop libsoup build in jenkins

GH actions is fast zero cost system that covers these build matrix
things well.  Let's keep our Fedora CI system doing more of the
qemu heavy lifting.

---


cgwalters:
OK I'm just dropping that patch for now.

PR #2622, opened 2022-05-31
3 CI patches

cgwalters:
ci: Drop an unused files

Cruft from previous CI flows.

---

ci: Add a codestyle.sh

This matches what we do in rpm-ostree.  Move the `glnx_fd_close`
bits out of `make syntax-check` in preparation for dropping that.

---

gitignore: Add ci-build/

Generated by our scripts.

---


cgwalters:
OK right, the current GHA builds don't require Rust, which is a property we actually want to keep.  Fixed.

jlebon:
`Tests / Debian Stable 32 bit` flaked on https://github.com/ostreedev/ostree/issues/2038.

PR #2624, opened 2022-06-02
Rename ostree-cmdprivate to drop out of introspection

cgwalters:
I was looking at our `.gir` and noticed we had the cmdprivate bits
because the pattern for excluding headers is `-private.h`, which
didn't match `cmdprivate.h`.


jmarrero (approved):
lgtm

PR #2625, opened 2022-06-02
Drop `OSTREE_BUILT_FEATURES` from introspection

cgwalters:
It inherently depends on the individual build, and can't
really be an official stable API for introspection users.
I've noticed the value of this flip flop when doing local builds.

I'm fairly certain no one is trying to use it from a higher level
language.

It'd probably make sense to even drop from the official C API,
but I'm trying to be conservative with that.


jmarrero (approved):
lgtm

PR #2627, opened 2022-06-03
Updates to rust-bindings/ infrastructure, and update to latest git

cgwalters:
rust-bindings: Delete committed gir files

These seem to have been accidentally committed to git; they should
not be here.  The Gir should be derived from source code.

---

rust-bindings: Use OSTree-1.0.gir from our own builddir

Now that the projects are unified.

---

rust-bindings: Update to latest git

---


jlebon (approved):
LGTM.

Given that the bindings are auto-generated, I think a follow-up we need here is adding CI to check that the bindings are not out of date. Edit: filed https://github.com/ostreedev/ostree/issues/2628.

PR #2632, opened 2022-06-06
lib/prune: speed up pruning by retrieving only commits

saqibali-2k:
After landing the new --commit-only functionality, we still noticed
exceedingly long pruning times in large repos. Lets add an optimization
that will only retrieve commit objects when --commit-only flag is used.

### DIFF
--- a/src/libostree/ostree-repo-prune.c
+++ b/src/libostree/ostree-repo-prune.c
@@ -189,7 +189,6 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
   return TRUE;
 }
 
-
 /**
  * ostree_repo_prune_static_deltas:
  * @self: Repo
@@ -437,8 +436,17 @@ ostree_repo_prune (OstreeRepo        *self,
         return FALSE;
     }
 
-  objects = ostree_repo_list_objects_set (self, OSTREE_REPO_LIST_OBJECTS_ALL | OSTREE_REPO_LIST_OBJECTS_NO_PARENTS,
+  if (commit_only)
+    {
+      if (!ostree_repo_list_commit_objects_starting_with (self, "", &objects, cancellable, error))
+        return FALSE; 
+    }
+  else 
+    {
+      objects = ostree_repo_list_objects_set (self, OSTREE_REPO_LIST_OBJECTS_ALL | OSTREE_REPO_LIST_OBJECTS_NO_PARENTS,
                                           cancellable, error);
+    }
+
   if (!objects)
     return FALSE;
 
@@ -508,9 +516,20 @@ ostree_repo_prune_from_reachable (OstreeRepo        *self,
   if (!lock)
     return FALSE;
 
-  g_autoptr(GHashTable) objects =
-    ostree_repo_list_objects_set (self, OSTREE_REPO_LIST_OBJECTS_ALL | OSTREE_REPO_LIST_OBJECTS_NO_PARENTS,
-                                  cancellable, error);
+  g_autoptr(GHashTable) objects = NULL;
+  OstreeRepoPruneFlags flags = options->flags;
+  gboolean commit_only = (flags & OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY) > 0;
+  if (commit_only) 
+    {
+      if (!ostree_repo_list_commit_objects_starting_with (self, "", &objects, cancellable, error))
+        return FALSE;  
+    } 
+  else
+    {
+      objects =
+        ostree_repo_list_objects_set (self, OSTREE_REPO_LIST_OBJECTS_ALL | OSTREE_REPO_LIST_OBJECTS_NO_PARENTS,
+                                      cancellable, error);
+    }
   if (!objects)
     return FALSE;
 
--- a/src/libostree/ostree-repo.h
+++ b/src/libostree/ostree-repo.h
@@ -1261,7 +1261,7 @@ void ostree_repo_commit_traverse_iter_cleanup (void *p);
  * OstreeRepoPruneFlags:
  * @OSTREE_REPO_PRUNE_FLAGS_NONE: No special options for pruning
  * @OSTREE_REPO_PRUNE_FLAGS_NO_PRUNE: Don't actually delete objects
- * @OSTREE_REPO_PRUNE_FLAGS_REFS_ONLY: Do not traverse individual commit objects, only follow refs
+ * @OSTREE_REPO_PRUNE_FLAGS_REFS_ONLY: Do not traverse individual commit objects, only follow refs for reachability calculations
  * @OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY: Only traverse commit objects.  (Since 2022.2)
  */
 typedef enum {


src/libostree/ostree-repo-prune.c
```diff
@@ -189,6 +189,44 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
...
+  if (!ostree_repo_list_refs (repo, NULL, &all_refs,
+                              cancellable, error))
+    return FALSE;
+  
+  /* Iterate over all refs and add commit objects to 
+   * table of prunable objects.
+   */
+  GLNX_HASH_TABLE_FOREACH (all_refs, const char *, ref)
+    {
+      const char *checksum = g_hash_table_lookup (all_refs, ref);
```
cgwalters, line 215: We can get the key and the value together with `GLNX_HASH_TABLE_FOREACH_KV`.

src/libostree/ostree-repo-prune.c
```diff
@@ -189,6 +189,44 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
...
+        return FALSE;
+    }
+
+  GLNX_HASH_TABLE_FOREACH (ret_objects, GVariant *, key)
+    {
+      GVariant *value = g_variant_new ("(b@as)", TRUE, g_variant_new_strv (NULL, 0));
+      /* transfer ownership */
+      g_hash_table_replace (*objects, g_variant_ref(key), 
+                            g_variant_ref_sink (value));
+    }
```
cgwalters, line 226: Can't we directly pass `*objects` to `ostree_repo_traverse_commit_with_flags()`?

src/libostree/ostree-repo-prune.c
```diff
@@ -515,10 +567,23 @@ ostree_repo_prune_from_reachable (OstreeRepo        *self,
     return FALSE;
 
   g_autoptr(GHashTable) objects = NULL;
-
-  if (!ostree_repo_list_objects (self, OSTREE_REPO_LIST_OBJECTS_ALL | OSTREE_REPO_LIST_OBJECTS_NO_PARENTS,
+  OstreeRepoPruneFlags flags = options->flags;
+  gboolean list_refs_only = flags & OSTREE_REPO_PRUNE_FLAGS_LIST_REFS_ONLY;
```
cgwalters, line 571: Very minor nit but I usually coerce bitflags to true booleans via e.g. `(flags & OSTREE_REPO_PRUNE_FLAGS_LIST_REFS_ONLY) > 0` or so; today `gboolean` is just a typedef for `int` and in C, anything non-zero is true...but it can be confusing to have its value be `4` or whatever.  Modern languages like Rust don't have this implicit coercion and have a true `bool` type and e.g. the [bitflags crate has contains()](https://docs.rs/bitflags/1.3.2/bitflags/example_generated/struct.Flags.html#method.contains) which is IMO definitely better.

cgwalters (commented):
A few quick minor things, but I'm thinking more about the non-superficial stuff.

src/libostree/ostree-repo-prune.c
cgwalters: Oh nevermind, I see...there's some weird historical ostree tech debt here.

cgwalters: https://github.com/ostreedev/ostree/pull/2634 - let's rebase this on that PR and do the same thing.

dbnicholson:
I think I'm getting confused by the recent prune options, but if you're not scanning all the commits in the repo, don't you risk corrupting it by removing objects that a commit requires?

src/libostree/ostree-repo-prune.c
```diff
@@ -189,6 +189,44 @@ _ostree_repo_prune_tmp (OstreeRepo *self,
   return TRUE;
 }
 
+static gboolean
+get_objects_from_refs_only (OstreeRepo   *repo,
+                            OstreeRepoCommitTraverseFlags flags,
+                            GHashTable   **objects,
```
cgwalters, line 195: This is OK as is, but I did change the new internal API https://github.com/ostreedev/ostree/pull/2635/files#diff-140b1d44c51cd554eceeef9700a401e26e72f310df9694d61c655b5d6f089e47R4949 to directly return the `GHashTable` which I think is cleaner.

src/ostree/ot-builtin-prune.c
```diff
@@ -191,6 +191,16 @@ ostree_builtin_prune (int argc, char **argv, OstreeCommandInvocation *invocation
...
   if (opt_commit_only)
     pruneflags |= OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY;
+  if (opt_refs_only && (opt_retain_branch_depth || opt_keep_younger_than))
+    /* The flag below will tell ostree to look only at refs when gathering
+     * objects. This will speed up operations when trying to truncate branch history.
+     *
+     * We don't want to blindly pass this flag. Consider branch: HEAD->A->B->C,
+     * if we delete commit B with --delete-commit flag, we will not be able to 
+     * retrieve orphaned commit C if we only look at refs. Thus C will not be pruned,
+     * which may be different from what the user expects.
```
cgwalters, line 201: So I think `opt_refs_only` is really the common case.  And from that PoV, we *are* changing behavior in this way by default.

Now, the fact that you're doing this in the CLI does limit the "blast radius" to some degree.  But I think it may lead to confusion if the CLI has special behavior not directly reflected in the API.

Hmmm.  Thinking through this...are there any cases where one wants `OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY` but *not* `OSTREE_REPO_PRUNE_FLAGS_LIST_REFS_ONLY`? 

To me what is special here is `OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY` - that's not something we do in a "normal prune" (as used by e.g. the default cleanup after `ostree admin upgrade`).

What if we dropped this public flag and turned on this behavior when `OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY` is provided?

src/libostree/ostree-repo-prune.c
saqibali-2k: I agree, thats a lot cleaner. I'll make that change :+1:

src/ostree/ot-builtin-prune.c
dbnicholson: That's what I was just thinking. I don't understand why a new flag is needed, but I've never really used these modes. It seems to me that `COMMIT_ONLY` should have one simple effect - it filters the objects in both the full set and the wanted set. That allows a big optimization by not having to traverse commits. Really you don't have to traverse anything - you can simply scan the repo for commit objects. Shouldn't that just be the behavior for `COMMIT_ONLY` without any other flags? Am I missing something?

saqibali-2k: > Thinking through this...are there any cases where one wants OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY but not OSTREE_REPO_PRUNE_FLAGS_LIST_REFS_ONLY

So currently, `OSTREE_REPO_PRUNE_FLAGS_LIST_REFS_ONLY` gathers all objects by traversing refs (*not* just commit objects, unless `--commit-only` is provided). In the case that a user would want to truncate the history of a branch (`ostree prune --refs-only --keep-younger-than=<DATE>`) , but they don't want to only delete commit objects, it would make sense to only gather objects via traversal of refs.

> What if we dropped this public flag and turned on this behavior when OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY is provided?

I'm OK with ignoring the case mentioned above and switching to only supporting this for `--commit-only`. I felt that supporting the new optimization for  `ostree prune --refs-only` instead of only `--commit-only` would help generalise it. If we were to remove `OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY` and instead just turn on the functionality when `--commit-only` is passed, we would have to disallow certain combinations of `--commit-only` and other flags. An example would be the case mentioned in the comment referenced above.

src/libostree/ostree-repo-prune.c
```diff
@@ -515,10 +567,23 @@ ostree_repo_prune_from_reachable (OstreeRepo        *self,
...
+    {
+       OstreeRepoCommitTraverseFlags traverse_flags = 0;
+      if (commit_only)
+        traverse_flags |= OSTREE_REPO_COMMIT_TRAVERSE_FLAG_COMMIT_ONLY;
+      if (!get_objects_from_refs_only (self, traverse_flags, &objects, cancellable, error))
+        return FALSE;      
+    } 
+  else
+    {
+      if (!ostree_repo_list_objects (self, OSTREE_REPO_LIST_OBJECTS_ALL | OSTREE_REPO_LIST_OBJECTS_NO_PARENTS,
```
dbnicholson, line 583: It seems to me that the much safer thing to do is to use `ostree_repo_list_commit_objects_starting_with` with `OSTREE_REPO_PRUNE_FLAGS_COMMIT_ONLY` and `ostree_repo_list_objects` otherwise. You're still scanning all the objects, but the set of objects is much smaller, which will significantly reduce memory pressure.

I feel like the idea that you can consider the full set of objects by only looking at refs is bound to cause corruption sooner or later.

jlebon:
That said, another path I think could be to rework the pruner to use the libostree APIs directly so we can build up our own reachables and do a single `ostree_repo_prune_from_reachable`. This would reimplement a bunch of code in `ot-builtin-prune.c` for handling depth-based and time-based options. Or maybe it's possible to fully express the operation in a single `ostree prune` invocation (and without going over the CLI arg limit).

But OTOH, I think the optimization introduced here would be really nice to have baked in.

src/ostree/ot-builtin-prune.c
saqibali-2k: > It seems to me that COMMIT_ONLY should have one simple effect - it filters the objects in both the full set and the wanted set. That allows a big optimization by not having to traverse commits. Really you don't have to traverse anything - you can simply scan the repo for commit objects. Shouldn't that just be the behavior for COMMIT_ONLY without any other flags? Am I missing something?

Currently, `--commit-only` filters the reachable set such that it only includes commit objects. However, it does not filter the set of all objects. We want the flag to do both by adding this change. At the same time, we want to avoid the cost of using readdir on objects that we don't need (thus why we are traversing refs).  The reason why I added another flag to the prune API is because I want to disallow certain cases ( I.E using the new optimization with `--delete-commit`). If we don't add a flag to turn the optimization off, then we need to disallow the use of `--commit-only` with `--delete-commit` which is a little awkward since we allowed that possibility in the last release.

Sorry if I'm over-explaining here - this is also just to clear my head.

dbnicholson: OK, I understand that. What I'm saying is that you can also achieve the filtering on the set of all objects by using `list_commits_starting_with` instead of `list_repo_objects`. That can be used without introducing flags or changing the definition of "all objects". However, it may not achieve enough of an optimization since you're still scanning all the objects.

src/libostree/ostree-repo-prune.c
dbnicholson: I was curious about `ostree_repo_list_objects` vs `ostree_repo_list_commit_objects_starting_with`. At least on our server with our production OS repo (as well as my laptop with `/var/lib/flatpak/repo`), it makes a big difference.

[Here](https://gist.github.com/dbnicholson/4c1898deae258b07d207aa7254ac51e0) is a test program that simply does the scan and prints the number objects it found. The results were surprising:

```
$ time ~/count-objects
Number of objects: 1835291

real	0m9.183s
user	0m7.850s
sys	0m1.332s
$ time ~/count-objects --commits
Number of commits: 660

real	0m1.095s
user	0m0.248s
sys	0m0.847s
```

That's a 9x speedup by throwing out all but the commit objects from the scan. I ran that a few times to make sure it was repeatable. So, I really think you should just try that before introducing new flags.

dustymabe: > @dbnicholson Thanks, that's instructive. We're going to try that same test on the Fedora repos to see how it compares and report back.

Here is the script:

```
$ cat ~/script.sh 
cd /mnt/fedora_koji_prod/koji/ostree/repo/
time ~/count-objects
time ~/count-objects --commits

cd /mnt/fedora_koji_prod/koji/compose/ostree/repo/
time ~/count-objects
time ~/count-objects --commits
```

and here are the results:

```
[dustymabe@compose-x86-01 ~][STG]$ bash -x ~/script.sh 
+ cd /mnt/fedora_koji_prod/koji/ostree/repo/
+ /home/fedora/dustymabe/count-objects
Number of objects: 22243384

real    2m32.035s
user    2m15.591s
sys     0m15.070s
+ /home/fedora/dustymabe/count-objects --commits
Number of commits: 25760

real    0m7.071s
user    0m1.392s
sys     0m5.489s
+ cd /mnt/fedora_koji_prod/koji/compose/ostree/repo/
+ /home/fedora/dustymabe/count-objects
Number of objects: 6579571

real    30m12.614s
user    0m37.902s
sys     0m8.075s
+ /home/fedora/dustymabe/count-objects --commits
Number of commits: 9006

real    0m1.752s
user    0m0.391s
sys     0m1.215s
```

> 
> One thing worth pointing out is that these repos have never been pruned so far (this is what we're working on fixing!). So the first prune will be extremely expensive. But once we're over that hump, subsequent prunes should be much cheaper.

One correction here. We did prune the "compose" repo a few months back and dropped out about 2TB of objects/commits. We went ahead and did this repo back then because we have a simple time based prune for the whole repo and so we could prune in a single run and we needed to save disk space because we were running out.

dbnicholson: I think the `--commits` results there are probably bring overstated because the directory cache is warm, but I'm pretty sure you're going to see a noticeable difference regardless.

dbnicholson: > One thing worth pointing out is that these repos have never been pruned so far (this is what we're working on fixing!). So the first prune will be extremely expensive. But once we're over that hump, subsequent prunes should be much cheaper.

I was in this situation a few years ago at Endless pre-repo-locking but now we have a daily prune that looks like it took 5 minutes in our worst case repo yesterday. I feel you're pain, though. The initial prune took a really long time after having years of commits piled up in it.

cgwalters:
Super minor nit; commit subject is long.  How about:

> lib/prune: speed up pruning by retrieving only commits

cgwalters (approved):
OK, so we ended up with a much simpler patch here; makes sense to me.

I think the combined optimizations here are likely to make a large difference.

cgwalters:
CI failure looks like an unrelated flake:

```
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + assert_file_has_content /run/ostree-boot-complete-status.txt 'error: ostree-finalize-staged.service failed on previous boot.*Operation not permitted'
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + fpath=/run/ostree-boot-complete-status.txt
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + shift
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + for re in "$@"
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + grep -q -e 'error: ostree-finalize-staged.service failed on previous boot.*Operation not permitted' /run/ostree-boot-complete-status.txt
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + _fatal_print_file /run/ostree-boot-complete-status.txt 'File '\''/run/ostree-boot-complete-status.txt'\'' doesn'\''t match regexp '\''error: ostree-finalize-staged.service failed on previous boot.*Operation not permitted'\'''
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + file=/run/ostree-boot-complete-status.txt
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + shift
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + ls -al /run/ostree-boot-complete-status.txt
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1155]: -rw-r--r--. 1 root root 587 Jun 29 15:45 /run/ostree-boot-complete-status.txt
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + sed -e 's/^/# /'
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: # ● ostree-boot-complete.service - OSTree Complete Boot
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #      Loaded: loaded (/usr/lib/systemd/system/ostree-boot-complete.service; enabled-runtime; vendor preset: disabled)
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #      Active: activating (start) since Wed 2022-06-29 15:45:59 UTC; 136ms ago
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #        Docs: man:ostree(1)
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #    Main PID: 774 (ostree)
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #       Tasks: 1 (limit: 1047)
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #      Memory: 2.0M
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #         CPU: 12ms
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #      CGroup: /system.slice/ostree-boot-complete.service
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #              └─ 774 /usr/bin/ostree admin boot-complete
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: #
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1156]: # Jun 29 15:45:59 localhost systemd[1]: Starting ostree-boot-complete.service - OSTree Complete Boot...
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + fatal 'File '\''/run/ostree-boot-complete-status.txt'\'' doesn'\''t match regexp '\''error: ostree-finalize-staged.service failed on previous boot.*Operation not permitted'\'''
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + echo File ''\''/run/ostree-boot-complete-status.txt'\''' 'doesn'\''t' match regexp ''\''error:' ostree-finalize-staged.service failed on previous 'boot.*Operation' not 'permitted'\'''
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: File '/run/ostree-boot-complete-status.txt' doesn't match regexp 'error: ostree-finalize-staged.service failed on previous boot.*Operation not permitted'
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 kola-runext-staged-deploy.sh[1104]: + exit 1
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 systemd[1]: kola-runext.service: Main process exited, code=exited, status=1/FAILURE
[2022-06-29T15:46:08.234Z] Jun 29 15:46:06 qemu0 systemd[1]: kola-runext.service: Failed with result 'exit-code'.
[2022-06-29T15:46:10.827Z] --- FAIL: ext.ostree.destructive.staged-deploy.sh (96.27s)
```

Hmm, wonder how that could flake though.

PR #2633, opened 2022-06-08
rust: Bump semver, add feature for current release

cgwalters:
There were some changes to the sys API for introspection fixes.
And add a feature for the current release, which is something
I'll add to the checklist for releases.

### DIFF
--- a/rust-bindings/sys/Cargo.toml
+++ b/rust-bindings/sys/Cargo.toml
@@ -56,6 +56,7 @@ v2021_3 = ["v2021_2"]
 v2021_4 = ["v2021_3"]
 v2021_5 = ["v2021_4"]
 v2022_2 = ["v2021_5"]
+v2022_5 = ["v2022_2"]
 
 [lib]
 name = "ostree_sys"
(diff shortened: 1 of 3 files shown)


jmarrero:
``` 
Package dependency requirement 'ostree-1 >= 2022.5' could not be satisfied.
Package 'ostree-1' has version '2022.3', required version is '>= 2022.5'
```

cgwalters:
OK yeah, the Rust builds are only using ostree from current Fedora, not what's in git.  Hmm.  Tricky, because we do want to test most of the matrix from (libostree, rust-bindings) of (git, stable) (stable, git) (git, git)

Probably wants a new CI context.  For now I dropped out the bit that hard requires the current version in the Rust CI.

jmarrero (approved):
lgtm

rust-bindings/sys/Cargo.toml
```diff
@@ -71,7 +72,7 @@ license = "MIT"
 links = "ostree-1"
 name = "ostree-sys"
 repository = "https://github.com/ostreedev/ostree-rs"
-version = "0.9.2"
+version = "0.10.0"
```
jlebon, line 75: Side note: now that this is in-tree, it seems like we should release them together and have the version string reflect the libostree release. To make it semver, we can add `0.` in front for now, and eventually `1.`, etc...

jlebon (approved):

PR #2634, opened 2022-06-08
repo: Optimize memory use of `ostree_repo_list_objects()`

cgwalters:
I was looking at https://github.com/ostreedev/ostree/pull/2632
and confused at the usage of
`GVariant *value = g_variant_new ("(b@as)", TRUE, g_variant_new_strv (NULL, 0));`
which looked strange - why the empty strv?

It turns out that this is a historical legacy of the time when
ostree had pack files.  And nothing actually cares about the values
of these variants; we should have an API that returns a proper set,
and not a hash.

But...since all of these things have exactly the same value, instead
of allocating lots of redundant copies on the heap, just have
them all hold a refcount on a shared value.

This cuts the heap usage from 20MB to 13MB on a test FCOS repository
build.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3895,6 +3895,7 @@ ostree_repo_get_parent (OstreeRepo  *self)
 
 static gboolean
 list_loose_objects_at (OstreeRepo             *self,
+                       GVariant               *dummy_value,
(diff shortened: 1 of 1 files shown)


dbnicholson (approved):
Nice! I roll my eyes every time I look at that part.

PR #2635, opened 2022-06-08
repo: Further optimize `ostree_repo_list_objects_set()`

cgwalters:
In a prior change we discovered that for bad historical reasons
libostree was returning a mapping "object type+checksum" => "metadata"
but the "metadata" was redundant and pointless.

Optimize the prune API to use a (currently internal) object listing
API which returns a set, not a map.  This allows `GHashTable` to
avoid allocating a separate array for the values, neatly cutting
memory usage in half (from ~13MB to ~6MB) on my test case of a
dry-run prune of a FCOS build.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3974,8 +3974,11 @@ list_loose_objects_at (OstreeRepo             *self,
       key = ostree_object_name_serialize (buf, objtype);
 
       /* transfer ownership */
-      g_hash_table_replace (inout_objects, g_variant_ref_sink (key),
-                            g_variant_ref (dummy_value));
+      if (dummy_value)
+        g_hash_table_replace (inout_objects, g_variant_ref_sink (key),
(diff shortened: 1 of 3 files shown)


jmarrero (approved):
lgtm

src/libostree/ostree-repo.c
```diff
@@ -4905,6 +4906,44 @@ ostree_repo_load_commit (OstreeRepo            *self,
                                  out_variant, NULL, NULL, out_state, NULL, error);
 }
 
+GHashTable *
+ostree_repo_list_objects_set (OstreeRepo                  *self,
```
jlebon, line 4910: Can we dedupe this with `ostree_repo_list_objects`? E.g. have the latter call this one with `dummy_value` passed in and in the other callsite, pass `NULL`.

cgwalters: There wasn't *that* much code in both, but OK yes, I merged things even more.

PR #2636, opened 2022-06-08
ci: Add a flow that does a git libostree + git rust-bindings

cgwalters:
In https://github.com/ostreedev/ostree/pull/2633 I realized
that our CI only builds git of libostree or git of rust-bindings,
not git of both.  And we definitely want to test the latter too,
so e.g. the Rust tests *also* become tests for changes to the C code.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -12,6 +12,8 @@ on:
 env:
   CARGO_TERM_COLOR: always
   CARGO_PROJECT_FEATURES: "v2021_5"
+  # TODO: Automatically query this from the C side
+  LATEST_LIBOSTREE: "v2022_5"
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -47,6 +49,24 @@ jobs:
         uses: Swatinem/rust-cache@ce325b60658c1b38465c06cc965b79baf32c1e72
       - name: cargo build
         run: cargo build --features=${{ env['CARGO_PROJECT_FEATURES'] }}
+  build-git-libostree:
```
jlebon, line 52: Minor/optional: would be nice to give it a name following a similar convention as the other ones.

jlebon (approved):
Neat. One minor comment, but meh not worth wasting good CI.

PR #2637, opened 2022-06-08
prune: Also use object set API in `ostree_repo_prune_from_reachable()`

cgwalters:
I missed the second prune path when working on
https://github.com/ostreedev/ostree/pull/2635


jmarrero (approved):
lgtm

PR #2638, opened 2022-06-08
lib: Fix symbol versioning inheritance

cgwalters:
I messed this up; the last release should inherit from the previous
release (N-1) and not the previous to that (N-2).

I think (hope) this isn't an ABI break...

Just noticed this when I was going to add a new symbol.


jmarrero (approved):
lgtm

dbnicholson:
Well, it definitely changes the library. I'm not sure what the impact of that is.

dbnicholson:
After poking around glibc a bit, I believe this is an ABI break. The version dependency will be recorded in the `ostree_fs_get_all_xattrs` and `ostree_fs_get_all_xattrs_at` symbols. Now you're changing them, so (I believe) if someone runs something built against 2022.5 by run against 2022.4, those symbols won't resolve.

So, I think you should release 2022.5 ASAP with instructions to ignore 2022.4.

cgwalters:
OK well, that's not too hard to test:

```
$ git checkout b330f84172e87c62b3ef85539b93881569d39ae1  # commit before this PR
$ makesudoinstall
$ cat get-xattrs-test.c
#include <ostree.h>
#include <fcntl.h>
#include <err.h>

int main(int argc, char **argv)
{
  g_autoptr(GError) local_error = NULL;
  g_autoptr(GVariant) xattrs = ostree_fs_get_all_xattrs_at(
      AT_FDCWD, "/etc/resolv.conf", NULL, &local_error);
  if (!xattrs)
    err(1, "%s", local_error->message);
  g_autofree char *t = g_variant_print(xattrs, TRUE);
  g_print("%s\n", t);
}
$ gcc $(pkg-config --cflags --libs ostree-1) get-xattrs-test.c -o get-xattrs-test
$ ./get-xattrs-test 
[(b'security.selinux', b'system_u:object_r:container_file_t:s0:c1022,c1023')]
$ git checkout main
$ makesudoinstall
$ ./get-xattrs-test
[(b'security.selinux', b'system_u:object_r:container_file_t:s0:c1022,c1023')]
```

Inspecting the binary:

```
$ eu-readelf -a get-xattrs-test| grep ostree_fs_
  0x0000000000404048  X86_64_JUMP_SLOT 000000000000000000      +0 ostree_fs_get_all_xattrs_at
    9: 0000000000000000      0 FUNC    GLOBAL DEFAULT    UNDEF ostree_fs_get_all_xattrs_at@LIBOSTREE_2022.4 (4)
   78: 0000000000000000      0 FUNC    GLOBAL DEFAULT    UNDEF ostree_fs_get_all_xattrs_at@LIBOSTREE_2022.4
  [   65e]  ostree_fs_get_all_xattrs_at@LIBOSTREE_2022.4
  [    34]  ostree_fs_get_all_xattrs_at
```

And this doesn't change with the new version - in fact, the binary is bit for bit identical.

So...I'm not a symbol versioning expert but...I think we don't need to panic at least.  I am sure there's *some* reason we need to specify the "prior version" in the dependency table, but it's not obvious to me, because it seems like the main goal is to associate each symbol with its correct version, which we *did* do correctly before.

src/libostree/libostree-released.sym
```diff
@@ -681,7 +681,7 @@ LIBOSTREE_2022.4 {
 global:
   ostree_fs_get_all_xattrs;
   ostree_fs_get_all_xattrs_at;
-} LIBOSTREE_2021.5;
+} LIBOSTREE_2022.2;
```
cgwalters, line 684: To be extra clear, this section being changed here is the "prior version inheritance", *not* the version associated with the symbols, which is correctly `LIBOSTREE_2022.4`.

PR #2639, opened 2022-06-09
RFE: Add a hidden option to `ostree admin kargs edit-in-place` to update all existing deployments in place

HuijingHei:
Describe the enhancement

Add a hidden option to `ostree admin kargs edit-in-place` to update all existing deployments in place.

Example:

$ sudo ostree admin kargs edit-in-place --append-if-missing=rw
This would be useful for: https://fedoraproject.org/wiki/Changes/Silverblue_Kinoite_readonly_sysroot


src/ostree/ot-admin-builtin-kargs.c
```diff
@@ -0,0 +1,134 @@
+/*
+ * Copyright (C) 2015 Colin Walters <walters@verbum.org>
```
travier, line 2: You probably want to add your name here too. Maybe we could trim down those headers too as we have SPDX identifiers already.

The date should also reflect the file creation date.

src/ostree/ot-admin-builtin-kargs.c
```diff
@@ -0,0 +1,134 @@
...
+#include "ot-admin-functions.h"
+#include "ot-admin-kargs-builtins.h"
+#include "otutil.h"
+
+#include <glib/gi18n.h>
+
+static OstreeCommand admin_kargs_subcommands[] = {
+  { "edit-in-place", OSTREE_BUILTIN_FLAG_NO_REPO | OSTREE_BUILTIN_FLAG_HIDDEN,
+    ot_admin_kargs_builtin_edit_in_place,
+    "Set new kernel command line arguments in place(by default to all deployments)"  },
```
travier, line 33:
```suggestion
    "Set new kernel command line arguments in place (applies to all deployments by default)"  },
```

src/ostree/ot-admin-kargs-builtin-edit-in-place.c
```diff
@@ -0,0 +1,83 @@
+/*
+ * Copyright (C) 2014 Colin Walters <walters@verbum.org>
```
travier, line 2: idem

src/ostree/ot-admin-kargs-builtin-edit-in-place.c
```diff
@@ -0,0 +1,83 @@
...
+
+#include "ot-main.h"
+#include "ot-admin-kargs-builtins.h"
+
+#include "otutil.h"
+
+static char **opt_kargs_edit_in_place_append;
+
+static GOptionEntry options[] = {
+  { "append-if-missing", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_kargs_edit_in_place_append, "Append kernel argument if not exists; useful with e.g. console= that can be used multiple times", "NAME=VALUE" },
```
travier, line 28:
```suggestion
  { "append-if-missing", 0, 0, G_OPTION_ARG_STRING_ARRAY, &opt_kargs_edit_in_place_append, "Append kernel argument if they do not exist", "NAME=VALUE" },
```

Not sure the console reference is accurate here.

src/ostree/ot-admin-kargs-builtins.h
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright (C) 2015 Red Hat, Inc.
```
travier, line 2: idem

tests/test-admin-kargs.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+#
+# Copyright (C) 2011 Colin Walters <walters@verbum.org>
```
travier, line 3: idem

tests/test-admin-kargs.sh
```diff
@@ -0,0 +1,38 @@
...
+
+# Exports OSTREE_SYSROOT so --sysroot not needed.
+setup_os_repository "archive" "syslinux"
+
+echo "1..1"
+
+${CMD_PREFIX} ostree --repo=sysroot/ostree/repo pull-local --remote=testos testos-repo testos/buildmain/x86_64-runtime
+
+${CMD_PREFIX} ostree admin deploy --karg=root=LABEL=MOO --karg=quiet --os=testos testos:testos/buildmain/x86_64-runtime
+${CMD_PREFIX} ostree admin kargs edit-in-place --append-if-missing=TESTARG=TESTVALUE --append-if-missing=ARGWITHOUTKEY testos:testos/buildmain/x86_64-runtime
```
travier, line 33: We probably want to also test if they are added only once if called twice

travier (commented):
Not a full review but looks good.

tests/test-admin-kargs.sh
HuijingHei: I update test locally to call twice, the result is failed as `append the same value twice`.
IMU, `ostree_kernel_args_append_argv (kargs, opt_kargs_edit_in_place_append)` will append value if it is not in the hash table already, but the result is not, any suggestions?

HuijingHei: Seems `ostree_kernel_args_append` can not avoid duplicated case
https://github.com/ostreedev/ostree/blob/ca1fbc59cecbbb254897180ca6afeb6daed7a7e6/tests/test-kargs.c#L120-L126

src/ostree/ot-admin-kargs-builtin-edit-in-place.c
```diff
@@ -0,0 +1,85 @@
...
+
+  // set kargs for each deployment
+  for (i = 0; i < deployments->len; i++)
+    {
+      deployment = deployments->pdata[i];
+      OstreeBootconfigParser *bootconfig = ostree_deployment_get_bootconfig (deployment);
+      g_auto(GStrv) deployment_args = g_strsplit (ostree_bootconfig_parser_get (bootconfig, "options"), " ", -1);
+      
+      kargs = ostree_kernel_args_new ();
+      ostree_kernel_args_append_argv (kargs, deployment_args);
```
cgwalters, line 68: We should use `ostree_kernel_args_from_string (ostree_bootconfig_parser_get (bootconfig, "options"));` here...it's the same thing but this way we centralize splitting.

src/ostree/ot-admin-kargs-builtin-edit-in-place.c
```diff
@@ -0,0 +1,85 @@
...
+    {
+      deployment = deployments->pdata[i];
+      OstreeBootconfigParser *bootconfig = ostree_deployment_get_bootconfig (deployment);
+      g_auto(GStrv) deployment_args = g_strsplit (ostree_bootconfig_parser_get (bootconfig, "options"), " ", -1);
+      
+      kargs = ostree_kernel_args_new ();
+      ostree_kernel_args_append_argv (kargs, deployment_args); 
+
+      if (opt_kargs_edit_in_place_append)
+        ostree_kernel_args_append_argv_if_missing(kargs, opt_kargs_edit_in_place_append);
```
cgwalters, line 71: Missing space here, `ostree_kernel_args_append_argv_if_missing (kargs, `

I think here it'd also be simpler to open-code the loop on the strv, i.e. make it:

```
  for (char **strviter = opt_kargs_edit_in_place_append; strviter && *strviter; strviter++)
    {
      const char *arg = *strviter;
      ostree_kernel_args_append_if_missing (kargs, arg);
    }
```

src/ostree/ot-admin-kargs-builtin-edit-in-place.c
```diff
@@ -0,0 +1,85 @@
...
+                                          invocation, &sysroot, cancellable, error))
+    goto out;
+
+  deployments = ostree_sysroot_get_deployments (sysroot);
+  if (deployments->len == 0)
+    {
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                   "Unable to find a deployment in sysroot");
+      goto out;
+    }
```
cgwalters, line 58: I think this is fine, though...I wonder if the "no deployments" case should actually be an error or just a silent no-op.  Mmm...I guess it makes sense to error out.  But it also seems valid to no-op.

cgwalters (commented):

src/ostree/ot-admin-kargs-builtin-edit-in-place.c
HuijingHei: Not quite sure about this, maybe this can exit early before going into loop if no deployments?

cgwalters: Yeah, could just `return TRUE;`.

That said...I honestly don't have a strong opinion on this.  It's fine by me to keep it as an error too.

HuijingHei: Update code style according to https://github.com/ostreedev/ostree/pull/2649

src/libostree/ostree-kernel-args.c
```diff
@@ -804,3 +804,25 @@ ostree_kernel_args_get_last_value (OstreeKernelArgs *kargs, const char *key)
...
+ *
+ * Since: 2022.5
+ **/
+void
+ostree_kernel_args_append_if_missing (OstreeKernelArgs  *kargs,
+                                      const char *arg)
+{
+  char *duped = g_strdup (arg);
+
+  if (g_hash_table_contains (kargs->table, duped))
```
HuijingHei, line 824: - The value that has `=` like `TESTARG=TESTVALUE` will be appended again if call twice, for example the result will be like:

```
# options root=LABEL=MOO quiet ostree=/ostree/boot.1/testos/7bd091ad6955c0e7ef41bab65adb016275664b55efc0d10a0c8cb263da3c1e4e/0 ARGWITHOUTKEY TESTARG=TESTVALUE TESTARG=TESTVALUE
```

- But value like `ARGWITHOUTKEY` will not be appended twice as expected.

HuijingHei: Parse the key from `key=value` and call g_hash_table_contains (kargs->table, key)

src/libostree/ostree-kernel-args.c
```diff
@@ -804,3 +804,33 @@ ostree_kernel_args_get_last_value (OstreeKernelArgs *kargs, const char *key)
...
+void
+ostree_kernel_args_append_if_missing (OstreeKernelArgs  *kargs,
+                                      const char *arg)
+{
+  g_autofree char *duped = g_strdup (arg);
+  const char *val = split_keyeq (duped);
+  if (val)
+    {
+      g_autofree char *key = g_strndup (duped, val - duped - 1);
+      duped = g_strdup (key);
```
cgwalters, line 828: This looks like it's trying to parse out the key in key=val, but `split_keyeq()` already mutates the input by changing the `=` into a `\0`.

src/libostree/ostree-kernel-args.c
cgwalters: Right...I think the logic we want here is more like:

```
  g_autofree char *key = g_strdup (arg);
  const char *val = split_keyeq (key);

  // Don't insert a duplicate key.
  if (g_hash_table_contains (kargs->table, key))
    return; 

  ostree_kernel_args_append (kargs, arg);
```

or so?

Though, if the argument table already has e.g. `TESTARG=VAL1` and we say `append_if_missing (TESTARG=VAL2)` ...do we expect it to be ignored, or to add a new value, or to replace the existing one?  I guess in most cases we want it to be ignored, which is what will happen here.

HuijingHei:
You are right, this will not add duplicate key.

Update according to your suggestion, and make a little change for `const char *val = split_keyeq (key);` to `split_keyeq (key);` to avoid error
```
src/libostree/ostree-kernel-args.c:824:15: warning: unused variable 'val' [-Wunused-variable]
  824 |   const char *val = split_keyeq (key);
```

cgwalters:
```
 --- expected-documented.txt	2022-06-23 03:30:28.690917580 +0000
+++ found-documented.txt	2022-06-23 03:30:28.694917654 +0000
@@ -116,7 +116,6 @@
 ostree_kernel_args_append
 ostree_kernel_args_append_argv
 ostree_kernel_args_append_argv_filtered
-ostree_kernel_args_append_if_missing
 ostree_kernel_args_append_proc_cmdline
 ostree_kernel_args_cleanup
 ostree_kernel_args_delete
OK closing connection
ERROR: tests/test-symbols.sh - too few tests run (expected 3, got 1)
ERROR: tests/test-symbols.sh - exited with status 1
```

Need to also add the new symbol to `apidoc/ostree-sections.txt`.

HuijingHei:
> Need to also add the new symbol to `apidoc/ostree-sections.txt`.

Update is done, thanks!

cgwalters (approved):
Thanks!

travier:
🎉

HuijingHei:
I am much appreciated with the help of @cgwalters and @travier, thanks a lot!

PR #2640, opened 2022-06-09
Add initial composefs integration

alexlarsson:
This is some work in progress for supporting composefs in ostree.
The initial version just supports checking out a commit to a composefs image.

### DIFF
--- a/src/libostree/ostree-repo-composefs.c
+++ b/src/libostree/ostree-repo-composefs.c
@@ -1,0 +1,560 @@
+/*
+ * Copyright (C) Red Hat, Inc.
+ *
+ * SPDX-License-Identifier: LGPL-2.0+
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "config.h"
+
+#include <gio/gunixinputstream.h>
+#include <gio/gunixoutputstream.h>
+#include <sys/ioctl.h>
+
+#include "ostree-core-private.h"
+#include "ostree-repo-file.h"
+#include "ostree-repo-private.h"
+
+#ifdef HAVE_COMPOSEFS
+#include <libcomposefs/lcfs-writer.h>
+#endif
+
+#ifdef HAVE_LINUX_FSVERITY_H
+#include <linux/fsverity.h>
+#endif
+
+gboolean
+_ostree_repo_parse_composefs_config (OstreeRepo *self, GError **error)
+{
+  /* Currently experimental */
+  OtTristate use_composefs;
+
+  if (!ot_keyfile_get_tristate_with_default (self->config, _OSTREE_INTEGRITY_SECTION, "composefs",
+                                             OT_TRISTATE_NO, &use_composefs, error))
+    return FALSE;
+
+  self->composefs_wanted = use_composefs;
+#ifdef HAVE_COMPOSEFS
+  self->composefs_supported = TRUE;
+#else
+  self->composefs_supported = FALSE;
+#endif
+
+  if (use_composefs == OT_TRISTATE_YES && !self->composefs_supported)
+    return glnx_throw (error, "composefs required, but libostree compiled without support");
+
+  return TRUE;
+}
+
+struct OstreeComposefsTarget
+{
+#ifdef HAVE_COMPOSEFS
+  struct lcfs_node_s *dest;
+#endif
+  int ref_count;
+};
+
+/**
+ * ostree_composefs_target_new:
+ *
+ * Creates a #OstreeComposefsTarget which can be used with
+ * ostree_repo_checkout_composefs() to create a composefs image based
+ * on a set of checkouts.
+ *
+ * Returns: (transfer full): a new of #OstreeComposefsTarget
+ */
+OstreeComposefsTarget *
+ostree_composefs_target_new (void)
+{
+  OstreeComposefsTarget *target;
+
+  target = g_slice_new0 (OstreeComposefsTarget);
+
+#ifdef HAVE_COMPOSEFS
+  target->dest = lcfs_node_new ();
+  lcfs_node_set_mode (target->dest, 0755 | S_IFDIR);
+#endif
+
+  target->ref_count = 1;
+
+  return target;
+}
+
+/**
+ * ostree_composefs_target_ref:
+ * @target: an #OstreeComposefsTarget
+ *
+ * Increase the reference count on the given @target.
+ *
+ * Returns: (transfer full): a copy of @target, for convenience
+ */
+OstreeComposefsTarget *
+ostree_composefs_target_ref (OstreeComposefsTarget *target)
+{
+  gint refcount;
+  g_return_val_if_fail (target != NULL, NULL);
+  refcount = g_atomic_int_add (&target->ref_count, 1);
+  g_assert (refcount > 0);
+  return target;
+}
+
+/**
+ * ostree_composefs_target_unref:
+ * @target: (transfer full): an #OstreeComposefsTarget
+ *
+ * Decrease the reference count on the given @target and free it if the
+ * reference count reaches 0.
+ */
+void
+ostree_composefs_target_unref (OstreeComposefsTarget *target)
+{
+  g_return_if_fail (target != NULL);
+  g_return_if_fail (target->ref_count > 0);
+
+  if (g_atomic_int_dec_and_test (&target->ref_count))
+    {
+#ifdef HAVE_COMPOSEFS
+      g_clear_pointer (&target->dest, lcfs_node_unref);
+#endif
+      g_slice_free (OstreeComposefsTarget, target);
+    }
+}
+
+G_DEFINE_BOXED_TYPE (OstreeComposefsTarget, ostree_composefs_target, ostree_composefs_target_ref,
+                     ostree_composefs_target_unref);
+
+#ifdef HAVE_COMPOSEFS
+
+static ssize_t
+_composefs_read_cb (void *_file, void *buf, size_t count)
+{
+  GInputStream *in = _file;
+  gsize bytes_read;
+
+  if (!g_input_stream_read_all (in, buf, count, &bytes_read, NULL, NULL))
+    {
+      errno = EIO;
+      return -1;
+    }
+
+  return bytes_read;
+}
+
+static ssize_t
+_composefs_write_cb (void *file, void *buf, size_t len)
+{
+  int fd = GPOINTER_TO_INT (file);
+  const char *content = buf;
+  ssize_t res = 0;
+
+  while (len > 0)
+    {
+      res = write (fd, content, len);
+      if (res < 0 && errno == EINTR)
+        continue;
+
+      if (res <= 0)
+        {
+          if (res == 0) /* Unexpected short write, should not happen when writing to a file */
+            errno = ENOSPC;
+          return -1;
+        }
+
+      break;
+    }
+
+  return res;
+}
+
+#endif
+
+/**
+ * ostree_composefs_target_write:
+ * @target: an #OstreeComposefsTarget
+ * @fd: Write image here (or -1 to not write)
+ * @out_fsverity_digest: (out) (array fixed-size=32) (nullable): Return location for the fsverity
+ * binary digest, or %NULL to not compute it
+ * @cancellable: Cancellable
+ * @error: Error
+ *
+ * Writes a composefs image file to the filesystem at the
+ * path specified by @destination_dfd and destination_path (if not %NULL)
+ * and (optionally) computes the fsverity digest of the image.
+ *
+ * Returns: %TRUE on success, %FALSE on failure
+ */
+gboolean
+ostree_composefs_target_write (OstreeComposefsTarget *target, int fd, guchar **out_fsverity_digest,
+                               GCancellable *cancellable, GError **error)
+{
+#ifdef HAVE_COMPOSEFS
+  g_autoptr (GOutputStream) tmp_out = NULL;
+  g_autoptr (GOutputStream) out = NULL;
+  struct lcfs_node_s *root;
+  g_autofree guchar *fsverity_digest = NULL;
+  struct lcfs_write_options_s options = {
+    LCFS_FORMAT_EROFS,
+  };
+
+  root = lcfs_node_lookup_child (target->dest, "root");
+  if (root == NULL)
+    root = target->dest; /* Nothing was checked out, use an empty dir */
+
+  if (out_fsverity_digest)
+    {
+      fsverity_digest = g_malloc (OSTREE_SHA256_DIGEST_LEN);
+      options.digest_out = fsverity_digest;
+    }
+
+  if (fd != -1)
+    {
+      options.file = GINT_TO_POINTER (fd);
+      options.file_write_cb = _composefs_write_cb;
+    }
+
+  if (lcfs_write_to (root, &options) != 0)
+    return glnx_throw_errno (error);
+
+  if (out_fsverity_digest)
+    *out_fsverity_digest = g_steal_pointer (&fsverity_digest);
+
+  return TRUE;
+#else
+  g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED,
+               "Composeefs is not supported in this ostree build");
+  return FALSE;
+#endif
+}
+
+#ifdef HAVE_COMPOSEFS
+static gboolean
+_ostree_composefs_set_xattrs (struct lcfs_node_s *node, GVariant *xattrs, GCancellable *cancellable,
+                              GError **error)
+{
+  const guint n = g_variant_n_children (xattrs);
+  for (guint i = 0; i < n; i++)
+    {
+      const guint8 *name;
+      g_autoptr (GVariant) value = NULL;
+      g_variant_get_child (xattrs, i, "(^&ay@ay)", &name, &value);
+
+      gsize value_len;
+      const guint8 *value_data = g_variant_get_fixed_array (value, &value_len, 1);
+
+      if (lcfs_node_set_xattr (node, (char *)name, (char *)value_data, value_len) != 0)
+        return glnx_throw_errno_prefix (error, "Setting composefs xattrs for %s", name);
+    }
+
+  return TRUE;
+}
+
+static gboolean
+checkout_one_composefs_file_at (OstreeRepo *repo, const char *checksum, struct lcfs_node_s *parent,
+                                const char *destination_name, GCancellable *cancellable,
+                                GError **error)
+{
+  g_autoptr (GInputStream) input = NULL;
+  g_autoptr (GVariant) xattrs = NULL;
+  struct lcfs_node_s *existing;
+
+  /* Validate this up front to prevent path traversal attacks */
+  if (!ot_util_filename_validate (destination_name, error))
+    return FALSE;
+
+  existing = lcfs_node_lookup_child (parent, destination_name);
+  if (existing != NULL)
+    return glnx_throw (error, "Target checkout file already exist");
+
+  g_autoptr (GFileInfo) source_info = NULL;
+  if (!ostree_repo_load_file (repo, checksum, &input, &source_info, &xattrs, cancellable, error))
+    return FALSE;
+
+  const guint32 source_mode = g_file_info_get_attribute_uint32 (source_info, "unix::mode");
+  const guint32 source_uid = g_file_info_get_attribute_uint32 (source_info, "unix::uid");
+  const guint32 source_gid = g_file_info_get_attribute_uint32 (source_info, "unix::gid");
+  const guint64 source_size = g_file_info_get_size (source_info);
+  const char *source_symlink_target = g_file_info_get_symlink_target (source_info);
+  const gboolean is_symlink
+      = (g_file_info_get_file_type (source_info) == G_FILE_TYPE_SYMBOLIC_LINK);
+
+  struct lcfs_node_s *node = lcfs_node_new ();
+  if (node == NULL)
+    return glnx_throw (error, "Out of memory");
+
+  /* Takes ownership on success */
+  if (lcfs_node_add_child (parent, node, destination_name) != 0)
+    {
+      lcfs_node_unref (node);
+      return glnx_throw_errno (error);
+    }
+
+  lcfs_node_set_mode (node, source_mode);
+  lcfs_node_set_uid (node, source_uid);
+  lcfs_node_set_gid (node, source_gid);
+  lcfs_node_set_size (node, source_size);
+  if (is_symlink)
+    {
+      if (lcfs_node_set_payload (node, source_symlink_target) != 0)
+        return glnx_throw_errno (error);
+    }
+  else if (source_size != 0)
+    {
+      char loose_path_buf[_OSTREE_LOOSE_PATH_MAX];
+      _ostree_loose_path (loose_path_buf, checksum, OSTREE_OBJECT_TYPE_FILE, OSTREE_REPO_MODE_BARE);
+      if (lcfs_node_set_payload (node, loose_path_buf) != 0)
+        return glnx_throw_errno (error);
+
+      guchar *known_digest = NULL;
+
+#ifdef HAVE_LINUX_FSVERITY_H
+      /* First try to get the digest directly from the bare repo file.
+       * This is the typical case when we're pulled into the target
+       * system repo with verity on and are recreating the composefs
+       * image during deploy. */
+      char buf[sizeof (struct fsverity_digest) + OSTREE_SHA256_DIGEST_LEN];
+
+      if (G_IS_UNIX_INPUT_STREAM (input))
+        {
+          int content_fd = g_unix_input_stream_get_fd (G_UNIX_INPUT_STREAM (input));
+          struct fsverity_digest *d = (struct fsverity_digest *)&buf;
+          d->digest_size = OSTREE_SHA256_DIGEST_LEN;
+
+          if (ioctl (content_fd, FS_IOC_MEASURE_VERITY, d) == 0
+              && d->digest_size == OSTREE_SHA256_DIGEST_LEN
+              && d->digest_algorithm == FS_VERITY_HASH_ALG_SHA256)
+            known_digest = d->digest;
+        }
+#endif
+
+      if (known_digest)
+        lcfs_node_set_fsverity_digest (node, known_digest);
+      else if (lcfs_node_set_fsverity_from_content (node, input, _composefs_read_cb) != 0)
+        return glnx_throw_errno (error);
+    }
+
+  if (xattrs)
+    {
+      if (!_ostree_composefs_set_xattrs (node, xattrs, cancellable, error))
+        return FALSE;
+    }
+
+  g_clear_object (&input);
+
+  return TRUE;
+}
+
+static gboolean
+checkout_composefs_recurse (OstreeRepo *self, const char *dirtree_checksum,
+                            const char *dirmeta_checksum, struct lcfs_node_s *parent,
+                            const char *name, GCancellable *cancellable, GError **error)
+{
+  g_autoptr (GVariant) dirtree = NULL;
+  g_autoptr (GVariant) dirmeta = NULL;
+  g_autoptr (GVariant) xattrs = NULL;
+  struct lcfs_node_s *directory;
+
+  if (!ostree_repo_load_variant (self, OSTREE_OBJECT_TYPE_DIR_TREE, dirtree_checksum, &dirtree,
+                                 error))
+    return FALSE;
+  if (!ostree_repo_load_variant (self, OSTREE_OBJECT_TYPE_DIR_META, dirmeta_checksum, &dirmeta,
+                                 error))
+    return FALSE;
+
+  /* Parse OSTREE_OBJECT_TYPE_DIR_META */
+  guint32 uid, gid, mode;
+  g_variant_get (dirmeta, "(uuu@a(ayay))", &uid, &gid, &mode, &xattrs);
+  uid = GUINT32_FROM_BE (uid);
+  gid = GUINT32_FROM_BE (gid);
+  mode = GUINT32_FROM_BE (mode);
+
+  directory = lcfs_node_lookup_child (parent, name);
+  if (directory != NULL && lcfs_node_get_mode (directory) != 0)
+    {
+      return glnx_throw (error, "Target checkout directory already exist");
+    }
+  else
+    {
+      directory = lcfs_node_new ();
+      if (directory == NULL)
+        return glnx_throw (error, "Out of memory");
+
+      /* Takes ownership on success */
+      if (lcfs_node_add_child (parent, directory, name) != 0)
+        {
+          lcfs_node_unref (directory);
+          return glnx_throw_errno (error);
+        }
+    }
+
+  lcfs_node_set_mode (directory, mode);
+  lcfs_node_set_uid (directory, uid);
+  lcfs_node_set_gid (directory, gid);
+
+  /* Set the xattrs if we created the dir */
+  if (xattrs && !_ostree_composefs_set_xattrs (directory, xattrs, cancellable, error))
+    return FALSE;
+
+  /* Process files in this subdir */
+  {
+    g_autoptr (GVariant) dir_file_contents = g_variant_get_child_value (dirtree, 0);
+    GVariantIter viter;
+    g_variant_iter_init (&viter, dir_file_contents);
+    const char *fname;
+    g_autoptr (GVariant) contents_csum_v = NULL;
+    while (g_variant_iter_loop (&viter, "(&s@ay)", &fname, &contents_csum_v))
+      {
+        char tmp_checksum[OSTREE_SHA256_STRING_LEN + 1];
+        _ostree_checksum_inplace_from_bytes_v (contents_csum_v, tmp_checksum);
+
+        if (!checkout_one_composefs_file_at (self, tmp_checksum, directory, fname, cancellable,
+                                             error))
+          return FALSE;
+      }
+    contents_csum_v = NULL; /* iter_loop freed it */
+  }
+
+  /* Process subdirectories */
+  {
+    g_autoptr (GVariant) dir_subdirs = g_variant_get_child_value (dirtree, 1);
+    const char *dname;
+    g_autoptr (GVariant) subdirtree_csum_v = NULL;
+    g_autoptr (GVariant) subdirmeta_csum_v = NULL;
+    GVariantIter viter;
+    g_variant_iter_init (&viter, dir_subdirs);
+    while (
+        g_variant_iter_loop (&viter, "(&s@ay@ay)", &dname, &subdirtree_csum_v, &subdirmeta_csum_v))
+      {
+        /* Validate this up front to prevent path traversal attacks. Note that
+         * we don't validate at the top of this function like we do for
+         * checkout_one_file_at() becuase I believe in some cases this function
+         * can be called *initially* with user-specified paths for the root
+         * directory.
+         */
+        if (!ot_util_filename_validate (dname, error))
+          return FALSE;
+
+        char subdirtree_checksum[OSTREE_SHA256_STRING_LEN + 1];
+        _ostree_checksum_inplace_from_bytes_v (subdirtree_csum_v, subdirtree_checksum);
+        char subdirmeta_checksum[OSTREE_SHA256_STRING_LEN + 1];
+        _ostree_checksum_inplace_from_bytes_v (subdirmeta_csum_v, subdirmeta_checksum);
+        if (!checkout_composefs_recurse (self, subdirtree_checksum, subdirmeta_checksum, directory,
+                                         dname, cancellable, error))
+          return FALSE;
+      }
+  }
+
+  return TRUE;
+}
+
+/* Begin a checkout process */
+static gboolean
+checkout_composefs_tree (OstreeRepo *self, OstreeComposefsTarget *target, OstreeRepoFile *source,
+                         GFileInfo *source_info, GCancellable *cancellable, GError **error)
+{
+  if (g_file_info_get_file_type (source_info) != G_FILE_TYPE_DIRECTORY)
+    return glnx_throw (error, "Root checkout of composefs must be directory");
+
+  /* Cache any directory metadata we read during this operation;
+   * see commit b7afe91e21143d7abb0adde440683a52712aa246
+   */
+  g_auto (OstreeRepoMemoryCacheRef) memcache_ref;
+  _ostree_repo_memory_cache_ref_init (&memcache_ref, self);
+
+  g_assert_cmpint (g_file_info_get_file_type (source_info), ==, G_FILE_TYPE_DIRECTORY);
+
+  const char *dirtree_checksum = ostree_repo_file_tree_get_contents_checksum (source);
+  const char *dirmeta_checksum = ostree_repo_file_tree_get_metadata_checksum (source);
+  return checkout_composefs_recurse (self, dirtree_checksum, dirmeta_checksum, target->dest, "root",
+                                     cancellable, error);
+}
+
+static struct lcfs_node_s *
+ensure_lcfs_dir (struct lcfs_node_s *parent, const char *name, GError **error)
+{
+  struct lcfs_node_s *node;
+
+  node = lcfs_node_lookup_child (parent, name);
+  if (node != NULL)
+    return node;
+
+  node = lcfs_node_new ();
+  lcfs_node_set_mode (node, 0755 | S_IFDIR);
+  if (lcfs_node_add_child (parent, node, name) != 0)
+    {
+      lcfs_node_unref (node);
+      glnx_throw_errno (error);
+      return NULL;
+    }
+
+  return node;
+}
+#endif
+
+/**
+ * ostree_repo_checkout_composefs:
+ * @self: Repo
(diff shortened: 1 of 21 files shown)


alexlarsson:
Yeah, maybe composefs-generate is better. I'm not sure how important all the features are, but we want to support everything that rpm-ostree uses, which is at least the union stuff. You would know better than I here.

alexlarsson:
Basically, wherever rpm-ostree now does a checkout, we're likely to want to make a composefs instead, using the same features.

wmanley:
This is really interesting stuff :).

FUSE has a "passthrough" optimisation where userspace deals with permissions, etc. but then can return a file descriptor to the kernel such that all reads/writes don't have any FUSE overhead: https://lwn.net/Articles/843873/ .  I'd planned to use this in my ostree-fuse filesystem, but I've mostly lost interested due to the amount of work that would be required to bring the rust FUSE library (fuser) up to scratch.

`composefs` sounds like it's a much better approach - at least for solving these problems, as it'll be faster, can be used for rootfs and will be easier to manage the security properties in the kernel.  I've not looked in detail, but I'd imagine that in combination with overlayfs and a underlying filesystem that supports reflinks it could work quite well in RW mode too.

alexlarsson:
Rebased on top of the latest composefs tree. This version has a nicer integration with the build-system, and now has a `ostree composefs-generate` command instead of piling on top of the checkout command.

This is still WIP though. For this to be more complete we would need:
 * Support in `ostree commit` to store the fsverity digest of the composefs image of the commit in the commit metadata
 * Support for ostree/rpm-ostree as a system manager to use a composefs instead of a checkout for the root fs.
 * Optional support for extracting the fsverity digest (generated above) from the ostree commit and add that to the kernel command line in order to validate the root fs.

alexlarsson:
With the latest commit I was able to boot an ostree system build from the latest fedora coreos (built using coreos-assembler) with a bunch of workaround hacks:

1) Use rpm overlays to install a kernel with composefs support. Rpms here: https://alexl.fedorapeople.org/composefs-rpms/

2) ostree hack to load the composefs module (unclear why this is not automatic, probably some issue with the above rpms):
```
diff --git a/src/boot/ostree-prepare-root.service b/src/boot/ostree-prepare-root.service
index 510d866a..4385f0be 100644
--- a/src/boot/ostree-prepare-root.service
+++ b/src/boot/ostree-prepare-root.service
@@ -26,6 +26,7 @@ Before=initrd-root-fs.target
 
 [Service]
 Type=oneshot
+ExecStart=insmod /usr/lib/modules/5.19.0-0.rc4.33.fc36.x86_64/kernel/fs/composefs/composefs.ko.xz
 ExecStart=/usr/lib/ostree/ostree-prepare-root /sysroot
 StandardInput=null
 StandardOutput=journal
```

3. Install the ostree in the fcos rootfs overlay *and* the fcos build container.
4. Hack to allow selinux policy to work with xattrs for composfs:

```
diff --git a/manifests/fedora-coreos-base.yaml b/manifests/fedora-coreos-base.yaml
index 53c63de..db6d73b 100644
--- a/manifests/fedora-coreos-base.yaml
+++ b/manifests/fedora-coreos-base.yaml
@@ -64,6 +64,8 @@ postprocess:
     set -xeuo pipefail
     setsebool -P -N container_use_cephfs on  # RHBZ#1692369
     setsebool -P -N virt_use_samba on  # RHBZ#1754825
+    echo "(fsuse xattr composefs (system_u object_r fs_t ((s0) (s0))))" > /tmp/composefs.cil
+    semodule -i /tmp/composefs.cil
 
   # Mask dnsmasq. We include dnsmasq for host services that use the dnsmasq
   # binary but intentionally mask the systemd service so users can't easily
```

5. Disable rdcore in coreos-boot-edit as it gets confused by the composefs mount
```
diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
index cf2e9c3..649e5f1 100755
--- a/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/35coreos-ignition/coreos-boot-edit.sh
@@ -30,9 +30,9 @@ rm -vrf ${initramfs_firstboot_network_dir}
 # If root is specified, assume rootmap is already configured. Otherwise,
 # append rootmap kargs to the BLS configs.
 root=$(karg root)
-if [ -z "${root}" ]; then
-    rdcore rootmap /sysroot --boot-mount ${bootmnt}
-fi
+#if [ -z "${root}" ]; then
+#    rdcore rootmap /sysroot --boot-mount ${bootmnt}
+#fi
 
 # This does a few things:
 # 1. it puts the boot UUID in /run/coreos/bootfs_uuid which is used by the real
@@ -42,4 +42,4 @@ fi
 # 3. it create a .root_uuid stamp file on the bootfs or fails if one exists
 # 4. it adds GRUB bootuuid.cfg dropins so that GRUB selects the boot filesystem
 #    by UUID
-rdcore bind-boot /sysroot ${bootmnt}
+#rdcore bind-boot /sysroot ${bootmnt}
```
6. Disable` ignition-ostree-check-rootfs-size.service as it gets confused by the composefs mount
```
diff --git a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
index bf9a787..c377fd4 100755
--- a/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
+++ b/overlay.d/05core/usr/lib/dracut/modules.d/40ignition-ostree/module-setup.sh
@@ -96,9 +96,6 @@ install() {
     inst_script "$moddir/ignition-ostree-growfs.sh" \
         /usr/sbin/ignition-ostree-growfs
 
-    install_ignition_unit ignition-ostree-check-rootfs-size.service
-    inst_script "$moddir/coreos-check-rootfs-size" \
-        /usr/libexec/coreos-check-rootfs-size
 
     inst_script "$moddir/coreos-relabel" /usr/bin/coreos-relabel
 }

```

Note that the above still runs without fserity enabled for the files. That is gonna need more FCOS plumbing to ensure we get fsverity enabled on the files in the ostree repo.

alexlarsson:
Also, with the above rpm-ostreed.service fails with:

Status: "error: Couldn't start daemon: Error setting up sysroot: loading sysroot: Unexpected state: /run/ostree-booted found and in / sysroot, but bootloader entry not found"

But I believe this is fallout from the disabling of rdcore.

cgwalters:
Nice!  I think it is likely indeed that we'd need to adapt `rdcore rootmap` to understand composefs.  Concretely...we need to find the underlying backing block device behind it.  Hmm, has that come up for e.g. overlayfs?  (I guess in theory overlayfs can have multiple backing blockdevs for separate lower dirs vs upper)

alexlarsson:
Its not just that overlayfs can have multiple sources, they in turn can be weird filesystems (like another overlayfs) so you would have to recurse through the whole stack.

cgwalters:
Yep that's fine, the rootmap code already understands recursion because block devices can already be stacked in that way (dm-crypt on raid etc.)

alexlarsson:
@cgwalters How do you currently find stacked block devices like that? We might need to change composefs so that its possible to find the underlying fs easily. For example, by canonicalizing the mount source path (its currenly just what was given).

cgwalters:
WDYT about making this a build time option?  I think it'd make merging easier.

alexlarsson:
I rebased this work on the latest composefs userspace and switched it to use the new overlay+erofs approach. This means that this code works as-is on a non-modified kernel, although it will require kernel changes to use the fs-verity validating aspects of composefs.

For now, this is just the initial work of using composfs at all, there is not work on fs-verity. For that to happen we need further work to:
* Upstream overlayfs kernel patches
* Ensure the object files get fs-verity enabled
* Ensure the image file gets fs-verity enabled
* Somehow pass down a trusted id for the composefs image (digest or signature) into the initrd and verify it when mounting

Additionally we need to consider exactly how/when to enable composefs. Right now the patch just always uses it when built-in, but that may not always work or be wanted. However, we also don't want anyone to be able to just disable composefs if the OS relies on it for security.

Another question is about the deploy dir, currently we *both* deploy and create the composefs image. Maybe we should only deploy the /etc directory if composefs is used?

alexlarsson:
I've been testing this with fedora-coreos, but that needs some fixes to work properly:
https://github.com/coreos/fedora-coreos-config/pull/2404

alexlarsson:
Here is a sketch for how we can support trusted boot using composefs and fs-verity signatures.

During build:

1. Generate a public/private key pair
2. Copy the public key into the new rootfs for the commit (e.g. `/etc/cfs.pub`)
3. During initrd generation in the rootfs, pass `--install  /etc/cfs.pub` to dracut, which will copy the public key into the initrd
4. Generate a UKI or aboot image containing the above initrd, the kernel and kernel command line. The kernel command line uses a generic `ostree=latest` argument, because at this point we don't know the final deployment id. [1]
5. Sign the UKI with a private key that is trusted by your secureboot keyring.
6. Stuff the UKI into the rootfs next to the normal kernel
7. Save the rootfs as objects in the ostree repo, giving the digest of the rootdir
8. ostree commit normally just stores the above digest in the metadata. But now we also take the private key from step 1 (passed as argument to `ostree commit`)  and generate a composefs image file based on the rootdir digest. We sign this file with the private key and store the signature as extra metadata in the commit object.
9. The entire commit object is GPG signed and pushed to a repo.

During install:

1.Pull a new commit
2. Verify the GPG signature
3. When deploying we look at the metadata from the commit, in particular the rootdir digest and the signature. The rootdir digest (and the repo objects) is used to construct a new composefs file, and the signature is used to sign the composefs image file when enabling fs-verity on it.
4. The BLS files are created in /boot/loader.[01] that points to the deploy dir with the composefs file, and the /boot/loader symlink is atomically switched to new loader.[01] dir. This BLS file contains the deploy id we deployed into in the kernel `ostree=...` arg.
5. The UKI is put somewhere where the boot loader can find it. (EFI partition, aboot partition, etc)

During boot:

1. The firmware loads the UKI, verifies it according to the secureboot keyring and boots kernel+initrd.
2. The initrd mounts the disk partitions.
3. The initrd notices the kernel arg "ostree=latest" and looks for the BLS file in /boot/loader with index=1 (i.e. most recent deploy, or index=2 if we're in fallback mode).
4. The initrd parses the BLS file, which contains the full ostree=... argument. This lets us find the deploy directory (like `/ostree/deploy/fedora-coreos/deploy/443ae0cd86a7dd4c6f5486a2283471b3c8f76fc5dcc4766cf935faa24a9e3d34.0`). (Note at this point that we can't trust either the BLS file or the deploy dir.)
6. The initrd loads /etc/cfs.pub into the kernel keyring for fs-verity. (Trusted, as its in signed initrd.)
7. The initrd mounts the composefs with the `LCFS_MOUNT_FLAGS_REQUIRE_SIGNATURE` flag. This ensures that the file to be mounted has a signature, and thus can only be read if the matching public key is loaded in the keyring.
8. On top of the composefs we bind mount writable things like /var and /etc.
9. Pivot-root into the new composefs mount, which now will verify all further reads from the readonly parts of rootfs are valid.

[1] This is worked on in https://github.com/ostreedev/ostree/pull/2844

ericcurtin:
> Here is a sketch for how we can support trusted boot using composefs and fs-verity signatures.
> 
> During build:
> 
> 1. Generate a public/private key pair
> 2. Copy the public key into the new rootfs for the commit (e.g. `/etc/cfs.pub`)
> 3. During initrd generation in the rootfs, pass `--install  /etc/cfs.pub` to dracut, which will copy the public key into the initrd
> 4. Generate a UKI or aboot image containing the above initrd, the kernel and kernel command line. The kernel command line uses a generic `ostree=latest` argument, because at this point we don't know the final deployment id. [1]
> 5. Sign the UKI with a private key that is trusted by your secureboot keyring.
> 6. Stuff the UKI into the rootfs next to the normal kernel
> 7. Save the rootfs as objects in the ostree repo, giving the digest of the rootdir
> 8. ostree commit normally just stores the above digest in the metadata. But now we also take the private key from step 1 (passed as argument to `ostree commit`)  and generate a composefs image file based on the rootdir digest. We sign this file with the private key and store the signature as extra metadata in the commit object.
> 9. The entire commit object is GPG signed and pushed to a repo.
> 
> During install:
> 
> 1.Pull a new commit 2. Verify the GPG signature 3. When deploying we look at the metadata from the commit, in particular the rootdir digest and the signature. The rootdir digest (and the repo objects) is used to construct a new composefs file, and the signature is used to sign the composefs image file when enabling fs-verity on it. 4. The BLS files are created in /boot/loader.[01] that points to the deploy dir with the composefs file, and the /boot/loader symlink is atomically switched to new loader.[01] dir. This BLS file contains the deploy id we deployed into in the kernel `ostree=...` arg. 5. The UKI is put somewhere where the boot loader can find it. (EFI partition, aboot partition, etc)
> 
> During boot:
> 
> 1. The firmware loads the UKI, verifies it according to the secureboot keyring and boots kernel+initrd.
> 2. The initrd mounts the disk partitions.
> 3. The initrd notices the kernel arg "ostree=latest" and looks for the BLS file in /boot/loader with index=1 (i.e. most recent deploy, or index=2 if we're in fallback mode).
> 4. The initrd parses the BLS file, which contains the full ostree=... argument. This lets us find the deploy directory (like `/ostree/deploy/fedora-coreos/deploy/443ae0cd86a7dd4c6f5486a2283471b3c8f76fc5dcc4766cf935faa24a9e3d34.0`). (Note at this point that we can't trust either the BLS file or the deploy dir.)
> 5. The initrd loads /etc/cfs.pub into the kernel keyring for fs-verity. (Trusted, as its in signed initrd.)
> 6. The initrd mounts the composefs with the `LCFS_MOUNT_FLAGS_REQUIRE_SIGNATURE` flag. This ensures that the file to be mounted has a signature, and thus can only be read if the matching public key is loaded in the keyring.
> 7. On top of the composefs we bind mount writable things like /var and /etc.
> 8. Pivot-root into the new composefs mount, which now will verify all further reads from the readonly parts of rootfs are valid.
> 
> [1] This is worked on in #2844

I align with this view, I did call ostree=latest , ostree=aboot in #2844 but I'd be delighted if UKI's used the same approach rather than going down two separate code paths. So I'd be happy to change it to ostree=latest ...

ericcurtin:
I don't want to go there yet but there is an opportunity to try and move all bootloaders to the select ostree in initramfs model. It disassociates the kernel, cmdline and initramfs from userspace a little. But that's the way it would be in a non-ostree system anyway. Kernelspace is not supposed to break userspace anyway...

cgwalters:
>  It disassociates the kernel, cmdline and initramfs from userspace a little.

It can't just be disassociated "a little".  Things are either coupled, or they're not.

Specifically in the general case there's `/usr/lib/modules/$kver` in the userspace that needs to match the booted kernel.  So if we were trying to sever things here we'd need to teach ostree how to handle that too...seems tractable by having each userspace tree include the kmod directories.

alexlarsson:
> * Generate a public/private key pair
> 
> * Copy the public key into the new rootfs for the commit (e.g. `/etc/cfs.pub`)
> 
> * During initrd generation in the rootfs, pass `--install  /etc/cfs.pub` to dracut, which will copy the public key into the initrd

So, I've been thinking about this. And while this is one possible approach. Another is to fs-verity sign the composefs image using a key that is stored in the TPM trusted set of keys. In other words, the same key that also signs the ABOOT/UKI image.

This means any signed kernel can boot any signed image, whereas the throwaway key will require the initrd and the userspace to exactly match. I don't think this is a huge deal anyway, this is how e.g. aboot with a signed system partition works right now.

In fact, this will work better for ostree, as you may re-gain the ability to share a kernel/initrd between different userspaces.

Anywhow, In practice this matters little to ostree as such. In both cases what we need is the ability to pass a private key to "ostree commit" which is then used to sign the composefs image and embed the result in the commit metadata, and the abilility for ostree deploy to apply this signature. The loading of the right public key is all up to the initrd implementation anyway.

alexlarsson:
Initial work on a new version that doesn't have any public API. All it does is record info in commit and optionally deploy a composefs and pick up a deployed composefs at boot.

I've not fully tested this new version (didn't boot into it), and there are some remaining stuff I want to add. But, it shows the overall direction. I'll keep working on this tomorrow.

alexlarsson:
So, I ran the above version against fcos, by following the following step:
* Add to `overrides/`  the patched ostree + latest libcomposefs
* Copy `overrides` to / in the cosa container
* Add libcomposefs to the build kvm via this change to /usr/lib/coreos-assembler/cmdlib.sh:
```
--- cmdlib.sh~	2023-04-26 15:57:45.370547700 +0000
+++ cmdlib.sh	2023-04-26 16:00:14.604919612 +0000
@@ -731,6 +731,8 @@
     # and include all GPG keys
     find /etc/pki/rpm-gpg/ -type f >> "${vmpreparedir}/hostfiles"
 
+    echo /usr/lib64/libcomposefs.so* | tr " " "\n"  >> "${vmpreparedir}/hostfiles"
+
     # the reason we do a heredoc here is so that the var substition takes
     # place immediately instead of having to proxy them through to the VM
     cat > "${vmpreparedir}/init" <<EOF
```
* Ensure that the built cosa image has fs-verity and composefs enabled by patching /usr/lib/coreos-assembler/create_disk.sh like so:
```
--- create_disk.sh~	2023-05-18 10:19:04.170471734 +0000
+++ create_disk.sh	2023-05-18 10:36:18.022877764 +0000
@@ -310,6 +310,8 @@
 # Opt-in to https://github.com/ostreedev/ostree/pull/1767 AKA
 # https://github.com/ostreedev/ostree/issues/1265
 ostree config --repo $rootfs/ostree/repo set sysroot.readonly true
+ostree config --repo $rootfs/ostree/repo set ex-fsverity.required true
+ostree config --repo $rootfs/ostree/repo set ex-composefs.required true
 # Initialize the "stateroot"
 ostree admin os-init "$os_name" --sysroot $rootfs
```
* Add `rootfs: ext4verity` to src/config/image-base.yaml
* Enable composefs in the build repo so that ostree commits generates composefs commit metadata:
```
 sudo ostree config --repo cache/repo-build set ex-composefs.required true
```
* Also set some key/certs so that the build composefs image will be signed (passing the signature via the commit metadata):
```
$ openssl req -newkey rsa:4096 -nodes -keyout key.pem -x509 -out cert.pem
$ sudo ostree config --repo=cache/repo-build set ex-composefs.certfile /srv/cert.pem
$ sudo ostree config --repo=cache/repo-build set ex-composefs.keyfile /srv/key.pem
```

With this, after boot my mounts look like this:
```
$ cat /proc/self/mountinfo  | grep "ext4\|overl"
70 76 252:4 / /sysroot ro,relatime - ext4 /dev/vda4 rw,seclabel
76 1 0:34 / / ro,relatime shared:1 - overlay overlay ro,seclabel,lowerdir=/sysroot/ostree/deploy/fedora-coreos/deploy/c44ede8bf849ead5d22f6120097566452c1a70ca7c8aebc15e4898278d2f8244.0/.ostree.mnt:/sysroot/ostree/repo/objects,redirect_din
73 76 252:4 /ostree/deploy/fedora-coreos/deploy/c44ede8bf849ead5d22f6120097566452c1a70ca7c8aebc15e4898278d2f8244.0/etc /etc rw,relatime shared:2 - ext4 /dev/vda4 rw,seclabel
77 70 252:4 /ostree/deploy/fedora-coreos/var /sysroot/ostree/deploy/fedora-coreos/var rw,relatime - ext4 /dev/vda4 rw,seclabel
41 76 252:4 /ostree/deploy/fedora-coreos/var /var rw,relatime shared:4 - ext4 /dev/vda4 rw,seclabel
44 76 252:3 / /boot ro,nosuid,nodev,relatime shared:79 - ext4 /dev/vda3 rw,seclabel
```
And the composefs image has fs-verity enabled, and the signature applied:
```
$ fsverity measure /ostree/deploy/fedora-coreos/deploy/c44ede8bf849ead5d22f6120097566452c1a70ca7c8aebc15e4898278d2f8244.0/.ostree.cfs 
sha256:af03eab236d76f142f06e854b087782a5cd5b7b7fef640c30ce798a44eb50b6d /ostree/deploy/fedora-coreos/deploy/c44ede8bf849ead5d22f6120097566452c1a70ca7c8aebc15e4898278d2f8244.0/.ostree.cfs
$ fsverity dump_metadata signature /ostree/deploy/fedora-coreos/deploy/c44ede8bf849ead5d22f6120097566452c1a70ca7c8aebc15e4898278d2f8244.0/.ostree.cfs | od -x | head -3
0000000 8230 bd02 0906 862a 8648 0df7 0701 a002
0000020 0282 30ae 0282 02aa 0101 0f31 0d30 0906
0000040 8660 0148 0365 0204 0501 3000 060b 2a09
```
What is missing here for a complete setup is:
* Kernel support for the `overlay.verity` xattr (I have patches on the list)
* The fedora kernel doesn't have fsverity signatures enabled in the config
* I need to tweak the kernel command line to say ot-composefs=signed
* The public key has to be added to the initrd, and loaded into the .fs-verity keyring.

But, once those is done, it should be fully protected.

Further work that could be done in the patchset:
 * Do partial checkouts (only etc) when deploying using composefs
 * Cache fs-verity digests for file object somewhere for faster composefs metadata generation during ostree commit if the fs the repo is on doesn't use fs-verity (or it is not enabled).

cgwalters:
> Do partial checkouts (only etc) when deploying using composefs

Bigger picture though this leads to the question of whether one wants a mutable/persistent `/etc` at all right?  I think we should definitely add first-class support to ostree for making `/etc` a transient (writable but goes away on reboot) overlayfs at least.

cgwalters:
Hmm while we're here, probably `ex-fsverity` shouldn't be its own group, but be under an `integrity` group?  And `ex-composefs` should also imply `ex-verity`.

cgwalters:
IOW it's `ostree config --repo $rootfs/ostree/repo set integrity.composefs required` ?

cgwalters:
(Just rebased this on current main for the very minor reason that switching branches across this forces a rebuild due to configure.ac bumping the release)

cgwalters:
OK, awesome work on figuring out all the bits necessary in coreos-assembler!  I did a bit of work today "productizing" and streamlining things more with

- https://github.com/coreos/rpm-ostree/pull/4422
- https://github.com/coreos/coreos-assembler/pull/3483

Once we land some bits of this, that will help us test this in CI here conveniently.

alexlarsson:
> > Do partial checkouts (only etc) when deploying using composefs
> 
> Bigger picture though this leads to the question of whether one wants a mutable/persistent `/etc` at all right? I think we should definitely add first-class support to ostree for making `/etc` a transient (writable but goes away on reboot) overlayfs at least.

Yes, this is also something we're currently shopping for answers too. I'm not sure we want to mix it up in this PR though.

alexlarsson:
> Hmm while we're here, probably `ex-fsverity` shouldn't be its own group, but be under an `integrity` group? And `ex-composefs` should also imply `ex-verity`.

Yeah, having an integrity group seems good to me. 

However, note that the exact use of global settings in this PR is a bit short term. It makes a lot of sense to have global "deploy as composefs" switch, but this PR uses the same setting to determine at commit time whether to add the composefs digest/signature to the commit. IMHO that part should really be done using arguments to "ostree commit", but I took a shortcut in the PR to avoid having any public API changes. 

So, in the current setup we don't want composefs to imply verity is required, because we want to be able to set composefs to true on the build machine (where verity is not needed). However, for the deploy side, yes, i think composefs should at least imply fsverity=maybe.

alexlarsson:
Rebased again, because it seems some stuff was left out from your rebase.

alexlarsson:
So, proposed new confuration options:

```
[integrity]
fsverity=yes/no/maybe # This is same as ex-fsverity
composefs=yes/no/maybe # default no
composefs-apply-sig=yes/no # default yes

composefs-add-metadata=yes/no # default no
composefs-keyfiile=/a/path
composefs-certfile=/a/path
```

Details:
* Composefs implies ex-fsverity=maybe. We do want to support systems with no fs-verity and at least get something booted and maybe other features like file sharing with containers down the line
* We need a way to disable applying signatures to the composefs image, because we want to be able to use a signed image on a system without the public key. If the signature is set without the public key in the keyring, then the fsverity enable call will fail. It should be on by default though.
* composefs-add-metatdata and keyfile/certfile affects whether composefs metadata (digest/signature) is set during ostree commit. You'd set this only on build repos. Long term we probably want this to be options to "ostree commit", but that will require library API changes.

Opinions?

alexlarsson:
@cgwalters `ex-` prefix or not?

cgwalters:
Maybe it's `ex-integrity` for now?  That avoids a need to `ex-` all the individual keys.

cgwalters:
One thing I tinkered with but didn't get to work is linking libcomposefs as a static library.  Having dynamic shared library dependencies, particularly for ones that might change API/ABI IMO really need to carry their weight.  It'd help a lot if we could just link libcomposefs statically, at least optionally.

alexlarsson:
I don't want to stream it from a tar, because then I would need to stream all file content data during deploy which is very expensive. All I need is the fs-verity digests, and typically those are available for reading from the object files themselves (as ex-fsverity is enabled).

alexlarsson:
If you want to avoid dependency, we could set up libcomposefs as a submodule.

alexlarsson:
New version, now using the new config option name, and using composefs as a submodule.
To test this you need some changes to the above docs:
create_disk.sh now just needs:
```
ostree config --repo $rootfs/ostree/repo set ex-integrity.composefs true
```

And the build config we need are:
```
sudo ostree config --repo cache/repo-build set ex-integrity.composefs-add-metadata true
sudo ostree config --repo=cache/repo-build set ex-integrity.composefs-certfile /srv/cert.pem
sudo ostree config --repo=cache/repo-build set ex-integrity.composefs-keyfile /srv/key.pem
```

alexlarsson:
Hmm, seems like CI build work up until distcheck. Not sure why.

alexlarsson:
I copied the autogen.sh workaround for libglnx and bsdiff, and that seems to fix it. Can't say I understand what it does though.

cgwalters:
> That only works if we know the fs-verity digest for the object files though, otherwise we have to stream the data anyway, because the code to compute the fs-verity digest in that case is in libcomposefs.

Ah, right.  We can't just use the kernel code here.  OK so the general case of "shared library via a binary" here is speaking a cross-language-friendly IPC to it over a `socketpair` or so.  That's how we do container fetches from `skopeo`.

I'm not going to argue really strongly for this right now; you put a lot of work in the shared library.  But dealing with FFI (and we'll want that at some point for Rust and Go) is painful.  Static linking will help.

Hmm that said, the code in `lcfs-fsverity.c` is about 50% a copy/paste of sha256, the block computation logic wouldn't be too bad to rewrite in Rust/Go or whatever's driving this.

Anyways, the static linking for now will greatly improve ergonomics.

src/libostree/ostree-repo-composefs.c
```diff
@@ -0,0 +1,632 @@
...
+  g_autoptr (GBytes) sig = NULL;
+
+  certfile = g_key_file_get_string (self->config, integrity_section, "composefs-certfile", NULL);
+  keyfile = g_key_file_get_string (self->config, integrity_section, "composefs-keyfile", NULL);
+
+  if (certfile == NULL && keyfile == NULL)
+    return TRUE;
+
+  if (certfile == NULL)
+    return glnx_throw (error, "Error signing compoosefs: keyfile specified but certfile is");
```
cgwalters, line 581: s/is/is not/

src/libostree/ostree-repo-composefs.c
```diff
@@ -0,0 +1,632 @@
...
+
+  if (keyfile == NULL)
+    return glnx_throw (error, "Error signing compoosefs: certfile specified but keyfile is");
```
cgwalters, line 584: s/is/is not/

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -614,6 +615,83 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
                                 cancellable, error))
     return FALSE;
 
+#ifdef HAVE_COMPOSEFS
+  if (repo->composefs_wanted != OT_TRISTATE_NO)
+    {
+      static const char integrity_section[] = "ex-integrity";
```
cgwalters, line 621: This one is probably better as
`#define _OSTREE_INTEGRITY_SECTION "ex-integrity"` in ostree-repo-private.h

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -614,6 +615,83 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
...
+      if (!glnx_open_tmpfile_linkable_at (osdeploy_dfd, checkout_target_name, O_WRONLY | O_CLOEXEC,
+                                          &tmpf, error))
+        return FALSE;
+
+      if (!ostree_composefs_target_write (target, tmpf.fd, &fsverity_digest, cancellable, error))
+        return FALSE;
+
+      /* If the commit specified a composefs digest, verify it */
+      if (metadata_composefs
+          && g_variant_n_children (metadata_composefs) == OSTREE_SHA256_DIGEST_LEN
```
cgwalters, line 667: Shouldn't it be an error if the digest length doesn't match?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -614,6 +615,83 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
...
+          && memcmp (fsverity_digest, g_variant_get_data (metadata_composefs),
+                     OSTREE_SHA256_DIGEST_LEN)
+                 != 0)
+        return glnx_throw (error, "Generated composefs image digest doesn't match commit metadata");
```
cgwalters, line 671: It'd be a bit friendlier to output the mismatched hex digests potentially?  Also, with stuff like this I think it's worth having a helper function to compare digests because it's security relevant.  See e.g. `_ostree_compare_object_checksum()`.  And the logic for what to do if the length of the input byte array doesn't match would probably be there.

src/libotutil/ot-keyfile-utils.c
```diff
@@ -60,6 +60,62 @@ ot_keyfile_get_boolean_with_default (GKeyFile *keyfile, const char *section, con
...
+        }
+
+      g_propagate_error (error, temp_error);
+      return FALSE;
+    }
+
+  /* Count the number of non-whitespace characters */
+  for (i = 0; ret_value[i]; i++)
+    if (!g_ascii_isspace (ret_value[i]))
+      length = i + 1;
```
cgwalters, line 98: Should this use `g_strstrip()` instead?

alexlarsson:
@cgwalters I got some kind of failure in tests/test-concurrency.py. Is this a flake, or do you think it is related to my changes?

cgwalters (commented):
A few minor nits

src/libotutil/ot-keyfile-utils.c
alexlarsson: I just copied what parse_boolean() did, but strstrip seems useful.

cgwalters:
> @cgwalters I got some kind of failure in tests/test-concurrency.py. Is this a flake, or do you think it is related to my changes?

Flake unfortunately https://github.com/ostreedev/ostree/issues/2038

alexlarsson:
Would you be ok with at least building the composefs codepath in one of the CI jobs?

cgwalters:
Also did https://github.com/coreos/rpm-ostree/pull/4428

cgwalters:
OK, looks like this breaks the non-composefs path.  The ISO tests are failing with
> [    2.624613] ostree-prepare-root[589]: ostree-prepare-root: failed to bind mount /sysroot/boot to boot: No such file or directory

I think it'd help to factor out the non-composefs changes to ostree-prepare-root to a separate PR.

cgwalters:
Ah, I think I saw the typo.  Pushed a possible fix...

alexlarsson:
> Ah, I think I saw the typo. Pushed a possible fix...

Ah, good catch. Once that PR lands I will rebase this.

cgwalters:
So...I'm on vacation this week.  I'm OK to merge this mainly as is if you want to, but I think we should make some time to come up with a more proper design doc with different phases.  The comment here https://github.com/coreos/fedora-coreos-tracker/issues/1252#issuecomment-1560778408 seems valid.

I'd also like to more deeply investigate bits around https://github.com/containers/composefs/issues/125 

I do see the value in making this "incrementally adoptable".  Even for what I'd call "unlocked" systems (i.e. user-is-sysadmin, traditional Windows/Linux desktop) composefs definitely has value in simply making things more resilient - you can't just `mount -o remount,rw /usr` and corrupt things anymore.  (And corruption can be detected by the kernel directly).

But it's not clear to me that in order to achieve that "incrementality" we need to conform so much to what ostree did from the start.  We have an opportunity now to re-think and clean things up and I think we should take it.

A simple way to say this is that I see this more like "composefs *replacing* the ostree core" and not "ostree using composefs" if that makes sense.

alexlarsson:
So, I'm of two minds here. 

In my practical mind, I want to use composefs with ostree in a product that is rhel9 based, and soon. This means it can't really diverge too much from existing setups, especially as there are a bunch of other parts of this I have to juggle (like the upstream kernel work). 

But, in my ideal approach mine, I think that a fresh start of deploy with a pure composefs-only deploy mechanism could be so much nicer. To do that right will really be much more work though, especially if you think of it in more grandiose terms like https://github.com/containers/composefs/issues/125.

I think the best approach is to go with "use composefs from ostree" for now (i.e. this PR), changing as little as we have to, and then start doing the long term visioneering for the "replace ostree with composefs" in parallel. Having this split means we can maybe handle the fact that there may be warts in the current composefs approach (like the fact that we also do the checkout atm).

alexlarsson:
Rebased this on top of https://github.com/ostreedev/ostree/pull/2864 so I could get something testable on rhivos, which seems to work now with some handholding. We can discuss further next week.

alexlarsson:
So, when i boot this on a system without a `sysroot.readonly=true` configuration, i hit this part in ostree-remount.c:

```
  if (path_is_on_readonly_fs ("/"))
    {
      /* If / isn't writable, don't do any remounts; we don't want
       * to clear the readonly flag in that case.
       */
      exit (EXIT_SUCCESS);
```

This means the sysroot is not made writable, and thus /var.

I'm not sure exactly what this check is for? Clearly with composefs the root will always be read-only. I guess i need to special case composefs in this check.

alexlarsson:
There are some more trickery too. With normal ostree, i had to have / in the fstab, otherwise systemd-remount-fs.service will not remount / read-only. However, if that is there when composefs is active the remount will fail, because it is a read-only fs and can't be made read-write. So, to make it work I made the fstab contain /sysroot for the main fs in the composefs case.

alexlarsson:
Here is a WIP example of using this with centos9 automotive: 
  https://gitlab.com/CentOS/automotive/sample-images/-/merge_requests/333

cgwalters:
> I think the best approach is to go with "use composefs from ostree" for now (i.e. this PR), changing as little as we have to, and then start doing the long term visioneering for the "replace ostree with composefs" in parallel.

Yes, definitely agreed.  I'm overall OK with landing initial experimental support here.  Basically though I think we do need to also in parallel spend a bit more time fleshing out something closer to the "end state" - having that vision/roadmap will help inform shorter term decisions and avoid painting ourselves into a corner.

alexlarsson:
Cool. I'm almost to the point that I have a working automotive build including signatures, and expect to get it ready this week. I'd like to avoid merging this until that is all working, but I would like to merge it by then.

cgwalters:
OK I filed https://github.com/ostreedev/ostree/issues/2867 which tries to navigate this a bit.  We still need to fill in and flesh out some important details - one of the biggest in my mind being how rollbacks work.

Can we iterate on design questions on that issue and leave this PR for "phase 0" implementation nits/details?  Please do feel free to directly edit the proposed issue text in addition to commenting.

Let's aim to do a [public meeting](https://meet.gnome.org/col-hab-hej-e5i) say this week?  Maybe Friday 9:30am EST?  Or earlier could work for me.

alexlarsson:
I have a rebased version of this almost ready, but i'm finishing up some details. Hopefully done soon.

alexlarsson:
Ok, I verified this patchset both with FCOS as well as the rhivos manifest, the later including both a signed commit (but not actually a signed initrd) as well as an update cycle.

I think this is good to go in as a "version zero" experimental support.

src/libostree/ostree-repo-composefs.c
```diff
@@ -0,0 +1,631 @@
...
+      guchar *known_digest = NULL;
+
+#ifdef HAVE_LINUX_FSVERITY_H
+      /* First try to get the digest directly from the bare repo file.
+       * This is the typical case when we're pulled into the target
+       * system repo with verity on and are recreating the composefs
+       * image during deploy. */
+      char buf[sizeof (struct fsverity_digest) + OSTREE_SHA256_DIGEST_LEN];
+
+      if (G_IS_UNIX_INPUT_STREAM (input))
```
cgwalters, line 328: I think we should change this to an assertion, and then just error out very early on if we're not operating on a bare mode repository.

src/libostree/ostree-repo-composefs.c
```diff
@@ -0,0 +1,631 @@
...
+        return glnx_throw_errno (error);
+
+      guchar *known_digest = NULL;
+
+#ifdef HAVE_LINUX_FSVERITY_H
+      /* First try to get the digest directly from the bare repo file.
+       * This is the typical case when we're pulled into the target
+       * system repo with verity on and are recreating the composefs
+       * image during deploy. */
+      char buf[sizeof (struct fsverity_digest) + OSTREE_SHA256_DIGEST_LEN];
```
cgwalters, line 326: Maybe a best practice to use `FS_VERITY_MAX_DIGEST_SIZE` here instead of `OSTREE_SHA256_DIGEST_LEN`?

(We could also add a static assertion that the latter is less than the former)

src/libostree/ostree-repo-composefs.c
```diff
@@ -0,0 +1,631 @@
...
+
+static gboolean
+checkout_composefs_recurse (OstreeRepo *self, const char *dirtree_checksum,
+                            const char *dirmeta_checksum, struct lcfs_node_s *parent,
+                            const char *name, GCancellable *cancellable, GError **error)
+{
+  g_autoptr (GVariant) dirtree = NULL;
+  g_autoptr (GVariant) dirmeta = NULL;
+  g_autoptr (GVariant) xattrs = NULL;
+  struct lcfs_node_s *directory;
```
cgwalters, line 366: Here and everywhere - in newer ostree code we prefer C99-style declare-and-initialize.

src/libostree/ostree-repo-composefs.c
```diff
@@ -0,0 +1,631 @@
...
+ * how the image is created.
+ *
+ * Returns: %TRUE on success, %FALSE on failure
+ */
+gboolean
+ostree_repo_checkout_composefs (OstreeRepo *self, OstreeComposefsTarget *target,
+                                OstreeRepoFile *source, GCancellable *cancellable, GError **error)
+{
+#ifdef HAVE_COMPOSEFS
+  char *root_dirs[] = { "usr", "etc", "boot", "var", "sysroot" };
```
cgwalters, line 528: We'll probably want this to be more configurable in the future.

src/libostree/ostree-repo-composefs.c
```diff
@@ -0,0 +1,631 @@
...
+    {
+      dir = ensure_lcfs_dir (root, root_dirs[i], error);
+      if (dir == NULL)
+        return FALSE;
+    }
+
+  return TRUE;
+#else
+  g_set_error (error, G_IO_ERROR, G_IO_ERROR_NOT_SUPPORTED,
+               "Composeefs is not supported in this ostree build");
```
cgwalters, line 557: s/composeefs/composefs/

(Also this error is duplicated in a few places, let's add a single helper function to set the error for this case)

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -582,13 +582,45 @@ merge_configuration_from (OstreeSysroot *sysroot, OstreeDeployment *merge_deploy
...
+  const guchar *expected_digest;
+
+  if (metadata_composefs == NULL)
+    return TRUE;
+
+  if (g_variant_n_children (metadata_composefs) != OSTREE_SHA256_DIGEST_LEN)
+    return glnx_throw (error, "Expected composefs fs-verity in metadata has the wrong size");
+
+  expected_digest = g_variant_get_data (metadata_composefs);
+  if (memcmp (fsverity_digest, expected_digest, OSTREE_SHA256_DIGEST_LEN) != 0)
```
cgwalters, line 598: It may be good to have a helper function for this that centralizes our checksum comparisons.

src/libostree/ostree-sysroot.c
```diff
@@ -799,14 +799,26 @@ parse_deployment (OstreeSysroot *self, const char *boot_link, OstreeDeployment *
   if (looking_for_booted_deployment)
     {
       struct stat stbuf;
+      struct stat etc_stbuf = {};
```
cgwalters, line 802: Why the `{}` just for this one?

src/switchroot/ostree-mount-util.h
```diff
@@ -121,4 +129,65 @@ touch_run_ostree (void)
   (void)close (fd);
 }
 
+static inline unsigned char *
+read_file (const char *path, size_t *out_len)
+{
+  int fd;
+
+  fd = open (path, O_RDONLY);
```
cgwalters, line 137: `O_CLOEXEC` too...

src/switchroot/ostree-mount-util.h
```diff
@@ -121,4 +129,65 @@ touch_run_ostree (void)
   (void)close (fd);
 }
 
+static inline unsigned char *
+read_file (const char *path, size_t *out_len)
```
cgwalters, line 133: Hmm maybe we should just link in glib again here.

cgwalters (commented):
Just nits

cgwalters:
Test just failed on a single flake.

cgwalters:
> I think this is good to go in as a "version zero" experimental support.

Yes, agreed.  There'll be a lot of followup, but it'd be good to start making it easier to play with "hands on" in git main.

cgwalters (approved):

PR #2641, opened 2022-06-09
tests/inst: Bump the version of ostree-ext

cgwalters:
In the interest of cross-testing and keeping things up to date.

Hmm, I think we need to set up dependabot here.


jmarrero (approved):
lgtm

PR #2644, opened 2022-06-09
 fsck: Don't load all object names into memory

cgwalters:
repo: Document non-obvious way to list all commits

I was going to add an API for this and then realized the empty
string does it.

---

fsck: Don't load all object names into memory

We recently discovered `list_objects()` is inefficient with memory.
The more efficient `list_objects_set()` API isn't yet public, but
this fsck code actually just skips over non-commit objects, and
we already have an API to list just those.

---

### DIFF
--- a/src/ostree/ot-builtin-fsck.c
+++ b/src/ostree/ot-builtin-fsck.c
@@ -282,16 +282,15 @@ ostree_builtin_fsck (int argc, char **argv, OstreeCommandInvocation *invocation,
     }
 
   if (!opt_quiet)
-    g_print ("Enumerating objects...\n");
+    g_print ("Enumerating commits...\n");
 
-  g_autoptr(GHashTable) objects = NULL;
-  if (!ostree_repo_list_objects (repo, OSTREE_REPO_LIST_OBJECTS_ALL,
-                                 &objects, cancellable, error))
+  // Find all commit objects, including partial ones
+  g_autoptr(GHashTable) all_commits = NULL;
+  if (!ostree_repo_list_commit_objects_starting_with (repo, "", &all_commits, cancellable, error))
     return FALSE;
-
+  // And gather a set of non-partial commits for further analysis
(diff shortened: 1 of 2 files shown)


src/ostree/ot-builtin-fsck.c
```diff
@@ -313,7 +312,9 @@ ostree_builtin_fsck (int argc, char **argv, OstreeCommandInvocation *invocation,
 
       ostree_object_name_deserialize (serialized_key, &checksum, &objtype);
 
-      if (objtype == OSTREE_OBJECT_TYPE_COMMIT)
+      g_assert (objtype == OSTREE_OBJECT_TYPE_COMMIT);
+
+      if (TRUE)  // Temporarily avoiding the noise of de-indenting this whole thing
```
dbnicholson, line 317: I'd personally vote for taking the re-indentation noise since I don't think this is going to be temporary. Whether it's `list_commits_starting_with` now or `list_objects_set` later, I think this program is going to stick with some method that only scans for commit objects (with or without the above `assert`). Not a blocker, though.

dbnicholson:
I was also thinking to add a `list_commit_objects` API that's just a wrapper around `list_commit_objects_starting_with` using an empty string, but it's definitely not a blocker. I haven't seen this `list_objects_set` API yet, but I presume it takes an object types field? If that's imminent, then I definitely don't think it's worth adding API if there's something that totally supplants it.

dbnicholson (approved):
This looks good to me. Feel free to do the re-indentation or not.

cgwalters:
>  I haven't seen this list_objects_set API yet, but I presume it takes an object types field?

It's not public but see `ostree_repo_list_objects_set`.

(That said, in this fsck case and I think others, what we really want is an iterator, which would *entirely* obviate memory usage.)

cgwalters:
Will do the indentation as a followup to save good CI runs.

PR #2646, opened 2022-06-12
rust-bindings: Wire up `tests/`

cgwalters:
Because the source is in a subdirectory, we lose out on cargo target
autodiscovery.

I noticed this when I edited one of the tests in a way that
should have failed, but didn't...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,10 @@ features = ["dox"]
 name = "ostree"
 path = "rust-bindings/src/lib.rs"
 
+[[test]]
+name = "integration"
+path = "rust-bindings/tests/tests.rs"
+
 [workspace]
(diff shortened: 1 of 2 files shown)


lucab:
LGTM, but I do wonder why we don't have direct nesting of manifests instead? That is, leaving the top-level manifest almost empty and moving the real juice to `rust-bindings/Cargo.toml`.

lucab (approved):

cgwalters:
> LGTM, but I do wonder why we don't have direct nesting of manifests instead? That is, leaving the top-level manifest almost empty and moving the real juice to rust-bindings/Cargo.toml.

I think I tried that but ran into some problem, but you're probably right that we should try harder to make that work.

PR #2655, opened 2022-06-23
s390x: rename sd-boot to sdboot

nikita-dubrovskii:
same name is used by `coreos-installer`


jlebon (approved):

jlebon:
https://github.com/ostreedev/ostree/pull/2658 for the CI failure.

jlebon (approved):

jlebon:
CI now needs https://github.com/ostreedev/ostree/pull/2659. But... this is affecting s390x only and CI doesn't test that, so let's just get this in.

PR #2656, opened 2022-06-23
rust: Bump semver to 0.15 && bump cap-std 0.25

cgwalters:
ci/rust: Enable `cap-std-apis` in default build, add a no-feature build

Our CI was missing coverage of `cap-std-apis`.

---

ci/rust: Change MSRV to `cargo check`

No reason to codegen just to throw it away.  We could test here too,
but eh.

---

Fix clippy lint in cap-std bits

---

rust: Bump semver to 0.15

Prep for some breaking changes.

---

Bump to cap-std 0.25 and io-lifetimes 0.7

Prep for bumping ostree-rs-ext, which will help bump rpm-ostree,
which will get it out of having two copies of rustix.

---

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
-  CARGO_PROJECT_FEATURES: "v2021_5"
+  CARGO_PROJECT_FEATURES: "v2021_5,cap-std-apis"
   # TODO: Automatically query this from the C side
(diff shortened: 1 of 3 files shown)


jmarrero (approved):
lgtm

cgwalters:
Oh this doesn't work; I tested it locally with `cargo check --feature=cap-std`, not `--feature=cap-std-apis` which actually enables things.

jmarrero (approved):
new changes lgtm too

PR #2657, opened 2022-06-23
ci/rust: Enable `cap-std-apis` in default build && Change MSRV to `cargo check`

cgwalters:
ci/rust: Enable `cap-std-apis` in default build, add a no-feature build

Our CI was missing coverage of `cap-std-apis`.

---

ci/rust: Change MSRV to `cargo check`

No reason to codegen just to throw it away.  We could test here too,
but eh.

---

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


cgwalters:
Oh fun, the expanded coverage triggers clippy to hit a new lint.  Let's just roll this into https://github.com/ostreedev/ostree/pull/2656

PR #2659, opened 2022-06-23
tests/staged-deploy.sh: Hack around cosa systemd unit check

cgwalters:
https://github.com/coreos/coreos-assembler/pull/2921 broke this
test which is intentionally causing a systemd unit to fail.

As they say, necessity is the mother of invention.  They don't
say though that need always causes particularly *beautiful* things
to be invented...


jlebon (approved):
Just fixed an ordering issue. LGTM!

PR #2660, opened 2022-06-23
repo: Metadata return values from `load_file` are not nullable

cgwalters:
The pattern this API uses in C is to allow the input parameters
pointer targets to be `NULL`, and it doesn't return values in that
case.

A further complexity here is that the API will still return `NULL`
for symbolic links.

But Rust can't express this pattern as is, so we were always
returning values but in `Option<T>` wrappers that the caller needed
to unwrap for the metadata.

(We really want an even more efficient API here that avoids the glib
 objects entirely, e.g. no reason not to pass directly back a type
 that lets Rust directly read from the fd for bare repos, but
 that can come later)


jmarrero (approved):
lgtm

cgwalters:
Fixed in main

Issue #2647, opened 2022-01-17
rust-bindings/cargo: bump to gtk-rs 0.15


lucab:
Next iteration of the gtk-rs stack is out, see https://gtk-rs.org/blog/2022/01/16/new-release.html.
This effectively means bumping glib/gio/gobject to 0.15.

However, this comes with an hard requirement for a minimum Rust toolchain >= 1.56.
Holding this for now while we verify that such toolchain requirement is fine for the environments we target.
Also, this bumps the C native dependencies:
> GLib/GObject/GIO: minimum 2.48, latest 2.72

PR #2663, opened 2022-06-30
lib/static-delta: document and check parameters format

mstuehn:
This enhances the logic handling GVariant parameters within
`ostree_repo_static_delta_generate()`.
Several of those entries are expected to be zero-terminated values,
and this implicit assumption has been observed to be an hidden trap
in languages where strings and arrays may not carry a terminator value
(e.g. Rust).
In order to improve the situation, this makes the documentation more
explicit and actively tries to catch invalid input parameters.

Closes: https://github.com/ostreedev/ostree/issues/2728

---
[Original PR text below here]

This commit adds a test for the rust interface.
It should test the static-delta generation into a specified file.
Unfortunately it does not work because the ostree-lib throws an error:

```
---- repo::generate_static::should_generate_static_delta_at stdout ----
thread 'repo::generate_static::should_generate_static_delta_at' panicked at
'static delta generate: Error { domain: g-io-error-quark, code: 1, message: "renameat(./tmp.ReHre1, ): No such file or directory" }', 
rust-bindings/tests/repo/generate_static.rs:39:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It seems, that there is something wrong with the GVariant of filename, because
[`if (!g_variant_lookup (params, "filename", "^&ay", &opt_filename))`](https://github.com/ostreedev/ostree/blob/e527cdc5826ffb7c06ebf39fb3ab70482699eb8e/src/libostree/ostree-repo-static-delta-compilation.c#L1411) returns "true", but `opt_filename` is empty (see the empty 2nd parameter of [`renameat()`](https://gitlab.gnome.org/GNOME/libglnx/-/blob/master/glnx-fdio.h#L365) in the error message).

### DIFF
--- a/rust-bindings/tests/repo/generate_static.rs
+++ b/rust-bindings/tests/repo/generate_static.rs
@@ -14,8 +14,10 @@ fn should_generate_static_delta_at() {
     let delta_path = delta_dir.path().join("static_delta.file");
     let path_var = delta_path
         .to_str()
+        .map(std::ffi::CString::new)
         .expect("no valid path")
-        .as_bytes()
+        .unwrap()
+        .as_bytes_with_nul()
         .to_variant();
 
     let test_repo = TestRepo::new();
@@ -38,5 +40,5 @@ fn should_generate_static_delta_at() {
         )
         .expect("static delta generate");
 
-    assert!(std::fs::File::open(&delta_path).is_err());
+    assert!(delta_path.try_exists().unwrap());
 }
--- a/rust-bindings/tests/repo/generate_static.rs
+++ b/rust-bindings/tests/repo/generate_static.rs
@@ -1,0 +1,42 @@
+use crate::util::*;
+use gio::NONE_CANCELLABLE;
+use ostree::glib::prelude::*;
+use ostree::glib::Variant;
+use ostree::*;
+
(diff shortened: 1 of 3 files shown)


lucab:
Thanks for the report and the reproducer.
I think there is a problem with the `filename` bytestring variant, which is supposed to be nul-terminated.
As you directly transform path -> string -> bytes -> variant, there is no terminator in the value.

If you try using a `CString` instead it should get past the renaming error:
```rust
    let path_var = delta_path
        .to_str()
        .map(std::ffi::CString::new)
        .expect("no valid path")
        .unwrap()
        .as_bytes_with_nul()
        .to_variant();
```

I do agree this is not very ergonomic.

lucab:
Hey @mstuehn, I did push a couple of commits on top of your PR here. There is now some logic to detect cases where the zero-terminator is missing, and the test has been tweaked to show how to approach it in Rust.
Let me know your feedback, but I think we can go ahead and merge this.
I think we could gain better UX in the future once we update the bindings for https://github.com/ostreedev/ostree/issues/2647.

mstuehn:
Hi @lucab 
the patch looks good to me. Thanks for the merge.

cgwalters (approved):

PR #2665, opened 2022-07-01
deploy: Make bwrap dependency soft

cgwalters:
Let's support people who are building with SELinux enabled, but aren't
shipping bwrap.

(Though, those people probably want this to work)


cgwalters:
(only compile tested locally)

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2967,9 +2967,20 @@ sysroot_finalize_selinux_policy (int deployment_dfd, GError **error)
...
-                          SEMODULE_HELP_ARGC, &exit_status, &stdout, error))
-    return FALSE;
+  {
+    g_autoptr(GError) local_error = NULL;
+    if (!run_in_deployment (deployment_dfd, SEMODULE_HELP_ARGV,
+                            SEMODULE_HELP_ARGC, &exit_status, &stdout, &local_error))
+      {
+        if (g_error_matches (local_error, G_SPAWN_ERROR, G_SPAWN_ERROR_NOENT))
+          {
+            ot_journal_print (LOG_INFO, "No bwrap available");
```
jlebon, line 2977: A better log message here would be nice. Maybe

```suggestion
            ot_journal_print (LOG_INFO, "Skipping `semodule --rebuild-if-modules-changed`: no bwrap available");
```

?

PR #2666, opened 2022-07-06
ostree-repo: bls-append-except-default followup

saqibali-2k:
This PR is followup from https://github.com/coreos/coreos-assembler/pull/2863
Summary of changes:
- Moved bls-append-except-default parsing logic to `reload_sysroot_config()`
- Made sure heap allocated memory is being free'd

### DIFF
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -224,6 +224,7 @@ struct OstreeRepo {
   gint fs_support_reflink; /* The underlying filesystem has support for ioctl (FICLONE..) */
   gchar **repo_finders;
   OstreeCfgSysrootBootloaderOpt bootloader; /* Configure which bootloader to use. */
+  GHashTable *bls_append_values; /* Parsed key-values from bls-append-except-default key in config. */
 
   OstreeRepo *parent_repo;
 };
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -1238,6 +1238,7 @@ ostree_repo_finalize (GObject *object)
   g_mutex_clear (&self->txn_lock);
   g_free (self->collection_id);
   g_strfreev (self->repo_finders);
+  g_clear_pointer (&self->bls_append_values, g_hash_table_unref);
 
   g_clear_pointer (&self->remotes, g_hash_table_destroy);
   g_mutex_clear (&self->remotes_lock);
@@ -1412,6 +1413,9 @@ ostree_repo_init (OstreeRepo *self)
   self->remotes = g_hash_table_new_full (g_str_hash, g_str_equal,
                                          (GDestroyNotify) NULL,
                                          (GDestroyNotify) ostree_remote_unref);
+  self->bls_append_values = g_hash_table_new_full (g_str_hash, g_str_equal, 
+                                                   (GDestroyNotify) g_free, 
+                                                   (GDestroyNotify) g_free);
   g_mutex_init (&self->remotes_lock);
 
   self->repo_dir_fd = -1;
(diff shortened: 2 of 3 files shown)


src/libostree/ostree-repo.c
```diff
@@ -3509,16 +3512,57 @@ reload_sysroot_config (OstreeRepo          *self,
...
         {
           self->bootloader = (OstreeCfgSysrootBootloaderOpt) i;
-          return TRUE;
+          valid_bootloader = TRUE;
+        }
+    }
+  if (!valid_bootloader) 
+  {
+    return glnx_throw (error, "Invalid bootloader configuration: '%s'", bootloader);
+  }
```
cgwalters, line 3527: Can drop the braces here (which are off indentation)

src/libostree/ostree-repo.c
```diff
@@ -3509,16 +3512,57 @@ reload_sysroot_config (OstreeRepo          *self,
...
+  /* Parse bls-append-except-default string list. */
+  g_autoptr(GError) local_error = NULL;
+  g_auto(GStrv) read_values = g_key_file_get_string_list (self->config, "sysroot", "bls-append-except-default", NULL, &local_error);
+  /* We can ignore not found errors */
```
cgwalters, line 3531: There is also `ot_keyfile_get_string_list_with_default()` for this.

src/libostree/ostree-repo.c
```diff
@@ -3509,16 +3512,57 @@ reload_sysroot_config (OstreeRepo          *self,
...
+          g_clear_error (&local_error);
+        }
+      else
+        {
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return FALSE;
+        }
+    }
+  /* get all key value pairs in bls-append-except-default */
+  g_hash_table_unref (self->bls_append_values);
```
cgwalters, line 3547: It seems simpler to me to invoke [g_hash_table_remove_all](https://developer-old.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-remove-all) - then we don't have a period of time where `self->bls_append_values` is pointing to freed memory.

src/libostree/ostree-repo.c
```diff
@@ -3509,16 +3512,57 @@ reload_sysroot_config (OstreeRepo          *self,
...
+  self->bls_append_values = g_hash_table_new_full (g_str_hash, g_str_equal,
```
cgwalters, line 3548: And then we could drop this duplicate re-initialization.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2082,48 +2082,19 @@ install_deployment_kernel (OstreeSysroot   *sysroot,
...
-    }
-
   /* Only append to this BLS config if:
    * - this is not the default deployment
    */
    /* If deployment was prepended, it is the new default */
   gboolean is_new_default = (ostree_deployment_get_index (deployment) == 0);
   gboolean allow_append = !is_new_default;
-  if (allow_append)
+  if (repo && allow_append)
```
cgwalters, line 2091: Why the `if (repo)`?  Having a repository is a required part of ostree here, there should be no situations where it's `NULL`.

(In Rust we'd have references, which are non-nullable by default)

If you're concerned about this, it'd probably be OK to drop a `g_assert (repo != NULL)` near the top of the function...oh.  Oh no.  Right, now I see:

```
  /* Need the repo to try and extract the versions for deployments.
   * But this is a "nice-to-have" for the bootloader UI, so failure
   * here is not fatal to the whole operation.  We just gracefully
   * fall back to the deployment index. */
  g_autoptr(OstreeRepo) repo = NULL;
  (void) ostree_sysroot_get_repo (self, &repo, cancellable, NULL);
```

That code just needs to change.

cgwalters: Can you rebase on https://github.com/ostreedev/ostree/pull/2667 ?

src/libostree/ostree-repo-private.h
```diff
@@ -224,6 +224,7 @@ struct OstreeRepo {
   gint fs_support_reflink; /* The underlying filesystem has support for ioctl (FICLONE..) */
   gchar **repo_finders;
   OstreeCfgSysrootBootloaderOpt bootloader; /* Configure which bootloader to use. */
+  GHashTable *bls_append_values; /* Parsed key-values from bls-append-except-default key in config. */
```
cgwalters, line 227: We do however need a `g_hash_table_unref()` for this inside `ostree_repo_finalize()`.

src/libostree/ostree-repo.c
cgwalters: This one looks like it wasn't fixed, but again, let's do https://github.com/ostreedev/ostree/pull/2565 at some point.

cgwalters (approved):

travier:
This is a follow up for https://github.com/ostreedev/ostree/pull/2605

PR #2672, opened 2022-07-13
sysroot: Have `ensure_writable` also always initialize

cgwalters:
For historical reasons we have a fair bit of distinct sysroot
initialization going on.  A lot of code is calling *just* the new
`ensure_writable()` API, which does basically what you'd expect...
except if we're not using a mount namespace.

Which is the case in unit tests and legacy setups.
Change this API to also ensure the sysroot is fully initialized
even in those cases.  Specifically we'll have `self->sysroot_fd`.

For now, callers that need `/boot` also need to separately
call `_ensure_boot_fd()`.


jmarrero (approved):
lgtm

PR #2673, opened 2022-07-13
sysroot: Add a few more assertions about `boot_fd`

cgwalters:
These places are all safe, but it would catch bugs in the future
more clearly to trip an assertion here.


jmarrero (approved):
lgtm

PR #2681, opened 2022-07-29
RFC: Support `/etc/ostree/config.d/` and use it to support user-configured top-level symlinks

jlebon:
The first commit adds support for users to drop configs in `/etc/ostree/config.d`. The second commit makes use of it to support users configuring top-level symbolic links. See individual commit messages for details.

Example using Fedora CoreOS:

```
$ cat config-toplevel-symlinks.bu
variant: fcos
version: 1.4.0
storage:
  disks:
    - device: /dev/disk/by-id/coreos-boot-disk
      partitions:
        - start_mib: 8000
          label: foobar
  filesystems:
    - device: /dev/disk/by-partlabel/foobar
      format: xfs
      with_mount_unit: true
      path: /var/foobar
  files:
    - path: /etc/ostree/config.d/01-foobar.conf
      contents:
        inline: |
          [toplevel-links]
          foobar=/var/foobar

$ cosa run -B config-toplevel-symlinks.bu -x 'sudo touch /foobar/bazboo && ls -l /foobar /foobar/'
lrwxrwxrwx. 1 root root 11 Jul 29 22:37 /foobar -> /var/foobar

/foobar/:
total 0
-rw-r--r--. 1 root root 0 Jul 29 22:37 bazboo
```

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -3195,6 +3195,97 @@ min_free_space_size_validate_and_convert (OstreeRepo    *self,
   return TRUE;
 }
 
+static int
+compare_strings (gconstpointer a, gconstpointer b)
+{
+  const char **sa = (const char **)a;
+  const char **sb = (const char **)b;
+
+  return strcmp (*sa, *sb);
+}
+
+#define SYSCONF_CONFIGS SHORTENED_SYSCONFDIR "/ostree/config.d"
+
+
+static gboolean
+load_user_configs (OstreeRepo    *self,
+                   GCancellable  *cancellable,
+                   GError       **error)
+{
+  g_clear_pointer (&self->user_config, g_key_file_unref);
+  g_autoptr(OstreeSysroot) sysroot_ref = (OstreeSysroot*)g_weak_ref_get (&self->sysroot);
+
+  /* Only read from /etc/ostree/config.d if we are pointed at a deployment */
+  if (sysroot_ref == NULL || sysroot_ref->is_physical)
+    return TRUE;
+
+  g_auto(GLnxDirFdIterator) dfd_iter = { 0, };
+  gboolean exists;
+  if (!ot_dfd_iter_init_allow_noent (sysroot_ref->sysroot_fd, SYSCONF_CONFIGS, &dfd_iter, &exists, error))
+    return FALSE;
+  /* Note early return */
+  if (!exists)
+    return TRUE;
+
+  g_autoptr(GPtrArray) configs = g_ptr_array_new_with_free_func (g_free);
+
+  while (TRUE)
+    {
+      struct dirent *dent;
+      if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
+        return FALSE;
+      if (dent == NULL)
+        break;
+
+      /* match remotes.d semantics */
+      if (dent->d_type != DT_REG || !g_str_has_suffix (dent->d_name, ".conf"))
+        continue;
+
+      /* parse it just to validate it */
+      g_autoptr(GKeyFile) k = g_key_file_new ();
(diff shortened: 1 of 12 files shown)


jlebon:
I encountered https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/41 while hacking on this, which took me some time to figure out!

cgwalters:
Today, the ostree-ext container logic discards everything not in `/usr` (or `/etc`).  But, we could choose to make it as easy to do this as:

```dockerfile
FROM quay.io/coreos-assembler/fcos:testing-devel
RUN mkdir /nix
```

cgwalters:
(hmm also really need to fix `ostree container commit` to warn about content outside of `/usr` and `/etc`, not just special case `/var`)

jlebon:
> Today, the ostree-ext container logic discards everything not in `/usr` (or `/etc`). But, we could choose to make it as easy to do this as:

Sure, I think this could make sense to support in the container flow. Do you think it obviates native support for it in libostree? The UX described in the demo above is quite streamlined in comparison. (There's also the question of whether other downstreams not planning to ship their OS as a container would be interested in this.)

src/libostree/ostree-repo.c
```diff
@@ -3195,6 +3195,97 @@ min_free_space_size_validate_and_convert (OstreeRepo    *self,
...
+        return glnx_throw_errno_prefix (error, "copying %s/%s", SYSCONF_CONFIGS, config);
+
+      /* make sure there's a newline */
+      if (glnx_loop_write (user_keyfile.fd, "\n", 1) < 0)
+        return glnx_throw_errno_prefix (error, "writing newline to tmpfile");
+    }
+
+  self->user_config = g_key_file_new ();
+
+  /* we want g_key_file_load_from_fd() but that's private */
```
cgwalters, line 3280: Mmm...I think I would say it's simpler to read the configs into e.g. a `GString` or a `GMemoryOutputStream`.

cgwalters:
> Do you think it obviates native support for it in libostree? 

I'd say there's clearly room for both.

> The UX described in the demo above is quite streamlined in comparison. 

I find this a bit more debatable :smile: 

> (There's also the question of whether other downstreams not planning to ship their OS as a container would be interested in this.)

True!

Basically I'm roughly :+1: if you want to drive this in.

jlebon:
>> The UX described in the demo above is quite streamlined in comparison.
>
> I find this a bit more debatable :smile: 

Heh fair. The `mkdir` is definitely nicer. Just the barrier to entry to get there is higher.

> Basically I'm roughly :+1: if you want to drive this in.

OK cool! In that case, I'll enhance it to support directories too. It'd be good to support anyway, but also I think I've ironed out a not-too-hacky way to support it on first boot via Ignition.

cgwalters:
I know I wrote this somewhere else but I can't find it right now:

Another approach that would work both client and server side is to support e.g. `/etc/ostree/toplevel` and `/usr/lib/ostree/toplevel`.

Then the UX flow (instead of frobbing a config file) for making a toplevel dir is just:
`mkdir -p /etc/ostree/toplevel/nix`

(OK and...hm, we'd need a command to re-reconcile this, might as well support it live; don't want to wait to do an upgrade to process it)

A key idea here is that the description of the root directory is captured in `/usr`.

cgwalters:
I'm still increasingly thinking we should relax the "content in /usr" restriction in ostree and just allow people to put at least empty directories wherever (except `/var`).

That said, today one can do this via a systemd unit with `ExecStart` e.g. in a systemd unit:
```
[Service]
Type=oneshot
ExecStart=chattr -i / && mkdir /afs && chattr +i /
```

Then one can have a regular mount unit mount it that's ordered `After=` that unit (or just also run `mount` from the service unit).

jlebon:
Closing in favour of the derivation approach.

PR #2682, opened 2022-07-31
s390x: Avoid bash dependency

akiernan:
bash is GPLv3 licensed and there's nothing in this script which really
needs it.

### DIFF
--- a/src/libostree/s390x-se-luks-gencpio
+++ b/src/libostree/s390x-se-luks-gencpio
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/bin/sh
 # This script creates new initramdisk with LUKS config within
-set -euo pipefail
+set -eu
 
 old_initrd=$1
 new_initrd=$2


cgwalters (approved):

cgwalters:
It's tempting to rewrite this in C and avoid a shell dependency at all.

cgwalters:
cc @nikita-dubrovskii

cgwalters:
A rebase should fix CI on this PR

nikita-dubrovskii:
> It's tempting to rewrite this in C and avoid a shell dependency at all.

yes, this is in plan (as now we have everything merged in f-c-c , cosa & rdcore), so will do that

nikita-dubrovskii:
Here is a PR: https://github.com/ostreedev/ostree/pull/2708

cgwalters:
Obsoleted by https://github.com/ostreedev/ostree/pull/2708 - but that said, thanks for the patch!

PR #2683, opened 2022-08-01
deny.toml: Add `Unicode-DFS-2016`

cgwalters:
This is used by the unicode crate now and is definitely a compatible
FOSS license.


lucab (approved):

cgwalters:
Will look at that, but not related

PR #2688, opened 2022-08-13
Update to `libtest-mimic` 0.5.0

LukasKalbertodt:
I just released [a new version of libtest-mimic](https://github.com/LukasKalbertodt/libtest-mimic/releases/tag/v0.5.0). This is one of the bigger projects using it, so as a sanity check whether the new version still works for real world users, I just did this update. It allowed to remove some of the code here as the API changed a bit.

Two things to note:
- `libtest-mimic` requires Rust 1.58. 
- I do not know about this project at all and did not test anything. Please review carefully and/or test.


cgwalters (approved):
Nice, thank you so much!

/ok-to-test

PR #2689, opened 2022-08-15
lib/commit: Directly use FICLONE for payload link

jlebon:
The idea of payload linking is to reflink between objects where
possible. Instead of relying on `glnx_regfile_copy_bytes` to hit the
`FICLONE` path, just call `FICLONE` directly. At that point in the code,
we've already established that the source and dest repos are on the same
filesystem and that it supports `FICLONE`.

Related: https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/41
Related: https://github.com/ostreedev/ostree/pull/2684#issuecomment-1204068437

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -881,11 +881,8 @@ _try_clone_from_payload_link (OstreeRepo   *self,
       else
         {
           /* This undoes all of the previous writes; we want to generate reflinked data.   */
-          if (ftruncate (tmpf->fd, 0) < 0)
-            return glnx_throw_errno_prefix (error, "ftruncate");
-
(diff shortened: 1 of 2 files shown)


cgwalters (approved):

jlebon:
OK fun, `ext.ostree.nondestructive.itest-payload-link.sh` passed, but now `ext.ostree.destructive.staged-deploy.sh` broke. Digging.

cgwalters:
> but now ext.ostree.destructive.staged-deploy.sh broke

That isn't related to this PR, it's a regression/breakage that came in I think a week or two ago.

jlebon:
Ack yup, seeing it locally on `main` too. Chasing it down.

jlebon:
Added a commit to fix that too now!

cgwalters (approved):

PR #2691, opened 2022-08-16
 cli/rev-parse: Add `--single` option

cgwalters:
cli/rev-parse: Add `--single` option

In the current "ostree native container" flow, we're inserting
a commit object into the repo but with no refs.

We have hacks in a few places to find the commit digest via e.g.
`find repo/objects -name *.commit` but that's a horrible hack.
Add `ostree rev-parse --single` which will print the single commit,
and error out if there is not exactly one commit.

### DIFF
--- a/src/ostree/ot-builtin-rev-parse.c
+++ b/src/ostree/ot-builtin-rev-parse.c
@@ -25,13 +25,17 @@
 #include "ot-builtins.h"
 #include "ostree.h"
 #include "otutil.h"
+#include <stdbool.h>
 
 /* ATTENTION:
  * Please remember to update the bash-completion script (bash/ostree) and
  * man page (man/ostree-rev-parse.xml) when changing the option list.
  */
 
+static gboolean opt_single;
+
 static GOptionEntry options[] = {
+  { "single", 'S', 0, G_OPTION_ARG_NONE, &opt_single, "If the repository has exactly one commit, then print it; any other case will result in an error", NULL },
   { NULL }
 };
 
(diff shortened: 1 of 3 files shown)


cgwalters:
Depends https://github.com/ostreedev/ostree/pull/2690

jlebon:
Hmm, any reason we don't create a ref in the repo? I think this has come up before too.

cgwalters:
The thing is, part of the goal of ostree native containers is to really be "container native".  ostree remotes and refs have a 100% overlap with container image references.

For projects which are starting now (and not like e.g. FCOS transitioning from pure ostree) they won't have a ref.  We could just of course make one up...I'm not opposed to that.  If someone feels even mildly strongly about doing so, OK.  But this patch still helps us and is useful today for existing container images, and doesn't conflict with a future change to add a ref.

src/ostree/ot-builtin-rev-parse.c
```diff
@@ -25,47 +25,79 @@
...
+            return glnx_throw (error, "Multiple commit objects found");
+          found = true;
+        }
+      if (!found)
+        return glnx_throw (error, "No commit objects found");
+      g_assert (key != NULL);
+      GVariant *serialized_key = key;
+      const char *checksum;
+      OstreeObjectType objtype;
+      ostree_object_name_deserialize (serialized_key, &checksum, &objtype);
```
jlebon, line 79: Hmm, why not use `GLNX_HASH_TABLE_FOREACH` here? Ahh because you want the key to live longer.

OK how about:

```suggestion
      GVariant *found = NULL;
      GLNX_HASH_TABLE_FOREACH (objects, GVariant*, key)
        {
          if (found)
            return glnx_throw (error, "Multiple commit objects found");
          found = key;
        }
      if (!found)
        return glnx_throw (error, "No commit objects found");
      const char *checksum;
      OstreeObjectType objtype;
      ostree_object_name_deserialize (found, &checksum, &objtype);
```

cgwalters:
Addressed comments

jlebon (approved):

Issue #2679, opened 2022-07-25
`ostree admin kargs edit-in-place` assertion reached when deployments are pending


travier:
See https://github.com/ostreedev/ostree/issues/2664:

```
Jul 25 13:57:17 fedora ostree-sysroot-readonly[2417]: + ostree admin kargs edit-in-place --append-if-missing=rw
Jul 25 13:57:17 fedora audit[2427]: ANOM_ABEND auid=4294967295 uid=0 gid=0 ses=4294967295 subj=system_u:system_r:unconfined_service_t:s0 pid=2427 comm="ostree" exe="/usr/bin/ostree" sig=6 res=1
Jul 25 13:57:17 fedora ostree-sysroot-readonly[2427]: **
Jul 25 13:57:17 fedora ostree-sysroot-readonly[2427]: OSTree:ERROR:src/libostree/ostree-sysroot-deploy.c:3618:ostree_sysroot_deployment_set_kargs_in_place: assertion failed: (!ostree_deployment_is_staged (deployment))
Jul 25 13:57:17 fedora ostree-sysroot-readonly[2427]: Bail out! OSTree:ERROR:src/libostree/ostree-sysroot-deploy.c:3618:ostree_sysroot_deployment_set_kargs_in_place: assertion failed: (!ostree_deployment_is_staged (deployment))
Jul 25 13:57:17 fedora systemd[1]: Created slice system-systemd\x2dcoredump.slice - Slice /system/systemd-coredump.
Jul 25 13:57:17 fedora audit: BPF prog-id=131 op=LOAD
Jul 25 13:57:17 fedora audit: BPF prog-id=132 op=LOAD
Jul 25 13:57:17 fedora audit: BPF prog-id=133 op=LOAD
Jul 25 13:57:17 fedora audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 subj=system_u:system_r:init_t:s0 msg='unit=systemd-coredump@0-2428-0 comm="systemd" exe="/usr/lib/systemd/systemd" hostname=? addr=? terminal=? res=>
Jul 25 13:57:17 fedora systemd[1]: Started systemd-coredump@0-2428-0.service - Process Core Dump (PID 2428/UID 0).
Jul 25 13:57:17 fedora [2430]: [164B blob data]
Jul 25 13:57:17 fedora systemd-coredump[2429]: [🡕] Process 2427 (ostree) of user 0 dumped core.
                                               
[...]

PR #2695, opened 2022-08-19
Fix `ostree admin kargs edit-in-place` assertion when deployments are pending

HuijingHei:
This is to support pending deployments instead of rasing assertion.
For example:
```
$ sudo rpm-ostree kargs --append=foo=bar
$ sudo ostree admin kargs edit-in-place --append-if-missing=rw
```
After reboot we get both `foo=bar rw`.

Fix https://github.com/ostreedev/ostree/issues/2679

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -3614,25 +3614,55 @@ ostree_sysroot_deployment_set_kargs_in_place (OstreeSysroot     *self,
   if (!_ostree_sysroot_ensure_writable (self, error))
     return FALSE;
 
-  g_assert (!ostree_deployment_is_staged (deployment));
+  // handle staged deployment
+  if (ostree_deployment_is_staged (deployment))
+    {
+      /* Read the staged state from disk */
+      glnx_autofd int fd = -1;
+      if (!glnx_openat_rdonly (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED, TRUE, &fd, error))
+        return FALSE;
 
-  OstreeBootconfigParser *new_bootconfig = ostree_deployment_get_bootconfig (deployment);
-  ostree_bootconfig_parser_set (new_bootconfig, "options", kargs_str);
-
-  g_autofree char *bootconf_name =
-    g_strdup_printf ("ostree-%d-%s.conf",
-                     self->deployments->len - ostree_deployment_get_index (deployment),
-                     ostree_deployment_get_osname (deployment));
-
-  g_autofree char *bootconfdir = g_strdup_printf ("loader.%d/entries", self->bootversion);
-  glnx_autofd int bootconf_dfd = -1;
-  if (!glnx_opendirat (self->boot_fd, bootconfdir, TRUE, &bootconf_dfd, error))
-    return FALSE;
-
-  if (!ostree_bootconfig_parser_write_at (new_bootconfig,
-                                          bootconf_dfd, bootconf_name,
+      g_autoptr(GBytes) contents = ot_fd_readall_or_mmap (fd, 0, error);
+      if (!contents)
+        return FALSE;
+      g_autoptr(GVariant) staged_deployment_data =
+        g_variant_new_from_bytes ((GVariantType*)"a{sv}", contents, TRUE);
+      g_autoptr(GVariantDict) staged_deployment_dict =
+        g_variant_dict_new (staged_deployment_data);
+      
+      g_autoptr(OstreeKernelArgs) kargs = ostree_kernel_args_from_string (kargs_str);
+      g_auto(GStrv) kargs_strv = ostree_kernel_args_to_strv (kargs);
+      
+      g_variant_dict_insert (staged_deployment_dict, "kargs", "^a&s", kargs_strv);
+      g_autoptr(GVariant) new_staged_deployment_data = g_variant_dict_end (staged_deployment_dict);
+      
+      if (!glnx_file_replace_contents_at (fd, _OSTREE_SYSROOT_RUNSTATE_STAGED,
+                                          g_variant_get_data (new_staged_deployment_data), 
+                                          g_variant_get_size (new_staged_deployment_data),
+                                          GLNX_FILE_REPLACE_NODATASYNC,
                                           cancellable, error))
-    return FALSE;
+        return FALSE;
+    }
+  else
+    {
+      OstreeBootconfigParser *new_bootconfig = ostree_deployment_get_bootconfig (deployment);
+      ostree_bootconfig_parser_set (new_bootconfig, "options", kargs_str);
+
+      g_autofree char *bootconf_name =
+        g_strdup_printf ("ostree-%d-%s.conf",
+                        self->deployments->len - ostree_deployment_get_index (deployment),
+                        ostree_deployment_get_osname (deployment));
+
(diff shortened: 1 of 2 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3614,25 +3614,45 @@ ostree_sysroot_deployment_set_kargs_in_place (OstreeSysroot     *self,
...
+      g_autoptr(OstreeKernelArgs) kargs = ostree_kernel_args_from_string (kargs_str);
+      g_auto(GStrv) kargs_strv = ostree_kernel_args_to_strv (kargs);
+      
+      OstreeSysrootDeployTreeOpts opts = {
+        .override_kernel_argv = kargs_strv,
+      };
+      g_autoptr(OstreeDeployment) new_deployment = NULL;
+      const char *osname = ostree_deployment_get_osname (self->staged_deployment);
+      g_autoptr(OstreeDeployment) merge_deployment = ostree_sysroot_get_merge_deployment (self, osname);
+      if (!ostree_sysroot_stage_tree_with_options(self, osname, ostree_deployment_get_csum (self->staged_deployment),
```
jlebon, line 3631: This will replace the staged deployment instead of editing it in-place. So we'd possibly lose other attributes of the staged deployment in the process.

To elaborate a bit more on https://github.com/ostreedev/ostree/issues/2679#issuecomment-1194483321, we need to do part of what `_ostree_sysroot_reload_staged` does ourselves, which is read the contents of `_OSTREE_SYSROOT_RUNSTATE_STAGED` into a `GVariantDict`. Then we can modify the `kargs` member as necessary, and then rewrite `_OSTREE_SYSROOT_RUNSTATE_STAGED` with our modifications.

jlebon:
> ```
> $ sudo rpm-ostree kargs --append=foo=bar
> $ sudo ostree admin kargs edit-in-place --append-if-missing=rw
> ```
>
> After reboot we get both `foo=bar rw`.

This works because `ot_admin_kargs_builtin_edit_in_place` edits the kargs from the staged deployment's bootconfig. So kargs will be correctly updated, but other fields in the serialized staged deployment will be lost. Here's an example test:

```
[root@cosa-devsh ~]# rpm-ostree initramfs-etc --track /etc/hosts
Checking out tree 49035e1... done
Writing OSTree commit... done
Staging deployment... done
Run "systemctl reboot" to start a reboot
[root@cosa-devsh ~]# ostree show --print-variant-type='a{sv}' /run/ostree/staged-deployment
{
 'target': <{'name': <'09a04c700fcff1fc712c0df8f538dd5cb72d1bb8b6961b86e367e74a27c713a7.0'>, 'osname': <'fedora-coreos'>, 'bootcsum': <'9ae2b94922656a13d3b3e3c8674d4a2eebda84743d2bd2fc1230b4def95a99ed'>}>,
 'merge-deployment': <{'name': <'49035e1db9c4f3fc459f3e745c391b33a77062cd0f99314045fa1dcbd9979400.0'>, 'osname': <'fedora-coreos'>, 'bootcsum': <'9ae2b94922656a13d3b3e3c8674d4a2eebda84743d2bd2fc1230b4def95a99ed'>}>,
 'kargs': <['mitigations=auto,nosmt', 'ignition.platform.id=qemu', 'console=tty0', 'console=ttyS0,115200n8', '$ignition_firstboot', 'ostree=/ostree/boot.1/fedora-coreos/9ae2b94922656a13d3b3e3c8674d4a2eebda84743d2bd2fc1230b4def95a99ed/0', 'root=UUID=73fcba18-c2ba-44aa-bba7-bcf2d4de1e9b', 'rw', 'rootflags=prjquota', 'boot=UUID=628b7037-2139-44ba-947b-1c134c880197']>,
 'overlay-initrds': <['47e711619716843b36b2ca3cdc0fc3e6459cb6e2f1e7a67e8a7736a4ebae5ff4']>
}
```

Notice the `overlay-initrds` field.

Now running `edit-in-place`:

```
[root@cosa-devsh ~]# ostree admin kargs edit-in-place --append foobar
error: Unknown option --append
[root@cosa-devsh ~]# ostree admin kargs edit-in-place --append-if-missing foobar
[root@cosa-devsh ~]# ostree show --print-variant-type='a{sv}' /run/ostree/staged-deployment
{
 'target': <{'name': <'09a04c700fcff1fc712c0df8f538dd5cb72d1bb8b6961b86e367e74a27c713a7.1'>, 'osname': <'fedora-coreos'>, 'bootcsum': <'9ae2b94922656a13d3b3e3c8674d4a2eebda84743d2bd2fc1230b4def95a99ed'>}>,
 'merge-deployment': <{'name': <'49035e1db9c4f3fc459f3e745c391b33a77062cd0f99314045fa1dcbd9979400.0'>, 'osname': <'fedora-coreos'>, 'bootcsum': <'9ae2b94922656a13d3b3e3c8674d4a2eebda84743d2bd2fc1230b4def95a99ed'>}>,
 'kargs': <['mitigations=auto,nosmt', 'ignition.platform.id=qemu', 'console=tty0', 'console=ttyS0,115200n8', '$ignition_firstboot', 'ostree=/ostree/boot.1/fedora-coreos/9ae2b94922656a13d3b3e3c8674d4a2eebda84743d2bd2fc1230b4def95a99ed/0', 'root=UUID=73fcba18-c2ba-44aa-bba7-bcf2d4de1e9b', 'rw', 'rootflags=prjquota', 'boot=UUID=628b7037-2139-44ba-947b-1c134c880197', 'foobar']>
}
```

Notice `foobar` was added but the `overlay-initrds` field is gone.

src/libostree/ostree-sysroot-deploy.c
HuijingHei: Thanks @jlebon!
So I should write another function like `_ostree_sysroot_reload_staged`, only modify `kargs`, to rewrite 
 `_OSTREE_SYSROOT_RUNSTATE_STAGED` like https://github.com/ostreedev/ostree/blob/main/src/libostree/ostree-sysroot-deploy.c#L3355?

jlebon: Yes, exactly! You can see if there's something worth factoring out of those functions so we can use them here too. They might be too thin to be worth it.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3614,25 +3614,58 @@ ostree_sysroot_deployment_set_kargs_in_place (OstreeSysroot     *self,
...
+    {
+      /* Read the staged state from disk */
+      glnx_autofd int fd = -1;
+      if (!ot_openat_ignore_enoent (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED, &fd, error))
+        return FALSE;
 
-  OstreeBootconfigParser *new_bootconfig = ostree_deployment_get_bootconfig (deployment);
-  ostree_bootconfig_parser_set (new_bootconfig, "options", kargs_str);
+      if (fd == -1)
+        return FALSE;
```
cgwalters, line 3626: I think this case can't (shouldn't) happen, because we can only have a staged deployment if the file exists.

So I'd change this to use `glnx_openat` which does *not* ignore not found errors.

cgwalters (approved):
One minor comment, otherwise this is looking good to me.  Really great work diving into this and figuring it all out!

jlebon (approved):
LGTM too! Nice work!

PR #2698, opened 2022-08-22
rust: Update to latest git

cgwalters:
In the future I may try to add CI that requires this to be
sync'd...


jlebon (approved):
Strong agree on having a CI check for this!

PR #2699, opened 2022-08-22
ci: Also drop seccomp on debian testing

cgwalters:
I didn't deep dive into debugging this but I'm pretty sure it's
 https://gitlab.gnome.org/GNOME/glib/-/issues/2580
which is us having an older Docker in the middle here.

---

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -126,6 +126,7 @@ jobs:
 
           - name: Debian Testing
             image: debian:testing-slim
+            container-options: --security-opt seccomp=unconfined
             pre-checkout-setup: |
               apt-get update
               apt-get install -y git


dbnicholson (changes requested):
This seems like almost the same issue I saw on one of the Ubuntu builds awhile ago. Could you try running it with `container-options: --security-opt seccomp=unconfined`? See the `ubuntu:rolling` build below.

cgwalters:
> This seems like almost the same issue I saw on one of the Ubuntu builds awhile ago. Could you try running it with container-options: --security-opt seccomp=unconfined? See the ubuntu:rolling build below.

Likely, thanks!

dbnicholson (approved):

lucab:
It looks like Prow got stuck and it also fails to re-trigger, overriding.

/override ci/prow/sanity

PR #2704, opened 2022-08-30
finalize-staged: Don't listen to `SIGTERM`, just let kernel exit us

cgwalters:
Followup from discussion in
https://github.com/ostreedev/ostree/pull/2544#discussion_r958840936

This is more efficient; no need to have the kernel context switch
us in at shutdown time just so we can turn around and call
`exit()`.

### DIFF
--- a/src/ostree/ot-admin-builtin-finalize-staged.c
+++ b/src/ostree/ot-admin-builtin-finalize-staged.c
@@ -41,16 +41,6 @@ static GOptionEntry options[] = {
   { NULL }
 };
 
-static gboolean
-sigterm_cb (gpointer user_data)
-{
-  gboolean *running = user_data;
-  g_print ("Received SIGTERM, exiting\n");
-  *running = FALSE;
-  g_main_context_wakeup (NULL);
-  return G_SOURCE_REMOVE;
-}
-
 /* Called by ostree-finalize-staged.service, and in turn
  * invokes a cmdprivate function inside the shared library.
  */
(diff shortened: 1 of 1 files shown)


src/ostree/ot-admin-builtin-finalize-staged.c
```diff
@@ -94,13 +84,11 @@ ot_admin_builtin_finalize_staged (int argc, char **argv, OstreeCommandInvocation
...
+       * system shutdown, so block until we get SIGTERM which systemd will send
+       * when the unit is stopped.  The below is basically the equivalent of
+       * `sleep infinity` because nothing will wake the event loop up.
        */
-      gboolean running = TRUE;
-      g_unix_signal_add (SIGTERM, sigterm_cb, &running);
-      g_print ("Waiting for SIGTERM\n");
-      while (running)
+      while (TRUE)
         g_main_context_iteration (NULL, TRUE);
```
dbnicholson, line 92: If the loop isn't doing anything, why not just use `pause()`? That's actually exactly what `sleep infinity` does. Seems silly to have glib setup an event loop that does nothing.

cgwalters: Well, we're just optimizing a no-op at this point, but yes, done :smile: 
I didn't know about the `pause` system call actually.  It looks like it may be x86 only; glibc has a fallback that just uses `ppoll()` https://github.com/bminor/glibc/blob/02ca25fef2785974011e9c5beecc99b900b69fd7/sysdeps/unix/sysv/linux/pause.c#L31 with `NULL`s.

dbnicholson (approved):

src/ostree/ot-admin-builtin-finalize-staged.c
dbnicholson: I ran `strace` on `sleep infinity` and that's where it blocked. I think it is provided most everywhere on linux since it's in POSIX. `pause` is defined in most syscall tables in the kernel and there's a [generic implementation](https://github.com/torvalds/linux/blob/master/kernel/signal.c#L4602).

PR #2706, opened 2022-09-03
ostree-fetcher-curl: check for HTTP2 support before trying to use it

git-bruh:
https://github.com/flatpak/flatpak/pull/5074


lucab:
Thanks for the patch, it looks good.
Could you perhaps share more details about which OS/version is shipping a recent `libcurl` without HTTP2 support?

git-bruh:
@lucab I don't think any distro does that by default - I'm sort of rolling my own tiny/esoteric distro so I didn't have http2 enabled as I just wasn't aware of that libcurl option. Downloading seems to work fine after applying this patch.

lucab (approved):

PR #2708, opened 2022-09-05
s390x: drop shell code for `initrd` modification in SE case

nikita-dubrovskii:


src/libostree/ostree-bootloader-zipl.c
```diff
@@ -217,25 +221,38 @@ _ostree_secure_execution_generate_initrd (const gchar *initrd,
...
-  int status = 0;
-  if (!g_spawn_sync (NULL, (char**)argv, NULL, G_SPAWN_SEARCH_PATH,
-                     NULL, NULL, &out, &err, &status, error))
-    return glnx_prefix_error(error, "s390x SE: spawning %s", SECURE_EXECUTION_RAMDISK_TOOL);
-
-  if (!g_spawn_check_exit_status (status, error))
+  #ifdef HAVE_LIBARCHIVE
+  // creating new cpio gzip archive with LUKS keys
+  g_autoptr(OtAutoArchiveWrite) a = archive_write_new ();
+  g_assert_cmpint (0, ==, archive_write_set_format_cpio_newc (a));
```
lucab, line 227: I'd suggest turning this assertion (and all the ones below) into proper handled errors.

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -217,25 +221,38 @@ _ostree_secure_execution_generate_initrd (const gchar *initrd,
...
+            return glnx_prefix_error (error, "s390x SE: reading %s", path);
+          gsize size = 0;
+          const char *ptr = (const char *) g_bytes_get_data (data, &size);
+          g_assert_cmpint (size, ==, archive_write_data (a, ptr, size));
+        }
+      archive_entry_free (ae);
     }
-
+  #endif
   ot_journal_print(LOG_INFO, "s390x SE: luks keys added to initrd");
```
lucab, line 256: I guess this one should belong to the logic above too?

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -195,30 +199,61 @@ static gboolean
...
+  if (!glnx_open_anonymous_tmpfile (O_RDWR | O_CLOEXEC, out_initrd, error))
+    return glnx_prefix_error (error, "s390x SE: opening new ramdisk");
+  {
+    glnx_autofd int fd = -1;
+    glnx_openat_rdonly (AT_FDCWD, initrd, TRUE, &fd, error);
+    if (glnx_regfile_copy_bytes (fd, out_initrd->fd, (off_t) -1) < 0)
+      return glnx_throw_errno_prefix (error, "s390x SE: copying ramdisk");
+  }
+  #ifdef HAVE_LIBARCHIVE
+  // creating new cpio gzip archive with LUKS keys
```
lucab, line 225: Let's split this to its own function maybe?

src/libostree/ostree-bootloader-zipl.c
nikita-dubrovskii: agree, will do

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -195,73 +199,117 @@ static gboolean
...
+          if (!glnx_openat_rdonly (AT_FDCWD, path, TRUE, &fd, error))
+            return glnx_prefix_error (error, "s390x SE: opening %s", path);
+          g_autoptr(GBytes) data = glnx_fd_readall_bytes (fd, cancellable, error);
+          if (!data)
+            return glnx_prefix_error (error, "s390x SE: reading %s", path);
+
+          gsize size = 0;
+          const char *ptr = (const char *) g_bytes_get_data (data, &size);
+          if (archive_write_data (a, ptr, size) != size)
+            return glnx_prefix_error (error, "s390x SE: writting cpio entry: %s", archive_error_string (a));
```
lucab, line 246: Uhm, I think `archive_error_string(a)` only works for the case where we get a `-1`. I don't think we can use that in the case of short writes.
Overall, I think the two error cases should be handled each one separately.

lucab (commented):
Some minor comments on the new functions, looks good otherwise.

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -195,73 +199,117 @@ static gboolean
...
+        }
+      archive_entry_free (ae);
```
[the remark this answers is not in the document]
cgwalters: Yeah, unfortunately we don't have autoptr set up for this elsewhere either.

cgwalters:
Overall looks sane to me.  One thing I'd say though is we could try having ostree learn optional dependencies on https://github.com/ostreedev/ostree-rs-ext/ - for example in this case, all of this code could have been written in Rust, and then we just teach ostree to call out to e.g. `/usr/libexec/libostree/internals/s390x-se-initramfs` (which would just be a symlink to `/usr/bin/rpm-ostree`, which for us contains the ostree-rs-ext code)

This approach gets more interesting of course if we ever decide to take a *hard* dependency of the ostree-system-root stuff on the Rust code.  I'd definitely like to consider that.

nikita-dubrovskii:
> Overall looks sane to me. One thing I'd say though is we could try having ostree learn optional dependencies on https://github.com/ostreedev/ostree-rs-ext/ - for example in this case, all of this code could have been written in Rust, and then we just teach ostree to call out to e.g. `/usr/libexec/libostree/internals/s390x-se-initramfs` (which would just be a symlink to `/usr/bin/rpm-ostree`, which for us contains the ostree-rs-ext code)
> 
> This approach gets more interesting of course if we ever decide to take a _hard_ dependency of the ostree-system-root stuff on the Rust code. I'd definitely like to consider that.

it's already implemented in Rust )) `rdcore` knows how-to `zipl` and update `sdboot`

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -195,73 +199,123 @@ static gboolean
...
+          g_autoptr(GBytes) data = glnx_fd_readall_bytes (fd, cancellable, error);
+          if (!data)
+            return glnx_prefix_error (error, "s390x SE: reading %s", path);
+
+          gsize size = 0;
+          const char *ptr = (const char *) g_bytes_get_data (data, &size);
+          ssize_t written = archive_write_data (a, ptr, size);
+          if (written == -1)
+            return glnx_prefix_error (error, "s390x SE: writing cpio entry: %s", archive_error_string (a));
+          if (written == !size)
```
lucab, line 252: I think this is a typo for `!=`.

nikita-dubrovskii: Thx! good catch

lucab:
Code looks good to me now. I'll leave some space for comments to @cgwalters for further follow-ups on the ostree-ext discussion.

cgwalters:
> it's already implemented in Rust )) rdcore knows how-to zipl and update sdboot

Heh right so...we could make a small shared crate for that or something like that.  Anyways, this is a discussion for future changes, should not block this one.

cgwalters (approved):

PR #2709, opened 2022-09-06
otutil: add error handling to variant builders

lucab:
This enhances a bunch of helpers related to GVariant building, in
order to properly handle errors and avoid some potential cases of
unexpected NULL results.

### DIFF
--- a/src/libotutil/ot-variant-builder.c
+++ b/src/libotutil/ot-variant-builder.c
@@ -758,11 +758,9 @@ struct _OtVariantBuilder {
 static OtVariantBuilderInfo *
 ot_variant_builder_info_new (OtVariantBuilder *builder, const GVariantType *type)
 {
-  OtVariantBuilderInfo *info;
+  g_assert (g_variant_type_is_container (type));
 
-  g_return_val_if_fail (g_variant_type_is_container (type), NULL);
-
-  info = (OtVariantBuilderInfo *) g_slice_new0 (OtVariantBuilderInfo);
+  OtVariantBuilderInfo *info = (OtVariantBuilderInfo *) g_slice_new0 (OtVariantBuilderInfo);
 
   info->builder = builder;
   info->type = g_variant_type_copy (type);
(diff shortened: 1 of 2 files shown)


cgwalters:
AFAICS all inputs to that function come from compile-time values (i.e. never external/user input).  What's the benefit in making them runtime errors versus assertion failures?

cgwalters (approved):
That said, I don't oppose this either; I don't think it makes a big difference either way, so marking approved.

src/libotutil/ot-variant-builder.c
```diff
@@ -841,15 +847,16 @@ ot_variant_builder_info_free (OtVariantBuilderInfo *info)
...
+                        GError **error)
 {
   OtVariantBuilder *builder;
 
   g_return_val_if_fail (g_variant_type_is_container (type), NULL);
 
   builder = (OtVariantBuilder *) g_slice_new0 (OtVariantBuilder);
 
-  builder->head = ot_variant_builder_info_new (builder, type);
+  builder->head = ot_variant_builder_info_new (builder, type, error);
```
lucab, line 859: Ugh, I actually forgot to error-check here, as well as dropping the `g_return_val_if_fail` above.

lucab:
@cgwalters yes indeed, all the current usages are statically determined.
Assertions work too, all good as long as we drop the `g_return_val_if_fail` footgun. I wasn't sure which way to go, and this seemed the most future-proof.
I just realized I forgot one tiny bit and will need to respin this anyway; let me know if you'd prefer this PR based on assertions instead.

cgwalters:
If it can't come from user input then I think we should use assertions as a general rule - you agree with that right?  But honestly if it's easier for you in this micro-instance to just fix up what you've already done instead of redoing it, that seems totally fine too.

lucab:
Agreed, I think we are on the same page. What I can't generally foresee is how an helper can be used in the future, but here I think it's fair to assume that the signature type is unlikely to come from the outside. Will tweak the PR accordingly!

cgwalters (approved):

PR #2710, opened 2022-09-07
lib/mtree: drop redundant name checks

lucab:
This drops several NULL checks against filename input argument.
Those checks are both redundant (as filename validation already checks for that) and dangerous (as they return early without setting an error value).


lucab:
For reference, the NULL check is already performed here:
https://github.com/ostreedev/ostree/blob/73eda7966226718aa4ae6637fd43cd81cdbd5fdd/src/libotutil/ot-unix-utils.c#L40-L45

cgwalters (approved):

Issue #2712, opened 2022-09-12
Support character devices (whiteouts) in ostree commits


mangelajo:
This is an enhancement request to have ostree support the commit storage
of whiteout 0:0 character devices.

# Motivation

Embedding container storage into ostree images allows for offline, and fast container
startup (no need to contact a registry) for ostree based systems where the containers
can be stored in the native container storage format within the ostree image.

overlayfs in the linux kernel implements the deletion of a file from one layer to the
next by creating a char 0:0 file (whiteout) in the new layer.

But today, ostree does not support the storage of character devices: 

```bash
rm -rf rootfs test
ostree --repo=./test init --mode=bare-user
mkdir rootfs
mknod -m 0 rootfs/whiteout c 0 0
ostree commit --repo=./test ./rootfs --branch test/stable/x86_64
````
results:
```
error: Not a regular file or symlink: whiteout
```

This limits our ability to encode container storage into ostree images. 

# References

This necessity has been brought up in the past, and from other projects that build on os

- https://github.com/ostreedev/ostree/issues/2568
- https://github.com/ostreedev/ostree/issues/2310
- https://github.com/ostreedev/ostree-rs-ext/issues/246
- https://issues.redhat.com/browse/COMPOSER-1745
 
# Alternatives

1.  Compacting images into a single layer

This would remove the introduction of whiteout files and would not increase the storage size
since ostree is a content-addressed storage system, but would modify the sha256 hash
of images that sometimes are used as references.

i.e. `quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ba44dead03ea74107f90d58525106fb52d27a120b73c6cc8e2be31d37043ca1c`

2. Avoiding deletion between layers
 
This is technically non feasible in some cases (next layer or final containers expects a file
not to exist), and hard to manage over time, it also precludes including off-shelf existing
images where the builder of the ostree has no control over the image building.

# Possible issues

This will probably not be usable when a new ostree layers is being built from within a container, since you can't
create a whiteout inside an overlayfs mount (assumption here).

cgwalters:
> This would remove the introduction of whiteout files but would increase the storage size
exponentially (depending on the number of shared layers) for systems where many
images share common layers).

No it wouldn't, because ostree is a content-addressed storage system.

mangelajo:
> > This would remove the introduction of whiteout files but would increase the storage size
> > exponentially (depending on the number of shared layers) for systems where many
> > images share common layers).
> 
[...]

PR #2713, opened 2022-09-13
WIP/DNM / Support overlayfs whiteout character files

mangelajo:
This is yet incomplete and untested, it doesn't work yet.

Related-Issue: #2712

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -1732,6 +1732,7 @@ _ostree_stbuf_to_gfileinfo (const struct stat *stbuf)
   g_file_info_set_attribute_uint32 (ret, "unix::uid", stbuf->st_uid);
   g_file_info_set_attribute_uint32 (ret, "unix::gid", stbuf->st_gid);
   g_file_info_set_attribute_uint32 (ret, "unix::mode", mode);
+  g_file_info_set_attribute_uint32 (ret, "unix::rdev", stbuf->st_rdev);
 
   /* those aren't stored by ostree, but used by the devino cache */
   g_file_info_set_attribute_uint32 (ret, "unix::device", stbuf->st_dev);
@@ -1814,6 +1815,22 @@ _ostree_stbuf_equal (struct stat *stbuf_a, struct stat *stbuf_b)
   if (stbuf_a->st_gid != stbuf_b->st_gid)
     return FALSE;
   return TRUE;
+}
+
+/* Check if stbuf belongs to a whiteout character device */
+gboolean
+_ostree_stbuf_is_whiteout(struct stat *stbuf)
+{
+  return S_ISCHR(stbuf->st_mode) && stbuf->st_rdev == 0;
+}
+
+/* Check if GFileInfo belongs to a whiteout character device */
+gboolean
+_ostree_gfileinfo_is_whiteout(GFileInfo *file_info)
+{
+  struct stat stbuf;
+  _ostree_gfileinfo_to_stbuf(file_info, &stbuf);
+  return _ostree_stbuf_is_whiteout(&stbuf);
 }
 
 /* Many parts of libostree only care about mode,uid,gid - this creates
@@ -2014,7 +2031,7 @@ file_header_parse (GVariant         *metadata,
   mode = GUINT32_FROM_BE (mode);
   g_autoptr(GFileInfo) ret_file_info = _ostree_mode_uidgid_to_gfileinfo (mode, uid, gid);
 
-  if (S_ISREG (mode) || S_ISCHR(mode))
+  if (S_ISREG (mode) || _ostree_gfileinfo_is_whiteout(ret_file_info))
     {
       ;
     }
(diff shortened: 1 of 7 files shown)


lucab:
/hold

I'd like to push back on this, I don't think it is a good idea. The thread at https://github.com/ostreedev/ostree/issues/2712 is still ongoing, let's find better alternatives there.

mangelajo:
> /hold
> 
> I'd like to push back on this, I don't think it is a good idea. The thread at #2712 is still ongoing, let's find better alternatives there.

please participate on the thread and provide alternatives

lucab:
I'll let @cgwalters drive the conversation there to avoid too many voices, but I pretty much share the hints in https://github.com/ostreedev/ostree/issues/2712#issuecomment-1245928742.

src/libostree/ostree-core.c
```diff
@@ -2014,7 +2014,7 @@ file_header_parse (GVariant         *metadata,
   mode = GUINT32_FROM_BE (mode);
   g_autoptr(GFileInfo) ret_file_info = _ostree_mode_uidgid_to_gfileinfo (mode, uid, gid);
 
-  if (S_ISREG (mode))
+  if (S_ISREG (mode) || S_ISCHR(mode))
```
cgwalters, line 2017: I think we should have a common helper that checks specifically for whiteouts, which needs to take either a `GFileInfo` or a `struct stat`.

We still do want to error out for any character devices that are not whiteouts.

src/libostree/ostree-repo.c
```diff
@@ -4474,7 +4475,7 @@ ostree_repo_load_file (OstreeRepo         *self,
           if (S_ISLNK (stbuf.st_mode))
             g_file_info_set_symlink_target (*out_file_info, symlink_target);
           else
-            g_assert (S_ISREG (stbuf.st_mode));
+            g_assert (S_ISREG (stbuf.st_mode) || (S_ISCHR(stbuf.st_mode) && stbuf.st_rdev == 0));
```
cgwalters, line 4478: This is the check we want to generalize everywhere I think, as e.g. `_ostree_is_whiteout (&stbuf)`.

cgwalters:
I do want to note here the last bit of https://github.com/ostreedev/ostree/issues/2712#issuecomment-1245656497 

> There's a bunch of code that will need touching for this, not just in ostree core but also in e.g. https://github.com/ostreedev/ostree-rs-ext/tree/main/lib/src/tar

which is that we could land this in ostree and that would probably work for your use case, but all that code that lives in ostree-rs-ext, such as e.g. `ostree container encapsulate` would fail.

And that gets into the whole topic of being able to "nest" container storage layers in container layers, which I think we do want longer term, which leads to https://github.com/ostreedev/ostree/issues/2712#issuecomment-1249469298 or so.

src/libostree/ostree-core.c
```diff
@@ -2368,7 +2385,7 @@ _ostree_validate_bareuseronly_mode (guint32     content_mode,
         return glnx_throw (error, "Content object %s: invalid mode 0%04o with bits 0%04o",
                            checksum, content_mode, invalid_modebits);
     }
-  else if (S_ISLNK (content_mode))
+  else if (S_ISLNK (content_mode) || S_ISCHR(content_mode))
```
mangelajo, line 2388: I couldn't use the helper functions here since at this point it's only mode

src/libostree/ostree-core.c
```diff
@@ -2400,7 +2417,7 @@ gboolean
 ostree_validate_structureof_file_mode (guint32            mode,
                                        GError           **error)
 {
-  if (!(S_ISREG (mode) || S_ISLNK (mode)))
+  if (!(S_ISREG (mode) || S_ISLNK (mode) || S_ISCHR(mode)))
```
mangelajo, line 2420: same here.

src/libostree/ostree-repo-commit.c
```diff
@@ -301,7 +332,7 @@ commit_loose_regfile_object (OstreeRepo        *self,
             return FALSE;
         }
       else
-        g_assert (S_ISLNK (mode));
+        g_assert (S_ISLNK (mode) || S_ISCHR(mode));
```
mangelajo, line 335: and here.

mangelajo:
I'm working on another PR that follows the other suggested approach.

src/libostree/ostree-repo-checkout.c
```diff
@@ -385,9 +385,9 @@ create_file_copy_from_input_at (OstreeRepo     *repo,
                 if (g_str_equal (checksum, actual_checksum))
                   return TRUE;
 
-                /* Otherwise, fall through and do the link, we should
-                 * get EEXIST.
-                 */
+                  /* Otherwise, fall through and do the link, we should
```
mangelajo, line 388: This shouldn't be here

mangelajo:
Dropping this approach in favor of https://github.com/ostreedev/ostree/pull/2717

PR #2717, opened 2022-09-19
Support overlayfs whiteouts on checkout

mangelajo:
Introduces an intermediate format for overlayfs storage, where
.wh-ostree. prefixed files will be converted into char 0:0
whiteout devices used by overlayfs to mark deletions across layers.

Related-Issue: #2712

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -35,6 +35,8 @@
 #define WHITEOUT_PREFIX ".wh."
 #define OPAQUE_WHITEOUT_NAME ".wh..wh..opq"
 
+#define OVERLAYFS_WHITEOUT_PREFIX ".ostree-wh."
+
 /* Per-checkout call state/caching */
 typedef struct {
   GString *path_buf; /* buffer for real path if filtering enabled */
@@ -583,6 +585,117 @@ checkout_file_hardlink (OstreeRepo                          *self,
 }
 
 static gboolean
+_checkout_overlayfs_whiteout_at_no_overwrite (OstreeRepoCheckoutAtOptions    *options,
+                                              int                             destination_dfd,
+                                              const char                     *destination_name,
+                                              GFileInfo                      *file_info,
+                                              GVariant                       *xattrs,
+                                              gboolean                       *found_exant_file,
+                                              GCancellable                   *cancellable,
+                                              GError                        **error)
+{
+  if (found_exant_file != NULL)
+    *found_exant_file = FALSE;
+  guint32 file_mode = g_file_info_get_attribute_uint32 (file_info, "unix::mode");
+  if (mknodat(destination_dfd, destination_name, (file_mode & ~S_IFMT) | S_IFCHR, (dev_t)0) < 0)
+    {
+      if (errno == EEXIST && found_exant_file != NULL)
+        {
+          *found_exant_file = TRUE;
+          return TRUE;
+        }
+      return glnx_throw_errno_prefix (error, "Creating whiteout char device");
+    }
+  if (options->mode != OSTREE_REPO_CHECKOUT_MODE_USER)
+    {
+      if (xattrs != NULL &&
+          !glnx_dfd_name_set_all_xattrs(destination_dfd, destination_name, xattrs,
+                                          cancellable, error))
+          return glnx_throw_errno_prefix (error, "Setting xattrs for whiteout char device");
+
+      if (TEMP_FAILURE_RETRY(fchownat(destination_dfd, destination_name,
+                                      g_file_info_get_attribute_uint32 (file_info, "unix::uid"),
+                                      g_file_info_get_attribute_uint32 (file_info, "unix::gid"),
+                                      AT_SYMLINK_NOFOLLOW) < 0))
+          return glnx_throw_errno_prefix (error, "fchownat");
+      if (TEMP_FAILURE_RETRY (fchmodat (destination_dfd, destination_name, file_mode & ~S_IFMT, 0)) < 0)
+          return glnx_throw_errno_prefix (error, "fchmodat %s to 0%o", destination_name, file_mode & ~S_IFMT);
+    }
+
+  return TRUE;
+}
+
+static gboolean
+_checkout_overlayfs_whiteout_at (OstreeRepo                     *repo,
+                                 OstreeRepoCheckoutAtOptions    *options,
+                                 int                             destination_dfd,
+                                 const char                     *destination_name,
+                                 GFileInfo                      *file_info,
+                                 GVariant                       *xattrs,
+                                 GCancellable                   *cancellable,
+                                 GError                        **error)
+{
+  gboolean found_exant_file = FALSE;
+  if (!_checkout_overlayfs_whiteout_at_no_overwrite(options, destination_dfd, destination_name,
+                                                    file_info, xattrs,&found_exant_file,
+                                                    cancellable, error))
+    return FALSE;
+
+   if (!found_exant_file)
+    return TRUE;
+
+  guint32 uid = g_file_info_get_attribute_uint32 (file_info, "unix::uid");
+  guint32 gid = g_file_info_get_attribute_uint32 (file_info, "unix::gid");
+  guint32 file_mode = g_file_info_get_attribute_uint32 (file_info, "unix::mode");
+
+  struct stat dest_stbuf;
+
+  switch(options->overwrite_mode)
+    {
+      case OSTREE_REPO_CHECKOUT_OVERWRITE_NONE:
+        return FALSE;
+      case OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_FILES:
+        if (!ot_ensure_unlinked_at (destination_dfd, destination_name, error))
+          return FALSE;
+        return _checkout_overlayfs_whiteout_at_no_overwrite(options, destination_dfd, destination_name,
+                                                   file_info, xattrs, NULL, cancellable, error);
+      case OSTREE_REPO_CHECKOUT_OVERWRITE_ADD_FILES:
+        return TRUE;
+
+      case OSTREE_REPO_CHECKOUT_OVERWRITE_UNION_IDENTICAL:
+        if (!glnx_fstatat(destination_dfd, destination_name, &dest_stbuf, AT_SYMLINK_NOFOLLOW,
+                          error))
+          return FALSE;
+        if (!(repo->disable_xattrs || repo->mode == OSTREE_REPO_MODE_BARE_USER_ONLY))
+          {
+            g_autoptr(GVariant) fs_xattrs;
+            if (!glnx_dfd_name_get_all_xattrs (destination_dfd, destination_name,
+                                               &fs_xattrs, cancellable, error))
+              return FALSE;
+            if (!g_variant_equal(fs_xattrs, xattrs))
+              return glnx_throw(error, "existing destination file %s xattrs don't match",
+                                destination_name);
+          }
+        if (options->mode != OSTREE_REPO_CHECKOUT_MODE_USER)
+          {
+            if (gid != dest_stbuf.st_gid)
+              return glnx_throw(error, "existing destination file %s does not match gid %d",
+                                destination_name, gid);
+
+            if (uid != dest_stbuf.st_uid)
+              return glnx_throw(error, "existing destination file %s does not match uid %d",
+                                destination_name, gid);
+
+            if ((file_mode & ALLPERMS) != (dest_stbuf.st_mode & ALLPERMS))
+              return glnx_throw(error, "existing destination file %s does not match mode %o",
+                                destination_name, file_mode);
+          }
+        break;
+    }
+    return TRUE;
+}
+
+static gboolean
 checkout_one_file_at (OstreeRepo                        *repo,
                       OstreeRepoCheckoutAtOptions       *options,
                       CheckoutState                     *state,
@@ -603,7 +716,8 @@ checkout_one_file_at (OstreeRepo                        *repo,
 
   /* FIXME - avoid the GFileInfo here */
   g_autoptr(GFileInfo) source_info = NULL;
-  if (!ostree_repo_load_file (repo, checksum, NULL, &source_info, NULL,
+  g_autoptr(GVariant) source_xattrs = NULL;
+  if (!ostree_repo_load_file (repo, checksum, NULL, &source_info, &source_xattrs,
                               cancellable, error))
     return FALSE;
 
@@ -623,6 +737,7 @@ checkout_one_file_at (OstreeRepo                        *repo,
   const gboolean is_unreadable = (!is_symlink && (source_mode & S_IRUSR) == 0);
   const gboolean is_whiteout = (!is_symlink && options->process_whiteouts &&
                                 g_str_has_prefix (destination_name, WHITEOUT_PREFIX));
+  const gboolean is_overlayfs_whiteout = (!is_symlink && g_str_has_prefix (destination_name, OVERLAYFS_WHITEOUT_PREFIX));
   const gboolean is_reg_zerosized = (!is_symlink && g_file_info_get_size (source_info) == 0);
   const gboolean override_user_unreadable = (options->mode == OSTREE_REPO_CHECKOUT_MODE_USER && is_unreadable);
 
@@ -642,6 +757,18 @@ checkout_one_file_at (OstreeRepo                        *repo,
         return FALSE;
 
       need_copy = FALSE;
+    }
+  else if (is_overlayfs_whiteout && options->process_passthrough_whiteouts)
+    {
+      const char *name = destination_name + (sizeof (OVERLAYFS_WHITEOUT_PREFIX) - 1);
+
+      if (!name[0])
+        return glnx_throw (error, "Invalid empty overlayfs whiteout '%s'", name);
+
+      g_assert (name[0] != '/'); /* Sanity */
+
+      return _checkout_overlayfs_whiteout_at(repo, options, destination_dfd, name,
+                                             source_info, source_xattrs, cancellable, error);
     }
   else if (is_reg_zerosized || override_user_unreadable)
     {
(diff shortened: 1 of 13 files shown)


src/libostree/ostree-repo-checkout.c
```diff
@@ -643,6 +647,32 @@ checkout_one_file_at (OstreeRepo                        *repo,
 
       need_copy = FALSE;
     }
+  else if (is_overlayfs_whiteout)
+    {
+      const char *name = destination_name + (sizeof (OVERLAYFS_WHITEOUT_PREFIX) - 1);
```
mangelajo, line 652: ok, next update should refactor this out to a function where we handle selinux policy handling into the xattrs, and overwrite mode handling.

src/libostree/ostree-repo-checkout.c
```diff
@@ -35,6 +35,8 @@
 #define WHITEOUT_PREFIX ".wh."
 #define OPAQUE_WHITEOUT_NAME ".wh..wh..opq"
 
+#define OVERLAYFS_WHITEOUT_PREFIX ".wh-ostree."
```
cgwalters, line 38: How about
```suggestion
#define OSTREE_TARGET_WHITEOUT_PREFIX ".ostree-wh."
```

Here "ostree_" is first because it's a thing conceptually namespaced/scoped to ostree, not to overlayfs.

src/libostree/ostree-repo-checkout.c
```diff
@@ -643,6 +647,32 @@ checkout_one_file_at (OstreeRepo                        *repo,
...
+                                              cancellable, error))
+              return glnx_throw_errno_prefix (error, "Setting xattrs for whiteout char device");
+
+          if (TEMP_FAILURE_RETRY(fchownat(destination_dfd, name,
+                                          g_file_info_get_attribute_uint32 (source_info, "unix::uid"),
+                                          g_file_info_get_attribute_uint32 (source_info, "unix::gid"),
+                                          AT_SYMLINK_NOFOLLOW) < 0))
+              return glnx_throw_errno_prefix (error, "fchownat");
+        }
+      need_copy = FALSE;
```
cgwalters, line 674: We could probably just `return TRUE;` here too - what we're skipping over is the caching stuff which isn't relevant for this.

src/libostree/ostree-repo-checkout.c
```diff
@@ -643,6 +647,32 @@ checkout_one_file_at (OstreeRepo                        *repo,
...
+  else if (is_overlayfs_whiteout)
+    {
+      const char *name = destination_name + (sizeof (OVERLAYFS_WHITEOUT_PREFIX) - 1);
+
+      if (!name[0])
+        return glnx_throw (error, "Invalid empty overlayfs whiteout '%s'", name);
+
+      g_assert (name[0] != '/'); /* Sanity */
+
+      if (mknodat(destination_dfd, name, (source_mode & ~S_IFMT) | S_IFCHR, (dev_t)0) < 0)
```
cgwalters, line 659: One thing to verify here is whether we also need to do a `fchmodat` - I think the mode we set is modified by `umask`, so in most cases we always unconditionally chmod.

src/libostree/ostree-repo-checkout.c
```diff
@@ -643,6 +647,32 @@ checkout_one_file_at (OstreeRepo                        *repo,
 
       need_copy = FALSE;
     }
+  else if (is_overlayfs_whiteout)
```
cgwalters, line 650: I'm thinking we may want to make this behavior potentially opt-in?  So far ostree has no "magical filenames" - this would be the first.

A concern I have here is that e.g. flatpak would suddenly gain exposure to this.  And as we discussed it's *probably* fine...the worst I can think of is that someone using flatpak *on top of overlayfs* wouldn't be able to install an app that included a file with this name.

But still, by requiring an opt-in flag we're at least reducing the blast radius.

I think we can then hange the deploy path to pass this flag by default.  This feature would clearly be useful for everyone using ostree-for-bootable-systems.

Though I think I'd also be OK having it opt-out (i.e. we add a flag to disable this) instead; then if we discover other use cases don't want it they could turn it off.  But opt-in seems safer and only requires a little bit more work.

cgwalters (commented):
I think we should also add `ostree commit --retarget-whiteouts` or so which would automatically do the inverse side.  Now AIUI today osbuild uses `rpm-ostree compose commit` which could also use a similar API.

But neither of these are blockers.

cgwalters:
Have you been able to prove out that this approach seems viable via a post-processing script that renames the whiteouts in this way?  I'm OK landing a minimal patch that only does the checkout side.

src/libostree/ostree-repo-checkout.c
mangelajo: good point I will verify this

src/libostree/ostree-repo-checkout.c
mangelajo: I agree that opt-in as you said,may require changes in the deployment path when we install or upgrade the system. Let me talk to the RHEL4Edge team and try figure out how easy/hard it's to include the flag, probably it's easy or even under the control of the image creator so not a big deal, but I'm not familiar with that.

By looking at our composer installer image I see:
```
ostreesetup --nogpg --osname=rhel --remote=edge --url=file:///run/install/repo/ostree/repo --ref=rhel/8/x86_64/edge
```

Do you know where ostreesetup lives?


Note: It seems like any regular user with no special privileges can create char 0:0 devices , and that device always returns error.

As you said the issue would be if trying to checkout inside an overlayfs system.

mangelajo:
> I'm OK landing a minimal patch that only does the checkout side.

Thanks Colin

src/libostree/ostree-repo-checkout.c
mangelajo: Done

src/libostree/ostree-repo-checkout.c
mangelajo: Right, looking at mknodat documentation:
```
The file mode is modified by the process's umask in the usual way: in the absence of a default ACL, the permissions of the created node are (mode & ~umask).
```

So we also need fchmodat

src/libostree/ostree-repo-checkout.c
cgwalters: In my proposal here, the `ostree admin deploy` code would by default pass this option down when it does the checkout.  This means no modifications would be necessary for anything higher level than this project.

Conceptually ostree has two separate parts; there's a git-like core represented by `OstreeRepo`, and the sysroot/deploy bits in `OstreeSysroot`.  The split isn't 100% clean (the magical `ostree/1/0/1` type refs are part of that) but that's the idea.

Hmm actually looking at this we of course already have

```
typedef struct {
...
  gboolean process_whiteouts;
...
} OstreeRepoCheckoutAtOptions;
```

And...confusingly that's for a different model of whiteouts where we want ostree to handle them, not the container engine.  Anyways so...basically I'm saying we'd add `gboolean process_passthrough_whiteouts;` there, and then in `checkout_deployment_tree` set that boolean or so.  WDYT?

mangelajo: I added an opt-out for now, but happy to change into opt-in if we see it's doable in rhel4edge or for any other reason.

mangelajo: > I added an opt-out for now, but happy to change into opt-in if we see it's doable in rhel4edge or for any other reason.

oh sorry I didn't see you had written comments in the meantime, reading ^

mangelajo: > 

Ok, sounds like a good plan to me

Questions to make sure I understood right:

 * so, RHEL uses ostree admin underneath ? 
 * I guess I also must verify that switch/upgrade will use the right flag too

cgwalters: `ostree admin` is just a small CLI for the core `libostree` C shared library, which is also used by rpm-ostree (e.g. [here](https://github.com/coreos/rpm-ostree/blob/bc791794184802c3a405e3719f43ddbd168ae3ff/src/daemon/rpmostree-sysroot-upgrader.cxx#L1400)), which is what's used by most things in Fedora/RHEL today.  So modifying the code here will affect everything we want.

cgwalters: > I guess I also must verify that switch/upgrade will use the right flag too

Basically by design, every single ostree operation ends up reconstructing a new filesystem via hardlinks - it never tries to do an "in place" update.  `checkout_deployment_tree()` is the function that does that, and everything else calls it.

mangelajo: Thank you, undrestood

mangelajo: ostree_sysroot_deploy_tree -> ostree_sysroot_deploy_tree_with_options -> sysroot_initialize_deployment -> checkout_deploytment_tree -> ostree_repo_checkout_at

src/ostree/ot-builtin-checkout.c
```diff
@@ -193,6 +195,7 @@ process_one_checkout (OstreeRepo           *repo,
       checkout_options.force_copy = opt_force_copy;
       checkout_options.force_copy_zerosized = opt_force_copy_zerosized;
       checkout_options.bareuseronly_dirs = opt_bareuseronly_dirs;
+      checkout_options.process_passthrough_whiteouts = opt_process_passthrough_whiteouts;
```
mangelajo, line 198: I'm checking why this option is not getting down to the callee down the line in my local testing.

mangelajo:
```
+ ./ostree checkout --process-passthrough-whiteouts -v --repo=./testrepo-bare test/stable/x86_64 ./rootfs-out
.ostree-wh.whiteout is_overlayfs_whiteout: 1, enabled:0
not-a-whiteout-just-empty is_overlayfs_whiteout: 0, enabled:0
.ostree-wh.whiteout is_overlayfs_whiteout: 1, enabled:0
.ostree-wh.whiteout2 is_overlayfs_whiteout: 1, enabled:0
+ ls -la rootfs-out
total 8
drwxr-xr-x. 1 majopela majopela  102 Jan  1  1970 .
drwxr-xr-x. 1 majopela majopela 2058 Sep 20 14:24 ..
drwxr-xr-x. 1 majopela majopela   78 Jan  1  1970 another
-rw-r--r--. 1 majopela majopela    0 Sep 20 14:24 not-a-whiteout-just-empty
-rw-r--r--. 1 majopela majopela    0 Sep 20 14:24 .ostree-wh.whiteout
+ ls -la rootfs-out/another
total 0
drwxr-xr-x. 1 majopela majopela  78 Jan  1  1970 .
drwxr-xr-x. 1 majopela majopela 102 Jan  1  1970 ..
-rw-r--r--. 1 majopela majopela   0 Sep 20 14:24 .ostree-wh.whiteout
-rw-r--r--. 1 majopela majopela   0 Sep 20 14:24 .ostree-wh.whiteout2
+ getfattr -d -m '.*' rootfs-out/whiteout
getfattr: rootfs-out/whiteout: No such file or directory

```

I'm missing something here... ':D

mangelajo: I separated those for debugging:  
```
  const gboolean is_overlayfs_whiteout = (!is_symlink && g_str_has_prefix (destination_name, OVERLAYFS_WHITEOUT_PREFIX));

  printf("%s is_overlayfs_whiteout: %d, enabled:%d\n", destination_name, is_overlayfs_whiteout, options->process_passthrough_whiteouts);

```

mangelajo: oh wait... :) 
```
  /* This strange code structure is to preserve testing
   * coverage of both `ostree_repo_checkout_tree` and
   * `ostree_repo_checkout_at` until such time as we have a more
   * convenient infrastructure for testing C APIs with data.
   */
  if (opt_disable_cache || opt_whiteouts || opt_require_hardlinks ||
      opt_union_add || opt_force_copy || opt_force_copy_zerosized ||
      opt_bareuseronly_dirs || opt_union_identical ||
      opt_skiplist_file || opt_selinux_policy || opt_selinux_prefix ||
      opt_process_passthrough_whiteouts)
    {
```

mangelajo:
I used this as a small functional test:

```

[majopela@lenovo ostree]$ cat test-lifecycle.sh
#!/bin/sh
set -xe
sudo rm -rf ./tmp
podman pull --root ./tmp quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:ba44dead03ea74107f90d58525106fb52d27a120b73c6cc8e2be31d37043ca1c

find ./tmp -type d -exec getfattr -d {} \;
# this chokes on directories or filenames with spaces
find ./tmp -type c -exec sh -c 'touch "$(dirname $1)/.ostree-wh.$(basename $1)"; rm -f "$1"; echo $1' _ {} \;

mode=bare
repo="./testrepo-${mode}"
sudo rm -rf $repo rootfs-out
./ostree --repo=$repo init --mode=$mode
sudo ./ostree commit --repo=$repo ./tmp --branch test/stable/x86_64
sudo ./ostree checkout --process-passthrough-whiteouts -v --repo=$repo test/stable/x86_64 ./rootfs-out

find ./rootfs-out -type c

```

and it looks good, I´m going to feed it into crio as additional storage now and see if the whiteouts work.

mangelajo:
ok, running the image from an additionalImageStore extracted with ostree checkout and the new flag works correctly.

I guess remaining parts for this patch would be:
1) overwrite modes implemented
2) testing , I accept advice here userland testing would be ok?

and as additional PRs:

a) on this repository, adding the ability to encode whiteouts as .ostree-wh. files automatically
b) QUESTION: would this still need rpm-ostree support? or does rpm-ostree simply consume this library.

cgwalters:
> testing , I accept advice here userland testing would be ok?

Yeah that's fine

> b) QUESTION: would this still need rpm-ostree support? or does rpm-ostree simply consume this library.

I'm 97.3% sure no changes to rpm-ostree are necessary for this.

mangelajo:
> > b) QUESTION: would this still need rpm-ostree support? or does rpm-ostree simply consume this library.
> 
> I'm 97.3% sure no changes to rpm-ostree are necessary for this.

🤣

mangelajo:
Ok, this should be now feature complete, writing tests.

Later down I can squash or reorganize the commits as necessary

tests/test-admin-deploy-whiteouts.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+#
+# Copyright (C) 2020 Red Hat, Inc.
```
mangelajo, line 3:
```suggestion
# Copyright (C) 2022 Red Hat, Inc.
```

src/libostree/ostree-repo-checkout.c
```diff
@@ -582,6 +584,109 @@ checkout_file_hardlink (OstreeRepo                          *self,
...
+_checkout_overlayfs_whiteout_at_no_overwrite (OstreeRepoCheckoutAtOptions    *options,
+                                              int                             destination_dfd,
+                                              const char                     *destination_name,
+                                              GFileInfo                      *file_info,
+                                              GVariant                       *xattrs,
+                                              GCancellable                   *cancellable,
+                                              GError                        **error)
+{
+  guint32 file_mode = g_file_info_get_attribute_uint32 (file_info, "unix::mode");
+  if (mknodat(destination_dfd, destination_name, (file_mode & ~S_IFMT) | S_IFCHR, (dev_t)0) < 0)
```
cgwalters, line 597: Don't we store the file type in the ostree mode too?  Curious why you're explicitly forcing on `S_IFCHR` here.

cgwalters:
Hm, looks like all the tests are failing with
`2022-09-26T16:41:39.8776463Z error: Checking out deployment tree: fchmodat: Operation not supported`...I wonder if this is related to the file type bits above.

src/libostree/ostree-repo-checkout.c
mangelajo: In ostree it's stored as a regular file, otherwise in bare mode we would have to create character devices to get those bits preserved. Ghfn it would look more like my original approach here https://github.com/ostreedev/ostree/pull/2713 which was more complex

mangelajo: or are we extracting the S_IFCHR before this call? checking,..

mangelajo: ah, no, we just detect the .ostree-wh. prefix for a regular empty file, which removes the prefix and triggers this call

mangelajo:
> Hm, looks like all the tests are failing with `2022-09-26T16:41:39.8776463Z error: Checking out deployment tree: fchmodat: Operation not supported`...I wonder if this is related to the file type bits above.

it seemed to be that not all glibs support the NO_FOLLOW_SYMLINK flag which we don't use in other parts of the code.

src/libostree/ostree-repo-checkout.c
```diff
@@ -582,6 +584,108 @@ checkout_file_hardlink (OstreeRepo                          *self,
...
+                                 GVariant                       *xattrs,
+                                 GCancellable                   *cancellable,
+                                 GError                        **error)
+{
+  if (_checkout_overlayfs_whiteout_at_no_overwrite(options, destination_dfd, destination_name,
+                                                   file_info, xattrs, cancellable, error))
+    return TRUE;
+
+  if (!error || !g_error_matches(*error, G_IO_ERROR, G_IO_ERROR_EXISTS))
+    return FALSE;
```
cgwalters, line 633: This isn't allowed by GError conventions.  It's allowed to pass `NULL` to ignore errors, in the caller, though this isn't used much in the ostree stack.

This topic heavily relates to e.g. https://users.rust-lang.org/t/why-i-use-anyhow-error-even-in-libraries/68592 where I argue that we want to separate out the "expected/likely error" case from the "disk is failing" cases.

Specifically here, we need to pass back the "found an extant file" as a separate boolean.  There's precedent for this in APIs like `glnx_fstatat_allow_noent()` and `ot_openat_ignore_enoent` etc.

Something a bit more like:

```
  gboolean found_extant_file = FALSE;
  if (!_checkout_overlayfs_whiteout_at_no_overwrite(options, destination_dfd, destination_name,
                                                   file_info, xattrs, &found_extant_file, cancellable, error))
    return FALSE;
 
  if (!found_extant_file)
    return TRUE;

  /* fallthrough */
```

mangelajo: Thanks for the feedback, handling it

cgwalters (approved):
Thanks!

PR #2718, opened 2022-09-23
lib/repo: properly initialize boolean variable

lucab:
This initializes a boolean variable that was previously left uninitialized.
It was detected by a RHT internal static analyzer.


cgwalters (approved):
ouch

PR #2742, opened 2022-10-24
rust: regenerate bindings after latest release (2022.6)

lucab:
This updates Rust binding to latest libostree release 2022.6.
~~It also contains version bumps for crates `ostree-sys 0.11.0` and `ostree 0.16.0`~~


cgwalters (dismissed):
LGTM, though note the semver bump here ripples through higher level things.

In the future I'd like to consider trying to e.g. batch this change with things like updating to the latest gtk-rs (also a semver bump).

But, LGTM as is.

lucab:
Fair point. For the moment I've dropped the last commit from here, which was bumping the versions in `Cargo.toml`.
There is another semver bump upcoming for https://github.com/ostreedev/ostree/issues/2647.
I wasn't sure how much work that would be, but I think I already managed to complete the porting at https://github.com/ostreedev/ostree/pull/2743.
We can decide to directly jump to that and tag the new semver-breaking releases already with the new gtk-rs requirements.

cgwalters (approved):
Thanks!

PR #2743, opened 2022-10-25
rust-bindings: regenerate for gtk-rs 0.15

lucab:
This updates our Rust bindings to version 0.15 of the gtk-rs stack, see https://gtk-rs.org/blog/2022/01/16/new-release.html.
Notably, we now require a Rust toolchain >= 1.56 and GLib/GObject/GIO >= 2.48.

Closes: https://github.com/ostreedev/ostree/issues/2647


cgwalters:
That seems easier than I'd expected!  Did you hit anything surprising or problematic?

lucab:
> Did you hit anything surprising or problematic?

Notable things I spotted while assembling this PR:
 * `NONE_CANCELLABLE` is gone. This cascades quite a bit to consumers.
 * `ostree_repo_finder_avahi_start()` is actually missing the usual bool return value. May be worth replacing with a `_try_start()` with a better signature.
 * some misgenerated GString code on `ostree_gpg_verify_result_describe*()`, likely a `gir-rs` bug. Still need to look into it, just ignoring the two functions for now.

lucab:
Ok, this one should be ready to merge.
Once in, we can do the semver bumps and release crates `ostree-sys 0.11.0` and `ostree 0.16.0`.

cgwalters (approved):

lucab:
Ah yes, that one comes with a fairly large MSRV jump though.
On the gtk-rs there is also https://gtk-rs.org/blog/2022/10/18/new-release.html which just came out and also requires a toolchain >= 1.63.
I think we can do a semver bump here for this, and then bundle the new cap-std and gtk-rs together for another bump in a few months when the dust settles. How do you feel about this plan?

cgwalters:
SGTM

PR #2748, opened 2022-10-27
ci/rust: test from git using latest feature

lucab:
A late CI update with a followup for https://github.com/ostreedev/ostree/pull/2742.


jmarrero (approved):
/lgtm

lucab:
Prow didn't trigger, but we don't really need it here.

/override ci/prow/sanity

PR #2751, opened 2022-10-28
Replace calls to g_memdup() with g_memdup2()

smcv:
* Update submodule: libglnx
    
    - Backport g_memdup2 (see #2751)

* Replace calls to g_memdup() with g_memdup2()

    g_memdup() is subject to an integer overflow on 64-bit machines if the
    object being copied is larger than UINT_MAX bytes. I suspect none of
    these objects can actually be that large in practice, but it's easier
    to replace all the calls than it is to assess whether we need to
    replace them.

    A backport in libglnx is used on systems where GLib is older than 2.68.x.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -4858,7 +4858,7 @@ _ostree_repo_import_object (OstreeRepo           *self,
 static OstreeRepoTransactionStats *
 ostree_repo_transaction_stats_copy (OstreeRepoTransactionStats *stats)
 {
-  return g_memdup (stats, sizeof (OstreeRepoTransactionStats));
+  return g_memdup2 (stats, sizeof (OstreeRepoTransactionStats));
 }
 
 static void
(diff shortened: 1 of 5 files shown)


lucab:
Nice! Do we want to have a coccinelle semantic patch to enforce this for future code too?

smcv:
> Do we want to have a coccinelle semantic patch to enforce this for future code too?

I don't really know how coccinelle works; but I think that would be non-trivial, because libostree wants to be backportable to old GLib, which means files that use `g_memdup2()` have to include `libglnx.h` either directly or indirectly.

For the instances of `g_memdup` that I replaced here, I checked that the source file included either `libglnx.h` or `otutil.h`.

smcv:
There is one additional use of `g_memdup` which is not fixed by this PR, because it's in a source file that does not include `libglnx.h` and is not linked to libglnx (it's a `LD_PRELOAD` module). See #2752 for that.

smcv:
If the libostree team like this change, please could someone look at the libglnx MR?

cgwalters:
Rebased :surfer: this one to pick up the CI fixes

cgwalters (approved):

PR #2752, opened 2022-10-28
readdir-rand: Copy full size of struct dirent

smcv:
As noted in readdir(3), in the presence of long filenames it is possible for a directory entry to be larger than `sizeof (struct dirent)`. Copy the full length instead.

---

I happened to notice this while replacing `g_memdup` with `g_memdup2` for #2751.

### DIFF
--- a/tests/readdir-rand.c
+++ b/tests/readdir-rand.c
@@ -40,6 +40,33 @@
 # define READDIR_R "readdir_r"
 #endif
 
+/*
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Looks sane; going to merge this to give you credit for the work, but I think it's time to just remove this code.  It's only for testing and shouldn't be needed anymore.

PR #2754, opened 2022-10-31
Remove readdir-rand

cgwalters:
This was only there to reproduce a bug we hit long ago with bootloader file ordering.  We're extremely unlikely to reintroduce such a bug, and it's not worth carrying around this code.


cgwalters:
/override  continuous-integration/jenkins/pr-merge
Tracking the pxe/metal failure in https://github.com/ostreedev/ostree/issues/2757

JAORMX (approved):

lucab (approved):

Issue #2747, opened 2022-10-27
[RFE] Retry mechanism for HTTP 500 error when pulling ostree


oglok:
@mangelajo and I have been experiencing issues with pulling ostrees hosted in S3 buckets. This is framed in the context of the creation of ISO images, and the whole process would be interrupted if the request to pull one filez throws an HTTP 500 error.

We would like to extend the existing retrying mechanism to cover this type of http error.

Feedback would be much appreciated.

Thanks in advance.

mangelajo:
I have also found this on rhel4edge when triggering an update from the edge management console, i.e.: 
![image](https://user-images.githubusercontent.com/1176434/198263411-98e27a59-538e-4be9-9db6-73992c6f2325.png)

mangelajo:
@cgwalters @lucab Is the repo fetching also rust native in rpm-ostree, or is it consumed from here? If it is, I guess we should open a parallel RFE in rpm-ostree.

oglok:
ok, let me push a PR to tackle this. Thanks for the info

lucab:
Maybe I'm missing some details, but how did you deduce this happened in reaction to an HTTP 500 response?
The logs you pasted seem to point instead to `CURLE_RECV_ERROR`, which should be some kind of problem with the underlying network/transport instead, I think. At that point `libcurl` is possibly having issues with your whole networking I/O in general (i.e.possibly affecting DNS, TLS, etc.), not just with a HTTP `GET` to retry.

oglok:
> Maybe I'm missing some details, but how did you deduce this happened in reaction to an HTTP 500 response? The logs you pasted seem to point instead to `CURLE_RECV_ERROR`, which should be some kind of problem with the underlying network/transport instead, I think. At that point `libcurl` is possibly having issues with your whole networking I/O in general (i.e.possibly affecting DNS, TLS, etc.), not just with a HTTP `GET` to retry.

```
**CloudWatch Logs Insights**  
region: us-east-1  
log-group-names: workers-aoc-production  
start-time: 2022-10-25T09:31:10.551Z  
end-time: 2022-10-25T09:34:42.316Z  
query-string:
```
fields message
| filter(host like "worker-aoc-production-i-0519acbbbd2aa4eb7")
| sort @timestamp desc
| limit 3000
```
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
[...]

Issue #2757, opened 2022-11-03
kola:metal test consistently failing


cgwalters:
```
[2022-11-01T21:31:41.343Z] + set -xeuo pipefail
[2022-11-01T21:31:41.343Z] ++ umask
[2022-11-01T21:31:41.343Z] + '[' 0000 = 0000 ']'
[2022-11-01T21:31:41.343Z] + umask 0022
[2022-11-01T21:31:41.343Z] + cd /home/jenkins/agent/workspace/ostree_PR-2754
[2022-11-01T21:31:41.343Z] + cosa kola testiso -S --output-dir /home/jenkins/agent/workspace/ostree_PR-2754/tmp/kolaTestIso-8MZrB/kola-testiso-metal
[2022-11-01T21:31:41.343Z] kola --output-dir /home/jenkins/agent/workspace/ostree_PR-2754/tmp/kolaTestIso-8MZrB/kola-testiso-metal testiso -S
[2022-11-01T21:31:41.343Z] Testing scenarios: [pxe-install iso-offline-install pxe-offline-install iso-live-login iso-as-disk miniso-install miniso-install-nm]
[2022-11-01T21:31:41.343Z] Ignoring verification of signature on metal image
[2022-11-01T21:32:48.951Z] FAIL: pxe-install (bios + metal) (1m6.033s)
[2022-11-01T21:32:48.951Z]     Got EOF from completion channel, coreos-installer-test-OK expected
[2022-11-01T21:32:48.951Z] Error: Got EOF from completion channel, coreos-installer-test-OK expected
[2022-11-01T21:32:48.951Z] 2022-11-01T21:32:48Z cli: Got EOF from completion channel, coreos-installer-test-OK expected
[2022-11-01T21:32:48.951Z] error: failed to execute cmd-kola: exit status 1
script returned exit code 1
```

Not sure what's going on here still.  Looking at the console logs, I see us successfully doing the install, then booting from the hard disk...then the last message is

```
[    2.675247] systemd[1]: Started multipathd.service - Device-Mapper Multipath Device Controller.
```

Then...that's it?  We didn't complete anything else in the initramfs or transition to the real root.

cgwalters:
This turned out to be memory limits.

PR #2759, opened 2022-11-08
Enabling retry for HTTP 500 internal server error

oglok:
Signed-off-by: Ricardo Noriega <rnoriega@redhat.com>

Hello! This PR tries to address the issue #2747 . 

Manual testing shows that it follows the path of network retries, as other error codes:

```
OT: Should retry request (remaining: 5 retries), due to transient error: Server returned status 500: Internal Server Error
OT: starting fetch of be117dcfcf8b16c9ca2831a1dc9e6cb92944ee57d8e00adeb4cd769e95ac436e.file
OT: starting fetch of 582dda5ed1e27f82d001b220f96e43863c56d6143f0fdb930b2c808b514c45b4.dirtree
OT: fetch of 582dda5ed1e27f82d001b220f96e43863c56d6143f0fdb930b2c808b514c45b4.dirtree complete
OT: _ostree_fetcher_should_retry_request: error: 321:2 Server returned status 500: Internal Server Error, n_retries_remaining: 5
OT: Should retry request (remaining: 5 retries), due to transient error: Server returned status 500: Internal Server Error
OT: starting fetch of 582dda5ed1e27f82d001b220f96e43863c56d6143f0fdb930b2c808b514c45b4.dirtree
OT: fetch of 582dda5ed1e27f82d001b220f96e43863c56d6143f0fdb930b2c808b514c45b4.dirtree complete
OT: _ostree_fetcher_should_retry_request: error: 321:2 Server returned status 500: Internal Server Error, n_retries_remaining: 4
OT: Should retry request (remaining: 4 retries), due to transient error: Server returned status 500: Internal Server Error
```

I have a question though. The error code that ostree-trivial-httpd is returning consists of:

- Domain: G_IO_ERROR which corresponds to number 321
- Code: G_IO_ERROR_FAILED which corresponds to number 0. 

I changed this last code to G_RESOLVER_ERROR_INTERNAL which corresponds to number 2, because I thought it matches the nature of HTTP 500 Internal Server error. However, in ostree-trivial-httpd.c, domain cannot be changed. I would have expected to use G_RESOLVER_ERROR domain in [g_set_error()](https://github.com/ostreedev/ostree/blob/main/src/ostree/ostree-trivial-httpd.c#L593) , but it keeps throwing 321 (G_IO_ERROR). Am I missing something?

If you think it's fine like this, I'll remove the WIP. 

Thanks in advance!

### DIFF
--- a/src/libostree/ostree-fetcher-util.c
+++ b/src/libostree/ostree-fetcher-util.c
@@ -239,6 +239,7 @@ _ostree_fetcher_should_retry_request (const GError *error,
       g_error_matches (error, G_IO_ERROR, G_IO_ERROR_HOST_NOT_FOUND) ||
       g_error_matches (error, G_IO_ERROR, G_IO_ERROR_HOST_UNREACHABLE) ||
       g_error_matches (error, G_IO_ERROR, G_IO_ERROR_PARTIAL_INPUT) ||
+      g_error_matches (error, G_IO_ERROR, G_IO_ERROR_BUSY) ||
 #if !GLIB_CHECK_VERSION(2, 44, 0)
       g_error_matches (error, G_IO_ERROR, G_IO_ERROR_BROKEN_PIPE) ||
 #else
@@ -269,6 +270,8 @@ _ostree_fetcher_http_status_code_to_io_error (guint status_code)
       return G_IO_ERROR_NOT_FOUND;
     case 408:  /* SOUP_STATUS_REQUEST_TIMEOUT */
       return G_IO_ERROR_TIMED_OUT;
+    case 500:  /* SOUP_STATUS_INTERNAL_SERVER_ERROR */
+      return G_IO_ERROR_BUSY;
     default:
       return G_IO_ERROR_FAILED;
     }
--- a/tests/test-pull-repeated.sh
+++ b/tests/test-pull-repeated.sh
@@ -26,18 +26,32 @@ if has_gpgme; then
     COMMIT_SIGN="--gpg-homedir=${TEST_GPG_KEYHOME} --gpg-sign=${TEST_GPG_KEYID_1}"
 fi
 
-echo "1..4"
+echo "1..6"
 
-# Test pulling from a repo which gives error 500 (internal server error) a lot of the time.
+# Sanity check with no network retries and 500s given, pull should fail.
+rm ostree-srv httpd repo -rf
+setup_fake_remote_repo1 "archive" "${COMMIT_SIGN}" --random-500s=99
+
+pushd ${test_tmpdir}
+ostree_repo_init repo --mode=archive
+${CMD_PREFIX} ostree --repo=repo remote add --set=gpg-verify=false origin $(cat httpd-address)/ostree/gnomerepo
+assert_fail ${CMD_PREFIX} ostree --repo=repo pull --mirror origin --network-retries=0 main 2>err.txt
+assert_file_has_content err.txt "\(500.*Internal Server Error\)\|\(HTTP 500\)"
+
+popd
+echo "ok no retries after a 500"
+
+# Test pulling a repo which gives error 500 (internal server error) a lot of the time.
+rm ostree-srv httpd repo -rf
 setup_fake_remote_repo1 "archive" "${COMMIT_SIGN}" --random-500s=50
 
 pushd ${test_tmpdir}
 ostree_repo_init repo --mode=archive
 ${CMD_PREFIX} ostree --repo=repo remote add --set=gpg-verify=false origin $(cat httpd-address)/ostree/gnomerepo
-for x in $(seq 200); do
-    if ${CMD_PREFIX} ostree --repo=repo pull --mirror origin main 2>err.txt; then
-	echo "Success on iteration ${x}"
-	break;
+for x in $(seq 40); do
+    if ${CMD_PREFIX} ostree --repo=repo pull --mirror origin --network-retries=2 main 2>err.txt; then
+       echo "Success on iteration ${x}"
+       break;
     fi
     assert_file_has_content err.txt "\(500.*Internal Server Error\)\|\(HTTP 500\)"
 done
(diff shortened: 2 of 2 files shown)


dbnicholson:
@oglok it's a different error domain (`G_IO` vs `G_RESOLVER`). I don't know if gio has a good error match for a 500.

dbnicholson:
> Regarding my question, do you know why doing something like the following [code](https://github.com/ostreedev/ostree/blob/main/src/ostree/ostree-trivial-httpd.c#L593) in `ostree-trivial-httpd.c` keeps throwing an error with domain `G_IO_ERROR` (321) instead of `G_RESOLVER_ERROR` ?:
> 
> ```
>   if (!(opt_random_500s_percentage >= 0 && opt_random_500s_percentage <= 99))
>     {
>       g_set_error (error, G_RESOLVER_ERROR, G_RESOLVER_ERROR_INTERNAL,
>                    "Invalid --random-500s=%u", opt_random_500s_percentage);
>       goto out;
>     }
> ```

This error would never actually reach the client. trivial-httpd is just a testing HTTP server, and this part is just the argument parsing code for it. It's already doing what you want - return HTTP 500 status codes at random times. Any errors hit here would just make the server die. The fact that it was using `G_IO_ERROR_FAILED` was just a coincidence and you can leave trivial-httpd alone. The other part of your patch is the important part - make the HTTP client manage a 500 status code returned from the server in a different way.

src/libostree/ostree-fetcher-util.c
```diff
@@ -243,6 +243,7 @@ _ostree_fetcher_should_retry_request (const GError *error,
       g_error_matches (error, G_IO_ERROR, G_IO_ERROR_BROKEN_PIPE) ||
 #else
       g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CONNECTION_CLOSED) ||
+      g_error_matches (error, G_IO_ERROR, G_IO_ERROR_BUSY) ||
```
dbnicholson, line 246: You can move this above the `#if` since it's not glib version dependent.

dbnicholson:
For some related background, @wjt pointed me to #1847 from the past. I.e., there should really be a separate error domain instead of trying to overload `GIOErrorEnum`. However, that would technically be an API break.

oglok:
> 

I've been using this trivial-httpd server for testing (great tool btw). So ok, it's not returning anything to the client, so it was just a coincidence. I just wanted to make align the resulting code (and domain) from the test server with what I checked in the client when doing `ostree pull`.

Thanks for the info!

src/ostree/ostree-trivial-httpd.c
```diff
@@ -590,7 +590,7 @@ run (int argc, char **argv, GCancellable *cancellable, GError **error)
 
   if (!(opt_random_500s_percentage >= 0 && opt_random_500s_percentage <= 99))
     {
-      g_set_error (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+      g_set_error (error, G_IO_ERROR, G_IO_ERROR_BUSY,
```
dbnicholson, line 593: You can just drop this part. As mentioned in the PR conversation, it's just coincidental that it also uses `G_IO_ERROR_FAILED`. This represents an error in the HTTP server if you give it a bogus `--random-500s` and doesn't reflect what error the HTTP client will receive. The only changes needed are on the client side.

dbnicholson (changes requested):
I think this generally makes sense, but I'm not sure about further overloading `GIOErrorEnum`.

As to the changes, I think you can squash them all down into one patch. A test would also be good.

oglok:
> For some related background, @wjt pointed me to #1847 from the past. I.e., there should really be a separate error domain instead of trying to overload `GIOErrorEnum`. However, that would technically be an API break.

Yes, I first looked at that PR to implement the 500 error, but then I realized most of the logic was already in place, and it was just a matter of enabling the code number in `_ostree_fetcher_http_status_code_to_io_error`.

oglok:
> I think this generally makes sense, but I'm not sure about further overloading `GIOErrorEnum`.
> 
> As to the changes, I think you can squash them all down into one patch. A test would also be good.

Are these tests enough? https://github.com/ostreedev/ostree/blob/main/tests/test-pull-repeated.sh#L31-L49

dbnicholson:
> > I think this generally makes sense, but I'm not sure about further overloading `GIOErrorEnum`.
> > As to the changes, I think you can squash them all down into one patch. A test would also be good.
> 
> Are these tests enough? https://github.com/ostreedev/ostree/blob/main/tests/test-pull-repeated.sh#L31-L49

Sort of. Now that you've added retries, that test is going to be less useful since sometimes it will get a 500, retry and get a correct response. You're essentially moving the logic of repeated `ostree pull` into the pull logic itself. That would mask the original intent of making sure that `pull` can be repeated.

I think what you want to do is to copy the set of tests below that using 408s. In other words:

* With `--random-500s=99` and `--network-retries=0`, the error propagates out so `ostree pull` fails.
* With `--random-500s=50` and `--network-retries=2`, you can loop `ostree pull` and eventually succeed.
* With `--random-500s=50` and `--network-retries=100`, you can succeed on a single `ostree pull` run.

Probably that test could be refactored a bit so it's not repeating everything, but maybe another day.

oglok:
```
PASS: tests/test-pull-repeated.sh 1 no retries after a 500
PASS: tests/test-pull-repeated.sh 2 repeated pull after 500s
PASS: tests/test-pull-repeated.sh 3 big number of retries with one 500s
PASS: tests/test-pull-repeated.sh 4 no retries after a 408
PASS: tests/test-pull-repeated.sh 5 repeated pull after 408s
PASS: tests/test-pull-repeated.sh 6 big number of retries with one 408
============================================================================
Testsuite summary for libostree 2022.7
============================================================================
# TOTAL: 6
# PASS:  6
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

cgwalters:
Thanks so much for the patch!  Definitely an overdue change.

I'd say we should also retry on e.g. [502](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) and 504 at least - I've seen this from some CDNs, and I think they also qualify as "possibly transient" errors.

cgwalters (approved):

dbnicholson:
> I'd say we should also retry on e.g. [502](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/502) and 504 at least - I've seen this from some CDNs, and I think they also qualify as "possibly transient" errors.

See also #1847. That PR used `G_IO_ERROR_TIMED_OUT` for 504 Gateway Timeout like 408 Request Timeout. But the other conversation in that PR was to stop abusing `GIOErrorEnum` and instead create a specific `OSTREE_HTTP_ERROR` domain. If we continue to map to `GIOErrorEnum`, I'd go with:

500 Internal Server Error -> `G_IO_ERROR_BUSY`
502 Bad Gateway -> `G_IO_ERROR_BUSY`
503 Service Unavailable -> `G_IO_ERROR_BUSY`
504 Gateway Timeout -> `G_IO_ERROR_TIMED_OUT`

The former 3 feel a bit more like a transient error. The latter is explicitly a timeout.

cgwalters:
Yeah, it's hard to know the blast radius of changing the error domain.  I'm not opposed but personally I'd lean against doing so.

cgwalters:
@dbnicholson are you OK to merge this patch as is?  If so can you change your review status from "changes needed" to "approved"?

dbnicholson (approved):

dbnicholson:
> @dbnicholson are you OK to merge this patch as is? If so can you change your review status from "changes needed" to "approved"?

Sure. Sorry for holding it up.

cgwalters:
Blah, will dig into https://github.com/ostreedev/ostree/issues/2757 at some point...

oglok:
Thanks for merging this PR @cgwalters and for all your support @dbnicholson !

mangelajo:
Awesome!, nice work

Issue #2758, opened 2022-11-08
ostree fails with: bwrap: open /proc/192/ns/ns failed: No such file or directory


flaper87:
I'm trying to create a live-iso using livemedia-creator. This project runs anaconda in the background, which then calls `ostree`

https://github.com/rhinstaller/anaconda/blob/f0df778cc43e2494fdd0fad131aa8643590510ac/pyanaconda/modules/payloads/payload/rpm_ostree/installation.py#L511-L521

I captured the logs from ostree and this is what I got:

```
OT: Didn't find $sysroot/boot/loader symlink; assuming bootversion 0
OT: Didn't find $sysroot/ostree/boot.0 symlink; assuming subbootversion 0
OT: Preparing transaction in repository 0x55925a240140
OT: Pushing lock non-blocking with timeout 30
OT: Opening repo lock file
OT: Push lock: state=unlocked, depth=0
OT: Locking repo shared
OT: Using new tmpdir staging-3470f58f-7cec-4c35-b82a-a92470d1b924-gSlmOP
OT: Committing transaction in repository 0x55925a240140
OT: txn commit staging-3470f58f-7cec-4c35-b82a-a92470d1b924-gSlmOP
OT: Popping lock non-blocking with timeout 30
OT: Pop lock: state=shared, depth=1
OT: Unlocking repo
Regex version mismatch, expected: 10.37 2021-05-26 actual: 10.32 2018-09-10
Regex version mismatch, expected: 10.37 2021-05-26 actual: 10.32 2018-09-10
note: Deploying commit 031cdee1ba34753c862d3da109a6b6532540235977036865f80298dbcd0af905 which contains content in /var/lib that will be ignored.
bwrap: open /proc/192/ns/ns failed: No such file or directory
```

# Environment

This is a rhel9 VM

```
[cloud-user@rhel9 ~]$ dnf info ostree
Not root, Subscription Management repositories not updated

This system is not registered with an entitlement server. You can use subscription-manager to register.

Last metadata expiration check: 0:00:16 ago on Tue 08 Nov 2022 03:44:30 AM EST.
Installed Packages
Name         : ostree
Version      : 2022.5
Release      : 1.el9_0
Architecture : x86_64
Size         : 578 k
Source       : ostree-2022.5-1.el9_0.src.rpm
Repository   : @System
From repo    : rhel-9-for-x86_64-appstream-rpms
Summary      : Tool for managing bootable, immutable filesystem trees
[...]

PR #2760, opened 2022-11-09
deploy: Don't rebuild selinux policy on first deployment

cgwalters:
Basically, it should not be necessary - the policy should be up-to-date.  We don't want to force on continual policy rebuilds.

Even trying to run bwrap when we're *not* in a booted root can cause failures in nested containerization scenarios.

Closes: https://github.com/ostreedev/ostree/issues/2758


JAORMX (approved):

lucab (approved):

flaper87:
I'm not familiar with the CI failure. Is it related to this PR?

I could help testing it in my env.

lucab:
No, the failure was an infra-related flake, tracked at https://github.com/coreos/fedora-coreos-tracker/issues/1339.
I re-triggered it once more and it passed.

PR #2762, opened 2022-11-11
copr: only use libostree tags

lucab:
This adds a tag filter to the logic which emits version labels for COPR build, so that it avoids mistakenly picking up tags belonging to the Rust bindings.


cgwalters (approved):

cgwalters:
Nice!

lucab:
This didn't work exactly as planned :disappointed:. It looks like there is another bit of logic somewhere which still need to be tweaked to ignore the Rust tags.

PR #2763, opened 2022-11-11
ci: Bump memory for ISO testing flow

cgwalters:
It seems likely that we're allocating more RAM here.  What we really need to do is for cosa to express the necessary RAM requirements declaratively, then we compute pod requirements from that.


jmarrero (approved):
lgtm

Issue #2732, opened 2022-10-10
ostree pull dies via oom-killer on low-memory system


hgroover:
First unusual event: ostree pull (using libcurl, https) is reaped by the oom-killer on a very large delta pull.
Kernel version 4.9.217 mips
Total available physical memory 85mb (physical DRAM is 128mb, minus a large CMA chunk, minus kernel and wpa_supplicant).

For the most part, I am able to successfully run ostree pull. However, if the total quantity of deltas is high, ostree is likely to use up all available memory and get reaped by the oom-killer.
This happens less frequently if the thread limits in ostree-repo-private.h are set to lower values, but is still guaranteed to happen on a large set of deltas. I am currently investigating running under ulimit as a workaround but have to ask if there might be a better way to avoid this?

In the usual case I get an errorlevel 137 in a script and can retry the ostree pull repeatedly until it succeeds. In the worst case I will get an oom-killer murder spree that begins terminating every process on the system, and while looking for victims an I/O thread fails and results in ubifs corruption, resulting in /ostree left in a read-only state. In short, I need to avoid oom-killer deaths. I have very little memory on this system and enough NAND to pull the update (but not enough for a second rootfs).

jlebon:
Interesting specs. What values did you try for e.g. `_OSTREE_MAX_OUTSTANDING_FETCHER_REQUESTS`? Relatedly, I think it could make sense to make that into a config knob. Also, to confirm, is the remote using static deltas? If so, try with --disable-static-deltas` too.

hgroover:
The three values I've been changing are
_OSTREE_MAX_OUTSTANDING_FETCHER_REQUESTS 3
_OSTREE_MAX_OUTSTANDING_DELTAPART_REQUESTS 1
_OSTREE_MAX_OUTSTANDING_WRITE_REQUESTS 3

The write requests limit was changed to 3 in a release soon after 2020.3 (I've gone through everything up to and including 2022.2 - I think it was 2022.3 that went to a newer GLib version (and I'm running a very old uclibc build so updating GLib is going to be a long process).
I've also used 4 for fetcher requests.
What I've tried to do is limit threads because part of my problem seems to be that on the kernel version I have, the oom-reaper has a cadence of dumping about one candidate thread per second regardless of the VM setting for oom-killer dump, and if there are enough threads it SOMETIMES creates the conditions for an I/O thread to run out of memory before any process can get reaped - in short, a kind of oom-killer double-fault situation.

[...]

PR #2766, opened 2022-11-14
fetcher: Avoid too large queues for metadata processing

cgwalters:
We added backoff/queueing for fetching via HTTP, but we have another queue in the metadata scanning which can also grow up to the number of outstanding objects, which can be large.

Capping the scanning operation when we have hit our operation limit will avoid potentially large amounts of allocations in the case of e.g. a slow network.

Closes: https://github.com/ostreedev/ostree/issues/2732


cgwalters:
I only did basic unit testing of this locally.  I think what we want to do here is set up a CI flow that tries doing a large update inside a container image with limited RAM.

cgwalters:
I think we can probably merge this now, and let it soak a bit in CI.

jlebon (approved):
Haven't looked at this code in a while, but it looks sane to me!

PR #2770, opened 2022-11-16
ci: adjust git tags filtering

lucab:
Another attempt at fixing logic for version detection in COPR.

Followup to https://github.com/ostreedev/ostree/pull/2762.


cgwalters (approved):
Thanks so much for digging into this!

cgwalters:
This can't break CI here so

lucab:
Ok, this one properly built with correct versioning: https://copr.fedorainfracloud.org/coprs/g/CoreOS/continuous/build/5057720/

cgwalters:
Nice, thanks! :pray:

PR #2771, opened 2022-11-17
ostree_kernel_args_contains for OstreeKernelArgs

### LINKED ISSUE #2329: Enhance OstreeKernelArgs API for easier idempotent append and delete operations
For doing https://github.com/coreos/rpm-ostree/issues/2709, it'd be nice if `OstreeKernelArgs` supported e.g. `ostree_kernel_args_append_if_missing` and `ostree_kernel_args_delete_if_present`.

Or... maybe we just want a `ostree_kernel_args_contains` for now which is more general and avoids having to expose the underlying data structures.

Razaloc:
Check if an argument is present in OstreeKernelArgs. This is a way to make easier idempotent append and delete operations.

Closes #2329

Signed-off-by: Rafael Garcia Ruiz <rafael.garcia@collabora.com>

### DIFF
--- a/src/libostree/ostree-kernel-args.c
+++ b/src/libostree/ostree-kernel-args.c
@@ -820,12 +820,31 @@ void
 ostree_kernel_args_append_if_missing (OstreeKernelArgs  *kargs,
                                       const char *arg)
 {
-  g_autofree char *key = g_strdup (arg);
-  split_keyeq (key);
- 
   // Don't insert a duplicate key.
-  if (g_hash_table_contains (kargs->table, key))
+  if (ostree_kernel_args_contains (kargs, arg))
     return;
  
   ostree_kernel_args_append (kargs, arg);
+}
+
+/**
+ * ostree_kernel_args_contains:
+ * @kargs: a OstreeKernelArgs instance
+ * @arg: key or key/value pair to check
+ *
+ * Search for @arg which is in the form of key=value pair at the hash table kargs->table
+ * and returns true if finds it.
+ *
+ *Returns: %TRUE if @arg is contained in @kargs, %FALSE otherwise.
+ *
+ * Since: 2022.7
+ **/
+gboolean
+ostree_kernel_args_contains (OstreeKernelArgs  *kargs,
+                             const char *arg)
+{
+  g_autofree char *key = g_strdup (arg);
+  split_keyeq (key);
+
+  return g_hash_table_contains (kargs->table, key);
 }
--- a/src/libostree/ostree-kernel-args.h
+++ b/src/libostree/ostree-kernel-args.h
@@ -134,4 +134,8 @@ _OSTREE_PUBLIC
 void ostree_kernel_args_append_if_missing (OstreeKernelArgs *kargs, 
                                            const char *arg);
 
+_OSTREE_PUBLIC
+gboolean ostree_kernel_args_contains (OstreeKernelArgs *kargs,
+                                      const char *arg);
+
 G_END_DECLS
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -171,9 +171,9 @@ endif # USE_GPGME
 symbol_files = $(top_srcdir)/src/libostree/libostree-released.sym
 
 # Uncomment this include when adding new development symbols.
-#if BUILDOPT_IS_DEVEL_BUILD
-#symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
-#endif
+if BUILDOPT_IS_DEVEL_BUILD
+symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
+endif
 
 # http://blog.jgc.org/2007/06/escaping-comma-and-space-in-gnu-make.html
 wl_versionscript_arg = -Wl,--version-script=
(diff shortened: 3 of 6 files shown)


src/libostree/ostree-kernel-args.c
```diff
@@ -829,3 +829,27 @@ ostree_kernel_args_append_if_missing (OstreeKernelArgs  *kargs,
...
+                             const char *arg)
+{
+  g_autofree char *key = g_strdup (arg);
+  split_keyeq (key);
+
+  // Don't insert a duplicate key.
+  if (g_hash_table_contains (kargs->table, key))
+    return TRUE;
+
+  return FALSE;
```
dbnicholson, line 854: Why not just return the value from `g_hash_table_contains`? It's a `gboolean`.

src/libostree/ostree-kernel-args.h
```diff
@@ -134,4 +134,8 @@ _OSTREE_PUBLIC
 void ostree_kernel_args_append_if_missing (OstreeKernelArgs *kargs, 
                                            const char *arg);
 
+_OSTREE_PUBLIC
+gboolean ostree_kernel_args_contains (OstreeKernelArgs *kargs,
+                                           const char *arg);
```
dbnicholson, line 139: Misaligned.

src/libostree/libostree-released.sym
```diff
@@ -687,6 +687,7 @@ LIBOSTREE_2022.5 {
 global:
   ostree_kernel_args_append_if_missing;
   ostree_sysroot_deployment_set_kargs_in_place;
+  ostree_kernel_args_contains;
```
dbnicholson, line 690: This is wrong since 2022.5 is already released. You'll need to add a new section for `LIBOSTREE_2022.7` that starts from `LIBOSTREE_2022.5`. And actually that should be in `libostree-devel.sym` as noted in the comment below this.

src/libostree/ostree-kernel-args.c
```diff
@@ -829,3 +829,27 @@ ostree_kernel_args_append_if_missing (OstreeKernelArgs  *kargs,
...
+
+/**
+ * ostree_kernel_args_append_contains:
+ * @kargs: a OstreeKernelArgs instance
+ * @arg: key or key/value pair to check
+ *
+ * Search for @arg which is in the form of key=value pair at the hash table kargs->table
+ * and returns true if finds it.
+ *
+ * Since: 2022.5
```
dbnicholson, line 841: This should be 2022.7.

dbnicholson (changes requested):
Seems fine. It would be nice to implement `ostree_kernel_args_append_if_missing` with this instead of duplicating it. I imagine you're also going to hit a test error since the checksum for the symbol file will change.

src/libostree/ostree-kernel-args.c
```diff
@@ -829,3 +829,27 @@ ostree_kernel_args_append_if_missing (OstreeKernelArgs  *kargs,
...
   ostree_kernel_args_append (kargs, arg);
 }
+
+/**
+ * ostree_kernel_args_append_contains:
+ * @kargs: a OstreeKernelArgs instance
+ * @arg: key or key/value pair to check
+ *
+ * Search for @arg which is in the form of key=value pair at the hash table kargs->table
+ * and returns true if finds it.
```
dbnicholson, line 839: Also, use `%TRUE` and since this function returns a value this should probably have a line like:

```
Returns: %TRUE if @arg is contained in @kargs, %FALSE otherwise
```

src/libostree/ostree-kernel-args.c
```diff
@@ -829,3 +829,27 @@ ostree_kernel_args_append_if_missing (OstreeKernelArgs  *kargs,
  
   ostree_kernel_args_append (kargs, arg);
 }
+
+/**
+ * ostree_kernel_args_append_contains:
```
dbnicholson, line 834: Also, this is not correct.

Razaloc:
Thanks a lot for the review comments. What should I do about the checksum for the symbol file?
Do I have to change something in `tests/test-checksum.c` ?

dbnicholson:
> Thanks a lot for the review comments. What should I do about the checksum for the symbol file? Do I have to change something in `tests/test-checksum.c` ? Edit: I just added changes to `Makefile-libostree.am` but I don't think that's enough.

Actually, I think this now only happens on release commits. You can ignore that.

dbnicholson:
According to the [Jenkins build](https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/blue/organizations/jenkins/ostree/detail/PR-2771/6/pipeline), `ci/codestyle.sh` wants the rust declaration to look a little different:

```
 -    pub fn ostree_kernel_args_contains(kargs: *mut OstreeKernelArgs, arg: *const c_char) -> gboolean;
 +    pub fn ostree_kernel_args_contains(
 +        kargs: *mut OstreeKernelArgs,
 +        arg: *const c_char,
 +    ) -> gboolean;
```

Razaloc:
I'm not sure what the current jenkins error is about.

```
rofiles-fuse: error while loading shared libraries: libfuse3.so.3: cannot open shared object file: No such file or directory
Running pre scripts...done
error: Running %prein for libutempter: Creating bwrap instance
error: failed to execute cmd-build: exit status 1

script returned exit code 1
```

dbnicholson:
The jenkins build is very fragile. @cgwalters any ideas?

I think we can ignore it if it doesn't get straightened out as that error has nothing to do with this change.

Razaloc:
🎉

Issue #2750, opened 2022-10-27
rust-bindings/cargo: bump to gtk-rs 0.16


lucab:
Next iteration of the gtk-rs stack is out, see https://gtk-rs.org/blog/2022/10/18/new-release.html.
This effectively means bumping glib/gio/gobject to 0.16, with newer requirements on C native dependencies:
> GLib/GObject/GIO: minimum 2.56, and supports API up to version 2.74

There is also an hard requirement for a minimum Rust toolchain >= 1.63.
Holding this for now. We'd want to give some time for recent toolchains to spread around.
Also, we would like to have cap-std 1.0 available at the time we do this, so that we can do a single semver break for both at the same time. Tracking ticket for cap-std stabilization is https://github.com/bytecodealliance/cap-std/issues/272.

PR #2772, opened 2022-11-18
cargo: bump minimum Rust version to 1.63

lucab:
This bumps MSRV to 1.63, in order to prepare for the next version of gtk-rs stack.

Refs:
 * https://gtk-rs.org/blog/2022/10/18/new-release.html
 * https://github.com/ostreedev/ostree/issues/2750


cgwalters:
Hmm, not totally sure what's going on with CI.  It looks like the buildroot is f37 as expected and...ohhhh, I see, we're installing the ostree binaries on top of coreos-assembler too which is still f36, so we'd need https://github.com/coreos/coreos-assembler/pull/3212

Mmmph.  Well...we could I guess detect this situation and skip installing over cosa?

cgwalters (approved):

PR #2774, opened 2022-11-18
ci: Fix case where cosa != buildroot

cgwalters:
We can't override host binaries in this case.  We'll need to watch carefully and verify we're still testing things.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -34,6 +34,7 @@ stage("Build") {
         shwrap("""
           make install DESTDIR=\$(pwd)/installed/rootfs
           make -C tests/kolainst install DESTDIR=\$(pwd)/installed/tests
+          bash -c '. /usr/lib/os-release && echo \$VERSION_ID' >\$(pwd)/installed/buildroot-id
         """)
       }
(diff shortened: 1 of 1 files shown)


.cci.jenkinsfile
```diff
@@ -47,8 +48,12 @@ cosaPod(runAsUser: 0, memory: "9Gi", cpu: "4") {
     checkout scm
     unstash 'build'
     shwrap("""
-      # Move the bits into the cosa pod
-      rsync -rlv installed/rootfs/ /
+      # Move the bits into the cosa pod (but only if major versions match)
+      buildroot_id=\$(cat installed/.buildroot-id)
+      osver=(. /usr/lib/os-release && echo \$VERSION_ID)
```
jlebon, line 53: Shouldn't this be
```suggestion
      osver=\$(. /usr/lib/os-release && echo \$VERSION_ID)
```
?

jlebon (approved):
Yeah, I think we should probably have an `fcos-buildroot:cosa` tag. It'd require building the code twice though. We can probably add some coreos-ci-lib goop that incurs that cost only when they don't match.

PR #2777, opened 2022-11-18
Idempotent delete operation for  OstreeKernelArgs

Razaloc:
ostree_kernel_args_delete_if_present checks if an argument is present in OstreeKernelArgs and delete it.

#2329 

Signed-off-by: Rafael Garcia Ruiz <rafael.garcia@collabora.com>

### DIFF
--- a/src/libostree/ostree-kernel-args.c
+++ b/src/libostree/ostree-kernel-args.c
@@ -848,3 +848,25 @@ ostree_kernel_args_contains (OstreeKernelArgs  *kargs,
 
   return g_hash_table_contains (kargs->table, key);
 }
+
+/**
+ * ostree_kernel_args_delete_if_present:
+ * @kargs: a OstreeKernelArgs instance
+ * @arg: key or key/value pair to be deleted
+ * @error: an GError instance
+ *
+ * Deletes @arg which is in the form of key=value pair from the hash table kargs->table.
+ *
+ * Returns: %TRUE on success, %FALSE on failure
+ *
+ * Since: 2022.7
+ **/
+gboolean
+ostree_kernel_args_delete_if_present (OstreeKernelArgs  *kargs,
+                                      const char        *arg,
+                                      GError           **error)
+{
+  if (ostree_kernel_args_contains (kargs, arg))
+    return ostree_kernel_args_delete (kargs, arg, error);
+  return TRUE;
+}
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-kernel-args.c
```diff
@@ -848,3 +848,26 @@ ostree_kernel_args_contains (OstreeKernelArgs  *kargs,
...
+ * ostree_kernel_args_delete_if_present:
+ * @kargs: a OstreeKernelArgs instance
+ * @arg: key or key/value pair to be added
+ * @error: an GError instance
+ *
+ * Deletes @arg which is in the form of key=value pair from the hash table kargs->table.
+ *
+ * Returns: %TRUE on success, %FALSE on failure
+ *
+ * Since: 2022.5
```
dbnicholson, line 862: Should be 2022.7.

src/libostree/ostree-kernel-args.c
```diff
@@ -848,3 +848,26 @@ ostree_kernel_args_contains (OstreeKernelArgs  *kargs,
...
+ **/
+gboolean
+ostree_kernel_args_delete_if_present (OstreeKernelArgs  *kargs,
+                                      const char        *arg,
+                                      GError           **error)
+{
+  // Don't insert a duplicate key.
+  if (ostree_kernel_args_contains (kargs, arg))
+    ostree_kernel_args_delete (kargs, arg, error);
```
dbnicholson, line 871: The comment is wrong and this should return the value from `ostree_kernel_args_delete`. The other way to implement this would be to catch the error from `ostree_kernel_args_delete`, but it uses the generic `G_IO_ERROR_FAILED`, so you can't really tell the difference between non-existent key and some other error.

src/libostree/ostree-kernel-args.c
```diff
@@ -848,3 +848,26 @@ ostree_kernel_args_contains (OstreeKernelArgs  *kargs,
 
   return g_hash_table_contains (kargs->table, key);
 }
+
+/**
+ * ostree_kernel_args_delete_if_present:
+ * @kargs: a OstreeKernelArgs instance
+ * @arg: key or key/value pair to be added
```
dbnicholson, line 855: to be deleted

dbnicholson (changes requested):

dbnicholson (approved):

Issue #2769, opened 2022-11-16
thread 'main' panicked at 'failed to unpin deployment: Error { domain: g-io-error-quark, code: 21, message: "Writing out origin file: open(O_TMPFILE): Read-only file system" }', src/main.rs:15:14


juhp:
This is probably something elementary, but after upgrading to Fedora 37 Silverblue my tiny [ostree-pin](https://github.com/juhp/ostree-pin/blob/main/src/main.rs) utility started to panic:

```shellsession
$ RUST_BACKTRACE=1 ~/.cargo/bin/ostree-pin 
thread 'main' panicked at 'failed to unpin deployment: Error { domain: g-io-error-quark, code: 21, message: "Writing out origin file: open(O_TMPFILE): Read-only file system" }', src/main.rs:15:14
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: ostree_pin::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
Any ideas what might be wrong?  Basically it is crashing in `deployment_set_pinned()`.
(I am currently running Silverblue 37.20221116.0.)

I believe the code was working in SB36 (when I upgraded a couple of weeks ago) - so maybe something has changed?
I tried rebuilding with newer ostree/glib bindings but that didn't help.

Running `sudo ostree admin pin ...` by hands works fine.  I had a look at its C implementation but couldn't really see anything obvious stand out for me yet. Hopefully it is just something trivial I am missing.

cgwalters:
It's because https://fedoraproject.org/wiki/Changes/Silverblue_Kinoite_readonly_sysroot

So...we don't yet have an ergonomic way exposed in the Rust API to do the unshare bits.  Will work on it at some point.  Basically it should work if you do e.g.:

```
unshare -m /bin/sh -c 'mount -o remount,rw /sysroot && ~/.cargo/bin/ostree-pin`
```

juhp:
Ah I see thank you!  (I had missed that Change:)

juhp:
Thanks!

juhp:
Thanks again: I finally tried to use `ostree_sysroot_initialize_with_mount_namespace` in ostree-pin and it works nicely :pray:

PR #2779, opened 2022-11-21
sysroot: Add an API to initialize with mountns

cgwalters:
This lowers down into the C library some logic we
have in the binary/app logic, in prep for having more Rust-native CLI code in https://github.com/ostreedev/ostree-rs-ext/pull/412

Basically we want to *ensure* a mount namespace by invoking `unshare()` if necessary, instead of requiring our callers to do this dance.

This also helps fix e.g.
Closes: https://github.com/ostreedev/ostree/issues/2769

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -260,6 +260,63 @@ ostree_sysroot_set_mount_namespace_in_use (OstreeSysroot  *self)
 }
 
 /**
+ * ostree_sysroot_initialize_with_mount_namespace:
+ *
+ * Prepare the current process for modifying a booted sysroot, if applicable.
+ * This function subsumes the functionality of `ostree_sysroot_initialize`
+ * and may be invoked wherever that function is.
+ *
+ * If the sysroot does not appear to be booted, or where the current process is not uid 0,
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-sysroot.c
```diff
@@ -259,6 +259,63 @@ ostree_sysroot_set_mount_namespace_in_use (OstreeSysroot  *self)
...
+ * Otherwise, if the process is in the same mount namespace as pid 1, create
+ * a new namespace.
+ *
+ * If you invoke this function, it must be before ostree_sysroot_load(); it may
+ * be invoked before or after ostree_sysroot_initialize().
+ *
+ * Since: 2022.7
+ */
+gboolean
+ostree_sysroot_initialize_mount_namespace (OstreeSysroot *self, GError **error)
```
lucab, line 281: A more explicit `initialize_with_mount_namespace` maybe?

As we are adding a new API, do you want to add a `GCancellable` anyway? The `glnx` primitives would gladly take one (even if not really using it).

cgwalters: Done and done.

lucab (approved):

PR #2780, opened 2022-11-21
test-concurrency: Try to widen race conditions

cgwalters:
Right now we're spawning all the commit processes, then spawning all the pruners.  Change things to randomly interleave process execution to raise the likelihood of provoking bugs.

Further, randomize the number of commit and prune processes.


lucab:
@cgwalters do you want to merge this to `main` or just keep running it on the side until it triggers and exposes some more clues?

cgwalters:
Not sure.  Let's leave this open for a bit; we know we're seeing the test-concurrency failures in CI, and I don't want to unintentially *mask* them if the changes here somehow do so.

PR #2782, opened 2022-11-22
Testinst updates

cgwalters:
tests/inst: Update to latest ostree-ext

---

tests/inst: Drop nix dependency

It has various semver bumps and we were literally just using
it to duplicate what's already in `libc`.

---

tests/inst: Update cap-std-ext to 1.0

---

tests/inst: Update sh-inline

---


jmarrero (approved):
lgtm

PR #2789, opened 2022-11-24
configure: use pkg-config with newer gpgme and gpg-error

lucab:
This tweaks autoconf logic in order to use pkg-config for gpgme and gpg-error when available.
Recent versions of gpgme directly provide threaded support, and gpg-error started shipping a .pc file. Thus on recent distributions it is possible to directly use pkg-config for both. On older environments, the legacy logic is kept in place.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -2528,6 +2528,7 @@ out:
 #endif /* OSTREE_DISABLE_GPGME */
 }
 
+#ifndef OSTREE_DISABLE_GPGME
 static gboolean
 _ostree_repo_gpg_prepare_verifier (OstreeRepo         *self,
                                    const gchar        *remote_name,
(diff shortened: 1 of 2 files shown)


dbnicholson:
I didn't look closely, but I agree with this in principle. There's at least one open issue about this. I can take a closer look after the US holiday if no one gets to it earlier.

lucab:
This is meant to fix the recent CI failure on Debian testing, which I think is due to https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1023601.

src/libostree/ostree-repo.c
```diff
@@ -2528,6 +2528,7 @@ ostree_repo_remote_gpg_import (OstreeRepo         *self,
 #endif /* OSTREE_DISABLE_GPGME */
 }
 
+#ifndef OSTREE_DISABLE_GPGME
```
lucab, line 2531: Drive-by change: I was testing the `--with-gpme=no` path and there was an `unused-function` warning here.

dbnicholson: Could you sit that into a separate commit?

lucab: Yup, done!

cgwalters:
One nit from `maint.mk` which I'm honestly tempted to drop given how nit-picky it is:
> [2022-11-28T16:55:09.982Z] maint.mk: use "test C1 && test C2", not "test C1 -a C2"; use "test C1 || test C2", not "test C1 -o C2"

cgwalters (approved):

PR #2791, opened 2022-12-01
Enable `trust_return_value_nullability`

cgwalters:
Enable `trust_return_value_nullability`

Several commonly used APIs currently unnecessarily return
`Option<T>`, like `ostree_deployment_get_csum()`.

Flip the gir flag for this to on; I think our annotations are
correct.

---

Add docs and fix annotations for ostree-repo-file.c

The code here is not great, embarassing we've gone this
long without docs for some of these public API functions too.

I think this is right though.

---


cgwalters:
From chat on Matrix:

> yes, you need to set `(nullable)` on those
> I believe (not nullable) is useless and should be dropped as an annotation
the problem is a lot of APIs lack the nullable annotations, that is why the default is to assume everything returned by the C apis is nullable
so you have to carefully review all the API changes

So basically we need to go over the API surfaces affected here and choose whether they do need `nullable`.

rust-bindings/src/auto/bootconfig_parser.rs
```diff
@@ -26,14 +26,14 @@ impl BootconfigParser {
...
-    pub fn clone(&self) -> Option<BootconfigParser> {
+    pub fn clone(&self) -> BootconfigParser {
         unsafe {
             from_glib_full(ffi::ostree_bootconfig_parser_clone(self.to_glib_none().0))
         }
     }
 
     #[doc(alias = "ostree_bootconfig_parser_get")]
-    pub fn get(&self, key: &str) -> Option<glib::GString> {
+    pub fn get(&self, key: &str) -> glib::GString {
```
bilelmoussaoui, line 36: This should probably stay nullable and you can use something like `nullable_return_is_error = "The key doesn't exists"`

cgwalters:
OK rebased this on https://github.com/ostreedev/ostree/pull/2814 and lifting draft.

rust-bindings/src/auto/bootconfig_parser.rs
cgwalters: And thanks so much for spotting this!!

jlebon (approved):
Looks sane to me, though I didn't review the nullability of all the APIs.

jlebon:
What's the behaviour if an API we thought was not nullable actually is? Does the Rust code panic on encountering the null?

cgwalters:
> Does the Rust code panic on encountering the null?

Hmm I was going to say yes, but apparently the answer is no?  
https://docs.rs/glib/0.17.5/src/glib/shared.rs.html#550
That's not great...
I guess we could try to add `enable_trust_but_verify_return_value_nullability = true` to change `debug_assert!` -> `assert!`?

jlebon:
> > Does the Rust code panic on encountering the null?
> 
> Hmm I was going to say yes, but apparently the answer is no? [docs.rs/glib/0.17.5/src/glib/shared.rs.html#550](https://docs.rs/glib/0.17.5/src/glib/shared.rs.html#550) That's not great... I guess we could try to add `enable_trust_but_verify_return_value_nullability = true` to change `debug_assert!` -> `assert!`?

Could be interesting. I guess it depends how much we trust the current annotations to be correct? I think they should be too, but I haven't had to pay close attention to them before. I wouldn't be surprised if there's at least one or two that needs tweaking.

It doesn't look like there's *that many* APIs being changed, so I'm happy to split it up and sanity-check their nullability if you'd like. Otherwise, I'd vote to just ship this and do any fixups as follow-ups. At least it's early in the development cycle.

cgwalters:
OK.  I did a full audit and found a few more APIs that did need `nullable`.  I think this is now good to go!  It will be a nice ergonomic improvement in some cases, biggest one is `Sysroot::repo()` now correctly just giving `Repo`.

jlebon (approved):

cgwalters:
Yep this made things better https://github.com/ostreedev/ostree-rs-ext/pull/467

src/libostree/ostree-repo.c
```diff
@@ -982,7 +982,7 @@ _ostree_repo_remote_name_is_file (const char *remote_name)
  * underneath that group, or @default_value if the remote exists but not the
  * option name.  If an error is returned, @out_value will be set to %NULL.
  *
- * Returns: %TRUE on success, otherwise %FALSE with @error set
+ * Returns: (nullable): %TRUE on success, otherwise %FALSE with @error set
```
tintou, line 985: This can't be null as it is a gboolean

cgwalters: Right, I messed this up; PR incoming

cgwalters: https://github.com/ostreedev/ostree/pull/2838

PR #2792, opened 2022-12-02
lib/sysroot-upgrader: add some 'nullable' annotations

lucab:


cgwalters (approved):

cgwalters:
```
Run cd tests/inst && cargo build --verbose --release
    Updating crates.io index
    Updating git repository `[https://github.com/coreos/rpm-ostree`](https://github.com/coreos/rpm-ostree%60)
    Updating git submodule `[https://github.com/rpm-software-management/libdnf`](https://github.com/rpm-software-management/libdnf%60)
    Updating git submodule `[https://gitlab.gnome.org/GNOME/libglnx.git`](https://gitlab.gnome.org/GNOME/libglnx.git%60)
    Updating git repository `[https://github.com/cgwalters/with-procspawn-tempdir`](https://github.com/cgwalters/with-procspawn-tempdir%60)
error: failed to select a version for the requirement `hyper-staticfile = "^0.6.0"`
candidate versions found which didn't match: 0.9.2
location searched: crates.io index
required by package `ostree-test v0.1.0 (/__w/ostree/ostree/tests/inst)`
Error: Process completed with exit code 101.
```

Hum...it sure looks to me like the hyper-staticfile just yanked all prior versions.  That's rather unfriendly.

lucab:
@cgwalters it looks like the test crate here is an application, but it is lacking its lockfile. If you still have one around, you can commit it and the yanked library will work again. Otherwise we'll need to update the dependency first.

cgwalters:
True...we could think of it as an app.  That said, yanking IMO should only be used when there's a *reason* - I filed:
https://github.com/stephank/hyper-staticfile/issues/37

lucab:
Rebased.

PR #2793, opened 2022-12-13
bootloader: Add an aboot (Android) bootloader backend

ericcurtin:
aboot is special in that it packages kernel, initrd, cmdline dtb and signature one combined image (similar to upcoming unified kernel images). This is then loaded as an image into an aboot partition.

This image is signed by the OS vendor and covers everything in the image. So locally on the deployed system it should not be possible to boot an unsigned image (unless signature checking is turned off).

We call a shell script aboot-deploy when it is required to write a new image to the aboot partition (a file typically starting with aboot and ending in .img extension). This shell script may also read some configurations from a .cfg file.

Signed-off-by: Eric Curtin <ecurtin@redhat.com>

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -111,6 +111,8 @@ libostree_1_la_SOURCES = \
 	src/libostree/ostree-deployment.c \
 	src/libostree/ostree-bootloader.h \
 	src/libostree/ostree-bootloader.c \
+	src/libostree/ostree-bootloader-aboot.h \
+	src/libostree/ostree-bootloader-aboot.c \
 	src/libostree/ostree-bootloader-grub2.h \
 	src/libostree/ostree-bootloader-grub2.c \
 	src/libostree/ostree-bootloader-zipl.h \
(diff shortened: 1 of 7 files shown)


ericcurtin:
Related commit for aboot-deploy:

https://gitlab.com/CentOS/automotive/rpms/aboot-deploy/-/commit/e2acf0017fad1688792a012ae16b5d1842a6ed8c

ericcurtin:
Related PR for aboot-update:

https://gitlab.com/CentOS/automotive/rpms/aboot-update/-/merge_requests/1

which creates the image file to be written to the boot partition.

cgwalters (approved):
I wonder if it'd make sense to just call the thing to be booted `vmlinuz` or so?  But I guess there's probably precedent behind this `aboot.img` thing.

ericcurtin:
Thanks for looking at this @cgwalters we boot from an Android boot partition (which is unrecognizable in a normal Linux userspace, it's just data in a a partition), which is where the aboot.img file essentially gets written to. U-boot actually has some nice documentation on how it works, I test using that... Although this will really be loaded from the Qualcomm Android Bootloader in real life. u-boot has some documentation on this:

https://u-boot.readthedocs.io/en/latest/android/boot-image.html

PR #2799, opened 2023-01-03
Replace the radix64 crate with base64

jeamland:
The [radix64](https://crates.io/crates/radix64) crate was last updated over 3 years ago.

On the other hand the [base64](https://crates.io/crates/base64) crate appears to be far more actively maintained, supports all the needed features and has a few orders of magnitude more users.


cgwalters (approved):
Thanks!

PR #2807, opened 2023-01-24
tests/inst: Update hyper-staticfile to 0.9.4

dbnicholson:
Due to some vulnerabilities[1][2], all versions prior to 0.9.4 have been yanked from crates.io. There have been some API changes since 0.6.0, but none of them affect the simple `hyper_staticfile::Static` usage here.

1. https://rustsec.org/advisories/RUSTSEC-2022-0069.html
2. https://rustsec.org/advisories/RUSTSEC-2022-0072.html


cgwalters (approved):
Awesome!  Thank you so much!

PR #2808, opened 2023-01-30
ostree/prune: Calculate reachability under exclusive lock

jlebon:
When we calculate the reachability set in `ostree prune`, we do this without any locking. This means that between the time we build the set and when we call `ostree_repo_prune_from_reachable`, new content might've been added. This then causes us to immediately prune that content since it's not in the now outdated set.

Fix this by calculating the set under an exclusive lock.

I think this is what happened in
https://github.com/fedora-silverblue/issue-tracker/issues/405. While the pruner was running, the `new-updates-sync` script[1] was importing content into the repo. The newly imported commits were immediately deleted by the many `ostree prune --commit-only` calls the pruner does, breaking the refs.

[1] https://pagure.io/fedora-infra/ansible/blob/35b35127e444/f/roles/bodhi2/backend/files/new-updates-sync#_18

### DIFF
--- a/src/ostree/ot-builtin-prune.c
+++ b/src/ostree/ot-builtin-prune.c
@@ -208,6 +208,15 @@ ostree_builtin_prune (int argc, char **argv, OstreeCommandInvocation *invocation
     }
   else
     {
+      /* In this branch, we need to compute the reachability set manually.
+       * While we do this, we can't let new content in since it'll race with
+       * reachability calculations and we may immediately nuke it. So push an
+       * exclusive lock now. */
+      g_autoptr(OstreeRepoAutoLock) lock =
+        ostree_repo_auto_lock_push (repo, OSTREE_REPO_LOCK_EXCLUSIVE, cancellable, error);
+      if (!lock)
+        return FALSE;
+
       g_autoptr(GHashTable) all_refs = NULL;
       g_autoptr(GHashTable) reachable = ostree_repo_traverse_new_reachable ();
       g_autoptr(GHashTable) retain_branch_depth = g_hash_table_new_full (g_str_hash, g_str_equal, g_free, NULL);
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
Ah, yes.  Classic GC issue.  The obvious optimization here would be to skip pruning any commits created (filesystem timestamp I guess) after the scan started.

But for now, SGTM.

jlebon:
> Ah, yes. Classic GC issue. The obvious optimization here would be to skip pruning any commits created (filesystem timestamp I guess) after the scan started.
> 
> But for now, SGTM.

Good idea. Let me try this and see how hard it'd be. It's definitely unfortunate to be locking for the whole operation since it could take a long time to calculate reachability.

Hmm, though I guess the timestamp trick won't work if you're somehow using rsync for getting content into the repo and preserving timestamps (I think this has come up before IIRC). I wonder if what we actually want here is a new `OstreeRepoPruneOptions` field where the client also passes the list of objects to consider for pruning. So we'd list the objects, then calculate reachability, and then pass the objects and the set into the prune API.

src/libostree/ostree-repo.c
```diff
@@ -4975,8 +4975,19 @@ repo_list_objects_impl (OstreeRepo                  *self,
...
+/* ostree_repo_list_objects_set:
+ * @self: Repo
+ * @flags: Flags controlling enumeration
+ * @cancellable: Cancellable
+ * @error: Error
+ *
+ * This function is similar to ostree_repo_list_objects but returns a
+ * set and not a map (with a useless value).
+
+ * Returns: (transfer container) (element-type GVariant):
```
jlebon, line 4987: I did `transfer container` here to match `ostree_repo_list_objects` but I think both of them should probably be `transfer full` instead. But I guess it doesn't matter for `GHashTable` since the container is what owns the objects inside anyway.

jlebon:
Well this is awkward... `make gir` doesn't generate code that respects rustfmt. Fixed!

PR #2811, opened 2023-02-08
Various CLI improvements

dbnicholson:
Often I want to quickly inspect local and remote repos to see what's currently in them. While it's often better to write something that uses the API to get the precise desired information, the CLI is very handy for quickly assessing the situation. This adds some options to various CLI builtins to expose or improve that information.

### DIFF
--- a/src/ostree/ot-builtin-refs.c
+++ b/src/ostree/ot-builtin-refs.c
@@ -47,6 +47,16 @@ static GOptionEntry options[] = {
   { NULL }
 };
 
+static int
+collection_ref_cmp (OstreeCollectionRef *a,
+                    OstreeCollectionRef *b)
+{
+  int ret = g_strcmp0 (a->collection_id, b->collection_id);
+  if (ret == 0)
+    ret = g_strcmp0 (a->ref_name, b->ref_name);
+  return ret;
+}
+
 static gboolean
 do_ref_with_collections (OstreeRepo    *repo,
                          const char    *refspec_prefix,
@@ -66,10 +76,12 @@ do_ref_with_collections (OstreeRepo    *repo,
 
   if (!opt_delete && !opt_create)
     {
-      g_hash_table_iter_init (&hashiter, refs);
-      while (g_hash_table_iter_next (&hashiter, &hashkey, &hashvalue))
+      g_autoptr(GList) ordered_keys = g_hash_table_get_keys (refs);
+      ordered_keys = g_list_sort (ordered_keys, (GCompareFunc) collection_ref_cmp);
+
+      for (GList *iter = ordered_keys; iter != NULL; iter = iter->next)
         {
-          const OstreeCollectionRef *ref = hashkey;
+          OstreeCollectionRef *ref = iter->data;
           g_print ("(%s, %s)\n", ref->collection_id, ref->ref_name);
         }
     }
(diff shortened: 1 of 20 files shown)


src/ostree/ot-builtin-refs.c
```diff
@@ -203,6 +214,11 @@ static gboolean do_ref (OstreeRepo *repo, const char *refspec_prefix, GCancellab
               const char *alias = g_hash_table_lookup (refs, ref);
               g_print ("%s -> %s\n", ref, alias);
```
jlebon, line 215: Hmm, maybe in `--revision` mode we should just print aliases like regular refs. Otherwise, it throws a wrench in the machine-friendliness of this output.

jlebon (approved):
Nice work!

src/ostree/ot-builtin-refs.c
jlebon: Ahh gotcha, I mistakenly thought the `->` output could occur mixed with revision output.

> Would you like me to make it an error if they're combined?

That makes sense to me, though I'm not too worried about that corner case either.

jlebon (approved):
LGTM. Thanks for the well split-out patches!

Restarted CI. Will let @cgwalters have a look and merge.

src/ostree/ot-dump.c
```diff
@@ -407,6 +407,66 @@ ot_dump_summary_bytes (GBytes          *summary_bytes,
...
+gboolean
+ot_dump_summary_metadata_key (GBytes      *summary_bytes,
+                              const char  *key,
+                              GError     **error)
+{
+  g_autoptr(GVariant) summary = NULL;
+  g_autoptr(GVariant) metadata = NULL;
+  g_autoptr(GVariant) value = NULL;
+
+  summary = g_variant_new_from_bytes (OSTREE_SUMMARY_GVARIANT_FORMAT,
```
cgwalters, line 454: This is all fine as is but personally I've been trying to convert the whole codebase to declare-and-initialize because it's simply shorter, also matches how e.g. Rust works, and works better with `goto out` and not having implicitly nullable pointers.

I imagine you work on other codebases that still use C89 separation between declaration and init, but hopefully we can agree to keep this one using declare-and-initialize.

cgwalters (approved):

cgwalters:
Well, something clearly broke WRT our space checks, but it's also not related to this PR

src/ostree/ot-dump.c
dbnicholson: That's fair. I split it for 2 reasons.

One is that I personally think sometimes it gets pretty busy with the initializer on the same line with an autoptr declaration. The other is that I have had gcc whine about using using uninitialized variables before when autoptr variables aren't explicitly initialized to `NULL`. Which seems like a compiler bug but there is some logic to it in certain scenarios.

Anyways, I'll be sure to do the full initialization next time.

cgwalters: > The other is that I have had gcc whine about using using uninitialized variables before when autoptr variables aren't explicitly initialized to NULL. 

I'm talking about declare-and-initialize, so there's no point at which it's set to `NULL`.

dbnicholson: I hope I'm remembering this right, but I am talking about declare-and-initialize of a `g_autoptr`. I believe what happened is that the compiler decided that the initialization code could be run later, but then the `__attribute__((cleanup(func)))` code would warn that it was potentially using an uninitialized variable. As as said, likely a gcc bug since the presence of the cleanup attribute means it shouldn't split the initializer. I've only seen it a few times, though.

PR #2814, opened 2023-02-09
libostree: Enhance the annotation coverage

tintou:
Increase the coverage of the GObject Introspection annotation and most notably replace the use of "allow-none" with "out" parameters to "nullable" or "optional" as the previous annotation is deprecated and ambiguous.

### DIFF
--- a/src/libostree/ostree-bootconfig-parser.c
+++ b/src/libostree/ostree-bootconfig-parser.c
@@ -134,6 +134,14 @@ ostree_bootconfig_parser_parse (OstreeBootconfigParser  *self,
                                             cancellable, error);
 }
 
+/**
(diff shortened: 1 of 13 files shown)


cgwalters:
Thanks for the patch!  Some overlap here with https://github.com/ostreedev/ostree/pull/2791 right?

tintou:
This would probably help this PR indeed

cgwalters:
OK before we do another semver bump I'd like to get a release out with current fixes for 0.17 https://github.com/ostreedev/ostree/pull/2818

tintou:
I guess that this can be merged now?

cgwalters (approved):
Thanks!

PR #2821, opened 2023-02-27
tests: Ensure non-root users have access to libcap tools

smcv:
On Debian systems, by default only root has /{usr/,}sbin in PATH.

---

Similar to containers/bubblewrap#228. This should give us test coverage for rofiles-fuse on Ubuntu's autopkgtest infrastructure; previously that wasn't tested.

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -32,6 +32,9 @@ else
 fi
 . ${test_srcdir}/libtest-core.sh
 
+# Make sure /sbin/capsh etc. are in our PATH even if non-root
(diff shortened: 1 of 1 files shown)


smcv:
I'll undraft this when I've put it through a full set of Debian builds (which takes a while), but I think it's right.

smcv:
Yes, this seems to be enough to get the test to pass in `autopkgtest-virt-qemu`, while still being skipped under `autopkgtest-virt-lxc` (which doesn't give us `/dev/fuse`).

PR #2824, opened 2023-03-04
rust-bindings: Regenerate for 2023.1 && rust-bindings: `Sysroot` is `Send`

cgwalters:
rust-bindings: Regenerate for 2023.1

Still need to automate this...

---

rust-bindings: `Sysroot` is `Send`

I want to spawn tokio worker threads referencing sysroot objects
in bootc.

Just like the repo, there's nothing thread-local about it.

---


jmarrero (approved):
lgtm

PR #2831, opened 2023-03-14
Increase buffer size for create_regular_tmpfile_linkable_with_content

nanonyme:
The small buffer size results in really bad performance under any FUSE-based filesystems with round-trips.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -678,12 +678,13 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
        *    e.g. 10 bytes but is actually gigabytes.
        *  - Due to GLib bugs that pointlessly calls `poll()` on the output fd for every write
        */
-      char buf[8192];
+      gsize buf_size = MIN(length, 1048576);
+      g_autofree gchar * buf = g_malloc(buf_size);
       guint64 remaining = length;
       while (remaining > 0)
         {
           const gssize bytes_read =
-            g_input_stream_read (input, buf, MIN (remaining, sizeof (buf)), cancellable, error);
+            g_input_stream_read (input, buf, MIN (remaining, buf_size), cancellable, error);
           if (bytes_read < 0)
             return FALSE;
           else if (bytes_read == 0)


cgwalters:
Hmm...I wonder if higher level code should instead be creating a https://developer-old.gnome.org/gio/stable/GBufferedInputStream.html ?

What code paths were you seeing for this?  What is the FUSE filesystem involved here?

src/libostree/ostree-repo-commit.c
```diff
@@ -678,12 +678,13 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self,
        *    e.g. 10 bytes but is actually gigabytes.
        *  - Due to GLib bugs that pointlessly calls `poll()` on the output fd for every write
        */
-      char buf[8192];
+      gsize buf_size = 1048576;
```
cgwalters, line 681: Might as well `gsize buf_size = MIN(length, 1048576);`

nanonyme:
We're using buildbox-fuse as part of our build stack and running ostree inside it. We noticed the small reads and writes and polls result in very high overhead. It seems this is negligible with regular filesystems. Using larger buffer works consistently on both. This resulted in ballpark doubling filesystem performance.

src/libostree/ostree-repo-commit.c
nanonyme: I guess that would be fine. As per I'm reading documentation right, with zero-sized files buf would then be NULL and the loop would abort right away.

nanonyme:
> Hmm...I wonder if higher level code should instead be creating a https://developer-old.gnome.org/gio/stable/GBufferedInputStream.html ?

It's possible this could also use higher-level primitive. I was just trying to avoid larger code changes. The conversion to g_malloc was largely to avoid allocating larger amounts of memory from stack when increasing buffer size.

nanonyme:
Ah yes, so the codepath was what flatpak build-export calls on ostree. I can't recall specific call anymore but we traced the hot path to this code segment a couple of months back.

dbnicholson:
FWIW, I recall reading a comment from a kernel developer a few years ago that using a classic kB sized buffer is a mistake. He suggested a much bigger buffer like 10 or even 50 MB would have much better performance with no downside.

nanonyme:
I have no problem increasing it further. The bump to 1MB is tested to improve performance and I would like a bump to at least this. It's quite trivial to change the number after switching to g_malloc, it's quite literally just a number.

cgwalters (approved):

PR #2832, opened 2023-03-14
itest-pull-space: Use mkfs.ext4, align to at least 512b

cgwalters:
XFS now seems to want filesystems larger than 300MB, so switch to ext4.  Also use `20MiB` so we align to 512b sectors to squash a `losetup` warning.
Also tweak some of the numbers to still work.


nanonyme:
Doesn't this use dramatically different codepaths in ostree? XFS supports reflinks while Ext4 does not.

jlebon (approved):

PR #2834, opened 2023-03-16
core: Ensure glib standard::size attribute is always set

evan-a-a:
A recent [change in glib](https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3261) requires that the appropriate attribute be available when calling getters. ostree core only sets this attribute on regular files, and frequently triggers the critical warning. Solve this by setting standard::size to zero for non-regular files.

Fixes https://github.com/ostreedev/ostree/issues/2827


cgwalters:
Thanks so much @evan-a-a !
/ok-to-test

cgwalters (approved):

PR #2835, opened 2023-03-16
tests: Set size on fileinfo

cgwalters:
Closes: https://github.com/ostreedev/ostree/issues/2827

PR #2836, opened 2023-03-17
Cope with GLib 2.76 being more strict about GFileInfo standard::size

smcv:
* ostree_raw_file_to_content_stream: Make size default to 0
    
    Some existing code calls into ostree_raw_file_to_content_stream() with
    file objects that do not have the standard::size attribute. Since GLib
    2.76.0, attempting to access the size of such an object raises a
    critical warning. Handle this more gracefully by defaulting the size
    to 0, like earlier versions of GLib did.

* _ostree_zlib_file_header_new: Default size to 0
    
    Similar to the previous commit, but for
    ostree_raw_file_to_archive_z2_stream() and similar public APIs.

* write_content_object: Don't assume file info has standard::size
    
    The file info object for symlinks might validly not have this attribute.
    If not, behave as though it was 0, matching what happened with older
    versions of GLib.

---

The first commit avoids Flatpak test failures, and is an alternative to #2835 and flatpak/flatpak#5354 (although I think it might be best to address this both from the Flatpak side and the libostree side).

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -550,7 +550,9 @@ ostree_raw_file_to_content_stream (GInputStream       *input,
   g_autoptr(GBytes) file_header = _ostree_file_header_new (file_info, xattrs);
   *out_input = header_and_input_to_stream (file_header, input);
   if (out_length)
-    *out_length = g_bytes_get_size (file_header) + g_file_info_get_size (file_info);
+    *out_length = g_bytes_get_size (file_header);
+  if (out_length && g_file_info_has_attribute (file_info, "standard::size"))
+    *out_length += g_file_info_get_size (file_info);
   return TRUE;
 }
 
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -355,7 +355,7 @@ GBytes *
 _ostree_zlib_file_header_new (GFileInfo         *file_info,
                               GVariant          *xattrs)
 {
-  guint64 size = g_file_info_get_size (file_info);
+  guint64 size = 0;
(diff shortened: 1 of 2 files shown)


smcv:
I think the underlying problem here is that libostree never really documented its assumptions about the GFileInfo objects that are passed in to API entry points like ostree_raw_file_to_content_stream(), so it's unclear whether this is a Flatpak bug (because libostree always assumed Flatpak was responsible for populating the size field), or a libostree bug (because libostree was meant to be robust against incomplete GFileInfo).

cgwalters (approved):
LGTM, thanks so much!

cgwalters:
> I think the underlying problem here is that libostree never really documented its assumptions about the GFileInfo objects that are passed in to API entry points like ostree_raw_file_to_content_stream(), 

Agree.  

> or a libostree bug (because libostree was meant to be robust against incomplete GFileInfo).

Ultimately I think glib decided to make this a bug for everyone who was using it...and it probably makes sense to fix *both* ostree and flatpak.

smcv:
I've only dealt with the most obvious cases here, and there might be others.

If an API user submits a regular file accompanied by a GFileInfo with no size, that's probably going to trigger GLib warnings. Perhaps libostree should document that as just not valid, and throw its own criticals?

Similarly if an API user provides a GFileInfo with no file type, uid, gid or mode, I think libostree should just document that as not valid, and throw criticals.

cgwalters:
Agree that it makes sense to document some of the expectations here.  But let's get this patch in; that can be a separate PR.

I also want to make APIs that don't involve `GFileInfo` but just perhaps directly take the required parameters instead, because honestly `GFileInfo` being a whole GObject is inefficient when we just need a subset of `struct stat` which could just be stack allocated by callers.

PR #2837, opened 2023-03-17
Use g_steal_fd() in preference to glnx_steal_fd()

smcv:
Based on #2822.

* Use g_steal_fd() in preference to glnx_steal_fd()
    
    g_steal_fd() exists in GLib since 2.70, and libglnx has a backport for
    older GLib versions, equivalent to the libglnx-specific glnx_steal_fd().

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -687,7 +687,7 @@ ostree_content_file_parse_at (gboolean                compressed,
   if (!glnx_fstat (fd, &stbuf, error))
     return FALSE;
 
-  g_autoptr(GInputStream) file_input = g_unix_input_stream_new (glnx_steal_fd (&fd), TRUE);
+  g_autoptr(GInputStream) file_input = g_unix_input_stream_new (g_steal_fd (&fd), TRUE);
 
   g_autoptr(GFileInfo) ret_file_info = NULL;
   g_autoptr(GVariant) ret_xattrs = NULL;
@@ -1038,7 +1038,7 @@ ostree_checksum_file_at (int               dfd,
       glnx_autofd int fd = -1;
       if (!glnx_openat_rdonly (dfd, path, FALSE, &fd, error))
         return FALSE;
-      in = g_unix_input_stream_new (glnx_steal_fd (&fd), TRUE);
+      in = g_unix_input_stream_new (g_steal_fd (&fd), TRUE);
       if (canonicalize_perms)
         {
           g_file_info_set_attribute_uint32 (file_info, "unix::uid", 0);
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -1031,7 +1031,7 @@ content_fetch_on_complete (GObject        *object,
       if (!glnx_fstat (tmpf.fd, &stbuf, error))
         goto out;
       /* Non-mirroring path */
-      tmpf_input = g_unix_input_stream_new (glnx_steal_fd (&tmpf.fd), TRUE);
+      tmpf_input = g_unix_input_stream_new (g_steal_fd (&tmpf.fd), TRUE);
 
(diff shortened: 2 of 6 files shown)


cgwalters (approved):

dbnicholson:
@cgwalters can we give @smcv commit access (if he wants)? As the flatpak maintainer he's exercising ostree as much as anyone. Not to mention that he's made several valuable contributions over the years, has shown himself to be a responsible developer in several projects, and maintains the debian ostree package. I don't think he needs someone to approve his PRs to be tested.

cgwalters:
Yep, totally agree!  You have the power to do that right?  Please do if so!

dbnicholson:
> Yep, totally agree! You have the power to do that right? Please do if so!

Oh, I didn't realize that.

I have the power!

@smcv interested? At least I think it would be nice if you were a member of the ostreedev org so you don't need approval to test PRs. I sent you an invite.

smcv:
I don't mind being a member of the org, but I'm sorry, I'm not going to subscribe to ostree's issue tracker or consider myself to be a maintainer. I'm already responsible for too many projects, and I don't want to end up being the maintainer of last resort for this one too.

(If I remember correctly, I said similar things about bubblewrap, for which I now seem to be the single point of failure for releases, PR review, triaging, fielding issue reports from people who demand that bubblewrap should make incompatible changes in order to be impossibly secure by default, ... - please help if you can, because I'm doing my best but my best is demonstrably not good enough.)

dbnicholson:
> I don't mind being a member of the org, but I'm sorry, I'm not going to subscribe to ostree's issue tracker or consider myself to be a maintainer. I'm already responsible for too many projects, and I don't want to end up being the maintainer of last resort for this one too.

For sure. I didn't intend to make you a maintainer that's on the hook for the health of this project. I was thinking more in the spirit of contributor where you can make or approve changes as you need or want. There are several of those people here. Mostly I want to reduce barriers to your contributions since they're valuable. No problem if that's not something you're interested in, though.

cgwalters:
> (If I remember correctly, I said similar things about bubblewrap, for which I now seem to be the single point of failure for releases, PR review, triaging, fielding issue reports from people who demand that bubblewrap should make incompatible changes in order to be impossibly secure by default, ... - please help if you can, because I'm doing my best but my best is demonstrably not good enough.)

Yeah...sorry...

Anyways, I think I'd summarize this as:

If a person is contributing PRs on an even irregular basis and we expect that to continue, then I'd tend to grant review/approve access since one test suite requires it (probably unnecessarily) but also indeed with the hope that additionally, that person may review other PRs.  But obviously, it's not required nor could anyone somehow enforce that.

smcv:
> We no longer build on CentOS Stream 9 anymore

Any file that is calling `g_steal_fd()` and wants to be compilable on an old LTS operating system should `#include <libglnx.h>`. Probably the error you quoted means `src/libotutil/ot-fs-utils.c` should do the same.

(Also, if you want the latest libostree to compile successfully on an LTS operating system, please consider contributing a github workflow that compiles on that operating system, so that PRs breaking that won't be accepted.)

ericcurtin:
> > We no longer build on CentOS Stream 9 anymore
> 
> Any file that is calling `g_steal_fd()` and wants to be compilable on an old LTS operating system should `#include <libglnx.h>`. Probably the error you quoted means `src/libotutil/ot-fs-utils.c` should do the same.

I apologise I just needed to do a:

git submodule update --init

I was just out of date I think.

> 
> (Also, if you want the latest libostree to compile successfully on an LTS operating system, please consider contributing a github workflow that compiles on that operating system, so that PRs breaking that won't be accepted.)

This makes sense, I will look into this if we encounter an LTS breakage in future and we still need CentOS Steam 9 to be close to upstream at that point. I have the same policy in inotify-tools, please contribute your workflow to avoid breakages.

PR #2838, opened 2023-03-17
More binding fixes

cgwalters:
Revert "repo: Fix nullability for remote options"

This reverts commit c4d03d28500c75ad2bcf7b75383fc60fe7b83ca8.

---

rust-bindings: Regenerate

This picks up more of tintou's recent changes.

---


cgwalters:
To elaborate on 
> Revert "repo: Fix nullability for remote options"

I just got confused into thinking the output value was a return value.  But my comment still stands; I think I'm going to do a separate PR which implements this manually and just drops `default_value` from the API because it's not useful in Rust, where we have convenient `unwrap_or` etc. combinators.

rust-bindings/sys/src/lib.rs
```diff
@@ -3545,38 +3443,6 @@ extern "C" {
         out_success_message: *mut *mut c_char,
         error: *mut *mut glib::GError,
     ) -> gboolean;
-    pub fn ostree_sign_dummy_add_pk(
```
cgwalters, line 3548: And it's also good for these to go away.

rust-bindings/sys/src/lib.rs
```diff
@@ -309,62 +309,6 @@ impl ::std::fmt::Debug for OstreeAsyncProgressClass {
     }
 }
 
-#[repr(C)]
-pub struct _OstreeBootloader {
-    _data: [u8; 0],
-    _marker: core::marker::PhantomData<(*mut u8, core::marker::PhantomPinned)>,
-}
-
-pub type OstreeBootloader = *mut _OstreeBootloader;
```
cgwalters, line 318: BTW @tintou this looks good - we were leaking these pointless private structs into the API before.

rust-bindings/src/auto/functions.rs
```diff
@@ -82,7 +82,7 @@ pub fn commit_get_object_sizes(commit_variant: &glib::Variant) -> Result<Vec<Com
 }
 
 #[doc(alias = "ostree_commit_get_parent")]
-pub fn commit_get_parent(commit_variant: &glib::Variant) -> glib::GString {
+pub fn commit_get_parent(commit_variant: &glib::Variant) -> Option<glib::GString> {
```
cgwalters, line 85: Verified we want this API change, and the other ones here too.

I think I accidentally dropped a commit to regenerate the Rust bindings in my previous PR.

cgwalters:
```
[2023-03-17T16:33:59.814Z] Mar 17 16:33:57 qemu0 kola-runext-destructive::itest_transactionality[1758]: Generating update from ecb557526ba34ae4fd24307a29107dc4019b0ad8614104d19bbf6c8da67f5f89
[2023-03-17T16:33:59.814Z] Mar 17 16:33:57 qemu0 kola-runext-destructive::itest_transactionality[1758]: thread 'tokio-runtime-worker' panicked at 'assertion failed: mutated > 0', src/treegen.rs:138:5
[2023-03-17T16:33:59.814Z] Mar 17 16:33:57 qemu0 kola-runext-destructive::itest_transactionality[1758]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Hmm, looks like an unrelated race; retest passed.  May dig into this at some point.

jmarrero (approved):
lgtm

Issue #2570, opened 2022-03-24
"Failure when receiving data from from the peer" is not retried


champtar:
Doing kickstart install from an Alma 8.5 iso / Alma 8.5 based tree ( libostree/2021.3 |  libcurl/7.61.1),
pulling content using http2 I'm getting `Failure when receiving data from from the peer`.
Disabling http2 make the install succeed, so root cause is likely a curl bug, but we might want to consider retrying more errors in _ostree_fetcher_should_retry_request

cgwalters:
OK right, this also got filed as https://issues.redhat.com/browse/RHELPLAN-146197

The relevant code here is in `_ostree_fetcher_should_retry_request`.

I think what we may be missing here is a mapping from libcurl errors to GError.

Today we do:

```
      else if (curlres != CURLE_OK)
        {
          if (is_file && curlres == CURLE_FILE_COULDNT_READ_FILE)
            {
              /* Handle file not found */
              g_task_return_new_error (task, G_IO_ERROR, G_IO_ERROR_NOT_FOUND,
                                       "%s", curl_easy_strerror (curlres));
            }
          else
            {
              g_task_return_new_error (task, G_IO_ERROR, G_IO_ERROR_FAILED,
```

But...arguably perhaps we should just map all libcurl errors here to `G_IO_ERROR_TIMED_OUT` - this would force a retry.

PR #2843, opened 2023-04-03
ostree-fetcher-curl: handle non 404 errors as G_IO_ERROR_TIMED_OUT

jmarrero:
This introduces the retry-all-network-errors option which
is enabled by default. This is a behavior change as now
ostree will retry on requests that fail except when
they fail with NOT_FOUND. It also introduces the options
low-speed-limit-bytes and low-speed-time-seconds these
map to CURL options only at the moment. Which have defaults
set following librepo:
https://github.com/rpm-software-management/librepo/blob/7c9af219abd49f8961542b7622fc82cfdaa572e3/librepo/handle.h#L90
https://github.com/rpm-software-management/librepo/blob/7c9af219abd49f8961542b7622fc82cfdaa572e3/librepo/handle.h#L96
Currently this changes only apply when using libcurl.

Closes: https://github.com/ostreedev/ostree/issues/2570

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -79,6 +79,10 @@ struct OstreeFetcher
   int tmpdir_dfd;
   bool force_anonymous;
   char *custom_user_agent;
+  guint32 opt_low_speed_limit;
+  guint32 opt_low_speed_time;
+  gboolean opt_retry_all;
+  guint32 opt_max_outstanding_fetcher_requests;
 
   GMainContext *mainctx;
   CURLM *multi;
@@ -330,7 +334,13 @@ check_multi_info (OstreeFetcher *fetcher)
             }
           else
             {
-              g_task_return_new_error (task, G_IO_ERROR, G_IO_ERROR_FAILED,
+              /* When it is not a file, we want to retry the request.
+               * We accomplish that by using G_IO_ERROR_TIMED_OUT.
+               */
+              gboolean opt_retry_all = req->fetcher->opt_retry_all;
+              int g_io_error_code
+                  = (is_file || !opt_retry_all) ? G_IO_ERROR_FAILED : G_IO_ERROR_TIMED_OUT;
+              g_task_return_new_error (task, G_IO_ERROR, g_io_error_code,
                                        "While fetching %s: [%u] %s", eff_url, curlres,
                                        curl_easy_strerror (curlres));
               _ostree_fetcher_journal_failure (req->fetcher->remote_name, eff_url,
@@ -665,6 +675,31 @@ _ostree_fetcher_set_proxy (OstreeFetcher *self, const char *http_proxy)
 }
 
 void
+_ostree_fetcher_set_low_speed_time (OstreeFetcher *self, guint32 opt_low_speed_time)
+{
+  self->opt_low_speed_time = opt_low_speed_time;
+}
+
+void
+_ostree_fetcher_set_low_speed_limit (OstreeFetcher *self, guint32 opt_low_speed_limit)
+{
+  self->opt_low_speed_limit = opt_low_speed_limit;
+}
+
+void
+_ostree_fetcher_set_retry_all (OstreeFetcher *self, gboolean opt_retry_all)
+{
+  self->opt_retry_all = opt_retry_all;
+}
+
+void
+_ostree_fetcher_set_max_outstanding_fetcher_requests (OstreeFetcher *self,
+                                                      guint32 opt_max_outstanding_fetcher_requests)
+{
+  self->opt_max_outstanding_fetcher_requests = opt_max_outstanding_fetcher_requests;
+}
+
+void
 _ostree_fetcher_set_cookie_jar (OstreeFetcher *self, const char *jar_path)
 {
   g_free (self->cookie_jar_path);
@@ -912,14 +947,10 @@ initiate_next_curl_request (FetcherRequest *req, GTask *task)
   g_assert_cmpint (rc, ==, CURLM_OK);
   rc = curl_easy_setopt (req->easy, CURLOPT_CONNECTTIMEOUT, 30L);
   g_assert_cmpint (rc, ==, CURLM_OK);
-  /* We used to set CURLOPT_LOW_SPEED_LIMIT and CURLOPT_LOW_SPEED_TIME
-   * here, but see https://github.com/ostreedev/ostree/issues/878#issuecomment-347228854
-   * basically those options don't play well with HTTP2 at the moment
-   * where we can have lots of outstanding requests.  Further,
-   * we could implement that functionality at a higher level
-   * more consistently too.
-   */
-
+  rc = curl_easy_setopt (req->easy, CURLOPT_LOW_SPEED_LIMIT, req->fetcher->opt_low_speed_limit);
+  g_assert_cmpint (rc, ==, CURLM_OK);
+  rc = curl_easy_setopt (req->easy, CURLOPT_LOW_SPEED_TIME, req->fetcher->opt_low_speed_time);
+  g_assert_cmpint (rc, ==, CURLM_OK);
   /* closure bindings -> task */
   rc = curl_easy_setopt (req->easy, CURLOPT_PRIVATE, task);
   g_assert_cmpint (rc, ==, CURLM_OK);
--- a/src/libostree/ostree-fetcher-soup.c
+++ b/src/libostree/ostree-fetcher-soup.c
@@ -75,6 +75,8 @@ typedef struct
   /* Also protected by output_stream_set_lock. */
   guint64 total_downloaded;
 
+  guint32 opt_max_outstanding_fetcher_requests;
+
   GError *oob_error;
 
 } ThreadClosure;
@@ -361,6 +363,12 @@ session_thread_set_tls_database_cb (ThreadClosure *thread_closure, gpointer data
 }
 
 static void
+session_thread_set_max_outstanding_fetcher_requests (ThreadClosure *thread_closure, gpointer data)
+{
+  thread_closure->opt_max_outstanding_fetcher_requests = GPOINTER_TO_UINT (data);
+}
+
+static void
 session_thread_set_extra_user_agent_cb (ThreadClosure *thread_closure, gpointer data)
 {
   const char *extra_user_agent = data;
@@ -375,6 +383,33 @@ session_thread_set_extra_user_agent_cb (ThreadClosure *thread_closure, gpointer 
       g_object_set (thread_closure->session, SOUP_SESSION_USER_AGENT,
                     OSTREE_FETCHER_USERAGENT_STRING, NULL);
     }
+}
+
+void
+_ostree_fetcher_set_low_speed_time (OstreeFetcher *self, guint32 opt_low_speed_time)
+{
+  // TODO
+}
+
+void
+_ostree_fetcher_set_low_speed_limit (OstreeFetcher *self, guint32 opt_low_speed_limit)
+{
+  // TODO
+}
+
+void
+_ostree_fetcher_set_retry_all (OstreeFetcher *self, gboolean opt_retry_all)
+{
+  // TODO
+}
+
+void
+_ostree_fetcher_set_max_outstanding_fetcher_requests (OstreeFetcher *self,
+                                                      guint32 opt_max_outstanding_fetcher_requests)
+{
+  session_thread_idle_add (self->thread_closure,
+                           session_thread_set_max_outstanding_fetcher_requests,
+                           GUINT_TO_POINTER (opt_max_outstanding_fetcher_requests), NULL);
 }
 
 static void on_request_sent (GObject *object, GAsyncResult *result, gpointer user_data);
@@ -511,7 +546,7 @@ ostree_fetcher_session_thread (gpointer data)
    * by spreading requests across mirrors. */
   gint max_conns;
   g_object_get (closure->session, "max-conns-per-host", &max_conns, NULL);
-  if (max_conns < _OSTREE_MAX_OUTSTANDING_FETCHER_REQUESTS)
+  if (max_conns < closure->opt_max_outstanding_fetcher_requests)
     {
       /* We download a lot of small objects in ostree, so this
        * helps a lot.  Also matches what most modern browsers do.
@@ -522,7 +557,7 @@ ostree_fetcher_session_thread (gpointer data)
        * currently conservative #define SOUP_SESSION_MAX_CONNS_PER_HOST_DEFAULT 2 (as of
        * 2018-02-14).
        */
-      max_conns = _OSTREE_MAX_OUTSTANDING_FETCHER_REQUESTS;
+      max_conns = closure->opt_max_outstanding_fetcher_requests;
       g_object_set (closure->session, "max-conns-per-host", max_conns, NULL);
     }
 
--- a/src/libostree/ostree-fetcher.h
+++ b/src/libostree/ostree-fetcher.h
@@ -99,6 +99,16 @@ void _ostree_fetcher_set_proxy (OstreeFetcher *fetcher, const char *proxy);
 void _ostree_fetcher_set_client_cert (OstreeFetcher *fetcher, const char *cert_path,
                                       const char *key_path);
 
+void _ostree_fetcher_set_low_speed_limit (OstreeFetcher *self, guint32 opt_low_speed_limit);
+
+void _ostree_fetcher_set_low_speed_time (OstreeFetcher *self, guint32 opt_low_speed_time);
+
+void _ostree_fetcher_set_retry_all (OstreeFetcher *self, gboolean opt_retry_all);
+
+void
+_ostree_fetcher_set_max_outstanding_fetcher_requests (OstreeFetcher *self,
+                                                      guint32 opt_max_outstanding_fetcher_requests);
+
 void _ostree_fetcher_set_tls_database (OstreeFetcher *self, const char *tlsdb_path);
 
 void _ostree_fetcher_set_extra_headers (OstreeFetcher *self, GVariant *extra_headers);
--- a/src/libostree/ostree-repo-pull-private.h
+++ b/src/libostree/ostree-repo-pull-private.h
@@ -54,6 +54,10 @@ typedef struct
 
   GVariant *extra_headers;
   char *append_user_agent;
+  guint32 low_speed_limit;
+  guint32 low_speed_time;
+  gboolean retry_all;
+  guint32 max_outstanding_fetcher_requests;
 
   gboolean dry_run;
   gboolean dry_run_emitted_progress;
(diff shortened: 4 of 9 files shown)


cgwalters:
See the context in this function; this code is only run when libcurl fails for a non-HTTP reason (e.g. TCP level errors).  We already have specific error handling for HTTP codes just below this.

jmarrero:
I likely misguided Dan which my commit message talking 404s and http. Thanks for clarifying, will fix the commit msg.

cgwalters (approved):
LGTM

src/libostree/ostree-fetcher-curl.c
```diff
@@ -329,7 +329,11 @@ check_multi_info (OstreeFetcher *fetcher)
             }
           else
             {
-              g_task_return_new_error (task, G_IO_ERROR, G_IO_ERROR_FAILED,
+              /* When it is not a file, we want to retry the request.
+               * We accomplish that by using G_IO_ERROR_TIMED_OUT.
+               */
+              int g_io_error_code = is_file ? G_IO_ERROR_FAILED : G_IO_ERROR_TIMED_OUT
```
cgwalters, line 335: Need a `;` at the end here

jmarrero: man, not passing the ` --with-curl` flag got me again. fixed.

Also it builds OK with:
```
env CFLAGS='-ggdb -Og' CXXFLAGS='-ggdb -Og' ./autogen.sh \ 
    --disable-silent-rules --enable-gtk-doc --with-selinux --with-curl \ 
    --with-openssl --with-dracut=yesbutnoconf --prefix=/usr \
    --libdir=/usr/lib64 --sysconfdir=/etc && make
```

Will get it out of draft as soon as I test this actually works.

cgwalters:
How did testing go for this?

jmarrero:
I have not had time to test this, in the last two weeks sorry. Will give it some time this week if the other bugs allow.

cgwalters:
I've rebased :surfer: this on main
And I added an official `Closes: https://github.com/ostreedev/ostree/issues/2570`
/test all

This issue just came up again in https://gitlab.apertis.org/infrastructure/apertis-issues/-/issues/349

cgwalters:
I was playing with
`iptables -A OUTPUT -p tcp --dport 8000 -m statistic --mode random --probability 0.005 -j REJECT --reject-with tcp-reset`
but this seems to have the strange semantic that at some point, the pull just hangs.  And digging in briefly, the reason seems to be that curl has a "connection cache" and at some point it reaches "too many closed connections to this host" and just stops trying to make new ones.  That's kind of annoying.

cgwalters:
I was also trying `iptables -F OUTPUT; iptables -A OUTPUT -p tcp --dport 8000 -m statistic --mode random --probability 0.05 -j DROP` but the problem with this is that TCP papers over the problem and we don't see errors at the HTTP layers (where this code would mainly come into play)...it's just really slow.

To really test this and reproduce the originating issue I think we need a HTTP/2 webserver which is intentionally returning errors.  I'm sure it can be done, but I'm having trouble in a quick search finding anything...I would think one of the clients would have a setup for this.

Dunno...I was originally going to say we should merge this, but the "pulls just hang on too many errors" effect I discovered is concerning.  With this we may potentially switch from erroring out too easily to just hanging indefinitely, which is definitely worse.

champtar:
@cgwalters if it hangs forever it's just a bug to be fixed in curl

For the HTTP/2 webserver it was an nginx running on Fedora (latest version available when opening #2570), no http2 specific tunings, don't know if it's still happening with newer versions

jmarrero:
@lucas-akira this is very helpful, thank you. I will resume working on this PR, hopefully next week.

jmarrero:
I re-tested with both:
```
iptables -A OUTPUT -p tcp --dport 8000 -m statistic --mode random --probability 0.005 -j REJECT --reject-with tcp-reset
```
and
```
iptables -F OUTPUT; iptables -A OUTPUT -p tcp --dport 8000 -m statistic --mode random --probability 0.05 -j DROP
```
on fedora and don't see the infinite hangs described on:  https://github.com/ostreedev/ostree/pull/2843#issuecomment-1593228323

The process completes successfully and we see the retries:
```
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.282: write of a823ebad56c043dd9bd17560d65938dc36e79b40c4dfdf76e37349b72dd92aac.file complete
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.297: fetch of b43c71d5ef6993bb1bba0b005159f6dedf3446af97f5bf7cde8947b12ef44a76.file complete
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.298: _ostree_fetcher_should_retry_request: error: unset, n_retries_remaining: 5
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.300: fetch of b5736b0253c681c70bea3f68dc4f6bd76aee325c0d6694cba99d6d1579903d47.file complete
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.300: _ostree_fetcher_should_retry_request: error: unset, n_retries_remaining: 5

...
b43c71d5ef6993bb1bba0b005159f6dedf3446af97f5bf7cde8947b12ef44a76.file complete
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.303: pull: idle, exiting mainloop
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.303: Committing transaction in repository 0x66c620
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.736: txn commit staging-db81f3da-910d-418d-b019-fb3f6ac5ce4a-hUkJjC
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.747: Popping lock non-blocking with timeout 30
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.747: Pop lock: state=shared, depth=1
(/run/hostsrv/github/ostreedev/ostree/.libs/ostree pull:3149): OSTree-DEBUG: 21:16:43.747: Unlocking repo
1180 metadata, 6047 content objects fetched; 106853 KiB transferred in 25 seconds; 118.1 MB content written                                               
```

Curl version is:
```
curl-8.0.1-2.fc38.x86_64.fc38.x86_64
```

Going to try centos stream now and see if I can replicate the issue.

jmarrero:
I guess that is not seeing the re-tries... It's just the debug output. mmm I am not seeing error: set to anything other than unset... @cgwalters Is there something specific I should be seeing here?

jmarrero:
> Those iptables invocations are targeting a local webserver I set up running on port 8000, if you're not doing that they won't have any effect. I think what we could do is try to set this up to run programmatically as a kola test. (And actually some of what are currently kola tests could also be GH actions tests)

I had a proxy which made no sense... after our chat I dropped the `--dport 8000` inside a VM and sadly was able to replicate the hang on the latest FCOS. As you recommended I am attempting to use a runtime environment variable on the latest push. Hopefully I am using that `g_getenv` with a boolean correctly? I just saw us use something similar here: https://github.com/ostreedev/ostree/blob/main/src/libostree/ostree-bootloader-grub2.c#L168

champtar:
@jmarrero @cgwalters was the hang reported to curl ?

src/libostree/ostree-fetcher-curl.c
```diff
@@ -330,7 +331,14 @@ check_multi_info (OstreeFetcher *fetcher)
             }
           else
             {
-              g_task_return_new_error (task, G_IO_ERROR, G_IO_ERROR_FAILED,
+              /* When it is not a file and the OSTREE_CURL_IO_RETRY
+               * environment variable is set, we want to retry the request.
+               * We accomplish that by using G_IO_ERROR_TIMED_OUT.
+               */
+              should_retry = g_getenv ("OSTREE_CURL_IO_RETRY") != NULL;
```
cgwalters, line 338: General style in ostree is to use declare-and-initialize where possible, so this can be `gboolean should_retry = ...` instead of declaring above.

cgwalters, line 338: So in GLib there is this handy API `g_parse_debug_string` which is intended for things like this and we use it e.g. over here https://github.com/ostreedev/ostree/blob/92d25eb56bcb17ee117ef5b877de86dc828e823f/src/libostree/ostree-sysroot.c#L192

The advantage of this is it provides a more uniform way to set multiple options/flags from a single environment variable; the UX would be e.g. `env OSTREE_FETCHER_FLAGS=retry-all` or so.

And then let's say later we want to add *another* flag (actually, supporting disabling http2 e.g. might be a good one) we can add `OSTREE_FETCHER_FLAGS=no-http2,retry-all`.

jmarrero:
> @jmarrero @cgwalters was the hang reported to curl ?

trying to get a nice reproducer first.

jmarrero:
OSTREE_DEBUG_HTTP might be enough to show the curl team the issue. So I am trying to get logs with that enabled now. 2 hrs running no luck yet. But *soon* hopefully.

jmarrero:
created this curl issue with the logs: https://github.com/curl/curl/issues/11873

cgwalters:
The upstream response was helpful!  This reminds me that we should probably look at librepo (the thing that fetches rpms using curl) as inspiration, and I do see https://github.com/rpm-software-management/librepo/blob/7c9af219abd49f8961542b7622fc82cfdaa572e3/librepo/handle.c#L72

If we have to just copy-paste their defaults, that seems fine to me.

cgwalters:
Now obviously we could and probably should make some of these things tunable via the ostree repository configuration too.

jmarrero:
yeah they were super helpful!, Thanks for the links that makes a lot of sense. I will give it a first pass and update the PR.

src/libostree/ostree-repo-pull.c
```diff
@@ -2995,6 +2998,9 @@ _ostree_repo_remote_new_fetcher (OstreeRepo *self, const char *remote_name, gboo
       }
   }
 
+  _ostree_fetcher_set_low_speed_limit (fetcher, (low_speed_limit) ? low_speed_limit : OPT_LOWSPEEDLIMIT_DEFAULT);
```
cgwalters, line 3001: `(low_speed_limit) ? *low_speed_limit : OPT_LOWSPEEDLIMIT_DEFAULT` will probably help

jmarrero:
Failing to build:
```
In file included from src/libostree/ostree-repo-pull-private.h:22,
                 from src/libostree/ostree-repo-pull.c:29:
src/libostree/ostree-fetcher-util.h:35:1: error: storage class specified for parameter '_ostree_fetcher_tmpf'
   35 | _ostree_fetcher_tmpf (int dfd, GLnxTmpfile *tmpf, GError **error)
      | ^~~~~~~~~~~~~~~~~~~~
src/libostree/ostree-fetcher-util.h:35:1: warning: parameter '_ostree_fetcher_tmpf' declared 'inline'
src/libostree/ostree-fetcher-util.h:36:1: error: expected ';', ',' or ')' before '{' token
   36 | {
      | ^
make[2]: *** [Makefile:5536: src/libostree/libostree_1_la-ostree-repo-pull.lo] Error 1
```
I bet It's something super basic, but I did not touch that file and I am not sure how I am affecting `/ostree-fetcher-util.h:35` @cgwalters can you see what I am doing wrong?

src/libostree/ostree-repo-pull.c
jmarrero: I removed all the pointers from the 3 options I added. Hopefully it looks better now?

src/libostree/ostree-fetcher.h
```diff
@@ -99,6 +99,12 @@ void _ostree_fetcher_set_proxy (OstreeFetcher *fetcher, const char *proxy);
 void _ostree_fetcher_set_client_cert (OstreeFetcher *fetcher, const char *cert_path,
                                       const char *key_path);
 
+void _ostree_fetcher_set_low_speed_limit (OstreeFetcher *self, long *opt_low_speed_limit);
+
+void _ostree_fetcher_set_low_speed_time (OstreeFetcher *self, long *opt_low_speed_time);
+
+void _ostree_fetcher_set_retry_all (OstreeFetcher *self, gboolean *opt_retry_all;
```
cgwalters, line 106: You're missing a trailing `)` here at least

src/libostree/ostree-repo-pull.c
```diff
@@ -4895,6 +4914,9 @@ ostree_repo_pull_with_options (OstreeRepo *self, const char *remote_name_or_base
   g_free (pull_data->remote_refspec_name);
   g_free (pull_data->remote_name);
   g_free (pull_data->append_user_agent);
+  g_free (pull_data->low_speed_limit);
+  g_free (pull_data->low_speed_time);
+  g_free (pull_data->retry_all);
```
cgwalters, line 4919: Since these aren't heap allocated we shouldn't call `free()` on them either.

src/libostree/ostree-repo-pull-private.h
```diff
@@ -54,6 +54,10 @@ typedef struct
 
   GVariant *extra_headers;
   char *append_user_agent;
+  long low_speed_limit;
+  long low_speed_time;
```
cgwalters, line 58: These two should be `guint32` - i.e. explicitly 32 bits because that's what the gvariant type `u` expects.

As is right now, this would cause undefined behavior on 64 bit architectures.

src/libostree/ostree-fetcher.h
jmarrero: lol omg

champtar:
Skimming through the change, I undert it's enabled by default ? (The commit message isn't explicit)

src/libostree/ostree-repo-pull.c
```diff
@@ -2995,6 +3000,12 @@ _ostree_repo_remote_new_fetcher (OstreeRepo *self, const char *remote_name, gboo
       }
   }
 
+  _ostree_fetcher_set_low_speed_limit (fetcher, (low_speed_limit) ? low_speed_limit
+                                                                  : OPT_LOWSPEEDLIMIT_DEFAULT);
```
cgwalters, line 3004: I think we don't need the `?` here because we're already effectively setting the default below, so this can just be:

```
  _ostree_fetcher_set_low_speed_limit (fetcher, low_speed_limit);
```
right?

And same for the other two.

src/libostree/ostree-repo-pull.c
```diff
@@ -3570,6 +3588,9 @@ ostree_repo_pull_with_options (OstreeRepo *self, const char *remote_name_or_base
                               &pull_data->timestamp_check_from_rev);
       (void)g_variant_lookup (options, "max-metadata-size", "t", &pull_data->max_metadata_size);
       (void)g_variant_lookup (options, "append-user-agent", "s", &pull_data->append_user_agent);
+      (void)g_variant_lookup (options, "low-speed-limit-bytes", "u", &pull_data->low_speed_limit);
```
cgwalters, line 3591: Though for this one let's also do what we do above for
```
  if (!opt_n_network_retries_set)
    pull_data->n_network_retries = DEFAULT_N_NETWORK_RETRIES;
```
or so to ensure the defaults are consistently inherited?

src/libostree/ostree-fetcher-curl.c
```diff
@@ -912,14 +939,10 @@ initiate_next_curl_request (FetcherRequest *req, GTask *task)
   g_assert_cmpint (rc, ==, CURLM_OK);
   rc = curl_easy_setopt (req->easy, CURLOPT_CONNECTTIMEOUT, 30L);
   g_assert_cmpint (rc, ==, CURLM_OK);
-  /* We used to set CURLOPT_LOW_SPEED_LIMIT and CURLOPT_LOW_SPEED_TIME
-   * here, but see https://github.com/ostreedev/ostree/issues/878#issuecomment-347228854
-   * basically those options don't play well with HTTP2 at the moment
-   * where we can have lots of outstanding requests.  Further,
```
cgwalters, line 918: The core question now is though...is that bug still present?  I am dimly remembering now that this just broke when we ended up doing hundreds of concurrent requests for objects.

But that itself may have already been fixed by 97e9bc7fd4a73c5c63e5390b3935cf986432ab6e !

cgwalters: 82e2150b98bd1ad5149c1e86db22f8ffb2dd6a41 came after c18628ecb8b2ef591db0444fa57052cba60807a8 so I guess that theory that the low speed limit bug is mitigated is probably wrong.

However...I wonder if it'd be helpful to drop our HTTP queuing from 8 to 2/3.  In practice I doubt that would slow things down much, and it'd avoid having curl thinking that the remote server is answering too slowly - or we've hit a hard cap on the number of concurrent streams.

For reference it looks like the latest Apache defaults too 100 concurrent streams: https://httpd.apache.org/docs/2.4/mod/mod_http2.html
which is quite large.  But I could imagine other servers have this much smaller.

cgwalters: Yeah let's definitely make the max concurrent configurable too, would make sense to do on top of this PR

src/libostree/ostree-fetcher-soup.c
```diff
@@ -360,6 +362,13 @@ session_thread_set_tls_database_cb (ThreadClosure *thread_closure, gpointer data
     }
 }
 
+static void
+session_thread_set_max_outstanding_fetcher_requests (ThreadClosure *thread_closure, gpointer data)
+{
+  guint32 n = GUINT_TO_POINTER (data);
```
cgwalters, line 368: Here we want to go the other way so try `GPOINTER_TO_UINT (data)`

src/libostree/ostree-fetcher-soup.c
```diff
@@ -360,6 +362,13 @@ session_thread_set_tls_database_cb (ThreadClosure *thread_closure, gpointer data
...
+  thread_closure->opt_max_outstanding_fetcher_requests = n;
```
cgwalters, line 369: Here we want to go the other way so try `GPOINTER_TO_UINT (data)`; how about just:
```suggestion
  thread_closure->opt_max_outstanding_fetcher_requests = GPOINTER_TO_UINT (data);
```
?

jmarrero: thanks!

jmarrero:
should be ready for review now, tested locally and can't see the hang. I also disabled the retries and that works too.

ell1e:
> but this seems to have the strange semantic that at some point, the pull just hangs

For what it's worth, I'm seeing this happen sometimes as well even without this retry patch on flatpak 1.14.4 on openSUSE Leap 15.5, not sure if related at all or a separate problem.

Issue #2753, opened 2022-10-31
Support UKI


cgwalters:
See https://github.com/uapi-group/specifications/blob/main/specs/unified_kernel_image.md
and
https://fedoraproject.org/wiki/Changes/Unified_Kernel_Support_Phase_1

There are two major points here:

## UEFI only

We'll need to add a UEFI backend to ostree, which explicitly controls the UEFI boot ordering via e.g. `efibootmgr` instead of using the `/boot/loader/entries` stuff.

## Kernel cmdline :arrow_right: rootfs

One goal of the UKI work is to have generic Linux distributions sign both the kernel and initramfs and stock kernel cmdline.  However, ostree today embeds the target rootfs in the kernel cmdline - this creates a recursion issue.

I think what we'd do instead is have the initramfs automatically find the latest symlink in `/sysroot/ostree` - we effectively do almost this with `/ostree/boot.[01]` today.

This problem here is really similar to "how to find the rootfs" which I think was discussed in that group.  The Discoverable Partitions spec was floated, but IMO that's not viable in the general case because it just blows up if one e.g. reinstalls the OS on a new disk, but leave the old rootfs around.  Or if one wants to create a bootable USB stick intending to install on another system, and you leave it in across a reboot.  xref https://github.com/coreos/fedora-coreos-tracker/issues/976
I think we will really need to continue to at least use machine-local UUIDs that are tied to the UKI to find the rootfs (and hence the ostree).

The most obvious thing here would be to use systemd-stub to make the rootfs UUID be a credential.

ricardosalveti:
> I think what we'd do instead is have the initramfs automatically find the latest symlink in /sysroot/ostree - we effectively do almost this with /ostree/boot.[01] today.

How would we know from the initramfs when a rollback was performed? Thinking on the use scenario in which the bootloader decides to rollback as the new deployment/update is not good enough (wasn't confirmed), as currently this can be done quite easily by booting the previous deployment (previous initramfs), which also has the previous ostree argument in place.

dbnicholson:
> ## UEFI only
> We'll need to add a UEFI backend to ostree, which explicitly controls the UEFI boot ordering via e.g. `efibootmgr` instead of using the `/boot/loader/entries` stuff.

[...]

PR #2844, opened 2023-04-05
Add ostree=aboot for signed Android Boot Images

ericcurtin:
Some kernel images are delivered in a signed kernel + cmdline +
initramfs + dtb blob. When this is added to the commit server side, only
after this do you know what the cmdline is, this creates a recursion
issue. To avoid this, in the case where we have ostree=aboot karg
set, do the bls parsing in the initramfs instead, so we can take
advantage of existing bls logic.

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -21,17 +21,20 @@
 #ifndef __OSTREE_MOUNT_UTIL_H_
 #define __OSTREE_MOUNT_UTIL_H_
 
+#include <dirent.h>
 #include <err.h>
-#include <stdlib.h>
-#include <sys/statvfs.h>
-#include <stdio.h>
-#include <unistd.h>
 #include <fcntl.h>
-#include <string.h>
 #include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/statvfs.h>
+#include <sys/utsname.h>
+#include <unistd.h>
 
 #define INITRAMFS_MOUNT_VAR "/run/ostree/initramfs-mount-var"
 #define _OSTREE_SYSROOT_READONLY_STAMP "/run/ostree-sysroot-ro.stamp"
+#define ABOOT_KARG "aboot"
 
 static inline int
 path_is_on_readonly_fs (const char *path)
@@ -72,6 +75,26 @@ out:
   return cmdline;
 }
 
+static inline void
+free_char (char **to_free)
+{
+  free (*to_free);
+}
+
+static inline void
+close_dir (DIR **dir)
+{
+  if (*dir)
+    closedir (*dir);
+}
+
+static inline void
+close_file (FILE **f)
+{
+  if (*f)
+    fclose (*f);
+}
+
 static inline char *
 read_proc_cmdline_ostree (void)
 {
@@ -104,6 +127,150 @@ read_proc_cmdline_ostree (void)
 
   free (cmdline);
   return ret;
+}
+
+static inline void
+cpy_and_null (char **dest, char **src)
+{
+  *dest = *src;
+  *src = NULL;
+}
+
+/* If the key matches the start of the line, copy line to out
+ */
+static inline bool
+cpy_if_key_match (char **line, const char *key, char **out)
+{
+  /* There should only be one of each key per BLS file, so check for NULL, we
+   * should only parse the first occurance of a key, if there's two, it's a
+   * malformed BLS file
+   */
+  if (!*out && strstr (*line, key) == *line)
+    {
+      cpy_and_null (out, line);
+      return true;
+    }
+
+  return false;
+}
+
+static inline bool
+has_suffix (const char *str, const char *suffix)
+{
+  if (!str || !suffix)
+    return false;
+
+  const size_t str_len = strlen (str);
+  const size_t suffix_len = strlen (suffix);
+  if (str_len < suffix_len)
+    return false;
+
+  return !strcmp (str + str_len - suffix_len, suffix);
+}
+
+/* On completion version and options will be set to new values if the version
+ * is more recent. Will loop line through line on the passed in open FILE.
+ */
+static inline void
+copy_if_higher_version (FILE *f, char **version, char **options)
+{
+  char __attribute__ ((cleanup (free_char))) *line = NULL;
+  char __attribute__ ((cleanup (free_char))) *version_local = NULL;
+  char __attribute__ ((cleanup (free_char))) *options_local = NULL;
+  char __attribute__ ((cleanup (free_char))) *linux_local = NULL;
+  /* Note getline() will reuse the previous buffer when not zero */
+  for (size_t len = 0; getline (&line, &len, f) != -1;)
+    {
+      /* This is an awful hack to avoid depending on GLib in the
+       * initramfs right now.
+       */
+      if (cpy_if_key_match (&line, "version ", &version_local))
+          continue;
+
+      if (cpy_if_key_match (&line, "options ", &options_local))
+          continue;
+
+      if (cpy_if_key_match (&line, "linux ", &linux_local))
+          continue;
+    }
+
+  /* The case where we have no version set yet */
+  if (!*version
+      || strverscmp (version_local + sizeof ("version"), (*version) + sizeof ("version")) > 0)
+    {
+      struct utsname buf;
+      uname (&buf);
+      strtok (linux_local + sizeof ("linux"), " \t\r\n");
+      if (!has_suffix (linux_local, buf.release))
+        return;
+
+      cpy_and_null (version, &version_local);
+      cpy_and_null (options, &options_local);
+    }
+
+  return;
+}
+
+static inline char *
+parse_ostree_from_options (const char *options)
+{
+  if (options)
+    {
+      options += sizeof ("options");
+      char *start_of_ostree = strstr (options, "ostree=");
+      if (start_of_ostree > options)
+        {
+          start_of_ostree += sizeof ("ostree");
+          /* trim everything to the right */
+          strtok (start_of_ostree, " \t\r\n");
+          return strdup (start_of_ostree);
+        }
+    }
+
+  return NULL;
+}
+
+/* This function is for boot arrangements where it is not possible to use a
+ * karg/cmdline, this is the case when the cmdline is part of the signed
+ * boot image, alternatively this function takes the karg from the bls entry
+ * which will have the correct ostree= karg set. This bls entry is not parsed
+ * from the bootloader but from the initramfs instead.
+ */
+static inline char *
+bls_parser_get_ostree_option (const char *sysroot)
+{
+  char out[PATH_MAX] = "";
+  int written = snprintf (out, PATH_MAX, "%s/boot/loader/entries", sysroot);
+  DIR __attribute__ ((cleanup (close_dir))) *dir = opendir (out);
+  if (!dir)
+    {
+      fprintf (stderr, "opendir(\"%s\") failed with %d\n", out, errno);
+      return NULL;
+    }
+
+  char __attribute__ ((cleanup (free_char))) *version = NULL;
+  char __attribute__ ((cleanup (free_char))) *options = NULL;
+  for (struct dirent *ent = 0; (ent = readdir (dir));)
+    {
+      if (ent->d_name[0] == '.')
+        continue;
+
+      if (!has_suffix (ent->d_name, ".conf"))
+        continue;
+
+      snprintf (out + written, PATH_MAX - written, "/%s", ent->d_name);
+
+      FILE __attribute__ ((cleanup (close_file))) *f = fopen (out, "r");
+      if (!f)
+        {
+          fprintf (stderr, "fopen(\"%s\", \"r\") failed with %d\n", out, errno);
+          continue;
+        }
+
+      copy_if_higher_version (f, &version, &options);
+    }
+
+  return parse_ostree_from_options (options);
 }
 
 /* This is an API for other projects to determine whether or not the
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -124,9 +124,13 @@ resolve_deploy_path (const char * root_mountpoint)
 {
   char destpath[PATH_MAX];
   struct stat stbuf;
-  char *ostree_target, *deploy_path;
+  char __attribute__ ((cleanup (free_char))) *ostree_target;
+  char *deploy_path;
 
   ostree_target = read_proc_cmdline_ostree ();
+  if (!strcmp (ostree_target, ABOOT_KARG))
+    ostree_target = bls_parser_get_ostree_option (root_mountpoint);
+
   if (!ostree_target)
     errx (EXIT_FAILURE, "No OSTree target; expected ostree=/ostree/boot.N/...");
 
--- a/src/switchroot/ostree-system-generator.c
+++ b/src/switchroot/ostree-system-generator.c
@@ -62,7 +62,10 @@ main(int argc, char *argv[])
    * exit so that we don't error, but at the same time work where switchroot
    * is PID 1 (and so hasn't created /run/ostree-booted).
    */
-  char *ostree_cmdline = read_proc_cmdline_ostree ();
+  char __attribute__ ((cleanup (free_char))) *ostree_cmdline = read_proc_cmdline_ostree ();
+  if (!strcmp (ostree_cmdline, ABOOT_KARG))
+    ostree_cmdline = bls_parser_get_ostree_option ("/sysroot");
+
   if (!ostree_cmdline)
     exit (EXIT_SUCCESS);
 


ericcurtin:
Related to conversations in https://github.com/ostreedev/ostree/issues/2753 ... Boots fine

```
[root@localhost ~]# cat /proc/cmdline
root=LABEL=root ro loglevel=4 efi=runtime libahci.ignore_sss=1 console=ttyAMA0 rw
[root@localhost ~]# rpm-ostree status
State: idle
Deployments:
● auto-sig:cs9/aarch64/abootqemu-minimal
                  Version: 9 (2023-04-05T15:27:58Z)
                   Commit: 210167d098a681908be52d0ea79b87c6b96570584c7c7ee643bf722c2160d653
```

Although I want to do some more testing with some greenboot changes.

src/boot/ostree-remount.service
```diff
@@ -17,7 +17,6 @@
 Description=OSTree Remount OS/ Bind Mounts
 Documentation=man:ostree(1)
 DefaultDependencies=no
-ConditionKernelCommandLine=ostree
```
cgwalters, line 20: This should change to `ConditionPathExists=/run/ostree` instead; we don't want to run it unconditionally.

That's true of basically everything that is `After=ostree-prepare-root.service` which creates it.

That change can land now, independent of this change.  Also, we can start porting things in [in fedora-coreos-config](https://github.com/search?q=repo%3Acoreos%2Ffedora-coreos-config%20ConditionKernelCommandLine%3Dostree&type=code) etc.

src/boot/ostree-prepare-root.service
```diff
@@ -17,7 +17,6 @@
 Description=OSTree Prepare OS/
 Documentation=man:ostree(1)
 DefaultDependencies=no
-ConditionKernelCommandLine=ostree
```
cgwalters, line 20: This one is tricky.  What we wanted to defend against here is anyone who has the ostree dracut module installed but isn't booting ostree.

This is really the fundamental change; you're changing it here so that if we detect an ostree setup, we use it by default.  We may need to support something like `ostree=no` to disable it for oddball cases?

I'm not totally sure.

Maybe this isn't a real case anymore - quite a while ago we changed the ostree dracut module to be optionally opt-in (yes this is confusing).  See https://github.com/ostreedev/ostree/blob/c9271423b1c780ccb9ff47685d22d744968d9d19/configure.ac#L504
which is used by Fedora derivatives today.  This means we don't inject the module by default, but require invokers of dracut to use e.g. `--add ostree` explicitly, and this is done by rpm-ostree for example.

OK so, perhaps this is just OK to land.

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +120,69 @@ read_proc_cmdline_ostree (void)
...
+  if (!dir)
+    {
+      fprintf(stderr, "'%s' %d", strerror(errno), __LINE__);
+      return "";
+    }
+
+  struct dirent *ent = 0;
+  for (time_t latest = 0; (ent = readdir(dir)); )
+    {
+      if (ent->d_type == DT_LNK)
```
cgwalters, line 144: Usually I try to avoid nesting by doing e.g.:
```
  if (ent->d_type != DT_LNK)
    continue;
```

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +120,69 @@ read_proc_cmdline_ostree (void)
...
+              appended_len = snprintf(append_from, dir_str_cap, "/%s", ent->d_name);
+              latest = lsb.st_mtime;
+            }
+        }
+    }
+
+  dir_str_len += appended_len;
+  dir_str_cap = PATH_MAX - dir_str_len;
+
+  for (int i = 0; i < 3; ++i)
```
cgwalters, line 160: What's with the `3` here?  Usually things like this want at least e.g.
```
   const int retries = 3;  // Because...?
```

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +120,69 @@ read_proc_cmdline_ostree (void)
   return ret;
 }
 
+static inline char*
+gen_ostree_target (const char* prepend)
```
cgwalters, line 124: This function really really wants a documentation comment describing what it does.

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +120,69 @@ read_proc_cmdline_ostree (void)
...
+{
+  char dir_str[PATH_MAX];
+  int dir_str_len = snprintf(dir_str, sizeof(dir_str), "%s", prepend);
+  int dir_str_cap = PATH_MAX - dir_str_len;
+  int prepend_len = dir_str_len;
+  dir_str_len += snprintf(dir_str + prepend_len, sizeof(dir_str), "%s", "/ostree");
+  dir_str_cap = PATH_MAX - dir_str_len;
+  char* append_from = dir_str + dir_str_len;
+  int appended_len = 0;
```
cgwalters, line 133: There's a lot of hairy C string parsing and arithmetic here.  It really looks to me like this should simplify down to something more like:

`snprintf(dir_str, sizeof(dir_str), "%s/%s", prepend, ostree)`  ?  Not totally sure.

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +120,69 @@ read_proc_cmdline_ostree (void)
...
+    }
+
+  struct dirent *ent = 0;
+  for (time_t latest = 0; (ent = readdir(dir)); )
+    {
+      if (ent->d_type == DT_LNK)
+        {
+          struct stat lsb;
+          lsb.st_mtime = 0;
+          lstat(ent->d_name, &lsb);
```
cgwalters, line 148: Let's not ignore errors from `lstat()` here.

cgwalters (commented):
Thanks for starting this!

src/switchroot/ostree-mount-util.h
ericcurtin: It basically saved me doing an extra strlen, by appending prepend and ostree one by one. Because I need the length of the prepend for the very last line of the function.

I can do one sprintf and a strlen though, no biggie, same thing, more readable I guess! ☺️

src/boot/ostree-prepare-root.service
ericcurtin: Our oddball case for the record is we are gonna read some GPT flag to see if we want to boot the latest symlink or rollback once  (that's all that we require)... But I haven't got that far yet, so at the moment, it's just boot the latest symlink!

src/switchroot/ostree-mount-util.h
ericcurtin: This adds the 3 next dirs to the string, there are 5 directories in the string in total, I can add a comment to make more clear. It's not a retry, it's recursing down the directory.

src/switchroot/ostree-mount-util.h
ericcurtin: Will error handle this

src/switchroot/ostree-mount-util.h
ericcurtin: Makes sense

src/switchroot/ostree-mount-util.h
ericcurtin: Well it's actually 6 but the first one is "/sysroot" which we need to throw away at the end, we just need it here to open directories because the "/ostree" directory is at "/sysroot/ostree" at this point.

src/switchroot/ostree-mount-util.h
ericcurtin: Will add to give context

src/switchroot/ostree-mount-util.h
ericcurtin: This will probably change to something other than, just add 3 next directories onto the end of the string as we figure things out

ericcurtin:
More info of what a system booted like this looks like below. Just to point out, aboot bootloaders do not part BLS entries at all. But we can still use that data, that might be useful, we can just parse BLS in the initramfs to decide exactly where we need to boot from, once we've figured out the right entry.

BLS cmdline does not match the /proc/cmdline here, but I think that's ok. A person debugging this should know aboot does not parse BLS, it's a pretty simple booloader it boots the boot image in slot A or slot B, but the data there can still help keep the aboot technique consistent with a BLS bootloader like grub.

```
rpm-ostree status; echo; cat /proc/cmdline; echo; cat /boot/loader.1/entries/ostree-1-centos.conf
Deployments:
● auto-sig:cs9/aarch64/abootqemu-minimal
                  Version: 9 (2023-04-06T09:23:50Z)
                   Commit: 3bc0523161590c853b6aaefa43f81e1072226ab1ca49c0e188ec5768139ffd17

root=LABEL=root ro loglevel=4 efi=runtime libahci.ignore_sss=1 console=ttyAMA0 rw

title CentOS AutoSD 9 (ostree:0)
version 1
options root=LABEL=root ro loglevel=4 efi=runtime libahci.ignore_sss=1 console=ttyAMA0 ostree=/ostree/boot.1/centos/7cc01e1ce3c47a67d5a60a0c3e515dafc5bbaed2596a42c830f80234e782fea1/0 rw
linux /ostree/centos-7cc01e1ce3c47a67d5a60a0c3e515dafc5bbaed2596a42c830f80234e782fea1/vmlinuz-5.14.0-999.124.pre.ES2.CCU.el9iv.aarch64
initrd /ostree/centos-7cc01e1ce3c47a67d5a60a0c3e515dafc5bbaed2596a42c830f80234e782fea1/initramfs-5.14.0-999.124.pre.ES2.CCU.el9iv.aarch64.img
abootcfg /ostree/deploy/centos/deploy/3bc0523161590c853b6aaefa43f81e1072226ab1ca49c0e188ec5768139ffd17.0/usr/lib/ostree-boot/aboot.cfg
aboot /ostree/centos-7cc01e1ce3c47a67d5a60a0c3e515dafc5bbaed2596a42c830f80234e782fea1/aboot-5.14.0-999.124.pre.ES2.CCU.el9iv.aarch64.img
fdtdir /ostree/centos-7cc01e1ce3c47a67d5a60a0c3e515dafc5bbaed2596a42c830f80234e782fea1/dtb
```

src/boot/ostree-prepare-root.service
jlebon: `ConditionKernelCommandLine=ostree` is present in a lot of units outside of this repo as well. For alignment, it's probably cleaner to just leave it there.

One approach is to have something like `ostree=latest`. Then `ConditionKernelCommandLine=ostree` is still valid, but the kargs don't need to change on each update. It also makes the logic to boot the latest symlink explicit.

ericcurtin: ostree=latest works for me!

ericcurtin: A question for both of you @cgwalters @jlebon . The BLS files exist, but are not readable by the bootloader. But we can read them just fine in the initramfs, by changing this code up a little...

I was wondering what's the most bulletproof way of identifying the most recent BLS .conf file and the second most recent BLS .conf? Then we can just take the ostree= from there, I think this could be a simple solution to formulate the full path. (this might make more sense looking at the output below, BLS and /proc/cmdline are not consistent because it doesn't really matter when your bootloader cannot parse BLS)

@alexlarsson was also suggesting earlier, it could be possible that maybe we always make ostree-0-centos.conf the latest one, then we can guarantee ostree-1-centos.conf is the rollback one (or vice-versa if it's easier).

Could this work, without breaking anything existing in ostree? Still learning all the little nuances of ostree day by day :smile:

ericcurtin: The guaranteeing ostree-0-centos.conf and ostree-1-centos.conf are latest and the second most latest is kinda nice as you don't have to worry about timestamps getting messed up somehow.

cgwalters: I do still think though that we can add some sort of flag to the initramfs that says "require ostree"; maybe `/usr/lib/ostree-required` or whatever.

Then units in the initramfs can use `ConditionPathExists` on that.

Or, any units that should run in the initramfs iff ostree is enabled could be pulled in via `WantedBy=ostree-prepare-root.service`, though that gets a bit trickier to manage.

(But this is a distinct thread from the suggestion of having code in the initramfs parse the BLS configs)

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +120,96 @@ read_proc_cmdline_ostree (void)
...
+          appended_len = snprintf(append_from, *cap, "/%s", ent->d_name);
+          latest = lsb.st_mtime;
+        }
+    }
+
+  *len += appended_len;
+  *cap = PATH_MAX - *len;
+
+  return 0;
+}
```
bauen1, line 161: So, as I take it, this method searches `$prefix/ostree/` and appends the latest symlink.

Maybe it would be good to only look at `$prefix/ostree/boot*` (perhaps even limited to all possible combinations) ?

Otherwise creating a link in there (even if you shouldn't do so) seems to have potential to break any attempt to boot.

Sorting by modification timestamps seems a bit fragile to me, can't that be done by simply looking at the number in the link ?

bauen1:
So the current logic simply picks the latest, as defined by modification timestamp, file from `/sysroot/ostree/boot.$x/$stateroot/$bootcsum/$checkoutnum` as checkout to use ?

I think that will lead to unexpected results at best, and unexpected vulnerabilities at worst, and I guess that's why this is a Draft currently.

Basing the decision on the _modified timestamp_ could have the very slim chance of actually booting an older, but later modified, deployment, I guess that situation would be rare.

It would also mean, that you can boot a (signed) kernel with a different checkout than it was deployed with, which could open up new vulnerabilities, e.g. new deployment relies on kernel feature / initramfs changes that aren't present.

But the `$bootcsum` should be computable by the loader (or maybe even the kernel/initramfs itself), so while you can't embed the hash inside the images (which is the data being hashed), perhaps it could be passed by the UKI-stub or loader.
Then you wouldn't need to "guess" it, and can ensure that a deployment would only ever be booted by a image it was build with.
(Side-Note: That could even work if the bootcsum ends up being the hash of the UKI, but in that case it could be preferable to use the PECOFF hash that would be signed).

Rollbacks / multiple deployments, seem to be impossible with this logic, since even with an older image the latest checkout is used.
I think the information of _which deployment_ to use comes inherently from an unverifiable source and the best you can do is probably to limit the image+deployment combinations to ones where the image booted and the image of the deployment match (bootcsum).
And I believe that is true even if you implement automatic health checks, since at some point you need to trust something, and realistically you want to give users the option of choosing to boot rollback image+deployment (e.g. by means of UEFI boot selection / sd-boot).

In any case you're putting your trust in nobody tampering with the actual filesystem, which is quite a far shot by today's standards ...

src/switchroot/ostree-mount-util.h
ericcurtin: > So the current logic simply picks the latest, as defined by modification timestamp, file from `/sysroot/ostree/boot.$x/$stateroot/$bootcsum/$checkoutnum` as checkout to use ?

Yes.

> 
> I think that will lead to unexpected results at best, and unexpected vulnerabilities at worst, and I guess that's why this is a Draft currently.

Yes again, I'm just trying things at the moment.

> 
> Basing the decision on the _modified timestamp_ could have the very slim chance of actually booting an older, but later modified, deployment, I guess that situation would be rare.

Yes I am worried about this too, expect the implementation to change.

> 
> It would also mean, that you can boot a (signed) kernel with a different checkout than it was deployed with, which could open up new vulnerabilities, e.g. new deployment relies on kernel feature / initramfs changes that aren't present.
> 
> But the `$bootcsum` should be computable by the loader (or maybe even the kernel/initramfs itself), so while you can't embed the hash inside the images (which is the data being hashed), perhaps it could be passed by the UKI-stub or loader. Then you wouldn't need to "guess" it, and can ensure that a deployment would only ever be booted by a image it was build with. (Side-Note: That could even work if the bootcsum ends up being the hash of the UKI, but in that case it could be preferable to use the PECOFF hash that would be signed).

I am using a very limited bootloader that boots either an A or B boot partition, that contains an Android Boot Image. But there is very limited data available in the partition that I can be able to use in order to identify if I should use the latest or the latest-1 version. If that makes sense.

UKI's probably have much more options.

> 
> Rollbacks / multiple deployments, seem to be impossible with this logic, since even with an older image the latest checkout is used. I think the information of _which deployment_ to use comes inherently from an unverifiable source and the best you can do is probably to limit the image+deployment combinations to ones where the image booted and the image of the deployment match (bootcsum). And I believe that is true even if you implement automatic health checks, since at some point you need to trust something, and realistically you want to give users the option of choosing to boot rollback image+deployment (e.g. by means of UEFI boot selection / sd-boot).

I only want Android style A/B rollbacks, nothing else.

> 
> In any case you're putting your trust in nobody tampering with the actual filesystem, which is quite a far shot by today's standards ...

Will need to be hardened.

I think you can take this PR with a pinch of salt at the minute, it will be drastically changed. But thanks for the early feedback.

bauen1: > I am using a very limited bootloader that boots either an A or B boot partition, that contains an Android Boot Image. But there is very limited data available in the GPT partition table that I can be able to use in order to identify if I should use the latest or the latest-1 version. If that makes sense.

Ah, that makes things a bit easier :thinking: 
I think it might end up being a very similar trade-off like you have to do with UKI, since the information on what to boot (A or B) comes from an untrusted / unverified source.

> > In any case you're putting your trust in nobody tampering with the actual filesystem, which is quite a far shot by today's standards ...
> Will need to be hardened.

It was more of a general comment, a lot of things currently depend on the filesystem not having been tampered with.

> I think you can take this PR with a pinch of salt at the minute, it will be drastically changed. But thanks for the early feedback.

It is very nice to see some attempts at implementing it, even if it wouldn't fit my use case in the end :rocket: 
As always, an actual (incomplete) implementation is much easier to argue about than some abstract idea.

cgwalters:
The code seems tractable...it's the maintenance and testing that seems like the hard part here.  Having two ways to do it permanently seems like a problem.

ericcurtin:
Ok pushed tonight instead had twenty minutes to spare

src/boot/ostree-remount.service
ericcurtin: This file is no longer changed

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +128,115 @@ read_proc_cmdline_ostree (void)
...
+  char __attribute__ ((cleanup (free_char))) *line = NULL;
+  char version_local[PATH_MAX] = "";
+  char options_local[PATH_MAX] = "";
+  /* Note getline() will reuse the previous buffer */
+  for (size_t len = 0; getline (&line, &len, f) != -1;)
+    {
+      /* This is an awful hack to avoid depending on GLib in the
+       * initramfs right now.
+       */
+      if (rm_key_if_match (line, "version ", sizeof ("version"), version_local))
```
alexlarsson, line 162: Please pass in sizeof(version_local) to ensure we're not overflowing the destination.

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +128,115 @@ read_proc_cmdline_ostree (void)
...
+    {
+      fprintf (stderr, "opendir(\"%s\") failed with %d\n", out, errno);
+      return strdup ("");
+    }
+
+  char version[PATH_MAX] = "";
+  char options[PATH_MAX] = "";
+  for (struct dirent *ent = 0; (ent = readdir (dir));)
+    {
+      if (ent->d_name[0] == '.')
```
alexlarsson, line 211: Maybe should check `.conf` suffix too, to avoid autosave files, etc. Might wanna stat and check for dirs too?

alexlarsson:
So, overall I like this approach. However I'm not sure I agree with the use of the BLS version field. You may use the ostree to switch to an older version, and you want if that fails to fall back to the one that was booted when you deployed the older version. Version ordering doesn't help you here. What does help is the ostree deployment index, which will be "1" for the newly deployed thing and "2" for the previous one. 

This is encoded in the filename, like ostree-1-fedora-coreos.conf for index 1 and ostree-2-fedora-coreos.conf for index 2. Should we not just always boot the file that starts with "ostree-1-" and fall back to "ostree-2-"?

src/switchroot/ostree-mount-util.h
```diff
@@ -106,6 +128,115 @@ read_proc_cmdline_ostree (void)
...
+      if (rm_key_if_match (line, "options ", sizeof ("options"), options_local))
+        {
+          if (version_local[0])
+            break;
+
+          continue;
+        }
+    }
+
+  const int ret = strverscmp (version_local, version);
```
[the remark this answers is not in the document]
ericcurtin: I did this this way to be consistent with the other BLS parser in ostree:

https://github.com/ostreedev/ostree/blob/b4dcfd7ede821c2577c04accb37ee8d5a25a5ab8/src/libostree/ostree-sysroot.c#L595

Although I could do a more fully fledged version comparison if we wanted, there's a good example implementation in systemd. I guess my question is then, should I create a shared version comparison function like the systemd one to be used in both places in ostree (here and in ostree-sysroot.c)?

https://github.com/systemd/systemd/blob/91010bd6ccd363be9e09f22c60d45f5e27c1d2a8/src/fundamental/string-util-fundamental.c#LL82C5-L82C24

src/switchroot/ostree-mount-util.h
ericcurtin: I guess you could theoretically overflow version_local if the version field in "line" was longer that 4096 bytes. That would be nuts, but no hard checking, I'll do similar for options_local, since it's the exact same logic.

ericcurtin:
> So, overall I like this approach. However I'm not sure I agree with the use of the BLS version field. You may use the ostree to switch to an older version, and you want if that fails to fall back to the one that was booted when you deployed the older version. Version ordering doesn't help you here. What does help is the ostree deployment index, which will be "1" for the newly deployed thing and "2" for the previous one.
> 
> This is encoded in the filename, like ostree-1-fedora-coreos.conf for index 1 and ostree-2-fedora-coreos.conf for index 2. Should we not just always boot the file that starts with "ostree-1-" and fall back to "ostree-2-"?

I'll test a few more rollbacks before continuing to avoid wasted effort, you are probably right. My first version of this yesterday compared filenames before I changed it to the "version " field. Literally copying and pasting my logic from yesterday:

```
Which is more correct when deciding what is the latest BLS file to pick, do a vercmp of the filename or do a vercmp of the actual "version " field in the file text? Or are they guaranteed to be the same (because if they are guaranteed to be the same case I'll just check the filename it's quicker and shorter code, don't have to open in the file in C)? See my grep below of two BLS files on an ostree system

grep "version" /boot/loader/entries/ostree-*
/boot/loader/entries/ostree-1-centos.conf:version 1
/boot/loader/entries/ostree-2-centos.conf:version 2

Nevermind actually I'll open the file to be consistent with some other ostree code.
```



Funnily enough my first version yesterday confime

alexlarsson:
Maybe @cgwalters knows best what is right here.,

src/switchroot/ostree-mount-util.h
ericcurtin: Added ".conf" check

src/switchroot/ostree-mount-util.h
ericcurtin: I actually did this a different way now, I didn't need to copy strings at all, I could just the pointer to the string allocated by getline and if you set the line char* to NULL, getline will allocate a new buffer next time around.

I copy the pointer to a string instead of strcpy now.

cgwalters:
I'm a little confused here; so ostree is both telling aboot via `aboot-deploy` to synchronize the kernel, and then now we're also trying to read the bootloader entries in the initramfs.  It's not clear to me how rollbacks work in this scenario.  Or actually, how atomicity of updates would work.

I think what I'm getting stuck on here is kmods in the userspace root. Do you have those in your use case?  Does Android?

If the flow here is:

1. running tree with kver=6.1
2. deploy new tree with kver=6.2
3. Run `aboot-deploy` to copy new kernel to updated aboot partition
4. Write bootloader configuration

Now if we're interrupted at step 3...won't we be trying to boot into the new kernel with the old userspace?

(BTW where's the source to `aboot-deploy`?)

alexlarsson:
@cgwalters So, in aboot the deploy is not really the writing of the partition. There are two boot partitions, boot_a and boot_b, when you are booted to boot_a, you write the new deployment to boot_b, and the switch is not the writing of the partition. Instead it is at the end when you flip a bit in the partition table making boot_b the default boot partition. This would be done after writing the boot loader files. 

In case of a rollback we would boot the old boot_a partition, and the idea is to have the initrd notice with partition we booted from, and if it is A (instead of the expected B), then it would pick the BLS file that points to the older userspace.


BTW. I wonder if we could just generate a "build id" (essentially a random number) for each new ostree commit  build, and encode that in the kernel command line (but no deploy id). Then we could use that (instead of the version) to find the matching BLS file and find the deloy id from it.

ericcurtin:
> I'm a little confused here; so ostree is both telling aboot via `aboot-deploy` to synchronize the kernel, and then now we're also trying to read the bootloader entries in the initramfs. It's not clear to me how rollbacks work in this scenario. Or actually, how atomicity of updates would work.

Rollbacks are not accounted for in this PR yet. I was hoping to do that in a follow on PR. This is just boot the latest. Our firmware writers are gonna give us some bits in the GPT partition table to detect when we need to fallback but they haven't delivered this to us yet. I also have tested the above code without greenboot for the moment. @eric-ch is working on greenboot for this platform.

> 
> I think what I'm getting stuck on here is kmods in the userspace root. Do you have those in your use case? Does Android?
> 
> If the flow here is:
> 
> 1. running tree with kver=6.1
> 2. deploy new tree with kver=6.2
> 3. Run `aboot-deploy` to copy new kernel to updated aboot partition
> 4. Write bootloader configuration
> 
> Now if we're interrupted at step 3...won't we be trying to boot into the new kernel with the old userspace? If there are no loadable kmods, that's perhaps not _fatal_ but we'd clearly need a reconcilation process to reboot again.

Well if we get interrupted at any point, we won't tell the system to switch slots, so we would still be booting old kernel and userspace. Although I'm only testing single slot at the minute.

> 
> This also means that anyone shipping this would need to account for testing skewed kernel-userspace combinations (in reality that _should_ work but...)

I'm gonna push for not skewing the kernel-userspace combinations. One thing I can do right now to further harden the code above, is I can compare a uname against the "linux " entry in the BLS to check if they align. But I think we should be able to avoid all skewed kernel-userspace combinations with the various pieces of information we will have in the end.

> 
> If there are loadable kmods, then we're just broken.
> 
> (BTW where's the source to `aboot-deploy`?)

It's here:

https://gitlab.com/CentOS/automotive/rpms/aboot-deploy/-/blob/main/aboot-deploy

It doesn't do all that much except for dd'ing an Android Boot Image to a boot partition. Once I get this PR done and a few others, it will have to read the GPT partition table etc. to detect whether it should write to boot_a or boot_b, but that's not done yet.

ericcurtin:
> @cgwalters So, in aboot the deploy is not really the writing of the partition. There are two boot partitions, boot_a and boot_b, when you are booted to boot_a, you write the new deployment to boot_b, and the switch is not the writing of the partition. Instead it is at the end when you flip a bit in the partition table making boot_b the default boot partition. This would be done after writing the boot loader files.

Yup that flipping the bit is the switch. Although now we have two switches, because we use BLS also.

> 
> In case of a rollback we would boot the old boot_a partition, and the idea is to have the initrd notice with partition we booted from, and if it is A (instead of the expected B), then it would pick the BLS file that points to the older userspace.

Yeah it's gonna be similar to this.

> 
> BTW. I wonder if we could just generate a "build id" (essentially a random number) for each new ostree commit build, and encode that in the kernel command line (but no deploy id). Then we could use that (instead of the version) to find the matching BLS file and find the deloy id from it.

This sounds like a good idea, it does mean we would have to write to the Android Boot Partitions everytime, even for userspace upgrades, which is actually ok, not a big deal. Lets keep this in mind if we need it. It could simplify things quite a bit.

ericcurtin:
> > So, overall I like this approach. However I'm not sure I agree with the use of the BLS version field. You may use the ostree to switch to an older version, and you want if that fails to fall back to the one that was booted when you deployed the older version. Version ordering doesn't help you here. What does help is the ostree deployment index, which will be "1" for the newly deployed thing and "2" for the previous one.
> > This is encoded in the filename, like ostree-1-fedora-coreos.conf for index 1 and ostree-2-fedora-coreos.conf for index 2. Should we not just always boot the file that starts with "ostree-1-" and fall back to "ostree-2-"?
> 
> I'll test a few more rollbacks before continuing to avoid wasted effort, you are probably right. My first version of this yesterday compared filenames before I changed it to the "version " field. Literally copying and pasting my logic from yesterday:
> 
> ```
> Which is more correct when deciding what is the latest BLS file to pick, do a vercmp of the filename or do a vercmp of the actual "version " field in the file text? Or are they guaranteed to be the same (because if they are guaranteed to be the same case I'll just check the filename it's quicker and shorter code, don't have to open in the file in C)? See my grep below of two BLS files on an ostree system
> 
> grep "version" /boot/loader/entries/ostree-*
> /boot/loader/entries/ostree-1-centos.conf:version 1
> /boot/loader/entries/ostree-2-centos.conf:version 2
> 
> Nevermind actually I'll open the file to be consistent with some other ostree code.
> ```

Btw @cgwalters did you see this question. Which is the correct way to do a vercmp check in ostree, via filename or "version " field or are they the same?

src/switchroot/ostree-prepare-root.c
```diff
@@ -124,9 +124,13 @@ resolve_deploy_path (const char * root_mountpoint)
 {
   char destpath[PATH_MAX];
   struct stat stbuf;
-  char *ostree_target, *deploy_path;
+  char __attribute__ ((cleanup (free_char))) *ostree_target;
```
ericcurtin, line 127: Pretty sure this was leaking here and in src/switchroot/ostree-system-generator.c all along, but both are short lived processes so you wouldn't really notice.

ericcurtin:
Added a uname check

jlebon:
> In case of a rollback we would boot the old boot_a partition, and the idea is to have the initrd notice with partition we booted from, and if it is A (instead of the expected B), then it would pick the BLS file that points to the older userspace.

I'm not sure I follow this. If the initrd knows from which partition it booted from, can't it pick the latest BLS in *its* partition? Or maybe that's what you're saying but in different words.

It feels confusing to have both multiple boot partitions *and* multiple BLS entries per partition. IIUC, I think ideally we'd have ostree just emit a single BLS entry instead in that case, but that's clearly a more invasive patch. It seems to me though like we should just pretend the older BLS entries don't exist (this rules out any potential mismatching issues).

And then, rather than reimplementing a BLS parser, we could have some bootloader option that writes out the deployment root path in some file (or maybe as a symlink) under the `/boot/loader` symlink.

ericcurtin:
> > In case of a rollback we would boot the old boot_a partition, and the idea is to have the initrd notice with partition we booted from, and if it is A (instead of the expected B), then it would pick the BLS file that points to the older userspace.
> 
> I'm not sure I follow this. If the initrd knows from which partition it booted from, can't it pick the latest BLS in _its_ partition? Or maybe that's what you're saying but in different words.

This is a possible technique, we do not have A/B system partitions (because it doesn't make as much sense when it's managed by ostree that can keep multiple version. But both Android Boot Images (the are like UKIs) can be from booted from in either boot partition A or B.

> 
> It feels confusing to have both multiple boot partitions _and_ multiple BLS entries per partition. IIUC, I think ideally we'd have ostree just emit a single BLS entry instead in that case, but that's clearly a more invasive patch. It seems to me though like we should just pretend the older BLS entries don't exist (this rules out any potential mismatching issues).

We plan on having one set for BLS entries for both partitions.

> 
> And then, rather than reimplementing a BLS parser, we could have some bootloader option that writes out the deployment root path in some file (or maybe as a symlink) under the `/boot/loader` symlink.

Could do, but BLS is easy to parse, it's a nice simple format tbh, even if we started using a separate file, BLS file format would make sense as it's easy to parse in C.

cgwalters:
> Btw @cgwalters did you see this question. Which is the correct way to do a vercmp check in ostree, via filename or "version " field or are they the same?

Well, the right way to answer this question is to look at the consumers.  For e.g. zipl, it's 
https://github.com/ibm-s390-linux/s390-tools/blob/e6be1c1b793915e001ceaf3f6f93a00c396618dc/zipl/src/scan.c#L757

which is parsing the `version` field.  But AIUI the original spec versions called for sorting by filenames too?  I may be wrong.
Reading the code for grub2 in https://src.fedoraproject.org/rpms/grub2/blob/rawhide/f/0021-blscfg-add-blscfg-module-to-parse-Boot-Loader-Specif.patch it looks like it's sorting by filename...

jlebon:
> > I'm not sure I follow this. If the initrd knows from which partition it booted from, can't it pick the latest BLS in _its_ partition? Or maybe that's what you're saying but in different words.
> 
> This is a possible technique, we do not have A/B system partitions (because it doesn't make as much sense when it's managed by ostree that can manage multiple versions). But both Android Boot Images (the are like UKIs) can be from booted from in either boot partition A or B.
> 
> > It feels confusing to have both multiple boot partitions _and_ multiple BLS entries per partition. IIUC, I think ideally we'd have ostree just emit a single BLS entry instead in that case, but that's clearly a more invasive patch. It seems to me though like we should just pretend the older BLS entries don't exist (this rules out any potential mismatching issues).
> 
> We plan on having one set for BLS entries for both partitions.

Ack thanks, this helps a lot. So IIUC, the A/B boot partitions are purely for the UKI-like blobs, but otherwise are separate from `/boot` where the BLS entries are written to as usual?

So then my next question is re. this bit:

> Well if we get interrupted at any point, we won't tell the system to switch slots, so we would still be booting old kernel and userspace.

There's two critical events: the symlink update in the bootfs for pointing at the new set of BLS entries, and the slot switch. If we get interrupted after the symlink update but before the slot switch, isn't the system state inconsistent? I.e. won't the active slot (containing the old kernel) try to boot the now latest deployment (containing the new userspace)?

Sorry if this is obvious to everyone else. I'm just trying to understand how this all fits since we're bound to get more UKI-related things in the future.

alexlarsson:
> > > I'm not sure I follow this. If the initrd knows from which partition it booted from, can't it pick the latest BLS in _its_ partition? Or maybe that's what you're saying but in different words.
> > 
> > 
> > This is a possible technique, we do not have A/B system partitions (because it doesn't make as much sense when it's managed by ostree that can manage multiple versions). But both Android Boot Images (the are like UKIs) can be from booted from in either boot partition A or B.
> > > It feels confusing to have both multiple boot partitions _and_ multiple BLS entries per partition. IIUC, I think ideally we'd have ostree just emit a single BLS entry instead in that case, but that's clearly a more invasive patch. It seems to me though like we should just pretend the older BLS entries don't exist (this rules out any potential mismatching issues).
> > 
> > 
> > We plan on having one set for BLS entries for both partitions.
> 
> Ack thanks, this helps a lot. So IIUC, the A/B boot partitions are purely for the UKI-like blobs, but otherwise are separate from `/boot` where the BLS entries are written to as usual?

Yeah. The boot_a and boot_b partitions are "raw" data partitions with just the UKI blob. They have no filesystem on them at all. Typically in aboot system boot_a would be paired with a read-only dm-verity:ed userspace fs image in the system_a partition and boot_b would be paired with system_b partition. However, since our system partition is writable with the ostree repo on it we will only have one system partition.
 
> So then my next question is re. this bit:
> 
> > Well if we get interrupted at any point, we won't tell the system to switch slots, so we would still be booting old kernel and userspace.
> 
> There's two critical events: the symlink update in the bootfs for pointing at the new set of BLS entries, and the slot switch. If we get interrupted after the symlink update but before the slot switch, isn't the system state inconsistent? I.e. won't the active slot (containing the booted kernel) try to boot the now latest deployment (containing the new userspace)?
> 
> Sorry if this is obvious to everyone else. I'm just trying to understand how this all fits since we're bound to get more UKI-related things in the future.

Yeah, this is true. This is why I proposed adding a "randomly generated" build identifier to the initrd, which would be paired with the BLS file. Then whichever initrd gets booted it can finds its own correct BLS file in either of the boot directories.

ericcurtin:
> > > > I'm not sure I follow this. If the initrd knows from which partition it booted from, can't it pick the latest BLS in _its_ partition? Or maybe that's what you're saying but in different words.
> > > 
> > > 
> > > This is a possible technique, we do not have A/B system partitions (because it doesn't make as much sense when it's managed by ostree that can manage multiple versions). But both Android Boot Images (the are like UKIs) can be from booted from in either boot partition A or B.
> > > > It feels confusing to have both multiple boot partitions _and_ multiple BLS entries per partition. IIUC, I think ideally we'd have ostree just emit a single BLS entry instead in that case, but that's clearly a more invasive patch. It seems to me though like we should just pretend the older BLS entries don't exist (this rules out any potential mismatching issues).
> > > 
> > > 
> > > We plan on having one set for BLS entries for both partitions.
> > 
> > 
> > Ack thanks, this helps a lot. So IIUC, the A/B boot partitions are purely for the UKI-like blobs, but otherwise are separate from `/boot` where the BLS entries are written to as usual?
> 
> Yeah. The boot_a and boot_b partitions are "raw" data partitions with just the UKI blob. They have no filesystem on them at all. Typically in a aboot system, boot_a would be paired with a read-only dm-verity:ed userspace fs image in the system_a partition and boot_b would be paired with system_b partition. However, since our system partition is writable with the ostree repo on it we will only have one system partition.
> 
> > So then my next question is re. this bit:
> > > Well if we get interrupted at any point, we won't tell the system to switch slots, so we would still be booting old kernel and userspace.
> > 
> > 
> > There's two critical events: the symlink update in the bootfs for pointing at the new set of BLS entries, and the slot switch. If we get interrupted after the symlink update but before the slot switch, isn't the system state inconsistent? I.e. won't the active slot (containing the booted kernel) try to boot the now latest deployment (containing the new userspace)?
> > Sorry if this is obvious to everyone else. I'm just trying to understand how this all fits since we're bound to get more UKI-related things in the future.
> 
> Yeah, this is true. This is why I proposed adding a "randomly generated" build identifier to the initrd, which would be paired with the BLS file. Then whichever initrd gets booted it can finds its own correct BLS file in either of the boot directories.

Going along with the idea of a "random generated" build identifier, we could actually just make this a checksum of the diff before we commit and store it as a non-random thing, that way the id is reproducible in builds and it's more verifiable it's correct. We could use something like sha1sum, we want a checksumming algorithm that's really fast. This way we might be able to afford to get away with no fallback flag written by the firmware into the GPT partition table because you would know from this build identifier to keep rolling back commits until the build id matches.

I'm setting up an environment at the moment to test all different kinds of fake upgrades etc. The above works with an "rpm-ostree install" kind of upgrade. But I want to try a normal userspace upgrade, a kernel upgrade etc. So whatever technique we use I can test it covers all bases.

jlebon:
I think I have an idea on how to resolve the race condition here more systematically. The approach is essentially to never change the position of the booted deployment during an update. Let's say you're booted in the default deployment and you're ready to upgrade. Then you would:
- deploy the update in rollback position (if using the CLI, this can be achieved via `ostree admin deploy --not-as-default`)
- do the slot switch

If we're interrupted in between the two steps, we'll just boot right back in the old deployment. As soon as we do the slot switch, we're set to boot into the new deployment.

In the next update, you'd do the opposite (deploy to the pending deployment position).

Essentially, this makes the deployment index not really matter. 0 is always matched to slot A and 1 is always matched to slot B.

Obviously if you do something like `rpm-ostree rollback`, it messes up the order, but it wouldn't be hard to add some sysroot option in libostree that enforces that the booted deployment never changes index too.

ericcurtin:
I am heading on a short break for a few days, but I will be back next week, just to give an update. I am working with the upstream maintainer of this tool:

https://gitlab.com/sdm845-mainline/qbootctl/-/merge_requests

to do slot switching on a real Qualcomm based solution, up to now I'd been testing in single slot mode.

Expect there to be another re-spin of this, thanks for the feedback so far.

cgwalters:
This one could probably benefit from a realtime conversation.  I'm looking at time slots for this now, tentatively thinking May 31 @ 9:30am EST at https://meet.gnome.org/col-hab-hej-e5i

ericcurtin:
I am out next week 29 May-2 June, I spoke to @alexlarsson and we have something in mind that should require no changes to ostree-prepare-root, thanks to the design of composefs, it really simplified things. I should be able to complete in aboot-deploy abstraction requiring minimal to no further changes to ostree.

I'm gonna close this PR, but would be more than happy to have the conversation to keep everyone sync'd up.

Issue #2670, opened 2022-07-08
Add a repo option to auto-prune other deployments (e.g. rollback) when starting upgrade


cgwalters:
In RHCOS we're running up against space constraints https://bugzilla.redhat.com/show_bug.cgi?id=2104619

I think we should support something like
```
[sysroot]
deployments-max=2
```

This would tell ostree to auto-prune the rollback deployment (and others) when starting an upgrade.

travier:
This would also help with an often requested feature where some folks want to keep more than 2 deployments by default.

jmarrero:
I think the intent is to have the number of deployments to keep. So if it's set to 1, you still would temporary balloon to 2 prior rebooting.

jmarrero:
Ahh, I guess that makes sense. But that means the minimum number is 2 right?

dbnicholson:
Bikeshedding, but I think I'd expect the maximum deployments to be the non-ballooned value. The ballooned number of deployments is a temporary implementation detail. As a user it would bug me that I said `deployments-max=2` but there was only 1 deployment under normal circumstances.

Also, when you're in the ballooned situation, one of the deployments is shown as staged. I think it would be reasonable to interpret `deployments-max` as the maximum number of non-staged deployments on disk.

In other words, I think the `deployments-max` minimum can be 1 and that means you'll normally have 1 deployment on disk and temporarily 2 during an upgrade.

travier:
I think the idea is to have `2` mean that we keep 2 deployments on the disk, and when performing an upgrade, just before deploying the new one, we remove the old one to keep ourselves under 2 deployments for disk usage.

dbnicholson:
Right, I thought of that while after I walked away.

They're actually 2 orthogonal concepts to me. Specifying the maximum number of deployments allows you to say you want no rollback deployment or more than 1 rollback deployment. Saying that you want to delete a rollback deployment before upgrading so that the number of deployments is strictly capped is slightly different.

For example, does `deployments-max=2` mean that you should pre-delete the rollback deployment or that you want to allow ballooning to 2 deployments and only have 1 under normal circumstances.

So, I really think there are 2 knobs you want:

* `deployments-max` - The number of active + rollback deployments to keep under normal circumstances.
* `pre-delete-rollback` (can't think of a better name at the moment) - Whether to delete a rollback deployment prior to pulling a new deployment.

[...]

PR #2847, opened 2023-04-13
lib/sysroot-deploy: Add experimental support for automatic early prune

jlebon:
During the early design of FCOS and RHCOS, we chose a value of 384M
for the boot partition. This turned out to be too small: some arches
other than x86_64 have larger initrds, kernel binaries, or additional
artifacts (like device tree blobs). We'll likely bump the boot partition
size in the future, but we don't want to abandon all the nodes deployed
with the current size.[[1]]

Because stale entries in `/boot` are cleaned up after new entries are
written, there is a window in the update process during which the bootfs
temporarily must host all the `(kernel, initrd)` pairs for the union of
current and new deployments.

This patch determines if the bootfs is capable of holding all the
pairs. If it can't but it could hold all the pairs from just the new
deployments, the outgoing deployments (e.g. rollbacks) are deleted
*before* new deployments are written. This is done by updating the
bootloader in two steps to maintain atomicity.

Since this is a lot of new logic in an important section of the
code, this feature is gated for now behind an environment variable
(`OSTREE_EXP_AUTO_EARLY_PRUNE`). Once we gain more experience with it,
we can consider turning it on by default.

This strategy increases the fallibility of the update system since one
would no longer be able to rollback to the previous deployment if a bug
is present in the bootloader update logic after auto-pruning. This is
however mitigated by the fact that the heuristic is opportunistic: the
rollback is pruned *only if* it's the only way for the system to update.

[1]: https://github.com/coreos/fedora-coreos-tracker/issues/1247

Closes: #2670


jlebon:
I still need to add tests for this. Prep patches split out in https://github.com/ostreedev/ostree/pull/2848.

jlebon:
Now with a test!

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2516,6 +2744,12 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
   if (!_ostree_sysroot_ensure_writable (self, error))
     return FALSE;
 
+  /* for now, this is gated on an environment variable */
+  const char *opted_in = getenv ("OSTREE_EXP_AUTO_EARLY_PRUNE");
+  if (opted_in && *opted_in != '\0' && !opts->disable_auto_early_prune &&
+        !auto_early_prune_old_deployments (self, new_deployments, cancellable, error))
```
cgwalters, line 2750: We should also no-op this if there's not a distinct `/boot` filesystem right?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2487,6 +2489,232 @@ write_deployments_finish (OstreeSysroot *self,
...
+      if (!_ostree_sysroot_parse_bootdir_name (bootdir, NULL, &bootcsum))
+        g_assert_not_reached (); /* checked in _ostree_sysroot_list_all_boot_directories() */
+
+      guint64 bootdir_size;
+      g_autofree char *ostree_bootdir = g_build_filename ("ostree", bootdir, NULL);
+      if (!ot_get_dir_size (self->boot_fd, ostree_bootdir, &bootdir_size, cancellable, error))
+        return FALSE;
+
+      /* for our purposes of sizing bootcsums, it's highly unlikely we need a
+       * guint64; cast it down to guint so we can more easily store it */
```
cgwalters, line 2602: Yeah, ghashtable is annoying for this.  Really what we want here is Rust which would transparently optimize this case...

cgwalters (commented):
Gave this a skim, seems sane.

dustymabe:
> During the early design of FCOS and RHCOS, we chose a value of 384M for the boot partition. This turned out to be too small: some arches other than x86_64 have larger initrds, kernel binaries, or additional artifacts (like device tree blobs). We'll likely bump the boot partition size in the future, but we don't want to abandon all the nodes deployed with the current size.[[1](https://github.com/coreos/fedora-coreos-tracker/issues/1247)]
> 
> Because stale entries in `/boot` are cleaned up after new entries are written, there is a window in the update process during which the bootfs temporarily must host all the `(kernel, initrd)` pairs for the union of current and new deployments.
> 
> This patch determines if the bootfs is capable of holding all the pairs. If it can't but it could hold all the pairs from just the new deployments, the outgoing deployments (e.g. rollbacks) are deleted _before_ new deployments are written. This is done by updating the bootloader in two steps to maintain atomicity.
> 
> Since this is a lot of new logic in an important section of the code, this feature is gated for now behind an environment variable (`OSTREE_EXP_AUTO_EARLY_PRUNE`). Once we gain more experience with it, we can consider turning it on by default.

If we lift it out of experimental in the future, but leave it off by
default should we consider removing `EXP` from the name of the var
here?

Also, should we consider allowing the var to be used to turn off the
behavior (i.e. if we flip it to the the default in the future and
someone finds they don't want the behavior).

> 
> This strategy increases the fallibility of the update system since one would no longer be able to rollback to the previous deployment if a bug is present in the bootloader update logic after auto-pruning. This is however mitigated by the fact that the heuristic is opportunistic: the rollback is pruned _only if_ it's the only way for the system to update.

So basically if you have `X`,`Y` -> `Z` and there is a logic problem
in `Y` you can't get back to `X` in order to update to `Z`? Your words
are a bit hard to decipher but this simple example makes it clear to me
(assuming the simple example is correct).

src/libostree/ostree-sysroot-deploy.c
jlebon: Fixed!

jlebon:
> If we lift it out of experimental in the future, but leave it off by default should we consider removing `EXP` from the name of the var here?

The idea with the `EXP` variable is that it's purposely temporary. If it's stabilized but left off by default, I think we'd want this controlled by e.g. a repo config knob instead. That said, I've removed the `EXP` wording since it could make sense even long-term to have a variable override for it to get out of a pickle (given that e.g. repo configs are unmanaged, but systemd dropins aren't).

> Also, should we consider allowing the var to be used to turn off the behavior (i.e. if we flip it to the the default in the future and someone finds they don't want the behavior).

Since it's currently not the default, this would be decided at the time we do decide to flip it to be the default. We could e.g. keep the variable around and slightly tweak the existing checks. Related to this, I've updated the check so that we not only skip it if the variable is defined but empty, but also if it's set to `0`.

> So basically if you have `X`,`Y` -> `Z` and there is a logic problem in `Y` you can't get back to `X` in order to update to `Z`? Your words are a bit hard to decipher but this simple example makes it clear to me (assuming the simple example is correct).

I didn't want to dig too much into this in the commit message because indeed it's a bit subtle and would make it even longer. I linked to https://github.com/ostreedev/ostree/issues/2670#issuecomment-1179341883 instead which with the following comments should help I think.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2516,6 +2755,12 @@ ostree_sysroot_write_deployments_with_options (OstreeSysroot     *self,
   if (!_ostree_sysroot_ensure_writable (self, error))
     return FALSE;
 
+  /* for now, this is gated on an environment variable */
+  const char *opted_in = getenv ("OSTREE_ENABLE_AUTO_EARLY_PRUNE");
```
cgwalters, line 2759: We do have `OSTREE_SYSROOT_OPTS` today which is basically intended for things like this.

So it'd look more like `OSTREE_SYSROOT_OPTS=early-prune`

cgwalters (changes requested):
A few minor changes

src/libostree/ostree-sysroot-deploy.c
jlebon: Yeah, much nicer!

jlebon:
Updated for comments!

cgwalters (approved):

jlebon:
I forgot to update the commit message on this to reference the new environment variable knob (`OSTREE_SYSROOT_OPTS=early-prune`) rather than the initial `OSTREE_ENABLE_AUTO_EARLY_PRUNE` one. I've updated the description, but it'll forever be in git history, lying to readers... :cry:

PR #2848, opened 2023-04-13
Prep patches for automatic early prune

jlebon:
See individual commits.


cgwalters (approved):
Nicely split up patches!

PR #2850, opened 2023-04-26
ci: two fixes

cgwalters:
ci: Turn off errors for deprecated-declarations

Having `-Werror` on in CI only by default has generally worked OK,
but I don't think it's worth trying to immediately scramble to port
when they deprecate APIs.

Motivated in this case by
```
 src/libostree/ostree-fetcher-curl.c: In function 'initiate_next_curl_request':
src/libostree/ostree-fetcher-curl.c:876:3: error: 'CURLOPT_PROTOCOLS' is deprecated: since 7.85.0. Use CURLOPT_PROTOCOLS_STR [-Werror=deprecated-declarations]
  876 |   rc = curl_easy_setopt (req->easy, CURLOPT_PROTOCOLS, (long)(CURLPROTO_HTTP | CURLPROTO_HTTPS | CURLPROTO_FILE));
      |   ^~
```

---

ci: Drop workaround for fedora-release-container

I think this isn't necessary anymore, and is now actively broken
with f38.

---


jmarrero (approved):
/lgtm

PR #2851, opened 2023-04-28
Add .clang-format file which defines coding style as GNU

ericcurtin:
This is particularly useful for those that work on multiple projects
with different styles so they can easily auto-format their code before
submission. Luckily libostree looks like it's pretty standard GNU style
so this file is quite small.

### DIFF
--- a/ci/codestyle.sh
+++ b/ci/codestyle.sh
@@ -1,7 +1,27 @@
 #!/usr/bin/env bash
 # Tests that validate structure of the source code;
 # can be run without building it.
-set -euo pipefail
+set -eo pipefail
+
+if command -v clang-format >/dev/null && [ -n "$GITHUB_ACTOR" ]; then
+    git remote add github_actor https:///github.com/$GITHUB_ACTOR/ostree.git
+    git remote add https_upstream https://github.com/ostreedev/ostree.git
+    git fetch --recurse-submodules=no github_actor &
+    git fetch --recurse-submodules=no https_upstream main &
+    LOG_LINE=$(git show -s --format=%B | grep Merge)
+    wait
+    THIS_COMMIT=$(echo $LOG_LINE | awk '{print $2}')
+    THAT_COMMIT=$(echo $LOG_LINE | awk '{print $4}')
+    git reset --hard $THIS_COMMIT
+    git rebase $THAT_COMMIT
+    
+    echo -n "Formatting with git clang-format... "
+    if ! git clang-format $THAT_COMMIT; then
+        echo "git clang-format $THAT_COMMIT failed; run: please commit the changes"
+        exit 1
+    fi
+    echo "ok"
+fi
 
 # Don't hard require Rust
 if command -v cargo >/dev/null; then
--- a/ci/gh-build.sh
+++ b/ci/gh-build.sh
@@ -27,7 +27,7 @@ set -euo pipefail
 set -x
 
 # First, basic static analysis
-./ci/codestyle.sh
+./ci/codestyle.sh clang-format
 
 NOCONFIGURE=1 ./autogen.sh
 
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,3 @@
+Language: Cpp
+BasedOnStyle: GNU
+ColumnLimit: 99
--- a/ci/gh-install.sh
+++ b/ci/gh-install.sh
@@ -71,6 +71,7 @@ case "$ID" in
             build-essential
             bubblewrap
             ca-certificates
+            clang-format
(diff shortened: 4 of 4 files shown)


ericcurtin:
Example of code formatted using git clang-format/clang-format and this file:

https://github.com/ostreedev/ostree/pull/2844

I find this useful, because when different codebases use different style, you can just write it in whatever style you like and auto-format in the end before submission. Before I was spending a lot of time manually editing before submission to ostree. Didn't realize clang-format had GNU these days baked in, nice!

ericcurtin:
If we wanted we could hook this up to the "Code style" CI build so that all new changes are enforced to use this style, I did that with inotify-tools and never regretted it.

But for me I'm just happy to have this file here.

ericcurtin:
I think the "Debian Testing" failure is just flakey CI

ericcurtin:
I also addressed this comment by @dbnicholson from #2565:

> The other part that seems important to me is to have a workflow that enforces it. Without that people will forget to do it and we'll continue to get commits fixing up formatting.

Although I didn't use a plugin to do it, I try to avoid CI plugins because if you ever have to migrate off GitHub or somebody stops maintaining the plugin, it's a pain.

But if you guys would prefer to use a pre-baked plugin, that's fine with me!

ci/codestyle.sh
```diff
@@ -1,7 +1,27 @@
 #!/usr/bin/env bash
 # Tests that validate structure of the source code;
 # can be run without building it.
-set -euo pipefail
+set -eo pipefail
```
cgwalters, line 4: I'd prefer we keep this and instead...

ci/codestyle.sh
```diff
@@ -1,7 +1,27 @@
...
+
+if command -v clang-format >/dev/null && [ -n "$GITHUB_ACTOR" ]; then
```
cgwalters, line 6: ...use `[ -n "${GITHUB_ACTOR:-}" ]`

ci/codestyle.sh
```diff
@@ -1,7 +1,27 @@
...
+    git fetch --recurse-submodules=no https_upstream main &
+    LOG_LINE=$(git show -s --format=%B | grep Merge)
+    wait
+    THIS_COMMIT=$(echo $LOG_LINE | awk '{print $2}')
+    THAT_COMMIT=$(echo $LOG_LINE | awk '{print $4}')
+    git reset --hard $THIS_COMMIT
+    git rebase $THAT_COMMIT
+    
+    echo -n "Formatting with git clang-format... "
+    if ! git clang-format $THAT_COMMIT; then
```
cgwalters, line 19: What makes this a git subcommand?

ci/gh-build.sh
```diff
@@ -27,7 +27,7 @@ set -euo pipefail
 set -x
 
 # First, basic static analysis
-./ci/codestyle.sh
+./ci/codestyle.sh clang-format
```
cgwalters, line 30: What does this argument do?

cgwalters (commented):
Not opposed but I think it'd actually be better to just try to land the PR to restyle everything at once.

ci/gh-build.sh
ericcurtin: I can remove the "clang-format" arg if needs be. I added it because I felt the ci/codestyle.sh shouldn't do git operations like fetch, rebase, etc. and apply reformats by default. But it's the only way I could thing of, of getting git clang-format to just look at the changes in a PR. If we reformat the whole codebase though, you can simply call clang-format to check the whole codebase.

Another thing I've seen happen, just to be mindful of if we start seeing it, is sometimes you have to do version checks eventually like only check if clang-format version is greater than 12. Because sometimes how it reformats between versions can change slightly, and you get into the situation where so some builds thing the style is ok, and some don't depending on clang-format version.

This is the way I did it in inotify-tools CI, is we only support clang-format greater than 11 because earlier versions reformat a little different:

https://github.com/inotify-tools/inotify-tools/blob/8367bdd1eda71c10ab0a49c694846a622e543807/build_and_test.sh#L80

I didn't reformat the whole codebase in inotify-tools. Like I am proposing here, all new changes to inotify-tools are mandated to correspond with the .clang-format style. It avoids the "git blame" being confusing problem.

ericcurtin:
> Not opposed but I think it'd actually be better to just try to land the PR to restyle everything at once.

Yeah happy to close this

cgwalters:
Closing in favor of https://github.com/ostreedev/ostree/pull/2565 which has https://github.com/ostreedev/ostree/pull/2854 as a preparatory fix series

PR #2854, opened 2023-05-01
Clang format prep

cgwalters:
build-sys: Squash automake conditional warning re `.PHONY`

---

Add clang formatting infrastructure

Prep for reformatting the codebase.

---

lib: clang-format `ostree.h`

In order to make this work, we need to move the autocleanup
definitions after the other headers.

---

tree-wide: Fix various include ordering issues

This fixes the build with `clang-format`.

---


.clang-format
```diff
@@ -0,0 +1,3 @@
+Language: Cpp
+BasedOnStyle: GNU
+ColumnLimit: 99
```
jlebon, line 3: Can we add https://github.com/coreos/rpm-ostree/blob/deb8b3b155abcb546712d7f610f68d032d6e1366/.clang-format#L5 here? (See https://github.com/coreos/rpm-ostree/pull/3604.)

jlebon (approved):

.clang-format
cgwalters: Sure, I synchronized them exactly to also bump the column limit to 100 just for consistency.

jlebon (approved):

PR #2856, opened 2023-05-10
tests/inst: Add xshell and use it in one place

cgwalters:
I've deprecated sh-inline; in the end I think it is better to minimize the amount of bash code we have.  xshell solves the core convenience problem of taking local variables and mapping them to command arguments.

A full port would be nontrivial; this just starts the ball rolling.


jlebon (approved):
Seems sane to me. Should we track the removal of `sh-inline` in an issue so we don't lose it?

cgwalters:
Still hitting https://github.com/ostreedev/ostree/issues/2038

cgwalters:
> Should we track the removal of sh-inline in an issue so we don't lose it?

https://github.com/ostreedev/ostree/issues/2857

PR #2864, opened 2023-05-23
prepare-root: Move sysroot.tmp creation earlier

cgwalters:
Main motivation is prep for composefs in
https://github.com/ostreedev/ostree/pull/2640
In the interest of that, we add a `bool using_composefs` but it's currently always `false`.

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -36,11 +36,16 @@
  * hardlinked chroot() targets are maintained, this one does the equivalent
  * of chroot().
  *
+ * # ostree-prepare-root.service
+ *
  * If using systemd, an excellent reference is `man bootup`.  This
  * service runs Before=initrd-root-fs.target.  At this point it's
  * assumed that the block storage and root filesystem are mounted at
  * /sysroot - i.e. /sysroot points to the *physical* root before
- * this service runs.  After, `/` is the deployment root.
+ * this service runs.  After, `/` is the deployment root, and /sysroot is
+ * the physical root.
+ *
+ * # Running as pid 1
  *
  * There is also a secondary mode for this service when an initrd isn't
  * used - instead the binary must be statically linked (and the kernel
@@ -77,6 +82,9 @@
 #define OSTREE_PREPARE_ROOT_DEPLOYMENT_MSG \
   SD_ID128_MAKE (71, 70, 33, 6a, 73, ba, 46, 01, ba, d3, 1a, f8, 88, aa, 0d, f7)
 #endif
+
+// A temporary mount point
+#define TMP_SYSROOT "/sysroot.tmp"
 
 #include "ostree-mount-util.h"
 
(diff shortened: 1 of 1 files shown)


src/switchroot/ostree-prepare-root.c
```diff
@@ -371,32 +388,22 @@ main (int argc, char *argv[])
...
-        err (EXIT_FAILURE, "failed to MS_MOVE '%s' to '/sysroot.tmp'", deploy_path);
-
       if (mount (root_mountpoint, "sysroot", NULL, MS_MOVE | MS_SILENT, NULL) < 0)
         err (EXIT_FAILURE, "failed to MS_MOVE '%s' to 'sysroot'", root_mountpoint);
 
       if (mount (".", root_mountpoint, NULL, MS_MOVE | MS_SILENT, NULL) < 0)
-        err (EXIT_FAILURE, "failed to MS_MOVE %s to %s", deploy_path, root_mountpoint);
+        err (EXIT_FAILURE, "failed to MS_MOVE %s to %s", TMP_SYSROOT, root_mountpoint);
+
+      if (chdir ("/sysroot") < 0)
```
alexlarsson, line 402: Shouldn't this be using `root_mountpoint`, not `"/sysroot"`. They will mostly be the same, but all the other places use root_mountpoint.

alexlarsson:
Minor comment, but otherwise LGTM

src/switchroot/ostree-prepare-root.c
cgwalters: Yes.  That said, I had to re-read and re-re-read this code and think through things because I realized this change overall was sufficiently large that it had removed my confidence in knowing how things work.

I added some more comments, and updated.

cgwalters:
I find myself wondering though if we shouldn't instead create `ostree-prepare-composefs.c` that is dedicated to just that.  Among other things, there we can not care about the "no-initramfs" path, we can hardcode the sysroot_readonly flag, etc.

But bigger picture also, I think with composefs we should stop doing the hardlinked checkouts at all.  And that also leads into the discussions around having `ostree=1|2` and not checksums, etc., which also drops out the need to call this `resolve_deploy_path()` etc.

Also related to this, I'm thinking that instead of parsing options from the target repo, we should inject a config file into the initramfs (e.g. `/usr/lib/ostree/initramfs/config`).  In fact I'm pretty sure that's what we must do because it chains into the SB story.  And if we do that it's again different from what `ostree-prepare-root.c` is doing today and an argument to make a new file.

alexlarsson:
The current PR setup is kinda incrementally adding composefs. It deploys both styles, and the code picks up whichever works. I agree that a more opinionated version may be better where we *just* do the composefs during the deploy if that is configured.

Don't we still need a partial checkout for the etc merging to work though?

alexlarsson:
Hmm, also it feels like we shouldn't embedd details about how the image is deployed in the initrd, because you may want to have a commit that works in both kinds of deployment.

cgwalters:
> As for ostree-prepare-composefs. What would decide to run this instead of ostree-prepare-root?

With systemd in the initramfs, there's an immense amount of flexibilty.  In the general case we can install a generator which parses an injected config file and chooses whether to activate `ostree-prepare-root.service` or `ostree-prepare-composefs.service`.  Or we can just have a single binary that parses the config itself and dispatches.

I think what seems simplest though is having one binary with two source files, and very early on in `ostree-prepare-root.c` we detect the composefs case and just dispatch to that - explicitly erroring out if we're pid 1.

cgwalters:
@alexlarsson mind adding an official "approve" on this PR?  Branch protection here requires it.

alexlarsson (approved):
ack

PR #2865, opened 2023-05-25
Increase the metadata size limit to 128MB

### LINKED ISSUE #2715: Metadata size limit too low
The metadata size limit is currently [defined](https://github.com/ostreedev/ostree/blob/c6c3c5ae9fbeaaa2207149ed348109400b88314e/src/libostree/ostree-core.h#L38) as 10 MB and is documented as an arbitrary limit. Unfortunately, flathub has hit that limit for the summary file and has had to resort to filtering refs out of it. It's not totally dire since flatpak has moved to its own subsummary format broken up by architecture, but eventually they'll hit the limit on those, too.

The [commit](https://github.com/ostreedev/ostree/commit/47610b45c2ac91d7d9749bc0a1ea5b3150a09a70) points to this [old bug](https://bugzilla.gnome.org/show_bug.cgi?id=725921). As mentioned in the commit message, the 10 MB is arbitrary, so I think it could just be raised to something bigger. However, I also think it would be possible to skip the limit when the remote URL is https:// under the assumption that there's no reasonable way to MITM that connection and TLS authenticates the source.

Obviously, changing the limit will not immediately fix anything since that change would have to roll out to clients, but eventually that would happen.

CC @barthalion

barthalion:
Flathub has hit the 10MB limit in 2022, and we had to drop some CPU architectures from the main summary to subsummaries, effectively cutting off users running too old Flatpak version. Despite that, the main summary containing only x86_64 is already at 7MB. As this is eventually going to happen to subsummaries as well, preemptively bump the limit 12 times.

It takes between 2 and 3 years for a change like this to roll out across Linux distibutions so the best time for this was yesterday.

fixes #2715

### DIFF
--- a/tests/test-pull-large-metadata.sh
+++ b/tests/test-pull-large-metadata.sh
@@ -25,10 +25,11 @@ setup_fake_remote_repo1 "archive"
 
 echo '1..1'
 
-# Overwrite the commit object with 20 M of 
+# Overwrite the commit object with 121 M of zeroes. This is based on the
+# OSTREE_MAX_METADATA_SIZE constant in src/libostree/ostree-core.h
 cd ${test_tmpdir}
 rev=$(cd ostree-srv && ${CMD_PREFIX} ostree --repo=gnomerepo rev-parse main)
-dd if=/dev/zero bs=1M count=20 of=ostree-srv/gnomerepo/objects/$(echo $rev | cut -b 1-2)/$(echo $rev | cut -b 3-).commit
+dd if=/dev/zero bs=1M count=130 of=ostree-srv/gnomerepo/objects/$(echo $rev | cut -b 1-2)/$(echo $rev | cut -b 3-).commit
 
 cd ${test_tmpdir}
 mkdir repo
--- a/src/libostree/ostree-core.h
+++ b/src/libostree/ostree-core.h
@@ -35,7 +35,7 @@ G_BEGIN_DECLS
  * objects). This is an arbitrary number intended to mitigate disk space
  * exhaustion attacks.
  */
-#define OSTREE_MAX_METADATA_SIZE (10 * 1024 * 1024)
+#define OSTREE_MAX_METADATA_SIZE (128 * 1024 * 1024)
 
 /**
  * OSTREE_MAX_METADATA_WARN_SIZE:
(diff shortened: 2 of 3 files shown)


barthalion:
Looks like a single test hardcodes the size to test against the modified variable, I will get that fixed later today.

barthalion:
I don't think I understand how test-pull-large-metadata is supposed to work. It overwrites a commit object with zeroes, so it ends up complaining about a corrupted object, although only if I change its size from 20MB?

dbnicholson:
> I don't think I understand how test-pull-large-metadata is supposed to work. It overwrites a commit object with zeroes, so it ends up complaining about a corrupted object, although only if I change its size from 20MB?

The pull is going to fail regardless because the commit object is corrupted. However, what the test is checking for is that it explicitly fails with a size too big error. That check happens before the commit object is read, so you'd get an `exceeded maximum` error in the case that the object is too big. If it's not too big (> 128MB with your change), then you'll get a `Corrupted commit object` error.

tests/test-pull-large-metadata.sh
```diff
@@ -25,10 +25,11 @@ setup_fake_remote_repo1 "archive"
 
 echo '1..1'
 
-# Overwrite the commit object with 20 M of 
+# Overwrite the commit object with 121 M of zeroes. This is based on the
+# OSTREE_MAX_METADATA_SIZE constant in src/libostree/ostree-core.h
 cd ${test_tmpdir}
 rev=$(cd ostree-srv && ${CMD_PREFIX} ostree --repo=gnomerepo rev-parse main)
-dd if=/dev/zero bs=1M count=20 of=ostree-srv/gnomerepo/objects/$(echo $rev | cut -b 1-2)/$(echo $rev | cut -b 3-).commit
+dd if=/dev/zero bs=1M count=121 of=ostree-srv/gnomerepo/objects/$(echo $rev | cut -b 1-2)/$(echo $rev | cut -b 3-).commit
```
dbnicholson, line 32: Doesn't this need to be greater than 128? 121 MB is within your updated limit and the pull will not error with `exceeded maxiumum size`.

dbnicholson (changes requested):

tests/test-pull-large-metadata.sh
barthalion: Oh, a typo makes sense. That's what I get for multitasking…

cgwalters (dismissed):
Marking as needs work

dbnicholson (approved):

PR #2872, opened 2023-06-02
Composefs followups

cgwalters:
mount-util: Add missing O_CLOEXEC

Seen in review.

---

composefs: Factor out a shared helper for setting error

To keep the error messages consistent.

---

### DIFF
--- a/src/libostree/ostree-repo-composefs.c
+++ b/src/libostree/ostree-repo-composefs.c
@@ -178,6 +178,16 @@ _composefs_write_cb (void *file, void *buf, size_t len)
   return res;
 }
 
+#else /* HAVE_COMPOSEFS */
+
+static gboolean
+composefs_not_supported (GError **error)
+{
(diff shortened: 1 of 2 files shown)


src/switchroot/ostree-prepare-root.c
```diff
@@ -397,6 +413,12 @@ main (int argc, char *argv[])
       if (mount (deploy_path, TMP_SYSROOT, NULL, MS_BIND | MS_SILENT, NULL) < 0)
         err (EXIT_FAILURE, "failed to make initial bind mount %s", deploy_path);
     }
+  else
+    {
+#ifdef USE_LIBSYSTEMD
+      sd_journal_send ("MESSAGE=Mounted composefs (signed: %d)", (int)using_composefs_signature, NULL);
```
alexlarsson, line 419: This is not quite right. using_composefs_signature is only set if we applied the signature this boot, which only happens on the first boot of this deployment.

alexlarsson: Maybe just drop that part, i'm looking at reworking the signature handling.

PR #2874, opened 2023-06-06
commit: fix ostree deployment on 64-bit inode fs

aospan:
This commit addresses a bug that was causing ostree deployment to become corrupted on the large fs, when any package was installed using 'rpm-ostree install'.

In such instances, multiple files were assigned the same inode. For example, the '/home' directory and a regular file 'pkg-get' were assigned the same inode (2147484070), making the deployment unusable.

A root cause analysis was performed, running the process under gdb, which revealed a lossy conversion from guint64 to guint32, for example 6442451366 converted to 2147484070:

(gdb) p name
$10 = 0x7fe9224d2d70 "home"

(gdb) p inode
$73 = 6442451366

(gdb) s
    device=66311, modifier=0x7fe914791840) at
src/libostree/ostree-repo-commit.c:1590

The conversion resulted in entirely independent files potentially receiving the same inode.

The issue was discovered on PoC machine equipped with a large NVME (3.4TB), but the bug can be easily reproduced using `cosa run -m 4000 --qemu-size +3TB', followed by installation of any package using `rpm-ostree install`. The resulting deployment will be unusable due to many files being "corrupted" by the aforementioned issue.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1510,7 +1510,7 @@ scan_loose_devino (OstreeRepo *self, GHashTable *devino_cache, GCancellable *can
  * checksum. */
 static const char *
 devino_cache_lookup (OstreeRepo *self, OstreeRepoCommitModifier *modifier, guint32 device,
-                     guint32 inode)
+                     guint64 inode)
 {
   OstreeDevIno dev_ino_key;
   OstreeDevIno *dev_ino_val;


cgwalters (approved):
Oh wow, ouch.  Nice find.  This bug has been lurking for a long time...I think we've mostly been saved by how most Linux filesystems keep inodes generally below 32 bits unless there's pressure to go higher.

cgwalters:
I want to do a bit of a "risk analysis" on this.  It's important to emphasize that I believe right now that this bug *does not affect* "plain ostree" usage, where the ostree client is fetching objects.

This issue affects "builds" - that's where we do the device/inode caching.  rpm-ostree's local installs here count as a "build".

But, if one was doing an ostree based build system and using `--link-checkout-speedup` e.g. on a filesystem that uses >32 bit inode numbers, this might result in corrupted (i.e. incorrect files) in a resulting commit.

cgwalters:
Thanks so much for finding this.  It turns out that I was wrong...in the ostree-container flow today, we do [use a devino cache](https://github.com/ostreedev/ostree-rs-ext/blob/c2a292cfb5b6c9c8c0f20ad1d457d4655221ed79/lib/src/container/store.rs#L824) to "flatten" container image layers, and so anything using the ostree-container flow is subject to this.

cgwalters:
I wrote a lot of new code to detect this situation (to prove it's happening): https://github.com/ostreedev/ostree-rs-ext/pull/504

Issue #2571, opened 2022-03-25
pull_with_options uses /var/tmp as tmp space, preventing pulling big files during kickstart installs


champtar:
Using libostree/2021.3 (Alma 8.5)
I'm trying to install a tree over https containing big files (~1.5G) using anaconda/kickstart
While anaconda is running `/var/tmp` only has 800 Mo free (`/dev/mapper/live-rw`), and once `pull_with_options` pulls a big file, running `df -h`, I can see / fill up until I get a `No space left on device`.
`lsof /var/tmp` shows a deleted file:
![image](https://user-images.githubusercontent.com/3536755/160044755-6760c98a-2a7c-48bc-a394-c711eb69eb87.png)

`file://` install are working but not `https://` (so either libcurl or the code around it need love)
If possible tmp space should be in the ostree repo dir by default IMO.

As a workaround we can just mount a tmpfs on top of `/var/tmp`:
```
%pre
mount -t tmpfs -o size=100% tmpfs /var/tmp/
%end
```

cgwalters:
> If possible tmp space should be in the ostree repo dir by default IMO.

Yes; hmm, AFAICS this should be the case right now.  This needs a bit more debugging; there's various places where we pass the tempdir down, and it *should* point to the repository tmpdir i.e. `/ostree/repo/tmp`.

champtar:
Do you know if it shipped with 8.7, IE should I try again ?

NeilHanlon:
Heya - Just a note as I am experiencing this as well in the current 8.8 releases with the following ostree/rpm-ostree.

```
ostree-2022.2-6.el8
rpm-ostree-2022.10.115.g15eba7b1-2.el8_8
```

More than happy to try and debug this -- I was looking into the code which performs the download and found my way toward the libGObject interfaces for ostree and rpm-ostree, but would appreciate any pointers you have to where I can look further, @cgwalters.

FYI the versions above are from Rocky Linux 8.8, but I did confirm it occurs w/ RHEL versions, too.

dbnicholson:
I'm pretty sure this is from the fetcher. Unless `OSTREE_FETCHER_REQUEST_LINKABLE` is set (which is not the common case AFAICT), `_ostree_fetcher_tmpf_from_flags` uses `glnx_open_anonymous_tmpfile`, which uses `TMPDIR` or `/var/tmp` as a fallback.

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/2875

[...]

PR #2875, opened 2023-06-06
fetcher: Always open tmpfiles in repo location

cgwalters:
In an installation environment (like a live ISO) we may not have significant space outside of the target installation repository.

There's no reason not to always open a linkable tempfile.  In the future we should fix the pull path to verify the checksum and then just directly link in the object instead of copying.

Closes: https://github.com/ostreedev/ostree/issues/2571

### DIFF
--- a/src/libostree/ostree-fetcher-util.h
+++ b/src/libostree/ostree-fetcher-util.h
@@ -35,14 +35,8 @@ static inline gboolean
 _ostree_fetcher_tmpf_from_flags (OstreeFetcherRequestFlags flags, int dfd, GLnxTmpfile *tmpf,
                                  GError **error)
 {
-  if ((flags & OSTREE_FETCHER_REQUEST_LINKABLE) > 0)
-    {
-      if (!glnx_open_tmpfile_linkable_at (dfd, ".", O_RDWR | O_CLOEXEC, tmpf, error))
-        return FALSE;
-    }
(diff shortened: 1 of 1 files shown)


cgwalters:
(Only compile tested locally)

src/libostree/ostree-fetcher-util.h
```diff
@@ -35,14 +35,8 @@ static inline gboolean
 _ostree_fetcher_tmpf_from_flags (OstreeFetcherRequestFlags flags, int dfd, GLnxTmpfile *tmpf,
                                  GError **error)
 {
-  if ((flags & OSTREE_FETCHER_REQUEST_LINKABLE) > 0)
```
dbnicholson, line 38: It looks like this is a dead flag with this removal. Since this is internal API (I think), can it be removed?

dbnicholson (approved):
Looks good to me. Feel free to pickup the comment or not.

src/libostree/ostree-fetcher-util.h
cgwalters: Yeah, I felt it was clearer to keep the flag even if unused just to clarify the simplicity of the change.

May do some other cleanup though after this.

cgwalters:
This was reverted in https://github.com/ostreedev/ostree/pull/2901/commits/4e61e6f7d0d6aebd6abcdc455ec53164afe39e8d

PR #2877, opened 2023-06-08
Add ostree=aboot for signed Android Boot Images

ericcurtin:
Some kernel images are delivered in a signed kernel + cmdline + initramfs + dtb blob. When this is added to the commit server side, only after this do you know what the cmdline is, this creates a recursion issue. To avoid this, in the case where we have ostree=aboot karg set, create a symlink after deploy to the correct ostree target in the rootfs, as the cmdline can't be malleable and secured client-side at the same time.

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -40,6 +40,9 @@
 #define _OSTREE_SYSROOT_READONLY_STAMP "/run/ostree-sysroot-ro.stamp"
 #define _OSTREE_COMPOSEFS_ROOT_STAMP "/run/ostree-composefs-root.stamp"
 
+#define autofree __attribute__ ((cleanup (cleanup_free)))
+#define steal_pointer(pp) steal_pointer_impl ((void **)pp)
+
 static inline int
 path_is_on_readonly_fs (const char *path)
 {
@@ -79,6 +82,22 @@ out:
   return cmdline;
 }
 
+static inline void
+cleanup_free (void *p)
+{
+  void **pp = (void **)p;
+  free (*pp);
+}
+
+static inline void *
+steal_pointer_impl (void **to_steal)
+{
+  void *ret = *to_steal;
+  *to_steal = NULL;
+
+  return ret;
+}
+
 static inline char *
 read_proc_cmdline_key (const char *key)
 {
@@ -112,6 +131,34 @@ read_proc_cmdline_key (const char *key)
 
   free (cmdline);
   return ret;
+}
+
+static inline char *
+get_aboot_root_slot (void)
+{
+  autofree char *slot_suffix = read_proc_cmdline_key ("androidboot.slot_suffix");
+  if (strcmp (slot_suffix, "_a") == 0)
+    return strdup ("/ostree/root.a");
+  else if (strcmp (slot_suffix, "_b") == 0)
+    return strdup ("/ostree/root.b");
+
+  errx (EXIT_FAILURE, "androidboot.slot_suffix invalid: %s", slot_suffix);
+
+  return NULL;
+}
+
+static inline char *
+get_ostree_target (void)
+{
+  autofree char *ostree_cmdline = read_proc_cmdline_key ("ostree");
+
+  if (!ostree_cmdline)
+    errx (EXIT_FAILURE, "No ostree= cmdline");
+
+  if (strcmp (ostree_cmdline, "aboot") == 0)
+    return get_aboot_root_slot ();
+
+  return steal_pointer (&ostree_cmdline);
 }
 
 /* This is an API for other projects to determine whether or not the
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -246,7 +246,7 @@ main (int argc, char *argv[])
     }
 
   OstreeComposefsMode composefs_mode = OSTREE_COMPOSEFS_MODE_MAYBE;
-  char *ot_composefs = read_proc_cmdline_key ("ot-composefs");
+  autofree char *ot_composefs = read_proc_cmdline_key ("ot-composefs");
   char *composefs_digest = NULL;
   if (ot_composefs)
     {
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -154,11 +154,8 @@ resolve_deploy_path (const char *root_mountpoint)
 {
   char destpath[PATH_MAX];
   struct stat stbuf;
-  char *ostree_target, *deploy_path;
-
-  ostree_target = read_proc_cmdline_key ("ostree");
-  if (!ostree_target)
-    errx (EXIT_FAILURE, "No OSTree target; expected ostree=/ostree/boot.N/...");
+  char *deploy_path;
+  autofree char *ostree_target = get_ostree_target ();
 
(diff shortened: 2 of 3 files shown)


ericcurtin:
So this is #2844 reworked, much simpler

src/switchroot/ostree-mount-util.h
```diff
@@ -114,6 +120,46 @@ read_proc_cmdline_key (const char *key)
   return ret;
 }
 
+static inline char *
+get_aboot_root_slot (void)
+{
+  char __attribute__ ((cleanup (free_char))) *slot_suffix
```
alexlarsson, line 126: I think if we start using cleanup attribute in prepare-root we should add a define for this.

ericcurtin: It's actually recommended in the coding style "Use GCC `__attribute__((cleanup))` wherever possible. ", not sure how many read it, no extra dependencies and is nicer than manual cleanup.

I'll add some sort of define, the real string is long winded.

src/switchroot/ostree-mount-util.h
```diff
@@ -114,6 +120,46 @@ read_proc_cmdline_key (const char *key)
...
+  else if (slot_suffix[1] == 'b')
+    return strdup ("/ostree/root.b");
+
+  errx (EXIT_FAILURE, "androidboot.slot_suffix invalid: %s", slot_suffix);
+
+  return NULL;
+}
+
+static inline char *
+move (char **to_move)
```
alexlarsson, line 139: This is typically called steal_pointer()

src/switchroot/ostree-mount-util.h
```diff
@@ -114,6 +120,46 @@ read_proc_cmdline_key (const char *key)
...
+
+static inline char *
+get_ostree_target (void)
+{
+  char __attribute__ ((cleanup (free_char))) *ostree_cmdline = read_proc_cmdline_key ("ostree");
+
+  if (!ostree_cmdline)
+    errx (EXIT_FAILURE, "No ostree= cmdline");
+
+  if (strcmp (ostree_cmdline, "aboot"))
```
alexlarsson, line 155: Personally i like strcmp calls to explicitly compare with zero. so `== 0` or ` != 0` just because the result really isn't a traditional boolean.

src/switchroot/ostree-prepare-root.c
```diff
@@ -249,7 +246,7 @@ main (int argc, char *argv[])
     }
 
   OstreeComposefsMode composefs_mode = OSTREE_COMPOSEFS_MODE_MAYBE;
-  char *ot_composefs = read_proc_cmdline_key ("ot-composefs");
+  char __attribute__ ((cleanup (free_char))) *ot_composefs = read_proc_cmdline_key ("ot-composefs");
```
alexlarsson, line 249: This is cool, but ideally split up different fixes in separate commits

src/switchroot/ostree-mount-util.h
ericcurtin: move is a C++ism for the same thing, can change to steal_pointer

alexlarsson: Well, it is g_steal_pointer() in glib which is used in the rest of ostree.

alexlarsson:
So, where are the symlinks created?

src/switchroot/ostree-mount-util.h
alexlarsson: Yeah, this is used in the rest of ostree but called `g_autofree`. So, maybe just #define it to `autofree`.

ericcurtin:
So at deploy time on the client-side, we shell out to aboot-deploy, this knows where to symlink to and can create the symlink. Doesn't need a change in this repo

ericcurtin:
All review comments addressed

src/switchroot/ostree-mount-util.h
```diff
@@ -40,6 +40,8 @@
 #define _OSTREE_SYSROOT_READONLY_STAMP "/run/ostree-sysroot-ro.stamp"
 #define _OSTREE_COMPOSEFS_ROOT_STAMP "/run/ostree-composefs-root.stamp"
 
+#define autofree_char char __attribute__ ((__cleanup__ (free_char)))
```
alexlarsson, line 43: I don't think you need the _char part here, or in free_char. This can free anything that malloc returned. I would propose just "autofree" define and "free_ptr()".

ericcurtin:
On a slightly related note, I was thinking of replacing https://github.com/ostreedev/ostree/blob/main/src/libostree/ostree-bootloader-aboot.c with a generic c file that calls a systemd unit, I thought of doing something similar and I actually saw a comment that hinted towards that in the code base here somewhere....

Would mean you don't have to write a load of bolierplate C to shell out to something, it would just become a systemd unit file...

src/switchroot/ostree-mount-util.h
```diff
@@ -114,6 +131,34 @@ read_proc_cmdline_key (const char *key)
   return ret;
 }
 
+static inline char *
+get_aboot_root_slot (void)
+{
+  autofree_char *slot_suffix = read_proc_cmdline_key ("androidboot.slot_suffix");
+  if (strcmp (slot_suffix, "_a") == 0)
+    return strdup ("/ostree/root.a");
```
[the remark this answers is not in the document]
ericcurtin: So we already create links in /ostree:

```
$ ls -ltr /ostree/
total 12
drwxr-xr-x. 3 root root 4096 May 25 17:29 boot.1.1
lrwxrwxrwx. 1 root root    8 May 25 17:29 boot.1 -> boot.1.1
drwxr-xr-x. 7 root root 4096 May 25 17:29 repo
drwxr-xr-x. 3 root root 4096 May 25 17:29 deploy
```

/ostree/deploy/ normally contains osname:

```
ls -ltr /ostree/deploy/
total 4
drwxr-xr-x. 4 root root 4096 May 25 17:29 centos
```

would rather keep in /ostree/ as it's not standardised to store osnames and it's a shorter path.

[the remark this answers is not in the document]
ericcurtin: ostree expects everything in /ostree/deploy to be an osname (centos, fedora)

src/switchroot/ostree-mount-util.h
```diff
@@ -79,6 +81,21 @@ read_proc_cmdline (void)
...
 }
 
+static inline void
+free_char (char **to_free)
+{
+  free (*to_free);
+}
+
+static inline char *
+steal_pointer_char (char **to_steal)
```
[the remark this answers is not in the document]
ericcurtin: Made this generic

src/switchroot/ostree-mount-util.h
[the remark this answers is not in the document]
alexlarsson: makes sense

src/switchroot/ostree-prepare-root.c
ericcurtin: They are separate now

alexlarsson (approved):
LGTM

cgwalters:
I gave this a quick skim and it looks sane.  I do have a vague feeling it might be nicer to try to abstract away from aboot in some way because there's certain to be some alignment as discussed between UKIs and aboot (and potentially other A/B bootloaders) but this is fine.

A procedural note: In this organization we only need one approving reviewer, so in the general case it's not just me that can merge things.  I'm certainly fine if Alex or Dan had merged this too.

PR #2879, opened 2023-06-08
composefs: Change how we do signatures

alexlarsson:
Currently we generate a signature for the actual composefs image, and
then we apply that when we enable fsverity on the composefs
image. However, there are some issues with this.
    
First of all, such a signed fs-verity image file can only be read if
the corresponding puiblic keyring is loaded into the fs-verity
keyring. In a typical secure setup we will have a per-commit key that
is loaded from the initrd. Additionally, the keyring is often sealed
to avoid loading more keys later.
    
This means you can only ever mount (or even look at) composefs images
from the current boot. While this is not a huge issue it is something
of a pain for example when debugging things.
    
Secondly, and more problematic, during a deploy we can't enable
fs-verity on the newly created composefs file, because and at that
point you need to pass in the signature. Unfortunately this will fail
if the matching public key is not in the keyring, which will fail for
similar reasons as the first issue.
    
The current workaround is to *not* enable fs-verity during deploy, but
write the signature to a file. Then the first time the particular
commit is booted we apply the signature to the iamge. This works
around issue two, but not issue one. But it causes us to do a lot of
writes and computation during the first boot as we need to write the
fs-verity merkle tree to disk. It would be much better and robust if
the merkle tree could be written during the deployment of the update
(i.e. before boot).
    
The new apporach is to always deploy an unsigned, but fs-verity
enabled composefs image. Then we create separate files that contain
the expected digest, and a signature of that file. On the first boot
we sign the digest file, and on further boots we can just verify
that it is signed before using it.
    
This fixes issue 1, since all deploys are always readable, and it
makes the workaround for issue 2 much less problematic, as we only
need to change a much smaller file on the first boot.
    
Long term I would like to avoid the first-boot writing totally, and
I've been chatting with David Howells (kernel keyring maintainer) and
he proposed adding a new keyring syscall that verifies a PKCS#7
signature from userspace directly. This would be exactly what
fs-verity does, except we wouldn't have to write the digest to disk
during boot, we would just read the digest file and the signature file
each boot and ask the kernel to verify it.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -649,7 +649,6 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
 #ifdef HAVE_COMPOSEFS
   if (repo->composefs_wanted != OT_TRISTATE_NO)
     {
-      gboolean apply_composefs_signature;
       g_autofree guchar *fsverity_digest = NULL;
       g_auto (GLnxTmpfile) tmpf = {
         0,
@@ -657,11 +656,6 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
       g_autoptr (GVariant) commit_variant = NULL;
 
       if (!ostree_repo_load_commit (repo, revision, &commit_variant, NULL, error))
-        return FALSE;
-
-      if (!ot_keyfile_get_boolean_with_default (repo->config, _OSTREE_INTEGRITY_SECTION,
-                                                "composefs-apply-sig", TRUE,
-                                                &apply_composefs_signature, error))
         return FALSE;
 
       g_autoptr (GVariant) metadata = g_variant_get_child_value (commit_variant, 0);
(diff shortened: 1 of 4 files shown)


src/switchroot/ostree-prepare-root.c
```diff
@@ -183,6 +184,130 @@ resolve_deploy_path (const char *root_mountpoint)
...
+  if ((fd_flags & FS_VERITY_FL) == 0)
+    {
+      close (digest_fd);
+
+      apply_digest_signature (digestfile, sigfile);
+
+      /* Reopen */
+      digest_fd = open (OSTREE_COMPOSEFS_NAME ".digest", O_RDONLY | O_CLOEXEC);
+      if (digest_fd < 0)
+        err (EXIT_FAILURE, "failed to reopen %s", OSTREE_COMPOSEFS_NAME ".digest");
```
cgwalters, line 280: digestfile

cgwalters (approved):

cgwalters:
> we wouldn't have to write the digest to disk during boot, we would just read the digest file and the signature file each boot and ask the kernel to verify it.

That sounds *way* better indeed.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -698,25 +692,33 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
...
-          /* We can't apply the signature during deploy, because the corresponding public key for
-             this commit is not loaded into the keyring. So, we delay fs-verity application to the
-             first boot. */
+      if (!_ostree_tmpf_fsverity (repo, &tmpf, NULL, error))
+        return FALSE;
 
+      if (metadata_composefs && metadata_composefs_sig)
+        {
+          g_autofree char *composefs_digest_path
+              = g_strdup_printf ("%s/.ostree.cfs.digest", checkout_target_name);
```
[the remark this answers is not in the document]
alexlarsson: Well, that define is in ostree-prepare-root.c.

src/switchroot/ostree-prepare-root.c
```diff
@@ -183,6 +184,130 @@ resolve_deploy_path (const char *root_mountpoint)
...
+static char *
+read_signed_digest (const char *digestfile, const char *sigfile)
+{
+  unsigned fd_flags;
+  int digest_fd;
+  unsigned char buf[LCFS_DIGEST_SIZE];
+  char *digest;
+  ssize_t bytes_read;
+
+  digest_fd = open (OSTREE_COMPOSEFS_NAME ".digest", O_RDONLY | O_CLOEXEC);
```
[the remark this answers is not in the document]
alexlarsson: yeah

alexlarsson:
Seems to have hit the same flake again...

cgwalters:
Yeah it's embarrassing, there's a race condition there but I haven't been able to figure it out or even reproduce yet outside of CI.  

src/switchroot/ostree-prepare-root.c
```diff
@@ -183,6 +184,130 @@ resolve_deploy_path (const char *root_mountpoint)
...
+
+      /* Reopen */
+      digest_fd = open (digestfile, O_RDONLY | O_CLOEXEC);
+      if (digest_fd < 0)
+        err (EXIT_FAILURE, "failed to reopen %s", digestfile);
+    }
+
+  /* By now we know its fs-verify enabled, also ensure it is signed
+   * with a key in the keyring */
+  ensure_digest_fd_is_signed (digest_fd);
```
ebiggers, line 285: Why aren't you just storing the signature in the contents of the `digestfile` and verifying it in userspace here?

Unfortunately I wasn't able to get in front of the train wreck that is CONFIG_FS_VERITY_BUILTIN_SIGNATURES fast enough, but in summary it should not be used, as it is not a good way to do signatures.

cgwalters: That's at the end of the commit message, right?

> Long term I would like to avoid the first-boot writing totally, and
I've been chatting with David Howells (kernel keyring maintainer) and
he proposed adding a new keyring syscall that verifies a PKCS#7
signature from userspace directly. This would be exactly what
fs-verity does, except we wouldn't have to write the digest to disk
during boot, we would just read the digest file and the signature file
each boot and ask the kernel to verify it.

Or are you suggesting that it's simpler to just do full userspace verification here and not use the kernel keyring?

ebiggers: Yes, just doing full userspace verification would be much simpler.

cgwalters: Thanks for looking at this by the way!  Conceptually I think this is a composefs-level issue and not an ostree issue so I've moved this to https://github.com/containers/composefs/issues/151

PR #2885, opened 2023-06-14
composefs: Add some basic docs

cgwalters:
Let's describe the state of things at a high level, independent of the tracking issue which has a lot more detail (and hence noise).

This document keeps things at a high level and describes how to enable things today.


jmarrero (approved):
/lgtm

PR #2886, opened 2023-06-15
ostree-fetcher-curl: explicitly use HTTP1.1 when HTTP2 is disabled

daissi:
Currently, trying to disable HTTP2 with `ostree config get` doesn't work as HTTP2 is actually used anyway (see https://gitlab.apertis.org/infrastructure/apertis-issues/-/issues/349 for the context).

The current logic to select the HTTP version to be use relies on the fact that curl don't use by default HTTP2. This assumption seems wrong with recent versions of curl. To fix this option, this PR adds a specific check if HTTP2 is disabled and explicitely switch to HTTP1.1 in this case.

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -879,6 +879,12 @@ initiate_next_curl_request (FetcherRequest *req, GTask *task)
 #endif
     }
 
+  if (self->config_flags & OSTREE_FETCHER_FLAGS_DISABLE_HTTP2)
+    {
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
Code LGTM, just needs a `make clang-format`

daissi:
> Code LGTM, just needs a `make clang-format`

Oups, fixed! thanks

cgwalters (approved):

cgwalters:
flaked on https://github.com/coreos/fedora-coreos-tracker/issues/1511

PR #2891, opened 2023-06-16
lib: Rework composefs metadata, drop custom signatures

cgwalters:
lib: Rework composefs metadata, drop custom signatures

We will be switching to handling signature verification of
the target ostree commit.

---


docs/composefs.md
```diff
@@ -22,12 +22,53 @@ At the current time, integration of composefs and ostree is experimental.
...
+
+## Requirements
+
+The current default composefs integration in ostree does not have any requirements
+from the underlying kernel and filesystem other than having the following
+kernel options set:
+
+- `CONFIG_OVERLAY_FS`
+- `CONFIG_BLK_DEV_LOOP`
+- `CONFIG_EROFS_FS`
```
alexlarsson, line 67: Need config for fs-verity too

alexlarsson (approved):
lgtm

docs/composefs.md
cgwalters: It's not technically a *requirement* right now is what I was getting at.  (Hmm, we should probably have this in upstream composefs actually)

PR #2893, opened 2023-06-19
ci: Add some composefs testing

cgwalters:
tests: Add a sanity check for composefs

Prep for adding some coverage of this flow when booting with
composefs.

---

ci: Sync prow config with rpm-ostree, enable composefs there

I want to gain testing over the composefs path; but without
yet changing the main Jenkins build.  Because we have duplicate/overlapping
CI systems, we can take advantage of this by testing the composefs
flow via Prow.

Sync the Prow build scripts with what's in coreos/rpm-ostree
right now.

Then let's test the composefs flow a bit more e2e there.

---


jmarrero (approved):
lgtm

PR #2904, opened 2023-06-27
ci/prow: Build tests before trying to install

cgwalters:
Our `install` target doesn't depend on `all` unlike the rpm-ostree version; not doing so is arguably better, so let's explicitly build before installing.


cgwalters:
xref https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_release/40485/rehearse-40485-pull-ci-ostreedev-ostree-main-fcos-e2e/1673586518762983424/build-log.txt

PR #2905, opened 2023-06-27
Separate prepare-root static path + link to glib

cgwalters:
Separate prepare-root static path

We should have done this a long time ago.  We don't have any test
coverage for the no-initramfs path, and I think it's not long
term supportable as we want to add more features like composefs.

Particularly now that there's good support for embedding an
initramfs in a kernel image, I see little value in a path for
having custom static linking for this prepare root flow.

That said, we will continue to make a best-effort "it compiles"
attempt to support it.

Fork the "pid 1" prepare root code into a new
`ostree-prepare-root-static.c` file, and drop the runtime conditionals.

We can drop the composefs logic from `-static.c` which ends up
keeping that file much smaller.

A further next step here will be to actually fold the
`prepare-root.c` logic into the main `ostree` binary which we
can then just include in the initramfs.

---

prepare-root: Link to glib

Since we've split off the "prepare root as init" code
into a separate file, we can now use glib to parse
the config file again, which is a lot less hacky.

This is particularly motivated by composefs, where
we want to do more in the initramfs.  Future patches
may also link to parts of libostree.

---

### DIFF
--- a/Makefile-switchroot.am
+++ b/Makefile-switchroot.am
@@ -15,6 +15,12 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library. If not, see <https://www.gnu.org/licenses/>.
 
+ostree_prepare_root_SOURCES = \
+    src/switchroot/ostree-mount-util.h
+ostree_prepare_root_CFLAGS =
+ostree_prepare_root_CPPFLAGS = $(AM_CPPFLAGS)
+ostree_prepare_root_LDADD =
+
 if BUILDOPT_SYSTEMD
 ostree_boot_PROGRAMS += ostree-remount
 else
@@ -23,15 +29,9 @@ else
 check_PROGRAMS += ostree-remount
 endif
 
-ostree_prepare_root_SOURCES = \
-    src/switchroot/ostree-mount-util.h \
-    src/switchroot/ostree-prepare-root.c \
-    $(NULL)
-ostree_prepare_root_CFLAGS =
-ostree_prepare_root_CPPFLAGS = $(AM_CPPFLAGS)
-ostree_prepare_root_LDADD =
-
 if BUILDOPT_USE_STATIC_COMPILER
+ostree_prepare_root_SOURCES += src/switchroot/ostree-prepare-root-static.c
+
 # ostree-prepare-root can be used as init in a system without a populated /lib.
 # To support this use case we need to link statically as we will be unable to
 # locate libc.so at run time if it's not installed in /lib.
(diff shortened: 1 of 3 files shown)


cgwalters:
cc @akiernan who I think uses the static path; we don't have CI upstream for this.

akiernan:
> cc @akiernan who I think uses the static path; we don't have CI upstream for this.

A previous life I'm afraid... though from memory it saved us ~2s on boot since we knew exactly what we were booting on/into to an initramfs was just overhead (and we weren't on the most performant hardware!). @aj-bagwell are you guys still running like this?

cgwalters:
> A previous life I'm afraid... though from memory it saved us ~2s on boot since we knew exactly what we were booting on/into to an initramfs was just overhead (and we weren't on the most performant hardware!). 

Right; but did you investigate minimizing the initramfs?  Did you use a "UKI" or similar approach of embedding the initramfs in the kernel image?

akiernan:
> > A previous life I'm afraid... though from memory it saved us ~2s on boot since we knew exactly what we were booting on/into to an initramfs was just overhead (and we weren't on the most performant hardware!).
> 
> Right; but did you investigate minimizing the initramfs? Did you use a "UKI" or similar approach of embedding the initramfs in the kernel image?

initramfs would be a grandiose term for it... it was basically busybox + a shell script which ended up in a call to ostree-prepare-root (something a whole lot like this https://github.com/intel/intel-iot-refkit/blob/786a4de29c30b47f885d8ad9cb2d110a08919ebd/meta-refkit-core/recipes-ostree/initramfs-framework-ostree/files/ostree). It could certainly have been optimised more, but since literally the only thing we needed to do was call ostree-prepare-root, using the static boot option was just easier.

The initramfs was delivered as part of a U-Boot FIT image, so packaged/signed alongside the kernel, but loaded separately.

cgwalters:
OK.  Well, I tried to do this PR relatively carefully and I *think* the static path will continue to work.

But after this, it will be in a "no new features" mode unless someone makes a strong argument otherwise.

(I doubt after composefs integration there will be many more "features" here anyways, but it's hard to say for sure)

cgwalters:
@dbnicholson any concerns from your end on this?

alexlarsson:
This looks good to me, but does it maybe make sense to build both ostree-prepare-root  and ostree-prepare-root-static if STATIC_COMPILER is true? Then you can make a single binary package that can be used for both approaches. Also, that would make it easier to keep both keep building.

cgwalters:
Hmm.  Yes, I could see that making sense, though it would require packaging changes for users of the static compiler today.  Unsure about the cost/benefit there.

@alexlarsson mind adding an official review/approval here?

alexlarsson (approved):
LGTM

PR #2907, opened 2023-06-27
tests: Source libtest before exiting

cgwalters:
I am not sure why this is failing on older Debian systems, but I'm wildly guessing that something being done in `libtest.sh` is setting up automake in a way that we need.  This is done in other tests.


jmarrero (approved):
/lgtm

Issue #2909, opened 2023-06-27
Ostree-system-generator failed with exit status 1


Freyon77:
Arch x86_64

Every boot error message appear in journalctl. Sometimes error message shows randomly, but not sure what trigger it.

(sd-execu[4034]: /usr/lib/systemd/system-generators/ostree-system-generator failed with exit status 1.

Freyon77:
Affected fedora users too

chrisawi:
I saw a user report this earlier and meant to relay it here, but I got distracted by the flatpak issue.

Since https://github.com/ostreedev/ostree/commit/aa72caffb540114a0e8635ab2990b9c8b0b8f9db, that generator no longer exits 0 when not ostree-booted. The comment describing the intended behavior [is still present](https://github.com/ostreedev/ostree/blob/main/src/switchroot/ostree-system-generator.c#L61-L65):

```
  /* If we're installed on a system which isn't using OSTree for boot (e.g.
   * package installed as a dependency for flatpak or whatever), silently
   * exit so that we don't error, but at the same time work where switchroot
   * is PID 1 (and so hasn't created /run/ostree-booted).
   */
```

Freyon77:
So as I understand it this is intentional and will not be corrected? It would be nice if there was an option to mask the error. Well I guess the only solution is to go back to the previous version

dbnicholson:
@ericcurtin looks like the aboot systemd generator changes in aa72caffb540114a0e8635ab2990b9c8b0b8f9db cause the generator to fail on non ostree booting systems. Can you have a look?

PR #2911, opened 2023-06-28
Fix return value of generator on non-ostree systems

cgwalters:
Commit aa72caffb540114a0e8635ab2990b9c8b0b8f9db
regressed the `ostree-system-generator` on non-ostree systems. If there's no `ostree=` karg, we need to just exit 0.

Closes: https://github.com/ostreedev/ostree/issues/2909
Fixes: aa72caffb540114a0e8635ab2990b9c8b0b8f9db

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -153,7 +153,7 @@ get_ostree_target (void)
   autofree char *ostree_cmdline = read_proc_cmdline_key ("ostree");
 
   if (!ostree_cmdline)
-    errx (EXIT_FAILURE, "No ostree= cmdline");
+    return NULL;
 
   if (strcmp (ostree_cmdline, "aboot") == 0)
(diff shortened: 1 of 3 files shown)


dbnicholson (approved):

dbnicholson:
I don't know what to make of the kola failure. I can't tell if it's transient or related to this change.

cgwalters:
I think that's just an instance of a low-frequency flake we've seen; https://github.com/ostreedev/ostree/pull/2912 will help a little in debugging it hopefully.

cgwalters:
Flaked on https://github.com/coreos/fedora-coreos-tracker/issues/1511

ericcurtin:
This still fails in the Android Boot case, because this regex basically doesn't apply:

`regex = g_regex_new ("^/ostree/boot.[01]/([^/]+)/", 0, 0, NULL);`

but I'm gonna fix it up.

PR #2912, opened 2023-06-29
tests/inst: A few small patches

cgwalters:
tests/transactionality: Port a bit to xshell

This will give us more useful error messages which should
help debug a flake.

---

tests: Drop unused alias

---

tests: Enable mtime test

I think this just accidentally was never enabled.

While looking at the code, add a sleep here to be resilient to
filesystems with only second mtime granularity.

---


cgwalters:
@HuijingHei mind stamping this one?

HuijingHei (approved):
LGTM

Issue #2900, opened 2023-06-23
2023.4 appears to have made Flatpak regress


smcv:
As reported on https://github.com/flatpak/flatpak/issues/5452, multiple Fedora users seem to be seeing Flatpak update failures since their OS was upgraded from libostree 2023.3 to 2023.4. @yuntaz0 reports in https://github.com/flatpak/flatpak/issues/5452#issuecomment-1604140400 that rolling back to a previous OS image resolves this for them; only a few packages were involved in that rollback, of which libostree seems like the most likely cause for a Flatpak regression.

I haven't packaged 2023.4 for my OS (Debian) yet, so I have no particular insight into this myself.

Jan200101:
I've bisected this down to 59a653c9dd3aa7aa5f0b5b49c6bec288c6cd747a, reverting it on `2023.4` fixes this behavior.

dbnicholson:
Interesting. Can you tell exactly where in flatpak this starts to fail? Flatpak does some interesting things with temporary and parent/child repos that might not get good test coverage in ostree.

Jan200101:
I have no experience with all the different glib abstractions so I haven't found anything yet.

The best I have found so far is that it happens somewhere in or after `write_content_object()` since that is where the error prefix `Writing content object` is set.

dbnicholson:
The error comes from [GZlibDecompressor](https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gzlibdecompressor.c#L365). Why that would happen from creating the file in the repo's `tmp` directory rather than in `/var/tmp`... I don't know. @cgwalters any ideas?

It would be interesting to run flatpak with `G_MESSAGES_DEBUG=all` (or maybe just `--verbose --ostree-verbose`) and try to see where the repo is and if there's anything special about it.

Jan200101:
running with both `G_MESSAGES_DEBUG=all` and `--verbose --ostree-verbose` gives

<details>
<summary>Partial Log</summary>

```
F: fetch of 298c2271055571ec60ba06b071c49b7f3d6e1ca3eca3d14beccd8ffd093e8965.file complete
F: _ostree_fetcher_should_retry_request: error: unset, n_retries_remaining: 5
F: Request caught error: Writing content object: Need more input
Installing 1/8… ██                    10%  12.8 MB/s
F: fetch static delta part c000a200f6ec1a27427e6de616dacbfb5d0ed6102bf8e4a868f1aaad5bd73008 complete
[...]

PR #2913, opened 2023-06-29
fetcher: Always open tmpfiles in repo (except on FUSE)

cgwalters:
This reverts commit 4e61e6f7d0d6aebd6abcdc455ec53164afe39e8d and re-instates the fix for ensuring that we download temporary files into the repository location.

However in order to ensure we don't re-introduce
https://github.com/ostreedev/ostree/issues/2900
we detect the case where we're writing to a FUSE mount and keep the prior behavior.

I've verified that this works with flatpak.

Note a downside of this is the change needs to be triplicated across the 3 http backends.

This then again
Closes: https://github.com/ostreedev/ostree/issues/2571


jmarrero (approved):
lgtm

PR #2914, opened 2023-06-29
docs: Update user and group section

cgwalters:
- mention `DynamicUser=yes`
- mention the recent systemd JSON bits
- mention sysusers.doc

And briefly talk about the tradeoffs in these.


cgwalters:
Hm, `test-commit-sign.sh` failed here; that's concerning.  Oh interesting...trying this out locally it seems to be flaky somehow.  Will investigate...

HuijingHei (approved):
/LGTM

cgwalters:
/override ci/prow/fcos-e2e
doc only change

PR #2922, opened 2023-07-06
Implement ed255519 using openssl too

alexlarsson:
This adds an openssl-based implementation to ostree-sign-ed25519.c which is used if libsodium was not manually configured.

The target usedcase for this is: https://github.com/ostreedev/ostree/pull/2921

### DIFF
--- a/src/libostree/ostree-sign-ed25519.c
+++ b/src/libostree/ostree-sign-ed25519.c
@@ -26,8 +26,17 @@
 #include "ostree-sign-ed25519.h"
 #include <libglnx.h>
 #include <ot-checksum-utils.h>
+
 #ifdef HAVE_LIBSODIUM
 #include <sodium.h>
+#define USE_LIBSODIUM
+#else
+
+#if defined(HAVE_OPENSSL)
+#include <openssl/evp.h>
+#define USE_OPENSSL
+#endif
+
 #endif
 
 #undef G_LOG_DOMAIN
@@ -99,12 +108,13 @@ _ostree_sign_ed25519_init (OstreeSignEd25519 *self)
   self->public_keys = NULL;
   self->revoked_keys = NULL;
 
-#ifdef HAVE_LIBSODIUM
+#if defined(USE_LIBSODIUM)
   if (sodium_init () < 0)
     self->state = ED25519_FAILED_INITIALIZATION;
+#elif defined(USE_OPENSSL)
 #else
   self->state = ED25519_NOT_SUPPORTED;
-#endif /* HAVE_LIBSODIUM */
+#endif
 }
 
 static gboolean
@@ -117,7 +127,7 @@ _ostree_sign_ed25519_is_initialized (OstreeSignEd25519 *self, GError **error)
     case ED25519_NOT_SUPPORTED:
       return glnx_throw (error, "ed25519: engine is not supported");
     case ED25519_FAILED_INITIALIZATION:
-      return glnx_throw (error, "ed25519: libsodium library isn't initialized properly");
+      return glnx_throw (error, "ed25519: crypto library isn't initialized properly");
     }
 
   return TRUE;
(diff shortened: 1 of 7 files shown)


alexlarsson:
The debian testing build is failing with:
```
2023-07-06T16:08:14.3477143Z ERROR: tests/test-composefs.sh - missing test plan
2023-07-06T16:08:14.3477941Z ERROR: tests/test-composefs.sh - exited with status 1
```

I wonder what this is about. I didn't really change that...

alexlarsson:
Ok, got everything working. But I don't understand how the composefs test worked before?

src/libostree/ostree-sign-ed25519.c
```diff
@@ -230,10 +251,28 @@ ostree_sign_ed25519_data_verify (OstreeSign *self, GBytes *data, GVariant *signa
...
+#if defined(USE_LIBSODIUM)
+          valid = crypto_sign_verify_detached ((guchar *)g_variant_get_data (child),
+                                               g_bytes_get_data (data, NULL),
+                                               g_bytes_get_size (data), public_key->data)
+                  == 0;
+#elif defined(USE_OPENSSL)
+          EVP_MD_CTX *ctx = EVP_MD_CTX_new ();
+          EVP_PKEY *pkey = EVP_PKEY_new_raw_public_key (EVP_PKEY_ED25519, NULL, public_key->data,
+                                                        OSTREE_SIGN_ED25519_PUBKEY_SIZE);
+          valid = ctx != NULL && pkey != NULL
```
cgwalters, line 263: How about e.g.
```
if (!ctx)
  return glnx_throw (error, "openssl: failed to allocate context");
if (!pkey)
  {
    EVP_MD_CTX_free (ctx);
    return glnx_throw (error, "openssl: Failed to initialize ed5519 key");
  }
```
or so?

alexlarsson:
> OK, awesome! My biggest question here is: Are the signatures from libsodium and openssl interoperable?

Yes, I tried it both ways.

src/libostree/ostree-sign-ed25519.c
alexlarsson: sure

alexlarsson:
And, yes:
PASS: tests/test-pre-signed-pull.sh 1 pre-signed pull

cgwalters:
OK nice work here, I think let's get https://github.com/ostreedev/ostree/pull/2923 in and then rebase this on top and merge.  (Alternatively you could drop the overlapping fixes from the composefs fix from this PR, because that test isn't currently gating, but eh)

src/libostree/ostree-sign-ed25519.c
```diff
@@ -124,40 +141,53 @@ ostree_sign_ed25519_data (OstreeSign *self, GBytes *data, GBytes **signature,
...
+  if (!pkey)
     {
-      return glnx_throw (error, "Not able to sign: fail to sign the object");
+      EVP_MD_CTX_free (ctx);
+      return glnx_throw (error, "openssl: Failed to initialize ed5519 key");
     }
 
-  *signature = g_bytes_new_take (sig, sig_size);
+  size_t len;
+  if (ctx != NULL && pkey != NULL && EVP_DigestSignInit (ctx, NULL, NULL, NULL, pkey)
```
cgwalters, line 170: We can now drop the `ctx != NULL && pkey != NULL` checks.  (In theory maybe `EVP_DigestSignInit` too wants the gerror treatment...eh.  Not a big deal.  We can look at perhaps adding autocleanup wrappers for this stuff later)

src/libostree/ostree-sign-ed25519.c
```diff
@@ -124,40 +141,53 @@ ostree_sign_ed25519_data (OstreeSign *self, GBytes *data, GBytes **signature,
...
+      && EVP_DigestSign (ctx, sig, &len, g_bytes_get_data (data, NULL), g_bytes_get_size (data)))
+    sig_size = len;
+
+  EVP_PKEY_free (pkey);
+  EVP_MD_CTX_free (ctx);
+
+#endif
+
+  if (sig_size == 0)
+    return glnx_throw (error, "Not able to sign: fail to sign the object");
```
cgwalters, line 180: I think just "Failed to sign" is less redundant and sufficient (well, there's a whole thing around openssl error strings we could use right?  But not important to me at least right now since I don't think these things will fail except in OOM right?)

alexlarsson: I guess OOM, or if your openssl build is configured in some weird way maybe without support for some things.

alexlarsson:
Feel free to rebase this when #2923 lands, should work then.

cgwalters (approved):

cgwalters:
BTW nice work here!

PR #2924, opened 2023-07-11
build: Drop `make syntax-check`

cgwalters:
As of lately it emits a ton of errors from `grep` about having `*` at the start of a line, but more generally it's only generally found papercut-style issues that aren't worth carrying a distinct checking system for.


cgwalters:
(It does look like it's getting updates in https://github.com/coreutils/gnulib/blob/master/top/maint.mk but eh...)

PR #2926, opened 2023-07-11
prepare-root: A few cleanups

cgwalters:
Add an internal constant for the composefs image name

Just a minor cleanup.

---

prepare-root: Use otutil and g_print

Now that we link to libotutil (and glib) we don't need separate
handling for conditionalizing on the presence of systemd.

Further, there's no value in `sd_journal_send(MESSAGE=)` over
just printing to stdout.

---

prepare-root: Drop unused verity flag querying

This logic got pushed down into libcomposefs.

---


alexlarsson (approved):
lgtm

alexlarsson:
test-concurrency.py strikes again!

alexlarsson:
Oh, and clang-format

PR #2928, opened 2023-07-13
More prepare-root cleanups

cgwalters:
prepare-root: Use constant for ed25519 signature

Minor cleanup.

---

prepare-root: Add metadata for composefs to `/run/ostree-booted`

Particularly for the signature case, having this metadata
acts as a reliable "proof of execution" of the signature verification
code (as opposed to parsing a log file or so).

Besides that, this is also just a stronger check for "we're using
composefs" instead of checking for "overlayfs on /".

---

remount: Don't overwrite /run/ostree-booted

Since it should always be written in the initramfs.

---

remount: Use new metadata in `/run/ostree-booted` for composefs

Since we now have a generalized more structured way of serializing
state in the initramfs instead of "stamp files", use it for
passing the composefs state.

---

Use /run/ostree-booted metadata for sysroot-ro state passing

Just like we did with composefs, use the new metadata instead
of a "stamp file".

---

### DIFF
--- a/Makefile-switchroot.am
+++ b/Makefile-switchroot.am
@@ -60,8 +60,8 @@ ostree_remount_SOURCES = \
     src/switchroot/ostree-mount-util.h \
     src/switchroot/ostree-remount.c \
     $(NULL)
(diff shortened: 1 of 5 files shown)


alexlarsson (commented):
lgtm

cgwalters:
@alexlarsson you need to officially "approve" btw

alexlarsson (approved):

PR #2929, opened 2023-07-13
prepare-root: Drop dead `pivot_root` code

cgwalters:
I think this was only ever used in the "static pid1" path which is now split out into a separate file.

We always expect that e.g. systemd does the switchroot in the initramfs, so drop this dead code.


ericcurtin:
Was playing around with very similar code to this recently. LGTM

ericcurtin (approved):

PR #2930, opened 2023-07-13
prepare-root: Introduce ostree/prepare-root.conf && sysroot.readonly improvements

cgwalters:
prepare-root: Introduce `ostree/prepare-root.conf`

Using the repository configuration for configuration of this
program was always a bit hacky.

But actually with composefs, we really must validate
the target root *before* we parse anything in it.

Let's add a config file for `ostree-prepare-root` that can live
in the initramfs, which will already have been verified.

In the future we'll also add configuration for composefs here.

We expect OS builders to drop this in `/usr/lib/ostree/prepare-root.conf`,
but system local configuration can live in `/etc`.

---

prepare-root: Default sysroot.readonly=true if composefs

Not because it's logically required or anything, but because
it's just a good idea.

---

prepare-root: Don't parse target root when composefs enabled

We shouldn't load anything from the target root filesystem *before*
verifying its integrity if composefs is enabled.

In effect, we want to force composefs users to migrate to
`/usr/lib/ostree/prepare-root.conf` which lives in the initramfs.
(But because we enable sysroot.readonly=true if composefs is enabled
 too, they don't actually need to)

---

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -73,7 +73,15 @@
 #include <ostree-core.h>
 #include <ostree-repo-private.h>
 
+#include "ot-keyfile-utils.h"
 #include "otcore.h"
+
+// The path to the config file for this binary
+const char *config_roots[] = { "/usr/lib", "/etc" };
+#define PREPARE_ROOT_CONFIG_PATH "ostree/prepare-root.conf"
+
+#define SYSROOT_KEY "sysroot"
+#define READONLY_KEY "readonly"
 
 // The kernel argument we support to configure composefs.
 #define OT_COMPOSEFS_KARG "ot-composefs"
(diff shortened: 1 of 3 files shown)


cgwalters:
Discussion on this is saying that we also need the readonly sysroot stuff in the initramfs, which is an argument for a config file.

cgwalters:
Leaning towards an `/etc/ostree/prepare-root.conf` keyfile that gets installed by our dracut module.

cgwalters:
OK I've reworked this PR a bit to not (yet) be about configuring composefs via keyfile, but instead just the preparatory work around correctly handling `sysroot.readonly` w/composefs.

cgwalters:
Rebased :surfer: for textual (not semantic) merge conflict.

ericcurtin (approved):
LGTM

src/switchroot/ostree-prepare-root.c
```diff
@@ -317,10 +363,18 @@ main (int argc, char *argv[])
...
+   * https://github.com/ostreedev/ostree/pull/1767
    */
-  const bool sysroot_readonly = sysroot_is_configured_ro (root_arg);
+  if (!sysroot_readonly)
+    {
+      sysroot_readonly = sysroot_is_configured_ro (root_arg);
+      // Encourage porting to the new config file
+      if (sysroot_readonly)
+        g_print ("Found legacy sysroot.readonly flag, not configured in %s\n",
+                 PREPARE_ROOT_CONFIG_PATH);
```
alexlarsson, line 388: Note: The fedora readonly migration service will enable this and trigger this warning.

cgwalters: Yeah, we will need to have a long period of migrating things to the new API.  It doesn't actually contain the string "warning".

PR #2931, opened 2023-07-13
man: Add ostree-prepare-root

cgwalters:
Add an overdue man page that describes this.  Prep for also documenting composefs things here.


ericcurtin (approved):
LGTM

PR #2932, opened 2023-07-14
mount: Fix gcc -fanalyzer warning for parsing androidboot.slot_suffix

cgwalters:
If the karg wasn't present, we'd do a NULL deref which is undefined behavior.

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -137,6 +137,9 @@ static inline char *
 get_aboot_root_slot (void)
 {
   autofree char *slot_suffix = read_proc_cmdline_key ("androidboot.slot_suffix");
+  if (!slot_suffix)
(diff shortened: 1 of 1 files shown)


cgwalters:
cc @ericcurtin

ericcurtin (approved):
LGTM, thanks for catching this.

cgwalters:
This repository requires a reviewer with write access to merge PRs; since you've contributed multiple things and will continue to do so (I hope) I've invited you to the organization.  Welcome!

ericcurtin:
This can be my first merge, it's a low-hanging fruit thanks.

Also thanks for making me aware of "gcc -fanalyzer", I must add that to the builds of inotify-tools and other projects. I recently removed some static analysis tools from inotify-tools because of flakiness of false positives. But I try and use as many of the tools with high true positive rates as possible.

ericcurtin (approved):

PR #2934, opened 2023-07-14
build-sys: Enable composefs at *build time* by default

cgwalters:
There's no additional dependencies, and it's a small amount of new code.

The riskiest thing is the changes to ostree-prepare-root, but I believe that things are in a good state now there.

Again, this just enables it at *build time* - it's still off at runtime by default.


ericcurtin (approved):
LGTM

PR #2935, opened 2023-07-14
prepare-root: Refactor composefs config handling

cgwalters:
- Convert the current enum into a struct, using `OtTristate` and two member variables (expected signature and digest)
- Factor out a helper function to parse this config
- Clean up the logging by consistently using `composefs:` as a prefix
- Add more assertions to more strictly verify our runtime state since this is security relevant

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -75,6 +75,9 @@
 
 #include "otcore.h"
 
+// The kernel argument we support to configure composefs.
+#define OT_COMPOSEFS_KARG "ot-composefs"
+
 #define OSTREE_PREPARE_ROOT_DEPLOYMENT_MSG \
   SD_ID128_MAKE (71, 70, 33, 6a, 73, ba, 46, 01, ba, d3, 1a, f8, 88, aa, 0d, f7)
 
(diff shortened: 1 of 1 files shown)


src/switchroot/ostree-prepare-root.c
```diff
@@ -385,54 +413,67 @@ main (int argc, char *argv[])
...
   if (!using_composefs)
     {
-      if (composefs_mode > OSTREE_COMPOSEFS_MODE_MAYBE)
-        err (EXIT_FAILURE, "Failed to mount composefs");
-
       /* The deploy root starts out bind mounted to sysroot.tmp */
       if (mount (deploy_path, TMP_SYSROOT, NULL, MS_BIND | MS_SILENT, NULL) < 0)
         err (EXIT_FAILURE, "failed to make initial bind mount %s", deploy_path);
     }
-  else
```
ericcurtin, line 432: I wonder should we have log levels in prepare-root, "Mounted composefs" would be a nice example of a verbose or debug thing

ericcurtin (approved):

src/switchroot/ostree-prepare-root.c
cgwalters: Yes, fair point.  I did move the "mounted composefs" metadata into the `/run/ostree-booted` in the previous PR where it's introspectable in a more reliable fashion than log messages.  But this one is certainly worthy of a log message...I'll re-add.

PR #2936, opened 2023-07-15
commit: Add `--sign-from-file`

cgwalters:
Passing the private key via a direct command line argument is just a bad idea because it's highly likely to get logged or appear in `ps`.
Spotted in review of work for composefs signatures.

### DIFF
--- a/man/ostree-commit.xml
+++ b/man/ostree-commit.xml
@@ -311,10 +311,22 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 </para></listitem>
 
             </varlistentry>
+
+            <varlistentry>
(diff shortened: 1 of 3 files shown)


ericcurtin (approved):
Good idea, we should also do a OPENSSL_cleanse (or an equivalent): 

https://github.com/openssl/openssl/blob/master/crypto/mem_clr.c

of sensitive buffers also, because the even after the memory is free'd, our of scope, etc. The sensitive data is still going to be in that memory until that memory space is written over again. OPENSSL_cleanse nulls it out to protect against this, if you call it just before a free or before a buffer goes out of scope.

PR #2937, opened 2023-07-15
android-boot: Remove dependency on ostree= karg, use androidboot.slot_suffix=

ericcurtin:
Given the kernel argument androidboot.slot_suffix= is required in Android AB updates [1] there is no need to check for ostree= or ostree=aboot at all in the aboot case. This also ensures if the dependancy on ostree= is removed, the android-boot technique will work regardless.

[1]: https://source.android.com/docs/core/ota/ab/ab_implement

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -134,12 +134,8 @@ read_proc_cmdline_key (const char *key)
 }
 
 static inline char *
-get_aboot_root_slot (void)
+get_aboot_root_slot (const char *slot_suffix)
 {
-  autofree char *slot_suffix = read_proc_cmdline_key ("androidboot.slot_suffix");
-  if (!slot_suffix)
-    errx (EXIT_FAILURE, "Missing androidboot.slot_suffix");
-
   if (strcmp (slot_suffix, "_a") == 0)
     return strdup ("/ostree/root.a");
   else if (strcmp (slot_suffix, "_b") == 0)
@@ -153,15 +149,11 @@ get_aboot_root_slot (void)
 static inline char *
 get_ostree_target (void)
 {
-  autofree char *ostree_cmdline = read_proc_cmdline_key ("ostree");
+  autofree char *slot_suffix = read_proc_cmdline_key ("androidboot.slot_suffix");
+  if (slot_suffix)
+    return get_aboot_root_slot (slot_suffix);
 
-  if (!ostree_cmdline)
-    return NULL;
-
-  if (strcmp (ostree_cmdline, "aboot") == 0)
-    return get_aboot_root_slot ();
-
-  return steal_pointer (&ostree_cmdline);
+  return read_proc_cmdline_key ("ostree");
 }
 
 /* This is an API for other projects to determine whether or not the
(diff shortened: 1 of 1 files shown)


src/switchroot/ostree-mount-util.h
```diff
@@ -134,12 +134,8 @@ read_proc_cmdline_key (const char *key)
 }
 
 static inline char *
-get_aboot_root_slot (void)
+get_aboot_root_slot (const char *slot_suffix)
 {
-  autofree char *slot_suffix = read_proc_cmdline_key ("androidboot.slot_suffix");
```
ericcurtin, line 139: We don't need this recently added EXIT FAILURE now, instead, if androidboot.slot_suffix exists we do the aboot way, or else we do the standard ostree way, or else we return null, which is handled at the caller:

```
ag -A2 -ri get_ostree_target src/switchroot/ostree-system-generator.c src/switchroot/ostree-prepare-root.c src/switchroot/ostree-prepare-root-static.c
src/switchroot/ostree-system-generator.c
66:  autofree char *ostree_target = get_ostree_target ();
67-  if (!ostree_target)
68-    exit (EXIT_SUCCESS);

src/switchroot/ostree-prepare-root.c
115:  autofree char *ostree_target = get_ostree_target ();
116-  if (!ostree_target)
117-    errx (EXIT_FAILURE, "No ostree= cmdline");

src/switchroot/ostree-prepare-root-static.c
119:  autofree char *ostree_target = get_ostree_target ();
120-
121-  if (snprintf (destpath, sizeof (destpath), "%s/%s", root_mountpoint, ostree_target) < 0)
```

src/switchroot/ostree-mount-util.h
```diff
@@ -153,15 +149,15 @@ get_aboot_root_slot (void)
...
-    return NULL;
+  autofree char *slot_suffix = read_proc_cmdline_key ("androidboot.slot_suffix");
+  if (slot_suffix)
+    return get_aboot_root_slot (slot_suffix);
 
-  if (strcmp (ostree_cmdline, "aboot") == 0)
-    return get_aboot_root_slot ();
+  autofree char *ostree_cmdline = read_proc_cmdline_key ("ostree");
+  if (ostree_cmdline)
+    return steal_pointer (&ostree_cmdline);
```
cgwalters, line 158: I think this is just `return read_proc_cmdline_key ("ostree")` now.

cgwalters (approved):
Nice, this is definitely way simpler and better

src/switchroot/ostree-mount-util.h
ericcurtin: Ah yes it is, thanks

ericcurtin:
Re-pushed I also removed steal_pointer , I believe we started linking in glib now, so that function is irrelevant.

cgwalters:
You incorrectly rolled in a change to the composefs git submodule (an easy trap to fall into)
> + echo 'error: Commit message for 79f1ad42bf7be53501372d5887d62087f18097e5 changes a submodule, but does not match regex Update submodule: composefs'

ericcurtin:
@cgwalters I re-pushed without it thanks

PR #2938, opened 2023-07-17
generator: Deduplicate ostree= karg parsing

cgwalters:
Avoid having two copies of a regular expression for parsing the `ostree=` kernel argument.  Because the `ostree-system-generator` binary already has access to the internals because it's implemented in the shared library, expose the sysroot version internally and use that.

Motivated by an attempt to change one of these copies but not the other.


ericcurtin (approved):
LGTM

cgwalters:
OK I decided to just squash the now-unnecessary wrapper function.

jmarrero (approved):

ericcurtin (approved):

PR #2939, opened 2023-07-18
sign-ed25519: Minor cleanups

cgwalters:
tests: Remove dead references to "SEED"

Nothing uses this.

---

sign-ed25519: More verbose errors for invalid length

Add a helper to compare lengths which includes the found
and expected values.  Use it consistently.

The error message "Incorrect" is misleading; one might think
one had the *wrong* key.  "Ill-formed" makes it clearer that
we're not parsing it at all.

---

sign-ed25519: Add some comments for data structure

Like poor man's generics.

---

sign-ed25519: Don't set sk unless we've validated it

The semantics of this function now keep the key assigned
even if we fail to validate it, which is ugly.  Only assign
the key after verifying its length.

---


jmarrero (approved):
lgtm

ericcurtin (approved):

PR #2940, opened 2023-07-18
autofree to g_autofree

ericcurtin:
Now that we link in glib into ostree-prepare-root, etc. We don't need this reimplementation of g_autofree.

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -23,6 +23,7 @@
 
 #include <err.h>
 #include <fcntl.h>
+#include <glib.h>
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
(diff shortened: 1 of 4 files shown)


ericcurtin:
These are real build failures, must fix Debian build

ericcurtin:
So we don't link glib in the static-prepareroot build:

```
/usr/bin/ld: /tmp/ccDewzYu.o: in function `g_autoptr_cleanup_generic_gfree':
ostree-prepare-root-static.c:(.text+0x1f): undefined reference to `g_free'
collect2: error: ld returned 1 exit status
```

maybe this should just be closed...

cgwalters:
Right, we'd have to bifurcate/fork more between the `-static.c` path and the initramfs path.

At a practical level though, `g_malloc == malloc` *always* for us nowadays, so it's fine as is in the current code.

PR #2943, opened 2023-07-19
prepare-root: Drop code mounting `/proc`

cgwalters:
This must be done by the init process here; it was always dead code in the initramfs path.

We keep the copy of this code in the now-forked -static.c.


jmarrero (approved):
lgtm

PR #2944, opened 2023-07-19
prepare-root: Drop more dead code

cgwalters:
Most of this was used for the old composefs signature model.  We now reuse the core signature code and link to glib, so we don't need reimplementations of hex strings and reading files.


jmarrero (approved):
lgtm

PR #2946, opened 2023-07-20
Add an always-on `inode64` feature

cgwalters:
As I (and others) will be backporting the fix in
https://github.com/ostreedev/ostree/pull/2874/commits/de6fddc6adee09a93901243dc7074090828a1912 pretty far, I want a way for sysadmins and OS builders to be able to reliably see when their version of ostree has this fix (Because comparing version numbers isn't portable).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -82,7 +82,10 @@ AM_COND_IF([BUILDOPT_TSAN],
 LT_PREREQ([2.2.4])
 LT_INIT([disable-static])
 
-OSTREE_FEATURES=""
+dnl We have an always-on feature now to signify the fix for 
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):

ericcurtin:
Looks like these are not flakes, tests say no:

image

cgwalters:
Yeah, need to use `$OSTREE` in some of the CI tests to get the uninstalled binary...for some reason we're not overrding `PATH` properly?  Kind of a bug.  The whole "test things uninstalled" is really a mess.  But anyways this should be right now.

cgwalters:
Well now I'm worried that this is somehow turning up "some CI tests aren't testing the code in the PR" which would be very exciting...  I think it may relate to something in the "distcheck" flow and us not having `$PATH` set correctly.

Anyways, let's just get the feature in...

PR #2948, opened 2023-07-22
composefs: Use lowerdir in /run

cgwalters:
I just noticed that this was another constant string duplicated between prepare-root.c and libostree-1.so, and I went to make it a common `#define` in libotcore.la.

But then I thought "it's ugly to have this directory mixed into the deployment namespace" because in some theoretical world it could also be in the ostree commit, which would cause weird behavior.

I think this is transient state that is better in `/run`, so move it there.


ericcurtin (approved):
LGTM

ericcurtin:
I know sometimes we keep transient bits in "/var/tmp/" and sometimes we keep them in "/run" do we have a clear distinction on when we use "/run" vs "/var/tmp/"? Just curious...

cgwalters:
Two things.  First, at this point, `/var` might not be mounted (it could be mounted in the real root via `/etc/fstab` for example).

Second, `/var` is commonly a *persistent* filesystem, whereas we really want the semantic that this is transient data tied to a particular boot, which is where `/run` is correct.

Finally and related to these, the best way to pass state from the initramfs to the real root is via `/run`.

PR #2949, opened 2023-07-23
kernel-args: Move private functions out of public header

cgwalters:
External users can't call `_` prefixed APIs, so move them
to a `-private.h`.

---


ericcurtin (approved):
LGTM

PR #2950, opened 2023-07-23
generator: Some cleanup

cgwalters:
generator: Stop creating `/run/ostree-booted`

This must have always been dead code.  We're trying to iterate
towards a place where it's only `ostree-prepare-root.c` which
parses the `ostree=` kernel argument, and canonically sets up
`/run/ostree-booted`.

---

src/generator: Move all logic into libostree-1.so

This pushes down the code for parsing the `ostree=` cmdline
in the generator into code that's part of libostree-1.so.

This is prep for using logic shared in libotcore.la.

But in general it's just cleaner to also keep the binary
entrypoint to just be a trampoline into the C library.

---


ericcurtin (approved):
LGTM

PR #2951, opened 2023-07-24
sysroot: Add a bit more error prefixing

cgwalters:
While an error message I saw was pretty clear, this would be even more explicit.


jmarrero (approved):
lgtm

ericcurtin (approved):
LGTM

PR #2952, opened 2023-07-25
tree-wide: Consistently `(void)g_variant_lookup()`

cgwalters:
Coverity warns when we're checking the return value in most-but-not-all instances.  The code is correct in these instances; we're initializing the values to defaults.  So add a `(void)` cast like we are doing in many other places.


ericcurtin:
Failed on style, grrr, just needs a quick "git clang-format -f HEAD~"

ericcurtin (approved):

PR #2954, opened 2023-07-26
Harden gvariant get data

cgwalters:
core, switchroot: Harden a bit against `g_variant_get_data() == NULL`

I'm not totally sure this is the cause of
https://bugzilla.redhat.com/show_bug.cgi?id=2217401
but analyzing the code a bit it seems the most likely.

---

checksum-utils: Add an assertion that `buf != NULL`

Another hardening against https://bugzilla.redhat.com/show_bug.cgi?id=2217401

---


ericcurtin (approved):
LGTM

PR #2956, opened 2023-07-27
deploy: Be way more verbose about what we're doing

cgwalters:
This will help us debug bugs like https://issues.redhat.com/browse/OCPBUGS-15917 in the future.


ericcurtin (approved):
LGTM, reviewing from phone, not saying it's an issue or anything but when you leave links to issues.redhat.com in the commit message, it's not viewable without logging in.

PR #2957, opened 2023-07-28
tests/destructive: Turn off global sync()

cgwalters:
Let's verify that things work with that off, as they should.

Previously:
https://github.com/ostreedev/ostree/pull/2519/commits/cb731294837736e957ee595ce11ab115277dbb36 "deploy: Add a 5s max timeout on global filesystem sync()"

But we may still have problems even with that, see https://issues.redhat.com/browse/OCPBUGS-15917
where it might be that even a thread doesn't work because we're locked in the kernel.


ericcurtin (approved):
LGTM

Issue #2868, opened 2023-05-31
Want an option for transient /etc


alexlarsson:
For the automotive usecase, and for the general extension of the composefs work (https://github.com/ostreedev/ostree/issues/2867) it would be nice to have a mode where /etc is mounted as an overlayfs on top of /usr/etc from the commit and a tmpfs directory (say in /run). This way you can have a system that is validated by composefs, where you cannot persist system changes between boots.

In theory this should not be very complex, but my initial experiments ran into some issues with selinux contexts which need to be resolved. 

In addition, there are some files that we really want to per persist in /etc that are host machine specific. In particular, the `machine-id` file.  This would have to be solved by copying in it from some persistent location during early boot.

Additionally, and vaguely related:  

In the automotive project we are interested in being able to extend the system dynamically, but in a controlled fashion. In particular, we want to install container images and add systemd service files, but only for a trusted set of apps. Such service files would have to be persisted in /etc/systemd/system, but allowing files to be added there is very risky. The plan we have for this is to have a systemd generator that copies these files from a persisted location, after having verified that they are all signed (for example, using fs-verity signatures) against a trusted key. 

For full chain of trust, this also needs a container image backend using composefs, which is also in the works.

cgwalters:
Some WIP for this in https://github.com/ostreedev/ostree/pull/2970

cgwalters:
This was fixed by https://github.com/ostreedev/ostree/pull/3062

PR #2958, opened 2023-07-29
deploy: Support an empty `/etc` and populated `/usr/etc`

cgwalters:
In preparation for support for a transient `/etc`: https://github.com/ostreedev/ostree/issues/2868
particularly in combination with composefs.

Basically it's just much more elegant if we can directly mount an overlayfs on the *empty* `etc` directory, using `usr/etc` as the lower.

In the composefs case, we'd have to mount the composefs overlayfs itself writable (and call `mkdir`) *just* so we can make that empty `etc` directory which is ugly.


ericcurtin (approved):
LGTM

cgwalters:
@alexlarsson 
I didn't understand that code at all and I meant to circle back to it.

I find it confusing to have these hardcoded directories.  Why don't we just require them to be in the ostree commit as we were doing before?

PR #2962, opened 2023-08-01
os-init: Create a mount namespace

cgwalters:
Today on anything using readonly sysroot `os-init` fails, because we don't create a mount namespace if the `UNLOCKED` flag is specified because we assume it's a readonly operation.

Since technically this is a mutation, let's just lock the sysroot and use the tested path.


ericcurtin (approved):
LGTM

Issue #2965, opened 2023-08-02
Add `ostree admin set-next-boot`


cgwalters:
We should have a CLI verb that allows arbitrary changes to the bootloader entries.  Today we only have `ostree admin undeploy` etc.  

Maybe the simplest here is to offer `ostree admin set-next-boot 3` to start?

PR #2966, opened 2023-08-02
Add `admin set-default`

cgwalters:
A core underlying primitive in the C library is the ability to arbitrarily reorder bootloader entries.

Let's expose the basic functionality here with the ability to pick an arbitrarily deployment for the next boot.

Closes: https://github.com/ostreedev/ostree/issues/2965


ericcurtin (approved):
LGTM

PR #2967, opened 2023-08-02
More fully drop `trivial-httpd` entrypoint

cgwalters:
It's just part of the tests and we should no longer support `ostree trivial-httpd`.

This is a followup cleanup to previous work.


ericcurtin (approved):
LGTM

PR #2968, opened 2023-08-03
deploy: Remove global `sync` by default

cgwalters:
Our previous change here was not actually sufficient for the ceph case, because what (I think) is happening is that our other `syncfs()` invocation is getting blocked on some kernel mutexes that are used in `sync`, and that's causing the process to fully block.

We should not be dependent on a full filesystem `sync`, only on the sync of the sysroot and boot filesystems.

Anyone who *does* want this behavior could inject an override for `ostree-finalize-staged.service` that overrides `ExecStop` to add a run of `sync`.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1634,43 +1634,7 @@ typedef struct
 {
   guint64 root_syncfs_msec;
   guint64 boot_syncfs_msec;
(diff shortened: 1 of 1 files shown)


cgwalters:
OK I think we should do this PR at some point, but some more investigation here narrowed in on https://github.com/ostreedev/ostree/pull/2969

jmarrero (approved):
lgtm

PR #2970, opened 2023-08-04
WIP: transient /etc

cgwalters:


cgwalters:
This currently depends on https://github.com/ostreedev/ostree/pull/2958 (and changing how we build ostree images to include an empty `/etc` directory in addition to `/usr/etc`)

But, we can and probably should sever that dependency by detecting the missing `etc` and instead making the whole `sysroot` an overlayfs and making the `etc` directory there.

src/switchroot/ostree-prepare-root.c
```diff
@@ -326,6 +327,34 @@ load_composefs_config (GError **error)
   return g_steal_pointer (&ret);
 }
 
+static gboolean
+find_etc (const char **out_path, GError **error)
+{
+  // Look for /etc
+  if (fstatat (AT_FDCWD, "etc") == 0)
```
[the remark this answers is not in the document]
alexlarsson: If using plain glnx_fstatat you need to use g_error_matches(error, ...), not errno. Easiest fix is to use 
glnx_fstatat_allow_noent() which keeps errno for ENOENT.

src/switchroot/ostree-prepare-root.c
```diff
@@ -569,18 +598,58 @@ main (int argc, char *argv[])
         }
     }
 
+  g_autofree char *etc_config = NULL;
+  if (!ot_keyfile_get_value_with_default (config, SYSROOT_KEY, ETC_KEY, "persistent", &etc_config,
+                                          error))
+    errx (EXIT_FAILURE, "failed to parse %s.%s: %s", SYSROOT_KEY, ETC_KEY, error->message);
+  bool etc_transient = FALSE;
```
alexlarsson, line 605: Weird mix of `FALSE` and `false`/`true`.

src/switchroot/ostree-prepare-root.c
```diff
@@ -569,18 +598,58 @@ main (int argc, char *argv[])
...
-      if (mount (TMP_SYSROOT "/etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_REMOUNT | MS_SILENT,
-                 NULL)
-          < 0)
-        err (EXIT_FAILURE, "failed to make writable /etc bind-mount at /sysroot.tmp/etc");
+      if (etc_transient)
+        {
+          /* Do we have a persistent overlayfs for /usr?  If so, mount it now. */
+          g_autofree char *etc_ovldir
+              = g_build_filename (OTCORE_RUN_OSTREE_PRIVATE, "etc-transient", NULL);
+          if (mkdirat (AT_FDCWD, etc_ovldir, 0700) < 0)
```
alexlarsson, line 628: Did you boot test this? Because when I tried something like this the selinux context on the initramfs at this point was uninitialized, which was carried over into boot for the overlayfs lowerdir, and without `etc_t` selinux context on `/tmp` a bunch of stuff broke at boot.

src/switchroot/ostree-prepare-root.c
```diff
@@ -569,18 +598,58 @@ main (int argc, char *argv[])
...
+            err (EXIT_FAILURE, "Failed to create %s", etc_ovldir);
+          g_autofree char *upper = g_build_filename (etc_ovldir, "upper", NULL);
+          g_autofree char *work = g_build_filename (etc_ovldir, "work", NULL);
+
+          const char *etc_lower = NULL;
+          if (!find_etc (&etc_lower, &error))
```
alexlarsson, line 634: Isn't this looking in the wrong place (root) for the etc dir? Shouldn't it look in TMP_SYSROOT?

alexlarsson: Actually, not in sysroot i guess, but deploy dir?

alexlarsson: Ah, which is CWD, so this is ok.

raballew: as you already said the deploy dir is the cwd so this should be correct.

cgwalters:
Closing in favor of https://github.com/ostreedev/ostree/pull/2972

PR #2972, opened 2023-08-07
WIP: transient /etc

raballew:
Working copy of https://github.com/ostreedev/ostree/pull/2970

### DIFF
--- a/man/ostree-prepare-root.xml
+++ b/man/ostree-prepare-root.xml
@@ -114,6 +114,10 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <listitem><para>A boolean value; the default is <literal>false</literal>.  If this is set to <literal>true</literal>, then the <literal>/sysroot</literal> mount point is mounted read-only.</para></listitem>
             </varlistentry>
             <varlistentry>
+                <term><varname>sysroot.etc</varname></term>
+                <listitem><para>A string value; the default is <literal>persitent</literal>.  If this is set to <literal>transient</literal>, then the <literal>/sysroot</literal> mount point is mounted transiently i.e. a non-persistent location.</para></listitem>
+            </varlistentry>
+            <varlistentry>
                 <term><varname>composefs.enabled</varname></term>
                 <listitem><para>This can be <literal>yes</literal>, <literal>no</literal>. <literal>maybe</literal> or
                 <literal>signed</literal>. The default is <literal>maybe</literal>.  If set to <literal>yes</literal> or
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -84,6 +84,7 @@ const char *config_roots[] = { "/usr/lib", "/etc" };
 
 #define SYSROOT_KEY "sysroot"
 #define READONLY_KEY "readonly"
+#define ETC_KEY "etc" // Possible values = "persistent" "transient"
 
 #define COMPOSEFS_KEY "composefs"
 #define ENABLED_KEY "enabled"
@@ -349,6 +350,29 @@ load_composefs_config (GKeyFile *config, GError **error)
   return g_steal_pointer (&ret);
 }
 
+static gboolean
+copy_selinux_context (const char *src_path, const char *dst_path, GError **error)
+{
+  ssize_t bytes_read, real_size;
+
+  if (TEMP_FAILURE_RETRY (bytes_read = lgetxattr (src_path, "security.selinux", NULL, 0)) < 0)
+    {
+      if (errno == ENODATA || errno == ENOTSUP)
+        return TRUE; /* no selinux context, we're done */
+      return glnx_throw_errno_prefix (error, "lgetxattr(security.selinux)");
+    }
+
+  g_autofree guint8 *buf = g_malloc (bytes_read);
+  if (TEMP_FAILURE_RETRY (real_size = lgetxattr (src_path, "security.selinux", buf, bytes_read))
+      < 0)
+    return glnx_throw_errno_prefix (error, "lgetxattr(security.selinux)");
+
+  if (lsetxattr (dst_path, "security.selinux", buf, real_size, 0) < 0)
+    return glnx_throw_errno_prefix (error, "lsetxattr(security.selinux)");
+
+  return TRUE;
+}
+
 int
 main (int argc, char *argv[])
 {
@@ -582,18 +606,71 @@ main (int argc, char *argv[])
         }
     }
 
+  g_autofree char *etc_config = NULL;
+  if (!ot_keyfile_get_value_with_default (config, SYSROOT_KEY, ETC_KEY, "persistent", &etc_config,
+                                          &error))
+    errx (EXIT_FAILURE, "failed to parse %s.%s: %s", SYSROOT_KEY, ETC_KEY, error->message);
+  bool etc_transient = false;
+  if (g_str_equal (etc_config, "persistent"))
+    etc_transient = false;
+  else if (g_str_equal (etc_config, "transient"))
+    etc_transient = true;
+  else
+    errx (EXIT_FAILURE, "Invalid %s.%s: %s", SYSROOT_KEY, ETC_KEY, etc_config);
+
+  // In theory these could be distinct, but no reason to try to support it.
+  if (etc_transient && !sysroot_readonly)
+    errx (EXIT_FAILURE, "Must specify %s.%s for %s.%s=transient", SYSROOT_KEY, READONLY_KEY,
+          SYSROOT_KEY, ETC_KEY);
+
   /* Prepare /etc.
    * No action required if sysroot is writable. Otherwise, a bind-mount for
    * the deployment needs to be created and remounted as read/write. */
   if (sysroot_readonly || using_composefs)
     {
-      /* Bind-mount /etc (at deploy path), and remount as writable. */
-      if (mount ("etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_SILENT, NULL) < 0)
-        err (EXIT_FAILURE, "failed to prepare /etc bind-mount at /sysroot.tmp/etc");
-      if (mount (TMP_SYSROOT "/etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_REMOUNT | MS_SILENT,
-                 NULL)
-          < 0)
-        err (EXIT_FAILURE, "failed to make writable /etc bind-mount at /sysroot.tmp/etc");
+      if (etc_transient)
+        {
+          /* Do we have a persistent overlayfs for /usr?  If so, mount it now. */
+          g_autofree char *etc_ovldir
+              = g_build_filename (OTCORE_RUN_OSTREE_PRIVATE, "etc-transient", NULL);
+          g_autofree char *upper = g_build_filename (etc_ovldir, "upper", NULL);
+          g_autofree char *work = g_build_filename (etc_ovldir, "work", NULL);
+
+          if (mkdirat (AT_FDCWD, etc_ovldir, 0700) < 0)
+            err (EXIT_FAILURE, "Failed to create %s", etc_ovldir);
+          if (mkdirat (AT_FDCWD, upper, 0755) < 0)
+            err (EXIT_FAILURE, "Failed to create %s", upper);
+          if (mkdirat (AT_FDCWD, work, 0755) < 0)
+            err (EXIT_FAILURE, "Failed to create %s", work);
+
+          g_autofree char *etc_ovl_options = g_strdup_printf ("lowerdir=%s,upperdir=%s,workdir=%s",
+                                                              TMP_SYSROOT "/usr/etc", upper, work);
+          if (mount ("overlay", TMP_SYSROOT "/etc", "overlay", MS_SILENT, etc_ovl_options) < 0)
+            err (EXIT_FAILURE, "failed to mount transient etc overlayfs");
+
+          g_autoptr (GError) local_error = NULL;
+          if (!copy_selinux_context (TMP_SYSROOT "/usr/etc", TMP_SYSROOT "/etc", &local_error))
+            err (EXIT_FAILURE, "failed to copy /usr/etc selinux label: %s", local_error->message);
+
+          /* We make ovldir read-only to avoid it being relabeled to
+           * var_run_t when /run is relabeled */
+          if (mount (etc_ovldir, etc_ovldir, NULL, MS_BIND | MS_SILENT, NULL) < 0)
+            err (EXIT_FAILURE, "failed to bind mount (class:readonly) %s", etc_ovldir);
+          if (mount (etc_ovldir, etc_ovldir, NULL, MS_BIND | MS_REMOUNT | MS_RDONLY | MS_SILENT,
+                     NULL)
+              < 0)
+            err (EXIT_FAILURE, "failed to bind mount (class:readonly) %s", etc_ovldir);
+        }
+      else
+        {
+          /* Bind-mount /etc (at deploy path), and remount as writable. */
+          if (mount ("etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_SILENT, NULL) < 0)
+            err (EXIT_FAILURE, "failed to prepare /etc bind-mount at /sysroot.tmp/etc");
+          if (mount (TMP_SYSROOT "/etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_REMOUNT | MS_SILENT,
+                     NULL)
+              < 0)
+            err (EXIT_FAILURE, "failed to make writable /etc bind-mount at /sysroot.tmp/etc");
+        }
     }
 
   /* Prepare /usr.


alexlarsson:
Actually, why are we using `usr/etc` as the lower anyway? Shouldn't we be using `TMP_SYSROOT/usr/etc`, which is the copy that is (possibly) on the composefs mount, rather than the one in the deploy dir?

alexlarsson:
Booted this and got:
```
[    5.063620] ostree-prepare-root[996]: Loading /etc/ostree/prepare-root.conf
[    5.066518] ostree-prepare-root[996]: sysroot.readonly configuration value: 1 (fs writable: 1)
[    5.068433] ostree-prepare-root[996]: composefs: Mounting with no digest or signature check
[    5.072483] ostree-prepare-root[996]: Resolved OSTree target to: /sysroot/ostree/deploy/fedora-coreos/deploy/3fe9ea20c2e61e182502cc32d34bb93214a81deacbfbd08375f7ae3f12df9e12.0
[    5.123628] ostree-prepare-root[996]: ostree-prepare-root: Failed to find etc: failed to stat etc: Success
[    4.943158] ostree-prepare-root[996]: ostree-prepare-root: Failed to find etc: failed to stat etc: Success
[    5.128123] ostree-prepare-root[996]: composefs: mounted successfully
[    5.136962] systemd[1]: ostree-prepare-root.service: Failed with result 'exit-code'.
[    5.142865] systemd[1]: Failed to start ostree-prepare-root.service - OSTree Prepare OS/.
```

alexlarsson:
Fixing that I got:
```
[    5.222235] ostree-prepare-root[1005]: sysroot.readonly configuration value: 1 (fs writable: 1)
[    5.224991] ostree-prepare-root[1005]: composefs: Mounting with no digest or signature check
[    5.227061] ostree-prepare-root[1005]: Resolved OSTree target to: /sysroot/ostree/deploy/fedora-coreos/deploy/b0a3b28af0e50d195a87b4fcb32427f180eb837dbec83f91ffc217ab44cfb921.0
[    5.268330] overlayfs: empty lowerdir
[    5.269682] overlayfs: failed to resolve '/run/ostree/.private/etc-transient/upper': -2
[    5.273042] ostree-prepare-root[1005]: ostree-prepare-root: failed to mount transient etc overlayfs: No such file or directory
[    5.091384] ostree-prepare-root[1005]: ostree-prepare-root: failed to mount transient etc overlayfs: No such file or directory
[    5.277826] ostree-prepare-root[1005]: composefs: mounted successfully
[    5.282528] systemd[1]: ostree-prepare-root.service: Main process exited, code=exited, status=1/FAILURE
[    5.284302] systemd[1]: ostree-prepare-root.service: Failed with result 'exit-code'.
[    5.293295] systemd[1]: Failed to start ostree-prepare-root.service - OSTree Prepare OS/.
```

raballew:
> I don't quite understand the handling of `/etc` here.
> 
> In the typical, current, case, an ostree commit will contain `/usr/etc`, which during deploy will end up in `$deploydir/usr/etc` (as well as in the composefs image). During `prepare_deployment_etc()` we extract this into `usr/etc`, and then we copy it back to `etc` . Then, during `merge_configuration_from()`, we take `diff usr/etc etc` on the old deployement, and apply the result to the new `etc`.
> 
> Given the code in this MR, if transient etc is enabled, it always prefers the `etc` directory in the deploy dir, over the `usr/etc` directory. This means that if you persist anything in `/sysroot/ostree/deploy/..../$previous_commit/etc, then it will be merged into the new etc dir, which will be used as the lower dir in the overlayfs, and thus be persisted to the new boot.
> 
> Isn't the point of this all that with transient etc, we only _ever_ make `/etc` an overlayfs with `/usr/etc` from the composefs and a tmpfs dir? Thats what I think of when I hear transient anyway.

I think the answer to this partially is given by https://github.com/ostreedev/ostree/pull/2970#issuecomment-1665539676 which implies that there can be and empty /etc which mean find_etc will return /usr/etc (or NULL).

alexlarsson:
> > Isn't the point of this all that with transient etc, we only _ever_ make `/etc` an overlayfs with `/usr/etc` from the composefs and a tmpfs dir? Thats what I think of when I hear transient anyway.
> 
> I think the answer to this partially is given by [#2970 (comment)](https://github.com/ostreedev/ostree/pull/2970#issuecomment-1665539676) which implies that there can be an empty /etc which mean find_etc will return /usr/etc (or NULL on error).

I think that quote is just confused. We always create an /etc in the composefs image (as per comment  https://github.com/ostreedev/ostree/pull/2958#issuecomment-1668073965 ), so there is no need for that commit at all from what i can see.

raballew:
If thats the case, let me drop the find_etc function and use TMP_SYSROOT/etc instead.

alexlarsson:
With this version (squashed back to one commit): 
https://github.com/alexlarsson/ostree/commits/prepare-root-transient-etc

I get it to boot and start most services, including with selinux enabled.
However, there is one failure remaining:

```
× rpm-ostreed.service - rpm-ostree System Management Daemon
     Loaded: loaded (/usr/lib/systemd/system/rpm-ostreed.service; static)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: failed (Result: exit-code) since Tue 2023-08-08 17:37:46 UTC; 259ms ago
       Docs: man:rpm-ostree(1)
    Process: 1720 ExecStart=rpm-ostree start-daemon (code=exited, status=1/FAILURE)
   Main PID: 1720 (code=exited, status=1/FAILURE)
     Status: "error: Couldn't start daemon: Error setting up sysroot: loading sysroot: Unexpected state: /run/ostree-booted found and in / sysroot, but bootloader entry not found"
        CPU: 25ms

Aug 08 17:37:46 cosa-devsh systemd[1]: Starting rpm-ostreed.service - rpm-ostree System Management Daemon...
Aug 08 17:37:46 cosa-devsh rpm-ostree[1720]: Reading config file '/etc/rpm-ostreed.conf'
Aug 08 17:37:46 cosa-devsh rpm-ostree[1720]: error: Couldn't start daemon: Error setting up sysroot: loading sysroot: Unexpected state: /run/ostree-booted found and in / sysroot, but bootloader entry not found
Aug 08 17:37:46 cosa-devsh systemd[1]: rpm-ostreed.service: Main process exited, code=exited, status=1/FAILURE
Aug 08 17:37:46 cosa-devsh systemd[1]: rpm-ostreed.service: Failed with result 'exit-code'.
Aug 08 17:37:46 cosa-devsh systemd[1]: Failed to start rpm-ostreed.service - rpm-ostree System Management Daemon.
```

Not sure what `Unexpected state: /run/ostree-booted found and in / sysroot, but bootloader entry not found` mean. @cgwalters ?

raballew:
@alexlarsson I am trying to reproduce your results and I am only able to do so with additional kargs set to enforce=0 otherwise the following units fail:

dbus-broker.service
NetworkManager.service
rpm-ostreed.service
systemd-homed.service
dbus.socket

alexlarsson:
@raballew Sorry, I can't get it to work, must have forgotten i was in permissive mode.

I've done some research here, and here is what I think happens:

* The selinux label of an overlayfs mount is stored as the `security.selinux` xattr on the upper directory of the overlayfs mount. In our case this is  `/run/ostree/.private/etc-transient/upper`.

* In the  ostree-prepare-root service, we copy the real selinux label (xattr) from the sysroot usr/etc onto this upper dir. 

* At a later time systemd loads the selinux policy. At this point, all tmpfs inodes in memory  gets labeled with the associated label `tmpfs_t`. This overwrites the copy we did before.

* During mount setup, systemd manually calls for relabeling of /run, which will recursively traverse /run (not crossing mounts) and relabel files according to the policy that was loaded from the sysroot. This will replace the  `tmpfs_t` with `var_run_t` for the upper dir.

* Now the  `/etc` overlayfs mount root has the overwritten label from the upper dir. This causes various services to fail because of selinux policy.

A manual run of `restorecon /etc` will fix this. But `/etc` is not typically relabeled at boot, because its generally part of a filesystem with persistent selinux labels.

I've tried to make the upper dir a separate mount to stop the var_run_t relabeling across devices, but that just results in getting tmpfs_t instead.

I also tried to mount the overlayfs with `rootcontext=...`, but that fails to mount at this point because selinux is not enabled.

I'm not sure there is anything else we can do other than expressing the expected labels for/run/ostree/.private/etc-transient/upper in the selinux policy, which would make systemd relabel get the right `etc_t label.

alexlarsson:
I see we have a service (ostree-remount) that runs early after switching to the sysroot before /etc is used. I think we can probably fix up /etc in that.

alexlarsson:
> I see we have a service (ostree-remount) that runs early after switching to the sysroot before /etc is used. I think we can probably fix up /etc in that.

Unfortunately several AVCs we hit in /etc happen before this, so it won't work. I guess we really need an ostree selinux module.

raballew:
@alexlarsson So, dropping dcache with `echo 1 > /proc/sys/vm/drop_caches` should do the trick then but it comes at a cost of significant amount of I/O and CPU to recreate the dropped objects.

alexlarsson:
@raballew I dunno if that is even enough. Some inodes are kept alive even over that if they are references elsewhere.

In fact, looking at this in more detail, it is even more complex.
After ostree-prepare-root runs, /sysroot/etc is now a tmpfs sourced from /run, and the selinux policy is not yet loaded. 
For whatever reason, some files are changed in this directory in the initramfs before we switchroot to /sysroot and reexec systemd which triggetrs the loading of the policy and the relabeling of /run.

Directly after a permissive boot we can see that even if we relabeled the upper dir after policy load, some files in the upper are still owned by var_run_t: 
```
# ls -ldZ /run/ostree/.private/etc-transient/upper/
drwxr-xr-x. 11 root root system_u:object_r:etc_t:s0 640 Aug 10 10:04 /run/ostree/.private/etc-transient/upper/
# ls -lZ /run/ostree/.private/etc-transient/upper/
total 64
d---------. 2 root root system_u:object_r:etc_t:s0         40 Jan  1  1970 credstore
d---------. 2 root root system_u:object_r:etc_t:s0         40 Jan  1  1970 credstore.encrypted
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0    108 Aug 10 10:04 group
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0     79 Jan  1  1970 group-
-r--------. 1 root root system_u:object_r:var_run_t:s0    408 Aug 10 10:04 gshadow
-r--------. 1 root root system_u:object_r:var_run_t:s0    391 Jan  1  1970 gshadow-
drwxr-xr-x. 2 root root system_u:object_r:etc_t:s0        120 Aug 10 10:04 issue.d
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0    16243 Aug 10 10:04 ld.so.cache
-rw-r--r--. 1 root root system_u:object_r:locale_t:s0      81 Jan  1  1970 locale.conf
drwxr-xr-x. 3 root root system_u:object_r:etc_t:s0         60 Aug 10 10:04 lvm
-r--r--r--. 1 root root system_u:object_r:tmpfs_t:s0       33 Aug 10 10:04 machine-id
lrwxrwxrwx. 1 root root system_u:object_r:etc_t:s0         19 Jan  1  1970 mtab -> ../proc/self/mounts
lrwxrwxrwx. 1 root root system_u:object_r:etc_t:s0         21 Jan  1  1970 os-release -> ../usr/lib/os-release
drwxr-xr-x. 2 root root system_u:object_r:etc_t:s0         40 Jan  1  1970 pam.d
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0     93 Aug 10 10:04 passwd
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0     38 Jan  1  1970 passwd-
drwxr-xr-x. 2 root root system_u:object_r:var_run_t:s0     80 Aug 10 10:04 profile.d
lrwxrwxrwx. 1 root root system_u:object_r:net_conf_t:s0    39 Jan  1  1970 resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
-r--------. 1 root root system_u:object_r:var_run_t:s0    423 Aug 10 10:04 shadow
-r--------. 1 root root system_u:object_r:var_run_t:s0    397 Jan  1  1970 shadow-
drwxr-xr-x. 2 root root system_u:object_r:etc_t:s0        160 Aug 10 10:04 ssh
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0     18 Aug 10 10:04 subgid
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0      0 Jan  1  1970 subgid-
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0     18 Aug 10 10:04 subuid
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0      0 Jan  1  1970 subuid-
drwxr-xr-x. 4 root root system_u:object_r:var_run_t:s0     80 Aug 10 10:04 systemd
drwxr-xr-x. 2 root root system_u:object_r:etc_t:s0         60 Aug 10 10:04 udev
```
I think this means they were created by overlayfs due to writes to /sysroot/etc before the policy was loaded, and thus the label was applied when labeling /run. Thats not good, for example shadow should be shadow_t.

This is probably some ignition thing fixing up the /etc labels, because the /usr/etc/shadow file is not shadow_t.

Additionally I see ld.so.cache and machine-id are even tmpfs_t instead of var_run_t. Where these not relabeled as var_run_t?

To make things even worse, various files are cached in the overlayfs mount before the relabel happen, which mean that emkdir /ven if the upper dir files get relabled, they are still tmpfs_t when viewed via the overlayfs /etc:

```
# ls -lZ /etc/ | grep tmpfs_t
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0       108 Aug 10 10:04 group
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0        79 Jan  1  1970 group-
-r--------. 1 root root system_u:object_r:tmpfs_t:s0       408 Aug 10 10:04 gshadow
-r--------. 1 root root system_u:object_r:tmpfs_t:s0       391 Jan  1  1970 gshadow-
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0     16243 Aug 10 10:04 ld.so.cache
-r--r--r--. 1 root root system_u:object_r:tmpfs_t:s0        33 Aug 10 10:04 machine-id
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0        93 Aug 10 10:04 passwd
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0        38 Jan  1  1970 passwd-
drwxr-xr-x. 1 root root system_u:object_r:tmpfs_t:s0        80 Aug 10 10:04 profile.d
-r--------. 1 root root system_u:object_r:tmpfs_t:s0       423 Aug 10 10:04 shadow
-r--------. 1 root root system_u:object_r:tmpfs_t:s0       397 Jan  1  1970 shadow-
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0        18 Aug 10 10:04 subgid
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0         0 Jan  1  1970 subgid-
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0        18 Aug 10 10:04 subuid
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0         0 Jan  1  1970 subuid-
drwxr-xr-x. 1 root root system_u:object_r:tmpfs_t:s0        80 Aug 10 10:04 systemd
```

I guess what happened is that since these files were created before selinux policy was added they were in dcache, and the overlayfs inodes that correspond to the upper files were created before the /run relabel, so they aren't aware of the label change. (Really, we shouldn't change the files in the upper dir behind the back of overlayfs like this).

And indeed, after `echo 3 > /proc/sys/vm/drop_caches` the new labels are recoverd:
```
# ls -lZ /etc/ | grep -v etc_t
total 1616
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0     108 Aug 10 10:04 group
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0      79 Jan  1  1970 group-
-r--------. 1 root root system_u:object_r:var_run_t:s0     408 Aug 10 10:04 gshadow
-r--------. 1 root root system_u:object_r:var_run_t:s0     391 Jan  1  1970 gshadow-
-rw-r--r--. 1 root root system_u:object_r:tmpfs_t:s0     16243 Aug 10 10:04 ld.so.cache
-rw-r--r--. 1 root root system_u:object_r:locale_t:s0       81 Jan  1  1970 locale.conf
-r--r--r--. 1 root root system_u:object_r:tmpfs_t:s0        33 Aug 10 10:04 machine-id
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0      93 Aug 10 10:04 passwd
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0      38 Jan  1  1970 passwd-
drwxr-xr-x. 1 root root system_u:object_r:var_run_t:s0      80 Aug 10 10:04 profile.d
lrwxrwxrwx. 1 root root system_u:object_r:net_conf_t:s0     39 Jan  1  1970 resolv.conf -> ../run/systemd/resolve/stub-resolv.conf
-r--------. 1 root root system_u:object_r:var_run_t:s0     423 Aug 10 10:04 shadow
-r--------. 1 root root system_u:object_r:var_run_t:s0     397 Jan  1  1970 shadow-
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0      18 Aug 10 10:04 subgid
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0       0 Jan  1  1970 subgid-
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0      18 Aug 10 10:04 subuid
-rw-r--r--. 1 root root system_u:object_r:var_run_t:s0       0 Jan  1  1970 subuid-
drwxr-xr-x. 1 root root system_u:object_r:var_run_t:s0      80 Aug 10 10:04 systemd
``´
I guess this is what happened to e.g. ld.so.cache. It was created after the relabel, but the root overlayfs had the pre-relabel tmpfs_t label in cache, so the newly created ld.so.cache file inherited that label.

We could avoid having the files created in etc be var_run_t by having the selinux policy specify etc_t also for files inside the upper dir. Then the files created before policy load would get etc_t. But we can't really fix the fact that the policy load "loses" the labels added on tmpfs file pre-policy-load (such as shadow_t), and we can't (easily) fix the fact that the relabel of upper doesn't affect the /etc overlayfs files.

**Comment by alexlarsson** [comment]
Ok, I finally figured out why this is happening:

ext4 (etc) uses xattrs (SECURITY_FS_USE_XATTR) to manage the selinux context. On initialization of an inode it reads the security.selinux xattr, parses it according to the policy, and then stores this (broken down) in the `struct inode_security_struct` for the inode. This struct is used to make security checks. It also hooks into setxattr to update this.

eg:
```
fs_use_xattr ext4 gen_context(system_u:object_r:fs_t,s0);
```

tmpfs uses a different mechanism, `SECURITY_FS_USE_TRANS`, where it never looks at any xattr. It just initializes the `struct inode_security_struct`  with the default value, and then updates this directly when changing the label on a file. TRANS means "transition" and I guess this means it transitions from unableled to this, and then gets set to whatever you update it too when relabeling.

eg:
```
fs_use_trans tmpfs gen_context(system_u:object_r:tmpfs_t,s0);
```

**Comment by alexlarsson** [comment]
Hmm... So, there is another issue too.  When ostree deploy creates the merged etc directory after it copies /usr/etc to /etc it does a relabel of etc, because the selinux labels are different in /usr/etc and /etc. This means that a plain overlayfs of /usr/etc will not work as /etc without doing a full relabel.

**Comment by alexlarsson** [comment]
Ugh, even more issues. When anonymous temporary files are created with overlayfs they get created in the work dir and that has a different selinux context, so i'm getting avcs like:

```
avc:  denied  { create } for  pid=1 comm="systemd" name="#7" scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:unlabeled_t:s0 tclass=chr_file permissive=1
avc:  denied  { link } for  pid=1 comm="systemd" name="#7" dev="vda4" ino=125239 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:unlabeled_t:s0 tclass=chr_file permissive=1
avc:  denied  { rename } for  pid=1 comm="systemd" name="#8" dev="vda4" ino=125239 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:unlabeled_t:s0 tclass=chr_file permissive=1
```

Where the unlabeled_t is the work dir (i think).

**Comment by alexlarsson** [comment]
Also, even with this all worked around, I keep getting AVCs like this far after switchroot:
 ```
AVC avc:  denied  { relabelfrom } for  pid=1302 comm="systemd-hwdb" name="hwdb.bin" dev="vda4" ino=125254 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:systemd_hwdb_etc_t:s0 tclass=file permissive=1
AVC avc:  denied  { relabelto } for  pid=1302 comm="systemd-hwdb" name="hwdb.bin" dev="vda4" ino=125254 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:systemd_hwdb_etc_t:s0 tclass=file permissive=1
```

At this point things should not still be running as `kernel_t`, as that is the context in the initrd before the policy is loaded. systemd should be running as init_t.

I think the problem here is that the overlayfs mount happened before policy was loaded, and it saves the context from that point and uses it when writing things to the upper dir.

alexlarsson:
Ok, so to summarize the problems I've found with using a transient etc
where the basic idea is to have ostree, in the initrd, setup /etc be
an overlayfs mount of /usr/etc in the image backed on a transient
directory (for example in /run):

* The selinux contexts of files in /usr/etc in the image are all etc_t,
  which is not right for when it is moved to /etc. Ostree deploy will
  relabel the files to the right thing when usr/etc is copied to etc.

  We could use the merged relabeled "etc" directory instead as the
  overlayfs lower. To do this with composefs we would have to make the
  compoosefs image contain a relabeled etc in it somewhere.

* Overlayfs stores the calling context at mount time, including things
  like uid and selinux context. If a user is allowed (based on its
  uid/context) to do a file operation on the overlayfs file, the
  overlayfs filesystem applies the equivalent operation on the "upper"
  directory. At this point the overlayfs context is applied against
  the selinux context in the upper dir. This will always have the
  internal selinux context kernel_t, because the overlayfs was mounted
  in the initramfs before the policy was loaded.

  In practice, this means many operations fail against the /etc mount,
  because kernel_t is not very powerful.

* On overlayfs, anonymous files (like tmpfiles) are created first in
  the work dir before being linked into the upper dir. The selinux
  context of this dir is unlabeled due to the early mount, which
  causes AVCs.  And, even if it was labeled there may be problems, as
  whatever label was used may not have the permissions to do some
  things we need.

* systemd relabels /run very early after re-execing into the sysroot
  and loading the selinux policy. If the overlayfs upper dir is in
  /run, then this will be relabeled as var_run_t. This can be worked
  around by having special selinux policy rules for the upper dir.

* /run is on a tmpfs, and the selinux policy for tmpfs filesystems is
  fs_use_trans instead of the more typical fs_use_xattr. This means
  that the initial selinux context for files (i.e. at selinux policy
  load time, or later at file creation time) are set to the fs-wide
  default (tmpfs_t here). This means we can't affect the selinux
  context for any files in the upper dir that are created in the
  initrd. This includes the upper dir itself, the files created
  directly by ignition, and the directories indirectly created by
  ignition when overlayfs creates parent directories in the upper as
  needed for the directly created files.

  Ignition has code to "fix up" these direct labels, but the fixups
  are set on xattrs before policy load, so the fixups will be
  discarded.

Almost all these problems are related to things happening before
loading the selinux policy (i.e. in the initrd). I don't think it is
possible to work around all of these issues, and thus the only
workable approach is to mount the overlayfs after policy
load. However, systemd accessess (and writes to) /etc very early after
loading the policy, and there is no way to intercept before that to
mount the overlay.

So, the summary is that I don't think ostree can implement transient
/etc. But systemd can. and in fact there are transient rootfs options
in systemd which we need to research. The `systemd.volatile=overlay`
option seems to make everything a tmpfs+overlay, and we can maybe use
that.

Issues I forsee with this approach:

 * For the composefs case we would need to be able to expose a
   relabeled /etc, as we don't want to use the relabeled merged
   deploy/etc (since that is persistent and unverified).

 * It is impossible for ignition to modify this etc dir in the initrd.
   Any such changes would have to be made in the image itself, or
   possibly using bind mount tricks.

alexlarsson:
Ok, I tried `systemd.volatile=overlay`. I had to add `Before=systemd-volatile-root.service` to `ostree-prepare-root.service`. But there is an issue with the /sysroot overlayfs mount covering the other submounts like /sysroot/sysroot and the var/etc mounts.

cgwalters:
> The selinux contexts of files in /usr/etc in the image are all etc_t,

Right, but they don't have to be.  At build time (e.g. rpm-ostree) we could force /usr/etc to have labels as if it's `etc`.

I am honestly not sure I can think of real downsides to that...suddenly we don't need to relabel `etc` client side by default.

I think doing this is going to be way better than trying to do relabeling at runtime.

cgwalters:
@raballew why did you close this?

raballew:
@cgwalters I am not sure. I did a force push to the fork and that seems to have auto-closed the PR. It was not my intention but I can not reopen the PR either. This is weird.

raballew:
Reopen due to mistake.

cgwalters:
That's very strange...I don't have permission to reopen it either, and I'm a repository administrator.  I didn't think that was possible.

cgwalters:
Reopening

cgwalters:
Wait, the "reopen and comment" button *did* turn green for that last comment, but did nothing?  I'm pretty sure this sort of Github glitch.

Anyways, not a fatal problem - can you just open a new one?

cgwalters:
Ahh I see, if you hover over it, the Github UI is showing "There are no new commits on the raballew:prepare-root-transient-etc branch" - but I didn't think that was required to reopen a PR?  Wonder if it'd work if you force pushed again though?

raballew:
@cgwalters It seems pushing to the branch resolved the issue. Anyhow, as mentioned in https://github.com/ostreedev/ostree/pull/2972#issuecomment-1674558770 a transient etc is not possible at the moment.

> So, the summary is that I don't think ostree can implement transient
/etc. But systemd can. and in fact there are transient rootfs options
in systemd which we need to research. The systemd.volatile=overlay
option seems to make everything a tmpfs+overlay, and we can maybe use
that.

> Issues I forsee with this approach:

> For the composefs case we would need to be able to expose a
relabeled /etc, as we don't want to use the relabeled merged
deploy/etc (since that is persistent and unverified).

> It is impossible for ignition to modify this etc dir in the initrd.
Any such changes would have to be made in the image itself, or
possibly using bind mount tricks.

For the systemd volatile path I have opened https://github.com/ostreedev/ostree/pull/2986 which has its own challenges.

cgwalters:
Closing in favor of https://github.com/ostreedev/ostree/pull/3062 now...

PR #2974, opened 2023-08-08
Read composefs configuration from initrd instead of commandline

alexlarsson:
This drops the `ot-composefs` kernel commandline in favour of a `[composefs]` section in the `prepare-rootfs.conf` file.

You can set `composefs.enabled` to `yes`, `no` or `maybe`, with `maybe` being the default.

You can also set `composefs.keyfile` to point to an ed25519 key, which the commit must be signed with, or boot fails.

NOTE: This drop the option to define a digest in the commandline. However, that was currently unused (i.e. ComposefsConfig.expected_digest was never read).

Additionally it very hard to actually store the composefs digest in the initrd, as the initrd is typically part of the commit and thus the composefs. It may be possible to handle this, but lets add it back when we know exactly how that will work.

### DIFF
--- a/docs/composefs.md
+++ b/docs/composefs.md
@@ -30,21 +30,13 @@ have a `.ostree.cfs` file in the deployment directory which is a mountable
 composefs metadata file, with a "backing store" directory that is
 shared with the current `/ostree/repo/objects`.
 
-### Kernel argument ot-composefs
+### composefs configuration
 
-The `ostree-prepare-root` binary will look for a kernel argument called `ot-composefs`.
+The `ostree-prepare-root` binary will look for `ostree/prepare-root.conf` in `/etc` and
+`/usr/lib` in the initramfs. Using that configuration file you can enable composefs,
+and specify an Ed25519 public key to validate the booted commit.
 
-The default value is `maybe` (this will likely become a build and initramfs-configurable option)
-in the future too.
-
-The possible values are:
-
-- `off`: Never use composefs
-- `maybe`: Use composefs if supported and there is a composefs image in the deployment directory
-- `on`: Require composefs
-- `digest=<sha256>`: Require the mounted composefs image to have a particular digest
-- `signed=<path>`: Require that the commit is signed as validated by the ed25519 public key specified
-   by `path` (the path is resolved in the initrd).
+See the manpage for `ostree-prepare-root` for details of how to configure it.
 
 ### Injecting composefs digests
 
@@ -56,20 +48,20 @@ covering the composefs fsverity digest with a signature.
 
 ### Signatures
 
-If a commit is signed with a ed25519 private key (see `ostree
---sign`), and `signed=/path/to/public.key` is specified on the
-commandline, then the initrd will find the commit being booted in the
-system repo and validate its signature against the public key. It will
-then ensure that the composefs digest being booted has an fs-verity
-digest matching the one in the commit. This allows a fully trusted
-read-only /usr.
+If a commit is signed with an Ed25519 private key (see `ostree
+--sign`), and `composefs.keyfile` is specified in `prepare-root.conf`,
+then the initrd will find the commit being booted in the system repo
+and validate its signature against the public key. It will then ensure
+that the composefs digest being booted has an fs-verity digest
+matching the one in the commit. This allows a fully trusted read-only
+/usr.
 
 The exact usage of the signature is up to the user, but a common way
-to use it with transien keys. This is done like this:
+to use it with transient keys. This is done like this:
  * Generate a new keypair before each build
  * Embed the public key in the initrd that is part of the commit.
- * Ensure the kernel commandline has `ot-signed=/path/to/key`
- * After commiting, run `ostree --sign` with the private key.
+ * Ensure the initrd has a `prepare-root.conf` with `keyfile=/path/to/key`
+ * After committing, run `ostree --sign` with the private key.
  * Throw away the private key.
 
 When a transient key is used this way, that ties the initrd with the
--- a/man/ostree-prepare-root.xml
+++ b/man/ostree-prepare-root.xml
@@ -32,7 +32,7 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <surname>Walters</surname>
                 <email>walters@verbum.org</email>
             </author>
-        </authorgroup>g
+        </authorgroup>
     </refentryinfo>
 
     <refmeta>
@@ -111,8 +111,25 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
         <variablelist>
             <varlistentry>
                 <term><varname>sysroot.readonly</varname></term>
-                <listitem><para>A boolean value; the default is false.  If this is set to <literal>true</literal>, then the <literal>/sysroot</literal> mount point is mounted read-only.</para></listitem>
-          </varlistentry>
+                <listitem><para>A boolean value; the default is <literal>false</literal>.  If this is set to <literal>true</literal>, then the <literal>/sysroot</literal> mount point is mounted read-only.</para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><varname>composefs.enabled</varname></term>
+                <listitem><para>This can be <literal>yes</literal>, <literal>no</literal>. <literal>maybe</literal> or
+                <literal>signed</literal>. The default is <literal>maybe</literal>.  If set to <literal>yes</literal> or
+                <literal>signed</literal>, then composefs is always used, and the boot fails if it is not
+                available. Additionally if set to <literal>signed</literal>, boot will fail if the image cannot be
+                validated by a public key. If set to <literal>maybe</literal>, then composefs is used if supported.
+                </para></listitem>
+            </varlistentry>
+            <varlistentry>
+                <term><varname>composefs.keypath</varname></term>
+                <listitem><para>Path to a file with Ed25519 public keys in the initramfs, used if
+                <literal>composefs.enabled</literal> is set to <literal>signed</literal>. The default value for this is
+                <literal>/etc/ostree/initramfs-root-binding.key</literal>. For a valid signed boot the target OSTree
+                commit must be signed by at least one public key in this file, and the commitfs digest listed in the
+                commit must match the target composefs image.</para></listitem>
+            </varlistentry>
         </variablelist>
     </refsect1>
 
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -80,11 +80,14 @@
 const char *config_roots[] = { "/usr/lib", "/etc" };
 #define PREPARE_ROOT_CONFIG_PATH "ostree/prepare-root.conf"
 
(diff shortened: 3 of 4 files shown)


docs/composefs.md
```diff
@@ -30,21 +30,13 @@ have a `.ostree.cfs` file in the deployment directory which is a mountable
 composefs metadata file, with a "backing store" directory that is
 shared with the current `/ostree/repo/objects`.
 
-### Kernel argument ot-composefs
+### composefs configuration
 
-The `ostree-prepare-root` binary will look for a kernel argument called `ot-composefs`.
+The `ostree-prepare-root` binary will look for `ostree/prepare-root.conf` in `/etc` and
+`/usr/lib` in the initramfs. Using that configuration file you can enable composefs,
+and specify a ed25519 public key to validate the booted commit.
```
raballew, line 37:
```suggestion
and specify an ed25519 public key to validate the booted commit.
```

docs/composefs.md
```diff
@@ -57,18 +49,18 @@ covering the composefs fsverity digest with a signature.
 ### Signatures
 
 If a commit is signed with a ed25519 private key (see `ostree
```
raballew, line 51:
```suggestion
If a commit is signed with an ed25519 private key (see `ostree
```

docs/composefs.md
```diff
@@ -57,18 +49,18 @@ covering the composefs fsverity digest with a signature.
...
+/usr.
 
 The exact usage of the signature is up to the user, but a common way
-to use it with transien keys. This is done like this:
+to use it with transient keys. This is done like this:
  * Generate a new keypair before each build
  * Embed the public key in the initrd that is part of the commit.
- * Ensure the kernel commandline has `ot-signed=/path/to/key`
+ * Ensure the initrd has a `prepare-root.conf` with `keyfile=/path/to/key`
  * After commiting, run `ostree --sign` with the private key.
```
raballew, line 64:
```suggestion
 * After committing, run `ostree --sign` with the private key.
```

man/ostree-prepare-root.xml
```diff
@@ -113,6 +113,14 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <term><varname>sysroot.readonly</varname></term>
                 <listitem><para>A boolean value; the default is false.  If this is set to <literal>true</literal>, then the <literal>/sysroot</literal> mount point is mounted read-only.</para></listitem>
           </varlistentry>
+            <varlistentry>
+                <term><varname>composefs.enable</varname></term>
+                <listitem><para>This can be yes, no or maybe, the default is false.  If this is set to <literal>yes</literal>, then composefs is always used, and boot fail if it is not available. If set to <literal>maybe</literal>, then composefs is used if supported.</para></listitem>
```
raballew, line 118:
```suggestion
                <listitem><para>This can be yes, no or maybe, the default is no.  If this is set to <literal>yes</literal>, then composefs is always used, and the boot fails if it is not available. If set to <literal>maybe</literal>, then composefs is used if supported.</para></listitem>
```

man/ostree-prepare-root.xml
```diff
@@ -113,6 +113,14 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
...
+          </varlistentry>
+            <varlistentry>
+                <term><varname>composefs.keyfile</varname></term>
+                <listitem><para>Path to a ED25519 public key in the initramfs. If this is specified, the boot fails unless the commit is signed with this key and the composefs image digest matches the one specified in the digest.</para></listitem>
```
raballew, line 122:
```suggestion
                <listitem><para>Path to an ed25519 public key in the initramfs. If this is specified, the boot fails unless the commit is signed with this key and the composefs image digest matches the one specified in the digest.</para></listitem>
```

raballew (changes requested):

docs/composefs.md
alexlarsson: Surely it is "a key", not "an key"?

man/ostree-prepare-root.xml
alexlarsson: Accepted captialization seem to be "Ed25519"

docs/composefs.md
cgwalters: I think rabellew is right here, because ed25519 starts with a vowel.  It's about the immediately following word, not the logical subject.  I think basically because it's about smoothing out pronunciation.

man/ostree-prepare-root.xml
```diff
@@ -113,6 +113,14 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <term><varname>sysroot.readonly</varname></term>
                 <listitem><para>A boolean value; the default is false.  If this is set to <literal>true</literal>, then the <literal>/sysroot</literal> mount point is mounted read-only.</para></listitem>
           </varlistentry>
+            <varlistentry>
+                <term><varname>composefs.enable</varname></term>
```
cgwalters, line 117: `enabled`

man/ostree-prepare-root.xml
```diff
@@ -113,6 +113,14 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
...
+                <listitem><para>This can be yes, no or maybe, the default is no.  If this is set to <literal>yes</literal>, then composefs is always used, and the boot fails if it is not available. If set to <literal>maybe</literal>, then composefs is used if supported.</para></listitem>
```
cgwalters, line 118: For consistency let's wrap as `<literal>yes</literal>` etc here too

man/ostree-prepare-root.xml
```diff
@@ -113,6 +113,14 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
...
+          </varlistentry>
+            <varlistentry>
+                <term><varname>composefs.keyfile</varname></term>
```
cgwalters, line 121: (None of the below is blocking)

I'm fine with this, however...we could also introduce `/etc/ostree/initramfs-root-binding.key` or so and if it exists we automatically honor it  It's a bit easier to use.

If we did this it does probably argue to also support `composefs.enable=signed` again to require signature validation.

But I do also see this "initramfs root binding" as something necessary for composefs, but also logically distinct.  We could also do this even if composefs isn't enabled - basically if the key is present, use it for verification.

A final thing - it looks like the main ostree-sign ed25519 code supports basic "keyrings" that are newline separated files.

It could be beneficial to do so here too and reuse the same code.

man/ostree-prepare-root.xml
```diff
@@ -113,6 +113,14 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
...
+                <listitem><para>Path to a Ed25519 public key in the initramfs. If this is specified, the boot fails unless the commit is signed with this key and the composefs image digest matches the one specified in the digest.</para></listitem>
```
cgwalters, line 122: > composefs image digest matches the one specified in the digest.

composefs image digest matches the one embedded in the target OSTree commit metadata.

?

man/ostree-prepare-root.xml
```diff
@@ -32,7 +32,7 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <surname>Walters</surname>
                 <email>walters@verbum.org</email>
             </author>
-        </authorgroup>g
```
cgwalters, line 35: Oops...I'm sure this one was my fault :smile:

src/switchroot/ostree-prepare-root.c
```diff
@@ -83,8 +83,9 @@ const char *config_roots[] = { "/usr/lib", "/etc" };
 #define SYSROOT_KEY "sysroot"
 #define READONLY_KEY "readonly"
 
-// The kernel argument we support to configure composefs.
-#define OT_COMPOSEFS_KARG "ot-composefs"
+#define COMPOSEFS_KEY "composefs"
+#define ENABLED_KEY "enabled"
+#define KEYFILE_KEY "keyfile"
```
cgwalters, line 88: I originally paused here for a second in confusion because we're parsing a "keyfile" from a (G)"keyfile".  Perhaps `keypath` for the key key?

cgwalters (commented):
Thanks for doing this!  I was actually starting to work on a similar patch - 22b8e4f953ed183a368ec97ce7f74337b1574045 was prep for this.

I think we're agreed that we want to support a config file for this.

However...and I know I'm disagreeing with my past self here...I started to think about the case of changing the configuration locally more.   Kernel arguments are just so nice for this because they don't require mutating the initramfs - another way to look at this is that there's only one copy of kernel arguments (in the bootloader config) that can be read by both the real root and the initramfs directly.

Whereas in the case of an "unlocked system" user (user is root on their local computer) it's less ergonomic to e.g. `vi /etc/ostree/prepare-root.conf` and then also trigger initramfs regeneration (e.g. `rpm-ostree initramfs-etc --track /etc/ostree`).

Anyways, we can figure out whether we also want to support kernel arguments after this lands.

alexlarsson:
> Whereas in the case of an "unlocked system" user (user is root on their local computer) it's less ergonomic to e.g. `vi /etc/ostree/prepare-root.conf` and then also trigger initramfs regeneration (e.g. `rpm-ostree initramfs-etc --track /etc/ostree`).
> 
> Anyways, we can figure out whether we also want to support kernel arguments after this lands.

We can support e.g. `ostree.systroot.readonly=true` syntax to override individual lines in the conf file.

alexlarsson:
New version with default value for the path and a enabled=`signed` option. I also added the new default path to the dracut module for installing.

Still no support for kernel arg overrides though. But, by thinking is that they could be applied generally for the entire prepare-root.conf keyfile.

alexlarsson:
I tested the signatures with this change:
https://gitlab.com/CentOS/automotive/sample-images/-/merge_requests/362
Seems to work fine

src/switchroot/ostree-prepare-root.c
```diff
@@ -274,53 +280,76 @@ validate_signature (GBytes *data, GVariant *signatures, const guchar *pubkey, si
...
+      g_autofree char *pubkeys = NULL;
+      gsize pubkeys_size;
+
+      /* Load keys */
+
+      if (!g_file_get_contents (ret->signature_pubkey, &pubkeys, &pubkeys_size, error))
+        return glnx_prefix_error_null (error, "Reading public key file '%s'",
+                                       ret->signature_pubkey);
+
+      /* Raw binary form if right size */
```
cgwalters, line 329: Having two different file formats seems like it needs some motivation.  If we support this we should document it.  I'd vote we drop it and only support base64+newline separated format, which is the same thing supported by the main ostree+ed25519 format.

(There's a larger picture question here of whether we should actually also leverage openssl here and parse PEM for example too)

alexlarsson: Both file formats are supported by ostree_sign_ed25519_add_pk() though, based on the variant type (bytestring vs string).

alexlarsson: I guess for the ostree sign --keys-file we only ever use the base64 version though.
The current format type in the autosd images is raw, but we can easily change this.

cgwalters (approved):
We can also do the file format tweak afterward.

alexlarsson:
@cgwalters Lemme update the automotive code to use base64 keys first.

PR #2979, opened 2023-08-14
README.md: Drop dead mailing list, link to GH discussions

cgwalters:
While I resisted taking the next step in binding ourselves more to GH with discussions...it's way, way better than answering questions out of band in private (also proprietary) chats.

We haven't been successful in using the GNOME discussion forums.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,9 +158,8 @@ make install DESTDIR=/path/to/dest
 
 ## Contact and discussion forums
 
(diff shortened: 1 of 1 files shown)


jmarrero (approved):
lgtm

cgwalters:
/override ci/prow/fcos-e2e
Doesn't run for docs-only changes

PR #2980, opened 2023-08-14
Prepare root minor

cgwalters:
prepare-root: Use declare-and-initialize

This is our default style.

---

prepare-root: Check for empty string, not strlen > 0

No point in doing a full strlen, we can just check the first byte.
Also, invert the conditional using `continue` to avoid another
level of indentation.

---

prepare-root: Use ptrarray, not linked list

Linked lists are a data structure with only very obscure
use cases, and this is a classic one where since we're appending
it's O(N^2) behavior.

Also we were leaking the memory.

It's more ergonomic, clearer and efficient to use a ptrarray.

---


alexlarsson (approved):
lgtm

PR #2985, opened 2023-08-16
switchroot,generator: Only read /proc/cmdline once

cgwalters:
Change the helper function to parse an existing cmdline instead of potentially reading `/proc/cmdline` multiple times.


ericcurtin (approved):
LGTM, optional change autofree -> g_autofree ... The implementation is basically the same...

PR #2987, opened 2023-08-16
deploy: Add some error prefixing

cgwalters:
We saw a bare
`Txn Rebase on <osname> failed: Failed to find kernel in /usr/lib/modules, /usr/lib/ostree-boot or /boot` which isn't bad, but it'd be better to be a bit more specific.


ericcurtin (approved):
LGTM

PR #2989, opened 2023-08-17
repo: Bump lock timeout to 5 minutes

cgwalters:
And update the doc text to talk about having a timeout at all by default being a mistake.

Timeouts are really best handled at a higher level; if two processes are contending for the ostree lock and one is actually frozen, resolving this is something an admin may want to handle and introspect/debug instead of having the waiter error out.

Most people using ostree are doing it in a way in which they have higher level timeouts (e.g. on a container pod).


ericcurtin (approved):
LGTM

PR #2990, opened 2023-08-17
Two s/osname/stateroot/ patches

cgwalters:
Add `ostree admin stateroot-init` as alias for `os-init`

To further help deprecate the confusing "osname" terminology.

---

admin-deploy: Add `--stateroot` as alias for `--os`

To further help deprecate the confusing "osname" terminology.

---


jmarrero (approved):
lgtm

PR #2991, opened 2023-08-17
Misc c99 style

cgwalters:
admin: Port to c99 style

- Drop `goto out`
- Use declare-and-initialize

---

remote-add: Port to c99 style

- Drop `goto out`
- Use declare-and-initialize

---


ericcurtin (approved):
LGTM, I find the new style better to read and the old code did seem to use goto's just for the sake of it

PR #2993, opened 2023-08-18
Misc c99 style 2

cgwalters:
lzma: Port to C99 style

Just keeping up momentum.

---

checkout: Port to C99 style

Just keeping up momentum.

---

### DIFF
--- a/src/ostree/ot-builtin-checkout.c
+++ b/src/ostree/ot-builtin-checkout.c
@@ -260,15 +260,9 @@ static gboolean
 process_many_checkouts (OstreeRepo *repo, const char *target, GCancellable *cancellable,
                         GError **error)
 {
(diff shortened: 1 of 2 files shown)


src/ostree/ot-builtin-checkout.c
```diff
@@ -293,37 +290,33 @@ process_many_checkouts (OstreeRepo *repo, const char *target, GCancellable *canc
 
       /* Read the null byte */
       (void)g_data_input_stream_read_byte (datastream, cancellable, NULL);
-      g_free (subpath);
```
ericcurtin, line 296: A  good example of why C99 is better, declare when you need helps prevent silliness like this.

ericcurtin (approved):
LGTM

PR #2994, opened 2023-08-21
Refactor composefs warnings

cgwalters:
prepare-root: Drop redundant print about signature/digest

We print if we're doing a signature+digest verification; its absence is
sufficient in the other case.  The goal here is to avoid polluting
the logs when signatures are not enabled.

---

prepare-root: Init composefs options earlier

Prep for a later patch.

---

prepare-root: Fold together composefs signature cases

Now that we don't support digest-but-not-signature verification
for composefs, the logic here was unnecessarily complicated.
With a prior prep patch that moved the composefs option
initialization up, we can just have everything related to signature
verification in a single conditonal.

---

composefs: Hard error except on ENOENT even in "optional" case

Since we enabled composefs at build time, the default (non-composefs)
case now always prints
`composefs: Optional support failed: No such file or directory`
But that's normal and expected.

Rework things here so that in the very special case where
we are in "maybe/optional" mode and we get ENOENT, then we
output a much more normal-looking message that doesn't include
the string "failed".

Now on the flip side - if I have explicitly enabled signature
checking, I think we *do* want to make that fatal even if
composefs is in "maybe" mode.

(This part is more debatable; perhaps we should just disallow
 the case of "maybe" + signatures at all; but I think this is
 an improvement in that direction)

---

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -493,7 +493,6 @@ main (int argc, char *argv[])
           // Or stated in reverse: if signature verification is enabled, then digest verification
           // must also be.
           g_assert (!composefs_config->is_signed);
-          g_print ("composefs: Mounting with no digest or signature check\n");
         }
 
       if (lcfs_mount_image (OSTREE_COMPOSEFS_NAME, TMP_SYSROOT, &cfs_options) == 0)
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -479,20 +479,10 @@ main (int argc, char *argv[])
 
           expected_digest = g_malloc (OSTREE_SHA256_STRING_LEN + 1);
           ot_bin2hex (expected_digest, cfs_digest_buf, g_variant_get_size (cfs_digest_v));
-        }
 
-      if (expected_digest != NULL)
-        {
           cfs_options.flags |= LCFS_MOUNT_FLAGS_REQUIRE_VERITY;
           g_print ("composefs: Verifying digest: %s\n", expected_digest);
           cfs_options.expected_fsverity_digest = expected_digest;
-        }
-      else
-        {
-          // If we're not verifying a digest, then we *must* also have signatures disabled.
-          // Or stated in reverse: if signature verification is enabled, then digest verification
-          // must also be.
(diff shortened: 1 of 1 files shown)


cgwalters:
I guess we can hold this one for the next release, since it could use a bit more testing and in the end it's mostly cosmetic.

cgwalters:
OK release is done, lifting draft here

src/switchroot/ostree-prepare-root.c
```diff
@@ -506,29 +513,14 @@ main (int argc, char *argv[])
...
-            case ENOSIGNATURE:
-              errmsg = "Missing signature for fsverity in composefs image";
-              break;
-            default:
-              errmsg = strerror (errno);
-              break;
-            }
-          if (composefs_config->enabled == OT_TRISTATE_MAYBE)
+          g_assert (composefs_config->enabled != OT_TRISTATE_NO);
+          if (composefs_config->enabled == OT_TRISTATE_MAYBE && errsv == ENOENT)
```
ericcurtin, line 517: Are we sure we want to errx in the ENOVERITY, EWRONGVERITY, ENOSIGNATURE cases when composefs is set to maybe?

ericcurtin: Also, why not have ENOENT in the switch statement with all the other error messages?

```
    case ENOENT:
      return "No image present";
```

cgwalters: > Are we sure we want to errx in the ENOVERITY, EWRONGVERITY, ENOSIGNATURE cases when composefs is set to maybe?

[The commit message discusses this](
https://github.com/ostreedev/ostree/pull/2994/commits/38880bff01a4ed858b8b3d53aa00057b27d5bfc9); I'm open to debate but the rationale is there.

cgwalters: > Also, why not have ENOENT in the switch statement with all the other error messages?

Two reasons:

- `ENOENT` is an "expected error", and we handle it specially anyways
- "No such file or directory" is already an OK explanation of the error

Ah yes and now that I look, the other errors don't actually exist, see:

```
#define ENOVERITY ENOTTY
#define EWRONGVERITY EILSEQ
#define ENOSIGNATURE EBADMSG
```

So the `strerror` for them would just be misleading and confusing.

PR #2996, opened 2023-08-22
cli/set-origin: Port to C99 style

cgwalters:
Just keeping up momentum.


ericcurtin (approved):
LGTM

cgwalters:
Hmm https://status.crates.io/ stays everything is good, but definitely seeing failures to fetch from crates.io in CI

PR #2998, opened 2023-08-22
build-sys: Disable composefs on too-old Linux headers

cgwalters:
This should fix the build with Google OSS-fuzz which currently uses an old Ubuntu.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -274,7 +274,17 @@ AS_IF([test x$have_gpgme = xyes],
 )
 AM_CONDITIONAL(USE_GPGME, test "x$have_gpgme" = xyes)
 
-# These are needed by libcomposefs
+dnl composefs won't work at all without this
+AC_MSG_CHECKING([for MOUNT_ATTR_IDMAP])
+AC_COMPILE_IFELSE(
+	[AC_LANG_SOURCE([[
+			#include <sys/mount.h>
+      int foo = MOUNT_ATTR_IDMAP;
+	]])],
+	[AC_MSG_RESULT(yes)
+	 have_mount_attr_idmap=yes],
+	[AC_MSG_RESULT(no)])
+dnl These are needed by libcomposefs to use the new mount API optionally
 AC_MSG_CHECKING([for new mount API (fsconfig)])
 AC_COMPILE_IFELSE(
 	[AC_LANG_SOURCE([[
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):
LGTM

configure.ac
```diff
@@ -276,6 +276,15 @@ AM_CONDITIONAL(USE_GPGME, test "x$have_gpgme" = xyes)
...
 AC_MSG_CHECKING([for new mount API (fsconfig)])
+AC_COMPILE_IFELSE(
+	[AC_LANG_SOURCE([[
+			#include <sys/mount.h>
+      int foo = MOUNT_ATTR_IDMAP;
+	]])],
+	[AC_MSG_RESULT(yes)
+	 have_mount_attr_idmap=yes],
+	[AC_MSG_RESULT(no)])
+AC_MSG_CHECKING([for new mount API (fsconfig)])
```
dbnicholson, line 287: I think you want to use a different message here. Maybe `for new mount API (idmap)`?

dbnicholson (changes requested):

configure.ac
cgwalters: Thanks, fixed

dbnicholson (approved):
That should be good enough to keep moving.

ericcurtin:
@cgwalters @alexlarsson we built an rpm based on the CentOS Stream 9 one and composefs was off (it was on by default a few days ago). Responding from my phone so this is a hunch but maybe "sys/mount.h" isn't a build dependency in the cs9 .spec file, but just something to watch out for if we do another CentOS Stream 9 release and we want composefs support built-in, but runtime support off for those that don't explicitly enable it.

cgwalters:
Hmm yes...we're going to need a proper c9s CI check here.  It looks like the [COPR build is broken](https://download.copr.fedorainfracloud.org/results/@CoreOS/continuous/centos-stream-9-x86_64/06341555-ostree/builder-live.log.gz) for a different reason, argh.

Hmm.  Yes, going to have to dig through this stuff before we do a release.

cgwalters:
I think I see the problem, `MOUNT_ATTR_IDMAP` is in both glibc and kernel-headers in current Fedora, but in c9s it's only in `linux/mount.h`.

cgwalters:
PR in https://github.com/ostreedev/ostree/pull/3009

PR #3003, opened 2023-08-23
Move prepare-root karg helpers into otcore, add unit tests

cgwalters:
Add long overdue unit testing coverage for this, which at least slightly closes out the android boot CI gap.

Actually, this *copies* the karg parsing code into otcore because it now uses glib, which we're not yet using in the static prepare-root.  It's pretty tempting to drop support for the static prepare root entirely.  But for now we'll live with some code duplication.


cgwalters:
This one can be held until after the next release.

ericcurtin (approved):
LGTM

PR #3006, opened 2023-08-24
Misc c99 style 5

cgwalters:
cmd/export: Port to C99 style

Just keeping up momentum.

---

cmd/export: Fold libarchive error handling

The helper function here doesn't quite carry its weight; this
brings us inline with having single line (no braces) error
return paths.

---

cmd/log: Port to C99 style

Just keeping up momentum.

---

cmd/show-remote-url: Port to C99 style

Just keeping up momentum.

---


cgwalters:
Lifting draft since this is a fine change to merge post-release.

ericcurtin (approved):
LGTM

PR #3008, opened 2023-08-24
prepare-root: If composefs is configured as "maybe" don't fail

ericcurtin:
If composefs is configured as "maybe", we should continue even if composefs support is not built in.

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -424,11 +424,11 @@ main (int argc, char *argv[])
   // Tracks if we did successfully enable it at runtime
   bool using_composefs = false;
 
+#ifdef HAVE_COMPOSEFS
   /* We construct the new sysroot in /sysroot.tmp, which is either the composfs
      mount or a bind mount of the deploy-dir */
   if (composefs_config->enabled != OT_TRISTATE_NO)
     {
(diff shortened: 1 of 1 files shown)


cgwalters (changes requested):
Oh yes, this would have been a big regression, good catch.

This has an incorrect change to the composefs submodule, but otherwise LGTM

src/switchroot/ostree-prepare-root.c
```diff
@@ -533,7 +533,9 @@ main (int argc, char *argv[])
             }
         }
 #else
-      errx (EXIT_FAILURE, "composefs: enabled at runtime, but support is not compiled in");
+      /* if composefs is configured as "maybe", we should continue */
+      if (composefs_config->enabled == OT_TRISTATE_YES)
```
cgwalters, line 537: Also, this is a perfectly fine fix, but...it may actually be more readable to move the `#ifdefs` around such that we do it more like:

```
#ifdef HAVE_COMPOSEFS
  if (composefs_config->enabled != OT_TRISTATE_NO)
    {
    }
#else
  if (composefs_config->enabled == OT_TRISTATE_YES)
    errx (...)
#endif
```
or so

cgwalters (approved):

PR #3009, opened 2023-08-24
C9s CI && composefs build detection fix

cgwalters:
ci: Add c9s build

Since it's an active target of this project.

---

build-sys: Look for both linux/mount.h and sys/mount.h

Since that's what composefs uses.

---


cgwalters:
xref https://github.com/ostreedev/ostree/pull/2998#issuecomment-1692393017

ericcurtin (approved):
LGTM

PR #3010, opened 2023-08-24
build-sys: Really fix composefs check

cgwalters:
The ordering of the includes apparently matters...and I didn't actually check that the previous change enables composefs on c9s. But I did now.  For reals.

While we have the patient open, I switched to `AC_LANG_PROGRAM` because I originally thought the bug had something to do with that. As far as I understand, more cleanly separating the includes from the injected body text is a useful thing in `AC_LANG_PROGRAM`.


jmarrero (approved):
lgtm

PR #3012, opened 2023-08-25
sysroot: Promote the "early prune" behavior to default

cgwalters:
I think we have enough testing for this, let's default it to on, and change the variable to allow opt-out in case it does introduce a regression.


ericcurtin (approved):
LGTM

cgwalters:
OK, found the other place in the test that needed to opt-out to verify the old behavior.

ericcurtin (approved):
LGTM

PR #3013, opened 2023-08-25
2023.6 coverity minor fixes

cgwalters:
switchroot: Lower config parser to otcore, add unit tests

Part of the continuation of unit testing coverage.

---

commit: Drop dead code

Coverity points out that since we're now *always* initializing
metadata, the `create_empty_gvariant_dict()` is unreachable.

---

sign/ed25519: Fix two memory leaks

Spotted by coverity.

---

switchroot: Use g_new/g_free consistently

Coverity complains about this, even though they're the same thing.

---

### DIFF
--- a/src/libostree/ostree-sign-ed25519.c
+++ b/src/libostree/ostree-sign-ed25519.c
@@ -415,17 +415,18 @@ ostree_sign_ed25519_add_pk (OstreeSign *self, GVariant *public_key, GError **err
   if (!_ostree_sign_ed25519_is_initialized (sign, error))
     return FALSE;
 
-  gpointer key = NULL;
+  g_autofree guint8 *key_owned = NULL;
+  const guint8 *key = NULL;
   gsize n_elements = 0;
 
   if (g_variant_is_of_type (public_key, G_VARIANT_TYPE_STRING))
     {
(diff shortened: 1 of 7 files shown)


src/libostree/ostree-sign-ed25519.c
```diff
@@ -415,17 +415,18 @@ ostree_sign_ed25519_add_pk (OstreeSign *self, GVariant *public_key, GError **err
   if (!_ostree_sign_ed25519_is_initialized (sign, error))
     return FALSE;
 
-  gpointer key = NULL;
+  g_autofree guint8 *key_owned = NULL;
+  const guint8 *key = NULL;
```
ericcurtin, line 419: Why not:

`g_autofree const guint8 *key = NULL;`

key_owned is unused, just set.

cgwalters: Because `g_variant_get_fixed_array()` returns a `const` pointer into its data, whereas `g_base64_decode` returns a new value.  Notice that we do set `key_owned` only in the latter case.

ericcurtin: Ah I see, so you want it free'd at the end of scope in one case but not the other... Makes sense

ericcurtin (approved):
LGTM

PR #3019, opened 2023-08-29
mutable-tree: Quiet clang-analyzer warning

cgwalters:
The warning here was about a use-after-free which is serious, but I don't think it was reachable in practice, because we'd only enter the second conditional if `contents_checksum != self->contents_checksum`, yet we made them the same thing here.

Nevertheless, refactor the code to do a direct return in this case because that's basically what happened before, and doing so is a lot clearer.


jmarrero (approved):
lgtm

PR #3020, opened 2023-08-29
mutable-tree: Change some `g_return_if_fail` to `g_assert()`

cgwalters:
We did this in some prior patches because it's better for static analyzers; code is either reachable or not, and we don't want the default desktop-style behavior of "try to stumble on without crashing".  It's just likely to lead to a crash somewhere else.


jmarrero (approved):
lgtm

PR #3021, opened 2023-08-29
tests: Add an integration test for composefs signatures

cgwalters:
Ensure we have some automated test coverage for this.


cgwalters:
OK this doesn't fail but it doesn't quite work either; to do this sanely we want the new rpm-ostree, and I discovered we hadn't built it in Fedora yet.

alexlarsson:
I can't say I know these integration tests, and how this works. But from a glance, this looks ok to me.

jmarrero (approved):
lgtm

PR #3024, opened 2023-08-30
Clang analyzer fixes 5

cgwalters:
keyfile-utils: Quiet a clang-analyzer warning

It complains that we could leak memory if the return value
pointer isn't set.  That's actually a nonsensical case, there's
no reason to call this and ignore the return value.

So change things to require it be set, and also change the
`g_return_val_if_fail` to be hard assertions per our new policy.

---

commit: Quiet clang-analyzer warning

Another conditional ownership.

---


jmarrero (approved):
lgtm

PR #3027, opened 2023-08-31
Drop cap-std from our public APIs

cgwalters:
Since it bumped semver (when I didn't expect it to; xref https://github.com/bytecodealliance/cap-std/commit/963eebf3ab52b04a2e8b9ba88ce6308bbed5cbd0#r121651362

It's not load-bearing enough here to matter versus just passing an untyped file descriptor.

This mainly means that it will be the `glib` ecosystem which forces transitive semver bumps for us, not both.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ on:
 
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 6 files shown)


jmarrero (approved):
lgtm

PR #3029, opened 2023-09-01
rust: Port to glib 0.18

cgwalters:
rust: Port to glib 0.18

- Also rolls up a few other changes in the C API
- Drop `&mut` from `OstreeKernelArgs`; it now confuses type inference.
  More generally while using `&mut` here was well intentioned, it
  goes against the glib standard of mapping everything to `&` and
  accepting interior mutability.

---

rust: Bump semver to 0.19

- New glib 0.18 dependency
- Drop cap-std APIs and features
- Other minor changes

---


cgwalters:
This one should be good to go

ericcurtin (approved):
LGTM, although I'm at a beginner level of Rust knowledge.

PR #3039, opened 2023-09-08
rust: Switch to using `include`

cgwalters:
This way we don't randomly pick up bits from the C library unintentionally as things change on that side.

I think the support for `!` in `include` may be relatively new and that's why the original author here chose to do things via `exclude`.  But using `include` with a few specific exclusions is just way better.


jlebon (approved):
Seems sane.

PR #3048, opened 2023-09-19
Add erofs and overlayfs kernel modules to initramfs

ericcurtin:
These kernel modules are required for composefs usage in the initramfs.

The composefs use-case as of today uses an overlayfs on top of a erofs.

### DIFF
--- a/src/boot/dracut/module-setup.sh
+++ b/src/boot/dracut/module-setup.sh
@@ -19,6 +19,10 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library. If not, see <https://www.gnu.org/licenses/>.
 
+installkernel() {
+    instmods erofs overlay
+}
+
 check() {
     if [[ -x $systemdutildir/systemd ]] && [[ -x /usr/lib/ostree/ostree-prepare-root ]]; then
        return 255


ericcurtin:
We have been adding these kernel modules to out Automotive image via our manifests:

https://gitlab.com/CentOS/automotive/sample-images/-/blob/main/osbuild-manifests/include/computed-vars.ipp.yml#L101

doing it this way alternatively ensures dependant modules of composefs are inserted into the initramfs on all OSes with ostree installed. Generally this is how we install dependant kernel modules anyway for packages, via a module-setup.sh file.

src/boot/dracut/module-setup.sh
```diff
@@ -19,6 +19,10 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with this library. If not, see <https://www.gnu.org/licenses/>.
 
+installkernel() {
+    hostonly='' instmods erofs overlay
```
cgwalters, line 23: This is a complex topic because I am sure there are ostree users who are not using composefs today and this will break their initramfs builds.

We clearly need to do this when we make composefs the default, but it isn't yet.

I think anyone who wants to do this can pretty easily configure dracut externally.

ericcurtin: I know most of these dracut commands you can add a "-o" which means "optional" and "optional" means, include it if it's there otherwise don't worry about it and skip. Would that be appropriate?

ericcurtin: How many platforms is this "module-setup.sh" file used for, all the CentOS Stream variants, all the Fedora variants... Does it cover even more than this?

ericcurtin: RHEL 9 kernel would have this not included but Automotive kernel would, so we would need this to be optional at least for that https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/merge_requests/2620/diffs

cgwalters: There's other Linux systems that use ostree and dracut, I think at least some Debian variants may do so.

> I know most of these dracut commands you can add a "-o" which means "optional" and "optional" means, include it if it's there otherwise don't worry about it and skip. Would that be appropriate?

I'm fine with that, it seems like it'd fix this case.  It is a minor size penalty for those not using composefs now, but we want them to do so in the future for sure.

ericcurtin: I removed `hostonly=''`, I'm not sure we need it. I did various testing of options today to try and create this behaviour, and none of the options suited what I wanted to do.

But, it turns out, it does the behaviour we desire by default. It skips these kernel modules, spits out an error message, includes everything it can and exits 0. I couldn't silence the error message, but maybe it's not a bad thing we are explicitly told these modules can't be found.

```
$ sudo compress=pigz dracut -f; echo $?; sudo lsinitrd | grep -i "overlay.ko\|erofs.ko"
dracut-install: could not get modinfo from 'erofs': No such file or directory
dracut-install: could not get modinfo from 'overlay': No such file or directory
0
```

PR #3051, opened 2023-09-19
rust/tests: Adjust for new ostree

cgwalters:
The rust tests do two things:

- Parse the error output from the binary, which broke when the new ostree hit Fedora stable
- Source the libtest.sh test suite, which broke when we dropped the trivial-httpd binary by default


cgwalters:
Gotta love when CI technical debt just roars out of nowhere and slams into you

cgwalters:
OK, I think we have a winner now...

ericcurtin (approved):
This LGTM

PR #3054, opened 2023-09-20
Demonstrate Rust code from sysroot

cgwalters:
This is a demo/prototype for starting to add new Rust code only for `OstreeSysroot`.

https://github.com/ostreedev/ostree/issues/2427#issuecomment-1728210940

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ name = "integration"
 path = "rust-bindings/tests/tests.rs"
 
 [workspace]
-members = [".", "rust-bindings/sys"]
+members = [".", "src/sysroot-rs", "rust-bindings/sys"]
 
 [dependencies]
 base64 = "0.20.0"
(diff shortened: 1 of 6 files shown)


cgwalters:
I started thinking about this and couldn't help trying it out; it was easier than I thought.

However, a few obvious things:

- We definitely want to at least use https://github.com/mozilla/cbindgen; and even with that, error handling still sucks.  Maybe we can cbindgen about GError...
- The dynamic linking thing will suck for ergonomics

cgwalters:
Other binding tool options:

 - https://rust-diplomat.github.io/book/intro.html (needs investigation, but claims some C support)
 - https://cxx.rs/ - used in rpm-ostree, it's good but the C++ requirement is a hump, but the main annoyance is that it really wants to use exceptions for errors

cgwalters:
> [2023-09-20T19:41:39.924Z] ostree: failed to initialize libostreesysrs.so: /usr/libexec/libostree/libostreesysrs.so: cannot open shared object file: No such file or directory


This is because the supermin tooling only knows how to find things shipped as an RPM.

(Hmmm....maybe we could figure out some way to actually ship this thing as part of libostree-1.so...)

Or I guess, the more I think about it, pull off the band aid and just hard require Rust.

PR #3059, opened 2023-09-27
repo: Default bootloader to zipl on s390x

cgwalters:
This will allow us to drop code like
https://github.com/coreos/coreos-assembler/blob/87fc693c115eae1e7ff0e1621ce26c9167af6e84/src/create_disk.sh#L503 which is really just unnecessary since there aren't any other bootloaders we care about on this architecture.

### DIFF
--- a/src/libostree/ostree-bootloader-zipl.c
+++ b/src/libostree/ostree-bootloader-zipl.c
@@ -63,10 +63,11 @@ static gboolean
 _ostree_bootloader_zipl_query (OstreeBootloader *bootloader, gboolean *out_is_active,
                                GCancellable *cancellable, GError **error)
 {
-  /* We don't auto-detect this one; should be explicitly chosen right now.
-   * see also https://github.com/coreos/coreos-assembler/pull/849
-   */
+#if defined(__s390x__)
+  *out_is_active = TRUE;
+#else
   *out_is_active = FALSE;
+#endif
   return TRUE;
 }
 
--- a/src/libostree/ostree-repo-private.h
+++ b/src/libostree/ostree-repo-private.h
@@ -143,6 +143,13 @@ typedef enum
   /* Non-exhaustive */
 } OstreeCfgSysrootBootloaderOpt;
 
+#if !defined(__s390x__)
+#define CFG_SYSROOT_BOOTLOADER_DEFAULT_STR "auto"
+#else
+// There's nothing else on s390x.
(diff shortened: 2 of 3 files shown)


jlebon (commented):
Though I think we also want `auto` on s390x to just use zipl, no? Hmm, should we do something like this instead:

```diff
diff --git a/src/libostree/ostree-bootloader-zipl.c b/src/libostree/ostree-bootloader-zipl.c
index 4ff093ff..c0a2a145 100644
--- a/src/libostree/ostree-bootloader-zipl.c
+++ b/src/libostree/ostree-bootloader-zipl.c
@@ -63,10 +63,11 @@ static gboolean
 _ostree_bootloader_zipl_query (OstreeBootloader *bootloader, gboolean *out_is_active,
                                GCancellable *cancellable, GError **error)
 {
-  /* We don't auto-detect this one; should be explicitly chosen right now.
-   * see also https://github.com/coreos/coreos-assembler/pull/849
-   */
+#if defined(__s390x__)
+  *out_is_active = TRUE;
+#else
   *out_is_active = FALSE;
+#endif
   return TRUE;
 }

diff --git a/src/libostree/ostree-sysroot.c b/src/libostree/ostree-sysroot.c
index 91b63f94..76912d0e 100644
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -1493,6 +1493,7 @@ _ostree_sysroot_query_bootloader (OstreeSysroot *sysroot, OstreeBootloader **out
   if (bootloader_config == CFG_SYSROOT_BOOTLOADER_OPT_AUTO)
     {
       OstreeCfgSysrootBootloaderOpt probe[] = {
+        CFG_SYSROOT_BOOTLOADER_OPT_ZIPL,
         CFG_SYSROOT_BOOTLOADER_OPT_SYSLINUX,
         CFG_SYSROOT_BOOTLOADER_OPT_GRUB2,
         CFG_SYSROOT_BOOTLOADER_OPT_UBOOT,
```

?

cgwalters:
> Though I think we also want auto on s390x to just use zipl, no?

This seems pretty theoretical given it'd require someone *explicitly* setting `sysroot.bootloader=auto`.  But I think indeed there's no reason not to also fix the other place, so I combined the changes and hoisted the constant to the same place as other bootloader bits.

jlebon (approved):

Issue #2925, opened 2023-07-11
ostree export and hard links


owtaylor:
I tried converting https://hub.docker.com/_/busybox into a OCI-format Flatpak, and got instead of a 2MB result, a 240MB result.
The busy box filesystem has all of the /usr/bin images hardlinked together - they were all separate files in the layer tarfile.

I think the problem is that ostree_repo_export_tree_to_archive(), which Flatpak and 'ostree export' use simply never writes a hard link. It probably should keep track of what checksums have been seen and when the same checksum is seen again, write a hardlink rather than a regular file.

Since ostree doesn't maintain a difference between "duplicated file" and "hardlinked file", this would also reduce archive size when the original filesystem tree had non-hardlinked duplicated files.

If hardlinking duplicated files in the 'ostree export' output is seen as a problem or compatibility break, then it would be useful if OstreeRepoExportArchiveOptions had a new option that Flatpak could use when creating OCI layer archives.

cgwalters:
Note the new ostree-container stuff that lives instead in https://github.com/ostreedev/ostree-rs-ext/ does deduplicate things - though it *also* forces on an ostree-style filesystem layout.

owtaylor:
Having a common format between ostree-rs-ext and Flatpak would be awesome - and maybe that common format could just be the ostree-rs-ext format... I have not analyzed the situation at all.[*] But that's definitely a long term thing, since even if we implemented support for reading ostree-rs-ext style commits today, we wouldn't be able to deploy them until that was available in most Flatpak clients.

So this is definitely worthwhile in the short term. If the busybox example seems contrived - here's the stats for the Fedora Flatpak runtime: 

   compressed layer size as we write it today: 1345M
   After untar, hardlink(1), tar:  712M 

So this is almost doubling the size!

[*]  off the top of my head: the ostree format makes things harder for delta generation, because tardiff uses filenames as hints - but that could be overcome with some level of complexity.

PR #3060, opened 2023-09-29
When exporting, use hardlinks for duplicated files

owtaylor:
For ostree_repo_export_tree_to_archive(), and 'ostree export', when the exported tree contains multiple files with the same checksum, write an archive with hard links.

Without this, importing a tree, then exporting it again breaks hardlinks.

As an example of savings: this reduces the (compressed) size of the Fedora Flatpak Runtime image from 1345MiB to 712MiB.

Resolves: #2925

As noted in #2925, if this is considered insufficiently compatible, it could be put behind an option. If someone is untarring an 'ostree export'  tarball and then using it read-write, then hardlinking coincidentally the same files (like all empty files) could be quite surprising. For typical usage of ostree, however, using hardlinks in the exported tar is less surprising.

There's a fair bit of memory usage during export from keeping all the OstreeRepoFile objects. Making the hash table be 'checksum => path string' might save some of that, though it would depend on the paths. I'm not sure which way is better.

### DIFF
--- a/src/libostree/ostree-repo-libarchive.c
+++ b/src/libostree/ostree-repo-libarchive.c
@@ -943,15 +943,10 @@ ostree_repo_write_archive_to_mtree_from_fd (OstreeRepo *self, int fd, OstreeMuta
 
 #ifdef HAVE_LIBARCHIVE
 
-static gboolean
-file_to_archive_entry_common (GFile *root, OstreeRepoExportArchiveOptions *opts, GFile *path,
-                              GFileInfo *file_info, struct archive_entry *entry, GError **error)
+static char *
+file_to_pathstr (GFile *root, OstreeRepoExportArchiveOptions *opts, GFile *path)
 {
-  gboolean ret = FALSE;
   g_autofree char *pathstr = g_file_get_relative_path (root, path);
-  g_autoptr (GVariant) xattrs = NULL;
-  time_t ts = (time_t)opts->timestamp_secs;
-
   if (opts->path_prefix && opts->path_prefix[0])
     {
       g_autofree char *old_pathstr = pathstr;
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-repo-libarchive.c
```diff
@@ -1086,9 +1094,26 @@ write_directory_to_libarchive_recurse (OstreeRepo *self, OstreeRepoExportArchive
...
+
+                archive_entry_set_hardlink (entry, old_pathstr);
+                if (!write_header_free_entry (a, &entry, error))
+                  goto out;
+
+                break;
+              }
+            else
+              {
+                g_hash_table_insert (seen_checksums, (char *)checksum, g_object_ref (path));
```
cgwalters, line 1114: This may be worth a comment like:

`/* The checksum string is owned by the repo file object */` or so; I paused for a few seconds here thinking about the memory management.

cgwalters (approved):
Thanks for working on this!  This looks sane to me.

tests/libtest.sh
```diff
@@ -249,6 +249,9 @@ setup_test_repository () {
     mkdir baz/another/
     echo x > baz/another/y
 
+    echo "SAME_CONTENT" > baz/duplicate_a
+    echo "SAME_CONTENT" > baz/duplicate_b
```
cgwalters, line 253: Let's also add a test case for hardlinks across directories?

owtaylor:
Pushed a new version with clang-format applied.

cgwalters:
> There's a fair bit of memory usage during export from keeping all the OstreeRepoFile objects. Making the hash table be 'checksum => path string' might save some of that, though it would depend on the paths. I'm not sure which way is better.

The other alternative would be to walk the tree twice; the first time to find hardlinked objects.  That may be a good CPU/memory tradeoff.

owtaylor:
> > There's a fair bit of memory usage during export from keeping all the OstreeRepoFile objects. Making the hash table be 'checksum => path string' might save some of that, though it would depend on the paths. I'm not sure which way is better.
> 
> The other alternative would be to walk the tree twice; the first time to find hardlinked objects. That may be a good CPU/memory tradeoff.

You mean, the first time, keep a checksum => seen count hash table, then use that to save only selected OstreeRepoFile objects on the second pass?

The maximum savings over the "save the paths" approach is <average length of path * number of files in tree> - for a *very* large tree (million files), that might be around 100 MiB? I'm less sure about the savings compared to the "save the OstreRepoFiles" approach.

I think I'd rather avoid the complexity and the chance of getting it wrong, but if you feel strongly, I'm happy to do it that way too.

cgwalters:
Definitely let's keep what you've tested, it's in the merge queue!

PR #3062, opened 2023-10-02
Support transient /etc

alexlarsson:
This is a new version of https://github.com/ostreedev/ostree/pull/2972
There was a lot of complexities, so this is essentially a clean restart. So, rather than reusing that PR I created a new one.

This PR does multiple things:

 *  Adds a `etc.transient=true` option to `prepare-root.conf`, which causes /etc to be and overlayfs mount with a tmpfs-based upper.
 * Relabels /usr/etc as if it was /etc in the generated composefs image, allowing the use of usr/etc as the lower for the above /etc overlayfs mount.
 * Since we're changing the composefs image format anyway (due to the relabeling), we're copying the mkcomposefs change that inlines files at 64 bytes or smaller to avoid overhead.

The actual transient etc overlay mount details are quite complex due to selinux interactions between tmpfs, selinux policy loading, and overlayfs. See https://github.com/ostreedev/ostree/pull/2972#issuecomment-1674558770 for some details. In the end, after a discussion with @rhatdan I ended up with a new ostree selinux module that will be needed for transient mode to work.

### DIFF
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -35,6 +35,9 @@
 #include <sys/stat.h>
 #include <sys/statvfs.h>
 #include <unistd.h>
+#ifdef HAVE_SELINUX
+#include <selinux/restorecon.h>
+#endif
 
 #include "ostree-mount-util.h"
 #include "otcore.h"
@@ -74,6 +77,50 @@ do_remount (const char *target, bool writable)
     }
 
   printf ("Remounted %s: %s\n", writable ? "rw" : "ro", target);
+}
+
+/* Relabel the directory $real_path, which is going to be an overlayfs mount,
+ * based on the content of an overlayfs upperdirectory that is in use by the mount.
+ * The goal is that we relabel in the overlay mount all the files that have been
+ * modified (directly or via parent copyup operations) since the overlayfs was
+ * mounted. This will be used for the /etc overlayfs mount where no selinux labels
+ * are set before selinux policy is loaded.
+ */
+static void
+relabel_dir_for_upper (const char *upper_path, const char *real_path, gboolean is_dir)
+{
+#ifdef HAVE_SELINUX
+  if (selinux_restorecon (real_path, 0))
+    err (EXIT_FAILURE, "Failed to relabel %s", real_path);
+
+  if (!is_dir)
+    return;
+
+  g_auto (GLnxDirFdIterator) dfd_iter = {
+    0,
+  };
+
+  if (!glnx_dirfd_iterator_init_at (AT_FDCWD, upper_path, FALSE, &dfd_iter, NULL))
+    err (EXIT_FAILURE, "Failed to open upper directory %s for relabeling", upper_path);
+
+  while (TRUE)
+    {
+      struct dirent *dent;
+
+      if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent, NULL, NULL))
+        {
+          err (EXIT_FAILURE, "Failed to read upper directory %s for relabelin", upper_path);
+          break;
+        }
+
+      if (dent == NULL)
+        break;
+
+      g_autofree char *upper_child = g_build_filename (upper_path, dent->d_name, NULL);
+      g_autofree char *real_child = g_build_filename (real_path, dent->d_name, NULL);
+      relabel_dir_for_upper (upper_child, real_child, dent->d_type == DT_DIR);
+    }
+#endif
 }
 
 int
@@ -119,6 +166,52 @@ main (int argc, char *argv[])
   if (mount ("none", "/sysroot", NULL, MS_REC | MS_PRIVATE, NULL) < 0)
     perror ("warning: While remounting /sysroot MS_PRIVATE");
 
+  const char *transient_etc = NULL;
+  g_variant_dict_lookup (ostree_run_metadata, OTCORE_RUN_BOOTED_KEY_TRANSIENT_ETC, "&s",
+                         &transient_etc);
+
+  if (transient_etc)
+    {
+      /* If the initramfs created any files in /etc (directly or via overlay copy-up) they
+       * will be unlabeled, because the selinux policy is not loaded until after the
+       * pivot-root. So, for all files in the upper dir, relabel the corresponding overlay
+       * file.
+       *
+       * Also, note that during boot systemd will create a /run/machine-id ->
+       * /etc/machine-id bind mount (as /etc is read-only early on). It will then later
+       * replace this mount with a real one (in systemd-machine-id-commit.service).
+       *
+       * We need to label the actual overlayfs file, not the temporary bind-mount. To do
+       * this we unmount the covering mount before relabeling, but we do so in a temporary
+       * private namespace to avoid affecting other parts of the system.
+       */
+
+      glnx_autofd int initial_ns_fd = -1;
+      if (g_file_test ("/run/machine-id", G_FILE_TEST_EXISTS)
+          && g_file_test ("/etc/machine-id", G_FILE_TEST_EXISTS))
+        {
+          initial_ns_fd = open ("/proc/self/ns/mnt", O_RDONLY | O_NOCTTY | O_CLOEXEC);
+          if (initial_ns_fd < 0)
+            err (EXIT_FAILURE, "Failed to open initial namespace");
+
+          if (unshare (CLONE_NEWNS) < 0)
+            err (EXIT_FAILURE, "Failed to unshare initial namespace");
+
+          /* Ensure unmount is not propagated */
+          if (mount ("none", "/etc", NULL, MS_REC | MS_PRIVATE, NULL) < 0)
+            err (EXIT_FAILURE, "warning: While remounting /etc MS_PRIVATE");
+
+          if (umount2 ("/etc/machine-id", MNT_DETACH) < 0)
+            err (EXIT_FAILURE, "Failed to unmount machine-id");
+        }
+
+      g_autofree char *upper = g_build_filename (transient_etc, "upper", NULL);
+      relabel_dir_for_upper (upper, "/etc", TRUE);
+
+      if (initial_ns_fd != -1 && setns (initial_ns_fd, CLONE_NEWNS) < 0)
+        err (EXIT_FAILURE, "Failed to join initial namespace");
+    }
+
   gboolean root_is_composefs = FALSE;
   g_variant_dict_lookup (ostree_run_metadata, OTCORE_RUN_BOOTED_KEY_COMPOSEFS, "b",
                          &root_is_composefs);
@@ -140,8 +233,9 @@ main (int argc, char *argv[])
   do_remount ("/sysroot", !sysroot_configured_readonly);
 
   /* And also make sure to make /etc rw again. We make this conditional on
-   * sysroot_configured_readonly because only in that case is it a bind-mount. */
-  if (sysroot_configured_readonly)
+   * sysroot_configured_readonly && !transient_etc because only in that case is it a
+   * bind-mount. */
+  if (sysroot_configured_readonly && !transient_etc)
     do_remount ("/etc", true);
 
   /* If /var was created as as an OSTree default bind mount (instead of being a separate
--- a/Makefile-switchroot.am
+++ b/Makefile-switchroot.am
@@ -63,6 +63,11 @@ ostree_remount_SOURCES = \
 ostree_remount_CPPFLAGS = $(AM_CPPFLAGS) $(OT_INTERNAL_GIO_UNIX_CFLAGS) -Isrc/switchroot -I$(srcdir)/src/libotcore -I$(srcdir)/src/libotutil -I$(srcdir)/libglnx
 ostree_remount_LDADD = $(AM_LDFLAGS) $(OT_INTERNAL_GIO_UNIX_LIBS) libotcore.la libotutil.la libglnx.la
 
+if USE_SELINUX
+ostree_remount_CPPFLAGS += $(OT_DEP_SELINUX_CFLAGS)
+ostree_remount_LDADD += $(OT_DEP_SELINUX_LIBS)
+endif
+
 if USE_COMPOSEFS
 ostree_prepare_root_LDADD += libcomposefs.la
 endif
--- a/man/ostree-prepare-root.xml
+++ b/man/ostree-prepare-root.xml
@@ -114,6 +114,10 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <listitem><para>A boolean value; the default is <literal>false</literal>.  If this is set to <literal>true</literal>, then the <literal>/sysroot</literal> mount point is mounted read-only.</para></listitem>
             </varlistentry>
             <varlistentry>
+                <term><varname>etc.transient</varname></term>
+                <listitem><para>A boolean value; the default is <literal>false</literal>.  If this is set to <literal>true</literal>, then the <literal>/etc</literal> mount point is mounted transiently i.e. a non-persistent location.</para></listitem>
+            </varlistentry>
+            <varlistentry>
                 <term><varname>composefs.enabled</varname></term>
                 <listitem><para>This can be <literal>yes</literal>, <literal>no</literal>. <literal>maybe</literal> or
                 <literal>signed</literal>. The default is <literal>maybe</literal>.  If set to <literal>yes</literal> or
--- a/src/libotcore/otcore.h
+++ b/src/libotcore/otcore.h
@@ -72,3 +72,5 @@ GKeyFile *otcore_load_config (int rootfs, const char *filename, GError **error);
 #define OTCORE_RUN_BOOTED_KEY_COMPOSEFS_SIGNATURE "composefs.signed"
 // This key will be present if the sysroot-ro flag was found
 #define OTCORE_RUN_BOOTED_KEY_SYSROOT_RO "sysroot-ro"
+
+#define OTCORE_RUN_BOOTED_KEY_TRANSIENT_ETC "transient-etc"
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -87,6 +87,9 @@
 #define SYSROOT_KEY "sysroot"
 #define READONLY_KEY "readonly"
 
+#define ETC_KEY "etc"
+#define TRANSIENT_KEY "transient"
+
 #define COMPOSEFS_KEY "composefs"
 #define ENABLED_KEY "enabled"
 #define KEYPATH_KEY "keypath"
@@ -547,13 +550,51 @@ main (int argc, char *argv[])
    * the deployment needs to be created and remounted as read/write. */
   if (sysroot_readonly || using_composefs)
     {
-      /* Bind-mount /etc (at deploy path), and remount as writable. */
-      if (mount ("etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_SILENT, NULL) < 0)
-        err (EXIT_FAILURE, "failed to prepare /etc bind-mount at /sysroot.tmp/etc");
-      if (mount (TMP_SYSROOT "/etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_REMOUNT | MS_SILENT,
-                 NULL)
-          < 0)
-        err (EXIT_FAILURE, "failed to make writable /etc bind-mount at /sysroot.tmp/etc");
+      gboolean etc_transient = FALSE;
+      if (!ot_keyfile_get_boolean_with_default (config, ETC_KEY, TRANSIENT_KEY, FALSE,
+                                                &etc_transient, &error))
+        errx (EXIT_FAILURE, "Failed to parse etc.transient value: %s", error->message);
+
+      if (etc_transient)
+        {
+          char *ovldir = "/run/ostree/transient-etc";
+
+          g_variant_builder_add (&metadata_builder, "{sv}", OTCORE_RUN_BOOTED_KEY_TRANSIENT_ETC,
+                                 g_variant_new_string (ovldir));
+
+          char *lowerdir = "usr/etc";
+          if (using_composefs)
+            lowerdir = TMP_SYSROOT "/usr/etc";
+
+          g_autofree char *upperdir = g_build_filename (ovldir, "upper", NULL);
+          g_autofree char *workdir = g_build_filename (ovldir, "work", NULL);
+
+          struct
+          {
+            const char *path;
+            int mode;
+          } subdirs[] = { { ovldir, 0700 }, { upperdir, 0755 }, { workdir, 0755 } };
+          for (int i = 0; i < G_N_ELEMENTS (subdirs); i++)
+            {
+              if (mkdirat (AT_FDCWD, subdirs[i].path, subdirs[i].mode) < 0)
+                err (EXIT_FAILURE, "Failed to create dir %s", subdirs[i].path);
+            }
+
+          g_autofree char *ovl_options
+              = g_strdup_printf ("lowerdir=%s,upperdir=%s,workdir=%s", lowerdir, upperdir, workdir);
+          if (mount ("overlay", TMP_SYSROOT "/etc", "overlay", MS_SILENT, ovl_options) < 0)
+            err (EXIT_FAILURE, "failed to mount transient etc overlayfs");
+        }
+      else
+        {
+          /* Bind-mount /etc (at deploy path), and remount as writable. */
+          if (mount ("etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_SILENT, NULL) < 0)
+            err (EXIT_FAILURE, "failed to prepare /etc bind-mount at /sysroot.tmp/etc");
+          if (mount (TMP_SYSROOT "/etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_REMOUNT | MS_SILENT,
+                     NULL)
+              < 0)
+            err (EXIT_FAILURE, "failed to make writable /etc bind-mount at /sysroot.tmp/etc");
+        }
     }
 
   /* Prepare /usr.


alexlarsson:
@cgwalters Sure. That approach works too, although only with ostree commits built with a new rpm-ostree.

alexlarsson:
@cgwalters And btw, we're not relabeing /etc on boot for this reason. We relabel /usr/etc during deploy, when we generate the composefs image. The on-boot stuff (the part in ostree-remount) is to fix up the labels for things that changed in upperdir during the initramfs.

cgwalters:
Hmm; changing the labels just in the composefs flow is making it asymmetric with the legacy case though.  I'm not strictly opposed to that but...it seems better to support setting up the labels in this way in both cases so we can test it more uniformly.

alexlarsson:
@cgwalters Just so we're on the same page. Is this what you mean:

With this PR, this is how transient-etc works:
* rpm-ostree labels usr/etc "wrong"
* If not composefs:
  * ostree deploy relabels the $deploydir/etc during 
  * This is then 3-way merged with previous /etc during update
  * /etc is an overlayfs with lower=$deploydir/etc
* else:
  * ostree deploy relabels usr/etc in the composefs image
  * /etc is an overlayfs with lower=$cfsmount/usr/etc

And your proposal is:
* rpm-ostree labels usr/etc as /etc
* We don't care about $deploydir/etc
* if not composefs
  * /etc is an overlayfs with lower=$deploydir/usr/etc
 * else:
   * /etc is an overlayfs with lower=$cfsmount/usr/etc

That makes a lot of sense to me.

alexlarsson:
I'll have a look at the rpm-ostree side, then we can just drop the first two commits in this PR. I want to test the current code against the autosd images though, to ensure it works there. So far I've only tried against fcos.

alexlarsson:
New version following this approach. To test it I just commented out the flag check, and it seems to work well. Hoever, the real fix is to make rpm-ostree pass this flag. Do we always pass it?

src/switchroot/ostree-remount.c
```diff
@@ -119,6 +159,49 @@ main (int argc, char *argv[])
   if (mount ("none", "/sysroot", NULL, MS_REC | MS_PRIVATE, NULL) < 0)
     perror ("warning: While remounting /sysroot MS_PRIVATE");
 
+  const char *transient_etc = NULL;
+  g_variant_dict_lookup (ostree_run_metadata, OTCORE_RUN_BOOTED_KEY_TRANSIENT_ETC, "&s",
+                         &transient_etc);
+
+  if (transient_etc)
+    {
+      /* Systemd will create a /run/machine-id -> /etc/machine-id bind mount if /etc is
```
cgwalters, line 168: Historically `ostree-remount.service` has just been a big source of race conditions.

In general the direction has been to ensure that the target `/` is set up fully in the initramfs.


In particular the historical rationale for `ostree-remount.service` was compatibility with the legacy behavior of remounting `/` as writable only in the real root early boot.

If that's done in the initramfs, there is no reason to run `ostree-remount.service`.  And I think we should aim to do that uniformly - and actually disable `ostree-remount.service` by default.

I'd need to dig in here into what point the `/etc/machine-id` is being dealt with in systemd, but it really seems to me like the relevant code here should be run in the initramfs.

selinux/ostree.te
```diff
@@ -0,0 +1,20 @@
...
+# against the overlayfs upper/work dirs.
+#
+# Unfortunately, for historical reasons kernel_t doesn't really have any permissions in
+# the selinux, so the overlayfs will not be able to do most operations against the etc_t
+# (and similar) upper files.
+#
+# To fix this we make the kernel context unconfined. It essentially is anyway, as
+# the kernel is the entity that validates the permissions anyway.
+
+unconfined_domain(kernel_t)
```
cgwalters, line 20: OK I think we need to debate this.  I hope you'd agree with me that logically this change isn't really related to ostree.

I think if we're going to do this it should probably be just changed in the base SELinux policy package.

Isn't this effectively undoing some effort to confine `kernel_t` actually?  IIRC there was some...something for constraining usermode helpers that could be executed, etc.

I haven't dug in deep to this but could we do something like add an overlayfs option that disables LSM checks for it?  And then when we switch to the real root and selinux policy is loaded, we remount it, setting the mounter credentials to `unconfined_t` and re-enabling LSM checks.

cgwalters:
>  Hoever, the real fix is to make rpm-ostree pass this flag. Do we always pass it?

See https://github.com/coreos/rpm-ostree/pull/4640

src/switchroot/ostree-remount.c
cgwalters: > If that's done in the initramfs, there is no reason to run ostree-remount.service. 

Ah wait, there is one load bearing thing left that it's doing, which is to reset the propagation mode of `/sysroot`.

Blah...I wonder if we can work with systemd upstream to add some way to opt out of that for particular mounts.

Alternatively...I could imagine actually hiding the `/sysroot` mount in a separate mount namespace entirely, held open by some process that we keep running from the initramfs.

That would be much stronger isolation, but it'd mean every tool that wants to operate on that state would need to `nsenter -m -t $(pidof ostree-mountns-holder)`.

selinux/ostree.te
rhatdan: This is a lot of work, which I don't find that useful, we can hack around this, but the kernel is enforcing the rules, and confining it just makes for hacks to get around it while adding no real security benefit.
@wrabcak @zpytela WDYT?

cgwalters: Alternatively, I could imagine building something like `initramfs_t` into the kernel itself, and use it for things we run from the initramfs.

jlebon: Also not a fan of this approach, or having it live here.

What kind of SELinux denials are you getting BTW? There was a related BZ to this a while back: https://bugzilla.redhat.com/show_bug.cgi?id=2154428.

At that time, it was the overlay created by `ostree admin hotfix` causing issues (so on `/usr`, not `/etc`), but it lead to https://github.com/fedora-selinux/selinux-policy/pull/1574 which greatly loosened what `kernel_t` can do. I'm assuming you're running your tests with that patch there? If so, it might just need to tweaked a bit further.

> Alternatively, I could imagine building something like `initramfs_t` into the kernel itself, and use it for things we run from the initramfs.

I like it. That's a similar (but better) idea to what I mentioned in https://bugzilla.redhat.com/show_bug.cgi?id=2154428#c19 to be able to differentiate between pre and post-pivot.

Also /cc @WOnder93

rhatdan: The issues are that initramfs launches processes before the policy is loaded.  When systemd loads the policy any process launched prior to systemd is going to be labeled kernel_t, if those processes do anything that is not currently defined, then they blow up. It becomes a big mess.

Since these processes exists before systemd load_policy, they can only be labeled kernel_t not initramfs_t.

src/switchroot/ostree-remount.c
alexlarsson: It is impossible to do this relabeling in the initrd, becase at that point the selinux policy is not loaded.

selinux/ostree.te
alexlarsson: This was the recommended solution from @rhatdan, but yes, i agree that it is something that should be discussed. His opinion is that kernel_t is already unconfined (being as the kernel is the entity that does the validation of the selinux contexts), and that it is a huge mistake to not do this as it keeps causing issues.

alexlarsson: I did at one point have code that re-mounted the /etc overlay mount. It was massively more complex and had some limitations and problems, but it is *doable*. In the end that failed too because it got mounted init_t instead of kernel_t , which isn't unconfined anyway. So it had the same problem. It would be possible to instead do the mounting in a separate process that has is unconfined_t, but then we would have to add a rule allowing init_t to go to unconfined_t. That would be more complex, and probably more risky.

As to adding overlayfs options to disable LSM checks, I doubt that is going to happen. And also, not selinux based mount options work in the initramfs anyway, because the selinux policy has not been loaded.

Anyway, @rhatdan care to comment?

src/switchroot/ostree-remount.c
alexlarsson: And, for reasons described in https://github.com/ostreedev/ostree/pull/2972#issuecomment-1674558770 it isn't even possible to manually fake the relabeling in the initrd. It has to be done after the policy is loaded.

selinux/ostree.te
rhatdan: I think it is fairly insane to work around issues of SELinux errors by kernel processes that happen early in boot, and make the system more fragile.  We could add a bunch of allow rules to kernel_t to make it closer to unconfined_domain, but in my opinion is kind of a waste of time.  I want @wrabcak @zpytela to comment.

cgwalters: > but then we would have to add a rule allowing init_t to go to unconfined_t.

That has to already be supported though at least *indirectly* because clearly init_t ends up launching user login shells as unconfined_t...

wrabcak: Allow me to discuss this with whole SELinux team. I understand comments from @rhatdan  that it would result in many SELinux policy work, however similar discussion were here for many years also about making init_t as unconfined etc.  Maybe I missed the log but do we have any list of SELinux denials related to the kernel_t issue? It would give us some idea what needs to be modified if we won't make kernel_t as unconfined_domain.

cgwalters: >  Maybe I missed the log but do we have any list of SELinux denials related to the kernel_t issue? It would give us some idea what needs to be modified if we won't make kernel_t as unconfined_domain.

As far as I understand it, because overlayfs captures the mounting process credentials (at the time of the initramfs, `kernel_t`) it would mean that in the general case, `kernel_t` would need access to perform *all* file operations on all file types in `/etc` at least.

I still think though while the short term making `kernel_t` unconfined is probably simpler, medium term what seems cleanest to me is simply turning off the mounting process LSM checks for this overlayfs mount.  Something like `mount -t overlayfs -o mount-credential-check=disabled` or so (which would be gated obviously by `CAP_SYS_ADMIN` the root userns, etc).

rhatdan:
This is not really a persistent process issue, it is more of what Colin said.  If I am running in initramfs and I mount an overlay, from the kernel's point of view kernel_t mounted the file system. When a user tries to read/or write content in the overlay, the way the kernel works is it checks whether the "mounter" (kernel_t) is able to read/execute the file that the process in userspace is attempting to use.  The AVC's are things like kernel_t trying to read /etc/foobar, and getting an AVC.  We could write a policy that says the kernel_t can read/execute all file_types to fix this issue, but I think this ends up being a slippery path. Eventually we end up with an AVC saying the kernel is not allowed to do something, and user say WTH?

rhatdan:
I don't thing the solution to this is to change the kernel. to tell overlay to work differently from a security point of view. But to simply fix the SELinux policy to give broader permissions to kernel_t.  We can work more closely on fixing the issues with init_t, if necessary.

selinux/ostree.te
alexlarsson: @cgwalters That seems a lot more risky to me. I mean, it will get equivalent results in this particular usecase, but having the ability to later create these kinds of mounts that essentially neuter selinux rules sort of goes against the robustness of the global selinux policy, does it not? 

kernel_t behaviour seems harder to misuse once you're in a full selinux policy.

Imagine trying to argue this upstream agains the potential security risks. We already have CAP_DAC_OVERRIDE, so overriding user permissions is not a new thing, but I don't think we ever have had any options to override MAC checks, so this would be a first.

alexlarsson: Since there was interest i booted without the module, in permissive mode, to show the boot problems. Here are [the results](https://github.com/ostreedev/ostree/files/12801082/avc-boot.txt)

However, that is just the most basic of things. On top of that, almost any kind of use of /etc in the booted system is denied.  For example, if I run `mkdir /etc/foo` as unconfined_t, then I get:
```
avc:  denied  { create } for  pid=2101 comm="mkdir" name="foo" scontext=system_u:system_r:kernel_t:s0 tcontext=unconfined_u:object_r:etc_t:s0 tclass=dir permissive=1
```
So, basically, /etc is useless.

alexlarsson:
Yeah, I think my example from a comment above is a simple way to understand it.

I log in as root, so I am `unconfined_t`, then i run `mkdir /etc/foo`, and since the label on /etc is `etc_t` the mkdir syscall entry is allowed and the makedir code proceeds.

However, then somewhere deep inside overlayfs it creates the backing directory in the upperdir (which is also `etc_t`). This is done in the context of the mounter (`kernel_t`), and `kernel_t` is not allowed to create a directory in `etc_t`, so it fails:

`
avc:  denied  { create } for  pid=2101 comm="mkdir" name="foo" scontext=system_u:system_r:kernel_t:s0 tcontext=unconfined_u:object_r:etc_t:s0 tclass=dir permissive=1
`

Now, I'm supposedly unconfined, so I should be able to do *anything*. However, for that to work, the overlayfs mount must also be unconfined. Anything more limited means that there are things unconfined_t cannot do in /etc.

selinux/ostree.te
WOnder93: > I still think though while the short term making `kernel_t` unconfined is probably simpler,

It's okay to make it unconfined with the exception of the `execute_no_trans` permission (to restrict which userspace binaries the kernel can execute, protecting against vulnerabilities that allow an attacker to overwrite usermode helper paths - see https://issues.redhat.com/browse/RHEL-1466). The problem is that the simple `unconfined_domain(kernel_t)` would grant it this permission universally, so that's a no-go. The alternative is to either add rules broad enough to cover the early boot overlayfs needs (and that don't include `execute_no_trans`), or to make a version of `unconfined_domain()` that grants everything but `execute_no_trans` (also possible, but would require some careful refactoring of multiple interfaces, I think).

> medium term what seems cleanest to me is simply turning off the mounting process LSM checks for this overlayfs mount. Something like `mount -t overlayfs -o mount-credential-check=disabled` or so (which would be gated obviously by `CAP_SYS_ADMIN` the root userns, etc).

No, I don't this would fly. It shouldn't be possible to circumvent MAC through a DAC privilege (be it `CAP_SYS_ADMIN` or something else). If you wanted to go this way, the mount credential check exemption would need to be also gate-able via the MAC policy (i.e. so that you would also need something like `allow kernel_t self:overlayfs override_mounter_check`), but that would be ugly as well and I don't know how the LSM/SELinux community would feel about it...

> Alternatively, I could imagine building something like initramfs_t into the kernel itself, and use it for things we run from the initramfs.

This is actually something I've been trying to implement recently and already had a patch accepted upstream for a while, but it got reverted shortly after because of an unforeseen regression. But I have a goal to rework it and get it upstream by the end of the year, so long/medium term this should also help with the issue (but it probably won't make much difference, since there would still be reasons to keep kernel_t close to or equal to unconfined-except-execute_no_trans).

cgwalters: "kernel_t = unconfined_t - execute_no_trans" sounds perfectly fine to me.

cgwalters: > No, I don't this would fly. It shouldn't be possible to circumvent MAC through a DAC privilege (be it CAP_SYS_ADMIN or something else).

Yes, though we could support this *only* when SELinux policy isn't loaded actually...

but anyways this point seems moot if we're in agreement about the kernel_t change.

rhatdan: unconfined_domain(kernel_t) == ("kernel_t = unconfined_t - execute_no_trans")

rhatdan: Well not quite, but I don't think it is an issue.
allow unconfined_domain_type container_domain:process { dyntransition transition };
allow unconfined_domain_type container_runtime_t:process transition;
allow unconfined_domain_type qm_container_domain:process { dyntransition transition };

rhatdan: If you want to do something like
setexeccon(unconfined_t) before mounting
This is only possible after the policy is loaded.

rhatdan: unconfined_t does not exists before policy is loaded, kernel_t also does not exists before policy is loaded.

When policy is loaded, the kernel is told that all procesess identifiers that existed before the policy load are labeled as kernel_t.

rhatdan:
I don't think 
avc:  denied  { create } for  pid=2101 comm="mkdir" name="foo" scontext=system_u:system_r:kernel_t:s0 tcontext=unconfined_u:object_r:etc_t:s0 tclass=dir permissive=1

Is coming from the overlay, if it is, then I don't understand.  The mounter as I understand it should only be requesting read/execute permissions on types, and the creation should be done by the unconfined_t process, since it should be created in the upper directory.

selinux/ostree.te
WOnder93: > unconfined_domain(kernel_t) == ("kernel_t = unconfined_t - execute_no_trans")

No, it is not, because of `files_unconfined_type` and `filesystem_unconfined_type`.

> Well not quite, but I don't think it is an issue.
> allow unconfined_domain_type container_domain:process { dyntransition transition };
> allow unconfined_domain_type container_runtime_t:process transition;
> allow unconfined_domain_type qm_container_domain:process { dyntransition transition };

How is that relevant? Once you allow read + execute + execute_no_trans on all files, then SELinux won't prevent an attacker exploiting the modprobe path vulnerability from running a binary they control (most likely labeled user_tmp_t or user_home_t). When you eliminate execute_no_trans from kernel_t, then the upper bound of what file types it can execute is those that it has a type_transition defined for, which is easy to check and control. And it is best to avoid execute_no_trans even for selected file types, as you would get userspace processes running as root (always - systemd hardening features can't be used here) + kernel_t, which is risky even for binaries considered "safe" (system binaries). Plus the task being labeled kernel_t would be confusing as it's no longer running in kernel space.

alexlarsson: @WOnder93 So, in practical terms, how would one go about creating rules for this?

cgwalters:
> Is coming from the overlay, if it is, then I don't understand. The mounter as I understand it should only be requesting read/execute permissions on types, and the creation should be done by the unconfined_t process, since it should be created in the upper directory.

See https://docs.kernel.org/filesystems/overlayfs.html#permission-model

I saw at one point the maintainer of overlayfs once described it as a "just in time cp -a".  Hence, the mounter credentials are captured and checked for each access in *addition* to the calling task.

So what we're hitting is this:

> task creating the overlay mount MUST NOT gain additional privileges

WOnder93:
> `avc:  denied  { create } for  pid=2101 comm="mkdir" name="foo" scontext=system_u:system_r:kernel_t:s0 tcontext=unconfined_u:object_r:etc_t:s0 tclass=dir permissive=1`

@alexlarsson This AVC is sus, because it should be allowed since commit https://github.com/fedora-selinux/selinux-policy/commit/f5438002d7... What was the selinux-policy RPM version on the system where you got this? In fact, I think that commit should fully suffice for the early boot overlayfs mounts to work.

alexlarsson:
So, I looked into what it would mean to have kernel_t be a fully unconfined domain, and it seems *way* too much. Like its got the ability to talk to systemd and all sorts of stuff. Instead I tried to make the minimal policy that I got to boot the system, and let me create some files in /etc as unconfined_t root.

This is what I came up with:

```
allow kernel_t file_type:dir_file_class_set { read write append map open create link unlink rename getattr setattr relabelfrom relabelto };
allow kernel_t file_type:dir { add_name remove_name reparent rmdir search };

domain_role_change_exemption(kernel_t)
domain_obj_id_change_exemption(kernel_t)
```
(Also in the latest commit here)

Basically, this allows all the file operations that I see overlayfs doing while it is in a section between ovl_override_creds() and revert_creds(). I also had to add the excemptions, because the mount ends up running as system_u:system_r:kernel_t and /etc files may not have those roles or object ids.

This seems a lot safer to me...

WOnder93:
@alexlarsson Please see my previous comment. This is all already in the policy since at least February... (except the `domain_role_change_exemption()`, but that one is not needed here, because files can only have `object_r` as the role).

alexlarsson:
@WOnder93 I am testing with some fedora coreos build that has selinux-policy-38.20-1.fc38

That change is similar to what I ended up with (although i don't think it allows relabeling and maybe some other corner things?). But, I think maybe my problem was not that, but rather the lack of domain_obj_id_change_exemption().

alexlarsson:
@WOnder93 I don't see domain_obj_id_change_exemption() in the commit you pointed at, was it added later?

alexlarsson:
@WOnder93 Also, I need this in centos9, what is the status of this there?

WOnder93:
CentOS 9 doesn't have the usermode helper execution hardening (i.e. `kernel_t` is still a fully unconfined domain there). And if we end up backporting that work there (not sure if that will happen as it would be quite risky...), we would take care to backport the mentioned commits as well, so the overlay mounts should remain working.

alexlarsson:
@WOnder93 We still need relabeling, that was not in the manage_files change.
When booting i get:
```
avc:  denied  { relabelfrom } for  pid=1416 comm="ostree-remount" name="upper" dev="tmpfs" ino=811 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:var_run_t:s0 tclass=dir permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="upper" dev="tmpfs" ino=811 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=1
avc:  denied  { relabelfrom } for  pid=1416 comm="ostree-remount" name="machine-id" dev="tmpfs" ino=922 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:tmpfs_t:s0 tclass=file permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="machine-id" dev="tmpfs" ino=922 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:machineid_t:s0 tclass=file permissive=1
avc:  denied  { relabelfrom } for  pid=1416 comm="ostree-remount" name=".ignition-result.json" dev="tmpfs" ino=874 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:var_run_t:s0 tclass=file permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name=".ignition-result.json" dev="tmpfs" ino=874 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="system" dev="tmpfs" ino=865 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:systemd_unit_file_t:s0 tclass=dir permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="var-mnt-workdir\x2dtmp.mount" dev="tmpfs" ino=871 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:systemd_unit_file_t:s0 tclass=file permiss1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="10-autologin.conf" dev="tmpfs" ino=869 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:getty_unit_file_t:s0 tclass=file permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="profile.d" dev="tmpfs" ino=862 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:bin_t:s0 tclass=dir permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="autoresize.sh" dev="tmpfs" ino=863 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:bin_t:s0 tclass=file permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="gshadow" dev="tmpfs" ino=857 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:shadow_t:s0 tclass=file permissive=1
avc:  denied  { relabelto } for  pid=1416 comm="ostree-remount" name="group" dev="tmpfs" ino=855 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:passwd_file_t:s0 tclass=file permissive=1
avc:  denied  { relabelfrom } for  pid=1400 comm="systemd-hwdb" name="hwdb.bin" dev="tmpfs" ino=1036 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:systemd_hwdb_etc_t:s0 tclass=file permissive=1
avc:  denied  { relabelto } for  pid=1400 comm="systemd-hwdb" name="hwdb.bin" dev="tmpfs" ino=1036 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:systemd_hwdb_etc_t:s0 tclass=file permissive=1
avc:  denied  { relabelfrom } for  pid=1457 comm="systemd-tmpfile" name="console-login-helper-messages-profile.sh" dev="tmpfs" ino=1515 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:bin_t:s0 tclass=lnk_file p1
avc:  denied  { relabelto } for  pid=1457 comm="systemd-tmpfile" name="console-login-helper-messages-profile.sh" dev="tmpfs" ino=1515 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:bin_t:s0 tclass=lnk_file per1
avc:  denied  { relabelfrom } for  pid=1457 comm="systemd-tmpfile" name="credstore" dev="tmpfs" ino=1517 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=1
avc:  denied  { relabelto } for  pid=1457 comm="systemd-tmpfile" name="credstore" dev="tmpfs" ino=1517 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=1
avc:  denied  { relabelfrom } for  pid=1457 comm="systemd-tmpfile" name="os-release" dev="tmpfs" ino=1522 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=lnk_file permissive=1
avc:  denied  { relabelto } for  pid=1457 comm="systemd-tmpfile" name="os-release" dev="tmpfs" ino=1522 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=lnk_file permissive=1
avc:  denied  { relabelfrom } for  pid=1457 comm="systemd-tmpfile" name="locale.conf" dev="tmpfs" ino=1524 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:locale_t:s0 tclass=file permissive=1
avc:  denied  { relabelto } for  pid=1457 comm="systemd-tmpfile" name="locale.conf" dev="tmpfs" ino=1524 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:locale_t:s0 tclass=file permissive=1
avc:  denied  { relabelfrom } for  pid=1457 comm="systemd-tmpfile" name="resolv.conf" dev="tmpfs" ino=1542 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:net_conf_t:s0 tclass=lnk_file permissive=1
avc:  denied  { relabelto } for  pid=1457 comm="systemd-tmpfile" name="resolv.conf" dev="tmpfs" ino=1542 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:net_conf_t:s0 tclass=lnk_file permissive=1
avc:  denied  { relabelfrom } for  pid=1627 comm="restorecon" name="21_clhm_ssh_host_keys.issue" dev="tmpfs" ino=1690 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:NetworkManager_dispatcher_console_var_run_t:1
avc:  denied  { relabelto } for  pid=1627 comm="restorecon" name="21_clhm_ssh_host_keys.issue" dev="tmpfs" ino=1690 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=1
avc:  denied  { relabelfrom } for  pid=1739 comm="restorecon" name="22_clhm_ens4.issue" dev="tmpfs" ino=1750 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:NetworkManager_dispatcher_console_var_run_t:s0 tclas1
avc:  denied  { relabelto } for  pid=1739 comm="restorecon" name="22_clhm_ens4.issue" dev="tmpfs" ino=1750 scontext=system_u:system_r:kernel_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=1
```

The ostree-remount stuff is from the PR, but the other stuff is stuff that should affect everyone. This is with selinux-policy-38.28-1.fc38.noarch.

alexlarsson:
Actually, if i added relabel, i got these too:

```
AVC avc:  denied  { map } for  pid=1379 comm="systemd-network" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_network_generator_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1383 comm="systemd-hwdb" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_hwdb_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1383 comm="systemd-hwdb" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:systemd_hwdb_t:s0 tcontext=system_u:object_r:etc_t:s0 tcla0
AVC avc:  denied  { map } for  pid=1383 comm="systemd-hwdb" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:systemd_hwdb_t:s0 tcontext=system_u:object_r:etc_0
AVC avc:  denied  { map } for  pid=1384 comm="systemd-sysctl" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_sysctl_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1388 comm="systemd-userdbd" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_userdbd_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1390 comm="systemd-userwor" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_userdbd_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1389 comm="systemd-userwor" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_userdbd_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1391 comm="systemd-userwor" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_userdbd_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1392 comm="systemd-tmpfile" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_tmpfiles_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1392 comm="systemd-tmpfile" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:systemd_tmpfiles_t:s0 tcontext=system_u:object_r:etc_t:0
AVC avc:  denied  { map } for  pid=1392 comm="systemd-tmpfile" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:systemd_tmpfiles_t:s0 tcontext=system_u:object0
AVC avc:  denied  { unlink } for  pid=1383 comm="systemd-hwdb" name="hwdb.bin" dev="overlay" ino=15884 scontext=system_u:system_r:systemd_hwdb_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1421 comm="systemd-sysctl" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_sysctl_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1438 comm="systemd-tmpfile" path="/etc/ld.so.cache" dev="overlay" ino=6566 scontext=system_u:system_r:systemd_tmpfiles_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=file permissive=0
AVC avc:  denied  { map } for  pid=1438 comm="systemd-tmpfile" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:systemd_tmpfiles_t:s0 tcontext=system_u:object_r:etc_t:0
AVC avc:  denied  { map } for  pid=1438 comm="systemd-tmpfile" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:systemd_tmpfiles_t:s0 tcontext=system_u:object0
AVC avc:  denied  { map } for  pid=1473 comm="systemd-resolve" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:systemd_resolved_t:s0 tcontext=system_u:object_r:etc_t:0
AVC avc:  denied  { map } for  pid=1473 comm="systemd-resolve" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:systemd_resolved_t:s0 tcontext=system_u:object0
AVC avc:  denied  { watch } for  pid=1485 comm="dbus-broker-lau" path="/etc/dbus-1/system.d" dev="overlay" ino=68 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=0
AVC avc:  denied  { watch } for  pid=1488 comm="dbus-broker-lau" path="/etc/dbus-1/system.d" dev="overlay" ino=68 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=0
AVC avc:  denied  { watch } for  pid=1491 comm="dbus-broker-lau" path="/etc/dbus-1/system.d" dev="overlay" ino=68 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=0
AVC avc:  denied  { watch } for  pid=1494 comm="dbus-broker-lau" path="/etc/dbus-1/system.d" dev="overlay" ino=68 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=0
AVC avc:  denied  { watch } for  pid=1497 comm="dbus-broker-lau" path="/etc/dbus-1/system.d" dev="overlay" ino=68 scontext=system_u:system_r:system_dbusd_t:s0-s0:c0.c1023 tcontext=system_u:object_r:etc_t:s0 tclass=dir permissive=0
AVC avc:  denied  { map } for  pid=1572 comm="restorecon" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:setfiles_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=fil0
AVC avc:  denied  { map } for  pid=1572 comm="restorecon" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:setfiles_t:s0 tcontext=system_u:object_r:etc_t:s0 t0
AVC avc:  denied  { map } for  pid=1573 comm="restorecon" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:setfiles_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=fil0
AVC avc:  denied  { map } for  pid=1573 comm="restorecon" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:setfiles_t:s0 tcontext=system_u:object_r:etc_t:s0 t0
AVC avc:  denied  { map } for  pid=1574 comm="systemd-logind" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:systemd_logind_t:s0 tcontext=system_u:object_r:etc_t:s0 0
AVC avc:  denied  { map } for  pid=1574 comm="systemd-logind" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:systemd_logind_t:s0 tcontext=system_u:object_r:0
AVC avc:  denied  { map } for  pid=1604 comm="systemd-logind" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:systemd_logind_t:s0 tcontext=system_u:object_r:etc_t:s0 0
AVC avc:  denied  { map } for  pid=1604 comm="systemd-logind" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:systemd_logind_t:s0 tcontext=system_u:object_r:0
AVC avc:  denied  { map } for  pid=1620 comm="restorecon" path="/etc/selinux/targeted/contexts/files/file_contexts.bin" dev="overlay" ino=115780 scontext=system_u:system_r:setfiles_t:s0 tcontext=system_u:object_r:etc_t:s0 tclass=fil0
AVC avc:  denied  { map } for  pid=1620 comm="restorecon" path="/etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin" dev="overlay" ino=115792 scontext=system_u:system_r:setfiles_t:s0 tcontext=system_u:object_r:etc_t:s0 t0
```
I'll try to come up with a minimal set of things needed on top of selinux-policy-38.28 tomorrow.
I

selinux/ostree.te
cgwalters: One thing that's somewhat tangentially related to this is I think usermode helper execution is a bit better constrained via other LSMs, like e.g. [LoadPin](https://lwn.net/Articles/895781/) etc...that's something we can more strongly push for in an image-based world.

alexlarsson:
Sorry, the later avcs was wrong, as i forgot to relabel usr/etc. But I update the PR with the minimal ruleset we need added:

```
files_relabel_all_files(kernel_t)
manage_chr_files_pattern(kernel_t, file_type, file_type)
manage_blk_files_pattern(kernel_t, file_type, file_type)
```

I actually didn't hit blk_file management in the boot, just chr_file, but it seems weird to just add the one without the other.

alexlarsson:
@WOnder93 Would adding these to selinux-policy make sense?

WOnder93:
Yes, we should add that to the system policy - I opened a PR: https://github.com/fedora-selinux/selinux-policy/pull/1893

cgwalters:
> Can we also get list of services started from initramfs which need to persist switchroot?

@zpytela I'm pretty sure there are zero userspace processes which persist today; it's *possible* we'd add something in the future but I'd like to avoid it.  (And if we do we can probably have it save its own state and re-exec after policy load or so?)

The "thing that persists from the initramfs" here is the overlayfs credentials cached in the kernel.

alexlarsson:
Unmarked this as draft, as the selinux part now works out of the box (although needs selinux policy master for now). It should however work as is with older selinux-policy that have a fully unconfined kernel_t including, i believe, rhel9.

alexlarsson:
I don't undestand the "Adding composefs metadata: composefs is not supported in this ostree build" CI failures. These checks passed before, did we end up with an older ostree version in the CI?

cgwalters:
See https://github.com/ostreedev/ostree/pull/3067

zpytela:
FYI the selinux-policy rawhide build is now available at https://bodhi.fedoraproject.org/updates/FEDORA-2023-5be6d65113

alexlarsson:
After rebase it passed CI, @cgwalters can you review it now?

cgwalters:
I just merged https://github.com/ostreedev/ostree/pull/3063 - this will need to drop the other patch for it now.

src/switchroot/ostree-remount.c
```diff
@@ -119,6 +166,52 @@ main (int argc, char *argv[])
...
+            err (EXIT_FAILURE, "warning: While remounting /etc MS_PRIVATE");
+
+          if (umount2 ("/etc/machine-id", MNT_DETACH) < 0)
+            err (EXIT_FAILURE, "Failed to unmount machine-id");
+        }
+
+      g_autofree char *upper = g_build_filename (transient_etc, "upper", NULL);
+      relabel_dir_for_upper (upper, "/etc", TRUE);
+
+      if (initial_ns_fd != -1 && setns (initial_ns_fd, CLONE_NEWNS) < 0)
```
cgwalters, line 211: Ahhh I see.  OK.  Nevermind

src/switchroot/ostree-remount.c
```diff
@@ -76,6 +79,50 @@ do_remount (const char *target, bool writable)
...
 
+/* Relabel the directory $real_path, which is going to be an overlayfs mount,
+ * based on the content of an overlayfs upperdirectory that is in use by the mount.
+ * The goal is that we relabel in the overlay mount all the files that have been
+ * modified (directly or via parent copyup operations) since the overlayfs was
+ * mounted. This will be used for the /etc overlayfs mount where no selinux labels
+ * are set before selinux policy is loaded.
+ */
+static void
+relabel_dir_for_upper (const char *upper_path, const char *real_path, gboolean is_dir)
```
cgwalters, line 90: Do we still need this though if we land https://github.com/coreos/rpm-ostree/pull/4640 and use it?

alexlarsson: Yes. The problem is not the labels on the lowerdir, those are fixed by that commit. The problem is the labels in upperdir and in the overlayfs itself (i.e. in-memory overlayfs inodes cache the selinux label from the upper).

Basically, in the initramfs all the files created in the tmpfs upper will get their xattrs reset at selinux policy load. So, what we do is to look at upperdir (in tmpfs), and for any file that exists there (and there should be few, only for files created in initramfs), we relabel the corresponding overlayfs file via /etc. This will set the correct selinux label *both* on the in-memory overlayfs inode, and the backing upperdir file.

alexlarsson: Note: We still need that commit tho.

cgwalters (approved):
Nice work!

I think we can add an installed test for this, but not gating.

alexlarsson:
@cgwalters The installed test will probably be easier once the selinux changes have propagated too.

zpytela:
> @cgwalters The installed test will probably be easier once the selinux changes have propagated too.

Which changes do you have in mind? Rawhide selinux-policy build is available since 3 days ago.

alexlarsson:
@zpytela Yes, but it has to propagate to the fedora coreos images used in the CI here.

cgwalters:
I want to try spinning up testing against ELN across our repositories soon, which will fix that.

PR #3067, opened 2023-10-05
ci: Ensure composefs is enabled on Fedora

cgwalters:
For some reason we're not picking this up in the Prow build, which breaks things because now rpm-ostree hard requires it.

Let's make this a fatal build time error for more clear debugging.

PR #3071, opened 2023-10-10
configure: fix composefs default to on, if appropriate

ericcurtin:
composefs should be included by default. If the appropriate headers are
not present, have_mount_attr_idmap should be set to no. If this is no,
don't build composefs as part of the build.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -284,7 +284,8 @@ AC_COMPILE_IFELSE(
 	)],
 	[AC_MSG_RESULT(yes)
 	 have_mount_attr_idmap=yes],
-	[AC_MSG_RESULT(no)])
+	[AC_MSG_RESULT(no)
+	 have_mount_attr_idmap=no])
 dnl These are needed by libcomposefs to use the new mount API optionally
 AC_MSG_CHECKING([for new mount API (fsconfig)])
 AC_COMPILE_IFELSE(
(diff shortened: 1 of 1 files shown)


alexlarsson:
I don't understand how this changes anything? Like we were checking  `x"$have_mount_attr_idmap" != xyes`, and if $have_mount_attr_idmap is unset we should have taken this if, because `x != xyes`, no?

ericcurtin:
I suspect have_mount_attr_idmap was never getting set correctly, so composefs_default was always being set as no. Because != yes covers the unset case also. I'm not confident that this is the best fix, this was tested, after this change composefs is on because checking for == no is a more explicit stricter check than != yes.

But a full test would be to test on an older kernel also, to see if it's no in that case, an explanation of why this conditional stuff was introduced is here:

https://github.com/ostreedev/ostree/pull/2998

ericcurtin:
The build being red probably proves composefs is always on with this change. Maybe I can take a look at this when I'm back, although I'm not an autotools expert.

cgwalters (approved):
I just tested this quickly, running a default build without `--with-composefs` does enable composefs.

The real problem is in https://github.com/ostreedev/ostree/pull/3067

But I'm fine to land this too, just needs a rebase.

But, this change is perhaps clearer so I'm also fine to land it.

ericcurtin:
You are right @cgwalters ... Closing...

PR #3073, opened 2023-10-12
deploy: Improve error message for nonexistent stateroot

cgwalters:
Came up on an internal chat; previously we were only erroring out when trying to do the SELinux labeling for `/var` which was really misleading.

Add some other error prefixing while we have the patient open.


jmarrero (approved):
lgtm

PR #3077, opened 2023-10-13
deploy: Remove lock when re-staging

cgwalters:
This closes the biggest foot-gun when doing e.g.
`rpm-ostree rebase` when zincati is running on a FCOS system.

Previously if zincati happened to have staged + locked a deployment, we'd keep around the lock which is definitely not what is desired.


jmarrero (approved):
lgtm

PR #3078, opened 2023-10-17
karg-delete: support multiple times

HuijingHei:
Fix https://github.com/ostreedev/ostree/pull/2612#issuecomment-1764412141


src/ostree/ot-admin-builtin-deploy.c
```diff
@@ -207,8 +207,12 @@ ot_admin_builtin_deploy (int argc, char **argv, OstreeCommandInvocation *invocat
 
   if (opt_kernel_argv_delete)
```
cgwalters, line 208: Minor nit; this conditional can now be dropped because it's redundant with the `for` loop.

cgwalters (approved):

cgwalters (approved):

PR #3080, opened 2023-10-19
lib/bootloader-zipl: Check for Secure Boot before zipl

nikita-dubrovskii:
Starting with z/VM Version 7 Release 3.0, service level 2301 we can run Linux with Secure Boot enabled

### DIFF
--- a/src/libostree/ostree-bootloader-zipl.c
+++ b/src/libostree/ostree-bootloader-zipl.c
@@ -35,6 +35,7 @@
 #define SECURE_EXECUTION_LUKS_ROOT_KEY "/etc/luks/root"
 #define SECURE_EXECUTION_LUKS_BOOT_KEY "/etc/luks/boot"
 #define SECURE_EXECUTION_LUKS_CONFIG "/etc/crypttab"
+#define SECURE_BOOT_SYSFS_FLAG "/sys/firmware/ipl/secure"
 
 #if !(defined HAVE_LIBARCHIVE) && defined(__s390x__)
 #error libarchive is required for s390x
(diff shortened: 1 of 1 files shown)


nikita-dubrovskii:
Here is a how-to: https://github.com/coreos/enhancements/pull/10

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -112,6 +113,47 @@ _ostree_bootloader_zipl_write_config (OstreeBootloader *bootloader, int bootvers
...
+  // [    0.023200] setup: 0000000000880000 - 0000000003f98000 (not signed)
+  int status = 0;
+  const char *const cmd[] = { "dmesg", "--kernel", NULL };
+  g_autofree gchar *output = NULL;
+  if (!g_spawn_sync (NULL, (char **)cmd, NULL, G_SPAWN_SEARCH_PATH, NULL, NULL, &output, NULL,
+                     &status, error))
+    return glnx_prefix_error (error, "spawning dmesg");
+  if (!g_spawn_check_exit_status (status, error))
+    return glnx_prefix_error (error, "`dmesg` failed");
+  *out_enabled = (output != NULL && strstr (output, "Secure-IPL enabled") != NULL);
```
jlebon, line 152: Feels weird to be grepping for strings in logs for something like Secure Boot. I think `--kernel` prevents a simple e.g. `echo Secure-IPL enabled > /dev/kmsg` from throwing this off, but my vote would be to stick with the sysfs flag and open a issue against RHEL 9 to have it backported.

nikita-dubrovskii: Dropped `dmesg` and now read `kmsg` directly. I don't know when kernel fix comes, so it'd be better to have this workaround until former happens.

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -112,6 +113,51 @@ _ostree_bootloader_zipl_write_config (OstreeBootloader *bootloader, int bootvers
...
+  // [    0.023194] setup: 0000000000009000 - 000000000000a000 (not signed)
+  // [    0.023196] setup: 000000000000a000 - 000000000000e000 (signed, verified)
+  // [    0.023197] setup: 0000000000010000 - 0000000000866000 (signed, verified)
+  // [    0.023198] setup: 0000000000867000 - 0000000000868000 (not signed)
+  // [    0.023199] setup: 0000000000877000 - 0000000000878000 (not signed)
+  // [    0.023200] setup: 0000000000880000 - 0000000003f98000 (not signed)
+  fd = openat (AT_FDCWD, "/dev/kmsg", O_NONBLOCK | O_RDONLY);
+  if (fd == -1)
+    return glnx_throw_errno_prefix (error, "openat(/dev/kmsg)");
+  int max_lines = 5; // no need to read dozens of messages, ours comes really early
```
cgwalters, line 148: Let's make this an `unsigned int`.

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -112,6 +113,51 @@ _ostree_bootloader_zipl_write_config (OstreeBootloader *bootloader, int bootvers
...
+  while (*out_enabled != TRUE && --max_lines)
```
cgwalters, line 149: My general preference here is to avoid side effects in iteration, let's do e.g. `max_lines > 0` here, and then:

src/libostree/ostree-bootloader-zipl.c
```diff
@@ -112,6 +113,51 @@ _ostree_bootloader_zipl_write_config (OstreeBootloader *bootloader, int bootvers
...
+    {
+      char buf[1024];
+      ssize_t len = read (fd, buf, sizeof (buf));
+      if (len == -EAGAIN)
+        break;
+      *out_enabled = strstr (buf, "Secure-IPL enabled") != NULL;
```
cgwalters, line 155: Add `max_lines -= 1;` here.

cgwalters (approved):
Looks sane, just some minor nits.

cgwalters:
One thing I don't quite understand btw...isn't zipl in the position to check for secure boot itself?  

Would it make more sense for the code here to *always* run `zipl --auto` itself?

cgwalters:
Followup fixes in https://github.com/ostreedev/ostree/pull/3119 - can you review?

PR #3081, opened 2023-10-19
lib/deploy: Log SELinux policy refresh

jlebon:
I have a suspicion that the `semodule -N --refresh` we do here is involved in https://github.com/coreos/fedora-coreos-tracker/issues/1597. Let's log when we execute it and include its time.

While we're here, also log the time it takes to `syncfs()` and the fsfreeze/thaw cycling. It's logged in a structured journal entry, but that's less accessible than just having it in the journal message field.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1646,25 +1646,28 @@ full_system_sync (OstreeSysroot *self, SyncStats *out_stats, GCancellable *cance
                   GError **error)
 {
   GLNX_AUTO_PREFIX_ERROR ("Full sync", error);
-  guint64 start_msec = g_get_monotonic_time () / 1000;
   ot_journal_print (LOG_INFO, "Starting syncfs() for system root");
+  guint64 start_msec = g_get_monotonic_time () / 1000;
   if (syncfs (self->sysroot_fd) != 0)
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1646,25 +1646,28 @@ full_system_sync (OstreeSysroot *self, SyncStats *out_stats, GCancellable *cance
...
   GLNX_AUTO_PREFIX_ERROR ("Full sync", error);
-  guint64 start_msec = g_get_monotonic_time () / 1000;
   ot_journal_print (LOG_INFO, "Starting syncfs() for system root");
+  guint64 start_msec = g_get_monotonic_time () / 1000;
   if (syncfs (self->sysroot_fd) != 0)
     return glnx_throw_errno_prefix (error, "syncfs(sysroot)");
-  ot_journal_print (LOG_INFO, "Completed syncfs() for system root");
   guint64 end_msec = g_get_monotonic_time () / 1000;
+  ot_journal_print (LOG_INFO, "Completed syncfs() for system root in %" G_GUINT64_FORMAT " ms",
+                    end_msec - start_msec);
```
cgwalters, line 1655: This is completely fine but I previously changed it because in the end the journal keeps timestamps so one can just diff those to see.

cgwalters (approved):

Issue #3088, opened 2023-11-07
Sanitize input of ostree admin undeploy


jmarrero:
Adding a string instead of an index results in ostree undeploy deployment 0.
We should error out.
```
jmarrero@fedora ~ [1]> sudo ostree admin undeploy trash_input
[sudo] password for jmarrero: 
Transaction complete; bootconfig swap: yes; bootversion: boot.1.0, deployment count change: -1
Freed objects: 660.2 kB
Deleted deployment 48f4dc9dc1ef0ac8e990cdcc2fd693bcf6506b5abc578e32703c1e463469ac22.0

```

This came from: https://github.com/openshift-kni/lifecycle-agent/pull/25#discussion_r1384247963

PR #3089, opened 2023-11-08
undeploy: Parse integer more properly

cgwalters:
`atoi` doesn't offer any error checking.

Closes: https://github.com/ostreedev/ostree/issues/3088


ericcurtin (approved):
LGTM

Issue #3025, opened 2023-08-31
Promote deployment finalization to stable


cgwalters:
In https://github.com/ostreedev/ostree/pull/1841 we added "locking" for deployment finalization.  It's way past time to promote this to a stable API.

One thing I think would be helpful here is to have the deployment lock change to become a *symlink* to the deployment that's locked.  If that has changed, the lock no longer applies.  This would solve an issue I continually hit:

- `systemctl stop zincati`
- rpm-ostree rebase
- systemctl reboot

And discover the rebase didn't work because we had a lingering lockfile.

Something like `ostree admin unlock-finalize` or so, and perhaps `ostree admin reboot --execute` that would itself invoke `systemctl reboot`.  Incidentally it may be nice to lower the "check for systemd inhibitors" here too.

This is a bit more important now because I'd like to change things to have the deployments be source-of-truth (xref https://github.com/coreos/rpm-ostree/issues/4391 )

Basically, this would make it all much easier to set up a flow (like zincati) which does:

- pull and stage (*and lock*) updates
- opt-in unlock+apply

jlebon:
Might be worth updating the description here now that we have https://github.com/ostreedev/ostree/pull/3077.

PR #3090, opened 2023-11-08
sysroot: Stabilize deployment finalization, add API

cgwalters:
It's about time we do this; deployment finalization locking is a useful feature.  An absolutely key thing here is that we've slowly been moving towards the deployments as the primary "source of truth".

Specifically in bootc for example, we will GC container images not referenced by a deployment.

This is then neecessary to support a "pull but don't apply automatically" model.

Closes: https://github.com/ostreedev/ostree/issues/3025


ericcurtin (approved):
LGTM

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3722,11 +3726,22 @@ _ostree_sysroot_finalize_staged_inner (OstreeSysroot *self, GCancellable *cancel
...
-    return FALSE;
-  if (errno == 0)
+  gboolean locked = false;
+  (void)g_variant_lookup (self->staged_deployment_data, _OSTREE_SYSROOT_STAGED_KEY_LOCKED, "b",
+                          &locked);
+  if (locked)
+    g_debug ("staged is locked via metadata");
+  else
+    {
+      if (!glnx_fstatat_allow_noent (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED_LOCKED, NULL, 0,
```
jlebon, line 3736: As part of stabilization, shouldn't we make this file path public? And document how to unlock it.

Also, should we add `ostree admin finalize-staged --unlock`?

cgwalters: Right thanks, this was not fully baked.  I've updated this more, though needs some testing.

cgwalters: > As part of stabilization, shouldn't we make this file path public? 

Here we're moving the locked state into the staged deployment data itself (which would have avoided the earlier problem) but it does mean we need an API and CLI to mutate that state.

man/ostree-admin-unlock-finalization.xml
```diff
@@ -0,0 +1,80 @@
...
+    </refsynopsisdiv>
+
+    <refsect1>
+        <title>Description</title>
+
+        <para>
+            This command requires a staged deployment.  If it has been "finalization locked",
+            which means it will not be queued for the next boot by default, then this command
+            will unlock it.  If it is already finalization-unlocked, then this command will
+            be a no-op.
```
jlebon, line 63: The API implemented below isn't idempotent as described here.

man/ostree-admin-unlock-finalization.xml
```diff
@@ -0,0 +1,80 @@
...
+Lesser General Public License for more details.
+
+You should have received a copy of the GNU Lesser General Public
+License along with this library. If not, see <https://www.gnu.org/licenses/>.
+-->
+
+<refentry id="ostree">
+
+    <refentryinfo>
+        <title>ostree admin unlock-finalization</title>
```
jlebon, line 27: Should we doc the `--lock-finalization` option in the `ostree admin deploy` page and  cross-reference this manpage/command?

jlebon (commented):
Looks sane overall! CI failure looks legit.

cgwalters:
> CI failure looks legit.

Hmm yes possibly, though I'm not reproducing this here.  One thing this really brings to the fore is the huge messiness around gpg verification vs direct deployments and how `ostree admin status` bails out with an command-fatal error versus what rpm-ostree does in just showing the error as part of state gathering.

man/ostree-admin-deploy.xml
```diff
@@ -121,6 +121,16 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 </para></listitem>
             </varlistentry>
 
+            <varlistentry>
+                <term><option>--lock-finalization</option></term>
+
+                <listitem><para>
+                    The deployment will not be "finalized" by default on shutdown; to later
+                    queue it, use <literal>ostree admin unlock-finalization</literal>.
```
jlebon, line 129:
```suggestion
                    queue it, use <literal>ostree admin lock-finalization --unlock</literal>.
```

man/ostree-admin-lock-finalization.xml
```diff
@@ -0,0 +1,92 @@
...
+
+    <refsect1>
+        <title>Description</title>
+
+        <para>
+            This command requires a staged deployment.  By default, this command ensures the deployment
+            will be set into a "finalization locked" state, which means it will not be queued for the next boot by default.
+        </para>
+        <para>
+            This is the same as the <literal>--lock-finalization</literal> argument for <literal>ostree admin deploy</literal>.
```
jlebon, line 64: Feels like we should be more explicit about the race condition of not doing this. Maybe e.g.

```suggestion
            This is the same as the <literal>--lock-finalization</literal> argument for <literal>ostree admin deploy</literal>, which is the recommended way to use this feature in a race-free way.
```

?

jlebon (approved):
Nice!

Some docs comments, but fine to leave as follow-ups to not waste CI.

man/ostree-admin-deploy.xml
cgwalters: Good catch, thanks for the careful review!

cgwalters:
Followup in https://github.com/ostreedev/ostree/pull/3100

PR #3094, opened 2023-11-13
Add `ostree admin post-copy` command

alexlarsson:
This command will apply fs-verity on all objects that need it and needs to be called when an ostree deployment has been copied on a file-by-file basis, which would loose information such as fs-verity.

This is needed by osbuild which works by creating the final image in a rootfs, and then separately copying that rootfs file-by-file to a loopback mounted filesystem image.

This comes from https://github.com/osbuild/osbuild/pull/1343

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -175,9 +175,9 @@ endif # USE_GPGME
 symbol_files = $(top_srcdir)/src/libostree/libostree-released.sym
 
 # Uncomment this include when adding new development symbols.
-#if BUILDOPT_IS_DEVEL_BUILD
-#symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
-#endif
+if BUILDOPT_IS_DEVEL_BUILD
+symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
+endif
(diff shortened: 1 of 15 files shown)


cgwalters (approved):
This looks generally sane.  I would like osbuild ultimately to stop doing the copy though and let ostree do what it was designed to do...

cgwalters:
CI failing on
> make: *** No rule to make target 'man/ostree-admin-post-copy.xml', needed by 'man/html/ostree-admin-post-copy.html'.  Stop.

alexlarsson:
> This looks generally sane. I would like osbuild ultimately to stop doing the copy though and let ostree do what it was designed to do...

That is quite a rearchitecture of osbuild though.

alexlarsson:
It seems fcos-e2e fails with:
```
+ ostree --version
ostree: /lib64/libc.so.6: version `GLIBC_2.38' not found (required by ostree)
```
Do we have some version skew in the CI?

cgwalters:
> Do we have some version skew in the CI?

Yeah see https://github.com/coreos/coreos-assembler/pull/3653
/override ci/prow/fcos-e2e

PR #3098, opened 2023-11-21
ci: cancel previous build on PR update

jlebon:
This is an easy way to save CI resources; when a PR is updated, abort any previous build for that PR to focus on testing the latest push.


dustymabe (approved):
LGTM

PR #3101, opened 2023-11-28
main: Drop some redundant includes of `ot-main.h`

cgwalters:
When backporting a patch recently we hit a non-obvious dependency on another fix for `ot-main.h` includes.  Clean this up a bit by dropping the redundant includes.


jmarrero (approved):
/lgtm

thank you so much for this

PR #3103, opened 2023-11-30
tmpfiles: Copy `/usr/share/factory/var` to `/var`

cgwalters:
This is a pattern we want to encourage.  It's honestly just way simpler than what rpm-ostree is doing today in auto-synthesizing individual tmpfiles.d snippets.


cgwalters:
Pairs with https://github.com/ostreedev/ostree-rs-ext/pull/569

jmarrero (approved):
lgtm

PR #3106, opened 2023-12-02
commit: Try reflinks for local commits by default

cgwalters:
I think we originally used to do this, but at some point in a code refactoring, this optimization got lost.

It's a quite important optimization for the case of writing content generated by an external system into an ostree repository.


src/libostree/ostree-repo-commit.c
```diff
@@ -601,40 +601,44 @@ create_regular_tmpfile_linkable_with_content (OstreeRepo *self, guint64 length,
...
-  if (G_IS_FILE_DESCRIPTOR_BASED (input))
+  // Try to do a reflink if possible; if we hit this case we're operating on trusted local input.
+  gboolean did_clone = FALSE;
+  if (G_IS_FILE_DESCRIPTOR_BASED (original_input))
     {
-      int infd = g_file_descriptor_based_get_fd ((GFileDescriptorBased *)input);
-      if (glnx_regfile_copy_bytes (infd, tmpf.fd, (off_t)length) < 0)
-        return glnx_throw_errno_prefix (error, "regfile copy");
+      int infd = g_file_descriptor_based_get_fd ((GFileDescriptorBased *)original_input);
+      if (ioctl (tmpf.fd, FICLONE, infd) == 0)
```
cgwalters, line 609: While `glnx_regfile_copy_bytes` will try reflinks, that will never work because we've wrapped the input with a checksum stream.  And actually since we have to checksum, we can't do `copy_file_range` either, etc.

Ultimately this issue will be moot with a switch to composefs where we'll use fsverity in kernel to do checksums.

jmarrero (approved):

PR #3107, opened 2023-12-03
tests: Drop bare-user on tmpfs negative test

cgwalters:
As of https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=2daf18a7884dc03d5164ab9c7dc3f2ea70638469 the restriction is lifted, FINALLY.


jlebon:
Ahhh, is support for it gated behind a config? We could grep for it and skip it if it's not on.

jlebon (approved):
> > Ahhh, is support for it gated behind a config?
> 
> I don't think so, see the linked commit. It changed to unconditional support.

In that case, I don't understand https://github.com/ostreedev/ostree/pull/3107#issuecomment-1838833464.

I'm just pointing out that this test can trivially be turned from a negative to a positive test.

> Is it really worth keeping our CI broken over this?

Not really. I hadn't realized this was blocking CI. Feel free to merge!

cgwalters:
> I'm just pointing out that this test can trivially be turned from a negative to a positive test.

Not if we want to run the tests on c9s, where that commit isn't backported right?

PR #3108, opened 2023-12-03
Switch to external composefs

cgwalters:
Since there's now a stable shared library, let's use it.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -18,7 +18,7 @@ stage("Build") {
           git fetch origin --tags
           git submodule update --init
 
-          env MAKE_JOBS=${n} SKIP_INSTALLDEPS=1 ./ci/build.sh
+          env MAKE_JOBS=${n} ./ci/build.sh
         """)
       }
       stage("Unit tests") {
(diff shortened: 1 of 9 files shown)


ericcurtin (approved):

cgwalters:
I've belatedly realized that not having composefs shipped in current c9s is going to make this more logistically complicated.  I think that will likely happen, but hard to say when.

alexlarsson:
Yeah, for practical reason, maybe hold off on this a bit.

cgwalters:
OK composefs is in c9s, rebased :surfer:

alexlarsson (approved):

cgwalters:
Yeah, various CI failures because our build/runtime deps come from packaging and that breaks when dropping in new binaries.  I did <https://src.fedoraproject.org/rpms/ostree/c/b64d73f3718115cb9d6926275e21f5b16c3f14f9?branch=rawhide> as one fix.

cgwalters:
Ah, this needs https://github.com/coreos/fedora-coreos-config/pull/2900 at least

Issue #3113, opened 2023-12-07
Add support for full overlayfs for `/`


cgwalters:
It'd greatly improve compatibility with things like RPMs that install in `/opt` if we supported a full "original docker" style model where `/` is a transient overlayfs.  We'd still keep our semantics for `/etc` and `/var` by default, but e.g. we'd stop recommending `/opt` :arrow_right: `/var/opt`, so `/opt` would be on the overlayfs.

Note this all aligns with composefs, where we'd actually be making `/` a *read-only* overlayfs by default; it'd be really nice of course to *implement* this by just making the composefs overlayfs writable, but I am not sure we can hard require composefs for this right now.

Something like 
```
root.transient = "true"
```

in `/usr/lib/ostree/prepare-root.conf`.

## Downsides

The major downside is that people could be surprised if files they write to e.g. `/opt` don't persist across upgrades.  But, that's already again how it works since Docker started.

alexlarsson:
I worry about this in general, because it breaks a basic container-like workload, because you can't (safely) rebase to a new version while keeping the existing overlay uppper-dir. (Any random change to a file will always replace a new version of the file from the image.

This works in containers, because you start new containers whenever you rebase, which means throwing away the upper layer. Do we intend to do the same here?

cgwalters:
@alexlarsson On upgrade indeed the old overlayfs data will be gone.  But that's how it works since Docker first came out.

alexlarsson:
> I think applications like crowstrike will struggle as their client performs a type of "activation" and maintains state with the controller all under the /opt/crowdstrike/ directory. ...but if I understand this correctly the software should still work, but they'd have to reestablish that activation with each os update. I _think_ we can make this work, but this is an area where more feedback will hopefully help.

I feel like this would be better solved by adding to the image a bind mount uniit from /opt/crowdstrike into /var instead, so that the activation can be persisted? Or am I missing something here?

mrguitar:
I agree that would be much better. I think the challenge is mainly around users knowing these details like this about their applications, and also knowing rpm-ostree/ostree. Docker makes them care more about where things are persisted, traditional rpm systems do not. 

Earlier today I think @cgwalters or @stefwalter mentioned using the VOLUME key in the containerfile to help. Would it makes sense to create bindmounts like this, or something similar, for volumes that users declare?

alexlarsson:
[...]

PR #3114, opened 2023-12-07
prepare-root: Add support for root.transient

cgwalters:
sysroot: Drop unused prototype

This function doesn't exist anymore.

---

prepare-root: Add an autofree

This doesn't matter at all, but I just noticed this while working
on the code.

---

prepare-root: Add support for root.transient

Closes: https://github.com/ostreedev/ostree/issues/3113

It'd greatly improve compatibility with things like RPMs that install
in `/opt` if we supported a full "original docker" style model where
`/` is a transient overlayfs.  We'd still keep our semantics for `/etc`
and `/var` by default, but e.g. we'd stop recommending
`/opt` :arrow_right: `/var/opt`, in this model,
so `/opt` would be on the overlayfs.

Note this all aligns with composefs, where we'd actually be making
`/` a *read-only* overlayfs by default; it'd be really nice of course
to *implement* this by just making the composefs overlayfs writable,
but I am not sure we can hard require composefs for this right now.

So this change adds support for `root.transient = true`
in `/usr/lib/ostree/prepare-root.conf`.

The major downside is that people could be surprised if files they
write to e.g. `/opt` don't persist across upgrades.  But, that's
already again how it works since Docker started.

Note as part of the implementation of this, we need to add a whole
new "backing" directory distinct from the deployment directories.

(Tangentially related to this, it's tempting to switch to always
 using a *read-only* overlay mount by default.

---

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -3078,6 +3078,10 @@ sysroot_initialize_deployment (OstreeSysroot *self, const char *osname, const ch
   if (!require_stateroot (self, osname, error))
     return FALSE;
 
+  g_autofree char *stateroot_backing = g_strdup_printf ("ostree/deploy/%s/backing", osname);
+  if (!glnx_shutil_mkdir_p_at (self->sysroot_fd, stateroot_backing, 0700, cancellable, error))
+    return glnx_prefix_error (error, "Creating backing directory");
+
   OstreeRepo *repo = ostree_sysroot_repo (self);
 
   gint new_deployserial;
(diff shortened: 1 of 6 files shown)


cgwalters:
So this seems to work well:

```
[root@cosa-devsh ~]# findmnt /
TARGET
  SOURCE  FSTYPE  OPTIONS
/ overlay overlay rw,relatime,seclabel,lowerdir=.,upperdir=../../backing/4b42f60ddbb87dbcabdf17e15b1414234ce03eefab0b2f0c2e7972dc1bb816bf.0/root-transient/upper,workdir=../../backing/4b42f60ddbb87dbcabdf17e15b1414234ce03eefab0b2f0c2e7972dc1bb816bf.0/root-transient/work,uuid=on
```

Then I did `rm /opt && mkdir /opt`, changes there persist across reboot.

A next step here is to change rpm-ostree to detect this configuration and [stop doing the opt :arrow_right: var/opt](https://github.com/coreos/rpm-ostree/blob/fc39c53715e665a0ca12dac7dad1c0626f0dceb5/rust/src/composepost.rs#L74) link.

cgwalters:
This also wants https://github.com/ostreedev/ostree-rs-ext/pull/570

cgwalters:
BTW this change *also* requires https://github.com/coreos/coreos-installer/pull/1203 for FCOS for the same reason as composefs.
And will also trip over [os-prober failing](https://github.com/coreos/rpm-ostree/issues/4664#issuecomment-1843618120).  os-prober is just the worst :cry: 
Anyways though, I think I'd like to just get this (and all its dependent patches) in since it should be inert code until explicitly turned on ([example](https://github.com/CentOS/centos-bootc/pull/98)).

rhatdan:
LGTM
@vrothberg @jlebon PTAL

ericcurtin (approved):
LGTM

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3360,6 +3407,9 @@ sysroot_finalize_deployment (OstreeSysroot *self, OstreeDeployment *deployment,
   if (!selinux_relabel_var_if_needed (self, sepolicy, os_deploy_dfd, cancellable, error))
     return FALSE;
 
+  if (!sysroot_initialize_deployment_backing (self, deployment, sepolicy, error))
```
jlebon, line 3410: Shouldn't this be gated on `root.transient`?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3078,6 +3078,10 @@ sysroot_initialize_deployment (OstreeSysroot *self, const char *osname, const ch
   if (!require_stateroot (self, osname, error))
     return FALSE;
 
+  g_autofree char *stateroot_backing = g_strdup_printf ("ostree/deploy/%s/backing", osname);
+  if (!glnx_shutil_mkdir_p_at (self->sysroot_fd, stateroot_backing, 0700, cancellable, error))
```
jlebon, line 3082: And this too.

src/libostree/ostree-sysroot-deploy.c
cgwalters: You're right to call this out, but this was semi-intentional because one thing I want to do is change `usroverlay` to write here instead of the `/var/tmp` it does today - because that would just be better; the content would be clearly "lifecycle bound" to the deployment and we wouldn't need that hacky `tmpfiles.d` snippet we carry:

```
# https://github.com/ostreedev/ostree/issues/393
R! /var/tmp/ostree-unlock-ovl.*
```

(That said to do that note we'd need to also change some logic to ensure the overlay content for it is actually removed on reboot by default)

This all said, yes we could still create it dynamically in either case.  I think we actually *should* have logic here in ostree to do what is being replicated in ostree-rs-ext and rpm-ostree around querying the initramfs config.  Then we could drop the `ex-integrity.composefs` repo config.

cgwalters:
BTW for people seeing this now, it's also worth looking at https://github.com/ostreedev/ostree/pull/3120

travier:
It's unclear to me reading the commit message and the issues here what exactly happens when we turn this knob on. Could you write a bit of docs about it? Thanks

PR #3120, opened 2023-12-14
Add concept of state overlays

jlebon:
In the OSTree model, executables go in `/usr`, state in `/var` and configuration in `/etc`. Software that lives in `/opt` however messes this up because it often mixes code *and* state, making it harder to manage.

More generally, it's sometimes useful to have the OSTree commit contain code under a certain path, but still allow that path to be writable by software and the sysadmin at runtime (`/usr/local` is another instance).

Add the concept of state overlays. A state overlay is an overlayfs mount whose upper directory, which contains unmanaged state, is carried forward on top of a lower directory, containing OSTree-managed files.

In the example of `/usr/local`, OSTree commits can ship content there, all while allowing users to e.g. add scripts in `/usr/local/bin` when booted into that commit.

Some reconciliation logic is executed whenever the base is updated so that newer files in the base are never shadowed by a copied up version in the upper directory. This matches RPM semantics when upgrading packages whose files may have been modified.

For ease of integration, this is exposed as a systemd template unit which any downstream distro/user can enable. The instance name is the mountpath in escaped systemd path notation (e.g. `ostree-state-overlay@usr-local.service`).

See discussions in https://github.com/ostreedev/ostree/issues/3113 for more details.

### DIFF
--- a/src/ostree/ot-admin-builtin-state-overlay.c
+++ b/src/ostree/ot-admin-builtin-state-overlay.c
@@ -1,0 +1,284 @@
+/* Copyright (C) 2023 Red Hat, Inc.
+ *
+ * SPDX-License-Identifier: LGPL-2.0+
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <https://www.gnu.org/licenses/>.
+ */
+
+#include "config.h"
+
+#include <fcntl.h>
+#include <glib-unix.h>
+#include <sched.h>
+#include <stdlib.h>
+#include <sys/mount.h>
+
+#include "glnx-errors.h"
+#include "glnx-fdio.h"
+#include "glnx-local-alloc.h"
+#include "glnx-shutil.h"
+#include "glnx-xattrs.h"
+#include "ostree-core.h"
+#include "ostree-deployment.h"
+#include "ot-admin-builtins.h"
+
+#define OSTREE_STATEOVERLAYS_DIR "/var/ostree/state-overlays"
+#define OSTREE_STATEOVERLAY_UPPER_DIR "upper"
+#define OSTREE_STATEOVERLAY_WORK_DIR "work"
+
+#define OSTREE_STATEOVERLAY_XATTR_DEPLOYMENT_CSUM "user.ostree.deploymentcsum"
+
+/* https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html */
+#define OVERLAYFS_DIR_XATTR_OPAQUE "trusted.overlay.opaque"
+
+static GOptionEntry options[] = { { NULL } };
+
+static gboolean
+ensure_overlay_dirs (const char *overlay_dir, int *out_overlay_dfd, GCancellable *cancellable,
+                     GError **error)
+{
+  glnx_autofd int overlay_dfd = -1;
+  if (!glnx_shutil_mkdir_p_at_open (AT_FDCWD, overlay_dir, 0700, &overlay_dfd, cancellable, error))
+    return FALSE;
+
+  if (!glnx_shutil_mkdir_p_at (overlay_dfd, OSTREE_STATEOVERLAY_WORK_DIR, 0700, cancellable, error))
+    return FALSE;
+  if (!glnx_shutil_mkdir_p_at (overlay_dfd, OSTREE_STATEOVERLAY_UPPER_DIR, 0700, cancellable,
+                               error))
+    return FALSE;
+
+  *out_overlay_dfd = glnx_steal_fd (&overlay_dfd);
+  return TRUE;
+}
+
+/* XXX: upstream to libglnx */
+static gboolean
+lgetxattrat_allow_noent (int dfd, const char *path, const char *attribute, GBytes **out_bytes,
+                         GError **error)
+{
+  g_autoptr (GError) local_error = NULL;
+  *out_bytes = glnx_lgetxattrat (dfd, path, attribute, &local_error);
+  if (!*out_bytes)
+    {
+      if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_INVALID_DATA))
+        return TRUE;
+      g_propagate_error (error, g_steal_pointer (&local_error));
+      return FALSE;
+    }
+  return TRUE;
+}
+
+static gboolean
+is_opaque_dir (int dfd, const char *dname, gboolean *out_is_opaque, GError **error)
+{
+  g_autoptr (GBytes) data = NULL;
+  if (!lgetxattrat_allow_noent (dfd, dname, OVERLAYFS_DIR_XATTR_OPAQUE, &data, error))
+    return FALSE;
+
+  if (!data)
+    *out_is_opaque = FALSE;
+  else
+    {
+      gsize size;
+      const guint8 *buf = g_bytes_get_data (data, &size);
+      *out_is_opaque = (size == 1 && buf[0] == 'y');
+    }
+  return TRUE;
+}
+
+static gboolean
+prune_upperdir_recurse (int lower_dfd, int upper_dfd, GCancellable *cancellable, GError **error)
+{
+  g_auto (GLnxDirFdIterator) dfd_iter = { 0 };
+  if (!glnx_dirfd_iterator_init_at (upper_dfd, ".", FALSE, &dfd_iter, error))
+    return FALSE;
+
+  while (TRUE)
+    {
+      struct dirent *dent = NULL;
+      if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent, cancellable, error))
+        return FALSE;
+      if (dent == NULL)
+        break;
+
+      /* do we have an entry of the same name in the lowerdir? */
+      struct stat stbuf;
+      if (!glnx_fstatat_allow_noent (lower_dfd, dent->d_name, &stbuf, AT_SYMLINK_NOFOLLOW, error))
+        return FALSE;
+      if (errno == ENOENT)
+        continue; /* state file (i.e. upperdir only); carry on */
+
+      /* ok, it shadows; are they both directories? */
+      if (dent->d_type == DT_DIR && S_ISDIR (stbuf.st_mode))
+        {
+          /* is the directory opaque? */
+          gboolean is_opaque = FALSE;
+          if (!is_opaque_dir (upper_dfd, dent->d_name, &is_opaque, error))
+            return FALSE;
+
+          if (!is_opaque)
+            {
+              /* recurse */
+              glnx_autofd int lower_subdfd = -1;
+              if (!glnx_opendirat (lower_dfd, dent->d_name, FALSE, &lower_subdfd, error))
+                return FALSE;
+              glnx_autofd int upper_subdfd = -1;
+              if (!glnx_opendirat (upper_dfd, dent->d_name, FALSE, &upper_subdfd, error))
+                return FALSE;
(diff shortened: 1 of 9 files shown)


cgwalters:
There's a lot of appeal to this for sure.  I mean, it opens up the question a bit if we were to do this, what would it look like to do it for *everything*, having it also take over handling of `/etc` and `/var` for example too.

We had a realtime chat on this, but for the record: The thing that I fear a bit is basically that while I agree it will largely Just Work for things like RPMs/debs that install in `/opt` and want to write extra data there...the flip side is anywhere in these directories you can "leak state".  Of course, state can be leaked in `/etc` and `/var` too today!  But a big part of the idea is that admins only have those two places to save/restore.  With this, there's potentially new important persistent state in these mount points.

However there are some other things:

> Some reconciliation logic is executed whenever the base is updated so that newer files in the base are never shadowed by a copied up version in the upper directory. This matches RPM semantics when upgrading packages whose files may have been modified.

There's a lot of potentially subtle corner cases in this.  For example, what happens when the lower replaces a directory with a symlink? (A classic RPM problem).  It looks like your logic unconditionally deletes away all content in the upper, which could include extra "local" state files too.  That isn't an incorrect behavior, but it's one case which could result in some data loss, depending on the scenario.  To be clear there *isn't* a magic solution here; the crude hammer of `root.transient` just forces the "data loss" to always happen on every OS update, instead of just sometimes.  (Really it forces admins to manually choose what persists by symlinking/mounting those files to `/var`).

Note that the ostree `/etc` semantic handles it differently; the "upper" always wins, meaning you can instead silently not get updates in this scenario.  Now, changing the "type" of an entry in `/etc`  between dir and !dir is IME exceedingly rare, i.e. I don't know of a time it has happened.  But while it is rare, it definitely happens to change between dir and !dir in `/usr`.



---

Now, if I'm understanding things right, state overlays are orthogonal to `root.transient`, which would be great.  I could imagine for example wanting to enable `root.transient` in an OS build, but then also add a state overlay for some specific `/opt/someapp`.

---

Visualization/debuggability: overlayfs is a pretty raw kernel feature; but with e.g. container tooling there's tools to introspect the "layers" (that usually map to overlayfs dirs).  I am sure we'd need to have something like `ostree state-overlay diff` like we have `ostree admin config-diff` too in the future.

cgwalters:
I think my biggest concern in a nutshell is the "unknown unknowns" of this at least to me.  The implementation of `root.transient` is extremely simple, and its semantics very well known and easy to understand and explain due to its prevalence since the creation of docker.

This is a more complex approach, and its semantics required me to think carefully and read the code.  I wonder if anyone else is using this approach for anything?  I can't imagine it's actually novel, but it'd really help me to see something else doing this for some part of a system and how they document it.  

I think https://kubic.opensuse.org/documentation/man-pages/transactional-update.8.html *might* use something like this for `/etc`?

May see if I can reach out on some social media.

cgwalters:
Also of note: It looks like in this implementation that metadata (ownership, mode, xattrs) are not updated when the base image changes but there exists an upper entry.  I would probably say that's a bug.

ericcurtin:
Wouldn't transactional-update give us a dependency on btrfs? I know in Automotive we use, ext4, vfat, erofs, overlays, composefs (even though this is erofs backed). I'd start to get worried if we added even more, every extra filesystem is another possible point of failure.

Do we even support btrfs in RHEL?

ericcurtin:
Or are we thinking to try and do transactional-update with overlays?

jlebon:
> There's a lot of appeal to this for sure. I mean, it opens up the question a bit if we were to do this, what would it look like to do it for _everything_, having it also take over handling of `/etc` and `/var` for example too.

Yeah, definitely could be interesting. Obviously, it's got a much larger impact (and e.g. the fact that overlayfs isn't fully POSIX compliant has a higher chance to rear its head if all OS state is on it).

> We had a realtime chat on this, but for the record: The thing that I fear a bit is basically that while I agree it will largely Just Work for things like RPMs/debs that install in `/opt` and want to write extra data there...the flip side is anywhere in these directories you can "leak state". Of course, state can be leaked in `/etc` and `/var` too today! But a big part of the idea is that admins only have those two places to save/restore. With this, there's potentially new important persistent state in these mount points.

The nice thing though is that the upperdir is on `/var` and because of the beauty of overlayfs, the dir is quite usable from a backup perspective. There are the whiteout character devices which are weird, but a backup service could pretty much ignore those. Backing up the mountpoint is OK too, but would pick up base content too.

> There's a lot of potentially subtle corner cases in this. For example, what happens when the lower replaces a directory with a symlink? (A classic RPM problem). It looks like your logic unconditionally deletes away all content in the upper, which could include extra "local" state files too. That isn't an incorrect behavior, but it's one case which could result in some data loss, depending on the scenario. To be clear there isn't a magic solution here; the crude hammer of root.transient just forces the "data loss" to always happen on every OS update, instead of just sometimes. (Really it forces admins to manually choose what persists by symlinking/mounting those files to /var).
>
> Note that the ostree /etc semantic handles it differently; the "upper" always wins, meaning you can instead silently not get updates in this scenario. Now, changing the "type" of an entry in /etc between dir and !dir is IME exceedingly rare, i.e. I don't know of a time it has happened. But while it is rare, it definitely happens to change between dir and !dir in /usr.

Indeed. I'll note though that the condition here would have to be that it's a file in e.g. `/opt` or `/usr/local` which changes from dir to !dir *and* that it's in the upper dir (i.e. users/apps are expected to write to that dir). Given that RPM itself errors out, I wonder how likely this is. Of course, users could do whatever they want and write to places they shouldn't, so definitely something to be aware of. We could also instead of deleting, rename the directory out of the way and warn.

> Now, if I'm understanding things right, state overlays are orthogonal to `root.transient`, which would be great. I could imagine for example wanting to enable `root.transient` in an OS build, but then also add a state overlay for some specific `/opt/someapp`.

Correct!

> Visualization/debuggability: overlayfs is a pretty raw kernel feature; but with e.g. container tooling there's tools to introspect the "layers" (that usually map to overlayfs dirs). I am sure we'd need to have something like `ostree state-overlay diff` like we have `ostree admin config-diff` too in the future.

Could be interesting down the line. In the vanilla additive case, `tree /usr/ostree/state-overlays/$overlay/upper` gives you that, but definitely whiteouts and opaque dirs could benefit from better visualization.

> I think my biggest concern in a nutshell is the "unknown unknowns" of this at least to me. The implementation of `root.transient` is extremely simple, and its semantics very well known and easy to understand and explain due to its prevalence since the creation of docker.
> 
> This is a more complex approach, and its semantics required me to think carefully and read the code.

I think this is worth teasing apart a bit more. Semantics-wise, I would describe it as "package manager-like", which is of course familiar to a lot of people in this space. If you have to describe it to an end-user, that would suffice. Where it's more complex is in its implementation, which yes, if you hit some corner case you might have to peak under the covers and understand it. We certainly should have docs for that, but the goal is certainly that most people will not need to care.

Agreed re. "unknown unknowns". [I'm optimistic about this](https://github.com/coreos/rpm-ostree/pull/4728#issuecomment-1856719460), but we really need people testing it in realistic scenarios (this is the motivation behind the last commit in https://github.com/coreos/rpm-ostree/pull/4728). Then we can see how well this works and whether we should move forward trying to polish and stabilize it.

> I wonder if anyone else is using this approach for anything? I can't imagine it's actually novel, but it'd really help me to see something else doing this for some part of a system and how they document it.

Yeah, hard to find prior art on this specifically. Definitely also interested to hear from overlayfs SMEs. Maybe @rhvgoyal?

cgwalters:
@ericcurtin I was just referencing that project as potential prior art to look at for an small implementation detail, not proposing depending on it.

src/ostree/ot-admin-builtin-state-overlay.c
```diff
@@ -0,0 +1,243 @@
...
+#define OVERLAYFS_DIR_XATTR_OPAQUE "trusted.overlay.opaque"
+
+static GOptionEntry options[] = { { NULL } };
+
+static gboolean
+ensure_overlay_dirs (const char *overlay_dir, int *out_overlay_dfd, GCancellable *cancellable,
+                     GError **error)
+{
+  glnx_autofd int overlay_dfd = -1;
+  if (!glnx_shutil_mkdir_p_at_open (AT_FDCWD, overlay_dir, 0755, &overlay_dfd, cancellable, error))
```
cgwalters, line 48: This (and below) should be `0700` I think, we don't want to support unprivileged code traversing into it.

src/ostree/ot-admin-builtin-state-overlay.c
```diff
@@ -0,0 +1,243 @@
...
+
+      /* any other case, we prune (this also implicitly covers whiteouts and opaque dirs) */
+      if (dent->d_type == DT_DIR)
+        {
+          if (!glnx_shutil_rm_rf_at (upper_dfd, dent->d_name, cancellable, error))
+            return FALSE;
+        }
+      /* just unlinkat(); saves one openat() call */
+      else if (!glnx_unlinkat (upper_dfd, dent->d_name, 0, error))
+        return FALSE;
```
cgwalters, line 145: This is a super minor style point but I'd find this more readable like:

```
if (is_dir)
  {
  }
else
  {
    if (!glnx_unlinkat (upper_dfd, dent->d_name, 0, error))
      return FALSE;
  }
```

I know some style guides argue the opposite but basically the above makes it crystal clear that it's fully expected to drop into the `else` here, and the `if` is our normal error handling (like a standin for `?` in Rust).

src/ostree/ot-admin-builtin-state-overlay.c
```diff
@@ -0,0 +1,243 @@
...
+
+  /* We don't use "/" directly here because that may have e.g. an overlay
+   * slapped on from root.transient or composefs. */
+  g_autofree char *deployment_path
+      = ostree_sysroot_get_deployment_dirpath (sysroot, booted_deployment);
+  struct stat stbuf_lower;
+  if (!glnx_fstatat (ostree_sysroot_get_fd (sysroot), deployment_path, &stbuf_lower, 0, error))
+    return FALSE;
+
+  if (stbuf_upper.st_mtime < stbuf_lower.st_mtime)
```
cgwalters, line 234: Operating on mtimes seems OK, but I think we have the ability here to operate on versions instead too - e.g. stash an xattr on the upper with the value of `ostree_deployment_get_csum()` and re-sync if they're different.

src/ostree/ot-builtin-admin.c
```diff
@@ -42,6 +42,8 @@ static OstreeCommand admin_subcommands[] = {
     "Change the finalization locking state of the staged deployment" },
   { "boot-complete", OSTREE_BUILTIN_FLAG_NO_REPO | OSTREE_BUILTIN_FLAG_HIDDEN,
     ot_admin_builtin_boot_complete, "Internal command to run at boot after an update was applied" },
+  { "state-overlay", OSTREE_BUILTIN_FLAG_NO_REPO | OSTREE_BUILTIN_FLAG_HIDDEN,
```
cgwalters, line 45: The CLI is hidden, but the systemd unit is not; so this would become insta-stable; is that the intention?  I'm OK with that, just checking.

cgwalters (approved):
I'm OK to merge this to just make it easier to experiment with.  We'll definitely want a man page for this though.

I'm not totally sure if this should be marked as experimental or not.

jlebon:
> I'm OK to merge this to just make it easier to experiment with. We'll definitely want a man page for this though.

Sure, can add one.

> The CLI is hidden, but the systemd unit is not; so this would become insta-stable; is that the intention? I'm OK with that, just checking.
> ...
> I'm not totally sure if this should be marked as experimental or not.

My intent was to not declare it as stabilized yet, because I really feel it needs validation in the real world first. I didn't think about how to gate the systemd unit itself. I guess we could name it e.g. `ostree-experimental-state-overlay@.service` and then rename it once it's stabilized? A bit awkward. We'd also have to carry the old unit name for a while.

In practice, I think people aren't going to know to turn this unit on without looking at the docs, where we can explicitly say this is still experimental. (And even more realistically, I think the primary user of this before stabilizing will be the environment client-side knob demo'ed in https://github.com/coreos/rpm-ostree/issues/233#issuecomment-1856720559, where it's more clear that it's experimental.)

src/ostree/ot-admin-builtin-state-overlay.c
jlebon: Good idea. This handles the case where we're redeploying the same checksum; it would be counterintuitive to prune in those cases.

jlebon:
Updated!

Though this now requires https://gitlab.gnome.org/GNOME/libglnx/-/merge_requests/52.

PR #3125, opened 2023-12-21
switchroot: Be explicit about what could cause /sysroot to be ro

ericcurtin:
If you don't have rw in the kernel cmdline or have ro in it, often you hit this issue. This is just to be really explicit about that in the error messages so people can check.


ericcurtin:
I do wonder should we make src/switchroot/ostree-prepare-root-static.c and src/switchroot/ostree-prepare-root.c one combined src/switchroot/ostree-prepare-root.h file and just rewrite the functions that are different in separate .c files sometimes.

ericcurtin:
On a second look the two files have diverged enough to make the above comment redundant.

cgwalters (approved):

PR #3127, opened 2024-01-02
docs/composefs: Add note about toplevel dirs

cgwalters:
Just thinking about trying to land this, I think some people may hit this one.


jmarrero (approved):
lgtm

jmarrero:
I guess worst case scenario people ask more about this, or are worried the move to composefs might negatively impact them even if they are not adding top level dirs using that method. But I think having it documented outweighs any of that.

PR #3128, opened 2024-01-03
switchroot: use shared constant for unlock --hotfix

cgwalters:
So it's easier to link the two bits of code.


ericcurtin (approved):
Yeah these ifdef's and compile time concatenations are handy

PR #3131, opened 2024-01-04
status: Pass correct remote name when verifying

cgwalters:
The default `ostree admin status` tries to do signature verification, but doesn't error out if that fails.

However, an extremely longstanding bug here is that we passed the "osname" aka stateroot instead of the remote name.  This happens to work for e.g. Fedora Silverblue today, where they are the same.  It doesn't work on FCOS (remote = fedora, stateroot = fedora-coreos).

### DIFF
--- a/src/ostree/ot-admin-builtin-status.c
+++ b/src/ostree/ot-admin-builtin-status.c
@@ -33,35 +33,6 @@ static gboolean opt_verify;
 static GOptionEntry options[] = { { "verify", 'V', 0, G_OPTION_ARG_NONE, &opt_verify,
                                     "Print the commit verification status", NULL },
                                   { NULL } };
-
-#ifndef OSTREE_DISABLE_GPGME
-static gboolean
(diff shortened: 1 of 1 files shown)


ericcurtin (approved):
LGTM, gonna run a test on this branch first, because why not.

ericcurtin:
Tested, also superfast:

```
time LD_LIBRARY_PATH=.libs ./.libs/ostree admin status
* fedora b85477d99c978596689e50e47e3a571816c0a93ea35b994e1c84a98cc3318695.0
    Version: 38.20240104.0
    Unlocked: development
    origin: <unknown origin type>
  fedora 2f0b019864dbc2ff2be71f64f80b64a2499c612b41d6c6f738658495cd76c9e9.0 (rollback)
    Version: 38.20240104.0
    origin: <unknown origin type>

real	0m0.016s
user	0m0.006s
sys	0m0.010s
```

ericcurtin:
Sorry this failed one of the builds because of ifdef differences, gonna fix.

PR #3132, opened 2024-01-05
status: Fix build without GPGME

ericcurtin:
If OSTREE_DISABLE_GPGME is not built in set remote to NULL.

The ostree_repo_signature_verify_commit_data path is irrelevant in the
no gpg case anyway. Having this set as NULL ensures an error gets
thrown early.

### DIFF
--- a/src/ostree/ot-admin-builtin-status.c
+++ b/src/ostree/ot-admin-builtin-status.c
@@ -152,6 +152,8 @@ deployment_print_status (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploym
 
       g_print ("%s", output_buffer->str);
     }
+#else
+  g_autofree char *remote = NULL;
 #endif /* OSTREE_DISABLE_GPGME */
   if (opt_verify)
     {


ericcurtin:
Fixes broken build from:

https://github.com/ostreedev/ostree/pull/3131

src/ostree/ot-admin-builtin-status.c
```diff
@@ -152,6 +152,8 @@ deployment_print_status (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploym
 
       g_print ("%s", output_buffer->str);
     }
+#else
+  g_autofree char *remote = ostree_deployment_get_osname (deployment);
```
cgwalters, line 156: No, the osname/stateroot is never right for this.  Let's just do instead:
`const char *remote = NULL;`
instead.

cgwalters:
BTW procedural point; we try to use "topic prefixes" in this repo, so the commit message could look more like (but also rewording with the new suggestion):

> status: Fix build without GPGME
>
> Regression from previous change.

src/ostree/ot-admin-builtin-status.c
ericcurtin: Sounds good, this whole ostree_repo_signature_verify_commit_data path seems irrelevant with gpg anyway, setting this as null nips it in the bud alright.

ericcurtin:
Repushed with changes.

cgwalters (approved):

PR #3133, opened 2024-01-05
systemd/ostree-boot-complete: Start earlier

cgwalters:
systemd/ostree-boot-complete: Add `ConditionKernelCommandLine=ostree`

In practice we don't enable this unit except via our generator,
but let's do this on general principle.

---

systemd/ostree-boot-complete: Start earlier

Prep for changing this service to perform state computations
such as "is this boot the default, or did we get rolled back"
that can be used by higher level tools.

---


jmarrero (approved):
lgtm

PR #3134, opened 2024-01-05
status: Introduce tool to quickly check if we are booted as default

ericcurtin:
Generally in ostree based systems you would expect to boot into deployment 0, in rollback conditions triggered by greenboot-related rollbacks this might not be the case. This is a tool to detect this.

### DIFF
--- a/man/ostree-admin-status.xml
+++ b/man/ostree-admin-status.xml
@@ -62,6 +62,65 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
     </refsect1>
 
     <refsect1>
+        <title>Options</title>
+
+        <variablelist>
+            <varlistentry>
+                <term><option>--sysroot</option>="PATH"</term>
+
+                <listitem><para>
+                    Create a new OSTree sysroot at PATH
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>-V, --verify</option></term>
+
+                <listitem><para>
+                    Print the commit verification status
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>-S, --skip-signatures</option></term>
+
+                <listitem><para>
+                    Skip signatures in output
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>-Q, --query-booted</option></term>
+
+                <listitem><para>
+                    Output the string <literal>default</literal> if the default deployment
+                    is the booted one, <literal>not-default</literal> if we are booted in
+                    a non-default deployment (e.g. the user interactively chose a
+                    different entry in the bootloader menu, or the bootloader rolled back
+                    automatically, etc.). If we are not in a booted OSTree system, an
+                    error is returned.
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>-v, --verbose</option></term>
+
+                <listitem><para>
+                    Print debug information during command processing
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>--version</option>--version</term>
+
+                <listitem><para>
+                    Print version information and exit
+                </para></listitem>
+            </varlistentry>
+        </variablelist>
+    </refsect1>
+
+    <refsect1>
         <title>Example</title>
         <para><command>$ ostree admin status</command></para>
 <programlisting>
--- a/src/ostree/ot-admin-builtin-status.c
+++ b/src/ostree/ot-admin-builtin-status.c
@@ -30,13 +30,21 @@
 
 static gboolean opt_verify;
 static gboolean opt_skip_signatures;
+static gboolean opt_query_booted;
 
-static GOptionEntry options[]
-    = { { "verify", 'V', 0, G_OPTION_ARG_NONE, &opt_verify, "Print the commit verification status",
-          NULL },
-        { "skip-signatures", 'S', 0, G_OPTION_ARG_NONE, &opt_skip_signatures,
-          "Print the commit verification status", NULL },
-        { NULL } };
+static GOptionEntry options[] = {
+  { "verify", 'V', 0, G_OPTION_ARG_NONE, &opt_verify, "Print the commit verification status",
+    NULL },
+  { "skip-signatures", 'S', 0, G_OPTION_ARG_NONE, &opt_skip_signatures,
+    "Print the commit verification status", NULL },
+  { "query-booted", 'Q', 0, G_OPTION_ARG_NONE, &opt_query_booted,
+    "Output the string \"default\" if the default deployment is the booted one, \"not-default\" if "
+    "we are booted in a non-default deployment (e.g. the user interactively chose a different "
+    "entry in  the bootloader menu, or the bootloader rolled back automatically, etc.). If we are "
+    "not in a booted OSTree system, an error is returned.",
+    NULL },
+  { NULL }
+};
 static gboolean
 deployment_print_status (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeployment *deployment,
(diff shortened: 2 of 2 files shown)


src/ostree/ot-admin-builtin-status.c
```diff
@@ -204,6 +207,11 @@ ot_admin_builtin_status (int argc, char **argv, OstreeCommandInvocation *invocat
     {
       g_print ("No deployments.\n");
     }
+  else if (opt_query_booted)
+    {
+      const gboolean is_default_booted = deployments->pdata[0] == booted_deployment;
+      g_print ("%s\n", is_default_booted ? "default" : "not-default");
```
ericcurtin, line 213: I had a:

```
      if (!is_default_booted)
        return FALSE;
```

here also initially, but this causes it to coredump and everything:

```
ostree admin status -Q
not-default
**
OSTree:ERROR:src/ostree/ot-main.c:323:ostree_run: assertion failed: (success || error)
Bail out! OSTree:ERROR:src/ostree/ot-main.c:323:ostree_run: assertion failed: (success || error)
Aborted (core dumped)
```

would have been nice because you could check the exit code of the binary, than a string comparison for "default" in systemd.

cgwalters: The classic problem with using exit codes like that then is there's no way to distinguish "failed to compute state because we encountered e.g. an I/O error or the process segfaulted" from "not default".

src/ostree/ot-admin-builtin-status.c
```diff
@@ -204,6 +207,11 @@ ot_admin_builtin_status (int argc, char **argv, OstreeCommandInvocation *invocat
     {
       g_print ("No deployments.\n");
     }
+  else if (opt_query_booted)
```
cgwalters, line 210: I think this option should come at the very first thing; otherwise in the degenerate case of "failed to find any deployments" the caller would see an unexpected "No deployments" string.  Because we expect callers to parse this we should be strict about what we output.

Also...I would say it becomes an *error* when using this option if we don't find any deployments, right?

src/ostree/ot-admin-builtin-status.c
```diff
@@ -30,12 +30,15 @@
...
 static gboolean opt_skip_signatures;
+static gboolean opt_query_booted;
 
 static GOptionEntry options[]
     = { { "verify", 'V', 0, G_OPTION_ARG_NONE, &opt_verify, "Print the commit verification status",
           NULL },
         { "skip-signatures", 'S', 0, G_OPTION_ARG_NONE, &opt_skip_signatures,
           "Print the commit verification status", NULL },
+        { "query-booted", 'Q', 0, G_OPTION_ARG_NONE, &opt_query_booted,
+          "Print whether the default deployment is booted or not", NULL },
```
cgwalters, line 41: Remember to update the man page too.

ericcurtin: Yeah the man page for this has zero command line options documented, I guess we can change that man/ostree-admin-status.xml

src/ostree/ot-admin-builtin-status.c
ericcurtin: Done

src/ostree/ot-admin-builtin-status.c
cgwalters: Ah yes, we are not great at it.  Thank you for doing this!

man/ostree-admin-status.xml
```diff
@@ -61,6 +61,60 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
...
+                <listitem><para>
+                    Skip signatures in output
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>-Q, --query-booted</option></term>
+
+                <listitem><para>
+                    Print whether the default deployment is booted or not
```
cgwalters, line 96:
```suggestion
                    Output the string <literal>default</literal> if the default deployment is the booted one, <literal>not-default</literal> if we are booted in a non-default deployment (e.g. the user interactively chose a different entry in  the bootloader menu, or the bootloader rolled back automatically, etc.).  If we are not in a booted OSTree system, an error is returned.

File: src/ostree/ot-admin-builtin-status.c [hunk 5]
```diff
@@ -200,10 +203,16 @@ ot_admin_builtin_status (int argc, char **argv, OstreeCommandInvocation *invocat
   if (booted_deployment)
     ostree_sysroot_query_deployments_for (sysroot, NULL, &pending_deployment, &rollback_deployment);
 
-  if (deployments->len == 0)
+  if (opt_query_booted)
     {
-      g_print ("No deployments.\n");
+      if (deployments->len == 0)
+        return FALSE;
```
cgwalters, line 209: This must be
```suggestion
        return glnx_throw (error, "Not in a booted OSTree system");
```
or so.  We can't just `return FALSE` without setting an error.

src/ostree/ot-admin-builtin-status.c
```diff
@@ -200,10 +203,16 @@ ot_admin_builtin_status (int argc, char **argv, OstreeCommandInvocation *invocat
...
+
+      const gboolean is_default_booted = deployments->pdata[0] == booted_deployment;
+      g_print ("%s\n", is_default_booted ? "default" : "not-default");
     }
+  else if (deployments->len == 0)
+    g_print ("No deployments.\n");
```
cgwalters, line 215: Our style usually has all components of an `if` either have braces or not.

cgwalters (changes requested):
Overall LGTM, one bug and some nits.  Thanks for this!

ericcurtin:
This one is ready for re-review @cgwalters I believe everything is addressed

man/ostree-admin-status.xml
```diff
@@ -61,6 +61,65 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
...
+            <varlistentry>
+                <term><option>-v, --verbose</option></term>
+
+                <listitem><para>
+                    Print debug information during command processing
+                </para></listitem>
+            </varlistentry>
+
+            <varlistentry>
+                <term><option>--version</option>--version</term>
```
jlebon, line 114: Is the syntax correct here?

src/ostree/ot-admin-builtin-status.c
```diff
@@ -30,13 +30,21 @@
...
-          "Print the commit verification status", NULL },
-        { NULL } };
+static gboolean opt_query_booted;
+
+static GOptionEntry options[] = {
+  { "verify", 'V', 0, G_OPTION_ARG_NONE, &opt_verify, "Print the commit verification status",
+    NULL },
+  { "skip-signatures", 'S', 0, G_OPTION_ARG_NONE, &opt_skip_signatures, "Skip signatures in output",
+    NULL },
+  { "query-booted", 'Q', 0, G_OPTION_ARG_NONE, &opt_query_booted,
```
jlebon, line 40: This is pure bikeshed, but I think this would be better named e.g. `--is-default`. Somewhat precedence are the `systemctl is-*` commands that also print strings. `--query-booted` is a bit ambiguous.

src/ostree/ot-admin-builtin-status.c
```diff
@@ -30,13 +30,21 @@
...
+    "Output the string \"default\" if the default deployment is the booted one, \"not-default\" if "
+    "we are booted in a non-default deployment (e.g. the user interactively chose a different "
+    "entry in  the bootloader menu, or the bootloader rolled back automatically, etc.). If we are "
+    "not in a booted OSTree system, an error is returned.",
```
jlebon, line 44: Minor: this is duplicating the manpage. These docstrings are usually more brief. Maybe e.g.

> Output \"default\" if booted into the default deployment, otherwise \"not-default\"

?

PR #3135, opened 2024-01-05
doc: Add section about ostree and aboot

ericcurtin:
Android Bootloader is a standard of how Android devices should implement their bootloaders, we also use it in CentOS Automotive Stream Distribution for some ARM boards. Here is some documentation on how ostree works with this.


ericcurtin:
This can also be reviewed/collaborated on in https://hackmd.io/@7wAdmxHWRI6dhoDAwOtrpw/ByAeouBua if preferred.

cgwalters (approved):
This is opening up more questions and concerns I have but let's start with this!

PR #3136, opened 2024-01-09
status: Rename query-booted to is-default

ericcurtin:
This is a tool to check if we are booted as default or not, just a rename before it becomes widely used. We also shortened the '-h' output for this.


ericcurtin:
Addressing review feedback from https://github.com/ostreedev/ostree/pull/3134

jlebon (approved):
Thanks!

PR #3143, opened 2024-01-22
deploy: Ignore sockets, fifos in /etc/<subdir> during merge

yummypeng:
It's a followup of commit e6a560b40797324aa8b90e7100c6d50bff91f14d. We should also ignore sockets and fifos in the subdir of /etc.


cgwalters (approved):
Thanks, this looks good.  Would be nice to have a test but we can do that after, and I don't expect you to learn our test suite for things like this.

Thanks again!

PR #3144, opened 2024-01-22
Add commands to get index of booted, pending and previous deployments

p5:
Closes #3142 

Right now, it's difficult to pin the current deployment without needing to look into the output of some other tooling (like rpm-ostree) to get the index of each deployment.  This index also is not consistent - the current deployment could be 0 when you first boot the system then 1 shortly after. 

This change makes it easy to get the index of the current or future deployment so it can be piped into other commands (such as ostree admin pin).

(Please note: I have not yet built or tested this PR - I am working on figuring out how to compile it)

Huge thanks to Eric for authoring this patch for me!

TODO before merge:
- [ ] Squash all commits into one
- [ ] Get confirmation for flag descriptions & man pages
- [x] Confirm the new functionality works locally

Co-authored-by: Eric Curtin <ecurtin@redhat.com>

PR #3146, opened 2024-01-24
admin/pin: Add commands to pin booted, pending and rollbacks deployments

ericcurtin:
Add new commands to pin the current, staged and previous deployment for
use in automation and scripting. Right now, it's difficult to pin the
current deployment without needing to look into the output of some other
tooling (like rpm-ostree) to get the index of each deployment. This
index also is not consistent - the current deployment could be 0 when
you first boot the system then 1 shortly after. This change makes it
easy to pin the current or future deployment.

Co-authored-by: Robert Sturla <robertsturla@outlook.com>
Signed-off-by: Eric Curtin <ecurtin@redhat.com>

### DIFF
--- a/src/ostree/ot-admin-builtin-pin.c
+++ b/src/ostree/ot-admin-builtin-pin.c
@@ -31,6 +31,53 @@ static gboolean opt_unpin;
 static GOptionEntry options[]
     = { { "unpin", 'u', 0, G_OPTION_ARG_NONE, &opt_unpin, "Unset pin", NULL }, { NULL } };
 
+static gint64
+get_deployment_index_for_type (OstreeSysroot *sysroot, const char *deploy_index_str)
+{
+  g_autoptr (GPtrArray) deployments = ostree_sysroot_get_deployments (sysroot);
+  OstreeDeployment *booted_deployment = ostree_sysroot_get_booted_deployment (sysroot);
+  if (booted_deployment && deploy_index_str[0] == 'b')
+    return ostree_deployment_get_index (booted_deployment);
+
+  g_autoptr (OstreeDeployment) pending_deployment = NULL;
+  g_autoptr (OstreeDeployment) rollback_deployment = NULL;
+  if (booted_deployment)
+    ostree_sysroot_query_deployments_for (sysroot, NULL, &pending_deployment, &rollback_deployment);
+
+  if (pending_deployment && deploy_index_str[0] == 'p')
+    return ostree_deployment_get_index (pending_deployment);
+  else if (rollback_deployment && deploy_index_str[0] == 'r')
+    return ostree_deployment_get_index (rollback_deployment);
+
+  return -1;
+}
+
+static gboolean
+do_pinning (OstreeSysroot *sysroot, const gint64 deploy_index, GError **error)
+{
+  g_autoptr (OstreeDeployment) target_deployment
+      = ot_admin_get_indexed_deployment (sysroot, deploy_index, error);
+  if (!target_deployment)
+    return FALSE;
+
+  const gboolean current_pin = ostree_deployment_is_pinned (target_deployment);
+  const gboolean desired_pin = !opt_unpin;
+  if (current_pin == desired_pin)
+    {
+      g_print ("Deployment %lld is already %s\n", (long long int)deploy_index,
+               current_pin ? "pinned" : "unpinned");
+      return TRUE;
+    }
+  else if (ostree_sysroot_deployment_set_pinned (sysroot, target_deployment, desired_pin, error))
+    {
+      g_print ("Deployment %lld is now %s\n", (long long int)deploy_index,
+               desired_pin ? "pinned" : "unpinned");
+      return TRUE;
+    }
+
+  return FALSE;
+}
+
 gboolean
 ot_admin_builtin_pin (int argc, char **argv, OstreeCommandInvocation *invocation,
                       GCancellable *cancellable, GError **error)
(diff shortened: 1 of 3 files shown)


ericcurtin:
Linked to https://github.com/ostreedev/ostree/pull/3144 , must complete man pages.

cgwalters (commented):
I'm OK with this, but how about extending our syntax to accept an integer *or* `booted|pending|rollback`?  IOW it'd just be

```
ostree admin pin rollback
```

ericcurtin:
@cgwalters ok, I'll just change it to a string comparison, it will be 95% the same code

ericcurtin:
This one is ready for review

src/ostree/ot-admin-builtin-pin.c
```diff
@@ -28,9 +28,74 @@
...
+} OstreeDeploymentType;
+
 static GOptionEntry options[]
     = { { "unpin", 'u', 0, G_OPTION_ARG_NONE, &opt_unpin, "Unset pin", NULL }, { NULL } };
 
+static gint64
+get_deployment_index (const OstreeDeployment *target_deployment, const GPtrArray *deployments)
+{
+  for (gint i = 0; i < deployments->len; ++i)
+    if (deployments->pdata[i] == target_deployment)
```
cgwalters, line 46: BTW our style generally prefers braces `{` in these types of cases, really braceless is only for a simple `if`, but this is irrelevant per above

src/ostree/ot-admin-builtin-pin.c
```diff
@@ -54,32 +119,23 @@ ot_admin_builtin_pin (int argc, char **argv, OstreeCommandInvocation *invocation
       char *endptr = NULL;
 
       errno = 0;
-      const guint64 deploy_index = g_ascii_strtoull (deploy_index_str, &endptr, 10);
+      gint64 deploy_index = g_ascii_strtoull (deploy_index_str, &endptr, 10);
       if (*endptr != '\0')
-        return glnx_throw (error, "Invalid index: %s", deploy_index_str);
-      if (errno == ERANGE)
+        {
+          if (!g_strcmp0 (deploy_index_str, "booted"))
```
cgwalters, line 125: Hmm I'd say we should check the string values first, and then only fall back to trying to parse as an integer.

src/ostree/ot-admin-builtin-pin.c
```diff
@@ -28,9 +28,74 @@
...
+  OSTREE_DEPLOYMENT_BOOTED = 1,
+  OSTREE_DEPLOYMENT_PENDING = 2,
+  OSTREE_DEPLOYMENT_ROLLBACK = 3
+} OstreeDeploymentType;
+
 static GOptionEntry options[]
     = { { "unpin", 'u', 0, G_OPTION_ARG_NONE, &opt_unpin, "Unset pin", NULL }, { NULL } };
 
+static gint64
+get_deployment_index (const OstreeDeployment *target_deployment, const GPtrArray *deployments)
```
cgwalters, line 43: This is just `ostree_deployment_get_index`

src/ostree/ot-admin-builtin-pin.c
```diff
@@ -54,32 +119,23 @@ ot_admin_builtin_pin (int argc, char **argv, OstreeCommandInvocation *invocation
...
-        return glnx_throw (error, "Invalid index: %s", deploy_index_str);
-      if (errno == ERANGE)
+        {
+          if (!g_strcmp0 (deploy_index_str, "booted"))
+            deploy_index = get_deployment_index_for_type (sysroot, OSTREE_DEPLOYMENT_BOOTED);
+          else if (!g_strcmp0 (deploy_index_str, "pending"))
+            deploy_index = get_deployment_index_for_type (sysroot, OSTREE_DEPLOYMENT_PENDING);
+          else if (!g_strcmp0 (deploy_index_str, "rollback"))
+            deploy_index = get_deployment_index_for_type (sysroot, OSTREE_DEPLOYMENT_ROLLBACK);
+
```
cgwalters, line 131: Feels like we should do something to explicitly handle the notfound aka `-1` case here?  Like:

```
if (deploy_index < 0)
  return glnx_throw (error, "Deployment type not found: %s", deploy_index_str);
```

ericcurtin: It's already handled inside ot_admin_get_indexed_deployment function, so I didn't worry about handling it again, but we can handle it here also if needs be.

src/ostree/ot-admin-builtin-pin.c
ericcurtin: Sure...

src/ostree/ot-admin-builtin-pin.c
ericcurtin: Add both to for and if I assume?

src/ostree/ot-admin-builtin-pin.c
ericcurtin: SGTM

ericcurtin: I should call this `ostree_match_deployment`, thanks for pointing this out, there's another very similarly named function...

src/ostree/ot-admin-builtin-pin.c
cgwalters: That will tell the user `error: Invalid index -1` which seems much less friendly than `error: Deployment type not found: rollback` right?

src/ostree/ot-admin-builtin-pin.c
cgwalters: The sort of vague unspecified rule I follow is that if you have any conditional/looping instructions that nest, then use braces at least on the outer one. 

So
```
for (...) {
  if (something)
    return i;
}
```
is OK too.  The rationale behind this is the same as that for [-Wmisleading-indentation](https://developers.redhat.com/blog/2016/02/26/gcc-6-wmisleading-indentation-vs-goto-fail?source=sso).

src/ostree/ot-admin-builtin-pin.c
ericcurtin: I see what you mean now, will use that function

ericcurtin:
Looks like an unrelated build failure maybe :shrug:

cgwalters (approved):

cgwalters:
For some reason `iso-live-login.uefi-secure` is flaky.

PR #3147, opened 2024-01-26
generator: Exit if there's no `/run/ostree`

cgwalters:
Currently if run in a container image under systemd, we will incorrectly synthesize a `var.mount` unit even if `ostree-prepare-root` hasn't run.

The comment here said why we didn't do that before, but that's for the really legacy embedded-only "ostree-prepare-root-static" path, and even then I'm pretty sure it was wrong because the generator here only runs in the *real* root, and we should have `/run/ostree` at that point.


ericcurtin (approved):

jlebon:
Follow-up on this in https://github.com/ostreedev/ostree/pull/3158.

PR #3149, opened 2024-01-31
composefs: Bump composefs max version to 1

alexlarsson:
This generates the new format for whiteout markers which was added in  6.8 (and which will be backported to 6.7). Without this whiteouts  will not work anymore.
    
This is a slight format change, but will only affect ostree commits that already were broken (i.e that had whiteouts), and since the composefs code is still marked experimental I think it is fine to do this without introducing another format version on the ostree side.


cgwalters (approved):

alexlarsson:
@cgwalters Is this supposed to be auto-merged with the override? It doesn't seem to be.

Issue #2804, opened 2023-01-13
Add support for forcing on selinux labels


cgwalters:
Working on https://github.com/containers/bootc/issues/1 I was experimenting with testing a FCOS install from the default GHA Ubuntu VM where SELinux is disabled.  ostree today doesn't work for this, but it should.

Basically, we shouldn't care about the host state, only the target.  If the target system has selinux labels, we should write them.  However, we have to battle libselinux which looks at `/sys/fs/selinux` etc.  Our own sepolicy class also dispatches on that.

Now, when we're operating on a host system that *does* have selinux enabled, I think it's better to use `setfscreatecon()` where we can to ensure atomicity of the security labels.  But if not...I think the patches will look something like this:

```
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -210,7 +210,7 @@ create_file_copy_from_input_at (OstreeRepo     *repo,
   g_autoptr(GVariant) modified_xattrs = NULL;
 
   /* If we're doing SELinux labeling, prepare it */
-  if (sepolicy_enabled)
+  if (sepolicy_enabled && options->sepolicy->host_enabled())
     {
       /* If doing sepolicy path-based labeling, we don't want to set the
        * security.selinux attr via the generic xattr paths in either the symlink
```

Basically, if the host doesn't have selinux enabled, we just write the `security.selinux` xattr the same as all other xattrs (which works fine with selinux disabled on the host).

cgwalters:
Some more on this in e.g. https://github.com/containers/bootc/pull/397 - we have a warning test there and it reveals tons of missing things for this:

- repo object directories `ostree/repo/objects/a1`
- All repository metadata objects `.dirtree`, `.dirmeta`, `.commit` etc.
- Everything in `ostree/repo/refs`
- Everything under `ostree/deploy`
- Very ironically, the `ostree/deploy/default/var/.ostree-selabeled` file

PR #3151, opened 2024-01-31
libostree: write selinux xattr when on non-selinux systems

mvo5:
Currently when writing data for selinux systems on a non-selinux system there will be no labels. This is because
`ostree_sepolicy_setfscreatecon()` just returns TRUE on non-selinux systems and xattr writing for `security.seliux` is filtered out.

This patches uses the suggestion of Colin Walters (thanks!) from https://github.com/ostreedev/ostree/issues/2804 and detects if the host has selinux enabled and if not just skips filtering the xattrs for selinux.

[draft as this will need a test (and validation of the basic approach by someone experienced) and I will need a hint how to best test this]

### DIFF
--- a/.github/workflows/bootc.yaml
+++ b/.github/workflows/bootc.yaml
@@ -35,7 +35,9 @@ jobs:
       - name: bootc install
         run: |
           set -xeuo pipefail
-          sudo podman run --rm -ti --privileged -v /:/target --pid=host --security-opt label=disable \
+          sudo podman run --env BOOTC_SKIP_SELINUX_HOST_CHECK=1 --rm -ti --privileged -v /:/target --pid=host --security-opt label=disable \
             -v /var/lib/containers:/var/lib/containers \
             localhost/test:latest bootc install to-filesystem --skip-fetch-check \
-            --disable-selinux --replace=alongside /target
+             --replace=alongside /target
+          # Verify labeling for /etc
+          sudo ls -dZ /ostree/deploy/default/deploy/*.0/etc |grep :etc_t:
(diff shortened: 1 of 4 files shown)


cgwalters:
> [draft as this will need a test (and validation of the basic approach by someone experienced) and I will need a hint how to best test this]

I can help with a test case

cgwalters:
https://github.com/ostreedev/ostree/pull/3152 will help verify this, though to make it work we'll want to add a "no really try to use selinux even on non-selinux" host arg to `bootc install`.

Alternatively we can just do a direct ubuntu deploy here too, I can look at that.

mvo5:
> #3152 will help verify this, though to make it work we'll want to add a "no really try to use selinux even on non-selinux" host arg to `bootc install`.
> 
> Alternatively we can just do a direct ubuntu deploy here too, I can look at that.

Thank you!, I really appreciate the quick reply here I have no opinion about the best approach here and will happily follow your lead. I will just need a few hints how to best approach this. Alternatively feel free to push directly to this PR or I can mark it ready and we test indirectly via the approaches outlined. I am happy either way :)

cgwalters:
OK https://github.com/containers/bootc/pull/293 helps us get farther here for testing.

And yes...trying this out I notice that yep, we are not handling the labeling for all of the "basic infrastructure dirs" here, like `/ostree/deploy` etc.  I will look at that in the background, it's going to be a bit of a can of worms though.

cgwalters:
I did https://github.com/containers/bootc/pull/294 push this farther, but there's definitely still that can of worms.

cgwalters:
I've updated this with a test, and this verifies that it improves things as we now correctly have labels for `etc` in the deployment root where we didn't before.

cgwalters (approved):

mvo5:
> I've updated this with a test, and this verifies that it improves things as we now correctly have labels for `etc` in the deployment root where we didn't before.

Thank you! Very nice how straightforward it was to add a small test for this.

ericcurtin (approved):

cgwalters:
@mvo5 just one more bug to fix here, we need to handle the build without selinux enabled.

PR #3152, opened 2024-01-31
ci: Add a bootc/c9s workflow

cgwalters:
This obsoletes the "just build" on c9s flow, and actually runs though a bootc install, which exercises more of things.

### DIFF
--- a/.github/workflows/bootc.yaml
+++ b/.github/workflows/bootc.yaml
@@ -1,0 +1,41 @@
+name: ci-bootc
+
+permissions:
+  actions: read
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+  workflow_dispatch: {}
+
+concurrency:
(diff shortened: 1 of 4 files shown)


ericcurtin (approved):

ericcurtin:
I am curious how:

```
bootc install to-filesystem
```

Identifies the partitions to write to, or does it expect esp, rootfs and boot partitions to already be mounted in a certain way.

Hoping to use it for the Fedora Asahi immutable variants post Fedora 40 release and when my PC refresh comes around in the second half of 2024 so I have a second Apple Silicon machine to iterate on without dumping my main ARM environment.

cgwalters:
> or does it expect esp, rootfs and boot partitions to already be mounted in a certain way.

This, it expects `/boot` and `/boot/efi` relative to the root.

ericcurtin:
Does

```
bootc install to-filesystem
```

work on a pre-existing Fedora Workstation and/or Silverblue install?

PR #3157, opened 2024-02-05
syslinux: Avoid double `/boot` if bootprefix is enabled

cgwalters:
This backend always explicitly emitted a `/boot` - but if the global `sysroot.bootprefix` is enabled, then we can rely on the outer code doing it.

Luckily this was caught by the unit tests here failing when enabling `sysroot.bootprefix` by default.


jlebon (approved):

ericcurtin (approved):

cgwalters:
`iso-live-login.uefi` timeout again, unrelated

PR #3156, opened 2024-02-05
Enable `sysroot.bootprefix` by default

cgwalters:
I've been testing this in various places and not seen any fallout, so let's finally enable this by default and have the situation where `/boot` is on the root `/` filesystem work out of the box.

PR #3159, opened 2024-02-07
Revert "Enable `sysroot.bootprefix` by default"

cgwalters:
This reverts commit 8627c8afa15fa0b2dc2dc261a217dd043a991a7d.

See discussion in https://github.com/ostreedev/ostree/pull/3156 ; we think this breaks s390x in some cases at least, and that warrants further investigation.


ericcurtin (approved):

ericcurtin:
Clicked auto-merge to expedite this, since it's a regression

dustymabe (approved):

PR #3162, opened 2024-02-08
ostree-grub-generator: allow adding custom scripts to grub.cfg

EdTheBearded:
Hello!
I would like to propose this change here, since this may help someone else.

First the motivation for this change: rollback
We wanted to be able to perform a rollback if our update was unsuccessful in some way. And for that we needed to add some logic to `grub.cfg` script to get it done.

With that in mind, I tried to do this without changing too much or just manually editing `grub.cfg`. I took some inspiration on what `grub-mkconfig` does: check `/etc/grub.d` for custom scripts, and add those to the final `grub.cfg` script, parsing the files in their alphabetical order.

So with this change, whatever files found under `/etc/ostree.d` (just went with whatever name at the time) are added (in alphabetical order) between the cfg header and its menuentries, allowing to add custom logic to the bootloader in a non intrusive way.

NOTE: we're using this on Yocto, so there's also a minor change to be done in ostree recipe, to create the `/etc` folder being used.

### DIFF
--- a/src/boot/grub2/ostree-grub-generator
+++ b/src/boot/grub2/ostree-grub-generator
@@ -25,6 +25,7 @@ script=$(basename ${0})
 # Atomically safe location where to generete grub.cfg when executing system upgrade.
 new_grub2_cfg=${2}
 entries_path=$(dirname $new_grub2_cfg)/entries
+custom_scripts="/etc/ostree.d"
 
 read_config()
 {
@@ -105,10 +106,22 @@ timeout=10
 EOF
 }
 
+populate_custom_section()
+{
+if [ -d "$custom_scripts" ]; then
+    for script in $(ls -v $custom_scripts/*); do
+        echo -e "\n### BEGIN ${script} ###" >> ${new_grub2_cfg} 
+        cat ${script} >> ${new_grub2_cfg}
+        echo -e "\n### END ${script} ###\n" >> ${new_grub2_cfg}
+    done
+fi
+}
+
 generate_grub2_cfg()
 {
     populate_warning
     populate_header
+    populate_custom_section
     populate_menu
 }
 


src/boot/grub2/ostree-grub-generator
```diff
@@ -25,6 +25,7 @@ script=$(basename ${0})
 # Atomically safe location where to generete grub.cfg when executing system upgrade.
 new_grub2_cfg=${2}
 entries_path=$(dirname $new_grub2_cfg)/entries
+custom_scripts="/etc/ostree.d"
```
cgwalters, line 28: This seems like it should be e.g. `/etc/ostree-grub.d` or so.

cgwalters:
I'm not opposed exactly, but I don't quite understand.  grub has a fully generic system for injecting code, why can't you drop your extra stuff into the existing `/etc/grub.d`?

On the topic of configuring rollbacks and grub configs, you may also want to look into https://github.com/coreos/bootupd/ which is intended to pair with ostree - bootupd supports an opinionated "static grub config" model where we don't regenerate any grub configs when updating the kernel, assuming GRUB has been patched to read from `/boot/loader/entries` directly.

Also, there is https://github.com/fedora-iot/greenboot/ (currently relies on rpm-ostree, not ostree, but that is conceptually a bug) which also now injects some grub logic via bootupd, see https://github.com/fedora-iot/greenboot/pull/129

champtar:
@EdTheBearded can you address the review and rebase ?

src/boot/grub2/ostree-grub-generator
EdTheBearded: I've changed the name of the folder for `ostree-grub.d`.

PR #316, opened 2016-05-31
[merged] Glnx round x

cgwalters:

PR #3163, opened 2024-02-08
Fix booted deploy detection with transient /etc

alexlarsson:
When /etc is transient, and composefs is root, there is currently no bind-mount of the deploy directory so the checks for finding if a deploy is booted are failing.

We fix this by bind-mounting the deploy dir as /run/ostree/deploy and use that to look for the booted deploy directory.

This fixes: https://github.com/ostreedev/ostree/issues/316

### DIFF
--- a/src/libostree/ostree-sysroot-private.h
+++ b/src/libostree/ostree-sysroot-private.h
@@ -74,8 +74,8 @@ struct OstreeSysroot
   /* The device/inode for / and /etc, used to detect booted deployment */
   dev_t root_device;
   ino_t root_inode;
-  dev_t etc_device;
-  ino_t etc_inode;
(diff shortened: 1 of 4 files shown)


cgwalters:
I'm OK with this probably as is, but WDYT about just injecting the device/inode pair into the `/run/ostree-booted` metadata file?  That file being non-empty is a relatively new thing but I think it just makes sense to read it instead of trying to re-infer the state of what was set up in ostree-prepare-root (or adding another bind mount just so we can stat it).

alexlarsson:
That would work to, lemme try it out and see how it ends up-

alexlarsson:
Given the weird error in the CI here i prefer the #3164 approach

PR #3164, opened 2024-02-08
Track deployment root/inode from prepare root

cgwalters:
When we added composefs, it broke the logic for detecting the booted
deployment which was previously a direct (device, inode) comparison.
So the code there started looking at `etc`.  However, that in
turns breaks with `etc.transient = true` enabled.

Fix all of this by tracking the real deployment directory's
(device,inode) that we found in `ostree-prepare-root`, and inject
it into the extensible metadata we have in `/run/ostree-booted`
which is designed exactly to pass state between the initramfs
and the real root.

---

### DIFF
--- a/src/libostree/ostree-sysroot-private.h
+++ b/src/libostree/ostree-sysroot-private.h
@@ -74,8 +74,9 @@ struct OstreeSysroot
   /* The device/inode for / and /etc, used to detect booted deployment */
   dev_t root_device;
   ino_t root_inode;
-  dev_t etc_device;
-  ino_t etc_inode;
+
+  // The parsed data from /run/ostree
+  GVariantDict *run_ostree_metadata;
 
   gboolean is_physical; /* TRUE if we're pointed at physical storage root and not a deployment */
   GPtrArray *deployments;
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -83,6 +83,7 @@ ostree_sysroot_finalize (GObject *object)
 
   g_clear_object (&self->path);
   g_clear_object (&self->repo);
+  g_clear_pointer (&self->run_ostree_metadata, g_variant_dict_unref);
   g_clear_pointer (&self->deployments, g_ptr_array_unref);
   g_clear_object (&self->booted_deployment);
   g_clear_object (&self->staged_deployment);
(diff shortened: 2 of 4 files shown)


src/switchroot/ostree-prepare-root.c
```diff
@@ -430,6 +430,11 @@ main (int argc, char *argv[])
   GVariantBuilder metadata_builder;
   g_variant_builder_init (&metadata_builder, G_VARIANT_TYPE ("a{sv}"));
 
+  if (lstat (".", &stbuf) < 0)
+    err (EXIT_FAILURE, "lstat(TMP_SYSROOT)");
```
alexlarsson, line 434: The error message seems wrong, it's the deploy path, not the tmp sysroot

src/libostree/ostree-sysroot.c
```diff
@@ -812,6 +813,17 @@ parse_deployment (OstreeSysroot *self, const char *boot_link, OstreeDeployment *
       if (!glnx_fstat (deployment_dfd, &stbuf, error))
         return FALSE;
 
+      g_assert (self->run_ostree_metadata);
```
alexlarsson, line 816: I wouldn't use an assert here. You may run a newer ostree userspace on an older booted ostree which doesn't have the metadata in the ostree-booted file. We have a fallback anway for the dict not having the data, so just use the same if the file is missing.

cgwalters: This is about the dict being constructed, not about the key present.  Note that previously `/run/ostree-booted` was just an empty file, and I believe creating a GVariantDict out of zero data is a perfectly cromulent empty dict.

src/libostree/ostree-sysroot.c
```diff
@@ -826,7 +838,7 @@ parse_deployment (OstreeSysroot *self, const char *boot_link, OstreeDeployment *
        * be relied on anyways.
        */
       is_booted_deployment
-          = (stbuf.st_dev == self->root_device && stbuf.st_ino == self->root_inode)
+          = (stbuf.st_dev == expected_root_dev && stbuf.st_ino == expected_root_inode)
             || (etc_stbuf.st_dev == self->etc_device && etc_stbuf.st_ino == self->etc_inode);
```
cgwalters, line 842: We should also be able to drop this conditional too, let's try to do that after some more testing.

src/libostree/ostree-sysroot.c
cgwalters: (Just tested by doing a `cp /dev/null /run/ostree-booted`, this works ok)

alexlarsson:
I tried this, but it didn't seem to work:
```
$ ostree admin status
  fedora-coreos cb8c90c9c52d6b7cd4a08024435fe7cfa993b90ba445d8959b8ca1da99044b4d.0
    Version: 38.20240208.dev.6
    origin refspec: fedora:fedora/x86_64/coreos/testing-devel
```
The should have been an asterisk, no?

src/libostree/ostree-sysroot.c
```diff
@@ -1016,10 +1030,21 @@ ostree_sysroot_initialize (OstreeSysroot *self, GError **error)
        * we'll use it to sanity check that we found a booted deployment for example.
        * Second, we also find out whether sysroot == /.
        */
-      if (!glnx_fstatat_allow_noent (AT_FDCWD, OSTREE_PATH_BOOTED, NULL, 0, error))
+      glnx_autofd int booted_state_fd = -1;
```
alexlarsson, line 1033: Should this be != -1?

alexlarsson:
I tried it with
```
 -      const gboolean ostree_booted = booted_state_fd == -1;
+      const gboolean ostree_booted = booted_state_fd != -1;
```
and it works.

cgwalters:
OK I've done some more cleanup, and this definitely works with composefs.  Enabling transient etc blocks ssh logins in the current cosa qemu bits...fixing that is going to need some reworking of the test suite.

src/libostree/ostree-sysroot-private.h
```diff
@@ -77,6 +77,9 @@ struct OstreeSysroot
   dev_t etc_device;
   ino_t etc_inode;
```
alexlarsson, line 78: These can be dropped too

cgwalters:
OK fixed; we had a green CI though before that at least.  Not totally sure about waiting for a full cycle again.

alexlarsson (approved):

PR #3165, opened 2024-02-09
deploy: Honor prepare-root.conf at deploy time for composefs

cgwalters:
lib: Move parsing of composefs config into otcore

So it can be shared with the deployment path.  Prep for dropping
`ex-integrity.composefs`.

---

prepare-root: Fix crash if no keys were found

Handle a NULL pointer.

---

switchroot: Move a define into library too

---

deploy: Honor prepare-root.conf at deploy time

I want to try to get away from the "repository global" configuration
in the repo config.

A major problem is that there's not an obvious way to configure
it as part of an ostree commit/container build - it needs
to be managed "out of band".

With this change, we parse the `usr/lib/ostree/prepare-root.conf`
in the deployment root, and if composefs is enabled there,
then we honor it.

We do still honor `ex-integrity.composefs` but that I think
we can schedule to remove.

---

### DIFF
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -104,6 +104,7 @@ _installed_or_uninstalled_test_scripts = \
 	tests/test-admin-upgrade-systemd-update.sh \
 	tests/test-admin-deploy-syslinux.sh \
 	tests/test-admin-deploy-bootprefix.sh \
+	tests/test-admin-deploy-composefs.sh \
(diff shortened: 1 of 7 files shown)


ericcurtin (approved):
LGTM

cgwalters:
Moving this to https://github.com/ostreedev/ostree/issues/3202

PR #3169, opened 2024-02-12
Expose MOUNT_ATTR_IDMAP detection result to C code

rborn-tx:
This is a follow up to https://github.com/containers/composefs/pull/253.


cgwalters (approved):
LGTM, but note that https://github.com/ostreedev/ostree/pull/3108 will obviate this.

PR #3170, opened 2024-02-12
 prepare-root: Unify root.transient with composefs

cgwalters:
prepare-root: Add missing newline

This is ugly in the output.

---

prepare-root: Unify root.transient with composefs

First, I was totally wrong and composefs handles being passed
an upperdir itself, we don't need to stack overlayfs.

Next, there's really no reason to support `root.transient`
*without* a backing composefs.  The legacy ostree bind mount
and readonly `/usr` is just that - legacy.

Finally, we actually *must* do this to enable both composefs
and transient root, because the prepare-root flow assumes
that it just needs to `MS_MOVE` a *single* mount for the root,
not a stack.

---


cgwalters:
Arguably now it should be something like
`composefs.overlay = true`
instead of `root.transient`...but eh.

PR #3171, opened 2024-02-13
docs/atomic-rollbacks: Add a section on rollbacks

ericcurtin:
Describing how different types of rollbacks work.

### DIFF
--- a/docs/adapting-existing.md
+++ b/docs/adapting-existing.md
@@ -1,5 +1,5 @@
 ---
-nav_order: 6
+nav_order: 70
 ---
 
 # Adapting existing mainstream distributions
--- a/docs/atomic-rollbacks.md
+++ b/docs/atomic-rollbacks.md
@@ -1,0 +1,176 @@
+---
+nav_order: 60
+---
+
+# Atomic Rollbacks
+{: .no_toc }
+
+1. TOC
+{:toc}
+
+## Automatic rollbacks
+
+See [greenboot](https://github.com/fedora-iot/greenboot/blob/main/README.md) for information on automatic rollbacks and how to integrate
+without your bootloader.
+
+## Manual rollbacks
+
+Ostree writes bootloader entries that are interpreted by the bootloader. To
+manually rollback, for bootloaders such as GRUB and syslinux that have an
+interactive UI, it is possible to select a previous boot entry. In the case of
+an Android bootloader, a slot switch may be triggererd using an AB switching
+tool. This may be useful for testing purposes.
+
+## Rollbacks
+
+```
+                        +------------------------------------------------+
++------------------+    |                                                |
+|                  |    |                                                |
+|                  |    |                                                |
+|                  |    |      (ostree:0) latest     (multi-user.target) |
+|                  |    |                                                |
+| Bootloader       |--->+ root                                           |
+|                  |    |                                                |
+|                  |    |      (ostree:1) latest - 1 (multi-user.target) |
+|                  |    |                                                |
+|                  |    |                                                |
++------------------+    |                                                |
+                        +------------------------------------------------+
+```
+
+Bootloaders have multiple boot entries to choose from after upgrade. On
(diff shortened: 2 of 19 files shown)


ericcurtin:
Here is how is looks in .pdf format:

[Atomic Rollbacks _ ostreedev_ostree.pdf](https://github.com/ostreedev/ostree/files/14261110/Atomic.Rollbacks._.ostreedev_ostree.pdf)

docs/adapting-existing.md
```diff
@@ -1,5 +1,5 @@
 ---
-nav_order: 6
```
ericcurtin, line 2: This is a pain, I wonder should we add a 0 to all the nav_orders, like "70" in this case, that way it's easier to stick pages in between previous ones.

cgwalters: Yep SGTM

docs/atomic-rollbacks.md
```diff
@@ -0,0 +1,174 @@
...
+{:toc}
+
+## Automatic rollbacks
+
+See [greenboot](https://github.com/fedora-iot/greenboot/blob/main/README.md) for information on automatic rollbacks and how to integrate
+without your bootloader.
+
+## Manual rollbacks
+
+To manually rollback, an older entry can be selected via GRUB bootloader or in
```
cgwalters, line 18: ostree isn't tied to grub, so I would say e.g.

> ostree writes bootloader entries that are interpreted by the bootloader.  For bootloaders such as GRUB and syslinux that have an interactive UI, it is possible to select a previous boot entry.  For the Android bootloader...

docs/atomic-rollbacks.md
```diff
@@ -0,0 +1,174 @@
...
+
+In the happy path, when we have booted the latest version
+`ostree admin status -D` would output `default`.
+
+ostree-rollback-to-rescue:
+
+```
+#!/usr/bin/bash
+
+set -ex
```
cgwalters, line 142: I'd use exactly the [bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/) setup by default.

docs/atomic-rollbacks.md
```diff
@@ -0,0 +1,174 @@
...
+
+if [ "$(ostree admin status -D)" = "not-default" ]; then
```
cgwalters, line 144: This looks fine, that said...now that we've done all this I am still coming around to the idea that ostree should automatically inject something like `ostree.fallback=/path/to/root` for all non-default entries, that way we can reliably detect this from the initramfs and properly set up the default target from the initramfs.

Anyways, something to just consider for later.

ericcurtin: Well, we normally fallback at the bootloader level right? So if we have loaded a kernel+initrd+cmdline from a fallback boot entry the `ostree=` karg would have the fallback root and `ostree.fallback=` would have the next fallback if that makes sense.

Also... cmdlines are generated server-side, this wouldn't be able to account for clients that missed an upgrade. 

You can skip an upgrade in ostree right, if you haven't upgraded in a while?

That's why it's kinda nice to have the /boot/loader data client-side, fallbacks seem like something that should be client-side. Because each client may have a different upgrade path (and different fallback versions, or the fallback could be on A on some clients and B on others).

docs/atomic-rollbacks.md
ericcurtin: SGTM

docs/atomic-rollbacks.md
ericcurtin: SGTM

docs/atomic-rollbacks.md
ericcurtin: I wanted to do this using systemd only (and no bash), seems to be no string comparison in systemd, so you need something, basically just to do a single string comparison.

ericcurtin:
Addressed comments and re-pushed

cgwalters (approved):

PR #3175, opened 2024-02-15
rofiles-fuse: Check fsverity flag for copyup

cgwalters:
rofiles-fuse: Port to statx

This allows us to query fsverity efficiently.

---

rofiles-fuse: Check fsverity flag for copyup

We need to do a copyup if fsverity is enabled.
Sadly to do this we can't just use ostree_break_hardlink
as is.

---

### DIFF
--- a/ci/gh-install.sh
+++ b/ci/gh-install.sh
@@ -79,6 +79,7 @@ case "$ID" in
             docbook-xsl
             e2fslibs-dev
             elfutils
+            fsverity
             fuse
             gnupg
             gobject-introspection
--- a/ci/installdeps.sh
+++ b/ci/installdeps.sh
@@ -19,7 +19,7 @@ pkg_install_buildroot
 pkg_builddep ostree
 pkg_install sudo which attr fuse strace \
     libubsan libasan libtsan redhat-rpm-config \
-    elfutils
+    elfutils fsverity-utils
 if test -n "${CI_PKGS:-}"; then
     pkg_install ${CI_PKGS}
 fi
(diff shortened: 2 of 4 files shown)


alexlarsson:
This looks good to me, and it works on fedora and CS9, and I think using statx (supported since glibc 2.28) is fine. However STATX_ATTR_VERITY was added in Linux 5.5 and glibc 2.32, which was in 2020. Maybe we should add an ifdef for that to at least build on older systems?

alexlarsson (approved):

alexlarsson:
Btw, this test depends on the "fsverity" tool from fsverity-utils, which is not currently in CS9 (it is in epel though). Maybe the test should check for it?

cgwalters:
Also reworked this with some more cleanups; I started using `ostree_break_hardlink` for the symlink case to notably lessen code duplication.

alexlarsson:
> BTW...I have this lingering vague concern about the changes here to `do_open()`. Previously we had a complex dance of potentially opening the file twice to deal with `O_TRUNC` etc. I maybe what I was trying to do before is avoid race conditions, but in practice we don't need to worry about concurrency on the mount, and the lower shouldn't be concurrently mutated. Does it seem sane to you?

The only two reasons I can see for the original code was to either avoid races, or as an optimization to avoid resolving the path twice. However, given that we don't really care about concurrent mutators of the lower, I don't see anything wrong with the new approach. What would the worry be? The stat seeing a non-hardlinked file, and then something replaces the filename with a hard-link to the repo that we open writable. In some kind of generic usecase it may be a problem, but as used in ostree where the lower dir is not meant to change it doesn't seem like a real risk

And anyway, I don't see how the original code would help much against concurrent mutators anyway in a more "generic" usecase. Sure, you decide whether to copy-up or not based on the single resolve done during open(), but someone could create a second link to the file between the open and the writes to it.

alexlarsson:
So, lgtm

PR #3179, opened 2024-02-16
docs: Add webrick dependancy for building site locally

ericcurtin:
This mimics the GitHub Pages environment so that you can build and serve the site locally for testing. It requires webrick these days.


cgwalters (approved):
Unrelated directly to this but personally I'd be happiest to switch to a Go or Rust static site generator

PR #3181, opened 2024-02-17
README: Add Red Hat In-Vehicle Operating System

ericcurtin:
RHIVOS is a derivative of CentOS Automotive Stream Distribution that uses OSTree, it's closest Fedora derivative is Fedora IoT although it was created as it's own distribution.


dougsland:
LGTM

Issue #3183, opened 2024-02-19
installed-tests fail with 2024.3: test-admin-deploy-var.sh: line 24: OSTREE_FEATURES: unbound variable


smcv:
While running libostree's "as-installed" tests, via ginsttest-runner and Debian's autopkgtest framework:

```
189s Executing: libostree/test-admin-deploy-var.sh.test
189s ++ _have_user_xattrs=
189s ++ _have_systemd_and_libmount=
189s ++ _have_strace_fault_injection=
189s ++ libtest_exit_cmds+=(libtest_cleanup_gpg)
189s ++ ED25519PUBLIC=
189s ++ ED25519SEED=
189s ++ ED25519SECRET=
189s /usr/libexec/installed-tests/libostree/test-admin-deploy-var.sh: line 24: OSTREE_FEATURES: unbound variable
```

Probably tests that are meant to work "as-installed" should use `ostree --version | grep initial-var` instead?

PR #3184, opened 2024-02-19
test-admin-deploy-var: Don't rely on OSTREE_FEATURES

smcv:
* tests: Generalize has_gpgme, has_sign_ed25519 into has_ostree_feature

* tests: Use skip_without_ostree_feature to detect libarchive, composefs
    
    This avoids false negatives from `ostree --version | grep -q ...` exiting with failure under `set -o pipefail` because `grep -q` can exit as soon as it sees the desired string, leaving `ostree --version` to be terminated by `SIGPIPE` next time it writes to stdout.

* test-admin-deploy-var: Don't rely on OSTREE_FEATURES
    
    This is set during build-time testing, but unset during "as-installed" tests.
    
    Resolves: https://github.com/ostreedev/ostree/issues/3183

### DIFF
--- a/tests/libtest.sh
+++ b/tests/libtest.sh
@@ -696,18 +696,20 @@ skip_without_fuse () {
     [ -e /etc/mtab ] || skip "no /etc/mtab"
 }
 
-has_gpgme () {
-    local ret
+has_ostree_feature () {
+    local ret=0
+    # Note that this needs to write to a file and then grep the file, to
+    # avoid ostree --version being killed with SIGPIPE and exiting with a
+    # nonzero status under `set -o pipefail`.
     ${CMD_PREFIX} ostree --version > version.txt
-    grep -q -e '- gpgme' version.txt
-    ret=$?
+    grep -q -e "- $1\$" version.txt || ret=$?
     rm -f version.txt
     return ${ret}
 }
 
-skip_without_gpgme() {
-    if ! has_gpgme; then
-        skip "no gpg support compiled in"
+skip_without_ostree_feature () {
+    if ! has_ostree_feature "$1"; then
+        skip "no $1 support compiled in"
     fi
 }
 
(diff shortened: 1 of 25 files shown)


smcv:
This was more involved than I had expected!

jlebon (approved):
Very nice, thanks!

tests/libtest.sh
```diff
@@ -696,18 +696,20 @@ skip_without_fuse () {
     [ -e /etc/mtab ] || skip "no /etc/mtab"
 }
 
-has_gpgme () {
-    local ret
+has_ostree_feature () {
+    local ret=0
+    # Note that this needs to write to a file and then grep the file, to
+    # avoid ostree --version being killed with SIGPIPE and exiting with a
+    # nonzero status under `set -o pipefail`.
```
cgwalters, line 703: Oh man...wow.

smcv: As I said, more involved than I expected. This PR brought to you by the dept. of Unix arcana.

dbnicholson: Even though it's eminently useful, I stopped using `pipefail` years ago after being burned by the same issue in `tar`. It might be nice to install a `SIGPIPE` handler in the [version printing code](https://github.com/ostreedev/ostree/blob/1b3ffe54e87c108b576a528efa6493f935a80f27/src/ostree/ot-main.c#L511) since this is likely to affect other `ostree` consumers.

cgwalters: One thing I've said many times with coworkers in live video chats, but rarely externally is that a nontrivial component of why I still (20+ years later) wake up and log into my computer for this job is that I just really enjoy collaborating with lots of smart people across the internet.  Also there's the continual "surprise" factor in computers as in "what did someone figure out how to make them do this year" (and right now it's AI).

Anyways this little bit combines those two - I just want to say thanks @smcv as one of those smart people on the Internet for everything you do :pray: - and this little bit of pipefail is one of those surprises :smile:

dbnicholson: See #3203.

PR #3186, opened 2024-02-20
ostree-prepare-root: Amend comment about shared mounts

rborn-tx:
This fixes a comment mentioning a kernel bug which is not really a bug but expected behavior.


jlebon (approved):

jlebon:
Thanks!

Issue #3187, opened 2024-02-20
Hotfix functionality incompatible with composefs+verity


rborn-tx:
I noticed the hotfix functionality (related to `ostree admin unlock --hotfix`) is available even when one is using composefs (at least the part that applies the hotfix).

Specifically, in this [stretch of code](https://github.com/ostreedev/ostree/blob/v2024.3/src/switchroot/ostree-prepare-root.c#L576), `ostree-prepare-root` checks for the existence of a "hotfix overlay" by looking at the deployment directory even when composefs is in use. Unless I'm missing something, this would allow one to apply a hotfix on top of a composefs+verity image and hence bypass the integrity/authenticity guarantees you would normally expect. I suppose this is not intended, right?

Since the hotfix is made of overlay directories (workdir, upperdir) which aren't signed I suppose composefs and hotfixing are not compatible...

alexlarsson:
Yeah. I don't think should allow this for a signed composefs.

cgwalters:
Thanks for noticing this!

PR #3194, opened 2024-02-22
prepare-root: Disallow hotfixes if using signed composefs images

alexlarsson:
As mentioned in https://github.com/ostreedev/ostree/issues/3187, we can't allow a hotfix overlay of /usr when using signed composefs images as that would allow an attacker to persist something used across boots.


jmarrero (approved):
lgtm

PR #3205, opened 2024-02-27
bootloader/grub2: Don't do anything if we have static configs

cgwalters:
This builds on top of https://github.com/coreos/bootupd/pull/609/commits/fa9924e4fe403c3751392c041cd98614a2cc3611 (But in a very hacky way because we don't currently link to a JSON library)

Basically, bootupd supports injecting static configs, and this is the currently least hacky way for us to detect this and understand that we shouldn't try to run `grub2-mkconfig`.

A further patch I'd like to do here is also change the probing logic to gracefully no-op if `grub2-mkconfig` doesn't exist, but that has a bit more risk and involvement.

### DIFF
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -26,6 +26,11 @@
 
 #include <string.h>
 
+// Written by bootupd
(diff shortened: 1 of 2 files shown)


jmarrero (approved):
lgtm

Issue #2961, opened 2023-07-31
grub and ostree disagree on BLS versioning


cgwalters:
OK so I was playing around with alternative stateroots (i.e. `ostree admin os-init backup; ostree admin deploy --os=backup $commit`) and made an unfortunate discovery.

Today, ostree writes BLS files with a `version` entry in them, and that's what it thinks the bootloader order should be.

grub2 though seems to be parsing the *filenames*...as if they were RPM packages actually.  See this code: https://github.com/rhboot/grub2/blob/7de33f4072abf5f14c4fdad6e566c58bdbf3b26e/grub-core/commands/blscfg.c#L400

And what's ugly here is what ostree writes into the filenames looks like `ostree-$version-$stateroot`.  Except...that parsing code splits on `-`.  So things go badly if the stateroot name contains `-`...as it does for `fedora-coreos`.

We end up with (I think)

name=ostree-$version
version=fedora
release=coreos

This isn't exactly problem if one only has one stateroot...we'll still correctly sort on the ordering of "name" here...it's just ugly.

However, it does completely break the ability to use alternative stateroots because then the parsing of things just breaks.

I think we could probably change the filenames we emit to match this schema.  Hmm...probably we need the version to go first to reliably fix this, and then encode everything else into the version field or so?

cgwalters:
I think last time I looked at this it looked like zipl at least did read the `version` field.

PR #3206, opened 2024-02-28
sysroot: Turn on bootloader-naming-2 by default

cgwalters:
I think it's about time we flipped this on by default; like the bootprefix I was a bit too chicken.  We still have a `bootloader-naming-1` that can be flipped on in case of some regression.

Closes: https://github.com/ostreedev/ostree/issues/2961

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1853,6 +1853,30 @@ parse_os_release (const char *contents, const char *split)
   return ret;
 }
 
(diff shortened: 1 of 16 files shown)


ericcurtin (approved):
This LGTM, although I am only familiar with grub BLS-style OSTree and the Android Bootloader variant.

I have wondered this in the past, do we rely on more the filename for versioning or the version number in the files, it didn't seem 100% clear when I read the code to determine that. I wonder should we document what the actual index is.

PR #3208, opened 2024-03-04
kargs: parse spaces in kargs input and keep quotes

HuijingHei:
According to Jonathan's suggestion, should fix the code from
ostree repo.

With this patch:
- kargs input like "init_on_alloc=1 init_on_free=1", will be
parsed as 2 seperated args `init_on_alloc=1` and `init_on_free=1`,
instead of whole;
- According to https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html,
need to keep spaces in double-quotes, like `param="spaces in here"`
will be parsed as whole instead of 3.

Fixes https://github.com/coreos/rpm-ostree/issues/4821

### DIFF
--- a/src/libostree/ostree-kernel-args.c
+++ b/src/libostree/ostree-kernel-args.c
@@ -163,6 +163,44 @@ strcmp0_equal (gconstpointer v1, gconstpointer v2)
   return g_strcmp0 (v1, v2) == 0;
 }
 
+/* Split string with spaces, but keep quotes.
+   For example, "test=\"1 2\"" will be saved as whole.
+ */
+static char **
+split_kernel_args (const char *str)
+{
+  gboolean quoted = FALSE;
+  g_return_val_if_fail (str != NULL, NULL);
+
+  GPtrArray *strv = g_ptr_array_new ();
+
+  size_t len = strlen (str);
+  // Skip first spaces
+  const char *start = str + strspn (str, " ");
+  for (const char *iter = start; iter && *iter; iter++)
+    {
+      if (*iter == '"')
+        quoted = !quoted;
+      else if (*iter == ' ' && !quoted)
+        {
+          g_ptr_array_add (strv, g_strndup (start, iter - start));
+          start = iter + 1;
+        }
+    }
+
+  // Add the last slice
+  if (!quoted)
+    g_ptr_array_add (strv, g_strndup (start, str + len - start));
+  else
+    {
+      g_debug ("Missing terminating quote in '%s'.\n", str);
+      g_assert_false (quoted);
+    }
+
+  g_ptr_array_add (strv, NULL);
+  return (char **)g_ptr_array_free (strv, FALSE);
+}
+
 /**
  * ostree_kernel_args_new: (skip)
  *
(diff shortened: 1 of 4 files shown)


HuijingHei:
Do local testing and failed `free(): double free detected in tcache 2`, the error is vague, need to do more investigation, and any pointer is appreciated, thanks!
```
# cosa build-fast
# cosa run --qemu-image xx
# rpm-ostree kargs --append "test=\"1 2\""
[root@cosa-devsh ~]# rpm-ostree kargs --delete "test=\"1 2\""
error: Bus owner changed, aborting. This likely means the daemon crashed; check logs with `journalctl -xe`.

[root@cosa-devsh ~]# journalctl -u rpm-ostreed.service
...
Mar 04 08:54:33 cosa-devsh rpm-ostree[1904]: free(): double free detected in tcache 2

# journalctl -xe
Stack trace of thread 2121:
#0  0x00007fd005f5a834 __pthread_kill_implementation (libc.so.6 + 0x90834)
#1  0x00007fd005f088ee raise (libc.so.6 + 0x3e8ee)
#2  0x00007fd005ef08ff abort (libc.so.6 + 0x268ff)
#3  0x00007fd005ef17d0 __libc_message.cold (libc.so.6 + 0x277d0)
#4  0x00007fd005f647a5 malloc_printerr (libc.so.6 + 0x9a7a5)
#5  0x00007fd005f66c6f _int_free (libc.so.6 + 0x9cc6f)
#6  0x00007fd005f693de free (libc.so.6 + 0x9f3de)
#7  0x00007fd006cdf865 g_free (libglib-2.0.so.0 + 0x5e865)
#8  0x00007fd0067bd4d1 g_autoptr_cleanup_generic_gfree (libostree-1.so.1 + 0x7a4d1)
#9  0x000056187a05d7b3 kernel_arg_apply_patching (rpm-ostree + 0x5ad7b3)
#10 0x000056187a059f7b transaction_execute_thread (rpm-ostree + 0x5a9f7b)
#11 0x00007fd006e82774 g_task_thread_pool_thread (libgio-2.0.so.0 + 0xb7774)
#12 0x00007fd006d11db2 g_thread_pool_thread_proxy.lto_priv.0 (libglib-2.0.so.0 + 0x90db2)
#13 0x00007fd006d0d523 g_thread_proxy (libglib-2.0.so.0 + 0x8c523)
#14 0x00007fd005f58897 start_thread (libc.so.6 + 0x8e897)
#15 0x00007fd005fdf80c __clone3 (libc.so.6 + 0x11580c)
```

ericcurtin:
Might be worth taking a peak at kernel/systemd and other karg parsing implementations, this kind of thing has been done in other places before.

Sometimes for this kinda thing it's worth testing loads of random things in a small implementation for quick iterations:

```
#include <stdio.h>

static char **
split_whitespace (const char *str)
{
  gboolean quoted = FALSE;
  g_return_val_if_fail (str != NULL, NULL);

  GPtrArray *strv = g_ptr_array_new ();

  guint len = strlen (str);
  guint start;
  // Skip first whitespaces
  for (start = 0; start < len && str[start] == ' '; start++)
    {
    }
  for (guint i = start; str[i] != '\0'; i++)
    {
      if (str[i] == ' ' && !quoted)
        {
          g_ptr_array_add (strv, g_strndup (str + start, i - start));
          start = i + 1;
        }
      if (str[i] == '"')
        {
          quoted = !quoted;
        }
      if (i == (len - 1))
        {
          if (!quoted)
            g_ptr_array_add (strv, g_strndup (str + start, len - start));
        }
    }
  g_ptr_array_add (strv, NULL);
  return (char **)g_ptr_array_free (strv, FALSE);
}

int main() {
  split_whitespace(argv[1]);
}
```

then add as an automated test. I didn't look at the code in any great detail, but on first glance it looks like it could be a string manipulation bug based on the double free.

seems to me that we don't exactly want to split by whitespace here.

ericcurtin:
I also wonder should you just try and edit the BLS files directly rather than change this feature.

grubby has a similar feature which I've had trouble with in the past, these days I find editing the BLS directly easier.

tests/test-admin-deploy-karg.sh
```diff
@@ -77,3 +77,10 @@ assert_not_file_has_content sysroot/boot/loader/entries/ostree-2-testos.conf 'op
 assert_not_file_has_content sysroot/boot/loader/entries/ostree-2-testos.conf 'options.*APPENDARG=VALAPPEND'
 
 echo "ok deploy --karg-delete"
+
+${CMD_PREFIX} ostree admin deploy  --os=testos --karg-append="test=\"1 2\"" testos:testos/buildmain/x86_64-runtime
+assert_file_has_content sysroot/boot/loader/entries/ostree-2-testos.conf 'options.*test="1 2"'
+${CMD_PREFIX} ostree admin deploy  --os=testos --karg-delete="test=\"1 2\"" testos:testos/buildmain/x86_64-runtime
```
ericcurtin, line 83: The delete interface might be a little simpler/cleaner if it expected key only (the test part)

HuijingHei:
Thanks @ericcurtin for the review, the error is fixed with the help of @jmarrero , it is because the previous change in `ostree_kernel_args_delete()`, it is for debug and remove that part, then no error.

ericcurtin (approved):
LGTM, might want to run this through gcc's "-fanalyzer" to see if it identifies any potential problems, string manipulation is always tricky.

src/libostree/ostree-kernel-args.c
```diff
@@ -163,6 +163,49 @@ strcmp0_equal (gconstpointer v1, gconstpointer v2)
...
+          start = i + 1;
+        }
+
+      if (i == (len - 1))
+        {
+          if (!quoted)
+            g_ptr_array_add (strv, g_strndup (str + start, len - start));
+          else
+            {
+              g_printerr ("Error: Missing terminating quote in '%s'.\n", str);
```
cgwalters, line 199: We shouldn't print to stderr from a library as a general rule; it's either a fatal error (`GError`) or it's something we accept and log with `g_debug()` by default.

The question here is what does the Linux kernel do?  If it accepts it then we should too.  I suspect it just accepts it.

src/libostree/ostree-kernel-args.c
```diff
@@ -163,6 +163,49 @@ strcmp0_equal (gconstpointer v1, gconstpointer v2)
...
+ */
+static char **
+split_whitespace (const char *str)
+{
+  gboolean quoted = FALSE;
+  g_return_val_if_fail (str != NULL, NULL);
+
+  GPtrArray *strv = g_ptr_array_new ();
+
+  guint len = strlen (str);
```
cgwalters, line 177: Technically this should be `size_t` not `guint`; they're different sizes.  A giant pitfall in C is that it will happily silently cast numeric types.  Of course, we shouldn't be dealing with >4G strings but still.

src/libostree/ostree-kernel-args.c
```diff
@@ -163,6 +163,49 @@ strcmp0_equal (gconstpointer v1, gconstpointer v2)
...
+  guint start;
+  // Skip first whitespaces
+  for (start = 0; start < len && str[start] == ' '; start++)
+    {
+    }
+  for (guint i = start; str[i] != '\0'; i++)
```
cgwalters, line 183: So using a pointer:

```
const char *start = str + strspn (str, " ");

`for (const char *iter = start; iter; iter++) {`

File: src/libostree/ostree-kernel-args.c [hunk 5]
```diff
@@ -163,6 +163,49 @@ strcmp0_equal (gconstpointer v1, gconstpointer v2)
...
+    {
+      if (str[i] == '"')
```
cgwalters, line 185: Then this is `*iter` instead of `str[i]`.

src/libostree/ostree-kernel-args.c
```diff
@@ -163,6 +163,49 @@ strcmp0_equal (gconstpointer v1, gconstpointer v2)
...
+        quoted = !quoted;
+      else if (str[i] == ' ' && !quoted)
+        {
+          g_ptr_array_add (strv, g_strndup (str + start, i - start));
+          start = i + 1;
+        }
+
+      if (i == (len - 1))
```
cgwalters, line 193: I believe this condition can actually be outside of the loop.

ericcurtin (approved):

PR #3212, opened 2024-03-12
deployment: enable adding custom metadata

jmarrero:
This is a Work in progress, still need to re-work it with GVariant and test it, but it builds locally.

### DIFF
--- a/src/libostree/ostree-deployment.c
+++ b/src/libostree/ostree-deployment.c
@@ -18,6 +18,7 @@
 #include "config.h"
 
 #include "ostree-deployment-private.h"
+#include "ostree-sysroot-private.h"
 #include "ostree.h"
 #include "otutil.h"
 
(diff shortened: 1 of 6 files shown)


src/ostree/ot-admin-builtin-metadata.c
```diff
@@ -0,0 +1,77 @@
...
+      ostree_deployment_set_ext_metadata (first_deployment, key, value, error);
+    }
+
+  if (opt_get)
+    {
+      ostree_deployment_get_ext_metadata (first_deployment, *opt_get, error);
+    }
+
+  ret = TRUE;
+out:
```
cgwalters, line 75: In brand new code we can avoid use of `goto` and just use `return FALSE` above

src/ostree/ot-admin-builtin-metadata.c
```diff
@@ -0,0 +1,77 @@
...
+    {
+      char *key = strtok (*opt_set, "=");
+      char *value = strtok (NULL, "=");
+      // ^^ This needs error checking and probably is wrong... but builds!
+      ostree_deployment_set_ext_metadata (first_deployment, key, value, error);
+    }
+
+  if (opt_get)
+    {
+      ostree_deployment_get_ext_metadata (first_deployment, *opt_get, error);
```
cgwalters, line 71: Need to handle errors here

src/libostree/ostree-deployment.c
```diff
@@ -476,3 +477,62 @@ ostree_deployment_is_finalization_locked (OstreeDeployment *self)
...
+ * @error: a #GError
+ * Returns: The value of a extended attribute(metadata) checksum given the key.
+ */
+const char *
+ostree_deployment_get_ext_metadata (OstreeDeployment *self, const char *metadata_key,
+                                    GError **error)
+{
+  g_autofree char *backing_relpath = _ostree_sysroot_get_deployment_backing_relpath (self);
+  g_autofree char *metadata_value = NULL;
+  g_autofree int len = getxattr (backing_relpath, metadata_key, NULL, 0);
```
cgwalters, line 517: In the general case ostree wants to support using fd-relative paths, so we need to open a fd for the dir (relative to `self->sysroot_fd`), then use `fgetxattr()`.

cgwalters: Also
> g_autofree int len

Definitely don't want to call `free()` on an integer.  Also, `getxattr` returns a `ssize_t`, not an `int` (they're different sizes).

src/libostree/ostree-deployment.h
```diff
@@ -107,4 +107,12 @@ const char *ostree_deployment_unlocked_state_to_string (OstreeDeploymentUnlocked
 _OSTREE_PUBLIC
 OstreeDeploymentUnlockedState ostree_deployment_get_unlocked (OstreeDeployment *self);
 
+_OSTREE_PUBLIC
+gboolean ostree_deployment_set_ext_metadata (OstreeDeployment *self, const char *metadata_key,
```
cgwalters, line 111: New API needs to be added to `libostree-devel.sym`; there's some ceremony to do that.  See the comments around that and `git log`.

jmarrero:
Closing as we have moved on from this work. If we do something like this in future the we can revisit.

PR #3216, opened 2024-03-15
curl: Also map HTTP errors for retries

cgwalters:
When we added the retry logic, the intention here was definitely to do it not just for network errors but also e.g. HTTP 500s and the like.

xref https://pagure.io/releng/issue/11439
where we rather painfully debugged that this was missing.


jmarrero (approved):
lgtm

Issue #3219, opened 2024-03-20
prepare-root: Avoid retries


cgwalters:
I've seen some failure cases where when `ostree-prepare-root.service` fails, it ends up getting retried (not totally sure how)...and what happens is we get:

`error: Failed to create /run/ostree: File exists`

And that combined with other noise makes the *real* error scroll off the TTY...

So...I think we need to dig in and try to;

- Ensure our unit doesn't get rerun
- Try to add cleanup/idempotence on failure so we *can* rerun (a good bit harder)

jlebon:
https://github.com/ostreedev/ostree/pull/3220 should definitely help. Hard to tell if it fixes it completely unless you can get a reproducer for the specific issue you've hit.

michaelst0:
We've run into this issue a couple of times `error: Failed to create /run/ostree: File exists`.

We're running on RPi 4 compute module with Fedora IoT 41 using SD card for storage. We _think_ that the cause is powering off during the boot stage. The next boot _can_ end up in the failed state. 

Here's a screenshot my colleague took:
![Image](https://github.com/user-attachments/assets/eaea6533-0866-4285-b176-8806b8844a1d)

PR #3220, opened 2024-03-21
ostree-prepare-root.service: add OnFailureJobMode=isolate

jlebon:
This is stronger than the default (`replace`) because it tells systemd to *stop everything* and go to `emergency.target`. In other codebases, this has definitely helped me with the problem of "systemd keeps going even after a failure".

Likely addresses #3219.

See also e.g. https://github.com/coreos/ignition-dracut/commit/3d2e165f97f30c1e62577357f27f32e60e6add18.


cgwalters (approved):
I actually tried this yesterday and it wasn't enough.  Spent a bit of time staring at systemd debug logging but couldn't figure it out.

That said, it definitely doesn't hurt so...

PR #3227, opened 2024-04-08
_ostree_ensure_fsverity: Properly check for errors

alexlarsson:
If fs_verity_wanted == _OSTREE_FEATURE_YES we should fail if !suported, but we were checking !supported where supported is a pointer, not a boolean. This caused us to miss errors when the kernel didn't support fs-verity that lead to lots of debugging.

### DIFF
--- a/src/libostree/ostree-repo-verity.c
+++ b/src/libostree/ostree-repo-verity.c
@@ -224,9 +224,10 @@ _ostree_tmpf_fsverity (OstreeRepo *self, GLnxTmpfile *tmpf, GBytes *signature, G
 
 gboolean
 _ostree_ensure_fsverity (OstreeRepo *self, gboolean allow_enoent, int dirfd, const char *path,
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
LGTM. 

(Another one where Rust wouldn’t have let this happen)

ericcurtin:
Could be a good C/C++ compiler warning/error addition that you could turn on optionally, ensure pointers get checked against NULL in C or nullptr in C++ and that booleans get checked against true/false etc. For new C or C++ codebases to switch on from the start.

ericcurtin (approved):

ericcurtin:
Tagging @davidmalcolm in case it gives inspiration to a feature (although I'm sure this kinda check has been discussed many times before)

ericcurtin:
I guess the header files from libraries might make these warnings difficult among other things :sweat_smile:

PR #3229, opened 2024-04-12
init-fs: Add --epoch

cgwalters:
I want to add another variant here, and `--modern` is now old.  Let's acknowledge that we may want to make even more changes in the future.  So `--modern == --epoch=1` but I will add `--epoch=2` after this.

### DIFF
--- a/src/ostree/ot-admin-builtin-init-fs.c
+++ b/src/ostree/ot-admin-builtin-init-fs.c
@@ -28,9 +28,13 @@
 #include <glib/gi18n.h>
 
 static gboolean opt_modern;
+static gint opt_epoch;
 
 static GOptionEntry options[]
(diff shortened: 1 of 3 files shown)


src/ostree/ot-admin-builtin-init-fs.c
```diff
@@ -62,12 +66,19 @@ ot_admin_builtin_init_fs (int argc, char **argv, OstreeCommandInvocation *invoca
...
+  if (opt_modern && opt_epoch == 0)
+    opt_epoch = 1;
+
   /* See https://github.com/coreos/coreos-assembler/pull/688
    * For Fedora CoreOS at least, we have this now to the point where we don't
    * need this stuff in the physical sysroot.  I'm not sure we ever really did,
    * but to be conservative, make it opt-in to the new model of just boot/ and ostree/.
    */
-  if (!opt_modern)
+  if (opt_epoch == 0)
```
ericcurtin, line 81: Type correct comparisons I see! :smile:

PR #3231, opened 2024-04-15
prepare-root: Handle non-AB aboot properly

alexlarsson:
otcore_get_ostree_target() should set is_aboot for android boot systems, but currently it only does this on A/B boot systems, not single-boot-partition systems. Fix this by setting it in the second case.


ericcurtin (approved):
There's three types of aboot I guess:

1. Normal AB switching mode
2. androidboot.slot_suffix=_a but no AB switching tool available, so we always boot into a
3. similar to 2, but we have specified a root= karg, booting in development mode, causing androidboot.slot_suffix to not be set at all.

PR #3233, opened 2024-04-25
Sync summary times

dbnicholson:
I was looking into an HTTP caching issue in a different project and it occurred to me that there's no reason the `summary` and `summary.sig` files can't have synchronized mtimes since they're now created in a temporary directory. The first commit is unrelated and I can split it out if preferred.

### DIFF
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -6092,6 +6092,23 @@ regenerate_metadata (OstreeRepo *self, gboolean do_metadata_commit, GVariant *ad
       && !_ostree_sign_summary_at (sign, self, summary_tmpdir.fd, sign_keys, cancellable, error))
     return FALSE;
 
+  /* If a signature was made, sync the summary times to it. This way an
+   * HTTP client will consider the files expired at the same time.
+   */
+  if (gpg_key_ids != NULL || sign_keys != NULL)
+    {
+      struct stat stbuf;
+      if (!glnx_fstatat (summary_tmpdir.fd, "summary.sig", &stbuf, AT_SYMLINK_NOFOLLOW, error))
+        return glnx_prefix_error (error, "Unable to get summary.sig status");
+
+      struct timespec ts[2];
+      ts[0] = stbuf.st_atim;
+      ts[1] = stbuf.st_mtim;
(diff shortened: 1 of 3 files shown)


cgwalters (approved):

dbnicholson:
Oof, forgot about `clang-format`. Can't say I agree with it's decision, but I will appease it.

dbnicholson:
Looks like `clang-format` in rawhide or whatever (clang 18?) disagrees with older versions. I think this is WRT to [AlignOperands](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignoperands) and [BreakBeforeBinaryOperators](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#breakbeforebinaryoperators).

dbnicholson:
Should I just override the failure? It's in a part of code that I didn't touch and I don't know how to get different versions of `clang-format` to agree on what the right thing is:

```
[2024-04-25T15:39:35.622Z] checking clang-format... src/libostree/ostree-sysroot.c:2241:56: error: code should be clang-formatted [-Wclang-format-violations]
[2024-04-25T15:39:35.622Z]                   ? _ostree_sysroot_get_runstate_path (
[2024-04-25T15:39:35.622Z]                                                        ^
[2024-04-25T15:39:35.622Z] src/libostree/ostree-sysroot.c:2243:56: error: code should be clang-formatted [-Wclang-format-violations]
[2024-04-25T15:39:35.622Z]                   : _ostree_sysroot_get_runstate_path (
[2024-04-25T15:39:35.622Z]                                                        ^
```

Possibly `ci/codestyle.sh` should only be run from one target so there's only one version of `clang-format` in use.

cgwalters:
> Possibly ci/codestyle.sh should only be run from one target so there's only one version of clang-format in use.

Yeah, this has bit us at least twice in the past I believe.  It's so messy...even if we pick just one target we force everyone to use that specific version which is not great.

For now...yes I'll follow up to drop the clang-format from everything except just one OS version.

dbnicholson:
> Yeah, this has bit us at least twice in the past I believe. It's so messy...even if we pick just one target we force everyone to use that specific version which is not great.

I was thinking the same thing. A simple thing would be a wrapper script that runs `clang-format` using a container image for whatever is the agreed upon target. Then it would be simpler for developers to match whatever was going to happen in CI. My gut says rawhide because that's always going to have the most recent clang version, but there might be too much chance of random distro breakage.

PR #3234, opened 2024-04-30
ostree-sysroot-deploy: check if deployments are in the same stateroot.

jmarrero:
Our code currently does not check for stateroot as a differentiator between deployments, however given that we allow multiple stateroots, we should see them as different deployments in the same way we check for kargs for example.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2383,6 +2383,12 @@ deployment_bootconfigs_equal (OstreeRepo *repo, OstreeDeployment *a, OstreeDeplo
   if (g_strcmp0 (a_version, b_version) != 0)
     return FALSE;
 
+  /* same stateroot? */
+  const char *a_stateroot = ostree_deployment_get_osname (a);
+  const char *b_stateroot = ostree_deployment_get_osname (b);
+  if (g_strcmp0 (a_stateroot, b_stateroot) != 0)
+    return FALSE;
+
   return TRUE;
 }
 
--- a/Makefile-tests.am
+++ b/Makefile-tests.am
@@ -125,6 +125,7 @@ _installed_or_uninstalled_test_scripts = \
 	tests/test-admin-locking.sh \
 	tests/test-admin-deploy-clean.sh \
 	tests/test-admin-kargs.sh \
+        tests/test-admin-stateroot.sh \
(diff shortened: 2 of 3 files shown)


ericcurtin (approved):
I agree with this, but I am curious what is the use case where all the previous checks have not returned FALSE already. The case where this happens in real-life. We can't depend on changing kargs this is correct, we use ostree=true in our Android Bootloader images for example.

jmarrero:
thanks for the review!
Single Node Openshift uses stateroots for managing faster upgrades. And they hit this.

See OCPBUGS-30276

just trying to confirm locally this works as intended before moving this out of draft.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2383,6 +2383,12 @@ deployment_bootconfigs_equal (OstreeRepo *repo, OstreeDeployment *a, OstreeDeplo
   if (g_strcmp0 (a_version, b_version) != 0)
     return FALSE;
 
+  /* same stateroot? */
+  g_autofree char *a_stateroot = ostree_deployment_get_osname (a);
```
cgwalters, line 2387: These should both just be `const char *`

cgwalters (changes requested):
Ooh, good catch.  Fun bug.  I am actually looking forward to us trying to reuse the stateroot stuff more broadly as it's a very powerful thing.

I think it'd be relatively straightforward to do a test case for this in tests/admin-test.sh or so, I can help with that

jmarrero:
I can see the failure reported in the original issue with the test when I remove my changes:

```
error: Reloading deployments after commit: loading sysroot: Parsing deployment /ostree/boot.0/testos2/7bd091ad6955c0e7ef41bab65adb016275664b55efc0d10a0c8cb263da3c1e4e/0 in stateroot 'testos2': readlinkat: No such file or directory
```

cgwalters (approved):

cgwalters:
> [2024-05-02T00:14:50.093Z] FAIL: tests/test-libarchive-import 11 /libarchive/selinux - OSTree:ERROR:tests/test-libarchive-import.c:681:test_libarchive_selinux: assertion failed (error == NULL): Child process exited with code 1 (g-spawn-exit-error-quark, 1)


Hmm that one is weird.

cgwalters:
@jmarrero I think this just needs a rebase :surfer:

PR #3236, opened 2024-05-02
ci: Only run clang-format on ubuntu-stable GH runner

cgwalters:
Previously we were running clang-format across multiple operating system versions and hence clang versions, and it turns out clang has changed the preferred formatting multiple times.

We could *probably* dig in and try to pin things more strongly but eh...for now let's arbitrarily just use whatever's in the default GH Action ubuntu-latest runner as that should be equally painful for everyone.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -26,6 +26,21 @@ jobs:
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
       - name: Test style
         run: ./ci/ci-commitmessage-submodules.sh
+  clang-format:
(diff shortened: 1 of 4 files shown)


jmarrero (approved):
lgtm

ericcurtin (approved):

dbnicholson (approved):
That makes sense and obviously the test failure is unrelated (but worth looking at closer).

cgwalters:
> the test failure is unrelated (but worth looking at closer).

Yes...something changed recently with the `container` environment variable that's triggering this.

PR #3237, opened 2024-05-02
Container env

cgwalters:


cgwalters:
OK yes, somehow the pods launched by coreosci lost the `container=` environment variable.

cgwalters:
https://github.com/coreos/fedora-coreos-config/pull/2992

Issue #3022, opened 2023-08-30
`ostree admin` fails if `/ostree/deploy` has mtime 0


fwilhe:
Hi everyone, I have an issue that seems hard to describe, I'll do my best to explain what I did:

I'm building a debian-based OSTree system.
I'm using the [ostree package from debian testing](https://packages.debian.org/trixie/ostree).

My system does boot successfully and it looks good on the first sight.

I can run `ostree` commands like:

```
root@e931861b90bf:~# ostree refs
gardenlinux/today/arm64
ostree/1/1/0
root@e931861b90bf:~# ostree log gardenlinux/today/arm64
commit 54d0898411d9a709685bc1ff6b3b4fafbdc16402c8c4e014ad84488b7e39fb59
ContentChecksum:  9a7c454a3150e9d42e9ccf1743c5b51c4e588799423e13235d4c95882f4bc993
Date:  2023-08-28 14:42:53 +0000

    Gardenlinux build Mon Aug 28 14:42:50 UTC 2023

root@e931861b90bf:~# ostree ls 54d0898411d9a709685bc1ff6b3b4fafbdc16402c8c4e014ad84488b7e39fb59
d00755 0 0      0 /
l00777 0 0      0 /bin -> usr/bin
l00777 0 0      0 /home -> var/home
l00777 0 0      0 /lib -> usr/lib
l00777 0 0      0 /ostree -> sysroot/ostree
l00777 0 0      0 /root -> var/roothome
l00777 0 0      0 /sbin -> usr/sbin
... (output abbreviated)
```

But as soon as I'm trying to run any of the `ostree admin` subcommands, I get an error because of a failed assertion:

```
root@e931861b90bf:~# ostree admin status
**
OSTree:ERROR:src/libostree/ostree-sysroot.c:1333:ostree_sysroot_get_deployments: assertion failed: (self->loadstate == OSTREE_SYSROOT_LOAD_STATE_LOADED)
Bail out! OSTree:ERROR:src/libostree/ostree-sysroot.c:1333:ostree_sysroot_get_deployments: assertion failed: (self->loadstate == OSTREE_SYSROOT_LOAD_STATE_LOADED)
Aborted
```

This does seem to happen with any sub-command of `ostree admin`.

```
root@e931861b90bf:~# ostree admin cleanup
**
[...]

PR #3243, opened 2024-05-15
sysroot: Handle `/ostree/deploy` having epoch 0

cgwalters:
Ironically we break if the timestamp there is zero.

Closes: https://github.com/ostreedev/ostree/issues/3022


dbnicholson (approved):
Hilarious. Fix looks good to me.

PR #3244, opened 2024-05-15
Fix _ostree_ensure_fsverity reporting of supports in early exit

alexlarsson:
If supported_out is passed to _ostree_ensure_fsverity and we successfully exit early, for example because the file is a symlink, then *supported_out is not initialized.

This is problematic in the case of ostree_sysroot_update_post_copy(), because it passes in an uninitialized supported, and on successfull return of _ostree_ensure_fsverity() it assumes that it is iniialized.

In case supported happened to be initialized to non-zero it will take this branch:

      if (!supported)
        break; /* If not supported, skip rest */

Which means *all* further objects will not get fs-verity enabled.


cgwalters:
Ug, another one rust would have caught.

This looks sane to me.

cgwalters (approved):

PR #3245, opened 2024-05-16
docs: Describe offline updates with static deltas

cgwalters:
This one isn't immediately obvious that it's possible.

### DIFF
--- a/docs/copying-deltas.md
+++ b/docs/copying-deltas.md
@@ -1,0 +1,32 @@
+---
+nav_order: 101
+---
+
+# Static deltas for offline updates
+{: .no_toc }
+
+1. TOC
+{:toc}
+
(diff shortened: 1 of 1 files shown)


docs/copying-deltas.md
```diff
@@ -0,0 +1,32 @@
...
+<!-- SPDX-License-Identifier: (CC-BY-SA-3.0 OR GFDL-1.3-or-later) -->
+
+OSTree supports generating "self-contained" static delta files, via an
+invocation similar to the following:
+
+```
+$ ostree --repo=/path/to/repo static-delta generate --min-fallback-size=0 --filename=delta-update-file --from=<from> <to>
+```
+
+Note the usage of `--min-fallback-size=0` to ensure that the delta is self-cotnained.
```
jlebon, line 20:
```suggestion
Note the usage of `--min-fallback-size=0` to ensure that the delta is self-contained.
```

jlebon (approved):

PR #3249, opened 2024-05-21
docs: add webOS as users of libostree

ericcurtin:
Updated the README.md to include information about webOS, an OS for smart devices such as smart TVs, which from version 2.0.0 supports Firmware-Over-the-Air (FOTA) based on libostree. Added relevant links for further details on webOS and FOTA setup.


cgwalters (approved):

cgwalters:
Generally my preference actually is that these PRs come from the project/company maintaining it, but...

PR #3252, opened 2024-05-22
checkout: Add API to directly checkout composefs

cgwalters:
We were missing the simple, obvious API and CLI to go from ostree commit -> composefs.

Internally, we had `ostree_repo_checkout_composefs` with the right "shape" mostly, except it had more code in the deploy path to turn that into a composefs.

Add a straightforward public API that does what
the deploy code did before, and then the old
API becomes an explicitly internal helper with an `_` prefix.

Goals:

- Lead towards a composefs-oriented future
- This makes the composefs logic more testable directly


cgwalters:
It's hard to overstate the extent to which we just really really need to do a hard switch to composefs, it will make so many things better.

ericcurtin (approved):

PR #3263, opened 2024-06-12
2023.8-3 coverity minor fixes

lukewarmtemp:
checksum: Null terminate buffer
    
Coverity points out that `buffer` is an unterminated string.
g_checksum_update() expects a null-terminated string.

---

tree: autofree `name` var
    
Coverity points out that we have a memory leak from `g_strdup(name)

---

commit: Null terminate `target_buf` var
    
Coverity points out that we are passing an unterminated string to
sprintf().

---

repo: Return copy of local var `ret`
    
Coverity points out that we are returning `stbuf` which is a local
variable.

---

repo: Free actual pointer instead of stolen pointer
    
Coverity points out that ostree_repo_finder_result_free() is freeing
an incorrect pointer `(gpointer)g_steal_pointer(&results->pdata[i])`.
Since we are not returning this value, we shouldn't need to transfer
ownership of the pointer. Instead, we should free the actual pointer.

---

repo: Autofree `ret_rollsum`
    
Coverity points out that we have a memory leak from `ret_rollsum`.

---

repo: Autofree `dir_or_file_path`
    
Coverity points out that we have a memory leak from
g_strdup(dir_or_file_path).

---

repo: Autofree contents of `input_mfile`
    
Coverity points out that we have memory leak from
g_mapped_file_get_contents(input_mfile).

---

sysroot: Autofree `devpath`
    
Coverity points out that we have a memory leak from g_strdup(devpath).

---

prepare: Change mount `etc` to absolute path
    
Coverity points out that ""/sysroot.tmp/etc"" could be a copy-paste
error.

### DIFF
--- a/src/libostree/ostree-checksum-input-stream.c
+++ b/src/libostree/ostree-checksum-input-stream.c
@@ -132,7 +132,11 @@ ostree_checksum_input_stream_read (GInputStream *stream, void *buffer, gsize cou
 
   res = g_input_stream_read (fself->base_stream, buffer, count, cancellable, error);
   if (res > 0)
-    g_checksum_update (self->priv->checksum, buffer, res);
+    {
+      guchar *char_buffer = (guchar *)buffer;
+      char_buffer[res] = '\0';
+      g_checksum_update (self->priv->checksum, (const guchar *)char_buffer, res);
+    }
 
   return res;
 }
--- a/src/libostree/ostree-mutable-tree.c
+++ b/src/libostree/ostree-mutable-tree.c
@@ -452,7 +452,8 @@ ostree_mutable_tree_ensure_parent_dirs (OstreeMutableTree *self, GPtrArray *spli
           invalidate_contents_checksum (subdir);
           next = ostree_mutable_tree_new ();
           ostree_mutable_tree_set_metadata_checksum (next, metadata_checksum);
-          insert_child_mtree (subdir, g_strdup (name), next);
+          g_autofree char *name_copy = g_strdup (name);
+          insert_child_mtree (subdir, name_copy, next);
         }
 
       subdir = next;
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -815,6 +815,7 @@ _try_clone_from_payload_link (OstreeRepo *self, OstreeRepo *dest_repo, const cha
       if (size < OSTREE_SHA256_STRING_LEN + _OSTREE_PAYLOAD_LINK_PREFIX_LEN)
         return glnx_throw (error, "invalid data size for %s", loose_path_buf);
 
+      target_buf[size] = '\0';
       sprintf (target_checksum, "%.2s%.62s", target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN,
                target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN + 3);
 
--- a/src/libostree/ostree-repo-libarchive.c
+++ b/src/libostree/ostree-repo-libarchive.c
@@ -244,7 +244,7 @@ aic_get_final_path (OstreeRepoArchiveImportContext *ctx, const char *path, GErro
       char *ret = ctx->opts->translate_pathname (ctx->repo, &stbuf, path,
                                                  ctx->opts->translate_pathname_user_data);
       if (ret)
-        return ret;
+        return g_strdup (ret);
       /* Fall through */
     }
   else if (ctx->opts->use_ostree_convention)
--- a/src/libostree/ostree-repo-pull.c
+++ b/src/libostree/ostree-repo-pull.c
@@ -5952,7 +5952,7 @@ find_remotes_cb (GObject *obj, GAsyncResult *async_result, gpointer user_data)
             {
               g_debug ("%s: Omitting remote ‘%s’ from results as none of its refs are new enough.",
                        G_STRFUNC, result->remote->name);
-              ostree_repo_finder_result_free (g_steal_pointer (&g_ptr_array_index (results, i)));
+              ostree_repo_finder_result_free (g_ptr_array_index (results, i));
               continue;
             }
         }
--- a/src/libostree/ostree-repo-static-delta-compilation.c
+++ b/src/libostree/ostree-repo-static-delta-compilation.c
@@ -608,7 +608,7 @@ try_content_rollsum (OstreeRepo *repo, DeltaOpts opts, const char *from, const c
                   (unsigned long long)matches->match_size);
     }
 
-  ContentRollsum *ret_rollsum = g_new0 (ContentRollsum, 1);
+  g_autofree ContentRollsum *ret_rollsum = g_new0 (ContentRollsum, 1);
   ret_rollsum->from_checksum = g_strdup (from);
   ret_rollsum->matches = g_steal_pointer (&matches);
   ot_transfer_out_value (out_rollsum, &ret_rollsum);
--- a/src/libostree/ostree-repo-static-delta-core.c
+++ b/src/libostree/ostree-repo-static-delta-core.c
@@ -409,7 +409,8 @@ ostree_repo_static_delta_execute_offline_with_signature (OstreeRepo *self, GFile
         return glnx_throw_errno_prefix (error, "openat(O_DIRECTORY)");
       else
         {
-          g_autofree char *dir = dirname (g_strdup (dir_or_file_path));
+          g_autofree char *dir_or_file_path_copy = g_strdup (dir_or_file_path);
+          g_autofree char *dir = dirname (dir_or_file_path_copy);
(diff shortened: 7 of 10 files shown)


src/switchroot/ostree-prepare-root.c
```diff
@@ -570,7 +570,7 @@ main (int argc, char *argv[])
       else
         {
           /* Bind-mount /etc (at deploy path), and remount as writable. */
-          if (mount ("etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_SILENT, NULL) < 0)
+          if (mount ("/etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_SILENT, NULL) < 0)
```
lukewarmtemp, line 573: I'm not sure if this will actually resolve the coverity scan message.

ericcurtin: This is a bug, coverity is wrong here

ericcurtin: Will need to change this back.

If we insist on shutting up coverity here, this might work:

`#define TMP_SYSROOT_ETC TMP_SYSROOT "/etc"`

or

`static const char *tmp_sysroot_etc = TMP_SYSROOT "/etc"`

then just use that read-only global variable everywhere instead. The "static const char *" approach is probably slightly better, it's typesafe, but both are fine.

ericcurtin:
I worry about static analysis suggestions sometimes, we better thoroughly review these.

src/libostree/ostree-checksum-input-stream.c
```diff
@@ -132,7 +132,11 @@ ostree_checksum_input_stream_read (GInputStream *stream, void *buffer, gsize cou
 
   res = g_input_stream_read (fself->base_stream, buffer, count, cancellable, error);
   if (res > 0)
-    g_checksum_update (self->priv->checksum, buffer, res);
+    {
+      guchar *char_buffer = (guchar *)buffer;
+      char_buffer[res] = '\0';
+      g_checksum_update (self->priv->checksum, (const guchar *)char_buffer, res);
```
dbnicholson, line 138: This can't be right. Data is being read from a stream to a buffer and being checksummed. We can't just arbitrarily add a NUL to it and expect to get the right checksum. I don't know why coverity thinks this is a string that needs terminating.

src/libostree/ostree-mutable-tree.c
```diff
@@ -452,7 +452,8 @@ ostree_mutable_tree_ensure_parent_dirs (OstreeMutableTree *self, GPtrArray *spli
           invalidate_contents_checksum (subdir);
           next = ostree_mutable_tree_new ();
           ostree_mutable_tree_set_metadata_checksum (next, metadata_checksum);
-          insert_child_mtree (subdir, g_strdup (name), next);
+          g_autofree char *name_copy = g_strdup (name);
+          insert_child_mtree (subdir, name_copy, next);
```
dbnicholson, line 456: That is indeed a leak. However, `insert_child_mtree` takes a `const char *` and dupes it. So, this is just extra allocation and instead `name` can be passed directly to `insert_child_mtree`.

src/libostree/ostree-repo-commit.c
```diff
@@ -815,6 +815,7 @@ _try_clone_from_payload_link (OstreeRepo *self, OstreeRepo *dest_repo, const cha
       if (size < OSTREE_SHA256_STRING_LEN + _OSTREE_PAYLOAD_LINK_PREFIX_LEN)
         return glnx_throw (error, "invalid data size for %s", loose_path_buf);
 
+      target_buf[size] = '\0';
       sprintf (target_checksum, "%.2s%.62s", target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN,
                target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN + 3);
```
dbnicholson, line 820: I think this is pretty careful about the lengths being used, but it doesn't hurt.

src/libostree/ostree-repo-libarchive.c
```diff
@@ -244,7 +244,7 @@ aic_get_final_path (OstreeRepoArchiveImportContext *ctx, const char *path, GErro
       char *ret = ctx->opts->translate_pathname (ctx->repo, &stbuf, path,
                                                  ctx->opts->translate_pathname_user_data);
       if (ret)
-        return ret;
+        return g_strdup (ret);
```
dbnicholson, line 247: Won't this just introduce a leak? `translate_pathname` is defined as returning a `char *` that the caller takes ownership of, so I think this was right. Maybe coverity is confused by the conditional part?

src/libostree/ostree-repo-pull.c
```diff
@@ -5952,7 +5952,7 @@ find_remotes_cb (GObject *obj, GAsyncResult *async_result, gpointer user_data)
             {
               g_debug ("%s: Omitting remote ‘%s’ from results as none of its refs are new enough.",
                        G_STRFUNC, result->remote->name);
-              ostree_repo_finder_result_free (g_steal_pointer (&g_ptr_array_index (results, i)));
+              ostree_repo_finder_result_free (g_ptr_array_index (results, i));
```
dbnicholson, line 5955: Coverity is confused here. [g_steal_pointer](https://docs.gtk.org/glib/func.steal_pointer.html) sets pointer it was passed to `NULL` and returns the pointed to value. So, the actual pointer is being returned. However, the address of the pointer is being set to `NULL`. That's actually important since the `results` pointer array is declared to free all the values in the array when the pointer array itself is freed. Without stealing the pointer, it will result in a double free.

src/libostree/ostree-repo-static-delta-compilation.c
```diff
@@ -608,7 +608,7 @@ try_content_rollsum (OstreeRepo *repo, DeltaOpts opts, const char *from, const c
                   (unsigned long long)matches->match_size);
     }
 
-  ContentRollsum *ret_rollsum = g_new0 (ContentRollsum, 1);
+  g_autofree ContentRollsum *ret_rollsum = g_new0 (ContentRollsum, 1);
   ret_rollsum->from_checksum = g_strdup (from);
   ret_rollsum->matches = g_steal_pointer (&matches);
   ot_transfer_out_value (out_rollsum, &ret_rollsum);
```
dbnicholson, line 614: That's not really going to work as `g_autofree` is only going to free the structure without the contents. If you really wanted to do this, you'd define a `g_autoptr` cleanup function pointing to `content_rollsums_free`. However, this almost certainly isn't needed since the allocation is transferred to the caller in `ot_transfer_out_value`.

It is good practice to declare a `g_autoptr` and then `NULL` the pointer before transferring it to the caller so that you can be sure that any early returns free the memory. However, just sticking in `g_autofree` isn't going to suffice here.

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -409,7 +409,8 @@ ostree_repo_static_delta_execute_offline_with_signature (OstreeRepo *self, GFile
         return glnx_throw_errno_prefix (error, "openat(O_DIRECTORY)");
       else
         {
-          g_autofree char *dir = dirname (g_strdup (dir_or_file_path));
+          g_autofree char *dir_or_file_path_copy = g_strdup (dir_or_file_path);
+          g_autofree char *dir = dirname (dir_or_file_path_copy);
```
dbnicholson, line 413: [dirname](https://man7.org/linux/man-pages/man3/dirname.3.html) doesn't make a copy of the passed in string, so this is almost certainly going to be a double free with the second `g_autofree`. Since the string is already being copied, it would probably me more straightforward to use [g_path_get_dirname](https://docs.gtk.org/glib/func.path_get_dirname.html) like the `g_path_get_basename` right below this. That API does what you'd expect by not modifying the string in place and coverity should be able to understand it.

src/libostree/ostree-sysroot.c
```diff
@@ -2243,7 +2243,8 @@ ostree_sysroot_deployment_unlock (OstreeSysroot *self, OstreeDeployment *deploym
                         deployment, _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_DEVELOPMENT)
                   : _ostree_sysroot_get_runstate_path (
                         deployment, _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_TRANSIENT);
-        g_autofree char *devpath_parent = dirname (g_strdup (devpath));
+        g_autofree char *devpath_copy = g_strdup (devpath);
+        g_autofree char *devpath_parent = dirname (devpath_copy);
```
dbnicholson, line 2247: I don't think that's really a leak like in the previous patch. However, I think it's clearer to just use `g_path_get_dirname` since it acts the way you expect instead of modifying the string in place like `dirname` does.

lukewarmtemp:
PR containing only valid coverity scans: https://github.com/ostreedev/ostree/pull/3265

PR #3265, opened 2024-06-13
2023.8-3 coverity scan

lukewarmtemp:
tree: Fix `name` memory leak
    
Coverity points out that we have a memory leak from `g_strdup(name)`.
`insert_child_mtree()` takes a const char * and duplicates it.
`name` can be passed directly to `insert_child_mtree()`.

---

commit: Null terminate `target_buf` var
    
Coverity points out that we are passing an unterminated string to
sprintf().

---

repo: Fix `dir_or_file_path` memory leak
    
Coverity points out that we have a memory leak from
`g_strdup(dir_or_file_path)`. We use `g_path_get_dirname()` which
modifies the string in place, to which coverity should understand.

---

prepare: Create global var for tmp_sysroot_etc
    
Coverity points out that ""/sysroot.tmp/etc"" could be a copy-paste
error. This is mistake from coverity, but to supress the warning,
we create a global var, tmp_sysroot_etc, which replaces all
instances of TMP_SYSROOT "/etc".

### DIFF
--- a/src/libostree/ostree-repo-static-delta-core.c
+++ b/src/libostree/ostree-repo-static-delta-core.c
@@ -409,7 +409,8 @@ ostree_repo_static_delta_execute_offline_with_signature (OstreeRepo *self, GFile
         return glnx_throw_errno_prefix (error, "openat(O_DIRECTORY)");
       else
         {
-          g_autofree char *dir = dirname (g_strdup (dir_or_file_path));
+          g_autofree char *tmpbuf = g_strdup (dir_or_file_path);
+          const char *dir = dirname (tmpbuf);
           basename = g_path_get_basename (dir_or_file_path);
 
           if (!glnx_opendirat (AT_FDCWD, dir, TRUE, &dfd, error))
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -537,6 +537,7 @@ main (int argc, char *argv[])
                                                 &etc_transient, &error))
         errx (EXIT_FAILURE, "Failed to parse etc.transient value: %s", error->message);
 
+      static const char *tmp_sysroot_etc = TMP_SYSROOT "/etc";
       if (etc_transient)
         {
           char *ovldir = "/run/ostree/transient-etc";
@@ -564,15 +565,15 @@ main (int argc, char *argv[])
 
           g_autofree char *ovl_options
               = g_strdup_printf ("lowerdir=%s,upperdir=%s,workdir=%s", lowerdir, upperdir, workdir);
-          if (mount ("overlay", TMP_SYSROOT "/etc", "overlay", MS_SILENT, ovl_options) < 0)
+          if (mount ("overlay", tmp_sysroot_etc, "overlay", MS_SILENT, ovl_options) < 0)
             err (EXIT_FAILURE, "failed to mount transient etc overlayfs");
         }
       else
         {
           /* Bind-mount /etc (at deploy path), and remount as writable. */
-          if (mount ("etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_SILENT, NULL) < 0)
+          if (mount ("etc", tmp_sysroot_etc, NULL, MS_BIND | MS_SILENT, NULL) < 0)
             err (EXIT_FAILURE, "failed to prepare /etc bind-mount at /sysroot.tmp/etc");
-          if (mount (TMP_SYSROOT "/etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_REMOUNT | MS_SILENT,
+          if (mount (tmp_sysroot_etc, tmp_sysroot_etc, NULL, MS_BIND | MS_REMOUNT | MS_SILENT,
                      NULL)
               < 0)
             err (EXIT_FAILURE, "failed to make writable /etc bind-mount at /sysroot.tmp/etc");
--- a/src/libostree/ostree-mutable-tree.c
+++ b/src/libostree/ostree-mutable-tree.c
@@ -452,7 +452,7 @@ ostree_mutable_tree_ensure_parent_dirs (OstreeMutableTree *self, GPtrArray *spli
           invalidate_contents_checksum (subdir);
           next = ostree_mutable_tree_new ();
           ostree_mutable_tree_set_metadata_checksum (next, metadata_checksum);
-          insert_child_mtree (subdir, g_strdup (name), next);
+          insert_child_mtree (subdir, name, next);
         }
 
       subdir = next;
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -815,8 +815,8 @@ _try_clone_from_payload_link (OstreeRepo *self, OstreeRepo *dest_repo, const cha
       if (size < OSTREE_SHA256_STRING_LEN + _OSTREE_PAYLOAD_LINK_PREFIX_LEN)
         return glnx_throw (error, "invalid data size for %s", loose_path_buf);
 
-      sprintf (target_checksum, "%.2s%.62s", target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN,
-               target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN + 3);
+      snprintf (target_checksum, size, "%.2s%.62s", target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN,
+                target_buf + _OSTREE_PAYLOAD_LINK_PREFIX_LEN + 3);
 
       _ostree_loose_path (loose_path_target_buf, target_checksum, OSTREE_OBJECT_TYPE_FILE,
                           self->mode);


src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -409,7 +409,7 @@ ostree_repo_static_delta_execute_offline_with_signature (OstreeRepo *self, GFile
         return glnx_throw_errno_prefix (error, "openat(O_DIRECTORY)");
       else
         {
-          g_autofree char *dir = dirname (g_strdup (dir_or_file_path));
+          g_autofree char *dir = g_path_get_dirname (g_strdup (dir_or_file_path));
```
dbnicholson, line 412: In this case, you don't need to duplicate `dir_or_file_path`. `g_path_dirname` takes care of it. This is essentially copying the buffer twice with the first copy being leaked.

src/libostree/ostree-sysroot.c
```diff
@@ -2243,7 +2243,7 @@ ostree_sysroot_deployment_unlock (OstreeSysroot *self, OstreeDeployment *deploym
                         deployment, _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_DEVELOPMENT)
                   : _ostree_sysroot_get_runstate_path (
                         deployment, _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_TRANSIENT);
-        g_autofree char *devpath_parent = dirname (g_strdup (devpath));
+        g_autofree char *devpath_parent = g_path_get_dirname (g_strdup (devpath));
```
dbnicholson, line 2246: Same as previous `g_path_get_dirname` comment.

dbnicholson (changes requested):
This seems more reasonable, but now there are lots of failing tests. For instance, one of the tests fails like this:

```
++ ostree --repo=ostree-srv/gnomerepo static-delta generate main
modified: 1
new reachable: metadata=3 content regular=2 symlink=0
rollsum for 0/1 modified
processing bsdiff: [0/1]
fallback for c5efd04e28fd1f9ae16cd119bb79766cdf042b3f1ef89a282f0b3ca68c062b8c (21.9 MB)
part 1 n:3 compressed:585 uncompressed:524
uncompressed=524 compressed=585 loose=484
rollsum=0 objects, 0 bytes
bsdiff=1 objects
++ ostree --repo=ostree-srv/gnomerepo summary -u
++ cd /test-tmp/tap-test.KzJhuj
++ ostree --repo=repo pull origin main
/__w/ostree/ostree/ci-build/../tests/pull-test.sh: line 573: 37143 Segmentation fault      (core dumped) ${CMD_PREFIX} ostree --repo=repo pull origin main
```

If I had to guess, this is from the change in `ostree_repo_static_delta_execute_offline_with_signature`. Possibly `g_path_get_dirname` has slightly different semantics than `dirname`.

src/switchroot/ostree-prepare-root.c
```diff
@@ -570,9 +570,10 @@ main (int argc, char *argv[])
       else
         {
           /* Bind-mount /etc (at deploy path), and remount as writable. */
-          if (mount ("etc", TMP_SYSROOT "/etc", NULL, MS_BIND | MS_SILENT, NULL) < 0)
+          static const char *tmp_sysroot_etc = TMP_SYSROOT "/etc";
```
ericcurtin, line 573: There's another occurrence of TMP_SYSROOT "/etc" a few lines up if you want to catch that too.

src/libostree/ostree-repo-static-delta-processing.c
```diff
@@ -423,8 +423,9 @@ dispatch_bspatch (OstreeRepo *repo, StaticDeltaExecutionState *state, GCancellab
       struct bspatch_stream stream;
       stream.read = bspatch_read;
       stream.opaque = &opaque;
-      if (bspatch ((const guint8 *)g_mapped_file_get_contents (input_mfile),
-                   g_mapped_file_get_length (input_mfile), buf, state->content_size, &stream)
+      g_autofree guint8 *old = (guint8 *)g_mapped_file_get_contents (input_mfile);
```
cgwalters, line 426: This is definitely wrong; the whole point of `GMappedFile` aka `mmap()` on Unix is that it's a view into memory owned by the kernel. It'd be useful to see the Coverity warning here, but regardless we need to skip it.

Really `g_mapped_file_get_contents` should have returned a `const char *`...but it'd be an API break now.

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -409,7 +409,15 @@ ostree_repo_static_delta_execute_offline_with_signature (OstreeRepo *self, GFile
         return glnx_throw_errno_prefix (error, "openat(O_DIRECTORY)");
       else
         {
-          g_autofree char *dir = dirname (g_strdup (dir_or_file_path));
+          g_autofree char *dir = "";
```
cgwalters, line 412: Initializing to `""` is wrong - that's a static buffer and we must not call `free()` on it. Set to `NULL` instead.

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -409,7 +409,15 @@ ostree_repo_static_delta_execute_offline_with_signature (OstreeRepo *self, GFile
...
+          if (dir_or_file_path[strlen (dir_or_file_path) - 1] == G_DIR_SEPARATOR)
```
cgwalters, line 413: Hold on, now our logic is getting more complicated here...why the conditional? Is this because `g_path_get_dirname()` has different semantics than `dirname`?

src/libostree/ostree-repo-static-delta-core.c
```diff
@@ -409,7 +409,15 @@ ostree_repo_static_delta_execute_offline_with_signature (OstreeRepo *self, GFile
...
+            {
+              g_autofree char *modifed_dir = g_strdup (dir_or_file_path);
+              modifed_dir[strlen (modifed_dir) - 1] = '\0';
+              dir = g_path_get_dirname (modifed_dir);
```
cgwalters, line 417: BTW you say in your commit message: 
> We use `g_path_get_dirname()` which modifies the string in place, to which coverity should understand.

But AFACIS that's wrong, it's actually `dirname` that modifies in place, not `g_path_get_dirname`.

src/libostree/ostree-repo-static-delta-core.c
cgwalters: Can we go back to the top here...I don't think the original code was wrong. But I think what we should do is simply:

```
g_autofree char *tmpbuf = g_strdup (dir_or_file_path);
const char *dir = dirname (tmpbuf);
```

Coverity should understand that (if it doesn't, it's arguably broken) and the intent is clearer.

src/libostree/ostree-sysroot.c
```diff
@@ -2240,10 +2240,19 @@ ostree_sysroot_deployment_unlock (OstreeSysroot *self, OstreeDeployment *deploym
...
             = unlocked_state == OSTREE_DEPLOYMENT_UNLOCKED_DEVELOPMENT
                   ? _ostree_sysroot_get_runstate_path (
-                        deployment, _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_DEVELOPMENT)
+                      deployment, _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_DEVELOPMENT)
                   : _ostree_sysroot_get_runstate_path (
-                        deployment, _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_TRANSIENT);
-        g_autofree char *devpath_parent = dirname (g_strdup (devpath));
+                      deployment, _OSTREE_SYSROOT_DEPLOYMENT_RUNSTATE_FLAG_TRANSIENT);
+
+        g_autofree char *devpath_parent = "";
```
cgwalters, line 2247: Same comment here.

cgwalters (changes requested):

src/libostree/ostree-repo-static-delta-core.c
lukewarmtemp: > Is this because g_path_get_dirname() has different semantics than dirname?

Yeah. `dirname` will ignore `/` at the end of a filepath while `g_path_get_dirname` will not. For example, `dirname("test/path/example/")` will return `"test/path"` while `g_path_get_dirname("test/path/example/")` will return `test/path/example`

src/libostree/ostree-repo-static-delta-core.c
lukewarmtemp: Thoughts on a previous review from @dbnicholson? https://github.com/ostreedev/ostree/pull/3263#discussion_r1638435360. Your suggested change is similar to my old one, but I don't have the second `g_autofree`, so will it solve the concern mentioned in the previous review?

cgwalters: > Your suggested change is similar to my old one, but I don't have the second g_autofree, so will it solve the concern mentioned in the previous review?

Yes, not having the second `g_autofree` solves the concern.

Now, using `g_path_get_dirname()` sounded nice to me at the start, but being surprised by semantic differences is less nice.

So I definitely lean towards my suggestion above instead of `g_path_get_dirname` now.

cgwalters (approved):

PR #3266, opened 2024-06-18
remount: ignore ENOENT error during SELinux relabeling

ericcurtin:
Ignore ENOENT error in selinux_restorecon to avoid failures when temporary files created by systemd-sysusers in /etc are missing during relabeling. This prevents errors such as:

  "Failed to relabel /etc/.#gshadowJzu4Rx: No such file or directory"

and allows the process to continue.

### DIFF
--- a/src/boot/ostree-remount.service
+++ b/src/boot/ostree-remount.service
@@ -25,7 +25,7 @@ After=-.mount var.mount
 After=systemd-remount-fs.service
 # But we run *before* most other core bootup services that need write access to /etc and /var
 Before=local-fs.target umount.target
-Before=systemd-random-seed.service plymouth-read-write.service systemd-journal-flush.service
+Before=systemd-random-seed.service plymouth-read-write.service systemd-journal-flush.service systemd-sysusers.service
 Before=systemd-tmpfiles-setup.service systemd-rfkill.service systemd-rfkill.socket
 
 [Service]
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -90,8 +90,18 @@ static void
 relabel_dir_for_upper (const char *upper_path, const char *real_path, gboolean is_dir)
 {
 #ifdef HAVE_SELINUX
+  /* Ignore ENOENT, because if there is no file to relabel we can continue,
+   * systemd-sysusers runs in parallel and can create temporary files in /etc
+   * causing failures like:
+   * "Failed to relabel /etc/.#gshadowJzu4Rx: No such file or directory"
+   */
   if (selinux_restorecon (real_path, 0))
-    err (EXIT_FAILURE, "Failed to relabel %s", real_path);
+    {
+      if (errno == ENOENT)
+        return;
+
+      err (EXIT_FAILURE, "Failed to relabel %s", real_path);
+    }
 
   if (!is_dir)
     return;


src/switchroot/ostree-remount.c
```diff
@@ -90,7 +90,12 @@ static void
...
 {
 #ifdef HAVE_SELINUX
-  if (selinux_restorecon (real_path, 0))
+  /* Ignore ENOENT, because if there is no file to relabel we can continue,
+   * systemd-sysusers runs in parallel and can create temporary files in /etc
+   * causing failures like:
+   * "Failed to relabel /etc/.g#shadowJzu4Rx: No such file or directory"
+   */
+  if (selinux_restorecon (real_path, 0) && errno != ENOENT)
     err (EXIT_FAILURE, "Failed to relabel %s", real_path);
```
alexlarsson, line 99: I think we should do:
if (...) {
 if (errno = ENOENT)
   return; 
  err(...);
}
That way we also handle ENOENT on directories right (would otherwise fail below).-

alexlarsson:
lgtm

cgwalters (approved):
Hmm, so we'd then rely on any process doing concurrent writes to have initialized the labels on its own?

I don't think that can really work longer term because it'd still be racy on things like the parent directory, right?

ISTM we need to move this logic out of `ostree-remount.service` into something that happens basically *immediately* after the systemd policy load.

Ah, digging in a bit I see there's `/run/systemd/relabel-extra.d`, I think all we need to do here is synthesize a variant of that in the transient-etc case that contains simply `/etc`.

I think that'd be equally easy and much more reliable, but I'm approving this PR as is as I think it will *mostly* work too.

alexlarsson:
/run/systemd/relabel-extra.d isn't good enough as-is, as doing so would cause a massive relabel of all of /etc, instead of the smarter approach of just relabeling the very few files that have a file in the upper.

ericcurtin:
Made a very minor change, there was a small typo in the comment

cgwalters:
> /run/systemd/relabel-extra.d isn't good enough as-is, as doing so would cause a massive relabel of all of /etc, instead of the smarter approach of just relabeling the very few files that have a file in the upper.

OK, right. So to fix this we'd need to extend `relabel-extra.d` to support something that separates "physical vs logical root" to have e.g. `/run/ostree/transient-etc /etc` (relabel that dir as if it was `/etc`.

That said, the other fix here is to also add `Before=systemd-sysusers.service` - we already have a big pile of those.

Actually, I think we could split *this* logic out into something like its own distinct `ostree-relabel-etc.service` that could run *very* early in the real root, literally just `After=etc.mount` and `Before=systemd-remount-fs.service` even.

alexlarsson:
> Actually, I think we could split _this_ logic out into something like its own distinct `ostree-relabel-etc.service` that could run _very_ early in the real root, literally just `After=etc.mount` and `Before=systemd-remount-fs.service` even.

Yeah, thats probably a good idea.

cgwalters:
Tangentially related, we don't have CI coverage for the transient-etc case in this repo. Something on the list.

ericcurtin:
I'll work on this is a follow on PR:

ostree-relabel-etc.service

ericcurtin:
Shouldn't be so tricky

ericcurtin:
Although I can also put a Before=systemd-remount-fs.service in this PR also if it is requested

cgwalters:
> Although I can also put a Before=systemd-remount-fs.service in this PR also if it is requested

No, in this PR it needs to be `Before=systemd-sysusers.service`

ericcurtin:
> > Although I can also put a Before=systemd-remount-fs.service in this PR also if it is requested
> 
> No, in this PR it needs to be `Before=systemd-sysusers.service`

Pushing, sorry I meant that one, copy pasted the wrong one in here, re-pushing in seconds

src/boot/ostree-remount.service
```diff
@@ -25,7 +25,7 @@ After=-.mount var.mount
 After=systemd-remount-fs.service
 # But we run *before* most other core bootup services that need write access to /etc and /var
 Before=local-fs.target umount.target
-Before=systemd-random-seed.service plymouth-read-write.service systemd-journal-flush.service
+Before=systemd-random-seed.service plymouth-read-write.service systemd-journal-flush.service systemd-sysusers.service
```
cgwalters, line 28: Hmm, this actually created an ordering cycle it looks like; my bad for not digging into this. I merged over red because we had other unrelated tests that were broken.

PR #3267, opened 2024-06-19
relabel-etc: Split remount into, remount and relabel-etc

ericcurtin:
Split these into two different binaries we two different resonsibilities. They don't share so much functionality.


ericcurtin:
Incomplete and untested right now

src/boot/ostree-relabel-etc.service
```diff
@@ -0,0 +1,40 @@
...
+ConditionKernelCommandLine=ostree
+OnFailure=emergency.target
+Conflicts=umount.target
+# Run after core mounts
+After=-.mount var.mount
+After=systemd-remount-fs.service
+# But we run *before* most other core bootup services that need write access to /etc and /var
+Before=local-fs.target umount.target
+Before=systemd-random-seed.service plymouth-read-write.service systemd-journal-flush.service systemd-sysusers.service
+Before=systemd-tmpfiles-setup.service systemd-rfkill.service systemd-rfkill.socket
```
cgwalters, line 29: We can drop all this stuff, as all of that is `After=systemd-remount-fs.service`.

But I'm fine going belt-and-suspenders here with e.g. `Before=ostree-remount.service` - which already has all those `Before=` itself.

To combine all this I'd suggest:

```
Before=local-fs-pre.target systemd-remount-fs.service
# This one is already itself After=systemd-remount-fs.service, but for redundancy
Before=ostree-remount.service
```

src/switchroot/ostree-relabel-etc.c
```diff
@@ -0,0 +1,189 @@
...
+int
+main (int argc, char *argv[])
+{
+  g_autoptr (GError) error = NULL;
+  g_autoptr (GVariant) ostree_run_metadata_v = NULL;
+  {
+    glnx_autofd int fd = open (OTCORE_RUN_BOOTED, O_RDONLY | O_CLOEXEC);
+    if (fd < 0)
+      {
+        /* We really expect that nowadays that everything is done in the initramfs,
```
cgwalters, line 108: Don't need this copy here.

src/switchroot/ostree-relabel-etc.c
```diff
@@ -0,0 +1,189 @@
...
+  }
+  g_autoptr (GVariantDict) ostree_run_metadata = g_variant_dict_new (ostree_run_metadata_v);
+
+  /* The /sysroot mount needs to be private to avoid having a mount for e.g. /var/cache
+   * also propagate to /sysroot/ostree/deploy/$stateroot/var/cache
+   *
+   * Today systemd remounts / (recursively) as shared, so we're undoing that as early
+   * as possible.  See also a copy of this in ostree-prepare-root.c.
+   */
+  if (mount ("none", "/sysroot", NULL, MS_REC | MS_PRIVATE, NULL) < 0)
```
cgwalters, line 139: Also drop this

src/boot/ostree-relabel-etc.service
```diff
@@ -0,0 +1,40 @@
...
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStart=/usr/lib/ostree/ostree-relabel-etc
```
cgwalters, line 34: Hmm, so far we haven't introduced many new little binaries, but instead we've been tending to dispatch via hidden CLI verbs, see e.g. `ostree admin finalize-staged`. I'd slightly prefer that, i.e. we introduce a new `ostree admin relabel-etc` (which should barf if it's not run under systemd, i.e. we check `INVOCATION_ID` etc).

src/libostree/ostree-impl-system-generator.c
```diff
@@ -109,6 +109,11 @@ require_internal_units (const char *normal_dir, const char *early_dir, const cha
       < 0)
     return glnx_throw_errno_prefix (error, "symlinkat");
 
+  if (symlinkat (SYSTEM_DATA_UNIT_PATH "/ostree-relabel-etc.service", normal_dir_dfd,
```
cgwalters, line 112: I'm OK with this as is, but just a note: I think we can change things so that the initramfs code itself acts as a generator and writes the enablement symlink into `/run` directly then.

As is it's a bit silly because the initramfs writes the file in `/run` which says that transient etc is enabled, then we boot in the real root, run a binary which reads that file and exits if the flag isn't found...but we can bypass all that and be slightly more efficient by just telling systemd whether or not to run the binary.

cgwalters (changes requested):

src/boot/ostree-relabel-etc.service
```diff
@@ -0,0 +1,40 @@
...
+[Unit]
+Description=OSTree Relabel etc
+Documentation=man:ostree(1)
+DefaultDependencies=no
+ConditionKernelCommandLine=ostree
+OnFailure=emergency.target
+Conflicts=umount.target
+# Run after core mounts
+After=-.mount var.mount
+After=systemd-remount-fs.service
```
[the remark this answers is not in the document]
cgwalters: OK I was thinking about this more and I think we're in a situation where actually we tell people using composefs in particular to just mount the rootfs writable in the initramfs, so the whole role of systemd-remount-fs.service is a no-op from our PoV basically.

At least...I thought we enforced that, but now I can't find it.

src/switchroot/ostree-relabel-etc.c
```diff
@@ -0,0 +1,179 @@
+/*
+ * Copyright (C) 2011 Colin Walters <walters@verbum.org>
```
cgwalters, line 2: Tangentially related, I think you can update this to be you and or/alex, or just say (C) Red Hat, Inc.

src/switchroot/ostree-relabel-etc.c
```diff
@@ -0,0 +1,179 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Author: Colin Walters <walters@verbum.org>
```
cgwalters, line 19: I've been dropping this in new code BTW, "git log" is source of truth and I don't think I wrote any of this code.

src/switchroot/ostree-relabel-etc.c
```diff
@@ -0,0 +1,179 @@
...
+ * based on the content of an overlayfs upperdirectory that is in use by the mount.
+ * The goal is that we relabel in the overlay mount all the files that have been
+ * modified (directly or via parent copyup operations) since the overlayfs was
+ * mounted. This will be used for the /etc overlayfs mount where no selinux labels
+ * are set before selinux policy is loaded.
+ */
+static void
+relabel_dir_for_upper (const char *upper_path, const char *real_path, gboolean is_dir)
+{
+#ifdef HAVE_SELINUX
```
cgwalters, line 55: Optional: I think this entire binary could become conditional on SELinux actually, i.e. we don't even compile or ship it if built without selinux support.

jlebon:
Hmm, why aren't we doing the relabeling from the initramfs instead? Current kernels allow relabeling before a policy is loaded. This is notably used by Ignition: https://github.com/coreos/ignition/blob/b2ea35258bf5367a3522a4bea7f71063954b740c/internal/exec/util/selinux.go#L80.

From my experience, doing relabeling from the real root causes a lot of issues and quickly becomes a game of whack-a-race condition.

cgwalters:
Relabeling from the initramfs indeed sounds way better. The only thing I'm a little uncertain about is the semantics for how e.g. `selinux=0` or other flags for configuring it in the real root are handled. Maybe running `setfiles` handles this already?

ericcurtin:
I like the idea of using setfiles or something similar, less code to maintain...

I think selinux=0, etc. is easy to handle in systemd .service file:

ConditionKernelCommandLine=!selinux=0

We may be able to achieve this with zero C code here, just write a well crafted systemd .service file.

Although, I would rather we relabelled from rootfs, even if it meant running this .service file in it's own exclusive time slot.

We put a lot of effort in Automotive for example to make our initramfs super tiny like < 10M .

Ideally we wouldn't relabel on boot at all tbh, it's another thing to do during boot and again we are trying to cut down on doing tasks at boot. Any reason we can't do this at image composition time or shutdown?

/.autorelabel for example seems to perform relabeling on shutdown

cgwalters:
> Ideally we wouldn't relabel on boot at all tbh, it's another thing to do during boot and again we are trying to cut down on doing tasks at boot to make that as fast as possible. Any reason we can't do this at image composition time or shutdown?

Automotive is using transient etc, this is about the in-memory inodes for the overlayfs. There's no real way to shortcut this today, we have a kind of complex interlock between selinux policy loading vs the mount.

However as far as the cost...slimming down what's in `/etc` by default will inherently make this cheaper.

PR #3271, opened 2024-06-25
u-boot: add 'bootdir' to the generated uEnv.txt

ChenQi1989:
When doing a full copy of:

$deployment/usr/lib/ostree-boot -> /boot/ostree/$os-$bootcsum/

U-Boot bootscript can use the 'bootdir' to find, for example, the Device Tree (dtb) file, as in:

load ${dtype} ${disk}:${bootpart} ${a_fdt} ${bootdir}${dtbname}

Or u-boot external bootscript:

load ${dtype} ${disk}:${bootpart} ${a_scr} ${bootdir}${scriptname}

It could also be possible to point 'bootdir' directly to the $deployment/usr/lib/ostree-boot, but this would add unnecessary restrictions on what file system can be used for rootfs as u-boot, for example, can not read from BTRFS. So having
bootdir=/boot/ostree/$os-$bootcsum/ is a better approach here, as /boot can be on a separate partition with its own file system type.


ericcurtin:
@martinezjavier I think you wrote this code, could we borrow your eyes for a quick review?

src/libostree/ostree-bootloader-uboot.c
```diff
@@ -106,6 +106,7 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot *self, int bootver
   g_autoptr (GPtrArray) boot_loader_configs = NULL;
   OstreeBootconfigParser *config;
   const char *val;
+  g_autofree char *bootdir = NULL;
```
ericcurtin, line 109: It might be better if we moved the scope of this variable into the for loop, like the "index_suffix" variable.

martinezjavier:
> @martinezjavier I think you wrote this code, could we borrow your eyes for a quick review?

The patch makes sense to me. I wonder if `kernel_image` and `ramdisk_image` should also use this `bootdir` variable then for consistency.

src/libostree/ostree-bootloader-uboot.c
```diff
@@ -129,6 +130,9 @@ create_config_from_boot_loader_entries (OstreeBootloaderUboot *self, int bootver
         }
       g_ptr_array_add (new_lines, g_strdup_printf ("kernel_image%s=/boot%s", index_suffix, val));
 
+      bootdir = strndup (val, strrchr(val, '/') - val);
```
cgwalters, line 133: Isn't this effectively `dirname()`? We do this across the codebase:

```
          g_autofree char *tmpbuf = g_strdup (dir_or_file_path);
          const char *dir = dirname (tmpbuf);
```

PR #3275, opened 2024-07-10
libostree: Remove compatibility code with GLib < 2.44

ueno:
As the build system has required GLib 2.44 since commit eb09207e1abd7499bd92866cce1de6148d659a4a, the manual expansion of G_DECLARE_INTERFACE and G_DECLARE_FINAL_TYPE is no longer necessary.


cgwalters:
Thanks!
/approve
/ok-to-test

ericcurtin (approved):

PR #3276, opened 2024-07-10
keyfile-utils: Add API to parse tristate strings

cgwalters:
Prep for using this in multiple places. Add unit tests.

### DIFF
--- a/src/libotutil/ot-keyfile-utils.c
+++ b/src/libotutil/ot-keyfile-utils.c
@@ -60,6 +60,51 @@ ot_keyfile_get_boolean_with_default (GKeyFile *keyfile, const char *section, con
   return TRUE;
 }
 
+// Keep this in sync with
+// https://gitlab.gnome.org/GNOME/glib/-/blob/4a73fbda8be6f80f14b05983eb575c1eb1329c2c/glib/gkeyfile.c?page=5#L4585
+// Except for some reason at some point we added "yes" or "no" as possible values too...
+gboolean
+_ostree_parse_boolean (const char *s, gboolean *out_val, GError **error)
+{
+  g_assert (s);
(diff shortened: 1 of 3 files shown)


src/libotutil/ot-keyfile-utils.c
```diff
@@ -60,6 +60,51 @@ ot_keyfile_get_boolean_with_default (GKeyFile *keyfile, const char *section, con
...
+      *out_val = FALSE;
+      return TRUE;
+    }
+  return glnx_throw (error, "Invalid boolean: %s", s);
+}
+
+gboolean
+_ostree_parse_tristate (const char *s, OtTristate *out_tri, GError **error)
+{
+  if (strcmp (s, "maybe") == 0)
```
jlebon, line 88: This isn't new, but `"maybe"` is a very confusing value to support. Feels like `"auto"` would capture it better?

src/libotutil/ot-keyfile-utils.c
```diff
@@ -60,6 +60,51 @@ ot_keyfile_get_boolean_with_default (GKeyFile *keyfile, const char *section, con
...
+  if (_ostree_parse_boolean (s, &bool_value, NULL))
+    {
+      if (bool_value)
+        *out_tri = OT_TRISTATE_YES;
+      else
+        *out_tri = OT_TRISTATE_NO;
+      return TRUE;
+    }
+  // If it's invalid, be clear a tristate was expected.
+  return glnx_throw (error, "Invalid tri-state value: %s", s);
```
jlebon, line 105: Minor: looks like we're losing the `G_KEY_FILE_ERROR_INVALID_VALUE` semantics here.

jlebon (approved):

PR #3277, opened 2024-07-10
prepare-root: Add ostree.prepare-root.composefs

cgwalters:
We have a use case for overriding the composefs state via
the kernel commandline; see e.g.
https://gitlab.com/fedora/bootc/tracker/-/issues/27

### DIFF
--- a/src/libotcore/otcore-prepare-root.c
+++ b/src/libotcore/otcore-prepare-root.c
@@ -24,6 +24,8 @@
 // in use, the ostree commit metadata will contain the composefs image digest,
 // which can be used to fully verify the target filesystem tree.
 #define BINDING_KEYPATH "/etc/ostree/initramfs-root-binding.key"
+// The kernel argument to configure composefs
+#define CMDLINE_KEY_COMPOSEFS "ostree.prepare-root.composefs"
 
 static bool
 proc_cmdline_has_key_starting_with (const char *cmdline, const char *key)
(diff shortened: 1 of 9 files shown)


src/libotcore/otcore-prepare-root.c
```diff
@@ -214,5 +217,23 @@ otcore_load_composefs_config (GKeyFile *config, gboolean load_keys, GError **err
         return glnx_null_throw (error, "public key file specified, but no public keys found");
     }
 
+  g_autofree char *ostree_composefs
+      = otcore_find_proc_cmdline_key (cmdline, "ostree.prepare-root.composefs");
```
jlebon, line 221: Minor: would be good to have that be a top-level define.

src/libotcore/otcore-prepare-root.c
```diff
@@ -214,5 +217,23 @@ otcore_load_composefs_config (GKeyFile *config, gboolean load_keys, GError **err
...
+        {
+          ret->enabled = OT_TRISTATE_YES;
+          ret->is_signed = true;
+        }
+      else
+        {
+          // The other states force off signatures
+          ret->is_signed = false;
+          if (!_ostree_parse_tristate (ostree_composefs, &ret->enabled, error))
+            return FALSE;
```
jlebon, line 234:
```suggestion
            return glnx_prefix_error (error, "handling karg " COMPOSEFS_KARG);
```

jlebon (approved):
Thanks for working on this!

PR #3278, opened 2024-07-12
sign: Support generic "spki" type of commit signatures

ueno:
The current "ed25519" signing type assumes raw Ed25519 key format for
both public and private keys. This patch generalizes it by adding a
new signature type "spki" which uses the X.509 SubjectPublicKeyInfo
format for public keys. Keys in this format can easily be created with
openssl tools and provide crypto agility[1] as the format embeds
algorithm identifier.

The supposed use-case of this feature is to attach multiple signatures
with different algorithms to a single commit, so even if an algorithm
turned vulnerable, the signatures made with other algorithms can still
be used as a fallback. For instance, signer can create an Ed25519
signature along with a quantum-resistent ML-DSA signature.

The following are a couple of implementation notes:

- The private keys shall be stored in the PKCS#8 format, though future
  extensions may support other format such as opaque key handles on a
  hardware token.

- The "spki" signature type prefers the keys to be encoded in the PEM
  format on disk, while it still accepts base64 encoded keys when given
  through the command-line.

1. https://en.wikipedia.org/wiki/Cryptographic_agility


src/libostree/ostree-sign-ed25519.c
```diff
@@ -157,6 +175,8 @@ ostree_sign_ed25519_data (OstreeSign *self, GBytes *data, GBytes **signature,
 
   size_t len;
   if (EVP_DigestSignInit (ctx, NULL, NULL, NULL, pkey)
+      && EVP_DigestSign (ctx, NULL, &len, g_bytes_get_data (data, NULL), g_bytes_get_size (data))
+      && (sig = g_malloc0 (len)) != NULL
```
cgwalters, line 179: Allocating in the middle of a conditional feels weird. How about splitting the conditionals:

```
if (digestsigninit && digestsign)
  {
    g_autofree guchar *sig = g_malloc0 (len);
    if (digestsign)
      ...
  }
```

or so?

src/libostree/ostree-sign-ed25519.c
```diff
@@ -320,22 +335,23 @@ ostree_sign_ed25519_clear_keys (OstreeSign *self, GError **error)
   /* Clear secret key */
   if (sign->secret_key != NULL)
     {
-      memset (sign->secret_key, 0, OSTREE_SIGN_ED25519_SECKEY_SIZE);
-      g_free (sign->secret_key);
+      gsize size;
+      gpointer data = g_bytes_unref_to_data (sign->secret_key, &size);
+      memset (data, 0, size);
```
cgwalters, line 340: (Unrelated to your change, but I think we should probably use `explicit_bzero` here or so)

src/libostree/ostree-sign-ed25519.c
```diff
@@ -222,29 +236,27 @@ ostree_sign_ed25519_data_verify (OstreeSign *self, GBytes *data, GVariant *signa
       g_autoptr (GVariant) child = g_variant_get_child_value (signatures, i);
       g_autoptr (GBytes) signature = g_variant_get_data_as_bytes (child);
 
+#if defined(USE_LIBSODIUM)
```
cgwalters, line 239: This is minor but I think things would be slightly cleaner if we did something at the top like:

```
#define FIXED_KEY_SIZES (defined(USE_LIBSODIUM))
```

And here did `#if FIXED_KEY_SIZES` or so? (Or alternatively invert with `DYNAMIC_KEY_SIZES`...no strong opinion here)

src/libotcore/otcore-ed25519-verify.c
```diff
@@ -64,21 +69,24 @@ otcore_validate_ed25519_signature (GBytes *data, GBytes *public_key, GBytes *sig
...
+  gsize public_key_size;
+  const guint8 *public_key_buf = g_bytes_get_data (public_key, &public_key_size);
 
+  gsize signature_size;
+  const guint8 *signature_buf = g_bytes_get_data (signature, &signature_size);
 #endif
 
 #if defined(HAVE_LIBSODIUM)
+  // Strictly verify pubkey and signature lengths, as libsodium can
+  // only handle raw ed25519 public key and signatures.
```
cgwalters, line 81: Right but...it makes me uncomfortable that we now accept arbitrary sized keys and signatures in the openssl case. Does openssl expose any constants that could cap this?

The only things I can really think of are DoS here, and even those feel hard to reach as ostree caps things at `pub const OSTREE_MAX_METADATA_SIZE: c_int = 134217728;`.

Still though...it'd be nice to have *some* sanity check (actually, we could safely cap these things to MAX_METADATA_SIZE) and that way we wouldn't need to special-case libsodium - it would just be the *size* that would vary.


I think in practice, there'

cgwalters (changes requested):
Thank you so much for this well written patch!

Can you elaborate a bit more on why specficially you chose to work on this now? The rationale makes sense, but...are you planning to actually use Ed448 somewhere?

src/libotcore/otcore-ed25519-verify.c
```diff
@@ -92,16 +100,23 @@ otcore_validate_ed25519_signature (GBytes *data, GBytes *public_key, GBytes *sig
...
-  EVP_PKEY *pkey = EVP_PKEY_new_raw_public_key (EVP_PKEY_ED25519, NULL, public_key_buf,
-                                                OSTREE_SIGN_ED25519_PUBKEY_SIZE);
+
+  // Try SubjectPublicKeyInfo encoded public key first.
+  const unsigned char *p = public_key_buf;
+  EVP_PKEY *pkey = d2i_PUBKEY (NULL, &p, public_key_size);
+
+  // Try raw ed25519 public key if the length matches.
+  if (pkey == NULL && public_key_size == OSTREE_SIGN_ED25519_PUBKEY_SIZE)
+    pkey = EVP_PKEY_new_raw_public_key (EVP_PKEY_ED25519, NULL, public_key_buf, public_key_size);
```
cgwalters, line 110: I'd be curious if *in practice* other key types happen to exactly match the ed25519 size...not saying you should spend a lot of time digging to find out, but if you happen to know offhand, that'd be interesting.

dbnicholson:
I haven't reviewed the code yet, but this is awesome and how it should have been done initially. If you can inspect the key algorithm from a standard format, then there's no reason to specify what type of key it is.

Again, I haven't looked at the changes, but my question now becomes what to do with the "sign types". On the one hand, I kind of want to get rid of it entirely since the only reason to specify the type is to know how to interpret the keys. But if the key type can be automatically determined, then there's no need for that. However, there's also the dummy sign type, which is just used for testing.

I think the way I would handle this is:

* Keep the existing `ed25519` and `dummy` types. There are already people using the raw ed25519 keys and presumably we want to keep the dummy type for testing.
* Add this new functionality in a new type called either `pkcs8` or `auto`. I can't really think of what other form of signing would be desired besides a PKCS8 encoded asymmetric key, but maybe there could be something later. Theoretically, you could move the GPG signing under ostree-sign as a distinct `pgp` type. The presence of the 2 signing infrastructures is awkward, but I suspect this would be a lot of work and require a long migration period.
* Deprecate the `ed25519` type with directions for how to convert the key to `pkcs8`. We'd need to keep the `ed25519` type for quite some time as both signers and clients need to be updated to handle PKCS8 keys.

src/libotcore/otcore-ed25519-verify.c
dbnicholson: I doubt it, but this is one of the reasons I think the PKCS8 handling should be in a separate backend and we should leave the existing raw ed25519 handling alone. Since signers and clients will need to be updated to handle the PKCS8 encoded keys, it might as well be "opt-in" in the form of a new signing type. Then the existing ed25519 signer can be left alone and there's no ambiguity about the key format it expects.

Certainly I think you'd want to have some common code for signing and verification once you've loaded the keys, though.

dbnicholson:
> * Add this new functionality in a new type called either `pkcs8` or `auto`. I can't really think of what other form of signing would be desired besides a PKCS8 encoded asymmetric key, but maybe there could be something later.

I guess _technically_ it's PKCS8 formatted private keys and X.509 SubjectPublicKeyInfo (SPKI) formatted public keys. So, `pkcs8` wouldn't be totally accurate. I can't think of a better name, though.

ueno:
Thank you for the reviews and comments so far!

> Can you elaborate a bit more on why specficially you chose to work on this now? The rationale makes sense, but...are you planning to actually use Ed448 somewhere?

My main motivation here is to prepare for the upcoming PQC [transition](https://research.redhat.com/blog/2024/07/05/fedora-linux-transition-for-quantum-resistant-cryptography/). While encryption (key encapsulation) is the current priority as it affects forward secrecy, signatures will come next, and it would be nice if ostree could support PQC signatures transparently.

src/libotcore/otcore-ed25519-verify.c
ueno: Yes, I added the guard unconditionally.

src/libotcore/otcore-ed25519-verify.c
cgwalters: > I doubt it, but this is one of the reasons I think the PKCS8 handling should be in a separate backend 

Yeah...this makes total sense to me. What do you think @ueno ?

cgwalters:
> My main motivation here is to prepare for the upcoming PQC [transition](https://research.redhat.com/blog/2024/07/05/fedora-linux-transition-for-quantum-resistant-cryptography/).

Got it! Makes sense...I think that's the kind of thing that's good to have in the commit message too.

src/libotcore/otcore-ed25519-verify.c
ueno: I agree; if we go that way, the following might be worth considering:
- Keys are not always restricted to particular signing parameters (e.g., hash in ECDSA); perhaps it might make sense to embed signing parameters in the signature itself in that case
- The current key format (one key per line) is not easy to deal with for longer keys (~2K with ML-DSA); maybe the new backend should support PEM/DER blocks instead (this could require additional API in OsteeSign)

ueno: It's still in progress but I've updated the PR with a new signing backend "x509". As a next step, I'll try to add the PEM based keysfile and make it default for the new backend (if that sounds ok).

ueno:
@cgwalters @dbnicholson There is still room for refactoring, but this is mostly feature complete from my point of view. PQ signatures can be used through oqsprovider in both pure/hybrid schemes:
```console
# Create SPHINCS+ public and secret keys
$ openssl genpkey -provider oqsprovider -algorithm sphincssha2128fsimple -outform PEM -out sphincssha2128fsimple-sk.pem
$ openssl pkey -provider oqsprovider -outform PEM -pubout -in sphincssha2128fsimple-sk.pem -out sphincssha2128fsimple-pk.pem

# Set up an OSTree repo
$ mkdir repo work
$ cd work
$ ../ostree --repo=../repo init

# Make a signed commit
$ echo > a.txt
$ ../ostree --repo=../repo commit -b main -s "Signed with x509 module" --sign-from-file ../sphincssha2128fsimple-sk.pem --sign-type=x509
fac8d8df984644e6b814be093c3a5f60f746d90f1aacea020cd73dc22b2c33ce
$ COMMIT=$(../ostree --repo=../repo rev-parse main)
$ ../ostree --repo=../repo sign --verify --sign-type=x509 --keys-file=../sphincssha2128fsimple-pk.pem $COMMIT
x509: Signature verified successfully with key '302d300806062bce0f06040d0321007aac408c97612b9974785e2fcbcc35bfd4b31f7b95b0fc1026485dc8df66c95b'
```
For hybrid schemes, e.g., Ed25519+ML-KEM (Dilithium), you can simply sign the same commit twice for each algorithm.

src/libostree/ostree-sign-x509.c
```diff
@@ -0,0 +1,651 @@
...
+
+  return ret;
+}
+
+static gboolean
+_load_pk_from_file (OstreeSign *self, const gchar *filename, gboolean trusted, GError **error)
+{
+  g_debug ("Processing file '%s'", filename);
+
+  g_autoptr (GFile) keyfile = NULL;
```
cgwalters, line 522: Here (and a lot of other places) - our new preferred coding style is declare-and-initialize as an atomic unit. In this specific function we clearly can do so.

Not a blocker at all, just a general note.

src/libostree/ostree-blob-reader-pem.c
```diff
@@ -0,0 +1,225 @@
...
+}
+
+enum PemInputState
+{
+  PEM_INPUT_STATE_OUTER,
+  PEM_INPUT_STATE_INNER
+};
+
+#define PEM_SUFFIX "-----"
+#define PEM_PREFIX_BEGIN "-----BEGIN "
```
cgwalters, line 136: OpenSSL really doesn't have something for this? A lot of new string parsing in C is pretty unfortunate.

tests/test-pem.c
```diff
@@ -0,0 +1,70 @@
...
+      expected_bytes = g_bytes_new_static (tests[i].data, tests[i].size);
+      g_assert (g_bytes_equal (bytes, expected_bytes));
+    }
+}
+
+int
+main (int argc, char **argv)
+{
+  g_test_init (&argc, &argv, NULL);
+  g_test_add_func ("/ostree_read_pem_block", test_ostree_read_pem_block);
```
cgwalters, line 68: Thanks for unit testing this! We should also add some corrupted/invalid test cases...this type of code is very fuzzable and we do have a fuzzer wired up here https://github.com/google/oss-fuzz/tree/master/projects/ostree

tests/test-signed-commit-dummy.sh
```diff
@@ -0,0 +1,61 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with this library. If not, see <https://www.gnu.org/licenses/>.
+
+set -euo pipefail
+
+. $(dirname $0)/libtest.sh
+
+echo "1..4"
```
cgwalters, line 24: Fine as is but newer tests we've been using `tap_ok` + `tap_end`

src/libostree/ostree-blob-reader-base64.c
```diff
@@ -0,0 +1,80 @@
...
+      g_propagate_error (error, g_steal_pointer (&local_error));
+      return NULL;
+    }
+
+  if (line == NULL)
+    return NULL;
+
+  gsize n_elements;
+  g_base64_decode_inplace (line, &n_elements);
+  explicit_bzero (line + n_elements, len - n_elements);
```
cgwalters, line 77: This is worthy of the same comment you have in a similar place `/* Don't leak the trailing encoded bytes */` later in the code (maybe also worthy of a shared helper function?).

But also...I don't fully grasp the threat model here...aren't the decoded bytes equally sensitive? In what cases would the trailing data in the malloc buffer be somehow accessible when the full buffer wouldn't?

Don't get me wrong, not arguing against it, but I would just like to understand a little bit more.

src/libostree/ostree-sign-ed25519.c
```diff
@@ -477,42 +489,46 @@ _ed25519_add_revoked (OstreeSign *self, GVariant *revoked_key, GError **error)
...
         }
 
-      if (line == NULL)
+      if (blob == NULL)
         return ret;
 
       /* Read the key itself */
-      /* base64 encoded key */
-      pk = g_variant_new_string (line);
+      pk = g_variant_new_fixed_array (G_VARIANT_TYPE ("y"), g_bytes_get_data (blob, NULL),
```
cgwalters, line 521: Hmm isn't this...oh, I see `ostree_sign_ed25519_add_pk` already accepted both strings and byte arrays.

src/libostree/ostree-sign-x509.c
```diff
@@ -0,0 +1,651 @@
...
+  g_assert (OSTREE_IS_SIGN (self));
+  OstreeSignX509 *sign = _ostree_sign_x509_get_instance_private (OSTREE_SIGN_X509 (self));
+
+  if (!_ostree_sign_x509_is_initialized (sign, error))
+    return FALSE;
+
+  if (sign->secret_key == NULL)
+    return glnx_throw (error, "Not able to sign: secret key is not set");
+
+#if defined(USE_OPENSSL)
```
cgwalters, line 126: I'm ok with this as is, but it would seem to me to be cleaner to basically `#ifdef` the whole function and have the `#else` block there be a "return unsupported" error...I think we have an example of that in `otcore_validate_ed25519_signature`.

src/libostree/ostree-sign-x509.c
```diff
@@ -0,0 +1,651 @@
...
+  if (data == NULL)
+    return glnx_throw (error, "x509: unable to verify NULL data");
+
+  OstreeSignX509 *sign = _ostree_sign_x509_get_instance_private (OSTREE_SIGN_X509 (self));
+
+  if (!_ostree_sign_x509_is_initialized (sign, error))
+    return FALSE;
+
+  if (signatures == NULL)
+    return glnx_throw (error, "x509: commit have no signatures of my type");
```
cgwalters, line 183: Under what scenarios would this happen out of curiosity? Don't we just want `g_assert (signatures != NULL);`?

src/libostree/ostree-sign-x509.h
```diff
@@ -0,0 +1,56 @@
+/* vim:set et sw=2 cin cino=t0,f0,(0,{s,>2s,n-s,^-s,e2s: */
+
+/*
+ * Copyright © 2019 Collabora Ltd.
```
cgwalters, line 4: I think you can add a different copyright here

src/libostree/ostree-sign-x509.h
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Authors:
```
cgwalters, line 21: And in new code we've been dropping `Authors` because git is canonical

cgwalters (commented):
Thanks so much for this! It overall looks really good! I mostly have nits and questions.

cgwalters:
> My main motivation here is to prepare for the upcoming PQC transition. While encryption (key encapsulation) is the current priority as it affects forward secrecy, signatures will come next, and it would be nice if ostree could support PQC signatures transparently.

But can you elaborate just a bit more - I'm just guessing here: did you do an audit of code shipped e.g. in RHEL linking to openssl? Do you happen to know about ostree specifically beforehand?

src/libostree/ostree-sign-x509.c
ueno: Fixed.

src/libostree/ostree-blob-reader-pem.c
ueno: OpenSSL provides PEM_* functions indeed, though they don't fit in our use-case reading as many PEM blocks incrementally from a file (stream); `PEM_read` expects a single complete PEM block, while `PEM_read_bio` requires to implement a BIO method backed by GInputStream (that wouldn't be difficult, but cumbersome as it also requires to use correct memory allocators: OpenSSL vs. GLib).

tests/test-pem.c
ueno: I added a couple more tests exercising invalid cases. Would it be OK to add fuzzing in a separate PR?

tests/test-signed-commit-dummy.sh
ueno: Thank you for the pointer; I've rewritten with those shell functions.

src/libostree/ostree-sign-x509.c
ueno: I tried this, but got an test failure:
```console
ERROR: tests/test-signed-pull.sh - Bail out! OSTreeSign:ERROR:src/libostree/ostree-sign-x509.c:169:ostree_sign_x509_data_verify: assertion failed: (signatures != NULL)
```
Looks like `ostree_sign_commit_verify` may call `ostree_sign_data_verify` with NULL `signatures`. Is there any case where this is supposed to succeed?

src/libostree/ostree-sign-x509.c
ueno: Done.

src/libostree/ostree-sign-x509.h
ueno: Fixed.

dbnicholson:
Sorry, I keep meaning to give some time to reviewing this. The one thing I'll say now is that I don't think `x509` is an appropriate name for the backend. That implies X.509 certificates, which would be interesting, but it's not what's happening here. The only connection to X.509 is using the `SubjectPublicKeyInfo` representation of public keys. This is a straightforward asymmetric digital signature scheme without certificates.

My suggested names would be:

* `pkey` - Borrow openssl's naming for public/private key handling.
* `pkcs8` - A misnomer since PKCS#8 representation only applies to the private key.
* `pem` - Still a little inaccurate since we're really talking about [RFC 7468](https://www.rfc-editor.org/rfc/rfc7468) textual encodings rather than strictly PEM. This also implies arbitrary types of PEM encodings are used, but we're only talking about [private](https://www.rfc-editor.org/rfc/rfc7468#section-11) and [public](https://www.rfc-editor.org/rfc/rfc7468#section-13) keys.
* `key`, `pubkey` or `keypair` - Straightforward names representing a signature using only public key cryptography rather than certificates or something else. It would just be documented that the "PEM" encodings from RFC 7468 are required.

src/libostree/ostree-sign-x509.c
cgwalters: Probably not but yeah let's continue to throw an error

tests/test-pem.c
cgwalters: Yep fuzzing is nonblocking

cgwalters:
Mild preference for `pkey` given that list

ueno:
I don't have strong opinion about the naming, but colleagues suggested `spki` or `pkcs` as well.

On a slightly different note, while this PR only supports SubjectPublicKeyInfo (SPKI) based raw public keys (as in [RFC 7250](https://www.rfc-editor.org/rfc/rfc7250)), I was wondering if it could eventually support X.509 certificates as well to take advantage of system trust store (either backed by p11-kit or file/directory based storage on `/etc/pki/`), instead of the custom directories, i.e., `/etc/ostree/{trusted,revoked}.*`.

ueno:
After a second thought, I'm leaning to `spki`, as we are more concerned about public keys here and private keys could be in any form, e.g., PKCS#8, something opaque on a hardware token, etc., with future extensions.

ueno:
Sorry for not having followed up recently. I rebased the PR against the latest main branch and also updated commit message with the use-case. @cgwalters @dbnicholson  could you take a look?

> > My main motivation here is to prepare for the upcoming PQC transition. While encryption (key encapsulation) is the current priority as it affects forward secrecy, signatures will come next, and it would be nice if ostree could support PQC signatures transparently.
> 
> But can you elaborate just a bit more - I'm just guessing here: did you do an audit of code shipped e.g. in RHEL linking to openssl? Do you happen to know about ostree specifically beforehand?

This is part of the effort to make software signatures to be compliant with the ([CNSA 2.0](https://media.defense.gov/2022/Sep/07/2003071836/-1/-1/0/CSI_CNSA_2.0_FAQ_.PDF)) guidance. There are also RPM (see https://github.com/rpm-software-management/rpm/issues/3363 and https://github.com/rpm-software-management/rpm/issues/3385) and container images (skopeo) to be covered, though I thought ostree commits are also in scope. Let me know if anything else is unclear.

cgwalters (approved):
Thanks, this looks great! Nice work!

PR #3282, opened 2024-08-14
deploy: Log to journal for boot space, not stderr

cgwalters:
bootc doesn't run as a systemd unit, and this pollutes stderr.


jlebon (approved):

cgwalters:
OK this build had a transient failure in the jenkins run, but actually the CI test does need to be updated to account for us not logging to stderr anymore...ug.

PR #3283, opened 2024-08-15
sysroot: Make coverity happy with dirname+strdup

cgwalters:
Similar to d528083cae3492f9b9424f3c9830869af7b4cbd0 - I don't believe we actually had a leak here because `dirname` always returns the same start pointer, but this makes Coverity happy.


jmarrero (approved):
lgtm

PR #3284, opened 2024-08-15
commit/payload-link: Ensure we don't overrun target_checksum size

cgwalters:
This is another warning from recently changed code from Coverity:

```
1. Defect type: OVERRUN
16. libostree-2024.7/src/libostree/ostree-repo-commit.c:823:7: overrun-buffer-arg: Overrunning array "target_checksum" of 65 bytes by passing it to a function which accesses it at byte offset 258 using argument "size" (which evaluates to 259). [Note: The source code implementation of the function has been overridden by a builtin model.]
```

I think this can only happen if the repository is corrupt; the data shouldn't be that long. But fix this by passing the max length we expect; this will ignore the rest currently.


jmarrero (approved):
lgtm

cgwalters:
CI failure is unrelated, see https://github.com/ostreedev/ostree/pull/3282#issuecomment-2291838101

PR #3285, opened 2024-08-15
tests: Attempt to update auto-prune test

cgwalters:
We need to look in the journal since
https://github.com/ostreedev/ostree/pull/3282/commits/3d07a808ce9a764c98560279fc444d2e4fb901db

### DIFF
--- a/tests/kolainst/destructive/auto-prune.sh
+++ b/tests/kolainst/destructive/auto-prune.sh
@@ -5,6 +5,34 @@ set -xeuo pipefail
 
 . ${KOLA_EXT_DATA}/libinsttest.sh
 
+journal_cursor() {
+    journalctl -o json -n 1 | jq -r '.["__CURSOR"]'
+}
+
+assert_journal_grep() {
+    local cursor re
+    cursor=$1
+    shift
+    re=$1
+    shift
+
+    if ! journalctl -t ostree --after-cursor "${cursor}" --grep="$re" "$@" >/dev/null; then
+        fatal "failed to find in journal: $re"; exit 1
+    fi
+}
+
+assert_not_journal_grep() {
+    local cursor re
+    cursor=$1
+    shift
+    re=$1
+    shift
+
+    if journalctl -t ostree --after-cursor "${cursor}" --grep="$re" "$@"; then
+        fatal "found in journal: $re"; exit 1
+    fi
+}
+
 # make two fake ostree commits with modified kernels of about the same size
 cd /root
 mkdir -p rootfs/usr/lib/modules/`uname -r`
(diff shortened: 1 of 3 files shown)


ericcurtin (approved):

cgwalters:
Ugh of course our journal grep command pulls up *itself* because the test is actually logging to the journal...

tests/kolainst/destructive/auto-prune.sh
```diff
@@ -5,6 +5,47 @@ set -xeuo pipefail
...
+    cat >/run/systemd/system/ostree-finalize-staged.service.d/10-no-early-prune.conf << 'EOF'
+[Service]
+Environment=OSTREE_SYSROOT_OPTS=no-early-prune
+EOF
+    systemctl daemon-reload
+}
+
+remove_no_early_prune() {
+    rm -vf /run/systemd/system/ostree-finalize-staged.service.d/10-no-early-prune.conf
+}
```
jlebon, line 47: Meh, probably not worth the complexity from doing this from the service unit. We can just keep running `ostree admin finalize-staged` directly and the logs we need should still go to the journal.

cgwalters:
Thanks for working on this, honestly what I hit is my cosa devenv bitrotted...too many build/test environments

tests/kolainst/destructive/auto-prune.sh
```diff
@@ -71,11 +100,11 @@ bootloader_orig=$(sha256sum /boot/loader/entries/*)
...
 consume_bootfs_space
 rpm-ostree rebase :modkernel2
-if OSTREE_SYSROOT_OPTS=no-early-prune ostree admin finalize-staged |& tee out.txt; then
+cursor=$(journal_cursor)
+if OSTREE_SYSROOT_OPTS=no-early-prune ostree admin finalize-staged; then
     assert_not_reached "successfully wrote kernel without auto-pruning"
 fi
-assert_file_has_content out.txt "No space left on device"
-rm out.txt
+assert_journal_grep "$cursor" "updating bootloader in two steps"
```
jlebon, line 107: I think this one needs to be e.g.

```bash
rm -vf err.txt
if OSTREE_SYSROOT_OPTS=no-early-prune ostree admin finalize-staged 2>err.txt; then
    assert_not_reached "successfully wrote kernel without auto-pruning"
fi
assert_file_has_content err.txt "No space left on device"
```

?

jlebon (commented):
Sorry, meant to circle back on this!

tests/kolainst/destructive/auto-prune.sh
cgwalters: I did that in the last push right?

cgwalters: But yeah I still haven't re-setup cosa, I've just been playing the game of "look at the CI result, blindly update the test and push, go do other things" rinse repeat...

cgwalters: The key change was using `journalctl -t ostree` to filter on just output from that binary and avoid the [grep finding itself](https://github.com/ostreedev/ostree/pull/3285#issuecomment-2292401986) problem.

cgwalters:
> [2024-08-30T18:39:17.547Z] Aug 30 18:39:15 qemu0 kola-runext-auto-prune.sh[2375]: /usr/local/bin/kola-runext-auto-prune.sh: line 81: assert_file_has_context: command not found

Whee hooray for interpreted code :cry:

tests/kolainst/destructive/auto-prune.sh
jlebon: Hmm, in the diff I'm seeing, this is changing the check we're doing on this section of the test. Before we were checking for "No space left on device" (which would show up on stderr), now we're checking for "updating bootloader in two steps" in the journal.

> The key change was using `journalctl -t ostree` to filter on just output from that binary and avoid the [grep finding itself](https://github.com/ostreedev/ostree/pull/3285#issuecomment-2292401986) problem.

Yeah... I walked right into that one too.

cgwalters: You're right, I have fixed that thanks!

Issue #2086, opened 2020-04-27
/sysroot private mount and /home -> /sysroot/home


cgwalters:
Related: https://github.com/ostreedev/ostree/issues/2085

Reported on IRC, today gnome-continuous uses `/home` :arrow_right: `/sysroot/home` but this conflicts with https://github.com/ostreedev/ostree/pull/1438 if the user creates mounts in their home directory.

I think basically everyone should use `/var/home`, but probably ostree should either disable the private mount for `/sysroot` if we detect these symlinks.

valentindavid:
I have noticed that if I remove totally `/home` :arrow_right: `var/home`, then Flatpak applications with `--filesystem=host` do not have access to home directory, unless they also have `--filesystem=home`. 

So it seems that home should be defined within `/var/home` in `/etc/passwd`, but the symlink should still exist for flatpak application to work correctly. (I still have to test that).

jlebon:
> So it seems that home should be defined within `/var/home` in `/etc/passwd`

Yes, note rpm-ostree for example does this: https://github.com/coreos/rpm-ostree/pull/1726.

valentindavid:
I can confirm that for Flatpak applications like org.flatpak.Builder need to have both the `/home` :arrow_right: `var/home` AND the home directory defined in `/etc/passwd` should be using `/var/home`. I have just tested it.

It is very peculiar that removing the symlink breaks flatpak.

dbnicholson:
We hit a very related issue in Endless OS recently that's caused by `/sysroot`'s private mount propagation. We have a test mode that mounts an overlay on /sysroot (and other disk backed mounts) so that you can make broad changes to the
system and have them discarded when the system is shut down. If a process is started with a separate mount namespace before test mode is invoked, then that process will not inherit the /sysroot overlay. This was causing new user home directories to persist in /sysroot/home since accountsservice started using systemd sandboxing features a couple years ago, but it just hit Endless OS early this year and apparently no one has run the test mode since.

[...]

PR #3292, opened 2024-08-30
switchroot: Stop making /sysroot mount private

dbnicholson:
Back in 2b8d586c5, /sysroot was changed to be a private mount so that
submounts of /var do not propagate back to the stateroot /var. That's
laudible, but it makes /sysroot different than every other shared mount
in the root namespace. In particular, it means that submounts of
/sysroot do not propagate into separate mount namespaces.

Rather than make /sysroot private, make /var a slave+shared mount so
that it receives mount events from /sysroot but not vice versa. That
achieves the same effect of preventing /var submount events from
propagating back to /sysroot while allowing /sysroot mount events to
propagate forward like every other system mount.

The mount propagation flags are applied as options in the generated
var.mount unit. This depends on a mount(8) feature that has been present
since util-linux 2.23. That's available in RHEL 7 and every non-EOL
Debian and Ubuntu release. Applying the propagation from var.mount fixes
a small race, too. Previously, if a /var submount was added before
/sysroot was made private, it would have propagated back into /sysroot.
That was possible since ostree-remount.service orders itself after
var.mount but not before any /var submounts.

Fixes: #2086

---

I included the added test separately in the first commit so you can see the difference in behavior.


dbnicholson:
Here's a comparison of mount propagation before and after this change using a cosa built VM. I mounted a tmpfs on `/var/foo` and `/sysroot/boot` to compare how those would propagate.

Current:
```
[core@cosa-devsh ~]$ findmnt -l -o TARGET,PROPAGATION,OPT-FIELDS
TARGET                                   PROPAGATION OPT-FIELDS
/sysroot                                 private     
/                                        shared      shared:1
/etc                                     shared      shared:2
/usr                                     shared      shared:3
/sysroot/ostree/deploy/fedora-coreos/var private     
/dev                                     shared      shared:6
/dev/shm                                 shared      shared:7
/dev/pts                                 shared      shared:8
/sys                                     shared      shared:9
/sys/kernel/security                     shared      shared:10
/sys/fs/cgroup                           shared      shared:11
/sys/fs/pstore                           shared      shared:12
/sys/fs/bpf                              shared      shared:13
/sys/kernel/config                       shared      shared:14
/proc                                    shared      shared:16
/run                                     shared      shared:17
/sys/fs/selinux                          shared      shared:15
/proc/sys/fs/binfmt_misc                 shared      shared:18
/dev/hugepages                           shared      shared:19
/dev/mqueue                              shared      shared:20
/sys/kernel/debug                        shared      shared:21
/sys/kernel/tracing                      shared      shared:22
/tmp                                     shared      shared:23
/sys/fs/fuse/connections                 shared      shared:24
/var                                     shared      shared:5
/boot                                    shared      shared:84
/var/mnt/workdir-tmp                     shared      shared:27
/var/mnt/workdir                         shared      shared:91
/var/lib/nfs/rpc_pipefs                  shared      shared:95
/run/user/1000                           shared      shared:401
/var/foo                                 shared      shared:362
/sysroot/boot                            private     
```

With this PR:
```
[core@cosa-devsh ~]$ findmnt -l -o TARGET,PROPAGATION,OPT-FIELDS
TARGET                                   PROPAGATION  OPT-FIELDS
/sysroot                                 shared       shared:4
/                                        shared       shared:1
/etc                                     shared       shared:2
/usr                                     shared       shared:3
/sysroot/ostree/deploy/fedora-coreos/var shared       shared:5
/dev                                     shared       shared:6
/dev/shm                                 shared       shared:7
/dev/pts                                 shared       shared:8
/sys                                     shared       shared:9
/sys/kernel/security                     shared       shared:10
/sys/fs/cgroup                           shared       shared:11
/sys/fs/pstore                           shared       shared:12
/sys/fs/bpf                              shared       shared:13
/sys/kernel/config                       shared       shared:14
/proc                                    shared       shared:16
/run                                     shared       shared:17
/sys/fs/selinux                          shared       shared:15
/proc/sys/fs/binfmt_misc                 shared       shared:18
/dev/hugepages                           shared       shared:19
/dev/mqueue                              shared       shared:20
/sys/kernel/debug                        shared       shared:21
/sys/kernel/tracing                      shared       shared:22
/tmp                                     shared       shared:23
/sys/fs/fuse/connections                 shared       shared:24
/var                                     shared,slave shared:58 master:5
/boot                                    shared       shared:86
/var/lib/nfs/rpc_pipefs                  shared       shared:163
/run/user/1000                           shared       shared:379
/var/foo                                 shared       shared:432
/sysroot/boot                            shared       shared:442
```

src/libostree/ostree-impl-system-generator.c
```diff
@@ -243,7 +250,7 @@ fstab_generator (const char *ostree_target, const bool is_aboot, const char *nor
                                "[Mount]\n"
                                "Where=%s\n"
                                "What=%s\n"
-                               "Options=bind\n",
+                               "Options=bind,slave,shared\n",
```
cgwalters, line 253: I'm confused, aren't the states of being shared and slave mutually exclusive?

cgwalters: Also unrelated to the above, we have a few places we mount `/var` unfortunately. There's some bits in `ostree-prepare-root.c`, especially if `INITRAMFS_MOUNT_VAR` is present.

I thought that at least some users relied on those and weren't actually running the fstab generator, FCOS included?

tests/kolainst/destructive/mount-propagation.sh
```diff
@@ -0,0 +1,127 @@
...
+require_writable_sysroot
+
+has_mount() {
+  local target=${1:?No target specified}
+  local pid=${2:?No PID specified}
+  local mounts="/proc/${pid}/mounts"
+
+  # Unlike grep, awk doesn't exit based on whether it matched anything or not,
+  # so that needs to be emulated.
+  awk 'BEGIN {ret = 1}; $2 == "'"${target}"'" {print; ret = 0}; END {exit ret}' "${mounts}"
```
cgwalters, line 16: We could instead parse JSON output from e.g. `findmnt -J --task`...

src/libostree/ostree-impl-system-generator.c
dbnicholson: Sorry, I should have been more explicit about this. The idea is to make the mount a slave of `/sysroot` and then create a separate shared group underneath that. From [mount_namespaces(7)](https://man7.org/linux/man-pages/man7/mount_namespaces.7.html#SHARED_SUBTREES):

> MS_SLAVE
> [mount(2)](https://man7.org/linux/man-pages/man2/mount.2.html) and [umount(2)](https://man7.org/linux/man-pages/man2/umount.2.html) events propagate into this mount
> from a (master) shared peer group.  [mount(2)](https://man7.org/linux/man-pages/man2/mount.2.html) and [umount(2)](https://man7.org/linux/man-pages/man2/umount.2.html)
> events under this mount do not propagate to any peer.
>
> Note that a mount can be the slave of another peer group
> while at the same time sharing [mount(2)](https://man7.org/linux/man-pages/man2/mount.2.html) and [umount(2)](https://man7.org/linux/man-pages/man2/umount.2.html)
> events with a peer group of which it is a member.  (More
> precisely, one peer group can be the slave of another peer
> group.)

What this means is that when you make a child mount of `/var`, it will be part of a shared group just like before. If you just make `/var` a slave to `/sysroot`, then a child mount of `/var` would become private. That's not what we want since then `/var/cache` would not propagate into an existing mount namespace.

This is show systemd sets up mounts when it creates a separate mount namespace. For example:

```
$ sudo nsenter -m -t $(systemctl show -P MainPID accounts-daemon.service) -- findmnt -l -o TARGET,PROPAGATION,OPT-FIELDS
TARGET                                                              PROPAGATION   OPT-FIELDS
/                                                                   shared,slave  shared:109 master:1
/run                                                                shared,slave  shared:125 master:13
/run/lock                                                           shared,slave  shared:126 master:14
/tmp                                                                shared,slave  shared:131 master:20
/efi                                                                shared,slave  shared:132 master:22
/dev                                                                shared        shared:133
/dev/pts                                                            shared,slave  shared:134 master:4
/dev/shm                                                            shared,slave  shared:135 master:3
/dev/mqueue                                                         shared,slave  shared:136 master:17
/dev/hugepages                                                      shared,slave  shared:137 master:16
/etc                                                                shared,slave  shared:138 master:1
/home                                                               shared,slave  shared:139 master:1
/proc                                                               shared        shared:140
/proc/sys/fs/binfmt_misc                                            shared,slave  shared:141 master:46
/proc/acpi                                                          shared        shared:142
/proc/asound                                                        shared        shared:143
/proc/bus                                                           shared        shared:144
/proc/fs                                                            shared        shared:145
/proc/irq                                                           shared        shared:146
/proc/kallsyms                                                      shared,slave  shared:147 master:13
/proc/kcore                                                         shared,slave  shared:148 master:13
/proc/mtrr                                                          shared        shared:149
/proc/scsi                                                          shared        shared:150
/proc/556/loginuid                                                  shared        shared:151
/proc/sys                                                           shared        shared:152
/proc/sys/fs/binfmt_misc                                            shared,slave  shared:153 master:46
/proc/sysrq-trigger                                                 shared        shared:154
/root                                                               shared,slave  shared:155 master:1
/run/credentials                                                    shared,slave  shared:127 master:13
/run/systemd/incoming                                               private,slave master:13
/run/systemd/seats                                                  shared,slave  shared:129 master:13
/run/user                                                           shared,slave  shared:130 master:13
/sys                                                                shared        shared:156
/sys/firmware/efi/efivars                                           shared,slave  shared:157 master:9
/sys/fs/bpf                                                         shared,slave  shared:158 master:10
/sys/fs/cgroup                                                      shared,slave  shared:159 master:7
/sys/fs/fuse/connections                                            shared,slave  shared:161 master:21
/sys/fs/pstore                                                      shared,slave  shared:162 master:8
/sys/kernel/config                                                  shared,slave  shared:163 master:11
/sys/kernel/debug                                                   shared,slave  shared:164 master:18
/sys/kernel/security                                                shared,slave  shared:165 master:6
/sys/kernel/tracing                                                 shared,slave  shared:166 master:19
/sys/fs/cgroup/system.slice/accounts-daemon.service/memory.pressure shared,slave  shared:160 master:7
/usr/lib/modules                                                    shared,slave  shared:167 master:13
/var/lib/AccountsService                                            shared,slave  shared:168 master:1
/var/log/lastlog                                                    shared,slave  shared:169 master:1
/var/mail                                                           shared,slave  shared:170 master:1
/efi                                                                shared,slave  shared:364 master:335
/run/user/1001                                                      shared,slave  shared:416 master:377
/run/user/1001                                                      shared,slave  shared:415 master:377
/run/user/1001/gvfs                                                 shared,slave  shared:593 master:587
/run/user/1001/gvfs                                                 shared,slave  shared:592 master:587
/run/user/1001/doc                                                  shared,slave  shared:826 master:794
/run/user/1001/doc                                                  shared,slave  shared:825 master:794
```

As for the `/var` mounting within prepare-root, I'm happy to fix that up. From my testing with cosa, it appears the generated `var.mount` is created and applied regardless of whether `/var` was mounted in the initramfs, but I agree it should DTRT out of the box.

tests/kolainst/destructive/mount-propagation.sh
dbnicholson: Yes, totally fair. The unfortunate part there is `findmnt` always reads `/proc/self/mountinfo` whereas I'm trying to read the mounts of an arbitrary process. You can do the same with `nsenter -m -t $pid findmnt`, but it's a bit more verbose.

src/libostree/ostree-impl-system-generator.c
dbnicholson: I added some more info in the comments and commit message about slave+shared propagation. I also updated the initramfs mounted `/var` case, too.

tests/kolainst/destructive/mount-propagation.sh
dbnicholson: I changed this to use `nsenter findmnt --json | jq`. It's wordier but a little nicer. I think I'm in the minority of understanding `awk` :smile:

src/switchroot/ostree-remount.c
```diff
@@ -167,15 +167,6 @@ main (int argc, char *argv[])
   }
   g_autoptr (GVariantDict) ostree_run_metadata = g_variant_dict_new (ostree_run_metadata_v);
 
-  /* The /sysroot mount needs to be private to avoid having a mount for e.g. /var/cache
```
cgwalters, line 170: It is very compelling to be able to delete this code...longer term what I think we want is for `ostree-remount.c` to basically go away.

cgwalters (approved):
I think I need some quality time with this, but overall makes sense to me enough that I think it's better to merge and give this some soak time in git main, for the very few people that run from git main (happens to include my workstation though).

Thanks for this!

Issue #3295, opened 2024-09-10
Incorrect check for static configs with bootupd


travier:
https://github.com/ostreedev/ostree/pull/3205 added a check that looks for `"static-configs"` in bootupd state file: `/boot/bootupd-state.json`.

Unfortunately, bootup will always write this entry (even if null) since https://github.com/coreos/bootupd/pull/609.

This is likely the source of https://github.com/fedora-silverblue/issue-tracker/issues/595.

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=2312453

PR #3300, opened 2024-09-13
bootloader/grub2: Handle empty static configs

travier:
In #3205, we introduced a check to skip re-generating the GRUB config if we detect that static configs are in used by looking at bootupd's state.

Unfortunately this check is incomplete and does not account for present but null entries in the JSON state file.

A proper fix would be to parse the JSON but this requires a larger code change.

Fixes: https://github.com/ostreedev/ostree/issues/3295
Fixes: https://github.com/ostreedev/ostree/pull/3205

### DIFF
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -30,6 +30,7 @@
 #define BOOTUPD_CONFIG "boot/bootupd-state.json"
 // Horrible hack, to avoid including a JSON parser we just grep for this
 #define BOOTUPD_CONFIG_STATIC_JSON_FRAGMENT "\"static-configs\""
+#define BOOTUPD_CONFIG_STATIC_JSON_FRAGMENT_NULL "\"static-configs\":null"
 
 /* Maintain backwards compatibility with legacy GRUB
  * installations that might rely on the -16 suffix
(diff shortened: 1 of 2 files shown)


travier:
Still test / investigating if this is really the issue.

travier:
https://github.blog/changelog/2024-02-13-deprecation-notice-v1-and-v2-of-the-artifact-actions/

travier:
CI fixes in https://github.com/ostreedev/ostree/pull/3301

src/libostree/ostree-bootloader-grub2.c
```diff
@@ -30,6 +30,7 @@
 #define BOOTUPD_CONFIG "boot/bootupd-state.json"
 // Horrible hack, to avoid including a JSON parser we just grep for this
 #define BOOTUPD_CONFIG_STATIC_JSON_FRAGMENT "\"static-configs\""
+#define BOOTUPD_CONFIG_STATIC_JSON_FRAGMENT_NULL "\"static-configs\":null"
```
cgwalters, line 33: Ouch :cry: 

I dunno, maybe overdue for us to link in a json library, I had just hoped to do that with rust (that's in https://github.com/ostreedev/ostree/pull/3054 but nontrivial)

cgwalters (approved):

src/libostree/ostree-bootloader-grub2.c
travier: Yeah, I agree it's really not great :/

travier:
This is working from initial testing.

champtar:
Just stumbled on this bug with EL 9.4, is it already planned for backport to EL 9.4 (or 9.5) or should I open an issue/case ?

PR #3304, opened 2024-09-15
Redo pages workflow

dbnicholson:


cgwalters (approved):
🙏 Thank you so much for fixing this!

dbnicholson:
Whoops, I was actually trying to make the PR against my own repo for some more testing 😀. I'm pretty sure it's good as is, but I might come back with another tweak or two.

PR #3305, opened 2024-09-15
workflow/docs: Fix deployments

dbnicholson:
A couple fixes to make PRs and non-PRs work correctly:

* In a conditional expression, `true` or `false` are returned unless you terminate both sides in a ternary. That was causing 2 strings to be suffixed with `false` instead of an empty string.
* For a PR, we do actually want to cancel in progress runs since there's no danger of breaking an in progress deployment.
* For PRs, just use the same `github-pages-pr` name for the artifact. The important part is that it's not called `github-pages` where an in progress deployment could pick it up. Otherwise it can use the same name all the time.

I've tested this now in my repo in all 3 triggers:

* [PR](https://github.com/dbnicholson/ostree/actions/runs/10873928054)
* [Push](https://github.com/dbnicholson/ostree/actions/runs/10873968648)
* [Manual](https://github.com/dbnicholson/ostree/actions/runs/10874004920)

Pages deployed successfully at https://dbnicholson.github.io/ostree/.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -15,10 +15,11 @@ on:
 # want to allow these production deployments to complete.
 #
 # Since pull requests use a unique artifact name and won't be deployed, they
-# shouldn't be limited. Use a unique group name in that case.
+# shouldn't be limited. Use a unique group name in that case and let in
+# progress runs be cancelled.
 concurrency:
-  group: "pages${{ github.event_name == 'pull_request' && format('-pr{0}', github.event.number) }}"
-  cancel-in-progress: false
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

cgwalters:
Nice!

dbnicholson:
That failed one last time [on push](https://github.com/ostreedev/ostree/actions/runs/10874149132) because Pages was still setup to deploy from the [gh-pages branch](https://github.com/ostreedev/ostree/tree/gh-pages). I changed the settings to deploy from GitHub Actions, and then [triggered a manual run](https://github.com/ostreedev/ostree/actions/runs/10874232137), which succeeded.

This all appears to be working again and we should delete the `gh-pages` branch.

Issue #3299, opened 2024-09-13
`flatpak update` produces ostree error with curl 8.10.0 (+ patch)


adamthiede:
When running `flatpak update` on Alpine Linux edge, the following error message is produced:

```
OSTree:ERROR:src/libostree/ostree-fetcher-curl.c:534:sock_cb: code should not be reached
Bail out! OSTree:ERROR:src/libostree/ostree-fetcher-curl.c:534:sock_cb: code should not be reached
Aborted
```

This began when [this patch](https://github.com/curl/curl/commit/48f61e781a01e6a8dbc4a347e280644b1c68ab6a) was added to Alpine's curl package; the patch addresses [this issue](https://github.com/curl/curl/issues/14860) and is already in curl's master branch.

Let me know if I can provide any more info.

See here for more info. https://gitlab.alpinelinux.org/alpine/aports/-/issues/16444

cgwalters:
Hmm fun, thanks for the report. This will require some analysis. If you happen to have the environment set up it'd be quite useful to know if this reproduces with `make check` in the ostree unit tests.

adamthiede:
Sure thing.

For reference, here's how Alpine builds ostree: https://git.alpinelinux.org/aports/tree/community/ostree/APKBUILD

The "build" step has the configuration options in it. I've built it myself and confirmed that removing the '--with-curl' option does fix this, but I'm unaware of any downsides to that.

Attached is the output of `make check` after the configure and build step that's in that APKBUILD. (In this instance I did not remove '--with-curl'.) Which test am I looking for?

[make_check.txt](https://github.com/user-attachments/files/16998911/make_check.txt)

adamthiede:
In the error it mentions this line:
https://github.com/ostreedev/ostree/blob/main/src/libostree/ostree-fetcher-curl.c#L534
I built ostree with that line commented out, and it worked. I assume that's because it was falling back to libsoup though?

vixalien:
The issue has been "fixed" in alpine linux: https://gitlab.alpinelinux.org/alpine/aports/-/issues/16444

All you need to do is downgrade libcurl by doing an `apk upgrade -a`

adamthiede:
Yes, it's working again, but unless this bug is fixed in curl, then it'll likely be relevant again soon.

adamthiede:
[...]

PR #3306, opened 2024-09-18
curl: Assert that curl_multi_assign worked

cgwalters:
ref https://github.com/ostreedev/ostree/issues/3299

This won't fix that issue, but *if* this assertion triggers it should give us a better idea of the possible codepaths where it is happening.


jmarrero (approved):
lgtm

PR #3307, opened 2024-09-18
curl: Make socket callback during cleanup into no-op

cgwalters:
Because curl_multi_cleanup may invoke callbacks, we effectively have
some circular references going on here. See discussion in

https://github.com/curl/curl/issues/14860

Basically what we do is the socket callback libcurl may invoke into a no-op when
we detect we're finalizing. The data structures are owned by this object and
not by the callbacks, and will be destroyed below. Note that
e.g. g_hash_table_unref() may itself invoke callbacks, which is where
some data is cleaned up.

---

### DIFF
--- a/src/libostree/ostree-fetcher-curl.c
+++ b/src/libostree/ostree-fetcher-curl.c
@@ -78,6 +78,7 @@ struct OstreeFetcher
   struct curl_slist *extra_headers;
   int tmpdir_dfd;
   bool force_anonymous;
+  bool finalizing; // Set if we're in the process of teardown
   char *custom_user_agent;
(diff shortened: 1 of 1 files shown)


jmarrero (approved):
lgtm

cgwalters:
For people who are distributing ostree: in my opinion this patch is quite safe to cherry pick. The affected code here is during the teardown path, so realistically the worst case is memory leaks. But for leaks: ASAN is still happy with this change, and from my auditing of the code again we're always relying on our own teardown, not on the libcurl socket callback.

ThatOneCalculator (approved):

dbnicholson (approved):

Issue #3032, opened 2023-09-06
extensible metadata on deployments


cgwalters:
As prep for https://github.com/ostreedev/ostree/issues/2725 but also in general, I think it'd be useful to have an official "metadata" API for deployments - surfacing things like "did this deployment boot before" in a way indepdenent of the journal.

I think this could be a pretty thin wrapper for extended attributes on the deployment root directory.  (Need to verify whether the immutable bit breaks this; if it does we could just document using the origin file for this)

cgwalters:
BTW if we do this we should also teach the OpenShift MCO to do this, to describe which rendered configs it used for the deployment.

cgwalters:
One thought related to this is that with composefs, we automatically gain a new `.cfs` file that would a natural target for xattrs.

cgwalters:
(And maybe we should just turn on composefs *generation* unconditionally by default?)

jmarrero:
we discussed turning composefs for Fedora ELN first.

cgwalters:
One thing that happened with https://github.com/ostreedev/ostree/pull/3114/#discussion_r1420552680 is we gained a whole new directory created unconditionally now - so we can trivially drop in a new file there in whatever format we want (but probably gvariant) or xattrs on that directory.

cgwalters:
This one also tangentially relates to https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/ as the success/failure of a boot is something that could make sense here - or in the bootloader configs.

jmarrero:
I see that is targeted for `kernel-install` to start the process. Since we are not doing a kernel install but a deployment, we could:

- check what kernel is part of the deployment
- ostree exposes a boot_retry count option
- ostree creates: `/etc/kernel/tries` with the boot_retry value.
- ostree creates: /boot/loader/entries/KERNEL_VER.fc27.x86_64+BOOT_RETRY.conf 

Then the process described in the **Walkthrough section** of https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/ will take over. We can embed additional information on `/boot/loader/entries/KERNEL_VER.fc27.x86_64+BOOT_RETRY.conf` for greenboot to consume either in the file itself as text as long systemd does not care or extended attributes if systemd cares about file content.

cgwalters:
xref https://github.com/systemd/systemd/blob/0630069f0bdba4535c2c659fbf2c9aa632268272/src/kernel-install/90-loaderentry.install.in#L96

[...]

PR #3310, opened 2024-09-19
[RFC] sysroot: Support boot counting for boot entries

igoropaniuk:
```
Add support for boot counting for bootloader entries [1]. The boot counting data is stored in the name of the boot loader entry. A boot loader entry file name may contain a plus (+) followed by a number. This may optionally be followed by a minus (-) followed by a second number. The dot (.) and file name suffix (conf or efi) must immediately follow.

All "pending" entries (in the middle of boot counting) are automatically removed during creation of new boot entries for new deployments.

Testing:
$ ostree admin deploy 91fc19319be9e79d07159303dff125f40f10e5c25614630dcbed23d95e36f907 Copying /etc changes: 2 modified, 3 removed, 4 added bootfs is sufficient for calculated new size: 0 bytes Transaction complete; bootconfig swap: yes; bootversion: boot.0.1, deployment count change: 1

$ ls /boot/loader/entries
ostree-1+3.conf  ostree-2+3.conf

[1] https://uapi-group.org/specifications/specs/boot_loader_specification/#boot-counting
```

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -62,6 +62,8 @@
   SD_ID128_MAKE (e8, 64, 6c, d6, 3d, ff, 46, 25, b7, 79, 09, a8, e7, a4, 09, 94)
 #endif
 
+#define BOOT_COUNT_MAX_RETRIES 3
+
 /* How much additional space we require available on top of what we accounted
  * during the early prune fallocate space check. This accounts for anything not
  * captured directly by `get_kernel_layout_size()` like writing new BLS entries.
@@ -1776,12 +1778,12 @@ parse_os_release (const char *contents, const char *split)
   return ret;
 }
 
-/* Generate the filename we will use in /boot/loader/entries for this deployment.
+/* Generate the entry name we will use in /boot/loader/entries for this deployment.
  * The provided n_deployments should be the total number of target deployments (which
  * might be different from the cached value in the sysroot).
  */
 static char *
-bootloader_entry_filename (OstreeSysroot *sysroot, guint n_deployments,
+bootloader_entry_name (OstreeSysroot *sysroot, guint n_deployments,
                            OstreeDeployment *deployment)
 {
   guint index = n_deployments - ostree_deployment_get_index (deployment);
@@ -1792,12 +1794,50 @@ bootloader_entry_filename (OstreeSysroot *sysroot, guint n_deployments,
   if (use_old_naming)
     {
       const char *stateroot = ostree_deployment_get_osname (deployment);
-      return g_strdup_printf ("ostree-%d-%s.conf", index, stateroot);
+      return g_strdup_printf ("ostree-%d-%s", index, stateroot);
     }
   else
     {
-      return g_strdup_printf ("ostree-%d.conf", index);
+      return g_strdup_printf ("ostree-%d", index);
     }
+}
+
+/* Drop all temporary entries in /boot/loader/entries for this deployment,
+ * which were created during automatic boot assesment
+ * https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/
+ */
+static gboolean
+bootloader_remove_tmp_entries (int dfd, const char *entry_name, GCancellable *cancellable,
+                               GError **error)
+{
+  g_auto(GLnxDirFdIterator) dfd_iter = { 0, };
+  g_autofree char *entry_name_init = g_strdup_printf ("%s+%d", entry_name, BOOT_COUNT_MAX_RETRIES);
+
+  if (!glnx_dirfd_iterator_init_at (dfd, ".", FALSE, &dfd_iter, error))
+    return FALSE;
+
+  while (TRUE)
+    {
+      struct dirent *dent = NULL;
+
+      if (!glnx_dirfd_iterator_next_dent_ensure_dtype (&dfd_iter, &dent, cancellable, error))
+        return FALSE;
+      if (dent == NULL)
+        break;
+
+      /* Don't remove default boot entry (with +3 suffix) */
+      if (g_str_has_prefix (dent->d_name, entry_name_init))
+        continue;
+
+      if (g_str_has_prefix (dent->d_name, entry_name))
+        {
+          if (!glnx_shutil_rm_rf_at (dfd_iter.fd, dent->d_name, cancellable, error))
+            return FALSE;
+        }
+
+    }
+
+    return TRUE;
 }
 
 /* Given @deployment, prepare it to be booted; basically copying its
@@ -1834,7 +1874,7 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
   const char *bootcsum = ostree_deployment_get_bootcsum (deployment);
   g_autofree char *bootcsumdir = g_strdup_printf ("ostree/%s-%s", osname, bootcsum);
   g_autofree char *bootconfdir = g_strdup_printf ("loader.%d/entries", new_bootversion);
-  g_autofree char *bootconf_name = bootloader_entry_filename (sysroot, n_deployments, deployment);
+  g_autofree char *bootconf_name = bootloader_entry_name (sysroot, n_deployments, deployment);
 
   if (!glnx_shutil_mkdir_p_at (sysroot->boot_fd, bootcsumdir, 0775, cancellable, error))
     return FALSE;
@@ -2146,8 +2186,13 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
   if (!glnx_opendirat (sysroot->boot_fd, bootconfdir, TRUE, &bootconf_dfd, error))
     return FALSE;
 
+  g_autofree char *bootconf_filename = g_strdup_printf ("%s+%d.conf", bootconf_name, BOOT_COUNT_MAX_RETRIES);
+
+  if (!bootloader_remove_tmp_entries(bootconf_dfd, bootconf_name, cancellable, error))
+    return FALSE;
+
   if (!ostree_bootconfig_parser_write_at (ostree_deployment_get_bootconfig (deployment),
-                                          bootconf_dfd, bootconf_name, cancellable, error))
+                                          bootconf_dfd, bootconf_filename, cancellable, error))
     return FALSE;
 
   return TRUE;
@@ -4176,14 +4221,19 @@ ostree_sysroot_deployment_set_kargs_in_place (OstreeSysroot *self, OstreeDeploym
       ostree_bootconfig_parser_set (new_bootconfig, "options", kargs_str);
 
       g_autofree char *bootconf_name
-          = bootloader_entry_filename (self, self->deployments->len, deployment);
+          = bootloader_entry_name (self, self->deployments->len, deployment);
 
       g_autofree char *bootconfdir = g_strdup_printf ("loader.%d/entries", self->bootversion);
       glnx_autofd int bootconf_dfd = -1;
       if (!glnx_opendirat (self->boot_fd, bootconfdir, TRUE, &bootconf_dfd, error))
         return FALSE;
 
-      if (!ostree_bootconfig_parser_write_at (new_bootconfig, bootconf_dfd, bootconf_name,
+      g_autofree char *bootconf_filename = g_strdup_printf ("%s+%d.conf", bootconf_name, BOOT_COUNT_MAX_RETRIES);
+
+      if (!bootloader_remove_tmp_entries(bootconf_dfd, bootconf_name, cancellable, error))
+        return FALSE;
+
+      if (!ostree_bootconfig_parser_write_at (new_bootconfig, bootconf_dfd, bootconf_filename,
                                               cancellable, error))
         return FALSE;
     }


cgwalters:
I love the idea of integrating here, thanks for starting this! This came up before in at least https://github.com/ostreedev/ostree/issues/3032

Hmm don't we need to consider how we make this configurable? Currently systemd kernel-install parses `/etc/kernel/tries`. (I think we should also support a `/usr/lib` variant of this, but that's mostly an aside). I guess we could in theory just start honoring that file too?

igoropaniuk:
@cgwalters thanks for the feedback, my concern regarding parsing `/etc/kernel/tries` is that it's systemd-specific config, which is not described in `https://uapi-group.org/specifications/specs/boot_loader_specification/`

A can add an optional autoconf flag for that

igoropaniuk:
@cgwalters added support for `/etc/kernel/tries` as suggested

configure.ac
```diff
@@ -362,6 +362,14 @@ AS_IF([test "$enable_man_html" = yes && test "$enable_man" = no], [
 ])
 AM_CONDITIONAL(ENABLE_MAN_HTML, test "$enable_man_html" = yes)
 
+AC_ARG_ENABLE(boot-count,
```
cgwalters, line 365: I think instead of a build-time option this should be a runtime option in the sysroot; something like 
```
[sysroot]
boot-counting = true
```
or so?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1792,14 +1792,78 @@ bootloader_entry_filename (OstreeSysroot *sysroot, guint n_deployments,
...
 
+#ifdef ENABLE_BOOT_COUNT
+
+#define BOOT_COUNT_MAX_RETRIES 3
+
+static gint
+bootloader_get_max_boot_tries (OstreeSysroot *self, OstreeDeployment *deployment,
+                               GCancellable *cancellable, GError **error)
+{
+  g_autofree char *tries_file_path = ostree_deployment_get_boot_tries_relpath (deployment);
```
cgwalters, line 1811: Hmmm...one super messy thing here is that at this point we may not have actually written out the deployment's `/etc`. We really want a version of this that's in `/usr`.

AFAICS, the uapi standard doesn't specify `/etc/kernel/tries` - that's an invention of `kernel-install`.

To combine with "make this configurable via sysroot" we could have e.g.:

```
[sysroot]
boot_counting=3
```

But OTOH I've been trying to get away from the sysroot configuration for other reasons...so we may need to invent `/usr/lib/ostree/sysroot.conf` or so as a pair with `/usr/lib/ostree/prepare-root.conf`?

cgwalters (commented):

cgwalters:
BTW, which bootloader are you looking at using this with?

igoropaniuk:
@cgwalters EDK2 + systemd-boot

configure.ac
igoropaniuk: yes, makes sense, will fix. thanks!

igoropaniuk: fixed

src/libostree/ostree-sysroot-deploy.c
igoropaniuk: added support for `boot_counting` property in repo config

igoropaniuk:
@cgwalters I've addressed all comments,  thanks!

igoropaniuk:
@cgwalters looks like some CI tests are still failing, but all of them are unrelated to this change, for example:
```
2024-10-30T16:06:09.9538403Z ERROR: tests/test-commit-sign.sh - too few tests run (expected 7, got 2)
2024-10-30T16:06:09.9539247Z ERROR: tests/test-commit-sign.sh - exited with status 1
```

cgwalters:
Yes CI failures are unrelated, will look at this soon

man/ostree.repo-config.xml
```diff
@@ -416,6 +416,14 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
         </listitem>
       </varlistentry>
 
+      <varlistentry>
+        <term><varname>boot_counting</varname></term>
```
jlebon, line 420: Bikeshed: for consistency with other knobs in this file, `boot-counting`?

Also, should this be e.g. `boot-counting-tries` to make it clearer?

man/ostree.repo-config.xml
```diff
@@ -416,6 +416,14 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
...
+        <listitem><para>Integer value controlling the number of maximum boot attempts. The boot counting data is stored in the name of the
+        boot loader entry. A boot loader entry file name may contain a plus (+) followed by a number. This may optionally be followed by
+        a minus (-) followed by a second number. The dot (.) and file name suffix (conf or efi) must immediately follow.
```
jlebon, line 423: Feel like a link to the spec here would be helpful.

src/libostree/ostree-repo.c
```diff
@@ -3297,6 +3297,20 @@ reload_remote_config (OstreeRepo *self, GCancellable *cancellable, GError **erro
...
 reload_sysroot_config (OstreeRepo *self, GCancellable *cancellable, GError **error)
 {
+  {
+    g_autofree char *boot_counting_str = NULL;
+
+    (void)ot_keyfile_get_value_with_default_group_optional (self->config, "sysroot", "boot_counting", "0",
+                                                            &boot_counting_str, NULL);
+
+    if (boot_counting_str)
+      /* Ensure boot count value is in [0,5] */
```
jlebon, line 3307: Hmm, doesn't seem required by the spec. Is that a systemd-boot requirement?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1792,14 +1792,72 @@ bootloader_entry_filename (OstreeSysroot *sysroot, guint n_deployments,
...
+}
+
+static gboolean
+bootloader_is_boot_count_enabled (OstreeSysroot *self, GCancellable *cancellable, GError **error)
+{
+  g_autoptr (OstreeRepo) repo = NULL;
+  if (!ostree_sysroot_get_repo (self, &repo, cancellable, error))
+    return FALSE;
+
+  return (repo->boot_counting != 0 ? TRUE : FALSE);
```
jlebon, line 1820:
```suggestion
  return repo->boot_counting != 0;
```

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2146,8 +2204,22 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
...
+        return FALSE;
+    }
+  else
+    {
+      bootconf_filename = g_strdup_printf ("%s.conf", bootconf_name);
+    }
+
   if (!ostree_bootconfig_parser_write_at (ostree_deployment_get_bootconfig (deployment),
-                                          bootconf_dfd, bootconf_name, cancellable, error))
+                                          bootconf_dfd, bootconf_filename, cancellable, error))
```
jlebon, line 2222: Hmm, but note that this function is called for _every_ deployment on _every_ update (that results in a BLS write). So this would effectively "unbless" a boot that was previously blessed (or reset to the max boots that did have some failures).

ISTM like the boot count needs to be part of e.g. the `OstreeDeployment` object so that we can leave it untouched when rewriting BLS entries.

igoropaniuk: Technically we can have just a boolean flag (`bootable` field for example) for that, and check if a filename of an old BLS entry (stored in `/boot/loader.%d/entries`) still has a boot counter suffix. And if it doesn't -> set OstreeDeployment->bootable, so then we can create all new BLS entries for this particular deployment without any boot counters.

The problem is that deployment index for OstreeDeployment is not static and there is no deterministic way to match existing boot entry with a specific OstreeDeployment, maybe only parsing BLS entry contents could help

src/libostree/ostree-repo.c
igoropaniuk: well, I took the initiative here :) . I personally don't see any reasons why we should have support for more then 5 boot attempts.

I can drop this limitation if there are objections.

src/libostree/ostree-sysroot-deploy.c
jlebon: > The problem is that deployment index for OstreeDeployment is not static and there is no deterministic way to match existing boot entry with a specific OstreeDeployment, maybe only parsing BLS entry contents could help

There is, but that state is currently not kept around. If you look at e.g. `sysroot_load_from_bootloader_configs()` is where we go through the BLS entries and convert those to an ordered list of `OstreeDeployment`s. So we could parse the counter information from the filename there and keep that state in the deployment objects.

igoropaniuk:
The PR is not stalled, I'll get back to it tomorrow

man/ostree.repo-config.xml
igoropaniuk: Changed to `boot-counting-tries`

man/ostree.repo-config.xml
igoropaniuk: Added a link to the sepc

src/libostree/ostree-sysroot-deploy.c
igoropaniuk: Re-implemented as you suggested, please take a look

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -4213,14 +4253,35 @@ ostree_sysroot_deployment_set_kargs_in_place (OstreeSysroot *self, OstreeDeploym
...
       if (!glnx_opendirat (self->boot_fd, bootconfdir, TRUE, &bootconf_dfd, error))
         return FALSE;
 
-      if (!ostree_bootconfig_parser_write_at (new_bootconfig, bootconf_dfd, bootconf_name,
+      g_autofree char *bootconf_filename;
+      if (bootloader_is_boot_count_enabled(self, cancellable, error))
+        {
+          guint max_tries = bootloader_get_max_boot_tries (self, cancellable, error);
+
+          if (!ostree_bootconfig_parser_is_parsed (new_bootconfig))
```
igoropaniuk, line 4268: Just realized that it makes sense to put that into a separate function to avoid code duplication

igoropaniuk: fixed

igoropaniuk:
@jlebon @cgwalters all comments фку addressed, any additional suggestions/objections are welcome

BTW, I've cross-checked CI checks that failed, looks like they are unrelated again

src/libostree/ostree-bootconfig-parser.h
```diff
@@ -67,4 +67,13 @@ void ostree_bootconfig_parser_set_overlay_initrds (OstreeBootconfigParser *self,
 _OSTREE_PUBLIC
 char **ostree_bootconfig_parser_get_overlay_initrds (OstreeBootconfigParser *self);
 
+_OSTREE_PUBLIC
+guint64 ostree_bootconfig_parser_get_tries_left (OstreeBootconfigParser *self);
```
cgwalters, line 71: If we want these to actually be public they'll need to be added to the `libostree-devel.sym` file.

src/libostree/ostree-bootconfig-parser.c
```diff
@@ -56,6 +59,94 @@ ostree_bootconfig_parser_clone (OstreeBootconfigParser *self)
...
+
+  self->tries_left = tries_left;
+  self->tries_done = tries_done;
+}
+
+/**
+ * ostree_bootconfig_parser_get_tries_left:
+ * @self: Parser
+ *
+ * Returns: Amount of boot tries left
```
cgwalters, line 118: Please also add:
`Since: 2025.2` (here and below)

src/libostree/ostree-bootconfig-parser.c
```diff
@@ -56,6 +59,94 @@ ostree_bootconfig_parser_clone (OstreeBootconfigParser *self)
...
+/**
+ * ostree_bootconfig_parser_parse_tries:
+ * @self: Parser
+ * @path: File path
+ *
+ * Parses a suffix of two counters in the form "+LEFT-DONE" from the end of the
+ * filename (excluding file extension).
+ */
+static void
+ostree_bootconfig_parser_parse_tries (OstreeBootconfigParser *self,  const char *filename)
```
cgwalters, line 71: I think this one would be easier to unit test if it was just
```
parse_bootloader_tries (const char *filename, guint32 *out_left, guint32 *out_done, GError **error)
```

(I also added an error so we can error if we fail to parse the numbers which seems like a good idea?)

Then it'd be good to have some unit tests for that. Actually, I will try to do a prep PR to add unit tests for some other code in this file just to start that ball.

cgwalters (commented):
A few comments, thanks again for this work!

igoropaniuk:
@cgwalters Thank you for your review. I have addressed all your recently posted comments. Please let me know if there's anything else I need to do.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1797,14 +1798,35 @@ parse_os_release (const char *contents, const char *split)
...
 }
 
+static guint
+bootloader_get_max_boot_tries (OstreeSysroot *self, GCancellable *cancellable, GError **error)
+{
+  g_autoptr (OstreeRepo) repo = NULL;
+  if (!ostree_sysroot_get_repo (self, &repo, cancellable, error))
+    return 0;
+
+  return repo->boot_counting;
```
cgwalters, line 1808: This function doesn't adhere to the GError conventions (I know, it's messy) - or at least the caller down below just ignores a return value of `0` and continues (potentially leaving the `error` set).

But I think there's a simple solution; we don't need this wrapper function at all. It's safe to just deref from `self->repo->boot_counting` in callers (if the sysroot is loaded).

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1797,14 +1798,35 @@ parse_os_release (const char *contents, const char *split)
...
+}
+
+static gboolean
+bootloader_is_boot_count_enabled (OstreeSysroot *self, GCancellable *cancellable, GError **error)
```
cgwalters, line 1812: Ditto for this

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1797,14 +1798,35 @@ parse_os_release (const char *contents, const char *split)
...
 /* Generate the filename we will use in /boot/loader/entries for this deployment.
  * The provided n_deployments should be the total number of target deployments (which
  * might be different from the cached value in the sysroot).
  */
 static char *
 bootloader_entry_filename (OstreeSysroot *sysroot, guint n_deployments,
-                           OstreeDeployment *deployment)
+                           OstreeDeployment *deployment, GCancellable *cancellable, GError **error)
 {
+  g_autofree char *bootconf_name;
```
cgwalters, line 1829: Initialize to `NULL` for safety

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1797,14 +1798,35 @@ parse_os_release (const char *contents, const char *split)
...
+}
+
 /* Generate the filename we will use in /boot/loader/entries for this deployment.
  * The provided n_deployments should be the total number of target deployments (which
  * might be different from the cached value in the sysroot).
  */
 static char *
 bootloader_entry_filename (OstreeSysroot *sysroot, guint n_deployments,
-                           OstreeDeployment *deployment)
+                           OstreeDeployment *deployment, GCancellable *cancellable, GError **error)
```
cgwalters, line 1827: After we drop the fallibility (use of GError) from those helpers we don't need this change either.

src/libostree/ostree-bootconfig-parser.c
```diff
@@ -56,6 +59,96 @@ ostree_bootconfig_parser_clone (OstreeBootconfigParser *self)
...
+          /* Look for the last "+" symbol in the filename */
+          counter = plus + 1;
+          continue;
+        }
+      if (counter)
+        break;
+
+      /* No boot counter found */
+      return;
+    }
```
cgwalters, line 92: I had to stare at this a bit...I think it's correct as is, but wouldn't it be a lot clearer to use `strrchr ()` here (note the extra `r` to search from the end)?

i.e.
```
const char *counter = strrchr (filename, '+');
if (!counter)
  return;
counter += 1;
```

src/libostree/ostree-bootconfig-parser.c
```diff
@@ -56,6 +59,96 @@ ostree_bootconfig_parser_clone (OstreeBootconfigParser *self)
...
+ * @path: File path
+ *
+ * Parses a suffix of two counters in the form "+LEFT-DONE" from the end of the
+ * filename (excluding file extension).
+ */
+static void
+parse_bootloader_tries (const char *filename, guint64 *out_left, guint64 *out_done, GError **error)
+{
+  gchar *counter = NULL;
+  gchar *old_counter = NULL;
```
cgwalters, line 74: Since these are pointing into `filename` they should also be `const`

src/libostree/ostree-bootconfig-parser.c
```diff
@@ -56,6 +59,96 @@ ostree_bootconfig_parser_clone (OstreeBootconfigParser *self)
...
+        break;
+
+      /* No boot counter found */
+      return;
+    }
+
+  guint64 tries_left, tries_done = 0;
+
+  old_counter = counter;
+  tries_left = g_ascii_strtoull (old_counter, &counter, 10);
```
cgwalters, line 97: I think this is also OK but it would seem clearer to me to leave the value of `counter` pointing where it is and then assign to a new `tries` variable? Like:

```
const char *tries = NULL;
tries_left = g_ascii_strtoull (counter, &tries, 10);
...
```
or so?

cgwalters (changes requested):
Thanks so much for working on this! I added some comments but I don't mind helping out myself by pushing updates here if you prefer, let me know!

One thing that I think would be helpful is some new unit tests, but I am also happy to do that.

src/libostree/ostree-sysroot-deploy.c
igoropaniuk: removed this function

src/libostree/ostree-sysroot-deploy.c
igoropaniuk: removed this function

src/libostree/ostree-sysroot-deploy.c
igoropaniuk: done

src/libostree/ostree-sysroot-deploy.c
igoropaniuk: sure, dropped these params

src/libostree/ostree-bootconfig-parser.c
igoropaniuk: fixed, thanks!

src/libostree/ostree-bootconfig-parser.c
igoropaniuk: I quickly looked into that and created a prototype, but it seems to only make the code more complex IMO

igoropaniuk:
@cgwalters Apologies for the delay. I've addressed all the comments except the last one concerning the "usage" variable.

Could we get this PR to land first (as it seems closer to being merged in terms of its maturity), and then I'll focus on addressing all issues in the other one (https://github.com/ostreedev/ostree/pull/3310) afterwards?

cgwalters:
OK I

- Addressed the gemini review comments
- Added unit tests for the tries parsing
- Added a unit-admin test that verifies we write the bootloader entry files w/boot counting initalized

src/libostree/libostree-devel.sym
```diff
@@ -1,16 +1,16 @@
 /* DEVEL symbol file - add new symbols here
   Copyright (C) 2016 Colin Walters <walters@verbum.org>
- 
+
   This library is free software; you can redistribute it and/or
   modify it under the terms of the GNU Lesser General Public
   License as published by the Free Software Foundation; either
   version 2 of the License, or (at your option) any later version.
- 
+
```
cgwalters: these are spurious changes, we should remove them

src/libostree/ostree-bootconfig-parser.h
```diff
@@ -67,4 +67,13 @@ void ostree_bootconfig_parser_set_overlay_initrds (OstreeBootconfigParser *self,
...
+
+_OSTREE_PUBLIC
```
cgwalters, line 73: Why are these public APIs? I'm not saying it's wrong, but it's not immediately obvious to me why they should be.

Well, I guess actually we probably should be rendering this in e.g. `ostree admin status` at least.

But right now only `_tries_left` is listed in the symbols file, so if we're making these public we should ensure that `_tries_done` is too.

src/libostree/ostree-bootconfig-parser.c
```diff
@@ -56,6 +59,87 @@ ostree_bootconfig_parser_clone (OstreeBootconfigParser *self)
...
+}
+
+/**
+ * ostree_bootconfig_parser_is_parsed:
+ * @self: Parser
+ *
+ * Returns: TRUE if the boot config was parsed from existing boot config file
+ */
+gboolean
+ostree_bootconfig_parser_is_parsed (OstreeBootconfigParser *self)
```
cgwalters: Hmm how about storing the original filename instead? i.e. `ostree_bootconfig_parser_get_filename`. It seems more generally useful and easy to describe than what "is_parsed" means.

cgwalters (changes requested):
Forgot to also mention I went ahead and rebased this on git main.

I have just a few other minor comments but I think we can get this in by tomorrow - what do you think?

igoropaniuk:
@cgwalters, it seems we are simultaneously working to address comments.
Would you like me to finish addressing all the issues, or would you like to handle that on your own?

cgwalters:
> @cgwalters, it seems we are simultaneously working to address comments :)

Ahhh sorry! Based on your location I thought you'd be sleeping now :smile: 

OK I did just one more force push but I will hand over the PR to you! What do you think of the changes?

igoropaniuk:
@cgwalters 

> Ahhh sorry! Based on your location I thought you'd be sleeping now 😄
> OK I did just one more force push but I will hand over the PR to you! What do you think of the changes?

Great, thanks for covering this functionality with unit tests and addressing the reported issues by Gemini!
Please let me know if you have any additional comments; I'll review them tomorrow morning and try to address them right away + address build issues reported by CI as well

cgwalters:
Offhand I don't have any other issues!

It looks like I broke CI by adding `ostree_bootconfig_parser_get_tries_done` to public API without also listing it in `apidoc/` - that should be an easy fix.

igoropaniuk:
> It looks like I broke CI by adding ostree_bootconfig_parser_get_tries_done to public API without also listing it in apidoc/ - that should be an easy fix.

Fixed, thanks for the hint!

cgwalters:
/override ci/prow/fcos-e2e
I'll probably drop this one as it's always duplicated coreos ci jenkins

PR #3311, opened 2024-09-19
curl: Add more assertions for curl return values

cgwalters:
Followup to the previous curl fixes; if we'd had an assertion earlier debugging the failure would have been more obvious.

All of these are "should not fail" cases so asserting is right.


ericcurtin:
Sometimes I prefer using exec(curl ...) rather than libcurl as I find it easier to debug and more memory safe, I guess OSTree does so many requests the forking gets too expensive, LGTM...

jlebon (approved):

Issue #3315, opened 2024-09-28
Request a kernel commandline for `sysroot.readonly`


ruihe774:
I request a kernel commandline to override `sysroot.readonly`, just like the existing `ostree.prepare-root.composefs`.

My Fedora Silverblue 40 becomes unbootable after upgrading to Fedora Silverblue 41. Because my disk is not writable during boot (I'm using TCG OPAL, FWIW), I have to use `ro` in kernel commandline. `sysroot.readonly=true` requires `rw`, so in f40 I simply set it to false in `/ostree/repo/config`. However, f41 bakes `sysroot.readonly=true` into `/usr/lib/ostree/prepare-root.conf` in the initramfs, which has precedence over `/ostree/repo/config` and makes my system unbootable. Because the initramfs in Silverblue is not generated in my local machine but provided by ostree repo, I have no way to modify `prepare-root.conf` in it.

Because I cannot change the initramfs, I think the only way is to provide a kernel commandline to override it. I'd appreciate it if it could be taken into consideration and be fixed before f41 releases.

ruihe774:
Another solution is to allow `sysroot.readonly=true` with `ro` and in this case postpone the bind mount of `/etc` and `/var` to ostree-remount.

cgwalters:
> Because my disk is not writable during boot (I'm using TCG OPAL, FWIW),

Can you explain a bit more? I only skimmed the spec and such but in your use case, how is the key provided? Do you enter a passphrase during boot, like LUKS?

cgwalters:
> Another solution is to allow sysroot.readonly=true with ro and in this case postpone the bind mount of /etc and /var to ostree-remount.

No, we learned that's just full of race conditions. The rootfs needs to be fully set up in the initramfs before we switch to it.

In your case is the drive really only set to writable mode during the real root boot, it isn't done in the initramfs? If not, can it be?

ruihe774:
> Can you explain a bit more? I only skimmed the spec and such but in your use case, how is the key provided? Do you enter a passphrase during boot, like LUKS?
>
> In your case is the drive really only set to writable mode during the real root boot, it isn't done in the initramfs? If not, can it be?

[...]

PR #3316, opened 2024-09-28
prepare-root: allow `sysroot.readonly=true` with kernel cmdline `ro`

ruihe774:
As described in #3315.

### DIFF
--- a/src/switchroot/ostree-mount-util.h
+++ b/src/switchroot/ostree-mount-util.h
@@ -36,10 +36,6 @@
 #define _OSTREE_SYSROOT_READONLY_STAMP "/run/ostree-sysroot-ro.stamp"
 #define _OSTREE_COMPOSEFS_ROOT_STAMP "/run/ostree-composefs-root.stamp"
 
-#define OTCORE_SYSROOT_NOT_WRITEABLE \
-  "sysroot.readonly=true requires %s to be writable at this point, the cmdline should contain rw " \
-  "but not ro, if that is not the case this is likely the issue"
-
 #define autofree __attribute__ ((cleanup (cleanup_free)))
 
 static inline int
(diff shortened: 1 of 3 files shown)


cgwalters:
I'm not opposed to this, but I'd like to try a bit harder to reach another design where we can mount the root writable in the initramfs. Let's debate that in the issue https://github.com/ostreedev/ostree/issues/3315 I think to keep it in once place.

cgwalters (approved):

ruihe774:
Does `MS_BIND | MS_REMOUNT` create readonly bind-mount for readonly source dir? If not, it is necessary to check `sysroot_currently_writable` and add `MS_RDONLY` when performing bind-mounts.

ruihe774:
Are there test cases that cover things in initramfs? I think I'd better add some tests to prevent regression. Especially, I have not played with root_transient, etc_transient, and using_composefs yet.

cgwalters:
> Ok. It just works. Feel free to merge it.

Yeah, bind mounts will inherit the writability by default.

> Are there test cases that cover things in initramfs? I think I'd better add some tests to prevent regression. 

There's definitely a coverage gap in some cases; we only test some of this implicitly. I've been meaning to try to make it easier to run ostree-prepare-root outside of a proper initramfs (and boot process), testing it as a container which would dramatically simplify complexity for testing.

> Especially, I have not tried ro with root_transient, etc_transient, and using_composefs yet.

I think though your use case in the future is going to be better addressed by adding integrity and measurement at the OS level and not trying to enforce it via drive-specific writability - which gets into the composefs path specifically so we'd effectively be obsoleting the case of "ro with composefs we mount writable later" for example.

ruihe774:
Thx

Issue #2568, opened 2022-03-18
What if char devices are being used as regular files?


petersenna:
I am trying to commit Podman's local storage to OSTree. Now the issue is that Overlayfs uses char devices to represent specific states. From: https://www.kernel.org/doc/Documentation/filesystems/overlayfs.txt
--
whiteouts and opaque directories
--------------------------------
In order to support rm and rmdir without changing the lower
filesystem, an overlay filesystem needs to record in the upper filesystem
that files have been removed.  This is done using whiteouts and opaque
directories (non-directories are always opaque).

A whiteout is created as a character device with 0/0 device number.
When a whiteout is found in the upper level of a merged directory, any
matching name in the lower level is ignored, and the whiteout itself
is also hidden.
--

And now I am stuck because OSTree wants only regular files and links, and trying to commit a tree that contains char devices does not work.

I am trying to convince OSTree that, hey these char devices are actually regular files, because they represent the state of the filesystem. But OSTree did not listen to me. Do I have any hope to use OSTree to also commit char devices for cases such as this one?

cgwalters:
Hi, this is related to https://github.com/ostreedev/ostree-rs-ext/issues/246

I see the desire to simply ship podman's storage, but trying to change ostree to accept character devices at this time would be a large change.  In the end, inherently the "ostree nature" of e.g. not shipping timestamps will also leak through into the containers, so it's really better for us to handle it by flattening I think.

cgwalters:
OK I looked at this briefly but it touches some of the most complex and hairy code in ostree...the `write_quoted_device` bit is just a stub and we need to handle all sorts of things like xattrs, uid/gid in a consistent way there but still do the "quoting". We'd need to define an encoding (filename like `.ostree-quoted-device.[c|b].<major>.<minor>`) that we serialize/deserialize etc.

```
From fdfeb0ba7b45b821c23fd3bd21f386d93dea6296 Mon Sep 17 00:00:00 2001
From: Colin Walters <walters@verbum.org>
Date: Tue, 1 Oct 2024 17:07:17 -0400
Subject: [PATCH 1/2] checkout: Add commentary around whiteout "quoting"

Signed-off-by: Colin Walters <walters@verbum.org>
---
 src/libostree/ostree-repo-checkout.c | 13 ++++++++++---
 1 file changed, 10 insertions(+), 3 deletions(-)

diff --git a/src/libostree/ostree-repo-checkout.c b/src/libostree/ostree-repo-checkout.c
[...]

PR #3318, opened 2024-10-02
wip: support for quoting devices

cgwalters:
This generalizes the whiteout support to handle block/chardev and FIFOs.

Closes: https://github.com/ostreedev/ostree/issues/2568


cgwalters:
At this point it's not worth doing. No one who is using "pure ostree" is doing these nonsensical things of having these devices in ostree commits. It's coming from poorly built containers. And what we really want down the line is to go OCI+composefs native on the bootc side.

Issue #3319, opened 2024-10-04
OSTree seems to be causing issues with flatpak update?


AtiusAmy:
I got this error when running Flatpak update (Alpine Edge):

```
OSTree:ERROR:src/libostree/ostree-repo-commit.c:3191:_ostree_repo_commit_modifier_apply: code should not be reached
Bail out! OSTree:ERROR:src/libostree/ostree-repo-commit.c:3191:_ostree_repo_commit_modifier_apply: code should not be reached
```
It originally looked like [#3299](https://github.com/ostreedev/ostree/issues/3299), but on closer inspection it's different. I'm not sure if this is a problem with packaging or what, but I'm on Ostree 2024.8 (Flatpak 1.14.10)

adamthiede:
Could you provide more details? (user or system remotes for example) I'm not able to reproduce this on Alpine Edge.

dbnicholson:
@AtiusAmy could you try again with ostree main? Or, if you can apply the f11e6a4ae0ae5ed4e5c0bd466620f8947b8ee2b3 as a patch to your distro's ostree, that would be also work. The commit code is intending to get either a regular file, directory or symlink at that spot. I don't know why flatpak would be trying to commit an object that isn't one of those types. The other types are documented [here](https://docs.gtk.org/gio/enum.FileType.html).

AtiusAmy:
It seems like this is an issue with the Fedora flatpak remote. Flathub updates have no issue with this?

I uh have no idea how patch it though

dbnicholson:
That seems plausible as Flathub builds with flatpak-builder, and I don't think anything has changed there that would suddenly start committing non-files. Fedora's flatpaks are built by a separate process that converts RPMs. That seems more likely that some non-file object would slip through the process.

For Fedora, @cgwalters can suggest the best way to patch ostree.

cgwalters:
Is there a specific app that's failing to update?

AtiusAmy:
> Please be as specific as possible. Is this `runtime/org.fedoraproject.KDE5Platform/x86_64/f41` and/or `app/org.kde.krita/x86_64/stable` ?

I believe it was runtime/org.fedoraproject.KDE5Platform/x86_64/f40 which at the time Krita depended on

PR #3322, opened 2024-10-10
commit: Give a better error message for unhandled file type

cgwalters:
xref https://github.com/ostreedev/ostree/issues/3319

It'd be useful to know what file type is being hit here; I believe this code path should be unreachable.


dbnicholson (approved):
Indeed. Would have been nice to have this there before.

PR #3323, opened 2024-10-18
deploy: Don't copy xattrs for devicetree

cgwalters:
xref: https://github.com/coreos/fedora-coreos-tracker/issues/1808

For the kernel/initramfs that we copy to `/boot`
we use an explicit relabeling today, ignoring the source SELinux
context.

When we added handling for devicetree it reuse the `copy_dir_recurse`
we have for `etc` handling, and that copied the source xattrs.

Let's ensure that the devicetree is also `boot_t` by *not* copying
xattrs and relying on the default labeling.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -184,8 +184,8 @@ install_into_boot (OstreeRepo *repo, OstreeSePolicy *sepolicy, int src_dfd, cons
 /* Copy ownership, mode, and xattrs from source directory to destination */
 static gboolean
 dirfd_copy_attributes_and_xattrs (int src_parent_dfd, const char *src_name, int src_dfd,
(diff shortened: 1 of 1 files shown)


cgwalters:
Only compile tested so far

HuijingHei:
Do testing on aarch64 with this patch, after finalize-staged, the `/boot/ostree/fedora-coreos-*/dtb` are labeled with `boot_t`.

with the patch:
```
$ sudo /usr/bin/ostree admin finalize-staged
$ sudo restorecon -vnr /boot/ostree/
```

without this patch:
```
$ sudo /usr/bin/ostree admin finalize-staged
$ sudo restorecon -vnr /boot/ostree/ | wc -l
1866
```

HuijingHei (approved):
LGTM, thank you for working on this.

PR #3326, opened 2024-10-27
deploy: Don't recompute verity checksums if not enabled

cgwalters:
This fixes a truly horrific performance bug when
composefs is enabled, but fsverity is not supported by the filesystem. We'd fall back to doing *userspace* checksumming of all files at deployment time which was absolutely not expected or required.

There's really an immense amount of technical debt here, such as the confusion between `ex-integity.composefs` vs the prepare-root config, how we handle "torn" states where some objects don't have verity enabled but some do, etc.

The ostree composefs state has two modes:

- signed: We need to enforce fsverity
- unsigned: Best effort resilience

So we fix this by making the deploy path to make verity "opportunistic" - if the ioctl gives us the data, then we add it to the composefs.

However, this code path is also invoked when we're computing the expected composefs digest to inject
as commit metadata, and *that* API must work regardless of whether the target repo has fsverity enabled as it may operate on a build server.

One lucky thing in all of this: When I went to add the "checkout composefs" API I added a stub `GVariant` for options extensibility, which we now use.

### DIFF
--- a/src/libostree/ostree-repo-checkout.c
+++ b/src/libostree/ostree-repo-checkout.c
@@ -1273,14 +1273,18 @@ compare_verity_digests (GVariant *metadata_composefs, const guchar *fsverity_dig
 /**
  * ostree_repo_checkout_composefs:
  * @self: A repo
- * @options: (nullable): Future expansion space; must currently be %NULL
+ * @options: (nullable): If non-NULL, must be a GVariant of type a{sv}. See below.
  * @destination_dfd: Parent directory fd
  * @destination_path: Filename
  * @checksum: OStree commit digest
  * @cancellable: Cancellable
  * @error: Error
  *
- * Create a composefs filesystem metadata blob from an OSTree commit.
+ * Create a composefs filesystem metadata blob from an OSTree commit. Supported
+ * options:
+ *
+ *  - verity: `u`: 0 = disabled, 1 = set if present on file, 2 = enabled; any other value is a fatal
+ * error
  */
 gboolean
 ostree_repo_checkout_composefs (OstreeRepo *self, GVariant *options, int destination_dfd,
@@ -1288,8 +1292,31 @@ ostree_repo_checkout_composefs (OstreeRepo *self, GVariant *options, int destina
                                 GCancellable *cancellable, GError **error)
 {
 #ifdef HAVE_COMPOSEFS
-  /* Force this for now */
-  g_assert (options == NULL);
+  OtTristate verity = OT_TRISTATE_YES;
+
+  if (options != NULL)
+    {
+      g_auto (GVariantDict) options_dict;
+      g_variant_dict_init (&options_dict, options);
+      guint32 verity_v = 0;
+      if (g_variant_dict_lookup (&options_dict, "verity", "u", &verity_v))
+        {
+          switch (verity_v)
+            {
+            case 0:
+              verity = OT_TRISTATE_NO;
+              break;
+            case 1:
+              verity = OT_TRISTATE_MAYBE;
+              break;
+            case 2:
+              verity = OT_TRISTATE_YES;
+              break;
+            default:
+              g_assert_not_reached ();
+            }
+        }
+    }
 
   g_auto (GLnxTmpfile) tmpf = {
     0,
@@ -1311,8 +1338,8 @@ ostree_repo_checkout_composefs (OstreeRepo *self, GVariant *options, int destina
 
   g_autoptr (OstreeComposefsTarget) target = ostree_composefs_target_new ();
 
-  if (!_ostree_repo_checkout_composefs (self, target, (OstreeRepoFile *)commit_root, cancellable,
-                                        error))
+  if (!_ostree_repo_checkout_composefs (self, verity, target, (OstreeRepoFile *)commit_root,
+                                        cancellable, error))
     return FALSE;
 
   g_autofree guchar *fsverity_digest = NULL;
--- a/src/libostree/ostree-repo-composefs.c
+++ b/src/libostree/ostree-repo-composefs.c
@@ -265,9 +265,9 @@ _ostree_composefs_set_xattrs (struct lcfs_node_s *node, GVariant *xattrs, GCance
 }
 
 static gboolean
-checkout_one_composefs_file_at (OstreeRepo *repo, const char *checksum, struct lcfs_node_s *parent,
-                                const char *destination_name, GCancellable *cancellable,
-                                GError **error)
+checkout_one_composefs_file_at (OstreeRepo *repo, OtTristate verity, const char *checksum,
+                                struct lcfs_node_s *parent, const char *destination_name,
+                                GCancellable *cancellable, GError **error)
 {
   g_autoptr (GInputStream) input = NULL;
   g_autoptr (GVariant) xattrs = NULL;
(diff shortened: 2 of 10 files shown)


src/libostree/ostree-repo-checkout.c
```diff
@@ -1273,23 +1273,50 @@ compare_verity_digests (GVariant *metadata_composefs, const guchar *fsverity_dig
...
+              verity = OT_TRISTATE_NO;
+              break;
+            case 1:
+              verity = OT_TRISTATE_MAYBE;
+              break;
+            case 2:
+              verity = OT_TRISTATE_YES;
+              break;
+            default:
+              g_assert_not_reached ();
```
allisonkarlitskaya, line 1316: Are you sure?  Where is this dict coming from?  Probably this needs some better error handling than just aborting...

src/libostree/ostree-repo-checkout.c
```diff
@@ -1311,17 +1338,20 @@ ostree_repo_checkout_composefs (OstreeRepo *self, GVariant *options, int destina
...
   g_autofree guchar *fsverity_digest = NULL;
   if (!ostree_composefs_target_write (target, tmpf.fd, &fsverity_digest, cancellable, error))
     return FALSE;
 
   /* If the commit specified a composefs digest, verify it */
-  if (!compare_verity_digests (metadata_composefs, fsverity_digest, error))
-    return FALSE;
+  if (verity == OT_TRISTATE_YES)
+    {
+      if (!compare_verity_digests (metadata_composefs, fsverity_digest, error))
```
allisonkarlitskaya, line 1352: Wasn't this the case we always wanted to do, even if verity was disabled for the underlying files?

src/libostree/ostree-repo-composefs.c
```diff
@@ -322,30 +322,37 @@ checkout_one_composefs_file_at (OstreeRepo *repo, const char *checksum, struct l
...
-              && d->digest_size == OSTREE_SHA256_DIGEST_LEN
-              && d->digest_algorithm == FS_VERITY_HASH_ALG_SHA256)
-            known_digest = d->digest;
-        }
+#ifdef HAVE_LINUX_FSVERITY_H
+          /* First try to get the digest directly from the bare repo file.
+           * This is the typical case when we're pulled into the target
+           * system repo with verity on and are recreating the composefs
+           * image during deploy. */
+          char buf[sizeof (struct fsverity_digest) + OSTREE_SHA256_DIGEST_LEN];
```
allisonkarlitskaya, line 332: Can you move this into the block with the `ioctl`?  It has no meaning outside of that block.

allisonkarlitskaya: Oh.  I see.  `known_digest` potentially ends up pointing into this buffer.  Then *definitely* bring `known_digest` in to this block to reduce its scope to the same scope as `buf`.  In the original code they were at the same logical level (although one of them was inside an `#ifdef` and the other not).  After the change, they are in two separate `block { }s` which is definitely suspect.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -665,8 +665,23 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
...
-      if (!ostree_repo_checkout_composefs (repo, NULL, ret_deployment_dfd, OSTREE_COMPOSEFS_NAME,
-                                           csum, cancellable, error))
+      // TODO: Clean up our mess around composefs/fsverity...we have duplication
+      // between the repo config and the sysroot config, *and* we need to better
+      // handle skew between repo config and repo state (e.g. "post-copy" should
+      // support transitioning verity on and off in general).
+      // For now we configure things such that the fsverity digest is only added
+      // if present on disk in the unsigned case, and in the signed case unconditionally
+      // require it.
+      g_autoptr(GVariantBuilder) cfs_checkout_opts_builder = g_variant_builder_new (G_VARIANT_TYPE ("a{sv}"));
```
allisonkarlitskaya: You could use a stack-allocated `GVariantBuilder` for this case:

```c
  g_auto(GVariantBuilder) builder = G_VARIANT_BUILDER_INIT (G_VARIANT_TYPE_VARDICT);
```

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -665,8 +665,23 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
...
+      guint32 composefs_requested = 1;
+      if (composefs_config->is_signed)
+        composefs_requested = 2;
+      g_variant_builder_add (cfs_checkout_opts_builder, "{sv}", "verity",
+                               g_variant_new_uint32 (composefs_requested));
+      g_debug ("composefs requested: %u", composefs_requested);
+      g_autoptr (GVariant) cfs_checkout_opts = g_variant_builder_end (cfs_checkout_opts_builder);
```
allisonkarlitskaya, line 688: Be careful: the return value of `g_variant_builder_end()` is floating.  I was young and impressionable and `g_autoptr()` didn't exist yet...

src/ostree/ot-builtin-checkout.c
```diff
@@ -145,12 +148,18 @@ process_one_checkout (OstreeRepo *repo, const char *resolved_commit, const char
...
 
   /* If we're doing composefs, then this is it */
-  if (opt_composefs)
+  if (opt_composefs || opt_composefs_noverity)
     {
       if (new_options_set)
         return glnx_throw (error, "Specified options are incompatible with --composefs");
-      return ostree_repo_checkout_composefs (repo, NULL, AT_FDCWD, destination, resolved_commit,
-                                             cancellable, error);
+      g_autoptr (GVariantBuilder) checkout_opts_builder
```
allisonkarlitskaya, line 155: Same comments as above: stack-allocated goes here better.  Or `g_variant_new_parsed()`.  And be careful about the floating return value of `_end()`.

src/libostree/ostree-sysroot-deploy.c
allisonkarlitskaya: Or, honestly, just:

```c
g_variant_new_parsed("{'verity': <%u>}", composefs_requested);
```

or whatever the syntax there is.

allisonkarlitskaya (changes requested):
I don't have a good high-level picture of what this code is trying to do, but I have some C/GVariant-related comments.

src/libostree/ostree-repo-checkout.c
cgwalters: Yeah...a unit test wanted this but for bad reasons. I will fix it.

src/libostree/ostree-repo-composefs.c
cgwalters: Definitely good to fix, though since the variable wasn't *used* outside that inner block I don't think there's anything suspect in practice right now.

src/libostree/ostree-repo-checkout.c
cgwalters: The dict is coming from inside our own code (or in theory external code), but not reachable from external *input*. In most cases we abort in that situation today. (We used to `g_return_if_fail` but that basically becomes untenable with static analyzers which reasonably want to care about what happens in that runtime situation too)

src/ostree/ot-builtin-checkout.c
cgwalters: Switched to stack allocations

cgwalters:
One possible followup in https://github.com/ostreedev/ostree/issues/3327

cgwalters:
Re this code alex wrote most of it originally... @alexlarsson mind giving this a quick skim from a high level? But we can't always get his time.

src/libostree/ostree-repo-composefs.c
```diff
@@ -320,32 +320,38 @@ checkout_one_composefs_file_at (OstreeRepo *repo, const char *checksum, struct l
...
+                  && d->digest_algorithm == FS_VERITY_HASH_ALG_SHA256)
+                known_digest = d->digest;
+            }
 #endif
 
-      if (known_digest)
-        lcfs_node_set_fsverity_digest (node, known_digest);
-      else if (lcfs_node_set_fsverity_from_content (node, input, _composefs_read_cb) != 0)
-        return glnx_throw_errno (error);
+          if (known_digest)
```
cgwalters, line 346: Actually we may want to pass in a composefs config here even...because in the signed case we should really fail if we find we're missing fsverity here.

jmarrero (approved):
lgtm

cgwalters:
@allisonkarlitskaya you need to at least change your review status from "changes requested" to "approve"

allisonkarlitskaya (approved):

Issue #3330, opened 2024-10-30
composefs: Don't verify checksum if !target-fsverity


cgwalters:
A long time ago we changed rpm-ostree to default to adding composefs metadata. I hadn't noticed at the time how expensive that is if the build system doesn't have verity enabled which is going to be the common case, which is its own issue. But build time performance here isn't as relevant as client side, and the semantics.

In https://github.com/ostreedev/ostree/pull/3326 I changed things so we only add the verity digest in the generated composefs if we have it.

In the original version of that patch that broke `checkout --composefs` in the !verity case when we have an expected digest in the commit (as every rpm-ostree generated commit will have).

Obviously bigger picture we have a mess here because the bootc based testing isn't using this flow as it's doing a client side commit and deployment from that.

In the immediate short term I think I want to back out adding the composefs metadata to rpm-ostree by default, but that's a different thing.

On the ostree side here I think we should just stop checking the generated composefs when we're !verity because there's no integrity here.

I'm thinking we should never do "add fsverity if it exists" into the composefs as that previous PR was doing; either it's enabled or it's not, and half-measures are broken.

PR #3331, opened 2024-10-30
checkout: Only verify digest if repo requires fsverity

cgwalters:
Fixes a regression from the previous commit; in
the case where the target repo doesn't have composefs in signed mode there's no reason to verify the digest at checkout time because we aren't verifying it at boot time either.

The regression is in cases that use rpm-ostree e.g. where as of recently we unconditionally add the composefs digest, but for e.g. FCOS we aren't deploying with fsverity enabled.

Closes: https://github.com/ostreedev/ostree/issues/3330


cgwalters:
For the record apparently the reason this got through FCOS CI is because https://github.com/ostreedev/ostree/pull/3326 landed literally *just before* the switch to f41 where composefs is enabled by default there.

But CI on this PR should run through with composefs enabled for FCOS.

jmarrero (approved):
/lgtm

jmarrero (changes requested):
code lgtm, but still test is complaining.

cgwalters:
Some of the test failures here are unrelated to this PR, it just also happens to be the first one after the FCOS default switch.

I put one test update in https://github.com/ostreedev/ostree/pull/3332

jmarrero (approved):
ok, lgtm. Looking at the test next.

PR #3333, opened 2024-10-31
tests: Work around GPG 2.2.45 error behaviour when revoking an expired key

smcv:
In GPG 2.2.45, a diagnostic message about the only trusted key having already expired causes this import to produce exit status 2, but the import still succeeds (the key is still revoked).

Bug: https://dev.gnupg.org/T7351  
Bug-Debian: https://bugs.debian.org/1086140


cgwalters (approved):
Thanks!

PR #3334, opened 2024-11-04
prepare-root: Fix composefs docs

cgwalters:
In practice in ostree-sysroot-deploy.c we only react to having `composefs = yes`; the docs mention `maybe` but that never did anything.

The value is wrong in the code too, but I'm not touching that here to avoid conflating changes - the main thing to fix is the docs because here `maybe == no`.


jmarrero (approved):
lgtm

Issue #3339, opened 2024-11-14
FS_IOC_MEASURE_VERITY ioctl can read into an unaligned buffer


smcv:
While updating from 2024.8 to 2024.9 in Debian, I noticed that the composefs code saves the result of an FS_IOC_MEASURE_VERITY ioctl into `char buf[sizeof (struct fsverity_digest) + OSTREE_SHA256_DIGEST_LEN]`, and dereferences that pointer via `(struct fsverity_digest *)&buf`.

If I understand alignment requirements correctly, that cast is not necessarily safe, because `struct fsverity_digest` starts with a `__u16` which (on at least some architectures) requires "natural" 2-byte alignment, but a `char[]` is not guaranteed to have any specific alignment.

This might also violate Standard C aliasing rules (I think optimizing compilers are allowed to assume that a `struct fsverity_digest *` does not alias an array of a different type), but I'm not 100% sure whether the exception that is made in the standard for `char *` applies here.

Possible solutions that I can see:

1. use a `union { struct fsverity_digest; char[...]; }` which is aligned to the stricter requirement of the `struct fsverity_digest` but has enough space for the larger `char[]`, which solves both the alignment and the aliasing
2. use the `char[]`, but with `alignas(struct fsverity_digest)` to ensure it is correctly aligned; this does not solve the aliasing but perhaps that isn't a problem because `char *` has special rules
3. have the ioctl write into a `char[]`, then `memcpy` out the `struct fsverity_digest` into a correctly-aligned variable

The union seems simplest and most obviously-correct.

cgwalters:
> While updating from 2024.8 to 2024.9 in Debian, I noticed that the composefs code

But were you running a static analyzer? Just perusing `git diff`?

smcv:
> Just perusing git diff?

Yes. This is a dependency of the component I'm actually interested in (Flatpak), so I try to be a responsible maintainer, but there's a rather small limit to how much time and resources I can spend on libostree - especially the parts that Flatpak doesn't make use of.

libcompose is an optional dependency of a dependency of the component I'm actually interested in, and as far as I can tell is only useful for use-cases that are orthogonal to the one I'm interested in, so I don't have the time and resources available to package it for Debian. Sorry, I wish I had more hours in a day, but I don't!

cgwalters:
> Yes. This is a dependency of the component I'm actually interested in (Flatpak), so I try to be a responsible maintainer,

And thank you for all your work there!

[...]

PR #3340, opened 2024-11-14
composefs: Ensure buffer is suitably aligned for struct fsverity_digest

smcv:
struct fsverity_digest starts with a __u16, so it will normally require 16-bit alignment, which is not guaranteed for a char array.

Resolves: https://github.com/ostreedev/ostree/issues/3339

---

This is untested: sorry, I don't currently have any filesystems with fsverity enabled.

### DIFF
--- a/src/libostree/ostree-repo-composefs.c
+++ b/src/libostree/ostree-repo-composefs.c
@@ -327,19 +327,22 @@ checkout_one_composefs_file_at (OstreeRepo *repo, OtTristate verity, const char 
            * This is the typical case when we're pulled into the target
            * system repo with verity on and are recreating the composefs
            * image during deploy. */
-          char buf[sizeof (struct fsverity_digest) + OSTREE_SHA256_DIGEST_LEN];
+          union
+            {
+              struct fsverity_digest d;
+              char buf[sizeof (struct fsverity_digest) + OSTREE_SHA256_DIGEST_LEN];
+            } result;
           guchar *known_digest = NULL;
 
           if (G_IS_UNIX_INPUT_STREAM (input))
             {
               int content_fd = g_unix_input_stream_get_fd (G_UNIX_INPUT_STREAM (input));
-              struct fsverity_digest *d = (struct fsverity_digest *)&buf;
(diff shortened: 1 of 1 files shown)


smcv:
On closer inspection, I don't have libcomposefs in my build environment either ([we don't have it in Debian yet](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1064457) and I don't have enough mental bandwidth to be responsible for it) so I'm doing this completely blind. (Thanks, CI!)

cgwalters:
Thanks for this! What I'd hoped to do is land https://github.com/containers/composefs/pull/394 and then we can switch to it.

Now the point that arises here is that the libcomposefs code is doing the same thing.

Of course fsverity-utils has a different copy of this code; it seems to heap allocate https://github.com/ebiggers/fsverity-utils/blob/2543e6e5037c0b1b448282c2793f8a6ce8b12d71/programs/cmd_measure.c#L47
which I think means it's always aligned properly.

Hmm anyways I think you're right. So I will also update the composefs code.

cgwalters:
> This is untested: sorry, I don't currently have any filesystems with fsverity enabled.

It's pretty easy to make a temporary loopback, no? Not to mention temporary VMs.

cgwalters (approved):

cgwalters:
We also should double check the alignment for https://github.com/containers/composefs-rs/blob/7571eb0dd5789f84cfb1515e0a0b4f3ab603f17d/src/fsverity/ioctl.rs#L60

smcv:
> it seems to heap allocate

As per Standard C, that should be sufficiently well-aligned for any built-in type, so that's fine. It's stack and static allocations that are the usual problem with alignment.

> It's pretty easy to make a temporary loopback, no? Not to mention temporary VMs.

I'm sure that's pretty easy if you have an immutable/image-based OS and a basic knowledge of how fsverity works; but, sorry, I have neither of those things, and my time is already spread more thinly than I can really cope with, so there's a pretty low limit to how much "can't you just" I am able to add here.

Issue #3343, opened 2024-11-26
SELinux relabling during commit causes the wrong content hash to be stored if ACLs or user xattrs are present


prydom:
Putting a pointer to https://github.com/containers/bootc/issues/920#issuecomment-2501967928 here.

I originally filed in bootc since I thought the bug was in the ostree-ext-rs crate but based on the following analysis I think the issue is in OSTree.

> I think I can answer why this happens for the "pure ostree" case, @cgwalters. It seems the root cause has been in https://github.com/ostreedev/ostree for a while and it may just be some coincidental changes in libdnf5, librpm or librepo which caused this to start failing container pulls now. 
> 
> libglnx will always sort the extended attributes returned from `glnx_dfd_name_get_all_xattrs` with [`canonicalize_xattrs()`](https://github.com/GNOME/libglnx/blob/87f2768fab168ded186bd6b973ad6acf4cd09f8b/glnx-xattrs.c#L44). This ensures we should always get the same content hash as long as the same file contents + uid/gid/perms/xattrs match even if the underlying filesystem returns a different order.
> 
> However, during commit the way [`get_final_xattrs()`](https://github.com/ostreedev/ostree/blob/111a45fa5c3d115074d80b0e01f419969221fb63/src/libostree/ostree-repo-commit.c#L3226) works is to first delete the `security.selinux` attribute with `_ostree_filter_selinux_xattr()` and then append the new label to the end of the gvariant array.
> 
> Linux capabilities have been working because "security.capabilities" always comes before "security.selinux". However the existence of any "user.*" attributes means that list will no longer be sorted and therefore the hash calculated at commit time will not match what is stored in the bare repo and what is copied into the encapsulated commit container image.
> 
> So, preserving `system.*` (e.g. ACLs) or `trusted.*` extended attributes would be bit by this same bug. I have reproduced the same failure to pull the container by adding an ACL.
> 
> ```bash
> # Create a new commit with just a single file with an extended attribute
> mkdir commit
> cd commit
> echo "hello world" > test.txt
> setfattr -n security.selinux -v "system_u:object_r:etc_runtime_t:s0" test.txt
> setfacl -m "u:root:rw-" test.txt
> setcap 'cap_net_bind_service=ep' test.txt
> getfattr -d -m - -- *
> stat test.txt
> cd ..
> ```

cgwalters:
[...]

PR #3346, opened 2024-11-27
core: Always sort incoming xattrs

cgwalters:
When recomputing selinux attrs during commit, we weren't sorting, which could cause various issues like fsck failures.

This is a big hammer; change things so we always canonicalize (i.e. sort) the incoming xattrs when creating a file header.

I think almost all places in the code were already keeping things sorted, but it's better to ensure correctness first. If we ever have some performance issue (I'm doubtful) we could add something like `_ostree_file_header_known_canonicalized` or so.

Closes: https://github.com/ostreedev/ostree/issues/3343

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -43,6 +43,7 @@ G_STATIC_ASSERT (OSTREE_REPO_MODE_BARE_USER_ONLY == 3);
 G_STATIC_ASSERT (OSTREE_REPO_MODE_BARE_SPLIT_XATTRS == 4);
 
 static GBytes *variant_to_lenprefixed_buffer (GVariant *variant);
+static GVariant *canonicalize_xattrs (GVariant *xattrs);
 
 #define ALIGN_VALUE(this, boundary) \
   ((((unsigned long)(this)) + (((unsigned long)(boundary)) - 1)) \
@@ -302,6 +303,47 @@ ostree_validate_collection_id (const char *collection_id, GError **error)
   return TRUE;
 }
 
+static int
+compare_xattrs (const void *a_pp, const void *b_pp)
+{
+  GVariant *a = *(GVariant **)a_pp;
+  GVariant *b = *(GVariant **)b_pp;
+  const char *name_a;
+  const char *name_b;
+  g_variant_get (a, "(^&ay@ay)", &name_a, NULL);
+  g_variant_get (b, "(^&ay@ay)", &name_b, NULL);
+
+  return strcmp (name_a, name_b);
+}
+
+// Sort xattrs by name
+static GVariant *
+canonicalize_xattrs (GVariant *xattrs)
+{
+  // We always need to provide data, so NULL is canonicalized to the empty array
+  if (xattrs == NULL)
+    return g_variant_ref_sink (g_variant_new_array (G_VARIANT_TYPE ("(ayay)"), NULL, 0));
+
+  g_autoptr (GPtrArray) xattr_array
+      = g_ptr_array_new_with_free_func ((GDestroyNotify)g_variant_unref);
+
+  const guint n = g_variant_n_children (xattrs);
+  for (guint i = 0; i < n; i++)
+    {
+      GVariant *child = g_variant_get_child_value (xattrs, i);
+      g_ptr_array_add (xattr_array, child);
+    }
+
+  g_ptr_array_sort (xattr_array, compare_xattrs);
+
+  GVariantBuilder builder;
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("a(ayay)"));
+  for (guint i = 0; i < xattr_array->len; i++)
+    g_variant_builder_add_value (&builder, xattr_array->pdata[i]);
+
+  return g_variant_ref_sink (g_variant_builder_end (&builder));
+}
+
 /* The file header is part of the "object stream" format
  * that's not compressed.  It's comprised of uid,gid,mode,
  * and possibly symlink targets from @file_info, as well
@@ -321,13 +363,12 @@ _ostree_file_header_new (GFileInfo *file_info, GVariant *xattrs)
   else
     symlink_target = "";
 
-  g_autoptr (GVariant) tmp_xattrs = NULL;
-  if (xattrs == NULL)
-    tmp_xattrs = g_variant_ref_sink (g_variant_new_array (G_VARIANT_TYPE ("(ayay)"), NULL, 0));
+  // We always sort the xattrs now to ensure everything is in normal/canonical form.
+  g_autoptr (GVariant) tmp_xattrs = canonicalize_xattrs (xattrs);
 
   g_autoptr (GVariant) ret
       = g_variant_new ("(uuuus@a(ayay))", GUINT32_TO_BE (uid), GUINT32_TO_BE (gid),
-                       GUINT32_TO_BE (mode), 0, symlink_target, xattrs ?: tmp_xattrs);
+                       GUINT32_TO_BE (mode), 0, symlink_target, tmp_xattrs);
   return variant_to_lenprefixed_buffer (g_variant_ref_sink (ret));
 }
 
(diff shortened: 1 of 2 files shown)


tests/test-basic-c.c
```diff
@@ -193,12 +201,49 @@ test_object_writes (gconstpointer data)
...
+    g_variant_builder_init (&builder, G_VARIANT_TYPE ("a(ayay)"));
+    g_variant_builder_add (&builder, "(^ay^ay)", "security.selinux", "foo_t");
+    g_variant_builder_add (&builder, "(^ay^ay)", "user.blah", "somedata");
+    g_autoptr (GVariant) xattrs = g_variant_ref_sink (g_variant_builder_end (&builder));
+    ostree_repo_write_regfile_inline (repo, expected_sha256_with_xattrs, 0, 0, S_IFREG | 0644,
+                                      xattrs, (guint8 *)"hi", 2, NULL, &error);
+    g_assert_no_error (error);
+  }
+
+  /* And now with a swapped order */
```
cgwalters, line 237: And I also verified that this test successfully fails if we drop the xattr canonicalization.

prydom:
I get the following error when pulling an image into ostree layered on top of quay.io/fedora/fedora-kinoite. It has user.* xattrs in the container tar stream.

```
2024-11-27T18:39:25.0514353Z layers already present: 0; layers needed: 89 (4.5 GB)
2024-11-27T18:40:54.5755174Z error: Importing: Parsing layer blob sha256:4c7aee33c4535de34dfdde394e64174d0177144339e7d35846be7ce1663722f4: error: ostree-tar: Failed to handle directory: Incorrectly sorted xattr name, index=2: Processing tar: Failed to commit tar: ExitStatus(unix_wait_status(256))
```

cgwalters:
Thanks for testing this out! I pushed an update that also covers directory metadata - only unit tested here so far.

prydom:
I've done another round of end-to-end testing and the changes work well for me.

On github actions:
- Updated ostree-libs to c8635586c21714656a833ac31a8c86d6e1642901 on CI build container.
- Built multiple layers upon upstream bootable container image which introduced the librepo extended attributes in tar stream.
- Pulled container image into empty bare ostree repository (this step is what failed earlier)
- Checked out image with ostree, made some further changes that are not appropriate to do with image layering
- Checked image back into new ostree ref and used `--selinux-policy=` to relabel based on policy in image 
- Ran `rpm-ostree compose container-encapsulate` to re-encapsulate commit as image and upload to registry
- Added an additional container image layer with even more extended attributes.

In `podman` on dev machine:
- In build container (same as CI) with updated ostree, created empty bare-user repo. Pulled re-encapsulated image. `ostree fsck` passes.
- Pulled encapsulated image, run `ostree fsck` on the sysroot in the container (bare-split-xattrs). Passes.

On target machine:
- Used bootc to switch/rebase to new image. Existing host with bootc and the new image does not have updated ostree library. Image pulls successfully.
- Booted into staged image successfully.
- Confirmed that introduced user xattrs are preserved.
- `ostree fsck` run from the booted image on sysroot fails on files that have user xattrs. This is because the ostree library hasn't been updated in the image.
- Activate usr-overlay, install rpms with ostree and ostree-libs from this PR, create new bare repo, pull container into fresh repo and `ostree fsck` passes.

src/libostree/ostree-core.c
```diff
@@ -302,6 +303,47 @@ ostree_validate_collection_id (const char *collection_id, GError **error)
...
 }
 
+static int
+compare_xattrs (const void *a_pp, const void *b_pp)
+{
+  GVariant *a = *(GVariant **)a_pp;
+  GVariant *b = *(GVariant **)b_pp;
+  const char *name_a;
+  const char *name_b;
+  g_variant_get (a, "(^&ay@ay)", &name_a, NULL);
```
allisonkarlitskaya, line 313: Random: you could just `memcmp()` the data of the GVariants directly: the key starts at byte 0.  That's pretty evil, though.

Otherwise it might be a bit nicer to use `g_variant_get_child(a, 0, "&ay", &name_a)`

src/libostree/ostree-core.c
```diff
@@ -302,6 +303,47 @@ ostree_validate_collection_id (const char *collection_id, GError **error)
...
+  g_variant_get (b, "(^&ay@ay)", &name_b, NULL);
+
+  return strcmp (name_a, name_b);
+}
+
+// Sort xattrs by name
+static GVariant *
+canonicalize_xattrs (GVariant *xattrs)
```
allisonkarlitskaya, line 321: Yikes.  This function is complicated, but I honestly can't think of any other way to do it more simply...

One thing I'd consider is scanning the "input" to see if it's already in order (trivially true also for the common case of only one item) and return a ref directly in that case.

src/libostree/ostree-core.c
```diff
@@ -2278,21 +2321,30 @@ ostree_validate_structureof_file_mode (guint32 mode, GError **error)
...
 _ostree_validate_structureof_xattrs (GVariant *xattrs, GError **error)
 {
   const guint n = g_variant_n_children (xattrs);
+  const char *previous = NULL;
   for (guint i = 0; i < n; i++)
     {
-      const guint8 *name;
+      const char *name;
       g_autoptr (GVariant) value = NULL;
       g_variant_get_child (xattrs, i, "(^&ay@ay)", &name, &value);
```
allisonkarlitskaya, line 2336: Pulling the (unused) `value` out here just to have it auto-freed at the end of the loop is wasteful.  You could just give `NULL` there.  This whole loop could also be converted to a `GVariantIter`.

tests/test-basic-c.c
```diff
@@ -193,12 +201,67 @@ test_object_writes (gconstpointer data)
...
+  g_assert (!ostree_repo_write_regfile_inline (repo, invalid_hi_sha256, 0, 0, S_IFREG | 0644, NULL,
+                                               (guint8 *)"hi", 2, NULL, &error));
+  g_assert (error != NULL);
+  g_clear_error (&error);
+
+  static const char expected_sha256_with_xattrs[]
+      = "72473a9e8ace75f89f1504137cfb134feb5372bc1d97e04eb5300e24ad836153";
+
+  GVariantBuilder builder;
+  g_autoptr (GVariant) inorder_xattrs = NULL;
```
allisonkarlitskaya, line 226: Could move the declaration down to the same line as the assignment, like you do for `unsorted...`

tests/test-basic-c.c
```diff
@@ -193,12 +201,67 @@ test_object_writes (gconstpointer data)
...
+  g_variant_builder_add (&builder, "(^ay^ay)", "security.selinux", "foo_t");
+  g_variant_builder_add (&builder, "(^ay^ay)", "user.blah", "somedata");
+  inorder_xattrs = g_variant_ref_sink (g_variant_builder_end (&builder));
+  g_variant_builder_init (&builder, G_VARIANT_TYPE ("a(ayay)"));
+  g_variant_builder_add (&builder, "(^ay^ay)", "user.blah", "somedata");
+  g_variant_builder_add (&builder, "(^ay^ay)", "security.selinux", "foo_t");
+  g_autoptr (GVariant) unsorted_xattrs = g_variant_ref_sink (g_variant_builder_end (&builder));
+
+  /* Now test with xattrs */
+  ostree_repo_write_regfile_inline (repo, expected_sha256_with_xattrs, 0, 0, S_IFREG | 0644,
```
allisonkarlitskaya, line 237: Did you also want to check the return value?

tests/test-basic-c.c
```diff
@@ -193,12 +201,67 @@ test_object_writes (gconstpointer data)
...
+                                    inorder_xattrs, (guint8 *)"hi", 2, NULL, &error);
+  g_assert_no_error (error);
+
+  /* And now with a swapped order */
+  ostree_repo_write_regfile_inline (repo, expected_sha256_with_xattrs, 0, 0, S_IFREG | 0644,
+                                    unsorted_xattrs, (guint8 *)"hi", 2, NULL, &error);
+  g_assert_no_error (error);
+
```
allisonkarlitskaya, line 245: I notice that you don't have a testcase for what happens if someone tries to repeat an xattr.  If I understand the code above, you'd sort the incoming list, but fail to detect the duplicate, which would later fail to validate.

Do you want to add that check as well?

allisonkarlitskaya: You also don't check for incoming empty xattr key values.  That might be another nice thing to figure out, and then add a test for.

src/libostree/ostree-core.c
cgwalters: > Yikes. This function is complicated, but I honestly can't think of any other way to do it more simply...

Yeah...this is where GVariant in C specifically is really painful. At one point Alex was working on a "gvariant compiler" IIRC that generated accessors. In Rust there's https://docs.rs/gvariant/latest/gvariant/ but none of this code is going to be in Rust soon sadly.

tests/test-basic-c.c
cgwalters: Yeah, this normalization also raises the question of what happens in those edge cases. However, this code in the write path isn't returning errors today, doing so would be a notably bigger change.

We don't have active issues with duplicate or zero-key-length xattrs; one would have to explicitly craft things to make that happen today. I just added some sanity checking which is more intended for `fsck` type operations. I think we can leave it as a followup - (really first ensure it's fixed in composefs...which hmm, it looks like we're not sanity checking for empty xattr names there either in a quick look)

cgwalters: https://github.com/containers/composefs/pull/400

cgwalters:
OK, updated this one, should be good to go!

allisonkarlitskaya (approved):

PR #3351, opened 2024-12-06
man: Note semantics combining `root.transient` with `composefs.enabled`

cgwalters:
It's all quite confusing having to reason about both the pre-composefs ostree and the composefs version. But hopefully soon we more firmly leave behind that first legacy.


ruihe774:
Line 124 also needs to be modified:
> with the upper directory being a hidden directory (in the underlying system root filesystem) that will persist across reboots by default.

cgwalters:
> Line 124 also needs to be modified:
Thanks! Indeed. And now that I look more closely we actually just error out with `root.transient` when composefs isn't enabled; I think the doc was written about an earlier version and then we backed out of that.

Reworded this entirely.

jeckersb (approved):

Issue #2867, opened 2023-05-30
ex-integrity.composefs: Tracking issue


cgwalters:
# composefs/ostree (and beyond)

## Background

A key design goal of ostree at its creation was to not require any new functionality in the Linux kernel.  The baseline mechanisms of hard links and read-only bind mounts suffice to manage views of read-only filesystem trees.

However, for Docker and then podman, `overlayfs` was created to more efficiently support copy-on-write semantics - also crucially, overlayfs is a *layered* filesystem; it can work with any underlying (modern) Linux filesystem as a backend.

More recently, [composefs](https://github.com/containers/composefs) was created which builds on overlayfs with more integrity features.  This tracking issue is for the integration of composefs and ostree.

## System integrity

ostree does not provide significant support for truly immutable system state; a simple `mount -o remount,rw /usr` will allow direct persistent modification of the underlying files.

There is `ostree fsck`, but this is inefficient and manual, and further still today does not cover the checked-out deployment roots (so e.g. newly added binaries in the deployment root aren't found).

### Accidental damage protection

It is important to ostree to support "user owns machine" scenarios, where the user is root on their own computer and must have the ability to make persistent changes.

But it's still useful to have stronger protection against *accidental* damage.  Due to the way composefs works using [fs-verity](https://www.kernel.org/doc/html/next/filesystems/fsverity.html), a simple `mount -o remount,rw` can no longer silently modify files.  First, the mounted `composefs` is always read-only; there is no write support in composefs.  Access to the distinct underlying persistent root filesystem can be more strongly separated and isolated.

### Support for "sealed" systems

It's however also desirable to support a scenario where an organization wants to produce computing devices that are "sealed" to run only code produced (or signed) by that organization.  These devices should not support persistent unsigned code.

ostree does not have strong support for this model today, and composefs should fix it.

## Phase 0: Basic integration (experimental)

In this phase, we will land [an outstanding pull request](https://github.com/ostreedev/ostree/pull/2640) which adds basic integration that enables booting a system using composefs as a root filesystem.  In this phase, a composefs image is *dynamically created* on the client using the ostree metadata.

This has already led us to multiple systems integration issues.  So far, all tractable.

- https://github.com/coreos/fedora-coreos-config/pull/2404

A good milestone to mark completion of this phase is landing a CI configuration to ostree which builds and deploys a system using composefs, and verifies it can be upgraded.

In this phase, there is no direct claimed support for "sealed" systems (i.e. files are not necessarily signed).

[...]

PR #3353, opened 2024-12-11
libostree/deploy: enable composefs by default

jlebon:
The composefs libostree integration has been supported for a while now
and is actively in use in various ostree/bootc-based systems. Let's
turn it on by default.

This has no effect if composefs support is not compiled in. Note also
that this does not change the default value of the `composefs.enabled`
tristate to `true`. The default is still `maybe`, but the deploy API
will now also create composefs images for `maybe`.

The reason for doing it this way is so that systems upgrading from
old libostree versions (which may either not have composefs support or
may have composefs-related bugs) will still be able to upgrade and not
trip `ostree-prepare-root` in the new deployment (which allows missing
composefs images for `maybe`).

We may in the future change the default value to `true`.

See also: https://github.com/ostreedev/ostree/issues/2867

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -640,9 +640,6 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
   if (!glnx_opendirat (osdeploy_dfd, checkout_target_name, TRUE, &ret_deployment_dfd, error))
     return FALSE;
 
-  guint64 composefs_start_time = 0;
-  guint64 composefs_end_time = 0;
-#ifdef HAVE_COMPOSEFS
   /* TODO: Consider changing things in the future to parse the deployment config from memory, and
    * if composefs is enabled, then we can check out in "user mode" (i.e. only have suid binaries
    * enabled in composefs, etc.)
@@ -667,6 +664,10 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
   g_debug ("composefs enabled by config: %d repo: %d", composefs_enabled, repo->composefs_wanted);
   if (repo->composefs_wanted == OT_TRISTATE_YES)
     composefs_enabled = repo->composefs_wanted;
+
+  guint64 composefs_start_time = 0;
+  guint64 composefs_end_time = 0;
+#ifdef HAVE_COMPOSEFS
   if (composefs_enabled != OT_TRISTATE_NO)
     {
       composefs_start_time = g_get_monotonic_time ();
@@ -694,6 +695,9 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
     }
   else
     g_debug ("not using composefs");
+#else
+  if (composefs_enabled == OT_TRISTATE_YES)
+    return glnx_throw (error, "composefs: enabled at runtime, but support is not compiled in");
 #endif
 
   *checkout_elapsed = (checkout_end_time - checkout_start_time);
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -667,7 +667,7 @@ checkout_deployment_tree (OstreeSysroot *sysroot, OstreeRepo *repo, OstreeDeploy
   g_debug ("composefs enabled by config: %d repo: %d", composefs_enabled, repo->composefs_wanted);
   if (repo->composefs_wanted == OT_TRISTATE_YES)
     composefs_enabled = repo->composefs_wanted;
-  if (composefs_enabled == OT_TRISTATE_YES)
+  if (composefs_enabled != OT_TRISTATE_NO)
     {
       composefs_start_time = g_get_monotonic_time ();
       // TODO: Clean up our mess around composefs/fsverity...we have duplication
--- a/tests/admin-test.sh
+++ b/tests/admin-test.sh
@@ -71,9 +71,10 @@ assert_not_file_has_content status.txt "pending"
 assert_not_file_has_content status.txt "rollback"
 validate_bootloader
 
-# Someday probably soon we'll turn this on by default, but for now
-if test -f sysroot/ostree/deploy/testos/deploy/*.0/.ostree.cfs; then
-    fatal "found composefs unexpectedly"
+if has_ostree_feature composefs; then
+    if ! test -f sysroot/ostree/deploy/testos/deploy/*.0/.ostree.cfs; then
+        fatal "missing composefs"
+    fi
 fi
 
 # Test the bootable and linux keys
--- a/tests/test-admin-deploy-composefs.sh
+++ b/tests/test-admin-deploy-composefs.sh
@@ -26,14 +26,32 @@ skip_without_ostree_feature composefs
 # Exports OSTREE_SYSROOT so --sysroot not needed.
 setup_os_repository "archive" "syslinux"
 
+# check disablement
 cd osdata
 mkdir -p usr/lib/ostree
 cat > usr/lib/ostree/prepare-root.conf << 'EOF'
 [composefs]
+enabled=false
+EOF
+cd -
+
+${CMD_PREFIX} ostree --repo=${test_tmpdir}/testos-repo commit --add-metadata-string version=1.composefs -b testos/buildmain/x86_64-runtime osdata
+${CMD_PREFIX} ostree --repo=sysroot/ostree/repo pull-local --remote=testos testos-repo testos/buildmain/x86_64-runtime
+
+${CMD_PREFIX} ostree admin deploy --os=testos --karg=root=LABEL=foo --karg=testkarg=1 testos:testos/buildmain/x86_64-runtime
+if test -f sysroot/ostree/deploy/testos/deploy/*.0/.ostree.cfs; then
+    fatal "found composefs unexpectedly"
+fi
+
+# check explicit enablement
+cd osdata
+cat > usr/lib/ostree/prepare-root.conf << 'EOF'
+[composefs]
 enabled=true
 EOF
-${CMD_PREFIX} ostree --repo=${test_tmpdir}/testos-repo commit --add-metadata-string version=1.composefs -b testos/buildmain/x86_64-runtime
 cd -
+
+${CMD_PREFIX} ostree --repo=${test_tmpdir}/testos-repo commit --add-metadata-string version=1.composefs -b testos/buildmain/x86_64-runtime osdata
 ${CMD_PREFIX} ostree --repo=sysroot/ostree/repo pull-local --remote=testos testos-repo testos/buildmain/x86_64-runtime
 
 ${CMD_PREFIX} ostree admin deploy --os=testos --karg=root=LABEL=foo --karg=testkarg=1 testos:testos/buildmain/x86_64-runtime


jlebon:
Haven't tested this yet.

jlebon:
OK so after looking at this a bit more, this doesn't quite do what I thought. I was planning to use `enabled = maybe` in the config file, but `maybe` is already the default value if it's missing. So this has a much larger impact; i.e. all libostree clients with composefs compiled in and no config specified would now start creating and mounting composefs images. Which... maybe it has baked enough now? But we'd need to raise awareness.

If we don't think we're ready for that, then we might need another config knob to express what we want here.

cgwalters:
>  If you're compiling libostree today with composefs enabled, then you're already tracking this tech and possibly using it already. 

Maybe. I think it's probably fine to do that change and release note it heavily, but it *is* a semantic change.

>Without this, it's hard to have new nodes use composefs while still allowing old nodes with ancient libostree to upgrade to the same image (without composefs).

That said, how does patching the *current* ostree help at all for this problem? Unless we back port this patch to y-stream branches...but I dunno, that seems like a big task to do.

I think we need to investigate instead fixing this up post-deploy pre-reboot via code execution which we're already tracking in other venues (e.g. the discussions of using `podman run <target image> bootc install`) which already give us a way to ensure we're running the *new* OS image.

cgwalters (changes requested):
Marking as requested changes just to signal it's reviewed

cgwalters:
If we were to try to cherry pick this patch to older releases, it'd definitely also want https://github.com/ostreedev/ostree/pull/3326/commits/a6d07b6cc3a6030307c6e29aaf33944ff0ad8c8b for example

jlebon:
> > If you're compiling libostree today with composefs enabled, then you're already tracking this tech and possibly using it already.
> 
> Maybe. I think it's probably fine to do that change and release note it heavily, but it _is_ a semantic change.

OK cool, I'll test it in earnest. Definitely at least some tests need to be adjusted it seems.

> > Without this, it's hard to have new nodes use composefs while still allowing old nodes with ancient libostree to upgrade to the same image (without composefs).
> 
> That said, how does patching the _current_ ostree help at all for this problem? Unless we back port this patch to y-stream branches...but I dunno, that seems like a big task to do.

Backporting is not necessary.

With this patch, the node eventually converges to composefs enabled. The first upgrade to an image which has this patch will not have composefs. The second upgrade starting from this patch and onward will.

> I think we need to investigate instead fixing this up post-deploy pre-reboot via code execution

For the OCP/OKD case, we clearly need to eventually be able to just assume that composefs is on to unblock other feature work that rely on it. There's indeed workarounds we could do (though actually with this patch, one workaround is just redeploy the same commit after updating) though we're also close I think to be able to addressing the old bootimages problem which would just solve... so much. At least until then, we can get the hardening benefits of composefs.

travier:
> With this patch, the node eventually converges to composefs enabled. The first upgrade to an image which has this patch will not have composefs. The second upgrade starting from this patch and onward will.

If I understand correctly, this means that we do not explicitly enable composefs on RHCOS yet but any system that updates using an ostree with this code will start using composefs?

travier:
Hum, I'm just realizing that we might need this for the Atomic Desktops as well and now I don't understand how we could make it work in Fedora CoreOS anymore without this 😕

cgwalters:
The more I think about this the more I lean towards trying to cleanly split "generate" vs "consume".


Today `composefs = maybe` is not really different from `composefs = no` right? So maybe the cleanest is the default for composefs is set to `no` in the code, but if composefs is compiled in we default to generating a composefs unconditionally.

jlebon:
> If I understand correctly, this means that we do not explicitly enable composefs on RHCOS yet but any system that updates using an ostree with this code will start using composefs?

Yeah, exactly. RHCOS would ship with `enabled = maybe` for now.

> Hum, I'm just realizing that we might need this for the Atomic Desktops as well and now I don't understand anymore how we could make it work in Fedora CoreOS without this 😕

It worked in FCOS because the previous version has new enough libostree. FCOS has periodic barriers. Upgrade tests would've hit this if it weren't covered by one of the existing barriers. It _should_ equally work in the Atomic Desktops given that we require that nodes go through every major at least once.

> So maybe the cleanest is the default for composefs is set to `no` in the code, but if composefs is compiled in we default to generating a composefs unconditionally.

I'm OK with that. If we do that, then we should also make the deploy logic match the `ostree-prepare-root` logic where we always parse the config and correctly error out if composefs is not compiled in but is enforced.

jlebon:
> If we do that, then we should also make the deploy logic match the `ostree-prepare-root` logic where we always parse the config and correctly error out if composefs is not compiled in but is enforced.

Hmm, based on https://github.com/ostreedev/ostree/blob/7daeaa2311abb813003151f4abcb2e02cf8f0315/src/libostree/ostree-sysroot-deploy.c#L657-L658 I think that was an unintended gap in the current code. Will fix!

jlebon:
Updated!

cgwalters:
But won't this cause the opposite problem where we will just bomb out on upgrades because the new prepare-root won't find a composefs for systems that didn't have it configured on previously?

jlebon:
> But won't this cause the opposite problem where we will just bomb out on upgrades because the new prepare-root won't find a composefs for systems that didn't have it configured on previously?

I think you're right.

So... I think we either have to leave the default on `maybe` (but still change the deploy threshold down to `maybe`), or add a new knob to control the deploy and ostree-prepare-root behaviours separately?

cgwalters:
Wouldn't https://github.com/ostreedev/ostree/pull/3353?new_mergebox=true#issuecomment-2541751214 work?

jlebon:
> Wouldn't [my previous suggestion](https://github.com/ostreedev/ostree/pull/3353?new_mergebox=true#issuecomment-2541751214) work?

Gotcha, I had misunderstood your suggestion.

Hmm... it's confusing though that we'd still be generating composefs images even if it's been explicitly turned off. I feel like this is quite close to the `maybe` suggestion, but goes one step further than it needs to and lowers the creation threshold to `no`?

Another way to look at the `maybe` approach I think is that it makes the deploy and mount logic consistent. The mount logic is "if the composefs image is available, then mount it" and the deploy logic becomes "if composefs is available, then generate it". This is also closer to the usual semantic of `maybe`.

src/libotcore/otcore-prepare-root.c
```diff
@@ -27,6 +27,12 @@
 // The kernel argument to configure composefs
 #define CMDLINE_KEY_COMPOSEFS "ostree.prepare-root.composefs"
 
+#ifdef HAVE_COMPOSEFS
+#define COMPOSEFS_TRISTATE_DEFAULT_VALUE OT_TRISTATE_YES
```
cgwalters, line 31: This still needs to change back to `OT_TRISTATE_MAYBE` right?

jlebon:
Updated this now but haven't tested it yet!

jlebon:
OK yup, can confirm this allows ratcheting from older libostree versions. :heavy_check_mark:

```
$ cosa run --qemu-image /srv/imgs/rhcos-412.86.202412161907-0-qemu.x86_64.qcow2
[root@cosa-devsh ~]# rpm-ostree rebase --experimental ostree-unverified-image:oci-archive:$(ls /mnt/workdir/builds/latest/x86_64/*.ociarchive)
[root@cosa-devsh ~]# ls -la /ostree/deploy/rhcos/deploy/*/.*.cfs
ls: cannot access '/ostree/deploy/rhcos/deploy/*/.*.cfs': No such file or directory
[root@cosa-devsh ~]# reboot
[root@cosa-devsh ~]# rpm-ostree status -v
State: idle
AutomaticUpdates: disabled
Deployments:
● ostree-unverified-image:oci-archive:/mnt/workdir/builds/latest/x86_64/scos-9.202412171851-0-ostree.x86_64.ociarchive (index: 0)
                   Digest: sha256:24eef5695d7c95bf07219093c8f6183596fe3b4144af38d673b8ebc7fe88ec2e
                  Version: 9.202412171851-0 (2024-12-17T18:55:00Z)
                   Commit: 35341846a7951dd16124f50a068c4e0d5cfc47d854fbc8ad7efedeb93c8867c9
                   Staged: no
                StateRoot: rhcos

  a99301b69bcdc27b5dba71aa103fdfd896e6ceb1345db9a108a4ee90b113bcd2 (index: 1)
                  Version: 412.86.202412161907-0 (2024-12-16T19:11:04Z)
                   Commit: a99301b69bcdc27b5dba71aa103fdfd896e6ceb1345db9a108a4ee90b113bcd2
                           ├─ rhel-8.6-appstream (2024-12-16T18:35:25Z)
                           ├─ rhel-8.6-baseos (2024-12-11T00:13:42Z)
                           ├─ rhel-8.6-fast-datapath (2024-12-10T00:08:42Z)
                           └─ rhel-8.6-server-ose-4.12 (2024-12-16T19:04:04Z)
                StateRoot: rhcos
[root@cosa-devsh ~]# rpm-ostree rebase :35341846a7951dd16124f50a068c4e0d5cfc47d854fbc8ad7efedeb93c8867c9
[root@cosa-devsh ~]# ls -la /ostree/deploy/rhcos/deploy/*/.*.cfs                                                                                                                                                                                                                                                            
-rw-r--r--. 1 root root 5849088 Dec 17 18:56 /ostree/deploy/rhcos/deploy/35341846a7951dd16124f50a068c4e0d5cfc47d854fbc8ad7efedeb93c8867c9.1/.ostree.cfs
[root@cosa-devsh ~]# reboot
[core@cosa-devsh ~]$ findmnt /
TARGET SOURCE    FSTYPE  OPTIONS
/      composefs overlay ro,relatime,seclabel,lowerdir=/run/ostree/.private/cfsroot-lower::/sysroot/ostree/repo/objects,redirect_dir=on,metacopy=on
```

cgwalters (approved):

travier:
This will likely impact Fedora IoT & RHEL 4 Edge as well cc @pcdubs

cgwalters:
In conversations here we should start trying to use terms like "generated" vs "runtime enabled" instead of just "enabled" - a root problem here is we're conflating those two things with the single flag right now.

Hmmm...you're trying to avoid having composefs runtime-enabled for stable releases just because ostree updated? That makes sense if so.

The simplest thing that comes to mind here is we can just unconditionally always generate the composefs blob, even if `composefs = no`. It doesn't have a large cost.

Then the flag becomes purely about runtime state, which is far easier to reason about.

travier:
> In conversations here we should start trying to use terms like "generated" vs "runtime enabled" instead of just "enabled" - a root problem here is we're conflating those two things with the single flag right now.

👍🏻

> Hmmm...you're trying to avoid having composefs runtime-enabled for stable releases just because ostree updated? That makes sense if so.

Yes

> The simplest thing that comes to mind here is we can just unconditionally always generate the composefs blob, even if `composefs = no`. It doesn't have a large cost.
> 
> Then the flag becomes purely about runtime state, which is far easier to reason about.

Sounds good to me.

> EDIT: If we go that route though, it would mean that people who _don't_ set the composefs configuration would actually switch to be enabled. I think that's probably fine per discussion previously; people who don't want composefs at all for custom OSes can already compile it out, and they can also set the runtime config.

👍🏻

cgwalters:
:arrow_right: https://github.com/ostreedev/ostree/pull/3366

cgwalters:
In the intermediate time though, I think we should *also* start telling everyone who doesn't want composefs to explicitly disable it in their base images.

Obviously the goal of https://github.com/ostreedev/ostree/pull/3366 is to be compatible, and I think we'll do that, but maybe at some point in the future we do push for it to be on by default.

PR #3354, opened 2024-12-16
prepare-root: Add composefs.enabled=verity

ruihe774:
In the current implementation, a composefs deploy is either:
- signed by key and verified by fs-verity, or
- not signed and not verified.

This PR adds a new supported value `verity` to `composefs.enabled`, which:
- requires the composefs to be verified by fs-verity;
- does not require it to be signed.

The idea is that:
- If we do not specify `LCFS_MOUNT_FLAGS_REQUIRE_VERITY` (the underlying of which is `verity=require` when mounting overlayfs), overlayfs does not verify file data even if fs-verity digests are present in the image.
- It is difficult and troublesome to have the ostree commit signed in some circumstances. For example, when deploying with custom layering.
- In some cases, the integrity of the composefs image is already protected by other mechanisms, for example EVM. It is not necessary to verify the image by a key, but it is still necessary to verify file data referenced by the image using fs-verity to protect file integrity from disk corruption.

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -452,9 +452,14 @@ main (int argc, char *argv[])
           expected_digest = g_malloc (OSTREE_SHA256_STRING_LEN + 1);
           ot_bin2hex (expected_digest, cfs_digest_buf, g_variant_get_size (cfs_digest_v));
 
+          g_assert (composefs_config->require_verity);
           cfs_options.flags |= LCFS_MOUNT_FLAGS_REQUIRE_VERITY;
           g_print ("composefs: Verifying digest: %s\n", expected_digest);
           cfs_options.expected_fsverity_digest = expected_digest;
+        }
+      else if (composefs_config->require_verity)
+        {
+          cfs_options.flags |= LCFS_MOUNT_FLAGS_REQUIRE_VERITY;
         }
 
       if (lcfs_mount_image (OSTREE_COMPOSEFS_NAME, TMP_SYSROOT, &cfs_options) == 0)
--- a/docs/composefs.md
+++ b/docs/composefs.md
@@ -47,11 +47,6 @@ before the content of a file in the mounted composefs is read,
 the integrity of its backing OSTree object in `/ostree/repo/objects` is validated by the digest stored in `.ostree.cfs`.
 This can ensure the integrity of the "backing store".
 
-The digests in `.ostree.cfs` are read from fsverity digests of OSTree objects when deploying.
-It is necessary to ensure all OSTree objects referenced have digests stored in `.ostree.cfs`.
-This can be achieved when [committing](#injecting-composefs-digests),
-or you have to set `ex-integrity.fsverity` to `true` for the OSTree repo.
-
 ### Injecting composefs digests
 
 When generating an OSTree commit, there is a CLI switch `--generate-composefs-metadata`
--- a/docs/composefs.md
+++ b/docs/composefs.md
@@ -40,6 +40,18 @@ and specify an Ed25519 public key to validate the booted commit.
 
 See the manpage for `ostree-prepare-root` for details of how to configure it.
 
+### Integrity of backing OSTree objects
+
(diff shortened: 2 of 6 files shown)


src/switchroot/ostree-prepare-root.c
```diff
@@ -449,12 +449,19 @@ main (int argc, char *argv[])
           if (!cfs_digest_buf)
             errx (EXIT_FAILURE, "Failed to query digest: %s", error->message);
 
-          expected_digest = g_malloc (OSTREE_SHA256_STRING_LEN + 1);
-          ot_bin2hex (expected_digest, cfs_digest_buf, g_variant_get_size (cfs_digest_v));
+          if (composefs_config->require_verity)
```
cgwalters, line 452: I'm confused by this change which is underneath `if is_signed` which always implies `require_verity` right? 

I think I'd just add a `g_assert (composefs_config->require_verity)` here instead or so.

cgwalters (changes requested):
ACK to the overall idea though the way I'd describe this is more "external signing". One thing I could imagine doing here is checking if a stamp file like `/run/ostree/composefs-signed` or so is present, then we skip our own signature checks and assume that an external systemd unit that ran before this code executed.

cgwalters (commented):
Thanks, looks sane. Can you also update `man/ostree-prepare-root.xml`?

ruihe774:
> One thing I could imagine doing here is checking if a stamp file like /run/ostree/composefs-signed or so is present, then we skip our own signature checks and assume that an external systemd unit that ran before this code executed.

I think it's a bit of complicated and over-designed. If we do that, the "external" verification tool needs to explicitly cooperate with ostree and write the stamp file, or the end user needs to write a custom systemd service and add it to initramfs, which is troublesome.

ruihe774:
> Can you also update `man/ostree-prepare-root.xml`?

Done. PTAL.

cgwalters (approved):
Looks good!

Our CI just fell over, I filed https://github.com/ostreedev/ostree/pull/3356 to get some things fixed and I think we need to have that resolved before merging more things. But I'll get this in after that, thanks again!

champtar:
To help with upgrades, would a `verity-maybe` make sense ?

ruihe774:
> To help with upgrades, would a `verity-maybe` make sense ?

You mean setting option `verity=on` on the overlayfs mount? Unfortunately, libcompose does not support it.

It only supports either no option (equivalent to `verity=off`) or `verity=require`.

https://www.kernel.org/doc/html/latest/filesystems/overlayfs.html#fs-verity-support
https://github.com/containers/composefs/blob/2e98bef7a783a850b431273fd6928c1d5ff280f8/libcomposefs/lcfs-mount.h#L25

BTW, `LCFS_MOUNT_FLAGS_TRY_VERITY` is to try whether the kernel support the `verity` mount option, not to try whether fs-verity is enabled.

To upgrade, you can simply enable fs-verity for all files under `repo/objects` (There are existing tools to do that) before you deploy with `composefs.enabled=verity`.

allisonkarlitskaya:
I'm late to the party here but is it too late to change the option?  Before this change, afaik, all options were boolean and the config file was effectively toml.  This takes us into a somewhat more freeform direction.... probably ini-ish.... that might make things more difficult in the future if we want to add more advanced features.

Something more like

```toml
[composefs]
enable=true
verity=true
```

instead?

cgwalters:
> This takes us into a somewhat more freeform direction.... probably ini-ish

It's always been ini, we've parsed using GKeyFile; and especially booleans in ini/GKeyFile accept values like "yes/no" and "0/1" that aren't valid in TOML. 

You're right of course regardless that it'd definitely have been cleaner to use a separate boolean for `verity=true`. (That said, `verity=true` should probably imply `enable=true` just to simplify things further) Happy to do that now too.

I'm assuming you're asking this because you want to reimplement it in composefs-rs, but note there is already Rust code in bootc that's parsing this too, and I did have to adjust it for this in e.g. https://github.com/bootc-dev/bootc/commit/06933ed3a2b069bad1f9a002452e2e025a48c0ba

Issue #3211, opened 2024-03-11
`ProtectHome` does not really protect home with ostree


ruihe774:
`ProtectHome=yes` in systemd (see `man:systemd.exec`) hides user directories to "ensure they cannot get access to private user data". However, this setting does not really protect home with ostree. Services can still access user directories from `/sysroot/ostree/deploy/fedora/var/home/` with `ProtectHome=yes`. An example is as follows.

Suppose we have a `secret` file to protect:
```shell
echo "my secret" > /home/test/secret
```

And a malicious service:
```ini
[Unit]
Description=Secret thief

[Service]
Type=oneshot
ProtectHome=true
ExecStart=/usr/bin/cat /sysroot/ostree/deploy/fedora/var/home/test/secret

[Install]
WantedBy=default.target
```

After it is started, the service can still steal the content of `secret`.

To enforce the protection as expected, one has to also add something like `InaccessiblePaths=/sysroot`, which I think is counterintuitive and does not compatible with environment without ostree.

I don't know if I shall post this issue here or to systemd. Please let me know if this is more relevant to systemd.

Maybe related: #2082

cgwalters:
More closely related is https://github.com/ostreedev/ostree/issues/907

I would say we should aim to switch to having the `/sysroot` directory be mode 0700 to start.  You can of course just make this change in any systems you're using; the main fallout right now is it breaks unprivileged read-only direct use of the `ostree` CLI for reading system state; but many or even most use cases may not care about that.  (e.g. things using rpm-ostree already have a DBus intermediate API)

---

This is a valid issue *however* do be aware that there's a really giant difference between using `ProtectHome` as a safeguard mainly against accidental damage versus protecting against malicious (or fully compromised) code running as a systemd unit.  In your example of just using `ProtectHome` on its own...well, the unit can just unmount the masking mounts and regain access...so it needs to be combined with other things.

systemd offers a whole lot of isolation knobs but the most reliable and complete in my opinion starts with `DynamicUser=yes` which will *also* reliably fix access to user home directories, no matter where they're mounted.  I am sure upstream systemd will agree; note the description for `ProtectSystem=strict` says "This setting cannot ensure protection in all cases.".  The other flow of course is to start with a podman-style isolated container (especially with using user namespaces there).

ruihe774:
[...]

PR #3358, opened 2024-12-18
[RFC] Hide /sysroot in a private mount namespace

ruihe774:
Related: #3211

This PR implements a new option `sysroot.invisible` in `prepare-root.conf` to hide `/sysroot` in a private mount namespace `/run/ostree/.private/sysroot-ns` to prevent the rest of the system from accessing it, while ostree admin commands can still operate on it inside a mount namespace.

This PR also add a new admin command "nsenter" that runs program in the mount namespace where `/sysroot` is present. This can wrap tools that currently do not support invisible sysroot. As an example, below is an drop-in override for `rpm-ostreed.service`:

```ini
[Unit]
ConditionPathExists=

[Service]
ExecStart=
ExecStart=+/usr/bin/ostree admin nsenter /usr/bin/rpm-ostree start-daemon
```

I know this is a big change, and there may be corner cases that I haven't considered. However, I have tested it on my local machine, and basic functionalities (booting, deploying, integration with rpm-ostree) work fine. And my code is ready for review.


src/libostree/ostree-impl-system-generator.c
```diff
@@ -126,34 +126,6 @@ require_internal_units (const char *normal_dir, const char *early_dir, const cha
 #endif
 }
 
-// Resolve symlink to return osname
-static gboolean
```
ericcurtin, line 130: How is the Android Boot implementation gonna work if we remove this?

ruihe774:
The test failed because only very new glibc has wrappers for `open_tree` and `move_mount`. I'm going to switch to use raw syscall instead.

src/libostree/ostree-impl-system-generator.c
ericcurtin: I'm a fan of this feature, I'm not convinced if we tested this with the aboot feature it would work though.

Maybe it's just the sake of adding a configuration to opt-in to use this feature and not delete the old way.

src/switchroot/ostree-prepare-root.c
```diff
@@ -77,6 +78,7 @@
 #include "otcore.h"
 
 #define SYSROOT_KEY "sysroot"
+#define INVISIBLE_KEY "invisible"
```
cgwalters, line 81: Minor bikeshed; since this overlaps/replaces `sysroot.readonly` how about we have just
`sysroot.visibility = readonly | hidden` or so?

We'd continue to parse `sysroot.readonly = true` and just turn it into `sysroot.visibility = readonly`.

In general "enums" (as expressed as strings) here feel more extensible than bools.

src/switchroot/ostree-prepare-root.c
```diff
@@ -684,12 +715,60 @@ main (int argc, char *argv[])
...
+
+      const gsize stack_size = 0x8000;
+      g_autofree void *stack = g_malloc (stack_size);
+
+      /* Block signals */
+      sigset_t oldset, newset;
+      sigfillset (&newset);
+      sigprocmask (SIG_SETMASK, &newset, &oldset);
+
+      int pid = clone (invisible_helper, (char*)stack + stack_size, CLONE_VM | CLONE_NEWNS | SIGCHLD, NULL);
```
cgwalters, line 736: I am fine with this as is, but I am wondering if it's really worth using the low level `clone()` API here instead of just invoking `fork() + setns()` ?

Hmm you know we could potentially use a `safe_fork()` helper like systemd has which also wraps the signal handling stuff. Anyways not a blocker, this code looks right to me as is.

src/switchroot/ostree-prepare-root.c
```diff
@@ -684,12 +715,60 @@ main (int argc, char *argv[])
...
+
+      sigprocmask (SIG_SETMASK, &oldset, NULL);
+
+      if (pid < 0)
+          err (EXIT_FAILURE, "failed to create child process");
+
+      /* Bind mount the private mount namespace */
```
cgwalters, line 743: Huh funky...this implementation is different than I thought it'd be. I thought we'd have a "holder process" that has a file descriptor for the mount and it could either have an API to pass that fd back over a socketpair(), or callers could just `nsenter()` to its mount namespace.

(See also https://github.com/containers/bootc/pull/919 for code related to the former)

I didn't realize bind mounting a `/proc/<pid>ns/mnt` like that would work. It seems OK though to me a downside of this is that it still appears in the global filesystem state.

Hmmm...could we actually just have a stub systemd unit in the initramfs and use e.g. https://systemd.io/FILE_DESCRIPTOR_STORE/ to store the mount fd?

src/libostree/ostree-sysroot.c
```diff
@@ -251,9 +266,46 @@ ostree_sysroot_set_mount_namespace_in_use (OstreeSysroot *self)
...
+    {
+      if (unshare (CLONE_NEWNS) < 0)
+        return glnx_throw_errno_prefix (error, "Failed to invoke unshare(CLONE_NEWNS)");
+
+      /* Ensure what we do in our mount namespace do not leak to outside */
+      if (mount (NULL, "/", NULL, MS_PRIVATE | MS_REC | MS_SILENT, NULL) < 0)
+        return glnx_throw_errno_prefix (error, "Failed to set the mount propagation to private");
+    }
+  else
+    g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```
cgwalters, line 302: This error handling doesn't match the GError conventions. Unfortunately the equivalent of `Result<bool>` as it'd be spelled in Rust needs to be encoded with glib+C as a separate `gboolean *out_somevalue` pointer passed in. See e.g. `_ostree_bootloader_query` or `_ostree_fetcher_request_to_tmpfile_finish` for examples of this pattern.

src/libostree/ostree-sysroot.c
```diff
@@ -374,11 +401,75 @@ remount_writable (const char *path, gboolean *did_remount, GError **error)
...
+  /* If we aren't operating on a booted system, then we don't
+   * do anything with mounts.
+   */
+  if (!self->root_is_ostree_booted)
+    return TRUE;
+
+  /* Handle invisible sysroot */
+  if (is_sysroot_invisible ())
+    {
+      glnx_autofd int sysroot_ns_fd = open (OTCORE_RUN_OSTREE_PRIVATE "/sysroot-ns", O_RDONLY);
```
cgwalters, line 438: `| O_CLOEXEC` here and elsewhere

src/libostree/ostree-sysroot.c
```diff
@@ -374,11 +401,75 @@ remount_writable (const char *path, gboolean *did_remount, GError **error)
...
+      if (sysroot_ns_fd < 0)
+        return FALSE;
+
+      glnx_autofd int cur_ns_fd = open("/proc/self/ns/mnt", O_RDONLY);
+      if (cur_ns_fd < 0)
+        return FALSE;
+
+      if (setns (sysroot_ns_fd, CLONE_NEWNS) < 0)
+        return FALSE;
```
cgwalters, line 447: As `setns` doesn't set `error` this needs to be `return glnx_throw_errno_prefix (error, "setns")`.

src/libostree/ostree-sysroot.c
```diff
@@ -374,11 +401,75 @@ remount_writable (const char *path, gboolean *did_remount, GError **error)
...
+
+      glnx_autofd int tree_fd = open_tree (AT_FDCWD, "/", OPEN_TREE_CLONE);
+      if (tree_fd < 0)
+        return FALSE;
```
cgwalters, line 451: Ditto per above, and please audit for other places

src/libostree/ostree-sysroot.c
```diff
@@ -374,11 +401,75 @@ remount_writable (const char *path, gboolean *did_remount, GError **error)
...
+    {
+      glnx_autofd int sysroot_ns_fd = open (OTCORE_RUN_OSTREE_PRIVATE "/sysroot-ns", O_RDONLY);
+      if (sysroot_ns_fd < 0)
+        return FALSE;
+
+      glnx_autofd int cur_ns_fd = open("/proc/self/ns/mnt", O_RDONLY);
+      if (cur_ns_fd < 0)
+        return FALSE;
+
+      if (setns (sysroot_ns_fd, CLONE_NEWNS) < 0)
```
cgwalters, line 446: One thing that will be a serious problem for several extant callers is one can't `setns` with a multithreaded process. And both rpm-ostree and bootc for example are multithreaded.

To fix this we'll need to do the fork() + socketpair passing FD I mention below I think.

src/libostree/ostree-sysroot.c
```diff
@@ -374,11 +401,75 @@ remount_writable (const char *path, gboolean *did_remount, GError **error)
...
+        return FALSE;
+
+      glnx_autofd int tree_fd = open_tree (AT_FDCWD, "/", OPEN_TREE_CLONE);
+      if (tree_fd < 0)
+        return FALSE;
+
+      if (setns (cur_ns_fd, CLONE_NEWNS) < 0)
+        abort (); // it's unsafe to continue if we cannot switch back
+
+      if (move_mount (tree_fd, "", AT_FDCWD, "/sysroot", MOVE_MOUNT_F_EMPTY_PATH) < 0)
```
cgwalters, line 456: Ideally we can drop this...everything in libostree and its callers should hopefully only access things using file-descriptors. There's some legacy tech debt in sysroot because the openat() stuff was only added later (and there's some GFile usage which was a mistake), but...maybe what we can do is just fail if anything tries to access the path property when we're in this mode.

cgwalters (commented):

One struggle we have right now is that the prepare-root code is not unit or integration tested well; I wouldn't block this on that, but adding more features there compounds that issue. I may find some time to work on that...

---

On a different note, in theory this code could be implemented as a service that runs `After=ostree-prepare-root.service` right? I wonder if it'd be helpful for us to structure it that way. But eh, just a thought.

cgwalters:
> The test failed because only very new glibc has wrappers for open_tree and move_mount. I'm going to switch to use raw syscalls instead.

I'm also OK just not having this feature work on older glibc for now.

(Though it's also of course worth noting that in https://github.com/containers/bootc/pull/919 I used the super nice rustix bindings for this which worked fine on C9S even though glibc doesn't have the headers there...and this touches on the ongoing thought of how we might start to use more Rust in libostree in general...or maybe put some of this in bootc to start...)

src/libostree/ostree-sysroot.c
cgwalters: Oh, I see your suggestion of doing
> ExecStart=+/usr/bin/ostree admin nsenter --exec /usr/bin/rpm-ostree start-daemon

Would work around this by having the remounting happen external. And this is also why we'd need to mount `/sysroot` into the new namespace.

But...we could also require callers like rpm-ostree and bootc to invoke the shared library API before becoming multithreaded. Unfortunately a fair bit of stuff happens in `ostree_sysroot_new_default() + ostree_sysroot_set_mount_namespace_in_use (true)`...

src/switchroot/ostree-prepare-root.c
ruihe774: > I am fine with this as is, but I am wondering if it's really worth using the low level `clone()` API here instead of just invoking `fork() + setns()` ?

No.

Please take a look at the following lines:
```C
/* Bind mount the private mount namespace */
g_autofree char *ns = g_strdup_printf ("/proc/%d/ns/mnt", pid);
if (mount (ns, sysroot_ns, NULL, MS_BIND | MS_SILENT, NULL) < 0)
    err (EXIT_FAILURE, "failed to bind-mount sysroot-ns");
```

After spawning the child process, the parent process immediately bind mount its mount namespace to `sysroot_ns`. If we use `fork() + setns()`, the bind mounting may be executed after `fork()` but before `setns()`, which is a race condition. By using `clone()`, it is guaranteed that the child process is created with a new mount namespace and we are always bind mounting the new ns.

src/switchroot/ostree-prepare-root.c
ruihe774: > Hmmm...could we actually just have a stub systemd unit in the initramfs and use e.g. https://systemd.io/FILE_DESCRIPTOR_STORE/ to store the mount fd?

I did not know it before. Let me look into it.

src/switchroot/ostree-prepare-root.c
ruihe774: Good idea.

src/libostree/ostree-sysroot.c
ruihe774: I think after 160bf13 and e0eb4df, the library can work out of box in a multi-threading environment. Mount namespace is a per-thread attribute, which is tricky to handle. In these two commits, I wrote some code to ensure threads that call the library API can all enter private mount namespaces (not the same one, though).

ruihe774: An alternative way is to only use sysroot_fd and boot_fd everywhere and not use paths. We can grab the fd using `open_tree` directly and do not need to enter a mount namespace. However, I'm afraid the caller may still use paths somewhere and it will break the compatibility.

src/libostree/ostree-sysroot.c
ruihe774: Sure. We can grab the fd using open_tree directly and do not need to enter a mount namespace. However, I'm afraid the caller may still use paths somewhere and it will break the compatibility.

src/switchroot/ostree-prepare-root.c
ruihe774: > I thought we'd have a "holder process" that has a file descriptor for the mount and it could either have an API to pass that fd back over a socketpair(), or callers could just nsenter() to its mount namespace.

A problem is to ensure that the "holder process" can live long enough. It must be alive when ostree-finalize-staged runs.

Maybe we can upload the fd to fd store and use socket activation to restart the holder when the mount namespace is required.

> It seems OK though to me a downside of this is that it still appears in the global filesystem state.

If we use `socketpair()` or `nsenter()`, every `root` process can still obtain sysroot, which does not make a difference from placing the nsfd in the global filesystem. If we want to introduce some policies, we may need dbus...

And we can use SELinux or other LSMs to prevent accessing to `/run/ostree/.private`. From a security perspective, I don't think it's really a downside.

ruihe774: Anyway, I think it's too much work for ostree-prepare-root to be a holder process or communicate with systemd or dbus. If you think we need one, it should be a service outside initramfs that consumes sysroot-ns and serves it somehow.

ruihe774:
> > The test failed because only very new glibc has wrappers for open_tree and move_mount. I'm going to switch to use raw syscalls instead.
> 
> I'm also OK just not having this feature work on older glibc for now.
> 
> (Though it's also of course worth noting that in [containers/bootc#919](https://github.com/containers/bootc/pull/919) I used the super nice rustix bindings for this which worked fine on C9S even though glibc doesn't have the headers there...and this touches on the ongoing thought of how we might start to use more Rust in libostree in general...or maybe put some of this in bootc to start...)

I think in first step we can have individual binaries in `switchroot` rewritten in Rust. Currently we have to maintain two `ostree-prepare-root`, one dynamically linked and one static. Using Rust we can easily merge them and always generate static builds.

One blocking point is the dependency on libcomposefs. Maybe we can make a pure-Rust re-implementation of it. I think it's not very difficult.

I'll try to make some POC later if I have spare time.

src/switchroot/ostree-prepare-root.c
ruihe774: I also added an alias `sysroot.protect`. IMO `sysroot.protect=invisible` looks better.

src/switchroot/ostree-prepare-root.c
ruihe774: I added some comments in code to explain it.

cgwalters:
> I think in first step we can have individual binaries in switchroot rewritten in Rust. Currently we have to maintain two ostree-prepare-root, one dynamically linked and one static. Using Rust we can easily merge them and always generate static builds.

We do have a [composefs-rs](https://github.com/containers/composefs-rs/) project going on *but* honestly...I would like to try to remove the static ostree-prepare-root in favor of UKIs basically.

cgwalters:
A totally different approach:

Today systemd when using volatile root [sets up a special symlink](https://github.com/systemd/systemd/blob/3cbf00a30c0a2fae5c02be50e0e47d397a3846e8/src/volatile-root/volatile-root.c#L171) that points to the root block device.

On Linux, it's possible to mount a block device multiple times. So instead of having `/sysroot` mounted at all, we just create that symlink to the real backing block device (which would be [generally useful](https://github.com/containers/composefs/issues/280)) and then anything which wants to operate on the physical root (including ostree/bootc) just...mounts it on their own, wherever the way.

I have only spent 3 minutes and 5 seconds thinking about this and maybe there's something I'm missing but it would be dramatically simpler than anything discussed so far.

cgwalters:
And to be clear I think we should create that symlink unconditionally (should be a separate PR).

Then this PR is just about whether or not to leave `/sysroot` mounted effectively.

cgwalters:
> I don't think this approach can work with Btrfs subvolumes. Sometimes system roots are Btrfs subvolumes, not at the top of a block device.

Yes, a valid point; it is also the setup for Fedora btrfs (at least with Anaconda). The general way to handle this would be to emit a `.mount` unit that describes how to mount the sysroot. 

Actually...we already should have a `sysroot.mount` from the initramfs, we can just copy it to the final root.

(This intersects a bit with https://github.com/containers/bootc/pull/972 - for the btrfs and other cases we could attempt to re-synthesize the necessary mount information from kargs, but I think the most general fix is to honor however the root was set up in the initramfs, which would be `sysroot.mount`).

> And what if the sysroot is not backed by a block device at all? For example, through network?

I don't know that there's a real use case for accessing the physical root in the network case though. Although it depends on what you mean by "network". iSCSI for example is still network, but is also a block device. NFS-as-root...I think is a bad idea. See also https://github.com/containers/bootc/issues/898 which I think is better for those desiring diskless.

cgwalters:
> Bind-mount /sysroot to /run/ostree/.private/sysroot, and unmount /sysroot

I would call this: "Move /sysroot to /run/ostree/.private/sysroot" and list its advantages/disadvantages as basically "Makes it somewhat less likely for processes to find and traverse it, otherwise same as status quo".

Re option 3:

> Privileged processes may steal the fd through /proc. 

Sure but that's really unlikely to happen accidentally, which is about half of what we're trying to improve. A historical issue we've had with the `/sysroot` mount is that backup tools and things like `setfiles` for selinux relabeling traverse it by default. But tools like that know not to traverse `/run` or `/proc`. And as far as *malicious* access, well you have to already be uid 0 to get to it; and yes that's where process sandboxing and LSMs come in, there's not much more we can do here.

src/libotutil/ot-fs-utils.c
```diff
@@ -277,3 +277,26 @@ ot_get_dir_size (int dfd, const char *path, guint64 blocksize, guint64 *out_size
 
   return TRUE;
 }
+
+/* Check whether a path exists */
+gboolean
+ot_path_exists (const char *path, gboolean *out_val, GError **error)
```
cgwalters, line 283: I'm fine with this but note there is a lot of other usage of `glnx_fstatat_allow_noent` in the code for a similar pattern.

src/ostree/ot-admin-builtin-nsenter.c
```diff
@@ -0,0 +1,126 @@
+/*
+* Copyright (C) 2024 Colin Walters <walters@verbum.org>
```
cgwalters, line 2: I think this is actually Copyright to you

src/ostree/ot-admin-builtin-nsenter.c
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library. If not, see <https://www.gnu.org/licenses/>.
+ *
+ * Author: Misaki Kasumi <misakikasumi@outlook.com>
```
cgwalters, line 19: In newer code I've been tending to drop this one as it quickly gets out of date (`git log` is real source of truth); but combined with the above we can basically move this one up

src/ostree/ot-builtin-admin.c
```diff
@@ -121,11 +123,6 @@ ostree_builtin_admin (int argc, char **argv, OstreeCommandInvocation *invocation
             }
         }
 
-      else if (g_str_equal (argv[in], "--"))
```
cgwalters, line 124: Can you motivate this change?

src/switchroot/ostree-prepare-root.c
```diff
@@ -393,11 +448,8 @@ main (int argc, char *argv[])
       // however, see
       // https://github.com/systemd/systemd/blob/604b2001081adcbd64ee1fbe7de7a6d77c5209fe/src/basic/mountpoint-util.h#L36
       // which bumps up these defaults for the rootfs a bit.
-      g_autofree char *root_upperdir
-          = root_transient ? g_build_filename (OTCORE_RUN_OSTREE_PRIVATE, "root/upper", NULL)
-                           : NULL;
-      g_autofree char *root_workdir
-          = root_transient ? g_build_filename (OTCORE_RUN_OSTREE_PRIVATE, "root/work", NULL) : NULL;
+      const char *root_upperdir = OTCORE_RUN_OSTREE_PRIVATE "/root/upper";
```
cgwalters, line 451: Wait isn't this going to turn on `root.transient` by default?

src/switchroot/ostree-prepare-root.c
ruihe774: Let's mark this resolved and discuss this topic in main thread if you want.

src/libostree/ostree-sysroot.c
ruihe774: Sorry but I need to point out that you may want to re-pull it. The code you comment on is outdated.

src/libostree/ostree-sysroot.c
ruihe774: I made it in 96e2fb8. PTAL.

src/libostree/ostree-sysroot.c
```diff
@@ -437,7 +437,8 @@ _ostree_sysroot_ensure_visible (OstreeSysroot *self, GError **error)
       if (sysroot_ns_fd < 0)
         return FALSE;
 
-      glnx_autofd int cur_ns_fd = open("/proc/self/ns/mnt", O_RDONLY);
+      g_autofree char *cur_ns = g_strdup_printf ("/proc/%d/ns/mnt", gettid ());
+      glnx_autofd int cur_ns_fd = open(cur_ns, O_RDONLY);
```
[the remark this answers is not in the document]
ruihe774: It's outdated code.

ruihe774:
In a big commit 96e2fb8 I try to avoid entering a namespace and keep using dir fd for sysroot and boot. The logic is somewhat complicated, especially when involved with remounting in `_ostree_sysroot_ensure_writable`. PTAL.

src/ostree/ot-builtin-admin.c
ruihe774: We need to support command patterns like this:
```console
$ ostree admin nsenter -- sh -c "echo hello"
```

In this command, if `--` is omitted, `-c` is processed by ostree and it fails.

So, nsenter need to support `--` processing. If `--` is processed here, nsenter cannot get the arguments after `--`.

Maybe deleting the logic here may break things. We need to add a OSTREE_BUILTIN_FLAG_ to specify whether a command needs to process `--` by itself. I'll do that later.

src/switchroot/ostree-prepare-root.c
ruihe774: These variables are not used if `root_transient` is false. The code is misleading. I'll put these variable definitions into the `if` block.

ruihe774:
> In a big commit [96e2fb8](https://github.com/ostreedev/ostree/commit/96e2fb8717eadf187c8f7c8b749ff7578bce8392) I try to avoid entering a mount namespace and keep using dir fd for sysroot and boot. The logic is somewhat complicated, especially when involved with remounting in `_ostree_sysroot_ensure_writable`. Now libostree only enters a mount namespace if `OSTREE_ADMIN_BUILTIN_FLAG_ENTER_NS` is specified. PTAL.

Ok. Now it's not necessary to wrap rpm-ostree with nsenter anymore!

src/ostree/ot-admin-builtin-nsenter.c
ruihe774: Hmm... If we consider it seriously, then every file I have modified needs to add my name. I don't want that. I think maybe a better approach is to write "Copyright (C) 2024 Colin Walters and contributors"

src/ostree/ot-admin-builtin-nsenter.c
ruihe774: Maybe we can change the word: "Author" -> "Code owner"

src/ostree/ot-admin-builtin-nsenter.c
cgwalters: Actually on second thought I'd personally vote to just drop `Copyright` and `Author`. Both linux and systemd for example seem to have just the SPDX license identifier in new files and that's it.

ruihe774: We can do that in another PR. Let's keep this for now.

src/ostree/ot-admin-builtin-nsenter.c
ruihe774: We can do that in another PR. Let's keep this for now.

ruihe774:
Hmm... Looks like `_ostree_sysroot_run_in_deployment()` does not work without entering a mount namespace. I'm still trying to figure out how to solve this.

ruihe774:
> In a big commit [96e2fb8](https://github.com/ostreedev/ostree/commit/96e2fb8717eadf187c8f7c8b749ff7578bce8392) I try to avoid entering a mount namespace and keep using dir fd for sysroot and boot. The logic is somewhat complicated, especially when involved with remounting in `_ostree_sysroot_ensure_writable`. Now libostree only enters a mount namespace if `OSTREE_ADMIN_BUILTIN_FLAG_ENTER_NS` is specified. PTAL.

It turned out bringing a lot of trouble. I reverted it.

src/ostree/ot-builtin-admin.c
cgwalters: > Maybe deleting the logic here may break things. We need to add a OSTREE_BUILTIN_FLAG_ to specify whether a command needs to process -- by itself. I'll do that later.

Agreed

ruihe774:
Seems it is too difficult. I'm not working on it anymore.

PR #3359, opened 2024-12-19
sysroot: Support for directories instead of symbolic links in boot part

igoropaniuk:
Allow manipulating and updating /boot/loader entries under a normal directory, as well as using symbolic links.

For directories this uses `renameat2` to do atomic swap of the loader directory in the boot partition. It fallsback to non-atomic rename. This stays atomic on filesystems supporting links but also provide a non-atomic behavior when filesystem does not provide any atomic alternative.

/boot/loader as a normal directory is needed by systemd-boot support, and can be stored under the EFI ESP vfat partition.

Tests were duplicated for simplicity reasons.

Based on the original implementation done by Valentin David [1].

[1] https://github.com/ostreedev/ostree/pull/1967

### DIFF
--- a/src/libostree/ostree-bootloader-grub2.c
+++ b/src/libostree/ostree-bootloader-grub2.c
@@ -411,8 +411,19 @@ _ostree_bootloader_grub2_write_config (OstreeBootloader *bootloader, int bootver
     }
   else
     {
-      new_config_path = ot_gfile_resolve_path_printf (self->sysroot->path,
-                                                      "boot/loader.%d/grub.cfg", bootversion);
+      /* In directory mode boot/loader is a real directory; staging happens in loader.tmp.
+       * In symlink mode staging happens in loader.N.
+       */
+      struct stat loader_stbuf;
+      if (!glnx_fstatat_allow_noent (self->sysroot->sysroot_fd, "boot/loader", &loader_stbuf,
+                                     AT_SYMLINK_NOFOLLOW, error))
+        return FALSE;
+      gboolean loader_is_dir = (errno == 0 && S_ISDIR (loader_stbuf.st_mode));
+      new_config_path
+          = loader_is_dir
+                ? ot_gfile_resolve_path_printf (self->sysroot->path, "boot/loader.tmp/grub.cfg")
+                : ot_gfile_resolve_path_printf (self->sysroot->path, "boot/loader.%d/grub.cfg",
+                                                bootversion);
     }
 
   const char *grub_argv[4] = { NULL, "-o", NULL, NULL };
--- a/src/libostree/ostree-bootloader-syslinux.c
+++ b/src/libostree/ostree-bootloader-syslinux.c
@@ -111,7 +111,17 @@ _ostree_bootloader_syslinux_write_config (OstreeBootloader *bootloader, int boot
 {
   OstreeBootloaderSyslinux *self = OSTREE_BOOTLOADER_SYSLINUX (bootloader);
 
-  g_autofree char *new_config_path = g_strdup_printf ("boot/loader.%d/syslinux.cfg", bootversion);
+  /* In directory mode boot/loader is a real directory; staging happens in loader.tmp.
+   * In symlink mode staging happens in loader.N.
+   */
+  struct stat loader_stbuf;
+  if (!glnx_fstatat_allow_noent (self->sysroot->sysroot_fd, "boot/loader", &loader_stbuf,
+                                 AT_SYMLINK_NOFOLLOW, error))
+    return FALSE;
+  gboolean loader_is_dir = (errno == 0 && S_ISDIR (loader_stbuf.st_mode));
+  g_autofree char *new_config_path
+      = loader_is_dir ? g_strdup ("boot/loader.tmp/syslinux.cfg")
+                      : g_strdup_printf ("boot/loader.%d/syslinux.cfg", bootversion);
 
   /* This should follow the symbolic link to the current bootversion. */
   g_autofree char *config_contents = glnx_file_get_contents_utf8_at (
--- a/src/libostree/ostree-bootloader-uboot.c
+++ b/src/libostree/ostree-bootloader-uboot.c
@@ -171,7 +171,17 @@ _ostree_bootloader_uboot_write_config (OstreeBootloader *bootloader, int bootver
   if (!create_config_from_boot_loader_entries (self, bootversion, new_lines, cancellable, error))
     return FALSE;
 
-  g_autofree char *new_config_path = g_strdup_printf ("boot/loader.%d/uEnv.txt", bootversion);
+  /* In directory mode boot/loader is a real directory; staging happens in loader.tmp.
+   * In symlink mode staging happens in loader.N.
+   */
+  struct stat loader_stbuf;
+  if (!glnx_fstatat_allow_noent (self->sysroot->sysroot_fd, "boot/loader", &loader_stbuf,
+                                 AT_SYMLINK_NOFOLLOW, error))
+    return FALSE;
+  gboolean loader_is_dir = (errno == 0 && S_ISDIR (loader_stbuf.st_mode));
+  g_autofree char *new_config_path = loader_is_dir
+                                         ? g_strdup ("boot/loader.tmp/uEnv.txt")
+                                         : g_strdup_printf ("boot/loader.%d/uEnv.txt", bootversion);
   g_autofree char *new_config_contents = _ostree_sysroot_join_lines (new_lines);
   if (!glnx_file_replace_contents_at (self->sysroot->sysroot_fd, new_config_path,
                                       (guint8 *)new_config_contents, strlen (new_config_contents),
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -2186,6 +2186,53 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
   return TRUE;
 }
 
+/* Determine whether an existing directory implements the semantics described in
+ * https://uapi-group.org/specifications/specs/boot_loader_specification/#type-1-boot-loader-entry-keys
+ */
+static gboolean
+is_bootconfig_type1_semantics (OstreeSysroot *sysroot, GCancellable *cancellable, GError **error)
+{
+  struct stat stbuf;
+
+  if (!_ostree_sysroot_ensure_boot_fd (sysroot, error))
+    return FALSE;
+
+  if (!glnx_fstatat_allow_noent (sysroot->boot_fd, "loader/entries.srel", &stbuf,
+                                 AT_SYMLINK_NOFOLLOW, error))
+    return FALSE;
+
+  if (errno == ENOENT)
+    {
+      g_debug ("Didn't find loader/entries.srel file");
+      return FALSE;
+    }
+  else
+    {
+      /* Get semantics type by reading loader/entries.srel */
+      gsize len;
+      g_autofree char *type = glnx_file_get_contents_utf8_at (
+          sysroot->boot_fd, "loader/entries.srel", &len, cancellable, NULL);
+      if (type == NULL)
+        {
+          g_debug ("Invalid loader/entries.srel file");
+          return FALSE;
+        }
+
+      if (g_str_has_prefix (type, "type1"))
+        {
+          g_debug ("type1 semantics is found in loader/entries.srel file");
+          return TRUE;
+        }
+      else
+        {
+          g_debug ("Unsupported semantics type ('%s') in loader/entries.srel file", type);
+          return FALSE;
+        }
+    }
+
+  return FALSE;
+}
+
 /* We generate the symlink on disk, then potentially do a syncfs() to ensure
  * that it (and everything else we wrote) has hit disk. Only after that do we
  * rename it into place.
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -107,6 +107,13 @@ sysroot_flags_to_copy_flags (GLnxFileCopyFlags defaults, OstreeSysrootDebugFlags
   return defaults;
 }
 
+/* Create the symlink boot/boot. Sets errno on failure. */
+static int
+create_bootself_link (OstreeSysroot *self)
+{
+  return TEMP_FAILURE_RETRY (symlinkat (".", self->sysroot_fd, "boot/boot"));
+}
+
 /* Try a hardlink if we can, otherwise fall back to copying.  Used
  * right now for kernels/initramfs/device trees in /boot, where we can just
  * hardlink if we're on the same partition.
@@ -1831,8 +1838,9 @@ bootloader_entry_filename (OstreeSysroot *sysroot, guint n_deployments,
  */
 static gboolean
 install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
-                           OstreeDeployment *deployment, guint n_deployments, gboolean show_osname,
-                           GCancellable *cancellable, GError **error)
+                           const char *boot_entries_dir, OstreeDeployment *deployment,
+                           guint n_deployments, gboolean show_osname, GCancellable *cancellable,
+                           GError **error)
 
 {
   GLNX_AUTO_PREFIX_ERROR ("Installing kernel", error);
@@ -1858,7 +1866,9 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
   const char *osname = ostree_deployment_get_osname (deployment);
   const char *bootcsum = ostree_deployment_get_bootcsum (deployment);
   g_autofree char *bootcsumdir = g_strdup_printf ("ostree/%s-%s", osname, bootcsum);
-  g_autofree char *bootconfdir = g_strdup_printf ("loader.%d/entries", new_bootversion);
+  /* boot_entries_dir is relative to boot_fd; in symlink mode it is loader.N/entries,
+   * in directory mode it is loader.tmp/entries. */
+  const char *bootconfdir = boot_entries_dir;
 
   if (!glnx_shutil_mkdir_p_at (sysroot->boot_fd, bootcsumdir, 0775, cancellable, error))
     return FALSE;
@@ -2188,11 +2198,10 @@ prepare_new_bootloader_link (OstreeSysroot *sysroot, int current_bootversion, in
   g_assert ((current_bootversion == 0 && new_bootversion == 1)
             || (current_bootversion == 1 && new_bootversion == 0));
 
-  /* This allows us to support both /boot on a seperate filesystem to / as well
+  /* This allows us to support both /boot on a separate filesystem to / as well
    * as on the same filesystem. */
-  if (TEMP_FAILURE_RETRY (symlinkat (".", sysroot->sysroot_fd, "boot/boot")) < 0)
-    if (errno != EEXIST)
-      return glnx_throw_errno_prefix (error, "symlinkat");
+  if (create_bootself_link (sysroot) < 0 && errno != EEXIST)
+    return glnx_throw_errno_prefix (error, "Creating bootself link");
 
   g_autofree char *new_target = g_strdup_printf ("loader.%d", new_bootversion);
 
@@ -2204,10 +2213,114 @@ prepare_new_bootloader_link (OstreeSysroot *sysroot, int current_bootversion, in
   return TRUE;
 }
 
+/* We generate the directory on disk (loader.tmp), then potentially do a syncfs() to ensure
+ * that it (and everything else we wrote) has hit disk. Only after that do we rename it into
+ * place (via renameat2 RENAME_EXCHANGE with loader).
+ *
+ * The ostree BLS entries are already written into loader.tmp/entries by install_deployment_kernel
+ * before this function is called; here we copy over the remaining content (loader.conf,
+ * entries.srel, and any non-ostree third-party BLS entries) so the new directory is complete.
+ */
+static gboolean
+prepare_new_bootloader_dir (OstreeSysroot *sysroot, GCancellable *cancellable, GError **error)
+{
+  GLNX_AUTO_PREFIX_ERROR ("Preparing bootloader directory", error);
+
+  if (!_ostree_sysroot_ensure_boot_fd (sysroot, error))
+    return FALSE;
+
+  /* This allows us to support both /boot on a separate filesystem to / as well
+   * as on the same filesystem. Allowed to fail with EPERM on ESP/vfat.
+   */
+  if (create_bootself_link (sysroot) < 0 && !G_IN_SET (errno, EEXIST, EPERM))
+    return glnx_throw_errno_prefix (error, "Creating bootself link");
+
+  /* Copy non-ostree content from the current loader directory (loader.conf) so
+   * that after the atomic exchange the new loader dir still has files that other
+   * bootloaders (e.g. systemd-boot) require.
+   * Paths are relative to boot_fd; "loader/" prefix is the current active directory.
+   */
+  {
+    struct stat stbuf;
+    const char *src = "loader/loader.conf";
+    if (!glnx_fstatat_allow_noent (sysroot->boot_fd, src, &stbuf, AT_SYMLINK_NOFOLLOW, error))
+      return FALSE;
+    if (errno != ENOENT)
+      {
+        if (!glnx_file_copy_at (sysroot->boot_fd, src, &stbuf, sysroot->boot_fd,
+                                "loader.tmp/loader.conf", GLNX_FILE_COPY_OVERWRITE, cancellable,
+                                error))
+          return FALSE;
+      }
+  }
+
+  /* Always write entries.srel to mark the directory as BLS type1 compliant.
+   * We write it unconditionally rather than copying from the current loader dir
+   * so that it is guaranteed to be present even on first deployment or when
+   * migrating from symlink mode where it may not have existed before.
+   */
+  {
+    static const char entries_srel_content[] = "type1\n";
+    if (!glnx_file_replace_contents_at (
+            sysroot->boot_fd, "loader.tmp/entries.srel", (const guint8 *)entries_srel_content,
+            strlen (entries_srel_content), GLNX_FILE_REPLACE_NODATASYNC, cancellable, error))
+      return FALSE;
+  }
+
+  /* Copy any third-party (non-ostree) BLS entries from loader/entries/ into
+   * loader.tmp/entries/ so they are preserved across the atomic directory swap.
+   * Ostree's own entries (ostree-*.conf) were already written by
+   * install_deployment_kernel and must not be duplicated here.
+   */
+  {
+    gboolean entries_exists;
+    g_auto (GLnxDirFdIterator) dfd_iter = {
+      0,
+    };
+    if (!ot_dfd_iter_init_allow_noent (sysroot->boot_fd, "loader/entries", &dfd_iter,
+                                       &entries_exists, error))
+      return FALSE;
+    if (entries_exists)
+      {
+        glnx_autofd int dest_entries_dfd = -1;
+        if (!glnx_opendirat (sysroot->boot_fd, "loader.tmp/entries", TRUE, &dest_entries_dfd,
+                             error))
+          return FALSE;
+
+        while (TRUE)
+          {
+            struct dirent *dent;
+            if (!glnx_dirfd_iterator_next_dent (&dfd_iter, &dent, cancellable, error))
+              return FALSE;
+            if (dent == NULL)
+              break;
+
+            /* Skip ostree-managed entries; they were written by install_deployment_kernel */
+            if (g_str_has_prefix (dent->d_name, "ostree-"))
+              continue;
+
+            struct stat stbuf;
+            if (!glnx_fstatat (dfd_iter.fd, dent->d_name, &stbuf, AT_SYMLINK_NOFOLLOW, error))
+              return FALSE;
+
+            if (!S_ISREG (stbuf.st_mode))
+              continue;
+
+            if (!glnx_file_copy_at (dfd_iter.fd, dent->d_name, &stbuf, dest_entries_dfd,
+                                    dent->d_name, GLNX_FILE_COPY_OVERWRITE, cancellable, error))
+              return glnx_prefix_error (error, "Copying non-ostree entry %s", dent->d_name);
+          }
+      }
+  }
+
+  return TRUE;
+}
+
 /* Update the /boot/loader symlink to point to /boot/loader.$new_bootversion */
 static gboolean
-swap_bootloader (OstreeSysroot *sysroot, OstreeBootloader *bootloader, int current_bootversion,
-                 int new_bootversion, GCancellable *cancellable, GError **error)
+swap_bootloader (OstreeSysroot *sysroot, OstreeBootloader *bootloader, gboolean loader_link,
+                 int current_bootversion, int new_bootversion, GCancellable *cancellable,
+                 GError **error)
 {
   GLNX_AUTO_PREFIX_ERROR ("Final bootloader swap", error);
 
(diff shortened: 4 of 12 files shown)


igoropaniuk:
This is based on the initial work done in the https://github.com/ostreedev/ostree/pull/1967 PR (looks like it stalled)
I'll push adjusted tests soon to cover this change.

cgwalters:
Thanks so much for picking this back up!! I am very interested in systemd-boot support. But this is just one part of the picture as far as [BLS type 1 spec](https://uapi-group.org/specifications/specs/boot_loader_specification/).

Note that that spec added the [srel file](https://uapi-group.org/specifications/specs/boot_loader_specification/#standard-conformance-marker-file) which I think we should use to dispatch on.

IOW if we find that file, it is an error if `/boot/loader` is a symlink. We should encourage installers to create that to signal compatibility (after this patch lands).

---

This is a tangentially related topic but I'd like to ask you: Have you investigated https://github.com/containers/bootc/ ? That's where most of my development work today is going. I will continue to help maintain ostree (and bootc hard depends on it today) but medium term I do plan to slowly sever that dependency (as we head towards composefs in particular, xref https://github.com/containers/composefs-rs/ for some cool PoC work). And especially that would mean here that we may end up doing BLS and UKI support in Rust in bootc, not here.

cgwalters:
@igoropaniuk Based on your contributions I'd like to offer you reviewer-level access to ostree. This would mean both the right and some responsibility to review/merge PRs made by others.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2212,10 +2212,60 @@ prepare_new_bootloader_link (OstreeSysroot *sysroot, int current_bootversion, in
...
+
+  if (!_ostree_sysroot_ensure_boot_fd (sysroot, error))
+    return FALSE;
+
+  /* This allows us to support both /boot on a seperate filesystem to / as well
+   * as on the same filesystem. Allowed to fail with EPERM on ESP/vfat.
+   */
+  if (TEMP_FAILURE_RETRY (symlinkat (".", sysroot->sysroot_fd, "boot/boot")) < 0)
+    if (errno != EPERM && errno != EEXIST)
+      return glnx_throw_errno_prefix (error, "symlinkat");
```
cgwalters, line 2238: This is the bit I'd like to have dispatch on the `srel` file in particular.

To be conformant, we shouldn't create a symlink even if we can - the spec just says that XBOOTLDR needs to be something "the firmware" can read, and maybe someone makes a firmware that handles ext4 or whatever.

To be clear there's no good reason I am aware of to make XBOOTLDR something other than VFAT today - but still, the srel file seems like the right way to dispatch.

cgwalters: Oh and as a big bonus, dispatching on that file means we can unit test this code without setting up a special hack to make symlink writing fail with `EPERM` etc.

igoropaniuk: @cgwalters makes sense, just to be sure we're on the same page:

1) `"type 1"` in srel file and `boot/loader` is a symlink -> report error
2) `"type 1"` in srel file and `boot/loader` is a directory -> use `prepare_new_bootloader_dir()`
3) no srel file exists/or not `"type 1"` and `boot/loader` is a directory -> use `prepare_new_bootloader_dir()`
4) no srel file exists/or not `"type 1"`  and `boot/loader` is a symlink -> use `prepare_new_bootloader_link()`

igoropaniuk:
> @igoropaniuk Based on your contributions I'd like to offer you reviewer-level access to ostree. This would mean both the right and some responsibility to review/merge PRs made by others.

Sounds great, I will be happy to help with reviews! 


> This is a tangentially related topic but I'd like to ask you: Have you investigated https://github.com/containers/bootc/ ? That's where most of my development work today is going. I will continue to help maintain ostree (and bootc hard depends on it today) but medium term I do plan to slowly sever that dependency (as we head towards composefs in particular, xref https://github.com/containers/composefs-rs/ for some cool PoC work). And especially that would mean here that we may end up doing BLS and UKI support in Rust in bootc, not here.

Actually I had, but our existing OTA solution is based on Uptane (Aktualizr fork as a client) + OSTree, so my primary focus is on this software stack for now.

igoropaniuk:
@cgwalters I've added additional commit for `entries.srel` support, let me know if you prefer to have everything in one commit (will squash them if needed)

igoropaniuk:
@cgwalters I've added additional tests to cover proposed changes

igoropaniuk:
@cgwalters gentle ping, let me know if there is anything else for me to do

cgwalters (commented):
So sorry for the delay here and thanks for working on this again!

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2180,6 +2180,56 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
...
+      g_autofree char *type = glnx_file_get_contents_utf8_at (
+          sysroot->boot_fd, "loader/entries.srel", &len, cancellable, error);
+      if (type == NULL)
+        {
+          g_debug ("Invalid loader/entries.srel file");
+          return FALSE;
+        }
+
+      /* Remove trailing newline symbol if there is any */
+      type[strcspn (type, "\n")] = 0;
```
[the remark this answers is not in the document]
igoropaniuk: fixed, thanks!

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2444,13 +2501,53 @@ write_deployments_bootswap (OstreeSysroot *self, GPtrArray *new_deployments,
...
+   */
+  struct stat stbuf;
+  gboolean loader_link = FALSE;
+  if (!glnx_fstatat_allow_noent (self->sysroot_fd, "boot/loader", &stbuf, AT_SYMLINK_NOFOLLOW,
+                                 error))
+    return FALSE;
+  if (errno == ENOENT)
+    {
+      /* When there is no loader, check if the fs supports symlink or not */
+      if (TEMP_FAILURE_RETRY (symlinkat (".", self->sysroot_fd, "boot/boot")) < 0)
```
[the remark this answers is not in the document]
igoropaniuk: done

igoropaniuk:
@cgwalters Thank you for your review. I have addressed all your recently posted comments. Please let me know if there's anything else I need to do.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2447,13 +2564,47 @@ write_deployments_bootswap (OstreeSysroot *self, GPtrArray *new_deployments,
...
+   * for recommended systemd type1 BLS
+   */
+  struct stat stbuf;
+  gboolean loader_link = FALSE;
+  gboolean force_type1_semantics = is_bootconfig_type1_semantics (self, cancellable, error);
+  if (!glnx_fstatat_allow_noent (self->sysroot_fd, "boot/loader", &stbuf, AT_SYMLINK_NOFOLLOW,
+                                 error))
     return FALSE;
+  if (errno == ENOENT)
+    loader_link = create_bootself_link (self, cancellable, error);
```
cgwalters, line 2577: Hmm. In an ideal world we wouldn't get here...it's a bit tempting to ask that for the case of a user deploying this *new* version of ostree targeting an ESP we ask them to pre-create the link.

Or...actually another tempting thing to do is check the filesystem type for `boot/loader` and unconditionally require this.

Anyways...for now I think what I'd ask is we continue to properly check errors here...

OK I thought more about this and came up with
https://github.com/ostreedev/ostree/pull/3405
to start.

Then a lot of this code can start checking `sysroot->boot_is_vfat` instead of the return value of `symlinkat`; WDYT?

cgwalters: https://github.com/ostreedev/ostree/pull/3405 merged! Let me know what you think!

cgwalters:
I pushed a rebased version of this to https://github.com/cgwalters/ostree/pull/new/symlinks-directories-support - I think I resolved the conflicts OK, but there's definitely some stuff here I want to look at more closely.

dbnicholson:
I'd like to bring my issues with #1967 here:

I've thought about this several times as I would really like to have this in Endless to support our systemd-boot systems. What always makes me anxious is trying to figure out how to handle compatibility with the vast majority of our systems that have symlink based deployments. There are 2 main issues I'm concerned with:

* The semantics of the deployment are different if loader is a symlink to a loader.N directory versus loader itself being a directory. What do the swapped loader.N directories mean in pure directory mode? I think the answer is that they don't really match the semantics of the symlink deployment. It's essentially a different algorithm even though it's seemingly only a small change from the symlink approach.
* What do you do with systems that already have symlink based deployments? Potentially you could write a migration to the directory scheme, but that means you can't roll back to a system with a libostree that doesn't understand the directory scheme.

So, to me this requires a couple additional pieces of implementation and policy.

* The directory deployment scheme is treated separately from the symlink deployment scheme. Your system can be in one scheme or the other, but the semantics are different and shouldn't be mixed.
* If your system is already using a symlink based deployment, it should stay that way. You can opt in to a migration to the directory scheme, but this means you may potentially lose roll back targets. At Endless we'd do that at some major version check point and probably have to delete some old deployments to ensure users didn't try to roll back to a system where they'd be screwed. Or maybe we'd never migrate anyone because the downsides are too great.
* If you're on a new deployment (i.e., no existing /boot/loader or /boot/ostree), then the directory scheme is preferred.

Does this PR consider any of these points?

src/libostree/ostree-sysroot.c
```diff
@@ -613,7 +618,16 @@ _ostree_sysroot_read_boot_loader_configs (OstreeSysroot *self, int bootversion,
...
+  g_autofree char *entries_path = NULL;
+  int current_version;
+  if (!read_current_bootversion (self, &current_version, cancellable, error))
+    return FALSE;
+
+  if (current_version == bootversion)
+    entries_path = g_strdup ("boot/loader/entries");
+  else
+    entries_path = g_strdup_printf ("boot/loader.%d/entries", bootversion);
+
```
champtar, line 711: Are we ever reading anything else than the current config ? ie `boot/loader/entries`

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2204,10 +2264,57 @@ prepare_new_bootloader_link (OstreeSysroot *sysroot, int current_bootversion, in
   return TRUE;
 }
 
+/* We generate the directory on disk, then potentially do a syncfs() to ensure
+ * that it (and everything else we wrote) has hit disk. Only after that do we
+ * rename it into place (via renameat2 RENAME_EXCHANGE).
+ */
+static gboolean
+prepare_new_bootloader_dir (OstreeSysroot *sysroot, int current_bootversion, int new_bootversion,
```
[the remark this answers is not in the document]
champtar: > That would have been harder to do with the symlink logic

The directory name could have been random, or a hash of the content, and for the cleanup you just need to readlink before you rename, but we can't change now

> I think it'd be quite a simplification to the current code here. Especially then we can stop writing the special ostree_bootversion file into the loader directory right?

Agreed, no need for bootversion when using directory swap

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2217,14 +2324,24 @@ swap_bootloader (OstreeSysroot *sysroot, OstreeBootloader *bootloader, int curre
...
+       * see https://bugzilla.gnome.org/show_bug.cgi?id=755595
+       */
+      if (!glnx_renameat (sysroot->boot_fd, "loader.tmp", sysroot->boot_fd, "loader", error))
+        return FALSE;
+    }
+  else
+    {
+      /* New target is currently under the old/current version */
+      g_autofree char *new_target = g_strdup_printf ("loader.%d", current_bootversion);
+      if (glnx_renameat2_exchange (sysroot->boot_fd, new_target, sysroot->boot_fd, "loader") != 0)
```
champtar: `glnx_renameat2_exchange` swap the 2 directories, so you need to delete `loader.%d` after `fsfreeze_thaw_cycle`
(`glnx_renameat` only leaves 1 file on success)

champtar:
@igoropaniuk can you rebase on main and address the current comments ?

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2176,6 +2191,53 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
...
+ */
+static gboolean
+is_bootconfig_type1_semantics (OstreeSysroot *sysroot, GCancellable *cancellable, GError **error)
+{
+  struct stat stbuf;
+
+  if (!_ostree_sysroot_ensure_boot_fd (sysroot, error))
+    return FALSE;
+
+  if (!glnx_fstatat_allow_noent (sysroot->boot_fd, "loader/entries.srel", &stbuf,
```
champtar, line 2200: We need to create / write `entries.srel` each time we create a new `loader` dir ? Or we already copy the content of the current directory first ?

champtar:
Looking at the PR I would say

> So, to me this requires a couple additional pieces of implementation and policy.
> 
>     * The directory deployment scheme is treated separately from the symlink deployment scheme. Your system can be in one scheme or the other, but the semantics are different and shouldn't be mixed.

yes

>     * If your system is already using a symlink based deployment, it should stay that way. You can opt in to a migration to the directory scheme, but this means you may potentially lose roll back targets. At Endless we'd do that at some major version check point and probably have to delete some old deployments to ensure users didn't try to roll back to a system where they'd be screwed. Or maybe we'd never migrate anyone because the downsides are too great.

yes (entries.srel containing "type1\n")

>     * If you're on a new deployment (i.e., no existing /boot/loader or /boot/ostree), then the directory scheme is preferred.

No, and this prevent downgrade, so at first I would personally prefer to use directories only when required (/boot is vfat)

src/libostree/ostree-sysroot-deploy.c
champtar: systemd-boot also reads `loader/loader.conf` and `loader/entries/*.conf` https://www.freedesktop.org/software/systemd/man/latest/loader.conf.html
either we need to just create `loader/entries.tmp` then rename to `loader/entries.tmp` (so we keep the content of `loader` at least), or we need to copy the full `loader` dir and just cleanup `loader/entries/ostree-*.conf` only to keep everything else

igoropaniuk:
Could you let me know if anything else should be addressed/improved in the context of this PR
@cgwalters @champtar

igoropaniuk:
Gentle ping on this @cgwalters @champtar

bam80:
Why closed?

cgwalters:
I know it's been a crazy delay here - this code touches the a very core component of the project and really needs careful review and testing. At the same time, obviously a lot of my time is now on bootc and https://github.com/composefs/composefs-rs 

I'm trying to take another look at this now though!

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2216,16 +2376,27 @@ swap_bootloader (OstreeSysroot *sysroot, OstreeBootloader *bootloader, int curre
...
+      if (!glnx_renameat (sysroot->boot_fd, "loader.tmp", sysroot->boot_fd, "loader", error))
+        return FALSE;
+    }
+  else
+    {
+      /* Atomically swap loader.tmp (the freshly-built directory) with the live
+       * loader directory.  After the exchange loader.tmp holds the old content
+       * which we remove below once the new state is safely on disk.
+       */
+      if (glnx_renameat2_exchange (sysroot->boot_fd, "loader.tmp", sysroot->boot_fd, "loader") != 0)
```
cgwalters, line 2397: Keeping this one but it looks wrong, I think we do expect the outer environment here to define whether we're using entries.srel or a symlink up front.

<details>
AI: *Important*: On first boot on a real vfat/ESP partition, `boot/loader` does not exist yet — `create_bootself_link` fails with EPERM so we enter directory mode, stage into `loader.tmp/`, but then `renameat2(RENAME_EXCHANGE)` returns ENOENT because `loader` does not exist (the kernel requires *both* paths to exist for RENAME_EXCHANGE). The three-step non-atomic fallback also fails for the same reason. Result: first deployment on a fresh vfat ESP always fails with "Exchanging loader.tmp and loader: No such file or directory".

The tests hide this because `admin-test.sh` always pre-creates `sysroot/boot/loader` as a directory before the first deploy.

Fix: detect the first-boot case (e.g. `errno == ENOENT` after the exchange attempt) and fall back to a plain `renameat(boot_fd, "loader.tmp", boot_fd, "loader")`; after the first deploy `loader` will always exist so subsequent deploys can use RENAME_EXCHANGE normally. Or ensure `loader/` is pre-created at sysroot init time.
</details>

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2401,8 +2582,58 @@ write_deployments_bootswap (OstreeSysroot *self, GPtrArray *new_deployments,
...
+   * In symlink mode the staging dir is boot/loader.N/entries (unchanged from before).
+   * In directory mode we always stage in boot/loader.tmp/entries and then atomically swap
+   * loader.tmp with loader via renameat2 RENAME_EXCHANGE.
+   */
+  if (!_ostree_sysroot_ensure_boot_fd (self, error))
+    return FALSE;
+
+  struct stat stbuf;
+  gboolean loader_link = FALSE;
+  gboolean force_type1_semantics = is_bootconfig_type1_semantics (self, cancellable, error);
```
cgwalters, line 2598: AI: *Important*: GError contract violation. `is_bootconfig_type1_semantics` sets `*error` and returns FALSE on genuine I/O errors (e.g. `_ostree_sysroot_ensure_boot_fd` failure). The result is used as a plain boolean here without checking whether `*error` was set. The very next call passes the same `error` pointer to `glnx_fstatat_allow_noent`, which will call `g_set_error` on an already-set GError — in debug builds (G_ENABLE_DEBUG) this triggers `g_return_if_fail(*error == NULL)` and aborts the process.

Fix: add `if (*error) return FALSE;` after this line, or redesign `is_bootconfig_type1_semantics` to use an `out_result` parameter so that FALSE unambiguously means error rather than "not type1".

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2401,8 +2582,58 @@ write_deployments_bootswap (OstreeSysroot *self, GPtrArray *new_deployments,
...
+      new_loader_entries_dir = g_strdup ("boot/loader.tmp/entries");
+      boot_entries_dir = g_strdup ("loader.tmp/entries");
+    }
+  else
+    {
+      new_loader_entries_dir = g_strdup_printf ("boot/loader.%d/entries", new_bootversion);
+      boot_entries_dir = g_strdup_printf ("loader.%d/entries", new_bootversion);
+    }
+
   if (!glnx_shutil_rm_rf_at (self->sysroot_fd, new_loader_entries_dir, cancellable, error))
```
cgwalters, line 2637: Needs evaluation.

---

AI: *Important*: In directory mode `new_loader_entries_dir` is `"boot/loader.tmp/entries"` — so only the `entries/` subdirectory is cleaned, leaving the rest of `loader.tmp/` intact. Stale files from a previous deployment (e.g. `grub.cfg`, `syslinux.cfg`, `uEnv.txt` from a different bootloader) remain in `loader.tmp/` and end up in the live `loader/` directory after RENAME_EXCHANGE. `prepare_new_bootloader_dir` only writes `loader.conf` and `entries.srel`; it does not scrub other stale top-level files.

Fix: when `!loader_link`, remove and recreate the full `"boot/loader.tmp"` tree rather than just the entries subdirectory.

src/libostree/ostree-sysroot.c
```diff
@@ -675,7 +676,44 @@ _ostree_sysroot_read_boot_loader_configs (OstreeSysroot *self, int bootversion,
...
+    {
+      /* Directory mode (or no loader at all): prefer loader.tmp/entries when it exists
+       * (i.e. during deployment staging), otherwise use the live loader/entries.
+       */
+      struct stat tmp_stbuf;
+      if (!glnx_fstatat_allow_noent (self->boot_fd, "loader.tmp/entries", &tmp_stbuf,
+                                     AT_SYMLINK_NOFOLLOW, error))
+        return FALSE;
+      if (errno == 0 && S_ISDIR (tmp_stbuf.st_mode))
+        entries_path = g_strdup ("loader.tmp/entries");
```
cgwalters, line 701: Needs evaluation.

---

AI: *Important*: If a previous deployment was interrupted after `loader.tmp/entries/` was created but before the RENAME_EXCHANGE swap, a stale staging directory persists on disk. On the next ostree invocation this path is unconditionally preferred over the live `loader/entries`, causing `ostree admin status`, rollback selection, and any subsequent bootloader writes to read phantom entries.

Consider writing a sentinel file (e.g. `loader.tmp/.ostree-staging`) at the very start of staging and deleting it as the last cleanup step; only prefer `loader.tmp` when that sentinel is present. At minimum, delete `loader.tmp/entries/` first during cleanup (before the rest of `loader.tmp/`) so that a partial `rm_rf` cannot leave a browsable-but-stale entries directory.

src/libostree/ostree-bootloader-grub2.c
```diff
@@ -411,8 +411,19 @@ _ostree_bootloader_grub2_write_config (OstreeBootloader *bootloader, int bootver
...
-      new_config_path = ot_gfile_resolve_path_printf (self->sysroot->path,
-                                                      "boot/loader.%d/grub.cfg", bootversion);
+      /* In directory mode boot/loader is a real directory; staging happens in loader.tmp.
+       * In symlink mode staging happens in loader.N.
+       */
+      struct stat loader_stbuf;
+      if (!glnx_fstatat_allow_noent (self->sysroot->sysroot_fd, "boot/loader", &loader_stbuf,
+                                     AT_SYMLINK_NOFOLLOW, error))
+        return FALSE;
+      gboolean loader_is_dir = (errno == 0 && S_ISDIR (loader_stbuf.st_mode));
```
cgwalters, line 421: AI: *Important*: This backend re-detects loader mode independently by statting `boot/loader`, but uses a different heuristic than `write_deployments_bootswap`. On first boot when `boot/loader` does not yet exist: `write_deployments_bootswap` enters directory mode (because `create_bootself_link` → EPERM), but here `errno == ENOENT` → `loader_is_dir = FALSE` → config written to `boot/loader.N/grub.cfg` instead of `boot/loader.tmp/grub.cfg`. The same inconsistency exists in the syslinux and uboot backends.

This is also a DRY violation — the detection is copy-pasted three times and can silently diverge. Consider passing `loader_link` (or an equivalent staging-path string) as a parameter to `_ostree_bootloader_write_config`, or caching the decision in the sysroot struct.

src/libostree/ostree-bootloader-syslinux.c
```diff
@@ -111,7 +111,17 @@ _ostree_bootloader_syslinux_write_config (OstreeBootloader *bootloader, int boot
...
 
-  g_autofree char *new_config_path = g_strdup_printf ("boot/loader.%d/syslinux.cfg", bootversion);
+  /* In directory mode boot/loader is a real directory; staging happens in loader.tmp.
+   * In symlink mode staging happens in loader.N.
+   */
+  struct stat loader_stbuf;
+  if (!glnx_fstatat_allow_noent (self->sysroot->sysroot_fd, "boot/loader", &loader_stbuf,
+                                 AT_SYMLINK_NOFOLLOW, error))
+    return FALSE;
+  gboolean loader_is_dir = (errno == 0 && S_ISDIR (loader_stbuf.st_mode));
```
cgwalters, line 121: AI: Same loader-mode detection issue as in grub2 backend — see comment there. On first boot where `boot/loader` does not yet exist, this evaluates to `FALSE` (not a directory) and writes the config to `boot/loader.N/syslinux.cfg`, disagreeing with the directory-mode decision made in `write_deployments_bootswap`.

src/libostree/ostree-bootloader-uboot.c
```diff
@@ -171,7 +171,17 @@ _ostree_bootloader_uboot_write_config (OstreeBootloader *bootloader, int bootver
...
 
-  g_autofree char *new_config_path = g_strdup_printf ("boot/loader.%d/uEnv.txt", bootversion);
+  /* In directory mode boot/loader is a real directory; staging happens in loader.tmp.
+   * In symlink mode staging happens in loader.N.
+   */
+  struct stat loader_stbuf;
+  if (!glnx_fstatat_allow_noent (self->sysroot->sysroot_fd, "boot/loader", &loader_stbuf,
+                                 AT_SYMLINK_NOFOLLOW, error))
+    return FALSE;
+  gboolean loader_is_dir = (errno == 0 && S_ISDIR (loader_stbuf.st_mode));
```
cgwalters, line 181: AI: Same loader-mode detection issue as in the grub2 and syslinux backends — see comment on grub2. On first boot where `boot/loader` does not yet exist, this disagrees with the directory-mode decision in `write_deployments_bootswap`.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2176,6 +2186,53 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
...
+    {
+      g_debug ("Didn't find loader/entries.srel file");
+      return FALSE;
+    }
+  else
+    {
+      /* Get semantics type by reading loader/entries.srel */
+      gsize len;
+      g_autofree char *type = glnx_file_get_contents_utf8_at (
+          sysroot->boot_fd, "loader/entries.srel", &len, cancellable, NULL);
```
cgwalters, line 2214: AI: `NULL` is passed as the GError argument, silently suppressing I/O errors when reading `entries.srel`. A permission error or corrupted file is treated identically to "file absent" — both cause the function to return FALSE with `*error` unset, making it impossible for the caller to distinguish the two cases. Pass `error` here instead.

---

Yes, looks valid - I think we have a helper for "open if exists" which is what we want here.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2176,6 +2186,53 @@ install_deployment_kernel (OstreeSysroot *sysroot, int new_bootversion,
...
+          return TRUE;
+        }
+      else
+        {
+          g_debug ("Unsupported semantics type ('%s') in loader/entries.srel file", type);
+          return FALSE;
+        }
+    }
+
+  return FALSE;
```
cgwalters, line 2233: AI: (low) This `return FALSE` is unreachable — every branch in the preceding if/else chain returns explicitly. It also obscures the API confusion: `is_bootconfig_type1_semantics` returns FALSE both for "not type1" (no error) and for actual I/O errors (error set). These two cases should be distinguishable; see the related comment at the call site (line ~2598).

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2204,10 +2260,114 @@ prepare_new_bootloader_link (OstreeSysroot *sysroot, int current_bootversion, in
...
+{
+  GLNX_AUTO_PREFIX_ERROR ("Preparing bootloader directory", error);
+
+  if (!_ostree_sysroot_ensure_boot_fd (sysroot, error))
+    return FALSE;
+
+  /* This allows us to support both /boot on a separate filesystem to / as well
+   * as on the same filesystem. Allowed to fail with EPERM on ESP/vfat.
+   */
+  if (create_bootself_link (sysroot) < 0 && !G_IN_SET (errno, EEXIST, EPERM))
```
cgwalters, line 2282: AI: (low) `create_bootself_link` is already called a few lines earlier in `write_deployments_bootswap` (which handles the EEXIST/EPERM cases). This call inside `prepare_new_bootloader_dir` will always hit EEXIST and is redundant. It is harmless but confusing to readers — consider removing it.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -2244,6 +2415,16 @@ swap_bootloader (OstreeSysroot *sysroot, OstreeBootloader *bootloader, int curre
   ot_journal_print (LOG_INFO, "Completed freeze/thaw cycle for boot in %" G_GUINT64_FORMAT " ms",
                     end_msec - start_msec);
 
+  /* In the directory case, glnx_renameat2_exchange swaps loader.tmp and loader atomically,
+   * leaving the old loader content in loader.tmp. Now that the new state is safely on disk
+   * (after fsfreeze/thaw), remove the now-stale loader.tmp directory.
+   */
+  if (!loader_link)
+    {
+      if (!glnx_shutil_rm_rf_at (sysroot->boot_fd, "loader.tmp", cancellable, error))
```
cgwalters, line 2424: Hmm, maybe.

---

AI: (low) If `rm_rf` is interrupted by a crash or signal, `loader.tmp/` may be partially deleted. If `loader.tmp/entries/` happens to survive the partial deletion, the "prefer loader.tmp" heuristic in `_ostree_sysroot_read_boot_loader_configs` will kick in on the next run and read stale data. A simple mitigation: delete `loader.tmp/entries/` first (before the rest of `loader.tmp/`), so the heuristic cannot be tricked by a partially-cleaned directory.

tests/admin-test.sh
```diff
@@ -63,6 +63,20 @@ export rev
...
 # timestamp of the deploy directory to the epoch as a regression test.
 touch -d @0 sysroot/ostree/deploy
+
+if [ "${folders_instead_symlinks_in_boot}" == "1" ]; then
+    # Have loader as a directory (simulate ESP-based deployments)
+    if [ -h sysroot/boot/loader ]; then
+        loader=`readlink sysroot/boot/loader`
+        rm -f sysroot/boot/loader
+        mv sysroot/boot/${loader} sysroot/boot/loader
+        echo -n ${loader} > sysroot/boot/loader/ostree_bootversion
```
cgwalters, line 73: AI: (low) `ostree_bootversion` is never read by the C code — it is dead test setup. In directory mode the current boot version is derived by parsing `ostree=` kargs from BLS entries, not from this file. The file will persist in `loader/` after deployment, which is harmless but confusing. Please remove both `echo` lines (here and line 76).

tests/test-admin-dir-deploy-syslinux.sh
```diff
@@ -0,0 +1,71 @@
...
+extra_admin_tests=3
+folders_instead_symlinks_in_boot=1
+
+. $(dirname $0)/admin-test.sh
+
+# Test the legacy dirs
+for test_bootdir in "boot" "usr/lib/ostree-boot"; do
+    cd ${test_tmpdir}
+    rm httpd osdata testos-repo sysroot -rf
+    setup_os_repository "archive" "syslinux" $test_bootdir
```
cgwalters, line 36: AI: *Important*: This loop calls `setup_os_repository` which creates `sysroot/boot/loader` as a symlink, then deploys without converting it to a real directory. Because the test filesystem supports symlinks, `create_bootself_link` succeeds → `loader_link = TRUE` → the deploy runs entirely in **symlink mode**. This defeats the purpose of `test-admin-dir-deploy-syslinux.sh`. The same issue affects the equivalent sections in `test-admin-dir-deploy-grub2.sh` and `test-admin-dir-deploy-uboot.sh`.

Fix: after `setup_os_repository`, add the same symlink-to-directory conversion that `admin-test.sh` does (remove the symlink, rename `loader.0` → `loader`, or simply `mkdir -p sysroot/boot/loader`).

tests/test-admin-dir-deploy-syslinux.sh
```diff
@@ -0,0 +1,71 @@
...
+    assert_file_has_content sysroot/ostree/boot.1/testos/${bootcsum}/0/etc/os-release 'NAME=TestOS'
+    ${CMD_PREFIX} ostree admin status
+    validate_bootloader
+    echo "ok kernel in $test_bootdir"
+done
+
+# And test that legacy overrides /usr/lib/modules
+cd ${test_tmpdir}
+rm httpd osdata testos-repo sysroot -rf
+setup_os_repository "archive" "syslinux" "usr/lib/ostree-boot"
```
cgwalters, line 57: AI: Same issue as the loop above — `setup_os_repository` creates `sysroot/boot/loader` as a symlink, and the subsequent deploy runs in symlink mode. Neither of these extra test sections tests directory mode.

cgwalters (commented):
OK, starting with a LLM-assisted draft here. Some of this is right, some needs further evalution.

Everything prefixed with AI: is Assisted-by: OpenCode (Mix: Sonnet/Opus 4.6 + Gemini 3.5 Flash + 3.1 Pro )

tests/test-admin-dir-deploy-syslinux.sh
lxnay: swap_bootloader does not handle the initial deployment case. The new code should be changed to something like this:

```
{
    /* First deployment: loader directory may not exist yet, fall back to rename. */
    if (glnx_renameat2_exchange (sysroot->boot_fd, "loader.tmp",
                                  sysroot->boot_fd, "loader") != 0)
      {
        if (errno != ENOENT)
          return glnx_throw_errno_prefix (error, "Exchanging loader.tmp and loader");
        /* loader doesn't exist yet (first deploy) — plain rename is fine. */
        if (!glnx_renameat (sysroot->boot_fd, "loader.tmp",
                             sysroot->boot_fd, "loader", error))
          return FALSE;
      }
}
```

PR #3364, opened 2024-12-20
chore: Use geteuid() instead of getuid() to check privilege

ruihe774:
I believe the correct way is to check `geteuid() == 0`, not `getuid() == 0`.

### DIFF
--- a/src/libotutil/ot-unix-utils.c
+++ b/src/libotutil/ot-unix-utils.c
@@ -32,6 +32,9 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/types.h>
+#include <sys/prctl.h>
+#include <linux/prctl.h>
+#include <linux/capability.h>
 #include <unistd.h>
 
 /* Ensure that a pathname component @name does not contain the special Unix
@@ -107,5 +110,12 @@ ot_util_path_split_validate (const char *path, GPtrArray **out_components, GErro
 gboolean
 ot_util_process_privileged (void)
 {
-  return geteuid() == 0;
+  if (geteuid() != 0)
+    return FALSE;
+
+  // https://github.com/containers/bootc/blob/c88fcfd6e145863408bde7d4706937dd323f64e2/lib/src/cli.rs#L621
+  if (prctl (PR_CAPBSET_READ, CAP_SYS_ADMIN) != 1)
+    return FALSE;
+
+  return TRUE;
 }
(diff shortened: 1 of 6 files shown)


ericcurtin (approved):
This looks correct to me, I tend to favour geteuid over getuid, most of the time that's what people actually want to do. But I'm not sure about every little use-case of OSTree.

What I will say, was there actually a bug or defect found here, etc.?

Or are we just changing for the sake of it because geteuid is the "better" one that's a more true test of rootness?

ruihe774:
> What I will say, was there actually a bug or defect found here, etc.?

Not appeared. But it is easy produce one. You can set the SUID bit of a program using libostree (e.g. `ostree` itself). In a program that gains privilege by SUID, `getuid()` returns the user that invokes the program, and `geteuid()` returns 0.

src/libotutil/ot-unix-utils.c
```diff
@@ -102,3 +102,10 @@ ot_util_path_split_validate (const char *path, GPtrArray **out_components, GErro
   ot_transfer_out_value (out_components, &ret_components);
   return TRUE;
 }
+
+/* Check if current process is privileged */
+gboolean
+ot_util_process_privileged (void)
+{
+  return geteuid() == 0;
```
cgwalters, line 110: This is totally fine, but what we actually want to check in almost all of these cases if if we have `CAP_SYS_ADMIN` - ref https://github.com/containers/bootc/blob/c88fcfd6e145863408bde7d4706937dd323f64e2/lib/src/cli.rs#L621

cgwalters (approved):
No one should be including libostree in a setuid binary. 
(Really there should be no setuid binaries at all, but that's another thing)

ruihe774:
> No one should be including libostree in a setuid binary. (Really there should be no setuid binaries at all, but that's another thing)

Yes. However, processes can drop privileges temporarily using `seteuid()` as well. In this case, `getuid()` returns 0 and `geteuid()` returns a unprivileged uid.

cgwalters:
> Yes. However, processes can drop privileges temporarily using seteuid() as well. In this case, getuid() returns 0 and geteuid() returns a unprivileged uid.

Yeah, though that kind of thing I personally consider a bad idea because most modern codebases are multithreaded, and changing uid affects all threads, which can have surprising effects. Basically if you do this type of thing you should fork off a dedicated helper process which runs always as that uid from the start - and that process shouldn't be doing anything related to ostree in general.

ruihe774:
> and changing uid affects all threads

This is implemented by libc (e.g. NPTL). If you use raw syscall, changing uid affects only current thread.

cgwalters:
> If you use raw syscall, changing uid affects only current thread.

Yes, but doing raw syscalls like that last I heard from the glibc developers often means you cannot invoke any later APIs from glibc, which becomes a giant pain in the butt. I'm not sure if there's a better reference but (searches for a minute) yeah see [this comment](https://github.com/rust-lang/rust/issues/89522#issuecomment-944636262) for example.

PR #3371, opened 2025-01-15
Release 2025.1

jmarrero:
This Release adds one new feature introduced on: https://github.com/ostreedev/ostree/pull/3362 which adds a new `--kexec` flag to `ostree admin upgrade` which will cause the deployment to be loaded into kexec after the upgrade completes.

Other than that it mostly a bugfix and small improement release with the more significant change being https://github.com/ostreedev/ostree/pull/3366 which relates to composefs and notably does:

- If composefs is enabled at build time, we always generate a composefs blob at deplyment time
- Configuring the prepare-root config now mostly only affects the runtime state.

```
Colin Walters (2):
      tree-wide: Rerun clang-format, update ci
      Always generate composefs blob, don't enable runtime by default

Joseph Marrero Corchado (1):
      Release 2025.1

Mary Strodl (1):
      bin/admin-upgrade: add kexec support

Misaki Kasumi (2):
      chore: Use geteuid() instead of getuid() to check privilege
      chore: Check CAP_SYS_ADMIN in ot_util_process_privileged
```
## New Contributors
* @Mstrodl  made their first contribution in https://github.com/ostreedev/ostree/pull/3362

**Full Changelog**: https://github.com/ostreedev/ostree/compare/v2024.10...v2025.1

### DIFF
--- a/Makefile-libostree.am
+++ b/Makefile-libostree.am
@@ -175,9 +175,9 @@ endif # USE_GPGME
 symbol_files = $(top_srcdir)/src/libostree/libostree-released.sym
 
 # Uncomment this include when adding new development symbols.
-if BUILDOPT_IS_DEVEL_BUILD
-symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
-endif
+#if BUILDOPT_IS_DEVEL_BUILD
+#symbol_files += $(top_srcdir)/src/libostree/libostree-devel.sym
+#endif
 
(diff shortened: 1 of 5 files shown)


cgwalters:
The release checklist wants two commits; one for the release and one which flips back release_build=no

jmarrero:
yeah, been a while. Pushed too early.

jmarrero:
@cgwalters please review the release notes here too

cgwalters (approved):

PR #3376, opened 2025-01-30
Update GIR

Mstrodl:
(And regenerate rust bindings)


cgwalters:
Thank you! I think this may just be missing a `cargo fmt`?

cgwalters:
Alternatively, hmm maybe we should just `#[rustfmt::skip]` on all the generated code?

cgwalters:
> Thank you! I think this may just be missing a cargo fmt?

I went ahead and pushed this.

cgwalters:
/override ci/prow/fcos-e2e
(duplicate)

cgwalters (approved):

Mstrodl:
Oops sorry. Only just saw this.

PR #3377, opened 2025-02-01
prepare-root: Log when we're mounting with verity required

cgwalters:
On general principle, but specifically as I wanted to cross-check this when debugging something else.


jmarrero (approved):
lgtm

PR #3379, opened 2025-02-03
zipl: remove 'sdboot' image before generating new one

nikita-dubrovskii:
Since s390-tools commit [f4cf4ae6ebb1](https://github.com/ibm-s390-linux/s390-tools/commit/195579cf0bcff320ed9ebe2cab6b59c9cc817f6e) (Remove genprotimg-C and switch to genprotimg-Rust implementation)
 `genprotimg` is just a symlink to `pvimg create`, which by default doesn't overwrite the output image,
  but introduces the `--overwrite` flag for this. Let's silently remove the file to support both tools.

Issue: https://github.com/openshift/os/issues/1731

### DIFF
--- a/src/libostree/ostree-bootloader-zipl.c
+++ b/src/libostree/ostree-bootloader-zipl.c
@@ -22,10 +22,12 @@
 #include "ostree-libarchive-private.h"
 #include "ostree-sysroot-private.h"
 #include "otutil.h"
+#include <fcntl.h>
 #include <gio/gunixinputstream.h>
 #include <string.h>
 #include <sys/mount.h>
 #include <sys/stat.h>
+#include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

cgwalters:
But is there something that can be done on the s390-tools side to remain compatible? Shouldn't this be viewed as a bug?

nikita-dubrovskii:
> But is there something that can be done on the s390-tools side to remain compatible? Shouldn't this be viewed as a bug?

This breaking change in `pvimg` tool was announced, so i doubt this new behaviour could be treated as a bug. Probably it was a bug-feature in the old `genprotimg`.

sdodson:
Still seems like we should file a RHEL bug, we may not be the only folks relying on the old behavior and RHEL should know about it even if we choose to work around the change.

sdodson:
https://issues.redhat.com/browse/RHEL-77592 Just so anyone finding this later sees we filed a bug

PR #3381, opened 2025-02-10
sepolicy: Add ostree_sepolicy_set_null_log

cgwalters:
Let's quiet the output of libselinux by default, all we get is the pointless "regex version mismatch" warnings. This is what e.g. systemd does.

This is exposed via the library; our binary opts in, then we expect higher level tools like e.g. bootc to do this too.


jlebon (approved):
Yes!

PR #3382, opened 2025-02-14
libostree: add private dependencies to pkg-config

alyssais:
This makes it possible to use pkg-config to link against libostree as a static library.  Unlike shared libraries, static libraries don't encode their dependencies in the library file, so we have to communicate them via pkg-config.


cgwalters:
Awesome, thanks!

cgwalters (approved):

PR #3385, opened 2025-02-25
core: Fix bare-user xattr canonicalization

cgwalters:
core: Print which xattrs are not sorted

I hit this in the rpm-ostree rechunker flow somehow
and with this change I now get:

`error: Generating commit from rootfs: Processing dir var: Writing content object: Incorrectly sorted xattr name (prev=user.Librepo.checksum.mtime, cur=security.selinux), index=2`

which is more useful, although I still need to figure out
and fix why that's happening (again?).

---

commit: Add errprefix for bareuser metadata

To aid debugging.

---

tests/basic: Add lots of user. xattrs

This exercises our requirement for xattr sorting.

---

core: canonicalize bare-user xattrs

Previously we were erroring out if xattrs were provided in
non-canonical (e.g. unsorted) form all the way down to just
the bare-user path. But for archive repos and dirmeta we
canonicalized.

Canonicalize bare-user xattrs on both read and write consistently
instead of erroring.

---


cgwalters:
Oh hmm fun I think we have some extra validation going on in the bare-user path that isn't happening in archive or bare.

jmarrero (approved):
lgtm

Issue #3386, opened 2025-02-26
test regression with recent gnupg patches from upstream: "public key not found" reported where expecting other errors


smcv:
Originally reported in Debian as <https://bugs.debian.org/1098951>.

Recently some patches were cherry-picked from the upstream stable branch of gpg, apparently to fix a denial of service (I'm sorry, I do not fully understand the situation or what these patches are aiming to prevent). https://salsa.debian.org/debian/gnupg2/-/commit/62d8d2f024d5e5c3289d5bf7892013dc18eac4b0 is believed to be the change that triggers this.

After that change, several tests in libostree fail. libostree is aiming to set up various situations in which a signing key or signature has become invalid, and it looks like in all cases signature verification does still fail, but it now fails in a different way that does not match the test's expectations:

* `/gpg-verify-result/valid-signature`, `/gpg-verify-result/missing-key`, `/gpg-verify-result/expired-signature` still have their expected results

* in `/gpg-verify-result/expired-key`, we expect `key_expired`, but what we actually now get is `key_missing`
    * similarly in `/gpg-verify-result/require-valid-signature-expired-key` we expect an expired key but get error message `Can't check signature: public key not found`

* in `/gpg-verify-result/revoked-key`, similarly, we get `key_missing`
    * similarly in `/gpg-verify-result/require-valid-signature-revoked-key` we get `Can't check signature: public key not found`

* and similarly `/gpg-verify-result/require-valid-signature-expired-missing-key` gets `Can't check signature: public key not found`, twice

Sorry, I don't know gpg well enough to know whether it is working as designed, or whether the patches contain a regression.

smcv:
See also https://dev.gnupg.org/T7527.

smcv:
Some of the shell-script-based tests are also failing with the new gpg, with symptoms that look similar ("public key not found" reported where it was not expected, and "Key expired" not reported where it was expected). So the changes to the C code above are not a complete answer to this.

ametzler:
See https://dev.gnupg.org/T7547 for the corresponding report against gnupg

PR #3387, opened 2025-02-26
test-gpg-verify-result: Show what the result was before asserting about it

smcv:
Helps: https://github.com/ostreedev/ostree/issues/3386

---

This does not do anything to solve the test failure, but hopefully it makes it easier to understand.

### DIFF
--- a/tests/test-gpg-verify-result.c
+++ b/tests/test-gpg-verify-result.c
@@ -211,6 +211,8 @@ test_attribute_basics (TestFixture *fixture, gconstpointer user_data)
       gboolean key_missing;
 
       tuple = ostree_gpg_verify_result_get_all (fixture->result, ii);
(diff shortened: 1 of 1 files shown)


cgwalters:
Just missing a `make clang-format`

smcv:
I think the result is less readable, but OK.

cgwalters:
If you feel even mildly strongly it's fine by me to slap [an ignore on it](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code) and manually format too.

cgwalters:
I guess either way it'd likely be more readable with a little helper `static inline yesno(v: uint8) -> char { v ? 'y' : 'n' }` or so? But again I have no strong opinion at this at all other than `make clang-format` is happy just so we have a default consistent baseline.

Issue #3242, opened 2024-05-13
support systemd soft reboots


cgwalters:
xref https://www.freedesktop.org/software/systemd/man/latest/systemd-soft-reboot.service.html

We could theoretically do this by default if we detect there's no kernel changes...but per the doc it will cause state leakage in corner cases around things like kernel sysctl.d, module loading blacklists, etc.

systemd (understandably) just made this an opt-in thing.  I think to start, we should do the same...just add the basic enablement features of setting up the next root.

Then users that want could invoke e.g.:

```
ostree prepare-next-root
```
And hmm... we should have a flag that detects the case where the kernel is skewed, and errors out by default.  But in theory we should also *allow* this, but ensure we've e.g. bind mounted in the old kernel data.

Mstrodl:
Kernel differences could be handled with kexec, see #435

cgwalters:
Implementing this is going to require some work on `ostree-prepare-root.service` to be able to run outside of the initramfs and target an explicit root. I don't think it'd be really hard but e.g. one thing we need to do in the general case is set up the composefs mount for example.

Mstrodl:
So, I've gotten ostree-prepare-root running on a booted system and appear to have successfully soft-rebooted!

Any opinions on how to handle kernel modules? I'd like to support the case (opt-in only!) where we soft-reboot with a mismatched kernel, but I'm not sure how best to do it. Some ideas:
* Bind mount from the old deployment
    * This is the obvious solution, but the problem is another upgrade could cause the old deployment to get deleted, which would cause issues, right?
* Copy modules from booted system into /run (preserved across soft reboots), then bind mount into the new deployment.
    * The disadvantage to this is that copying will be sorta slow
* Do a hardlink copy of /usr/lib/modules to somewhere on disk (maybe in /ostree/??), and bind mount that into /run/nextroot/usr/lib/modules
    * Disadvantage to this is that we'd need to build some cleanup infrastructure for the stored /usr/lib/modules
* Do a hardlink copy of /usr/lib/modules into the new deployment
    * Disadvantage to this is that now the deployment doesn't match the repo when we reboot...
* Ignore the problem, assume all the modules you need are already in memory
    * Obvious disadvantage is any modules that need to be dynamically loaded that the kernel doesn't already have in memory cannot be loaded, leading to confusing behavior for users

Let me know what you think!

cgwalters:
[...]

PR #3390, opened 2025-03-10
Preliminary soft reboot support

Mstrodl:
Most of the easy stuff from #3242

One big thing not handled here is detecting a difference in kernels and mounting in the old kernel's `/lib/modules/$(uname -r)`. Probably we shouldn't even be doing that much by default, and instead error out without the user explicitly asking for their kernel to have weird problems.

One issue I ran into is that the upgrade process creates a mount namespace so /sysroot can be written to, but we need to escape from it when preparing `/run/nextroot`. To work around this, I keep an FD of `/proc/self/ns/mnt` so we can `setns` to it. Unfortunately, we have to do this in a separate fork()'d process, because `setns` will fail if we have any threads (I couldn't figure out how to stop GIO's main loop...).

I'm not thrilled about the setns() stuff or needing to parse through `/proc/self/mounts`, so if anyone has any ideas on how to do it more elegantly, I'm all ears. One option is to set up `/run/nextroot` and switch namespaces in the setup part of `g_spawn_sync`, which would save a fork, but I'm not sure if it's worth it.


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -4259,6 +4259,102 @@ ostree_sysroot_deployment_set_mutable (OstreeSysroot *self, OstreeDeployment *de
...
+  gint exit_status;
+  g_autoptr (GPtrArray) args = g_ptr_array_new ();
+
+  /* We cannot just make a bind mount, because when the soft reboot happens, our
+     current root will be unmounted, and the bind mount will break. Therefore,
+     we have to recreate the mount at a different location. */
+
+  /* Read in mounted FSes */
+  g_autofree gchar *mounts_contents = NULL;
+  if (!g_file_get_contents ("/proc/self/mounts", &mounts_contents, NULL, error))
```
cgwalters, line 4291: There's libmount for parsing this stuff

src/libostree/ostree-sysroot-upgrader.c
```diff
@@ -628,6 +649,36 @@ ostree_sysroot_upgrader_deploy (OstreeSysrootUpgrader *self, GCancellable *cance
                                                      error))
             return FALSE;
         }
+
+      if ((self->flags & OSTREE_SYSROOT_UPGRADER_FLAGS_SOFT_REBOOT) > 0)
+        {
+          /* Need to fork because setns will fail if we have threads (GIO loop) */
+          pid_t child_pid = fork ();
```
cgwalters, line 656: Right, but the problem here is that in the general case now we can't call any function which may have relied on global state - especially like the glib worker thread.

I used to help maintain glib upstream and I probably would have been sympathetic to this situation but just closed as wontfix any bugs arising in this situation.

In bootc I did something *fairly* similar to this in https://github.com/bootc-dev/bootc/blob/f818b37a57a9cdaf657e55c84de8d6e773f64e97/lib/src/mount.rs#L171 and I was pretty careful about what APIs I called there.

And even there I was pretty tempted to rework to do a "re-exec self" dance (ref https://github.com/bootc-dev/bootc/issues/799#issuecomment-2567798152 )

---

OTOH this won't be our first instance of similar code...the `fork()` we also do for `fsfreeze` is related.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -4259,6 +4259,102 @@ ostree_sysroot_deployment_set_mutable (OstreeSysroot *self, OstreeDeployment *de
...
+    }
+
+  g_set_error_literal (error, G_IO_ERROR, G_IO_ERROR_FAILED,
+                       "Couldn't find mountpoint for /sysroot!");
+  return FALSE;
+
+end_loop:
+
+  /* At this point, /sysroot is mounted at /run/nextroot, do the pivot from there: */
+  g_ptr_array_add (args, "/usr/lib/ostree/ostree-prepare-root");
```
cgwalters, line 4332: Though maybe one alternative is to move this handling directly into `ostree-prepare-root`? Uncertain.

src/switchroot/ostree-prepare-root.c
```diff
@@ -259,13 +259,27 @@ main (int argc, char *argv[])
...
+    booted = (stbuf.st_mode & S_IFMT) == S_IFREG;
+
+  g_autofree char *kernel_cmdline;
+  if (argc < 3)
+    {
+      kernel_cmdline = read_proc_cmdline ();
+    }
+  else
+    {
+      kernel_cmdline = argv[2];
```
cgwalters, line 280: Awesome! This looks like a really nice change and should crucially really help us for testing ostree-prepare-root outside of the initramfs in general (should be totally doable to run in a container)

cgwalters (commented):
Thanks so much for working on this!! Just a preliminary review pass so far

src/libostree/ostree-sysroot-upgrader.c
Mstrodl: Yeah... I'm not thrilled about this. I couldn't find a good way to stop the GIO loop for this. Can you think of any better ways to do it? One option is to have ostree-prepare-root `nsenter` and set up the `/run/nextroot` mount. This might be doable by looking at `/proc/1/ns/mnt`, but I'm not sure. Or we could nsenter as part of the glib spawn (we're forking there anyways)

Mstrodl: Let me know what you think is best and we can do it :)
* Obviously keeping things the way they are is an option (forking isn't great though...)
* I have a suspicion the `/proc/1/ns/mnt` option could cause issues/weird behavior, so I'd prefer not to do that.
* ostree-prepare-root mounting `/run/nextroot` feels like it's a little out of scope for the tool (But also then we can nsenter in the fork created for the exec...)

cgwalters: Ultimately the replacement for `fork()` I think will be https://lwn.net/Articles/908268/
But Josh went back to working on Rust upstream I think and it doesn't have an active driver right now.

Anyways, the problem is there's a lot of APIs invoked here after `fork()` in `ostree_sysroot_deployment_prepare_next_root()` like e.g. `ot_journal_print()` and such.

I mentioned this above and we can try to prove this out but the most reliable thing is to do a re-exec of the current process. *Except* the massive complication is that libostree is a library...so we'd need to require all callers to give us the chance to intercept re-invocations in `main()` (doable, just ugly) or alternatively split out a little `/usr/libexec/libostree-helper` binary.

But probably the best alternative in this specific case is as you say

> One option is to have ostree-prepare-root nsenter and set up the /run/nextroot mount. 

If we can do as much parsing as possible before forking, and then just hand off to it the operations to do that seems likely to be tractable, right?

Mstrodl: One thing that worries me is that the setns call might not work from ostree-prepare-root, but I guess I can't really be certain until I try it :) Definitely the FD I keep open for the mountns won't get passed on, so it'd need to look at pid 1's mountns, which might be okay?

Ideally it would be nice to have a setns thing in io_uring. I suspect a lot of the container folks would like that too.

I agree reexec is really not the solution.

Having ostree-prepare-root do the mounting shouldn't be too crazy.

src/switchroot/ostree-prepare-root.c
ericcurtin: Simple and effective :)

cgwalters:
Let's commit to this one for the next release! I think  Joseph is going to help review too.  Definitely excited about landing this!

cgwalters:
Chatting about this with @jmarrero now, one thing I think would be helpful is (and this mirrors the kexec discussion IIRC) is basically:

- Add `ostree admin prepare-soft-reboot <deployment>` (where deployment is an integer like `undeploy` takes)
- That CLI verb calls an API exactly like `ostree_sysroot_deployment_prepare_next_root` except it's that API which auto-detects if it's in the mount namespace of pid1 (check `/proc/1/ns/mnt`) and if not, does the `fork()` dance internally or probably better (to avoid the `fork()` concerns) is it actually runs `ostree internals prepare-next-root-impl` or so?

cgwalters:
Also to write down a concern from the chat there is that putting this in `OstreeSysrootUpgrader` is a problem specifically because both rpm-ostree and bootc don't use that code (in rpm-ostree's case it has a [much more complex fork](https://github.com/coreos/rpm-ostree/blob/main/src/daemon/rpmostree-sysroot-upgrader.cxx) and in bootc's case because it *always* goes through the container flow and there's not much helpful in `OstreeSysrootUpgrader` for that.

So by moving the fork dance out of the upgrader and into the core `OstreeSysroot` API it is immediately usable by both of those projects (and probably more).

Bigger picture again I think we want to be able to use this even without "upgrading" - it could make total sense to soft-reboot into a rollback deployment for example.

Mstrodl:
Sorry I haven't had much time to work on this. Last time I was testing this I was seeing some weird behavior that I needed to debug, so it's possible this isn't 100% sound.

Issue #3391, opened 2025-03-12
unlock: Change to use deployment backing dir by default


cgwalters:
Talked with a customer who had a dedicated partition for `/var/tmp` that was relatively small; this means the scale of changes that can be done with `usroverlay` was limited. I think it'd be a good general cleanup to switch to use the deployment backing directory by default now that we have that.

PR #3397, opened 2025-03-20
unlock: Use deployment backing dir

cgwalters:
Closes: https://github.com/ostreedev/ostree/issues/3391


cgwalters:
Only compile tested so far

jlebon (approved):
Makes sense to me!

PR #3403, opened 2025-04-04
docs: removed unused argument

ChilloManiac:
Looking at the implementation [here](https://github.com/ostreedev/ostree/blob/main/src/ostree/ot-remote-builtin-delete-cookie.c#L48) we can see the $VALUE is not used. 

It would also be weird to have to supply a VALUE to a delete operation


jlebon (approved):
Thanks!

PR #3405, opened 2025-04-06
sysroot: Detect early on when /boot is on vfat

cgwalters:
We do want to support this (as part of supporing the Boot Loader Spec) but because we use symlinks in `/boot`, can't yet.

Error out very early on consistently if we detect
vfat for /boot, but also add a member variable to keep track of this in preparation for supporting it.

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -22,8 +22,10 @@
 
 #include "otutil.h"
 #include <err.h>
+#include <linux/magic.h>
 #include <sys/file.h>
 #include <sys/mount.h>
+#include <sys/vfs.h>
 #include <sys/wait.h>
 
 #include "ostree-bootloader-aboot.h"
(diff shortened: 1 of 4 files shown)


cgwalters:
cc @igoropaniuk - can you review this?

cgwalters:
OK I've rebased and tested this one.

src/libostree/ostree-sysroot.c
```diff
@@ -338,6 +340,19 @@ ensure_sysroot_fd (OstreeSysroot *self, GError **error)
   return TRUE;
 }
 
+static gboolean
+validate_boot_fd (OstreeSysroot *self, int fd, GError **error)
+{
+  g_assert_cmpint (fd, !=, -1);
+  struct statfs stbuf;
+  if (fstatfs (fd, &stbuf) < 0)
+    return glnx_throw_errno_prefix (error, "fstatfs");
```
champtar, line 349: don't we want to know which fstatfs failed ? 
```suggestion
    return glnx_throw_errno_prefix (error, "fstatfs(boot_fd)");
```

cgwalters: Sure, done

champtar (approved):

PR #3406, opened 2025-04-08
Revert "generator: Exit if there's no `/run/ostree`"

vilkov:
This reverts commit b9ce0e89801bbc92d50473d3620b3f41f1dbef9f.

>   The comment here said why we didn't do that before, but that's
>   for the really legacy embedded-only "ostree-prepare-root-static"
>   path, and even then I'm pretty sure it was wrong because
>   the generator here only runs in the *real* root, and we should
>   have `/run/ostree` at that point.

This is wrong assumption because in case of "ostree-prepare-root-static" path, "/run" does not exist yet, when generator runs.

According to SYSTEMD_FILE_HIERARCHY_REQUIREMENTS.md:
- `/run` must exist as an empty mount point and will automatically be mounted by systemd with a tmpfs.

### DIFF
--- a/src/libostree/ostree-impl-system-generator.c
+++ b/src/libostree/ostree-impl-system-generator.c
@@ -289,12 +289,6 @@ _ostree_impl_system_generator (const char *normal_dir, const char *early_dir, co
   if (unlinkat (AT_FDCWD, INITRAMFS_MOUNT_VAR, 0) == 0)
     return TRUE;
 
-  // If we're not booted via ostree, do nothing
-  if (!glnx_fstatat_allow_noent (AT_FDCWD, OTCORE_RUN_OSTREE, NULL, 0, error))
-    return FALSE;
-  if (errno == ENOENT)
-    return TRUE;
-
   g_autofree char *cmdline = read_proc_cmdline ();
   if (!cmdline)
     return glnx_throw (error, "Failed to read /proc/cmdline");


champtar:
@vilkov maybe you can detail a bit your setup

vilkov:
BTW, @cgwalters why do think systems without initramfs are legacy?
IMHO, these days we don't need initramfs at all. When you have UEFI, uBoot, etc which are able to boot your kernel directly from a bunch of devices and filesystems why would you need initramfs at all?

vilkov:
@ch

> @vilkov maybe you can detail a bit your setup

Sure, I use only "ostree-prepare-root-static" and don't need initramfs at all. Without this patch OSTree won't generate "/var" bind. And we can't create those directories in "ostree-prepare-root-static" because systemd will mount tmpfs on "/var" later on.

cgwalters:
> BTW, @cgwalters why do think systems without initramfs are legacy?

Well, obviously my perspective is a bit warped by the fact that most of my day is working on non-embedded systems; or at least large enough that we don't worry about an initramfs.

This all said...it seems to me that one could get a really small but still non-statically linked initramfs that's basically systemd (or even a very small init) that just does the bare minimum of things (mount API filesystems mainly?) and runs ostree-prepare-root and it wouldn't be much overhead.

If the initramfs is built into the kernel (UKI style instead of being a separately compressed file) there wouldn't be any (more) I/O overhead, so it's mostly just about CPU.

cgwalters:
@vilkov can you look at https://github.com/ostreedev/ostree/pull/3410 ?

vilkov:
Hi @cgwalters, sure I will have a look this weekend.
Sorry for delay - I'm testing my new distro for PC, built from scratch (OStree based btw :)), so it was a busy week tweaking it here and there.

vilkov:
> Well, obviously my perspective is a bit warped by the fact that most of my day is working on non-embedded systems; or at least large enough that we don't worry about an initramfs.
> 
> This all said...it seems to me that one could get a really small but still non-statically linked initramfs that's basically systemd (or even a very small init) that just does the bare minimum of things (mount API filesystems mainly?) and runs ostree-prepare-root and it wouldn't be much overhead.
> 
> If the initramfs is built into the kernel (UKI style instead of being a separately compressed file) there wouldn't be any (more) I/O overhead, so it's mostly just about CPU.

Thanks @cgwalters, got your point. You are right, from my point of view (distro developer) initramfs is just an extra step in build process which can easily be avoided considering modern bootloaders and firmwares.

vilkov:
> > This is wrong assumption because in case of "ostree-prepare-root-static" path, "/run" does not exist yet, when generator runs.
> 
> I'm pretty sure that systemd mounts the API filesystems (including `/run`) _before_ running generators. Are you sure that's the bug in your setup?
> 
> I think the real problem here is more that in this case of a static prepare root we never write `/run/ostree` at all.
> 
> It's a bit tricky as what we really want is a way to pass state from the static prepare-root path down to the generator, but the obvious ones I can think of don't exist in this flow...setting an environment variable would be too ugly; it'd propagate into child processes.
> 
> Note this commit was also attempting to fix something, which a revert re-introduces (and my bad for not adding an integration test for the fix at the time of the commit). It's pretty trivial to test as well, basically just run a bootc container (or really any container with ostree in it via systemd).

Yes, I think you are correct about the sequence, my description is probably wrong... Lack of sleep hasn't went unnoticed :)
Here is the whole thing:
  1. '/run/ostree' is being created by ostree-prepare-root
  2. In case of initramfs '/run' is already available when ostree-prepare-root runs
BUT, in case of ostree-prepare-root-static we can not touch '/run' yet - we are the first process in the system and systemd haven't mounted anything to '/run' yet.

vilkov:
I think this PR can be closed in favor of #3410 which is much better solution.

PR #3407, opened 2025-04-08
sysroot: Load bootloader configs via boot_fd

cgwalters:
This was a general principle cleanup, preparation
for handling VFAT for /boot for systemd-boot/BLS
support.

However I ran into an ugly corner case in our
unit tests that pointed at a sysroot without a
boot directory. The previous logic handled
ENOENT for boot/loader but not /boot.
Continue to cope with that degenerate situation.

### DIFF
--- a/src/libostree/ostree-sysroot.c
+++ b/src/libostree/ostree-sysroot.c
@@ -338,9 +338,38 @@ ensure_sysroot_fd (OstreeSysroot *self, GError **error)
   return TRUE;
 }
 
+/* Require that both self->sysroot_fd is set.
+ * If the sysroot has a boot/ subdirectory, it will be loaded.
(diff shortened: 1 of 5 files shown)


src/libostree/ostree-sysroot.c
```diff
@@ -338,9 +338,38 @@ ensure_sysroot_fd (OstreeSysroot *self, GError **error)
   return TRUE;
 }
 
+/* Require that both self->sysroot_fd is set.
```
allisonkarlitskaya, line 341: Both self->sysroot_fd and... what?

allisonkarlitskaya (approved):
Looks good, but you've got narrow text width in your commit messages again :)

PR #3410, opened 2025-04-09
generator: Still create /run/ostree in static prepareroot path

cgwalters:
Ref https://github.com/ostreedev/ostree/pull/3406

There's a combination of two commits here that broke the static prepareroot path:

ec1109c7a93a2ed07503b12ffecf7048cf7cc0d0
"generator: Stop creating `/run/ostree-booted`"
and more recently
b9ce0e89801bbc92d50473d3620b3f41f1dbef9f
generator: Exit if there's no `/run/ostree`

Basically when run via a non-static prepareroot we create `/run/ostree-booted` consistently in the initramfs, using the kernel argument presence as source of truth.

But for the static prepareroot, the generator checked the kernel argument, and had a fallback of creating it.

Except that's busted in the case of running in a
container, where with many runtimes we still
default to seeing the host's commandline (which
is basically wrong...but fixing that requires a
userspace virtualizer/interceptor for `/proc`
so it's not commonly done).

This should fix the static prepareroot path
by detecting the case where we're compiled
with a static prepareroot, and if so we then
hardcode creating the `/run/ostree-booted`
file in the generator. I think basically
everyone who is compiling ostree with
a static prepareroot *and* including it
in their filesystem trees can be pretty
much guaranteed to be actually using it.

### DIFF
--- a/src/libostree/ostree-impl-system-generator.c
+++ b/src/libostree/ostree-impl-system-generator.c
@@ -289,11 +289,18 @@ _ostree_impl_system_generator (const char *normal_dir, const char *early_dir, co
   if (unlinkat (AT_FDCWD, INITRAMFS_MOUNT_VAR, 0) == 0)
     return TRUE;
 
+#ifdef OSTREE_PREPARE_ROOT_STATIC
+  // Create /run/ostree-booted now, because other things rely on it.
+  // If the system compiled with a static prepareroot, then our generator
+  // makes a hard assumption that ostree is in use.
+  touch_run_ostree ();
+#else
   // If we're not booted via ostree, do nothing
   if (!glnx_fstatat_allow_noent (AT_FDCWD, OTCORE_RUN_OSTREE, NULL, 0, error))
     return FALSE;
   if (errno == ENOENT)
     return TRUE;
+#endif
 
   g_autofree char *cmdline = read_proc_cmdline ();
   if (!cmdline)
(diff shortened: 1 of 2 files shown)


champtar:
@vilkov do you think you can test this ?

vilkov (approved):
Works as expected, tested with a fresh OSTree-based image. Thanks!

src/libostree/ostree-impl-system-generator.c
```diff
@@ -289,11 +289,18 @@ _ostree_impl_system_generator (const char *normal_dir, const char *early_dir, co
   if (unlinkat (AT_FDCWD, INITRAMFS_MOUNT_VAR, 0) == 0)
     return TRUE;
 
+#ifdef OSTREE_PREPARE_ROOT_STATIC
+  // Create /run/ostree-booted now, because other things rely on it.
+  // If the system compiled with a static prepareroot, then our generator
+  // makes a hard assumption that ostree is in use.
+  touch_run_ostree ();
```
champtar, line 296: If the touch fails we don't log anything and we just continue running

cgwalters: True, but this was also the case before; and there's even a comment in the source code that that's intentional. I think the only real thing that could fail here is ENOSPC but if you hit that this early on with `/run` the system is going to fall over in a ton of ways regardless.

I don't mind changing this of course, but not sure about doing it in this PR.

champtar: I was thinking if /run is not created yet, or read-only because we have not mounted a tmpfs on top yet
I would at least log an error to help with debugging in any cases, but this can be done in another PR

champtar (approved):

PR #3419, opened 2025-05-11
Update ostree-prepare-root.c:: Need to use relative path instead of /

codeshravan:
Need to use relative path instead of absolute


codeshravan:
@cgwalters kindly review this

PR #3420, opened 2025-05-15
ot-builtin-admin: Add admin prepare-soft-reboot

jmarrero:
---

Work on top of #3390 specifically working on implementing:
https://github.com/ostreedev/ostree/pull/3390#issuecomment-2880991044

---

TODO list

- [x] optional but nice: factor out prep commit(s): option processing, mkdir idempotence
- [X] default deny soft rebooting to a deployment with a different kernel
- [x] `ostree admin status` may need some tweaking
- [x] `ostree admin prepare-soft-reboot` is too long
- [x] Need to ensure handling of /boot (this is why bootc status etc are broken post soft-reboot)
- [x] Also need to force deployment finalization for staged deployments to ensure we get the updated /etc but also so that we have bootloader entries across reboot etc. TODO *should* happen via ostree-finalizes
- [x] tests
- [x] docs (man ostree-soft-reboot?)
- [ ] TODO mechanism for tracking that we soft-rebooted (systemd is doing this a bit?); I saw a soft reboot counter but we should show it in "ostree admin status" or so...hmm how about adding a flag to `/run/ostree-booted` for this?
- [ ] support for opting in to kernel skew; should bind mount old kernel dir but also make very clear in `status` that skew has occurred (may want to drill into kernel vs initramfs change)
- [ ] handle detecting kernel argument changes (and this one gets into bootc kargs vs ostree)
- [ ] determine behavior with staged deployments present; should soft rebooting clear the staged deployment? I think so...
- [ ] (related to the above) possibly move soft reboot prep to finalization (shutdown) time; this would be cleaner conceptually


src/ostree/ot-main.c
```diff
@@ -288,6 +288,13 @@ ostree_run (int argc, char **argv, OstreeCommand *commands, GError **res_error)
       command++;
     }
 
+  int parent_mountns = open ("/proc/self/ns/mnt", O_RDONLY | O_NOCTTY | O_CLOEXEC);
```
cgwalters, line 291: The problem with this is that it will also need to be duplicated in bootc (if we were to use this as a shared library, which maybe we don't need to).

The more complex way to do this would be to have the library fork/exec itself, but that may not be worth it.

We have a flag for "mount namespace is unshared" and perhaps we should assert if it's set?

jmarrero:
```
[pid 54384] execve("/usr/lib/ostree/ostree-prepare-root", ["/usr/lib/ostree/ostree-prepare-root", "--mount", "/run/nextroot", "ostree/deploy/fedora/deploy/8b86ef33d4efb547ad8dd1562301a09cf7ad166f23cec4857bac69432b0d4417.0"]
```

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -4270,6 +4261,120 @@ ostree_sysroot_deployment_set_mutable (OstreeSysroot *self, OstreeDeployment *de
...
+  g_autofree char *deployment_path = ostree_sysroot_get_deployment_dirpath (self, deployment);
+  gint estatus;
+
+  glnx_autofd int deployment_dfd = -1;
+  if (!glnx_opendirat (self->sysroot_fd, deployment_path, FALSE, &deployment_dfd, error))
+    return FALSE;
+
+  const char *argv[] = { "/usr/lib/ostree/ostree-prepare-root", "--mount", "/run/nextroot",
+                         // we need the fd here instead of the path.
+                         deployment_path, NULL };
```
jmarrero, line 4290: I forgot removing this... but I also forgot if passing _ostree_sysroot_child_setup_fchdir was enough in the call.

jmarrero:
I feel I forgot how to properly pass the fd.
```
sudo strace -f -e execve ./ostree admin prepare-soft-reboot 0
```
```
[pid 58220] execve("/usr/lib/ostree/ostree-prepare-root", ["/usr/lib/ostree/ostree-prepare-root", "--mount", "/run/nextroot"], 0x10cad1b0 /* 26 vars */) = 0
ostree-prepare-root: realpath("--mount"): No such file or directory
[pid 58220] +++ exited with 1 +++
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=58220, si_uid=0, si_status=1, si_utime=0, si_stime=0} ---
error: Preparing /run/nextroot for a soft-reboot: Child process exited with code 1
+++ exited with 1 +++
```

Mstrodl:
Thanks for all the hard work everyone. Glad you guys were able to track down the weird behavior I was seeing. Sorry I didn't have time to give this the love it needed. Should I close my PR?

cgwalters:
@jmarrero can you review the latest two commits?

Also with this, the soft reboot test is passing!

jmarrero:
They both look good to me @cgwalters. Also I tested it manually and it works too.

cgwalters:
> That works for me, do you want to remove the WIP: from the commit?

Done!

> [2025-06-26T16:43:17.369Z] error: Installing packages: Packages not found: bootupd

I think this is a usual transient race condition with the FCOS koji pool, will wait and retry in a bit...

PR #3421, opened 2025-05-16
Move /run  from initramfs to real-rootfs(sysroot/run)

codeshravan:
Moved /run mount point from initramfs(/run) to realrootfs(sysroot/run) so that systemd will not mount it again if it finds a mount point during switchroot and all the contents of /run will be moved to realrootfs as well. It will fix the problem of mounting sysroot as read-only because during ostree-remount.service it checks in /run/ostree-booted if we need to mount it read-only only or not

### DIFF
--- a/src/switchroot/ostree-prepare-root.c
+++ b/src/switchroot/ostree-prepare-root.c
@@ -688,11 +688,12 @@ main (int argc, char *argv[])
   if (rmdir (TMP_SYSROOT) < 0)
     err (EXIT_FAILURE, "couldn't remove temporary sysroot %s", TMP_SYSROOT);
   
-  /* Moves /run mount point from initramfs(/run) to realrootfs(sysroot/run)
-  * so that systemd will not mount it again if it finds a mount point during switchroot.
-  * and all the contents of /run will be moved to realrootfs as well. It will fix the
-  * problem of mounting sysroot as read-only because during ostree-remount.service 
-  * it checks in /run/ostree-booted if we need to mount it read-only only or not
+  /* Move /run mount point from initramfs (/run) to the real root filesystem (sysroot/run).
+  * This ensures that systemd does not remount it during switchroot if it detects an existing mount point.
+  * Consequently, all contents of /run will be transferred to the real root filesystem.
+  * This adjustment resolves the issue of mounting sysroot as read-only. During the execution 
+  * of ostree-remount.service, the system checks /run/ostree-booted to determine if sysroot should be 
+  * mounted as read-only.
   */
   if (mount("/run", "run", NULL, MS_MOVE | MS_SILENT, NULL) < 0)
     err(EXIT_FAILURE, "failed to MS_MOVE '/run' to 'sysroot/run'");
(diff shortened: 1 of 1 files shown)


champtar:
@codeshravan if I follow correctly, you are using a "non classic" initrd, what happens in the case of a dracut initrd ? 
You code fix your issue but might regress everyone else, everything that runs between ostree-prepare-root and switch-root now has an empty /run, maybe even read only.

cgwalters:
Yes...I'm not inherently opposed to trying to make this work but it is quite hard to support what we have now with the static prepare root and a full systemd-in-initramfs, this middle ground is quite tricky.

Have you taken a look at running systemd in the initramfs? It can be trimmed down quite a bit IME. Alternatively, whatever you can do to emulate the minimal subset of it would be really helpful. In particular, as I understand it you could also do this move operation in your own initramfs code.

ricardosalveti:
> Have you taken a look at running systemd in the initramfs?

Having systemd in the initramfs is something we're experimenting in OE, but we will always have use cases for not running it there (due flexibility of the project).

cgwalters:
OK, I understand. But what about my suggestion of having whatever initramfs infrastructure is there today handle `/run` in the same way systemd does it? Basically it should move `/run` to the target root by default, in this case after `ostree-prepare-root` or whatever other bits are run.

codeshravan:
> @codeshravan if I follow correctly, you are using a "non classic" initrd, what happens in the case of a dracut initrd ? You code fix your issue but might regress everyone else, everything that runs between ostree-prepare-root and switch-root now has an empty /run, maybe even read only.



> OK, I understand. But what about my suggestion of having whatever initramfs infrastructure is there today handle `/run` in the same way systemd does it? Basically it should move `/run` to the target root by default, in this case after `ostree-prepare-root` or whatever other bits are run.

roshan20039:
Could someone explain the flow in any other initramfs (for eg dracut) and tell how this change would break functionality?
Because as per my understanding, the existing code was intended to create contents inside sysroot/run and then switchroot to sysroot so that we have all contents on /. Isn't it? Why would we need to access /run between ostree-prepare-root and switchroot? Can we not treat them together as an atomic operation?

champtar:
In an initramfs built with dracut / with systemd,
we have `/run/systemd` that contains a bunch of runtime conf for systemd, so moving /run is really a bad idea IMO

Looking at an EL 9.5 initramfs, ostree-prepare-root.service runs before
- initrd-parse-etc.service
- initrd-cleanup.service
- dracut-mount.service
- initrd-switch-root.service
(and likely other)
so there is a big gap between ostree-prepare-root and switch-root

initrd-switch-root.service just calls `systemctl --no-block switch-root /sysroot`
https://github.com/systemd/systemd/blob/main/src/shared/switch-root.c
Looking at the code it does a bind mount of /run, so we keep the content of /run

codeshravan:
> OK, I understand. But what about my suggestion of having whatever initramfs infrastructure is there today handle `/run` in the same way systemd does it? Basically it should move `/run` to the target root by default, in this case after `ostree-prepare-root` or whatever other bits are run.

@cgwalters If we change in initrd that also works fine because we mounted /sysroot/sysroot/ as read only not /sysroot, 
https://github.com/advancedtelematic/meta-updater/pull/834

@ricardosalveti Would you please review this and see if this is solving our purpose and will not impact anything as mentioned by @champtar .

codeshravan:
Moved the code to initrd

PR #3423, opened 2025-05-19
man: Document `ostree admin unlock --transient`

evan-goode:
I noticed `--transient` is missing from the [ostree-admin-unlock man page](https://ostreedev.github.io/ostree/man/ostree-admin-unlock.html).

Let me know whether the included `programlisting` is appropriate or not. It was helpful for me to see how to remount the overlay by reading `tests/kolainst/destructive/unlock-transient.sh`, so it may help a man page reader too.


cgwalters (approved):
Thanks!

Issue #2857, opened 2023-05-10
drop sh-inline


cgwalters:
It's deprecated https://github.com/cgwalters/rust-sh-inline/commit/dfeccfc247d1392074fb7bb3e0fd41f66a2ab6e8

See https://github.com/ostreedev/ostree/pull/2856 which started porting.  This is a low priority background task that can be done incrementally.

samzeter:
I've done some more work on this, although I had two questions:
- How to run the tests found in tests/inst/src/ ? I saw some mention in the git log about 'kola' ?
- I wasn't quite sure what the substitute for use sh_inline::{bash_command}; and it's unwrap() method should be

PR #3424, opened 2025-05-20
tests/inst: convert more sh-inline to xshell

samzeter:
Part of #2857


cgwalters:
Thank you for working on this! Can you rebase on git main to pick up recent CI fixes?

cgwalters:
Failure in https://github.com/ostreedev/ostree/actions/runs/15202426856/job/42758861826?pr=3424 needs fixing

PR #3426, opened 2025-05-21
ci: Drop --fast from buildextend-live

cgwalters:
This got dropped apparently with the osbuild rewrite.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -68,7 +68,7 @@ cosaPod(runAsUser: 0, memory: "9Gi", cpu: "4") {
       coreos-assembler build
       coreos-assembler buildextend-metal
       coreos-assembler buildextend-metal4k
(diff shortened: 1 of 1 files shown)


.cci.jenkinsfile
```diff
@@ -68,7 +68,7 @@ cosaPod(runAsUser: 0, memory: "9Gi", cpu: "4") {
       coreos-assembler build
       coreos-assembler buildextend-metal
       coreos-assembler buildextend-metal4k
-      coreos-assembler buildextend-live --fast
+      coreos-assembler buildextend-live
```
jlebon, line 71: Note nowadays this can be written as

```
coreos-assembler osbuild metal metal4k live
```

So it's all done in a single invocation which is a bit more efficient.

jlebon (approved):

jlebon (approved):

PR #3437, opened 2025-06-02
A few buildsystem fixes

cgwalters:
apidoc: Quiet many warnings

Should have done this long ago, this greatly reduces the spam
in the terminal logs from builds.

---

dockerignore: Add

This ensures we don't pick up things we shouldn't from the source.

---

ci: Updates for centos builds

- Do the modern way to enable the buildroot with crb and epel

---

tests/libtest: Just use python as a webserver if no libsoup

We only have a very few tests that actually need what
we have in ostree-trivial-httpd that supports things like serving
random 500 errors etc.

If we don't have libsoup, then just use a python webserver.

---

### DIFF
--- a/ci/libbuild.sh
+++ b/ci/libbuild.sh
@@ -6,6 +6,10 @@ OS_ID=$(. /etc/os-release; echo $ID)
 OS_ID_LIKE=$(. /etc/os-release; echo $ID ${ID_LIKE:-})
 OS_VERSION_ID=$(. /etc/os-release; echo $VERSION_ID)
 
+# Drop our content underneath target/ by default as it's already
+# ignored by rust
+BUILDDIR=target/c
+
 pkg_upgrade() {
     dnf -y distro-sync
 }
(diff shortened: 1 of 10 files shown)


ci/libbuild.sh
```diff
@@ -16,8 +20,11 @@ make() {
 
 build() {
     env NOCONFIGURE=1 ./autogen.sh
-    ./configure --sysconfdir=/etc --prefix=/usr --libdir=/usr/lib64 "$@"
-    make V=1
+    mkdir -p target/c
+    (cd target/c
+     ../../configure --sysconfdir=/etc --prefix=/usr --libdir=/usr/lib64 "$@"
```
jmarrero: builds seem to be complaining about no Makefile, this might be slightly off?

jmarrero (changes requested):

ci/libbuild.sh
cgwalters: Yeah, had to adjust various things that now need to know to run `make` in `target/c`

PR #3451, opened 2025-06-26
status: Add --json output

cgwalters:
This is way long overdue, and will be useful for a variety of things
but is especially motivated by soft reboot testing.


jmarrero (approved):
lgtm

PR #3452, opened 2025-06-26
status: Add `soft-reboot-target` to JSON

cgwalters:
It's there in the human readable status.

And update the soft-reboot test suite to assert things via JSON since it's *way* cleaner.


jmarrero (approved):
lgtm

PR #3457, opened 2025-06-30
deploy: Default quiet for forked systemctl

cgwalters:
I'm checking over the codebase for places we fork children; doing so from a library means we may end up printing to stdout/stderr of our caller, which we shouldn't do.

- Also add an error prefix with information about the service if we fail


jmarrero (approved):
lgtm

PR #3459, opened 2025-07-01
sysroot: Cache deployment device/inode

cgwalters:
This is a lot cleaner. In particular, it's prep
for having soft reboot finalize at shutdown time, because we weren't correctly handling the staged deployment in that case.


cgwalters:
OK this one got bigger, but I think it will really help clean up the soft reboot work, where identifying unique deployments that we've serialized is very important.

jmarrero (approved):
lgtm

PR #3464, opened 2025-07-08
docs: Remove <authorgroup>

cgwalters:
We've just been copy pasting this; for some commands it's true, but others have been written by other people. Most of the commands just call into the library code which is written by various people.

Because it isn't really accurate, just drop it.


jmarrero (approved):
lgtm

PR #3465, opened 2025-07-09
Soft reboot kargs check

cgwalters:


jmarrero (approved):
lgtm

cgwalters:
Yesterday it's centos stream that was broken because of rpm-md skew (ref https://lists.centos.org/hyperkitty/list/devel@lists.centos.org/thread/JZDLBY3EGNZDDVJQJVSXV7GHQQEDSMDB/ and now fixed!) and today it's FCOS lockfile skew with the rpm-md repos...

Issue #3471, opened 2025-07-16
Add `root.transient-ro = true`


cgwalters:
See https://github.com/bootc-dev/bootc/discussions/1036

Add `root = transient-ro`

Like `root = transient` but we'd allocate the overlayfs upper, but still keep it read-only by default. This would make it easy for code running in the real root to unshare the mount namespace, mount it writable and mutate it while still keeping it read-only for most use cases. This would be a pretty easy addition to ostree-prepare-root.

PR #3473, opened 2025-07-17
Add root.transient-ro

cgwalters:
An example use case for this is having privileged code add dynamic new toplevel mountpoints (that don't persist across reboots/upgrades), while still keeping the rootfs readonly for processes by default.

Closes: https://github.com/ostreedev/ostree/issues/3471


jmarrero (approved):
lgtm

PR #3494, opened 2025-08-06
deploy: Remove some dead aboot code

alexlarsson:
The code in get_kernel_from_tree_usrlib_modules() that checks for an "aboot.cfg" does nothing with it (and in fact may leak the fd in case its there). In practice, /usr/lib/modules never has an aboot.cfg, so this is just completely dead code.

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1211,10 +1211,6 @@ get_kernel_from_tree_usrlib_modules (OstreeSysroot *sysroot, int deployment_dfd,
     }
   glnx_close_fd (&fd);
 
-  /* look for a aboot.cfg file. */
-  if (!ot_openat_ignore_enoent (ret_layout->boot_dfd, "aboot.cfg", &fd, error))
-    return FALSE;
-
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1211,10 +1211,6 @@ get_kernel_from_tree_usrlib_modules (OstreeSysroot *sysroot, int deployment_dfd,
     }
   glnx_close_fd (&fd);
 
-  /* look for a aboot.cfg file. */
-  if (!ot_openat_ignore_enoent (ret_layout->boot_dfd, "aboot.cfg", &fd, error))
```
cgwalters, line 1215: I think this code was worse than dead - it always leaked a file descriptor if we found devicetree.

See line 1225 below.

cgwalters (approved):

src/libostree/ostree-sysroot-deploy.c
cgwalters: Oh sorry you said that

alexlarsson: yes, but there are no such files in practice, so you will always get ENOENT here.

PR #3504, opened 2025-08-18
deploy: call syncfs() for /ostree instead of /

### LINKED ISSUE #2283: OSTree boot links can get out of sync

champtar:
In full_system_sync we were calling syncfs(/) expecting all the recent modification in /ostree to be synced to disk. With / now being composefs, syncfs(/) is a noop, so call syncfs(/ostree) as that is what we really want.

This should hopefully fix #2283
This was only compile tested

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1634,21 +1634,22 @@ typedef struct
   guint64 boot_syncfs_msec;
 } SyncStats;
 
-/* First, sync the root directory as well as /var and /boot which may
- * be separate mount points.  Then *in addition*, do a global
- * `sync()`.
+/* sync /ostree and /boot which may be separate mount points.
  */
 static gboolean
 full_system_sync (OstreeSysroot *self, SyncStats *out_stats, GCancellable *cancellable,
                   GError **error)
 {
   GLNX_AUTO_PREFIX_ERROR ("Full sync", error);
-  ot_journal_print (LOG_INFO, "Starting syncfs() for system root");
+  ot_journal_print (LOG_INFO, "Starting syncfs() for /ostree");
   guint64 start_msec = g_get_monotonic_time () / 1000;
-  if (syncfs (self->sysroot_fd) != 0)
-    return glnx_throw_errno_prefix (error, "syncfs(sysroot)");
+  glnx_autofd int ostree_dfd = -1;
+  if (!glnx_opendirat (self->sysroot_fd, "ostree", TRUE, &ostree_dfd, error))
+    return glnx_throw_errno_prefix (error, "glnx_opendirat(/ostree)");
+  if (syncfs (ostree_dfd) != 0)
+    return glnx_throw_errno_prefix (error, "syncfs(/ostree)");
   guint64 end_msec = g_get_monotonic_time () / 1000;
-  ot_journal_print (LOG_INFO, "Completed syncfs() for system root in %" G_GUINT64_FORMAT " ms",
+  ot_journal_print (LOG_INFO, "Completed syncfs() for /ostree in %" G_GUINT64_FORMAT " ms",
                     end_msec - start_msec);
 
   out_stats->root_syncfs_msec = (end_msec - start_msec);


cgwalters (approved):
Thanks! I think a root problem we have is `sysroot_fd` means really two different things depending on whether the system is booted or not.

This code will do the right thing in both cases - if we're booted, then `/ostree` is a symlink and we'll follow it. If we're not booted, it will just be a directory.

In bootc recently though I changed things so we also have an explicit fd for the physical root:

https://github.com/bootc-dev/bootc/blob/147f16c32ade4721ab33d429ccd6c7b3ab427ad9/crates/lib/src/store/mod.rs#L80

We should probably do the same here; most uses of `sysroot_fd` should *really* be `physical_root_fd`.

cgwalters:
I did https://github.com/ostreedev/ostree/pull/3509 related to this

cgwalters:
> @cgwalters @jmarrero should we backport that to 2025-2-branch ?
> And also do a release so Fedora and derivative can pick this up ?

Yes to both, but let's aim for tomorrow - I want to do a bit more testing/analysis myself here before we start the whole train of releases and backporting (the above PR was a result of that).

PR #3509, opened 2025-08-21
Deduplicate repo+sysroot syncfs logic

cgwalters:
This is a followup to https://github.com/ostreedev/ostree/pull/3504/commits/6e5a27a29d33d50a2a4380c406405435d919b6b4
which I believe is correct as is. However, we already have a file
descriptor open for the ostree repo, which *must* be on
the same filesystem as `/sysroot/ostree` (the deployment
code forces hardlinking today).

It's hence cleaner to reuse that extant fd instead of opening
a new one - we know we did writes to that fd.

But going farther here, there already is logic to use syncfs
for the repo when downloading objects (in a common case
we actually syncfs twice).

Since these are really the same operation, unify them:

- Add journaling to the repo one syncfs case
- Change the sysroot case to just call it
- Since we log consistently to the journal for all syncfs/fsfreeze
  operations now, drop the SyncStats bits which was a way
  to add info about that to a later journal message

Additionally, let's add an extra check when we're
opening the repo that it's on the same device just on general
principle.

---

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -2227,14 +2227,8 @@ ostree_repo_commit_transaction (OstreeRepo *self, OstreeRepoTransactionStats *ou
   if ((self->test_error_flags & OSTREE_REPO_TEST_ERROR_PRE_COMMIT) > 0)
     return glnx_throw (error, "OSTREE_REPO_TEST_ERROR_PRE_COMMIT specified");
 
-  /* FIXME: Added OSTREE_SUPPRESS_SYNCFS since valgrind in el7 doesn't know
-   * about `syncfs`...we should delete this later.
-   */
-  if (!self->disable_fsync && g_getenv ("OSTREE_SUPPRESS_SYNCFS") == NULL)
-    {
-      if (syncfs (self->tmp_dir_fd) < 0)
-        return glnx_throw_errno_prefix (error, "syncfs(repo/tmp)");
-    }
+  if (!_ostree_repo_syncfs (self, error))
+    return FALSE;
 
   if (!rename_pending_loose_objects (self, cancellable, error))
     return FALSE;
--- a/src/libostree/ostree-repo.c
+++ b/src/libostree/ostree-repo.c
@@ -1521,6 +1521,32 @@ _ostree_repo_update_mtime (OstreeRepo *self, GError **error)
   return TRUE;
 }
 
+gboolean
+_ostree_repo_syncfs (OstreeRepo *self, GError **error)
+{
+
+  if (self->disable_fsync)
+    return TRUE;
+  /* FIXME: Added OSTREE_SUPPRESS_SYNCFS since valgrind in el7 doesn't know
+   * about `syncfs`...we should delete this later.
+   */
+  if (g_getenv ("OSTREE_SUPPRESS_SYNCFS") != NULL)
+    return TRUE;
+
+  gboolean is_system = ostree_repo_is_system (self);
+  if (is_system)
+    ot_journal_print (LOG_INFO, "Starting syncfs for system repo");
+  guint64 start_msec = g_get_monotonic_time () / 1000;
+  int repo_dfd = ostree_repo_get_dfd (self);
+  if (syncfs (repo_dfd) != 0)
+    return glnx_throw_errno_prefix (error, "syncfs(repo)");
+  guint64 end_msec = g_get_monotonic_time () / 1000;
+  if (is_system)
+    ot_journal_print (LOG_INFO, "Completed syncfs() for system repo in %" G_GUINT64_FORMAT " ms",
+                      end_msec - start_msec);
+  return TRUE;
+}
+
 /**
  * ostree_repo_get_config:
  * @self:
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -1628,44 +1628,31 @@ fsfreeze_thaw_cycle (OstreeSysroot *self, int rootfs_dfd, GCancellable *cancella
   return TRUE;
 }
 
-typedef struct
-{
-  guint64 root_syncfs_msec;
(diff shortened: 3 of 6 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1641,15 +1641,15 @@ full_system_sync (OstreeSysroot *self, SyncStats *out_stats, GCancellable *cance
                   GError **error)
 {
   GLNX_AUTO_PREFIX_ERROR ("Full sync", error);
-  ot_journal_print (LOG_INFO, "Starting syncfs() for /ostree");
+  ot_journal_print (LOG_INFO, "Starting syncfs() for physical root");
   guint64 start_msec = g_get_monotonic_time () / 1000;
-  glnx_autofd int ostree_dfd = -1;
-  if (!glnx_opendirat (self->sysroot_fd, "ostree", TRUE, &ostree_dfd, error))
```
champtar, line 1647: We use `ostree_dfd` in `create_new_bootlinks` and `swap_bootlinks`, that's why I went with it initially

src/libostree/ostree-sysroot.c
```diff
@@ -1076,10 +1076,25 @@ ensure_repo (OstreeSysroot *self, GError **error)
...
+  // Because the ostree model requires hardlinks, ensure up front
+  // here that /ostree is on the same filesystem as /ostree/repo.
+  // See `full_system_sync` which also requires this.
+  {
+    struct stat repo_stat, sysroot_stat;
+    if (!glnx_fstat (self->sysroot_fd, &sysroot_stat, error))
+      return FALSE;
+    if (!glnx_fstat (ostree_repo_get_dfd (self->repo), &repo_stat, error))
+      return FALSE;
+    if (sysroot_stat.st_dev != repo_stat.st_dev)
```
champtar: Here you are comparing / and /ostree/repo, but / might be composefs or overlayfs

champtar: And actually if `sysroot_stat.st_dev == repo_stat.st_dev`, then `syncfs(self->sysroot_fd)` would have worked all along

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1641,15 +1641,16 @@ full_system_sync (OstreeSysroot *self, SyncStats *out_stats, GCancellable *cance
                   GError **error)
 {
   GLNX_AUTO_PREFIX_ERROR ("Full sync", error);
-  ot_journal_print (LOG_INFO, "Starting syncfs() for /ostree");
+  // See also `ostree_repo_commit_transaction`
+  ot_journal_print (LOG_INFO, "Starting syncfs() for physical root");
```
champtar: we are doing `syncfs (repo_dfd)` in both cases, why name it `physical root` here ?
```suggestion
  ot_journal_print (LOG_INFO, "Starting syncfs() for system repo");
```

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -1641,15 +1641,16 @@ full_system_sync (OstreeSysroot *self, SyncStats *out_stats, GCancellable *cance
...
-  if (syncfs (ostree_dfd) != 0)
-    return glnx_throw_errno_prefix (error, "syncfs(/ostree)");
+  // We hard require that the ostree repo is on the same physical
+  // filesystem as the system root.
+  int repo_dfd = ostree_repo_get_dfd (self->repo);
+  if (syncfs (repo_dfd) != 0)
+    return glnx_throw_errno_prefix (error, "syncfs(repo)");
   guint64 end_msec = g_get_monotonic_time () / 1000;
-  ot_journal_print (LOG_INFO, "Completed syncfs() for /ostree in %" G_GUINT64_FORMAT " ms",
+  ot_journal_print (LOG_INFO, "Completed syncfs() for physical root in %" G_GUINT64_FORMAT " ms",
```
champtar:
```suggestion
  ot_journal_print (LOG_INFO, "Completed syncfs() for system repo in %" G_GUINT64_FORMAT " ms",
```

src/libostree/ostree-repo-commit.c
```diff
@@ -2232,8 +2232,20 @@ ostree_repo_commit_transaction (OstreeRepo *self, OstreeRepoTransactionStats *ou
    */
   if (!self->disable_fsync && g_getenv ("OSTREE_SUPPRESS_SYNCFS") == NULL)
     {
-      if (syncfs (self->tmp_dir_fd) < 0)
-        return glnx_throw_errno_prefix (error, "syncfs(repo/tmp)");
+      // See also the very similar `full_system_sync` function;
+      // at some point we may deduplicate these.
+      gboolean is_system = ostree_repo_is_system (self);
+      if (is_system)
+        ot_journal_print (LOG_INFO, "Starting syncfs for system repo");
```
champtar:
```suggestion
        ot_journal_print (LOG_INFO, "Starting syncfs() for system repo");
```

cgwalters:
OK I rewrote this one again, it's now one commit and syncs up the syncfs logic

src/libostree/ostree-repo.c
```diff
@@ -1521,6 +1521,32 @@ _ostree_repo_update_mtime (OstreeRepo *self, GError **error)
   return TRUE;
 }
 
+gboolean
+_ostree_repo_syncfs (OstreeRepo *self, GError **error)
+{
+
+  if (self->disable_fsync)
+    return TRUE;
+  /* FIXME: Added OSTREE_SUPPRESS_SYNCFS since valgrind in el7 doesn't know
```
champtar, line 1530: EL 7 is long dead, should we drop `OSTREE_SUPPRESS_SYNCFS` ?

champtar (approved):

src/libostree/ostree-repo.c
cgwalters: Yeah I just kept it on the principle of not changing too many things at once.

PR #3514, opened 2025-08-27
ostree-finalize-staged.service: RequiresMountsFor=/etc

cgwalters:
I've seen in some cases systemd try to unmount /etc quite early and then fail because it's in use.

It's confusing because I don't see this in all scenarios. But regardless, in the situations where it does occur, this fixes it.

### DIFF
--- a/src/boot/ostree-finalize-staged.service
+++ b/src/boot/ostree-finalize-staged.service
@@ -21,7 +21,7 @@ Documentation=man:ostree(1)
 ConditionPathExists=/run/ostree-booted
 DefaultDependencies=no
 
-RequiresMountsFor=/sysroot /boot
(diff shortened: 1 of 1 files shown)


champtar:
We are `After=local-fs.target`, so really makes no sense to me that we need `RequiresMountsFor`, it would make sense for automount, but not for /etc
Do you have logs / screenshot showing `/etc` unmount starting before `ostree-finalize-staged.service` is finished ?

champtar:
Actually `local-fs.target` has `After=sysroot.mount` but not `After=etc.mount`

champtar (approved):
We do require /etc to copy the changed files anyway, so it documents the requirement, but this is not a full fix IMO, we still need local-fs.target to be after etc.mount

PR #3515, opened 2025-08-28
fix: double free in usage of _ostree_filter_selinux_xattr

### LINKED ISSUE #3303: Usroverlay causes (non-fatal) reference counting fail
Probably the cousin of issue https://github.com/secureblue/secureblue/issues/369

Running `sudo rpm-ostree usroverlay --hotfix` causes an assertion error to be emitted for invalid reference counting. But the command completes successfully.

I think there is a chance the rpm-ostree version here is patched to fix the deployment bug that was found recently, but is not changed other than that.

```
bazzite@bazzite:~$ sudo rpm-ostree usroverlay --hotfix

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.069: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.179: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.416: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.418: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.418: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.423: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.439: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.440: g_atomic_ref_count_dec: assertion 'old_value > 0' failed

(ostree admin unlock:7956): GLib-CRITICAL **: 23:57:45.444: g_atomic_ref_count_dec: assertion 'old_value > 0' failed
Copying /etc changes: 19 modified, 2 removed, 90 added
bootfs is sufficient for calculated new size: 0 bytes
Transaction complete; bootconfig swap: no; bootversion: boot.0.1, deployment count change: 0
Hotfix mode enabled.  A writable overlayfs is now mounted on /usr
for this booted deployment.  A non-hotfixed clone has been created
as the non-default rollback target.

bazzite@bazzite:~$ ostree --version
libostree:
 Version: '2024.7'
 Git: 684652bdaa25ae16a551e5cfef82c8896cca5725
...
bazzite@bazzite:~$ rpm-ostree --version
rpm-ostree:
 Version: '2024.7'
 Git: 7d3dce6a6eaac546d9078d7d42f4d5df28f5f0fd
...
```

HastD:
In `create_file_copy_from_input_at` and `checkout_tree_at_recurse`, both `xattrs` and `modified_xattrs` are declared with `g_autoptr`, but later on, `xattrs` is simply assigned to be equal to `modified_xattrs`, meaning the automatic cleanup is a double-free.

This is fixed by instead declaring `modified_xattrs` as `GVariant *`.

I believe this is the cause of issue #3303, and this should fix #3303. (I can consistently reproduce the issue by attempting to deploy a rechunked image with bootc, and with this patch, the issue no longer occurs and the deployment succeeds.)


antheas:
Let's try to push this through soon

champtar (approved):

cgwalters:
> In create_file_copy_from_input_at and checkout_tree_at_recurse, both xattrs and modified_xattrs are declared with g_autoptr, but later on, xattrs is simply assigned to be equal to modified_xattrs, meaning the automatic cleanup is a double-free.

Not...quite. I believe there's a bug here, but my initial take is this patch is changing a double free to a memory leak. This merits deeper investigation.

Let's look at this code pattern:

```
static gboolean
create_file_copy_from_input_at (..., GVariant *xattrs, ...)
{
  g_autoptr (GVariant) modified_xattrs = NULL;

  if (condition)
    {
      modified_xattrs = _ostree_filter_selinux_xattr (xattrs);
      xattrs = modified_xattrs;
    }
```

A really important thing to understand is that in the C programming language, the assignment of `xattrs` *only happens inside that function*. The *calling function* does not see the change.
Just try running this for example:

```
#include <stdio.h>

void foo(char *p) {
  p = "foo";
}

int main() {
    const char *p = "bar";
    foo(p);
    printf("%s\n", p);
    return 0;
}
```

It will print `"bar"`.

cgwalters:
This pattern of "constant pointer and owned pointer" is used in many other places in the code. It is safe if used correctly - and again I could totally believe there's a bug here. 

(Insert obligatory "this wouldn't happen in Rust" thing and I am so looking forward to migrating most logic here to bootc/composefs which are all in Rust)

It's going to take me a bit but let me investigate before we merge please (cc @champtar ).

(Or of course anyone else can feel free to do so, but this just needs a careful analysis)

src/libostree/ostree-repo-checkout.c
```diff
@@ -198,7 +198,6 @@ create_file_copy_from_input_at (OstreeRepo *repo, OstreeRepoCheckoutAtOptions *o
                                 GError **error)
 {
   const gboolean sepolicy_enabled = options->sepolicy && !repo->disable_xattrs;
-  g_autoptr (GVariant) modified_xattrs = NULL;
```
cgwalters: OK I'm sorry I missed that there were two different chunks of this patch. I was focusing on this first one in all my comments so far. I still believe this one is introducing a memory leak.

src/libostree/ostree-repo-checkout.c
```diff
@@ -1055,7 +1053,7 @@ checkout_tree_at_recurse (OstreeRepo *self, OstreeRepoCheckoutAtOptions *options
     if (sepolicy_enabled && _ostree_sepolicy_host_enabled (options->sepolicy))
       {
         /* We'll set the xattr via setfscreatecon(), so don't do it via generic xattrs below. */
-        modified_xattrs = _ostree_filter_selinux_xattr (xattrs);
+        GVariant *modified_xattrs = _ostree_filter_selinux_xattr (xattrs);
         xattrs = modified_xattrs;
```
cgwalters: *This* patch is superficially similar to the above but very different in that in the *original code* we very clearly have two `g_autoptr` pointing to the same thing which is obviously wrong.

And so I think this is the key fix.

However - we're now leaking memory for the old value of `xattrs`. What we can do is instead:

```
g_autoptr(GVariant) old_xattrs = g_steal_pointer (&xattrs);
xattrs = _ostree_filter_selinux_xattr (old_xattrs);
```
or so.

(And yes, this is all incredibly subtle and easy to get wrong and C sucks...)

cgwalters (changes requested):

cgwalters:
TL;DR can you retest with reverting just the first hunk?

src/libostree/ostree-repo-checkout.c
HastD: Oops, wrote my below comment before I saw this. Yes, that pattern looks good—I'll retest with that.

src/libostree/ostree-repo-checkout.c
HastD: How is this one different from the second chunk? I agree my patch to both introduces a memory leak, but didn't this one also have the same double-free issue?

cgwalters:
I think the sole change we need is

```
diff --git i/src/libostree/ostree-repo-checkout.c w/src/libostree/ostree-repo-checkout.c
index ccddb488..1feb9be6 100644
--- i/src/libostree/ostree-repo-checkout.c
+++ w/src/libostree/ostree-repo-checkout.c
@@ -997,7 +997,6 @@ checkout_tree_at_recurse (OstreeRepo *self, OstreeRepoCheckoutAtOptions *options
   g_autoptr (GVariant) dirtree = NULL;
   g_autoptr (GVariant) dirmeta = NULL;
   g_autoptr (GVariant) xattrs = NULL;
-  g_autoptr (GVariant) modified_xattrs = NULL;
 
   if (!ostree_repo_load_variant (self, OSTREE_OBJECT_TYPE_DIR_TREE, dirtree_checksum, &dirtree,
                                  error))
@@ -1055,8 +1054,8 @@ checkout_tree_at_recurse (OstreeRepo *self, OstreeRepoCheckoutAtOptions *options
     if (sepolicy_enabled && _ostree_sepolicy_host_enabled (options->sepolicy))
       {
         /* We'll set the xattr via setfscreatecon(), so don't do it via generic xattrs below. */
-        modified_xattrs = _ostree_filter_selinux_xattr (xattrs);
-        xattrs = modified_xattrs;
+        g_autoptr(GVariant) old_xattrs = g_steal_pointer (&xattrs);
+        xattrs = _ostree_filter_selinux_xattr (old_xattrs);
 
         if (!_ostree_sepolicy_preparefscreatecon (&fscreatecon, options->sepolicy,
                                                   state->selabel_path_buf->str, mode, error))
```

src/libostree/ostree-repo-checkout.c
cgwalters: The difference is that we're only assigning to the *callee copy* of `xattrs`. I tried to describe this in https://github.com/ostreedev/ostree/pull/3515#issuecomment-3237742427

There's no "`g_autoptr()` effect" *inside* the function as it relates to `xattrs`.

HastD: Oh, right, I see what you're saying now.

cgwalters: This is all again extremely subtle and honestly I only learned and internalized all of this after being burned by messing things up over and over and over and over for many years and spending literally hours in gdb and valgrind (per week) and yeah it's why I just really want to end my career writing Rust.

HastD:
I just retested the revised version (which I've now pushed to this PR) in the same VM, and it does indeed fix the bug. Thanks @cgwalters for catching that memory leak in the earlier version.

cgwalters (approved):
Thanks for finally finding this bug!!

Issue #3513, opened 2025-08-27
Fix unmount of /etc and /sysroot


cgwalters:
In some scenarios I see `/sysroot` and sometimes `/etc` fail to unmount. In order to ensure we aren't causing or masking other problems, let's fix this in prep for fixing https://github.com/ostreedev/ostree/issues/3503

cgwalters:
Moving some tech discussion from https://github.com/ostreedev/ostree/pull/3514

>  but this is not a full fix IMO, we still need local-fs.target to be after etc.mount

Per
https://github.com/systemd/systemd/blob/1be088c2995f6f22767a4a58e7b51c3642930805/src/shared/fstab-util.c#L76
this is seemingly intentional.

cgwalters:
> As ostree-finalize-staged.service is After=systemd-journal-flush.service which is After=systemd-journald.socket, we might not have changed when we try to unmount with https://github.com/ostreedev/ostree/pull/3514

I'm not following what you think the problem is there. See the commit message in https://github.com/ostreedev/ostree/pull/1926 which clarifies that this ensures that our `ExecStop` runs *before* `systemd-journal-flush.service`. I don't see how that could cause issues with either `/etc` or `/sysroot`.

> To reproduce the /sysroot umount failure:

I tried that, it doesn't reproduce it for me in a Fedora Silverblue 42 (non-LUKS) setup.

champtar:
Never mind, mixed up my Before and After

> > To reproduce the /sysroot umount failure:
> 
> I tried that, it doesn't reproduce it for me in a Fedora Silverblue 42 (non-LUKS) setup.

Weird, I reproduce it 100% of the time

champtar:
I think that if we want to cleanly unmount everything, we must also umount /, so we need to fix/use dracut-initramfs-restore (https://github.com/dracut-ng/dracut-ng/issues/514)

Here some logs running `sleep.service` and various patches

- dependencies fix for `/etc`: `/etc` and `/sysroot` fails to unmount, then `systemd-shutdown` unmount `/etc` ok, fails to unmount `/sysroot`, but it is able to remount read-only
```
mkdir -p /run/systemd/system/etc.mount.d
cat > /run/systemd/system/etc.mount.d/deps.conf <<'EOS'
[Unit]
After=sysroot.mount
Before=local-fs.target umount.target
Conflicts=umount.target
EOS
systemctl daemon-reload
```

<details>

```
[...]

PR #3516, opened 2025-08-28
Add ostree-shutdown.service: hide /sysroot and make /etc read-only

cgwalters:
We have a lot of bind mounts; these are usually set up in the initramfs.
So far during shutdown we've let systemd just try to sort things out
via auto-generated mount units i.e. `sysroot.mount` and `etc.mount`
and so on.

systemd has some special casing for `-.mount` (i.e. `/`) and `etc.mount`
https://github.com/systemd/systemd/blob/e91bfad241799b449df73efc30d833b9c5937001/src/shared/fstab-util.c#L72

However it doesn't special case `/sysroot` - which is currently
an ostree-specific invention (when used in the real root).
We cannot actually unmount `/sysroot` while it's in use, and it
is because `/etc` is a bind mount into it. And we can't tear
down `/etc` because it's just expected that e.g. pid 1 and other
things hold open references to it - until things finally
transition into systemd-shutdown.

What we can do though is explicitly detach it during the shutdown
phase; this ensures that systemd won't try to clean it up then,
suppressing errors about its inability to do so.

While we're here, let's also remount `/etc` read-only; while
systemd itself will try to do so during systemd-shutdown.
Per comments if this service fails, it's a bug in something
else to be fixed.

Closes: https://github.com/ostreedev/ostree/issues/3513

---

### DIFF
--- a/src/boot/ostree-shutdown.service
+++ b/src/boot/ostree-shutdown.service
@@ -1,0 +1,22 @@
+# SPDX-License-Identifier: LGPL-2.0+
+
+[Unit]
+Description=OSTree Shutdown
+Documentation=man:ostree(1)
+DefaultDependencies=no
+# Only enabled via generator, but for good measure
+ConditionKernelCommandLine=ostree
+# Run after core mounts
+RequiresMountsFor=/etc /sysroot
+# However, we want to only shut down after `/var` has been umounted.
+# Since this runs via ExecStop, this Before= is actually After= at shutdown
+Before=var.mount
+Conflicts=umount.target
+Before=umount.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStop=/usr/lib/ostree/ostree-remount --shutdown
+
+# No [Install] section - we're only enabled via generator
--- a/src/switchroot/ostree-remount.c
+++ b/src/switchroot/ostree-remount.c
@@ -42,6 +42,12 @@
 #include "ostree-mount-util.h"
 #include "otcore.h"
 
+static gboolean opt_shutdown;
+
+static GOptionEntry options[] = { { "shutdown", 'S', 0, G_OPTION_ARG_NONE, &opt_shutdown,
+                                    "Perform shutdown unmounting", NULL },
+                                  { NULL } };
+
 static void
 do_remount (const char *target, bool writable)
 {
@@ -133,10 +139,61 @@ relabel_dir_for_upper (const char *upper_path, const char *real_path, gboolean i
 #endif
 }
 
+// ostree-prepare-root sets things up so that /sysroot points to the "physical" (real) root in the
+// initramfs, and then with composefs `/` is an overlay+EROFS that holds references to content in
+// that physical filesystem.
+//
+// In a typical mutable system where the OS is in a mutable `/` (or `/usr), systemd explicitly
+// skips unmounting both `/` and `/usr`. It will remount them read-only though - and that's
+// the semantic we want to match here.
+static void
+do_shutdown (void)
+{
+  const char *sysroot = "/sysroot";
+  if (mount (sysroot, sysroot, NULL, MS_REMOUNT | MS_SILENT | MS_RDONLY, NULL) < 0)
+    {
+      // Hopefully at this point nothing has any write references, but if they
+      // do we still want to continue.
+      perror ("Remounting /sysroot read-only");
+    }
+  // And fully detach it from the mountns because otherwise systemd thinks
+  // it can be unmounted, but it can't - it's required by `/` (and in a
+  // composefs setup `/etc`) and possibly `/var`. Again, we only really
+  // care that it got mounted read-only and hence outstanding data flushed.
+  // A better fix in the future would be to teach systemd to honor `-.mount`
+  // having a `Requires=sysroot.mount` meaning we can't unmount the latter.
+  if (umount2 (sysroot, MNT_DETACH) < 0)
+    err (EXIT_FAILURE, "umount(/sysroot)");
+
+  // And finally: /etc
+  // NOTE! This one is intentionally last in that we want to try to make
+  // this read-only, but if it fails, systemd-shutdown will have another
+  // attempt after a process killing spree. If anything happens to be
+  // holding a writable fd at this point, conceptually it would have
+  // created race conditions vs ostree-finalize-staged.service, and so
+  // having this service fail will be a signal that those things need
+  // to be fixed.
+  do_remount ("/etc", false);
+  // Don't add anything else after this.
+}
+
 int
 main (int argc, char *argv[])
 {
   g_autoptr (GError) error = NULL;
+  g_autoptr (GOptionContext) context = g_option_context_new ("");
+  g_option_context_add_main_entries (context, options, NULL);
+  if (!g_option_context_parse (context, &argc, &argv, &error))
+    errx (EXIT_FAILURE, "Error parsing options: %s", error->message);
+
+  // Handle the shutdown option
+  if (opt_shutdown)
+    {
+      do_shutdown ();
+      return 0;
+    }
+  // Otherwise fall through to the default startup
+
   g_autoptr (GVariant) ostree_run_metadata_v = NULL;
   {
     glnx_autofd int fd = open (OTCORE_RUN_BOOTED, O_RDONLY | O_CLOEXEC);
--- a/Makefile-boot.am
+++ b/Makefile-boot.am
@@ -38,6 +38,7 @@ endif
 if BUILDOPT_SYSTEMD
 systemdsystemunit_DATA = src/boot/ostree-prepare-root.service \
 	src/boot/ostree-remount.service \
+	src/boot/ostree-shutdown.service \
 	src/boot/ostree-boot-complete.service \
 	src/boot/ostree-finalize-staged.service \
 	src/boot/ostree-finalize-staged-hold.service \
@@ -69,6 +70,7 @@ EXTRA_DIST += src/boot/dracut/module-setup.sh \
 	src/boot/ostree-boot-complete.service \
 	src/boot/ostree-prepare-root.service \
 	src/boot/ostree-remount.service \
+	src/boot/ostree-shutdown.service \
 	src/boot/ostree-finalize-staged.service \
 	src/boot/ostree-finalize-staged-hold.service \
 	src/boot/ostree-state-overlay@.service \
--- a/src/libostree/ostree-impl-system-generator.c
+++ b/src/libostree/ostree-impl-system-generator.c
@@ -108,6 +108,10 @@ require_internal_units (const char *normal_dir, const char *early_dir, const cha
                  "local-fs.target.requires/ostree-remount.service")
       < 0)
     return glnx_throw_errno_prefix (error, "symlinkat");
+  if (symlinkat (SYSTEM_DATA_UNIT_PATH "/ostree-shutdown.service", normal_dir_dfd,
+                 "local-fs.target.wants/ostree-shutdown.service")
+      < 0)
+    return glnx_throw_errno_prefix (error, "symlinkat");
 
   if (!glnx_shutil_mkdir_p_at (normal_dir_dfd, "multi-user.target.wants", 0755, cancellable, error))
     return FALSE;


cgwalters:
An option I looked at here too that I think would work is overriding the auto-generated `sysroot.mount` with one that basically treats it the same as `etc.mount` - i.e. suppressing all default dependencies (including `Before=local-fs.target`) which is I think the main reason systemd is trying to unmount it during regular unit shutdown.

But in the end, this service already exists, already has code to remount read-only, already runs early (and hence late) enough to do cleanups, whereas touching the generator felt more invasive.

champtar:
Do we still have a writable mount of the rootfs open beneath the composefs even after the remount ro /sysroot ?
On the kernel cmdline we have `rw`, so I'm wondering if remounting `/sysroot` `ro` and/or detaching really flushes anything.
```
72 78 0:35 /root /sysroot ro,relatime shared:3 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 rw,seclabel,discard=async,space_cache=v2,subvolid=258,subvol=/root
78 2 0:39 / / ro,relatime shared:1 - overlay composefs ro,seclabel,lowerdir+=/run/ostree/.private/cfsroot-lower,datadir+=/sysroot/ostree/repo/objects,redirect_dir=on,metacopy=on
...
```

cgwalters:
> Do we still have a writable mount of the rootfs open beneath the composefs even after the remount ro /sysroot ?
On the kernel cmdline we have rw, so I'm wondering if remounting /sysroot ro and/or detaching really flushes anything.

Yes this is a good question to ask. There's multiple points here. First...note that composefs is a *read only* case - there's no open writable fds, so it's not going to keep the mount busy itself.

I think the real angle we want to look at this from is pretty simple: with e.g. ext4, it will flush the journal and mark the superblock as being "cleanly unmounted" when being truly remounted read-only, so we can check if that's the case.

champtar:
Here a little experiment

start @cgwalters bpftrace script https://github.com/ostreedev/ostree/pull/3504#issuecomment-3213810592

```
truncate -s10G rootfs
mkfs.xfs rootfs
mkdir mnt
mount rootfs mnt
# this is how we remount read only /sysroot
mount --bind -o remount,ro mnt
grep mnt /proc/self/mountinfo
# open another shell, cd into mnt
umount mnt
umount -l mnt
# no call to syncfs until the shell exit the dir
```

cgwalters:
> I think the real angle we want to look at this from is pretty simple: with e.g. ext4, it will flush the journal and mark the superblock as being "cleanly unmounted" when being truly remounted read-only, so we can check if that's the case.

Heh well...that was a whole rabbit hole. It turns out that ext4 in default journal mode, when inspected via `tune2fs -l` *always* shows the filesystem as clean.

Really...the important thing here is the journal. From the filesystem perspective there's no on-disk difference between "clean unmount" and "empty journal". If the composefs overlay mount still holds a reference to a rw mount, that doesn't matter.

Or to say it yet another way...in the end we're flushing the filesystem; remounting read-only (and here, detaching the `/sysroot` mount which was already read-only) is just part of ensuring processes don't keep writing to it.

(There's a whole side topic here of whether it'd make sense to change systemd-shutdown to actually *freeze* like we do for `/boot` because of xfs and journal...if we did that it'd bypass this whole "did we remount every mountpoint"; instead we just think if we froze each filesystem instance, regardless of the mount point)

---

Anyways so...given the question then becomes "do we have outstanding writes in the journal before reboot/poweroff" - as far as my testing goes the answer is "no" - which is as you'd expect. Things would have to pretty badly wrong for us to have a process which is actively writing past where systemd-shutdown's *two* different invocations of `sync_with_progress()` actually fail.

---

> start @cgwalters bpftrace script https://github.com/ostreedev/ostree/pull/3504#issuecomment-3213810592

Yes, but your test scenario isn't accounting for the above - that systemd-shutdown explicitly runs `sync()` which affects *all* mount points. So it's again really not so much about "did we successfully remount read-only or unmount".

Bottom line then, I plan to merge this patch tomorrow unless there's more followup.

cgwalters:
Tangential: It took some digging but to find out "was an xfs filesystem unmounted cleanly" is apparently best done via `xfs_logprint /dev/sda4` and verifying that the last log operation is e.g. `Oper (0): tid: 704c3ae5  len: 0  clientid: LOG  flags: UNMOUNT `.

champtar:
Actually my small test was not doing the `mount -o remount,ro mnt` ...,
it does remount the underlying mount and sync_fs

#### Test units:
```
cat > /run/sync.trace <<'EOF'
#!/usr/bin/bpftrace
#include <linux/fs.h>

tracepoint:btrfs:btrfs_sync_fs
{
  printf("btrfs sync impl: comm=%-16s pid=%-7d\n", comm, pid);
}
EOF
chmod +x /run/sync.trace

setenforce 0

mkdir -p /run/systemd/system/
cat > /run/systemd/system/trace.service <<'EOS'
[Unit]
DefaultDependencies=no
Before=final.target
Conflicts=final.target

[Service]
Type=simple
StandardOutput=journal+console
ExecStart=/run/sync.trace
KillSignal=SIGCONT
TimeoutStopSec=1m
EOS
systemctl daemon-reload
systemctl start trace.service

mkdir -p /run/systemd/system/ostree-remount.service.d
cat > /run/systemd/system/ostree-remount.service.d/test.conf <<'EOS'
[Unit]
After=-.mount etc.mount var.mount sysroot.mount

[Service]
StandardOutput=journal+console
ExecStop=/usr/bin/grep -E '/sysroot|/etc|/var' /proc/self/mountinfo
ExecStop=/usr/bin/mount -v -o remount,ro /sysroot
ExecStop=/usr/bin/grep -E '/sysroot|/etc|/var' /proc/self/mountinfo
ExecStop=/usr/bin/sleep 1
ExecStop=/usr/bin/umount -v -l /sysroot
ExecStop=/usr/bin/grep -E '/sysroot|/etc|/var' /proc/self/mountinfo
ExecStop=/usr/bin/sleep 1
ExecStop=/usr/bin/mount -v -o remount,ro /etc
ExecStop=/usr/bin/grep -E '/sysroot|/etc|/var' /proc/self/mountinfo
ExecStop=/usr/bin/sleep 1
EOS
systemctl daemon-reload
```

#### Logs

```
[  OK  ] Removed slice system-systemd\x2dfs…slice - Slice /system/systemd-fsck.
[ 1130.367926] grep[2880]: 73 79 0:35 /root /sysroot ro,relatime shared:3 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 rw,seclabel,discard=async,space_cache=v2,subvolid=258,subvol=/root
[ 1130.368018] grep[2880]: 79 2 0:39 / / ro,relatime shared:1 - overlay composefs ro,seclabel,lowerdir+=/run/ostree/.private/cfsroot-lower,datadir+=/sysroot/ostree/repo/objects,redirect_dir=on,metacopy=on
[ 1130.368051] grep[2880]: 76 79 0:35 /root/ostree/deploy/fedora/deploy/fafd1f1fff71690b088440d2dc42ad4f3b89dc05c5f8de6006dd82c24ede4a13.0/etc /etc rw,relatime shared:2 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 rw,seclabel,discard=async,space_cache=v2,subvolid=258,subvol=/root
[ 1130.368078] grep[2880]: 52 79 0:35 /var /var rw,relatime shared:116 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 rw,seclabel,discard=async,space_cache=v2,subvolid=256,subvol=/var
[ 1130.382115] sync.trace[2569]: btrfs sync impl: comm=mount            pid=2882
[ 1130.382266] sync.trace[2569]: btrfs sync impl: comm=mount            pid=2882
[ 1130.600572] mount[2882]: mount: /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 mounted on /sysroot.
[ 1130.622906] grep[2883]: 73 79 0:35 /root /sysroot ro,relatime shared:3 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 ro,seclabel,discard=async,space_cache=v2,subvolid=258,subvol=/root
[ 1130.622980] grep[2883]: 79 2 0:39 / / ro,relatime shared:1 - overlay composefs ro,seclabel,lowerdir+=/run/ostree/.private/cfsroot-lower,datadir+=/sysroot/ostree/repo/objects,redirect_dir=on,metacopy=on
[ 1130.623331] grep[2883]: 76 79 0:35 /root/ostree/deploy/fedora/deploy/fafd1f1fff71690b088440d2dc42ad4f3b89dc05c5f8de6006dd82c24ede4a13.0/etc /etc rw,relatime shared:2 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 ro,seclabel,discard=async,space_cache=v2,subvolid=258,subvol=/root
[ 1130.623487] grep[2883]: 52 79 0:35 /var /var rw,relatime shared:116 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 ro,seclabel,discard=async,space_cache=v2,subvolid=256,subvol=/var
[ 1131.656932] umount[2888]: umount: /sysroot (/dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1) unmounted
[  OK  ] Unmounted sysroot.mount - /sysroot.
[ 1131.662635] grep[2889]: 79 2 0:39 / / ro,relatime shared:1 - overlay composefs ro,seclabel,lowerdir+=/run/ostree/.private/cfsroot-lower,datadir+=/sysroot/ostree/repo/objects,redirect_dir=on,metacopy=on
[ 1131.662693] grep[2889]: 76 79 0:35 /root/ostree/deploy/fedora/deploy/fafd1f1fff71690b088440d2dc42ad4f3b89dc05c5f8de6006dd82c24ede4a13.0/etc /etc rw,relatime shared:2 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 ro,seclabel,discard=async,space_cache=v2,subvolid=258,subvol=/root
[ 1131.662721] grep[2889]: 52 79 0:35 /var /var rw,relatime shared:116 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 ro,seclabel,discard=async,space_cache=v2,subvolid=256,subvol=/var
[ 1132.680245] mount[2894]: mount: /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 mounted on /etc.
[ 1132.685649] grep[2895]: 79 2 0:39 / / ro,relatime shared:1 - overlay composefs ro,seclabel,lowerdir+=/run/ostree/.private/cfsroot-lower,datadir+=/sysroot/ostree/repo/objects,redirect_dir=on,metacopy=on
[ 1132.685717] grep[2895]: 76 79 0:35 /root/ostree/deploy/fedora/deploy/fafd1f1fff71690b088440d2dc42ad4f3b89dc05c5f8de6006dd82c24ede4a13.0/etc /etc ro,relatime shared:2 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 ro,seclabel,discard=async,space_cache=v2,subvolid=258,subvol=/root
[ 1132.685759] grep[2895]: 52 79 0:35 /var /var rw,relatime shared:116 - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 ro,seclabel,discard=async,space_cache=v2,subvolid=256,subvol=/var
[  OK  ] Stopped ostree-remount.service - OSTree Remount OS/ Bind Mounts.
         Unmounting etc.mount...
         Unmounting var.mount - /var...
[FAILED] Failed unmounting etc.mount.
[  OK  ] Unmounted var.mount - /var.
[  OK  ] Stopped target blockdev@dev-mapper…d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1.
```

src/switchroot/ostree-remount.c
```diff
@@ -133,10 +139,55 @@ relabel_dir_for_upper (const char *upper_path, const char *real_path, gboolean i
...
+//
+// In a typical mutable system where the OS is in a mutable `/` (or `/usr), systemd explicitly
+// skips unmounting both `/` and `/usr`. It will remount them read-only though - and that's
+// the semantic we want to match here.
+static void
+do_shutdown (void)
+{
+  // We expect this to be read-only by default on most modern systems, but
+  // in case it's not, make it read-only now.
+  do_remount ("/sysroot", false);
```
champtar: https://github.com/ostreedev/ostree/blob/a5a52e01edd565cca368b946d5e5e4a333b3f350/src/switchroot/ostree-remount.c#L61-L63
We want to unconditionally remount read-only, because the bind mount on top might be read only,
but not the original mount point under it.
In my tests using `mount -o remount,ro` works, so we need a `do_remount_force`

champtar:
> We cannot actually unmount /sysroot while it's in use, and it is because /etc is a bind mount into it.

There is also / (composefs) that prevents /sysroot from being unmounted

cgwalters:
> There is also / (composefs) that prevents /sysroot from being unmounted

Yes. But how is that different from a traditional mutable system on a single flat filesystem where systemd is running from `/` that it also doesn't unmount?

I guess you're commenting on my original commit message, which I didn't update for recent findings. I'll reword it to strengthen the argument there that this is about the journal.

src/boot/ostree-remount.service
```diff
@@ -21,9 +21,11 @@ ConditionKernelCommandLine=ostree
 OnFailure=emergency.target
 Conflicts=umount.target
 # Run after core mounts
-After=-.mount var.mount
+After=-.mount etc.mount var.mount sysroot.mount
```
champtar: As remouting read-only `/sysroot` also makes `/var` read-only, I think we should use a separate unit for shutdown, ordered
```
Before=var.mount
After=-.mount etc.mount sysroot.mount 
```
so everything depending explicitly on `var.mount` can write to it

champtar:
> But how is that different from a traditional mutable system on a single flat filesystem where systemd is running from `/` that it also doesn't unmount?

On a system without SELinux `/etc` might unmount just fine (which might be a problem), but `/` and `/sysroot` are really interlocked, you need to move `/sysroot` out of `/` and then unmount `/`, so you really need to `pivot_root`

> I guess you're commenting on my original commit message, which I didn't update for recent findings. I'll reword it to strengthen the argument there that this is about the journal.

yup, I really hate github (and gitlab) for not providing a way to comment on the commit message

cgwalters:
> On a system without SELinux /etc might unmount just fine (which might be a problem), but / and /sysroot are really interlocked, you need to move /sysroot out of / and then unmount /, so you really need to pivot_root

If we wanted to do that, it's supported via https://www.freedesktop.org/software/systemd/man/latest/bootup.html#The%20exitrd

But again though...someone would need to come up with a counter-argument to my argument that what matters is just mounting read-only and flushing outstanding writes (i.e. the journal is clean), not that we actually need to tear down the in-memory VFS structures - because on reboot *you can't tell the difference*.

champtar:
> > On a system without SELinux /etc might unmount just fine (which might be a problem), but / and /sysroot are really interlocked, you need to move /sysroot out of / and then unmount /, so you really need to pivot_root
> 
> If we wanted to do that, it's supported via https://www.freedesktop.org/software/systemd/man/latest/bootup.html#The%20exitrd
> 
> But again though...someone would need to come up with a counter-argument to my argument that what matters is just mounting read-only and flushing outstanding writes (i.e. the journal is clean), not that we actually need to tear down the in-memory VFS structures - because on reboot _you can't tell the difference_.

I don't think we want to bother with the exitrd, remount ro is enough.

I just wanted you when you reword the commit to mention that we can't unmount /sysroot also because it's interlocked with /, with the current wording  one can think `/etc` is the only blocker.

src/boot/ostree-remount.service
cgwalters: Making a new unit ups the complexity level a bit. I'm not opposed, but I think we need to weigh it vs the other approach of just trying to make sure `sysroot.mount` survives all the way into systemd-shutdown phase which I think is even more correct.

champtar: I agree it would be better to have systemd just skip the umount, but haven't found a way yet.
I've tried to play with `x-initrd.mount` in a unit drop-in but it's ignored,
tried to write a full unit with empty requires but the implicit dependencies on the luks volume still causes unmount
Tried `mount -o remount,x-initrd.mount /etc` but it's also ignored

cgwalters: > I agree it would be better to have systemd just skip the umount, but haven't found a way yet.

It works for me to do
```
cat >/run/systemd/generator/sysroot.mount <<EOF
[Unit]
DefaultDependencies=no
[Mount]
What=/sysroot
Where=/sysroot
Type=bind
EOF
```

Then there's no attempt to tear it down during the early shutdown phase.

I think that'd all we need do from ostree-system-generator.

However...I do still see
`[  245.612829] (sd-umount)[2620]: Failed to unmount /run/shutdown/mounts/9dcea02aab3b8be8: Device or resource busy`

which I haven't traced through but I suspect it's systemd-shutdown trying to unmount `/sysroot`, but again it can't.

champtar: Nice, we need the same one for `/etc` and we are good to go

cgwalters: > Nice, we need the same one for /etc and we are good to go

I dunno, we still end up with the sd-umount warning which is less noisy, but still undesirable.

So...in the end I think the only fix there would be a way to tell systemd-shutdown to skip unmounting `/sysroot` entirely, the same way it skips `/` and `/usr`. I don't think there's a way to do that right now.

Given the lack of that, I think my original approach of just hiding it makes sense.

champtar: If we want to fully fix the unmounting errors, we need your mount unit + we need to change how we do the initial mount
1) leave the rootfs mounted rw at /sysroot
2) prepare the composefs at /compose
3) mount --bind /sysroot /sysroot2
4) mount --move /compose /sysroot
5) mount --move /sysroot2 /sysroot/sysroot

That way / (composefs) only depends on hidden partitions, `/sysroot` bind mount can be unmouted, and systemd-shutdown remount read-only before unmounting so we will still have a clean state

I've played a bit booting with rd.break=pre-pivot and it seems to let me do the move

cgwalters:
OK now redone to add a new service

champtar:
> OK now redone to add a new service

See my latest comment (https://github.com/ostreedev/ostree/pull/3516#discussion_r2311014079), I think if we change slightly the mount we can keep /etc and /sysroot mounted till systemd-shutdown

src/boot/ostree-remount.service
cgwalters: Scoping in changing `ostree-prepare-root.c` adds a whole new dimension of risk here.

> If we want to fully fix the unmounting errors, 

I am claiming this PR fully fixes the errors. Do you disagree? Very specifically: There are no emitted error logs.

> That way / (composefs) only depends on hidden partitions, /sysroot bind mount can be unmouted, and systemd-shutdown remount read-only before unmounting so we will still have a clean state

I am not fully following this; what do you mean by "hidden partitions". Do you mean that `/sysroot` would not be mounted in the real root? That'd just break the ability to do in-place OS updates...

champtar: > > I agree it would be better to have systemd just skip the umount, but haven't found a way yet.
> 
> It works for me to do
> 
> ```
> cat > /run/systemd/generator/sysroot.mount <<EOF
> [Unit]
> DefaultDependencies=no
> [Mount]
> What=/sysroot
> Where=/sysroot
> Type=bind
> EOF
> ```
> 
> Then there's no attempt to tear it down during the early shutdown phase.

Do you create the .mount unit in the initramfs or during the normal boot ?
I've tried a simple generator
```
mkdir -p /etc/systemd/system-generators/
cat > /etc/systemd/system-generators/mount.sh <<'EOF'
#!/bin/bash

/bin/cat >/run/systemd/generator/sysroot.mount <<'EOU'
[Unit]
DefaultDependencies=no
[Mount]
What=/sysroot
Where=/sysroot
Type=bind
EOU

/bin/cat >/run/systemd/generator/etc.mount <<'EOU'
[Unit]
DefaultDependencies=no
[Mount]
What=/etc
Where=/etc
Type=bind
EOU

EOF
chmod +x /etc/systemd/system-generators/mount.sh
```

And systemd still tries to unmount `/sysroot`

src/boot/ostree-shutdown.service
```diff
@@ -0,0 +1,25 @@
...
+Conflicts=shutdown.target
+Before=shutdown.target
+
+[Service]
+Type=oneshot
+RemainAfterExit=yes
+ExecStop=/usr/lib/ostree/ostree-remount --shutdown
+StandardInput=null
+StandardOutput=journal
+StandardError=journal+console
```
champtar: As var is gone the journal is also gone
```suggestion
StandardOutput=console
StandardError=console
```

src/boot/ostree-shutdown.service
```diff
@@ -0,0 +1,25 @@
...
+DefaultDependencies=no
+# Only enabled via generator, but for good measure
+ConditionKernelCommandLine=ostree
+# Run after core mounts
+RequiresMountsFor=/etc /sysroot
+# However, we want to only shut down after `/var` has been umounted.
+# Since this runs via ExecStop, this Before= is actually After= at shutdown
+Before=var.mount
+Conflicts=shutdown.target
+Before=shutdown.target
```
champtar:
```suggestion
Conflicts=umount.target
Before=umount.target
```

src/boot/ostree-shutdown.service
cgwalters: > As var is gone the journal is also gone

Not quite; journald switches to logging to `/run`, see `systemd-journal-flush.service`.

cgwalters: Also of course there are cases where journald is set up to do remote logging via `ForwardToSocket=` and/or there isn't a persistent journal at all (i.e. the journal stays in `/run`).

src/boot/ostree-remount.service
champtar: > Scoping in changing `ostree-prepare-root.c` adds a whole new dimension of risk here.

100% agree

> > If we want to fully fix the unmounting errors,
> 
> I am claiming this PR fully fixes the errors. Do you disagree? Very specifically: There are no emitted error logs.

It does fix /sysroot (not /etc) by adding a small behavior change, we remount the rootfs readonly before systemd-shutdown, the open question is "what are we going to break" ?
for soft reboot, do we umount everything and remount the rootfs ? or we bind mounts /sysroot ?

> > That way / (composefs) only depends on hidden partitions, /sysroot bind mount can be unmouted, and systemd-shutdown remount read-only before unmounting so we will still have a clean state
> 
> I am not fully following this; what do you mean by "hidden partitions". Do you mean that `/sysroot` would not be mounted in the real root? That'd just break the ability to do in-place OS updates...

Here /proc/1/mountinfo in the initramfs
```
73 2 0:35 /root /sysroot ro,relatime - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 rw,discard=async,space_cache=v2,subvolid=258,subvol=/root
79 73 0:39 / /sysroot ro,relatime - overlay composefs ro,lowerdir+=/run/ostree/.private/cfsroot-lower,datadir+=/sysroot/ostree/repo/objects,redirect_dir=on,metacopy=on
76 79 0:35 /root/ostree/deploy/fedora/deploy/fafd1f1fff71690b088440d2dc42ad4f3b89dc05c5f8de6006dd82c24ede4a13.0/etc /sysroot/etc rw,relatime - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 rw,discard=async,space_cache=v2,subvolid=258,subvol=/root
80 81 0:35 /root/ostree/deploy/fedora/var /sysroot/sysroot/ostree/deploy/fedora/var rw,relatime - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 rw,discard=async,space_cache=v2,subvolid=258,subvol=/root
81 79 0:35 /root /sysroot/sysroot ro,relatime - btrfs /dev/mapper/luks-38d7d9c8-67d1-42c4-b3d4-58b3e1ee94e1 rw,discard=async,space_cache=v2,subvolid=258,subvol=/root
```

The original rootfs would stay at `/sysroot`, composefs on top of it also at `/sysroot`, and at `/sysroot/sysroot` we would have a bind mount of the rootfs, so no visible changes.
Haven't tested `rpm-ostree usroverlay` or `rpm-ostree install --apply-live` yet

src/boot/ostree-shutdown.service
cgwalters: Makes sense, done!

src/boot/ostree-shutdown.service
cgwalters: Anyways since `StandardInput=null` and `StandardOutput=journal` are actually just the defaults, I dropped them. I also dropped `StandardError=journal+console` for now because this service failing is not actually fatal.

src/boot/ostree-remount.service
cgwalters: > It does fix /sysroot (not /etc) by adding a small behavior change, we remount the rootfs readonly before systemd-shutdown, the open question is "what are we going to break" ? for soft reboot, do we umount everything and remount the rootfs ? or we bind mounts /sysroot ?

Well, for soft reboots the good thing is we do have tests here, and they pass. Remember that we set up `/run/nextroot` *before* shutdown, so we already have all the mounts there that remain unaffected by the shutdown process, including the writable reference in `/run/nextroot/etc` etc.

cgwalters: > The original rootfs would stay at /sysroot, composefs on top of it also at /sysroot, and at /sysroot/sysroot we would have a bind mount of the rootfs, so no visible changes.

I really need a bit more detail about what you're suggesting to change.

Can you clarify: Are you proposing that this blocks this PR or not? If not, can you click "auto merge" please? And then we can take any further investigations to followups.

champtar: See the logs https://github.com/ostreedev/ostree/pull/3516#issuecomment-3235588955, remounting /sysroot readonly remount the underlying rootfs for everyone, /etc and /var included
Do the tests try to do any writes after the boot ?

champtar: FWIW just tested `rpm-ostree install --apply-live tcpdump` after moving the mounts around and it works just fine, ie `sysroot.mount` unmount fine
The commands I run in initrd (`rd.break=pre-pivot`)
```
mkdir /tmp-rootfs /tmp-composefs /tmp-etc /tmp-var
mount --move /sysroot/etc /tmp-etc
mount --move /sysroot/sysroot/ostree/deploy/fedora/var /tmp-var
mount --move /sysroot/sysroot /tmp-rootfs
mount --move /sysroot /tmp-composefs

mount --move /tmp-rootfs /sysroot
mount --bind /sysroot /tmp-rootfs

mount --move /tmp-composefs /sysroot
mount --move /tmp-rootfs /sysroot/sysroot
mount --move /tmp-var /sysroot/sysroot/ostree/deploy/fedora/var
mount --move /tmp-etc /sysroot/etc
```

champtar: > > The original rootfs would stay at /sysroot, composefs on top of it also at /sysroot, and at /sysroot/sysroot we would have a bind mount of the rootfs, so no visible changes.
> 
> I really need a bit more detail about what you're suggesting to change.
> 
> Can you clarify: Are you proposing that this blocks this PR or not? If not, can you click "auto merge" please? And then we can take any further investigations to followups.

Pinging you on CNCF slack, Github is not the best chat :)
if you confirm you can actually write to `/var` after soft reboot I'm ok to merge this

cgwalters: > Do the tests try to do any writes after the boot ?

Yes, pretty sure a lot of things would fail if `/etc` and `/var` were read-only. I also interactively tested a soft reboot with this change just for good measure.

> See the logs https://github.com/ostreedev/ostree/pull/3516#issuecomment-3235588955, remounting /sysroot readonly remount the underlying rootfs for everyone, /etc and /var included

Hmmmm...I think you're conflating "remounting readonly" with "sync" a bit...but actually there's a further very important piece to understand which is that with bind mounts over a filesystem mounted read-only, remounting them writable still works without affecting the source mount. That's why we have *two* `mount` calls here:

https://github.com/ostreedev/ostree/blob/da93600f40b679cb11f136a50826691dc325415d/src/libotcore/otcore-prepare-root.c#L466

So in the soft reboot case, even if the shutdown remounting ro propagated to other mounts, we still force it back writable then.

champtar: Ok remounting rw also remount rw the underlying mount
```
root@pb14250:~# grep mnt /proc/1/mountinfo
841 75 7:1 / /root/mnt ro,relatime shared:1142 - xfs /dev/loop1 rw,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
root@pb14250:~# mount -o remount,ro mnt
root@pb14250:~# grep mnt /proc/1/mountinfo
841 75 7:1 / /root/mnt ro,relatime shared:1142 - xfs /dev/loop1 ro,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota
root@pb14250:~# mount -o remount,rw mnt
root@pb14250:~# grep mnt /proc/1/mountinfo
841 75 7:1 / /root/mnt rw,relatime shared:1142 - xfs /dev/loop1 rw,seclabel,attr2,inode64,logbufs=8,logbsize=32k,noquota

```

champtar (approved):
This fixes sysroot umount and hopefully doesn't break anything we could think of, LGTM

src/boot/ostree-remount.service
cgwalters: > Ok remounting rw also remount rw the underlying mount

Right...the way I understand things here is that the mounts (including ro/rw) are VFS level concept, but are references to an underlying filesystem. When a request comes in to mount writable, that "reaches through" to to the underlying fs (if it can be mounted writable).

src/switchroot/ostree-remount.c
```diff
@@ -133,10 +139,61 @@ relabel_dir_for_upper (const char *upper_path, const char *real_path, gboolean i
...
+// that physical filesystem.
+//
+// In a typical mutable system where the OS is in a mutable `/` (or `/usr), systemd explicitly
+// skips unmounting both `/` and `/usr`. It will remount them read-only though - and that's
+// the semantic we want to match here.
+static void
+do_shutdown (void)
+{
+  const char *sysroot = "/sysroot";
+  if (mount (sysroot, sysroot, NULL, MS_REMOUNT | MS_SILENT | MS_RDONLY, NULL) < 0)
```
jlebon, line 153: Since we're read-only by default already, I assume this is to handle old setups without `sysroot.readonly`? Seems worth a comment.

jlebon: OK, this is answered in the comment lower down I guess mentioning flushing.

champtar: As there is no `MS_BIND`, it actually remounts the underlying devices read-only, forcing a flush

PR #3518, opened 2025-09-03
Rework mounts to fix sysroot.mount umount

champtar:
- ostree-prepare-root: avoid mount cycle
    
Moving the physical root at /sysroot, we end up
with a mount cycle between / and /sysroot, forcing us to use
MS_DETACH during shutdown (d0c454c23637dceda6d7395dd2141b564e3efa47).

We can replace ostree-shutdown.service by reworking how we mount
/sysroot, in short use MS_BIND instead of MS_MOVE.

- ostree-soft-reboot: fix sysroot.mount umount

The composefs at /run/nextboot uses /sysroot, so systemd fails to
umount sysroot.mount during soft-reboot.
Create a temporary bind-mount, use it to prepare /run/nextboot
and MNT_DETACH it when we are done.

- Revert d0c454c23637dceda6d7395dd2141b564e3efa47

Notes:
- haven't touched ostree-prepare-root-static as I'm not sure if it's ever used with systemd


src/switchroot/ostree-prepare-root.c
```diff
@@ -325,6 +325,25 @@ main (int argc, char *argv[])
         err (EXIT_FAILURE, "failed to bind mount (class:readonly) /usr");
     }
 
+  /* Prepare /sysroot.
```
cgwalters, line 328: One thing that may help here is for us to use the new mount API. I am pretty sure we can reasonably hard require it now.

With that, we no loner need `sysroot.tmp` and these absolute paths - it becomes easier to reason about the mount setup. Also we can construct the entire mount as a transactional unit (detached fd) that we then attach in place.

cgwalters (commented):
I think this PR should also revert the ostree-shutdown commit right?

I didn't actually review the changes yet.

src/switchroot/ostree-prepare-root.c
champtar: That would require some prep work in libcomposefs no ?
https://github.com/composefs/composefs/blob/c34cd5109f64be4d50446ab80a87b8f13772fd07/libcomposefs/lcfs-mount.h#L47-L50

cgwalters: Yes I think so, both that and the composefs-rs reimplementation (https://github.com/containers/composefs-rs/blob/main/crates/composefs/src/mount.rs) are using the new mount API internally, just not exposing an API to mount at a FD.

champtar:
(still WIP, the soft-reboot part doesn't work)

champtar:
Now ready for review

src/libostree/ostree-soft-reboot.c
```diff
@@ -63,6 +63,22 @@ _ostree_prepare_soft_reboot (GError **error)
...
+   */
+  if (mount (sysroot_path, sysroot_path, NULL, MS_BIND | MS_SILENT, NULL) < 0)
+    err (EXIT_FAILURE, "failed to MS_BIND '%s'", sysroot_path);
+
+  /* Our curent working directory is in the old /sysroot,
+   * ie we are under the bind mount, so run 'cd $PWD'
+   * to move to the new /sysroot
+   */
+  g_autofree char *cwd = g_get_current_dir ();
+  if (chdir (cwd) < 0)
```
cgwalters, line 79: This could just be `chdir (".")` right?

src/libostree/ostree-soft-reboot.c
```diff
@@ -78,6 +94,14 @@ _ostree_prepare_soft_reboot (GError **error)
   if (!otcore_mount_etc (config, &metadata_builder, OTCORE_RUN_NEXTROOT, error))
     return FALSE;
 
+  // detach the temporary /sysroot bind-mount
+  if (umount2 (sysroot_path, MNT_DETACH) < 0)
+    err (EXIT_FAILURE, "failed to MS_DETACH '%s'", sysroot_path);
+
+  // run 'cd $PWD' again to go back to the old /sysroot
+  if (chdir (cwd) < 0)
```
cgwalters, line 102: Ditto

src/switchroot/ostree-prepare-root.c
```diff
@@ -325,6 +325,25 @@ main (int argc, char *argv[])
...
+   * The future / (currenyly at /sysroot.tmp) is an overlayfs or composefs that uses
```
cgwalters: Typo "currently"

src/libostree/ostree-soft-reboot.c
champtar: This is what I was doing at first and I lost a good amount of time on this one.
When you run `cd .`, it's actually equivalent to `cd $PWD`, ie it find's the current working dir and chdir to it.
When in C you run `chdir (".")`, nothing happens, you are still in the old `/sysroot` under the new bind mount, so if you use relative path you use the content of the old `/sysroot`, and if you use absolute path you use the new `/sysroot`

I tried to
```
chdir ("/sysroot")
mount ("/sysroot", "/sysroot", NULL, MS_BIND | MS_SILENT, NULL)
umount2 (".", MNT_DETACH)
```
but it seems to umount the upper mount

MOVE_MOUNT_BENEATH is only available since kernel 6.5 (https://github.com/brauner/man-pages-md/blob/main/move_mount.md)

When we switch to the new API we can `open_tree('/sysroot')` then `fchdir`

src/switchroot/ostree-prepare-root.c
champtar: Fixed locally, waiting a bit before pushing

src/switchroot/ostree-prepare-root.c
```diff
@@ -325,6 +325,25 @@ main (int argc, char *argv[])
...
+   */
+  if (mount (root_mountpoint, TMP_SYSROOT "/sysroot", NULL, MS_BIND | MS_SILENT, NULL) < 0)
+    err (EXIT_FAILURE, "failed to MS_BIND '%s' to 'sysroot'", root_mountpoint);
+
+  if (umount2 (root_mountpoint, MNT_DETACH) < 0)
+    err (EXIT_FAILURE, "failed to MS_DETACH '%s'", root_mountpoint);
+
+  /* Resolve deploy path again so we can use paths relative to the physical root bind-mount */
+  g_autofree char *deploy_path2 = resolve_deploy_path (kernel_cmdline, TMP_SYSROOT "/sysroot");
+  if (chdir (deploy_path2) < 0)
```
cgwalters, line 344: One thing I had to reassure myself on here is that since we're doing a `chdir` in a different place, that all the following code was ready. But basically we were already in `chdir` in the deploy path.

jlebon (commented):
Nice work on this!

PR #3520, opened 2025-09-05
staged-deploy: Verify we didn't fail to unmount /sysroot or /etc

cgwalters:
Followup to https://github.com/ostreedev/ostree/issues/3513

### DIFF
--- a/tests/kolainst/destructive/soft-reboot.sh
+++ b/tests/kolainst/destructive/soft-reboot.sh
@@ -76,6 +76,11 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   # Verify that soft-reboot state files are gone
   test '!' -f /run/ostree/nextroot-booted
   
+  # And verify there were no failures to unmount /etc or /sysroot
+  journalctl -b -1 -u sysroot.mount -u etc.mount >previous-journal.txt
+  assert_not_file_has_content previous-journal.txt Failed
+  rm -f previous-journal.txt
+
   echo "Soft reboot test completed successfully!"
 
   # Now soft reboot again into the rollback which is not staged,


tests/kolainst/destructive/staged-deploy.sh
```diff
@@ -98,6 +98,11 @@ EOF
     syncfs=$(journalctl -b -1 -u ostree-finalize-staged --grep='Completed syncfs.*for system repo' | wc -l)
     assert_streq "$syncfs" 2
 
+    # And verify there were no failures to unmount /etc or /sysroot
+    journalctl -b -1 -u sysroot.mount -u etc.mount >previous-journal.txt
```
jlebon: Minor: looks like it accepts globs, so `-u '*.mount'` also works and catches more.

jlebon (approved):

tests/kolainst/destructive/staged-deploy.sh
```diff
@@ -98,6 +98,11 @@ EOF
     syncfs=$(journalctl -b -1 -u ostree-finalize-staged --grep='Completed syncfs.*for system repo' | wc -l)
     assert_streq "$syncfs" 2
 
+    # And verify there were no failures to unmount /etc or /sysroot
```
champtar: When doing soft-reboot I think we keep the full journal, but on normal reboot this would only catch failures before /var umount.
I think we want to duplicate this test for the soft-reboot case.

Best would be to have the VM serial output, but I don't know what kola can do.

cgwalters: Yeah you're right, moved to soft reboot.

> Best would be to have the VM serial output, but I don't know what kola can do.

We do gather it, but it's currently outside what external tests can gate on with coreos-assembler. I'm working in the background on some things like https://github.com/bootc-dev/podman-bootc/issues/99 which can help form the core of a more powerful container-native testing framework.

tests/kolainst/destructive/soft-reboot.sh
```diff
@@ -76,6 +76,11 @@ case "${AUTOPKGTEST_REBOOT_MARK:-}" in
   # Verify that soft-reboot state files are gone
   test '!' -f /run/ostree/nextroot-booted
   
+  # And verify there were no failures to unmount /etc or /sysroot
+  journalctl -b -1 -u sysroot.mount -u etc.mount >previous-journal.txt
```
champtar, line 80: When you do a soft reboot, for the journal it's still the same boot, ie you must use `-b0` here

champtar: Thinking about it more, the journal is pretty small, we can just remove the `-b`

PR #3523, opened 2025-09-09
Support using composefs signatures also with bootc commits

alexlarsson:
When using bootc, if you convert a signed ostree commit into an OCI image `rpm-ostree compose container-encapsulate` you end up with a new commit that isn't signed. However, the base commit object, and its commitmeta are still in the image and will end up the repo.

The base commit id is available in the container image config as a Label. So, we change ostree-prepare-root to fall back to using this base commit+commitmeta to find the expected composefs digest if the main commit is not signed.

### DIFF
--- a/src/libotcore/otcore-prepare-root.c
+++ b/src/libotcore/otcore-prepare-root.c
@@ -278,19 +278,31 @@ otcore_load_rootfs_config (const char *cmdline, GKeyFile *config, gboolean load_
 }
 
 #ifdef HAVE_COMPOSEFS
-static GVariant *
+static gboolean
 load_variant (const char *root_mountpoint, const char *digest, const char *extension,
-              const GVariantType *type, GError **error)
+              const GVariantType *type, gboolean allow_noent, GVariant **out, GError **error)
 {
   g_autofree char *path = g_strdup_printf ("%s/ostree/repo/objects/%.2s/%s.%s", root_mountpoint,
                                            digest, digest + 2, extension);
 
   char *data = NULL;
   gsize data_size;
-  if (!g_file_get_contents (path, &data, &data_size, error))
-    return NULL;
+  g_autoptr (GError) local_error = NULL;
 
-  return g_variant_ref_sink (g_variant_new_from_data (type, data, data_size, FALSE, g_free, data));
+  if (!g_file_get_contents (path, &data, &data_size, &local_error))
+    {
+      if (allow_noent && g_error_matches (local_error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+        {
+          *out = NULL;
+          return TRUE;
+        }
+
+      g_propagate_error (error, g_steal_pointer (&local_error));
+      return FALSE;
+    }
+
+  *out = g_variant_ref_sink (g_variant_new_from_data (type, data, data_size, FALSE, g_free, data));
+  return TRUE;
 }
 
 // Given a mount point, directly load the .commit object.  At the current time this tool
@@ -299,27 +311,21 @@ static gboolean
 load_commit_for_deploy (const char *root_mountpoint, const char *deploy_path, GVariant **commit_out,
                         GVariant **commitmeta_out, GError **error)
 {
-  g_autoptr (GError) local_error = NULL;
   g_autofree char *digest = g_path_get_basename (deploy_path);
   char *dot = strchr (digest, '.');
   if (dot != NULL)
     *dot = 0;
 
-  g_autoptr (GVariant) commit_v
-      = load_variant (root_mountpoint, digest, "commit", OSTREE_COMMIT_GVARIANT_FORMAT, error);
-  if (commit_v == NULL)
+  g_autoptr (GVariant) commit_v = NULL;
+  g_autoptr (GVariant) commitmeta_v = NULL;
+
+  if (!load_variant (root_mountpoint, digest, "commit", OSTREE_COMMIT_GVARIANT_FORMAT, FALSE,
+                     &commit_v, error))
     return FALSE;
 
-  g_autoptr (GVariant) commitmeta_v = load_variant (root_mountpoint, digest, "commitmeta",
-                                                    G_VARIANT_TYPE ("a{sv}"), &local_error);
-  if (commitmeta_v == NULL)
-    {
-      if (g_error_matches (local_error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
-        glnx_throw (error, "No commitmeta for commit %s", digest);
-      else
-        g_propagate_error (error, g_steal_pointer (&local_error));
-      return FALSE;
-    }
+  if (!load_variant (root_mountpoint, digest, "commitmeta", G_VARIANT_TYPE ("a{sv}"), TRUE,
+                     &commitmeta_v, error))
+    return FALSE;
 
   *commit_out = g_steal_pointer (&commit_v);
   *commitmeta_out = g_steal_pointer (&commitmeta_v);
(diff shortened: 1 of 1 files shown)


src/libotcore/otcore-prepare-root.c
```diff
@@ -550,12 +602,31 @@
...
         return glnx_prefix_error (error, "Error loading signatures from repo");
 
+      if (main_commitmeta != NULL)
+        {
+          commit = main_commit;
+          commitmeta = main_commitmeta;
+        }
+      else if (base_commit != NULL && base_commitmeta != NULL)
+        {
+          g_print ("composefs+ostree: Validating composefs using bootc base commit\n");
```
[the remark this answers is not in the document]
alexlarsson: fixed

src/libotcore/otcore-prepare-root.c
```diff
@@ -314,15 +315,66 @@ load_commit_for_deploy (const char *root_mountpoint, const char *deploy_path, GV
...
+
+  const char *image_config = NULL;
+  g_autoptr (GVariant) metadata = g_variant_get_child_value (commit_v, 0);
+  if (g_variant_lookup (metadata, "ostree.container.image-config", "s", &image_config))
+    {
+      /* The image config is a json, but we don't want to bring in a
+         json parser into the dependencies as we want to minimize the
+         chance of this being exploited to gain trust.*/
+      g_autoptr (GRegex) commit_regex
+          = g_regex_new ("\"ostree.commit\":\"([0-9a-f]{64})\"", G_REGEX_MULTILINE, 0, NULL);
```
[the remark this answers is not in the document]
alexlarsson: fixed

src/libotcore/otcore-prepare-root.c
```diff
@@ -314,15 +315,66 @@
...
+              if (basecommitmeta_v == NULL)
+                {
+                  if (!g_error_matches (local_error, G_FILE_ERROR, G_FILE_ERROR_NOENT))
+                    {
+                      g_propagate_error (error, g_steal_pointer (&local_error));
+                      return FALSE;
+                    }
+                  g_clear_error (&local_error);
+                }
+            }
```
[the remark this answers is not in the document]
alexlarsson: fixed by adding allow_none argument to load_variant.

src/libotcore/otcore-prepare-root.c
```diff
@@ -550,12 +602,31 @@
...
+      if (!load_commit_for_deploy (root_mountpoint, deploy_path, &main_commit, &main_commitmeta,
+                                   &base_commit, &base_commitmeta, error))
         return glnx_prefix_error (error, "Error loading signatures from repo");
 
+      if (main_commitmeta != NULL)
+        {
+          commit = main_commit;
+          commitmeta = main_commitmeta;
+        }
+      else if (base_commit != NULL && base_commitmeta != NULL)
```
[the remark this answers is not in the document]
alexlarsson: fixed

cgwalters:
Not reviewing the code in detail yet...yeah I totally understand this and have thought about it myself...

But...there's obviously a lot of suboptimal things about this (many of which get fixed with the big composefs-native work of course).

In the short term though, I think we could change bootc to store the manifest and config data out of band if the input image is "fully" ostree (i.e. does not have any non-ostree layers).
That would dramatically simplify this logic right?

alexlarsson:
I wish we had signed the composefs digest instead of the complete commit. Then we could just have copied the composefs digest + its signature into the new commit.

src/libotcore/otcore-prepare-root.c
```diff
@@ -278,51 +278,94 @@ otcore_load_rootfs_config (const char *cmdline, GKeyFile *config, gboolean load_
...
 
-  g_autoptr (GVariant) commitmeta_v = load_variant (root_mountpoint, digest, "commitmeta",
-                                                    G_VARIANT_TYPE ("a{sv}"), &local_error);
-  if (commitmeta_v == NULL)
+  if (!load_variant (root_mountpoint, digest, "commitmeta", G_VARIANT_TYPE ("a{sv}"), TRUE,
+                     &commitmeta_v, error))
+    return FALSE;
+
+  /* In case the commit is one created by bootc when importing a container, it will not
+     be signed. However, we can still look at the base commit which may be. */
```
champtar: Haven't tried to follow the logic, but have you made sure this doesn't break layering:
- rpm-ostree local layering (not sure if metadata like `ostree.container.image-config` is copied between commit when doing local layering)
- container layering (`LABEL` are kept when adding layers I think)

cgwalters:
> IIUC, I'm not sure we could do this by default. Wouldn't this change the output of rpm-ostree status --json? This is used by at least CoreOS stuff to read manifests/configs via the base-commit-meta key (but I wouldn't be surprised there's other code out there that does this).

Probably but I think things that are doing that are broken and should have been instead using the `ostree container image metadata` command

src/libotcore/otcore-prepare-root.c
cgwalters: Remember that people doing sealed images don't want either of those things to work.

champtar: Sure, but silently ignoring the extra layers would be super surprising behavior.
You can imagine a vendor delivering a signed commit encapsulated in a container, and a customer layering their favorite security/monitoring agent.
This should error out (during deployment ?), telling the customer they must disable signature checks

cgwalters:
The more I think about this the more I feel this all just needs to be fixed on the bootc side.

src/libotcore/otcore-prepare-root.c
alexlarsson: Well, sealing fundamentally breaks layering. In this case, the boot will read the expected composefs digest from the base ostree commit, and if there were added layers the deployed composefs image will have a different commit, so it will fail to boot with an "invalid digest". But this is sort of the point of sealing it in the first place.

If you truly want to layer something you need to at the end run something that re-signs the completed image, and that could if you wanted re-create a full signed ostree commit to make it boot again.

alexlarsson:
I feel the opposite. This is a very localized hack that makes it work with minimal changes. Changing the format of the commits is a much more major change. Especially given that we want to do something else long-term.

alexlarsson:
One thing I was considering, is could we have bootc embedd the original commit and commitmeta variants as metadata keys in the bootc created commit as "base-commit" and "base-commitmeta"? That would make this workaround much cleaner.

alexlarsson:
> > Sure, but silently ignoring the extra layers would be super surprising behavior.
> 
> Yes. I think a clear next enhancement to bootc (ostree-ext) is something like `ostree.sealed` or something that disallows any non-ostree layers.

Not sure where the above comment came from, but that is not quite true. The local commit will be used to deploy. The base commit is **only** used to find the expected composefs digest, so what will happen is that you deploy something that will then not boot. I mean, this isn't exactly great either, but its not just "silently ignoring" some content.

src/libotcore/otcore-prepare-root.c
champtar: > If you truly want to layer something you need to at the end run something that re-signs the completed image, and that could if you wanted re-create a full signed ostree commit to make it boot again.

Or disable signature checking at runtime, because you just want to run a quick test and don't have the signing key.

cgwalters:
> One thing I was considering, is could we have bootc embedd the original commit and commitmeta variants as metadata keys in the bootc created commit as "base-commit" and "base-commitmeta"? That would make this workaround much cleaner.

See https://github.com/bootc-dev/bootc/pull/1600

cgwalters:
> I feel the opposite. This is a very localized hack that makes it work with minimal changes. Changing the format of the commits is a much more major change. Especially given that we want to do something else long-term.

I took a look at it, it's not *that* major...did some prep work in https://github.com/bootc-dev/bootc/pull/1602

Basically in this flow we'd skip `write_merge_commit_impl` and write the container metadata to detached instead, and change the lookup functions to look for this. 

Something to bear in mind of course here is that in theory in this case, the client side contents *should* be the same as the server side generated contents...but we're definitely not strictly validating that right now. Hmm, let me try out adding an assertion that they're content identical and see...

There's some potentially kind of tricky skew here if what we do at *mount* time is different from the filesystem/commit used by all of the rest of the code.

---

I'm OK to do something here but but the regex parsing JSON feels a bit too hacky for something security sensitive. How about the logic is:

- Detect if the target commit has a metadata key named `ostree.container.image-config` (without parsing it)
- If it does, resolve to the parent

?

alexlarsson:
> Detect if the target commit has a metadata key named ostree.container.image-config (without parsing it)
If it does, resolve to the parent

I think this is good. Although i think looking for the `ostree.importer.version` key makes it more clear what is going on.

cgwalters (approved):

PR #3525, opened 2025-09-18
remove $stateroot/var rw bind mount / improve do_remount()

champtar:
- ostree-remount: improve do_remount(), use warn()
    
    When we remount read only, we only want to change the vfs read only
    flag, not the fs one, so use MS_BIND.
    On the contrary when we remount read write, we want to change both (already ok).
    To check that we can write, we now use 'access'.
    
    Except when using old mount versions, we should never need do_remount(),
    as everything should have been properly setup by prepare-root,
    so start to warn() when we actually remounted /sysroot or /etc.

    - prepare-root/generator: remove $stateroot/var rw bind mount
    
    This allow to have /sysroot fully read-only.
    Recent versions of mount that use the new mount API (util-linux 2.39)
    will honor the 'bind,rw' mount option and create a writable bind mount
    even if the source was read-only.
    For older systems, ostree-remount will remount /var rw as needed.


champtar:
I've now added a small mount helper so var.mount is writable right away.

configure.ac
```diff
@@ -664,6 +664,29 @@ AM_COND_IF(BUILDOPT_SYSTEMD_AND_LIBMOUNT,
   AC_DEFINE([BUILDOPT_LIBSYSTEMD_AND_LIBMOUNT], 1, [Define if systemd and libmount]))
 if test x$with_libsystemd = xyes; then OSTREE_FEATURES="$OSTREE_FEATURES systemd"; fi
 
+AC_ARG_WITH(mounthelper,
```
cgwalters, line 667: Thanks for all the work on this!

I am really not excited about the idea of installing a globally accessible binary just to expose `Type=ostreebindrw` with `.mount` units.

I think instead couldn't we generate a `.mount` unit with a `Requires+After` on a `.service` which did a pre-mount or something?

champtar: I've played a bit more with this, if we mount something at `/var` before `var.mount` then `var.mount` will be enabled but it'll not run, and most importantly it'll print a confusing message (`/var was skipped because of an unmet condition check (ConditionKernelCommandLine=!systemd.volatile).`)

We can't use mount move without changing the mount propagation (`moving a mount residing under a shared mount is unsupported.`), so we could make `/sysroot` private, but this might have side effects.

Last solution is to have a first `.service` that prepare a temp bind mount already `rw` (at `$stateroot/var`), have `var.mount` bind that, then have a second `.service` to cleanup the temporary bind mount

Adding `mount.ostreebindrw` would only impact EL9 and is way simpler / cleaner than 2 extra `.service` IMO

@cgwalters what do you prefer ?

champtar: Another solution is also to keep thing as is for distro with old mount (EL9)

cgwalters: Well...yes, we could probably have a rhel9 branch here.

I guess bigger picture from my PoV we're trying really hard to work on the new bootc+composefs backend, and I'm hesitant on the risk/reward of further deeper changes here.

champtar: No need for a separate rhel9 branch, just a configure option

Will send the first cleanup commit in a separate PR and rework this one to remove the mount just for recent system, the changes should be smaller.

Will the new backend be usable with rpm-ostree ?

PR #3531, opened 2025-10-08
ci: use libsoup 3 by default

champtar:
libsoup-3.0-dev is present in all supported debian versions
libsoup2.4-dev is no present in testing anymore (forkie)


cgwalters (approved):
Thanks!

PR #3539, opened 2025-10-22
Add hooks for deployment steps to allow bootloader swapping

Mstrodl:
Can be used to provide hooks into a sysroot deploy for tools like bootupd to perform bootloader swaps.

Happy to do this a different way, but this is how we've been doing it on our debian-based system. I understand bootupd is supposed to work a different way. We do it this way because grub-mkconfig will generate a broken config for the older version of grub.

See #3496 for motivation


Mstrodl:
Hmm
Tests keep failing in different spots, but they look pretty unrelated. Is there some way to retry the jenkins one @cgwalters?

Also if it would be helpful I could share the source for our grub plugin. It's not perfectly atomic, but it's better than nothing :)

PR #3549, opened 2025-11-06
Add missing `:` to the gtk-doc in a few places

cgwalters:
This is SUCH a giant trap. I am not totally sure why it's working for me in a fedora-42 build env, but it seems like it may have broken in a different build environment in https://github.com/ostreedev/ostree/pull/3548#discussion_r2500278890

I used Sonnet to audit for similar instances beyond `read_blob` and it found some, fix those too.


jmarrero (approved):
lgtm

PR #3554, opened 2025-12-13
static-delta: Avoid copying bspatch output

razzeee:
Hey,

my C is very bad, so this was found and done with AI. After looking at the reports an https://github.com/flatpak/flatpak/issues/6255

I would expect this, to be slightly less performant, but work more reliably on systems with less RAM. Most of the things we do here, also seem to have been done in other code files here and there, so it doesn't seem to be unprecedented to go this route.

I haven't been able to runtime test this and I'm unsure how - especially validating, that it still works in the wild (e.g. doesn't break deltas)

### DIFF
--- a/src/libostree/ostree-repo-static-delta-processing.c
+++ b/src/libostree/ostree-repo-static-delta-processing.c
@@ -419,20 +419,34 @@ dispatch_bspatch (OstreeRepo *repo, StaticDeltaExecutionState *state, GCancellab
        * systems with limited RAM when processing large files.
        * See: https://github.com/flatpak/flatpak/issues/6255
        */
+      void *buf;
       g_auto (GLnxTmpfile) tmpf = {
         0,
       };
-      if (!glnx_open_anonymous_tmpfile (O_RDWR | O_CLOEXEC, &tmpf, error))
-        return FALSE;
 
-      /* Resize tmpfile to content_size */
-      if (ftruncate (tmpf.fd, state->content_size) < 0)
-        return glnx_throw_errno_prefix (error, "ftruncate");
+      if (state->content_size > 0)
+        {
+          if (!glnx_open_anonymous_tmpfile (O_RDWR | O_CLOEXEC, &tmpf, error))
+            return FALSE;
 
-      /* Memory-map the tmpfile for bspatch to write to */
-      void *buf = mmap (NULL, state->content_size, PROT_READ | PROT_WRITE, MAP_SHARED, tmpf.fd, 0);
-      if (buf == MAP_FAILED)
-        return glnx_throw_errno_prefix (error, "mmap");
+          /* Resize tmpfile to content_size */
+          if (ftruncate (tmpf.fd, state->content_size) < 0)
+            return glnx_throw_errno_prefix (error, "ftruncate");
+
+          /* Memory-map the tmpfile for bspatch to write to */
+          buf = mmap (NULL, state->content_size, PROT_READ | PROT_WRITE, MAP_SHARED, tmpf.fd, 0);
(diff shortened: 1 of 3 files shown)


src/libostree/ostree-repo-static-delta-processing.c
```diff
@@ -423,14 +456,48 @@ dispatch_bspatch (OstreeRepo *repo, StaticDeltaExecutionState *state, GCancellab
...
+      if (state->content_size > 0)
+        {
+          if (lseek (tmpf.fd, 0, SEEK_SET) < 0)
+            return glnx_throw_errno_prefix (error, "lseek");
+
+          /* Read and write in chunks to avoid loading entire file into memory */
+          guint64 bytes_remaining = state->content_size;
+          while (bytes_remaining > 0)
+            {
+              guchar chunk_buf[8192];
```
simonmicro: Have you considered bigger block sizes like experimented with in `src/libostree/ostree-repo-commit.c:1232`?

src/libostree/ostree-repo-static-delta-processing.c
```diff
@@ -423,14 +456,48 @@ dispatch_bspatch (OstreeRepo *repo, StaticDeltaExecutionState *state, GCancellab
...
+              gsize chunk_size = MIN (sizeof (chunk_buf), bytes_remaining);
+              gssize bytes_read;
+
+              do
+                bytes_read = read (tmpf.fd, chunk_buf, chunk_size);
+              while (G_UNLIKELY (bytes_read == -1 && errno == EINTR));
```
simonmicro: While reading, this chunk should not be "long enough" for a badly placed interrupt, you are going to retry it. Can you consider also patching `src/libostree/ostree-repo-commit.c:1240` for that?

src/libostree/ostree-repo-static-delta-processing.c
```diff
@@ -423,14 +456,48 @@ dispatch_bspatch (OstreeRepo *repo, StaticDeltaExecutionState *state, GCancellab
...
+      if (state->content_size > 0 && munmap (buf, state->content_size) < 0)
+        return glnx_throw_errno_prefix (error, "munmap");
+
+      if (bspatch_result < 0)
         return glnx_throw (error, "bsdiff patch failed");
 
-      if (!_ostree_repo_bare_content_write (repo, &state->content_out, buf, state->content_size,
-                                            cancellable, error))
-        return FALSE;
+      /* Now read from tmpfile and write to repository (if content_size > 0) */
```
simonmicro: Why was this done? Previously, the (potentially) large output buffer was used to be written into the repository via `_ostree_repo_bare_content_write`. Now, this code loads the patched content chunk-based into memory and then writes it to disk. But, as the source was changed to be potentially memory-backed, this is not needed - I would even say it introduced an unnecessary copy?

Furthermore, the `glnx_loop_write` part of `_ostree_repo_bare_content_write` now may performs additional calls to `write`, as the chuck sizes may not align with its retry-logic as from https://github.com/GNOME/libglnx/blob/de29c5f7d9df8d57b4f5caa9920f5d4caa7a8cfc/glnx-fdio.c#L752

simonmicro (changes requested):
I would be interested to learn more about your reasoning behind changing the code around `_ostree_repo_bare_content_write`.

cgwalters:
Sorry about the delay in review. This will need a rebase.

But bigger picture - I think what we should try to do here is actually pass the tempfile down into the commit path because that code is also going to allocate a tempfile and then commit, so we're just copying data from one tempfile to another...

razzeee:
The failing job seems to be unrelated

PR #3579, opened 2026-04-09
trivial-httpd: Fix const-correctness of slash pointer

kraj:
strrchr() returns a 'const char *' when passed a 'const char *' argument. Declare the local 'slash' variable as 'const char *' to match, fixing a build failure with clang when
-Werror,-Wincompatible-pointer-types-discards-qualifiers is active.


ericcurtin (approved):

ericcurtin:
@kraj you got unlucky with the build... Since const is only a build-time check, if it builds this is good

ericcurtin:
Build needs this to pass:

https://github.com/ostreedev/ostree/pull/3594

jmarrero (approved):

jmarrero:
Thank you for this @kraj I added your commit to #3599

PR #3583, opened 2026-04-13
kargs: Fix crash on quoted values in /proc/cmdline

### LINKED ISSUE #3582: 2026.1: Two tests fail on `OSTree:ERROR:src/libostree/ostree-kernel-args.c:197:split_kernel_args: 'quoted' should be FALSE`
Hi! 👋 

When trying to build 2026.1 as a package for Arch Linux I ran into two failing tests

```
[..]
ERROR: tests/test-admin-deploy-karg.sh - Bail out! OSTree:ERROR:src/libostree/ostree-kernel-args.c:197:split_kernel_args: 'quoted' should be FALSE
[..]
ERROR: tests/test-admin-instutil-set-kargs.sh - Bail out! OSTree:ERROR:src/libostree/ostree-kernel-args.c:197:split_kernel_args: 'quoted' should be FALSE
[..]
```

Full build and check log:

[ostree-2026.1-1-x86_64-build.log](https://github.com/user-attachments/files/26643704/ostree-2026.1-1-x86_64-build.log)
[ostree-2026.1-1-x86_64-check.log](https://github.com/user-attachments/files/26643705/ostree-2026.1-1-x86_64-check.log)

test-suite.log

[test-suite.log](https://github.com/user-attachments/files/26643714/test-suite.log)

jmarrero:
kargs: Fix crash on quoted values in /proc/cmdline
Bootloaders like GRUB may reformat quoted kernel arguments, producing
entries like "testparam=value with spaces" in /proc/cmdline. The
proc-cmdline parser was still using a naive space split (pre-dating the
quote-aware splitter added in https://github.com/ostreedev/ostree/commit/abc7d5b9a0eb312288fc9b31505aedba8c555e65), so these tokens got broken
apart and hit a g_assert_false() abort downstream.

Use the quote-aware splitter for /proc/cmdline too, and downgrade the
unterminated-quote assert to g_debug() since we can't control what
bootloaders emit.

Reported by Arch Linux maintainers building ostree 2026.1.

Closes: https://github.com/ostreedev/ostree/issues/3582
Assisted-by: OpenCode (Claude Opus 4.6)
Signed-off-by: Joseph Marrero Corchado <jmarrero@redhat.com>

### DIFF
--- a/src/libostree/ostree-kernel-args.c
+++ b/src/libostree/ostree-kernel-args.c
@@ -188,13 +188,14 @@ split_kernel_args (const char *str)
         }
     }
 
-  // Add the last slice
-  if (!quoted)
-    g_ptr_array_add (strv, g_strndup (start, str + len - start));
-  else
+  // Add the last slice if non-empty. If there's an unterminated quote
+  // (e.g. from bootloader-reformatted /proc/cmdline), treat the
+  // remainder as a single token rather than aborting.
+  if (str + len > start)
     {
-      g_debug ("Missing terminating quote in '%s'.\n", str);
-      g_assert_false (quoted);
+      if (quoted)
+        g_debug ("Missing terminating quote in '%s', treating remainder as single token.\n", str);
+      g_ptr_array_add (strv, g_strndup (start, str + len - start));
     }
 
   g_ptr_array_add (strv, NULL);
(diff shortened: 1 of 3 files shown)


cgwalters:
I know we probably need this in AGENTS.md but to me the commit message is overly verbose.

I have this in my system prompt: https://github.com/cgwalters/homegit/blob/main/dotfiles/.config/AGENTS.md#commit-messages

jeckersb:
I will plead ignorance to the details on how all of the rust<->C interop works on the ostree side, but is it feasible at all to swap to using the `kernel_cmdline` crate on the ostree side to handle this parsing?

cgwalters:
We only have Rust bindings for libostree; the C library does not call any Rust right now. Doing so would totally be possible, and I started some of that before but it would be a huge change and in practice we're already rewriting all of this in Rust with the bootc + composefs backend anyways so let's just focus on that.

jmarrero:
> I know we probably need this in AGENTS.md but to me the commit message is overly verbose.
> 
> I have this in my system prompt: https://github.com/cgwalters/homegit/blob/main/dotfiles/.config/AGENTS.md#commit-messages

done and rebased.

cgwalters (approved):

PR #3587, opened 2026-05-09
sysroot: Merge bootconfig-extra from previously staged deployment

jmarrero:
When a staged deployment is replaced by a new one (e.g. bootc sets source-tracked kargs, then rpm-ostree appends a karg before reboot), the bootconfig-extra keys from the first staging were lost. The new staging only checked the new deployment's bootconfig (always empty) and fell back to the merge deployment's bootconfig (the booted BLS file), which had no knowledge of what was in the previous staged GVariant.

Fix this by merging bootconfig-extra from three sources in priority order:

  1. Merge deployment's bootconfig (lowest priority)
  2. Previously staged deployment's bootconfig-extra
  3. New deployment's bootconfig (highest priority)

This ensures that x-options-source-* keys set by one consumer survive re-staging by another consumer on the same boot.

Assisted-by: OpenCode (Claude Opus 4.6)

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -3795,14 +3795,25 @@ _ostree_sysroot_ensure_finalize_staged_service (GError **error)
   return TRUE;
 }
 
+/* Merge all entries from an a{ss} GVariant into a string→string hash table.
+ * Used to accumulate bootconfig-extra keys from multiple sources during staging. */
+static void
+merge_extra_variant_into_table (GVariant *extra, GHashTable *table)
+{
+  GVariantIter iter;
+  const char *k, *v;
+  g_variant_iter_init (&iter, extra);
+  while (g_variant_iter_next (&iter, "{&s&s}", &k, &v))
+    g_hash_table_insert (table, g_strdup (k), g_strdup (v));
+}
+
 /**
  * ostree_sysroot_stage_tree_with_options:
  * @self: Sysroot
- * @osname: (allow-none): osname to use for merge deployment
+ * @osname: osname to use for merge deployment
  * @revision: Checksum to add
  * @origin: (allow-none): Origin to use for upgrades
- * @merge_deployment: (allow-none): Use this deployment for merge path
- * @opts: Options
+ * @opts: (nullable): Options
  * @out_new_deployment: (out): The new deployment path
  * @cancellable: Cancellable
  * @error: Error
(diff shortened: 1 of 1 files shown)


src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3888,27 +3888,97 @@ ostree_sysroot_stage_tree_with_options (OstreeSysroot *self, const char *osname,
...
+      if (!ot_openat_ignore_enoent (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED, &staged_fd, error))
+        return FALSE;
+      if (staged_fd != -1)
+        {
+          g_autoptr (GBytes) staged_contents = ot_fd_readall_or_mmap (staged_fd, 0, error);
+          if (!staged_contents)
+            return FALSE;
+          g_autoptr (GVariant) staged_data
+              = g_variant_new_from_bytes ((GVariantType *)"a{sv}", staged_contents, TRUE);
+          g_autoptr (GVariantDict) staged_dict = g_variant_dict_new (staged_data);
```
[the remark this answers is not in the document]
jmarrero: This is wrong.  We use the pattern in the in other places an it is because GLib g_variant_dict_init does NOT call g_variant_ref_sink or g_variant_ref. It just:

1. Creates a hash table
2. Iterates the input variant with g_variant_iter_init/g_variant_iter_next
3. Copies the key-value pairs into its own hash table

The dict holds no reference to the input variant. It copies the data out immediately.

Triple checked with Claude.

src/libostree/ostree-sysroot-deploy.c
```diff
@@ -3888,27 +3899,81 @@ ostree_sysroot_stage_tree_with_options (OstreeSysroot *self, const char *osname,
...
+      }
+
+    /* Priority 2: previously staged deployment's bootconfig-extra */
+    {
+      glnx_autofd int staged_fd = -1;
+      if (!ot_openat_ignore_enoent (AT_FDCWD, _OSTREE_SYSROOT_RUNSTATE_STAGED, &staged_fd, error))
+        return FALSE;
+      if (staged_fd != -1)
+        {
+          g_autoptr (GBytes) staged_contents = ot_fd_readall_or_mmap (staged_fd, 0, error);
```
cgwalters: Hmm I think we will have already loaded this data at this point and cached it in OstreeSysroot. Not big deal though.

cgwalters (approved):

src/libostree/ostree-sysroot-deploy.c
jmarrero: Did this.

jmarrero:
We need https://github.com/ostreedev/ostree/pull/3579 and some others fixes to fix CI here it seems. Probably F44 related too but I haven't investigated yet.

jmarrero:
OK tested this again, I think we are  good to merge the failure are unrelated and prow will go away soon I hope.

cgwalters:
This ended up breaking the tests added in bootc, see https://github.com/bootc-dev/bootc/pull/2277/changes/aa3a2fbad5433a654f413e720b9e35b7c0392b4b

This is...complicated. I think something we didn't consider enough when designing this is that bootc (because it has `/usr/lib/bootc/kargs.d` *always* passes `override_kernel_argv`), whereas that's not the default in ostree or rpm-ostree.

cgwalters:
Followup in https://github.com/ostreedev/ostree/issues/3609

PR #3594, opened 2026-06-08
static-delta: remove unused total_usize variable

ericcurtin:
In `_ostree_delta_get_endianness`, `total_usize` was accumulated across delta parts but never read or used in any heuristic or output. Remove the dead variable and its accumulation.

PR #3599, opened 2026-06-24
multiple CI fixes

jmarrero:
This picks up https://github.com/ostreedev/ostree/pull/3594 to fix prow failures and https://github.com/ostreedev/ostree/pull/3579 to fix Rust / build-git-libostree

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -77,7 +77,7 @@ RUN --network=none \
 FROM $base as rootfs
 COPY --from=rpmbuild /rpms/*.rpm /tmp/rpms/
 RUN ls /tmp/rpms/ && \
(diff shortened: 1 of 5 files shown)


cgwalters (approved):
Ideally in the future we do the "build derived container" instead of the cosa flow

cgwalters:
Let's just drop Prow. I want to get to a world where we only have GHA and Konflux as the main things we care about.

To drop it, we can basically make it a no-op to start, I think just have the Dockerfile be say `FROM scratch` and the tests do nothing? Then we drop it from openshift/release

cgwalters:
> Let's just drop Prow.

:arrow_right:  https://github.com/openshift/release/pull/81042

PR #3603, opened 2026-06-29
repo-commit: preserve existing .file object inode on consume commit

electricface:
repo-commit: preserve existing object inode when staging to objects/

    When ostree commit --consume is used and an object with the same
    checksum already exists in objects/, rename_pending_loose_objects() was
    unconditionally renaming the staging copy over it.  On Linux, renameat(2)
    atomically replaces the destination for two regular files, silently
    changing the inode of the existing repo object.

    Fix this by checking whether the object already exists in objects/
    before renaming.  If it does, the content is identical by definition
    (the object store is content-addressed by SHA256), so we can simply
    unlink the staging copy and keep the existing object with its original
    inode.

    Exception: .commitmeta objects are keyed by commit checksum rather than
    their own content, so they can be updated in place (e.g. when GPG
    signatures are added or deleted via ostree gpg-sign).  These are always
    renamed unconditionally.

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -1801,9 +1801,34 @@ rename_pending_loose_objects (OstreeRepo *self, GCancellable *cancellable, GErro
                                                     cancellable, error))
             return FALSE;
 
-          if (!glnx_renameat (child_dfd_iter.fd, loose_objpath + 3, self->objects_dir_fd,
-                              loose_objpath, error))
-            return FALSE;
+          /* For content-addressed objects, use RENAME_NOREPLACE so that an
+           * existing object keeps its inode.  If the destination already
+           * exists (EEXIST) the content is identical by definition, so just
+           * drop the staging copy.
+           *
+           * Exception: .commitmeta objects are keyed by commit checksum, not
+           * by their own content, so they can be updated in place (e.g. when
+           * GPG signatures are added or deleted).  Always overwrite those.
+           */
+          if (g_str_has_suffix (child_dent->d_name, ".commitmeta"))
+            {
+              if (!glnx_renameat (child_dfd_iter.fd, loose_objpath + 3, self->objects_dir_fd,
+                                  loose_objpath, error))
+                return FALSE;
+            }
+          else if (glnx_renameat2_noreplace (child_dfd_iter.fd, loose_objpath + 3,
+                                             self->objects_dir_fd, loose_objpath)
+                   < 0)
+            {
+              if (errno == EEXIST)
+                {
+                  /* Object already present with same content; drop staging copy */
+                  if (!glnx_unlinkat (child_dfd_iter.fd, loose_objpath + 3, 0, error))
+                    return FALSE;
+                }
+              else
+                return glnx_throw_errno_prefix (error, "renameat2(noreplace, %s)", loose_objpath);
+            }
         }
     }
 


src/libostree/ostree-repo-commit.c
```diff
@@ -1801,8 +1801,31 @@ rename_pending_loose_objects (OstreeRepo *self, GCancellable *cancellable, GErro
...
             return FALSE;
 
-          if (!glnx_renameat (child_dfd_iter.fd, loose_objpath + 3, self->objects_dir_fd,
-                              loose_objpath, error))
+          /* For .file objects: if one already exists in objects/ with the same
+           * checksum path, it has identical content (the store is
+           * content-addressed), so preserve the existing object and its inode
+           * rather than replacing it with the staging copy.
+           * All other object types (.commit, .dirtree, .dirmeta, .commitmeta,
+           * etc.) always go through rename so their content can be updated.
```
cgwalters: All objects are immutable, it does not make sense to me to just special case file objects.

src/libostree/ostree-repo-commit.c
```diff
@@ -1801,8 +1801,31 @@ rename_pending_loose_objects (OstreeRepo *self, GCancellable *cancellable, GErro
...
+                return FALSE;
+              obj_exists = (errno != ENOENT);
+            }
+          if (obj_exists)
+            {
+              /* Object already present with same content; drop staging copy */
+              if (!glnx_unlinkat (child_dfd_iter.fd, loose_objpath + 3, 0, error))
+                return FALSE;
+            }
+          else if (!glnx_renameat (child_dfd_iter.fd, loose_objpath + 3, self->objects_dir_fd,
```
cgwalters: Modern Linux has `renameat2(..., RENAME_NOREPLACE)` which is exactly what we want here, but we probably do need a fallback for old kernels if we get `EINVAL`.

cgwalters (changes requested):

src/libostree/ostree-repo-commit.c
cgwalters: Except for `commitmeta`, but...I don't think commitmeta goes through staging at all.

electricface: Thanks for the review! I wanted to follow up on the .commitmeta point, though I may be misreading the code — could you help confirm?

Looking at ostree_repo_write_commit_detached_metadata() in ostree-repo-commit.c:

int dest_dfd;
if (self->in_transaction)
dest_dfd = self->commit_stagedir.fd;
else
dest_dfd = self->objects_dir_fd;

It seems like when a pull is in progress (which runs inside a transaction), .commitmeta is written to the staging directory and would therefore go through rename_pending_loose_objects(). This is also consistent with the test failure we observed: without the .commitmeta exception, test-commit-sign.sh test 7 ("pull sig deleted") failed because the old .commitmeta in objects/ was preserved and the deleted GPG signature was never removed on the client side.

So my understanding is that .commitmeta does go through staging — at least during a pull. But I might be missing something about the overall flow. Could you take a look and let me know if I'm wrong?

As for the broader point about applying the optimization to all immutable object types (not just .file): that makes sense to me, and the current version does exactly that — all types except .commitmeta get the "skip rename if already exists" treatment. Happy to adjust if you see it differently.

src/libostree/ostree-repo-commit.c
electricface: Good point! I've updated the fix to use glnx_renameat2_noreplace() from libglnx, which already handles the old-kernel fallback internally — it tries renameat2(..., RENAME_NOREPLACE) first, and falls back to linkat + unlinkat when it gets EINVAL or ENOSYS.

The updated logic in rename_pending_loose_objects() is now:

For .commitmeta objects: always use a plain renameat (overwrite), since their content can change.

For all other content-addressed objects: use glnx_renameat2_noreplace(). If it returns EEXIST, the object is already present with identical content, so we just unlink the staging copy and keep the existing object — preserving its inode.

src/libostree/ostree-repo-commit.c
electricface: @cgwalters

cgwalters (approved):

cgwalters:
@electricface can you run clang-format?

electricface:
@cgwalters 
Executed clang-format.

PR #3604, opened 2026-06-29
Ci fixes3

jmarrero:
more CI fixes to prep for release


.cci.jenkinsfile
```diff
@@ -1,77 +0,0 @@
-// Documentation: https://github.com/coreos/coreos-ci/blob/main/README-upstream-ci.md
```
cgwalters, line 1: I missed you'd dropped this. Prow is gone now, but this one is still somewhat load bearing - yes a few tests are failing, but it gives us decent coverage.

That said, I bet we can convert it to GHA

PR #3610, opened 2026-07-07
Revert "sysroot: Merge bootconfig-extra from previously staged deployment"

### LINKED ISSUE #3609: 4c0d3a81 ("sysroot: Merge bootconfig-extra from previously staged deployment") broke bootc's tests
4c0d3a81 ("sysroot: Merge bootconfig-extra from previously staged deployment") regressed bootc's test here: https://github.com/bootc-dev/bootc/blob/54768712ee0308c51ccd27e8d7772d9c9f9aad39/tmt/tests/booted/test-loader-entries-source.nu

LLM-synthesized details:

<details>
## bootconfig-extra merge regression in ostree 2026.2

ostree commit 4c0d3a81 ("sysroot: Merge bootconfig-extra from previously
staged deployment"), included in v2026.2, introduced a 3-way merge of
`bootconfig-extra` extension keys during `stage_tree_with_options()`.
This breaks `bootc loader-entries set-options-for-source` when called
multiple times on the same boot, and is currently causing
`test-42-loader-entries-source` failures in bootc CI.

### Background

bootc tracks per-source kernel arguments via `x-options-source-*` BLS
extension keys (e.g. `x-options-source-dracut rd.driver.pre=vfio-pci`).
These keys are stored in ostree's `bootconfig-extra` dictionary, which
survives the staging roundtrip via the `/run/ostree/staged-deployment`
GVariant.

When staging, bootc sets the desired `x-options-source-*` keys on the
merge deployment's in-memory bootconfig, and passes the computed kargs
via `override_kernel_argv`. ostree picks up the extension keys from the
merge deployment and serializes them into the staged GVariant.

### The bug

Before 4c0d3a81, ostree used a simple fallback: if the new deployment
had extension keys, use those; otherwise use the merge deployment's.
This was effectively all-or-nothing — the merge deployment's keys were
the authoritative set.

4c0d3a81 changed this to a 3-way additive merge with priority ordering:

1. Merge deployment's bootconfig (lowest)
2. Previously staged deployment's `bootconfig-extra`
3. New deployment's bootconfig (highest)

The problem: bootc sets its keys on the merge deployment (Priority 1),
but the previously staged data (Priority 2) has higher priority and
overrides them.

### Reproduction

The `test-42-loader-entries-source` test does three stagings on one boot:

```
# Step 1: clear dracut kargs
bootc loader-entries set-options-for-source --source dracut --options ""

# Step 2: re-add dracut
bootc loader-entries set-options-for-source --source dracut --options "rd.driver.pre=vfio-pci"

# Step 3: delete dracut entirely
bootc loader-entries set-options-for-source --source dracut
```

Step 1 stages with `x-options-source-dracut=""` (tombstone). Step 2
sets `x-options-source-dracut="rd.driver.pre=vfio-pci"` on the merge
deployment, but ostree's merge picks up step 1's tombstone at Priority 2,
overriding it. The staged metadata now says dracut owns no kargs.

Step 3 reads the staged metadata, sees dracut owns no kargs, and has
nothing to remove from the options line. The resulting `override_kernel_argv`
still includes `rd.driver.pre=vfio-pci`. After reboot, `/proc/cmdline`
contains the karg that should have been removed.

### Fix

Restore the old default: the merge deployment's extension keys are
authoritative when present. Add an opt-in flag
`retain_pending_bootconfig_extra` for consumers like rpm-ostree that
re-stage without knowledge of extension keys and need to preserve
previously staged extras.
</details>

cgwalters:
This reverts commit 4c0d3a819033166ec9753ff8c5fdac5db1bac813.

The 3-way merge of bootconfig-extra during staging breaks bootc loader-entries set-options-for-source when called multiple times on the same boot.  Each call sets the desired x-options-source-* keys on the merge deployment's in-memory bootconfig, but the merge logic gives higher priority to the previously staged data, overriding the caller's updates.  This corrupts the source-tracking metadata that bootc relies on to compute override_kernel_argv, causing stale kargs to persist across reboots.

The scenario this commit was trying to fix (rpm-ostree re-staging after bootc on the same boot, preserving bootc's extension keys) will need a different approach that does not conflict with callers that update extension keys across multiple stagings.

Closes: https://github.com/ostreedev/ostree/issues/3609

### DIFF
--- a/src/libostree/ostree-sysroot-deploy.c
+++ b/src/libostree/ostree-sysroot-deploy.c
@@ -3795,25 +3795,14 @@ _ostree_sysroot_ensure_finalize_staged_service (GError **error)
   return TRUE;
 }
 
-/* Merge all entries from an a{ss} GVariant into a string→string hash table.
- * Used to accumulate bootconfig-extra keys from multiple sources during staging. */
-static void
-merge_extra_variant_into_table (GVariant *extra, GHashTable *table)
-{
-  GVariantIter iter;
-  const char *k, *v;
-  g_variant_iter_init (&iter, extra);
-  while (g_variant_iter_next (&iter, "{&s&s}", &k, &v))
-    g_hash_table_insert (table, g_strdup (k), g_strdup (v));
-}
-
 /**
  * ostree_sysroot_stage_tree_with_options:
  * @self: Sysroot
- * @osname: osname to use for merge deployment
+ * @osname: (allow-none): osname to use for merge deployment
  * @revision: Checksum to add
  * @origin: (allow-none): Origin to use for upgrades
- * @opts: (nullable): Options
+ * @merge_deployment: (allow-none): Use this deployment for merge path
+ * @opts: Options
  * @out_new_deployment: (out): The new deployment path
  * @cancellable: Cancellable
  * @error: Error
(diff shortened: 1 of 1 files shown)


cgwalters:
I tried to spend some time thinking about a fix, but honestly with the previous code kargs were starting to get beyond me understanding how they work.

Let's get a clean revert in, and then go back to thinking about a design that fixes the reported issue after.

jmarrero:
lgtm but I can also try to add fix on top of this. I got an agent with all the context from all this work. But up to you. Still this would require rpm work to make the tests pass regardless.

jmarrero (approved):

cgwalters:
> lgtm but I can also try to add fix on top of this. I got an agent with all the context from all this work. But up to you.

Let's do that as a separate PR. Given the breakage I want a new approach to get extra scrutiny.

>  Still this would require rpm work to make the tests pass regardless.

Do you mean rpm-ostree work?

PR #3611, opened 2026-07-07
sysroot: Fix bootconfig-extra merge for repeated staging

### LINKED ISSUE #3609: 4c0d3a81 ("sysroot: Merge bootconfig-extra from previously staged deployment") broke bootc's tests

jmarrero:
Reapply the intent of #3587 (reverted in #3610 because of #3609): extension BLS keys such as `x-options-source-NAME` that only exist in a previously staged deployment must survive when another tool re-stages on the same boot, e.g. `rpm-ostree kargs` after `bootc loader-entries set-options-for-source`.

The two kinds of caller need opposite priorities. bootc sets the complete desired key set on the merge deployment before staging, and that must win over stale previously-staged data (#3609). rpm-ostree passes the booted deployment through untouched, and for it the previously-staged data is newer than what's on disk. Preferring the on-disk keys whenever there are any doesn't work either: a removed source leaves an empty tombstone behind, so the on-disk set won and bootc's staged keys were dropped, orphaning kargs it could no longer diff against. The previous revision of this PR failed its own test that way.

So track it explicitly: `OstreeBootconfigParser` remembers whether a non-standard key was `set()` since it was loaded, and staging takes the merge deployment's keys if so, else the previously staged set, else the on-disk keys. Comparing against the on-disk entry instead would not work, since a caller may legitimately set a key back to its on-disk value after staging something else.

While here, treat `aboot`/`abootcfg`/`fdtdir` as standard keys like `devicetree`, since finalization rebuilds them.

The TMT plan takes ~50 minutes on a GitHub runner, so it gets its own matrix job (via a new `--exclude` for the tmt xtask) instead of running in series with the other plans. `docs/bootconfig-extra.md` has the full design.

Validated by the unit tests (every branch of the selection, including both regressions), the 6-boot cross-consumer TMT plan on stream9 and stream10, and the coreos-ci kola suite (with #3656 applied).

Closes: #3609

Generated-by: AI
I am knowledgeable in this problem domain and reviewed it carefully.


jmarrero:
We talked about this offline and next step is making sure that all test cases AI is running are covered in ostree and bootc tests. In the future we should also consider running the bootc tests as part of the gating here.

cgwalters:
A giant complexity in all of this is basically that:

- ostree only has relatively unopinionated kargs support; there's all the low level hooks to write bootloader entries with whatever we want, but the default is to inherit from booted
- bootc has `/usr/lib/bootc/kargs.d` which is a custom thing on top
- rpm-ostree wraps ostree with its own extra stuff

In particular a case I worry about is mixing `/usr/lib/bootc/kargs.d` with rpm-ostree or other local kargs - where the two worlds collide. This one has to work.

cgwalters:
OK output of a session I had will follow...one nonblocking bug the LLM claims

<details>
# ostreedev/ostree PR #3611 — Review Findings

PR: "sysroot: Fix bootconfig-extra merge for repeated staging"
https://github.com/ostreedev/ostree/pull/3611
Branch checked out locally as `pr-3611` (HEAD `724e9fea7`), cross-checked against
`/var/home/ai/src/github/bootc-dev/bootc` (main, `2e9413b6f9`).

## Background / history

`bootconfig-extra` is a mechanism to carry "extension" BLS keys (anything not in
`title, version, options, linux, initrd, devicetree`) — e.g. bootc's
`x-options-source-<name>` — through ostree's staged-deployment roundtrip
(`/run/ostree/staged-deployment`, tmpfs) so they survive to finalization at shutdown.
Logic lives in `ostree_sysroot_stage_tree_with_options()`,
`src/libostree/ostree-sysroot-deploy.c`.

1. **4c0d3a8 / PR #3587** (merged, shipped in ostree 2026.2): 3-way *additive* merge,
   priority new-deployment (high) > previously-staged (mid) > merge-deployment (low).
2. **Regression (#3609)**: because merge-deployment had *lowest* priority, stale
   previously-staged tombstones could override the caller's fresh, correct values.
   Broke bootc's `loader-entries set-options-for-source` when called multiple times
   in one boot. bootc CI (`tmt/tests/booted/test-loader-entries-source.nu`) had to add
   a version-skip guard for ostree "2026.2".
3. **PR #3610**: clean revert of 4c0d3a8.
4. **PR #3611** (current): re-adds the previously-staged fallback but as
   **all-or-nothing**: new deployment's extra keys (if any) → else merge deployment's
   extra keys (if any) → else previously-staged verbatim.

PR is 2 commits: first (`21bceb254`) mechanically re-reverts back to the *known-buggy*
3-way-merge code; second (`724e9fea7`) replaces it with the real fix. Intermediate
state is genuinely broken — poor for bisection/review; should be squashed.

Author's PR description notes this was AI-generated ("Tested using OpenCode
(Opus 4.6), all code done with Opus") with manual/external validation (5 regression
tests R1-R5, bootc's 10-scenario/6-boot test) that is **not** captured as automated
tests in the ostree repo (`changedFiles: 1`, only the `.c` file).

## Current code (PR #3611)

```c
/* Serialize any extension BLS keys (e.g. x-options-source-tuned). ... */
{
  g_autoptr (GVariant) extra = NULL;
  OstreeBootconfigParser *bootconfig = ostree_deployment_get_bootconfig (deployment);
  if (bootconfig)
    {
      GVariant *v = _ostree_bootconfig_parser_get_extra_keys_variant (bootconfig);
      if (v)
        extra = g_variant_ref_sink (v);
    }
  if (!extra && merge_deployment)
    {
      OstreeBootconfigParser *merge_bootconfig
          = ostree_deployment_get_bootconfig (merge_deployment);
      if (merge_bootconfig)
        {
          GVariant *v = _ostree_bootconfig_parser_get_extra_keys_variant (merge_bootconfig);
          if (v)
            extra = g_variant_ref_sink (v);
        }
    }
  /* Last resort: inherit from a previously staged deployment. ... */
  if (!extra && self->staged_deployment_data)
    {
      extra = g_variant_lookup_value (self->staged_deployment_data, "bootconfig-extra",
                                      (GVariantType *)"a{ss}");
    }
  if (extra)
    g_variant_builder_add (builder, "{sv}", "bootconfig-extra", extra);
}
```

`_ostree_bootconfig_parser_get_extra_keys_variant()` (unchanged): builds an `a{ss}`
dict of non-standard keys, returns **NULL if it would have zero entries** — so it
cannot distinguish "no extra keys ever set" from "extra keys explicitly cleared to
zero."

## Key verified findings

### A. Tier 1 of the fallback chain is dead code (confirmed by direct source reading)

`deployment` (the "new" deployment checked in tier 1) is created *inside*
`ostree_sysroot_stage_tree_with_options()` via
`sysroot_initialize_deployment()` → `_ostree_deployment_set_bootconfig_from_kargs()`:

```c
g_autoptr (OstreeBootconfigParser) bootconfig = ostree_bootconfig_parser_new ();  // blank
ostree_deployment_set_bootconfig (deployment, bootconfig);
if (override_kernel_argv)
  ostree_bootconfig_parser_set (bootconfig, "options", new_options);  // only "options" (standard key)
```

No caller ever gets a handle to this object before the bootconfig-extra block runs
later in the same function. So tier 1 is **structurally guaranteed to always be
empty/NULL** — it can never fire under the current implementation. The doc comment
("Extension keys are taken from the new deployment's bootconfig first...") describes
a path that cannot happen. Should either be removed (simplify to 2-tier: merge
deployment → previously-staged) or the comment corrected to explain why it's kept.

### B. `self->staged_deployment_data` scoping

- `GVariant*` (`a{sv}`) field on `OstreeSysrootPrivate`; populated by
  `_ostree_sysroot_reload_staged()` (part of `ostree_sysroot_load()`), read fresh from
  `/run/ostree/staged-deployment` (tmpfs; persists across separate process
  invocations within one boot; cleared at finalization/shutdown).
- `_ostree_sysroot_reload_staged()` *also* replays `bootconfig-extra` onto
  `self->staged_deployment`'s own in-memory bootconfig via
  `ostree_bootconfig_parser_set()` — separate from the raw GVariant used in the
  fallback.
- `ostree_sysroot_get_merge_deployment()` returns **`self->booted_deployment`** (not
  staged) when staging into the currently-booted osname — the common case. So a
  not-yet-finalized staging's extension keys are not automatically on
  `merge_deployment`'s bootconfig; only finalized (post-reboot) keys baked into the
  booted BLS file are.

### C. bootc's actual usage pattern (verified in `crates/lib/src/loader_entries.rs`)

`set_options_for_source_staged()` does **not** rely on ostree to merge/accumulate
keys across calls. On every invocation it:
1. Reads all known `x-options-source-*` sources from (a) the booted deployment's
   on-disk BLS entry and (b) if a staged deployment exists, directly parses
   `/run/ostree/staged-deployment`'s `bootconfig-extra` itself
   (`read_staged_bootconfig_extra_sources()`), bypassing ostree's C API.
2. Accumulates these into a `BTreeMap<name, value>` (monotonic; entries are only
   added, essentially never dropped from bootc's own bookkeeping).
3. Computes desired state: tombstones (`""`) ALL known source keys, re-sets all
   preserved sources' real values (except the one being modified), then
   sets/tombstones the target source per the CLI request.
4. Calls `merge_bootconfig.set(key, value)` for every one of these keys directly on
   `merge_deployment`'s (i.e. the booted deployment's) in-memory bootconfig — so by
   the time `stage_tree_with_options()` is called, `merge_deployment` already has the
   full, correct, non-empty desired key set (once any source has ever existed for the
   stateroot).
5. Passes this `merge_deployment` into `stage_tree_with_options`.

Because bootc tombstones with `""` rather than truly deleting keys,
`_ostree_bootconfig_parser_get_extra_keys_variant()` treats a `""`-valued key as
present (`has_entries = TRUE`; it only checks key membership, not value truthiness).
So bootc's own re-staging calls virtually always satisfy tier 2 and never reach tier
3. **Tier 3 (previously-staged fallback) is, in practice, only exercised by
genuinely unaware consumers like rpm-ostree** — exactly the cross-consumer case
#3611 says it's fixing.

### D. Latent API gap: can't force "truly zero" extension keys

Because the getter returns NULL for zero entries, ostree's fallback can't
distinguish "consumer never touched extension keys" from "consumer explicitly
cleared them to zero." A caller wanting to legitimately reach zero extension keys
within one boot can't — it will silently inherit stale previously-staged keys.
**Not reachable by bootc today** (due to its tombstone-forever convention), but a
real footgun for any future/other consumer that doesn't follow that convention.

### E. Semantic drift between `options` and `bootconfig-extra` (the concern behind
cgwalters' PR comment)

cgwalters' review comment raised: mixing bootc's `/usr/lib/bootc/kargs.d` with
rpm-ostree/other local kargs "where the two worlds collide... has to work."

- `/usr/lib/bootc/kargs.d/*.toml` (container-embedded static kargs, diffed
  old-image-vs-new-image directly against the plain `options` string via
  `bootc_kargs.rs`) is **completely separate** from `x-options-source-*` /
  `bootconfig-extra` — kargs.d never writes/reads `x-options-source-*` keys.
- `options` is in `is_standard_bls_key()`'s allowlist and is therefore **never**
  part of `bootconfig-extra`. The two channels can't corrupt each other's bytes —
  this part of the concern is addressed by construction (independent of #3611).
- However: `x-options-source-<name>` is pure bookkeeping metadata bootc maintains
  alongside `options`, with **no reconciliation** if a third party edits `options`
  directly. E.g. `rpm-ostree kargs --delete=isolcpus=1-3` removes the arg from
  `options` but leaves `x-options-source-tuned=isolcpus=1-3` untouched.
  `compute_merged_options()` in bootc calls `merged.remove_exact()` against the old
  recorded value — a safe no-op if already absent — so this **self-heals** the next
  time bootc touches that source, and is not boot-breaking/state-corrupting like the
  original #3609 regression. But metadata can go stale/inaccurate (e.g.
  `bootc loader-entries list` could report a source "owns" an arg an admin already
  stripped) until bootc next touches that source. This is a pre-existing property of
  the whole `x-options-source-*` design (since 4c0d3a8), not introduced or fixed by
  #3611 — but a real gap in test coverage: `test-loader-entries-source.nu` never
  invokes `rpm-ostree kargs`, and `test-image-pushpull-upgrade.nu` tests kargs.d +
  local kargs but never touches `bootconfig-extra`.

## How `options` and `bootconfig-extra` actually relate

A BLS entry file is flat `key value` lines. `options` (standard key) is the literal
kernel command line the kernel sees. `x-options-source-<name>` keys are non-standard
extension keys — pure bookkeeping bootc invented to record "this substring of
`options` belongs to source `<name>`". **`bootconfig-extra` is not a BLS key at
all** — it's a transient GVariant dict that exists solely to carry non-standard keys
through the staging roundtrip (`stage_tree_with_options()` → `/run/ostree/staged-deployment`
→ finalization writes them back out as `x-...` lines). `options` is excluded from it
by `is_standard_bls_key()`.

bootc's `set_options_for_source_staged()` treats `options` as ground truth and the
`x-options-source-*` map as "which substring I own": to update source `X`, remove
the *old* recorded value for `X` from `options`, add the *new* value, and re-set
`x-options-source-X`. **Nothing cross-checks the two** — ostree serializes/merges
`options` and `bootconfig-extra` completely independently. If something external
(rpm-ostree) edits `options` without updating `x-options-source-*`, the metadata
just goes stale until bootc next touches that source (self-healing, not
boot-breaking — see finding E above).

## Idea floated: unify kargs.d under `x-options-source-bootc`

Today `/usr/lib/bootc/kargs.d/*.toml` (container-embedded static kargs, in
`crates/lib/src/bootc_kargs.rs`) is a **third, separate** kernel-arg mechanism, not
using `x-options-source-*` at all. `get_kargs()` re-derives "old vs new" by
re-parsing kargs.d TOML files out of two actual container filesystem trees (the old
merge deployment's tree and the new pending commit's tree) each time, then diffs and
splices directly against `options`:

```rust
let existing_kargs = get_kargs_in_root(merge_root, sys_arch)?;         // re-parse OLD tree
let remote_kargs   = get_kargs_from_ostree(repo, &fetched_tree, ...)?; // re-parse NEW tree
compute_apply_kargs_diff(&existing_kargs, &remote_kargs, &mut kargs);   // remove old, add new
```

Proposal: record kargs.d's currently-applied set as `x-options-source-bootc` (or
`-kargsd`), reusing the same `compute_merged_options()`-style "remove old recorded
value, add new value" pattern used by `set-options-for-source`, instead of the
bespoke two-tree diff.

**Benefits:**
- One mechanism/one test surface instead of two independently-implemented,
  independently-tested kernel-arg-merging code paths that both write into `options`.
  Automatically inherits whatever `bootconfig-extra` fallback-chain fix/tests land
  from #3611.
- Cheaper: read a recorded string instead of checking out and re-parsing the old
  image's filesystem tree just to know what it previously contributed.
- Introspectable: kargs.d's contribution becomes labeled/visible in the BLS file and
  to any tooling enumerating `x-options-source-*` (e.g. future `loader-entries list`).

**Costs / risks (not free):**
1. **Migration**: existing systems have kargs.d args in `options` with no
   `x-options-source-bootc` key yet. Need a one-time fallback: if the key is absent,
   do today's re-derive-from-old-tree diff once, then start recording going forward.
2. **Name reservation**: `set-options-for-source --source <name>` currently accepts
   any charset-valid name (`SourceName::parse` only checks alphanumeric/`-`/`_`).
   Whatever name is chosen (`bootc`, `kargsd`, etc.) must be reserved/rejected by
   `SourceName::parse` so a user/other tool can't collide with or hijack it.
3. **Real rewrite required**: `get_kargs()` would need to move to the same
   "read all known sources → recompute full desired state → set full state on
   `merge_deployment`" pattern as `set_options_for_source_staged()`. A partial/
   incremental update relying on ostree to accumulate state across calls is exactly
   the pattern that caused #3609 — must avoid repeating it here.
4. **Doesn't fully close the rpm-ostree drift gap** — even unified, an external tool
   editing `options` directly still won't auto-update `x-options-source-bootc`; same
   self-healing-on-next-touch behavior as any other source. What it *does* fix is
   making kargs.d's drift risk identical to, and covered by the same fix/tests as,
   every other tracked source, rather than being a third untested interaction
   surface.

This would be a bootc-side follow-up, not something that should block ostree
PR #3611 (which is only about the ostree-side merge/fallback semantics that this
refactor would then also govern).

## Other review notes

- **Memory management**: correct. `g_variant_ref_sink()` properly sinks the two
  floating refs from `_ostree_bootconfig_parser_get_extra_keys_variant()`;
  `g_variant_lookup_value()` returns a full ref; all cleaned up via `g_autoptr`. No
  leaks or floating-ref bugs.
- **Style nit**: `(GVariantType *)"a{ss}"` casts away const non-idiomatically;
  should use `G_VARIANT_TYPE ("a{ss}")`.
- **Doc comment**: `ostree_sysroot_stage_tree_with_options()`'s public doc comment
  doesn't describe this 3-tier fallback or its limitations at all.
- **Test coverage**: zero automated tests added for a feature now on its second
  revert/fix cycle in a few weeks. Existing `tests/test-bootconfig-parser-internals.c`
  only unit-tests `_ostree_bootconfig_parser_get_extra_keys_variant()` in isolation,
  not the cross-call staging fallback logic.

## @architect-g verdict (independent review)

- Algorithm: **sound** for the stated goal (fixes #3609, preserves rpm-ostree
  cross-consumer inheritance).
- Empty-dict API gap: real but **not a blocker**; track as follow-up (getter should
  signal "explicitly empty" vs "absent" rather than NULL for both).
- Test coverage: **concerning** given repeat-regression history; recommends a
  minimal C test exercising the 3-tier fallback (unaware consumer inherits; aware
  consumer's set wins outright over stale previously-staged tombstones).
- Confirmed tier-1 dead-code finding independently from source inspection.
- Recommendation: **request changes** — squash the two commits, fix the
  `G_VARIANT_TYPE` cast, add doc comment describing the fallback chain and its
  limitations, drop or explain dead tier-1 check — then merge, with a follow-up
  issue for the automated test and the empty-dict ambiguity.

## Suggested PR reply (drafted, not yet posted)

> The `options` string and `bootconfig-extra` are serialized as independent GVariant
> dict keys, and `options` is explicitly excluded from `bootconfig-extra` handling
> (`is_standard_bls_key`), so they can't corrupt each other's bytes. But there's no
> reconciliation between them: `x-options-source-*` is pure bookkeeping bootc
> maintains on the side, so if `rpm-ostree kargs` (or anything else) edits `options`
> directly, that metadata can go stale until bootc next touches the same source — it
> self-heals on the next `set-options-for-source` call for that source, but there's
> no test today that exercises `set-options-for-source` interleaved with
> `rpm-ostree kargs`. Worth adding one alongside whatever coverage gets added for the
> R1-R5 scenarios.
>
> Separately: the new "tier 1" check (new deployment's bootconfig) can never actually
> fire — `_ostree_deployment_set_bootconfig_from_kargs()` always hands it a fresh
> bootconfig with at most the `options` key set, and no caller has a handle to it
> before this code runs. Suggest dropping tier 1 or documenting why it's kept.

## Status / open items

- Not yet posted to the PR — awaiting decision on whether to post as review comment
  and/or push a revised patch (drop dead tier 1, fix cast, update doc comment, squash
  commits, add C test).

</details>

jmarrero:
I think this should be in good state now @cgwalters , it will pair with: https://github.com/bootc-dev/bootc/pull/2330 which has some additional fixes but that will require the next ostree.

jmarrero:
The doc with the full tuned to ostree processes is in: https://gist.github.com/jmarrero/822fe7269783c8177f6e2d09f0007973  

I scoped the doc change here just for ostree and just bootc in https://github.com/bootc-dev/bootc/pull/2330

tmt/tests/booted/test-bootconfig-extra-staging.sh
```diff
@@ -0,0 +1,316 @@
...
+set -xeuo pipefail
+
+REBOOT_COUNT="${TMT_REBOOT_COUNT:-0}"
+
+# Save system kargs on first boot so we can verify they survive all phases
+SYSTEM_KARGS_FILE="/var/ostree-test-system-kargs.txt"
+
+parse_cmdline() {
+    cat /proc/cmdline
+}
```
cgwalters, line 49: Quite a parser 😄

docs/bootconfig-extra.md
```diff
@@ -0,0 +1,277 @@
...
+{:toc}
+
+<!-- SPDX-License-Identifier: (CC-BY-SA-3.0 OR GFDL-1.3-or-later) -->
+
+This document describes how OSTree carries *extension* keys of
+[Boot Loader Specification](https://uapi-group.org/specifications/specs/boot_loader_specification/)
+(BLS) entries through the staged-deployment lifecycle, and the contract
+tools that stage deployments must follow. It is aimed at people changing
+`ostree_sysroot_stage_tree_with_options()`, `OstreeBootconfigParser`, or
+a tool that stages deployments.
```
cgwalters, line 18: Hmmm....i think if it's about the C api it should be C docs, and the stuff we put in doc/ should roughly be for people using the CLI?

There's a bit of a fuzzy line of course. It definitely makes sense to document the *flow* including standard keys etc here.

cgwalters (commented):
I'll be honest I tried to read https://gist.github.com/jmarrero/822fe7269783c8177f6e2d09f0007973 and just found that my brain was unable to encompass it. Even though I know the background and terminology. It might be right. It might be completely wrong. It might as well be https://github.com/openai/NavierStokesAndEuler/blob/main/NavierStokes/ActivationCone.lean to me...

But anyways...I skimmed the tests and I don't see anything obviously wrong at least...

tmt/tests/booted/test-bootconfig-extra-staging.sh
jmarrero: Fable 5.1 is so good

jmarrero:
> I'll be honest I tried to read https://gist.github.com/jmarrero/822fe7269783c8177f6e2d09f0007973 and just found that my brain was unable to encompass it. Even though I know the background and terminology. It might be right. It might be completely wrong. It might as well be https://github.com/openai/NavierStokesAndEuler/blob/main/NavierStokes/ActivationCone.lean to me...
> 
> But anyways...I skimmed the tests and I don't see anything obviously wrong at least...

I guess having this be on the BLS and having to care about rpm-ostree and bootc on top is a bit painful trying to keep track of the states and callers.

I wonder if we close this PR, and move the implementation into bootc only and using the origin file to track the state there. Maybe that way we just have bootc write the kargs into any deployment taking the origin file as source of truth instead of trying to be merging states. 

What I am not sure is if rpm-ostree would still make this hard by touching the origin file.

docs/bootconfig-extra.md
```diff
@@ -0,0 +1,277 @@
...
+
+OSTree does not interpret extension keys. The consumer that defines them
+today is bootc, which records which kernel arguments a *source* (such as
+TuneD) owns in `x-options-source-NAME` keys; that side is described in
+[bootc's kernel arguments documentation](https://bootc-dev.github.io/bootc/building/kernel-arguments.html).
```
cgwalters, line 23: This doesn't seem to be there yet?

docs/bootconfig-extra.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Known limitations
+
+* A staged deployment is replaced wholesale. Two aware callers must each
+  read the current staged state before writing, or the earlier change is
+  lost together with its extension keys.
+* An aware caller cannot express "no extension keys at all" in a way
+  that overrides previously staged data: tier 1 only fires if a
+  non-standard key was set. Tombstoning with an empty value avoids the
+  question.
```
cgwalters, line 246: I think it's basically that replacing with an empty value is the API right?

docs/bootconfig-extra.md
```diff
@@ -0,0 +1,277 @@
...
+priorities:
+
+* An **aware caller** manages extension keys itself. It reads the
+  current state (booted entry *and* staged data), computes the complete
+  desired set, and sets every key on the merge deployment before
+  staging. That set must win: if a previously staged tombstone were
+  allowed to override it, replacing or re-adding a key twice in one boot
+  would silently keep the stale tombstone
+  ([#3609](https://github.com/ostreedev/ostree/issues/3609)).
+* An **unaware caller** (`rpm-ostree`, `bootc upgrade`, `ostree admin
```
cgwalters, line 164: Hmm is this right? I mean all of these call into libostree, isn't that the point of this?

cgwalters:
I am still trying to grasp all of this, will spend another 30 minutes after lunch today

PR #3614, opened 2026-07-22
lib/commit: Fix min-free-space accounting for duplicate content objects

jlebon:
First, in the _ostree_repo_bare_content_commit(), we never actually checked whether the file was already present in the repo before linking it in place. Do this so that we can no-op up front.

Second, and the actual bug this patch is fixing: the min-free-space accounting in write_content_object() and
_ostree_repo_bare_content_commit() optimistically reserves space from the running `txn.max_blocks` counter _before_ knowing whether we'll actually no-op or not based on the object already existing.

Over many duplicate writes within a single transaction, the counter diverges from the actual free space and eventually hits zero, causing a spurious "min-free-space would be exceeded" error even with plenty of disk space available.

This may be the source of the CI issue FCOS is hitting in https://github.com/coreos/fedora-coreos-config/pull/4265 due to the bootc SELinux relabeling done since https://github.com/bootc-dev/bootc/pull/2088.

This patch saves the number of blocks reserved and credits them back in both code paths when a duplicate object is detected.

This was heavily AI-guided, finding the bug, and then doing red-green testing towards the fix.

Assisted-by: AI


jlebon:
Leaving it as draft for now because this still needs a critical eye, even if my own in more depth. But it does look sane overall.

src/libostree/ostree-repo-commit.c
```diff
@@ -568,9 +570,25 @@ _ostree_repo_bare_content_commit (OstreeRepo *self, OstreeRepoBareContent *barew
                                            checksum_buf, error))
     return FALSE;
 
-  if (!commit_loose_regfile_object (self, checksum_buf, &real->tmpf, real->uid, real->gid,
-                                    real->mode, real->xattrs, cancellable, error))
+  /* Check if this object already exists.  If so, credit back the space
+   * reservation we made above — no new disk space is consumed.
+   */
+  gboolean have_obj;
+  if (!_ostree_repo_has_loose_object (self, checksum_buf, OSTREE_OBJECT_TYPE_FILE, &have_obj,
```
cgwalters: This is still racy in the presence of multiple threads - see https://github.com/ostreedev/ostree/pull/3603 for related discussion.

src/libostree/ostree-repo-commit.c
```diff
@@ -1100,11 +1126,14 @@ write_content_object (OstreeRepo *self, const char *expected_checksum, GInputStr
...
     return FALSE;
-  /* If we already have it, just update the stats. */
+  /* If we already have it, just update the stats.  Also credit back the
+   * free-space reservation we made above — no new disk space was consumed.
+   */
   if (have_obj)
     {
       g_mutex_lock (&self->txn_lock);
       self->txn.stats.content_objects_total++;
+      self->txn.max_blocks += object_blocks_reserved;
```
cgwalters, line 1138: Why can't this just be `+= object_blocks`? Why a separate variable?

tests/test-repo.c
```diff
@@ -541,6 +543,105 @@ test_repo_lock_multi_thread (Fixture *fixture, gconstpointer test_data)
...
+  /* Use bare-user-only so we don't need root privileges. */
+  g_autoptr (OstreeRepo) repo = ostree_repo_create_at (
+      fixture->tmpdir.fd, ".", OSTREE_REPO_MODE_BARE_USER_ONLY, NULL, NULL, &error);
+  g_assert_no_error (error);
+
+  /* Figure out how much free space the underlying filesystem has so we can
+   * set min-free-space-size to leave a tight-but-sufficient budget. There is a race here of course
+   * because this runs on a shared machine and other things could be filling up the disk
+   * so fast that between this fstatvfs and the one libostree does, that it shinks the max_blocks
+   * below what the test expects. In practice... It seems quite unlikely to be a serious concern.
```
cgwalters: I dunno, that does seem like it would be flaky on a multitenant environment.

What may help here is an API to override the repository's view of total free space?

src/libostree/ostree-repo-commit.c
```diff
@@ -568,9 +570,25 @@ _ostree_repo_bare_content_commit (OstreeRepo *self, OstreeRepoBareContent *barew
...
     return FALSE;
+  if (have_obj)
+    {
+      g_mutex_lock (&self->txn_lock);
+      self->txn.max_blocks += object_blocks_reserved;
+      g_mutex_unlock (&self->txn_lock);
+    }
+  else
+    {
+      if (!commit_loose_regfile_object (self, checksum_buf, &real->tmpf, real->uid, real->gid,
```
cgwalters: So what I think would be cleaner is if `commit_loose_regfile_object` returned a boolean value for whether it actually replaced the file or not. Then there's no extra syscall and no race.

jlebon:
OK a really obvious thing here is that this is a unit test so we can just... directly set `max_blocks` on the transaction object and avoid all that awkward quota guesstimation.

cgwalters (approved):

PR #3615, opened 2026-07-24
lib/commit: Fix min-free-space accounting for reflinked content objects

jlebon:
When writing a content object,
create_regular_tmpfile_linkable_with_content() opportunistically tries FICLONE to reflink data from the input fd. The reflinked file shares its data blocks via copy-on-write, so the actual disk cost is just metadata.

However, the min-free-space accounting in write_content_object() reserves the full file size in max_blocks upfront and never credits it back after a successful reflink. Over many files this causes max_blocks to diverge significantly from actual free space.

Propagate the reflink status from
create_regular_tmpfile_linkable_with_content back to write_content_object so it can credit back the reserved blocks, following the same pattern used for duplicate objects in 16e537cb ("lib/commit: Fix min-free-space accounting for duplicate content objects").

This should fix the CI issues in
https://github.com/coreos/fedora-coreos-config/pull/4265.

Heavily AI-guided investigation and resolution and red-green testing.

Assisted-by: AI

### DIFF
--- a/src/libostree/ostree-repo-commit.c
+++ b/src/libostree/ostree-repo-commit.c
@@ -617,8 +617,8 @@ _ostree_repo_bare_content_cleanup (OstreeRepoBareContent *regwrite)
 static gboolean
 create_regular_tmpfile_linkable_with_content (OstreeRepo *self, guint64 length,
                                               GInputStream *original_input, GInputStream *input,
-                                              GLnxTmpfile *out_tmpf, GCancellable *cancellable,
-                                              GError **error)
+                                              GLnxTmpfile *out_tmpf, gboolean *out_was_reflinked,
+                                              GCancellable *cancellable, GError **error)
 {
   g_auto (GLnxTmpfile) tmpf = {
     0,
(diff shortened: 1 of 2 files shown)


jlebon:
This was an interesting exercise in AI-guided investigation. FCOS could've easily worked around it I think once we understand what's going on. And ideally we somehow handle SELinux relabeling stuff _before_ handing it to `bootc install to-filesystem` so it doesn't have to do this two-pass import.

But also cool that the agent identified these genuine accounting bugs, with very minimal guidance from me.

The original prompt was:

> read the cosa docs, then create a new cosa workdir from the fcos local repo, then try to cosa build it and then cosa buildextend-qemu. you should hit the same issue as dusty in https://github.com/coreos/fedora-coreos-config/pull/4265. deeply investgate what the issue is and keep iterating until you get it working. understand the root cause, don't try to just add a workaround. take your time and be thorough.

I had to course correct perhaps twice where it did "fix" it, but not in the way I liked (once by trying to hack around it in chunkah, another time, by regularly re-running fstatvfs as a way to re-adjust max_blocks).

cgwalters (approved):

jlebon:
Filed https://github.com/coreos/chunkah/issues/159 related to this.

dustymabe:
> FCOS could've easily worked around it I think once we understand what's going on.

yes, but I'm glad we didn't and now everyone benefits.

AI is certainly help us here where digging into thorny problems (rather than working around them) previously required a significant investment, but now can be helped.

Nice work @jlebon!

PR #3618, opened 2026-07-31
static-delta: Account for the full part payload consistently

### LINKED ISSUE #3619: tests/test-delta.sh fails because object usize is used as the decompressed payload limit
I can reproduce a static-delta test failure from a clean build of current HEAD:

883cd757e81d2b2690e8f7ccc0d91cc31144d1e6

Running:

make check TESTS=tests/test-delta.sh

produces:

PASS: tests/test-delta.sh 1 generate
PASS: tests/test-delta.sh 2 show
PASS: tests/test-delta.sh 3 generate + show endian swapped
PASS: tests/test-delta.sh 4 heuristic endian detection
ERROR: tests/test-delta.sh - too few tests run (expected 14, got 4)

The test stops while pulling the initial empty-to-commit static delta:

error: Decompressed delta part exceeds configured limit of 2076299 bytes

The generator reported the same value as the part’s uncompressed size:

part 1 n:5 compressed:795133 uncompressed:2076299

The relevant usize field in OSTREE_STATIC_DELTA_META_ENTRY_FORMAT is documented as:

Uncompressed size of resulting objects on disk

The compiler populates it from part_builder->uncompressed_size. Commit a0ede4d33e0a0eb56052ef0a6ac5d8277da0057c then started passing that value to _ostree_static_delta_part_open() as the decompression limit.

Those sizes do not describe the same data. The decompressor produces a serialized static-delta payload with type:

(a(uuu)aa(ayay)ayay)

That payload includes mode and xattr tables, operation data, GVariant framing, and alignment. Its serialized size can therefore exceed the size of the resulting repository objects.

As a result, a static delta generated by OSTree itself is rejected by the corresponding OSTree consumer.

The existing usize appears to need to retain its resulting-object-size semantics because it is also used for progress and free-space accounting. It should not be used directly as the decompressed serialized-payload limit. A separate payload-size value would be needed if a limit more precise than the existing hard maximum is required.

cgwalters:
The previous commit adding a limit here didn't account correctly for metadata overhead; in trying to do it as a private security fix we didn't run all of the tests, which was a procedural mistake.

Assisted-by: AI

Closes: https://github.com/ostreedev/ostree/issues/3619

### DIFF
--- a/src/libostree/ostree-repo-finder-mount.c
+++ b/src/libostree/ostree-repo-finder-mount.c
@@ -339,8 +339,7 @@ ostree_repo_finder_mount_resolve_async (OstreeRepoFinder *finder,
 
 #if GLIB_CHECK_VERSION(2, 55, 0)
       G_GNUC_BEGIN_IGNORE_DEPRECATIONS /* remove once GLIB_VERSION_MAX_ALLOWED ≥ 2.56 */
-          g_autoptr (GUnixMountEntry) mount_entry
-          = g_unix_mount_at (mount_root_path, NULL);
+          g_autoptr (GUnixMountEntry) mount_entry = g_unix_mount_at (mount_root_path, NULL);
 
       if (mount_entry != NULL
           && (g_unix_is_system_fs_type (g_unix_mount_get_fs_type (mount_entry))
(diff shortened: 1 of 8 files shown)


jmarrero (commented):
lgtm

PR #3623, opened 2026-08-04
commit: Support mode stripping with '-' prefix in --statoverride files

### LINKED ISSUE #2368: Support `-<mode>` in file passed to `ostree commit --statoverride`
Right now, we can only add (`+` prefix or no prefix) or override (`=` prefix). Would make sense to have a `-` prefix too for stripping out mode bits. This would be useful in cosa at least: https://github.com/coreos/coreos-assembler/pull/2204.

lzwind:
## Summary

Add support for a `-` prefix in `--statoverride` files that strips the
specified permission bits from a file's mode (`mode &= ~specified`).
This complements the existing `+` (OR) and `=` (replace) prefixes and is
useful for removing specific permission bits without knowing the full
mode in advance.

## Changes

- Added `mode_removes` hash table to `CommitFilterData` and wired it
  through `handle_statoverride_line`, `commit_filter`, initialization,
  and the unmatched-path check.
- In `commit_filter`, a `-` entry applies `current_mode & ~mode_remove`.
- Updated `ostree-commit.xml` man page to document the `-` prefix.
- Added a test in `basic-test.sh` that strips all execute bits
  (0o111 = 73 decimal) from a 0755 file, expecting 0644.

Fixes #2368

### DIFF
--- a/man/ostree-commit.xml
+++ b/man/ostree-commit.xml
@@ -232,7 +232,9 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
                 <listitem><para>
                     File containing list of modifications to make permissions (file mode in
                     decimal, followed by space, followed by file path).  The specified mode
-                    is ORed with the file's original mode unless preceded by "=".
+                    is ORed with the file's original mode unless preceded by "=" in which
+                    case it replaces the mode, or "-" in which case the specified bits are
+                    stripped from the file's mode.
                 </para></listitem>
             </varlistentry>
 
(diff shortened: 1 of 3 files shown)


cgwalters:
Hi @lzwind why did you close this?

lzwind:
Hi @cgwalters , thanks for following up. I apologize for the confusion. After re-evaluating, I realized I might have misunderstood the root cause of the issue, and my previous fix may not have fully addressed it. I closed the PR to avoid introducing an incorrect solution. I'll take another look and update here if I find a proper fix.

PR #3633, opened 2026-08-06
ostree-gpg-verify-result: Strictly match subkeys if suffixed with `!`

pwithnall:
flatpak calls `ostree_gpg_verify_result_lookup()` when building an app commit to check whether it’s adding duplicate GPG signatures, for example to catch if the user accidentally specifies the same GPG key ID in `--gpg-sign` twice.

The code in `ostree_gpg_verify_result_lookup()` canonicalises its input and the fingerprint for each existing signature on a commit to the corresponding primary subkey fingerprint, and compares those.

This means that if `--gpg-sign` is passed twice to `flatpak build-export`, with the fingerprints of two *different* signing subkeys of the same primary key, the second will be rejected as a duplicate of the first, even if specified with `!` as a suffix. That suffix is accepted by `gpg` to force it to sign with a specific subkey rather than choosing the best subkey from the primary.

OSTree should follow `gpg`’s behaviour here, because it’s important for supporting subkey rotation. If we’re rotating subkeys, the primary key will have multiple valid signing subkeys at once, and we might want to legitimately sign a commit with all of them, in case some clients don’t have copies of all subkeys yet (e.g. they only have the old subkeys).

When doing a subkey rotation, it’s preferable to use an out-of-band mechanism to distribute the new subkey to all clients before they start trying to verify OSTree commits. Such a mechanism would be https://github.com/flatpak/flatpak/issues/6517 (for flatpak). Then we only have to sign commits with the new subkey. However, the approach allowed by this commit is also valid, and can be a useful fallback in case the out-of-band mechanism doesn’t work for some reason and we need a transition period for the rotation instead.

The existing behaviour of `ostree_gpg_verify_result_lookup()` is kept if the key ID does _not_ have a `!` suffix.

### DIFF
--- a/src/libostree/ostree-gpg-verify-result.c
+++ b/src/libostree/ostree-gpg-verify-result.c
@@ -278,10 +278,10 @@ ostree_gpg_verify_result_lookup (OstreeGpgVerifyResult *result, const gchar *key
       if ((match_only_subkey &&
            /* compare without the trailing `!`: */
            strlen (signature->fpr) == key_id_len - 1
-           && strncmp (key_id, signature->fpr, key_id_len - 1) == 0)
+           && g_ascii_strncasecmp (key_id, signature->fpr, key_id_len - 1) == 0)
           || (!match_only_subkey &&
               /* the first subkey in the list is the primary key: */
-              !g_strcmp0 (lookup_key->subkeys->fpr, signature_key->subkeys->fpr)))
+              g_ascii_strcasecmp (lookup_key->subkeys->fpr, signature_key->subkeys->fpr) == 0))
         {
 
           if (out_signature_index != NULL)
--- a/src/libostree/ostree-gpg-verify-result.c
+++ b/src/libostree/ostree-gpg-verify-result.c
@@ -242,7 +242,7 @@ ostree_gpg_verify_result_lookup (OstreeGpgVerifyResult *result, const gchar *key
   gboolean match_only_subkey;
 
   g_return_val_if_fail (OSTREE_IS_GPG_VERIFY_RESULT (result), FALSE);
-  g_return_val_if_fail (key_id != NULL, FALSE);
+  g_return_val_if_fail (key_id != NULL && ot_validate_gpg_key_id (key_id, NULL), FALSE);
 
   key_id_len = strlen (key_id);
   match_only_subkey = (key_id[key_id_len - 1] == '!');
--- a/src/libostree/ostree-gpg-verify-result.c
+++ b/src/libostree/ostree-gpg-verify-result.c
@@ -221,6 +221,11 @@ ostree_gpg_verify_result_count_valid (OstreeGpgVerifyResult *result)
  * Searches @result for a signature signed by @key_id.  If a match is found,
  * the function returns %TRUE and sets @out_signature_index so that further
  * signature details can be obtained through ostree_gpg_verify_result_get().
+ *
+ * If @key_id is a subkey fingerprint suffixed with `!`, the search will look
+ * for a signature from exactly that subkey. Otherwise, the search will look for
+ * a signature from any subkey in the primary key for @key_id.
+ *
  * If no match is found, the function returns %FALSE and leaves
  * @out_signature_index unchanged.
  *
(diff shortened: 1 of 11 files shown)


pwithnall:
This is a draft, currently looking for feedback about whether the high-level approach is acceptable. If it is, I’ll add unit tests and check the behaviour of `ostree gpg-sign` (which I think is the only non-testing use of `ostree_gpg_verify_result_lookup()` within ostree.git).

I’ve tested it briefly using flatpak, and it fixes the issue I was having with passing `--gpg-sign=$subkey1_fingerprint --gpg-sign=$subkey2_fingerprint` to `flatpak build-export`.

(I’m aware of the clang-format warning and will fix it when I next push.)

src/libostree/ostree-gpg-verify-result.c
```diff
@@ -233,35 +238,51 @@ ostree_gpg_verify_result_lookup (OstreeGpgVerifyResult *result, const gchar *key
...
+  size_t key_id_len;
+  gboolean subkey_match;
 
   g_return_val_if_fail (OSTREE_IS_GPG_VERIFY_RESULT (result), FALSE);
   g_return_val_if_fail (key_id != NULL, FALSE);
 
-  /* fetch requested key_id from keyring to canonicalise ID */
-  (void)gpgme_get_key (result->context, key_id, &lookup_key, 0);
+  key_id_len = strlen (key_id);
+  subkey_match = (key_id[key_id_len - 1] == '!');
```
cgwalters, line 248: Super minor but perhaps name this `expecting_subkey` or `match_only_subkey`?

src/libostree/ostree-gpg-verify-result.c
```diff
@@ -233,35 +238,51 @@ ostree_gpg_verify_result_lookup (OstreeGpgVerifyResult *result, const gchar *key
...
+          if (signature_key == NULL)
+            {
+              g_debug ("Could not find key when looking up signature from %s.", signature->fpr);
+              continue;
+            }
         }
 
-      /* the first subkey in the list is the primary key */
-      if (!g_strcmp0 (lookup_key->subkeys->fpr, signature_key->subkeys->fpr))
+      if ((subkey_match &&
```
cgwalters, line 278: It feels to me the code would be cleaner as 

```
if (expecting_subkey_match)
  {
    for (...) { ... }
  }
else 
  {
    for (...) { ... }
  }
```

cgwalters (commented):
Looks sane to me, apart from the LLM-found bug and some nits! Thanks for working on this!

src/libostree/ostree-gpg-verify-result.c
```diff
@@ -233,35 +238,51 @@ ostree_gpg_verify_result_lookup (OstreeGpgVerifyResult *result, const gchar *key
...
+           /* compare without the trailing `!`: */
+           strncmp (key_id, signature->fpr, key_id_len - 1) == 0) ||
```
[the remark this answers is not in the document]
pwithnall: Sigh, I explicitly mentioned that in my first comment.

src/libostree/ostree-gpg-verify-result.c
pwithnall: I tried that structure and wasn’t particularly happy with the repetition of the `for` conditions and exit code; I think the structure I went with is marginally clearer since it makes it clearer the difference is only within the loop iteration.

But if you insist I can rearrange it, just say :)

src/libostree/ostree-gpg-verify-result.c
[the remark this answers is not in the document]
pwithnall: The previous code didn’t do case-insensitive matching, and I think it’s fairly reasonable for the code in the PR to rely on the `key_id` being valid to work correctly (garbage in, garbage out). But sure, I can improve this.

src/libostree/ostree-gpg-verify-result.c
cgwalters: Nah, I'm fine with it as is, we have to repeat either the loop or the condition.

pwithnall: Great, I’ll leave it as-is. Currently addressing the other comments locally and adding some unit tests, then will push again. :)

pwithnall:
Ready for re-review

*Edit* And updated again to fix the build with `--with-gpgme=no` and `clang-format`

PR #3638, opened 2026-08-12
core: don't skip entries in _ostree_validate_structureof_xattrs

kali834x:
_ostree_validate_structureof_xattrs walks the xattr array with a for loop that already increments i, but the body increments it a second time, so only even-indexed entries are ever examined. the empty-name, duplicate-name and sort-order checks it exists to enforce quietly skip half the array, so a crafted dirmeta pulled from a remote can hide an unsorted or duplicate xattr in an odd slot and still pass ostree_validate_structureof_dirmeta on the write-metadata path. dropping the extra increment makes every entry get checked. the /dirmeta-xattrs test is extended with a duplicate name in the second slot, which is accepted before the change and rejected after.

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -2351,7 +2351,6 @@ _ostree_validate_structureof_xattrs (GVariant *xattrs, GError **error)
                                previous, name, i);
         }
       previous = name;
-      i++;
     }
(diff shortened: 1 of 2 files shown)


cgwalters (approved):
Thank you!

PR #3640, opened 2026-08-14
min-free-space: Use min(3%, 1GB) as default

### LINKED ISSUE #3616: Harmful min-free-space-percent these days
Percentage is useful for many things, the same as it's useless and harmful for many other things. In this particular case my attempt to run `flatpak-builder` resulted in:
```
error: Writing content object: min-free-space-percent '3%' would be exceeded, at least 4.1 kB requested

$ df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/nvme0n1p3  937G  865G   24G  98% /
```

Thus, the machine is full from 98%, with 24GB of the free space. The project I build does not need that much free space, it might need like 2GB or even less, I believe.

The percentage limit might make sense for small disks, but with 1TB disks and larger, which are getting common these days, the 3% limit does not make sense. Having 256GB disk the percentage would be a bit less than 10% free, for the same 24GB.

My suggestion: either drop the percentage limit completely, or use the percentage limit only if the actual value in bytes is smaller than the size limit.

lzwind:
The default minimum free space check uses 3%% of total filesystem space. On large disks (e.g., 1TB), this reserves 30GB which is excessive for typical operations.

Change the default to the lower of 3%% or 1GB, so the check is reasonable on both small and large filesystems.

Changes:
- Default fallback now sets both `min-free-space-percent = 3` and `min-free-space-mb = 1024`
- When both values are set, the minimum of the two is used (instead of size overriding percent)
- Updated man page documentation accordingly

Closes #3616

### DIFF
--- a/man/ostree.repo-config.xml
+++ b/man/ostree.repo-config.xml
@@ -135,13 +135,15 @@ License along with this library. If not, see <https://www.gnu.org/licenses/>.
           <para>
             Integer percentage value (0-99) that specifies a minimum percentage
             of total space (in blocks) in the underlying filesystem to keep
-            free. The default value is 3, which is enforced when neither this
-            option nor <varname>min-free-space-size</varname> are set.
+            free. The default value is 3, which is used when neither this
+            option nor <varname>min-free-space-size</varname> are set (together
+            with a default <varname>min-free-space-size</varname> of
+            <literal>1GB</literal>; the smaller of the two is enforced).
           </para>
           <para>
             If <varname>min-free-space-size</varname> is set to a non-zero
-            value, <varname>min-free-space-percent</varname> is ignored. Note
-            that, <varname>min-free-space-percent</varname> is not enforced on
+            value, the smaller of the two limits is used. Note that,
+            <varname>min-free-space-percent</varname> is not enforced on
             metadata objects. It is assumed that metadata objects are relatively
             small in size compared to content objects and thus kept outside the
             scope of this option.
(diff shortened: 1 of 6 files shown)


lzwind:
Thanks for the review! I've addressed all the comments and force-pushed a single updated commit.

Changes addressing the review:

1. **Stale defaults on reload** (`reload_core_config`): the cached `min_free_space_percent`/`min_free_space_mb` values are now reset before parsing the current config, so a reload no longer retains limits from a previous load (e.g. reloading `min-free-space-size=2GB` no longer leaves a stale 3% percent in effect, and setting the size to zero fully disables the check).

2. **Error reporting now reflects the enforced limit**: `min_free_space_calculate_reserved_bytes()` tracks which branch supplied the minimum (percent vs. size) and `throw_min_free_space_error()` reports that same option, so on filesystems below ~34GiB the default correctly reports `min-free-space-percent '3%'` (matching `itest-pull-space.sh`) instead of a misleading size message.

3. **Man page**: both `min-free-space-percent` and `min-free-space-size` entries now describe the same default (3% together with a 1GB cap, smaller of the two enforced), and the doubled percent sign (`3%%`) is fixed to `3%` (a percent sign is literal in DocBook).

4. **Comment typo**: fixed the doubled percent sign in the C comment (`3%%` → `3%`).

5. Added unit tests (`/repo/get_min_free_space_limits`) asserting the exact reserved bytes for size-only, percent-only, both-limits (min selection), the default, and the reload/stale-field regression.

lzwind:
Rebased onto the latest main (dfe9f7a1) and fixed the CI failures:
- Fixed a compile error in tests/test-repo.c (passed an extra `NULL` arg to `ostree_repo_write_config_and_reload`, which takes 3 args)
- Fixed a clang-format violation in tests/test-basic-c.c:605 that was introduced upstream by commit 7da79688 (moved the newly added `dup_dirmeta` GVariant construction to the GNU-style line break)
- Rebased to pick up new main

cgwalters (approved):
Thanks!

PR #3642, opened 2026-08-14
main: Don't pass subcommand verb to external CLI extensions

### LINKED ISSUE #3321: OSTree subcommand handler passes subcommand to itself
Like Git, the `ostree` command will look for a named command `ostree-$subcommand` in `PATH` when invoked with a subcommand it doesn't know. Unlike Git however, it passes its entire set of arguments, including the subcommand as `argv[1]`, to the child program.

What should happen, with `ostree-ext-cli` as an example:
```
$ ostree ext-cli
ostree-ext
Toplevel options for extended ostree functionality
...
```

What actually happens:
```
$ ostree ext-cli
error: Found argument 'ext-cli' which wasn't expected, or isn't valid in this context
...
```

wyu71:
When `ostree` dispatches to an external CLI extension binary (e.g. `ostree-ext-cli`), it replaces `argv[0]` with the extension path but leaves the subcommand verb (e.g. `ext-cli`) in `argv`, causing the external program to receive the verb as an unexpected argument:

```
$ ostree ext-cli
error: Found argument 'ext-cli' which wasn't expected, or isn't valid in this context
```

This strips the subcommand verb from `argv` before calling `execvp`, matching Git's behavior with external commands. The verb is identified as the first non-option argument after `argv[0]` (the same logic used by `ostree_command_lookup_external()`).

A test case in `tests/test-cli-extensions.sh` verifies the verb is not passed through.

Closes #3321

### DIFF
--- a/src/ostree/main.c
+++ b/src/ostree/main.c
@@ -87,6 +87,19 @@ main (int argc, char **argv)
   g_autofree gchar *ext_command = ostree_command_lookup_external (argc, argv, commands);
   if (ext_command != NULL)
     {
+      /* Strip the subcommand verb from argv so it is not passed to the
+       * external command as an unexpected argument (matching Git's
+       * behavior). The verb is the first non-option argument after argv[0]. */
+      for (guint i = 1; i < (guint)argc; i++)
+        {
+          if (argv[i] != NULL && !g_str_has_prefix (argv[i], "-")
+              && g_strcmp0 (argv[i], "") != 0)
+            {
+              for (guint j = i; j < (guint)argc; j++)
+                argv[j] = argv[j + 1];
+              break;
+            }
+        }
       argv[0] = ext_command;
       return ostree_command_exec_external (argv);
     }
--- a/tests/test-cli-extensions.sh
+++ b/tests/test-cli-extensions.sh
@@ -7,7 +7,7 @@ set -euo pipefail
 
 . $(dirname $0)/libtest.sh
 
-echo '1..2'
+echo '1..3'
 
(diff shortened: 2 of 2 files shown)


cgwalters:
Offhand sounds reasonable but doesn't this risk breaking existing extensions?

wyu71:
Good point — I thought about this, and I think the risk is low because the current behavior already breaks the canonical extension this feature is meant to support.

`ostree-ext-cli` (the ostree-rs-ext CLI, the primary external extension) parses its arguments with clap subcommands (`tar`, `container`, `ima-sign`, …) and has no `ext-cli` subcommand. With the current code, `ostree ext-cli container …` execs `ostree-ext-cli ext-cli container …`, and clap rejects the unexpected `ext-cli` — which is exactly the failure reported in #3321:

```
$ ostree ext-cli
error: Found argument 'ext-cli' which wasn't expected, or isn't valid in this context
```

After this change it execs `ostree-ext-cli container …`, which clap dispatches to the `container` subcommand as intended. (The ostree-rs-ext README's examples invoke `ostree-ext-cli` directly, so in practice the `ostree ext-cli` dispatch path has been unusable for clap-based extensions.)

This also matches Git's behavior, which #3321 explicitly cites as the reference: `git <verb>` dispatches to `git-<verb>` without forwarding the verb as an argument.

The residual risk would be an extension that deliberately skips `argv[1]` to work around the current behavior. That's an unusual pattern, and it would be relying on behavior that deviates from the documented Git-like contract. I didn't find such a case in the ecosystem, but happy to do a wider check if you'd like.

PR #3647, opened 2026-08-19
static-delta: validate bspatch payload offset and length

kali834x:
dispatch_bspatch reads the bspatch payload offset and length as varints from the delta opcode stream but never runs them through validate_ofs, unlike the other opcodes that index payload_data (open-splice-and-close, write, set-read-source). the only guard is the two asserts in bspatch_read, and opaque->offset + length overflows the second one for a 64-bit offset near UINT64_MAX, so a crafted offset passes the check and the memcpy reads before the mapped payload. add the same validate_ofs call the sibling opcodes use, right after the varints are read, which also returns a clean GError instead of aborting on an out-of-range window.

### DIFF
--- a/src/libostree/ostree-repo-static-delta-processing.c
+++ b/src/libostree/ostree-repo-static-delta-processing.c
@@ -403,6 +403,8 @@ dispatch_bspatch (OstreeRepo *repo, StaticDeltaExecutionState *state, GCancellab
     return FALSE;
   if (!read_varuint64 (state, &length, error))
     return FALSE;
+  if (!validate_ofs (state, offset, length, error))
+    return FALSE;
 
   if (state->stats_only)
     return TRUE; /* Early return */


cgwalters (approved):
Thank you!

cgwalters:
TL;DR I think this is just a DoS at most and can't be used to execute arbitrary code.

LLM generated content below <https://github.com/cgwalters/cgwalters#llms>

---

<details>

Why this is "just" a DoS and not a memory-disclosure bug: the OOB bytes
get mixed into the reconstructed file content, but before that content is
linked into the object store, `_ostree_repo_bare_content_commit()` recomputes
its SHA-256 and compares it to the checksum the delta itself declared. The
attacker can't predict unknown process memory ahead of time, so this check
essentially always fails, the temp file is discarded, and nothing leaked ever
gets persisted or returned to the attacker (the only residual is a SHA-256
*digest* of the leaked bytes in the error message — not reversible in
practice).

What *does* work reliably: the wild OOB pointer is likely to hit an unmapped
page and `SIGSEGV`, crashing whatever applies the delta (`ostree pull`,
`rpm-ostreed`, flatpak system helper) — a remote, unauthenticated,
no-interaction crash.

Same function had a related bug fixed as RHEL-189207 (CWE-680/122/190),
so this class of bug has precedent as CVE-worthy.

**Recommendation:** request a CVE, classified as Availability/DoS
(CWE-125, CWE-190), not confidentiality. CVSS ~7.5-8.2
(AV:N/AC:L/PR:N/UI:N/A:H/C:N).

</details>

PR #3649, opened 2026-08-25
fix: don't wedge deployment finalization on SELinux policy refresh

HuntedRaven7:
The /etc merge can carry stale or incomplete libsemanage state into new deployments (e.g. leftover tmp/ transaction directories from an interrupted image build, or orphaned policy modules), after which every semodule run fails and ostree-finalize-staged permanently wedges all updates until the module store is repaired manually.

Prune stale tmp/ transaction state under each policy store before running semodule, and treat refresh failures as non-fatal: log loudly to the journal and stderr, then finalize the deployment with the previously built binary policy carried over via /etc.

Closes [#3563](https://github.com/ostreedev/ostree/issues/3563#issuecomment-5403144016)


HuntedRaven7:
strange the clang stuff fails but the files I changed pass.

PR #3650, opened 2026-08-29
fetcher: Back off between HTTP retries

### LINKED ISSUE #3207: http retries should include backoff
Moving from https://pagure.io/releng/issue/11439#comment-898682

> So, one thing I notice poking through the ostree code a bit - is there any backoff logic on the retries? I don't think there is. Would it potentially help if we gave ostree some backoff logic, like curl itself has - initial backoff of one second, then double on each retry?

aashish-thapa:
Closes #3207.

Retries currently fire with no delay between them. This adds a capped exponential backoff with full jitter before each retry: the delay starts at 1s and doubles up to a 30s cap, and the actual wait is drawn randomly from [0, that] so that many clients hitting one mirror don't retry in lockstep. curl does the same doubling (it caps at 10 minutes); the jitter follows the usual AWS backoff guidance, and the cap is tighter here since pulls are often interactive.

The wait is cancellable, so cancelling a pull mid-backoff returns promptly rather than blocking for up to the delay.

The schedule lives in a small `ostree-fetcher-backoff.c` so the pure delay calculation can be unit tested (via the same compile-in pattern as the other internals tests). `test-pull-repeated.sh` sets `OSTREE_FETCHER_TEST_BACKOFF_MS=0` to skip the real waits.

The first commit adds a test for the schedule that fails against today's no-backoff behavior; the second implements it.


src/libostree/ostree-fetcher-backoff.c
```diff
@@ -0,0 +1,80 @@
...
+#include "config.h"
+
+#include <errno.h>
+
+#include "ostree-fetcher-backoff.h"
+
+/* Following curl, retries start at a 1 second delay and double each time. curl
+ * caps at 10 minutes; we cap tighter since pulls are frequently interactive. */
+#define OSTREE_FETCHER_RETRY_BASE_MS 1000
+#define OSTREE_FETCHER_RETRY_CAP_MS 30000
```
cgwalters, line 29: We should probably make this configurable via repo config?

src/libostree/ostree-fetcher-backoff.c
```diff
@@ -0,0 +1,80 @@
...
+      return;
+    }
+
+  gint64 deadline = g_get_monotonic_time () + (gint64)delay_ms * G_TIME_SPAN_MILLISECOND;
+  for (;;)
+    {
+      gint64 remaining_us = deadline - g_get_monotonic_time ();
+      if (remaining_us <= 0)
+        break;
+      gint n = g_poll (&pollfd, 1, (gint)(remaining_us / 1000));
```
cgwalters, line 73: Hmm is this really the best way to "wait for cancellable or a timeout"? Feels like perhaps a nested main loop would be cleaner?

(of course all of this makes me realize how much cleaner async rust is at all of this)

cgwalters (commented):
Thanks, this looks plausible.

PR #3653, opened 2026-08-31
docs: drop parameters that are no longer in the signatures

darkdi:
The `otcore_mount_rootfs` one is the interesting case: `root_transient` is now a field of `RootConfig`, which is declared a few lines above the block in the same header, so the parameter did not disappear, it moved. The other two describe signatures that predate the current ones, and the return text in `_ostree_stbuf_to_gfileinfo` already refers to `@stbuf`.

### DIFF
--- a/src/libostree/ostree-core.c
+++ b/src/libostree/ostree-core.c
@@ -1639,9 +1639,7 @@ _ostree_loose_path (char *buf, const char *checksum, OstreeObjectType objtype, O
 
 /**
  * _ostree_stbuf_to_gfileinfo:
- * @mode: File mode
- * @uid: File uid
- * @gid: File gid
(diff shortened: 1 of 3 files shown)


darkdi:
CI note: the two failures appear unrelated to this documentation-only patch. GitHub Actions Ubuntu Latest Release failed in tests/test-concurrency.py after opendir(staging-...): No such file or directory and reported too few tests run; the other 11 GitHub jobs passed. Jenkins failed later while rsync tried to change permissions on /, /usr/lib, and /usr/lib64 inside its container (exit 23). I do not have repository permission to rerun the failed job. Could a maintainer rerun the failed checks?

cgwalters (approved):

PR #3654, opened 2026-09-01
generator: Skip boot.mount if /etc/fstab has a /boot entry

babariviere:
## Problem

`boot_mount_generator()` links its unit with `GLNX_LINK_TMPFILE_NOREPLACE` and treats `EEXIST` as fatal. Because it runs *before* `fstab_generator()` in `_ostree_impl_system_generator()`:

```c
  if (!require_internal_units (normal_dir, early_dir, late_dir, error)) return FALSE;
  if (!sysroot_mount_generator (normal_dir, error))                     return FALSE;
  if (!boot_mount_generator (normal_dir, error))                        return FALSE;  /* aborts here */
  if (!fstab_generator (ostree_target, is_aboot, ...))                  return FALSE;  /* var.mount, never reached */
```

a `boot.mount` already written by `systemd-fstab-generator` aborts the whole generator, `var.mount` is never written, and `/var` is left unmounted. The system boots severely degraded: `local-fs.target` breaks and everything writing under `/var/lib` fails.

## Why this is reachable on a supported configuration

On systems where `/boot` lives on the root filesystem, an `/etc/fstab` entry for `/boot` is the *documented* way to opt out of Fedora CoreOS' `coreos-boot-mount-generator`:

```bash
# If there's already an /etc/fstab entries for /boot, then this is is a non-FCOS
# system, likely RHCOS pre-4.3 (which still used Anaconda).  In that case, we
# don't want to overwrite what the systemd-fstab-generator will do.
if findmnt --fstab /boot &>/dev/null; then
    exit 0
fi
```

That opt-out is required because `bootc install to-disk` does **not** create a partition labelled `boot` (its layout is BIOS-BOOT / EFI-SYSTEM / root, with `/boot` on the root fs). Without the fstab entry, the CoreOS generator emits a `boot.mount` that blocks forever on `/dev/disk/by-label/boot`:

```
[ *** ] Job dev-disk-by\x2dlabel-boot.device/start running (47s / 1min 30s)
```

But *with* the fstab entry, `systemd-fstab-generator` writes `boot.mount` first and this generator aborts. So on a CoreOS-derived image installed via `bootc install to-disk`, both paths fail, for different reasons.

## Reproducer

No special hardware needed; any ostree system with `/boot` on the root filesystem:

```console
$ mkdir -p /tmp/g /tmp/ge /tmp/gl
$ touch /tmp/g/boot.mount    # stand in for systemd-fstab-generator's unit
$ /usr/lib/systemd/system-generators/ostree-system-generator /tmp/g /tmp/ge /tmp/gl
ostree-system-generator: linkat: File exists
$ echo $?
1
$ ls /tmp/g
boot.mount  local-fs.target.requires  multi-user.target.wants  sysroot.mount.d
```

Note the absent `var.mount`. With an empty target directory the same invocation exits 0 and produces both units.

## Fix

Per review feedback: instead of probing the output directory for an existing `boot.mount` (which depends on generator ordering and is nondeterministic), check `/etc/fstab` directly. If it carries an entry for `/boot`, `systemd-fstab-generator` owns that unit, so we skip generating ours and continue to `var.mount`. The `GLNX_LINK_TMPFILE_NOREPLACE` behaviour is otherwise unchanged, so a genuinely unexpected conflict is still an error.

The fstab parsing already used to look for `/var` is factored out into `fstab_has_mount_entry()` and reused, so path normalization (`//boot/` -> `/boot`) and the missing-fstab handling are shared.

## Regression range

This looks like a regression from the soft-reboot work in #3487 / #3571. ostree 2025.7 did not generate `boot.mount` in the generator; the failure first appears in 2026.1 and is still present in `main`. Observed in the wild on ostree 2026.2 (Fedora 44 / uCore).

## Testing

Added `tests-unit-container/test-system-generator.sh`, picked up automatically by
`tests-unit-container/run.sh` (`just unitcontainer`). It fakes the generator's
environment (`/run/ostree`, a bind-mounted `/proc/cmdline` with a valid `ostree=`
bootlink, an `/etc/fstab`, and the `/sysroot/boot/loader` symlink) and covers:

1. No `/boot` entry in fstab: `boot.mount` (with `What=/sysroot/boot`), its
   `local-fs.target.requires` symlink, and `var.mount` are all generated.
2. `/etc/fstab` has a `/boot` entry: no `boot.mount` is generated, and
   `var.mount` still is. This is the regression. Also covered with a
   non-normalized `//boot/` path.
3. No `/etc/fstab` at all: both units are generated (ENOENT is not an error).
4. No `/sysroot/boot/loader` (separate `/boot` partition): no `boot.mount`,
   `var.mount` still generated.

The reproducer above was also verified on ostree 2026.2, before and after.

Note the same `GLNX_LINK_TMPFILE_NOREPLACE` pattern is used for `var.mount` and the soft-reboot drop-in. I left those alone since they're much less likely to collide, but say the word if you'd prefer consistency.

## Disclosure

```
Assisted-by: AI
```

Per `AGENTS.md` conventions: this was investigated and drafted with AI assistance after hitting the bug on a production host.

### DIFF
--- a/src/libostree/ostree-impl-system-generator.c
+++ b/src/libostree/ostree-impl-system-generator.c
@@ -89,6 +89,49 @@ path_kill_slashes (char *path)
 
 #endif
 
+/* Look for a mount entry for @where (e.g. "/var") in /etc/fstab. */
+static gboolean
+fstab_has_mount_entry (const char *where, gboolean *out_found, GError **error)
+{
+#ifdef HAVE_LIBMOUNT
+  static const char fstab_path[] = "/etc/fstab";
+
+  *out_found = FALSE;
+
+  g_autoptr (OtLibMountFile) fstab = setmntent (fstab_path, "re");
+  if (!fstab)
+    {
+      if (errno == ENOENT)
+        return TRUE;
+      return glnx_throw_errno_prefix (error, "Reading %s", fstab_path);
+    }
+
+  struct mntent *me;
+  while ((me = getmntent (fstab)))
+    {
+      g_autofree char *mnt_dir = g_strdup (me->mnt_dir);
+      if (is_path (mnt_dir))
+        path_kill_slashes (mnt_dir);
+
+      if (strcmp (mnt_dir, where) == 0)
+        {
+          *out_found = TRUE;
+          break;
+        }
+    }
+
+  return TRUE;
+#else
+  /* Without libmount we can't parse /etc/fstab; just report "no entry". The
+   * generator as a whole is not functional in this configuration anyway, since
+   * fstab_generator() below throws "Not implemented". */
+  (void)where;
+  (void)error;
+  *out_found = FALSE;
+  return TRUE;
+#endif
+}
+
 /* Forcibly enable our internal units, since we detected ostree= on the kernel cmdline */
 static gboolean
 require_internal_units (const char *normal_dir, const char *early_dir, const char *late_dir,
@@ -269,6 +312,22 @@ boot_mount_generator (const char *normal_dir, GError **error)
   if (!(lstat ("/boot", &stbuf) == 0 && S_ISDIR (stbuf.st_mode)))
     return TRUE;
 
+  /* If /etc/fstab has an entry for /boot, then systemd-fstab-generator owns
+   * boot.mount and we must not generate our own.
+   *
+   * An fstab entry for /boot is the documented way to opt out of Fedora CoreOS'
+   * coreos-boot-mount-generator on systems that have no separate partition
+   * labelled "boot" (for example the layout produced by "bootc install
+   * to-disk"). Without this check we unconditionally created our own unit and
+   * aborted the whole generator with EEXIST, which meant fstab_generator()
+   * below never ran and /var was never mounted at all.
+   */
+  gboolean fstab_has_boot = FALSE;
+  if (!fstab_has_mount_entry (boot_path, &fstab_has_boot, error))
+    return FALSE;
+  if (fstab_has_boot)
+    return TRUE;
+
   glnx_autofd int normal_dir_dfd = -1;
   if (!glnx_opendirat (AT_FDCWD, normal_dir, TRUE, &normal_dir_dfd, error))
     return FALSE;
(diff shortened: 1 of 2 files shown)


cgwalters:
xref https://github.com/bootc-dev/bootc/issues/2402 - I think while it helps to avoid this clash, we actually need to fix this so that there's only one thing generating `/boot` in general.

src/libostree/ostree-impl-system-generator.c
```diff
@@ -307,15 +324,18 @@ boot_mount_generator (const char *normal_dir, GError **error)
   g_clear_object (&outstream);
   if (!glnx_fchmod (tmpf.fd, 0644, error))
     return FALSE;
-  if (!glnx_link_tmpfile_at (&tmpf, GLNX_LINK_TMPFILE_NOREPLACE, normal_dir_dfd, "boot.mount",
-                             error))
+  /* Generators run in parallel, so another one may have created boot.mount
+   * between the check above and here; tolerate that rather than failing. */
```
cgwalters: But that's just going to make the whole thing nondeterministic right?

cgwalters:
Can you clarify: do you have `/boot` in your `/etc/fstab` or not?

babariviere:
@cgwalters yes I have to put a `/boot` entry in my `/etc/fstab` otherwise, `coreos-boot-mount-generator` block boot by trying to find `/boot` by disk label (`/dev/disk/by-label/boot`).
So I have this job that block when starting:

```
dev-disk-by\x2dlabel-boot.device/start
```

And I fallback to systemd emergency mode.

My /fstab looks like this:

```
/sysroot/boot /boot none bind,x-systemd.requires-mounts-for=/sysroot 0 0
```

src/libostree/ostree-impl-system-generator.c
babariviere: Indeed yes, maybe a solution would be to check if /etc/fstab have a /boot entry, and skip generating if that's the case, wdyt?

cgwalters: Yes, we should definitely do that.

babariviere: I have made the change

cgwalters (approved):
Thanks!

PR #3656, opened 2026-09-13
ci: Install built bits into the cosa pod with cp instead of rsync

jmarrero:
The coreos-ci Jenkins job — the only required status check on `main` — has failed on every PR since March with "This commit cannot be built". Two fixes:

1. Since c7265cce the cosa pod runs unprivileged, relying on coreos-assembler making `/usr` group-writable (coreos-assembler#4410, since narrowed to directories only in [533ac5be8](https://github.com/coreos/coreos-assembler/commit/533ac5be8)). `rsync -rlv installed/rootfs/ /` copies the files fine but also tries to chmod the root-owned directories it recurses into, fails with EPERM, and exits 23 ([example](https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/job/ostree/job/PR-3611/8/console)). Use `cp -dR --remove-destination` instead, which leaves existing directories alone. Reproduced and verified in a fedora:44 container as uid 1000/gid 0.

2. The build moved to `target/c`, so the unit test stage has been catting and archiving a nonexistent `test-suite.log` and we get no output when a test fails.

With both, Jenkins passes end to end (build, unit tests, cosa build, kola) — see the run on this PR.

Generated-by: AI
I am knowledgeable in this problem domain and reviewed it carefully.


jmarrero:
Jenkins build #1 on this PR failed before reaching the `Build FCOS` stage, in **Unit tests**: `tests/test-commit-sign.sh - too few tests run (expected 7, got 2)`. The same stage passed an hour earlier on #3611 with an otherwise identical tree, so this looks like a flake in the gpgme C helper (`test-commit-sign-sh-ext`).

We can't see why, because the Jenkinsfile still cats/archives `./test-suite.log` while the build now lives in `target/c/` (Jenkins even hints: *'test-suite.log' doesn't match anything, but 'target/c/test-suite.log' does*). Pushed a second commit fixing that path so the next failure is actually diagnosable.

jeckersb (approved):

